PR #1500, opened 2021-10-04
Queue stress test

### LINKED ISSUE #1476: Correct queue push errors
Right now push might fail for 3 reasons:
- queue is full
- tx expired
- tx in blockchain

But it will always report that queue is full.

### LINKED ISSUE #1483: Concurrent stress test for queue
In actual deployment txs can be added and popped concurrently to and from the queue. So this should be checked in a separate test in queue.rs module and the test should simulate high load conditions.

eadventurous:

### Description of the Change

- Queue stress test
- Some other minor tests added for queue cases
- Queue test fixes
- Fixes in the queue behavior due to improper rebase
- Unnecessary (due to removed functonaility) test with getting transactions on leader - removed

### Issue
Closes #1483 
Closes #1476

### DIFF
--- a/iroha/src/queue.rs
+++ b/iroha/src/queue.rs
@@ -4,10 +4,27 @@ use std::time::Duration;
 
 use crossbeam_queue::ArrayQueue;
 use dashmap::{mapref::entry::Entry, DashMap};
-use eyre::Result;
+use eyre::{Report, Result};
 
 use self::config::QueueConfiguration;
 use crate::{prelude::*, wsv::WorldTrait};
+
+/// Queue error.
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// Transaction expired.
+    #[error("Transaction expired.")]
+    TransactionExpired,
+    /// Transaction is already in blockchain.
+    #[error("Transaction is already in blockchain.")]
+    AlreadyInBlockchain,
+    /// Failure during signature condition execution.
+    #[error("Failure during signature condition execution: `{0}`")]
+    SignatureConditionExecution(#[source] Report),
+    /// The transaction queue is full.
+    #[error("The transaction queue is full.")]
+    QueueIsFull,
+}
 
 /// Lockfree queue for transactions
 ///
@@ -42,9 +59,26 @@ impl Queue {
         self.txs.iter().map(|t| t.value().clone()).collect()
     }
 
-    /// Pushes transaction into queue
+    fn check_tx<W: WorldTrait>(
+        &self,
+        tx: &VersionedAcceptedTransaction,
+        wsv: &WorldStateView<W>,
+    ) -> Result<(), Error> {
+        if tx.is_expired(self.ttl) {
+            Err(Error::TransactionExpired)
+        } else if tx.is_in_blockchain(wsv) {
+            Err(Error::AlreadyInBlockchain)
+        } else if let Err(err) = tx.check_signature_condition(wsv) {
+            Err(Error::SignatureConditionExecution(err))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Pushes transaction into queue.
+    ///
     /// # Errors
-    /// Returns transaction if queue is full
+    /// See [`Error`]
     #[allow(
         clippy::unwrap_in_result,
         clippy::expect_used,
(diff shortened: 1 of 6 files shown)


iroha/src/queue.rs
```diff
@@ -54,19 +92,15 @@ impl Queue {
...
-            iroha_logger::warn!("Transaction expired");
-            return Err(tx);
-        }
-        if tx.is_in_blockchain(wsv) {
-            iroha_logger::warn!("Transaction is already applied");
-            return Err(tx);
+    ) -> Result<(), Error> {
+        // Should always check transactions to prevent eternal gossiping of wrong txs.
+        if let Err(error) = self.check_tx(&tx, wsv) {
+            iroha_logger::warn!("{}", error);
```
i1i1, line 98: It is a bad idea to do both logging here and return error. It is better to add choose whether to report on caller side. For cases like transaction forwarding it makes sense to not report `InBlockchain` error. You should check out pr of mine. Some of the logs were improved there.

iroha/src/queue.rs
```diff
@@ -29,7 +46,11 @@ impl Queue {
     /// Makes queue from configuration
     pub fn from_configuration(cfg: &QueueConfiguration) -> Self {
         Self {
-            queue: ArrayQueue::new(cfg.maximum_transactions_in_queue as usize),
+            // See `get_transactions_for_block` implementation comment.
+            queue: ArrayQueue::new(
+                cfg.maximum_transactions_in_queue as usize
+                    + cfg.maximum_transactions_in_block as usize,
```
i1i1, line 52: Is it a bug though? Length of whole queue is measured by number of transactions in dashmap. So actually it should be always ok to push there. Did you encountered some error with that?

iroha/src/sumeragi/mod.rs
```diff
@@ -1378,13 +1378,15 @@ pub mod message {
                 .send_to(&sumeragi.broker, &self.peer)
                 .await;
             }
+            let hash = self.transaction.hash();
             sumeragi
                 .queue
                 .push(self.transaction, &*sumeragi.wsv)
-                .map_err(|tx| {
+                .map_err(|error| {
```
i1i1, line 1385:
```suggestion
                .map_err(|error| {
```

iroha/src/sumeragi/mod.rs
```diff
@@ -1378,13 +1378,15 @@ pub mod message {
...
                     eyre!(
-                        "Failed to push tx into queue. Queue is full. Hash = {:?}",
-                        tx.hash()
+                        "Failed to push tx into queue. Hash = {:?} Error: {}.",
+                        error,
+                        hash
```
i1i1, line 1389:
```suggestion
                .wrap_err_with(|| format!("Failed to push tx with hash {} into queue", hash))
```

iroha/src/queue.rs
eadventurous: Sure, I'll check. I had this in mind just did not plan to change a lot of code in this PR. But I guess I will add this logic.

iroha/src/queue.rs
eadventurous: There is a specific edge case - in the `get_transactions_for_block` we pop several transactions and sometimes push them back, assuming that queue has size to fit them. Concurrently with calling this function there might be queue pushes which will fill up the queue and our assumption will not be true for that case.

That's why public `push` will stop at `maximum_transactions_in_queue`. But queue will still have `maximum_transactions_in_block` places for the private push in `get_transactions_for_block`.

i1i1: This will never happen, as transactions are still will be in queue

eadventurous: Yes, I see now, it won't happen as the push checks by number of transactions in the map which does not decrease there.

iroha/src/queue.rs
```diff
@@ -4,11 +4,28 @@ use std::time::Duration;
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// Transaction expired.
+    #[error("Transaction expired.")]
+    TransactionExpired,
+    /// Transaction is already in blockchain.
+    #[error("Transaction is already in blockchain.")]
+    AlreadyInBlockchain,
+    /// Failure during signature condition execution.
+    #[error("Failure during signature condition execution: `{0}`")]
```
i1i1, line 22: Should we print internal error here?

eadventurous: To my mind yes, it will tell what is wrong with accounts signature condition configuration.

eadventurous: It will help our client developers and other developers building on Iroha a lot

eadventurous: We discussed in telegram: it is unnecessary as it will print it due to having this error as source.

i1i1 (approved):

iroha/src/queue.rs
eadventurous: After some more research we found a problem with this for our simple torii into response http converter for errors, as it doesn't follow the source. Planning to post an issue for this.

KalitaAlexey (changes requested):

iroha/src/queue.rs
```diff
@@ -488,7 +504,69 @@ mod tests {
...
+        let queue = Arc::new(Queue::from_configuration(&QueueConfiguration {
+            maximum_transactions_in_block: max_block_tx,
+            transaction_time_to_live_ms: 100_000,
+            maximum_transactions_in_queue: 100_000_000,
+        }));
+
+        let start_time = Instant::now();
+        let run_for = Duration::from_secs(5);
+
+        let queue_arc_clone_1 = Arc::clone(&queue);
```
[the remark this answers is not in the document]
eadventurous: To explicitly show that it is just a new reference and not a clone of the contents.

See https://rust-lang.github.io/rust-clippy/master/#clone_on_ref_ptr I think we have it enabled

KalitaAlexey (approved):

i1i1 (approved):

PR #1576, opened 2021-11-03
[Feat] Add the number of connected peers to `Status`

### LINKED ISSUE #1387: Endpoint for smoke testing network status
Sofiane, [29.08.21 18:05]
Feature Request:

We need and endpoint or a transaction that can check the state of the whole network. It has to be generic and independent from the genesis. Something like the number of connected peers in the network would be good for example.

Priority: Medium

s8sato:
### Description of the Change
* Number of connected peers, except for the reporting peer itself

Screenshot from 2021-11-16 01-54-03

### Issue
Closes #1387

### Possible Drawbacks


core/src/torii/mod.rs
```diff
@@ -364,6 +395,24 @@ async fn handle_subscription(events: EventsSender, stream: WebSocket) -> eyre::R
...
+/// Response body for get metrics request
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub struct Metrics {
+    /// Number of connected peers
+    peers: u64,
+    /// Number of committed blocks
+    blocks: u64,
+}
+
+#[allow(clippy::unused_async)]
```
appetrosyan, line 407: Since this function has no `await` can it be made a regular function, rather than `async`?

core/src/torii/mod.rs
```diff
@@ -892,4 +947,62 @@ mod tests {
             .assert()
             .await
     }
+    #[tokio::test]
```
eadventurous, line 950: Let's add a new line here just for consistency.

core/src/torii/mod.rs
eadventurous: Probably warp needs this signature, but, yes, we should check if it is possible to have it in other way.

core/src/torii/mod.rs
```diff
@@ -364,6 +395,24 @@ async fn handle_subscription(events: EventsSender, stream: WebSocket) -> eyre::R
...
+    /// Number of connected peers
+    peers: u64,
+    /// Number of committed blocks
+    blocks: u64,
+}
+
+#[allow(clippy::unused_async)]
+async fn handle_metrics<W: WorldTrait>(state: ToriiState<W>) -> Result<Json> {
+    let metrics: Metrics = Metrics {
+        peers: state.wsv.peers().len() as u64,
```
eadventurous, line 410: We need to get connected peers from the p2p crate somehow. In WSV we have the general number of peers that this peer knows about, but they are not necessarily connected at the moment. If they are connected or not knows only our p2p lib.

eadventurous (changes requested):

eadventurous:
Also as suggested by Sofaine let's rename the endpoint to `status`.

core/src/torii/mod.rs
appetrosyan: `iroha_p2p::network::NetworkBase::peers` has what you need.

client/tests/tests/status.rs
```diff
@@ -0,0 +1,91 @@
...
+    // Add a peer then #peers should be incremented
+    let (mut peer, _) = rt.block_on(network.add_peer());
+    thread::sleep(std::time::Duration::from_millis(5_000));
+    // FIXME 'assertion failed: `(left == right)` left: `3`, right: `4`'
+    assert_eq!(status(&client).peers, N_PEERS);
+
+    // Drop the peer then #peers should be decremented
+    peer.stop();
+    assert_eq!(status(&client).peers, N_PEERS - 1);
+}
```
s8sato, line 65: Both versions fail in the same assertion in the same way

https://github.com/hyperledger/iroha/blob/77c8461db84c784d97d790d4f8c94cd2913395d4/client/tests/tests/status.rs#L46-L64

client/src/client.rs
```diff
@@ -462,6 +470,30 @@ impl Client {
         self.get_config(&GetConfiguration::Value)
             .wrap_err("Failed to get configuration value")
     }
+
+    /// Gets network status seen from the peer
+    /// # Errors
+    /// Fails if sending request or decoding fails
+    pub fn get_status(&self) -> Result<serde_json::Value> {
```
eadventurous, line 477: Let's de-serialize status into `Status` here and not in the tests. Client always knows the API it is working with so it is better to de-serialize into structs instead of returning raw json values.

eadventurous: For this it might be better to move `Status` struct from torii into `data_model` crate.

client/tests/tests/status.rs
```diff
@@ -0,0 +1,91 @@
...
+    assert_eq!(status(&client).peers, N_PEERS - 1);
+
+    // Add a peer then #peers should be incremented
+    let (mut peer, _) = rt.block_on(network.add_peer());
+    thread::sleep(std::time::Duration::from_millis(5_000));
+    // FIXME 'assertion failed: `(left == right)` left: `3`, right: `4`'
+    assert_eq!(status(&client).peers, N_PEERS);
+
+    // Drop the peer then #peers should be decremented
+    peer.stop();
```
eadventurous, line 63: After this I think there should be a `thread::sleep`.

client/src/client.rs
```diff
@@ -462,6 +470,30 @@ impl Client {
...
+                std::str::from_utf8(resp.body()).unwrap_or(""),
+            ));
+        }
+        serde_json::from_slice(resp.body()).wrap_err("Failed to decode body")
+    }
+}
+
+fn url<L: Display, R: Display>(lhs: L, rhs: R) -> String {
+    format!("{}/{}", lhs, rhs)
 }
```
eadventurous, line 497: I think it is unnecessary the `format` there is quite brief and also easy to understand

client/tests/tests/status.rs
s8sato: Sure, was going to insert proper interval there after the 2nd assertion passed

core/src/torii/mod.rs
```diff
@@ -219,6 +228,12 @@ impl<W: WorldTrait> Torii<W> {
             .with(warp::trace::request())
             .recover(Torii::<W>::recover_arg_parse);
 
+        tokio::spawn(async move {
+            start_status(Arc::clone(&state))
+                .await
+                .wrap_err("Failed to start status service")
+        });
```
i1i1, line 235: There should be something done with an error. Smth like this can be done:
```suggestion
        tokio::spawn(async move {
            let res = start_status(Arc::clone(&state))
                .await
                .wrap_err("Failed to start status service");
            if let Err(err) = res {
                iroha_logger::error!(%error);
            }
        });
```

core/src/torii/mod.rs
```diff
@@ -391,13 +455,15 @@ pub mod config {
...
     pub const DEFAULT_TORII_API_URL: &str = "127.0.0.1:8080";
-    /// Default maximum size of single transaction.
+    /// Default socket for reporting internal metrics
+    pub const DEFAULT_TORII_STATUS_URL: &str = "127.0.0.1:8180";
+    /// Default maximum size of single transaction
     pub const DEFAULT_TORII_MAX_TRANSACTION_SIZE: usize = 2_usize.pow(15);
     /// Default maximum instruction number
     pub const DEFAULT_TORII_MAX_INSTRUCTION_NUMBER: u64 = 2_u64.pow(12);
-    /// Default maxiumum size of [`Sumeragi`] message size.
+    /// Default maximum size of [`Sumeragi`] message size
```
appetrosyan, line 466:
```suggestion
    /// Default maximum [`Sumeragi`] message size
```

data_model/src/lib.rs
```diff
@@ -387,6 +387,15 @@ pub fn current_time() -> Duration {
         .expect("Failed to get the current system time")
 }
 
+/// Response body for get status request
```
appetrosyan, line 390:
```suggestion
/// Response body for GET status request
```

docs/source/references/api_spec.md
```diff
@@ -151,6 +152,29 @@ Also returns current status of peer in json string:
...
+
+**Protocol**: HTTP
+
+**Encoding**: Json
+
+**Endpoint**: `/status`
+
+**Method**: `GET`
+
+**Expects**: -
```
appetrosyan, line 165: Maybe remove it altogether?

p2p/src/network.rs
```diff
@@ -190,28 +195,52 @@ where
...
+    T: Debug + Encode + Decode + BrokerMessage + Send + Sync + Clone + 'static,
+    K: KeyExchangeScheme + Send + 'static,
+    E: Encryptor + Send + 'static,
+{
+    type Result = ();
+
+    async fn handle(&mut self, DisconnectPeer(public_key): DisconnectPeer) {
+        let peer = match self.peers.remove(&public_key) {
+            Some(peer) => peer,
+            _ => return error!(?public_key, "Not found peer to disconnect"),
```
appetrosyan, line 239:
```suggestion
            _ => return warn!(?public_key, "Not found peer to disconnect"),
```

p2p/src/network.rs
```diff
@@ -248,40 +279,39 @@ where
     async fn handle(&mut self, msg: PeerMessage<T>) {
         use PeerMessage::*;
 
+        debug!(?msg);
```
appetrosyan, line 282:
```suggestion
        
```
Instead, just instrument the function with `#[log]`, which should produce trace level logs, (which this should be).

eadventurous:
What do you think about splitting this PR into 2 separate PRs:
1. Will introduce a status endpoint only with number of committed blocks.
2. Will add number of connected peers feature and corrections to p2p that you want to introduce.

I think we will be able to merge the 1st own with minor corrections almost instantly. Then you will have time to work on the second one in detail.

appetrosyan (dismissed):

appetrosyan (approved):

eadventurous (approved):

PR #1599, opened 2021-11-10
[Fix] 1502 Return first sources in error responses in Torii

### LINKED ISSUE #1502: Return full errors in response bodies in torii
Currently probably due to migration to thiserror and eyre, our torii error return bodies stopped giving back full error chain.

Only specifying `#[source]` there does not seem to be enough as is demonstrated by [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=346bce4c5ddfe2278b227741bae5d85b). It seems we need to add where appropriate `{0}` or other formatting specifiers to error descriptions, to have full chain in bodies.

appetrosyan:
### Issue

Closes #1502 

### Benefits

Torii produces more detailed error messages. 

### Possible Drawbacks

Error sources are added to all errors.

### DIFF
--- a/core/src/torii/mod.rs
+++ b/core/src/torii/mod.rs
@@ -17,7 +17,6 @@ use warp::{
     Filter, Reply,
 };
 
-#[macro_use]
 mod utils;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


core/src/torii/mod.rs
```diff
@@ -45,10 +44,10 @@ pub struct Torii<W: WorldTrait> {
 #[derive(Error, Debug)]
 pub enum Error {
     /// Failed to execute or validate query
-    #[error("Failed to execute or validate query")]
+    #[error("Failed to execute or validate query. {0}")]
```
i1i1, line 47: We shouldnt duplicate error output if we already have its source. It is printed here:
https://github.com/hyperledger/iroha/blob/iroha2-dev/core/src/torii/mod.rs#L98-L113

eadventurous: Then it seems that the PR is not needed?

appetrosyan: I'm closing this PR, and #1502 until there's an MWE of a missing error body.

PR #1603, opened 2021-11-11
[feature] #1238: Run-time log-level updates

### LINKED ISSUE #1238: Log-level (verbosity) should be changeable in runtime
Log-level (verbosity) should be changeable in runtime, example: system signal (e.g., SIGUSR1, SIGHUP) to reread configs(env variable) and change the log-level.
https://wiki.hyperledger.org/display/iroha/Logging

### LINKED ISSUE #1601: Configuration endpoint docs and fix.
The Configuration endpoint needs to have better documentation: something that can be copied and pasted, e.g. 
```curl -X GET -H 'content-type: application/json' http://127.0.0.1:8080/configuration -d '{"Value"}'```
Additionally it has problems producing documentation: 
image
it shouldn't expect a nested sequence, and should produce slightly more descriptive errors.

### LINKED ISSUE #1612: Runtime Upgrades: Sync actual behaviour to configuration state
After merging #1603 , the log level in the global configuration is no longer guaranteed to reflect the actual maximum log level.

### PR DESCRIPTION
**Pull request by appetrosyan**
### Description of the Change

Implemented a `POST` handler on the `/configuration` Torii endpoint, which changes the log-level at runtime. 

### Issue

Closes #1238
Closes #1601
Closes #1612

### Benefits

Log level is changeable at runtime. 

### Possible Drawbacks

None

### DIFF
--- a/.github/workflows/iroha2-dev-pr-static.yml
+++ b/.github/workflows/iroha2-dev-pr-static.yml
@@ -29,9 +29,12 @@ jobs:
         run: cargo +nightly-2021-10-22 fmt --all -- --check
       - name: Static analysis without features
         run: cargo lints clippy --workspace --benches --tests
+        if: always()
       - name: Static analysis with all features enabled
         run: cargo lints clippy --workspace --benches --tests --all-features
+        if: always()
       - name: Documentation check
         run: |
           cargo doc --no-deps
           ./scripts/check_docs.sh
+        if: always()
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1,3 +1,6 @@
+//! Contains the end-point querying logic building on top of
+//! [`crate::http_client`].  This is where you need to add any custom
+//! end-point related logic.
 use std::{
     collections::HashMap,
     convert::TryFrom,
@@ -11,7 +14,7 @@ use eyre::{eyre, Result, WrapErr};
 use http_client::WebSocketStream;
 use iroha_core::{
     smartcontracts::Query,
-    torii::{uri, GetConfiguration},
+    torii::{uri, GetConfiguration, PostConfiguration},
     wsv::World,
 };
 use iroha_crypto::{HashOf, KeyPair};
@@ -45,7 +48,8 @@ pub struct Client {
     pub account_id: AccountId,
     /// Http headers which will be appended to each request
     pub headers: http_client::Headers,
-    /// If `true` add nonce, which make different hashes for transactions which occur repeatedly and simultaneously
+    /// If `true` add nonce, which makes different hashes for
+    /// transactions which occur repeatedly and/or simultaneously
     pub add_transaction_nonce: bool,
 }
 
@@ -434,8 +438,9 @@ impl Client {
     }
 
     fn get_config<T: DeserializeOwned>(&self, get_config: &GetConfiguration) -> Result<T> {
-        let mut headers = self.headers.clone();
-        headers.insert("Content-Type".to_owned(), "application/json".to_owned());
+        let headers = [("Content-Type".to_owned(), "application/json".to_owned())]
+            .into_iter()
+            .collect();
         let get_cfg = serde_json::to_vec(get_config).wrap_err("Failed to serialize")?;
 
         let resp = http_client::get::<_, Vec<(&str, &str)>, _, _>(
@@ -454,16 +459,44 @@ impl Client {
         serde_json::from_slice(resp.body()).wrap_err("Failed to decode body")
     }
 
-    /// Gets documentation of some field on config
+    /// Send a request to change the configuration of a specified field.
+    ///
+    /// # Errors
+    /// If sending request or decoding fails
+    pub fn set_config(&self, post_config: PostConfiguration) -> Result<bool> {
+        let headers = [("Content-type".to_owned(), "application/json".to_owned())]
+            .into_iter()
+            .collect();
+        let resp = http_client::post::<_, Vec<(&str, &str)>, _, _>(
+            &format!("{}/{}", self.torii_url, uri::CONFIGURATION),
+            serde_json::to_vec(&post_config)
+                .wrap_err(format!("Failed to serialize {:?}", post_config))?,
+            vec![],
+            headers,
+        )?;
+        if resp.status() != StatusCode::OK {
+            return Err(eyre!(
+                "Failed to post configuration with HTTP status: {}. {}",
+                resp.status(),
+                std::str::from_utf8(resp.body()).unwrap_or(""),
+            ));
+        }
+        serde_json::from_slice(resp.body())
+            .wrap_err(format!("Failed to decode body {:?}", resp.body()))
+    }
+
+    /// Get documentation of some field on config
+    ///
     /// # Errors
     /// Fails if sending request or decoding fails
     pub fn get_config_docs(&self, field: &[&str]) -> Result<Option<String>> {
         let field = field.iter().copied().map(ToOwned::to_owned).collect();
-        self.get_config(&GetConfiguration::Docs { field })
+        self.get_config(&GetConfiguration::Docs(field))
             .wrap_err("Failed to get docs for field")
     }
 
-    /// Gets value of config on peer
+    /// Get value of config on peer
+    ///
     /// # Errors
     /// Fails if sending request or decoding fails
     pub fn get_config_value(&self) -> Result<serde_json::Value> {
--- a/client/tests/reload.rs
+++ b/client/tests/reload.rs
@@ -1,0 +1,19 @@
+#![allow(clippy::restriction)]
+use iroha_logger::Level::*;
+use test_network::Peer as TestPeer;
+
+#[test]
+fn reload_log_level() {
+    let (_dont_drop, _dont_drop_either, cl) = <TestPeer>::start_test_with_runtime();
+
+    let verify = |level| {
+        let field: bool = cl
+            .set_config(iroha_core::torii::PostConfiguration::LogLevel(level))
+            .unwrap();
+        assert!(field);
+    };
+    verify(ERROR);
+    verify(TRACE);
+    verify(WARN);
+    verify(DEBUG);
+}
--- a/config/Cargo.toml
+++ b/config/Cargo.toml
@@ -5,6 +5,9 @@ authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
 edition = "2021"
 
 [dependencies]
+iroha_config_derive = { path = "derive" }
+
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-iroha_config_derive = { path = "derive" }
+thiserror = "1.0.30"
+crossbeam = "0.8.1"
--- a/config/src/runtime_upgrades.rs
+++ b/config/src/runtime_upgrades.rs
@@ -1,0 +1,376 @@
+use std::{
+    fmt::Debug,
+    sync::{MutexGuard, PoisonError},
+};
+
+pub use serde::{Deserialize, Serialize};
+use thiserror::*;
+
+type Result<T, E = ReloadError> = std::result::Result<T, E>;
+
+/// Error which occurs when reloading a configuration fails.
+#[derive(Clone, Copy, Debug, Error)]
+pub enum ReloadError {
+    /// The resource held by the handle was poisoned by a panic in
+    /// another thread.
+    #[error("Resource poisoned.")]
+    Poisoned,
+
+    /// The resource held by the handle was dropped.
+    #[error("Resource dropped.")]
+    Dropped,
+
+    /// If the reload handle wasn't properly initialised (using
+    /// [`handle::Singleton::set`]), there's nothing to reload with.
+    #[error("Cannot reload an uninitialised handle.")]
+    NotInitialised,
+
+    /// Error not specified by the implementer of the [`Reload`]
+    /// traits. Use as last resort.
+    #[error("Unspecified reload failure.")]
+    Other,
+}
+
+type PoisonErr<'a, T> = PoisonError<MutexGuard<'a, Option<Box<(dyn ReloadMut<T> + Send + Sync)>>>>;
+
+impl<T> From<PoisonErr<'_, T>> for ReloadError {
+    fn from(_: PoisonErr<'_, T>) -> Self {
+        Self::Poisoned
+    }
+}
+
+/// The field needs to be mutably borrowed to be reloaded.
+pub trait ReloadMut<T>: Debug {
+    // TODO: When negative traits/specialisation stabilises, Debug ought not to be a requirement.
+
+    /// Reload `self` using provided `item`.
+    ///
+    /// # Errors
+    /// Fails with an appropriate variant of
+    /// [`ReloadError`]. [`ReloadError::Other`] can be used as a
+    /// **temporary** placeholder.
+    fn reload(&mut self, item: T) -> Result<()>;
+}
+
+/// The field can be immutably borrowed and reloaded.
+pub trait Reload<T> {
+    /// Reload `self` using provided `item`.
+    ///
+    /// # Errors
+    /// Fails with an appropriate variant of [`ReloadError`].
+    /// [`ReloadError::Other`] can be used as a **temporary** placeholder.
+    fn reload(&self, item: T) -> Result<()>;
+}
+
+/// Contains [`handle`] types: opaque wrappers around a reloadable
+/// configuration, used to embed reloading functionality into
+/// various [`iroha_config_derive::Configurable`] types.
+///
+/// # Architecture.
+///
+/// ## Desired behaviour
+///
+/// Given a value of type (`<T = LogLevel>` in this module), need to
+///
+/// -  Embed a handle into the configuration options, replacing a Value
+/// of type <T> with a handle.
+///
+/// - The handle gets (de)serialized as if it were `<T>`: no extra
+/// fields, no extra initialisation.
+///
+/// - The configuration as a whole is immutable. This is to ensure
+/// that you don't accidentally re-assign the handle.
+///
+/// - The last object that got instantiated from the configuration
+/// file is modified when we call [`Reload::reload`].
+///
+/// - The value used to [`Reload::reload`] the value, must be reflected in the
+/// configuration.
+///
+/// ## Additional considerations
+///
+/// - The handle might have internal mutable state, and be passed
+/// along several threads in both a `sync` and `async` context.
+///
+/// - The handle's state can be a global mutable static value behind a
+/// wrapper.
+///
+/// - The handle is almost never read. All interactions with the
+/// handle are writes.
+///
+/// - The handle can retain a reference to different types, depending
+/// on the configuration options. The types might not all be known
+/// ahead of time, or be impractically long (both true for
+/// `tracting_subscriber::reload::Handle`).
+///
+/// # Usage
+///
+/// Embed a `SyncValue<T, H: super::Reload<T>>`, in your
+/// configuration options.  When using the configuration to initialise
+/// components, call [`handle::SyncValue::set_handle`], on a value that
+/// implements [`ReloadMut`] (which you defined earlier). Call
+/// [`handle::SyncValue::reload`] to change the configuration at run-time.
+///
+/// If the type stored in `H` is a single simple type, it is
+/// recommended to use a custom tuple `struct`, and `impl`
+/// [`Reload`] for it.
+///
+/// If the types are too varied, or generic in arguments that change
+/// depending on run-time values, (as in
+/// e.g. `tracing_subscriber::reload::Handle`), it is recommended to
+/// instead use the provided opaque wrapper [`handle::Singleton`].
+///
+/// **NOTE** you shouldn't normally need to use either
+/// [`handle::Singleton`] or [`handle::Value`] directly.
+///
+/// # Examples
+/// ```rust,ignore
+/// use iroha_config_derive::Configurable;
+/// use serde::{Deserialize, Serialize};
+/// use iroha_config::runtime_upgrades::{handle, Reload, ReloadMut, ReloadError};
+/// use tracing::Level;
+/// use tracing_subscriber::{reload::Handle, filter::LevelFilter};
+/// use std::fmt::Debug;
+///
+/// struct Logger;
+///
+/// #[derive(Clone, Deserialize, Serialize, Debug, Configurable)]
+/// struct Configuration {
+///     pub max_log_level: handle::SyncValue<Level, handle::Singleton<Level>>,
+///     pub log_file_path: Option<std::path::PathBuf>,
+/// }
+///
+/// fn init(config: &Configuration) -> Logger {
+///     let level = config.max_log_level.value();
+///     let level_filter = tracing_subscriber::filter::LevelFilter::from_level(level);
+///     let (filter, handle) = reload::Layer::new(level_filter);
+///     config.max_log_level.set_handle(handle).unwrap();
+/// }
+///
+/// impl<T: Subscriber + Debug> ReloadMut<Level> for Handle<LevelFilter, T> {
+///    fn reload(&mut self, level: Level) -> Result<(), ReloadError> {
+///         let level_filter = LevelFilter::from_level(level);
+///         Handle::reload(self, level_filter).map_err(|_todo| ReloadError::Dropped)
+///    }
+/// }
+/// ```
+
+pub mod handle {
+    use std::{
+        fmt::{Debug, Formatter},
+        sync::{Arc, Mutex},
+    };
+
+    use crossbeam::atomic::AtomicCell;
+    use serde::{Deserialize, Serialize};
+
+    use super::{Reload, ReloadError, ReloadMut, Result};
+    // -----------------------------------------------------------------
+
+    /// An opaque handle for arbitrary [`super::ReloadMut<T>`], useful
+    /// when it is either impossible or impractical to specify a
+    /// single `enum` or generic type.  You shouldn't embed this into
+    /// your configuration, and instead use [`SyncValue`].
+    #[derive(Clone, Serialize, Deserialize)]
+    pub struct Singleton<T: Send + Sync> {
+        #[serde(skip)]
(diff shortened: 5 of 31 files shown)

### CODE REVIEW

File: config/src/runtime_upgrades.rs [hunk 1]
```diff
@@ -0,0 +1,136 @@
+use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
```
i1i1, line 1: May be it is a better idea to make it async? As using sync version (in async context) might lead to bugs

core/src/torii/mod.rs
```diff
@@ -71,6 +71,9 @@ pub enum Error {
     /// Failed to push into queue.
     #[error("Failed to push into queue")]
     PushIntoQueue(#[source] Box<queue::Error>),
+    /// Configuration change error.
+    #[error("Attempt to change configuration failed. {0}")]
+    ConfigurationReload(#[source] iroha_config::runtime_upgrades::ReloadFailure),
```
i1i1, line 76:
```suggestion
    #[error("Attempt to change configuration failed")]
    ConfigurationReload(#[source] iroha_config::runtime_upgrades::ReloadFailure),
```

core/src/torii/mod.rs
```diff
@@ -336,8 +355,10 @@ async fn handle_get_configuration<W: WorldTrait>(
     match get_cfg {
         Docs { field } => {
             Configuration::get_doc_recursive(field.iter().map(AsRef::as_ref).collect::<Vec<&str>>())
-                .wrap_err("Failed to get docs")
-                .and_then(|doc| serde_json::to_value(doc).wrap_err("Failed to serialize docs"))
+                .wrap_err("Failed to get docs {:?field}")
```
i1i1, line 358:
```suggestion
                .wrap_err_with(|| format!("Failed to get docs {:?field})")
```

logger/src/config.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub max_log_level: LevelEnv,
+    /// Capacity (or batch size) for telemetry channel
+    pub telemetry_capacity: usize,
+    /// Compact mode (no spans from telemetry)
+    pub compact_mode: bool,
+    /// If provided, logs will be copied to said file in the
+    /// format readable by [bunyan](https://lib.rs/crates/bunyan)
+    pub log_file_path: Option<std::path::PathBuf>,
+    /// Internal: handle to reload the log level.
+    pub log_reload_handle: ReloadHandle<LevelEnv>,
```
i1i1, line 83: Is there a need for whole mutex here? We can use just atomic for that (or `crossbeam::AtomicCell`)

config/src/runtime_upgrades.rs
eadventurous: I agree sync Mutex can freeze thread and tokio or any other async runtime does not account for it

config/Cargo.toml
```diff
@@ -8,3 +8,4 @@ edition = "2021"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 iroha_config_derive = { path = "derive" }
+thiserror = "1.0.30"
```
eadventurous, line 11: Minor one, but please add a new line here :)

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Error not specified by the implementors of the [`Reloadable`]
+    /// traits. Use as last resort.
+    #[error("Unspecified reload failure.")]
+    Other,
+}
+
+impl<T>
+    From<
+        PoisonError<
+            MutexGuard<'_, std::option::Option<Box<(dyn Reloadable<T> + Send + Sync + 'static)>>>,
```
eadventurous, line 29: Why not simply `Option`? I don't see any other `Option` in the context.

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,136 @@
...
+/// configuration. This handle must implement [`Reloadable<T>`] and be
+/// stored in [`ReloadHandle<T>`], using [`ReloadHandle::set`].
+///
+/// If the Handle is stored with the configuration, each cloned
+/// instance contains an [`std::sync::Arc`] with a [`Mutex`], so the
+/// containing object (configuration) doesn't need to be `mut`.
+///
+/// [`ReloadHandle`] uses dynamic dispatch (intentionally) in cases
+/// (e.g. [`tracing_subscriber`]) where the reloading process makes
+/// use of impractically many generic arguments. If this is not the
```
eadventurous, line 65: Yep, I agree that this is a good choice.

core/src/torii/mod.rs
```diff
@@ -347,6 +368,33 @@ async fn handle_get_configuration<W: WorldTrait>(
...
+async fn handle_post_configuration<W: WorldTrait>(
+    state: ToriiState<W>,
+    cfg: PostConfiguration,
+) -> Result<Json> {
+    use iroha_config::runtime_upgrades::InnerMutablyReloadable;
+    use PostConfiguration::*;
+
+    iroha_logger::debug!(?cfg);
+    match cfg {
+        // TODO: Now the configuration value and the actual value don't match.
```
eadventurous, line 391: Let's create an issue for it?

logger/src/config.rs
```diff
@@ -0,0 +1,99 @@
...
+use tracing_subscriber::{filter::LevelFilter, reload::Handle};
+
+const DEFAULT_MAX_LOG_LEVEL: LevelEnv = LevelEnv::DEBUG;
+
+use super::{debug, Level};
+
+/// Log level for reading from environment and (de)serializing
+#[allow(clippy::upper_case_acronyms)]
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum LevelEnv {
```
eadventurous, line 19: I am not sure about naming, right now we don't only read from env, but we also have post endpoint and config.json

eadventurous: So why name it `LevelEnv`...

core/src/torii/mod.rs
appetrosyan: #1612

logger/src/config.rs
appetrosyan: I didn't want to change too many things at once. Since it's now in a completely separate module I can just call it `Level` and use the appropriate namespace when ambiguous.

config/src/runtime_upgrades.rs
appetrosyan: https://docs.rs/tokio/1.13.1/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

My rationale was that in this particular case everything takes place within synchronous code. I have no problem with re-implementing using the `tokio_mutex`.

s8sato: As for `tokio::sync::Mutex`, I thought it should be avoided as much as possible in practice
https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex

core/src/torii/mod.rs
```diff
@@ -347,6 +368,33 @@ async fn handle_get_configuration<W: WorldTrait>(
     .map_err(Error::Config)
 }
 
+/// Message acceptable for `POST` requests to the configuration endpoint.
+#[derive(Clone, Debug, Deserialize, Serialize, Copy)]
+pub enum PostConfiguration {
+    /// Change the maximum logging level of logger.
+    ///
+    /// # Examples
+    /// `curl -X POST -H 'content-type: application/json' http://127.0.0.1:8080/configuration -d '{"ChangeLogLevel": "DEBUG"}' -i`
```
s8sato, line 377: This fails for now though. Am I missing something?

2021-11-17T183629+090000

core/src/torii/mod.rs
```diff
@@ -336,8 +355,10 @@ async fn handle_get_configuration<W: WorldTrait>(
     match get_cfg {
         Docs { field } => {
             Configuration::get_doc_recursive(field.iter().map(AsRef::as_ref).collect::<Vec<&str>>())
-                .wrap_err("Failed to get docs")
-                .and_then(|doc| serde_json::to_value(doc).wrap_err("Failed to serialize docs"))
+                .wrap_err_with(|| format!("Failed to get docs {:?}", field))
+                .and_then(|doc| {
+                    Context::wrap_err(serde_json::to_value(doc), "Failed to serialize docs")
+                })
```
s8sato, line 361:
```suggestion
                .and_then(|doc| serde_json::to_value(doc).wrap_err("Failed to serialize docs"))
```

.github/workflows/iroha2-dev-pr-heavy.yml
```diff
@@ -1,7 +1,7 @@
 name: I2::Dev::Tests (Heavy)
 
 on:
-  pull_request:
+  push:
```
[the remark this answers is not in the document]
appetrosyan: Hopefully it is temporary. It might be a good idea to move both the docker test and the benchmarks into the workflow for merging from `iroha2-dev` to `iroha2`. I'm doing it until we fix the docker test (or move it to a kubernetes test).

core/src/torii/mod.rs
appetrosyan: This is indeed a bug, but I can't seem to reproduce it reliably.

core/src/torii/mod.rs
```diff
@@ -432,463 +488,4 @@ pub mod config {
...
-            // .given(register_account("alice"))
-            // .keys(KeyPair::generate().unwrap())
-            // .deny_all()
-            .query(query())
-            .status(StatusCode::NOT_FOUND)
-            .assert()
-            .await
-    }
-}
+pub mod tests;
```
mversic, line 491: why public?

client/src/client.rs
```diff
@@ -632,31 +657,7 @@ pub mod transaction {
...
-    /// Guidance](https://prometheus.io/docs/instrumenting/writing_exporters/).
-    pub const METRICS: &str = "/metrics";
-    /// The URI used for block synchronization.
-    pub const BLOCK_SYNC: &str = "/block";
-    /// The web socket uri used to subscribe to pipeline and data events.
-    pub const SUBSCRIPTION: &str = "/events";
-    /// Get pending transactions.
-    pub const PENDING_TRANSACTIONS: &str = "/pending_transactions";
-}
+use iroha_core::torii::uri;
```
mversic, line 660: shouldn't this be at the top of the file? 
Why were api endpoints duplicated in both torii and client?

core/src/config.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
 };
 
 /// Configuration parameters container.
-#[derive(Clone, Default, Deserialize, Serialize, Debug, Configurable, PartialEq, Eq)]
```
mversic, line 22: why?

appetrosyan: When the configurations were a POD datatype, they could be hashed, compared for equality, among other things. 

Now the configuration (at least for the logger) contains both interior mutable state, as well as a `SideEffect` which cannot (and should not) be possible to compare for equality. So it's no longer true that in-general the configurations are `PartialEq` if you need this functionality for something, you can manually derive it on the configurations that have this property. 

Moreover, this is not a property that we made use of. There are no instances of code that broke because of lack of `PartialEq`.

client/src/client.rs
appetrosyan: > shouldn't this be at the top of the file?

No problem. 

> Why were api endpoints duplicated in both torii and client?

The only thing that I needed to change was add a `/` to some format statements. So probably some miscommunication.

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// (e.g. [`tracing_subscriber`]) where the reloading process makes
+    /// use of impractically many generic arguments. If this is not the
+    /// case, implement [`Reloadable`] on the type directly.
+    ///
+    /// Do not implement [`Reloadable`] for the configuration file as a
+    /// whole, instead explicitly reload the object stored by this handle.
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct SideEffect<T: Send + Sync> {
+        #[serde(skip)]
+        inner: Arc<Mutex<Option<Box<dyn Reloadable<T> + Send + Sync>>>>,
```
mversic, line 118: what about `OnceCell<Mutex<Box<dyn Reloadable ...>>>` vs `Arc`?

core/src/torii/mod.rs
```diff
@@ -347,6 +373,36 @@ async fn handle_get_configuration<W: WorldTrait>(
...
 
+/// Message acceptable for `POST` requests to the configuration endpoint.
+#[derive(Clone, Debug, Deserialize, Serialize, Copy)]
+pub enum PostConfiguration {
+    /// Change the maximum logging level of logger.
+    ///
+    /// # Examples
+    /// To silence all logging events that aren't `ERROR`s
+    /// `curl -X POST -H 'content-type: application/json' http://127.0.0.1:8080/configuration -d '{"ChangeLogLevel": "ERROR"}' -i`
+    ChangeLogLevel(iroha_logger::Level),
```
mversic, line 384: I would just call it `LogLevel`. `POST` http request indicates imperative mode

config/src/runtime_upgrades.rs
mversic: maybe I'm wrong here. Disregard if that's the case

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    ///
+    /// # Errors
+    /// Fails with an appropriate variant of
+    /// [`ReloadFailure`]. [`ReloadFailure::Other`] can be used as a
+    /// **temporary** placeholder.
+    fn reload(&mut self, item: T) -> Result<()>;
+}
+
+/// Implement this trait if the object can be immutably borrowed and `reload`-ed.
+pub trait InnerMutablyReloadable<T> {
```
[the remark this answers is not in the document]
mversic: also `able` suffix doesn't quite sound rusty to me :)

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Reload `self` using provided `item`.
+    ///
+    /// # Errors
+    /// Fails with an appropriate variant of [`ReloadFailure`].
+    /// [`ReloadFailure::Other`] can be used as a **temporary** placeholder.
+    fn reload(&self, item: T) -> Result<()>;
+}
+
+impl<T> std::fmt::Debug for dyn Reloadable<T> {
```
mversic, line 63: alternative could be to write a blanket implementation like:
`impl<T, K: Reload<T>> std::fmt::Debug for K`

.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -29,9 +29,12 @@ jobs:
...
         run: cargo lints clippy --workspace --benches --tests
+        if: always()
       - name: Static analysis with all features enabled
         run: cargo lints clippy --workspace --benches --tests --all-features
+        if: always()
       - name: Documentation check
         run: |
           cargo doc --no-deps
           ./scripts/check_docs.sh
+        if: always()
```
eadventurous, line 40: Isn't it semantically redundant?

config/src/runtime_upgrades.rs
mversic, line 63: I would maybe rather have `std::fmt::Debug` be a supertrait of `Reload`? like:
`pub trait Reload<T>: std::fmt::Debug`

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
+use std::sync::{MutexGuard, PoisonError};
+
+pub use serde::{Deserialize, Serialize};
+use thiserror::*;
+
+type Result<T, E = ReloadFailure> = std::result::Result<T, E>;
+
+/// Error which occurs when reloading a configuration fails.
+#[derive(Clone, Copy, Debug, Error)]
+pub enum ReloadFailure {
```
mversic, line 10: Wouldn't you rather name it `ReloadError`?

config/src/runtime_upgrades.rs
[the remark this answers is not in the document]
appetrosyan: I agree about the naming. `Reload`, `ReloadMut` and `Reconfigure` sound more rust-y (however our code-base doesn't seem to follow rust's naming convention anyway, preferring a Java-like convention).

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Locking the Mutex within [`ReloadHandle`] failed.
+    #[error("Cannot acquire Mutex Lock. Poisoned Mutex.")]
+    PoisonedMutex,
```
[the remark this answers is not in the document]
appetrosyan: We don't. Not yet anyway.

client/src/client.rs
```diff
@@ -446,16 +452,43 @@ impl Client {
...
     }
 
-    /// Gets documentation of some field on config
+    /// Reload a configuration for the field.
+    ///
+    /// # Errors
+    /// If sending request or decoding fails
+    pub fn set_config(&self, post_config: iroha_core::torii::PostConfiguration) -> Result<bool> {
+        let mut headers = self.headers.clone();
+        headers.insert("Content-type".to_owned(), "application/json".to_owned());
```
eadventurous, line 461: Let's move headers to const

config/src/runtime_upgrades.rs
appetrosyan: I actually considered it first. The issue is that it's a nightly-only experimental feature.

.github/workflows/iroha2-dev-pr-static.yml
appetrosyan: Nope. What this does is it forces these steps to work regardless if the previous step failed. So if someone needs to do all three: fix format, add clippy comments and generate docs, they can do that all at once.

client/tests/reload.rs
```diff
@@ -0,0 +1,20 @@
+#![allow(clippy::restriction)]
+use iroha_logger::Level::*;
+use test_network::Peer as TestPeer;
+
+#[test]
+fn reload_log_level() {
+    // The underscored variables must not be dropped until end of closure.
```
eadventurous, line 7: Generally the comment about not dropping the runtime I think can be moved into description of start test with runtime function.

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Error not specified by the implementors of the [`Reloadable`]
+    /// traits. Use as last resort.
+    #[error("Unspecified reload failure.")]
+    Other,
+}
+
+impl<T>
+    From<
+        PoisonError<
+            MutexGuard<'_, std::option::Option<Box<(dyn Reloadable<T> + Send + Sync + 'static)>>>,
```
eadventurous, line 29: Type aliasing a part of the inner type might help with readability

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// # Errors
+    /// Fails with an appropriate variant of [`ReloadFailure`].
+    /// [`ReloadFailure::Other`] can be used as a **temporary** placeholder.
+    fn reload(&self, item: T) -> Result<()>;
+}
+
+impl<T> std::fmt::Debug for dyn Reloadable<T> {
+    fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Ok(())
+    }
```
eadventurous, line 66: Any particular reason for not importing Debug and Formatter?

config/src/runtime_upgrades.rs
mversic: @appetrosyan there is a crate `once_cell` that gives you the same api in stable Rust and we have it as a dependency already AFAIK

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// configuration. This handle must implement [`Reloadable<T>`] and be
+    /// stored in [`SideEffect<T>`], using [`ReloadHandle::set`].
+    ///
+    /// If the Handle is stored with the configuration, each cloned
+    /// instance contains an [`std::sync::Arc`] with a [`std::sync::Mutex`], so the
+    /// containing object (configuration) doesn't need to be `mut`.
+    ///
+    /// [`SideEffect`] uses dynamic dispatch (intentionally) in cases
+    /// (e.g. [`tracing_subscriber`]) where the reloading process makes
+    /// use of impractically many generic arguments. If this is not the
```
mversic, line 110: did you consider using enum instead of trait object? They are interchangeable when you know ahead of time all possible implementors of your trait.

I have no idea if this is applicable here, just pointing this out

core/test_network/src/lib.rs
```diff
@@ -378,9 +375,9 @@ where
             },
             logger: LoggerConfiguration {
                 #[cfg(profile = "bench")]
-                max_log_level: LevelEnv::WARN,
+                max_log_level: Level::ERROR.into(),
                 #[cfg(not(profile = "bench"))]
-                max_log_level: LevelEnv::INFO,
+                max_log_level: Level::TRACE.into(),
```
eadventurous, line 380: I think it should stay at INFO level, and if during local debugging we need a TRACE level we can enable it temporarily. But in general especially for readability in CI, it is better to have INFO here.

core/src/torii/mod.rs
appetrosyan: imperative *mood* but yeah, I agree.

config/src/runtime_upgrades.rs
appetrosyan: Not sure it's a good idea to require that anything that is reloadable be `Debug`.

mversic: but you obviously want the `Debug` in your `Reloadable`s. Why would you be making this implementation otherwise?

config/src/runtime_upgrades.rs
appetrosyan: Old habit.

config/src/runtime_upgrades.rs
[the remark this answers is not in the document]
appetrosyan: I don't agree on making `ReloadMut` a supertrait. As is if you implement `Reload` you get `ReloadMut` for free. As a result you can use `Reload` and `ReloadMut` interchangeably for a `Reload` object. If you make it a supertrait, you'd force the user to implement both traits, (since it's now a cycle and there'd be conflicting `impl`s).

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,278 @@
...
+impl<T>
+    From<
+        PoisonError<
+            MutexGuard<'_, std::option::Option<Box<(dyn Reloadable<T> + Send + Sync + 'static)>>>,
+        >,
+    > for ReloadFailure
+{
+    fn from(
+        _: PoisonError<
+            MutexGuard<'_, std::option::Option<Box<(dyn Reloadable<T> + Send + Sync + 'static)>>>,
```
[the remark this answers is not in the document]
appetrosyan: `'static` isn't but the other two are necessary.

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// [`ReloadFailure::Other`] can be used as a **temporary** placeholder.
+    fn reload(&self, item: T) -> Result<()>;
+}
+
+/// Contains [`handle`] types: opaque wrappers around a reloadable
+/// configuration.  [`handle::Reconfigurable`] is the most commonly
+/// useful type of handle, as it deserializes into the inner object,
+/// as well as handles synchronisation of changes to the program state
+/// via the handle, to the changes in the configuration value.
+pub mod handle {
```
eadventurous, line 66: I think it would be very useful to have doc comments with minimal usage examples of each `Atomic`, `Singleton`, `Reconfigurable`.

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,265 @@
...
+        pub fn value(&self) -> T {
+            self.0 .0.load()
+        }
+    }
+
+    impl<T: Clone + Copy + Send + Sync + Debug> Reconfigure<T, Singleton<T>> {
+        /// Set the [`SideEffect`] handle
+        ///
+        /// # Errors
+        /// If [`SideEffect::handle`] fails.
```
eadventurous, line 254: It is no longer named SideEffect. There are several places where this needs to be replaced, if I understood it correctly.

config/src/runtime_upgrades.rs
eadventurous: Also their naming is a bit inconsistent with each other to my mind.

mversic (approved):

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// when it is either impossible or impractical to specify a
+    /// single `enum` or generic type.  You shouldn't embed this into
+    /// your configuration, and instead use [`SyncValue`].
+    #[derive(Clone, Serialize, Deserialize)]
+    pub struct Singleton<T: Send + Sync> {
+        #[serde(skip)]
+        inner: Arc<Mutex<Option<Box<dyn ReloadMut<T> + Send + Sync>>>>,
+    }
+
+    impl<T: Send + Sync> Default for Singleton<T> {
```
mversic, line 180: use derive

config/src/runtime_upgrades.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+
+    impl<T: Clone + Copy> From<T> for Value<T> {
+        fn from(value: T) -> Self {
+            Self(AtomicCell::new(value))
+        }
+    }
+
+    impl<T: Default + Clone + Copy> Default for Value<T> {
```
mversic, line 260: use derive

mversic, line 260: you can use `derive` and `T: Default` would be implied

eadventurous (approved):

config/src/runtime_upgrades.rs
appetrosyan: Unfortunately it doesn't seem to offer the interior mutability, as well as semantically accurate cloning behaviour.

eadventurous (approved):

mversic (approved):

PR #1613, opened 2021-11-15
poll instead of pushing latest blocks

### LINKED ISSUE #1236: Request block hash instead of flooding peer with it
**Pull vs. push for data syncing.**

Makoto: This is too much message passing. Instead, peers who want to sync should request randomly from peers that they know, rather than getting messages from other peers. It needs to be a "pull" modality, not push.

Egor:
So in terms of an algorithm:

    Every N seconds peer requests latest block hashes from a small random subset of all peers.
    Peers answer to this requests with their latest bock hashes
    Peer compares its own block hash to the received ones
    ...

Something like that?

mversic:
### Description of the Change

Each peer selects one partner peer at random and sends it `GetBlocksAfter` request(containing it's latest block hash) for latest blocks. Since the case of peers getting out of sync should be quite rare, there is a great chance for initiator peer to choose partner peer that has up to date blocks

Also, there was a bug with iterator of `Chain` which is backed by concurrent hash map. The bug was that iterator was returning chain blocks in unpredictable order since hash map doesn't have any order of it's entries. There was at least one place in the code where it was wrongly assumed chain iterator returns blocks in order. This PR fixes that by implementing `Chain` iterator which returns blocks in the chronological order

### Issue

Closes #1236 

As an improvement block height could be sent together with block hash in `GetBlocksAfter` message so that peer receiving the message can use that information to position itself in the blockchain instead of searching through it.

Once Hijiri is implemented we can use it to choose partner peer in a smarter way, i.e. to choose partner peer among peers which have chain sizes strictly greater than the peer requesting new blocks

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -539,7 +539,7 @@ impl ValidBlock {
     /// Creates dummy `ValidBlock`. Used in tests
     ///
     /// # Panics
-    /// Should never panic
+    /// If generating keys or block signing fails.
     #[cfg(test)]
     #[allow(clippy::restriction)]
     pub fn new_dummy() -> Self {
@@ -778,63 +778,63 @@ mod tests {
         let valid_block = ValidBlock::new_dummy();
         let committed_block = valid_block.clone().commit();
 
-        assert_eq!(valid_block.hash().transmute(), committed_block.hash())
+        assert_eq!(*valid_block.hash(), *committed_block.hash())
     }
 
     #[test]
     pub fn chain_iter_returns_blocks_ordered() {
-        const BLOCK_CNT: usize = 10;
+        const BLOCK_COUNT: usize = 10;
         let chain = Chain::new();
 
         let mut block = ValidBlock::new_dummy().commit();
 
-        for i in 1..=BLOCK_CNT {
+        for i in 1..=BLOCK_COUNT {
             block.header.height = i as u64;
             chain.push(block.clone().into());
         }
 
         assert_eq!(
-            (BLOCK_CNT - 5..=BLOCK_CNT)
+            (BLOCK_COUNT - 5..=BLOCK_COUNT)
                 .map(|i| i as u64)
                 .collect::<Vec<_>>(),
             chain
                 .iter()
-                .skip(BLOCK_CNT - 6)
+                .skip(BLOCK_COUNT - 6)
                 .map(|b| *b.key())
                 .collect::<Vec<_>>()
         );
 
-        assert_eq!(BLOCK_CNT - 2, chain.iter().skip(2).count());
+        assert_eq!(BLOCK_COUNT - 2, chain.iter().skip(2).count());
         assert_eq!(3, *chain.iter().nth(2).unwrap().key());
     }
 
     #[test]
     pub fn chain_rev_iter_returns_blocks_ordered() {
-        const BLOCK_CNT: usize = 10;
+        const BLOCK_COUNT: usize = 10;
         let chain = Chain::new();
 
         let mut block = ValidBlock::new_dummy().commit();
 
-        for i in 1..=BLOCK_CNT {
+        for i in 1..=BLOCK_COUNT {
             block.header.height = i as u64;
             chain.push(block.clone().into());
         }
 
         assert_eq!(
-            (1..=BLOCK_CNT - 4)
+            (1..=BLOCK_COUNT - 4)
                 .rev()
                 .map(|i| i as u64)
                 .collect::<Vec<_>>(),
             chain
                 .iter()
                 .rev()
-                .skip(BLOCK_CNT - 6)
+                .skip(BLOCK_COUNT - 6)
                 .map(|b| *b.key())
                 .collect::<Vec<_>>()
         );
 
         assert_eq!(
-            (BLOCK_CNT - 2) as u64,
+            (BLOCK_COUNT - 2) as u64,
             *chain.iter().nth_back(2).unwrap().key()
         );
     }
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -5,7 +5,7 @@
 
 use std::{collections::BTreeSet, iter, marker::PhantomData};
 
-use dashmap::{iter::Iter as MapIter, mapref::one::Ref as MapRef, DashMap};
+use dashmap::{mapref::one::Ref as MapRef, DashMap};
 use eyre::{Context, Result};
 use iroha_crypto::{HashOf, KeyPair, SignatureOf, SignaturesOf};
 use iroha_data_model::{current_time, events::prelude::*, transaction::prelude::*};
@@ -62,8 +62,8 @@ impl Chain {
     }
 
     /// Iterator over height and block.
-    pub fn iter(&self) -> MapIter<u64, VersionedCommittedBlock> {
-        self.blocks.iter()
+    pub fn iter(&self) -> ChainIterator {
+        ChainIterator::new(self)
     }
 
     /// Latest block reference and its height.
@@ -79,6 +79,48 @@ impl Chain {
     /// Whether blockchain is empty.
     pub fn is_empty(&self) -> bool {
         self.blocks.is_empty()
+    }
+}
+
+/// Chain iterator
+pub struct ChainIterator<'a> {
+    chain: &'a Chain,
+    pos: u64,
+}
+
+impl<'a> ChainIterator<'a> {
+    fn new(chain: &'a Chain) -> Self {
+        ChainIterator { chain, pos: 0 }
+    }
+}
+
+impl<'a> Iterator for ChainIterator<'a> {
+    type Item = MapRef<'a, u64, VersionedCommittedBlock>;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.pos += 1;
+        self.chain.blocks.get(&self.pos)
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        self.pos += n as u64;
+        self.next()
+    }
+
+    fn last(mut self) -> Option<Self::Item> {
+        self.pos = self.chain.len() as u64;
+        self.chain.blocks.get(&self.pos)
+    }
+
+    fn count(self) -> usize {
+        #[allow(clippy::cast_possible_truncation)]
+        let count = (self.chain.len() as u64 - self.pos) as usize;
+        count
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        #[allow(clippy::cast_possible_truncation)]
+        let height = (self.chain.len() as u64 - self.pos) as usize;
+        (height, Some(height))
     }
 }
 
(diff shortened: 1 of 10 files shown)


core/src/sumeragi/mod.rs
```diff
@@ -78,10 +78,10 @@ pub struct Init {
     pub height: u64,
 }
 
-/// Get sorted peers
+/// Get random amount of peers
```
appetrosyan, line 81:
```suggestion
/// Get a number of randomly selected peers.
```
"Random amount" means that you have a random usize.

core/src/block_sync.rs
```diff
@@ -94,8 +96,7 @@ pub struct ContinueSync;
 /// Message for getting updates from other peers
 ///
 /// Starts the `BlockSync`, meaning that every `gossip_period`
-/// the peers would gossip about latest block hashes
-/// and try to synchronize their blocks.
+/// peer would poll randomly selected `sync_peers_count` peers for their blocks
```
eadventurous, line 99:
```suggestion
/// peer would poll randomly selected `sync_peers_count` peers for their latest block hashes
/// and try to sync in case it has a different block hash.
```

core/src/block_sync.rs
```diff
@@ -242,22 +246,6 @@ pub mod message {
...
-        /// Peer id
-        pub peer_id: PeerId,
-    }
-
-    impl LatestBlock {
-        /// Default constructor
-        pub const fn new(hash: HashOf<VersionedCommittedBlock>, peer_id: PeerId) -> Self {
-            Self { hash, peer_id }
-        }
-    }
```
eadventurous, line 259: We still need this message. As mentioned in the task description peer should request latest block hashes from selected number of peers, then compare with its own latest block hash, after that in case of difference try to synchronize.

mversic: @eadventurous it seemed redundant to me to. I thought peer requesting latest blocks from other peers would just send it's latest block hash in `GetBlocksAfter` and other peers would respond with blocks after the latest block hash peer had sent

mversic: why wouldn't peers that provide latest blocks be put in the position to compare and respond if required?

mversic: because, otherwise, we are introducing another roundtrip of messaging

mversic: In the current PR, considering the communication between two nodes, there are 2 messages sent in total. In the case you are proposing there are 4 messages sent in total. Previously, there were 3 messages sent

eadventurous: Then there might be a situation:
1. New peer joins the network - it has 0 blocks
2. Network is already operating for a while and has 100 GB of blocks
3. Peer asks a set of N other randomly selected peers to give blocks after initial zero hash.
4. All N peers start to send 100 GB of blocks to it.

This seems to both create an extensive load on existing peers and on the new peer, with a lot of redundant bytes being sent.

eadventurous: Synchronization will be N times slower also in this case.

mversic: I see, payload is an issue, tnx

core/src/sumeragi/network_topology.rs
```diff
@@ -391,7 +391,7 @@ impl Topology {
         signatures: impl IntoIterator<Item = &'a SignatureOf<VersionedValidBlock>> + 'a,
     ) -> Vec<SignatureOf<VersionedValidBlock>> {
         let roles: HashSet<Role> = roles.iter().copied().collect();
-        let public_keys: Vec<_> = roles
+        let public_keys: HashSet<_> = roles
```
mversic, line 394: Is this irrelevant?

eadventurous: Sure, you can do this for clarity

core/src/block_sync.rs
```diff
@@ -108,20 +116,43 @@ impl<S: SumeragiTrait, W: WorldTrait> Actor for BlockSynchronizer<S, W> {
     async fn on_start(&mut self, ctx: &mut Context<Self>) {
         self.broker.subscribe::<Message, _>(ctx);
         self.broker.subscribe::<ContinueSync, _>(ctx);
-        ctx.notify_every::<ReceiveUpdates>(self.gossip_period);
+        ctx.notify_every::<PollBlockUpdates>(self.gossip_period);
+        ctx.notify_every::<PollHeightUpdates>(self.gossip_period);
```
eadventurous, line 120: This should probably be different configurable periods?

mversic: yes, done

core/src/sumeragi/network_topology.rs
mversic: not just for clarity, this collection is searched for an item in a loop so should be better to have set instead of a vec

core/src/block_sync.rs
```diff
@@ -137,11 +131,21 @@ impl<S: SumeragiTrait, W: WorldTrait> Handler<ContinueSync> for BlockSynchronize
...
-        message.handle(&mut self).await
+        message.handle(&mut self).await;
     }
 }
 
 impl<S: SumeragiTrait + Debug, W: WorldTrait> BlockSynchronizer<S, W> {
+    /// Sends request for latest blocks to a chosen peer
+    async fn request_latest_blocks_from_peer(&mut self, peer_id: PeerId) {
+        Message::GetBlocksAfter(GetBlocksAfter::new(
+            self.wsv.latest_block_hash(),
```
s8sato, line 142: If the chosen peer does not have the `latest_block_hash`,
while `GetBlocksAfter` handling it will log `ParentHashNotFound` error .
Should it be logged as an ERROR?

mversic: I reduced log level to `warn`

s8sato: I still don't know if that's okay.
Before the PR, where receiving `LatestBlock` was followed by `GetBlocksAfter` request,
`ParentHashNotFound` did indicate some kind of fraud and ERROR was appropriate.

Now the error includes both possibilities:
1. Not found because the hash is new to the responder -- common case
2. Not found because the hash is invalid -- real error case

mversic: As far as I can see both possibilities were there before this PR too. 
Peer would send it's latest block hash to all other peers and other peers would just check that block is not it's last block hash and not whether or not it actually exists in it's chain. If it did exist in it's chain as one of the previous blocks GetBlocksAfter would error out. I don't see any difference there

If you'd like, what I can propose is to also send height of block together with block hash in the `LatestBlock` message and in that way mitigate first error case.

s8sato: You are right.
I overlooked the possibility that the requester was a few blocks ahead.
Although it is a common case, ERROR seems to have been logged from before the PR

mversic: so I returned the `error` log level. Are we good here?

I don't think it's possible to differentiate whether a given hash is invalid or it's ahead in the chain from the perspective of one peer

appetrosyan (approved):

core/src/block_sync.rs
```diff
@@ -312,36 +291,24 @@ pub mod message {
...
                 Message::GetBlocksAfter(GetBlocksAfter { hash, peer_id }) => {
                     if block_sync.batch_size == 0 {
                         iroha_logger::warn!(
                             "Error: not sending any blocks as batch_size is equal to zero."
                         );
                         return;
                     }
+                    if *hash == block_sync.wsv.latest_block_hash() {
+                        return;
+                    }
```
eadventurous, line 303: Do we need to check it? I think `blocks_after` should just return an empty set if its latest block matches this hash?

core/src/sumeragi/mod.rs
```diff
@@ -377,12 +377,18 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait> ContextHandler<Init>
...
+    async fn handle(&mut self, GetRandomPeer: GetRandomPeer) -> Self::Result {
+        let rng = &mut rand::thread_rng();
+        #[allow(clippy::expect_used)]
+        self.topology
+            .sorted_peers()
+            .choose(rng)
+            .expect("No nodes in topology")
+            .clone()
     }
 }
```
eadventurous, line 393: I think Sumeragi should not concern itself with implementation specifics of block sync, so getting all sorted peers - which is a general case and then in block sync choosing the peer makes more sense to me.

Also I think we can get peers from wsv and it is also always up to date. I think it's better to get it from there as a single central point for other components.

core/src/block.rs
```diff
@@ -82,6 +82,77 @@ impl Chain {
...
+    fn count(self) -> usize {
+        #[allow(clippy::cast_possible_truncation)]
+        let count = (self.chain.len() as u64 - (self.pos_front - 1)) as usize;
+        count
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        #[allow(clippy::cast_possible_truncation)]
+        let height = (self.chain.len() as u64 - (self.pos_front - 1)) as usize;
+        (height, Some(height))
```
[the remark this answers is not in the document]
mversic: I thought it's not possible to insert more than `isize` elements into a collection and therefore this conversion would never fail? or am I wrong here?

core/src/block.rs
```diff
@@ -674,35 +771,71 @@ impl From<&CommittedBlock> for Vec<Event> {
...
-                rejected_transactions_hash: EmptyChainHash::default().into(),
-                view_change_proofs: view_change::ProofChain::empty(),
-                invalidated_blocks_hashes: Vec::new(),
-                genesis_topology: None,
-            },
-            rejected_transactions: vec![],
-            transactions: vec![],
-            signatures: BTreeSet::default(),
+    pub fn chain_iter_returns_blocks_ordered() {
+        const BLOCK_CNT: usize = 10;
```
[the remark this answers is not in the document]
mversic: ok

core/src/block_sync.rs
mversic: yes, because `blocks_after` compares previous hash with the given hash so it wouldn't match on the latest block and we would log error as if the block's hash is not present in the blockchain

core/src/sumeragi/mod.rs
mversic: you are correct, I missed that there is a method in `wsv`. I'll use that one. Would you like me to return the `GetPeers` message? it's not used anymore

core/src/block.rs
```diff
@@ -674,35 +771,71 @@ impl From<&CommittedBlock> for Vec<Event> {
...
+    #[test]
+    pub fn committed_and_valid_block_hashes_are_equal() {
+        let valid_block = ValidBlock::new_dummy();
+        let committed_block = valid_block.clone().commit();
 
-    use crate::{
-        block::{BlockHeader, EmptyChainHash, ValidBlock},
-        sumeragi::view_change,
-    };
+        assert_eq!(valid_block.hash().transmute(), committed_block.hash())
```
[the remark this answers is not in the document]
mversic: yes

core/src/block.rs
```diff
@@ -464,6 +535,32 @@ impl ValidBlock {
                 .iter()
                 .any(|transaction| transaction.is_in_blockchain(wsv))
     }
+
+    /// Creates dummy `ValidBlock`. Used in tests
+    ///
+    /// # Panics
+    /// Should never panic
+    #[cfg(test)]
+    #[allow(clippy::restriction)]
```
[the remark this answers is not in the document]
mversic: I needed a dummy constructor shared between test modules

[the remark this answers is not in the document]
eadventurous: Ok, haven't noticed it

core/src/sumeragi/mod.rs
eadventurous: Ok, then it is better to remove it I think.

mversic: done

core/src/block.rs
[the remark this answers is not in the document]
eadventurous: Ok, then I guess we will notice it.

eadventurous (approved):

appetrosyan (approved):

Issue #1492, opened 2021-09-30
Transaction status sometimes is not sent by peer


rkharisov:
Precondition: client subscribes to transaction status and as soon as peer responds with `SubscriptionAccepted` event the client sends transaction itself. 

Expected: Event with transaction status (`Committed`, `Rejected`, `Validating`) is received from peer.

Actual: No event about transaction is sent by peer

Note 1. Transaction eventually committed, just no status about it received

Note 2. Transactions are not MST

Note 3. Usually subscription is accepted by one peer, but transaction is sent to another

Note 4. This problem is raised from time to time, not with every transaction

eadventurous:
Issue closed as we could not reproduce the bug. Though we added better description and a more extensive test.

appetrosyan:
The issue seems to persist 

https://github.com/hyperledger/iroha/runs/4369233906?check_suite_focus=true

PR #1631, opened 2021-11-23
Transaction status events tested for each of 7 peers

eadventurous:

### Description of the Change
Transaction status events are tested:
- By connecting for events to each peer
- By submitting transactions in turn to each peer 

### Issue

#1492

### Benefits
Thoroughly testing potential bug cases described in the issue.

### DIFF
--- a/client/tests/tests/events.rs
+++ b/client/tests/tests/events.rs
@@ -6,49 +6,105 @@ use std::{
 };
 
 use eyre::Result;
+use iroha_client::client::Client;
 use iroha_core::config::Configuration;
 use iroha_data_model::prelude::*;
-use test_network::{Peer as TestPeer, *};
+use test_network::*;
+
+const PEER_COUNT: usize = 7;
 
 #[test]
-fn transaction_event_should_be_sent_after_it_is_committed() -> Result<()> {
-    let (_rt, _peer, mut iroha_client) = <TestPeer>::start_test_with_runtime();
-    let pipeline_time = Configuration::pipeline_time();
-
-    // Given
-    thread::sleep(pipeline_time);
-
+fn transaction_event_should_be_sent_to_all_peers_from_all_peers() -> Result<()> {
     let asset_definition_id = AssetDefinitionId::new("xor", "wonderland");
     let create_asset = RegisterBox::new(IdentifiableBox::AssetDefinition(
         AssetDefinition::new_quantity(asset_definition_id).into(),
     ));
-    let committed_event_received = Arc::new(RwLock::new(false));
-    let committed_event_received_clone = committed_event_received.clone();
-    let client_clone = iroha_client.clone();
-    let _handle = thread::spawn(move || {
-        client_clone.loop_on_events(
-            EventFilter::Pipeline(PipelineEventFilter::by_entity(
-                PipelineEntityType::Transaction,
-            )),
-            |event| {
-                if let Ok(Event::Pipeline(event)) = event {
-                    //TODO: check transaction hash
-                    if event.entity_type == PipelineEntityType::Transaction
-                        && event.status == PipelineStatus::Committed
-                    {
-                        *committed_event_received_clone
-                            .write()
-                            .expect("Failed to acquire lock.") = true;
-                    }
-                }
-            },
-        );
-    });
-    thread::sleep(pipeline_time * 2);
-    //When
-    iroha_client.submit(create_asset)?;
-    thread::sleep(pipeline_time * 2);
-    //Then
-    assert!(*committed_event_received.read().unwrap());
+    test_with_instruction_and_status(create_asset, true)?;
+    let fail = FailBox::new("Failing transaction to test Rejected event.");
+    test_with_instruction_and_status(fail, false)
+}
+
+#[allow(clippy::needless_range_loop)]
+fn test_with_instruction_and_status(
+    instruction: impl Into<Instruction>,
+    should_be_committed: bool,
+) -> Result<()> {
+    let instruction: Instruction = instruction.into();
+    for submitting_peer in 0..PEER_COUNT {
+        let (_rt, network, _) =
+            <Network>::start_test_with_runtime(PEER_COUNT.try_into().unwrap(), 1);
+        let pipeline_time = Configuration::pipeline_time();
+
+        // Given
+        thread::sleep(pipeline_time * 4);
+
+        let committed_event_received = Arc::new(RwLock::new([false; PEER_COUNT]));
+        let validating_event_received = Arc::new(RwLock::new([false; PEER_COUNT]));
+        let rejected_event_received = Arc::new(RwLock::new([false; PEER_COUNT]));
+        let peers: Vec<_> = network.peers().collect();
+        let mut submitter_client = Client::test(&peers[submitting_peer].api_address);
+        let transaction = submitter_client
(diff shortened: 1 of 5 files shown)


core/test_network/src/lib.rs
```diff
@@ -57,7 +57,7 @@ pub struct Network<
 {
     /// Genesis peer which sends genesis block to everyone
     pub genesis: Peer<W, G, K, S, B>,
-    /// peers
+    /// Peers excluding the `genesis` peer. Use [`peers()`] function to get all instead.
```
appetrosyan, line 60:
```suggestion
    /// [`Peer`]s excluding the genesis peer. Use [`Network::peers`]  to get all peers.
```

client/tests/tests/events.rs
```diff
@@ -6,49 +6,100 @@ use std::{
...
+                listener_client.for_each_event(
+                    EventFilter::Pipeline(PipelineEventFilter::by_entity(
+                        PipelineEntityType::Transaction,
+                    )),
+                    |event| {
+                        if let Ok(Event::Pipeline(event)) = event {
+                            if event.entity_type == PipelineEntityType::Transaction
+                                && event.hash == *hash
+                            {
+                                if event.status == PipelineStatus::Committed {
```
mversic, line 64: These branches should be mutually exclusive, so I would use `if-else`. Even better I would use match with arms instead of `if-else` because it wouldn't compile if another variant is added to the enum.

core/test_network/src/lib.rs
```diff
@@ -297,7 +297,7 @@ where
     B: BlockSynchronizerTrait<Sumeragi = S, World = W>,
 {
     fn drop(&mut self) {
-        iroha_logger::error!(
+        iroha_logger::warn!(
```
appetrosyan, line 300: Perhaps this is a `debug` message. We only need these data when something goes wrong.

core/test_network/src/lib.rs
```diff
@@ -325,7 +325,7 @@ where
     pub fn stop(&mut self) {
         if let Some(shutdown) = self.shutdown.take() {
             shutdown.abort();
-            iroha_logger::error!("Shutting down peer...");
+            iroha_logger::warn!("Shutting down peer...");
```
appetrosyan, line 328: This could be an `info` message. It's part of normal operation.

docs/source/references/api_spec.md
```diff
@@ -96,6 +96,18 @@ When server is ready to transmit events it sends: `SubscriptionAccepted`
 
 Server sends `Event` and expects `EventReceived` after each, before sending the next event.
 
+**Notes**:
+
+Most usually the client would want to wait for Transaction events.
```
appetrosyan, line 101:
```suggestion
Usually, the client  waits for Transaction events. 
```

docs/source/references/api_spec.md
```diff
@@ -96,6 +96,18 @@ When server is ready to transmit events it sends: `SubscriptionAccepted`
 
 Server sends `Event` and expects `EventReceived` after each, before sending the next event.
 
+**Notes**:
+
+Most usually the client would want to wait for Transaction events. 
+
+Transaction events statuses are: `Validating`, `Committed` or `Rejected`.
```
appetrosyan, line 103:
```suggestion
Transaction event statuses can be either `Validating`, `Committed` or `Rejected`.
```

docs/source/references/api_spec.md
```diff
@@ -96,6 +96,18 @@ When server is ready to transmit events it sends: `SubscriptionAccepted`
...
+
+Transactions status sequence for submitted transaction is: `Validating` -> `Committed` | `Rejected`
```
appetrosyan, line 105:
```suggestion
Transaction statuses proceed from `Validating` to either  `Committed` or `Rejected`. 
```

docs/source/references/api_spec.md
```diff
@@ -96,6 +96,18 @@ When server is ready to transmit events it sends: `SubscriptionAccepted`
...
+Though due to distributed nature of the network some peers might receive events out of order (e.g. first `Committed` then `Validating`).
```
appetrosyan, line 106:
```suggestion
However, due to the distributed nature of the network, some peers might receive events out of order (e.g. `Committed` before `Validating`).
```

docs/source/references/api_spec.md
```diff
@@ -96,6 +96,18 @@ When server is ready to transmit events it sends: `SubscriptionAccepted`
...
+
+There might be cases where some peers in the network might be offline for validation round and if the clients connects to them they might not send `Validating` status.
```
appetrosyan, line 108:
```suggestion
It's possible that some peers in the network are offline for the validation round. If the client connects to them while they are offline, the peers might not respond with the `Validating` status.
```

docs/source/references/api_spec.md
```diff
@@ -96,6 +96,18 @@ When server is ready to transmit events it sends: `SubscriptionAccepted`
...
+But if this peer gets up online it will synchronize the blocks and is guaranteed to send `Committed` or `Rejected` status depending on the information it finds in the block.
```
appetrosyan, line 109:
```suggestion
But when the offline peers come back online they will synchronize the blocks. They are then guaranteed to respond with the `Committed` (or `Rejected`) status depending on the information found in the block.
```

client/tests/tests/events.rs
```diff
@@ -6,49 +6,100 @@ use std::{
...
+}
+
+#[allow(clippy::needless_range_loop)]
+fn test_with_instruction_and_status(
+    instruction: impl Into<Instruction>,
+    should_be_committed: bool,
+) -> Result<()> {
+    let instruction: Instruction = instruction.into();
+    for submitting_peer in 0..PEER_COUNT {
+        let (_rt, network, _) =
```
mversic, line 34: why aren't you using client returned here and create new one instead?

client/tests/tests/events.rs
```diff
@@ -6,49 +6,100 @@ use std::{
...
+            <Network>::start_test_with_runtime(PEER_COUNT.try_into().unwrap(), 1);
+        let pipeline_time = Configuration::pipeline_time();
+
+        // Given
+        thread::sleep(pipeline_time * 4);
+
+        let committed_event_received = Arc::new(RwLock::new([false; PEER_COUNT]));
+        let validating_event_received = Arc::new(RwLock::new([false; PEER_COUNT]));
+        let rejected_event_received = Arc::new(RwLock::new([false; PEER_COUNT]));
```
mversic, line 43: funny thing, these array locations are not accessed concurrently, each thread accesses different index. But that would be a pain to express correctly

mversic (changes requested):

client/tests/tests/events.rs
eadventurous: That client is preconfigured to connect to specific peer. The test needs connection to different peers.

core/test_network/src/lib.rs
eadventurous: I think info message is enough - it is the same as in the next comment

client/tests/tests/events.rs
eadventurous: I think it doesn't matter here that much as it's just a test.

mversic (approved):

appetrosyan (approved):

appetrosyan (approved):

mversic (approved):

PR #1691, opened 2021-12-07
[refactor] #1693: remove Io derive

mversic:

### Description of the Change

`Io` derive was useless. Instead, use explicit `encode`/`decode` to convert to/from `Vec<u8>`.
No code is moved around (only removed) so this shouldn't affect port of bugfixes to release candidate

Also replaces `generics` fn in schema derive with `Generics::split_for_imp`.


s8sato (commented):
Please let it pass the check, then I can approve

data_model/src/query.rs
```diff
@@ -130,35 +126,21 @@ declare_versioned_with_scale!(VersionedSignedQueryRequest 1..2, Debug, Clone, ir
...
-        Response::new(Vec::from(self).into())
-    }
-}
-
-#[cfg(feature = "warp")]
-impl Reply for QueryResult {
-    fn into_response(self) -> Response {
-        (&self).into_response()
-    }
-}
```
eadventurous, line 160: Was it moved somewhere?

eadventurous: Or is it just unused in current torii impl?

mversic: removed because it is unused. I think this relates to the change I did about using `VersionedQueryResult` instead of `QueryResult`

eadventurous: Ok

eadventurous (approved):

s8sato (approved):

PR #1692, opened 2021-12-07
Feature/rdb metrics

iceseer:
### Description of the Change
Adds metrics for RocksDB


irohad/ametsuchi/impl/rocksdb_command_executor.cpp
```diff
@@ -53,6 +55,23 @@ RocksDbCommandExecutor::RocksDbCommandExecutor(
       vm_caller_{vm_caller},
       db_transaction_(db_context_) {
   assert(db_context_);
+
+  getSubscription()->dispatcher()->repeat(
+      SubscriptionEngineHandlers::kMetrics,
+      std::chrono::seconds(5ull),
```
baziorek, line 61: isn't better to have global variable instead of its?

baziorek (approved):
(Un)fortunately I didn't found any problem in the code

irohad/ametsuchi/impl/rocksdb_command_executor.cpp
iceseer: It doesn't use anywhere else. So it's better to add comment instead of global ns pollution.

baziorek: IMO Comment wouldn't help, because 5 seconds is understandable:)

PR #1708, opened 2021-12-09
[fix] #1707: UI tests format

### LINKED ISSUE #1707: Standardize compile tests format across crates
Fix compile tests and standardize their format. `data_model` compile tests aren't being run at all at the moment. There is also a case where `IntoSchema` derive macro doesn't work for generic structs with bounds on generics, e.g. `struct Foo<T: Clone>`
 
When using `trybuild`, proposed hierarchy would look like this:
```
tests/
  ui_fail/
    test_that_fails.rs
    test_that_fails.stderr
  ui_pass/
    test_that_passes.rs
  ui.rs
```

mversic:
### Description of the Change

* uniform format for compile tests
* enable `data_model` derive tests
* fix `IntoSchema` derive and add tests

It's quite easily possible to forget to include ui tests due to misconfiguration. Sticking with this convention would lower the possibility for that. We could also watch for `There are no trybuild tests enabled yet.` line in the CI as this would indicate no ui tests are actually being run even though `trybuild` is configured

### Issue

Closes #1707


macro/derive/tests/ui.rs
```diff
@@ -0,0 +1,8 @@
+use trybuild::TestCases;
+
+#[test]
+fn ui() {
+    let test_cases = TestCases::new();
+    test_cases.pass("tests/ui_pass/*.rs");
```
appetrosyan, line 6: Same here

schema/derive/Cargo.toml
```diff
@@ -9,6 +9,11 @@ publish = false
 proc-macro = true
 
 [dependencies]
-proc-macro2 = "1.0"
+syn = { version = "1.0.72", default-features = false, features = ["full"] }
 quote = "1.0"
-syn = "1.0.72"
+proc-macro2 = "1.0"
```
appetrosyan, line 14: With regards to your comment on the new `CONTRIBUTING.md` should we use 
```suggestion
proc-macro2 = "1.0.32"
```

appetrosyan (approved):

schema/derive/tests/ui_pass/derive_into_schema.rs
```diff
@@ -0,0 +1,29 @@
+use iroha_schema::IntoSchema;
+
+#[derive(IntoSchema)]
+pub struct Foo1<T> {
+    _value1: T,
+}
+
+#[derive(iroha_schema::IntoSchema)]
+pub struct Foo2<T, K> {
```
eadventurous, line 9:
```suggestion
#[derive(IntoSchema)]
pub struct Foo2<T, K> {
```

eadventurous:
> We could also watch for `There are no trybuild tests enabled yet.` line in the CI

Can we set for CI to fail when it detects this line?

schema/derive/tests/ui_pass/derive_into_schema.rs
mversic: yes

appetrosyan (approved):

mversic:
> > We could also watch for `There are no trybuild tests enabled yet.` line in the CI
> 
> Can we set for CI to fail when it detects this line?

I guess we can do that with something like `cargo test | grep <search_string>` or something along this line

> > We could also watch for `There are no trybuild tests enabled yet.` line in the CI
> 
> Can we set for CI to fail when it detects this line?



> > We could also watch for `There are no trybuild tests enabled yet.` line in the CI
> 
> Can we set for CI to fail when it detects this line?

I opened an issue #1712 if we may happen to want to consider this

PR #1721, opened 2021-12-12
[refactor] #1725: Remove global state in Torii

### LINKED ISSUE #1725: Remove Torii global state
As it is implemented at the moment, Torii has global state shared by all filters due to the limitations of the `add_state` method. The code is somewhat convoluted because there is the same set of variables in both `Torii` and `ToriiState`. 

Make changes to `add_state`(implement it as `macro_rules`) to support fine tuning of what filters receive as state

mversic:
### Description of the Change

Have `warp` filters in `Torii` only receive part of the state they require, not everything

### Issue

Closes  #1725

### DIFF
--- a/core/src/torii/mod.rs
+++ b/core/src/torii/mod.rs
@@ -32,14 +32,19 @@ use crate::{
     Addr, Configuration, IrohaNetwork,
 };
 
-/// Main network handler and the only entrypoint of the Iroha.
-pub struct Torii<W: WorldTrait> {
+#[derive(Clone)]
+pub struct ToriiState<W: WorldTrait> {
     iroha_cfg: Configuration,
     wsv: Arc<WorldStateView<W>>,
+    queue: Arc<Queue>,
     events: EventsSender,
     query_validator: Arc<IsQueryAllowedBoxed<W>>,
-    queue: Arc<Queue>,
     network: Addr<IrohaNetwork>,
+}
+
+/// Main network handler and the only entrypoint of the Iroha.
+pub struct Torii<W: WorldTrait> {
+    state: ToriiState<W>,
 }
 
 /// Torii errors.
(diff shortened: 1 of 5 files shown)


s8sato (dismissed):

core/src/torii/utils.rs
```diff
@@ -25,13 +25,13 @@ impl<T: Encode + Send> Reply for Scale<T> {
...
 
 /// Adds state to filter
-pub fn add_state<State: Send + Clone>(
-    state: State,
-) -> impl Filter<Extract = (State,), Error = Rejection> + Clone + Send {
-    warp::any().and_then(move || {
-        let state = state.clone();
-        async move { Ok::<_, Rejection>(state) }
-    })
+macro_rules! add_state {
```
appetrosyan, line 28: Any particular reason to use a declarative macro instead of a templated function?

mversic: I don't think it's possible to create variadic functions in rust

appetrosyan: https://github.com/rust-lang/rust/issues/44930

yeah. It's not been finished yet (much less stabilised).

appetrosyan (approved):

core/src/torii/mod.rs
```diff
@@ -240,92 +223,80 @@ impl<W: WorldTrait> Torii<W> {
...
-            Ok(())
-        }
-        Err(e) => {
-            iroha_logger::error!("Status address configuration parse error");
-            Err(eyre::Error::new(e))
+    /// Start status endpoint.
+    ///
+    /// # Errors
+    /// Can fail due to listening to network or if http server fails
+    async fn start_status(self) -> eyre::Result<()> {
```
eadventurous, line 259: We only need `self.wsv` and `self.network` here, maybe it is best not to move `self` into func? Just take them as arguments.

I think there is a high probability that we might need to start services on other ports.

core/src/torii/utils.rs
eadventurous: Yeah, macros are pretty much the standard way to handle these cases in Rust at the moment.

core/src/torii/mod.rs
mversic: mhm, I thought this was like a chain of calls to start services. One service is started and then forwards `self` to another to start itself and so on

eadventurous: But as the task is spawned, basically we move torii inside this spawned task and consume it, the way it is currently implemented.

s8sato: When it comes to further services on different ports in the future,
I prefer existing(flat) spawns to chained(nested) spawns for readability and modularity

mversic: I'm going to use flat spawns, will commit changes soon

appetrosyan (approved):

eadventurous (approved):

Issue #1751, opened 2021-12-23
Same transaction can be comitted more than 1 time


eadventurous:
Branch: `2.0.0-pre.1rc.1`
Commit: `e4f9863ca82e7cdd9745f5c9219a0e0f66a06e1c`
Configuration: local network started with docker compose with 4 peers

Action: 
Using this script https://github.com/eadventurous/iroha2-longevity-load-rs
A bit modified to send only 1000 tx instead of infinite amount. Other params are the same.

Expected outcome:
Iroha peers report 1001 committed transactions at `status` endpoint. +1 transaction due to genesis transaction.

Got outcome at `status` endpoint:
`{"peers":3,"blocks":893,"txs":1020,"uptime":{"secs":556,"nanos":777000000}}`

eadventurous:
Assumed cause is same transaction is committed several times.

PR #1752, opened 2021-12-23
[fix] #1751: Sumeragi awaits wsv block commit

eadventurous:

### Description of the Change
- Sumeragi awaits wsv block commit
- Update for network topology now directly from sumeragi

### Issue
#1751

### Benefits
Fixes double spend attack, and general bug with same transaction being committed more than once. For more detailed overview please read the issue.

Previously wsv was updated after kura when it received the message from sumeragi. The bug was in the fact that sumeragi assumed it committed the block and updated its state, therefore able to approve next blocks, but the message might arrive to Kura a lot later. This made possible for the `is_in_blockchain` checks for transactions to pass sometimes even as some of them were in previous block. And therefore leaders could create blocks with same txs and other peers might approve it.

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -27,11 +27,7 @@ use tokio::{
 use tokio_stream::wrappers::ReadDirStream;
 
 use crate::{
-    block::VersionedCommittedBlock,
-    block_sync::ContinueSync,
-    prelude::*,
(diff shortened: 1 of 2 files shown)


eadventurous:
I verified that the bug is gone with the same script as pointed in the issue.

mversic (approved):

appetrosyan (approved):

Issue #1425, opened 2021-09-08
Integrate WASM Runtime into Iroha2


e-ivkov:

eadventurous:
See https://wiki.hyperledger.org/display/iroha/Scripting+Languages+and+Runtimes+for+Iroha2+Smart+Contracts

eadventurous:
Prototype - https://github.com/soramitsu/iroha2-wasm-prototype

mversic:
Useful links:
https://www.w3.org/TR/wasm-core-1/#control-instructions%E2%91%A0
https://learnxinyminutes.com/docs/wasm/
https://blog.dbrgn.ch/2019/12/24/testing-for-no-std-compatibility/

PR #1759, opened 2021-12-27
[feature] #1425: add wasm runtime

mversic:

### Description of the Change

This PR adds support for executing wasm smartcontracts to Iroha. 

Function parameters and return values are exchanged through **linear memory** via offset and length of the allocation. Since dealing with linear memory is very low-level(it's just a contiguous array of bytes) Rust types/values must be serialized to binary format before writing them to wasm memory. 

Wasm binary to be able to execute must export:
* `WASM_MAIN_FN` - function which the host calls to execute smartcontract
* `WASM_MEMORY` - linear memory (used to exchange more complex data than integers/floats)
* `WASM_ALLOC_FN` - function which hosts uses to allocate memory from wasm's linear memory

Notes:
* Note that since wasm is sandboxed there is no way for the module to get access to hosts memory.
* Writing wat directly or dealing with raw memory and manual allocation in Rust code that compiles to wasm is not a reasonable user experience. Helper functions and macros are kept for the followup PR. 
* Limits like wasm binary size and wasm fuel are added in a separate PR.

Should feature flag be introduced to include wasm support optionally?

### Issue

#1425

Once [WebAssembly specification](https://github.com/webassembly/interface-types) adds support for `Interface Types` it would be much better to use those because no serialization would be required and types would just flow through wasm VM without VM actually understanding their representation.

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -90,25 +90,28 @@ impl Client {
         }
     }
 
-    /// Builds transaction out of supplied instructions.
+    /// Builds transaction out of supplied instructions or wasm.
     ///
     /// # Errors
     /// Fails if signing transaction fails
     pub fn build_transaction(
         &self,
-        instructions: Vec<Instruction>,
+        instructions: Executable,
         metadata: UnlimitedMetadata,
     ) -> Result<Transaction> {
-        let nonce = self
-            .add_transaction_nonce
-            .then(|| rand::thread_rng().gen::<u32>());
-        Transaction::with_metadata(
-            instructions,
+        let transaction = Transaction::new(
             self.account_id.clone(),
+            instructions,
             self.proposed_transaction_ttl_ms,
-            metadata,
-            nonce,
-        )
+        );
+
+        if self.add_transaction_nonce {
+            let nonce = rand::thread_rng().gen::<u32>();
+            transaction.with_nonce(nonce)
+        } else {
+            transaction
+        }
+        .with_metadata(metadata)
         .sign(&self.key_pair)
     }
 
@@ -171,7 +174,7 @@ impl Client {
         instructions: Vec<Instruction>,
         metadata: UnlimitedMetadata,
     ) -> Result<HashOf<VersionedTransaction>> {
-        self.submit_transaction(self.build_transaction(instructions, metadata)?)
+        self.submit_transaction(self.build_transaction(instructions.into(), metadata)?)
     }
 
     /// Submit a prebuilt transaction.
@@ -264,7 +267,7 @@ impl Client {
         let mut client = self.clone();
         let (event_sender, event_receiver) = mpsc::channel();
         let (init_sender, init_receiver) = mpsc::channel();
-        let transaction = self.build_transaction(instructions, metadata)?;
+        let transaction = self.build_transaction(instructions.into(), metadata)?;
         let hash = transaction.hash();
         let _handle = thread::spawn(move || -> eyre::Result<()> {
             let event_iterator = client
@@ -705,7 +708,7 @@ mod tests {
 
         let build_transaction = || {
             client
-                .build_transaction(vec![], UnlimitedMetadata::new())
+                .build_transaction(Vec::<Instruction>::new().into(), UnlimitedMetadata::new())
                 .unwrap()
         };
         let tx1 = build_transaction();
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -644,9 +644,7 @@ pub mod asset {
     }
 
     /// Get query to get all assets by account id
-    pub fn by_account_id(
-        account_id: impl Into<EvaluatesTo<<Account as Identifiable>::Id>>,
-    ) -> FindAssetsByAccountId {
+    pub fn by_account_id(account_id: impl Into<EvaluatesTo<AccountId>>) -> FindAssetsByAccountId {
         FindAssetsByAccountId::new(account_id)
     }
 
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -15,7 +15,6 @@ pub mod stream;
 pub mod sumeragi;
 pub mod torii;
 pub mod tx;
-mod wasm;
 pub mod wsv;
 
 use std::{path::PathBuf, sync::Arc, time::Duration};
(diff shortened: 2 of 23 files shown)


eadventurous:
> Should feature flag be introduced to include wasm support optionally?

@takemiyamakoto what do you think?

eadventurous:
> linear memory (used to exchange more complex data than integers/floats)

I remember during the prototyping I found that the max limit for linear memory for WASM was 5GB. At the time `wasmtime` did not support limiting it further in an easy way (you needed to write your own allocator for this). Is there a feature to limit it further now?

eadventurous:
Please create issues for the PRs that you plan to create and link them to this one.

core/src/lib.rs
```diff
@@ -15,6 +15,7 @@ pub mod stream;
 pub mod sumeragi;
 pub mod torii;
 pub mod tx;
+mod wasm;
```
eadventurous, line 18: As `wasm` execution is an alternative to `isi` direct execution. We might want to have these modules at the same level. I think we should either remove `smartcontracts` folder from hierarchy, so we would move down the `isi`. Or we should move `wasm` into `smartcontracts`.

core/src/tx.rs
```diff
@@ -195,46 +194,66 @@ impl AcceptedTransaction {
...
-            // Permission validation is skipped for genesis.
-            if !is_genesis {
-                #[cfg(feature = "roles")]
-                {
-                    let instructions = permissions::unpack_if_role_grant(instruction.clone(), wsv)
+        // WSV is cloned here so that instructions don't get applied to the blockchain
+        // Therefore, this instruction execution validates before actually executing
+        let wsv_temp = wsv.clone();
+        match &self.payload.instructions {
+            Executable::Instructions(instructions) => {
```
eadventurous, line 201: What do you think about moving the branch for instructions and the branch for wasm into separate functions?

It will probably make code more readable here.

core/src/wasm.rs
```diff
@@ -0,0 +1,384 @@
...
+
+                    // Accessing memory as a byte slice to avoid the use of unsafe
+                    let isi_mem_range = offset as usize..(offset + len) as usize;
+                    let mut isi_bytes = &memory.data(&caller)[isi_mem_range];
+                    let instruction = Instruction::decode(&mut isi_bytes)
+                        .map_err(|error| Trap::new(error.to_string()))?;
+
+                    instruction
+                        .execute(caller.data().account_id.clone(), caller.data().wsv)
+                        .map_err(|error| Trap::new(error.to_string()))?;
```
eadventurous, line 95: As I understand permissions are not checked here. We should also add these in the future PRs, better to create an issue for it.

core/src/wasm.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    fn create_linker(engine: &Engine) -> Result<Linker<State<'a, W>>, Error> {
+        let mut linker = Linker::new(engine);
+
+        linker
+            .func_wrap(
+                "iroha",
+                EXECUTE_ISI_FN_NAME,
+                |mut caller: Caller<State<_>>, offset: u32, len: u32| {
+                    let memory = Self::get_memory(&mut caller)?;
```
eadventurous, line 85: Maybe let's move these closures into fns? The function `create_linker` is really big.

mversic:
> I remember during the prototyping I found that the max limit for linear memory for WASM was 5GB. At the time `wasmtime` did not support limiting it further in an easy way (you needed to write your own allocator for this). Is there a feature to limit it further now?

@eadventurous Wasm memory is 32-bit addressable so you can't have more than 4GB of linear memory. Linear memory is allocated in 64Kb pages, limits can be defined in number of pages. Wasm module can import the memory from the host, rather than the other way around, and impose the limits on the host side

eadventurous:
> > I remember during the prototyping I found that the max limit for linear memory for WASM was 5GB. At the time `wasmtime` did not support limiting it further in an easy way (you needed to write your own allocator for this). Is there a feature to limit it further now?
> 
> @eadventurous Wasm memory is 32-bit addressable so you can't have more than 4GB of linear memory. Linear memory is allocated in 64Kb pages, limits can be defined in number of pages. Wasm module can import the memory from the host, rather than the other way around, and impose the limits on the host side

You are correct - 4GB. I was referring to this mostly https://github.com/bytecodealliance/wasmtime/issues/2273 but I guess this needs a separate issue.

core/src/lib.rs
mversic: I'll move wasm into `smartcontract`.

core/src/wasm.rs
mversic: no prob, I moved it

core/src/tx.rs
mversic: I've restructured this method a bit to make it more readable

eadventurous (approved):

core/src/wasm.rs
mversic: Opened issue

client/src/client.rs
```diff
@@ -90,25 +90,28 @@ impl Client {
...
 
-    /// Builds transaction out of supplied instructions.
+    /// Builds transaction out of supplied instructions or wasm.
     ///
     /// # Errors
     /// Fails if signing transaction fails
     pub fn build_transaction(
         &self,
-        instructions: Vec<Instruction>,
+        instructions: Executable,
```
appetrosyan, line 99:
```suggestion
        instructions: Execute,
```
Minor point, but -able traits are not very rust-like.

eadventurous (approved):

data_model/src/transaction.rs
```diff
@@ -179,67 +186,42 @@ pub struct Transaction {
...
-            instructions,
-            account_id,
-            proposed_ttl_ms,
-            UnlimitedMetadata::new(),
-            None,
-        )
-    }
-
-    /// [`Transaction`] constructor with nonce.
+    /// Constructs `Transaction`.
```
appetrosyan, line 189:
```suggestion
    /// Construct [`Transaction`].
```

appetrosyan (approved):

client/src/client.rs
eadventurous: The thing is this is not a trait. It is an enum. So `Executable` seems good to me.

PR #1765, opened 2021-12-28
Iroha 1 compile errors when compiling with g++11

baziorek:
I've tried to compile Iroha 1 with my g++ compiler
```
➜  ~ g++ --version
g++ (GCC) 11.1.0
```
And there were few compile errors because of missing headers. But of course those changes are not yer compile errors in earlier version of g++ (e.g. g++-10.*).

### Description of the Change
Added few C++ standard headers

### Benefits
Successful compilation with g++11 and newer.

### Possible Drawbacks
Rather none - there are just additional standard headers


iceseer (approved):

libs/common/common.hpp
```diff
@@ -9,6 +9,7 @@
 #include <chrono>
 #include <mutex>
 #include <shared_mutex>
+#include <condition_variable>
```
iceseer, line 12: use ```clang-format-7``` plz

irohad/pending_txs_storage/impl/pending_txs_storage_impl.cpp
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <mutex>
 #include "pending_txs_storage/impl/pending_txs_storage_impl.hpp"
 
 #include "ametsuchi/tx_presence_cache_utils.hpp"
```
iceseer, line 9:
```c++
#include "pending_txs_storage/impl/pending_txs_storage_impl.hpp"

#include <mutex>

#include "ametsuchi/tx_presence_cache_utils.hpp"

File: libs/common/radix_tree.hpp [hunk 3]
```diff
@@ -10,6 +10,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <optional>
```
iceseer, line 13: use ```clang-format-7``` plz

iceseer (changes requested):

libs/common/common.hpp
baziorek: After using `clang-format` headers got sorted, so I've moved ` #include <condition_variable> ` after `#include <chrono>`.
______
But `clang-format` changed also in different lines:
image
those formatting changes was ignored by me as not connected with the PR:)

irohad/pending_txs_storage/impl/pending_txs_storage_impl.cpp
baziorek: I've moved its between them. Could You explain me please for future what rule is here? In previous comment `clang-format` sorted headers, but here they are not sorted

libs/common/radix_tree.hpp
baziorek: It moved `#include <optional> ` after `#include <memory>`.
___________
There are other changes 13 changes of formatting after running the command -I've ignored its as not connected with the issue.

iceseer (approved):

Issue #1749, opened 2021-12-23
no_std data model


mversic:
1. smart contracts written in `rust` and compiled to `wasm` make use of `iroha_data_model` to get access to Iroha Special Instructions.
2. since these smart contracts are stored on blockchain as `.wasm` binary files, it is quite important their size to be limited

Therefore, it would be logical to have `iroha_data_model` be `no_std` and significantly reduce the size of the `.wasm` binary

Relates to: #1425

PR #1782, opened 2022-01-06
[feature] #1749: no_std support for iroha_crypto

### LINKED ISSUE #1794: Block not signed by leader when committed in single node deployment
Block should be signed before [this](https://github.com/hyperledger/iroha/blob/iroha2-dev/core/src/sumeragi/mod.rs#L861) otherwise it gets committed without leader signature.

mversic:
### Description of the Change

Make `iroha_crypto` `no_std` compatible
* remove `eyre` and implement custom error messages instead

### Issue

closes #1794
relates to #1749

### DIFF
--- a/core/src/sumeragi/mod.rs
+++ b/core/src/sumeragi/mod.rs
@@ -773,7 +773,8 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
             Entry::Occupied(mut occupied) => {
                 let proof_votes = occupied.get_mut();
                 let count = proof_votes.signatures().len();
-                proof_votes.merge_signatures(&proof);
+                proof_votes.merge_signatures(proof.signatures().clone());
+
                 if proof_votes.signatures().len() > count {
                     (true, proof_votes.clone())
                 } else {
@@ -794,7 +795,7 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
         let (count_increased, merged_proof) = self.merge_view_change_votes(proof.clone()).await;
         if count_increased {
             self.broadcast_msg(ViewChangeSuggested::new(
-                proof.clone(),
+                proof,
                 self.view_change_proofs().clone(),
             ))
             .await;
@@ -857,17 +858,20 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
             trace!(?event);
             drop(self.events_sender.send(event));
         }
+        let signed_block = block.sign(self.key_pair.clone())?;
         if !network_topology.is_consensus_required() {
-            self.commit_block(block).await;
+            self.commit_block(signed_block).await;
             return Ok(());
         }
 
-        let voting_block = VotingBlock::new(block.clone());
-        self.voting_block = Some(voting_block.clone());
-        self.broadcast_msg(BlockCreated::from(block.sign(self.key_pair.clone())?))
+        let voting_block = VotingBlock::new(signed_block.clone());
+        let voting_block_hash = voting_block.block.hash();
+
+        self.voting_block = Some(voting_block);
+        self.broadcast_msg(BlockCreated::from(signed_block.clone()))
             .await;
         self.start_commit_countdown(
-            voting_block.clone(),
+            voting_block_hash,
             *self.latest_block_hash(),
             self.latest_view_change_hash(),
             ctx,
@@ -878,16 +882,15 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
 
     /// Starts countdown for a period in which the `voting_block` should be committed.
     #[iroha_futures::telemetry_future]
-    #[log(skip(self, voting_block))]
+    #[log(skip(self, voting_block_hash))]
     #[allow(clippy::expect_used)]
     pub async fn start_commit_countdown(
         &self,
-        voting_block: VotingBlock,
+        voting_block_hash: HashOf<VersionedValidBlock>,
         latest_block: HashOf<VersionedCommittedBlock>,
         latest_view_change: HashOf<Proof>,
         ctx: &mut Context<Self>,
     ) {
-        let voting_block_hash = voting_block.block.hash();
         let proof = view_change::Proof::commit_timeout(
             voting_block_hash,
             latest_view_change,
@@ -1387,10 +1390,12 @@ pub mod message {
                 //TODO: send to set b so they can observe
             }
             let voting_block = VotingBlock::new(self.block.clone());
-            sumeragi.voting_block = Some(voting_block.clone());
+            let voting_block_hash = voting_block.block.hash();
+            sumeragi.voting_block = Some(voting_block);
+
             sumeragi
                 .start_commit_countdown(
-                    voting_block,
+                    voting_block_hash,
                     *sumeragi.latest_block_hash(),
                     sumeragi.latest_view_change_hash(),
                     ctx,
--- a/crypto/src/hash.rs
+++ b/crypto/src/hash.rs
@@ -1,4 +1,6 @@
-use std::{
+#[cfg(not(feature = "std"))]
+use alloc::{format, string::String, vec::Vec};
+use core::{
     fmt::{self, Debug, Display, Formatter},
     hash,
     marker::PhantomData,
@@ -8,6 +10,7 @@ use derive_more::{Deref, DerefMut, Display};
 use iroha_schema::IntoSchema;
 use parity_scale_codec::{Decode, Encode};
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "std")]
 use ursa::blake2::{
     digest::{Update, VariableOutput},
     VarBlake2b,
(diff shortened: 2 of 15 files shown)


core/src/sumeragi/mod.rs
```diff
@@ -857,17 +859,20 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
             info!(?event);
             drop(self.events_sender.send(event));
         }
+        let signed_block = block.sign(self.key_pair.clone())?;
         if !network_topology.is_consensus_required() {
-            self.commit_block(block).await;
```
mversic, line 861: this was a bug previously because block wasn't signed by the leader

crypto/src/signature.rs
```diff
@@ -198,52 +175,53 @@ impl<T> SignatureOf<T> {
...
-impl<T> Clone for SignatureOf<T> {
-    fn clone(&self) -> Self {
-        Self(self.0.clone(), PhantomData)
-    }
-}
-
-impl<T> IntoSchema for SignatureOf<T> {
-    fn type_name() -> String {
-        Signature::type_name()
-    }
```
mversic, line 234: Removed this method because it's inconsistent with `HashOf`.
Previously this serialized to:
```
iroha_schema::iroha_crypto::signature::SignatureOf
```

Now it serializes to (for every `T`):
```
iroha_schema::iroha_crypto::signature::SignatureOf<T>
```

crypto/src/hash.rs
```diff
@@ -69,24 +73,23 @@ impl AsRef<[u8]> for Hash {
...
+#[allow(clippy::default_trait_access)]
+#[derive(Deref, DerefMut, Display, Decode, Encode, Deserialize, Serialize)]
 #[display(fmt = "{}", _0)]
+#[serde(transparent)]
 pub struct HashOf<T>(
     #[deref]
     #[deref_mut]
     Hash,
-    PhantomData<T>,
+    #[codec(skip)] PhantomData<T>,
```
mversic, line 85: `PhantomData` is skipped by default with scale codec, but it doesn't hurt to be explicit

mversic: also, it removes this field from the generated schema

core/src/sumeragi/mod.rs
appetrosyan: Good catch.

mversic: test started failing all of a sudden when I enforced `SignaturesOf` invariant to contain at least one signature

crypto/src/varint.rs
```diff
@@ -7,18 +10,38 @@ pub struct VarUint {
     payload: Vec<u8>,
 }
 
+/// Error which may occur when converting to/from `VarUint`
```
appetrosyan, line 13:
```suggestion
/// Error which  occurs when converting to/from [`VarUint`]
```

crypto/src/varint.rs
```diff
@@ -72,30 +95,35 @@ macro_rules! from_uint(
...
 
         if empty.is_empty() {
-            Ok(Self { payload })
-        } else {
-            Err(eyre!("Last byte shouldn't be followed by anything"))
+            return Ok(Self { payload });
         }
+
+        Err(ConvertError::new(
+            "Last byte shouldn't be followed by anything",
```
appetrosyan, line 115: Could also print what follows the last byte, but I don't insist.

crypto/src/signature.rs
```diff
@@ -82,113 +87,91 @@ impl Signature {
...
-        Signature::encoded_fixed_size()
+impl<T> IntoSchema for SignatureOf<T> {
+    fn schema(metamap: &mut iroha_schema::MetaMap) {
+        Signature::schema(metamap)
     }
 }
 
 impl<T> SignatureOf<T> {
-    /// Verifies signature for this hash
+    /// Creates new [`SignatureOf`] by signing value via [`KeyPair::private_key`]
```
appetrosyan, line 159:
```suggestion
    /// Create new [`SignatureOf`] via signing `T` values by [`KeyPair::private_key`]
```

crypto/src/signature.rs
```diff
@@ -82,113 +87,91 @@ impl Signature {
...
-            })
+    /// Fails if decoding digest of key pair fails
+    #[cfg(feature = "std")]
+    pub fn from_hash(key_pair: KeyPair, hash: &HashOf<T>) -> Result<Self, Error> {
+        Ok(Self(Signature::new(key_pair, hash.as_ref())?, PhantomData))
     }
 
     /// Transmutes signature to some specific type
-    /// SAFETY:
-    /// Do at your own risk
```
appetrosyan, line 186: So this was a false-positive, right?

crypto/src/signature.rs
```diff
@@ -82,113 +87,91 @@ impl Signature {
...
     pub fn transmute<F>(self) -> SignatureOf<F> {
         SignatureOf(self.0, PhantomData)
     }
 
     /// Transmutes signature to some specific type
+    ///
     /// SAFETY:
```
appetrosyan, line 175:
```suggestion
    /// # Safety
```

crypto/src/signature.rs
```diff
@@ -198,52 +181,62 @@ impl<T> SignatureOf<T> {
         }
     }
 
-    /// Creates new [`SignatureOf`] by signing value via [`KeyPair::private_key`].
+    /// Verifies signature for this hash
```
appetrosyan, line 184:
```suggestion
    /// Verify signature for this hash
```

crypto/src/signature.rs
```diff
@@ -198,52 +181,62 @@ impl<T> SignatureOf<T> {
...
+    ///
     /// # Errors
-    /// Fails if decoding digest of key pair fails
-    pub fn from_hash(key_pair: KeyPair, hash: &HashOf<T>) -> Result<Self> {
-        Ok(Self(Signature::new(key_pair, hash.as_ref())?, PhantomData))
+    /// Fails if verification fails
```
appetrosyan, line 187:
```suggestion
    /// Forwards verification of the 0-th tuple variant verification. 
```

crypto/src/signature.rs
```diff
@@ -252,185 +245,200 @@ impl<T> Clone for SignaturesOf<T> {
...
+        let signatures: btree_set::BTreeSet<_> = iter.into_iter().collect();
+        signatures.try_into()
+    }
+}
+
 impl<T> SignaturesOf<T> {
     /// Transmutes signature generic type
-    /// SAFETY: Check complience of hashes of this types
+    ///
+    /// SAFETY: Check complience of hashes of these types
```
appetrosyan, line 318:
```suggestion
    /// # Safety
    /// Check compliance of hashes before use. Can result in X if used improperly.
```

crypto/src/signature.rs
```diff
@@ -252,185 +245,200 @@ impl<T> Clone for SignaturesOf<T> {
...
+    }
 }
 
+#[cfg(feature = "std")]
 impl<T: Encode> SignaturesOf<T> {
-    /// Creates new signatures container.
+    /// Creates new signatures container
+    ///
     /// # Errors
     /// Might fail in signature creation
```
appetrosyan, line 390:
```suggestion
    /// Forwards [`SignatureOf::new`] errors.
```

appetrosyan (approved):
LGTM. Minor cosmetic comments.

crypto/src/varint.rs
mversic: that's a good point, though I'd have to move from `&'static str` to `String`. Will add this

crypto/src/signature.rs
mversic: hmm, well there is no unsafe code here involved so what's the risk?
If type happens to carry some invariants that need to be upheld by the user then this is just another place in the code where user needs to watch out for those invariants. There is nothing special here I think

crypto/src/signature.rs
mversic: This is the only place in our codebase we use `unsafe` as of today. We should be careful here

I would even like to remove the warning from these transmutes. Since I've added `repr(transparent)` the ABI is guaranteed to match for all transmuted types so this code should be safe to compile. The only possible issue is if types need to uphold some invariants but that's not our concern anyways

crypto/src/signature.rs
```diff
@@ -82,113 +87,91 @@ impl Signature {
...
-    /// SAFETY:
-    /// Do at your own risk
     pub fn transmute<F>(self) -> SignatureOf<F> {
         SignatureOf(self.0, PhantomData)
     }
 
     /// Transmutes signature to some specific type
+    ///
     /// SAFETY:
     /// Do at your own risk
```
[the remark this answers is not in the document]
mversic: I can't think of any improper use as stated in the previous comment

appetrosyan (approved):

crypto/src/signature.rs
appetrosyan: Sounds good.

crypto/src/signature.rs
appetrosyan: Maybe just replace it with 
```
/// # Warning: unsafe
```
And be done with it?

crypto/src/signature.rs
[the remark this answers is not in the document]
appetrosyan: So same suggestion as previous comment?

appetrosyan (approved):

s8sato (dismissed):

appetrosyan (approved):

s8sato (approved):

Issue #1742, opened 2021-12-20
Concretise Errors in `iroha_core`


appetrosyan:
In Iroha core, many error types were implicitly converted into `eyre::Report`s which replaced transaction rejection reason with `Some other error happened`.

PR #1816, opened 2022-01-14
[feature] #1425: Wasm helper crate

### LINKED ISSUE #1425: Integrate WASM Runtime into Iroha2

mversic:
### Description of the Change

Smartcontract example:
``` rs
#![no_std]
#![no_main]

use iroha_wasm::{data_model::prelude::*, prelude::*};

#[iroha_wasm]
fn _entry_point(_account_id: AccountId) {
    let new_account_id = AccountId::test("mad_hatter", "wonderland");
    let register_isi = RegisterBox::new(NewAccount::new(new_account_id));

    Instruction::Register(register_isi).execute();
}
```

The pattern I used for testing to inject mocked objects is something I saw in the [mockall](https://asomers.github.io/mock_shootout/) crate. I like it because it doesn't infect the external API unlike the pattern we use with traits. I previously used this [crate](https://docs.rs/mockall/latest/mockall/) successfully, although I think I encountered some bugs but didn't investigate into it.

# Problems
* including `iroha_wasm` in the workspace seems to affect `cfg(test)` resolution
* `iroha_wasm` it has to be build with nightly compiler

### Issue

Closes #1425

### Benefits

Internal are hidden from the user. All user has to do is annotate one of their methods with `iroha_wasm`


wasm/derive/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+        block,
+    }: ItemFn = parse_macro_input!(item as ItemFn);
+
+    verify_function_signature(&sig);
+    let fn_name = &sig.ident;
+
+    quote! {
+        #[no_mangle]
+        unsafe extern "C" fn _iroha_wasm_main(ptr: u32, len: u32) {
+            #fn_name(iroha_wasm::_decode_from_raw::<AccountId>(ptr, len))
```
mversic, line 27: should the function `#fn_name` receive `account_id` as a parameter?

wasm/derive/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+        sig,
+        block,
+    }: ItemFn = parse_macro_input!(item as ItemFn);
+
+    verify_function_signature(&sig);
+    let fn_name = &sig.ident;
+
+    quote! {
+        #[no_mangle]
+        unsafe extern "C" fn _iroha_wasm_main(ptr: u32, len: u32) {
```
mversic, line 26: it is UB for two crates to export an item with the same name. When using `no_mangle` compiler cannot guarantee that this overlap will not occur. This usually means that it's not specified which of the items will be linked to. This is the reason why I prepended `_iroha_wasm` to all `no_mangle` function. `no_mangle` is a bit dangerous

https://github.com/rust-lang/rust/issues/28179

wasm/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+pub use iroha_wasm_derive::iroha_wasm;
+use parity_scale_codec::{Decode, Encode};
+
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[cfg(target_pointer_width = "32")]
+type WasmUsize = u32;
+#[cfg(target_pointer_width = "64")]
+type WasmUsize = u64;
```
mversic, line 21: Maybe only switch based on `cfg(test)`?

core/src/smartcontracts/wasm.rs
```diff
@@ -125,10 +132,19 @@ impl<'a, W: WorldTrait> Runtime<'a, W> {
...
+    /// # Warning
+    ///
+    /// This function doesn't take ownership of the provided allocation
+    /// but it does tranasfer ownership of the result to the caller
+    ///
     /// # Errors
     ///
     /// If decoding or execution of the ISI fails
-    fn execute_isi(mut caller: Caller<State<W>>, offset: u32, len: u32) -> Result<(), Trap> {
+    fn execute_instruction(
```
mversic, line 143: we can have only one generic execute method. Would that be preferred?

wasm/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    // Safety: - `host_execute_query` doesn't take ownership of it's pointer parameter
+    //         - ownership of the returned result is transfered into `_decode_from_raw`
+    unsafe {
+        let host::WasmQueryResult(res_ptr, res_len) = encode_and_execute(query, host_execute_query);
+        _decode_from_raw(res_ptr, res_len)
+    }
+}
+
+/// Execute the given instruction on the host environment
+pub fn execute_instruction(instruction: Instruction) {
```
mversic, line 147: we can have only one generic execute method instead of `execute_instruction` and `execute_query`. Would that be preferred?

wasm/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+    #[allow(clippy::cast_possible_truncation)]
+    let ptr = bytes.as_ptr() as u32;
+    #[allow(clippy::cast_possible_truncation)]
+    let len = bytes.len() as u32;
+
+    fun(ptr, len)
+}
+
+/// Executes the given query on the host environment
+pub fn execute_query(query: QueryBox) -> QueryResult {
```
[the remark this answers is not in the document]
mversic: the same goes for `execute_instruction`

wasm/src/lib.rs
appetrosyan: No. I think that this is precisely what it needs to be. Anything else, e.g. `cfg(test)` relies on some assumptions that may not be true in the future.

wasm/src/lib.rs
appetrosyan: I would. We already have `execute` for both impls of `ValidQuery` and `Execute`. While having separate `execute_instruction` will sometimes be clearer, I don't see why the same clarity can't be achieved using `Execute::execute(instruction)`. At the same time, in most cases it's obvious whether or not what you're dealing with is an instruction or a query.

wasm/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: I'm not sure, in that it *could* cause a deadlock, maybe bring this up during status with @e-ivkov ?

wasm/derive/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+//! Exposes macros which facilitate writing smartcontracts
```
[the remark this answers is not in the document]
mversic: done

wasm/src/lib.rs
```diff
@@ -0,0 +1,211 @@
+#![feature(alloc_error_handler)]
```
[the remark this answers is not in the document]
mversic: yes, added documentation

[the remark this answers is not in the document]
mversic: I think because this crate isn't part of the workspace some CI jobs didn't run for it. Still have to see what to do about that

core/src/smartcontracts/wasm.rs
```diff
@@ -125,10 +132,19 @@ impl<'a, W: WorldTrait> Runtime<'a, W> {
     /// Host defined function which executes ISI. When calling this function, module
     /// serializes ISI to linear memory and provides offset and length as parameters
     ///
+    /// # Warning
+    ///
+    /// This function doesn't take ownership of the provided allocation
```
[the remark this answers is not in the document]
mversic: the return type of this function is just `()`. `Result<(), Trap>` is just `wasmtime`'s sugar on top of it. For `()` it doesn't matter that ownership is transferred to the caller. I left this comment in case we return something else here

But you are correct, `#[must_use]` is very adequate both here and for `execute_query`. Nice catch

wasm/src/lib.rs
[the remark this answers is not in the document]
mversic: interesting idea but I don't understand how this could deadlock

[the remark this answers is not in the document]
appetrosyan: Suppose you only check things that can fail immediately? It can't. The problem arises in, can you guarantee that all checks that fail, do so immediately (which could stall Iroha) or that they don't depend on other instructions in the pipeline? Or that the other instruction in the pipeline doesn't use this as a trigger. 

We could get around this issue, by allowing only a specific kind of error to be returned: one that people aren't tempted to use everywhere, but for specific failure conditions. Kinda like in #1742 ?

appetrosyan (approved):

wasm/src/lib.rs
mversic: no, no, this is **target**`_pointer_width`, not host. It's the pointer size for the wasm target which is, at this moment still, 32-bit (`wasm32-unknown-unknown`). It is possible that wasm will support 64-bit targets in the future as the specification progresses.

Primary reason for this was because of tests which, if compiled for the entire workspace, are compiled for 64-bit architecture and start failing if 32-bit target architecture is expected. Of course tests can be compiled for 32-bit architecture for this crate specifically but that adds up some friction for the dev.

Also I think I'm going to have to add another CI job for building this crate, I don't think it can be part of the workspace because it introduces it's own panic handler which results in some conflict

appetrosyan: I think I may have misunderstood your original comment. 

I agree that a separate CI job is good. Actually a separate workflow, since everything related to WASM is in a few localised paths.

I'm not sure that your assertion about this being only useful for tests is going to hold true in the future. However, for now adding `#[cfg(test)]` is fine, and I doubt that 64-bit pointers are going to make it into WASM any time soon. I don't know if adding an extra layer of complexity via `#[cfg(test)]` is worth the few microseconds of debug build time, but it's up to you.

e-ivkov (approved):

wasm/src/lib.rs
mversic: changed, check the latest commit

.github/workflows/iroha2.yml
```diff
@@ -51,6 +51,23 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2021-02-20
+          target: wasm32-unknown-unknown
+          override: true
+      - name: Test
+        run: cargo test --tests --release --verbose -- --test-threads=1
```
appetrosyan, line 67: 1) Shouldn't this test be in the PR workflow. If this fails, we can't see that it did indeed fail. 
2) We do build the wasm crate, but it isn't saved as an artifact anywhere. Why are we building it then? 
3) This invocation is at odds with all other invocations: some comments (in the yaml file) would be useful:
	3.1) Why is it `cargo test --tests`
	3.2) Does it break with `mold -run cargo test --tests`, and if so, wouldn't it make sense to open an issue for `mold`?
	3.3) Why is this test run in `--release` mode. 
	3.4) Why is this test run in `--verbose` mode, most other tests are run with `--quiet` so that the logs are shorter. 
	3.5) Why `--test-threads=1`. Is there an issue we should be fixing?

.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -23,14 +26,10 @@ jobs:
         run: cargo lints clippy --workspace --benches --tests --examples --quiet --no-default-features
         if: always()
       - name: Static analysis with all features enabled
-        run: cargo lints clippy --workspace --benches --tests --examples --all-features --quiet
-        if: always()
-      - name: Verify iroha_data_model supports no_std
-        run: cargo nono check --package iroha_data_model --no-default-features
```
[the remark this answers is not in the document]
mversic: I moved it to `iroha_wasm` pipeline. We actually need `iroha_wasm` be `no_std` which will check `iroha_data_model` transitively

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1
+      - name: Set nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2021-02-20
+          target: wasm32-unknown-unknown
+          default: true
+      - name: Build for wasm32 target
+        run: mold -run cargo +nightly build --quiet --target wasm32-unknown-unknown
```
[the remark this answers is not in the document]
mversic: it's a lib, what's there to artifact? This was just a form of sanity check. I could remove it?

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -0,0 +1,75 @@
...
+  pull_request:
+    branches: [iroha2-dev]
+    paths:
+      - "wasm/**.rs"
+      - "wasm/**.json"
+      - "wasm/**.toml"
+      - "wasm/**.yml"
+
+jobs:
+  static-analysis:
```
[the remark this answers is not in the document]
mversic: ok, I'll move it to static analysis workflow

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1
+      - name: Set nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2021-02-20
+          target: i686-unknown-linux-gnu
+          default: true
+      - name: Run tests
+        run: mold -run +nightly cargo test --tests --quiet --no-fail-fast -- --skip unstable_network --test-threads 2
```
appetrosyan, line 75:
```suggestion
        run: mold -run cargo +nightly test --tests --quiet --no-fail-fast -- --skip unstable_network --test-threads 2
```

appetrosyan (changes requested):

wasm/derive/src/lib.rs
s8sato: Do you think `#[wasm_bindgen]` instead of `#[no_mangle]` bring something for this?
I [confirmed](https://github.com/s8sato/mangle/blob/main/src/main.rs) it allows to export items of the same name in the different modules:

FYI:

* https://docs.wasmtime.dev/wasm-rust.html?highlight=bindgen#exporting-rust-functionality

  > Rust functions must be #[no_mangle] to show up in the final binary, but if you're using #[wasm_bindgen] that will happen automatically for you.

  > Note that wasm-bindgen was originally developed for JS and usage in a browser, but a subset of its implementation (such as arguments which are strings) are supported for WebAssembly interface types.

* https://rustwasm.github.io/wasm-bindgen/contributing/design/describe.html?highlight=no_mangle#communicating-types-to-wasm-bindgen

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: [self-hosted, Linux]
+    container:
+      image: 7272721/i2-ci:latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1
+      - name: Set nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2021-02-20
```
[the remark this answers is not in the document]
mversic: no specific reason, I'm still figuring out CI. But nightly is required for sure

wasm/src/lib.rs
mversic: With the `compile_error`s added, I think this discussion can be closed

.github/workflows/iroha2-dev-pr-wasm.yaml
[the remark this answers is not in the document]
appetrosyan: FYI you can use unstable features with stable using `RUSTC_BOOTSTRAP=1`.  I was planning on removing the nightly toolchain to slim down the CI image.

wasm/derive/src/lib.rs
s8sato: No. `#[wasm_bindgen]` requires `cargo wasi build --release` and this does not allows that

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -0,0 +1,75 @@
...
+          target: i686-unknown-linux-gnu
```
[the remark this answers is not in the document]
mversic: let's test for `wasm32-unknown-unknown`

.github/workflows/iroha2.yml
mversic: I just literally copied this line from the job below. Would you like me to make these modifications for that  existing job?

.github/workflows/iroha2-dev-pr-wasm.yaml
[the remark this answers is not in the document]
mversic: ok, let's do that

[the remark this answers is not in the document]
mversic: except that `fmt` doesn't work then, only with nightly

.github/workflows/iroha2-dev-pr-wasm.yaml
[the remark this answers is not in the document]
mversic: since I've added testing against `wasm32-unknown-unknown` target, I'll remove this

wasm/derive/src/lib.rs
mversic: can we close this? and I don't think this would solve the issue. They would just attach `no_mangle` to the function but the linking problem would persist.

.github/workflows/iroha2-dev-pr-wasm.yaml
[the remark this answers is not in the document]
appetrosyan: On my machine with `rust-1.57` it does work, but ignores the `imports-granularity` directives. We should revisit this when we upgrade to `1.59`.

.gitignore
```diff
@@ -2,6 +2,7 @@
 **/blocks/
 **/*.rs.bk
 **/rusty-tags.vi
+/**/Cargo.lock
```
appetrosyan, line 5: That could break `cargo chef` in the `iroha2-dev` docker file. Are you sure you want to exclude it?

appetrosyan (approved):

.github/workflows/iroha2.yml
appetrosyan: We should, but I don't want to over-encumber this PR. Let's merge as is, and optimise later.

s8sato (approved):

PR #1899, opened 2022-02-10
[feature] #1195: Close a websocket connection cleanly

### LINKED ISSUE #1195: Closing a websocket connection cleanly
It is currently impossible to close the websocket connection normally. It should be done according to this spec, which is the version that browsers with current support implement:

    To close the connection cleanly, a frame consisting of just a 0xFF
    byte followed by a 0x00 byte is sent from one peer to ask that the
    other peer close the connection.

For now, the js client will force connections to terminate.

Arjentix:
### Description of the Change

* Now `torii` checks for WebSocket `Close` messages on `/events` path and gracefully handles them.
* Also added test for it.
* As a bonus I refactored some error types from `eyre::Report` to custom Error enums. I thought, that it will help me, but it doesn't. I can reset it, if everybody else think it wasn't necessary

### Issue

Resolves #1195 

I was working only with that `handle_subscription()` method, cause it's mentioned in the issue. I don't know if issue appears somewhere else

### Benefits

Closing WebSocket requires waiting for response. `torii` wasn't doing it, so clients had to interrupt connection.
Now this works as it shoud be.

### Possible Drawbacks

Any suggestions about additional tests will be helpful. May be it should be tested with js-client, as it was mentioned in #1195

See `torii/tests.rs`

### DIFF
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -14,18 +14,23 @@ pub type EventsSender = broadcast::Sender<Event>;
 /// Type of `Receiver<Event>` which should be used for channels of `Event` messages.
 pub type EventsReceiver = broadcast::Receiver<Event>;
 
+/// Type of error for `Consumer`
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
+    /// Error from provided stream/websocket
     #[error("Stream error: {0}")]
     Stream(#[from] stream::Error<<WebSocket as Stream<VersionedEventSubscriberMessage>>::Err>),
 
+    /// Error from converting received message to filter
     #[error("Can't retrieve subscription filter: {0}")]
     CantRetrieveSubscriptionFilter(#[from] ErrorTryFromEnum<EventSubscriberMessage, EventFilter>),
 
+    /// Error, that occurs when client answered not with `EventReceived` message
     #[error("Got unexpected response. Expected `EventReceived`")]
     ExpectedEventReceived,
 }
 
+/// Result type for `Consumer`
 pub type Result<T> = core::result::Result<T, Error>;
 
 /// Consumer for Iroha `Event`(s).
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -78,4 +78,11 @@ impl Consumer {
             Err(Error::ExpectedEventReceived)
         }
     }
+
+    /// Returns mut reference to stored `stream`
+    ///
+    /// Useful for performing other read/write stuff with `stream`
+    pub fn stream_mut(&mut self) -> &mut WebSocket {
+        &mut self.stream
+    }
 }
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -2,6 +2,7 @@
 //! This module contains descriptions of such an events and
 //! utility Iroha Special Instructions to work with them.
 
+use futures::TryStreamExt;
 use iroha_data_model::events::prelude::*;
 use iroha_macro::error::ErrorTryFromEnum;
 use tokio::sync::broadcast;
@@ -13,19 +14,34 @@ use crate::stream::{self, Sink, Stream};
 pub type EventsSender = broadcast::Sender<Event>;
 /// Type of `Receiver<Event>` which should be used for channels of `Event` messages.
 pub type EventsReceiver = broadcast::Receiver<Event>;
+/// Type of Stream error
+pub type StreamError = stream::Error<<WebSocket as Stream<VersionedEventSubscriberMessage>>::Err>;
 
 /// Type of error for `Consumer`
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     /// Error from provided stream/websocket
     #[error("Stream error: {0}")]
-    Stream(#[from] stream::Error<<WebSocket as Stream<VersionedEventSubscriberMessage>>::Err>),
+    Stream(Box<StreamError>),
     /// Error from converting received message to filter
     #[error("Can't retrieve subscription filter: {0}")]
     CantRetrieveSubscriptionFilter(#[from] ErrorTryFromEnum<EventSubscriberMessage, EventFilter>),
     /// Error, that occurs when client answered not with `EventReceived` message
     #[error("Got unexpected response. Expected `EventReceived`")]
     ExpectedEventReceived,
+    /// Error from provided websocket
+    #[error("WebSocket error: {0}")]
+    WebSocket(#[from] warp::Error),
+
+    /// Error that occurs than `WebSocket::next()` call returns `None`
+    #[error("Can't receive message from stream")]
+    CantReceiveMessage,
+}
+
+impl From<StreamError> for Error {
+    fn from(error: StreamError) -> Self {
+        Self::Stream(Box::new(error))
+    }
 }
 
 /// Result type for `Consumer`
@@ -82,10 +98,17 @@ impl Consumer {
         }
     }
 
-    /// Returns mut reference to stored `stream`
+    /// Listen for `Close` message in loop
     ///
-    /// Useful for performing other read/write stuff with `stream`
-    pub fn stream_mut(&mut self) -> &mut WebSocket {
-        &mut self.stream
+    /// # Errors
+    /// Can fail if can't receive message from stream for some reason
+    pub async fn stream_closed(&mut self) -> Result<()> {
+        while let Some(message) = self.stream.try_next().await? {
+            if message.is_close() {
+                return Ok(());
+            }
+            iroha_logger::trace!("Unexpected message received: {:?}", message);
+        }
+        Err(Error::CantReceiveMessage)
     }
 }
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -1,18 +1,32 @@
 //! Iroha is a quite dynamic system so many events can happen.
 //! This module contains descriptions of such an events and
-//! utilitary Iroha Special Instructions to work with them.
+//! utilitarian Iroha Special Instructions to work with them.
 
-use eyre::{eyre, Result};
 use iroha_data_model::events::prelude::*;
+use iroha_macro::error::ErrorTryFromEnum;
 use tokio::sync::broadcast;
 use warp::ws::WebSocket;
 
-use crate::stream::{Sink, Stream};
+use crate::stream::{self, Sink, Stream};
 
 /// Type of `Sender<Event>` which should be used for channels of `Event` messages.
 pub type EventsSender = broadcast::Sender<Event>;
 /// Type of `Receiver<Event>` which should be used for channels of `Event` messages.
 pub type EventsReceiver = broadcast::Receiver<Event>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Stream error: {0}")]
+    Stream(#[from] stream::Error<<WebSocket as Stream<VersionedEventSubscriberMessage>>::Err>),
+
+    #[error("Can't retrieve subscription filter: {0}")]
+    CantRetrieveSubscriptionFilter(#[from] ErrorTryFromEnum<EventSubscriberMessage, EventFilter>),
+
+    #[error("Got unexpected response. Expected `EventReceived`")]
+    ExpectedEventReceived,
+}
+
+pub type Result<T> = core::result::Result<T, Error>;
 
 /// Consumer for Iroha `Event`(s).
 /// Passes the events over the corresponding connection `stream` if they match the `filter`.
@@ -61,7 +75,7 @@ impl Consumer {
         if let EventSubscriberMessage::EventReceived = message.into_v1() {
             Ok(())
         } else {
-            Err(eyre!("Expected `EventReceived`."))
+            Err(Error::ExpectedEventReceived)
         }
     }
 }
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -1,6 +1,6 @@
 //! Iroha is a quite dynamic system so many events can happen.
 //! This module contains descriptions of such an events and
-//! utilitarian Iroha Special Instructions to work with them.
+//! utility Iroha Special Instructions to work with them.
 
 use iroha_data_model::events::prelude::*;
 use iroha_macro::error::ErrorTryFromEnum;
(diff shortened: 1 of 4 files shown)


core/src/event.rs
```diff
@@ -61,7 +80,14 @@ impl Consumer {
         if let EventSubscriberMessage::EventReceived = message.into_v1() {
             Ok(())
         } else {
-            Err(eyre!("Expected `EventReceived`."))
+            Err(Error::ExpectedEventReceived)
         }
     }
```
mversic, line 85: I assume that when socket is closed on the client side an error return value will be generated in this function. Couldn't you catch that error here and close the stream?

mversic (changes requested):

core/src/event.rs
Arjentix: Maybe I don't clearly understand what you are meaning, but I think it's not this place.

Do you suggest to remove `tokio::select!` call from `handle_subscription()` and make processing here? I think that would be incorrect.

In this function `Consumer` just waits client confirmation about successfully received event.
Does client need to wait for event to just say to server that it no longer needs new events and want to close conneciton? I think no. Client should be able to close connection at any time. So wee need to listen for new events from `events` and for new messages from client `stream` at the same time. That's why I used `tokio::select!`.

If you are talking about incorrect client behaviour when it does not send confirmation about received event and just closes its socket... When it should be processed normally:

1. `self.stream.recv().await?` will fail
2. `consumer.consume(evet).await?` also will fail
3. `handle_subscription()` will be ended with report about closed connection
4. API router will log it

core/src/torii/routing.rs
```diff
@@ -167,10 +167,21 @@ async fn handle_subscription(events: EventsSender, stream: WebSocket) -> eyre::R
     let mut consumer = Consumer::new(stream).await?;
 
     loop {
-        let event = events.recv().await?;
-
-        iroha_logger::trace!(?event);
-        consumer.consume(event).await?;
+        tokio::select! {
+            message_opt = consumer.stream_mut().next() => {
```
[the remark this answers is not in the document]
Arjentix: Yes, it consumes new message from stream.

Berfore that PR server was listening only for `EventReceived` messages from client and only if an event has occurred. If client was sending something berfore an event happened, server wasn't listening for that. So I assume that by design only one client behaviour was correct -- to send `EventReceived` when a new event was received.

Now I had to implement possibility to close connection cleanly. That's a new type of message (may be it should be presented on iroha message structures?), that can be created with `ws::Message::close()`. That message can be sent at any time during subscription, so I'm listening for that type of mesages here. There is no other types of messages right now.

Maybe it will be good to log if it's not `Close` message was received?

core/src/event.rs
```diff
@@ -1,19 +1,38 @@
 //! Iroha is a quite dynamic system so many events can happen.
 //! This module contains descriptions of such an events and
-//! utilitary Iroha Special Instructions to work with them.
+//! utilitarian Iroha Special Instructions to work with them.
```
appetrosyan, line 3:
```suggestion
//! utility Iroha Special Instructions to work with them.
```

core/src/event.rs
```diff
@@ -1,19 +1,38 @@
...
 /// Type of `Receiver<Event>` which should be used for channels of `Event` messages.
 pub type EventsReceiver = broadcast::Receiver<Event>;
 
+/// Type of error for `Consumer`
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// Error from provided stream/websocket
+    #[error("Stream error: {0}")]
+    Stream(#[from] stream::Error<<WebSocket as Stream<VersionedEventSubscriberMessage>>::Err>),
+
```
appetrosyan, line 23: Minor point. Our convention is to have no blank spaces between error variants. 

But this is a suggestion, not a requirement.

core/src/event.rs
appetrosyan: What Marin means here, is that while the result of the evaluation has an `Err` variant, it's not necessarily something we should propagate up the call chain. Closing a websocket connection is a normal part of the operation of Iroha, and it should probably be processed in-place.

core/src/stream.rs
```diff
@@ -13,14 +13,59 @@ const TIMEOUT: Duration = Duration::from_millis(10_000);
...
 const TIMEOUT: Duration = Duration::from_millis(1000);
 
-/// Represents messsage used by the stream
+/// Error type with generic for actual Stream/Sink error type
+#[derive(thiserror::Error, Debug)]
+pub enum Error<InternalStreamError>
+where
+    InternalStreamError: std::error::Error + Send + Sync + 'static,
+{
+    /// Error, that occurs when `recv()` timeout exceeds
```
appetrosyan, line 22:
```suggestion
    /// `recv()` timeout exceeded
```
It's a lot shorter, and most of our variant docs must fit on a single line for the doc generator to work.

core/src/stream.rs
```diff
@@ -13,14 +13,59 @@ const TIMEOUT: Duration = Duration::from_millis(10_000);
...
+    #[error("No message")]
+    NoMessage,
+
+    /// Error in internal stream representation (typically WebSocket)
+    ///
+    /// Made without `from` macro because it will break `IrohaVersion` variant conversion
+    #[error("Internal stream error: {err}")]
+    InternalStream {
+        /// Internal stream error representation
+        err: InternalStreamError,
```
appetrosyan, line 40:
```suggestion
       InternalStreamError,
```

appetrosyan: Use a tuple-enum here. The structure doesn't do anything, and it's not convention to have both struct and tuple enum variants in rust.

core/src/stream.rs
```diff
@@ -13,14 +13,59 @@ const TIMEOUT: Duration = Duration::from_millis(10_000);
...
+    NonBinaryMessage,
+
+    /// Error message during versioned message decoding
+    #[error("Iroha version error: {0}")]
+    IrohaVersion(#[from] iroha_version::error::Error),
+}
+
+/// Represents message used by the stream
 pub trait StreamMessage {
     /// Constructs new binary message
```
appetrosyan, line 58:
```suggestion
    /// Construct new binary message
```

core/src/event.rs
Arjentix: Then how should `handle_subscription()` stop working if it doesn't know about unexpected socket closing?

appetrosyan: Well, a good hint is that sometimes the closing of the socket is expected, and that is not an error. If the socket closure happens abruptly, it's a different case.

Arjentix: Okay, now I understand what you are talking about.

I didn't change part with client response receiving:

```rust
let message: VersionedEventSubscriberMessage = self.stream.recv().await?;
```

So, do you think I should process all parts like that for unexpected socket closing? That also can happen in `Consumer::new()`. If yes, then how can I tell calling function, that `stream` is no longer available? Returning `Result<bool>`? Kinda wierd for that purpose.
WebSocket *closing handshake* is not just socket closing, so I think it's an erroneous client behaviour. It's like unexpected connection failure

core/src/torii/routing.rs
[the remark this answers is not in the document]
s8sato: I think this code is all right.
The 1st branch is listening to external stream message.
The 2nd branch is listening to internal events

s8sato (approved):

core/src/torii/routing.rs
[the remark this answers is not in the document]
mversic: ok, I'm just pointing out what I think might be an issue. I'm not always correct. If you think all is good I trust you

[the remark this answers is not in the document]
mversic: but what happens if message other than close message is received here? it will be swept under the carpet. Isn't there some peek operation on the stream?

core/src/torii/tests.rs
```diff
@@ -533,3 +533,45 @@ async fn blocks_stream() {
...
+        .unwrap();
+
+    // Subscribing
+    let event_filter = EventFilter::Pipeline(pipeline::EventFilter::by_entity(
+        pipeline::EntityType::Block,
+    ));
+    let subscribe_message = VersionedEventSubscriberMessage::from(
+        EventSubscriberMessage::SubscriptionRequest(event_filter),
+    );
+    Sink::send(&mut client, subscribe_message).await.unwrap();
```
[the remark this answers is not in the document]
Arjentix: Can't do it, cause `WsClient` type has its own `send()` implementation for `ws::Message`. So I have to use fully quilified call syntax

core/src/torii/tests.rs
```diff
@@ -533,3 +533,45 @@ async fn blocks_stream() {
...
+
+    let confirmation_response: VersionedEventPublisherMessage =
+        Stream::recv(&mut client).await.unwrap();
```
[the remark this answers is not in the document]
Arjentix: Same thing as with `send()`. See my comment above

core/src/torii/routing.rs
```diff
@@ -167,10 +167,21 @@ async fn handle_subscription(events: EventsSender, stream: WebSocket) -> eyre::R
...
+                if let Some(message) = message_opt {
+                    if message?.is_close() {
+                        return Ok(());
+                    }
+                }
+            }
+            event = events.recv() => {
+                let event = event?;
+                iroha_logger::trace!(?event);
+                consumer.consume(event).await?;
```
mversic, line 181: since the branches are run concurrently, it's still possible that the code will return closed stream error here because it polls the underlying stream

mversic (changes requested):

core/src/torii/routing.rs
mversic: if this is correct, the same is true for `Consumer::new`

core/src/torii/routing.rs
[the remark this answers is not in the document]
Arjentix: `warp` WebSocket does not support peeking. So if we are gonna implement it ourselves on `Stream` trait, then what will be the difference if we still has to use the `next()` method?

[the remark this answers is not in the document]
mversic: What I meant here is first peek then take out of stream, but since you have concurrency issues as commented below this will not help. After you resolve concurrency issue this may become a non issue or we'll have to figure out a way how to first peek and then take out of stream.

core/src/event.rs
Arjentix: I've changed code to handle `Close` message in one place. So I'll mark this conversation as resolved and you can leave some comments for new code version

core/src/torii/routing.rs
```diff
@@ -161,16 +161,84 @@ async fn stream_blocks<W: WorldTrait>(
...
+                // This branch catches `Close` ans unexpected messages
+                message_opt = consumer.stream_mut().next() => {
+                    if let Some(message) = message_opt {
+                        let message = message?;
+                        if message.is_close() {
+                            return Err(Error::CloseMessage);
+                        }
+                        iroha_logger::trace!("Unexpected message received: {:?}", message);
+                    }
+                }
```
mversic, line 231: this is getting better, but you don't need this `select!` mechanics at all if you're relying on the returned error value. Just let any of the calls to consumer bubble up the error to where you can handle it

Arjentix: I think, that won't work for case when no events are received (or passed filter) but client want to close connection. In that case client has to wait until first `recv()` on server side that will happen only when server get new event, send it and call `recv` excpecting`EventReceived` message

mversic: if client wants to close connection, client can close connection regardless of what server is doing

Arjentix: Yeah, but that wouldn't be clean closing. From #1195 issue:

> For now, the js client will force connections to terminate.

I think the reason of this problem is that [Closing handshake](https://datatracker.ietf.org/doc/html/draft-ietf-hybi-thewebsocketprotocol#page-10) requires response with the same `Close` message.

Interesting, that `warp` WebSocket seems to do this response automatically, so I don't have to do anything. However I still has to call `recv()` (or `next()`) to retrieve that message. I've found this out during tests, but additional tests with some real js client would be good

mversic: I thought that closing connection abbruptly on the client side would produce an error on the server, not on the client

Yes, warp websocket handles close message automatically. I think that is documented

Arjentix: After talk in DM we decided to resolve this conversation and leave everything like it is

core/src/torii/routing.rs
```diff
@@ -161,16 +161,84 @@ async fn stream_blocks<W: WorldTrait>(
...
+    ///
+    /// Ideally should return `Result<!>` cause it either runs forever either returns `Err` variant
+    async fn subscribe_forever(events: EventsSender, stream: WebSocket) -> Result<()> {
+        let mut events = events.subscribe();
+        let mut consumer = Consumer::new(stream).await?;
+
+        loop {
+            tokio::select! {
+                // This branch catches `Close` ans unexpected messages
+                message_opt = consumer.stream_mut().next() => {
```
mversic, line 223: Would it be better/possible to have something like `consumer.stream_closed` which would encapsulate the logic instead of exposing `stream_mut`?

Totally up to you

core/src/stream.rs
```diff
@@ -81,12 +123,18 @@ impl StreamMessage for warp::ws::Message {
...
         self.as_bytes()
     }
+
     fn is_binary(&self) -> bool {
         self.is_binary()
     }
+
+    fn is_close(&self) -> bool {
+        self.is_close()
+    }
```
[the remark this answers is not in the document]
mversic: nevermind, I see it's used here

mversic (dismissed):

core/src/torii/routing.rs
s8sato: What is that conclusion?

mversic: `select!` is needed because while waiting on events a close message can be received on the websocket and not get handled until new event is received

s8sato: BTW what is the reason we don't use [`WebSocket::close`](https://docs.rs/warp/latest/warp/filters/ws/struct.WebSocket.html#method.close) ?

Arjentix: Cause we don't have to explicitly close socket like this. It's somehow being closed automatically by `warp` (if you try to send something after receiving `Close` message you'll get error with message `Connection closed normally`).
My test is working without `WebSocket::close()` on server side. May be there is any case which can catch this, but I don't know it

s8sato: This implies a contradiction with the warp docs. Worth reporting
> Close messages need to be handled explicitly

s8sato (approved):

core/src/torii/routing.rs
Arjentix: I'm gonna create minimal example for that soon and check it. If I reproduce this behaviour, then I'll report them

mversic (approved):

s8sato (approved):

PR #1905, opened 2022-02-12
[refactor] #1893: Split data_model lib.rs file

### LINKED ISSUE #1893: Split data_model file
Need to split data_model/src/lib.rs file into separate ones.

Arjentix:
### Description of the Change

Split `data_model/src/lib.rs` file into separate ones. Just moved some modules into its own files.

Reading `data_model/src/lib.rs` file shouldn't have gotten any worse, cause every module is completely independent from each other

### Issue

Resolves #1893 

### Benefits

Now it's easier to read and navigate in `data_model` crate


mversic (approved):
LGTM. `permissions` and `uri` may or may not be a separate module

appetrosyan (approved):

PR #1906, opened 2022-02-13
[fix] #1908: fix zero value handling for telemetry analysis script

### LINKED ISSUE #1908: Zero values are currently mishandled in the push workflow
The "print debug telemtry info" check is failing from each and every commit in iroha2-dev branch's history. The reason is because of this error:
`jq: error (at <stdin>:0): null (null) and number (0) cannot be divided`

In the `./scripts/analyze_telemetry.sh` file, there is a line that divides two values:
`jq -sr '[.[].duration] | add/length`

ahmedelkashev:
Signed-off-by: Ahmed Elkashef <ahmedelkashef2012@gmail.com>

### Description of the Change

add a select() function to filter out the null or zero values from the ./analyze-telemtry.sh script. 

### Issue

Closes #1908

and in some cases, the value can be 0 or null, and dividing by 0 is not allowed and causes an error.


appetrosyan (dismissed):
Good catch.

scripts/analyze_telemetry.sh
```diff
@@ -18,7 +18,7 @@ top() {
 
 get_func() { jq -r "select(.name == \"$1\")"; }
 
-mean() { jq -sr '[.[].duration] | add/length'; }
+mean() { jq -sr '[.[].duration] | add/length | select(.duration != null and .duration != 0)'; }
```
s8sato, line 21:
```suggestion
mean() { jq -sr '[.[].duration] | select(length > 0) | add/length'; }
```
Check `length` before the division
https://stackoverflow.com/questions/26195214/how-to-nicely-remove-empty-array-in-jq

s8sato (approved):

appetrosyan (approved):

PR #1909, opened 2022-02-14
[refactor]: Re-organise core

### LINKED ISSUE #1812: Add warnings when `iroha` is given unrecognised command-line flags.
Currently, the Iroha peer binary ignores the command line flags that are incorrect. Since the number of flags is limited, it should be possible to provide an efficient sanity check for the end user, warning of any unrecognised flags.

appetrosyan:
Signed-off-by: Aleksandr <a-p-petrosyan@yandex.ru>

# DO NOT REVIEW UNTIL AFTER RELEASE

### Description of the Change

- Split Sumeragi into smaller files, organised by function. 
- Add some checks to detect unrecognised CLI arguments in `cli`. 
- The Iroha instance type is moved into the `cli` crate. Rationale: needed for interface only. If you're using the core library directly, and don't care about the interface, you might want to override some functionality that's hard-coded into `Iroha<...>`. Plus `Iroha<…>` is specific to the CLI instance. 
- Move `torii` into the CLI crate. Rationale: `torii` is an interface element. Connections are (in principle) not the core functionality of Iroha, and we might change the net interface by moving to a different web/async runtime, but keep the core crate intact. Also None of the users of the `core` crate necessarily need to have access to `torii`. Separation makes sense. 
- `Name::new` now checks for both `#` and `@` symbols and rejects them with an appropriate error message. 
- Added feature flag for types used in testing by other crates. 
- Prevent unchecked variants of `Name::new` to leak into non-testing code. 
- Replaced unchecked `Name::new` with a checked panicking counter-part in benchmarks. This gives us an unrealistic expectation of what our performance is. Real-world applications will check their `Name` s. 

### Issue
Closes #1812 

### Benefits
- Less clutter. 
- Back-end — front-end separation.
- Isolation of commonly modified code (e.g. `torii`) from infrequently modified code (e.g. `kura`). 
- Smaller `core` crate. 

### Possible Drawbacks

None


core/src/stream.rs
```diff
@@ -103,8 +103,7 @@ impl<M: DecodeVersioned> Stream<M> for warp::ws::WebSocket {
     type Message = warp::ws::Message;
 }
 
-#[cfg(test)]
-mod tests {
+mod ws_client {
```
mversic, line 106: why would you remove `[cfg(test)]`?

appetrosyan: Temporary workaround. I'm planning on introducing a feature flag for cross-crate testing.

Issue #1480, opened 2021-09-28
Stop Iroha on panic


e-ivkov:
On panic in one of the threads we should stop Iroha2 with appropriate panic message. Currently it seemed that with `unreachable()!` panic Iroha peer was not stopped.

ales-tsurko:
How to reproduce it? When I call `unreachable!` after network initialization in tests, all of them failed/panicked.

PR #1910, opened 2022-02-14
[feature] #1898: share workdir as a volume with dev docker instances

### LINKED ISSUE #1898: Share working directory volume in docker
It's a neat feature to be able to access working directory from docker

mversic:

### Issue
Closes #1898


docker-compose-single.yml
```diff
@@ -14,4 +16,5 @@ services:
       - "1337:1337"
       - "8080:8080"
       - "8180:8180"
+    init: true
```
s8sato, line 19: Could you explain why this flag is necessary?

docker-compose-single.yml
```diff
@@ -1,8 +1,10 @@
-version: "3.3"
+version: "3.8"
 services:
   iroha0:
     build: .
     image: iroha0:debug
+    volumes:
+      - './:/root/soramitsu/iroha'
```
s8sato, line 7: The local repository path varies by developer though.
What is the reason we share it?

mversic: Yes, they vary. That is why `.`(current dir) is mapped to `/root/soramitsu/iroha` in container. It doesn't need to be mapped, I needed a way to copy locally built `iroha` into docker container and run it. I can get away with `docker cp` but maybe someone would like to build `iroha` inside container and now they have the possibility with shared volume

docker-compose-single.yml
mversic: There is no init system in docker and therefore, first process that is executed (`./iroha` in our case) gets PID = 1. This PID is special because it ignores signals sent to it like SIGKILL or SIGTERM. You can see this if you try to kill the process from inside container(and you cannot) or you can see that running `docker stop` takes 10sec to complete because it waits on graceful shutdown which isn't happening. 

This can be resolved in two ways, either by adding small init system called `tini` which is packed with docker or explicit signal handling needs to be included in the iroha binary(I haven't confirmed this)

appetrosyan: Explicit signal handling is planned after #1480. With that said, we don't gracefully exit regardless, so the implicit signal handling is fine for now.

appetrosyan (approved):

docker-compose-single.yml
s8sato: I mistook right for left of `:`

PR #1942, opened 2022-03-02
[feature] #1918: Implement basic authentication for `client`

### LINKED ISSUE #1918: Implement basic authentication for `client`
Implement basic authentication in `iroha_client` without backporting into `iroha_client_cli`.

Arjentix:
### Description of the Change

Added possibility to use [Basic Auth](https://www.twilio.com/docs/glossary/what-is-basic-authentication) for `client`

### Issue

Closes #1918

### Benefits

Simple use-case is to access Iroha, that is deployed on AWS servers, which requires Basic Auth

### Possible Drawbacks

None

### Usage Examples or Tests

See `iroha_client::client::tests::basic_auth` module

### Suggestion

May be we can use something like `iroha_doc` to generate docs for client's `config.json`.


client/src/client.rs
```diff
@@ -29,51 +29,63 @@ use crate::{
...
+    /// Fails if configuration contains only `login` or only `password`.
+    /// It should contain both or none of them
+    pub fn with_headers(
+        configuration: &Configuration,
+        mut headers: HashMap<String, String>,
+    ) -> Result<Self> {
+        match (&configuration.login, &configuration.password) {
+            (Some(login), Some(password)) => {
+                let credentials = format!("{}:{}", login, password);
+                let encoded = base64::encode(credentials);
```
mversic, line 77: why does `base64` come all the way from `iroha_crypto` just to be used here? I think it would be fine to make it a dependency of `iroha_client` directly

configs/client_cli/config.json
```diff
@@ -6,6 +6,8 @@
       "name": "wonderland"
     }
   },
+  "LOGIN": "mad_hatter",
```
appetrosyan, line 9:
```suggestion
  "WEB_LOGIN": "mad_hatter",
```

I'd add a comment here explaining the use case. Since most people can confuse their `LOGIN` with the `ACCOUNT` that they'll use with the blockchain, I think an explicit prefix of `WEB` makes sense here.

client/src/config.rs
```diff
@@ -27,6 +27,10 @@ pub struct Configuration {
     pub private_key: PrivateKey,
     /// User account id.
     pub account_id: AccountId,
+    /// Login for basic auth. Should be used with `password`.
+    pub login: Option<SmallStr>,
+    /// Password for basic auth. Should be used with `login`.
+    pub password: Option<SmallStr>,
```
mversic, line 33: logically, these could be grouped into a `Option<(SmallStr, SmallStr)>` and that way creating client couldn't result in an error. I grant that this might be inconvenient

client/src/client.rs
Arjentix: @appetrosyan suggested this approach. Well, it's done as a feature of `iroha_cryto`, so other crates don't get this library. But if any crate in future will need this, it will be easy to enable this feature.
Let's see what others think

appetrosyan: I actually suggested something closer to the first approach. You will raise a serialisation error when you can't find both the login and password. So you don't need an explicit check for whether you don't have either.

configs/client_cli/config.json
appetrosyan: I don't see any conflicts. You need to ensure that the user knows that this is not used for blockchain operations, just to get through basic authentication "firewalls". That comment is about something completely different.

client/src/client.rs
appetrosyan: While I think that  hiding it behind a feature flag is the right approach, it might be a premature optimization. When I made the suggestion, I assumed that Iroha will also need to process that information in order not to break, but since no changes are made beyond the client, we can do either.

client/src/config.rs
appetrosyan: I like the suggestion to make it a `BasicAuth` struct. There's a subtlety that I hadn't initially considered: 
If you specify one field and not the other it's a malformed config file. 

But in the spec, both `password` and the `login` can be empty: base64: `Og==` for `null`. These would deserialise to the same as a malformed config.

appetrosyan: Which reminds me, neither field should have whitespace nor the `:` character in them. @Arjentix I would follow the way it's done in #1909 , for the `Name` type. Don't bother with providing unchecked variants, it's always better to `expect` and `panic`.

Arjentix: Done! [RFC](https://datatracker.ietf.org/doc/html/rfc7617#section-2) says, that only `login` can't have `:` character.

I'm closing this conversation, cause it's outdated now

client/src/client.rs
Arjentix: I'll wait for more responses here

configs/client_cli/config.json
Arjentix: Done in new commit

client/src/client.rs
mversic: I should add that `base64` is encoding and not a form of encryption and if seen from that perspective shouldn't belong to `iroha_crypto`

Arjentix: Ok, I'll move it to `client`

mversic (approved):

Arjentix:
@s8sato , here is the thing -- we don't need one. Basic Auth is done not for Iroha itself, but for other services.
For example AWS, there Iroha peer is located, requires clients to use Basic Auth to pass requests to Iroha. So in `web_login` and `password` user will use it's AWS login and password

s8sato (approved):

Issue #1230, opened 2021-07-15
100 Peer Network for Testing


takemiyamakoto:
Can probably use vultr

mversic:
duplicate

Issue #1204, opened 2021-07-15
Setup distributed test running in CI


takemiyamakoto:
The iroha nodes should be physically on different peers.

appetrosyan:
There is some progress on this with Sofiane being assigned to the relevant Jira issue.

PR #1963, opened 2022-03-08
[feature] #1952: Add a TPS benchmark as a standard for optimizations

### LINKED ISSUE #1952: Add benchmark for network with 4 peers
# Goal
Establish a standard for optimizations about TPS
- Provide the difference before and after the code change, rather than the absolute value
- A benchmark to publish official TPS will be left to another issue
  - #1230
  - #1204

# How to
- Add one to `cargo bench`
- Use `test_network` crate

## What to be measured
`TPS := (txs/block)/(block/sec)`
- `iroha_telemetry::metrics::Metrics` has the source

## Issue
Generate a series of transactions with some diversity and reproduce it

# Related issues
- #1042
- #1230
- #1204
- #1590
- #1479
- #1225

s8sato:
### Description of the Change
Add a TPS benchmark for performance check during development

image

[Report sample](https://github.com/hyperledger/iroha/files/8395528/tps_4peers-80000interval_us-8192max_txs-5blocks-10samples.-.Criterion.rs.pdf)

### Issue
* Closes #1952
* Benchmarks to publish official TPS will be left to another issue:
  - #1230
  - #1204

### Benefits
New standard for TPS optimizations

### Possible Drawbacks
None

### Usage Examples or Tests

[README](https://github.com/s8sato/iroha/blob/feat/1952/client/benches/tps/README.md)

### DIFF
--- a/client/benches/tps/README.md
+++ b/client/benches/tps/README.md
@@ -1,0 +1,34 @@
+# tps
+
+## Usage
+
+1. Establish a baseline:
+
+    ```
+    git checkout iroha2-dev
+    ```
+    ```
+    cargo bench --bench tps-dev
+    ```
+
+1. Compare against the baseline:
+
+    ```
+    git checkout <your-optimization-commit>
+    ```
+    ```
+    cargo bench --bench tps-dev
+    ```
+
+1. See [the report](../../../target/criterion/report/index.html)
+
+* In case the benchmark fails, please try to take [`interval_us_per_tx`](config.json) longer.
+
+* Also single trial of the measurement would help:
+
+    ```
+    cd client
+    ```
+    ```
+    cargo run --example tps-oneshot
+    ```
--- a/client/benches/tps/README.md
+++ b/client/benches/tps/README.md
@@ -19,7 +19,7 @@
     ```
     cargo bench --bench tps-dev
     ```
-    * CAUTION: "improved" and "regressed" message are reversed due to the nature of Criterion.rs, which is supposed to measure time
+    * CAUTION: "improved" and "regressed" message are reversed due to the nature of Criterion.rs, which is supposed to measure time instead of throughput by default
 
 3. See [the report](../../../target/criterion/report/index.html)
 
--- a/client/benches/tps/README.md
+++ b/client/benches/tps/README.md
@@ -11,7 +11,7 @@
     cargo bench --bench tps-dev
     ```
 
-1. Compare against the baseline:
+2. Compare against the baseline:
 
     ```
     git checkout <your-optimization-commit>
@@ -19,8 +19,9 @@
     ```
     cargo bench --bench tps-dev
     ```
+    * CAUTION: "improved" and "regressed" message are reversed due to the nature of Criterion.rs, which is supposed to measure time
 
-1. See [the report](../../../target/criterion/report/index.html)
+3. See [the report](../../../target/criterion/report/index.html)
 
 * In case the benchmark fails, please try to take [`interval_us_per_tx`](config.json) longer.
 
--- a/client/benches/tps/config.json
+++ b/client/benches/tps/config.json
@@ -1,0 +1,7 @@
+{
+    "peers": 4,
+    "interval_us_per_tx": 80000,
+    "max_txs_per_block": 8192,
+    "blocks": 5,
+    "sample_size": 10
+}
--- a/client/benches/tps/oneshot.rs
+++ b/client/benches/tps/oneshot.rs
@@ -1,0 +1,11 @@
+//! Single trial of the benchmark
+
+mod lib;
+
+#[allow(clippy::expect_used)]
+fn main() {
+    let config = lib::Config::from_path("benches/tps/config.json").expect("Failed to configure");
+    let tps = config.measure().expect("Failed to measure");
+    iroha_logger::info!(?config);
+    iroha_logger::info!(%tps);
+}
--- a/data_model/src/events/pipeline.rs
+++ b/data_model/src/events/pipeline.rs
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 
 pub use crate::transaction::RejectionReason as PipelineRejectionReason;
 
-/// Event filter.
+/// [`Event`] filter.
 #[derive(
     Default,
     Debug,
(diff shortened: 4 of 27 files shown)


.github/workflows/iroha2-dev-pr-bench.yml
```diff
@@ -0,0 +1,29 @@
+name: I2::Dev::Bench
```
appetrosyan, line 1: In general I think that CI integration should be done in a separate PR by a DevOps. Running benchmarks on the CI runners is a bad idea, since the runners are "ondemand", meaning that the performance can vary. I would remove this workflow file entirely and focus on the actual benchmark.

appetrosyan (changes requested):

data_model/src/events/pipeline.rs
```diff
@@ -93,19 +100,19 @@ pub enum EntityType {
     Transaction,
 }
 
-/// Entity type to filter events.
+/// [`Event`].
```
appetrosyan, line 103:
```suggestion
/// Strongly-typed [`Event`], which tells the receiver the type of entity that changed, the change, and the hash of the entity. 
```

client/benches/tps/ci.rs
```diff
@@ -0,0 +1,22 @@
+//! One-shot benchmark by counting the CPU instructions and memory accesses
+//! using [iai](https://github.com/bheisler/iai)
+//! for checking performance improvements or regressions in CI environments
```
appetrosyan, line 3: Again. Running this kind of benchmark on the CI is going to be very unreliable. You want to run this on a stand that has repeatable performance. This is not the description of our AWS runners.

data_model/src/events/pipeline.rs
```diff
@@ -93,19 +100,19 @@ pub enum EntityType {
...
+    /// [`Status`] of the entity that caused this [`Event`].
     pub status: Status,
-    /// The hash of this entity.
+    /// [`Hash`] of the entity that caused this [`Event`].
     pub hash: Hash,
 }
 
 impl Event {
-    /// Constructs pipeline event.
+    /// Construct [`Event`].
```
appetrosyan, line 115: Why have this constructor? The `struct` literal syntax is going to be less verbose and more explicit.

data_model/src/events/pipeline.rs
```diff
@@ -126,11 +133,48 @@ pub enum Status {
...
+    /// Represents [`Status::Validating`].
+    Validating,
+    /// Represents [`Status::Rejected`].
+    Rejected,
+    /// Represents [`Status::Committed`].
+    Committed,
+}
+
+impl Status {
+    fn type_(&self) -> StatusType {
```
appetrosyan, line 162:
```suggestion
    fn kind(&self) -> StatusKind {
```
While this is not related to Haskell's `kind` system, I'd prefer not having names which collide with `rust` keywords. Especially since the standard way of naming a function called `type` is 
```rust 
r#type
```

data_model/src/events/pipeline.rs
s8sato:
```rust
PipelineEvent::new(
    PipelineEntityType::Transaction,
    PipelineStatus::Validating,
    transaction.hash().into(),
)
```

```rust
PipelineEvent {
    entity_type: PipelineEntityType::Transaction,
    status: PipelineStatus::Validating,
    hash: transaction.hash().into(),
}
```
I don't think so

.github/workflows/iroha2-dev-pr-bench.yml
s8sato: Yeah, I'm going to remove this workflow together with `tps-ci` on my final commit in that case

client/benches/tps/ci.rs
s8sato: Is it really useless even though [the doc](https://github.com/bheisler/iai#comparison-with-criterion-rs) says:

> Pro: Iai can work reliably in noisy CI environments or even cloud CI providers like GitHub Actions or Travis-CI, where Criterion-rs cannot.

appetrosyan: Interesting. I'd be a bit more skeptical. 

They claim that lai is good for noisy environments because they count the number of instructions. This is true in the general case, but since Iroha has a time-based consensus algorithm (and our tests frequently show instability if not enough performance is available), this kind of instability will add to the number of instructions that get executed. 

We should have benchmarks, but I think we'd need a dedicated test bench.

s8sato: Indeed, the instruction count will increase during timer countdowns

s8sato:
`cd client/benches/tps`
`cargo run --example tps-ex`
image

s8sato:
### WIP
```
cd client
```
```
cargo bench --bench tps-dev
```
image

### TODO
- [ ] Setup instructions

client/benches/tps/README.md
```diff
@@ -0,0 +1,34 @@
...
+1. Establish a baseline:
+
+    ```
+    git checkout iroha2-dev
+    ```
+    ```
+    cargo bench --bench tps-dev
+    ```
+
+1. Compare against the baseline:
```
appetrosyan, line 14:
```suggestion
2. Compare against the baseline:
```

client/benches/tps/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+    ```
+    git checkout <your-optimization-commit>
+    ```
+    ```
+    cargo bench --bench tps-dev
+    ```
+
+1. See [the report](../../../target/criterion/report/index.html)
```
appetrosyan, line 23:
```suggestion
3. See [the report](../../../target/criterion/report/index.html)
```

client/benches/tps/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "peers": 4,
```
appetrosyan, line 2: I'd add a comment, since these peers are `test_network` based entities. As my benchmarking shows the performance degrades significantly once the peers are in separate processes.

client/benches/tps/oneshot.rs
```diff
@@ -0,0 +1,11 @@
+//! Single trial of the benchmark
+
+mod lib;
+
+#[allow(clippy::expect_used)]
+fn main() {
+    let config = lib::Config::from_path("benches/tps/config.json").expect("Failed to configure");
+    let tps = config.measure().expect("Failed to measure");
+    iroha_logger::info!(?config);
```
appetrosyan, line 9:
```suggestion
    iroha_logger::debug!(?config);
```

appetrosyan (approved):

appetrosyan:
The word `Kind` in a computer-science context has a few connotations, but I think that the pattern `::new().hash()` is a little clearer than `::by_hash()`.

client/benches/tps/config.json
s8sato: Create another benchmark with separate processes instead of threads?

client/benches/tps/oneshot.rs
s8sato: I'd display the config by default for confirmation

appetrosyan (approved):

appetrosyan (dismissed):

client/benches/tps/config.json
appetrosyan: Ideally yes. But I think having a thread-based benchmark is also valuable.

client/benches/tps/oneshot.rs
appetrosyan: Another option is 
```rust
#[cfg(debug_assertions)]
iroha_logger::info!(?config);
#[cfg(not(debug_assertions))]
iroha_logger::debug!(?config);
```
If we're running the benchmark with `--nocapture` we might as well add `MAX_LOG_LEVEL=DEBUG` to see this kind of information.

client/benches/tps/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+2. Compare against the baseline:
+
+    ```
+    git checkout <your-optimization-commit>
+    ```
+    ```
+    cargo bench --bench tps-dev
+    ```
+    * CAUTION: "improved" and "regressed" message are reversed due to the nature of Criterion.rs, which is supposed to measure time
```
Arjentix, line 22: I can't understand why... At step 1 criterion saves its results into `target` directory. At step 2 it compares new results with previous. Why does Criterion swap this words?

data_model/src/events/pipeline.rs
```diff
@@ -29,48 +29,59 @@ pub use crate::transaction::RejectionReason as PipelineRejectionReason;
...
-    /// Filter by Hash if `Some`, if `None` all hashes are accepted.
+    /// If `Some::<EntityKind>` filters by the [`EntityKind`]. If `None` accepts all the [`EntityKind`].
+    pub entity_kind: Option<EntityKind>,
+    /// If `Some::<StatusKind>` filters by the [`StatusKind`]. If `None` accepts all the [`StatusKind`].
+    pub status_kind: Option<StatusKind>,
+    /// If `Some::<Hash>` filters by the [`Hash`]. If `None` accepts all the [`Hash`].
     pub hash: Option<Hash>,
 }
 
 impl EventFilter {
```
Arjentix, line 40: What about to add `inline` and `const` inside this impl block as much as possible?

Arjentix (approved):

appetrosyan (dismissed):

client/benches/tps/README.md
s8sato: Because in TPS time is the denominator

appetrosyan (approved):

appetrosyan (approved):

data_model/src/events/pipeline.rs
```diff
@@ -29,48 +29,65 @@ pub use crate::transaction::RejectionReason as PipelineRejectionReason;
...
-    /// Filter by entity
-    pub const fn by_entity(entity: EntityType) -> EventFilter {
-        EventFilter {
-            entity: Some(entity),
-            hash: None,
-        }
+    /// Construct [`EventFilter`].
+    #[must_use]
+    #[inline]
+    pub fn new() -> Self {
```
Arjentix, line 44: Hm..., you can't have it `const` because `Default` isn't const.
I'll just leave some interesting upcomming features here: https://stackoverflow.com/questions/70441495/what-is-impl-const-in-rust

Arjentix (approved):

PR #1989, opened 2022-03-21
[fix] #1897: Removed usize/isize from serialization

### LINKED ISSUE #1897: Remove `usize`/`isize` from serialization
it's possible for structures that are serialized through `serde` to contain `usize`/`isize` inside them. This is problematic because the size of these depends on the architecture and it's possible that deserialization could break. To fix this we should disallow the use of `usize`/`isize` in serialization and convert to `u64`/`i64`. For example, `usize` is used in `schema/src/lib.rs`, but there could be other places like `data_model`.

fortunately, parity SCALE codec doesn't implement encode/decode for `usize`/`isize` so only structures which implement `serde` serialization and not SCALE serialization are affected. There should be only a small number of these

SamHSmith:
### Description of the Change

In all the locations where a struct derives Serialize or Deserialize usize has been replaced
with either u64 or u32. No instances of isize existed. In the cases where code using the
data required a usize u32 was chosen as the replacement. There is no point in avoiding usize
in the on disk formats if 32-bit platforms can't read and use the values written.

### Issue

Resolves #1897 

### Benefits

No chance of crashing due to serializing on 64-bit then deserializing on 32-bit or vice versa.

### Possible Drawbacks

This change limits the size of certain configuration values to 4 billion. Probably a non issue.

### DIFF
--- a/actor/src/lib.rs
+++ b/actor/src/lib.rs
@@ -1,5 +1,6 @@
 //! Iroha simple actor framework.
 #![allow(clippy::same_name_method)]
+#![allow(clippy::expect_used)]
 
 #[cfg(feature = "deadlock_detection")]
 use std::any::type_name;
@@ -288,7 +289,7 @@ where
 #[async_trait::async_trait]
 pub trait Actor: Send + Sized + 'static {
     /// Capacity of actor queue
-    fn mailbox_capacity(&self) -> usize {
+    fn mailbox_capacity(&self) -> u32 {
         100
     }
 
@@ -305,7 +306,12 @@ pub trait Actor: Send + Sized + 'static {
     /// Initialize actor with its address.
     fn preinit(self) -> InitializedActor<Self> {
         let mailbox_capacity = self.mailbox_capacity();
-        InitializedActor::new(self, mailbox_capacity)
+        InitializedActor::new(
+            self,
+            mailbox_capacity
+                .try_into()
+                .expect("u32 should always fit in usize"),
+        )
     }
 
     /// Initialize actor with default values
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -69,10 +69,10 @@ pub struct Configuration {
 #[config(env_prefix = "IROHA_NETWORK_")]
 pub struct NetworkConfiguration {
     /// Actor mailbox size
-    pub mailbox: usize,
+    pub mailbox: u32,
 }
 
-const DEFAULT_MAILBOX_SIZE: usize = 100;
+const DEFAULT_MAILBOX_SIZE: u32 = 100;
 
 impl Default for NetworkConfiguration {
     fn default() -> Self {
--- a/cli/src/torii/config.rs
+++ b/cli/src/torii/config.rs
@@ -8,9 +8,9 @@ pub const DEFAULT_TORII_P2P_ADDR: &str = "127.0.0.1:1337";
 /// Default socket for reporting internal status and metrics
 pub const DEFAULT_TORII_TELEMETRY_URL: &str = "127.0.0.1:8180";
 /// Default maximum size of single transaction
-pub const DEFAULT_TORII_MAX_TRANSACTION_SIZE: usize = 2_usize.pow(15);
+pub const DEFAULT_TORII_MAX_TRANSACTION_SIZE: u32 = 2_u32.pow(15);
 /// Default upper bound on `content-length` specified in the HTTP request header
-pub const DEFAULT_TORII_MAX_CONTENT_LENGTH: usize = 2_usize.pow(12) * 4000;
+pub const DEFAULT_TORII_MAX_CONTENT_LENGTH: u32 = 2_u32.pow(12) * 4000;
 
 /// Structure that defines the configuration parameters of `Torii` which is the routing module.
 /// For example the `p2p_addr`, which is used for consensus and block-synchronisation purposes,
@@ -27,9 +27,9 @@ pub struct ToriiConfiguration {
     /// Torii URL for reporting internal status and metrics for administration.
     pub telemetry_url: String,
     /// Maximum number of bytes in raw transaction. Used to prevent from DOS attacks.
-    pub max_transaction_size: usize,
+    pub max_transaction_size: u32,
     /// Maximum number of bytes in raw message. Used to prevent from DOS attacks.
-    pub max_content_len: usize,
+    pub max_content_len: u32,
 }
 
 impl Default for ToriiConfiguration {
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -1,4 +1,5 @@
 //! Module with queue actor
+#![allow(clippy::expect_used)]
 
 use std::{sync::Arc, time::Duration};
 
@@ -80,12 +81,15 @@ impl<W: WorldTrait> Queue<W> {
     }
 
     /// Returns `n` randomly selected transaction from the queue.
-    pub fn n_random_transactions(&self, n: usize) -> Vec<VersionedAcceptedTransaction> {
+    pub fn n_random_transactions(&self, n: u32) -> Vec<VersionedAcceptedTransaction> {
         self.txs
             .iter()
             .filter(|e| self.is_pending(e.value()))
             .map(|e| e.value().clone())
-            .choose_multiple(&mut rand::thread_rng(), n)
+            .choose_multiple(
+                &mut rand::thread_rng(),
+                n.try_into().expect("u32 should always fit in usize"),
+            )
     }
 
     fn check_tx(&self, tx: &VersionedAcceptedTransaction) -> Result<(), Error> {
--- a/core/src/smartcontracts/wasm.rs
+++ b/core/src/smartcontracts/wasm.rs
@@ -1,6 +1,7 @@
 //! This module contains logic related to executing smartcontracts via
 //! `WebAssembly` VM Smartcontracts can be written in Rust, compiled
 //! to wasm format and submitted in a transaction
+#![allow(clippy::expect_used)]
 
 use std::sync::Arc;
 
@@ -140,7 +141,9 @@ impl<'a, W: WorldTrait> State<'a, W> {
             validator: None,
 
             store_limits: StoreLimitsBuilder::new()
-                .memory_size(config.max_memory)
+                .memory_size(config.max_memory.try_into().expect(
+                    "config.max_memory is a u32 so this can't fail on any supported platform",
+                ))
                 .instances(1)
                 .memories(1)
                 .tables(1)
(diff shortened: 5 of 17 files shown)


core/src/smartcontracts/wasm.rs
```diff
@@ -140,7 +140,7 @@ impl<'a, W: WorldTrait> State<'a, W> {
             validator: None,
 
             store_limits: StoreLimitsBuilder::new()
-                .memory_size(config.max_memory)
+                .memory_size(config.max_memory as usize)
```
appetrosyan, line 143: Which could panic. Or worse, truncate a u64. Some popular targets are still 32bit.

data_model/src/pagination.rs
```diff
@@ -129,10 +129,10 @@ impl From<Pagination> for Vec<(&'static str, usize)> {
     fn from(pagination: Pagination) -> Self {
         match (pagination.start, pagination.limit) {
             (Some(start), Some(limit)) => {
-                vec![(PAGINATION_START, start), (PAGINATION_LIMIT, limit)]
+                vec![(PAGINATION_START, start as usize), (PAGINATION_LIMIT, limit as usize)]
```
appetrosyan, line 132: Format with nightly cargo fmt.

appetrosyan (dismissed):

schema/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub trait IntoSchema {
 /// Applicable for types that represents decimal place of fixed point
 pub trait DecimalPlacesAware {
     /// decimal places of fixed point
-    fn decimal_places() -> usize;
+    fn decimal_places() -> u64;
```
appetrosyan, line 50: This can be safely assumed to be within u32. The fixed point number used here is a single i64, it can't possibly have more decimal places than u32. (It can have more decimal places than the mantissa of i64, but that fits into a single byte).

schema/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub trait IntoSchema {
 /// Applicable for types that represents decimal place of fixed point
```
mversic, line 47: inside `schema/src/bin/main.rs` we generate derive schemas for our structures. Would you be able to add a test there that would ensure no field is `usize`/`isize`?

cli/src/config.rs
```diff
@@ -69,10 +69,10 @@ pub struct Configuration {
 #[config(env_prefix = "IROHA_NETWORK_")]
 pub struct NetworkConfiguration {
     /// Actor mailbox size
-    pub mailbox: usize,
```
mversic, line 72: there is also the possibility to use serde field attributes to make custom serializations. With field attributes you can have a field on a struct be defined as `usize` and have it consistently serialized as `u32`(or `u64`). This might make sense if you will be converting `u32` into `usize` at the point of use anyway. @appetrosyan thoughts?

schema/src/lib.rs
mversic: if you go with serde fields I mentioned in the other comment, you may have to modify `IntoSchema` macro

cli/src/config.rs
appetrosyan: I think that making things fixed, portable and straightforward is preferable. We have plenty of structures that need a custom serialiser.

data_model/src/pagination.rs
Arjentix:
```bash
cargo +nightly-2021-12-02 fmt --all
```

might help

actor/src/lib.rs
```diff
@@ -347,8 +347,8 @@ pub struct InitializedActor<A: Actor> {
 
 impl<A: Actor> InitializedActor<A> {
     /// Constructor.
-    pub fn new(actor: A, mailbox_capacity: usize) -> Self {
-        let (sender, receiver) = mpsc::channel(mailbox_capacity);
+    pub fn new(actor: A, mailbox_capacity: u32) -> Self {
+        let (sender, receiver) = mpsc::channel(mailbox_capacity as usize);
```
Arjentix, line 351: I can't see why do we need this change, cause `InitializedActor` doesn't implement `Deserialize` and/or `Serialize`.
If something, that uses `InitializedActor`, also implements `Deserialize` and/or `Serialize`, then conversion should be done inside it

cli/src/torii/config.rs
```diff
@@ -8,9 +8,9 @@ pub const DEFAULT_TORII_P2P_ADDR: &str = "127.0.0.1:1337";
 /// Default socket for reporting internal status and metrics
 pub const DEFAULT_TORII_TELEMETRY_URL: &str = "127.0.0.1:8180";
 /// Default maximum size of single transaction
-pub const DEFAULT_TORII_MAX_TRANSACTION_SIZE: usize = 2_usize.pow(15);
+pub const DEFAULT_TORII_MAX_TRANSACTION_SIZE: u64 = 2_u64.pow(15);
 /// Default upper bound on `content-length` specified in the HTTP request header
-pub const DEFAULT_TORII_MAX_CONTENT_LENGTH: usize = 2_usize.pow(12) * 4000;
+pub const DEFAULT_TORII_MAX_CONTENT_LENGTH: u64 = 2_u64.pow(12) * 4000;
```
Arjentix, line 13: `u32` should be enough here?

Arjentix: Same for relevant fields in `ToriiConfiguration`

core/src/queue.rs
```diff
@@ -80,12 +80,12 @@ impl<W: WorldTrait> Queue<W> {
...
     /// Returns `n` randomly selected transaction from the queue.
-    pub fn n_random_transactions(&self, n: usize) -> Vec<VersionedAcceptedTransaction> {
+    pub fn n_random_transactions(&self, n: u32) -> Vec<VersionedAcceptedTransaction> {
         self.txs
             .iter()
             .filter(|e| self.is_pending(e.value()))
             .map(|e| e.value().clone())
-            .choose_multiple(&mut rand::thread_rng(), n)
+            .choose_multiple(&mut rand::thread_rng(), n as usize)
     }
```
Arjentix, line 89: I think it should be this method user responsibility to provide correct `usize` number. So I think this method should be left unchanged

core/src/smartcontracts/wasm.rs
```diff
@@ -140,7 +140,7 @@ impl<'a, W: WorldTrait> State<'a, W> {
             validator: None,
 
             store_limits: StoreLimitsBuilder::new()
-                .memory_size(config.max_memory)
+                .memory_size(config.max_memory as usize) // config.max_memory is a u32 so this can't fail on any supported platform
```
Arjentix, line 143: May be better use `try_from()` with `expect()` containing this comment?

data_model/src/pagination.rs
```diff
@@ -60,7 +60,7 @@ impl<I: Iterator> Iterator for Paginated<I> {
         #[allow(clippy::option_if_let_else)]
         // Required because of E0524. 2 closures with unique refs to self
         if let Some(start) = self.pagination.start.take() {
-            self.iter.nth(start)
+            self.iter.nth(start as usize)
```
Arjentix, line 63: I think every such place with `as usize` should be rewritten as `try_from(...).expect("This can't fail because ...")`

logger/src/telemetry.rs
```diff
@@ -136,10 +136,10 @@ impl<S: Subscriber> TelemetryLayer<S> {
...
-        channel_size: usize,
+        channel_size: u32,
     ) -> (impl Subscriber, Receiver<Telemetry>, Receiver<Telemetry>) {
-        let (sender, receiver) = mpsc::channel(channel_size);
-        let (sender_future, receiver_future) = mpsc::channel(channel_size);
+        let (sender, receiver) = mpsc::channel(channel_size as usize);
+        let (sender_future, receiver_future) = mpsc::channel(channel_size as usize);
         let telemetry = Self::from_senders(subscriber, sender, sender_future);
         (telemetry, receiver, receiver_future)
     }
```
Arjentix, line 145: `TelemetryLayer` doesn't implement `Deserialize` or `Serialize` so no need for changes here

schema/src/lib.rs
```diff
@@ -359,7 +359,7 @@ impl<T: IntoSchema, const L: usize> IntoSchema for [T; L] {
         let _ = map.entry(Self::type_name()).or_insert_with(|| {
             Metadata::Array(ArrayMeta {
                 ty: T::type_name(),
-                len: L,
+                len: L as u64,
```
Arjentix, line 362: Also think, that it's better to use `expect()`

```rust
try_from(L).expect("`usize` always fits in `u64`")
```

Arjentix (changes requested):

schema/src/lib.rs
SamHSmith: Good suggestion. I will take a look at this tomorrow.

schema/src/lib.rs
appetrosyan: It really is better, unfortunately Rust still supports some of C's cast-related functionality, even though it doesn't actually need most of it.

core/src/queue.rs
SamHSmith: I don't think we want to write code that supports certain behavior on certain platforms. Providing a value bigger than U32_MAX should not be allowed because that won't run on 32bit, better make the caller aware of this through a 32bit argument.

schema/derive/src/lib.rs
```diff
@@ -27,7 +27,20 @@ pub fn schema_derive(input: TokenStream) -> TokenStream {
...
+    if input_string.contains("usize") {
+        let to_add_tokens: TokenStream =
+            quote!(compile_error!("You may not use usize with IntoSchema");).into();
+        output.extend(to_add_tokens);
+    }
+    if input_string.contains("isize") {
+        let to_add_tokens: TokenStream =
+            quote!(compile_error!("You may not use isize with IntoSchema");).into();
+        output.extend(to_add_tokens);
+    }
```
Arjentix, line 42: In procedural macros you have to panic if something is wrong. No need to insert `compile_error!` inside the output code.

From [procedural macros reference](https://doc.rust-lang.org/reference/procedural-macros.html):
>  Panics are caught by the compiler and are turned into a compiler error

Arjentix: Also It seems to be a little bit hacky to scan input for `usize` or `isize` words such that. I think we should see, what @mversic thinks

Arjentix (changes requested):

schema/derive/src/lib.rs
appetrosyan: This is the best way to do it with Macros. However, Macros are not the best tool for the job. 

The real solution to this problem is a trait conflict that prevents users from implementing `IntoSchema` for a type `usize` altogether. 

So the macro either stays as is, and we then add the trait conflict, or the macro does a run-time check with proper type_id.

SamHSmith: I will open a new ticket for this, instead.

Arjentix: #2004

Arjentix (dismissed):
Good job!

appetrosyan (dismissed):

appetrosyan (dismissed):

Arjentix (dismissed):

Arjentix (dismissed):

Arjentix (approved):

Issue #2005, opened 2022-03-24
Fix `listen_for_events()` in `Client`


Arjentix:
When running some integration tests that use event subscription (for example `time_trigger_execution_count_error_should_be_less_than_10_percent`) an error message apears in log:

```
ERROR iroha::torii::routing: Failed to subscribe someone error=Event consuming error: WebSocket error: WebSocket protocol error: Connection reset without closing handshake
```

So `Client` uses API a bit wrong and doesn't close subscription connection properly. To make closing handshake a `WebSocket::close()` method should be called on `stream` field of `EventIterator`. 

Ideally this fix should provide the same good-looking API as before. User shouldn't close connection manualy

PR #2006, opened 2022-03-24
[feature] #1621: Introduce By Call Triggers

### LINKED ISSUE #1621: Introduce By Call Triggers
See [Trigger RFC](https://wiki.hyperledger.org/display/iroha/Triggers)

By Call Triggers - Triggers that are triggered by specific ISI call: ExecuteTrigger(Params)

Arjentix:
### Description of the Change

- Added new `ExecuteTrigger` *ISI*
- Added new event and filter: `ExecuteTriggerEvent` and `ExecuteTriggerEventFilter`
- Changed `WSV` so that every clone stores its own copy of events. No reference counting anymore
- Added some integration tests for a new instruction
- Opened #2005 

### Issue

Closes #1621 

### Benefits

Now we can execute trigger by ISI call

### Possible Drawbacks

Can't see any

### Usage Examples or Tests

See `client/test/integration/triggers/by_call_trigger.rs`


core/src/wsv.rs
```diff
@@ -807,6 +828,23 @@ impl<W: WorldTrait> WorldStateView<W> {
     {
         self.modify_world(|world| f(&world.triggers).map(WorldEvent::Trigger))
     }
+
+    /// Execute trigger with `trigger_id` as id and `authority` as owner
+    ///
+    /// Produces [`ExecuteTriggerEvent`].
+    /// Actual trigger execution will be in the next [`Self::apply()`] call
```
s8sato, line 835: It seems to me that within the same `apply()`, `ExecuteTriggerEvent` is picked up immediately after `execute_transactions()` and can pull the trigger

core/src/wsv.rs
```diff
@@ -107,6 +107,27 @@ impl<W: WorldTrait + Default> Default for WorldStateView<W> {
     }
 }
 
+impl<W: WorldTrait + Clone> Clone for WorldStateView<W> {
```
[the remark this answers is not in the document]
s8sato: Scratch that. We already have WSV: Clone

core/src/wsv.rs
Arjentix: Indeed, I've missed it. So there are two ways:

1. When `ExecuteTrigger` ISI is inside a *transaction*, *trigger* will be executed on the **same** block
2. When `ExecuteTrigger` ISI is inside a *trigger*, *trigger* will be executed on the **next** block

Is it okay for our purpose?

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -69,18 +68,14 @@ fn time_trigger_execution_count_error_should_be_less_than_10_percent() -> Result
 }
 
 /// Build register ISI for trigger which mints roses
-fn build_register_trigger_isi(
-    account_id: AccountId,
-    asset_id: AssetId,
-    schedule: TimeSchedule,
-) -> Result<RegisterBox> {
-    let instruction = MintBox::new(1_u32, asset_id);
+fn build_register_trigger_isi(asset_id: AssetId, schedule: TimeSchedule) -> Result<RegisterBox> {
```
appetrosyan, line 71: I'd like this function to be part of our public API. Most of our downstream clients may want something like this.

core/src/wsv.rs
appetrosyan: Yes. I’ll just need to add the information to the tutorial.

Arjentix: I'll mention this information in doc-comments and close this conversation

client/tests/integration/triggers/time_trigger.rs
Arjentix: Created #2009

s8sato (approved):

appetrosyan (dismissed):

appetrosyan (approved):

s8sato (approved):

Issue #1917, opened 2022-02-17
Use declarative macros in  ` impl FromStr` for `AssetValueType` 


appetrosyan:
This piece of code could be improved significantly by usage of declarative macros. 

```rust 
impl FromStr for AssetValueType {
    type Err = &'static str;

    fn from_str(value_type: &str) -> Result<Self, Self::Err> {
        // TODO: Could be implemented with some macro
        match value_type {
            "Quantity" => Ok(AssetValueType::Quantity),
            "BigQuantity" => Ok(AssetValueType::BigQuantity),
            "Fixed" => Ok(AssetValueType::Fixed),
            "Store" => Ok(AssetValueType::Store),
            _ => Err("Unknown variant"),
        }
    }
}
```

PR #2039, opened 2022-03-30
[fix] #1917: Use already existing FromStr macro

SamHSmith:
### Description of the Change

Replaced the FromStr implementation done by hand for the type AssetValueType with a derive macro from the rust crate strum.

### Issue

#1917 

### Benefits

Same functionality, less code.

### Possible Drawbacks

We depend on a foreign rust crate.

### DIFF
--- a/data_model/Cargo.toml
+++ b/data_model/Cargo.toml
@@ -36,6 +36,8 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.59", default-features = false }
 warp = { version = "0.3", default-features = false, optional = true }
 thiserror = { version = "1.0.28", optional = true }
+strum = "0.24.0"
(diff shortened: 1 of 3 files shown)


data_model/Cargo.toml
```diff
@@ -36,6 +36,8 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.59", default-features = false }
 warp = { version = "0.3", default-features = false, optional = true }
 thiserror = { version = "1.0.28", optional = true }
+strum = "0.24.0"
```
appetrosyan, line 39:
```suggestion
strum = { version = "=0.24.0", default-features = false }
```

appetrosyan (changes requested):

SamHSmith:
Well strum uses std so this PR is void.

PR #2054, opened 2022-04-04
[fix] #1917: Added easy_from_str_impl macro for use with AssetValueType

SamHSmith:
### Description of the Change

Created a variadic declarative macro that implements FromStr for C like enums.

### Issue

#1917 

### Benefits

Future FromStr implementations for C like enums can be a lot smaller codesize.

### Possible Drawbacks

May be harder to read and understand for some.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,8 @@ The following is a short set of guidelines for contributing to Iroha.
 * Fork [Iroha](https://github.com/hyperledger/iroha/tree/iroha2-dev).
 * Fix your issue of choice.
 * Write [tests](https://doc.rust-lang.org/cargo/commands/cargo-test.html). Ensure they all pass (`cargo test`).
-* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --tests --benches`
-* Format code `cargo fmt --all` and generate docs `cargo run --bin iroha_docs >"docs/source/references/config.md" && git add "docs/source/references/config.md"`
+* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --benches --tests --examples --all-features`
+* Format code `cargo +nightly fmt --all` and generate docs `cargo run --bin iroha_docs >"docs/source/references/config.md" && git add "docs/source/references/config.md"`
 * `git pull -r hyperledger iroha2-dev`, `git commit -s`, `git push <your-fork>`, and [create a pull request](https://github.com/hyperledger/iroha/compare) to the `iroha2-dev` branch on GitHub.
 
 ### Reporting Bugs
--- a/data_model/src/asset.rs
+++ b/data_model/src/asset.rs
@@ -150,20 +150,30 @@ pub enum AssetValueType {
     Store,
 }
 
-impl FromStr for AssetValueType {
-    type Err = &'static str;
-
-    fn from_str(value_type: &str) -> Result<Self, Self::Err> {
-        // TODO: Could be implemented with some macro
-        match value_type {
-            "Quantity" => Ok(AssetValueType::Quantity),
-            "BigQuantity" => Ok(AssetValueType::BigQuantity),
-            "Fixed" => Ok(AssetValueType::Fixed),
-            "Store" => Ok(AssetValueType::Store),
-            _ => Err("Unknown variant"),
+/// A declarative macro that implements `FromStr` for a given
+/// C like enumeration. The macro is invoked like follows:
+/// `easy_from_str_impl! { NameOfEnum, EnumVariation1, EnumVariation2, ... }`
+macro_rules! easy_from_str_impl {
+    (eval_to $cmp:expr, $enum_type:ty, $enum_value:tt) => {
+        if $cmp == stringify!($enum_value) {
+            return Ok(<$enum_type>::$enum_value);
         }
-    }
+    };
+    ($enum_type:ty, $( $enum_value:tt ),+ ) => {
+        impl FromStr for $enum_type {
+            type Err = &'static str;
+
+            fn from_str(value_type: &str) -> Result<Self, Self::Err> {
+                $(
+                    easy_from_str_impl!{eval_to value_type, $enum_type, $enum_value}
+                )+
+                return Err(concat!("Unknown variant for type ", stringify!($enum_type)));
+            }
+        }
+    };
 }
+
+easy_from_str_impl! {AssetValueType, Quantity, BigQuantity, Fixed, Store}
 
 /// Asset's inner value.
 #[derive(


appetrosyan (dismissed):

data_model/src/asset.rs
```diff
@@ -150,21 +150,31 @@ pub enum AssetValueType {
...
-            "BigQuantity" => Ok(AssetValueType::BigQuantity),
-            "Fixed" => Ok(AssetValueType::Fixed),
-            "Store" => Ok(AssetValueType::Store),
-            _ => Err("Unknown variant"),
+/// A declarative macro that implements `FromStr` for a given
+/// C like enumeration. The macro is invoked like follows:
+/// `easy_from_str_impl! { NameOfEnum, EnumVariation1, EnumVariation2, ... }`
+macro_rules! easy_from_str_impl {
+    (eval_to $cmp:expr, $enum_type:ty, $enum_value:tt) => {
+        if $cmp == stringify!($enum_value) {
```
mversic, line 158: it's better to use match statement because it will most likely be compiled into a jump table. With `if-else` statements I wouldn't be so sure of it. If you're having trouble with implementing it as a declarative macro, I think you can find examples of generating match statements with declarative macros in the codebase

CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@ The following is a short set of guidelines for contributing to Iroha.
 * Fork [Iroha](https://github.com/hyperledger/iroha/tree/iroha2-dev).
 * Fix your issue of choice.
 * Write [tests](https://doc.rust-lang.org/cargo/commands/cargo-test.html). Ensure they all pass (`cargo test`).
-* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --tests --benches`
-* Format code `cargo fmt --all` and generate docs `cargo run --bin iroha_docs >"docs/source/references/config.md" && git add "docs/source/references/config.md"`
+* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --benches --tests --examples --quiet --all-features`
+* Format code `cargo +nightly fmt --all` and generate docs `cargo run --bin iroha_docs >"docs/source/references/config.md" && git add "docs/source/references/config.md"`
```
mversic, line 16: is nightly required? I thought it's required only for wasm target

data_model/src/asset.rs
mversic: also, I would be more in favor of using procedural macro here which would guarantee exhaustive match if new variant is added. I don't think you can make such a guarantee with a declarative macro

I don't think we've got much from this macro in it's current form

appetrosyan: What Sam can do here is use this as a reference implementation for an Iroha macro procedural macro. He could also do that as a separate ticket.

CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@ The following is a short set of guidelines for contributing to Iroha.
 * Fork [Iroha](https://github.com/hyperledger/iroha/tree/iroha2-dev).
 * Fix your issue of choice.
 * Write [tests](https://doc.rust-lang.org/cargo/commands/cargo-test.html). Ensure they all pass (`cargo test`).
-* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --tests --benches`
-* Format code `cargo fmt --all` and generate docs `cargo run --bin iroha_docs >"docs/source/references/config.md" && git add "docs/source/references/config.md"`
+* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --benches --tests --examples --quiet --all-features`
```
appetrosyan, line 15:
```suggestion
* Fix [`clippy`](https://lib.rs/crates/cargo-lints) warnings: `cargo lints clippy --workspace --benches --tests --examples --all-features`
```
We only add quiet, so that the error message are easier to find in the interface. In general this is not a good idea.

data_model/src/asset.rs
SamHSmith: > it's better to use match statement because it will most likely be compiled into a jump table. With `if-else` statements I wouldn't be so sure of it. If you're having trouble with implementing it as a declarative macro, I think you can find examples of generating match statements with declarative macros in the codebase

It is not possible to produce a jump table for string comparison. Maybe if you were comparing string literals only and knew that the same string would have the same pointer into the .text block in the binary you could. But &str has no such guarantee and so the compiler output will be the same for match and multiple if's.

SamHSmith: I've created the seperate ticket #2056 for the tracking of the better macro.

mversic: > It is not possible to produce a jump table for string comparison. Maybe if you were comparing string literals only and knew that the same string would have the same pointer into the .text block in the binary you could. But &str has no such guarantee and so the compiler output will be the same for match and multiple if's.

ah well, perfect hashing approach would make it possible to produce jump table even for strings. Though it may be unlikely compiler would make such an "optimization"

mversic: it doesn't even have to be a perfect hash

appetrosyan: While this conversation is resolved, I would have to concede, that a procedural macro can use a hash map (or literal interning) to make the comparison a lightweight jump.

data_model/src/asset.rs
```diff
@@ -150,21 +150,31 @@ pub enum AssetValueType {
...
+            return Ok(<$enum_type>::$enum_value);
         }
-    }
+    };
+    ($enum_type:ty, $( $enum_value:tt ),+ ) => {
+        impl FromStr for AssetValueType {
```
Arjentix, line 163: I think you mean `$enum_type` here

data_model/src/asset.rs
```diff
@@ -150,21 +150,31 @@ pub enum AssetValueType {
...
+            type Err = &'static str;
+
+            fn from_str(value_type: &str) -> Result<Self, Self::Err> {
+                $(
+                    easy_from_str_impl!{eval_to value_type, $enum_type, $enum_value}
+                )+
+                return Err("Unknown variant");
```
Arjentix, line 170: Can you add here information about `$enum_type`?

Arjentix (changes requested):

appetrosyan (approved):

data_model/src/asset.rs
```diff
@@ -150,21 +150,31 @@ pub enum AssetValueType {
...
+                $(
+                    easy_from_str_impl!{eval_to value_type, $enum_type, $enum_value}
+                )+
+                return Err(concat!("Unknown variant for type ", stringify!($enum_type)));
+            }
+        }
+    };
 }
 
+easy_from_str_impl! {AssetValueType, Quantity, BigQuantity, Fixed, Store}
```
Arjentix, line 176: Just to mention for future macros. I think that's not a good tradition in Iroha to write such simple to implement but hard to read syntax.

Take a look at this [rule](https://rust-lang.github.io/api-guidelines/macros.html):

> Rust macros let you dream up practically whatever input syntax you want. Aim to keep input syntax familiar and cohesive with the rest of your users' code by mirroring existing Rust syntax where possible. Pay attention to the choice and placement of keywords and punctuation.

Arjentix (approved):

Issue #1983, opened 2022-03-18
Flattening Boxed structures general recommendations issue.


appetrosyan:
# Overview

The current implementation of `Encode`-able and `Decode`-able structures, namely `Instruction`, `Query`, `EventFilter` implementations suffer from having many associated boxed types like `RegisterBox`, `IdentifiableBox` etc. 

These pollute the code base. Moreover, the APIs become far too verbose to be used by anyone who's not a Rust developer. Moreover, the tagged-union-based `enum` implementation is very memory inefficient. For example, the `enum Value` is 120 Bytes. if we store a single `U32` value, we get only 12 bytes of useful data (4 bytes in the value, 8 for the tag), and waste approximately 90% of the RAM space. The percentage is higher, once we factor in that, a `Value` can only be one of 12 possible variants, an entropy of 4 bits, and not 64. 

## Problem

To avoid wasting memory, we need a way to
1. Safely encode/decode generic types directly into SCALE.
2. Store sets of heterogeneous generic types satisfying the same trait compactly. 
3. Avoid the Y2k bug. 

The 3-rd point is illustrated as follows. Suppose we wanted to store the following structure. 
```rust 
AssetFilter::new(BySome(IdFilter(tea.clone())), BySome(AssetEventFilter::ByRemoved));
```
This type has 1 bit of entropy for each `BySome` binary choice; 4 distinct combinations. Moreover, `ByRemoved` is only one of 6 variants, i.e. 3 bits of entropy. To enumerate all possible states this value can be in, we need approximately 5 bits of data. At worst, the entire space of possibilities is spanned by a single byte. 

In Parity scale one needs three bytes: one for each `FilterOpt` and one for `AssetEventFilter`. Decoded, one uses 3 machine words, i.e. 32 bytes. In other words, we waste 97% of the available space. 

This is similar to the bug of the year 2000, when the entropy used to store 6 characters was enough to count the number of milliseconds from the beginning of Unix time. 

## Proposed solution

The proposed solution to 1 and 3 is a compact stable type enumeration for all encodeable types, grouped by main category. 

### Example: 

All instructions implement a marker trait `OpCode` that encodes the type in a single fixed width integer (likely `u16` or `u8`). Where all possible combinations of instructions and operands are enumerated consecutively in alphabetical order. For example. 
```
Register<Account> => 1,
Register<Asset> =>2, 
…
Register<Peer> => n
```

[...]

PR #2079, opened 2022-04-08
[feature] #2003: Introduce Parity Scale Decoder tool

### LINKED ISSUE #2003: Parity Scale decoder tool
We need a tool that given a binary string and a type, attempts to decode the binary string according to the type. This should be used as a reference for SDKs as well as in debugging incompatibility issues.

Arjentix:
### Description of the Change

* Added new derive-macro `DumpDecoded` and corresponding trait
* New derive is used on pretty all `data_model` types. Exceptions: those, which don't implement `Debug + Decode`, and generics
* Added `tools/parity_scale_decoder`
* Added some binary samples for testing and demo

### Issue

* Closes #2003 

### Benefits

* With this tool we can debug incompatibility issues
* Useful for SDK developers

### Possible Drawbacks

That's the tricky part... Here I use 3 different procedurals macros and 2 of them should be executed in order. Rust have no guarantees for that and that's the problem. See [here](https://stackoverflow.com/questions/52910783/is-it-possible-to-store-state-within-rusts-procedural-macros) for details. Right now it works on my machine and I think it should be fine for others too.

Order:

1. Derive macro. Used in `data_model` multiple times. Generates implementation and adds type name at the global vector
2. `generate_dump_decoded_map!()`. Used in `data_model` once. Takes names from global vector and creates global Map with `Type Name -> dump_decoded() ptr` reflection
3. `get_dump_decoded_map!()`. Used in `parity_scale_decoder`. Just a getter for global map from step 2.

This approach allows not to hardcode every type by hand inside `parity_scale_decoder`... At least while it's working. **If sometime we will have a trouble with it, we can rewrite it using the less (or more?) elegant way, see the last section.**

### Usage Examples or Tests

See README.md

#### Hardcode types-map by hand
Pros:
  * Safe
  * Clear what types are presented
 
Cons:
  * Inflexible
  * A lot of code

#### Use `include!(data_model)`

Just an idea. Not even sure it's possible

### Another thing to note

We should be care when choosing #1983 implementation so that it don't rely on macro exectuion order as this one


Arjentix:
See alternative implementation in #2080

tools/parity_scale_decoder/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    decode_by_guess(map, &bytes, writer)
+}
+
+fn decode_by_type<W: io::Write>(
+    map: &DumpDecodedMap,
+    type_id: &str,
+    bytes: &[u8],
+    writer: &mut W,
+) -> Result<()> {
+    map.get(type_id).map_or_else(
```
s8sato, line 60: Instead of `DumpDecodedMap::get` here, is it feasible to have match arms such that each of them is a `TokenStream` from the derive macro, similar to `iroha_schema::schemas`?

s8sato: This was a concept of the 2nd implementation (#2080)

Cargo.toml
```diff
@@ -13,6 +13,8 @@ members = [
   "core/test_network",
   "crypto",
   "tools/crypto_cli",
+  "tools/kura_inspector",
+  "tools/parity_scale_decoder",
   "data_model",
```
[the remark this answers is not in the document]
Arjentix: Oh, no, just move it to the first *tool* in the list. I'll make it alphabetical

macro/derive/src/dump_decoded.rs
```diff
@@ -0,0 +1,170 @@
+//! Module with `DumpDecoded` macro implementation
+
+use super::*;
+
+#[cfg(feature = "dump_decoded")]
+static mut TYPES: Option<Vec<String>> = None;
```
mversic, line 6: why isn't it just an empty vector as a default value?

mversic (commented):
just one question out of curiosity since I prefer the 2nd approach. We cannot have code depending on macro expansion order

appetrosyan (changes requested):
Not a huge fan of this implementation. It's significantly longer, harder to maintain, and I managed to reproduce the instability related to macro expansion order. 

I don't think we should consider this approach at all.

tools/parity_scale_decoder/src/main.rs
s8sato: This requires manual listing of types so would be one of options for the 2nd implementation (#2080)

macro/derive/src/dump_decoded.rs
Arjentix: You are right, it can be just empty vector. I didn't notice that `Vec::new()` is *const*

Arjentix:
All of codeowners (including me) decided that #2080  implementation is better. I'm closing this PR

PR #2130, opened 2022-04-21
[feature] #2121: Check keypair is valid when constructed

### LINKED ISSUE #2121: `KeyPair::new` invariant check
When constructing a new `KeyPair` out of public and private key, it is not verified that the keys actually make a pair, i.e. the given public key doesn't have to correspond with the given private key. The `KeyPair::new` method should check that this invariant is maintained by the `KeyPair` container. I would suggest that this can be done by signing some dummy payload with the private key and verifying the signature with the public key. Performance of this check is not of big consideration because this constructor is rarely used and most often cached in the application

mversic:

### Description of the Change

I thought it might be useful to fail fast on invalid configuration. Should save us some debugging time. The gist of the change is in the `KeyPair::new` method
Closes #2121

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -123,13 +123,15 @@ impl Configuration {
         let mut configuration: Configuration = serde_json::from_reader(reader).wrap_err(
             format!("Failed to parse {:?} as Iroha peer configuration.", path),
         )?;
-        configuration.finalize();
+        configuration.finalize()?;
         Ok(configuration)
     }
 
-    fn finalize(&mut self) {
-        self.sumeragi.key_pair = self.key_pair();
+    fn finalize(&mut self) -> Result<()> {
+        self.sumeragi.key_pair = KeyPair::new(self.public_key.clone(), self.private_key.clone())?;
         self.sumeragi.peer_id = PeerId::new(&self.torii.p2p_addr, &self.public_key.clone());
+
+        Ok(())
     }
 
     /// Loads configuration from environment
@@ -139,13 +141,8 @@ impl Configuration {
     /// - Configuration `TrustedPeers` contains entries with duplicate public keys
     pub fn load_environment(&mut self) -> Result<()> {
         iroha_config::Configurable::load_environment(self)?;
-        self.finalize();
+        self.finalize()?;
         Ok(())
-    }
-
-    /// Get `public_key` and `private_key` configuration parameters.
-    pub fn key_pair(&self) -> iroha_crypto::KeyPair {
-        iroha_crypto::KeyPair::new(self.public_key.clone(), self.private_key.clone())
     }
 }
 
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -52,31 +52,37 @@ pub struct Client {
 /// Representation of `Iroha` client.
 impl Client {
     /// Constructor for client from configuration
-    pub fn new(configuration: &Configuration) -> Self {
+    ///
+    /// # Errors
+    /// If configuration isn't valid (e.g public/private keys don't match)
+    pub fn new(configuration: &Configuration) -> Result<Self> {
         Self::with_headers(configuration, http_client::Headers::default())
     }
 
     /// Constructor for client from configuration and headers
     ///
     /// *Authentication* header will be added, if `login` and `password` fields are presented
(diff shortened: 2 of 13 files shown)


cli/src/config.rs
```diff
@@ -139,14 +141,9 @@ impl Configuration {
...
         iroha_config::Configurable::load_environment(self)?;
-        self.finalize();
+        self.finalize()?;
         Ok(())
     }
-
-    /// Get `public_key` and `private_key` configuration parameters.
-    pub fn key_pair(&self) -> iroha_crypto::KeyPair {
-        iroha_crypto::KeyPair::new(self.public_key.clone(), self.private_key.clone())
-    }
```
mversic, line 149: do you think this method is useful in the public API?

appetrosyan: Adding it back in should be easy.

appetrosyan (dismissed):
LGTM

crypto/src/lib.rs
```diff
@@ -210,14 +210,52 @@ impl KeyPair {
         self.private_key.digest_function()
     }
 
-    /// Construct `KeyPair`
-    pub fn new(public_key: PublicKey, private_key: PrivateKey) -> Self {
+    /// Construct `KeyPair` from a matching pair of public and private key.
+    /// It is up to the user to ensure that the given keys indeed make a pair.
+    #[cfg(not(feature = "std"))]
+    pub fn new_unchecked(public_key: PublicKey, private_key: PrivateKey) -> Self {
```
Arjentix, line 216: I'd like to make it `unsafe` as `std` does for it's unchecked funtions (i.e. [`String::from_utf8_unchecked()`](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_unchecked))

crypto/src/lib.rs
```diff
@@ -210,14 +210,52 @@ impl KeyPair {
...
+                key_gen_option: Some(KeyGenOption::FromPrivateKey(private_key)),
+                algorithm,
+            })?;
+
+            if *key_pair.public_key() != public_key {
+                return Err(Error::KeyGen(String::from("Key pair mismatch")));
+            }
+
+            return Ok(key_pair);
+        }
```
Arjentix, line 245: Is this check with generating public key from secret key is possible only with this algorithms? Is it unavailable for BLS?

Arjentix: If so maybe it's better to do match here, so that when a new encryption algorithm will be added this will fail to compile without additional branch?

client/src/client.rs
```diff
@@ -52,39 +52,45 @@ pub struct Client {
...
-        Self {
+        Ok(Self {
             torii_url: configuration.torii_api_url.clone(),
             telemetry_url: configuration.torii_telemetry_url.clone(),
             transaction_limits: configuration.transaction_limits,
             key_pair: KeyPair::new(
                 configuration.public_key.clone(),
                 configuration.private_key.clone(),
-            ),
+            )?,
```
Arjentix, line 85: Maybe just store `KeyPair` in `configuration`? So that `Client` doesn't need to return `Result`?

crypto/src/lib.rs
mversic: I'd like that too, but that would be a misuse of the `unsafe`. `unsafe` means a possible violation of memory access. Either `unchecked` suffix doesn't imply unsafe in Rust std and it's good as it is, or I can rename the method to something else

client/src/client.rs
mversic: I think i's not stored as a key pair because that would affect environment variables and deserialization of configuration. I'd be happy to open an issue about this

crypto/src/lib.rs
mversic: sure, why not

client/src/client.rs
mversic: I opened an issue #2135 for it

appetrosyan: We should keep in mind that providing a private key automatically implies the keypair. The public key is an optional sanity check to keep track of which key was used in which place.

s8sato (approved):

appetrosyan (approved):

Issue #1974, opened 2022-03-16
Replace `blocks()` with `visit_blocks()` in WSV


Arjentix:
Right now `WorldStateView::blocks()` has locking behaviour if user stores somewhere an iterator from this function. This can be fixed by replacing it with `visit_blocks()` function, which looks something like this:

```rust
/// Visit and apply `f` for every block
///
/// # Errors
/// Throws up first `f` error
pub fn visit_blocks<F>(&self, f: F) -> Result<(), Error>
where
    F: Fn(&VersionedCommittedBlock) -> Result<(), Error>,
```

In this way user can't store reference to a block, cause it is passed as a function parameter and **Rust** won't allow to save references like that

Arjentix:
Now I think, that returning generic `R` instead of `Result<(), Error>` is better, cause we don't know the way user will use it

SamHSmith:
When implementing pr #2140 I didn't find that there was a particular need for something like this. If we do at a later point find that we do want this kind of api it can be implemented on the spot. If we do so in many places we can move it into the `LockedChainViewer`.

Issue #1737, opened 2021-12-17
Return references to blocks from WSV instead of cloning


mversic:
When interfacing WSV it is a good idea to return references to the blocks in the blockchain as the cloning can be quite expensive. At the moment we have `blocks_from_height` and `blocks_after_hash` which can/should both be replaced with something akin to the following:
```
    /// Returns iterator over blockchain blocks
    ///
    /// **Locking behaviour**: Holding references to blocks stored in the blockchain can induce
    /// deadlock. This limitation is imposed by the fact that blockchain is backed by [`Dashmap`]
    pub fn blocks(
        &self,
    ) -> impl Iterator<Item = impl Deref<Target = VersionedCommittedBlock> + '_> + '_ {
        self.blocks.iter()
    }
```

However, users of this function should be very careful not to store references(release them as soon as possible) to the blocks since blockchain is backed by the `dashmap` which can deadlock on insert in this case.

There may also be other methods on WSV which also return cloned blocks instead of references. Those should be modified as well to return references

appetrosyan:
COW?

mversic:
I don't think COW is necessary. Either all have to be cloned or none. If it is, I would consider it in another issue

Arjentix:
See also #1974

SamHSmith:
Sumeragi Simplification changes so much of this that it will have to be re-evaluated afterwards.

SamHSmith:
The blocks on kura has made this change.

PR #2140, opened 2022-04-23
[fix] #1737: Blockchain by reference instead of cloning.

SamHSmith:
### Description of the Change
Replace dashmap with spinlock based mutex in order to support referencing into the block chain.

### Issue

#1737 and #1974

### Benefits
Easier to debug deadlocks due to them being a lot less subtle.

### Possible Drawbacks
We might have slower performance due to locking but so far I haven't measured any significant change.


appetrosyan:
I strongly recommend running TPS benchmarks before and after these changes. I suspect that this will also significantly reduce the memory consumption in cases such as `million_accounts_tx`

core/src/wsv.rs
```diff
@@ -415,21 +429,27 @@ impl<W: WorldTrait> WorldStateView<W> {
...
     /// deadlock. This limitation is imposed by the fact that blockchain is backed by [`dashmap::DashMap`]
-    #[inline]
-    pub fn blocks(&self) -> crate::block::ChainIterator {
-        self.blocks.iter()
+    // #[inline]
+    // pub fn blocks(&self) -> crate::block::ChainIterator {
+    // self.blocks.lock().iter()
+    // }
+
+    pub fn lock_read_blocks(&self) -> LockedChainViewer {
```
mversic, line 437: I didn't think we'd be going for locking behavior, but @appetrosyan mentioned there may be some benefits.
The main question here is whether, if locking is a necessity, is it worth it in the sense that it offsets the price of cloning the elements out of a `DashMap`. Either in performance or memory usage

core/src/block.rs
```diff
@@ -95,74 +101,74 @@ impl Chain {
...
-        ChainIterator {
-            chain,
-            pos_front: 1,
-            pos_back: chain.len() as u64,
+impl<'chain> LockedChainViewer<'chain> {
+    /// Create new `LockedChainViewer` from `Chain` mutexguard.
+    pub fn new(chain_mutex_guard: MutexGuard<'chain, Chain>) -> Self {
+        Self { chain_mutex_guard }
+    }
+    /// Look for a block with the provided hash starting at the front
```
Arjentix, line 123: I'd add an empty line between this two

core/src/block.rs
```diff
@@ -95,74 +101,74 @@ impl Chain {
...
+            if self.chain_mutex_guard.blocks[index as usize]
+                .header()
+                .previous_block_hash
+                == hash
+            {
+                return Some(&self.chain_mutex_guard.blocks[index as usize..]);
+            }
+            index -= 1;
         }
         None
```
Arjentix, line 149:
```suggestion
self
    .chain_mutex_guard
    .blocks
    .iter()
    .rev()
    .position(|block| block.header().previous_block_hash == hash)
    .map(|index| &self.chain_mutex_guard.blocks[index..])
```

core/src/block.rs
```diff
@@ -95,74 +101,74 @@ impl Chain {
...
+        Some(
+            &self.chain_mutex_guard.blocks
+                [height.saturating_sub(1).try_into().expect("fits in usize")..],
+        )
     }
-
-    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
-        self.pos_back -= n as u64;
-        self.next_back()
+    /// Return a slice of all blocks in the `Chain`.
```
Arjentix, line 165: I'd add an empty line between this two

core/src/block.rs
```diff
@@ -95,74 +101,74 @@ impl Chain {
...
+    pub fn all_blocks(&self) -> &[VersionedCommittedBlock] {
+        self.chain_mutex_guard.blocks.as_slice()
+    }
+    /// Return the last block in the `Chain`.
```
Arjentix, line 169: I'd add an empty line between this two

core/src/block.rs
```diff
@@ -95,74 +101,74 @@ impl Chain {
...
+    pub fn last_block_iter(&self) -> Option<&VersionedCommittedBlock> {
```
Arjentix, line 170: I think `_iter` suffix is confusing, because this function returns `Option`, not an iterator

core/src/wsv.rs
```diff
@@ -85,7 +86,7 @@ pub struct WorldStateView<W: WorldTrait> {
     /// Configuration of World State View.
     pub config: Configuration,
     /// Blockchain.
-    blocks: Arc<Chain>,
+    blocks_mutex: Arc<Mutex<Chain>>,
```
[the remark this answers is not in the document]
Arjentix: `RwLock` allows for multiple read accesses when `Mutex` doesn't

Arjentix (changes requested):

core/src/wsv.rs
Arjentix: I think it should be possible to do read access without locking using Rust rules... But it will work only if functions that actually mutates `WorldStateView` will accept `&mut self`. Why they doesn't do that actually?

appetrosyan: In performance as of right now, it isn't. I think the largest benefit will come from stability. If the network is more stable given the same load, performance will degrade less in a larger system where each fault requires a view change.

core/src/wsv.rs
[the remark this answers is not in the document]
SamHSmith: `Mutex` is simpler. We can drop in `RWLock` in later if it is really necessary.

core/src/wsv.rs
```diff
@@ -415,21 +429,27 @@ impl<W: WorldTrait> WorldStateView<W> {
     ///
     /// **Locking behaviour**: Holding references to blocks stored in the blockchain can induce
     /// deadlock. This limitation is imposed by the fact that blockchain is backed by [`dashmap::DashMap`]
-    #[inline]
-    pub fn blocks(&self) -> crate::block::ChainIterator {
-        self.blocks.iter()
+    // #[inline]
+    // pub fn blocks(&self) -> crate::block::ChainIterator {
+    // self.blocks.lock().iter()
+    // }
```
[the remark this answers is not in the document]
SamHSmith: rookie mistake by me, thx

core/src/block.rs
SamHSmith: very good suggestion

SamHSmith: Well okay turns out this was a lot more complicated to get working than what I originally wrote so I think it's best to keep that.

core/src/block.rs
SamHSmith: yes

s8sato (commented):
One thing, I see a sign of some performance regression in [tps benchmark](https://github.com/hyperledger/iroha/tree/iroha2-dev/client/benches/tps#usage).
But I think this is not as big a deal as deadlock
[tps-4peers-80000interval_us-8192max_txs-5blocks-10samples-322afe8f_89af1542.pdf](https://github.com/hyperledger/iroha/files/8630694/tps-4peers-80000interval_us-8192max_txs-5blocks-10samples-322afe8f_89af1542.pdf)

appetrosyan:
In general I'd be careful with drawing conclusions from our graphs. We don't know how a single global lock compares to dash map for larger data sets. I've been planning to put this PR under a realistic load and finding the maximum cruising throughput (our previous design can be faster in short bursts but clog up for higher loads). 

@s8sato @SamHSmith I think we should leave this PR until after the release when we have the longevity stand to ourselves.

SamHSmith:
Sounds wise to leave it to after the release. Good performance will in the end require careful observation and optimization to *our* data and workload when deployed. A more manual approach will be easier to understand, avoid deadlocks and gives us optimization headroom.

core/src/block.rs
```diff
@@ -54,38 +54,44 @@ impl<T> From<EmptyChainHash<T>> for HashOf<T> {
...
-    pub fn iter(&self) -> ChainIterator {
-        ChainIterator::new(self)
+        assert_eq!(height, self.base_block_height + self.blocks.len() as u64);
+        self.blocks.push(block);
     }
 
     /// Latest block reference and its height.
-    pub fn latest_block(&self) -> Option<MapRef<u64, VersionedCommittedBlock>> {
-        self.blocks.get(&(self.blocks.len() as u64))
+    pub fn latest_block(&self) -> Option<&VersionedCommittedBlock> {
```
Arjentix, line 82: This comment is outdated. No more block height is returned

appetrosyan:
We will need to see what it does under the `longevity-load-rs` conditions. If it helps, I recommend merging with the regression, If not, we'd need to optimise it to reduce the performance regression to no more than 10% before merging.

core/src/block.rs
```diff
@@ -95,74 +101,76 @@ impl Chain {
...
-        #[allow(clippy::cast_possible_truncation)]
-        let count = (self.chain.len() as u64 - (self.pos_front - 1)) as usize;
-        count
-    }
-
-    fn size_hint(&self) -> (usize, Option<usize>) {
-        #[allow(clippy::cast_possible_truncation)]
-        let height = (self.chain.len() as u64 - (self.pos_front - 1)) as usize;
-        (height, Some(height))
+    /// Return a slice of all blocks in the `Chain` after and including
```
outoftardis, line 151:
```suggestion
    /// Return a slice of all blocks in the `Chain` starting with
```

core/src/block.rs
```diff
@@ -95,74 +101,76 @@ impl Chain {
...
 
-impl<'itm> Iterator for ChainIterator<'itm> {
-    type Item = MapRef<'itm, u64, VersionedCommittedBlock>;
-    fn next(&mut self) -> Option<Self::Item> {
-        if !self.is_exhausted() {
-            let val = self.chain.blocks.get(&self.pos_front);
-            self.pos_front += 1;
-            return val;
+    /// Look for a block with the provided hash starting at the front
+    /// of the chain going backward. If it is found return a slice
```
outoftardis, line 125:
```suggestion
    /// of the chain going backward. If the hash is found, return a slice
```

core/src/block.rs
```diff
@@ -95,74 +101,76 @@ impl Chain {
...
-    pos_front: u64,
-    pos_back: u64,
+/// Blockchain viewer.
+/// Keeps an internal lock on the blockchain in order
+/// for you to read it. Allows only read operations on
+/// the chain and will block everyone else from operating
+/// on the chain. Works across async and non async threads.
+///
+/// Try not to spam the lock, aka acquiring frequently, more
+/// than needed. If you want to keep long term access to information
```
outoftardis, line 111:
```suggestion
/// than needed. If you want to keep long-term access to information,
```

core/src/block.rs
```diff
@@ -95,74 +101,76 @@ impl Chain {
...
-/// Chain iterator
-pub struct ChainIterator<'itm> {
-    chain: &'itm Chain,
-    pos_front: u64,
-    pos_back: u64,
+/// Blockchain viewer.
+/// Keeps an internal lock on the blockchain in order
+/// for you to read it. Allows only read operations on
+/// the chain and will block everyone else from operating
+/// on the chain. Works across async and non async threads.
```
outoftardis, line 108:
```suggestion
/// on the chain. Works across async and non-async threads.
```

core/src/block.rs
```diff
@@ -59,38 +59,44 @@ impl<T> From<EmptyChainHash<T>> for HashOf<T> {
...
+            blocks: Vec::new(),
+            base_block_height: 1,
         }
     }
 
     /// Push latest block.
-    pub fn push(&self, block: VersionedCommittedBlock) {
+    /// # Panics
+    /// Panics when the block header's height does not match the
+    /// expected height of the next block.
```
outoftardis, line 79:
```suggestion
    /// Panics when the height of the block header does not match the
    /// expected height of the next block.
```

core/src/block.rs
```diff
@@ -100,74 +106,76 @@ impl Chain {
...
 
-impl<'itm> Iterator for ChainIterator<'itm> {
-    type Item = MapRef<'itm, u64, VersionedCommittedBlock>;
-    fn next(&mut self) -> Option<Self::Item> {
-        if !self.is_exhausted() {
-            let val = self.chain.blocks.get(&self.pos_front);
-            self.pos_front += 1;
-            return val;
+    /// Look for a block with the provided hash starting at the front
+    /// of the chain going backward. If the hash is found, return a slice
```
outoftardis, line 130:
```suggestion
    /// of the chain and going backwards. If the hash is found, return a slice
```

core/src/block.rs
```diff
@@ -100,74 +106,76 @@ impl Chain {
...
+/// Blockchain viewer.
+/// Keeps an internal lock on the blockchain in order
+/// for you to read it. Allows only read operations on
+/// the chain and will block everyone else from operating
+/// on the chain. Works across async and non-async threads.
+///
+/// Try not to spam the lock, aka acquiring frequently, more
+/// than needed. If you want to keep long-term access to information,
+/// consider copying it out of the chain and dropping the
+/// `LockedChainViewer`
```
outoftardis, line 118:
```suggestion
/// `LockedChainViewer`.
```

core/src/block.rs
```diff
@@ -100,74 +106,76 @@ impl Chain {
...
-    pos_front: u64,
-    pos_back: u64,
+/// Blockchain viewer.
+/// Keeps an internal lock on the blockchain in order
+/// for you to read it. Allows only read operations on
+/// the chain and will block everyone else from operating
+/// on the chain. Works across async and non-async threads.
+///
+/// Try not to spam the lock, aka acquiring frequently, more
+/// than needed. If you want to keep long-term access to information,
```
outoftardis, line 116:
```suggestion
/// Try not to use the `LockedChainViewer` too frequently, only lock the blockchain when you
/// need it. If you want to keep long-term access to information,
```

Issue #2161, opened 2022-04-28
`data_model` FFI bindings


mversic:
Generate FFI functions for accessing `data_model` structures behind an opaque pointer

PR #2164, opened 2022-04-28
[feature] #2161: generate FFI functions for `data_model`

mversic:


core/wasm/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// Forwards the call to the `host`
+///
+/// # Safety
+///
+/// Provided pointer and length must comprise a valid slice
+#[no_mangle]
+pub unsafe extern "C" fn dbg(ptr: *const u8, len: usize) {
+    host::dbg(ptr, len)
+}
```
mversic, line 90: the three functions above are just temporary placeholders. In the future smart contract will talk to host only through this module but there will be more meat in these functions. Namely, they'll convert opaque pointers received from the smart contract into scale encoded byte slice before passing them to the host for execution

wasm/src/lib.rs
```diff
@@ -179,14 +164,12 @@ pub unsafe fn _decode_from_raw<T: Decode>(ptr: WasmUsize, len: WasmUsize) -> T {
...
-#[allow(clippy::cast_possible_truncation)]
 unsafe fn _decode_from_raw_in_range<T: Decode>(
-    ptr: WasmUsize,
-    len: WasmUsize,
+    ptr: *const u8,
+    len: usize,
     range: RangeFrom<usize>,
 ) -> T {
-    let bytes = Box::from_raw(core::slice::from_raw_parts_mut(ptr as *mut _, len as usize));
+    let bytes = Box::from_raw(core::slice::from_raw_parts_mut(ptr as *mut _, len));
```
mversic, line 172: this is potentially UB, though I think it's not

appetrosyan: While I'd err on the side of caution, I think you're right in it not being intrinsically UB.

appetrosyan: If you really want to be pedantic, maybe add an `assert`, that checks for conversion errors.

core/wasm/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "iroha_core_wasm"
+version = "0.1.0"
```
appetrosyan, line 3:
```suggestion
version = "2.0.0-pre-rc.3"
```

core/wasm/build.rs
```diff
@@ -0,0 +1,5 @@
+//! Build script
+
+fn main() {
+    println!("cargo:rustc-link-arg=--export-table");
```
appetrosyan, line 4: Comment?

wasm/README.md
```diff
@@ -1,11 +1,25 @@
 # Iroha WASM
 
-The library crate that is used for writing Iroha-compliant smartcontracts in Rust using the WebAssembly format. 
+Library crate which is used for writing Iroha-compliant smartcontracts in Rust using the WebAssembly format.
 
 ## Usage
 
 Please follow the [WASM section of our tutorial](https://hyperledger.github.io/iroha-2-docs/guide/advanced/intro.html#wasm) for a detailed guide
 
+## How to build smart contract?
```
appetrosyan, line 9: Needs to be synchronised to the docs repo.

PR #2183, opened 2022-05-06
[feature] #1883: Remove embedded configuration files.

### LINKED ISSUE #1883: Remove sample configs from the Docker container
> Aleksandr Petrosyan, [2022/02/08 15:19]
> Another point that Bogdan raised is that we should stop including default configuration files inside containers. People boot up the container, see no error messages, assume everything is fine, when it isn't.
> 
> S Sato, [2022/02/08 16:30]
> [In reply to Aleksandr Petrosyan]
> The following would be one of the solutions:
> - Convert config.json to sample.env
> - Specify env_file: peer0.env .. for each peer in docker-compose.yml
> - Direct users to edit and rename sample.env and use docker compose
> https://docs.docker.com/compose/environment-variables/
> 
> S Sato, [2022/02/08 16:42]
> [In reply to S Sato]
> Not so much difference from directly writing environment section as current though
> 
> Aleksandr Petrosyan, [2022/02/08 16:42]
> [In reply to S Sato]
> This is a simplification, however.
> 
> Aleksandr Petrosyan, [2022/02/08 16:43]
> If we can reduce the number of places where Iroha can be configured to just one, I don’t see an issue.
> 
> S Sato, [2022/02/08 16:57]
> [In reply to Aleksandr Petrosyan]
> Yes, at least we can merge peer/config.json into environment variables,
> but instead, iroha_config will have to have dotenv

appetrosyan:

### Description of the Change

Removed sample configurations from the docker file. Adjusted docker compose to work from the root of the repository clone. 

### Issue

Closes #1883 

### Benefits

No problems with unspecified or incorrectly specified configuraion/genesis. 

### Possible Drawbacks

More involved process of running an iroha peer. 

```
docker compose up
```

Inside existing docker compose fields, please add a line with the following: 

```
volumes: 
 - <path-to-config-and-genesis>:config
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,8 +47,8 @@ COPY . .
 RUN cargo build $PROFILE --workspace
 
 FROM $BASE_IMAGE
-COPY configs/peer/config.json .
-COPY configs/peer/genesis.json .
+ARG CONFIG_DIR=config
+RUN mkdir -p $CONFIG_DIR
 ARG BIN=iroha
 ARG TARGET_DIR=debug
 COPY --from=builder /iroha/target/$TARGET_DIR/$BIN .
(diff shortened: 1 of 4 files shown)


mversic (approved):
Nice

PR #2185, opened 2022-05-09
[feature] #2186: Add transfer instructions for all asset types

### LINKED ISSUE #2036: Investigate odd logging behaviour
# MWE
### Add logging statements to an ISI
image

### Expected behaviour 
``` WARN: Non-mintable ```

### Actual output 

image

### LINKED ISSUE #2149: Expose `Mintable` from `iroha_data_model::prelude`

### LINKED ISSUE #2186: Allow transfers of more than just `Quantitites`
Currently only things that can be transferred are `Value::U32`. We want to support a bit more. 
- [ ] BigQuantity
- [ ] Fixed

### PR DESCRIPTION
**Pull request by appetrosyan**

### Description of the Change

- Expose `Mintable` and `MintabilityError` in `prelude`. 
- Replace `#[log]` annotations with explicit trace instructions. 
- Add `Transfer` implementations for `Asset`s of `Fixed` and `u128` types. 
- Refactor `Error` implementations. 
- Refactor `data_model`. 

### Issue

Closes #2149
Closes #2036
Closes #2186 

### Benefits

Smaller code.

### Drawbacks

None.

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -342,12 +342,12 @@ impl Client {
     ///
     /// # Errors
     /// Fails if sending transaction to peer fails or if it response with error
-    #[log]
     pub fn submit(
         &mut self,
         instruction: impl Into<Instruction> + Debug,
     ) -> Result<HashOf<VersionedTransaction>> {
-        self.submit_all(vec![instruction.into()])
+        let isi = instruction.into();
+        self.submit_all([isi])
     }
 
     /// Instructions API entry point. Submits several Iroha Special Instructions to `Iroha` peers.
@@ -368,13 +368,12 @@ impl Client {
     ///
     /// # Errors
     /// Fails if sending transaction to peer fails or if it response with error
-    #[log]
     pub fn submit_with_metadata(
         &mut self,
         instruction: Instruction,
         metadata: UnlimitedMetadata,
     ) -> Result<HashOf<VersionedTransaction>> {
-        self.submit_all_with_metadata(vec![instruction], metadata)
+        self.submit_all_with_metadata([instruction], metadata)
     }
 
     /// Instructions API entry point. Submits several Iroha Special Instructions to `Iroha` peers.
@@ -400,6 +399,7 @@ impl Client {
         &self,
         transaction: Transaction,
     ) -> Result<HashOf<VersionedTransaction>> {
+        iroha_logger::trace!(tx=?transaction);
         let (req, hash, resp_handler) =
             self.prepare_transaction_request::<DefaultRequestBuilder>(transaction)?;
         let response = req
@@ -611,7 +611,6 @@ impl Client {
     ///
     /// # Errors
     /// Fails if sending request fails
-    #[log]
     pub fn request_with_pagination<R>(
         &self,
         request: R,
@@ -621,6 +620,7 @@ impl Client {
         R: Query + Into<QueryBox> + Debug,
         <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>,
     {
+        iroha_logger::trace!(?request, %pagination);
         let (req, resp_handler) =
             self.prepare_query_request::<R, DefaultRequestBuilder>(request, pagination)?;
         let response = req.send()?;
@@ -631,7 +631,6 @@ impl Client {
     ///
     /// # Errors
     /// Fails if sending request fails
-    #[log]
     pub fn request<R>(&self, request: R) -> QueryHandlerResult<R::Output>
     where
         R: Query + Into<QueryBox> + Debug,
@@ -649,6 +648,7 @@ impl Client {
         &mut self,
         event_filter: FilterBox,
     ) -> Result<impl Iterator<Item = Result<Event>>> {
+        iroha_logger::trace!(?event_filter);
         EventIterator::new(
             &format!("{}/{}", &self.torii_url, uri::SUBSCRIPTION),
             event_filter,
@@ -793,6 +793,7 @@ impl Client {
     }
 
     /// Gets network status seen from the peer
+    ///
     /// # Errors
     /// Fails if sending request or decoding fails
     pub fn get_status(&self) -> Result<Status> {
--- a/client/tests/integration/transfer_asset.rs
+++ b/client/tests/integration/transfer_asset.rs
@@ -10,7 +10,46 @@ use test_network::{Peer as TestPeer, *};
 use super::Configuration;
 
 #[test]
-fn client_can_transfer_asset_to_another_account() {
+fn simulate_transfer_quantity() {
+    simulate_transfer(200_u32, 20_u32, |id| AssetDefinition::quantity(id).build())
+}
+
+#[test]
+fn simulate_transfer_big_quantity() {
+    simulate_transfer(200_u128, 20_u128, |id| {
+        AssetDefinition::big_quantity(id).build()
+    })
+}
+
+#[test]
+fn simulate_transfer_fixed() {
+    simulate_transfer(
+        Fixed::try_from(200_f64).expect("Valid"),
+        Fixed::try_from(20_f64).expect("Valid"),
+        |id| AssetDefinition::fixed(id).build(),
+    )
+}
+
+#[should_panic]
+#[test]
+#[ignore = "long"]
+fn simulate_insufficient_funds() {
+    simulate_transfer(
+        Fixed::try_from(20_f64).expect("Valid"),
+        Fixed::try_from(200_f64).expect("Valid"),
+        |id| AssetDefinition::fixed(id).build(),
+    )
+}
+
+// TODO add tests when the transfer uses the wrong AssetId.
+
+fn simulate_transfer<T: Into<AssetValue> + Clone, D: FnOnce(AssetDefinitionId) -> AssetDefinition>(
+    starting_amount: T,
+    amount_to_transfer: T,
+    value_type: D,
+) where
+    Value: From<T>,
+{
     let (_rt, _peer, mut iroha_client) = <TestPeer>::start_test_with_runtime();
     wait_for_genesis_committed(&vec![iroha_client.clone()], 0);
     let pipeline_time = Configuration::pipeline_time();
@@ -27,11 +66,9 @@ fn client_can_transfer_asset_to_another_account() {
     let create_account1 = RegisterBox::new(Account::new(account1_id.clone(), [public_key1]));
     let create_account2 = RegisterBox::new(Account::new(account2_id.clone(), [public_key2]));
     let asset_definition_id: AssetDefinitionId = "xor#domain".parse().expect("Valid");
-    let quantity: u32 = 200;
-    let create_asset =
-        RegisterBox::new(AssetDefinition::quantity(asset_definition_id.clone()).build());
+    let create_asset = RegisterBox::new(value_type(asset_definition_id.clone()));
     let mint_asset = MintBox::new(
-        Value::U32(quantity),
+        Value::from(starting_amount),
         IdBox::AssetId(AssetId::new(
             asset_definition_id.clone(),
             account1_id.clone(),
@@ -51,10 +88,9 @@ fn client_can_transfer_asset_to_another_account() {
     thread::sleep(pipeline_time * 2);
 
     //When
-    let quantity = 20;
     let transfer_asset = TransferBox::new(
         IdBox::AssetId(AssetId::new(asset_definition_id.clone(), account1_id)),
-        Value::U32(quantity),
+        Value::from(amount_to_transfer.clone()),
         IdBox::AssetId(AssetId::new(
             asset_definition_id.clone(),
             account2_id.clone(),
@@ -67,7 +103,7 @@ fn client_can_transfer_asset_to_another_account() {
             |result| {
                 result.iter().any(|asset| {
                     asset.id().definition_id == asset_definition_id
-                        && *asset.value() == AssetValue::Quantity(quantity)
+                        && *asset.value() == amount_to_transfer.clone().into()
                         && asset.id().account_id == account2_id
                 })
             },
--- a/client/tests/integration/triggers/time_trigger.rs
+++ b/client/tests/integration/triggers/time_trigger.rs
@@ -170,6 +170,7 @@ fn pre_commit_trigger_should_be_executed() -> Result<()> {
 }
 
 #[test]
+#[ignore = "Only on nightly"]
 fn mint_nft_for_every_user_every_1_sec() -> Result<()> {
     const TRIGGER_PERIOD_MS: u64 = 1000;
     const EXPECTED_COUNT: u64 = 4;
--- a/core/src/smartcontracts/isi/account.rs
+++ b/core/src/smartcontracts/isi/account.rs
@@ -240,13 +240,11 @@ pub mod isi {
 pub mod query {
 
     use eyre::{Result, WrapErr};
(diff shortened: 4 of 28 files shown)

### CODE REVIEW

**Review by Arjentix — DISMISSED** [review]

File: data_model/src/events/pipeline.rs [hunk 1]
```diff
@@ -61,7 +61,7 @@ impl EventFilter {
         self
     }
 
-    /// Filter by [`Hash`].
+    /// Filter by [`struct@Hash`].
```
mversic, line 64: what does this mean?

flakyness.org
```diff
@@ -0,0 +1,7 @@
+* Flaky tests
```
mversic, line 1: why do we have this file?

docs/source/references/config.md
```diff
@@ -1,858 +0,0 @@
-# Iroha Configuration reference
```
mversic, line 1: what happened here?

mversic (approved):

data_model/src/events/pipeline.rs
appetrosyan: in `cargo doc` whenever you have an ambiguous name, e.g. both a trait and a struct with the same identifier, you can disambiguate the two using the @ symbol. 

If you run cargo doc locally, it will suggest you to do that. the alternative is fully qualifying the path, but IMO docs should almost never do that.

flakyness.org
appetrosyan: Useful for people to know which checks need to be re-run and which - not. i'll remove it once we have `cargo nextest`.

data_model/src/pagination.rs
```diff
@@ -83,6 +83,17 @@ pub struct Pagination {
...
 }
 
+impl fmt::Display for Pagination {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let start = self.start.unwrap_or(0);
+        let end = self
+            .limit
+            .map(|n| n.to_string())
+            .unwrap_or(".inf".to_owned());
+        write!(f, "{}-{}", start, end)
```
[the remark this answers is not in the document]
appetrosyan: If we did that, the trace log 
`iroha_logger::trace!(%pagination)`. 
would contain "pagination= Pagination(XX-YY)". I think we want the option of having both less noise (with Display) and more standard data (Debug).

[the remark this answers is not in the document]
mversic: ok

mversic (approved):

client/src/client.rs
```diff
@@ -342,12 +342,12 @@ impl Client {
     ///
     /// # Errors
     /// Fails if sending transaction to peer fails or if it response with error
-    #[log]
```
mversic, line 345: was this removed due to some deadlock issues with it?

appetrosyan: Yes actually. This PR started by me investigating the problem that @Arjentix idenitified. It turned out to be related to a few others.

Arjentix: @appetrosyan have someone else also found this bug with `#[log]`? Am I not a mad man? Cause deadlock in span creating process is really wierd.
If so, maybe we need to spend some time creating issue for the [`tracing` repo](https://github.com/tokio-rs/tracing)? Or search for already openned one

client/src/client.rs
```diff
@@ -649,6 +648,7 @@ impl Client {
         &mut self,
         event_filter: FilterBox,
     ) -> Result<impl Iterator<Item = Result<Event>>> {
+        iroha_logger::trace!(?event_filter);
```
Arjentix, line 651: I think, rewriting our `log` macro to do something like that will be easy, so that we shouldn't write it by hand

client/tests/integration/transfer_asset.rs
```diff
@@ -10,7 +10,46 @@ use test_network::{Peer as TestPeer, *};
...
+}
+
+// TODO add tests when the transfer uses the wrong AssetId.
+
+fn simulate_transfer<T: Into<AssetValue> + Clone, D: FnOnce(AssetDefinitionId) -> AssetDefinition>(
+    starting_amount: T,
+    amount_to_transfer: T,
+    value_type: D,
+) where
+    Value: From<T>,
```
Arjentix, line 51: Didn't know we can add trait bounds to a type not from a function declaration...

core/src/smartcontracts/isi/account.rs
```diff
@@ -262,14 +261,14 @@ pub mod query {
...
     impl<W: WorldTrait> ValidQuery<W> for FindPermissionTokensByAccountId {
-        #[log]
         #[metrics(+"find_permission_tokens_by_account_id")]
         fn execute(&self, wsv: &WorldStateView<W>) -> Result<Self::Output, Error> {
             let account_id = self
                 .id
                 .evaluate(wsv, &Context::new())
                 .wrap_err("Failed to evaluate account id")
                 .map_err(|e| Error::Evaluate(e.to_string()))?;
+            iroha_logger::trace!(%account_id, accounts=?wsv.world.domains);
```
Arjentix, line 271: Accounts or domains?

flakyness.org
Arjentix: What does `.org` extension stand for? Emacs?

core/src/smartcontracts/isi/asset.rs
```diff
@@ -275,46 +271,49 @@ pub mod isi {
         }
     }
 
+    /// Assert that the two assets have the same asset `definition_id`.
+    fn assert_matching_definitions<W: WorldTrait>(
+        source: &<Asset as Identifiable>::Id,
+        destination: &<Asset as Identifiable>::Id,
+        wsv: &WorldStateView<W>,
+        value_type: AssetValueType,
+    ) -> Result<(), Error> {
```
Arjentix, line 280: Am I the only one who thinks about assertion functions like it should panic instead of returning `Result`? I'd prefer `check_` naming but that's a personal taste

core/src/smartcontracts/isi/asset.rs
```diff
@@ -335,7 +334,89 @@ pub mod isi {
...
-                Ok(AssetEvent::Added(destination_asset_id.clone()))
+                Ok(AssetEvent::Added(self.destination_id.clone()))
+            })
+        }
+    }
+
+    impl<W: WorldTrait> Execute<W> for Transfer<Asset, u128, Asset> {
+        type Error = Error;
+
+        #[metrics(+"transfer_qty_asset")]
```
Arjentix, line 345: Shouldn't this be changed? A copy-paste mistake, I believe

core/src/smartcontracts/isi/asset.rs
```diff
@@ -335,7 +334,89 @@ pub mod isi {
...
+
+                Ok(AssetEvent::Added(self.destination_id.clone()))
+            })
+        }
+    }
+
+    impl<W: WorldTrait> Execute<W> for Transfer<Asset, Fixed, Asset> {
+        type Error = Error;
+
+        #[metrics(+"transfer_qty_asset")]
```
Arjentix, line 388: Same as above

core/src/smartcontracts/isi/asset.rs
```diff
@@ -335,7 +334,89 @@ pub mod isi {
...
+            })?;
+            wsv.modify_asset(&self.destination_id, |asset| {
+                let quantity: &mut u128 = asset
+                    .try_as_mut()
+                    .map_err(eyre::Error::from)
+                    .map_err(|e| Error::Conversion(e.to_string()))?;
+                *quantity = quantity
+                    .checked_add(self.object)
+                    .ok_or(MathError::Overflow)?;
+                // wsv.metrics.tx_amounts.observe(f64::from(*quantity));
```
Arjentix, line 378: Commented code

core/src/smartcontracts/isi/asset.rs
```diff
@@ -335,7 +334,89 @@ pub mod isi {
                     .ok_or(MathError::Overflow)?;
                 wsv.metrics.tx_amounts.observe(f64::from(*quantity));
 
-                Ok(AssetEvent::Added(destination_asset_id.clone()))
+                Ok(AssetEvent::Added(self.destination_id.clone()))
+            })
+        }
+    }
+
+    impl<W: WorldTrait> Execute<W> for Transfer<Asset, u128, Asset> {
```
Arjentix, line 342: Transfer functions here look like it can be implemented with a generic function

core/src/smartcontracts/isi/mod.rs
```diff
@@ -171,47 +151,25 @@ pub mod error {
...
-    /// Asset type assertion error
-    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-    pub struct AssetTypeError {
-        /// Expected type
-        pub expected: AssetValueType,
-        /// Type which was discovered
-        pub got: AssetValueType,
+    #[derive(Debug, Clone, PartialEq, Eq, Error)]
+    #[error("Expected {expected:?},  actual {actual:?}")]
+    pub struct Mismatch<T> {
```
Arjentix, line 156: Docs needed

Arjentix: Shouldn't this structure be presented in the `Error` enum above?

Arjentix: Ah, it's presented inside `TypeError`, I see

client/src/client.rs
appetrosyan: You're definitely not going insane. Like i said, the likely culprit is the locking that happens with the global mutable state. I found it as a culprit for a different problem, though.

appetrosyan: I would recommend you create the issue, and get in touch with them via the Discord, I only managed to reproduce the issue by running `cargo flaky` and seeing a reduction in failed test cases (e.g. `7_peers_1_fault`), you might have a better MWE.

client/src/client.rs
appetrosyan: Fine as is for now. You're not saving a line of code either, since all I do is list the arguments that need to be traced (there aren't that many, and you can't do this automatically, since large objects take up valuable CPU time).

core/src/smartcontracts/isi/account.rs
appetrosyan: Good catch.

core/src/smartcontracts/isi/asset.rs
appetrosyan: Yes.

core/src/smartcontracts/isi/asset.rs
appetrosyan: I'm still thinking on the best way to do it. `u128` can be represented by `f64`, since `f64` uses scientific notation, has an 11 bit exponent and we don't care about precision. I don't want to hand-roll an implementation and see if there's a crate that has this conversion. If not, just right-shift by 64 bits, turn `u128 → u64` convert to `f64` add `64` to the exponent.

core/src/smartcontracts/isi/asset.rs
appetrosyan: So can be minting/burning. Do you prefer a marker trait for asset values, or explicit trait bounds on all objects?

core/src/smartcontracts/isi/asset.rs
appetrosyan: No a panic in this case doesn't signal the Iroha programmer's error it signals that the person who requested the ISI didn't consider the state of the blockchain. Since panics are not forwarded to clients as error messages, these `assert_` methods need to return a `Result`. 

`check_` and `assert_` have [different connotations](https://artoftesting.com/difference-between-assert-and-verify). 

> We use Assert when we have to validate critical functionality. If the feature fails then this makes the execution of further statements irrelevant. Hence, the test method is aborted as soon as failure occurs.

We cannot handle a `tx` that has the wrong type thinking it's the right type, we have no choice but to abort the execution. Same with `source` and `destination`. 

A check is usually something that can be recovered from: i.e. a `None` that can be replaced with `Some(Default::default())`.

core/src/smartcontracts/isi/asset.rs
appetrosyan: https://docs.rs/floatconv/0.2.8/floatconv/fast/index.html

appetrosyan: Nevermind. We don't need the extra precision so `*quantity as f64` works just as well.

core/src/smartcontracts/isi/asset.rs
appetrosyan: Created #2204 .

core/src/smartcontracts/isi/asset.rs
Arjentix: Makes sense

core/src/smartcontracts/isi/asset.rs
Arjentix: Hmm, let's an author of #2204 changes will decide the best apprroach

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -170,6 +170,7 @@ fn pre_commit_trigger_should_be_executed() -> Result<()> {
 }
 
 #[test]
+#[ignore = "Only on nightly"]
```
mversic, line 173:
```suggestion
#[ignore = "nightly"]
```
what is the meaning of the string attached?

data_model/src/lib.rs
```diff
@@ -659,11 +659,6 @@ where
     }
 }
 
-/// Marker trait for values.
-pub trait ValueMarker: Debug + Clone + Into<Value> {}
```
mversic, line 663: I'm glad we got rid of `ValueMarker`

mversic (approved):

core/src/smartcontracts/isi/asset.rs
```diff
@@ -275,46 +271,49 @@ pub mod isi {
...
     }
 
+    /// Assert that the two assets have the same asset `definition_id`.
+    fn assert_matching_definitions<W: WorldTrait>(
+        source: &<Asset as Identifiable>::Id,
+        destination: &<Asset as Identifiable>::Id,
+        wsv: &WorldStateView<W>,
+        value_type: AssetValueType,
+    ) -> Result<(), Error> {
+        if destination.definition_id != source.definition_id {
```
mversic, line 281: so the limitation is that it's not possible to transfer assets across domains? and they have to have the same name

client/tests/integration/triggers/time_trigger.rs
appetrosyan: If you run `cargo test` on 1.62+ (nightly) you'll get the string `ignored, nightly`, so people running the test know why it's ignored.

0x009922:
I think it makes sense to link #2149 as an issue to be resolved by this PR.

flakyness.org
```diff
@@ -0,0 +1,7 @@
...
+
+** test: integration::triggers::event_trigger::test_mint_asset_when_new_asset_definition_created (25/100)
+
+** test: multiple_networks stdout (1/100)
+
+** test: integration::roles::add_role_to_limit_transfer_count (6/100)
```
Arjentix, line 7: After merging of #2197 the results should be different. Can you check it one more time?

Arjentix (approved):

flakyness.org
appetrosyan: Created #2226. This process takes ~2 hours, so I will do this in a separate PR.

Issue #2118, opened 2022-04-18
Rethink `Default` configuration


mversic:
# Description
At the moment when deserializing config files into config structures if the field is missing the field will be taken from the `Default` implementation of the structure. E.g. for `cli::Configuration` if `private_key` is missing in the `config.json` file and is not defined as a `IROHA_PUBLIC_KEY` env variable, it's be taken from the `cli::Configuration::Default`. Needless to say, this a big potential security issue. In this case a more prudent course of action would be fail starting Iroha.

We can also say this happens because we implement `Default` for structures which don't have reasonable defaults

# Cause
This happens because configuration structures have `serde(default)` attached to them

# Solution
1. Remove `serde(default)`. This is a hazzard waiting to happen. However, if the annotation is removed, deserialization from a file would fail even if env variable defines the value. This suggests that deserializing from file and from environment have to be combined into a single step - they cannot be independent steps. 
2. Maybe it would be possible to use `serde(default = "path")` to force it to try to deserialize from environment if deserialization from file fails. It should be investigated which kinds of fallbacks serde offers
3. Consider removing `Default` implementation from configurations which don't have a sensible default such as `cli::Configuration` and rather rely on `serde(default = "path")` (or implement custom `config(default = "path")` attached on a field, not a whole container. This way default values will be defined only for fields which have sensible defaults.

mversic:
@appetrosyan suggests:
> Have you considered transparent wrappers with unique impl Default? Adding the delineation at the type level as well as in documentation?

mversic:
I would say, let's remove implementation of `Default` for `cli::Configuration`, `client::Configuration` and `sumeragi::Configuration` (because there isn't a sensible default, at least for the private key) and use `#[serde(default)]` attached to fields which have a default

mversic:
For structures which don't have a sensible default because of a minor number of fields (as is the case of `client::Configuration` it may be cumbersome to construct them if all you need to do is set one field and use the defaults for all others. I'm not sure anyone would be constructing `client::Configuration` by hand instead of deserializing it from a file, but if that happens to be the case I would suggest a builder is added for such structures. E.g:

```
#[derive(Debug, Clone, Deserialize, Serialize, Configurable)]
#[serde(rename_all = "UPPERCASE")]
#[config(env_prefix = "IROHA_")]
pub struct Configuration {
    /// Public key of the user account.
    #[config(serde_as_str)]
    pub public_key: PublicKey,
    /// Private key of the user account.
[...]

Issue #2101, opened 2022-04-13
`KeyPair` defaults


mversic:
After #2077 is merged implementation of the `KeyPair::Default` will contain empty vectors for public/private key pair payloads as can be seen in the implementation. I think we need better defaults. Let's agree on a specific valid public/private key pair that we'll use as a default. Also, maybe coordinate with other teams and have this keypair exposed in documentation

appetrosyan:
The de-facto standard is the key pair we use in `genesis` and `config`. 
```
ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0
```

appetrosyan:
Perhaps a similar trait that is not `Default` but acts like one.

appetrosyan:
Addressed in #2188

PR #2188, opened 2022-05-10
[fix] #1990: Enable peer startup via env vars in the absence of `config.json`

### LINKED ISSUE #1990: Support running Iroha without `config.json` using pure environment variables.

ilchu:
Signed-off-by: Ilia Churin <ilia.ch@netogami.net>

### Description of the Change

In the case when the minimal cli client (`iroha`) was provided with no `config.json` either in the same directory or as an environment variable, do not stop the execution, but instead require that all the variables from the default json be provided as environment variables too, and load them onto default configuration.

### Issue

Resolves #1990.

### Benefits
Presumably easier and more flexible setup of peer through cli.

### Possible Drawbacks
As the fix chooses to fall back on default impl for config, it has the same problems that were discussed in #2101 and  #2118. The question is which variables should be allowed to be set via environment, and which subset of them should be required when using default, if we're going to stick with it at all. As of now, the default configuration is generated with `Default` and all the variables from the `configs/peer/config.json` are required to proceed if no file was found.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -77,7 +77,17 @@ cargo run --bin kagami -- crypto --json
 
 For the Iroha peer binary to run, a configuration file must be provided. Iroha will not run with defaults if the configuration file is not available.
 
-The Iroha binary looks for either a file `config.json` in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the latter environment variable is defined, but not a valid configuration file, the Iroha peer binary will exit and do nothing.
+The Iroha binary looks for either a `config.json` file in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the configuration file is not valid, the Iroha peer binary exits and does nothing. If neither of these files is provided, all the fields from the default `config.json` should be specified as environment variables. Note that environment variables override the variables in their respective fields provided via `config.json`.
+
+The environment variables replacing `config.json` should be passed as JSON strings, meaning that any inner quotes should be properly escaped in the command line as shown in the rather unwieldy example below.
+
+<details>
+
+``` bash
+IROHA_TORII="{\"P2P_ADDR\": \"127.0.0.1:1339\", \"API_URL\": \"127.0.0.1:8080\"}" IROHA_SUMERAGI="{\"TRUSTED_PEERS\": [{\"address\": \"127.0.0.1:1337\",\"public_key\": \"ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b\"},{\"address\": \"127.0.0.1:1338\",\"public_key\": \"ed0120cc25624d62896d3a0bfd8940f928dc2abf27cc57cefeb442aa96d9081aae58a1\"},{\"address\": \"127.0.0.1:1339\",\"public_key\": \"ed0120faca9e8aa83225cb4d16d67f27dd4f93fc30ffa11adc1f5c88fd5495ecc91020\"},{\"address\": \"127.0.0.1:1340\",\"public_key\": \"ed01208e351a70b6a603ed285d666b8d689b680865913ba03ce29fb7d13a166c4e7f1f\"}]}" IROHA_KURA="{\"INIT_MODE\": \"strict\",\"BLOCK_STORE_PATH\": \"./blocks\"}" IROHA_BLOCK_SYNC="{\"GOSSIP_PERIOD_MS\": 10000,\"BATCH_SIZE\": 2}" IROHA_PUBLIC_KEY="ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b" IROHA_PRIVATE_KEY="{\"digest_function\": \"ed25519\",\"payload\": \"282ed9f3cf92811c3818dbc4ae594ed59dc1a2f78e4241e31924e101d6b1fb831c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b\"}" IROHA_GENESIS="{\"ACCOUNT_PUBLIC_KEY\": \"ed01204cffd0ee429b1bdd36b3910ec570852b8bb63f18750341772fb46bc856c5caaf\",\"ACCOUNT_PRIVATE_KEY\": {\"digest_function\": \"ed25519\",\"payload\": \"d748e18ce60cb30dea3e73c9019b7af45a8d465e3d71bcc9a5ef99a008205e534cffd0ee429b1bdd36b3910ec570852b8bb63f18750341772fb46bc856c5caaf\"}}" ./iroha 
+```
+
+</details>
 
 The  [configuration options reference](../docs/source/references/config.md) provides detailed explanations of each configuration variable. All variables defined in `config.json` can be overridden with environment variables. **We don't recommend using environment variables for configuration outside docker-compose and Kubernetes deployments**. Please change the values in the configuration file instead, so that we can better debug the problems that you might be having.
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -107,7 +107,10 @@ where
         query_validator: IsQueryAllowedBoxed<K::World>,
     ) -> Result<Self> {
         let broker = Broker::new();
-        let mut config = Configuration::from_path(&args.config_path)?;
+        let mut config = match Configuration::from_path(&args.config_path) {
+            Ok(config) => config,
+            Err(_) => Configuration::default(),
+        };
         config.load_environment()?;
 
         let genesis = G::from_configuration(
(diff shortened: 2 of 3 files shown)


cli/src/lib.rs
```diff
@@ -143,7 +146,9 @@ where
         broker: Broker,
     ) -> Result<Self> {
         if !config.disable_panic_terminal_colors {
-            color_eyre::install()?;
+            if let Err(e) = color_eyre::install() {
+                iroha_logger::error!("Tried to install eyre_hook twice: {:?}", e);
```
mversic, line 150: why does this happen?

cli/src/main.rs
```diff
@@ -27,6 +27,28 @@ async fn main() -> Result<(), color_eyre::Report> {
 
     if let Ok(config_path) = std::env::var("IROHA2_CONFIG_PATH") {
         args.config_path = std::path::PathBuf::from_str(&config_path)?;
+    } else {
+        if !Path::new("config.json").exists() {
```
mversic, line 31:
```suggestion
    } 
    if !args.config_path.exists() {
```

cli/src/main.rs
```diff
@@ -58,4 +80,17 @@ fn print_help() {
     println!("    IROHA2_CONFIG_PATH is the location of your `config.json`");
     println!("    IROHA2_GENESIS_PATH is the location of `genesis.json`");
     println!("If either of these is not provided, Iroha checks the current directory.");
+    println!(
+        "Additionally, in case of absense of both IROHA2_CONFIG_PATH and `config.json`
```
mversic, line 84:
```suggestion
        "Additionally, in case of absence of both IROHA2_CONFIG_PATH and `config.json`
```

cli/README.md
```diff
@@ -77,7 +77,7 @@ cargo run --bin kagami -- crypto --json
 
 For the Iroha peer binary to run, a configuration file must be provided. Iroha will not run with defaults if the configuration file is not available.
 
-The Iroha binary looks for either a file `config.json` in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the latter environment variable is defined, but not a valid configuration file, the Iroha peer binary will exit and do nothing.
+The Iroha binary looks for either a file `config.json` in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the latter environment variable is defined, but not a valid configuration file, the Iroha peer binary will exit and do nothing. If neither of those is provided, all the fields from default `config.json` should be specified as environment variables.
```
mversic, line 80: As I see from the implementation it seems that env variable `IROHA2_CONFIG_PATH` takes precedence over `config.json` in current directory

mversic:
```suggestion
The Iroha binary looks for either a file `config.json` in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the configuration file is not valid, the Iroha peer binary will exit and do nothing. If neither of those is provided, all the fields from default `config.json` should be specified as environment variables.
```

cli/src/main.rs
mversic: this would mean that env variables will be read from if `IROHA2_CONFIG_PATH` doesn't exist or `config.json` doesn't exist

mversic (changes requested):

cli/src/lib.rs
ilchu: Was not able to trace the exact reason, but somehow with this workflow when no config was provided color_eyre kept crashing. AFAICT the only other spots where `install()` is used are `kagami` and `logger`, but they don't seem to be called at that stage. Will double check the error again, but this was a stumbling point for me.

mversic: I'm asking because I think this also happened to sdk devs when bootstraping iroha

ilchu: Then at least it's not only me, but something in the code. I guess then as a workaround this will suffice for now, and I'll create a separate issue for this later today.

appetrosyan: Kagami is a completely unrelated executable, so this is probably the only place this happens. Log and ignore for now, but we ned to figure out why this happens. IMO there's a tokio task that tries to do the same thing in a different thread and breaks the install process.

cli/README.md
```diff
@@ -77,7 +77,7 @@ cargo run --bin kagami -- crypto --json
 
 For the Iroha peer binary to run, a configuration file must be provided. Iroha will not run with defaults if the configuration file is not available.
 
-The Iroha binary looks for either a file `config.json` in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the latter environment variable is defined, but not a valid configuration file, the Iroha peer binary will exit and do nothing.
+The Iroha binary looks for either a file `config.json` in the current directory, or for a JSON file `IROHA2_CONFIG_PATH`. If the configuration file is not valid, the Iroha peer binary will exit and do nothing. If neither of those is provided, all the fields from default `config.json` should be specified as environment variables.
```
Arjentix, line 80: I'd add the information, that if a config was found and some env variables are defined, then they will override the corresponding fields

mversic (dismissed):

cli/src/main.rs
```diff
@@ -58,4 +77,17 @@ fn print_help() {
...
+as follows:"
+    );
+    println!("    IROHA_TORII is the torii gateway config");
+    println!("    IROHA_SUMERAGI is the consensus config");
+    println!("    IROHA_KURA is block storage config");
+    println!("    IROHA_BLOCK_SYNC is block synchronization config");
+    println!("    IROHA_PUBLIC_KEY is the peer's public key");
+    println!("    IROHA_PRIVATE_KEY is the peers' private key");
+    println!("    IROHA_GENESIS is the genesis block config");
+    println!("Examples of these variables can be seen in the default `configs/peer/config.json`.")
```
outoftardis, line 92:
```suggestion
    println!("Examples of these variables can be found in the default `configs/peer/config.json`.")
```

cli/src/main.rs
```diff
@@ -2,6 +2,7 @@
 
 use std::str::FromStr;
 
+use eyre::WrapErr;
```
appetrosyan, line 5:
```suggestion
use eyre::WrapErr as _;
```

cli/src/main.rs
```diff
@@ -58,4 +77,17 @@ fn print_help() {
     println!("    IROHA2_CONFIG_PATH is the location of your `config.json`");
     println!("    IROHA2_GENESIS_PATH is the location of `genesis.json`");
     println!("If either of these is not provided, Iroha checks the current directory.");
+    println!(
+        "Additionally, in case of absence of both IROHA2_CONFIG_PATH and `config.json`
+in the current directory, all the variables from `config.json` should be set via the environment
+as follows:"
+    );
```
appetrosyan, line 84: Explain that they need to passed as a JSON string. I'd add an example of that could be run from the command line in the comments/docs.

cli/README.md
```diff
@@ -77,7 +77,13 @@ cargo run --bin kagami -- crypto --json
...
...94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b\"},{\"address\": \"127.0.0.1:1338\",\"public_key\": \"ed0120cc25624d62896d3a0bfd8940f928dc2abf27cc57cefeb442aa96d9081aae58a1\"},{\"address\": \"127.0.0.1:1339\",\"public_key\": \"ed0120faca9e8aa83225cb4d16d67f27dd4f93fc30ffa11adc1f5c88fd5495ecc91020\"},{\"address\": \"127.0.0.1:1340\",\"public_key\": \"ed01208e351a70b6a603ed285d666b8d689b680865913ba03ce29fb7d13a166c4e7f1f\"}]}" IROHA_KURA="{\"INIT_MODE\": \"strict\",\"BLOCK_STORE_PATH\": \"./blocks\"}" IROHA_BLOCK_SYNC="{\"GOSSIP_PERIOD_MS\": 10000,\"BATCH_SIZE\": 2}" IROHA_PUBLIC_KEY="ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b" IROHA_PRIVATE_KEY="{\"digest_function\": \"ed25519\",\"payload\": \"282ed9f3cf92811c3818dbc4ae594ed59dc1a2f78e4241e31924e101d6b1fb831c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b\"}" IROHA_GENESIS="{\"ACCOUNT_PUBLIC_KEY\": \"ed01204cffd0ee429b1bdd36b3910ec570852b8bb63f18750341772fb46bc856c5caaf\",\"ACCOUNT_PRIVATE_KEY\": {\"digest_function\": \"ed25519\",\"payload\": \"d748e18ce60cb30dea3e73c9019b7af45a8d465e3d71bcc9a5ef99a008205e534cffd0ee429b1bdd36b3910ec570852b8bb63f18750341772fb46bc856c5caaf\"}}" ./iroha
```
appetrosyan, line 85: This is a little ugly, but this is meant to be technical documentation. Add some `<details>` tags around it and we can merge.

appetrosyan (approved):

mversic (approved):

Issue #1149, opened 2021-07-13
Limit Block Files to 1 million Per Folder


takemiyamakoto:
Block files in the blocks storage should be limited to 1 million blocks. If more than 1 million files are in a block storage folder, new folders should be created so the 1 million file limit is not violated.

PR #2194, opened 2022-05-11
[feature] #1149: Restrict block number to 1 million per dir

SamHSmith:
### Description of the Change

Take what was the previous file name, prepend with zeroes, then interpret that as dir + file name. So block files are ex 2/003, 0/032 et cetera.

### Issue

#1149 

### Benefits

Most operating systems will freak out if there are too many files in a directory. This solves that.

### Possible Drawbacks

We still iterate over all existing files to figure out which block file we want to edit/create. I think this needs to change in the long run. We should using a simple formula map from block number to block file number. That way you only touch the the file you want to operate on. That is a seperate issue from #1149 though, so the point is irrelevant for this PR.

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -311,6 +311,9 @@ pub enum Error {
     InconsequentialBlockWrite,
 }
 
+/// Maximum length for block file paths
+const FILENAME_MAX_LENGTH: usize = 6;
+
 impl<IO: DiskIO> BlockStore<IO> {
     /// Initialize block storage at `path`.
     ///
@@ -350,8 +353,8 @@ impl<IO: DiskIO> BlockStore<IO> {
     /// - Filesystem access failure (HW or permissions)
     ///
     pub async fn get_block_path(&self, block_height: NonZeroU64) -> Result<PathBuf> {
-        let filename = self.get_block_filename(block_height).await?;
-        Ok(self.path.join(filename))
+        let file_string = self.get_block_filename(block_height).await?;
+        Ok(self.filename_to_path(&file_string))
     }
 
     /// Append block to latest (or new) file on the disk.
@@ -366,12 +369,17 @@ impl<IO: DiskIO> BlockStore<IO> {
         block: &VersionedCommittedBlock,
     ) -> Result<HashOf<VersionedCommittedBlock>> {
         let height = NonZeroU64::new(block.header().height).ok_or(Error::ZeroBlock)?;
-        let path = self.get_block_path(height).await?;
+        let file_path = self.get_block_path(height).await?;
+
+        let mut dir_path = file_path.clone();
+        dir_path.pop();
+
+        fs::create_dir_all(dir_path).await?;
         let mut file = BufWriter::new(
             fs::OpenOptions::new()
                 .create(true)
                 .append(true)
-                .open(path)
+                .open(&file_path)
                 .await?,
         );
         let hash = block.hash();
@@ -421,6 +429,25 @@ impl<IO: DiskIO> BlockStore<IO> {
         }
     }
 
+    #[allow(clippy::useless_let_if_seq)]
+    fn filename_to_path(&self, file_string: &str) -> PathBuf {
+        let dir_name;
+        let file_name;
+        if file_string.len() > FILENAME_MAX_LENGTH {
+            let (a, b) = file_string.split_at(file_string.len() - FILENAME_MAX_LENGTH);
+            dir_name = a.to_owned();
+            file_name = b.to_owned();
+        } else {
+            dir_name = "0".to_owned();
+            file_name = format!(
+                "{}{}",
+                "0".repeat(FILENAME_MAX_LENGTH - file_string.len()),
+                file_string
+            );
+        }
+        self.path.join(dir_name).join(file_name)
+    }
+
     /// Returns a stream of deserialized blocks that in order of reading from sorted storage files
     ///
     /// # Errors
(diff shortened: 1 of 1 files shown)


core/src/kura.rs
```diff
@@ -838,8 +942,12 @@ mod tests {
             DefaultIO,
         )
         .await
-        .unwrap();
-        let written = fs::write(dir.path().join("1"), vec![1; 40]).await;
+            .unwrap();
+        fs::DirBuilder::new()
+            .recursive(true)
+            .create(&dir.path().join("0"))
+            .await.unwrap();
```
Arjentix, line 949: `create_dir_all()`?

core/src/kura.rs
```diff
@@ -915,21 +1023,40 @@ mod tests {
...
+                                    }
+                                }
+                            } else {
+                                let error = std::io::Error::from_raw_os_error(42);
+                                Err(error)
+                            }
+                        })
                         .collect::<Vec<_>>()
                         .into_iter(),
                 ))
```
Arjentix, line 1055: Can you simplify it, please? One-letter variables are hard to read

core/src/kura.rs
```diff
@@ -915,21 +1023,40 @@ mod tests {
...
+                }
                 Ok(futures::stream::iter(
                     self.files
                         .keys()
                         .cloned()
-                        .map(Ok)
+                        .map(|f_os| {
+                            let f = f_os.to_str();
+                            if f.is_some() && f.unwrap().starts_with(path_str) {
+                                let path_buf : PathBuf = f.unwrap().strip_prefix(path_str).unwrap().into();
```
Arjentix, line 1040: [`strip_prefix()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix) docs shows, that you don't need to add `/` sign before, it should work just fine without it. So you can remove `path_srt` manipulation above

core/src/kura.rs
```diff
@@ -349,9 +352,23 @@ impl<IO: DiskIO> BlockStore<IO> {
...
+            file_string = format!("{}/{}", a, b);
+        } else {
+            file_string = format!(
+                "{}/{}{}",
+                0,
+                "0".repeat(FILENAME_MAX_LENGTH - file_string.len()),
+                file_string
+            );
+            dir_string = format!("{}", 0);
+        }
```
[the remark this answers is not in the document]
SamHSmith: good point.

core/src/kura.rs
```diff
@@ -429,21 +451,41 @@ impl<IO: DiskIO> BlockStore<IO> {
...
+                let (a, b) = file_string.split_at(file_string.len() - FILENAME_MAX_LENGTH);
+                file_string = format!("{}/{}", a, b);
+            } else {
+                file_string = format!(
+                    "{}/{}{}",
+                    0,
+                    "0".repeat(FILENAME_MAX_LENGTH - file_string.len()),
+                    file_string
+                );
+            }
```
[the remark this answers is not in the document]
SamHSmith: Yea your right. In this case we would never want them to diverge in the future so that makes sense.

core/src/kura.rs
SamHSmith: That's not why it's important. It's important that the / get's removed so that it doesn't get treated as a root directory. If you start with dir/dir1/file2 and remove dir you get /dir1/file2 which is not what we want.

Arjentix: Yeah, that's what I'm talking about. Look at [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=930966cdc057634a1a1e8920161fb7d1)

core/src/kura.rs
```diff
@@ -429,15 +456,20 @@ impl<IO: DiskIO> BlockStore<IO> {
...
+        for index in &base_indices {
+            files.push(self.filename_to_path(&index.to_string()));
+        }
+
+        let result = tokio_stream::iter(files)
+            .map(move |e| e)
             .then(move |e| {
                 let io = io.clone();
-                async move { io.open(e.into_os_string()).await }
+                // async move { io.open(e.into_os_string()).await }
```
Arjentix, line 471: Commented code

core/src/kura.rs
```diff
@@ -463,15 +505,39 @@ async fn storage_files_base_indices<IO: DiskIO>(
     path: &Path,
     io: &IO,
 ) -> Result<BTreeSet<NonZeroU64>> {
-    let bases = io
+    let dirs: Vec<(String, IO::Dir)> = io
         .read_dir(path.to_path_buf())
         .await?
         .filter_map(|item| async {
```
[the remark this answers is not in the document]
Arjentix: Don't you think it looks prettier? I'm talking about using `filter` and `map` instead of `filter_map`.

AFAIK we should use `filter_map` when we will apply an operation to parameter regardless of its value. Here is not this situation, here we apply operation only on `Ok` variant.

core/src/kura.rs
```diff
@@ -839,7 +929,12 @@ mod tests {
         )
         .await
         .unwrap();
-        let written = fs::write(dir.path().join("1"), vec![1; 40]).await;
+        fs::DirBuilder::new()
+            .recursive(true)
+            .create(&dir.path().join("0"))
+            .await
+            .unwrap();
```
Arjentix, line 936: `create_dir_all()`? Should be fine here. But if you like this style, keep it

core/src/kura.rs
```diff
@@ -429,15 +456,20 @@ impl<IO: DiskIO> BlockStore<IO> {
...
-    ) -> Result<impl Stream<Item = Result<VersionedCommittedBlock>> + 'static> {
+    ) -> Result<impl Stream<Item = Result<VersionedCommittedBlock>> + '_> {
         let io = self.io.clone();
         let base_indices = storage_files_base_indices(&self.path, &self.io).await?;
-        let dir_path = self.path.clone();
-        let result = tokio_stream::iter(base_indices)
-            .map(move |e| dir_path.join(e.to_string()))
+        let mut files = Vec::new();
+        for index in &base_indices {
+            files.push(self.filename_to_path(&index.to_string()));
```
Arjentix, line 464: What about renaming `filename_to_path()` function to `block_index_to_path()`?

core/src/kura.rs
```diff
@@ -350,8 +353,8 @@ impl<IO: DiskIO> BlockStore<IO> {
     /// - Filesystem access failure (HW or permissions)
     ///
     pub async fn get_block_path(&self, block_height: NonZeroU64) -> Result<PathBuf> {
-        let filename = self.get_block_filename(block_height).await?;
-        Ok(self.path.join(filename))
+        let file_string = self.get_block_filename(block_height).await?;
+        Ok(self.filename_to_path(&file_string))
     }
```
Arjentix, line 358: It's a little confusing for me that `get_block_filename()` does not return valid path and needs further processing... Maybe rename it to something else or refactor somehow?

Arjentix (approved):
Good job! But can you, please, also place somewhere in the code documentation about new files structure?

appetrosyan (approved):

PR #2195, opened 2022-05-11
[documentation] #1280: Document Iroha metrics

### LINKED ISSUE #1280: Document Iroha metrics
At the moment the only way to get list of metrics is to request them from the prometheus port and read text response containing help comments like that:

```
> curl http://127.0.0.1:5551/metrics
# HELP blocks_height Total number of blocks in chain
# TYPE blocks_height gauge
blocks_height 135543
# HELP peers_number Total number peers to send transactions and request proposals
# TYPE peers_number gauge
peers_number 7
# HELP number_of_domains Total number of domains in WSV
# TYPE number_of_domains gauge
number_of_domains 14
# HELP total_number_of_transactions Total number of transactions in blockchain
# TYPE total_number_of_transactions gauge
total_number_of_transactions 216499
# HELP number_of_signatures_in_last_block Number of signatures in last block
# TYPE number_of_signatures_in_last_block gauge
number_of_signatures_in_last_block 5
```

It would be convenient to add and maintaing list of all metrics in iroha readthedocs' documentation

outoftardis:
Signed-off-by: Ekaterina Mekhnetsova <mekkatya@gmail.com>

### Description of the Change

Add documentation for metrics: https://github.com/outoftardis/iroha/blob/doc-metrics/docs/source/guides/metrics.md

### Issue

Partially addresses #1280

### DIFF
--- a/docs/source/guides/metrics.md
+++ b/docs/source/guides/metrics.md
@@ -1,0 +1,39 @@
+# Metrics
+
+To conveniently and thoroughly monitor the performance of the network, we recommend using [`prometheus`](https://prometheus.io/). Prometheus is a program that can monitor your Iroha peer over a separate socket and provide different kinds of performance metrics.
+
+This data can help you find performance bottlenecks and optimise your Iroha configuration.
+
+## How to use metrics
+
+To use metrics, you need to configure the `/metrics` endpoint in the [Iroha configuration](../references/config.md). By default, the endpoint is exposed at `127.0.0.1:8180/metrics`. If the port is not available, Iroha will still start and work normally, but metrics won't be accessible.
+
+After that, use the IP address to access the data from the running Iroha instance. For example:
+
+```bash
+curl http://127.0.0.1:8080/metrics
+```
+
+This will give you a result like this:
+
+```bash
+# HELP blocks_height Total number of blocks in chain
+# TYPE blocks_height gauge
+blocks_height 135543
+# HELP peers_number Total number peers to send transactions and request proposals
+# TYPE peers_number gauge
+peers_number 7
+# HELP number_of_domains Total number of domains in WSV
+# TYPE number_of_domains gauge
+number_of_domains 14
+# HELP total_number_of_transactions Total number of transactions in blockchain
+# TYPE total_number_of_transactions gauge
+total_number_of_transactions 216499
+# HELP number_of_signatures_in_last_block Number of signatures in last block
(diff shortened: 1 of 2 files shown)


docs/source/guides/metrics.md
```diff
@@ -0,0 +1,39 @@
+# Metrics
+
+To conveniently and thoroughly monitor the performance of the network, you can use metrics: numeric data collected over time about your Iroha network.
```
appetrosyan, line 3:
```suggestion
To conveniently and thoroughly monitor the performance of the network, we recommend using [prometheus](https://prometheus.io/). It's a program that can monitor your Iroha peer over a separate socket, providing you different kinds of performance metrics. 
```

docs/source/guides/metrics.md
```diff
@@ -0,0 +1,39 @@
...
+
+You can then analyse the data to make your project even more efficient.
```
appetrosyan, line 5:
```suggestion
This data can help find performance bottlenecks, and optimise Iroha's configuration. 
```

docs/source/guides/metrics.md
```diff
@@ -0,0 +1,39 @@
...
+
+## How to use metrics
+
+To use metrics, you need to add it to your [Iroha configuration](../references/config.md).
```
appetrosyan, line 9:
```suggestion
To use metrics, you need to configure a metrics endpoint in the [Iroha configuration](../references/config.md). By default the endpoint is exposed at `127.0.0.1:8180/metrics`, if the port is not available, Iroha will still start and work normally, but metrics won't be accessible.
```

appetrosyan (changes requested):

appetrosyan:
I think we need a little more detail on what metrics are gathered exactly. Perhaps we should add a subcommand to `kagami` to generate a `metrics` reference, what do you think @Arjentix ?

Arjentix:
Still there is no way to read about all metrcis without requesting an end-point. So this PR doesn't close the issue, but maybe it shouldn't...

@appetrosyan , I don't know how exactly prometheus metrics are working in our project, but if it's possible to generate this reference, it would be very good.

After looking at code I think it might be possible with procedural macro and global variable magic. We can try it

Arjentix (approved):
I think the doc is alright and we can merge it now and update later, then there will be a reference generated by our `kagami` tool

appetrosyan (approved):

Arjentix (approved):

PR #2197, opened 2022-05-11
[feature] #2100: Add query to find all accounts with asset

### LINKED ISSUE #2100: Feature request: find all accounts that have an asset by its id
It is necessary for Iroha Blockchain Explorer: there is a situation when we want to know the exact asset holders.

Arjentix:
### Description of the Change

* Added `FindAccountsWithAsset` query
* Added test for it
* Fixed bug when block events (including `Committed` event) were sent before actual block applying to *WSV*
* Cause of this bugfix some of previously flaky tests are no longer such. Tested with `cargo flaky` running every test for **100** times for about 4 hours
* Smartcontract test is enabled back as it seems not being flaky anymore. But still takes about 50 secs to complete on my machine

Results of running ```cargo flaky``` (edited for readability):
```
====== FOUND 4 FAILING TESTS ======
- test: two_networks stdout
failures: 1/100 (1%)

- test: integration::unregister_peer::unstable_network_stable_after_add_and_after_remove_peer stdout
failures: 2/100 (2%)

- test: config::tests::parse_example_json stdout
failures: 100/100 (100%)

- test: ui stdout
failures: 600/100 (600%)
```

The **3'rd** case isn't a real problem. This test just depends on a folder it runs from, we should be able to fix it probably.
The **4'th** case is a compilation tests (as @mversic suggested). Not sure what does it means here.

### Issue

* Closes #2100 

### Benefits

* New query
* Less number of flaky tests

### Possible Drawbacks

I've enabled `mint_nft_for_every_user_every_1_sec` test back. I haven't run `cargo flaky` with it but I've run it mannually (bash script) for about 30 times and it seems to work fine. The only problem is a long execution time (~50 sec on my M1 Mac), so probably I need to ignore this test again.

### Usage Examples or Tests

```bash
cargo test --package iroha_client --test mod -- integration::queries::account::find_accounts_with_asset --exact --nocapture
```

### DIFF
--- a/client/build.rs
+++ b/client/build.rs
@@ -28,23 +28,23 @@ fn main() {
     assert!(fmt.success(), "Can't format smartcontract");
 
     let build = Command::new("cargo")
-            // Removing environment variable to avoid
-            // `error: infinite recursion detected` when running `cargo lints`
-            .env_remove("RUST_RECURSION_COUNT")
-            .env("CARGO_TARGET_DIR", out_dir)
-            .current_dir(smartcontract_path)
-            .args(&[
-                "+nightly-2022-04-20",
-                "build",
-                "--release",
-                "-Z",
-                "build-std",
-                "-Z",
-                "build-std-features=panic_immediate_abort",
-                "--target",
-                "wasm32-unknown-unknown",
-            ])
-            .status()
-            .expect("Failed to run `cargo build` on smartcontract");
+        // Removing environment variable to avoid
+        // `error: infinite recursion detected` when running `cargo lints`
+        .env_remove("RUST_RECURSION_COUNT")
+        .env("CARGO_TARGET_DIR", out_dir)
+        .current_dir(smartcontract_path)
+        .args(&[
+            "+nightly-2022-04-20",
+            "build",
+            "--release",
+            "-Z",
+            "build-std",
+            "-Z",
+            "build-std-features=panic_immediate_abort",
+            "--target",
+            "wasm32-unknown-unknown",
+        ])
+        .status()
+        .expect("Failed to run `cargo build` on smartcontract");
     assert!(build.success(), "Can't build smartcontract")
 }
--- a/client/build.rs
+++ b/client/build.rs
@@ -27,14 +27,7 @@ fn main() {
         .expect("Failed to run `rustfmt` on smartcontract");
     assert!(fmt.success(), "Can't format smartcontract");
 
-    let instrumenting_coverage = if let Ok(flags) = env::var("RUSTFLAGS") {
-        flags.contains("instrument-coverage")
-    } else {
-        false
-    };
-
-    if instrumenting_coverage {
-        let build = Command::new("cargo")
+    let build = Command::new("cargo")
             // Removing environment variable to avoid
             // `error: infinite recursion detected` when running `cargo lints`
             .env_remove("RUST_RECURSION_COUNT")
@@ -53,6 +46,5 @@ fn main() {
             ])
             .status()
             .expect("Failed to run `cargo build` on smartcontract");
-        assert!(build.success(), "Can't build smartcontract")
-    }
+    assert!(build.success(), "Can't build smartcontract")
 }
--- a/client/build.rs
+++ b/client/build.rs
@@ -27,6 +27,7 @@ fn main() {
         .expect("Failed to run `rustfmt` on smartcontract");
     assert!(fmt.success(), "Can't format smartcontract");
 
+    // TODO: Remove cargo invocation (#2152)
     let build = Command::new("cargo")
         // Removing environment variable to avoid
         // `error: infinite recursion detected` when running `cargo lints`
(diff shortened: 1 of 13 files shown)


client/build.rs
```diff
@@ -27,32 +27,28 @@ fn main() {
         .expect("Failed to run `rustfmt` on smartcontract");
     assert!(fmt.success(), "Can't format smartcontract");
 
-    let instrumenting_coverage = if let Ok(flags) = env::var("RUSTFLAGS") {
```
Arjentix, line 30: Note from [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts):
> Note that since Rust 1.55, RUSTFLAGS is removed from the environment; scripts should use CARGO_ENCODED_RUSTFLAGS instead.

core/src/smartcontracts/isi/query.rs
```diff
@@ -103,6 +103,7 @@ impl<W: WorldTrait> ValidQuery<W> for QueryBox {
             FindAccountById(query) => query.execute_into_value(wsv),
             FindAccountsByName(query) => query.execute_into_value(wsv),
             FindAccountsByDomainId(query) => query.execute_into_value(wsv),
+            FindAccountsWithAsset(query) => query.execute_into_value(wsv),
```
mversic, line 106: not really a big deal, but a small declarative macro could be implemented for this long enum

core/src/sumeragi/fault.rs
```diff
@@ -746,14 +746,15 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
         let block = block.commit();
         let block_hash = block.hash();
 
+        if let Err(error) = self.wsv.apply(block.clone()).await {
+            warn!(?error, %block_hash, "Failed to apply block on WSV");
+        }
+
         for event in Vec::<Event>::from(&block) {
             trace!(?event);
             drop(self.events_sender.send(event));
```
mversic, line 755: off topic comment. I've wondered before about this drop. No big deal

mversic (approved):

core/src/sumeragi/fault.rs
```diff
@@ -746,14 +746,15 @@ impl<G: GenesisNetworkTrait, K: KuraTrait, W: WorldTrait, F: FaultInjection>
         let block = block.commit();
         let block_hash = block.hash();
 
+        if let Err(error) = self.wsv.apply(block.clone()).await {
+            warn!(?error, %block_hash, "Failed to apply block on WSV");
+        }
```
mversic, line 751: so this change fixed flakiness?

client/build.rs
```diff
@@ -27,32 +27,28 @@ fn main() {
...
-        assert!(build.success(), "Can't build smartcontract")
-    }
+    let build = Command::new("cargo")
+        // Removing environment variable to avoid
+        // `error: infinite recursion detected` when running `cargo lints`
+        .env_remove("RUST_RECURSION_COUNT")
+        // Removing environment variable to avoid
+        // `error: `profiler_builtins` crate (required by compiler options) is not compatible with crate attribute `#![no_core]``
+        // when running with `-C instrument-coverage`
+        .env_remove("CARGO_ENCODED_RUSTFLAGS")
```
appetrosyan, line 37: I think we'll need to revisit this at some point.

core/src/smartcontracts/isi/account.rs
```diff
@@ -358,4 +358,31 @@ pub mod query {
             .ok_or_else(|| query::Error::Find(Box::new(FindError::MetadataKey(key))))
         }
     }
+
+    impl<W: WorldTrait> ValidQuery<W> for FindAccountsWithAsset {
+        #[log]
```
appetrosyan, line 363: I'll remove this in #2186, might as well do it now.

core/src/smartcontracts/isi/query.rs
appetrosyan: There's a crate that can forward implementations to enum variants. It can also remove the need for manually adding the `enum` variant.

appetrosyan (approved):

core/src/sumeragi/fault.rs
Arjentix: Yes

client/build.rs
mversic: let's attach a TODO then

core/src/smartcontracts/isi/query.rs
Arjentix: Create new issue?

appetrosyan (approved):

appetrosyan (approved):

core/src/smartcontracts/isi/query.rs
appetrosyan: I'll do that.

mversic (approved):

Issue #2103, opened 2022-04-14
Support querying for blocks and transactions


0x009922:
In Iroha Block Explorer, we need to show up to a user the list of latest blocks and transactions. Also it would be useful to query for some specific block/transaction by its id, height and hash relatively.

Thus, it would be nice to extend Query API with the following:

- `FindAllBlocks`, that returns a list of blocks. With pagination, of course. Ordered by height desc.

  There are several life stages of a block:

  1. pending
  2. chained
  3. valid
  4. committed

  AFAIK in explorer we only need pending and committed blocks, and we need to distinguish them from each other.

- `FindBlockByHeight`, that returns a single block if it has been found. I think we need to know the block status here as well. 
- `FindAllTransactions`, that returns a list of latest transactions, ordered by timestamp desc. **The most debatable query**. 

  I am not sure that this query is a good idea, because actually transactions are always contained within blocks, and its enough to query for blocks only to get transactions as well. But if we want to show the list of transactions decoupled from their blocks, how to paginate this data independently from blocks? Maybe we should drop the idea of "transactions list" decoupled from blocks.

- `FindTransactionByHash`, that returns a single transaction if it has been found. This is not as debatable as `FindAllTransactions` is, but I think it might be really hard to query for it in an efficient way. Am I right?

appetrosyan:
- `FindTransactionByHash` is there since #1517. 
- `FindAllTransactions` has some overlap with `FindAllBlocks`. Still, because of the difference in pagination and how it would be a nightmare to implement `FindAllTransactions` given only a paginated `FindAllBlocks` I think we should implement both. 
- `FindBlockByHeight` is redundant. Pagination has a lower and upper bound. `FindAllBlocks` can be used to get blocks of height _n_ to _n+1_ and it would have the same effect.

appetrosyan:
- `FindTransactionByHash` is there since #1517. 
- `FindAllTransactions` has some overlap with `FindAllBlocks`. Still, because of the difference in pagination and how it would be a nightmare to implement `FindAllTransactions` given only a paginated `FindAllBlocks` I think we should implement both. 
- `FindBlockByHeight` is redundant. Pagination has a lower and upper bound. `FindAllBlocks` can be used to get blocks of height _n_ to _n+1_ and it would have the same effect.

appetrosyan:
@QuentinI If you're looking for something new to implement, seems like a good place to start

s8sato:
[...]

PR #2210, opened 2022-05-13
[feature] #2103: support querying for blocks and transactions

QuentinI:
### Description of the Change

Added `FindAllBlocks` and `FindAllTransactions` queries as described in #2103 

### Issue

No way to query for e.g. block explorer to query for older blocks

### Benefits

Support querying for list of all blocks and transactions

### Possible Drawbacks

As implemented in this PR, FindAllBlocks query introduces a block representation in `iroha_data_schema` parallel to one already present in `iroha_core`.


data_model/src/events/mod.rs
```diff
@@ -88,7 +88,9 @@ pub enum EventSubscriberMessage {
 }
 
 /// Event.
-#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, FromVariant, IntoSchema)]
```
QuentinI, line 91: `Serialize` and `Deserialize` was missing from events. I added it, but I wonder if it was on purpose?

appetrosyan (approved):

data_model/src/block_value.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub height: u64,
+    /// Hash of a previous block in the chain.
+    /// Is an array of zeros for the first block.
+    pub previous_block_hash: Hash,
+    /// Hash of merkle tree root of the tree of valid transactions' hashes.
+    pub transactions_hash: HashOf<MerkleTree<VersionedTransaction>>,
+    /// Hash of merkle tree root of the tree of rejected transactions' hashes.
+    pub rejected_transactions_hash: HashOf<MerkleTree<VersionedTransaction>>,
+    /// Hashes of the blocks that were rejected by consensus.
+    pub invalidated_blocks_hashes: Vec<Hash>,
```
QuentinI, line 32: Since I can't use `VersionedCommitedBlock` here and `HashOf<BlockValue>` would be different to `HashOf<VersionedCommitedBlock>` it's just naked `Hash`. I don't think there's any good solution to be found without moving definition of blocks to separate crate entirely.

data_model/src/lib.rs
```diff
@@ -426,6 +430,7 @@ impl Value {
             U32(_) | U128(_) | Id(_) | PublicKey(_) | Bool(_) | Parameter(_) | Identifiable(_)
             | String(_) | Name(_) | Fixed(_) | TransactionValue(_) | PermissionToken(_)
             | Hash(_) => 1_usize,
+            Block(v) => v.nested_len() + 1_usize,
```
mversic, line 433: hm, where is +1 coming from?

data_model/src/block_value.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub header: BlockHeaderValue,
+    /// Array of transactions
+    pub transactions: Vec<TransactionValue>,
+    /// Event recommendations
+    pub event_recommendations: Vec<Event>,
+}
+
+impl BlockValue {
+    /// ...
+    pub fn nested_len(&self) -> usize {
```
QuentinI, line 60: I'm not sure I fully understand the semantics here, so I'm just adding lengths of everything having a length here :)

data_model/src/block_value.rs
mversic: yes, actually blocks should be moved to the `data_model` crate from the `core` crate. Because as far as I can see you are duplicating the block definition

data_model/src/lib.rs
QuentinI: From counting the block itself, I guess?
As I said, I'm not sure I fully understand the semantics of this.

data_model/src/block_value.rs
mversic: while this is what I would like to see the most I'm not sure everyone would agree

core/src/smartcontracts/isi/block.rs
```diff
@@ -0,0 +1,21 @@
+//! This module contains trait implementations related to block queries
+use eyre::Result;
+use iroha_data_model::block_value::BlockValue;
+use iroha_logger::prelude::*;
+use iroha_telemetry::metrics;
+
+use super::*;
+
+impl<W: WorldTrait> ValidQuery<W> for FindAllBlocks {
+    #[log]
```
mversic, line 10: please don't use `#[log]`. It causes deadlocks as far as I'm aware of

core/src/smartcontracts/isi/tx.rs
```diff
@@ -7,6 +7,16 @@ use iroha_telemetry::metrics;
 
 use super::*;
 
+impl<W: WorldTrait> ValidQuery<W> for FindAllTransactions {
+    #[log]
```
mversic, line 11: please don't use `#[log]`. It causes deadlocks as far as I'm aware of

data_model/src/block_value.rs
appetrosyan: I think that Block -> data_model is beyond the scope of a last minute change.

mversic (changes requested):

QuentinI:
@mversic re: https://github.com/hyperledger/iroha/pull/2210#discussion_r872538610
I agree that duplicating the block definition here is not ideal.
It's a quick solution we agreed on with @appetrosyan , as far as I understand it's planned to move block definitions to their own crate or do some other crate structure refactoring to that effect.

core/src/smartcontracts/isi/tx.rs
appetrosyan: Doesnt matter, I'll remove it in #2186 anyway.

data_model/src/block_value.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+/// Representation of block on blockchain
+#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Serialize, Deserialize, IntoSchema)]
+pub struct BlockValue {
+    /// Header
+    pub header: BlockHeaderValue,
+    /// Array of transactions
+    pub transactions: Vec<TransactionValue>,
```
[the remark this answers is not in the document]
QuentinI: I removed fields based on recommendations of @appetrosyan, although I may have mixed something up.
Not sure why I merged transactions into one vector though, I'll try to recall if I had any sensible reasoning for it and revert if not.

data_model/src/events/mod.rs
appetrosyan: We might need the two soon

appetrosyan (approved):

data_model/src/transaction.rs
```diff
@@ -424,7 +424,7 @@ impl PartialOrd for TransactionValue {
     }
 }
 
-declare_versioned_with_scale!(VersionedValidTransaction 1..2, Debug, Clone, FromVariant, IntoSchema);
+declare_versioned!(VersionedValidTransaction 1..2, Debug, Clone, PartialEq, Eq, FromVariant, IntoSchema);
```
mversic, line 427: do we need it encoded as both scale and json now?

mversic: and how come `PartialEq` and `Eq` are needed?

core/src/smartcontracts/isi/tx.rs
mversic: I'm think @appetrosyan PR got merged first, so this would reintroduce the `#[log]`

data_model/src/query.rs
```diff
@@ -1249,6 +1254,35 @@ pub mod transaction {
...
+        PartialOrd,
+        Ord,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    /// `FindAllTransactions` Iroha Query will list all transactions included in blockchain
+    pub struct FindAllTransactions;
```
mversic, line 1273: I think we should support n last or from hash filtering in the `FindAllTransactions` query

core/src/wsv.rs
```diff
@@ -684,6 +684,34 @@ impl<W: WorldTrait> WorldStateView<W> {
         self.new_block_notifier.subscribe()
     }
 
+    /// Get all transactions
+    pub fn transaction_values(&self) -> Vec<TransactionValue> {
```
mversic, line 688: I think transactions should be grouped by block number, i.e. function prototype should be something like this:
```suggestion
    pub fn transaction_values(&self) -> Vec<(u32, Vec<TransactionValue>)> {
```

or block number can be implicit in the position in the vector:
```suggestion
    pub fn transaction_values(&self) -> Vec<Vec<TransactionValue>> {
```

I somehow dislike the idea of mixing valid and rejected transactions so maybe this format could be favored:
 ```suggestion
    pub fn transaction_values(&self) -> Vec<(Vec<ValidTransaction>, Vec<RejectedTransaction>)> {
```

data_model/src/query.rs
QuentinI: I don't know much about the use-cases, so I'll defer to your judgement here.

core/src/wsv.rs
QuentinI: Oh, this function isn't actually needed anymore since I split the transactions into valid and rejected, so I should actually remove it.

That said, having block numbers in `FindAllTransactions` output makes sense to me, but again, not sure what the actual use-cases of the query are.

data_model/src/query.rs
appetrosyan: `from_hash` — good idea. `n_last` could be emulated with pagination, so I'd add a new ticket to the backlog.

mversic: yes, should be in a separate ticket. I'll resolve this then

core/src/wsv.rs
mversic: that's true, I too don't know the use-case of this feature so you could just open a ticket for this

mversic (dismissed):

QuentinI:
@appetrosyan @mversic  I applied minor suggestions and rebased.

Do I understand correctly that after this being merged I should open new issues for `from_hash` and block numbering in `FindAllTransactions` ?

mversic:
Yeah, let's do that

appetrosyan (approved):

mversic (approved):

PR #2213, opened 2022-05-15
[feature] #2056: Add a derive proc macro crate for AssetValueType enum

### LINKED ISSUE #2056: Create a iroha procedural macro for impl FromStr for AssetValueType and other C like enums
The following macro invokation
```
easy_from_str_impl! {AssetValueType, Quantity, BigQuantity, Fixed, Store}
```

does not automatically generate code for all variants. This should be replaced with a custom derive macro for FromStr.

ilchu:

### Description of the Change

Added a crate with a proc macro to `data_model` that allows automatically deriving `FromStr` impls for any enums exhaustively. 

### Issue

Resolves #2056.

### Benefits
`FromStr` derived automatically for all variants instead of having to list them manually as was done with a declarative macro.

### Possible Drawbacks
Doubtful if this new dependency should be placed into `data_model` or in the `macro` crate and then re-exported. Opted for the former.


appetrosyan (approved):
I'm happy we didn't need to add a custom implementation. Good job!

s8sato (approved):

PR #2223, opened 2022-05-17
[ci] #2222: Split tests by whether it involves coverage or not

### LINKED ISSUE #1683: Remove integration tests from coverage
Integration tests can provide an unreliable metric for coverage. In a real deployment of iroha, error handling is going to happen frequently. Occasional failures can improve code coverage in error recovery code, and test failures in unstable integration tests could produce dips in coverage. 

Our coverage should only depend on objective reproducible metrics, i.e. only on unit tests that explicitly test the error recovery mechanisms.

### LINKED ISSUE #2136: Fix Flaky tests
Cargo nextest is the more permanent solution to flaky tests, specifically it has a method of remembering and re-running tests that have failed previously. 

https://nexte.st/index.html 

We should move to use it in our CI image, and potentially generate test coverage and run tests in one go.

### LINKED ISSUE #2148: Fix `UI` tests under coverage
Currently, trybuild tests fail when instrumenting for coverage. They are disabled in #2104 , but should be re-enabled when we can guarantee that they run properly.

### LINKED ISSUE #2222: Combine  `coverage` job and unit tests
Related to:
- #1683
- #2136
- #2148

s8sato:
### Description of the Change
- Split tests into 3 jobs
  - `checks`
    - script checks
    - Wasm build check
  - `tests_with_coverage`
    - unit tests #2222
    - integration tests (in a broad sense; crate/tests/*) except for those to be tested in `integration_tests` job
  - `integration_tests` #1683
    - client/tests/integration
    - core/test_network/tests/integration
    - p2p/tests/integration
- Move from `grcov` to `cargo-llvm-cov`
- Include UI tests to coverage profiling #2148

### Issue
- Closes #1683
- Closes #2148
- Closes #2222
- Opens #2242

### Benefits
- Revival of coverage
- Visible coverage failure
- Accuracy of cover rate by #1683 #2148
- CI optimization by #2222

### Possible Drawbacks
- Appearance of 10% or more regression in coverage due to excluding some integration tests


appetrosyan:
https://github.com/BAStos525 Should be included in this development.

s8sato:
The `coverage` job passes even though it actually fails due to a missing component.
Probably `grcov` exit with code 0 despite of [the error](https://github.com/hyperledger/iroha/runs/6481913502?check_suite_focus=true#step:7:9).
There is a corresponding closed [issue](https://github.com/mozilla/grcov/issues/662), but it does not seem to have been resolved

s8sato:
Now `cargo-llvm-cov` [exits with code 1](https://github.com/hyperledger/iroha/runs/6486108316?check_suite_focus=true#step:7:10) with a missing component, and the `coverage` job *successfully* fails
image

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -16,67 +16,71 @@ env:
...
+      image: 7272721/i2-ci:latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1
+      - name: Run tests
+        if: always()
+        run: |
+          mold --run cargo test --test '*' --all-features -- \
+          integration:: \
```
s8sato, line 84: `'*'` pattern matches any integration tests, then `integration::` limits them to client/tests/integration/
https://doc.rust-lang.org/cargo/commands/cargo-test.html#examples

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -16,67 +16,68 @@ env:
...
+          --test-threads 2 \
```
appetrosyan, line 82: We should remove this at some point.

.github/workflows/iroha2-dev.yml
```diff
@@ -176,33 +176,27 @@ jobs:
...
-          grcov . --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore "/*" -o lcov.info
+      - name: Run tests and generate lcov report
         if: always()
+        run: mold --run cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v2
-        if: always()
         with:
           file: lcov.info
+          fail_ci_if_error: true
```
appetrosyan, line 202: This is dangerous. We have flaky tests so we wouldn't be able to push anything if the coverage fails.

.github/workflows/iroha2.yml
```diff
@@ -13,30 +13,24 @@ jobs:
...
+      - name: Run tests and generate lcov report
         if: always()
+        run: mold --run cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v2
-        if: always()
         with:
           file: lcov.info
-
+          fail_ci_if_error: true
```
appetrosyan, line 33: Same as in `iroha2-dev.yml`. We might want to be able to push despite a failing coverage

client/tests/integration/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(not(coverage))]
```
appetrosyan, line 1: This probably needs to be added to a few more crates. There's a few things that are just not coverable.

.github/workflows/iroha2-dev-pr.yml
s8sato: If you are mentioning line 82, I only inherited this and don't know exact reason to limit threads

.github/workflows/iroha2-dev.yml
s8sato: If it enters this uploading step, that means all the tests have passed in the previous step.
But I agree to remove this line because we would not like to withhold pushing while codecov.io is down

s8sato: But in that case we could dismiss the failing check and merge

s8sato: I think it affects only this `coverage` job, but not clear. I'll remove this line

appetrosyan (approved):

s8sato:
> Aleksandr Petrosyan, [2022/05/20 16:50]
> [In reply to S. Sato]
> Approved. 
> 
> There's one unresolved problem, though. 
> 
> Ideally we want to run the tests in parallel. We don't want to re-run cargo test fresh every time, because most of our pipeline time is taken by compiling Iroha and not actually running the tests. 
> 
> Because of the way that 1.61 cargo test --no-run prints the names of the executables that it generates, we could then build the test executables once, treat them as artifacts and test them in subsequent testing steps. The only drawback is that then cfg(not(coverage)) is more of a hindrance.

- Opened #2242

Arjentix (approved):
LGTM. You know a lot more about our CI/CD than me and I can't find any issues here

PR #2235, opened 2022-05-19
[feature] #1413: Add API version endpoint

### LINKED ISSUE #1413: Add API version endpoint
Clients need to verify if their implementation is compatible, so they need to check our API version.

It would be good to require upgrading version if our schema is changed.

ilchu:

### Description of the Change
Added a new endpoint that gives back the current API version as a JSON string. 

### Issue
Resolves #1413.

### Benefits
As we're about to introduce versioning, this will allow end users to check if their client and our API versions match.

### Possible Drawbacks
Set to grab the version from the genesis block, so the endpoint should be usable at all times when a minimal network is up. `ChainIterator` in `wsv`'s `blocks()` method was mentioning possible deadlock, but it should be safe in this case in all I can tell.

### Usage Examples or Tests
Added some documentation describing the change in `api_spec.md`.

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -344,6 +344,22 @@ mod subscription {
     }
 }
 
+#[iroha_futures::telemetry_future]
+#[cfg(feature = "telemetry")]
+async fn handle_version<W: WorldTrait>(wsv: Arc<WorldStateView<W>>) -> Json {
+    use iroha_version::Version;
+
+    #[allow(clippy::expect_used)]
+    reply::json(
+        &wsv.blocks()
+            .last()
+            .expect("At least genesis should always exist")
+            .value()
+            .version()
+            .to_string(),
+    )
+}
+
 #[cfg(feature = "telemetry")]
 async fn handle_metrics<W: WorldTrait>(
     wsv: Arc<WorldStateView<W>>,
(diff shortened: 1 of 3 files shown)


docs/source/references/api_spec.md
```diff
@@ -301,6 +301,26 @@ uptime_since_genesis_ms 54572974
...
+
+**Responses**:
+- 200 OK - The current version of API used by iroha returned as a json string.
+Grabbed from the genesis block's version, so at least a minimal subnet of 4 peers
+should be running and the genesis be submitted at the time of request.
+```
+"1"
+```
+
 Learn [how to use metrics](../guides/metrics.md).
```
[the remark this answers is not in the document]
ilchu: Wrongly thought metrics to cover all of telemetry, so placed it below, my bad. Fixed.

cli/src/torii/routing.rs
```diff
@@ -344,6 +344,22 @@ mod subscription {
...
 }
 
+#[iroha_futures::telemetry_future]
+#[cfg(feature = "telemetry")]
+async fn handle_version<W: WorldTrait>(wsv: Arc<WorldStateView<W>>) -> Json {
+    use iroha_version::Version;
+
+    #[allow(clippy::expect_used)]
+    reply::json(
+        &wsv.blocks()
```
appetrosyan, line 354: Bad idea. What if we changed the version after running Iroha for a while. `blocks().last()` on the other hand should contain either `genesis` since it's the first **and** last block, while it also shows the latest supported API version.

appetrosyan (changes requested):

cli/src/torii/routing.rs
ilchu: Fair enough, fixed that.

s8sato:
LGTM so far.
The requirements show that a client also needs to know the version the client expects.
One solution might be to hold a constant value somewhere shared by both peer and client e.g. `iroha_data_model`

appetrosyan (approved):

s8sato (approved):

appetrosyan (approved):

Issue #2170, opened 2022-04-29
make the docker build compatible with M1 machines 


appetrosyan:
```
Bogdan Mingela, [29.04.22 16:59]
JFYI, trying to produce a dockerized Iroha locally running M1 processor (step 17/19)
> [builder 5/5] RUN cargo build $PROFILE --workspace:

#17 24.80 error: failed to run custom build command for iroha_client v2.0.0-pre-rc.3 (/iroha/client)
#17 24.80
#17 24.80 Caused by:
#17 24.80   process didn't exit successfully: /iroha/target/debug/build/iroha_client-14d71321ee6f1108/build-script-build (exit status: 101)
#17 24.80   --- stderr
#17 24.80   error: toolchain 'nightly-2022-04-20-aarch64-unknown-linux-gnu' is not installed

installing the toolchain locally obviously does not help

as an idea for the future, we might provide some flexibility for the build process to determine and install the toolchains etc needed depending on the host configuration

Aleksandr Petrosyan, [29.04.22 17:00]
[In reply to Bogdan Mingela]
This is a known problem resolved by this PR (https://github.com/hyperledger/iroha/pull/2168).

Bogdan Mingela, [29.04.22 17:08]
[In reply to Aleksandr Petrosyan]
Interesting, I pulled the change and the cause remained the same, though a step and the lines around changed

Bogdan Mingela, [29.04.22 17:08]
ah, it's not the same, it requires also
toolchain 'nightly-2022-04-20-x86_64-unknown-linux-gnu' is not installed
now

```

appetrosyan:
@Arjentix If you have any updates, please share them here.

pesterev:
@appetrosyan @Arjentix 

Maybe this will help (it helped me). I saw this problem and fixed it [this way](https://github.com/hyperledger/iroha/pull/2237).

PR #2237, opened 2022-05-19
[fix] #2170: Fixes build in docker container on M1 machines

pesterev:
These changes work fine on my Apple Silicon (M1) machine.

I met this problem and I had to fix it in order to do my task.

The reason for the problem is that the toolchain is incorrect here:
`rustup component add rust-src --toolchain nightly-2022-04-20-x86_64-unknown-linux-gnu`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,10 +25,9 @@ RUN set -ex; \
     rm /tmp/*.sh
 
 RUN set -ex; \
-    rustup install --profile default nightly-2022-04-20; \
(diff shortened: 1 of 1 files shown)


Arjentix:
I wonder why other CI/CD checks weren't triggered?

Arjentix (approved):
I've also found this fix. Should be okay for now, I believe. @appetrosyan ?

appetrosyan (approved):

PR #2258, opened 2022-05-24
[feature] #2132: Add `endpointN` proc macro

### LINKED ISSUE #2132: Refactor `endpointN` macros
In torii routing we have a macro that produces `endpoint`s with a specific number of data items. 

This macro's usability can be greatly improved by refactoring it to be a procedural macro with more explicit (and readable) syntax.

ilchu:

### Description of the Change

Replaced the declarative macro generating `endpointN` functions with a procedural one in a separate crate.

### Issue
Resolves #2132.

### Benefits
Improved readability and extensibility.

### Possible Drawbacks
As the macro expansion uses `WarpResult` wrapper defined in `cli/src/torii/utils.rs`, all the users of macro should reexport that struct as well. Alternatively, the macro implementation could be amended to ditch the use of that altogether.

### DIFF
--- a/cli/derive/Cargo.toml
+++ b/cli/derive/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "iroha_cli_derive"
+version = "2.0.0-pre-rc.4"
+authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
+edition = "2021"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = {version = "1", default-features = false, features = ["parsing", "printing", "extra-traits"] }
+quote = "1.0"
+proc-macro2 = "1.0"
+proc-macro-error = "1.0"
+
+[dev-dependencies]
+iroha = { version = "=2.0.0-pre-rc.4", path = ".." }
+iroha_macro = { version = "=2.0.0-pre-rc.4", path = "../../macro" }
+trybuild = "1.0.53"
--- a/cli/derive/src/lib.rs
+++ b/cli/derive/src/lib.rs
@@ -1,0 +1,168 @@
+//! Crate with a proc macro for torii endpoint generation
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input,
+    punctuated::Punctuated,
+    Ident, LitInt, Result as SynResult, Token,
+};
+
+/// Generate warp filters for endpoints, accepting functions
+/// with any positive number of arguments within the range of `u8`.
+///
+/// Only the endpoint functions stated explicitly in the macro invocation
+/// are created.
+///
+/// There are two kinds of accepted arguments. One is supplying
+/// an integer literal denoting the number of arguments in a function
+/// that the endpoint accepts. The endpoint name is generated automatically
+/// in this case and will be in the shape of `endpoint{arg_count}`.
+///
+/// Another kind is a colon-separated string literal
+/// followed by an integer literal, denoting custom name of the endpoint being
+/// created and the number of arguments in a function that it accepts.
+///
+/// Also relies on `WarpResult` custom wrapper,
+/// and thus any module using this macro should also reexport
+/// the former, as well as some types from `warp` (see example).
+///
+/// # Panics:
+/// 1) When provided with neither a string nor integer literal.
+/// 2) When any of the argument count literals are not unique.
+/// 3) When the colon-separated form has spaces in the provided name.
+///
+/// # Examples:
+/// ```rust
+/// use warp::{Rejection, Filter};
+/// use std::convert::Infallible;
+/// use iroha_cli::torii::utils::WarpResult;
+///
+/// // An example with arguments of both acceptable kinds.
+/// // This would generate endpoints accepting functions with
+/// // 2, 3, 4 and 5 arguments. The first and the last of them
+/// // will have the custom names provided, whereas the other two will have
+/// // default ones such as `endpoint3`.
+/// generate_endpoints!(3, my_endpoint: 2, 4, anotherOne: 5, );
+/// ```
+#[proc_macro]
+pub fn generate_endpoints(input: TokenStream) -> TokenStream {
+    let EndpointList(list) = parse_macro_input!(input as EndpointList);
+    let arg_names = (1_u8..).map(|count| {
+        Ident::new(
+            format!("__endpoint_arg_{count}").as_str(),
+            Span::call_site(),
+        )
+    });
+    let arg_types = (1_u8..).map(|count| {
+        Ident::new(
+            format!("__Endpoint_Arg_{count}").as_str(),
+            Span::call_site(),
+        )
+    });
+    let mut endpoints = Vec::new();
+
+    for item in list {
+        let (fun_name, arg_names, arg_types) = match item {
+            EndpointItem::ArgCount(arg_count) => {
+                let fun_name = Ident::new(&format!("endpoint{arg_count}"), Span::call_site());
+                #[allow(clippy::expect_used)]
+                let count = arg_count
+                    .base10_parse::<usize>()
+                    .expect("Already checked at parse stage");
+                let arg_names = arg_names.clone().take(count).collect::<Vec<_>>();
+                let arg_types = arg_types.clone().take(count).collect::<Vec<_>>();
+                (fun_name, arg_names, arg_types)
+            }
+            EndpointItem::NameAndArgCount {
+                name: fun_name,
+                arg_count,
+            } => {
+                #[allow(clippy::expect_used)]
+                let count = arg_count
+                    .base10_parse::<usize>()
+                    .expect("Already checked at parse stage");
+                let arg_names = arg_names.clone().take(count).collect::<Vec<_>>();
+                let arg_types = arg_types.clone().take(count).collect::<Vec<_>>();
+                (*fun_name, arg_names, arg_types)
+            }
+        };
+
+        let expanded = quote! {
+            #[inline]
+            #[allow(clippy::redundant_pub_crate)]
+            pub(crate) fn #fun_name < O, E, F, Fut, Fil, #( #arg_types ),* > (
+                f: F,
+                router: Fil,
+            ) -> impl Filter<Extract = (WarpResult<O, E>,), Error = Rejection> + Clone
+            where
+                Fil: Filter<Extract = ( #( #arg_types ),* ), Error = Rejection> + Clone,
+                F: Fn( #( #arg_types ),* ) -> Fut + Copy + Send + Sync + 'static,
+                Fut: std::future::Future<Output = Result<O, E>> + Send,
+                #( #arg_types: Send ),*
+                {
+                    router.and_then(move | #( #arg_names ),* | async move {
+                        Ok::<_, Infallible>(WarpResult(f( #( #arg_names ),* ).await))
+                    })
+                }
+        };
+
+        endpoints.push(expanded);
+    }
+
+    quote! {
+        #( #endpoints )*
+    }
+    .into()
+}
+
+#[derive(Debug)]
+struct EndpointList(Vec<EndpointItem>);
+
+#[derive(Debug)]
+enum EndpointItem {
+    NameAndArgCount { arg_count: LitInt, name: Box<Ident> },
+    ArgCount(LitInt),
+}
+
+impl Parse for EndpointList {
+    fn parse(input: ParseStream) -> SynResult<Self> {
+        let items = Punctuated::<EndpointItem, Token![,]>::parse_terminated(input)?;
+        let mut seen_arg_counts = Vec::new();
+        for item in items.iter() {
+            match item {
+                EndpointItem::NameAndArgCount { arg_count, .. }
+                | EndpointItem::ArgCount(arg_count) => {
+                    let curr_count = arg_count.base10_parse::<u8>()?;
+                    if seen_arg_counts.contains(&curr_count) {
+                        return Err(syn::Error::new_spanned(
+                            arg_count.token(),
+                            "argument counts for all endpoints should be distinct",
+                        ));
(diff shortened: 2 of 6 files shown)


cli/derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+use proc_macro2::Span;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, Ident, LitInt, Result as SynResult,
+};
+
+/// Generates warp filters for endpoints accepting functions
+/// from 2 up to `n` arguments inclusively. Currently accepted
+/// maximum is 8. Also relies on `WarpResult` custom wrapper,
```
outoftardis, line 13: I'd suggest rewriting it like this:
```suggestion
/// Generates warp filters for endpoints, accepting functions
/// with the number of arguments from 2 to 8 (inclusive, current maximum).
/// Also relies on `WarpResult` custom wrapper,
```

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+struct EndpointCount(usize);
+
+impl Parse for EndpointCount {
+    fn parse(input: ParseStream) -> SynResult<Self> {
+        let lit: LitInt = input.parse()?;
+        let count = lit.base10_parse::<usize>()?;
+        if count > 8 || count < 2 {
+            Err(syn::Error::new_spanned(
+                lit,
+                "only endpoints with a number of arguments between two and eight are supported",
```
outoftardis, line 71:
```suggestion
                "only endpoints with the number of arguments between two and eight (inclusive) are supported",
```

though I'd use `from ... to` same as in the function description

appetrosyan:
Please amend the commit to conform to the guidelines.

cli/derive/src/lib.rs
ilchu: Thank you, all noted

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        .collect::<Vec<_>>();
+    let mut endpoints = Vec::new();
+
+    for count in 2..=max.0 {
+        let fun_name = Ident::new(format!("endpoint{count}").as_str(), Span::call_site());
+        let arg_names = &arg_names[..count];
+        let arg_types = &arg_types[..count];
+
+        let expanded = quote! {
+            pub fn #fun_name < O, E, F, Fut, Fil, #( #arg_types ),* > (
```
appetrosyan, line 36: Add `#[inline]` to the generated code. This very important for performance.

cli/src/torii/utils.rs
```diff
@@ -90,27 +91,4 @@ impl<O: Reply, E: Reply> Reply for WarpResult<O, E> {
...
-            )
-        }
-    }
-}
-
-// impl_custom_and_then!(endpoint1(a: A));
-impl_custom_and_then!(endpoint2(a: A, b: B));
-impl_custom_and_then!(endpoint3(a: A, b: B, c: C));
-impl_custom_and_then!(endpoint4(a: A, b: B, c: C, d: D));
+generate_endpoints_up_to!(4);
```
appetrosyan, line 94: I'd prefer a fn-like macro that works a little more like 
```
generate_endpoints!(2, 3, 4);
```

The reasoning is as follows: the function that you generate is public, so unless you have LTO enabled, the `endpoint1` function will be included in the binary and bloat it. We only want to have the `endpoint` functions which are actually used.

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+                f: F,
+                router: Fil,
+            ) -> impl Filter<Extract = (WarpResult<O, E>,), Error = Rejection> + Clone
+            where
+                Fil: Filter<Extract = ( #( #arg_types ),* ), Error = Rejection> + Clone,
+                F: Fn( #( #arg_types ),* ) -> Fut + Copy + Send + Sync + 'static,
+                Fut: std::future::Future<Output = Result<O, E>> + Send,
+                #( #arg_types: Send ),*
+                {
```
appetrosyan, line 45: Good scoping.

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+#[derive(Debug)]
+struct EndpointCount(usize);
+
+impl Parse for EndpointCount {
+    fn parse(input: ParseStream) -> SynResult<Self> {
+        let lit: LitInt = input.parse()?;
+        let count = lit.base10_parse::<usize>()?;
+        if (2..=8).contains(&count) {
```
appetrosyan, line 68: If we get a little more clever with the names (and we only chose single-character names for readability, there's no actual good reason to do it this way), we wouldn't need this restriction.

cli/src/torii/utils.rs
mversic: then I would suggest something along the lines of:
```suggestion
generate_endpoint!(endpoint_name1, 3);
generate_endpoint!(endpoint_name2, 4);
```

where you define the name of the endpoint, and the number of arguments it receives

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Generates warp filters for endpoints, accepting functions
+/// with the number of arguments from 2 to 8 (inclusive, current maximum).
+/// Also relies on `WarpResult` custom wrapper,
+/// and thus any module using this macro should also reexport
+/// the former.  
+#[proc_macro]
+pub fn generate_endpoints_up_to(input: TokenStream) -> TokenStream {
+    let max = parse_macro_input!(input as EndpointCount);
+    // Omitting 'e' and 'f' to allow their mnemonic use in the expansion
+    let arg_names = "abcdghij"
```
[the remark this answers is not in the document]
mversic: yeah, I would prefer some numbering scheme for naming

cli/derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "2.0.0-pre-rc.4"
+authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
+edition = "2021"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = {version = "1", default-features = false, features = ["full"] }
```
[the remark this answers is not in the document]
mversic: `full` feature is unlikely to be needed in this simple macro

mversic (changes requested):

cli/derive/Cargo.toml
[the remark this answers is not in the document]
ilchu: Do you mean that as "if all of our proc macro crates use full features of syn, then syn will only be compiled once instead of many times with different feature flags"? A few of our crates had full features enabled in syn, so I followed suit. There was at least one with minimal features only though.

cli/src/torii/utils.rs
ilchu: Do we really need the option of defining custom names for endpoints? This doesn't seem to bring any semantic value, as anyway they're wrapping handlers with more descriptive names, and there's a bit more boilerplate involved if we generate just one endpoint per macro invocation.

cli/derive/Cargo.toml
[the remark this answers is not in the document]
appetrosyan: @mversic Say **Full** was **A** and **B**. In crate **a** we use feature **A**, and in crate **b,** we use feature **B.** 

I was surprised to find that if you have resolver version 2, you might get better compilation times and smaller binary sizes (without LTO) if crates **a** and **b** both simultaneously depended on **Full**. 

It's the basis of `[cargo-hakari](https://crates.io/crates/cargo-hakari)`.

cli/src/torii/utils.rs
appetrosyan: I don't think we need the custom names just yet. But it is something that we should keep in mind. If we did do that, we could use a syntax that's a little more unambiguous: e.g. `name: 1, name_2: 2`...

cli/derive/Cargo.toml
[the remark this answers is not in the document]
mversic: Also from [here](https://rust-lang.github.io/rfcs/2957-cargo-features2.html#host-dependencies):
> Note that a dependency shared between a build-dependency and proc-macro are still unified. This is intended to help reduce build times, and is expected to be unlikely to cause problems that feature unification usually cause because they are both being built for the host platform, and are only used at build time.

appetrosyan (approved):

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+/// Also relies on `WarpResult` custom wrapper,
+/// and thus any module using this macro should also reexport
+/// the former.  
+///
+/// # Panics:
+/// 1) When provided with neither a string nor integer literal.
+/// 2) When any of the argument count literals are not unique.
+/// 3) When the colon-separated form has spaces in the provided name.
+#[proc_macro]
+pub fn generate_endpoints(input: TokenStream) -> TokenStream {
```
Arjentix, line 37: I'd add the usage examples, especially for the second synax, as it's not used anywhere

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+                    let curr_count = arg_count.base10_parse::<u8>()?;
+                    if seen_arg_counts
+                        .iter()
+                        .any(|&old_count| old_count == curr_count)
+                    {
+                        return Err(syn::Error::new_spanned(
+                            arg_count.token(),
+                            "argument counts for all endpoints should be distinct",
+                        ));
+                    }
```
Arjentix, line 149: This thing is duplicated in both match arms. I believe it can be a function

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+                        .any(|&old_count| old_count == curr_count)
+                    {
+                        return Err(syn::Error::new_spanned(
+                            arg_count.token(),
+                            "argument counts for all endpoints should be distinct",
+                        ));
+                    } else if name.value().contains(' ') {
+                        return Err(syn::Error::new_spanned(
+                            name.token(),
+                            "function name should not contain spaces",
```
mversic, line 134: had you parsed it as `Ident`, you would have to make no checks

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+                    if seen_arg_counts
+                        .iter()
+                        .any(|&old_count| old_count == curr_count)
+                    {
+                        return Err(syn::Error::new_spanned(
+                            arg_count.token(),
+                            "argument counts for all endpoints should be distinct",
+                        ));
+                    }
+                    seen_arg_counts.push(curr_count);
```
Arjentix, line 150: This is also duplicated in both match arms. Consider rewriting like this:

```rust
let count = match item {
    EndpointItem::NameAndArgCount { name, arg_count } => {
        // ...
        curr_count
    }
    EndpointItem::ArgCount(arg_count) => {
        // ...
        curr_count
    }
}?;
seen_arg_counts.push(count);
```
I think, `seen_arg_counts` modification is more clear here and match arms are doing only one thing -- parsing count

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+        let items = Punctuated::<EndpointItem, Token![,]>::parse_terminated(input)?;
+        let mut seen_arg_counts = Vec::new();
+        for item in items.iter() {
+            match item {
+                EndpointItem::NameAndArgCount { name, arg_count } => {
+                    let curr_count = arg_count.base10_parse::<u8>()?;
+                    if seen_arg_counts
+                        .iter()
+                        .any(|&old_count| old_count == curr_count)
+                    {
```
mversic, line 126:
```suggestion
                    if seen_arg_counts.contains(&curr_count) {
```

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+                        ));
+                    }
+                    seen_arg_counts.push(curr_count);
+                }
+                EndpointItem::ArgCount(arg_count) => {
+                    let curr_count = arg_count.base10_parse::<u8>()?;
+                    if seen_arg_counts
+                        .iter()
+                        .any(|&old_count| old_count == curr_count)
+                    {
```
mversic, line 144:
```suggestion
                    if seen_arg_counts.contains(&curr_count) {
```

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+            Span::call_site(),
+        )
+    });
+    let mut endpoints = Vec::new();
+
+    for item in list {
+        let (fun_name, arg_names, arg_types) = match item {
+            EndpointItem::ArgCount(arg_count) => {
+                let fun_name =
+                    Ident::new(format!("endpoint{arg_count}").as_str(), Span::call_site());
```
mversic, line 57:
```suggestion
                    Ident::new(&format!("endpoint{arg_count}"), Span::call_site());
```

mversic (approved):

cli/derive/src/lib.rs
ilchu: Got rid of duplication by introducing an "or" match pattern, see new changes

cli/derive/src/lib.rs
ilchu: Same as for duplication above

cli/derive/src/lib.rs
mversic: it's used, but in one place only. Yeah, example in the docs or a compile test or a runtime test. But because this is used in only one place and I don't think it will spread around, I'm ok with it as is

cli/derive/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+    .into()
+}
+
+#[derive(Debug)]
+struct EndpointList(Vec<EndpointItem>);
+
+#[derive(Debug)]
+enum EndpointItem {
+    NameAndArgCount { name: Ident, arg_count: LitInt },
```
mversic, line 126: would it be possible to have `arg_count` be `u8`? Since you're parsing it at the call site anyways

mversic (approved):

appetrosyan (approved):

PR #2261, opened 2022-05-24
shared memory dynamic linking

mversic:

POC implementation of dynamic linking with shared memory

### DIFF
--- a/core/src/smartcontracts/wasm.rs
+++ b/core/src/smartcontracts/wasm.rs
@@ -8,10 +8,10 @@ use std::sync::Arc;
 use config::Configuration;
 use eyre::WrapErr;
 use iroha_data_model::{prelude::*, ParseError};
-use iroha_logger::prelude::*;
 use parity_scale_codec::{Decode, Encode};
 use wasmtime::{
-    Caller, Config, Engine, Linker, Module, Store, StoreLimits, StoreLimitsBuilder, Trap, TypedFunc,
+    Caller, Config, Engine, Instance, Linker, Module, Store, StoreLimits, StoreLimitsBuilder, Trap,
+    TypedFunc,
 };
 
 use super::permissions::IsInstructionAllowedBoxed;
@@ -25,6 +25,7 @@ use crate::{
 
 type WasmUsize = u32;
 
+const IROHA_MODULE_NAME: &str = "iroha";
 /// Exported function to allocate memory
 pub const WASM_ALLOC_FN: &str = "_iroha_wasm_alloc";
 /// Name of the exported memory
@@ -52,8 +53,7 @@ pub enum Error {
     ExportNotFound(#[source] anyhow::Error),
     /// Call to the function exported from module failed
     ///
-    /// In Wasmtime v0.33, can also mean that max linear memory was
-    /// consumed
+    /// In Wasmtime v0.33, can also mean that max linear memory was consumed
     #[error("Exported function call failed")]
     ExportFnCall(#[from] Trap),
     /// Parse Error
@@ -136,7 +136,7 @@ struct State<'wrld, W: WorldTrait> {
 }
 
 impl<'wrld, W: WorldTrait> State<'wrld, W> {
-    fn new(wsv: &'wrld WorldStateView<W>, account_id: AccountId, config: Configuration) -> Self {
+    fn new(wsv: &'wrld WorldStateView<W>, account_id: AccountId, config: &Configuration) -> Self {
         Self {
             wsv,
             account_id,
@@ -146,9 +146,9 @@ impl<'wrld, W: WorldTrait> State<'wrld, W> {
                 .memory_size(config.max_memory.try_into().expect(
                     "config.max_memory is a u32 so this can't fail on any supported platform",
                 ))
-                .instances(1)
+                .instances(2)
                 .memories(1)
-                .tables(1)
+                .tables(2)
                 .build(),
         }
     }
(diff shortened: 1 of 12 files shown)


core/src/smartcontracts/wasm.rs
```diff
@@ -234,12 +275,17 @@ impl<'wrld, W: WorldTrait> Runtime<'wrld, W> {
...
     fn create_smart_contract(
-        &self,
+        &mut self,
         store: &mut Store<State<'wrld, W>>,
+        iroha_instance: Instance,
         bytes: impl AsRef<[u8]>,
-    ) -> Result<wasmtime::Instance, anyhow::Error> {
-        let module = Module::new(&self.engine, bytes)?;
-        self.linker.instantiate(store, &module)
+    ) -> Result<Instance, anyhow::Error> {
```
s8sato, line 282: This method links and instantiates the smartcontract module (as `bytes`) and the memory assistant module (as `iroha_instance`), right?

core/src/smartcontracts/wasm.rs
```diff
@@ -172,11 +172,55 @@ impl<'wrld, W: WorldTrait> State<'wrld, W> {
     }
 }
 
+struct IrohaModule {
+    module: Module,
+}
```
s8sato, line 177: This struct represents the intermediary module with the host to manage memory allocation requests from a smartcontract module, right?

core/src/smartcontracts/wasm.rs
```diff
@@ -316,7 +362,7 @@ impl<'wrld, W: WorldTrait> Runtime<'wrld, W> {
 
         let res_offset = {
             let res_offset = alloc_fn
-                .call(&mut caller, res_bytes_len)
+                .call(&mut caller, (res_bytes_len, 1))
```
s8sato, line 365: Is there any possibility that the byte alignment will take other than `1` somewhere else?
In other words, is the reason why the byte alignment can be specified in `iroha_core_wasm::alloc::_iroha_wasm_alloc` simply for consistency with the `GlobalAlloc` trait implementation, or is there another purpose?

s8sato (commented):
These are attempts to interpret the code rather than suggestions. Correct me if I am wrong

core/src/smartcontracts/wasm.rs
mversic: yes, that's it

core/src/smartcontracts/wasm.rs
mversic: Host and iroha module will exchange data serialized in scale format as a slice of bytes. Therefore, I don't think host would use alignment other than 1 and this is just for consistency

core/src/smartcontracts/wasm.rs
mversic: this instantiates only the smart contract, `iroha_instance` was instantiated earlier and is used here to import into smart contract all the exported items from `iroha_instance`

wasm/src/lib.rs
```diff
@@ -77,7 +71,7 @@ impl Execute for data_model::query::QueryBox {
     /// Executes the given query on the host environment
     fn execute(&self) -> Self::Result {
         #[cfg(not(test))]
-        use host::execute_query as host_execute_query;
+        use host::host__execute_query as host_execute_query;
```
appetrosyan, line 74: I like this convention of FFI functions having a `Struct__method` and `mod__fn` convention.

docs/source/references/config.md
```diff
@@ -773,6 +774,7 @@ Has type `WorldStateViewConfiguration`. Can be configured via environment variab
   },
   "WASM_RUNTIME_CONFIG": {
     "FUEL_LIMIT": 1000000,
+    "IROHA_WASM_MODULE": "/iroha_core_wasm.wasm",
```
appetrosyan, line 777: In this design, the WASM module that is being used can be hot-plugged. While I like the modularity, I think that this can come back to haunt us later. 

For example, how do we handle the modules' security patches. How do we ensure that the upgrade got propagated to all of the Iroha instances, without administrator intervention. What if the original WASM binary was located in a read-only file-system and is no-longer accessible. 

To me it seems like a better course of action would be to hard-code the WASM binary into Iroha itself. As such it should be impossible to update just the WASM, and we don't have a multitude of semver incompatible APIs, just one.

appetrosyan: I'd be interested to hear what @SamHSmith and @s8sato think about this.

core/wasm/src/lib.rs
```diff
@@ -0,0 +1,90 @@
+//! Library exposing memory and function table to smartcontract.
+//! Smartcontract must dynamically link against this library.
+
+// Required because of `unsafe` code and `no_mangle` use
+#![allow(unsafe_code)]
+
+#[cfg(all(not(test), not(target_pointer_width = "32")))]
+compile_error!("Target architectures other then 32-bit are not supported");
```
appetrosyan, line 8: This is something we need to revisit. WASM might not always be 32-bit.

core/wasm/src/alloc.rs
```diff
@@ -0,0 +1,30 @@
...
+///
+/// # Safety
+///
+/// Check [`alloc::dealloc`] documentation
+#[no_mangle]
+pub unsafe extern "C" fn _iroha_wasm_dealloc(ptr: *mut u8, size: usize, align: usize) {
+    if let Ok(layout) = Layout::from_size_align(size, align) {
+        alloc::dealloc(ptr, layout);
+    } else {
+        /* NOTE: Leak memory */
```
appetrosyan, line 28: And this is alright, since the smartcontract is loaded into a fresh VM and everything gets deallocated anyway?

core/wasm/build.rs
```diff
@@ -0,0 +1,5 @@
+//! Build script
+
+fn main() {
+    println!("cargo:rustc-link-arg=-zstack-size=2097152"); // 2 MiB
```
appetrosyan, line 4: I'd add an explanation for why we need this. IMO this variable needs fine-tuning. 2MiB is the lower end of what's acceptable for Rust. For context, the default on Linux is 8MiB. 

Since most of our work is loading the (ridiculously large) ISI and Queries, alongside the stack needed for the WASM dylib, we might need to adjust this number up.

core/src/smartcontracts/wasm.rs
```diff
@@ -544,13 +576,17 @@ pub mod config {
 
         /// Maximum amount of linear memory a given smartcontract can allocate
         pub max_memory: u32,
+
+        /// Path to the module [`Wasmtime`] runtime will link against
```
appetrosyan, line 580:
```suggestion
        /// Path to the module [`Wasmtime`] that the runtime will link against
```

core/src/smartcontracts/wasm.rs
```diff
@@ -459,25 +489,22 @@ impl<'wrld, W: WorldTrait> Runtime<'wrld, W> {
...
+        let state = State::new(wsv, account_id.clone(), &self.config);
         self.execute_with_state(account_id, bytes, state)
     }
 
     fn execute_with_state(
         &mut self,
         account_id: &AccountId,
         bytes: impl AsRef<[u8]>,
-        state: State<W>,
+        state: State<'wrld, W>,
```
appetrosyan, line 504: Interesting how this would interact with the changes introduced in #2284.

docs/source/references/config.md
s8sato: Since this design has the Wasm binary of the intermediary module on the Iroha side,
I think we can use API versioning for consistency between the two Wasm modules

core/wasm/build.rs
mversic: wasm's default is 1MB so I just doubled the value. But it can be tuned to our needs

core/wasm/src/alloc.rs
mversic: ok, yes every module instance is sandboxed. But the main reason is that there is no way to communicate the error to the caller and panicking over FFI is UB. Also, this is a developer error, It's unlikely that the user would call such low level APIs

core/src/smartcontracts/wasm.rs
appetrosyan: So effectively we have a `memcpy` of the functions. To me this seems the best that can be done under the circumstances.

mversic:
closing at the moment until FFI bindings generation is finished

Issue #2227, opened 2022-05-18
Register Store asset: "Unsupported Register instruction"


0x009922:
## Description

When trying to register store asset through genesis, it is not committed with the following log message:

> 2022-05-18T13:59:12.948865Z  WARN start_genesis_round: iroha_core::block: Transaction validation failed reason=Transaction rejected due to failure in instruction execution caused_by=Some(InstructionExecutionFail { instruction: Register(RegisterBox { object: EvaluatesTo { expression: Raw(Identifiable(Asset(Asset { id: Id { definition_id: DefinitionId { name: "Labadie", domain_id: Id { name: "Matthias_SCOP" } }, account_id: Id { name: "kpicard13", domain_id: Id { name: "Anicée_SCOP" } } }, value: Store(Metadata { map: {} }) }))), _value_type: PhantomData } }), reason: "Unsupported Register instruction" })

My genesis:

<details>
<summary><code>genesis.json</code></summary>

```json
{
  "transactions": [
    {
      "isi": [
        {
          "Register": {
            "object": {
              "Raw": {
                "Identifiable": {
                  "NewDomain": {
                    "id": {
                      "name": "Matthias_SCOP"
                    },
                    "logo": null,
                    "metadata": {}
                  }
                }
              }
            }
          }
        },
        {
          "Register": {
            "object": {
              "Raw": {
                "Identifiable": {
                  "NewDomain": {
                    "id": {
                      "name": "Anicée_SCOP"
                    },
                    "logo": null,
                    "metadata": {}
                  }
                }
              }
            }
          }
        },
        {
          "Register": {
            "object": {
              "Raw": {
                "Identifiable": {
                  "NewAccount": {
                    "id": {
                      "name": "kpicard13",
                      "domain_id": {
                        "name": "Anicée_SCOP"
                      }
                    },
                    "signatories": [
                      "ed01202a400f423ff42e19fe77dc2908d5b3f0d0793663ff7e1eb6e9093d798694c377"
                    ],
                    "metadata": {}
                  }
                }
              }
            }
          }
        },
        {
          "Register": {
            "object": {
              "Raw": {
                "Identifiable": {
                  "AssetDefinition": {
                    "id": {
                      "name": "Labadie",
                      "domain_id": {
                        "name": "Matthias_SCOP"
                      }
                    },
                    "value_type": "Store",
                    "mintable": "Once",
                    "metadata": {}
                  }
                }
              }
            }
          }
        },
        {
          "Register": {
[...]

PR #2270, opened 2022-05-26
[feature] #1998: Add filters to queries

### LINKED ISSUE #1998: Add filters to queries
Since Queries are already paginated, we should make filters more general so that they can be composed with queries.

### LINKED ISSUE #2233: Replace simple-ish implementations of debug and display with derive
Cases where an `impl Debug` is needed just to skip a field can be greatly simplified.

appetrosyan:

### Description of the Change

- [x] Add traits to track common filter paradigms. 
- [x] Add module `predicate` for various filters
- [x] Add sensible `fmt::Display` implementations to various structs #2310 
- [x] Decouple `RegisteredWith` from `Identifiable` relevant for #2227 
- [x] Add enum to process strings by `contains`, `is` and `starts/ends_with`
- [x] Add enum to target `Identifiables`, 
- [x] Add enum for Display
- [x] Add enum variant to target metadata values
- [x] Add predicate enum
- [x] Add combinator enum (optional) (and/or/not)
- [x] Add timestamp enum
- [x] Add filtering step before pagination
- [x] Add unit tests to cover filter behaviour

### Issue

Closes #1998 
Closes #2233

### Benefits

Queries are now filterable server-side

### Possible Drawbacks

Extra processing

TODO!


data_model/src/predicate.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub trait Predicate{
+    type Input;
+
+    fn applies(&self, input: &Self::Input) -> bool;
+}
+
+pub struct DisplayContains(pub String);
+
+pub struct IdContains(pub String);
```
s8sato, line 11: What will be the format of the query then?
Maybe it consists of a target (e.g. `ValueType`) and predicates?

appetrosyan: The idea is that predicates are built up like legos. A predicate applies to a `Value`. It has an enum of targets: `metadata`, `id (Display)`, `self (Display)` and `self (Debug)` (which I'm not sure is a good idea to include). And then it also has an enum of operations: `fn (&str) -> bool` that map onto the string representations e.g. `is`, `contains`, and `not`. This should give you 12 filters at the cost of 7 with the added benefit of being easy to extend.

appetrosyan: We could then add combinators into the mix to have more fine-grained control.

data_model/src/lib.rs
```diff
@@ -37,35 +37,24 @@ pub mod metadata;
...
 pub enum MintabilityError {
     /// Tried to mint an Un-mintable asset.
+    #[display(fmt = "This asset cannot be minted more than once and it was already minted.")]
     MintUnmintable,
     /// Tried to forbid minting on assets that should be mintable.
+    #[display(fmt = "This asset was set as infinitely mintable. You cannot forbid its minting.")]
     ForbidMintOnMintable,
 }
 
-impl fmt::Display for MintabilityError {
```
mversic, line 55: this is where deriving `Display` is actually worth doing. Nice

data_model/src/lib.rs
```diff
@@ -336,24 +331,25 @@ pub enum IdBox {
 )]
 pub enum RegistrableBox {
```
mversic, line 332: I'm thinking I would rename to `RegisteredWithBox` to be consistent or rename `RegisterWith` trait to `Registrable`

data_model/src/lib.rs
```diff
@@ -389,6 +385,28 @@ pub enum IdentifiableBox {
     Role(Box<role::Role>),
 }
 
+impl RegisteredWith for trigger::Trigger<FilterBox> {
```
mversic, line 388: shouldn't this be moved into `trigger` module?

data_model/src/lib.rs
```diff
@@ -389,6 +385,28 @@ pub enum IdentifiableBox {
...
+    type RegisteredWith = Self;
+}
+
+impl IdentifiableBox {
+    fn contains(&self, content: &str) -> bool {
```
mversic, line 393: why do we need this? I suppose this will be revealed when more stuff is implemented

data_model/src/lib.rs
```diff
@@ -449,6 +467,36 @@ pub enum Value {
     Block(BlockValue),
 }
 
+impl fmt::Display for Value {
```
mversic, line 470: seems like a good candidate for deriving. Maybe it's not possible?

data_model/src/lib.rs
```diff
@@ -696,10 +744,21 @@ where
...
     /// Type of entity's identification.
     type Id: Into<IdBox> + Debug + Clone + Eq + Ord;
-    /// Type used to register `Identifiable` entity
+
+    fn id(&self) -> &Self::Id;
+}
+
+pub trait HasMetadata {
+    // type Metadata = metadata::Metadata;
+    // Uncomment when stable.
```
mversic, line 756:
```suggestion
    // TODO: Uncomment when stable.
```

you're saying we could have more fine grained `Metadata` types? Because not all types having metadata allow for the same `Value` types. If that's the case I'm all for it

data_model/src/trigger/mod.rs
```diff
@@ -141,34 +158,22 @@ pub struct Id {
...
-            name: Name::from_str(name)?,
-        })
-    }
-}
+//     fn from_str(name: &str) -> Result<Self, Self::Err> {
+//         Ok(Self {
+//             name: Name::from_str(name)?,
+//         })
+//     }
+// }
```
mversic, line 169: forgot to remove?

data_model/src/account.rs
```diff
@@ -212,6 +244,19 @@ pub struct Account {
...
+    fn id(&self) -> &Self::Id {
+        &self.id
+    }
+}
+
+impl HasMetadata for Account {
+    fn metadata(&self) -> &crate::metadata::Metadata {
+        &self.metadata
+    }
+}
```
mversic, line 257: We could consider making a proc derive macro for the `Identifiable`/`HasMetadata` if we'll rely on the convention that id field is named `id` and metadata field is named `metadata`

mversic (changes requested):

data_model/src/lib.rs
```diff
@@ -696,10 +744,21 @@ where
 }
 
 /// This trait marks entity that implement it as identifiable with an `Id` type to find them by.
-pub trait Identifiable: Debug + Clone {
+pub trait Identifiable: Debug {
     /// Type of entity's identification.
     type Id: Into<IdBox> + Debug + Clone + Eq + Ord;
-    /// Type used to register `Identifiable` entity
+
+    fn id(&self) -> &Self::Id;
```
mversic, line 751: just an afterthought, if it weren't for this method we could make a blanket impl like:
```rs
impl<T: RegisteredWith> Identifiable for T::RegisteredWith {
    type Id = <T as Identifiable>::Id;
}
```

data_model/src/account.rs
mversic: AFAIK this is not true only for `NewRole` at the moment

data_model/src/lib.rs
appetrosyan: That's an interesting observation.

data_model/src/account.rs
appetrosyan: I was considering that myself. This would also resolve the issue with `PartialEq` and `PartialOrd`.

data_model/src/trigger/mod.rs
appetrosyan: Yes. Might bring back. Draft.

data_model/src/lib.rs
appetrosyan: GAT FTW.

data_model/primitives/src/fixed.rs
```diff
@@ -44,6 +44,12 @@ pub type FixNum = fixnum::FixedPoint<Base, fixnum::typenum::U9>;
 )]
 pub struct Fixed(FixNum);
 
+impl core::fmt::Display for Fixed {
```
appetrosyan, line 47: Derive.

data_model/src/account.rs
```diff
@@ -110,13 +122,24 @@ impl Default for SignatureCheckCondition {
...
+)]
+#[display(fmt = "[{id}]")]
 pub struct NewAccount {
-    id: Id,
+    id: <NewAccount as Identifiable>::Id,
     signatories: Signatories,
     metadata: Metadata,
 }
 
+impl Identifiable for NewAccount {
```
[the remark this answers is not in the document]
appetrosyan: I want to differentiate between Identifiable and `Into<IdentifiableBox>`. Perhaps the name needs to be changed?

data_model/src/domain.rs
```diff
@@ -169,8 +188,28 @@ pub struct Domain {
     metadata: Metadata,
```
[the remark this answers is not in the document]
appetrosyan: I'd very much like to interact with the bindgen a little. Since i'll mostly be working on this over the weekend, I'll have a bunch of questions by Monday.

data_model/src/expression.rs
```diff
@@ -32,8 +33,11 @@ pub type Context = btree_map::BTreeMap<ValueName, Value>;
 pub type ExpressionBox = Box<Expression>;
 
 /// Struct for type checking and converting expression results.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode, Serialize, Deserialize)]
+#[derive(
+    Debug, Display, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode, Serialize, Deserialize,
+)]
 #[serde(transparent)]
+#[display(fmt = "Expressions aren't `fmt::Display` yet :(")] // TODO: implement
```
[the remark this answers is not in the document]
appetrosyan: Draft PR. I think there might be a nice way of representing the entire chain.

data_model/src/lib.rs
appetrosyan: `Registrable::With`? Shorter and we almost always use the `<Self as Registrable>::With`.

data_model/src/lib.rs
appetrosyan: This is necessary for the predicates. I havent pushed the function that uses it yet, since I'm still playing around with the API.

data_model/src/lib.rs
appetrosyan: We want to turn Metadata into a trait, that's implemented for certain value enums. The overall `Value` enum can be cast into the specific kind with type checking. This adds strong typing to kinds of metadata. 

Moreover, this will pave the way to more precise metadata e.g. exchange rates, accounts and such.

mversic: exactly, I agree. Though I think we can do this without GAT support

data_model/src/lib.rs
mversic: sounds good

data_model/src/account.rs
mversic: yeah, except that I initially thought those would be implemented with a declarative macro. Not sure, about a particular macro implementation, but `Identifiable` and `PartialEq`/`PartialOrd` impls are rather closely related.

data_model/src/predicate.rs
```diff
@@ -0,0 +1,412 @@
...
+use string::*;
+
+use super::*;
+use crate::{IdBox, Identifiable, IdentifiableBox, Name, Value};
+
+/// Predicate combinator enum.
+#[derive(
+    Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema, PartialEq, Eq, PartialOrd, Ord,
+)]
+pub enum PredicateBox {
```
mversic, line 13: this module looks very interesting. Will we be able to replace expressions with this if it proves worthy? At first glance their functionality seems on par

mversic (changes requested):

data_model/src/predicate.rs
appetrosyan: Basically, this is the prototype for the composable filters. The expressions API has some overlap which I intend to remove. Since we have WASM, some things are now redundant (in addition to being woefully inadequate).

crypto/src/lib.rs
```diff
@@ -357,9 +357,23 @@ impl From<multihash::ConvertError> for KeyParseError {
...
 
+// Cannot be derived with `derive_more`.
+// impl fmt::Debug for PublicKey {
+//     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+//         f.debug_struct("PublicKey")
+//             .field("digest_function", &self.digest_function())
+//             .field("payload", &hex::encode_upper(self.payload().as_slice()))
+//             .finish()
+//     }
+// }
```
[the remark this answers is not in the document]
appetrosyan: Apparently it worked. I'll run this through cargo hack and if it passes -- remove.

data_model/src/lib.rs
```diff
@@ -389,6 +382,28 @@ pub enum IdentifiableBox {
     Role(Box<role::Role>),
 }
 
+impl RegisteredWith for trigger::Trigger<FilterBox> {
+    type RegisteredWith = Self;
+}
+
+impl From<IdentifiableBox> for IdBox {
```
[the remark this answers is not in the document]
appetrosyan: Because the conversion is lossy? The other option is to extend the `id` definition to include a lifetime. In that case the IdentifiableBox is also another `Identifiable`, and `<IdentifiableBox as Identifiaible>::Id` is `IdBox`.

data_model/src/account.rs
appetrosyan: We have an issue for it. Resolving

data_model/src/query.rs
```diff
@@ -171,6 +175,8 @@ declare_versioned_with_scale!(VersionedPaginatedQueryResult 1..2, Debug, Clone,
 pub struct PaginatedQueryResult {
     /// The result of the query execution.
     pub result: QueryResult,
+    /// The filter that was applied to the Query result
+    pub filter: PredicateBox,
```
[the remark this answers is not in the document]
appetrosyan: Explained in doc comment.

data_model/src/predicate.rs
```diff
@@ -0,0 +1,852 @@
...
+            PredicateBox::Or(vector) => vector.iter().any(|pred| pred.applies(input)),
+            PredicateBox::Not(predicate) => !predicate.applies(input),
+            PredicateBox::Raw(predicate) => predicate.applies(input),
+        }
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::print_stdout, clippy::use_debug)]
+pub mod test {
```
mversic, line 74:
```suggestion
pub mod test {
    #![allow(clippy::print_stdout, clippy::use_debug)]

```

data_model/src/predicate.rs
```diff
@@ -0,0 +1,852 @@
...
+    pub enum Predicate {
+        /// Forward to [`str::contains()`]
+        Contains(String),
+        /// Forward to [`str::starts_with()`]
+        StartsWith(String),
+        /// Forward to [`str::ends_with()`]
+        EndsWith(String),
+        /// Forward to [`String`] equality.
+        Is(String),
+    }
```
mversic, line 141: alternative design could have been:
```rs
    pub enum PredicateType {
        Contains,
        StartsWith,
        EndsWith,
        Is,
    }
    pub struct Predicate {
        type_: PredicateType,
        operates_on: String,
    }
```

I'm not saying it would be better, just noting. It is good either way

data_model/src/predicate.rs
```diff
@@ -0,0 +1,852 @@
...
+        /// 128-bit
+        U128(SemiInterval<u128>),
+        /// Fixed-precision
+        Fixed(SemiInterval<Fixed>),
+    }
+
+    /// A trait to mark objects which should be treated as bounded unsigned values.
+    pub trait UnsignedMarker {
+        /// The maximum attainable value
+        const MAX: Self;
```
mversic, line 435: what about min?

mversic (changes requested):

cli/src/torii/routing.rs
```diff
@@ -116,20 +120,28 @@ pub(crate) async fn handle_queries<W: WorldTrait>(
...
-        Value::Vec(value.into_iter().paginate(pagination).collect())
+
+    let filtered_result = QueryResult(if let Value::Vec(value) = original_result {
+        Value::Vec(
+            value
+                .into_iter()
+                .filter(|val| filter.applies(val))
+                .paginate(pagination)
+                .collect(),
+        )
```
mversic, line 137: why is only `Value::Vec` filtered?

data_model/src/domain.rs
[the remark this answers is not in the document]
appetrosyan: Yeah, I'm just curious why it doesn't cause conflicts in the FFI.

data_model/src/trigger/mod.rs
appetrosyan: Fixed.

data_model/src/domain.rs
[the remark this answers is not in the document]
appetrosyan: Done

client/tests/mod.rs
```diff
@@ -1,3 +1,4 @@
 #[cfg(not(coverage))]
 mod integration;
+mod reload;
```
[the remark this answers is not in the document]
appetrosyan: This module is `reload.rs` the test for reloading of log level. Before this change it compiled and linked to a separate executable, which lead to bloat in `target`.

core/src/genesis.rs
```diff
@@ -91,9 +92,10 @@ pub trait GenesisNetworkTrait:
 }
 
 /// [`GenesisNetwork`] contains initial transactions and genesis setup related parameters.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Deref)]
 pub struct GenesisNetwork {
```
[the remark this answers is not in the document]
appetrosyan: No. The macro gets confused. We shouldn't need this after #2284.

data_model/src/predicate.rs
appetrosyan: Good thinking, it should help if we decide to extract them into their own files.

data_model/src/predicate.rs
appetrosyan: It generates the exact same assembly, but now you can access the string without destructuring the `Predicate`. You are also forced to make all variants of the same type. Also, you have one fewer struct in the namespace. 

I'm not saying it would be worse, just nothing that I considered it, and decided to go with the algebraic form.

data_model/src/predicate.rs
appetrosyan: It's `UnsignedMarker::ZERO`.

cli/src/torii/routing.rs
appetrosyan: Added explanation to `predicate.rs`. Basically, it's ambiguous and not necessary.

client/src/client.rs
```diff
@@ -604,6 +612,27 @@ impl Client {
         ))
     }
 
+    /// Create a request with pagination and add the filter.
+    ///
+    /// # Errors
+    /// Forwards from [`prepare_query_request`].
+    pub fn request_with_pagination_and_filter<R>(
```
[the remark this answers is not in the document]
appetrosyan: Done.

data_model/src/lib.rs
```diff
@@ -368,6 +376,30 @@ pub enum IdentifiableBox {
     Role(Box<role::Role>),
 }
 
+impl RegisteredWith for trigger::Trigger<FilterBox> {
+    type RegisteredWith = Self;
+}
+
+// TODO: think of a way to `impl Identifiable for IdentifiableBox`.
+// The main problem is lifetimes and conversion cost.
```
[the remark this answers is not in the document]
appetrosyan: Undecided. You make a good case.

mversic (approved):

mversic (dismissed):

data_model/src/metadata.rs
```diff
@@ -122,6 +128,11 @@ impl Metadata {
         map.get(key)
     }
 
+    /// Return if the metadata map contains the given key.
+    pub fn contains(&self, key: &Name) -> bool {
+        self.map.contains_key(key)
+    }
+
     /// Returns iterator over key - value pairs stored in [`Metadata`]
```
outoftardis, line 136:
```suggestion
    /// Return the iterator over key-value pairs stored in [`Metadata`].
```

data_model/src/metadata.rs
```diff
@@ -122,6 +128,11 @@ impl Metadata {
         map.get(key)
     }
 
+    /// Return if the metadata map contains the given key.
```
outoftardis, line 131: maybe check instead of return?

```suggestion
    /// Check if the metadata map contains the given key.
```

if you want to say return, then you need to say "return the Boolean value" or something along these lines

data_model/src/lib.rs
```diff
@@ -663,11 +719,27 @@ where
...
+    // type Metadata = metadata::Metadata;
+    // Uncomment when stable.
+
+    /// The metadata associated to this object.
+    fn metadata(&self) -> &metadata::Metadata;
+}
+
+/// Trait for objects that are registered by proxy.
+pub trait Registered: Identifiable {
+    /// The type that is used to register this entity with.
```
outoftardis, line 741:
```suggestion
    /// The type that is used to register this entity.
```

don't really need the `with` here

data_model/src/lib.rs
```diff
@@ -663,11 +719,27 @@ where
...
 /// This trait marks entity that implement it as identifiable with an `Id` type to find them by.
-pub trait Identifiable: Debug + Clone {
+pub trait Identifiable: Debug {
     /// Type of entity's identification.
-    type Id: Into<IdBox> + Debug + Clone + Eq + Ord;
-    /// Type used to register `Identifiable` entity
-    type RegisteredWith: Into<RegistrableBox>;
+    type Id: Into<IdBox> + fmt::Display + fmt::Debug + Clone + Eq + Ord;
+
+    /// Return reference to the type's identification.
```
outoftardis, line 726:
```suggestion
    /// Return the reference to the type identification.
```

data_model/src/lib.rs
```diff
@@ -663,11 +719,27 @@ where
 }
 
 /// This trait marks entity that implement it as identifiable with an `Id` type to find them by.
-pub trait Identifiable: Debug + Clone {
+pub trait Identifiable: Debug {
     /// Type of entity's identification.
```
outoftardis, line 723:
```suggestion
    /// The type of entity identification.
```

core/src/sumeragi/view_change.rs
```diff
@@ -152,26 +153,19 @@ pub struct ProofPayload {
...
-#[derive(Debug, Clone, Decode, Encode, FromVariant, IntoSchema)]
+#[derive(Debug, Display, Clone, Decode, Encode, FromVariant, IntoSchema)]
 pub enum Reason {
-    /// Proxy tail have not committed a block in time.
+    /// Proxy tail have not committed the block in time.
+    #[display(fmt = "Commit timeout")]
     CommitTimeout(CommitTimeout),
-    /// Transaction was sent to leader, but no corresponding receipt was received from the leader for it.
+    /// Transaction was sent to the leader and no corresponding receipt was received from the leader for it.
+    #[display(fmt = "No transaction receipt received")]
```
outoftardis, line 162:
```suggestion
    #[display(fmt = "Received no transaction receipt")]
```
```suggestion
    #[display(fmt = "No transaction receipt was received")]
```

data_model/src/predicate.rs
```diff
@@ -0,0 +1,872 @@
...
+    }
+
+    /// Predicate useful for working with containers. Currently only [`Metadata`](crate::metadata::Metadata) and [`Vec<Value>`] are supported.
+    #[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema)]
+    pub enum Container {
+        /// Forward to [`Iterator::any`]
+        Any(Box<Predicate>),
+        /// Forward to [`Iterator::all`]
+        All(Box<Predicate>),
+        /// Apply predicate to the [`Value`] element at the index.
```
outoftardis, line 745:
```suggestion
        /// Apply the predicate to the [`Value`] element at the index.
```

data_model/src/predicate.rs
```diff
@@ -0,0 +1,872 @@
...
+    }
+
+    /// A predicate that targets the particular `key` of a collection.
+    #[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema)]
+    pub struct ValueOfKey {
+        key: Name,
+        predicate: Box<Predicate>,
+    }
+
+    /// Predicate useful for working with containers. Currently only [`Metadata`](crate::metadata::Metadata) and [`Vec<Value>`] are supported.
```
outoftardis, line 738:
```suggestion
    /// The predicate that is useful for working with containers. Currently only [`Metadata`](crate::metadata::Metadata) and [`Vec<Value>`] are supported.
```

data_model/src/predicate.rs
```diff
@@ -0,0 +1,872 @@
...
+    pub enum Predicate {
+        /// Apply predicate to the [`Identifiable::Id`] and/or [`IdBox`].
+        Identifiable(string::Predicate),
+        /// Apply predicate to the container.
+        Container(Container),
+        /// Apply predicate to the [`<Value as Display>::to_string`](ToString::to_string()) representation.
+        Display(string::Predicate),
+        /// Apply predicate to the numerical value.
+        Numerical(numerical::Range),
+        /// Timestamp (currently for `BlockValue`) only.
```
outoftardis, line 610:
```suggestion
        /// Timestamp (currently for `BlockValue` only).
```

data_model/src/predicate.rs
```diff
@@ -0,0 +1,872 @@
...
+    /// A predicate designed for general processing of `Value`.
+    #[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema)]
+    pub enum Predicate {
+        /// Apply predicate to the [`Identifiable::Id`] and/or [`IdBox`].
+        Identifiable(string::Predicate),
+        /// Apply predicate to the container.
+        Container(Container),
+        /// Apply predicate to the [`<Value as Display>::to_string`](ToString::to_string()) representation.
+        Display(string::Predicate),
+        /// Apply predicate to the numerical value.
```
outoftardis, line 608:
```suggestion
        /// Apply the predicate to the [`Identifiable::Id`] and/or [`IdBox`].
        Identifiable(string::Predicate),
        /// Apply the predicate to the container.
        Container(Container),
        /// Apply the predicate to the [`<Value as Display>::to_string`](ToString::to_string()) representation.
        Display(string::Predicate),
        /// Apply the predicate to the numerical value.
```

data_model/src/predicate.rs
```diff
@@ -0,0 +1,872 @@
...
+            PredicateBox::And(_)
+        ));
+    }
+}
+
+pub mod string {
+    //! String-related predicates and implementations.
+    use super::*;
+
+    /// Predicate useful for processing [`String`]s and [`Name`]s.
```
outoftardis, line 148:
```suggestion
    /// The predicate that is useful for processing [`String`]s and [`Name`]s.
```

data_model/src/metadata.rs
outoftardis: just a note: I've changed `returns` to `return` here just for consistency based on other files in this PR, but probably the 3rd person is more appropriate and it's better to make the change in other places instead (e.g. `construct` -> `constructs`)

I'm not sure what is more prevalent in your code base, I'm noticing the use of both verb forms, and we'd better be consistent

data_model/src/predicate.rs
```diff
@@ -0,0 +1,872 @@
...
+        /// Forward to [`str::ends_with()`]
+        EndsWith(String),
+        /// Forward to [`String`] equality.
+        Is(String),
+    }
+
+    impl Predicate {
+        /// Construct the [`Self::Contains`] variant
+        #[inline]
+        pub fn contains(predicate: &str) -> Self {
```
QuentinI, line 164: Why not something like

```suggestion
        pub fn contains(predicate: impl Into<String>) -> Self {
```

data_model/primitives/src/fixed.rs
```diff
@@ -7,6 +7,7 @@ use derive_more::Display;
 use fixnum::{
     ops::{CheckedAdd, CheckedSub, Zero},
     ArithmeticError,
+    _priv::Bounded,
```
QuentinI, line 10: It's available in [ops](https://docs.rs/fixnum/0.6.1/fixnum/ops/trait.Bounded.html), i think

data_model/src/predicate.rs
QuentinI: Yeah, I get that. What I meant is why not also accept owned `String` in short-hand functions?

mversic: I might not understand what is being proposed here but I think `impl Into<String>` is something you'll almost never need. It's just complicating the API. Function argument should either be owned `String` or `&str` slice. Converting structures into String is usually done just for display

What maybe could be done is to make use of the `Borrow` trait but that seems too restrictive

data_model/src/isi.rs
```diff
@@ -372,17 +428,17 @@ where
...
-    pub fn new(object: O::RegisteredWith) -> Self {
+    pub fn new(object: O::With) -> Self {
         Register { object }
     }
 }
 
 impl<O> Unregister<O>
 where
-    O: Identifiable,
+    O: Registered,
```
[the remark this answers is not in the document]
mversic: `Identifiable` is supertrait of `Registered` and therefore implied

data_model/src/predicate.rs
QuentinI: > Function argument should either be owned `String` or `&str` slice.

I'm proposing accepting both here. 
I don't actually have a strong opinion on this anyway, so feel free to mark resolved.

data_model/primitives/src/fixed.rs
appetrosyan: Good point

data_model/src/predicate.rs
appetrosyan: Again, if you want to use a string slice, you use the lowercase function and `.to_owned()` is implied. 

If you want to use the owned string object, you construct/`into()` in place using the upper-case function. 

We support both uses.

data_model/src/metadata.rs
appetrosyan: No. Functions do things ⇒ their docstrings should be imperative. This is the convention in Python, Lisp and the Linux kernel.

QuentinI (approved):

mversic (approved):

PR #2274, opened 2022-05-26
[feature]: #0000 Handle id and shared FFI fns

mversic:


ffi/src/lib.rs
```diff
@@ -25,3 +35,150 @@ pub enum FfiResult {
...
+        pub unsafe extern "C" fn __clone(
+            handle_id: $crate::HandleId,
+            handle_ptr: *const core::ffi::c_void,
+            output_ptr: *mut *mut core::ffi::c_void
+        ) -> $crate::FfiResult {
+            gen_ffi_impl!{@null_check_stmts handle_ptr, output_ptr}
+
+            // NOTE: Unreachable pattern occurs if at least two handle ids resolve to
+            // the same value. Denying it reduces chance of such implementation error
+            #[deny(unreachable_patterns)]
```
mversic, line 77: I don't understand why this lint doesn't activate when I'm compiling Iroha. I can implement this check manually, but this seemed really convenient and short

data_model/src/expression.rs
```diff
@@ -305,9 +286,7 @@ impl From<Mod> for ExpressionBox {
 
 /// Evaluates to the right expression in power of left expressions.
 /// Works only for `Value::U32`
-#[derive(
-    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Decode, Encode, Deserialize, Serialize, IntoSchema,
-)]
+#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
 pub struct RaiseTo {
```
appetrosyan, line 290: I always found the inclusion of this operation baffling. I'm 100% confident we need to groom the expression API.

data_model/src/lib.rs
```diff
@@ -458,8 +426,7 @@ impl Value {
         match self {
             U32(_) | U128(_) | Id(_) | PublicKey(_) | Bool(_) | Parameter(_) | Identifiable(_)
             | String(_) | Name(_) | Fixed(_) | TransactionValue(_) | PermissionToken(_)
-            | Hash(_) => 1_usize,
-            Block(v) => v.nested_len() + 1_usize,
+            | Hash(_) | Block(_) => 1_usize,
```
appetrosyan, line 429: So a block is a sing,e entity, rather than a collection of entitites. This was an extremely subtle mistake. Good catch.

mversic: yeah, I remember commenting on the PR but I wasn't certain at the time what's correct

appetrosyan (dismissed):

ffi/src/lib.rs
```diff
@@ -16,6 +18,8 @@ pub struct Pair<K, V>(pub K, pub V);
 // u32/i32, u64/i64 natively. Otherwise, `repr(i8)` would suffice
 #[repr(i32)]
 pub enum FfiResult {
+    /// Handle id doesn't identify any known handle
```
outoftardis, line 21:
```suggestion
    /// Handle id doesn't identify any known handles
```

ffi/tests/gen_shared_fns.rs
```diff
@@ -0,0 +1,83 @@
...
+
+handles! {0, FfiStruct1, FfiStruct2}
+gen_ffi_impl! {Drop: FfiStruct1, FfiStruct2}
+gen_ffi_impl! {Clone: FfiStruct1, FfiStruct2}
+gen_ffi_impl! {Eq: FfiStruct1, FfiStruct2}
+gen_ffi_impl! {Ord: FfiStruct1, FfiStruct2}
+
+#[ffi_bindgen]
+impl FfiStruct1 {
+    /// New
```
outoftardis, line 25: I'm curious if this type of description is enough. I'd suggest having something more descriptive

appetrosyan (approved):

appetrosyan (dismissed):

ffi/derive/src/bindgen.rs
```diff
@@ -272,17 +273,34 @@ fn gen_ptr_null_check_stmt(fn_arg_descriptor: &FnArgDescriptor) -> Option<syn::S
...
         gen_slice_len_arg_name(fn_arg_descriptor),
     );
 
     parse_quote! {
-        if #slice_len_arg_name == 0_usize {
-            // NOTE: `slice::from_raw_parts` takes a non-null aligned pointer
-            #arg_name = core::ptr::NonNull::dangling().as_ptr();
-        }
+        let #arg_name = if #slice_len_arg_name == 0_usize {
+            core::ptr::NonNull::dangling().as_ptr()
```
s8sato, line 302: It's up to the macro composer to properly initialize this pointer, right?

ffi/src/lib.rs
```diff
@@ -25,3 +29,144 @@ pub enum FfiResult {
...
+    ( $id:expr, $ty:ty $(, $other:ty)* $(,)? ) => {
+        impl Handle for $ty {
+            const ID: $crate::HandleId = $id;
+        }
+
+        $crate::handles! {$id + 1, $( $other, )*}
+    };
+    ( $id:expr, $(,)? ) => {
+        /// Represents handle in an FFI context
+        pub trait Handle {
```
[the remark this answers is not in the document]
mversic: > My understanding is that checking the handle ID serves as a monomorphization from generics to a specific type

exactly, I'll add the doc comment

ffi/derive/src/bindgen.rs
mversic: I would say yes. Consider the following example:

```rs
fn call_slicer() {
	let arr: [u8; 0] = [];

	// arr.as_ptr doesn't return null pointer. Therefore, it can be used with `core::slice::from_raw_parts`
	slicer(arr.as_ptr(), arr.len());
	
	// if null pointer is given then it cannot be used to construct zero-sized slice with `core::slice::from_raw_parts`
	// since this API is quite natural for a user who wishes to skip the use of the slice `slice_ptr` must be assigned non-null dangling value
	slicer(core::ptr::null(), 0);
}

extern fn slicer(slice_ptr: *const u8, slice_len: usize);
```

s8sato: Okay, 0-length slices are interpreted as dangling pointers and then always restored by `slice::from_raw_parts`, both in arguments and in return values

mversic: it would be a logical error to call `slice::from_raw_parts` with null pointer

s8sato: So we assign dangling pointers to 0-length slices so as not to be considered null

s8sato (approved):

appetrosyan (approved):

PR #2278, opened 2022-05-27
[ci]: DOPS-1789 Update Dockerfiles

Cre-eD:
# Description of Change

Move single-stage implementation of docker containers in Iroha v2 CI pipeline to a multistage assembly and testing with multiple layers. 

- [x] Introduce alpine final docker image
- [ ] Create env container for CI.
- [ ] Set up automatic action for building and pushing the env container from unprotected branches. 
- [ ] Build iroha binaries in GHA CI. 
- [ ] Test the built binaries using the `check.sh` and `test/*.sh`
- [ ] Build Iroha binaries with the `musl` target. 
- [ ] Copy and deploy the binaries in a single container versioned according to the branch it gets pushed from.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,60 +1,28 @@
-ARG BASE_IMAGE=ubuntu:20.04
-FROM $BASE_IMAGE AS rust-base
+FROM iroha2:build AS builder
 
-ENV CARGO_HOME=/cargo_home \
-    RUSTUP_HOME=/rustup_home \
-    DEBIAN_FRONTEND=noninteractive
-ENV PATH="$CARGO_HOME/bin:$PATH"
+FROM alpine:3.16
 
-RUN set -ex; \
-    apt-get update  -yq; \
-    apt-get install -y --no-install-recommends curl apt-utils; \
-    apt-get install -y --no-install-recommends \
-        build-essential \
-        ca-certificates \
-        libssl-dev \
-        clang \
-        pkg-config \
-        llvm-dev; \
-    rm -rf /var/lib/apt/lists/*
+ENV  GLIBC_REPO=https://github.com/sgerrand/alpine-pkg-glibc
+ENV  GLIBC_VERSION=2.30-r0
+ENV  BIN_PATH=/usr/local/bin/
+ENV  CONFIG_DIR=config
+ENV  IROHA2_CONFIG_PATH=$CONFIG_DIR/config.json
+ENV  IROHA2_GENESIS_PATH=$CONFIG_DIR/genesis.json
+ARG  TARGET_DIR=/iroha/target/release
 
-ARG TOOLCHAIN=stable
-RUN set -ex; \
-    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs >/tmp/rustup.sh; \
-    sh /tmp/rustup.sh -y --no-modify-path --default-toolchain "$TOOLCHAIN"; \
-    rm /tmp/*.sh
-
-RUN set -ex; \
-    rustup toolchain install --profile default nightly-2022-04-20; \
-    rustup target add wasm32-unknown-unknown; \
-    rustup component add rust-src
-
-FROM rust-base as cargo-chef
-RUN cargo install cargo-chef
-
-FROM cargo-chef as planner
-WORKDIR /iroha
-COPY . .
-RUN cargo chef prepare --recipe-path recipe.json
-
-FROM cargo-chef as builder
-ARG PROFILE
-WORKDIR /iroha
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,60 +1,28 @@
...
-    rm -rf /var/lib/apt/lists/*
-ENV IROHA_TARGET_BIN=$BIN
-ENV IROHA2_CONFIG_PATH=$CONFIG_DIR/config.json
-ENV IROHA2_GENESIS_PATH=$CONFIG_DIR/genesis.json
-CMD ./$IROHA_TARGET_BIN
+FROM iroha2:build AS builder
+
+FROM alpine:3.16
+
+ENV  GLIBC_REPO=https://github.com/sgerrand/alpine-pkg-glibc
```
appetrosyan, line 5: We need to adjust the compilation to add the musl libc instead. the GLibC is not necessary for Iroha.

Dockerfile.ci
```diff
@@ -23,45 +22,42 @@ RUN set -eux; \
...
-    esac; \
-    url="https://static.rust-lang.org/rustup/archive/1.24.3/${rustArch}/rustup-init"; \
-    wget "$url"; \
-    echo "${rustupSha256} *rustup-init" | sha256sum -c -; \
-    chmod +x rustup-init; \
-    ./rustup-init -y --no-modify-path --profile minimal --default-toolchain $RUST_VERSION --default-host ${rustArch}; \
-    rm rustup-init; \
-    chmod -R a+w $RUSTUP_HOME $CARGO_HOME; \
+    esac && \
+    url="https://static.rust-lang.org/rustup/archive/1.24.3/${rustArch}/rustup-init" && \
```
appetrosyan, line 34: Why use && instead of ;?

Dockerfile.ci
```diff
@@ -23,45 +22,42 @@ RUN set -eux; \
...
+    ./rustup-init -y --no-modify-path --profile minimal --default-toolchain $RUST_VERSION --default-host ${rustArch} && \
+    rm rustup-init && \
+    chmod -R a+w $RUSTUP_HOME $CARGO_HOME && \
     rm -rf /var/lib/apt/lists/*
 
-RUN git clone https://github.com/rui314/mold.git; \
-    git -C "./mold" checkout v1.2.0; \
-    make --directory="./mold" CXX=clang++; \
-    make --directory="./mold" install; \
+RUN git clone https://github.com/rui314/mold.git && \
```
appetrosyan, line 43: **Note to self**: Check if bullseye has mold in its repos.

docker-compose-local.yml
```diff
@@ -3,10 +3,8 @@ services:
   iroha0:
     build: .
     image: iroha0:debug
-    volumes:
-      - './configs/peer:/config'
-      - './:/root/soramitsu/iroha'
     environment:
+      KURA_BLOCK_STORE_PATH: /chain
```
appetrosyan, line 7: Double check if this needs to be quoted. AFAIK, serde requires it to be double quoted.

Dockerfile.ci
Cre-eD: it's best practice
In case with ; (Semi-colon) operator - the execution of the second command is independent of the exit status of the first command. If the first command does not get successfully executed, then also the second command will get executed.

docker-compose-local.yml
Cre-eD: It works

Dockerfile
```diff
@@ -1,60 +1,28 @@
...
+ENV  GLIBC_VERSION=2.30-r0
+ENV  BIN_PATH=/usr/local/bin/
+ENV  CONFIG_DIR=config
+ENV  IROHA2_CONFIG_PATH=$CONFIG_DIR/config.json
+ENV  IROHA2_GENESIS_PATH=$CONFIG_DIR/genesis.json
+ARG  TARGET_DIR=/iroha/target/release
+
+RUN  set -ex && \
+     apk --update add libstdc++ curl ca-certificates && \
```
Cre-eD, line 14:
```suggestion
RUN  apk --update add libstdc++ curl ca-certificates && \
```

Dockerfile.ci
Cre-eD: set is redundant, must be removed

appetrosyan: Good to know. Thank you for clarifying.

Dockerfile.ci
appetrosyan: No. Debian 11 does not have any version in its repos. Debian 12, Debian unstable and Devuan all have 1.2.1

appetrosyan:
Closed in favour of #2433

Issue #2279, opened 2022-05-27
Retain error types in the API response


s8sato:
Currently, only query errors retain their type in the API response.
If other Torii errors do as well instead of being stringified, the client can restore the error response to `torii::Error` or an equivalent type.
This allows for more reliable error handling

s8sato:
Could be replaced by #1455 in the future

Issue #2111, opened 2022-04-16
`roles` feature flag in `data_model` causes serialization disparity


mversic:
`roles` feature flag affects the layout of enums/structures in `data_model` because it conditionally includes/excludes variants/fields. Due to this, a difference can arise in serialized format on client as compared to the server which basically makes them unable to exchange some messages. Example:

```rs
pub enum WorldEvent {
    Peer(peer::PeerEvent),
    Domain(domain::DomainEvent),
    #[cfg(feature = "roles")]
    Role(role::RoleEvent),
    Trigger(trigger::TriggerEvent),
}
```

This can be fixed in following ways:
1. remove `roles` feature flag from all crates 
always include roles
2. remove `roles` feature flag from `data_model` crate only
don't include roles in your data structures but always include them in the serialized form. Queries can work without problems with server just empty lists objects for roles. Instructions on roles must fail with an error. I'm not sure this is possible without implying the 1st point to always include roles. This possibility requires more consideration
3. Define initialization handshake
Client will ask for the data layout information in the first message and adapt it's subsequent messages according to it. This effectively makes `roles` a compile flag on the server(as is currently) but a runtime flag on the client which adds strain on the sdk development
4. Include information about `roles` support in the schema
I mention this possibility just out of completeness. I maintain that the 3rd point should take precedence compared to this one.  

Relates to #1969 
Could relate to #2112

Issue #1854, opened 2022-01-30
CI checks don't catch unused imports


appetrosyan:
CI checks ignore unused imports in blocks marked with `#[cfg(not(feature = "std"))]`.

PR #2281, opened 2022-05-27
[refactor]: Refactor around `QueryError`

s8sato:
### Description of the Change
- [refactor]
    - d9e4712f3ce44bf6f3b62861e8a30e607e21c306 Remove unused `query::Error` variants
    - 3554726991d421c6800e93a7a7428aac4c9f9308 Remove recovery from useless `Rejection`
    - 1f1028045b91b27bcd07aafb295e241624a47742 Un-implement unused `Clone`
    - 770168c69a4f6215f73f258d06ade747830f12ae Replace `#[source]` in a single field with `#[from]`
- minor [test]
    - 30f262e2aa669a6b127d5e203ab437a0f55896ee Add to `torii::tests`
- minor [fix]
    - ab382ddbbf1e112ec5484ba0721133e7fbbec7f0 Resolve apparent mismatch between `VersionedPaginatedQueryResult` and `VersionedQueryResult`
- minor [documentation]
    - ab492d748334d7caeb2d15bdb8f505856524f2fb Fix description for `Status` response
### Issue
None

### Benefits
Trimmed code

### Possible Drawbacks
None

### DIFF
--- a/cli/src/torii/mod.rs
+++ b/cli/src/torii/mod.rs
@@ -18,7 +18,6 @@ use tokio::sync::Notify;
 use utils::*;
 use warp::{
     http::StatusCode,
-    reject::Rejection,
     reply::{self, Json, Response},
     ws::{WebSocket, Ws},
     Filter as _, Reply,
@@ -96,43 +95,57 @@ pub(crate) const fn query_status_code(query_error: &query::Error) -> StatusCode 
 
 impl Reply for Error {
     fn into_response(self) -> Response {
-        const fn status_code(err: &Error) -> StatusCode {
-            use Error::*;
-            match err {
-                Query(e) => query_status_code(e),
-                VersionedTransaction(_)
-                | AcceptTransaction(_)
-                | RequestPendingTransactions(_)
-                | DecodeRequestPendingTransactions(_)
-                | EncodePendingTransactions(_)
-                | ConfigurationReload(_)
-                | TxTooBig => StatusCode::BAD_REQUEST,
-                Config(_) => StatusCode::NOT_FOUND,
-                PushIntoQueue(err) => match **err {
-                    queue::Error::Full => StatusCode::INTERNAL_SERVER_ERROR,
-                    queue::Error::SignatureCondition(_) => StatusCode::UNAUTHORIZED,
-                    _ => StatusCode::BAD_REQUEST,
-                },
-                #[cfg(feature = "telemetry")]
-                Prometheus(_) | Status(_) => StatusCode::INTERNAL_SERVER_ERROR,
+        use Error::*;
+        match self {
+            Query(err) => {
+                reply::with_status(utils::Scale(&err), query_status_code(&err)).into_response()
+            }
+            // TODO Have a type-preserved response body instead of a stringified one #2279
+            VersionedTransaction(err)
+            | DecodeRequestPendingTransactions(err)
+            | EncodePendingTransactions(err) => {
+                reply::with_status(err.to_string(), err.status_code()).into_response()
+            }
+            _ => reply::with_status(Self::to_string(&self), self.status_code()).into_response(),
+        }
+    }
+}
+
+impl Error {
+    const fn status_code(&self) -> StatusCode {
+        use Error::*;
+        match self {
+            Query(e) => query_status_code(e),
+            VersionedTransaction(err)
+            | DecodeRequestPendingTransactions(err)
+            | EncodePendingTransactions(err) => err.status_code(),
+            AcceptTransaction(_)
+            | RequestPendingTransactions(_)
+            | ConfigurationReload(_)
+            | TxTooBig => StatusCode::BAD_REQUEST,
+            Config(_) => StatusCode::NOT_FOUND,
+            PushIntoQueue(err) => match **err {
+                queue::Error::Full => StatusCode::INTERNAL_SERVER_ERROR,
+                queue::Error::SignatureCondition(_) => StatusCode::UNAUTHORIZED,
+                _ => StatusCode::BAD_REQUEST,
+            },
+            #[cfg(feature = "telemetry")]
+            Prometheus(_) | Status(_) => StatusCode::INTERNAL_SERVER_ERROR,
+        }
+    }
+
+    fn to_string(mut err: &dyn std::error::Error) -> String {
+        let mut s = "Error:\n".to_owned();
+        let mut idx = 0_i32;
+
+        loop {
+            s += &format!("    {}: {}\n", idx, &err.to_string());
+            idx += 1_i32;
+            match err.source() {
+                Some(e) => err = e,
+                None => return s,
             }
         }
-
-        fn to_string(mut err: &dyn std::error::Error) -> String {
-            let mut s = "Error:\n".to_owned();
-            let mut idx = 0_i32;
-
-            loop {
-                s += &format!("    {}: {}\n", idx, &err.to_string());
-                idx += 1_i32;
-                match err.source() {
-                    Some(e) => err = e,
-                    None => return s,
-                }
-            }
-        }
-
-        reply::with_status(to_string(&self), status_code(&self)).into_response()
     }
 }
 
--- a/cli/src/torii/mod.rs
+++ b/cli/src/torii/mod.rs
@@ -86,7 +86,7 @@ pub enum Error {
 pub(crate) const fn query_status_code(query_error: &query::Error) -> StatusCode {
     use query::Error::*;
     match query_error {
-        Decode(_) | Version(_) | Evaluate(_) | Conversion(_) => StatusCode::BAD_REQUEST,
+        Decode(_) | Evaluate(_) | Conversion(_) => StatusCode::BAD_REQUEST,
         Signature(_) => StatusCode::UNAUTHORIZED,
         Permission(_) => StatusCode::FORBIDDEN,
         Find(_) => StatusCode::NOT_FOUND,
--- a/core/src/smartcontracts/isi/asset.rs
+++ b/core/src/smartcontracts/isi/asset.rs
@@ -74,7 +74,7 @@ pub mod isi {
                     .map_err(|e| Error::Conversion(e.to_string()))?;
                 *quantity = quantity
                     .checked_add(self.object)
-                    .ok_or(Error::Math(MathError::Overflow))?;
+                    .ok_or(MathError::Overflow)?;
                 wsv.metrics.tx_amounts.observe(f64::from(*quantity));
 
                 Ok(AssetEvent::Added(asset_id.clone()))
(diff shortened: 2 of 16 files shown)


version/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub mod error {
     use super::UnsupportedVersion;
 
     /// Versioning errors
-    #[derive(Debug, Clone, FromVariant, IntoSchema)]
```
appetrosyan, line 38: What's the rationale?

docs/source/references/api_spec.md
```diff
@@ -225,9 +225,9 @@ Also returns current status of peer in json string:
 - 200 OK - reports status:
   + Number of connected peers, except for the reporting peer itself
   + Number of committed blocks (block height)
-  + Total number of transactions
-  + `uptime` since creation of the genesis block in milliseconds.
-  + Number of view_changes in the current round
+  + Total number of accepted and rejected transactions
+  + Uptime since creation of the genesis block.
```
appetrosyan, line 229: Preserve the unit.

core/src/smartcontracts/isi/query.rs
```diff
@@ -33,33 +33,12 @@ impl ValidQueryRequest {
     }
 }
 
-/// Unsupported version error
```
appetrosyan, line 36: While I think we might need to bring this back when we have fully-functional versioning, removing it for the time being is fine.

core/src/smartcontracts/isi/mod.rs
```diff
@@ -213,14 +211,14 @@ pub mod error {
     impl From<FixedPointOperationError> for Error {
         fn from(err: FixedPointOperationError) -> Self {
             match err {
-                FixedPointOperationError::NegativeValue(_) => Self::Math(MathError::NegativeValue),
```
appetrosyan, line 216: On this occasion only, I would prefer remaining explicit and not relying on the `From` implementations generated by `FromVariant`.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -74,7 +74,7 @@ pub mod isi {
                     .map_err(|e| Error::Conversion(e.to_string()))?;
                 *quantity = quantity
                     .checked_add(self.object)
-                    .ok_or(Error::Math(MathError::Overflow))?;
+                    .ok_or(MathError::Overflow)?;
```
appetrosyan, line 77: While this is less explicit, the `?` does the `Into` conversion anyway.

version/src/lib.rs
appetrosyan: Also in a few places you removed the `Clone` implementation from Error variants. Why is that?

core/src/smartcontracts/isi/asset.rs
s8sato: Yeah, since `#[from] MathError` implementation is seen in `Self::Error` declaration, I think this is enough readable

version/src/lib.rs
s8sato: The intention of the removal of these `Clone` implementations is just a revert against what I added [earlier](https://github.com/hyperledger/iroha/pull/1867) but later found unnecessary.
Indeed, I did not touch `UnsupportedVersion` and `RawVersioned` at that time. Fixed

version/src/lib.rs
```diff
@@ -48,10 +47,8 @@ pub mod error {
         /// Cannot encode unsupported version from Parity SCALE to JSON
         UnsupportedScaleEncode,
         /// JSON (de)serialization issue
-        #[cfg(feature = "json")]
         Serde,
         /// Parity SCALE (de)serialization issue
-        #[cfg(feature = "scale")]
         ParityScale,
```
s8sato, line 52: Removed these attributes for the same reason as `#[cfg(feature = "role")]`, namely to fix the serialized format of the enum

core/src/smartcontracts/isi/query.rs
s8sato: Then we can utilize `iroha_version::error::Error::UnsupportedVersion`.
The idea is that Version error is considered a part of Decode error

version/src/lib.rs
```diff
@@ -8,10 +8,9 @@
 #[cfg(not(feature = "std"))]
 extern crate alloc;
 
-// TODO: #1854, CI doesn't catch errors with unused imports in this block.
```
s8sato, line 11: #1854 is resolved, right? @appetrosyan

version/src/lib.rs
s8sato: The intention of the removal of these `Clone` implementations is just a revert against what I added [earlier](https://github.com/hyperledger/iroha/pull/1867) but later found unnecessary.
Indeed, I did not touch `UnsupportedVersion` and `RawVersioned` at that time. Fixed

core/src/smartcontracts/isi/query.rs
s8sato: Then we can utilize `iroha_version::error::Error::UnsupportedVersion`.
The idea is that Version error is considered a part of Decode error

core/src/smartcontracts/isi/asset.rs
s8sato: Yeah, since `#[from] MathError` implementation is seen in `Self::Error` declaration, I think this is enough readable

version/src/lib.rs
```diff
@@ -48,10 +47,8 @@ pub mod error {
         /// Cannot encode unsupported version from Parity SCALE to JSON
         UnsupportedScaleEncode,
         /// JSON (de)serialization issue
-        #[cfg(feature = "json")]
         Serde,
         /// Parity SCALE (de)serialization issue
-        #[cfg(feature = "scale")]
```
mversic, line 54: @s8sato removed your comment accidentally. The comment said that this was removed to prevent structural inconsistencies between client and server as we've seen in #2111 

Good thinking

s8sato: > Removed these attributes for the same reason as `#[cfg(feature = "role")]`, namely to fix the serialized format of the enum

docs/source/references/api_spec.md
```diff
@@ -242,6 +242,7 @@ Also returns current status of peer in json string:
     "view_changes": 0
 }
 ```
+- __CAUTION__: Almost all fields are 64-bit integers and should be handled with care in JavaScript
```
outoftardis, line 245: can we be more specific about which fields are 64-bit int?

core/src/smartcontracts/isi/query.rs
appetrosyan: This is a bit of a problem in the code base in general. We have double coverage because one particular thing is not general-enough.

version/src/lib.rs
appetrosyan: Fixed since 1.60.

appetrosyan (dismissed):

core/src/smartcontracts/isi/query.rs
s8sato: I think keeping unused `QueryError::Version(UnsupportedVersionError)` is just misleading
because we already use `QueryError::Decode(iroha_version::error::Error::UnsupportedVersion(UnsupportedVersion))` for such cases

docs/source/references/api_spec.md
s8sato: Fixed to encourage readers to refer actual `Status` struct, but since the struct is defined nested, also added details here

docs/source/references/api_spec.md
```diff
@@ -242,6 +242,7 @@ Also returns current status of peer in json string:
     "view_changes": 0
 }
 ```
+- __CAUTION__: Almost all fields are 64-bit integers and should be handled with care in JavaScript. Only `nanos` field is 32-bit integer. See `iroha_telemetry::metrics::Status`
```
outoftardis, line 245:
```suggestion
__CAUTION__: Almost all fields are 64-bit integers and should be handled with care in JavaScript. Only the `nanos` field is 32-bit integer. See `iroha_telemetry::metrics::Status`.
```

cli/src/torii/mod.rs
```diff
@@ -96,43 +95,57 @@ pub(crate) const fn query_status_code(query_error: &query::Error) -> StatusCode
...
+                queue::Error::Full => StatusCode::INTERNAL_SERVER_ERROR,
+                queue::Error::SignatureCondition(_) => StatusCode::UNAUTHORIZED,
+                _ => StatusCode::BAD_REQUEST,
+            },
+            #[cfg(feature = "telemetry")]
+            Prometheus(_) | Status(_) => StatusCode::INTERNAL_SERVER_ERROR,
+        }
+    }
+
+    fn to_string(mut err: &dyn std::error::Error) -> String {
```
Arjentix, line 137: Is it something different from [`ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html)?

cli/src/torii/mod.rs
```diff
@@ -96,43 +95,57 @@ pub(crate) const fn query_status_code(query_error: &query::Error) -> StatusCode
...
-                    None => return s,
-                }
+        loop {
+            s += &format!("    {}: {}\n", idx, &err.to_string());
+            idx += 1_i32;
+            match err.source() {
+                Some(e) => err = e,
+                None => return s,
             }
         }
```
Arjentix, line 148:
```suggestion
        let mut err_opt = Some(err);
        while let Some(e) = err_opt {
            s += &format!("    {}: {}\n", idx, &e.to_string());
            idx += 1_i32;
            err_opt = e.source()
        }
```

Arjentix: Using explicit condition in loop is just my preference. This should do the same and it's more clear to me

core/src/smartcontracts/isi/query.rs
```diff
@@ -33,33 +33,12 @@ impl ValidQueryRequest {
...
-
-impl UnsupportedVersionError {
-    /// Expected version
-    pub const fn expected_version() -> u8 {
-        1
-    }
-}
-
 /// Query errors.
-#[derive(Error, Debug, Clone, Decode, Encode, IntoSchema)]
```
Arjentix, line 55: Not sure if it's a good idea: see this [guideline](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits)
From the other side we can always bring it back if we need it.

Me personally always implement `Debug` and `Clone` for all types

Arjentix (approved):

core/src/smartcontracts/isi/query.rs
s8sato: Yes, we can impl `Clone` when needed

cli/src/torii/mod.rs
s8sato:
```Rust
#[test]
fn test_to_string() {
    let err = Error::PushIntoQueue(Box::new(queue::Error::SignatureCondition(eyre!("msg"))));
    let str1 = Error::to_string(&err);
    let str2 = err.to_string();
    println!("{str1}");
    println!("{str2}");
}
```
`str1`:
```
Error:
    0: Failed to push into queue
    1: Failure during signature condition execution
    2: msg
```
`str2`:
```
Failed to push into queue
```
The equivalent of `str1` can be achieved by modifying `Display` impl, such as `#[error("Failed to push into queue: {0}")]`.
But I leave `Error::to_string` as it is because the string representation of Torii errors in general is off-topic of this PR and will be removed in #2279

appetrosyan (approved):

Arjentix (approved):

Issue #2282, opened 2022-05-27
improve FFI derive from getset


mversic:
Current implementation of deriving FFI functions from `getset` is inadequate. Improve the implementation

PR #2283, opened 2022-05-27
[fix] #2282: improve FFI derives from getset implementation

mversic:

### DIFF
--- a/ffi/derive/src/derive.rs
+++ b/ffi/derive/src/derive.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashSet, fmt};
+use std::collections::HashSet;
 
 use proc_macro2::TokenStream;
 use proc_macro_error::{abort, OptionExt};
@@ -7,7 +7,7 @@ use syn::{parse_quote, visit_mut::VisitMut, Ident, ItemStruct, Type};
 
 use crate::{get_ident, impl_visitor::SelfResolver};
 
-/// What does structure derives for public fields
+/// Type of accessor method derived for a structure
 #[derive(Clone, Copy, PartialEq, Eq, Hash)]
 enum Derive {
     Set,
@@ -19,7 +19,7 @@ enum Derive {
 pub fn gen_fns_from_derives(item: &ItemStruct) -> Vec<syn::ItemFn> {
     let struct_derives = parse_derives(&item.attrs).unwrap_or_default();
 
-    let mut ffi_derives = vec![];
+    let mut ffi_derives = Vec::new();
     match &item.fields {
         syn::Fields::Named(syn::FieldsNamed { named, .. }) => named.iter().for_each(|field| {
             if let Some(mut field_derives) = parse_derives(&field.attrs) {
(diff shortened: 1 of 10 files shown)


ffi/derive/src/derive.rs
```diff
@@ -0,0 +1,225 @@
+use std::collections::HashSet;
+
+use proc_macro2::TokenStream;
+use proc_macro_error::{abort, OptionExt};
+use quote::quote;
+use syn::{parse_quote, visit_mut::VisitMut, Ident, ItemStruct, Type};
+
+use crate::{get_ident, impl_visitor::SelfResolver};
+
+/// What does structure derives for public fields
```
appetrosyan, line 10: Docs a little unclear.

appetrosyan:
```suggestion
/// Enumeration of  which types of accessor methods could or have been derived by getset for this structure. 
```

ffi/derive/src/derive.rs
```diff
@@ -0,0 +1,225 @@
...
+    Set,
+    Get,
+    GetMut,
+}
+
+/// Generate FFI function equivalents of derived methods
+pub fn gen_fns_from_derives(item: &ItemStruct) -> Vec<syn::ItemFn> {
+    let struct_derives = parse_derives(&item.attrs).unwrap_or_default();
+
+    let mut ffi_derives = vec![];
```
appetrosyan, line 22:
```suggestion
    let mut ffi_derives = Vec::new();
```

ffi/derive/src/derive.rs
```diff
@@ -0,0 +1,225 @@
...
+            Derive::Get => format!("{}", field_name),
+            Derive::GetMut => format!("{}_mut", field_name),
+        },
+        proc_macro2::Span::call_site(),
+    )
+}
+
+fn gen_ffi_docs(struct_name: &Ident, derive_method_name: &syn::Ident) -> String {
+    format!(
+        " FFI function equivalent of [`{}::{}`]",
```
appetrosyan, line 95:
```suggestion
        "FFI function equivalent of [`{}::{}`]",
```
Or I may be wrong, but I don't think that the single space is intentional.

mversic: it's intentional because this without the leading space, the `#doc` macro is expanded into:
```
///FFI function equivalent of [`{}::{}`]"
```

ffi/derive/src/derive.rs
mversic: I don't think there is any difference. Do you think it's a bit faster to compile if it's not a macro expansion?

appetrosyan: I doubt it adds more than a few microseconds to the compile time. it's more about communicating the  intention.

ffi/derive/src/derive.rs
appetrosyan: Good to know. I'd add a comment about it.

ffi/derive/src/derive.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// Parses getset attributes to find out which methods it derives
+fn parse_derives(attrs: &[syn::Attribute]) -> Option<HashSet<Derive>> {
+    let mut getset_attrs = HashSet::new();
+
+    for attr in attrs {
+        if let Ok(syn::Meta::List(meta_list)) = attr.parse_meta() {
+            if !meta_list.path.is_ident("getset") {
+                continue;
+            }
```
Arjentix, line 49: What about something like this?

```rust
attrs
    .iter()
    .filter_map(|attr|
        if let Ok(syn::Meta::List(meta_list) == attr.parse_meta() {
            Some(meta_list)
        } else {
            None
        }
    )
    .filter(|meta_list| meta_list.path.is_ident("getset"))
```

ffi/tests/bindgen_derive.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let mut id = MaybeUninit::<*mut u32>::new(core::ptr::null_mut());
+    let mut name = MaybeUninit::<*const Name>::new(core::ptr::null());
+    let mut is_null = MaybeUninit::<*const Name>::new(core::ptr::null());
+
+    unsafe {
+        //FfiStruct__set_id(ffi_struct, 42_u32);
+        FfiStruct__id_mut(ffi_struct, id.as_mut_ptr());
+        let id = &mut *id.assume_init();
+        assert_eq!(&mut 0, id);
```
Arjentix, line 38: At first I was thinking that it's a strange way to check for a null pointer. And then I realized that it's just a reference to id. Maybe use another number instead of 0?

Arjentix (dismissed):
I was surprised, that I understood pretty everything in someone else procedural macro code. Good code quality!

ffi/tests/bindgen_derive.rs
mversic: ok, makes sense if it confuses the reader

ffi/derive/src/derive.rs
mversic: ok

Arjentix (approved):

Arjentix (approved):

Issue #2199, opened 2022-05-12
Add permissions to genesis block


appetrosyan:
We hard-code the `AcceptAll` permission mode, but Iroha supports multiple permission modes, which need to be specified somewhere. As the genesis round is the one where no permission validation takes place, it makes sense to deserialise the permissions set from the genesis block, either as a string (for built-in), or as a WASM binary blob (see #1618 )

PR #2284, opened 2022-05-27
[refactor]: Remove generics from `IsAllowedBoxed`

Arjentix:
### Description of the Change

While working on #2199 me and @appetrosyan have decided to split this work on several PRs, so that it will be easier to review.

We can't properly deserialize generics so we decided to make them be enum instead (at least there aren't much of things that needs permissions). It looks not so pretty, but it *much* easier to use with deserialization.

`permissions.rs` file now is very big, so I believe we should split it later. I don't do it now so that it will be easier to review (I hope)

Another important thing -- `World`s... The idea of mocking `World` with `WorldTrait` is quite painfull... I have created a new `MockWorld` but it's not used anywere yet. I'm not sure if we really need it.
Removing `W: WorldTrait` generic in some places and leaving it in another places causes me to use very wierd `W == World` runtime checks in 3 places. But it works fine for now.
The whole mocking architecture should be refactored here -- @appetrosyan  has some ideas.

### Next steps

After merging this PR I'm gonna work on actual validators deserialization. It's probably gonna be another config named `permissions.json`. It will be implemented with another enum containing all possible validators from `permission_validators` folder.

### Issue

* Refers to #2199

### Benefits

* Easier to deserialize validators in the future

### Possible Drawbacks

* More code doing pretty the same as before
* Wierd `W == World` checks in 3 places


config/derive/src/lib.rs
```diff
@@ -270,7 +270,7 @@ fn impl_get_recursive(
             fn get_recursive<'a, T>(
                 &self,
                 inner_field: T,
-            ) -> iroha_config::BoxedFuture<'a, Result<serde_json::Value, Self::Error>>
+            ) -> iroha_config::BoxedFuture<'a, std::result::Result<serde_json::Value, Self::Error>>
```
appetrosyan, line 273:
```suggestion
            ) -> iroha_config::BoxedFuture<'a, core::result::Result<serde_json::Value, Self::Error>>
```

config/derive/src/lib.rs
```diff
@@ -310,7 +310,7 @@ fn impl_get_recursive(
         fn get_recursive<'a, T>(
             &self,
             inner_field: T,
-        ) -> Result<serde_json::Value, Self::Error>
+        ) -> std::result::Result<serde_json::Value, Self::Error>
```
appetrosyan, line 313:
```suggestion
        ) -> core::result::Result<serde_json::Value, Self::Error>
```

macro/derive/src/lib.rs
```diff
@@ -8,20 +8,26 @@ use quote::quote;
 /// Attribute for skipping from attribute
 const SKIP_FROM_ATTR: &str = "skip_from";
 const SKIP_TRY_FROM_ATTR: &str = "skip_try_from";
+/// Attribute to skip inner container optimization. Useful for trait objects
```
appetrosyan, line 11: In the future, these sorts of small changes can be merged separately. It can help with the reviewing process.

cli/src/torii/routing.rs
```diff
@@ -110,14 +113,26 @@ pub(crate) async fn handle_instructions<W: WorldTrait>(
...
 #[iroha_futures::telemetry_future]
+#[allow(unsafe_code, clippy::unimplemented)]
 pub(crate) async fn handle_queries<W: WorldTrait>(
     wsv: Arc<WorldStateView<W>>,
-    query_validator: Arc<IsQueryAllowedBoxed<W>>,
+    query_validator: Arc<IsQueryAllowedBoxed>,
     pagination: Pagination,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>, warp::http::Response<warp::hyper::Body>> {
+    // TODO: Very dirty, need to do something with it
```
appetrosyan, line 123: Need a ticket. We also need to understand both why this was necessary, and what needs to be done better.

permissions_validators/src/public_blockchain/burn.rs
```diff
@@ -50,15 +50,17 @@ impl<W: WorldTrait> IsAllowed<W, Instruction> for OnlyAssetsCreatedByThisAccount
...
+                .into());
         }
         Ok(())
     }
 }
 
 /// Allows burning assets from a different account than the creator's of this asset if the corresponding user granted the permission token
 /// for a specific asset.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Copy, Clone, Serialize)]
```
appetrosyan, line 63:
```suggestion
#[derive(Debug, Clone, Copy, Serialize)]
```

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -20,65 +23,362 @@ impl NeedsPermission for QueryBox {}
...
+    ) -> std::result::Result<(), ValidatorTypeMismatch> {
+        if self != another {
+            return Err(ValidatorTypeMismatch::expected(another).found(self));
+        }
+
+        Ok(())
+    }
+}
+
+impl std::fmt::Display for ValidatorType {
```
appetrosyan, line 54: `derive_more`?

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -20,65 +23,362 @@ impl NeedsPermission for QueryBox {}
...
+    }
+
+    /// Wrong validator expectation error
+    ///
+    /// I.e. used when user tries to validate [`QueryBox`](super::QueryBox) with
+    /// [`IsAllowedBoxed`](super::IsAllowedBoxed) containing
+    /// [`IsAllowedBoxed::Instruction`](super::IsAllowedBoxed::Instruction) variant
+    #[derive(Debug, Copy, Clone, thiserror::Error)]
+    #[error("Expected `{expected}` validator type, but found `{found}`")]
+    pub struct ValidatorTypeMismatch {
```
appetrosyan, line 97: We also have a generic `Mismatch<T>` in the `data_model`. We could add a few trait bounds and perhaps make it produce the kind of output that you want, using e.g. `std::any::type`'s string representation.

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -20,65 +23,362 @@ impl NeedsPermission for QueryBox {}
...
+        expected: ValidatorType,
+        found: ValidatorType,
+    }
+
+    impl ValidatorTypeMismatch {
+        /// Start construction of [`ValidatorTypeMismatch`] by providing *expected* type
+        pub fn expected(expected: ValidatorType) -> Expected {
```
appetrosyan, line 104: This is why Rust would have benefitted from currying and partial application.

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -20,65 +23,362 @@ impl NeedsPermission for QueryBox {}
...
+}
+
+use error::*;
+
+/// Result type for permission validators
+pub type Result<T> = std::result::Result<T, DenialReason>;
 
 /// Implement this to provide custom permission checks for the Iroha based blockchain.
-pub trait IsAllowed<W: WorldTrait, O: NeedsPermission> {
+pub trait IsAllowed<W: WorldTrait, O: NeedsPermission>:
```
appetrosyan, line 132: !!! **Most important part of the PR**

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -20,65 +23,362 @@ impl NeedsPermission for QueryBox {}
...
+            #[cfg(test)]
+            IsExpressionAllowedBoxed::Mock(_) => unimplemented!(),
+        }
+    }
+}
 
 /// Trait for joining validators with `or` method, auto-implemented
-/// for all types which convert to [`IsAllowedBoxed`].
-pub trait ValidatorApplyOr<W: WorldTrait, O: NeedsPermission> {
+/// for all types which are convertible to something implementing [`IsAllowed`]
```
appetrosyan, line 303:
```suggestion
/// for all types which are convertible to a concrete type implementing [`IsAllowed`]
```

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -20,65 +23,362 @@ impl NeedsPermission for QueryBox {}
...
 {
-    fn or(self, another: impl Into<IsAllowedBoxed<W, O>>) -> Or<W, O> {
+    fn or(self, another: impl Into<V>) -> Or<W, O, V> {
+        Or::new(self, another)
+    }
+}
+
+/// `check` succeeds if either `first` or `second` validator succeeds.
+#[derive(Debug, Clone, Serialize)]
+pub struct Or<W: WorldTrait, O: NeedsPermission, V: IsAllowed<W, O>> {
```
appetrosyan, line 322: These will need to be replaced by generic, general-purpose combinators. We should start work on them after the LTS release.

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -368,52 +699,209 @@ fn check_query_in_instruction<W: WorldTrait>(
...
-pub struct AllShouldSucceed<W: WorldTrait, O: NeedsPermission> {
-    validators: Vec<IsAllowedBoxed<W, O>>,
+#[derive(Debug, Clone, Serialize)]
+pub struct AllShouldSucceed {
+    validators: Vec<IsAllowedBoxed>,
 }
 
-impl<W: WorldTrait, O: NeedsPermission> IsAllowed<W, O> for AllShouldSucceed<W, O> {
+impl AllShouldSucceed {
+    /// Creates new [`AllShouldSucceed`]
```
appetrosyan, line 728:
```suggestion
    /// Create new [`AllShouldSucceed`]
```

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -368,52 +699,209 @@ fn check_query_in_instruction<W: WorldTrait>(
...
+    /// Type of the first element in the `validators` is considered exemplary
+    pub fn new(validators: Vec<IsAllowedBoxed>) -> Result<Self> {
+        check_all_validators_have_the_same_type(&validators)?;
+        Ok(Self { validators })
+    }
+
+    fn check_type(&self, validator_type: ValidatorType) -> Result<()> {
+        if let Ok(self_type) = self.validator_type() {
+            if self_type != validator_type {
+                return Err(ValidatorTypeMismatch::expected(validator_type)
```
appetrosyan, line 741: The reason why I would prefer the usage of the generic is also because of the syntax: 
Compare: 
```rust
Mismatch<ValidatorType> {
   expected: validator_type,
   actual: self_type
}

```

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -422,74 +910,206 @@ impl<W: WorldTrait, O: NeedsPermission> IsAllowed<W, O> for AnyShouldSucceed<W,
...
+        .into())
     }
 }
 
-impl<W: WorldTrait, O: NeedsPermission + 'static> From<AnyShouldSucceed<W, O>>
-    for IsAllowedBoxed<W, O>
-{
-    fn from(validator: AnyShouldSucceed<W, O>) -> Self {
-        Box::new(validator)
+impl IsAllowed<World, QueryBox> for AnyShouldSucceed {
```
appetrosyan, line 918: This file should probably be split into modules (which could be their own files).

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -525,63 +1145,113 @@ impl<W: WorldTrait, O: NeedsPermission> IsAllowed<W, O> for AllowAll {
         _authority: &AccountId,
         _instruction: &O,
         _wsv: &WorldStateView<W>,
-    ) -> Result<(), DenialReason> {
+    ) -> Result<()> {
         Ok(())
     }
 }
 
+impl From<AllowAll> for IsInstructionAllowedBoxed {
```
appetrosyan, line 1153: These could (and probably should) be implemented in a declarative macro. Suggested syntax
```rust
impl_from_for_allowed_boxed! {
     AllowAll => IsInstructionAllowedBoxed,
     AllowAll => IsQueryAllowedBoxed,
     AllowAll => IsExpressionAllowedBoxed,
}

File: core/src/smartcontracts/isi/permissions.rs [hunk 16]
```diff
@@ -525,63 +1145,113 @@ impl<W: WorldTrait, O: NeedsPermission> IsAllowed<W, O> for AllowAll {
...
+pub trait HasToken<W: WorldTrait>: Debug + dyn_clone::DynClone + erased_serde::Serialize {
     /// This function should return the token that `authority` should
     /// possess, given the `instruction` they are planning to execute
     /// on the current state of `wsv`
     ///
     /// # Errors
     ///
     /// In the case when it is impossible to deduce the required token
-    /// given current data (e.g. unexistent account or unaplicable
+    /// given current data (e.g. inexistent account or unapplicable
```
appetrosyan, line 1219:
```suggestion
    /// given current data (e.g. non-existent account or inapplicable
```

core/src/smartcontracts/isi/permissions.rs
```diff
@@ -735,7 +1450,7 @@ fn unpack_if_role_grant<W: WorldTrait>(
 pub fn unpack_if_role_revoke<W: WorldTrait>(
     instruction: Instruction,
     wsv: &WorldStateView<W>,
-) -> Result<Vec<Instruction>> {
+) -> eyre::Result<Vec<Instruction>> {
```
appetrosyan, line 1453: Like this.

macro/derive/src/lib.rs
Arjentix: Okay, understood

permissions_validators/src/private_blockchain/query.rs
```diff
@@ -340,23 +361,23 @@ impl<W: WorldTrait> IsAllowed<W, QueryBox> for OnlyAccountsData {
...
-                Err("Only access to your account-local data is permitted.".to_owned())
+                Err("Only access to your account-local data is permitted.".to_owned().into())
             }
             FindTriggerById(query) => {
                 // TODO: should differentiate between global and domain-local triggers.
                 let id = query
                     .id
                     .evaluate(wsv, &context)
-                    .map_err(|e| e.to_string())?;
+                    .map_err(|err| err.to_string())?;
```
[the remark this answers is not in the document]
Arjentix: Unfortunately `ToString::to_string` does not work, cause closure should take parameter by value, but `ToString::to_string` takes it by reference

core/src/smartcontracts/isi/permissions.rs
Arjentix: Yeah, it adds `Debug`, `Clone` and `Serialize` bounds. Not sure if we need it now, but it can be useful later

cli/src/torii/routing.rs
Arjentix: Done: #2292

core/src/smartcontracts/isi/permissions.rs
Arjentix: I think `Mismatch<T>` should be fine just with additional info in `DenialReason`

core/src/smartcontracts/isi/permissions.rs
Arjentix: Yeah, I was thinking about accidental parameter swapping and completely forgot about this construction syntax

mversic (changes requested):

cli/src/lib.rs
```diff
@@ -187,11 +188,23 @@ where
         );
 
         let query_validator = Arc::new(query_validator);
+
+        // TODO #2292: Very dirty, need to do something with it
+        let world_wsv = if std::any::TypeId::of::<W>() == std::any::TypeId::of::<World>() {
```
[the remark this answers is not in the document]
mversic: why are you're getting rid of the generic here?

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+        ValidatorTypeMismatch(#[from] ValidatorTypeMismatch),
+        /// Variant for custom error
+        #[error("{0}")]
+        Custom(String),
+        /// Variant used when at least one [`Validator`](super::IsAllowed) should be provided
+        #[error("No validators provided")]
+        NoValidatorsProvided,
+    }
+
+    impl From<String> for DenialReason {
```
mversic, line 99: how about implementing `FromStr` instead?

mversic: it might conflict with `thiserror`?

mversic (changes requested):

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -0,0 +1,205 @@
...
+
+/// Checks the [`RevokeBox`] instruction.
+pub trait IsRevokeAllowed<W: WorldTrait>:
+    Debug + dyn_clone::DynClone + erased_serde::Serialize
+{
+    /// Checks the [`RevokeBox`] instruction.
+    ///
+    /// # Errors
+    /// If this validator doesn't approve this Revoke instruction.
+    fn check_revoke(
```
mversic, line 73: let's just call it `check`, revoke is implied by the type

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -0,0 +1,205 @@
...
+
+/// Checks the [`GrantBox`] instruction.
+pub trait IsGrantAllowed<W: WorldTrait>:
+    Debug + dyn_clone::DynClone + erased_serde::Serialize
+{
+    /// Checks the [`GrantBox`] instruction.
+    ///
+    /// # Errors
+    /// If this validator doesn't approve this Grant instruction.
+    fn check_grant(
```
mversic, line 28: let's just call it check, grant is implied by the type

core/src/smartcontracts/isi/permissions/roles.rs
mversic: if we have `IsAllowed` do we need this trait as well?

core/src/smartcontracts/isi/permissions/roles.rs
mversic: if we have `IsAllowed` do we need this trait as well?

mversic (changes requested):

cli/src/lib.rs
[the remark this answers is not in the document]
mversic: Why do we have `WorldTrait`/`KuraTrait`/`SumeragiTrait`/etc.? I thought this was to be able to swap out different parts of Iroha with different implementations, i.e. to give the blockchain user the ability to customize their blockchain

macro/derive/src/lib.rs
```diff
@@ -34,7 +40,7 @@ const SKIP_TRY_FROM_ATTR: &str = "skip_try_from";
 ///     }
 /// }
 /// ```
-#[proc_macro_derive(FromVariant, attributes(skip_from, skip_try_from))]
+#[proc_macro_derive(FromVariant, attributes(skip_from, skip_try_from, skip_container))]
```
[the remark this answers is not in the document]
appetrosyan: `derive_more` doesn't unbox structures, as such we would need to either have explicit `impl From<Box<Variant>>` or pre-box the structures at the call site.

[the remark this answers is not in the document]
appetrosyan: Relevant [docs](https://jeltef.github.io/derive_more/derive_more/from.html) and [discussion](https://github.com/JelteF/derive_more/issues/153)

cli/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > why are you're getting rid of the generic here?

Because it's impossible to work with generics and deserialization at the same time. And after removing in one place (or giving the ability to work only on some specific types implementing trait) it's unwinding through all the project. I'd like to remove `WorldTrait` at all. See also #2292 


> Why do we have WorldTrait/KuraTrait/SumeragiTrait/etc.? I thought this was to be able to swap out different parts of Iroha with different implementations, i.e. to give the blockchain user the ability to customize their blockchain

I can't say about others, but as I understand, `WorldTrait` was introduced just for mocking purposes...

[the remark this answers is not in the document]
mversic: If it's just mocking, I'd try to get rid of all these traits and do it through some form of dependency injection. But is it just for mocking?

core/src/smartcontracts/isi/permissions/builder.rs
```diff
@@ -0,0 +1,106 @@
+//! This module contains validator builder to construct complex validator
+
+use super::{
+    combinators::{AllShouldSucceed, AnyShouldSucceed, CheckNested},
+    *,
+};
+
+/// Builder to combine multiple validation checks into one.
+#[derive(Debug, Copy, Clone)]
+pub struct ValidatorBuilder;
```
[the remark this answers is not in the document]
Arjentix: Well, yes... It was like that before my changes. Although usage looks pretty clear. Building is done by `all_should_succeed()` or `any_should_succeed()`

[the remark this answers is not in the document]
Arjentix: Do you have an idea how to improve it?

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: I don't think it may help. It is often used with runtime formatted `String`.
So with `FromStr` it will be dereferenced as `&str` and then I have to clone it on heap with `to_owned()`. So it will be double memory allocation? I think it's not good. Better to have one allocation and one move as it is now

macro/derive/src/lib.rs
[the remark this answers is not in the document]
Arjentix: I like the idea of using this crate

cli/src/lib.rs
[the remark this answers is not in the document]
Arjentix: @appetrosyan , need your answers here, please

cli/src/torii/routing.rs
```diff
@@ -66,7 +69,7 @@ impl VerifiedQueryRequest {
         }
         query_validator
             .check(&self.payload.account_id, &self.payload.query, wsv)
-            .map_err(QueryError::Permission)?;
+            .map_err(|e| QueryError::Permission(e.to_string()))?;
```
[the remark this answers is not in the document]
Arjentix: I did it, it was possible. Good suggestion!
Actually, why `QueryError` doesn't derive `thiserror::Error`? I think, it would be usefull

cli/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: I don't think that we at any point will have more than one legitimate `World` implementation. They could be versioned, yes, but that system doesn't require generics and traits. 

Dependency injection is probably the right approach, but only if we decide to postpone this PR until after the LTS.

core/src/smartcontracts/isi/permissions/builder.rs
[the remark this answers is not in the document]
appetrosyan: Alias. `::default()` and `::new()` are often doing the same thing. Ideally we should have a `Builder` trait that has an associated type and an associated `build()` method, in that case the API is consistently coherent.

client/tests/integration/roles.rs
```diff
@@ -4,7 +4,7 @@ use std::{str::FromStr as _, time::Duration};
 
 use eyre::{eyre, Result};
 use iroha_client::client::{self, Client};
-use iroha_core::{prelude::AllowAll, smartcontracts::permissions::ValidatorBuilder};
+use iroha_core::{prelude::AllowAll, smartcontracts::permissions::builder::ValidatorBuilder};
```
appetrosyan, line 7: I think we shouldn't repeat the name-space in the name. So either 
`permissions::builder::Validator` or `permissions::ValidatorBuilder` is preferred.

config/derive/src/lib.rs
```diff
@@ -166,7 +166,7 @@ fn impl_load_env(
     quote! {
         fn load_environment(
             &'_ mut self
-        ) -> std::result::Result<(), iroha_config::derive::Error> {
+        ) -> core::result::Result<(), iroha_config::derive::Error> {
```
appetrosyan, line 169: This change isn't significant as `config::derive` doesn't need to be `no_std`.

cli/src/lib.rs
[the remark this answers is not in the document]
mversic: ok, and we should assume 3rd parties will not swap in their own implementations of parts of Iroha? What about other traits such as `SumeragiTrait` or `KuraTrait`? Should we aim to get rid of those. I would if their only purpose is mocking, or at least use `#[cfg(test)]` to exclude them from the public API

core/test_network/src/lib.rs
```diff
@@ -436,8 +436,8 @@ where
         &mut self,
         configuration: Configuration,
         genesis: Option<G>,
-        instruction_validator: impl Into<IsInstructionAllowedBoxed<W>> + Send + 'static,
-        query_validator: impl Into<IsQueryAllowedBoxed<W>> + Send + 'static,
+        instruction_validator: impl Into<IsInstructionAllowedBoxed> + Send + 'static,
```
pesterev, line 439: By the way, I've just realized that the `WorldTrait` generic is not needed inside `Network` and `Peer` structures and can be hard-coded [here](https://github.com/hyperledger/iroha/blob/iroha2-dev/core/test_network/src/lib.rs#L462) because the tests use the default `World` structure.

core/src/smartcontracts/isi/permissions/mod.rs
mversic: ok, as you wish. I don't think performance is crucial here for the construction of error variant and `FromStr` seemed more idiomatic to me

cli/src/lib.rs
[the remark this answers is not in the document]
s8sato: At least `SumeragiTrait` could be related to the [Consensus API](https://wiki.hyperledger.org/display/iroha/Consensus+API) idea in the future. In that case, we would have to extract the logics common to consensus to the trait methods

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: Hm.... Ok, let's do more ideomatic

Arjentix: I think, let's keep both

Arjentix: Because `FromStr` returns `Result` which is not very useful here

core/Cargo.toml
```diff
@@ -64,6 +66,12 @@ wasmtime = "0.36.0"
 
 # transitive dependencies
 anyhow = ">= 1.0"
+dyn-clone = "1.0.5"
+erased-serde = "0.3.20"
```
mversic, line 70: I don't like seeing introduced two completely new dependencies. Could you explain why it's required? It seems to me that now we have two approaches in the codebase for serialization of trait objects. The one where we make boxed enums with variants of all types implementing a trait like `IdentifiableBox` and the 2nd approach introduced here

core/Cargo.toml
```diff
@@ -27,6 +27,8 @@ dex = []
 dev-telemetry = ["telemetry", "iroha_telemetry/dev-telemetry"]
 # Support Prometheus metrics. See https://prometheus.io/.
 expensive-telemetry = ["iroha_telemetry/metric-instrumentation"]
+# Support for mocking World object
+mock_world = ["mockall"]
```
mversic, line 31: can we name it just `mock`? so that later we can add other mocks, not only `World`

core/src/wsv.rs
```diff
@@ -33,53 +27,112 @@ pub type NewBlockNotificationSender = tokio::sync::watch::Sender<()>;
...
+            fn clone(&self) -> Self;
+        }
     }
 }
 
-impl DerefMut for World {
-    #[inline]
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        self
+cfg_if::cfg_if! {
```
mversic, line 122: checkout [mockall_double](https://docs.rs/mockall_double/0.3.0/mockall_double/)

Arjentix: Yeah, I know about it, but I also need to `pub use World as OriginalWorld` which `mockall_double` does not provide

core/Cargo.toml
Arjentix: Yeah, you are right. This was my first attempt which failed at deserialization step. But `Clone` and `Serialize` for trait objects like `IsAllowedBoxed` should work fine.

Although it seems like we don't need it, cause it the future there will be an enum for Genesis permissions (de-)serialization.

I'll remove this things then

core/test_network/tests/integration/sumeragi_with_mock.rs
```diff
@@ -218,31 +220,12 @@ pub mod utils {
...
-        };
+        use std::str::FromStr as _;
 
-        use iroha_core::{prelude::*, tx::Domain, wsv::WorldTrait};
+        use iroha_core::{prelude::*, tx::Domain};
         use iroha_data_model::prelude::*;
         use once_cell::sync::Lazy;
 
-        #[derive(Debug, Clone, Default)]
-        pub struct WithAlice(World);
```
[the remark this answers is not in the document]
Arjentix: 1. Bacause `WithAlice` adds aditional steps in `with()` method
2. That's strange. You can try replace `<Network<WithAlice, ...>>` with `<Network<World, ...>>` and see if it changes anything

Anyway there is no `WithAlice` in my changes anymore

core/Cargo.toml
mversic: I think that if you're mocking an object defined in a create you depend on, the proper way is to make a trait inside the current crate and make you're mock implementation there, not in the dependency crate. Exporting cargo feature that defines mocked objects seems like a misuse of the cargo. In this case it would mean that all mock would be defined in the `iroha_client`. This doesn't mean that you cannot create mock objects inside `iroha_core`, it's just that those will only be usable in unit tests of that crate

core/test_network/src/lib.rs
Arjentix: Yeah, now I removed `WorldTrait` generic from everywhere and it works fine

core/src/wsv.rs
mversic: okay

core/test_network/tests/integration/sumeragi_with_mock.rs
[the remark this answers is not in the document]
mversic: 2. as you can see [here](https://github.com/hyperledger/iroha/pull/2347) all is green because `World` was never actually mocked. It was there just to fulfill some type requirement. So this is why I don't see a reason to mock the `World` at all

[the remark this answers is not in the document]
Arjentix: Hmm, can it be green just because this tests are ignoreed:

```rust
#[ignore = "mock"]
```
?

Although I'm agree. After this changes I can see that we don't need mocking for running integrations tests. Mocking was only done in `sumeragi_with_fault.rs` but was never runned in CI/CD (however I don't know why `cargo test --all-features` skips it).
If we don't need this tests I can remove them and World mocking at all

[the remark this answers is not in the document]
mversic: ah, I see now, you're right. These tests are ignored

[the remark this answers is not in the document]
mversic: does anyone know why these tests are ignored? if not, we'll need mocking most probably

core/src/wsv.rs
```diff
@@ -33,53 +28,54 @@ pub type NewBlockNotificationSender = tokio::sync::watch::Sender<()>;
...
-        trusted_peers_ids: impl IntoIterator<Item = PeerId>,
-    ) -> Self;
-}
-
 /// The global entity consisting of `domains`, `triggers` and etc.
 /// For example registration of domain, will have this as an ISI target.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default, Clone, Getters)]
+#[cfg_attr(feature = "mock", allow(dead_code))]
+#[getset(get = "pub")]
```
mversic, line 35: I'm very much against using getters/setters in rust unless it's not possible to get around. This pattern is just not well suited for Rust because you loose the ability to transfer ownership or destructure or take disjoint references, etc. It was a different case in the `data_model` because there getters/setters are required to translate into FFI calls in the background

core/src/wsv.rs
```diff
@@ -33,53 +28,54 @@ pub type NewBlockNotificationSender = tokio::sync::watch::Sender<()>;
...
 pub struct World {
     /// Iroha parameters.
-    pub parameters: Vec<Parameter>,
+    parameters: Vec<Parameter>,
```
mversic, line 38: If you insist on using `getset` here you will get `fn parameters(&self) -> &Vec<Parameter>` where I think the declaration should be `fn parameters(&self) -> &[Parameter]`

cli/src/lib.rs
```diff
@@ -150,11 +142,12 @@ where
     /// - Reading telemetry configs
     /// - telemetry setup
     /// - Initialization of [`Sumeragi`]
+    #[allow(unsafe_code, clippy::unimplemented)]
     pub async fn with_genesis(
```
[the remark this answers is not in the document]
mversic: remove `unimplemented`

core/src/wsv.rs
```diff
@@ -173,13 +143,14 @@ impl<W: WorldTrait> WorldStateView<W> {
...
         for role_id in account.roles() {
-            if let Some(role) = self.world.roles.get(role_id) {
+            if let Some(role) = self.world.roles().get(role_id) {
                 tokens.append(&mut role.permissions().cloned().collect());
             }
         }
         tokens
     }
 
+    #[allow(unsafe_code, clippy::panic_in_result_fn, clippy::unimplemented)]
```
mversic, line 153: are these allows necessary?

cli/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: Since Iroha is Open Source, I think that the `ConsensusAPI` doesn't necessarily require a single `SumeragiTrait`, but rather having the flexibility to easily swap one implementation for another. As such, we could retain the trait, but avoid the usage of generics. 

This can change when specialisation is stable.

core/src/wsv.rs
appetrosyan: I'd prefer this be a separate `impl` block. This is idiomatically a slice, which should be treated as such.

core/src/wsv.rs
appetrosyan: I totally agree. Furthermore, since scope can be tightly controlled, the `getset(get = "pub")` directive can really be replaced with `pub(crate)` or `pub(super)` for the fields.

permissions_validators/src/public_blockchain/unregister.rs
```diff
@@ -102,18 +104,18 @@ impl<W: WorldTrait> IsGrantAllowed<W> for GrantRegisteredByMeAccess {
 /// Validator that checks Revoke instructions, such that the access is
 /// revoked and the assets of the signer's account are no longer
 /// accessible.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Copy, Clone, Serialize)]
```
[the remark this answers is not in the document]
mversic: yeah, I didn't want to mention it because we don't enforce it

cli/src/lib.rs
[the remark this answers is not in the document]
mversic: I discussed this with @Arjentix . I think we can provide builder methods to `Iroha` such as `Iroha::with_sumeragi`, etc. and use current implementation as the default if the builder method isn't called

core/src/wsv.rs
Arjentix: Never thought that it's a bad idea in Rust...

Arjentix: I understand that we loose to much things like moving, I'll remove it, this getters were used for mocking.
 Although without getters it feels for me like we have no encapsulation

core/src/wsv.rs
```diff
@@ -33,53 +28,54 @@ pub type NewBlockNotificationSender = tokio::sync::watch::Sender<()>;
...
 
-impl DerefMut for World {
-    #[inline]
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        self
+    /// Creates a [`World`] with these [`Domain`]s and trusted [`PeerId`]s.
+    pub fn with<D, P>(domains: D, trusted_peers_ids: P) -> Self
+    where
+        D: IntoIterator<Item = Domain> + 'static,
+        P: IntoIterator<Item = PeerId> + 'static,
```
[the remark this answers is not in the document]
Arjentix: Yeah, it was needed for mocking. Useless for now

cli/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > so this function isn't used at all anymore.

It is used in `Iroha::new()` and in `Peer::start()` (test_network).

> Moreover, does it allow the user to define a custom genesis block?

Yes, this is done in `test_network`, take a look at `PeerBuilder::start_with_peer()` which creates custom genesis and passes it to `Peer::start()` which calls `Iroha::with_genesis()`

core/src/wsv.rs
Arjentix: No, just forgot to delete it, thank you

cli/src/lib.rs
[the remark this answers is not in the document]
mversic: ok, for some reason, I thought we were getting rid of `test_network`. I have the feeling that something is not quite right with that method, but if that's not the case, resolve this conversation

core/src/smartcontracts/isi/account.rs
```diff
@@ -242,43 +234,43 @@ pub mod query {
...
         #[metrics(+"find_roles_by_account_id")]
-        fn execute(&self, wsv: &WorldStateView<W>) -> Result<Self::Output, Error> {
+        fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, Error> {
             let account_id = self
                 .id
                 .evaluate(wsv, &Context::new())
                 .wrap_err("Failed to evaluate account id")
                 .map_err(|e| Error::Evaluate(e.to_string()))?;
-            iroha_logger::trace!(%account_id, roles=?wsv.world.roles);
+            iroha_logger::trace!(%account_id, roles=?wsv.world.roles());
```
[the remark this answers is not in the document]
Arjentix: No, it was like that before I introduced getters. Now I'll remove getters back

cli/src/lib.rs
```diff
@@ -150,11 +142,12 @@ where
     /// - Reading telemetry configs
     /// - telemetry setup
     /// - Initialization of [`Sumeragi`]
+    #[allow(unsafe_code, clippy::unimplemented)]
```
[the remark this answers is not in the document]
Arjentix: Yeah, it's outdated

cli/src/torii/routing.rs
```diff
@@ -114,9 +116,10 @@ pub(crate) async fn handle_instructions<W: WorldTrait>(
 }
 
 #[iroha_futures::telemetry_future]
-pub(crate) async fn handle_queries<W: WorldTrait>(
-    wsv: Arc<WorldStateView<W>>,
-    query_validator: Arc<IsQueryAllowedBoxed<W>>,
+#[allow(unsafe_code, clippy::unimplemented)]
```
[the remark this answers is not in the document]
Arjentix: Was there because of runtime `W` checking. Now it's outdated, yes

telemetry/derive/tests/ui_fail/return_nothing.rs
```diff
@@ -1,11 +1,12 @@
+#[allow(unused_imports)]
```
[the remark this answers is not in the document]
Arjentix: My fault, it's not needed here

permissions_validators/src/public_blockchain/unregister.rs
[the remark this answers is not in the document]
Arjentix: Didn't know it. Will be precise next time

core/src/wsv.rs
appetrosyan: Encapsulation =/= getters/setters. 

In rust, the main point is gone because visibility modifiers, interior mutability and borrow checking won't allow you to break any invariants. The only use case that is not covered is dynamic dispatch in dynamic linking, which is only used for data_model entities.

core/test_network/tests/integration/sumeragi_with_mock.rs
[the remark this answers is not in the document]
appetrosyan: These tests aren't ignored, they are singled-out so we can run them separately from the others. 

With that said, these tests don't actually do anything, so stripping them out could work just as easily.

cli/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: That works too.

appetrosyan (approved):

mversic (approved):

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+        ))
+    }
+
+    #[test]
+    pub fn multiple_validators_combined() {
+        let permissions_validator: IsInstructionAllowedBoxed =
+            ValidatorBuilder::with_validator(DenyBurn)
+                .with_validator(DenyAlice)
+                .all_should_succeed()
+                .build();
```
mversic, line 207: isn't  `all_should_succeed` a terminating operation? Do we need explicit call to `build` following it?

Arjentix: I think it was you who suggested to introduce `build()`?

Issue #2276, opened 2022-05-27
Include current Block hash into `BlockHeaderValue`


appetrosyan:
Currently, when we query for block headers we cannot reliably compute the hash of the block, since it's computed from the byte representation of the block and can be representation dependent. 

To simplify our work, I suggest including the block hash in the query result

omkar-mohanty:
Can I take this issue?

appetrosyan:
Thank you! Your help would be much appreciated.

appetrosyan:
@0x009922 Would you like to weigh in on the discussion.

PR #2285, opened 2022-05-28
[feature] #2276: Include current Block hash into BlockHeaderValue

omkar-mohanty:
### Description of the Change
This PR introduces `current_block_hash` field to `BlockHeaderValue`. This field contains the hash of the latest `VersionComittedBlock`. 
### Issue
When we query for block headers we cannot reliably compute the hash of the block, since it's computed from the byte representation of the block and can be architecture dependent.
### Benefits
The Hash of a block will no longer be architecture dependent. 
### Possible Drawbacks
Not that I know of.

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -725,6 +725,8 @@ impl VersionedCommittedBlock {
 
     /// Converts block to [`iroha_data_model`] representation for use in e.g. queries.
     pub fn into_value(self) -> BlockValue {
+        let current_block_hash = self.hash();
+
         let CommittedBlock {
             header,
             rejected_transactions,
@@ -732,7 +734,6 @@ impl VersionedCommittedBlock {
             event_recommendations,
             ..
         } = self.into_v1();
-
         let BlockHeader {
             timestamp,
             height,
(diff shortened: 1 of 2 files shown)


core/src/block.rs
```diff
@@ -730,7 +730,7 @@ impl VersionedCommittedBlock {
             transactions,
             event_recommendations,
             ..
-        } = self.into_v1();
+        } = self.clone().into_v1();
```
appetrosyan, line 733: This clone is copying a lot of data. I would prefer if you cloned the only moved value instead.

appetrosyan (changes requested):

appetrosyan:
Please observe that the expression `enhancement` is not one of the approved titles for feature-implementations, and amend both the commit message and the title of the PR.

appetrosyan (approved):

appetrosyan (approved):

Arjentix (approved):

Arjentix (approved):

appetrosyan (approved):

PR #2291, opened 2022-05-30
[feature] #1638: `configuration` return doc subtree

### LINKED ISSUE #1638: Have `configuration` endpoint GET return a subtree of docs
```
Aleksandr Petrosyan, [25.11.21 13:31]
The docs on the configure endpoint are working. But they're not too intuitive.

Aleksandr Petrosyan, [25.11.21 13:31]
" Logger configuration.\n\nHas type LoggerConfiguration. Can be configured via environment variable IROHA_LOGGER"

Egor Ivkov, [25.11.21 13:32]
[In reply to Aleksandr Petrosyan]
Full path to the type can help

Aleksandr Petrosyan, [25.11.21 13:33]
Which should be returned by the top level field. If I'm querying the configuration docs, I don't know what subtypes are available,

Aleksandr Petrosyan, [25.11.21 13:34]
I think it should return the whole subtree of configurations, and not just the one that it matches

Egor Ivkov, [25.11.21 13:35]
Yes, I think it's a good idea
```

ilchu:

### Description of the Change

The `configuration` endpoint now returns all the inner fields as well when queried for docs.

### Issue
Resolves #1638.

### Benefits

More helpful doc API calls.

### Possible Drawbacks
The only way to return a `&'static str` while also making it custom formatted that I found was by boxing a `String` and then leaking it explicitly. Hopefully it doesn't introduce security issues.

### DIFF
--- a/config/derive/src/lib.rs
+++ b/config/derive/src/lib.rs
@@ -183,7 +183,7 @@ fn impl_get_doc_recursive(
         return quote! {
             fn get_doc_recursive<'a>(
                 inner_field: impl AsRef<[&'a str]>,
-            ) -> std::result::Result<std::option::Option<&'static str>, iroha_config::derive::Error>
+            ) -> std::result::Result<std::option::Option<String>, iroha_config::derive::Error>
             {
                 Err(iroha_config::derive::Error::UnknownField(
                     inner_field.as_ref().iter().map(ToString::to_string).collect()
@@ -199,11 +199,16 @@ fn impl_get_doc_recursive(
         .map(|(((ident, inner_thing), documentation), ty)| {
             if inner_thing {
                 quote! {
-                    [stringify!(#ident)] => Some(#documentation),
+                    [stringify!(#ident)] => {
+                        let curr_doc = #documentation;
+                        let inner_docs = <#ty as iroha_config::Configurable>::get_inner_docs();
+                        let total_docs = format!("{}\n\nHas following fields:\n\n{}\n", curr_doc, inner_docs);
+                        Some(total_docs)
+                    },
                     [stringify!(#ident), rest @ ..] => <#ty as iroha_config::Configurable>::get_doc_recursive(rest)?,
                 }
             } else {
-                quote! { [stringify!(#ident)] => Some(#documentation), }
+                quote! { [stringify!(#ident)] => Some(#documentation.to_owned()), }
             }
         })
         // XXX: Workaround
@@ -213,7 +218,7 @@ fn impl_get_doc_recursive(
     quote! {
         fn get_doc_recursive<'a>(
             inner_field: impl AsRef<[&'a str]>,
-        ) -> std::result::Result<std::option::Option<&'static str>, iroha_config::derive::Error>
+        ) -> std::result::Result<std::option::Option<String>, iroha_config::derive::Error>
         {
             let inner_field = inner_field.as_ref();
             let doc = match inner_field {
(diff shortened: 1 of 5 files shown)


appetrosyan (commented):
Once merged, should notify the rest of API changes. I would also consider updating the docs (api-spec.md).

appetrosyan:
The changes also affect Kagami, so the docs are generated with many extras. I think this change might make things singificantly simpler in that regard.

ilchu:
> The changes also affect Kagami, so the docs are generated with many extras. I think this change might make things singificantly simpler in that regard.
I noticed that it duplicates the inner fields when the doc test CI crashed, so instead thought that keeping the old structure of `docs/source/references/config.md` would make more sense, see modified `kagami/src/main.rs`.

Arjentix (changes requested):

config/derive/src/lib.rs
```diff
@@ -227,6 +233,46 @@ fn impl_get_doc_recursive(
...
+
+            quote! {
+                inner_docs.push_str(stringify!(#ident));
+                inner_docs.push_str(": ");
+                inner_docs.push_str(#doc);
+                inner_docs.push_str("\n\n");
+            }
+        })
+        // XXX: Workaround
+        //Decription of issue is here https://stackoverflow.com/a/65353489
```
outoftardis, line 262:
```suggestion
        //Description of issue is here https://stackoverflow.com/a/65353489
```

docs/source/references/api_spec.md
```diff
@@ -147,15 +147,17 @@ continues to stream blocks as they are added to the blockchain.
 **Expects**:
 There are 2 variants:
 - It either expects json body `"Value"` and returns configuration value as json
```
outoftardis, line 149:
```suggestion
- Expects: a JSON body `"Value"`. Returns: a configuration value as JSON.
```

docs/source/references/api_spec.md
```diff
@@ -147,15 +147,17 @@ continues to stream blocks as they are added to the blockchain.
 **Expects**:
 There are 2 variants:
```
outoftardis, line 148:
```suggestion
There are two options:
```

docs/source/references/api_spec.md
```diff
@@ -147,15 +147,17 @@ continues to stream blocks as they are added to the blockchain.
...
 - It either expects json body `"Value"` and returns configuration value as json
-- Or it expects json body like below and returns documentation for specific field (as json string) or null (here for field `a.b.c`):
+- Or it expects json body like below and returns documentation for specific field (as json string) or null. If the requested
```
outoftardis, line 150:
```suggestion
- Expects: a JSON body that specifies the field (see example below). Returns: documentation for a specific field (as JSON string) or `null`.

Note that if the requested
```

config/derive/src/lib.rs
```diff
@@ -199,11 +199,16 @@ fn impl_get_doc_recursive(
         .map(|(((ident, inner_thing), documentation), ty)| {
             if inner_thing {
                 quote! {
-                    [stringify!(#ident)] => Some(#documentation),
+                    [stringify!(#ident)] => {
+                        let curr_doc = #documentation;
+                        let inner_docs = <#ty as iroha_config::Configurable>::get_inner_docs();
+                        let total_docs = format!("{}\n\nHas following fields:\n\n{}\n", curr_doc, inner_docs);
```
Arjentix, line 205: Sorry, I just remembered. What if you try to use [`concat`](https://doc.rust-lang.org/std/macro.concat.html)? Then you can leave `&'static str`, I think

Arjentix (approved):

config/derive/src/lib.rs
appetrosyan: I don't mind either way. The only place affected is kagami and it had a `to_owned()` somewhere, so semantically what we want is closer to a `String`. 

At some point we should consider making them immutable strings that don't store the capacity, but that's a bit far ahead.

appetrosyan (approved):

PR #2304, opened 2022-06-01
[refactor] #2238: Add peer builder for tests

### LINKED ISSUE #2238: Introduce Builder pattern for `Peer` in test_network
Right now there are too many constructors with too many fields for `Peer` in `test_network/lib.rs`. It looks ugly. We can improve it with *Builder* pattern

pesterev:
### Description of the Change

This replaces nested methods which creates `Peer` with `PeerBuilder` structure (Builder pattern).

- [x] MVP
- [x] CI checks pass
- [x] Add the docs/comments

### Issue

Resolves #2238

### Benefits

Makes creating a test environment more flexible and convenient.

### Possible Drawbacks

Idk may increase test run time.

Make `Peer` structure a builder itself. But in that design, will need to keep parameters (`genesis`, `configuration`, `instruction_validator`, `query_validator`, `temp_dir`) inside the `Peer` structure to start the peer's task - it will makes `Peer` structure a huge.

### DIFF
--- a/client/benches/torii.rs
+++ b/client/benches/torii.rs
@@ -12,7 +12,7 @@ use iroha_core::{
 };
 use iroha_data_model::prelude::*;
 use iroha_version::Encode;
-use test_network::{get_key_pair, Peer as TestPeer, TestRuntime};
+use test_network::{get_key_pair, Peer as TestPeer, PeerBuilder, TestRuntime};
 use tokio::runtime::Runtime;
 
 const MINIMUM_SUCCESS_REQUEST_RATIO: f32 = 0.9;
@@ -37,12 +37,11 @@ fn query_requests(criterion: &mut Criterion) {
     )
     .expect("genesis creation failed");
 
-    rt.block_on(peer.start_with_config_permissions(
-        configuration.clone(),
-        genesis,
-        AllowAll,
-        AllowAll,
-    ));
+    let builder = PeerBuilder::new()
+        .with_configuration(configuration.clone())
+        .with_into_genesis(genesis);
+
+    rt.block_on(builder.start_with_peer(&mut peer));
     configuration
         .logger
         .max_log_level
@@ -136,7 +135,10 @@ fn instruction_submits(criterion: &mut Criterion) {
         &configuration.sumeragi.transaction_limits,
     )
     .expect("failed to create genesis");
-    rt.block_on(peer.start_with_config_permissions(configuration, genesis, AllowAll, AllowAll));
+    let builder = PeerBuilder::new()
+        .with_configuration(configuration)
+        .with_into_genesis(genesis);
+    rt.block_on(builder.start_with_peer(&mut peer));
     let mut group = criterion.benchmark_group("instruction-requests");
     let domain_id: DomainId = "domain".parse().expect("Valid");
     let create_domain = RegisterBox::new(Domain::new(domain_id.clone()));
--- a/client/tests/integration/queries/account.rs
+++ b/client/tests/integration/queries/account.rs
@@ -5,11 +5,11 @@ use std::{collections::HashSet, str::FromStr as _};
 use eyre::Result;
 use iroha_client::client;
 use iroha_data_model::prelude::*;
-use test_network::{Peer as TestPeer, *};
+use test_network::*;
 
 #[test]
 fn find_accounts_with_asset() -> Result<()> {
-    let (_rt, _peer, test_client) = <TestPeer>::start_test_with_runtime();
+    let (_rt, _peer, test_client) = <PeerBuilder>::new().start_with_runtime();
     wait_for_genesis_committed(&vec![test_client.clone()], 0);
 
     // Registering new asset definition
--- a/core/test_network/src/lib.rs
+++ b/core/test_network/src/lib.rs
@@ -52,52 +52,6 @@ pub struct Network<
     pub peers: HashMap<PeerId, Peer<W, G, K, S, B>>,
 }
 
-impl From<Peer> for Box<iroha_core::tx::Peer> {
-    fn from(val: Peer) -> Self {
-        Box::new(iroha_core::tx::Peer { id: val.id.clone() })
-    }
-}
-
-/// Peer structure
-pub struct Peer<
-    W = World,
-    G = GenesisNetwork,
-    K = Kura<W>,
-    S = Sumeragi<G, K, W>,
-    B = BlockSynchronizer<S, W>,
-> where
-    W: WorldTrait,
-    G: GenesisNetworkTrait,
-    K: KuraTrait<World = W>,
-    S: SumeragiTrait<GenesisNetwork = G, Kura = K, World = W>,
-    B: BlockSynchronizerTrait<Sumeragi = S, World = W>,
-{
-    /// id of peer
-    pub id: PeerId,
-    /// api address
-    pub api_address: String,
-    /// p2p address
-    pub p2p_address: String,
-    /// telemetry address
-    pub telemetry_address: String,
-    /// Key pair of peer
-    pub key_pair: KeyPair,
-    /// Broker
-    pub broker: Broker,
-    /// Shutdown handle
-    shutdown: Option<JoinHandle<()>>,
-    /// Iroha itself
-    pub iroha: Option<Iroha<W, G, K, S, B>>,
-}
-
-impl std::cmp::PartialEq for Peer {
-    fn eq(&self, other: &Self) -> bool {
-        self.id == other.id
-    }
-}
-
-impl std::cmp::Eq for Peer {}
-
 /// Get a standardised key-pair from the hard-coded literals.
 ///
 /// # Panics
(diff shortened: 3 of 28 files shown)


Arjentix (commented):
Pretty the same I was thinking about!

core/test_network/src/lib.rs
```diff
@@ -422,48 +439,48 @@ where
...
+        let query_validator = arguments.query_validator.unwrap_or_else(|| AllowAll.into());
+
+        let peer = self;
+        configuration
+            .sumeragi
+            .trusted_peers
+            .peers
+            .insert(peer.id.clone());
+
+        let temp_dir = Arc::new(TempDir::new().expect("Failed to create temp dir."));
```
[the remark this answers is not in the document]
pesterev: This is a mistake and one of the reasons why the tests were falling, thanks.

client/tests/integration/queries/account.rs
```diff
@@ -5,11 +5,11 @@ use std::{collections::HashSet, str::FromStr as _};
 use eyre::Result;
 use iroha_client::client;
 use iroha_data_model::prelude::*;
-use test_network::{Peer as TestPeer, *};
+use test_network::*;
 
 #[test]
 fn find_accounts_with_asset() -> Result<()> {
-    let (_rt, _peer, test_client) = <TestPeer>::start_test_with_runtime();
+    let (_rt, _peer, test_client) = <PeerBuilder>::new().start_with_runtime();
```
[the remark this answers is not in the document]
pesterev: Yes, otherwise, we still need to specify generics.

client/benches/torii.rs
```diff
@@ -35,14 +35,14 @@ fn query_requests(criterion: &mut Criterion) {
...
-        configuration.clone(),
-        genesis,
-        AllowAll,
-        AllowAll,
-    ));
+    let builder = PeerBuilder::new()
+        .with_configuration(configuration.clone())
+        .with_genesis(genesis);
+
+    rt.block_on(builder.start_with_peer(&mut peer));
```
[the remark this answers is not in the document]
pesterev: That's why I think this design is a bit ugly.

Lets think about an alternative design. So, we have several facts:
1. Instance of `Peer` can starts and also restarts.
2. `Peer` uses only values (which keep inside `PeerBuilder`) when calling the `start` method.
3. Instance `Peer` doesn't means "an already running peer", that is, instance of `Peer` can be pre-created to keep inside only the ID (and other parameters) and can be starts later.

But the alternative design is to introduce a `build` method for `PeerBuilder` and to pass the values of `PeerBuilder` to `Peer` when calling the `build`. So `Peer` will have to store all these values internally and will look like this:
```rust
struct Peer {
    /// id of peer
    pub id: PeerId,
    /// api address
    pub api_address: String,
    /// bla bla bla ....
   
    /// ... and also these parameters
    genesis: WithGenesis<G>,
    configuration: Option<Configuration>,
    instruction_validator: Option<IsInstructionAllowedBoxed<W>>,
    query_validator: Option<IsQueryAllowedBoxed<W>>,
    temp_dir: Option<Arc<TempDir>>,
}

/// ....

impl PeerBuilder {
    /// ...
    fn build(self) -> Peer {
        Peer::from(self.genesis, self.configuration, self.instruction_validator, // ...)
    }
}
```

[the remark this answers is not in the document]
pesterev: I've just realized that I can rename `PeerBuilder` to `PeerArgsBuilder` or something. And make the `Peer::start` method accept the `PeerArgsBuilder` as argument. It will look like this:
```rust
Peer::new().start(PeerArgsBuilder::new().with_configuration(config).build()).await;
```

This is the same idea but more pretty interface.

client/benches/torii.rs
```diff
@@ -135,8 +134,12 @@ fn instruction_submits(criterion: &mut Criterion) {
         &configuration.genesis,
         &configuration.sumeragi.transaction_limits,
     )
-    .expect("failed to create genesis");
-    rt.block_on(peer.start_with_config_permissions(configuration, genesis, AllowAll, AllowAll));
+    .expect("failed to create genesis")
```
outoftardis, line 137: I think this message and the one on line 38 should be the same

core/test_network/src/lib.rs
```diff
@@ -422,24 +441,28 @@ where
...
-    /// Use *Builder* pattern (#2238)
-    pub async fn start_with_config_permissions_dir(
-        &mut self,
-        configuration: Configuration,
-        genesis: Option<G>,
-        instruction_validator: impl Into<IsInstructionAllowedBoxed<W>> + Send + 'static,
-        query_validator: impl Into<IsQueryAllowedBoxed<W>> + Send + 'static,
-        temp_dir: Arc<TempDir>,
-    ) {
+    /// Starts peer with arguments.
```
outoftardis, line 444:
```suggestion
    /// Starts a peer with arguments.
```

core/test_network/src/lib.rs
```diff
@@ -553,43 +533,202 @@ where
...
+            temp_dir: None,
+        }
+    }
+}
+
+/// `WithGenesis` structure.
+///
+/// Options for setting up the genesis network for `PeerBuilder`.
+pub enum WithGenesis<G> {
+    /// The option means that it will create the default genesis network.
```
outoftardis, line 570:
```suggestion
    /// Create the default genesis network.
```

core/test_network/src/lib.rs
```diff
@@ -553,43 +533,202 @@ where
...
+    Default,
+    /// The option means pass nothing.
+    None,
+    /// The option means that sets the genesis network
```
outoftardis, line 574:
```suggestion
    /// Set the given genesis network.
```

core/test_network/src/lib.rs
```diff
@@ -553,43 +533,202 @@ where
...
+    }
+}
+
+/// `WithGenesis` structure.
+///
+/// Options for setting up the genesis network for `PeerBuilder`.
+pub enum WithGenesis<G> {
+    /// The option means that it will create the default genesis network.
+    Default,
+    /// The option means pass nothing.
```
outoftardis, line 572: not clear to me what the option means

```suggestion
    /// Pass nothing.
```

pesterev: This literally means that there is no genesis. So the Iroha peer will start without a given genesis.
I made this custom enum from `Option<Option<G>>`.

core/test_network/src/lib.rs
```diff
@@ -553,43 +531,202 @@ where
...
+/// `WithGenesis` structure.
+///
+/// Options for setting up the genesis network for `PeerBuilder`.
+pub enum WithGenesis<G> {
+    /// Create the default genesis network.
+    Default,
+    /// Pass nothing.
+    None,
+    /// Set the given genesis network.
+    Has(G),
```
outoftardis, line 573: ok maybe this is a better way to describe this

```suggestion
    /// Use the default genesis network.
    Default,
    /// Do not use any genesis networks.
    None,
    /// Use the given genesis network.
    Has(G),
```

core/test_network/src/lib.rs
```diff
@@ -422,24 +441,26 @@ where
...
-    pub async fn start_with_config_permissions_dir(
-        &mut self,
-        configuration: Configuration,
-        genesis: Option<G>,
-        instruction_validator: impl Into<IsInstructionAllowedBoxed<W>> + Send + 'static,
-        query_validator: impl Into<IsQueryAllowedBoxed<W>> + Send + 'static,
-        temp_dir: Arc<TempDir>,
-    ) {
+    /// Starts a peer with arguments.
+    async fn start(&mut self, arguments: PeerArgs<W, G>) {
```
Arjentix, line 445: What about to implement `Option` unwrapping not inside `Peer::start()` but inside `PeerBuilder`, as we discussed on the call? I think, that will be more logical

core/test_network/src/lib.rs
```diff
@@ -553,43 +531,202 @@ where
...
+    fn from(x: Option<G>) -> Self {
+        match x {
+            None => Self::None,
+            Some(genesis) => Self::Has(genesis),
+        }
+    }
+}
+
+/// `PeerBuilder` structure
+pub struct PeerBuilder<W = World, G = GenesisNetwork>(PeerArgs<W, G>)
```
Arjentix, line 592: This comment doesn't give us any new information

core/test_network/src/lib.rs
```diff
@@ -553,43 +531,202 @@ where
...
+where
+    W: WorldTrait,
+    G: GenesisNetworkTrait;
+
+impl<W, G> PeerBuilder<W, G>
+where
+    W: WorldTrait,
+    G: GenesisNetworkTrait,
+{
+    /// Creates `PeerBuilder`.
```
Arjentix, line 602:
```suggestion
    /// Creates [`PeerBuilder`].
```

Arjentix (commented):
When all tests will pass, I'll approve

core/test_network/src/lib.rs
pesterev: I forgot about it. Added.

outoftardis (approved):

Arjentix (approved):

PR #2305, opened 2022-06-01
[refactor] #1896: Simplify `produce_event` implementation

### LINKED ISSUE #1896: Investigate what happens when events sender is not configured
This [piece of code](https://github.com/hyperledger/iroha/blob/iroha2-dev/core/src/wsv.rs#L223) in `core/src/wsv.rs` should be investigated:
```rs
    fn produce_events(&self, events: impl IntoIterator<Item = DataEvent>) {
        let events = events.into_iter().map(Event::from);
        let events_sender = if let Some(sender) = &self.events_sender {
            sender
        } else {
            return warn!("wsv does not equip an events sender");
        };
        for event in events {
            drop(events_sender.send(event))
        }
    }
```

I can't say why this compiles, maybe it `panics` inside the macro. Probably an error should be returned or events just dropped

Erigara:

### Description of the Change

Issue #1896 occured because `produce_event` function has misleading statement: 

```rust
return warn!("wsv does not equip an events sender");
```
So to avoid confusion it was decided to refactor `produce_event`.

### Issue

Resolves #1896.

### Benefits

No confusion.

### Possible Drawbacks

None.

### Alternate Designs

Firstly alternative `produce_event` implementation was considered, but clippy rejects it.

Clippy error: ```error: use Option::map_or_else instead of an if let/else```

Alternative implementation: 

```rust
/// Send [`Event`]s to known subscribers.
fn produce_event(&self, event: impl Into<Event>) {
    if let Some(events_sender) = &self.events_sender {
        drop(events_sender.send(event.into()));
    } else {
        warn!("wsv does not equip an events sender");
    }
}
```

### DIFF
--- a/core/src/wsv.rs
+++ b/core/src/wsv.rs
@@ -307,13 +307,14 @@ impl<W: WorldTrait> WorldStateView<W> {
 
     /// Send [`Event`]s to known subscribers.
     fn produce_event(&self, event: impl Into<Event>) {
-        let events_sender = if let Some(sender) = &self.events_sender {
-            sender
-        } else {
-            return warn!("wsv does not equip an events sender");
-        };
-
-        drop(events_sender.send(event.into()))
+        self.events_sender.as_ref().map_or_else(
+            || {
+                warn!("wsv does not equip an events sender");
+            },
+            |events_sender| {
+                drop(events_sender.send(event.into()));
+            },
+        )
     }
 
     /// Tries to get asset or inserts new with `default_asset_value`.


pesterev (approved):
LGTM

core/src/wsv.rs
```diff
@@ -307,13 +307,14 @@ impl<W: WorldTrait> WorldStateView<W> {
...
     /// Send [`Event`]s to known subscribers.
     fn produce_event(&self, event: impl Into<Event>) {
-        let events_sender = if let Some(sender) = &self.events_sender {
-            sender
-        } else {
-            return warn!("wsv does not equip an events sender");
-        };
-
-        drop(events_sender.send(event.into()))
+        self.events_sender.as_ref().map_or_else(
```
appetrosyan, line 310: So how exactly does this resolve the issue?

mversic: There was no issue to begin with. The initial code was not understandable (to me) so I thought there must be some quirk with the tracing crate

s8sato: Personally, I prefer [early returns](https://doc.rust-lang.org/rust-by-example/error/result/early_returns.html) to avoid nesting

mversic: > Personally, I prefer [early returns](https://doc.rust-lang.org/rust-by-example/error/result/early_returns.html) to avoid nesting

I think we can all agree in favor of early returns

Erigara: > Personally, I prefer [early returns](https://doc.rust-lang.org/rust-by-example/error/result/early_returns.html) to avoid nesting

I agree that early returns are great to avoid nesting, but in this case we have the same level of nesting.
In my opinion, the problem with initial code was that [return](https://doc.rust-lang.org/std/keyword.return.html) is usually used to return some value from function, but in this case we combine it with `warn` macro that is not supposed to return any value.

Issue #2225, opened 2022-05-18
Fix benchmarks


appetrosyan:
## MWE

run 

```
cargo bench
```

## Expected

Works

## Actual 

Error messages in logs complaining about not being able to find `alice@1@wonderland`.

s8sato:
Disambiguation.
If this issue refers to WARN logs such as below on running `cargo bench --bench tps-dev`, then #2078 is responsible for it. This occurs when the `interval_us_per_tx` configuration is too short.

![image](https://user-images.githubusercontent.com/49983831/171624756-9d836e7d-f7fc-4ebf-adbd-0d951f469575.png)
![image](https://user-images.githubusercontent.com/49983831/171624773-5773425e-aa78-4f98-bac6-a1345b6a64e1.png)

Apart from that, I found another issue by accident. The `tps-dev` benchmark seems to have broken at some point since #1963. As of efa9a35692a949707ae0752119e4c4a1a08342a1, it fails with
```
thread 'main' panicked at 'The block is not yet in WSV. Need more sleep?', client/benches/tps/lib.rs:94:22
```
message, which would not be resolved by increasing [the sleep](https://github.com/hyperledger/iroha/blob/efa9a35692a949707ae0752119e4c4a1a08342a1/client/benches/tps/lib.rs#L80) (at least in 32 seconds, I've confirmed). This means the same number of blocks as the configuration are never applied to WSV. One possible cause is that the event counters are detecting more events than the actual number of blocks committed, and exiting prematurely

s8sato:
I think so too. Created #2310

SamHSmith:
I believe this is actually a consensus issue. It is probably the case that sumeragi currently can't handle high load due to some bug. This bug is however only triggered during the benchmark since it is the only time sumeragi is put under high tx load.

appetrosyan:
Related #2078

SamHSmith:
This issue no longer happens and so it can be closed.

PR #2306, opened 2022-06-01
[fix] #2225: Fix benchmarks

ales-tsurko:

### Description of the Change

I've changed the waiting time (`thread::sleep`) for blocks to appear in WSV. The value is twice bigger, so I expect it will work on machines twice slower.

### Issue

As described in the ticket, `cargo bench` panicked. It turned out the panic caused by too little time of waiting for blocks to appear in WSV.

### Benefits

It should fix the bug.

### Possible Drawbacks

The needed time may vary from machine to machine. As for now I don't know how everything works exactly, I can only guess that it's because of the time of blocks validation. So it may fail on slower machines, and then we need to increase the value again. Also, it slows down the benchmarks.

None.

We may provide an optional config for benchmarks through an environment variable, for example, with the sleep time.

### DIFF
--- a/client/benches/tps/lib.rs
+++ b/client/benches/tps/lib.rs
@@ -77,7 +77,7 @@ impl Config {
         }
         // END
         let elapsed_secs = timer.elapsed().as_secs_f64();
-        thread::sleep(core::time::Duration::from_secs(2));
+        thread::sleep(core::time::Duration::from_secs(4));
         let blocks_out_of_measure = 1 + 2 * self.peers;
         let mut blocks = network
             .genesis


s8sato (commented):
- I can't reproduce the error in issue #2225. Is the panic message you find the same as the original issue?

- Apart from that, it seems that the `tps-dev` benchmark has indeed broken at some point since #1963. As of efa9a35692a949707ae0752119e4c4a1a08342a1, no matter how long I [wait](https://github.com/hyperledger/iroha/blob/efa9a35692a949707ae0752119e4c4a1a08342a1/client/benches/tps/lib.rs#L80) (at least in 32 seconds, I've confirmed that it fails. It's abnormal) the same number of blocks as the configuration are never applied to WSV:
    ```
    thread 'main' panicked at 'The block is not yet in WSV. Need more sleep?', client/benches/tps/lib.rs:94:22
    ```
    One possible cause is that the event counters are detecting more events than the actual number of blocks committed, and exiting prematurely

client/benches/tps/lib.rs
```diff
@@ -77,7 +77,7 @@ impl Config {
         }
         // END
         let elapsed_secs = timer.elapsed().as_secs_f64();
-        thread::sleep(core::time::Duration::from_secs(2));
+        thread::sleep(core::time::Duration::from_secs(4));
```
s8sato, line 80:
```suggestion
        thread::sleep(core::time::Duration::from_secs(self.block_apply_lag_secs));
```
and add the corresponding to the configuration

ales-tsurko:
> * I can't reproduce the error in issue #2225. Is the panic message you find the same as the original issue?

Well, I have the warnings described in the issue, but they aren't the cause of the panic, which is actually the one you described below.

> * Apart from that, it seems that the `tps-dev` benchmark has indeed broken at some point since #1963. As of [efa9a35](https://github.com/hyperledger/iroha/commit/efa9a35692a949707ae0752119e4c4a1a08342a1), no matter how long I [wait](https://github.com/hyperledger/iroha/blob/efa9a35692a949707ae0752119e4c4a1a08342a1/client/benches/tps/lib.rs#L80) (at least in 32 seconds, I've confirmed that it fails. It's abnormal) the same number of blocks as the configuration are never applied to WSV:
>   ```
>   thread 'main' panicked at 'The block is not yet in WSV. Need more sleep?', client/benches/tps/lib.rs:94:22
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   One possible cause is that the event counters are detecting more events than the actual number of blocks committed, and exiting prematurely

Hmm Then changing this value won't fix anything. Will investigate further then. Thanks!

PR #2313, opened 2022-06-03
[fix] #2017: Fix role unregistration

### LINKED ISSUE #2017: Integration test to check if role is removed from accounts after unregistering
There was a bug, when role was removed only from domain and not from accounts. This is fixed in #1984. Now we need a test for it

Erigara:
Signed-off-by: Shanin Roman [shahin1000@yandex.ru](mailto:shanin1000@yandex.ru)

### Description of the Change

- Add `Role`-related match arm in `UnregisterBox` execution;
- Fix `Unregister<Role>::execute` by: 
    - removing double deletion of role from account;
    - returning correct event after revoking role from account;
    - correctly early exiting if role doesn't exist;
- Add `Role`-related queries to client;
- Add tests:
    - to check that role removed from account after unregistering;
    - to check `Role`-related queries. 

### Issue

Resolves #2017 

### Benefits

- New tests;
- Bug fixes.

### Possible Drawbacks

None.

### Usage Examples or Tests

```bash
cargo test --package iroha_client --test mod -- integration::roles::unregistred_role_removed_from_account --exact --nocapture
```

```bash
cargo test --package iroha_client --test mod -- integration::queries::role
```

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1136,12 +1136,12 @@ pub mod role {
         FindAllRoleIds::new()
     }
 
-    /// Get query to retrieve role by id
+    /// Get query to retrieve a role by its id
     pub fn by_id(role_id: impl Into<EvaluatesTo<RoleId>>) -> FindRoleByRoleId {
         FindRoleByRoleId::new(role_id)
     }
 
-    /// Get query to retrieve all roles for account
+    /// Get query to retrieve all roles for an account
     pub fn by_account_id(account_id: impl Into<EvaluatesTo<AccountId>>) -> FindRolesByAccountId {
         FindRolesByAccountId::new(account_id)
     }
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1122,6 +1122,31 @@ pub mod transaction {
     }
 }
 
+pub mod role {
+    //! Module with queries for roles
+    use super::*;
+
+    /// Get query to retrieve all roles
+    pub const fn all() -> FindAllRoles {
+        FindAllRoles::new()
+    }
+
+    /// Get query to retrieve all role ids
+    pub const fn all_ids() -> FindAllRoleIds {
+        FindAllRoleIds::new()
+    }
+
+    /// Get query to retrieve role by id
+    pub fn by_id(role_id: impl Into<EvaluatesTo<RoleId>>) -> FindRoleByRoleId {
+        FindRoleByRoleId::new(role_id)
+    }
+
+    /// Get query to retrieve all roles for account
+    pub fn by_account_id(account_id: impl Into<EvaluatesTo<AccountId>>) -> FindRolesByAccountId {
+        FindRolesByAccountId::new(account_id)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     #![allow(clippy::restriction)]
--- a/client/tests/integration/queries/role.rs
+++ b/client/tests/integration/queries/role.rs
@@ -1,0 +1,145 @@
+#![allow(clippy::restriction)]
+
+use std::collections::HashSet;
+
+use eyre::Result;
+use iroha_client::client;
+use iroha_core::smartcontracts::isi::query::Error as QueryError;
+use iroha_data_model::prelude::*;
+use test_network::{Peer as TestPeer, *};
+
+fn create_role_ids() -> [RoleId; 5] {
+    [
+        "a".parse().expect("Valid"),
+        "b".parse().expect("Valid"),
+        "c".parse().expect("Valid"),
+        "d".parse().expect("Valid"),
+        "e".parse().expect("Valid"),
+    ]
+}
+
+#[test]
+fn find_roles() -> Result<()> {
+    let (_rt, _peer, test_client) = <TestPeer>::start_test_with_runtime();
+    wait_for_genesis_committed(&vec![test_client.clone()], 0);
+
+    let role_ids = create_role_ids();
+
+    // Registering roles
+    let register_roles = role_ids
+        .iter()
+        .cloned()
+        .map(|role_id| RegisterBox::new(NewRole::new(role_id).build()).into())
+        .collect::<Vec<_>>();
+    test_client.submit_all_blocking(register_roles)?;
+
+    let role_ids = HashSet::from(role_ids);
+
+    // Checking results
+    let found_roles = test_client.request(client::role::all())?;
+    let found_role_ids = found_roles
+        .into_iter()
+        .map(|role| role.id().clone())
+        .collect::<HashSet<_>>();
+
+    assert_eq!(found_role_ids, role_ids);
+
+    Ok(())
+}
+
+#[test]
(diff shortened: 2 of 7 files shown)


client/tests/integration/roles.rs
```diff
@@ -141,3 +141,34 @@ fn register_metadata_role() -> Result<()> {
...
+
+    // Check that Alice has root role
+    let found_alice_roles = test_client.request(client::role::by_account_id(alice_id.clone()))?;
+    assert!(found_alice_roles.contains(&role_id));
+
+    // Unregister root role
+    let unregister_role = UnregisterBox::new(role_id.clone());
+    test_client.submit_blocking(unregister_role)?;
+
+    // Check that Alice hasn't root role
```
outoftardis, line 169:
```suggestion
    // Check that Alice doesn't have the root role
```

client/src/client.rs
```diff
@@ -1122,6 +1122,31 @@ pub mod transaction {
...
+    pub const fn all_ids() -> FindAllRoleIds {
+        FindAllRoleIds::new()
+    }
+
+    /// Get query to retrieve role by id
+    pub fn by_id(role_id: impl Into<EvaluatesTo<RoleId>>) -> FindRoleByRoleId {
+        FindRoleByRoleId::new(role_id)
+    }
+
+    /// Get query to retrieve all roles for account
```
outoftardis, line 1144:
```suggestion
    /// Get query to retrieve all roles for an account
```

client/src/client.rs
```diff
@@ -1122,6 +1122,31 @@ pub mod transaction {
...
+    pub const fn all() -> FindAllRoles {
+        FindAllRoles::new()
+    }
+
+    /// Get query to retrieve all role ids
+    pub const fn all_ids() -> FindAllRoleIds {
+        FindAllRoleIds::new()
+    }
+
+    /// Get query to retrieve role by id
```
outoftardis, line 1139:
```suggestion
    /// Get query to retrieve a role by its id
```

client/tests/integration/roles.rs
Erigara: fixed

client/src/client.rs
Erigara: fixed

client/src/client.rs
Erigara: fixed, thank you for your suggested changes.

client/tests/integration/roles.rs
```diff
@@ -141,3 +141,34 @@ fn register_metadata_role() -> Result<()> {
...
     Ok(())
 }
+
+#[test]
+fn unregistred_role_removed_from_account() -> Result<()> {
+    let (_rt, _peer, test_client) = <TestPeer>::start_test_with_runtime();
+    wait_for_genesis_committed(&vec![test_client.clone()], 0);
+
+    let role_id: RoleId = "root".parse().expect("Valid");
+    let alice_id: AccountId = "alice@wonderland".parse().expect("Valid");
```
Arjentix, line 151: I'm not very sure why, but we use another style for id. We use `<SomeThing as Idenfifiable>::Id>` type, not `SomeThingId`

Arjentix: Please, check for it also in other your changes

client/tests/integration/queries/role.rs
```diff
@@ -0,0 +1,145 @@
...
+        .collect::<Vec<_>>();
+    test_client.submit_all_blocking(register_roles)?;
+
+    let role_ids = HashSet::from(role_ids);
+
+    // Checking results
+    let found_roles = test_client.request(client::role::all())?;
+    let found_role_ids = found_roles
+        .into_iter()
+        .map(|role| role.id().clone())
```
Arjentix, line 42: Can it be moved? As I can see we don't need `found_roles` after that

Arjentix (approved):

core/src/smartcontracts/isi/world.rs
```diff
@@ -171,22 +171,16 @@ pub mod isi {
...
                 if world.roles.remove(&role_id).is_none() {
-                    return Ok(RoleEvent::Deleted(role_id).into());
+                    return Err(FindError::Role(role_id).into());
                 }
 
-                Err(FindError::Role(role_id).into())
+                Ok(RoleEvent::Deleted(role_id).into())
             })
         }
     }
```
s8sato, line 186: Might be refactored as follows:
```rust
    impl<W: WorldTrait> Execute<W> for Unregister<Role> {
        type Error = Error;

        #[metrics("unregister_role")]
        fn execute(
            self,
            _authority: <Account as Identifiable>::Id,
            wsv: &WorldStateView<W>,
            let role_id = self.object_id;

            wsv.modify_world(|world| {
                if world.roles.remove(&role_id).is_none() {
                    return Err(FindError::Role(role_id.clone()).into());
                }
                Ok(RoleEvent::Deleted(role_id.clone()).into())
            })?;

            for domain in wsv.domains() {
                for account in domain.accounts() {
                    if account.contains_role(&role_id) {
                        wsv.modify_account(account.id(), |account| {
                            account.remove_role(&role_id);
                            Ok(AccountEvent::RoleRevoked(account.id().clone()))
                        })?;
                    }
                }
            }

            Ok(())
        }
    }
```

s8sato (approved):

core/src/smartcontracts/isi/world.rs
s8sato: This suggestion is valid as far as the chain does not already have that inconsistency (phantom roles attached to accounts). So I think we have 2 ways:
- Adopt this suggestion as is and encourage people to resolve existing inconsistencies by revoking phantom roles for each account. This option promises to return early from invalid unregistration calls
- Reverse the 2 blocks in the suggestion -- world modification and accounts modification. This option promises to wipe out if any phantom roles brought in the future

Erigara: Also i believe that such implementation has some kind of lock, it may happen because we are trying to modify account in closure while outer scope has reference to it.
Maybe it explains why we have `accounts_with_role` in the first place.

Erigara: It's possible to rewrite account role removal part in a more functional style, but I don't know if it's worth it.

```rust
wsv.domains()
    .iter()
    .flat_map(|domain| {
        domain
            .accounts()
            .filter_map(|account| {
                if account.contains_role(&role_id) {
                    return Some(account.id().clone());
                }
                None
            })
            .collect::<Vec<_>>()
    })
    .try_for_each(|account_id| {
        wsv.modify_account(&account_id.clone(), |account| {
            if !account.remove_role(&role_id) {
                error!(%role_id, "role not found - this is a bug");
            }

            Ok(AccountEvent::RoleRevoked(account_id))
        })
    })?;
```

s8sato: I don't think we need to. You can resolve this conversation

Issue #2252, opened 2022-05-23
FFI error handling


mversic:
implement better error handling over FFI as is described [here](https://michael-f-bryan.github.io/rust-ffi-guide/errors/return_types.html). Currently, only error codes are transferred over the FFI boundary

mversic:
resuming panics should also be considered: https://doc.rust-lang.org/std/panic/fn.resume_unwind.html

ales-tsurko:
Moved back to **to do** until #2314 is merged, because it heavily affects this task.

Arjentix:
@ales-tsurko , #2314 was merged, do you plan to work on this?

ales-tsurko:
> @ales-tsurko , #2314 was merged, do you plan to work on this?

Hi! I'm on vacation until August, 29. So no at least until this date

mversic:
When an `extern` function fails an error code `FfiReturn` is returned. The user can then query `iroha_ffi` to get more details on the error via another `extern` function named `take_error`/`take_last_error`/`last_error` or similar. IMO this should best be done via `ffi_export` annotation to expose the function for retrieving errors as follows:

```rs
#[ffi_export]
fn take_last_error() -> Option<Box<Error>> {
    // ...
}
```

there should never be any need to write code for FFI serialization/deserialization.

appetrosyan:
It should be made clear that this approach is 
1. Not exactly standard Rust. 
2. Not particularly ergonomic, since the user has to remember to check for the error, it would require considerably code being added to the FFI user which defeats the purpose of dynamic linkage.

appetrosyan:
It should be noted that the user of the FFI is only expected to handle a limited number of errors, the rest can **just** be logged for debugging purposes. 

I see no reason why the few errors that can be handled couldn't be passed like any other FFI struct, and in some cases just converted to a string with the `impl Debug`.

mversic:
>     1. Not exactly standard Rust.
>     2. Not particularly ergonomic, since the user has to remember to check for the error, it would require considerably code being added to the FFI user which defeats the purpose of dynamic linkage.

the idea was to autogenerate the error handling code on the smart contract side and hide the complexity behind a know API so the user wouldn't bear the burden. To the user this would all look like a standard Rust code. However, in the background, the shim code would incur another FFI call for **every** error returned which IMHO shouldn't be noticeable unless we were to fetishize error handling mechanism as a control flow mechanism, which we don't.

[...]

PR #2314, opened 2022-06-03
[feature] #2434: FFI bindgen library

### LINKED ISSUE #2327: Add catch unwind to prevent UB over FFI calls

mversic:

293320603_1100305063913355_1700096561062333955_n

### Description of the Change

I've found that often times I need to traverse the type and produce conversions from source to FFI compliant type and vice versa.
There have been a lot of code duplication and convoluted logic which this PR aims to resolve

Traits which enable this happen:  `ReprC`, `AsReprC`, `TryFromReprC` `IntoFfi`. A `Rust` structure is converted into an intermediary FFI struct via `IntoFfi` which implements the `AsReprC` and can be converted into a type that can cross FFI boundary. This goes the other way as well - FFI `ReprC` type is converted into a struct via `TryFromReprC`

* 1st rule of this lib - there is no ownership transfer except for opaque pointer types
* Added `catch_unwind` to prevent UB when panicking over FFI

### Issue

Closes #2327


ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,509 @@
...
+                }
+            }
+            _ => {
+                if self.is_input {
+                    self.ffi_type = Some(parse_quote! { *const #node });
+                    self.src_to_ffi = Some(parse_quote! {
+                        let #arg_name: *const _ = #arg_name;
+                    });
+                    self.ffi_to_src = Some(parse_quote! {
+                        let #arg_name = Clone::clone(&*#arg_name);
```
mversic, line 351: at the moment I'm doing an in depth clone of the given argument. This may be substituted with ownership transfer instead

ffi/derive/src/bindgen.rs
```diff
@@ -175,63 +221,53 @@ fn gen_output_assignment_stmts(fn_descriptor: &FnDescriptor) -> Vec<syn::Stmt> {
...
 
             stmts.push(parse_quote! {{
-                let #output_arg_name = core::slice::from_raw_parts_mut(#output_arg_name, #slice_len_arg_name);
-
-                #slice_elems_arg_name.write(method_res.len());
-                for (i, elem) in method_res.take(#slice_len_arg_name).enumerate() {
-                    #output_arg_name[i] = elem;
+                #slice_elems_arg_name.write(#output_arg_name.len());
+                for (i, elem) in #output_arg_name.take(#slice_len_arg_name).enumerate() {
+                    let offset = i.try_into().expect("allocation too large");
```
mversic, line 235: I could have used `i as isize` here. Thoughts?

mversic, line 235: I could have used:
```suggestion
                    let offset = i as isize;
```

relevant [text](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset)

ffi/derive/src/derive.rs
```diff
@@ -203,22 +175,48 @@ fn gen_ffi_fn_body(method_name: &Ident, field_ty: &Type, derive: Derive) -> Toke
...
-    let ffi_fn_body = gen_ffi_fn_body(&derive_method_name, &field_ty, derive);
+    let ffi_fn_args = gen_ffi_fn_args(&handle, &arg, derive);
+    let ffi_fn_body = gen_ffi_fn_body(&derive_method_name, &handle, &arg, derive);
 
     parse_quote! {
         #[doc = #ffi_fn_doc]
         #[no_mangle]
         pub unsafe extern "C" fn #ffi_fn_name(#ffi_fn_args) -> iroha_ffi::FfiResult {
-            #ffi_fn_body
+            let res = std::panic::catch_unwind(|| {
```
mversic, line 209: I'm not sure where a panic might hit, so I thought it's better to be safe then sorry

ffi/derive/tests/ui_fail/derive_skip_field.rs
```diff
@@ -15,16 +15,16 @@ pub struct FfiStruct {
 fn main() {
     let s: *mut _ = &mut FfiStruct { a: 42, b: 32 };
 
-    let a = MaybeUninit::<*const u32>::uninit();
-    let b = MaybeUninit::<*const u32>::uninit();
+    let mut a = MaybeUninit::<*const u32>::uninit();
+    let mut b = MaybeUninit::<*const u32>::uninit();
```
Arjentix, line 19: Why does it need to be mutable now? Because setters now writes result to them or what?

ffi/derive/tests/ui_fail/mutable_input_arg.rs
```diff
@@ -1,30 +0,0 @@
-//! Test which assures that methods taking mutable references as input arguments are not
```
Arjentix, line 1: So now we accepts mutable input args? What it is done for and why it wasn't like that before?

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,522 @@
...
+    }
+    fn visit_type_impl_trait(&mut self, node: &'ast syn::TypeImplTrait) {
+        if node.bounds.len() > 1 {
+            abort!(
+                node.bounds,
+                "Only one trait is allowed for the `impl trait` argument"
+            );
+        }
+
+        if let syn::TypeParamBound::Trait(trait_) = &node.bounds[0] {
```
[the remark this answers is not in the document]
mversic: it's the convention used in `syn` crate because you cannot use `trait` as a variable name since it's a keyword

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,522 @@
...
+                    self.src_to_ffi = Some(parse_quote! {let #arg_name = #arg_name as u8;});
+                    self.ffi_to_src = Some(parse_quote! {let #arg_name = #arg_name != 0;});
+                } else {
+                    self.ffi_type = Some(node.clone().into());
+                    self.src_to_ffi = Some(parse_quote! {();});
+                    self.ffi_to_src = Some(parse_quote! {();});
+                }
+            }
+            "u8" | "u16" => {
+                self.ffi_type = Some(parse_quote! {u32});
```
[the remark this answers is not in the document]
mversic: because wasm doesn't supports only `u32`/`i32` and `f32`/`f64`. In the future we can introduce a flag to distinguish between wasm FFI and the usually supported FFI types

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,522 @@
...
+                self.ffi_to_src = Some(parse_quote! {();});
+            }
+            "Option" => self.visit_type_option(generic_arg_types(last_seg)[0]),
+            "Result" => {
+                if self.is_input {
+                    abort!(node, "Type not supported as input type")
+                }
+
+                let args = generic_arg_types(last_seg);
+                let (ok_type, _) = (args[0], args[1]);
```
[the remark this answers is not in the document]
mversic: this is kinda on purpose. In the future when error handling is better implemented in #2252 the error variant will have to be extracted

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,522 @@
...
+                            #ok_src_to_ffi
+                            #arg_name
+                        },
+                        Err(_) => {
+                            // TODO: Implement error handling (https://github.com/hyperledger/iroha/issues/2252)
+                            return iroha_ffi::FfiResult::ExecutionFail;
+                        }
+                    };
+                });
+                self.ffi_to_src = Some(parse_quote! { #ok_ffi_to_src });
```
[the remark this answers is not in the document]
mversic: yeah, Ok is missing. Null is not allowed for `Result`. It should be checked for nullability beforehand as is done for other types

ffi/derive/tests/ui_fail/mutable_input_arg.rs
mversic: yeah, it's not used anywhere except for input mutable handles. Previously, it was a nuisance to support this

ffi/derive/src/bindgen.rs
```diff
@@ -126,43 +149,54 @@ fn gen_ffi_to_src_conversion_stmts(fn_descriptor: &FnDescriptor) -> Vec<syn::Stm
...
 
-        vec![arg.ffi_name.clone()]
+        vec![arg.name.clone()]
     });
 
-    let fn_arg_names = fn_descriptor.input_args.iter().map(|arg| &arg.ffi_name);
-    parse_quote! { let method_res = #self_type::#method_name(#(#self_arg_name,)* #(#fn_arg_names),*); }
+    let fn_arg_names = fn_descriptor.input_args.iter().map(|arg| &arg.name);
+    let method_call = quote! {
+    #self_type::#method_name(#(#self_arg_name,)* #(#fn_arg_names),*)};
```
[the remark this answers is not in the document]
mversic: don't understand

ffi/derive/src/arg.rs
mversic: yeah, because ownership transfer is a bit difficult to do correct

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub name: Ident,
+
+    /// Rust type
+    pub src_type: Type,
+    /// FFI compliant type
+    pub ffi_type: Type,
+
+    /// Conversion statement from rust to FFI compliant type
+    pub src_to_ffi: syn::Stmt,
+    /// Conversion statement from FFI compliant to rust type
```
outoftardis, line 27:
```suggestion
    /// Conversion statement from Rust to FFI compliant type
    pub src_to_ffi: syn::Stmt,
    /// Conversion statement from FFI compliant to Rust type
```

ffi/derive/src/arg.rs
mversic: opened #2325

ffi/derive/src/bindgen.rs
mversic: it could also be converted into an explicit error, instead of panicking

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,522 @@
...
+                    });
+                    self.ffi_to_src = Some(parse_quote! {
+                        let #arg_name = if !#arg_name.is_null() {
+                            Some(&*#arg_name)
+                        } else {
+                            None
+                        };
+                    });
+                }
+            }
```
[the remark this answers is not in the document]
mversic: doesn't have to be a macro, I can extract differences into tokens and stick them into one generalized expression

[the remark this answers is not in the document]
mversic: tnx, I deduplicated this code

ffi/derive/src/derive.rs
appetrosyan: Ideally we should have proper error handling and no panics.  #2252 is relevant.

ffi/derive/src/arg.rs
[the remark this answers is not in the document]
appetrosyan: Rust allows you to create what's effectively `trait` as the variable name using `r#trait`, but I prefer the syn_ convention.

ffi/derive/src/arg.rs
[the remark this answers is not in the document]
appetrosyan: I'd add a comment.

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,521 @@
...
+    fn visit_type_path(&mut self, node: &'ast syn::TypePath) {
+        let last_seg = node.path.segments.last().expect_or_abort("Defined");
+
+        let arg_name = self.name;
+        match last_seg.ident.to_string().as_str() {
+            "bool" => {
+                if self.is_input {
+                    self.ffi_type = Some(parse_quote! { u8 });
+                    self.src_to_ffi = Some(parse_quote! {let #arg_name = #arg_name as u8;});
+                    self.ffi_to_src = Some(parse_quote! {let #arg_name = #arg_name != 0;});
```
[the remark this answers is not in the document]
appetrosyan: Yes. Memory corruption and bit flips are not unheard of.

ffi/derive/tests/ui_fail/mutable_input_arg.rs
appetrosyan: While being more general is good IMO, since we're not using it, you could also leave it as is.

data_model/src/domain.rs
```diff
@@ -163,8 +163,12 @@ pub struct Domain {
     #[getset(skip)]
     asset_definitions: AssetDefinitionsMap,
     /// IPFS link to the `Domain` logo
+    // FIXME: Getter implemented manually because `getset`
```
appetrosyan, line 166: Relevant issue https://github.com/Hoverbear/getset/issues/78

ffi/derive/src/arg.rs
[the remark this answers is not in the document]
mversic: It's a question of API design. Nothing catastrophic will happen if bits are flipped or memory corrupted. You can consider this a serialization format where all values in u8 are representable as bool. So, in a sense, this expands the Rust bool to accept 256 values instead of only 2 as is the case with C

ffi/derive/tests/ui_fail/mutable_input_arg.rs
mversic: I wanted to leave it as is, but it ended up being complicated because now all types go through this conversion and handle type can be received as `&mut Self`. So API got expanded for free

ffi/derive/src/derive.rs
mversic: yeah, this is a form of precaution and I wouldn't remove it. It's hard to guarantee that no downstream library will ever panic

ffi/derive/src/bindgen.rs
appetrosyan: This will fail silently, that wualifies as undefined behaviour (technically just a truncation error). However the truncation error only occurs if somewhere else we also have UB. To me this cast seems reasonable, but I would link to the text inside the code.

mversic: added the link in the code. I've left it as explicit conversion, not a cast. Maybe the compiler would be smart enough to optimize it out on some platforms

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,494 @@
...
+impl<'ast> SelfResolver<'ast> {
+    pub fn new(self_ty: &'ast syn::Path) -> Self {
+        Self { self_ty }
+    }
+}
+
+impl VisitMut for SelfResolver<'_> {
+    fn visit_path_mut(&mut self, node: &mut syn::Path) {
+        if node.leading_colon.is_some() {
+            // NOTE: It's irrelevant
```
QuentinI, line 478: What's the purpose of this `if`? I don't get it from the comment.

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,494 @@
...
+                        return true;
+                    }
+
+                    return mutability && trait_name == "ExactSizeIterator";
+                }
+            }
+            _ => return false,
+        }
+
+        false
```
QuentinI, line 88: Could make result of match itself be the return value.

QuentinI (dismissed):
LGTM, but I'm not well-versed in FFI and proc macros, so I'm mostly nitpicking here anyway 😅

ffi/derive/src/arg.rs
mversic: I could remove it, it's there as a reminder that visitor could parse the item at hand such as `leading_colon` but it's irrelevant in the current implementation

ffi/derive/src/arg.rs
mversic: Some ifs don't have an else, so the match statement doesn't always return a value

appetrosyan (approved):

ffi/tests/ffi_bindgen.rs
```diff
@@ -62,6 +62,16 @@ impl FfiStruct {
         self.params.iter()
     }
 
+    /// Tokens
+    pub fn tokens(&self) -> &[Value] {
+        &self.tokens
+    }
+
+    /// Tokens mut
```
outoftardis, line 70:
```suggestion
    /// Mutable tokens
```
?

ffi/derive/src/lib.rs
```diff
@@ -47,6 +49,398 @@ pub fn ffi_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    }
+
+    if !input.generics.params.is_empty() {
+        abort!(input.generics, "Generics not supported");
+    }
+
+    match input.data {
+        syn::Data::Struct(_) => derive_into_ffi_for_struct(&input.ident, &input.attrs),
+        syn::Data::Enum(item) => derive_into_ffi_for_enum(&input.ident, &item, &input.attrs),
+        syn::Data::Union(item) => abort!(item.union_token, "Unions not supported"),
```
outoftardis, line 68:
```suggestion
        syn::Data::Union(item) => abort!(item.union_token, "Unions are not supported"),
```

ffi/derive/src/lib.rs
```diff
@@ -47,6 +49,398 @@ pub fn ffi_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[proc_macro_error::proc_macro_error]
+pub fn try_from_ffi_derive(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as syn::DeriveInput);
+
+    if !matches!(input.vis, syn::Visibility::Public(_)) {
+        abort!(input.vis, "Only public items supported");
+    }
+
+    if !input.generics.params.is_empty() {
+        abort!(input.generics, "Generics not supported");
```
outoftardis, line 83:
```suggestion
        abort!(input.vis, "Only public items are supported");
    }

    if !input.generics.params.is_empty() {
        abort!(input.generics, "Generics are not supported");
```

ffi/derive/src/lib.rs
```diff
@@ -47,6 +49,398 @@ pub fn ffi_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    }
+
+    match input.data {
+        syn::Data::Struct(_) => derive_try_from_ffi_for_struct(&input.ident, &input.attrs),
+        syn::Data::Enum(item) => derive_try_from_ffi_for_enum(&input.ident, &item, &input.attrs),
+        syn::Data::Union(item) => abort!(item.union_token, "Unions not supported"),
```
outoftardis, line 89:
```suggestion
        syn::Data::Union(item) => abort!(item.union_token, "Unions are not supported"),
```

ffi/derive/src/lib.rs
```diff
@@ -47,6 +49,398 @@ pub fn ffi_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[proc_macro_error::proc_macro_error]
+pub fn into_ffi_derive(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as syn::DeriveInput);
+
+    if !matches!(input.vis, syn::Visibility::Public(_)) {
+        abort!(input.vis, "Only public items supported");
+    }
+
+    if !input.generics.params.is_empty() {
+        abort!(input.generics, "Generics not supported");
```
outoftardis, line 62:
```suggestion
        abort!(input.vis, "Only public items are supported");
    }

    if !input.generics.params.is_empty() {
        abort!(input.generics, "Generics are not supported");
```

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,513 @@
...
+                    });
+                    self.ffi_to_src = Some(parse_quote! {
+                        let #arg_name = *Box::from_raw(#arg_name);
+                    });
+                }
+            }
+        }
+    }
+    fn visit_type_ptr(&mut self, node: &'ast syn::TypePtr) {
+        abort!(node, "Raw pointers not supported")
```
outoftardis, line 360:
```suggestion
        abort!(node, "Raw pointers are not supported")
```

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    fn visit_type_trait_object(&mut self, _: &'ast syn::TypeTraitObject) {
+        unimplemented!("Not needed as of yet")
+    }
+    fn visit_type_tuple(&mut self, node: &'ast syn::TypeTuple) {
+        let arg_name = self.name;
+
+        if node.elems.len() != 2 {
+            abort!(node, "Only tuple pairs supported as of yet");
```
outoftardis, line 398:
```suggestion
            abort!(node, "Only tuple pairs are supported as of yet");
```

ffi/derive/src/arg.rs
```diff
@@ -0,0 +1,513 @@
...
+            }
+        }
+
+        return bindings;
+    };
+
+    abort!(seg, "Binding not found in the given path segment")
+}
+
+/// Visitor for path types which replaces all occurrences of `Self` with a fully qualified type
```
outoftardis, line 479:
```suggestion
/// Visitor replaces all occurrences of `Self` in a path type with a fully qualified type
```

crypto/src/merkle.rs
```diff
@@ -125,16 +125,15 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
     fn schema(map: &mut MetaMap) {
+        HashOf::<T>::schema(map);
+
+        // NOTE: Leaf nodes in the order of insertion
         map.entry(Self::type_name()).or_insert_with(|| {
-            // BFS ordered list of leaf nodes
             Metadata::Vec(VecMeta {
                 ty: HashOf::<T>::type_name(),
-                sorted: true,
+                sorted: false,
```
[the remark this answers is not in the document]
mversic: this was a bug previously. This should not affect SDKs because the `sorted` requirement is relaxed, not strengthened

data_model/src/block_value.rs
```diff
@@ -38,18 +37,6 @@ pub struct BlockHeaderValue {
     pub current_block_hash: Hash,
 }
 
-impl PartialOrd for BlockHeaderValue {
```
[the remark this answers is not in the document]
mversic: yes, we do now

ffi/src/primitives.rs
```diff
@@ -0,0 +1,217 @@
+use crate::{
+    slice::{AsFfiSlice, SliceMut, SliceRef},
+    AsFfi, FfiRef, FfiResult, FfiType, FfiWriter, FromOption, IntoFfi, OptionWrapped, ReprC,
+    TryAsRust, TryFromFfi,
+};
+
+impl FfiType for bool {
+    type FfiType = u8;
```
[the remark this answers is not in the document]
mversic: `bool` size is [guaranteed](https://doc.rust-lang.org/reference/type-layout.html#primitive-data-layout) to be 1byte

ffi/src/lib.rs
```diff
@@ -1,23 +1,136 @@
...
+    type FfiType: ReprC;
+}
+
+pub struct Boxed<T>(Box<T>);
+
+pub struct IntoWrapper<T: Into<U>, U>(T, PhantomData<U>);
 
 /// Result of execution of an FFI function
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 // NOTE: Enum is `repr(i32)` becasuse WebAssembly supports only
```
outoftardis, line 128:
```suggestion
// NOTE: Enum is `repr(i32)` because WebAssembly supports only
```

ffi/src/lib.rs
```diff
@@ -1,23 +1,136 @@
...
+pub trait TryFromFfi: Sized {
+    type Item: FromReprC;
+
+    /// Performs the fallible conversion
+    ///
+    /// # Errors
+    ///
+    /// * given pointer is null
+    /// * given id doesn't identify any known handle
+    /// * given id is not a valid enum discriminant
```
outoftardis, line 94: I'd add the corresponding FfiResult enum variant for each error to create links in docs

ffi/src/lib.rs
```diff
@@ -1,23 +1,190 @@
...
-pub struct Pair<K, V>(pub K, pub V);
+pub struct IntoWrapper<T: Into<U>, U>(T, PhantomData<U>);
 
 /// Result of execution of an FFI function
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 // NOTE: Enum is `repr(i32)` becasuse WebAssembly supports only
 // u32/i32, u64/i64 natively. Otherwise, `repr(i8)` would suffice
 #[repr(i32)]
 pub enum FfiResult {
+    /// Indicates that the FFI function execution panicked
```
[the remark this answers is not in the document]
outoftardis: Here's a suggestion on how to reword the errors:

```
    /// FFI function execution panicked.
    UnrecoverableError = -5_i32,
    /// Provided handle id doesn't match any known handles.
    UnknownHandle = -4_i32,
    /// FFI function failed during the execution of the wrapped method on the provided handle.
    ExecutionFail = -3_i32,
    /// The input argument provided to FFI function is a null raw pointer.
    /// The pointer must not be null.
    ArgIsNull = -2_i32,
    /// The input argument provided to FFI function is not a valid UTF-8 string.
    /// Cannot derive the entity name from it.
    Utf8Error = -1_i32,
    /// FFI function executed successfully.
    Ok = 0_i32,
```

Though I'm not sure how to expand the `UnrecoverableError`

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+/// Conversion into an FFI compatible representation that consumes the input value
+pub trait IntoFfi: Sized {
+    /// Type which can be converted into a [`ReprC`] type and sent over FFI
+    type Item: AsReprC;
+
+    /// Performs the conversion from [`Self`] into [`Self::Item`]
+    fn into_ffi(self) -> Self::Item;
+}
+
+/// Conversion from an FFI compatible representation that consumes the input value
```
outoftardis, line 73:
```suggestion
/// Conversion from an FFI-compatible representation that consumes the input value
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+    /// Transferring ownership over FFI is not permitted, except for [`Opaque`] types
+    fn from_repr_c(source: Self::Item) -> Self;
+}
+
+/// Conversion into an FFI compatible representation that consumes the input value
+pub trait IntoFfi: Sized {
+    /// Type which can be converted into a [`ReprC`] type and sent over FFI
+    type Item: AsReprC;
+
+    /// Performs the conversion from [`Self`] into [`Self::Item`]
```
outoftardis, line 69:
```suggestion
    /// Convert from [`Self`] into [`Self::Item`]
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+///
+/// If two structures implement the same id, this may result in void pointer being casted to the wrong type
+pub unsafe trait Handle {
+    /// Unique identifier of the handle. Most commonly, it is
+    /// used to facilitate generic monomorphization over FFI
+    const ID: handle::Id;
+}
+
+/// Robust type which conforms to C ABI and can be safely shared across FFI boundaries. This does
+/// not ensure the ABI of the referent for pointers in which case the pointer should be made opaque
```
outoftardis, line 37:
```suggestion
/// Robust type that conforms to C ABI and can be safely shared across FFI boundaries. This does
/// not ensure the ABI compatibility of the referent for pointers, but making the pointers opaque does.
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+pub trait FromReprC {
+    /// Robust C ABI compliant representation of [`Self`]
+    type Item: ReprC;
+
+    /// Performs the converstion from [`Self::Item`] into [`Self`].
+    /// Transferring ownership over FFI is not permitted, except for [`Opaque`] types
+    fn from_repr_c(source: Self::Item) -> Self;
+}
+
+/// Conversion into an FFI compatible representation that consumes the input value
```
outoftardis, line 64:
```suggestion
/// Conversion into an FFI-compatible representation that consumes the input value
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+    /// used to facilitate generic monomorphization over FFI
+    const ID: handle::Id;
+}
+
+/// Robust type which conforms to C ABI and can be safely shared across FFI boundaries. This does
+/// not ensure the ABI of the referent for pointers in which case the pointer should be made opaque
+///
+/// # Safety
+///
+/// Type implementing the trait must be a robust type with a guaranteed C ABI
```
outoftardis, line 41:
```suggestion
/// The type implementing the trait must be a robust type with a guaranteed C ABI
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+pub unsafe trait ReprC {}
+
+/// Used to do a conversion into a type which implements [`ReprC`].
```
outoftardis, line 44:
```suggestion
/// Conversion into a type that implements [`ReprC`].
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+pub trait AsReprC {
+    /// Robust C ABI compliant representation of [`Self`]
+    type Item: ReprC;
+
+    /// Performs the converstion from [`Self`] into [`Self::Item`].
+    /// Transferring ownership over FFI is not permitted, except for [`Opaque`] types
```
outoftardis, line 50:
```suggestion
    /// Convert from [`Self`] into [`Self::Item`].
    /// Transferring ownership over FFI is not permitted, except for [`Opaque`] types.
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+    fn as_repr_c(&mut self) -> Self::Item;
+}
+
+/// Used to do a conversion from a type which implements [`ReprC`].
```
outoftardis, line 54:
```suggestion
/// Conversion from a type that implements [`ReprC`].
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+pub trait FromReprC {
+    /// Robust C ABI compliant representation of [`Self`]
+    type Item: ReprC;
+
+    /// Performs the converstion from [`Self::Item`] into [`Self`].
+    /// Transferring ownership over FFI is not permitted, except for [`Opaque`] types
```
outoftardis, line 60:
```suggestion
    /// Convert from [`Self::Item`] into [`Self`].
    /// Transferring ownership over FFI is not permitted, except for [`Opaque`] types.
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+    fn from_repr_c(source: Self::Item) -> Self;
+}
+
+/// Conversion into an FFI compatible representation that consumes the input value
+pub trait IntoFfi: Sized {
+    /// Type which can be converted into a [`ReprC`] type and sent over FFI
```
outoftardis, line 66:
```suggestion
    /// Type that can be converted into a [`ReprC`] type and sent over FFI
```

ffi/src/lib.rs
```diff
@@ -1,173 +1,360 @@
...
+    type Item: AsReprC;
+
+    /// Performs the conversion from [`Self`] into [`Self::Item`]
+    fn into_ffi(self) -> Self::Item;
+}
+
+/// Conversion from an FFI compatible representation that consumes the input value
+pub trait TryFromFfi: Sized {
+    /// Type which can be converted from a [`ReprC`] type that was sent over FFI
```
outoftardis, line 75:
```suggestion
    /// Type that can be converted from a [`ReprC`] type that was sent over FFI
```

ffi/src/slice.rs
```diff
@@ -0,0 +1,271 @@
+use crate::{AsReprC, FfiWriteOut, IntoFfi, ReprC};
+
+/// Trait which facilitates implementing [`IntoFfi`] for slices of foreign types
```
outoftardis, line 3:
```suggestion
/// Trait that facilitates the implementation of [`IntoFfi`] for slices of foreign types
```

mversic: done

ffi/src/lib.rs
mversic: done

crypto/src/lib.rs
```diff
@@ -358,21 +360,39 @@ impl From<multihash::ConvertError> for KeyParseError {
 impl std::error::Error for KeyParseError {}
 
 /// Public Key used in signatures.
-#[derive(DebugCustom, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Getters, Encode, IntoSchema)]
-#[getset(get = "pub")]
+// TODO: Find a way to remove `PartialOrd` since it
```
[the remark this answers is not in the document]
mversic: removed the comment

ffi/tests/gen_shared_fns.rs
```diff
@@ -57,17 +57,19 @@ fn gen_shared_fns() {
             cloned
         };
 
-        let mut is_equal = MaybeUninit::<bool>::new(false);
+        // TODO: Fix
```
appetrosyan, line 60: I'd suggest creating a ticket. This PR is in limbo for a month more less.

crypto/Cargo.toml
```diff
@@ -11,10 +11,9 @@ default = ["std"]
 # Enable static linkage of the rust standard library.
 # Please refer to https://docs.rust-embedded.org/book/intro/no-std.html
 std = ["ursa"]
+ffi = ["iroha_ffi"]
```
mversic, line 14: I have to introduce this feature flag temporarily because `iroha_ffi` requires `std`. The only reason it requires `std` is because of `catch_unwind`

ffi/src/owned.rs
```diff
@@ -0,0 +1,228 @@
...
+    T::Target: ReprC,
+{
+    type Target = LocalSlice<T::Target>;
+
+    fn into_ffi(self) -> Self::Target {
+        self.into_iter().map(IntoFfi::into_ffi).collect()
+    }
+}
+
+//impl<T: IntoFfi, const N: usize> TryFromReprC for [T; N]
```
Erigara, line 219: Commented code

ffi/src/primitives.rs
```diff
@@ -0,0 +1,222 @@
...
+
+impl<'itm> TryFromReprC<'itm> for bool {
+    type Source = <u8 as TryFromReprC<'itm>>::Source;
+    type Store = ();
+
+    unsafe fn try_from_repr_c(source: Self::Source, _: &mut ()) -> Result<Self, FfiResult> {
+        let source: u8 = TryFromReprC::try_from_repr_c(source, &mut ())?;
+
+        match source {
+            0 | 1 => Ok(source != 0),
```
Erigara, line 19: Consider extracting into function like `is_valid_bool`.

ffi/src/primitives.rs
```diff
@@ -0,0 +1,222 @@
...
+
+            fn into_ffi(source: &[Self]) -> Self::Target {
+                SliceRef::from_slice(source)
+            }
+        }
+        unsafe impl<'slice> IntoFfiSliceMut<'slice> for $ty {
+            type Target = SliceMut<'slice, $ty>;
+
+            fn into_ffi(source: &mut [Self]) -> Self::Target {
+                SliceMut::from_slice(source)
```
Erigara, line 216: Are you sure that it is ok for `f32` and `f64`? they are non-robust as i understend.

ffi/src/slice.rs
```diff
@@ -0,0 +1,414 @@
...
+impl<'slice, T: 'slice> AsReprCRef<'slice> for SliceRef<'slice, T> {
+    type Target = Self;
+
+    fn as_ref(&self) -> Self::Target {
+        *self
+    }
+}
+
+//impl<'slice, T: TryFromReprCSliceRef<'slice> + Clone> TryFromReprCSliceRef<'slice>
+//    for &'slice [T]
```
Erigara, line 292: Commented code

ffi/src/handle.rs
```diff
@@ -0,0 +1,174 @@
+//! Logic related to opaque pointer handles and functions that are common to multiple handle types
+
+/// Type of the handle id
+pub type Id = u8;
+
+/// Implement [`$crate::Handle`] for given types with the given initial handle id.
+#[macro_export]
+macro_rules! handles {
```
Erigara, line 8: I wonder why we need handles?
Why not have methods like A__clone, B__clone, ...?
Having an IDs seems a bit error prone.

schema/src/lib.rs
```diff
@@ -211,9 +211,8 @@ impl<I: IntoSchema, P: DecimalPlacesAware> IntoSchema for fixnum::FixedPoint<I,
                 decimal_places: P::decimal_places(),
             })
         });
-        if !metamap.contains_key(&I::type_name()) {
-            I::schema(metamap);
-        }
+
+        I::schema(metamap);
```
Erigara, line 215: I think we can introduce infinite recursion this way.

ffi/src/lib.rs
```diff
@@ -1,173 +1,381 @@
...
+    type Store = ();
+
+    unsafe fn try_from_repr_c(source: Self::Source, _: &mut ()) -> Result<Self, FfiResult> {
+        source.as_mut().ok_or(FfiResult::ArgIsNull)
+    }
+}
+
+impl<T: ReprC> OutPtr for *mut T {
+    fn is_valid(&self) -> bool {
+        (*self).is_null()
```
Erigara, line 187: I think should be !(*self).is_null(), no?

Erigara (changes requested):
Please, check `impl<T: ReprC> OutPtr for *mut T` i think it has bug in implementation.

data_model/src/name.rs
```diff
@@ -86,27 +90,40 @@ impl FromStr for Name {
...
 /// All of the given pointers must be valid
-#[cfg(feature = "ffi_api")]
+#[no_mangle]
+#[cfg(feature = "ffi")]
 #[allow(non_snake_case, unsafe_code)]
-pub unsafe extern "C" fn Name__from_str(
-    candidate: *const u8,
-    candidate_len: usize,
-    output: *mut *mut Name,
+pub unsafe extern "C" fn Name__from_str<'itm>(
```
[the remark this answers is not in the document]
mversic: Because macro doesn't supprt associated types yet. I opened a ticket for this

ffi/src/primitives.rs
mversic: Good point, I'll check

schema/src/lib.rs
mversic: I think you are correct. I should revert these

ffi/src/lib.rs
mversic: Yes

ffi/src/handle.rs
Erigara: I mean that that for example for inherit methods we generate methods like `Type__method_name`, could that be the case for `Clone`, `Drop`, ...
So insead of having handle ids we would have separate functions for each type.

mversic: Yes, that's the more safe alternative. And you might be correct it's a better approach. However, it introduces function declaration bloat in the binary/smartcontract so I went with multiplexing approach. I can turn this into an issue to discuss further and change the implementation

mversic: Handle ids are essentially a method multiplexing mechanism

ffi/src/lib.rs
```diff
@@ -1,173 +1,381 @@
...
+    /// All conversions from a pointer must ensure pointer validity beforehand
+    unsafe fn write(self, dest: Self::OutPtr) -> Result<(), FfiResult>;
+}
 
 /// Result of execution of an FFI function
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-// NOTE: Enum is `repr(i32)` becasuse WebAssembly supports only
-// u32/i32, u64/i64 natively. Otherwise, `repr(i8)` would suffice
-#[repr(i32)]
+#[repr(i8)]
```
SamHSmith, line 108: Why where you able to make this change?

ffi/tests/ffi_export.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+//#[test]
+//fn return_result() {
+//    let mut output = MaybeUninit::new(0);
+//
+//    unsafe {
+//        assert_eq!(
+//            FfiResult::ExecutionFail,
+//            FfiStruct__fallible_int_output(u8::from(false), output.as_mut_ptr())
```
SamHSmith, line 292: Commented Code.

SamHSmith (approved):
Other than the comments that exist I see no problems.

ffi/src/handle.rs
Erigara: Okay, I see it makes sense to reduce the memory footprint of the smart contract.

Thank you for the clarification.

> I can turn this into an issue to discuss further and change the implementation 

If you think this is necessary. 

Last question: 
Types in different crates could have overlapping handle ids because underling `__clone` is different for different crates, right?

ffi/src/lib.rs
mversic: I'm not, I decided against introducing wasm specifics at the moment. This will be restored

mversic: https://github.com/hyperledger/iroha/issues/2511

ffi/src/primitives.rs
mversic: https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/hyperledger/iroha/2512

ffi/derive/src/lib.rs
```diff
@@ -47,6 +48,405 @@ pub fn ffi_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+                Ok(store)
+            }
+        }
+    }
+}
+
+fn derive_into_ffi_for_item(_: &Ident) -> TokenStream2 {
+    quote! {
+        // TODO:
```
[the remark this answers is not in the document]
mversic: done

s8sato (approved):

ffi/derive/src/export.rs
```diff
@@ -0,0 +1,276 @@
...
+                    unsafe { &*(#arg_name as *const #elem) }
+                }
+            });
+        }
+
+        return stmt;
+    }
+
+    quote! {
+        let mut store = Default::default();
```
Erigara, line 175: Since proc-macros are unhygienic, consider using fully quilified paths like `core::default::Default`, to avoid possibility of using different `Default` defined for crate.

ffi/src/slice.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// All conversions from a pointer must ensure pointer validity beforehand
+    unsafe fn try_from_repr_c(
+        source: Self::Source,
+        store: &'slice mut Self::Store,
+    ) -> Result<&'slice mut [Self], FfiResult>;
+}
+
+/// Immutable slice with a defined C ABI layout. Consists of data pointer and length
+#[repr(C)]
+pub struct SliceRef<'data, T>(*const T, usize, PhantomData<&'data T>);
```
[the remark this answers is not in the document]
mversic: I believe that `&'data T` implies `T: 'data`

ffi/src/primitives.rs
```diff
@@ -0,0 +1,227 @@
...
+            unsafe fn try_from_repr_c(source: Self::Source, _: &mut Self::Store) -> Result<Self, FfiResult> {
+                Ok(source)
+            }
+        }
+
+        impl<'itm> TryFromReprCSliceRef<'itm> for $ty {
+            type Source = SliceRef<'itm, $ty>;
+            type Store = ();
+
+            unsafe fn try_from_repr_c(source: Self::Source, _: &'itm mut Self::Store) -> Result<&[Self], FfiResult> {
```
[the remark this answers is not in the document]
mversic: sometimes they are inferred, the trait signature enforces correctness

ffi/src/owned.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn into_ffi(self) -> Self::Target {
+        self.into_iter().map(IntoFfi::into_ffi).collect()
+    }
+}
+
+impl<'itm, T: TryFromReprC<'itm>> TryFromReprC<'itm> for Vec<T> {
+    type Source = SliceRef<'itm, T::Source>;
+    type Store = Vec<T::Store>;
+
+    // There will always be at least one element in the subslice
```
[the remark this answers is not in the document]
mversic: we place comments above allowed lints to know why they don't take place

ffi/src/lib.rs
```diff
@@ -1,173 +1,329 @@
...
+        Ok(())
+    }
+}
+impl<T: ReprC> OutPtrOf<Local<T>> for *mut T {
+    unsafe fn write(self, source: Local<T>) -> Result<(), FfiResult> {
+        if self.is_null() {
+            return Err(FfiResult::ArgIsNull);
+        }
+
+        self.write(source.0);
```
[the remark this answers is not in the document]
mversic: yes, ownership is never transferred across FFI. At least with the current design

ffi/derive/src/lib.rs
```diff
@@ -1,20 +1,21 @@
...
 
-mod bindgen;
 mod derive;
+mod export;
 mod impl_visitor;
 
 #[proc_macro_attribute]
 #[proc_macro_error::proc_macro_error]
-pub fn ffi_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn ffi_export(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
mversic: yes, I should have attached docs here

ffi/derive/src/export.rs
mversic: yeah, I tried to do that mostly, sometimes I missed it. These will be found if conflicting with something else

ffi/src/owned.rs
```diff
@@ -0,0 +1,206 @@
...
+#[derive(Clone)]
+#[repr(transparent)]
+pub struct Local<T>(pub(crate) T);
+
+/// Wrapper around [`Option<Box<[T]>>`] that is local to the conversion site. This structure
+/// carries ownership and care must be taken not to let it transfer ownership into FFI function
+// NOTE: It's not possible to mutably reference local context
+#[derive(Debug)]
+#[repr(C)]
+pub struct LocalSlice<T>(*mut T, usize);
```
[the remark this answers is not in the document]
mversic: that's true, I think I left it like this on purpose because I wasn't sure that it can be covariant. I figured this is more safe and that it can always be expanded to covariance if necessary. Do you think there might be a reason why it couldn't be covariant?

ffi/src/lib.rs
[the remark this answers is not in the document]
mversic: I added the `Copy` bound

ffi/src/slice.rs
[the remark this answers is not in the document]
s8sato: I think you are right @Erigara , from the doc `&'data T` in the 3rd field would not mean `T: 'data` in the 1st field

ffi/src/owned.rs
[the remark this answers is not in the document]
Erigara: > Do you think there might be a reason why it couldn't be covariant?

Uff, it's hard to tell really, i will try to do a bit more research about this.

Another concern that i have is drop checker, we don't signal that we own `T`.
[Nomicom](https://doc.rust-lang.org/nomicon/phantom-data.html) has relevant example.
And if we look at [`Unique`](https://docs.rs/rust-libcore/latest/src/core/ptr.rs.html#981-989) pointer which is used inside `Box`, we can see that it uses `PhantomData<T>` to make it clear that owns the data.
Is this relevant to our case? What do you think?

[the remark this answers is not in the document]
mversic: the ownership of data is a concern because of the drop check, I should add `PhantomData<T>` to `LocalSlice`

ffi/src/slice.rs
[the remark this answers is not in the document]
mversic: ok, I haven't looked at the `docs` initially :). I'll add the bound, but I'm surprised with this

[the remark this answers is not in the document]
mversic: I've added the explicit bound, but I'll point you to [this](https://rust-lang.github.io/rfcs/2093-infer-outlives.html)

ffi/src/owned.rs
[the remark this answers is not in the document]
mversic: > Uff, it's hard to tell really, i will try to do a bit more research about this.

I've made it covariant

ffi/src/slice.rs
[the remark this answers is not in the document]
mversic: actually, I just got lint telling me this bound is redundant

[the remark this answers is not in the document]
Erigara: ok, false alarm from me :)
i think the docs are not updated on time.

ffi/derive/src/impl_visitor.rs
```diff
@@ -1,18 +1,136 @@
...
+}
+
+pub struct ReturnArg<'ast> {
+    self_ty: &'ast syn::Path,
+    name: Ident,
+    type_: &'ast Type,
+}
 
 pub struct ImplDescriptor<'ast> {
+    /// Associated types used by this method
```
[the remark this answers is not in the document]
s8sato:
```suggestion
    /// Associated types in the impl block
```

The same for `ImplVisitor` @mversic

ffi/derive/src/impl_visitor.rs
```diff
@@ -254,26 +319,29 @@ impl<'ast> Visit<'ast> for ImplVisitor<'ast> {
...
+                _ => abort!(
+                    node,
+                    "Only methods or types are supported inside impl blocks"
+                ),
             }
-            _ => abort!(node, "Only methods are supported inside impl blocks"),
         }
     }
+    fn visit_impl_item(&mut self, _: &'ast syn::ImplItem) {
+        unreachable!("You souldn't have used this method")
```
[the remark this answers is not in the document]
s8sato: We use `unreachable!()` like this when the method will never called but its implementation is required by the trait, right @mversic ?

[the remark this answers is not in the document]
mversic: I removed it

s8sato (approved):

Erigara (approved):

PR #2318, opened 2022-06-06
[fix] #2128: Fix `MerkleTree` construction and iteration

### LINKED ISSUE #2128: Fix `MerkleTree` construction/destruction from/into iterators
See https://github.com/hyperledger/iroha/pull/2115#discussion_r852943513

s8sato:
#### when
```rust
    #[test]
    fn leaves_5() {
        let tree = (1..=5)
            .map(|i| Hash::prehashed([i; Hash::LENGTH]).typed())
            .collect::<MerkleTree<()>>();

        for node in tree.iter() {
            dbg!(node);
        }
        // after:
        // for leaf_hash in tree.into_iter() {
        //     dbg!(leaf_hash);
        // }
    }
```

#### before
```
              #0: iteration order
              c7: first 2 hex of hash
        ______/\______
      #8              #1
      f6              c0
    __/\__          __/\__
  #a      #9      #5      #2
  05      00      fc      17
                  /\      /\
                #7  #6  #4  #3
                01  02  03  04
```

#### after

```
              #-: iteration order
              9b: first 2 hex of hash
        ______/\______
      #-              #-
      0a              05
    __/\__          __/\__
  #-      #-      #-      #-
  fc      17      05      **
  /\      /\      /
#0  #1  #2  #3  #4
01  02  03  04  05
```

### Issue
- Closes #2128

### Benefits
- Fixes and optimizes `MerkleTree::get_leaf`
- Optimizes `MerkleTree::add`
- Fixes internal alignment and ordering of `MerkleTree`

### Possible Drawbacks
Changes the root hash, because of `nodes_pair_hash` logic change


crypto/src/merkle.rs
```diff
@@ -346,16 +364,81 @@ mod tests {
...
+    }
+
+    impl<T> MerkleTree<T> {
+        fn size(&self) -> usize {
+            self.iter().count()
+        }
+
+        fn depth(&self) -> u32 {
+            self.size().trailing_ones()
+        }
```
Arjentix, line 389: Is this always true? I can't get it...

crypto/src/merkle.rs
```diff
@@ -99,36 +108,38 @@ impl<U> FromIterator<HashOf<U>> for MerkleTree<U> {
...
-    /// Constructs new instance of the merkle tree
+    /// Construct [`MerkleTree`].
     pub const fn new() -> Self {
         MerkleTree {
             root_node: Node::Empty,
         }
     }
 
-    /// Returns leaf node
+    /// Get the hash of the `idx`th leaf node.
```
outoftardis, line 118:
```suggestion
    /// Get the hash of the `idx`-th leaf node.
```

crypto/src/merkle.rs
```diff
@@ -35,40 +34,40 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
 #[allow(clippy::module_name_repetitions)]
 pub enum Node<T> {
-    /// Node is root of a subtree
+    /// [`Subtree`] node.
     Subtree(Subtree<T>),
-    /// Leaf node
+    /// [`Leaf`] node.
     Leaf(Leaf<T>),
-    /// Empty node
+    /// Empty node.
```
outoftardis, line 63:
```suggestion
    /// An empty node.
```

crypto/src/merkle.rs
```diff
@@ -35,40 +34,40 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
+/// Represents a leaf node.
 #[derive(Debug)]
 pub struct Leaf<T> {
-    /// Hash of the node
+    /// Hash of this node.
     hash: HashOf<T>,
 }
 
-/// Binary Tree's node with possible variants: Subtree, Leaf (with data or links to data) and Empty.
+/// Binary tree's node with possible variants: [`Subtree`], [`Leaf`] (with data or links to data) and `Empty`.
```
outoftardis, line 55:
```suggestion
/// A binary tree node: [`Subtree`], [`Leaf`] (with data or links to data), or `Empty`.
```

crypto/src/merkle.rs
```diff
@@ -35,40 +34,40 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
     Leaf(Leaf<T>),
-    /// Empty node
+    /// Empty node.
     Empty,
 }
 
 #[derive(Debug)]
-/// BFS iterator over the Merkle tree
+/// BFS iterator over [`MerkleTree`].
 pub struct BreadthFirstIter<'itm, T> {
```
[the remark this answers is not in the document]
mversic: it is, this is a misnomer :face_exhaling:

crypto/src/merkle.rs
```diff
@@ -200,34 +198,50 @@ impl<T> Node<T> {
...
 }
 
 impl<'itm, T> BreadthFirstIter<'itm, T> {
     fn new(root_node: &'itm Node<T>) -> Self {
-        BreadthFirstIter {
-            queue: vec![root_node],
-        }
+        let mut nodes = root_node.bfs();
+        nodes.reverse();
+        BreadthFirstIter { stack: nodes }
```
mversic, line 232: I don't like how this isn't lazily evaluated. I am of opinion that iterators should be lazily evaluated, because sometimes iteration can finish before iterator is fully depleted.

mversic: This is another drawback. Can we find an algorithm that would make iterator lazily evaluated?

mversic: I think it should be possible

crypto/src/merkle.rs
s8sato: Indeed, the method assumes that the tree is constructed `from_iterator` and has a neat triangular shape. Its only usage is `geometry` test, which is rather a description of implementation than a test

crypto/src/merkle.rs
s8sato: I don't know if I should add an article to the noun at the beginning of the doc comment

appetrosyan: Referring back to the contributing guide, Articles are not needed if the doc string is referring to the object that it's describing, e.g. 

```
/// Node of Merkle (binary) tree
struct Node
``` 

the reasoning is to think about how it would look on docs.rs.

crypto/src/merkle.rs
```diff
@@ -80,9 +79,19 @@ impl<U> FromIterator<HashOf<U>> for MerkleTree<U> {
             .into_iter()
             .map(|hash| Node::Leaf(Leaf { hash }))
             .collect::<VecDeque<_>>();
```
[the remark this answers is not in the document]
mversic: I'd be careful to conclude that doing one iteration is faster than doing two iterations if you take into account cache locality. Also, it's hard to say with certainty how the high level source code would map to compiled code. In general, I would aim for a more beautiful code, i.e. simplistic, understandable, concise

I'm in favor of your suggestion. I t makes sense to implement `Ord` on a `Node` and the code would look a bit simpler

crypto/src/merkle.rs
Arjentix: I think at least comment about that would be helpful then

crypto/src/merkle.rs
appetrosyan: I wouldn't be so eager to jump into making this lazily evaluated. I'd wait before we had profiling data.

crypto/src/merkle.rs
```diff
@@ -35,54 +34,64 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
-        hashes.sort_unstable();
-        let mut nodes = hashes
+impl<T> FromIterator<HashOf<T>> for MerkleTree<T> {
+    fn from_iter<I: IntoIterator<Item = HashOf<T>>>(iter: I) -> Self {
+        let mut nodes = iter
             .into_iter()
             .map(|hash| Node::Leaf(Leaf { hash }))
             .collect::<VecDeque<_>>();
-        if nodes.len() % 2 != 0 {
+        nodes.make_contiguous().sort_unstable();
```
s8sato, line 81: Addressed https://github.com/hyperledger/iroha/pull/2318#discussion_r892530521, but haven't benchmarked

crypto/src/merkle.rs
```diff
@@ -35,54 +34,64 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
 }
 
+#[cfg(feature = "std")]
 #[derive(Debug)]
-/// BFS iterator over the Merkle tree
+/// BFS iterator over [`MerkleTree`].
 pub struct BreadthFirstIter<'itm, T> {
-    queue: Vec<&'itm Node<T>>,
+    queue: VecDeque<&'itm Node<T>>,
 }
```
s8sato, line 72: Addressed https://github.com/hyperledger/iroha/pull/2318#discussion_r893620502.
Inspired by [`traversal::Bft`](https://docs.rs/traversal/latest/src/traversal/bft.rs.html)

crypto/src/merkle.rs
mversic: am I correct that, in this case. `make_contiguous` won't make any changes to the queue since all the elements are returned sorted from the input iterator

mversic: which would mean that the `sort_unstable` is unnecessary?

If this is not the case, shouldn't `MerkleTree::IntoIterator` return elements in sorted order?

mversic: > If this is not the case, shouldn't `MerkleTree::IntoIterator` return elements in sorted order?

looking at it, the answer is no, BFS will not return elements in sorted order for `MerkleTree`. Now I'm wondering, shouldn't the `IntoIterator` return just the leaves? In which case the return type could be just the hash, not the entire `Node` which, I maintain, is an implementation detail

s8sato: - Some sorting is necessary in the construction, otherwise the same set of leaves results in different root hashes.
After consideration I decided not to impl `Ord` for entire `Node` because it's confusing with the BFS iteration order

- I agree `MerkleTree::iter` is only used in `MerkleTree::leaves` at the moment and there would be no drawbacks to combine them, I don't think `FromIterator` and `IntoIterator` should be always symmetrical.
If I made it symmetrical, `BreadthFirstIter` would turn into `LeavesIter`, which would bring little to developers who want to search the tree and do something with it

appetrosyan (dismissed):

crypto/src/merkle.rs
```diff
@@ -1,7 +1,7 @@
 //! Merkle tree implementation.
 
 #[cfg(not(feature = "std"))]
-use alloc::{boxed::Box, format, string::String, vec, vec::Vec};
+use alloc::{boxed::Box, collections::vec_deque::VecDeque, format, string::String, vec};
 #[cfg(feature = "std")]
 use std::collections::VecDeque;
```
Arjentix, line 6: So maybe you don't need the second `VecDeque` and `std` feature at all?

crypto/src/merkle.rs
```diff
@@ -35,54 +34,63 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
 #[derive(Debug)]
-/// BFS iterator over the Merkle tree
-pub struct BreadthFirstIter<'itm, T> {
-    queue: Vec<&'itm Node<T>>,
+/// BFS iterator over [`MerkleTree`].
+struct BreadthFirstIter<'itm, T> {
+    queue: VecDeque<&'itm Node<T>>,
 }
 
 #[cfg(feature = "std")]
```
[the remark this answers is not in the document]
s8sato: Yes. `MerkleTree::from_iter` depends on `Node::from_nodes`, which is enabled by `std` feature.
Similarly, if you go back to `Hash::new`, you can see [the need for static linkage with `ursa`](https://github.com/hyperledger/iroha/blob/916a2cc584ba110775eafeae69e9658157f704c9/crypto/Cargo.toml#L11-L13)

s8sato:
Converted to draft. Let me try to exclude empty nodes from hash calculation

s8sato:
Done https://github.com/hyperledger/iroha/pull/2318#issuecomment-1155979452.
This is also a step to [remove `Hash::zeroed`](https://github.com/hyperledger/iroha/blob/916a2cc584ba110775eafeae69e9658157f704c9/crypto/src/hash.rs#L47) and I'm actually trying to, but in this PR I'd like to [keep the impacted area to _transaction_ trees](https://github.com/s8sato/iroha/blob/5ae1ac85edc8b8c8e39bebb61779b243f4569b54/core/src/block.rs#L342)

crypto/src/merkle.rs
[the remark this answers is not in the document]
appetrosyan: Also, `#[cfg(feature = "std")]` is not only gating things that depend on the rust standard library, but also things that shouldn't be included in WASM. I doubt that we'd be manipulating Merkle trees from inside a smart-contract without FFI.

crypto/src/merkle.rs
s8sato: You're right, I said the wrong thing.
Indeed, I resolved the asymmetry between from/into iterator by removing `impl<'itm, T> IntoIterator for &'itm MerkleTree<T>`.

I'll make another suggestion which is probably close to your idea.
I found this would be even better because ordinary trees iterate leaves

s8sato: Done by just [renaming](https://github.com/hyperledger/iroha/pull/2318/commits/9afdc6440381435a07b2cd8ce142497d210ad201).

I considered
```rust
impl<T> IntoIterator for MerkleTree<T> {
    type Item = HashOf<T>;
    type IntoIter = LeafHashIterator<T>;
    //
}
```
, but left as is because an equivalent can be achieved by filtering `BreadthFirstNodeIterator`

crypto/src/merkle.rs
```diff
@@ -99,36 +107,41 @@ impl<U> FromIterator<HashOf<U>> for MerkleTree<U> {
...
         MerkleTree {
             root_node: Node::Empty,
         }
     }
 
-    /// Returns leaf node
+    /// Get the hash of the `idx`-th leaf node.
     pub fn get_leaf(&self, idx: usize) -> Option<HashOf<T>> {
-        self.root_node.get_leaf_inner(idx).ok()
+        self.iter().nth(idx)
```
mversic, line 119: this search can, most probably, be optimized to `log(n)`

mversic (changes requested):

crypto/src/merkle.rs
```diff
@@ -189,173 +189,142 @@ impl<T> Node<T> {
...
 
-impl<'itm, T> BreadthFirstIter<'itm, T> {
-    fn new(root_node: &'itm Node<T>) -> Self {
-        BreadthFirstIter {
-            queue: vec![root_node],
+impl<'itm, T> BreadthFirstNodeIterator<'itm, T> {
+    #[inline]
+    fn new(root: &'itm Node<T>) -> Self {
+        Self {
+            queue: VecDeque::from(vec![root]),
```
mversic, line 220: another nitpick, you could determine the max size of the queue and use `VecDeque::with_capacity`

crypto/src/merkle.rs
mversic: I would leave a comment at least

crypto/src/merkle.rs
```diff
@@ -99,36 +107,41 @@ impl<U> FromIterator<HashOf<U>> for MerkleTree<U> {
...
-        self.root_node.get_leaf_inner(idx).ok()
+        self.iter().nth(idx)
     }
 
-    /// Return the `Hash` of the root node.
-    pub const fn root_hash(&self) -> HashOf<Self> {
-        self.root_node.hash().transmute()
+    /// Get the hashes of the leaf nodes.
+    pub fn iter(&self) -> impl Iterator<Item = HashOf<T>> + '_ {
+        self.nodes().filter_map(Node::leaf_hash)
```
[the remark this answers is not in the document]
s8sato: Because I don't know how to realize both keeping `Node` private and implementing `IntoIter<Item = Node>` for pub `MerkleTree`

crypto/src/merkle.rs
s8sato: "ordinary" was overstated. I don't know trees so much

crypto/src/merkle.rs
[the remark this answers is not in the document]
mversic: so you don't think it should implement IntoIterator<Item = HashOf<T>>? :)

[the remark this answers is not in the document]
s8sato: Yeah, at the moment

crypto/src/merkle.rs
s8sato: Addressed in 6c93c3c6992dac9e4c880f2a87d83c9b1e089f6d

crypto/src/merkle.rs
s8sato: Addressed in 6c93c3c6992dac9e4c880f2a87d83c9b1e089f6d

s8sato: Do you think we should use DFS to iterate over nodes/leaves in the first place?
In that case, we would use `log(n_leaves)` capacity of `Vec` instead of `n_leaves` capacity of `VecDeque`.
However, this does not affect the tree structure and could be commented out and left to another optimizing PR

crypto/src/merkle.rs
```diff
@@ -189,173 +196,134 @@ impl<T> Node<T> {
...
+            (None, None) => return None,
+            (None, Some(r_hash)) => return Some(r_hash),
+            (Some(l_hash), None) => return Some(l_hash),
+            (Some(l_hash), Some(r_hash)) => (l_hash, r_hash),
+        };
         let sum: Vec<_> = left_hash
             .as_ref()
             .iter()
             .zip(right_hash.as_ref().iter())
             .map(|(l, r)| l.saturating_add(*r))
```
mversic, line 210: why don't we use `wrapping_add`? Seems as if the `saturating_add` would hit a ceiling

s8sato:
Converted to draft again to move to [array implementation](https://opendsa-server.cs.vt.edu/ODSA/Books/Everything/html/CompleteTree.html)

crypto/src/merkle.rs
```diff
@@ -35,106 +35,135 @@ impl<T: IntoSchema> IntoSchema for MerkleTree<T> {
...
             root_node: Node::Empty,
+            height: 0,
         }
     }
 
-    /// Returns leaf node
+    /// Get the hash of the `idx`-th leaf node.
     pub fn get_leaf(&self, idx: usize) -> Option<HashOf<T>> {
-        self.root_node.get_leaf_inner(idx).ok()
+        if 2_usize.pow(self.height) <= idx {
```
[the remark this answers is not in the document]
s8sato: I defined `height` to be 0 for an one-node tree (`MerkleTree::new`). So the total number of leaves is `2 ^ height`

crypto/src/merkle.rs
```diff
@@ -9,262 +9,256 @@ use iroha_schema::prelude::*;
...
 }
 
 #[cfg(feature = "std")]
 impl<T> FromIterator<HashOf<T>> for MerkleTree<T> {
     fn from_iter<I: IntoIterator<Item = HashOf<T>>>(iter: I) -> Self {
-        let mut nodes = iter
-            .into_iter()
-            .map(|hash| Node::Leaf(Leaf { hash }))
-            .collect::<VecDeque<_>>();
-        nodes.make_contiguous().sort_unstable_by_key(Node::hash);
```
s8sato: The sorting of leaves is not implemented in either `MerkleTree::from_iterator` or `MerkleTree::add`.
AFAICT there is no need to keep the root hash unique to a SET of leaves

crypto/src/merkle.rs
```diff
@@ -1,242 +1,256 @@
...
 use crate::HashOf;
 
-/// [Merkle Tree](https://en.wikipedia.org/wiki/Merkle_tree) used to validate and prove data at
-/// each block height.
-/// Our implementation uses binary hash tree.
+/// [Merkle Tree](https://en.wikipedia.org/wiki/Merkle_tree) used to validate `T`
 #[derive(Debug)]
-pub struct MerkleTree<T> {
-    root_node: Node<T>,
+pub struct MerkleTree<T>(Vec<Option<HashOf<T>>>);
```
mversic, line 14: if `Hash` is made `[NonNullU8, SIZE]` maybe compiler would be able to use niche optimization to represent `None` here. I'm saying this because that would shave of byte for every node of the merkle tree

mversic: This is something that can be considered for future work, not for this PR

appetrosyan: The problem with making it `NonZeroU8` is that then all bytes have to be non-zero. The odds of such a key not occurring are astronomically low.

mversic: I don't think you need `Option` for elements of the tree. I think that there should be no empty nodes if you know the size of the tree?

s8sato: If we want to take advantage of the properties of complete trees, we need some kind of "empty node".
Another way is to use `Hash::zeroed().typed()`

mversic: I don't understand why. Here is how I see it. Complete tree should:
* be backed with a vec of capacity `2^height`
* double the size every time `vec.len` is about to become `== vec.cap - 1`. Last element is always unused. Due to this, you'd have to be careful when calculating tree height out of `vec`'s capacity.
* number of nodes in the tree should always be `== vec.len` and all "empty" nodes are logically after `vec.len` and never accessed, i.e. in undefined state

crypto/src/merkle.rs
mversic: > I only see `Vec` around the merkle trees, but would there still be a problem?

`Vec` does guarantee ordering, and there is no `Ord` or `Eq` which would benefit from this (maybe there should be?).

> In that case I have to revert the `MerkleTree::add` optimization.

Our tree is being created out of iterators in the code. I don't think we're ever inserting just one element

Hypothetical case: there is a `HashSet<T>` out of which you create two `MerkleTree<T>`. You sign them with the same key but you find that the signatures don't match

I'm not pushing for sorting per se. It's a tradeoff we should discuss

mversic: One more thing. When creating from an iterator complexity should be O(nlog(n)). Therefore sorting will not affect complexity

crypto/src/merkle.rs
s8sato: For example, could I represent the tree in my description differently than `[9b, 0a, 05, fc, 17, 05, **, 01, 02, 03, 04, 05]`?
I can't think of a way to achieve both having no empty nodes in the middle of the array and having the tree be complete (or having utility methods equivalent to `parent()`, `l_child()`, etc.)

mversic: Are there empty nodes in the middle? I didn't think there should be any for complete tree. I'll take a closer look into implementation then

mversic: The tree in your example isn't a complete tree

crypto/src/merkle.rs
mversic: Yes, I've corrected myself

mversic: If `add` weren't necessary we could have the nodes sorted without paying any price. Do we need merkletree in kura?

s8sato: https://github.com/hyperledger/iroha/issues/2388

crypto/src/merkle.rs
s8sato: Since a node is represented as `Option<HashOf<T>>` at the moment, it is a complete tree in my mind if I believe the following definitions:
https://en.wikipedia.org/wiki/Binary_tree#Types_of_binary_trees
https://towardsdatascience.com/5-types-of-binary-tree-with-cool-illustrations-9b335c430254

> consider using `zeroed` hash

Yes. We would explore tradeoffs between time and space efficiency (skipping hash calculations and not having wrappers)

crypto/src/merkle.rs
mversic: I had a change of heart. I think it's better to preserve insertion order. You can reslove this as far as I am concerned

crypto/src/merkle.rs
appetrosyan: I'd create an issue for the remaining comments and merge the PR.

appetrosyan (approved):

crypto/src/merkle.rs
appetrosyan: I guess we can't reach consensus after all.

crypto/src/merkle.rs
s8sato: #2396

crypto/src/merkle.rs
```diff
@@ -111,6 +111,15 @@ impl<T> FromIterator<HashOf<T>> for MerkleTree<T> {
     }
 }
 
+impl<'item, T> IntoIterator for &'item MerkleTree<T> {
+    type Item = HashOf<T>;
+    type IntoIter = LeafHashIterator<'item, T>;
```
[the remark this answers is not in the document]
s8sato: I'll leave as is -- consume the tree and iterate over leaves.
Later if necessary we can revive `iter()` which does not consume the tree

crypto/src/merkle.rs
s8sato: Added a comment https://github.com/hyperledger/iroha/issues/2388#issuecomment-1166194214
Anyway keeping the insertion order improves (not completely purify) `Handler<GetBlockHash>::handle` since this was totally broken before the PR -- it was getting an ordering in hashes for the block height request

mversic (approved):

Issue #2294, opened 2022-05-31
Produce flamecharts from local benchmarks


SamHSmith:
The benchmarks we have are useful to measure the performance of our code. But if we want to improve the performance we need more information about how the code is running. That is why we need the option to produce a flamechart of the benchmark's execution. We already use `tracing` so we should be able to use `tracing-flame` to produce our flamecharts.

https://github.com/tokio-rs/tracing/tree/master/tracing-flame

PR #2324, opened 2022-06-06
[fix] #2294: Add flamegraph generation to oneshot.rs

SamHSmith:
### Description of the Change

Adds the ability to generate folded stack information from the oneshot tps benchmark. That data can then be used in performance analysis.

### Issue
#2294 

### Benefits
We can now generate flamecharts.

### Possible Drawbacks
There are no drawbacks per say. But we don't currently instrument that much so the charts generated are not that informative. That does not mean we should instrument everything since that would slow the application to a grinding halt. It simply means there is more work to do when optimizing, but this is a good start.

### DIFF
--- a/client/benches/tps/oneshot.rs
+++ b/client/benches/tps/oneshot.rs
@@ -2,10 +2,41 @@
 
 mod lib;
 
-#[allow(clippy::expect_used)]
+use std::{fs::File, io::BufWriter};
+
+use tracing_flame::{FlameLayer, FlushGuard};
+use tracing_subscriber::prelude::*;
+
+#[allow(clippy::expect_used, clippy::print_stdout, clippy::use_debug)]
 fn main() {
+    let args: Vec<String> = std::env::args().collect();
+    let mut flush_guard: Option<FlushGuard<BufWriter<File>>> = None;
+
+    if args.len() >= 2 {
+        let file = File::create(&args[1]).expect("valid path");
+
+        let flame_layer = FlameLayer::new(BufWriter::new(file))
+            .with_threads_collapsed(true)
+            .with_empty_samples(true);
+        flush_guard = Some(flame_layer.flush_on_drop());
+
+        tracing_subscriber::registry().with(flame_layer).init();
+        iroha_logger::disable_logger();
+    }
+
     let config = lib::Config::from_path("benches/tps/config.json").expect("Failed to configure");
     let tps = config.measure().expect("Failed to measure");
-    iroha_logger::info!(?config);
-    iroha_logger::info!(%tps);
+
+    match flush_guard {
+        Some(guard) => {
+            guard.flush().expect("Flushed data without errors");
+            println!("Tracing data outputted to file: {}", &args[1]);
+            println!("TPS was {}", tps);
+            println!("Config was {:?}", config);
+        }
+        None => {
+            iroha_logger::info!(?config);
+            iroha_logger::info!(%tps);
+        }
+    }
 }
--- a/logger/src/lib.rs
+++ b/logger/src/lib.rs
@@ -56,6 +56,17 @@ pub fn init(configuration: &Configuration) -> Result<Option<Telemetries>> {
     Ok(Some(setup_logger(configuration)?))
 }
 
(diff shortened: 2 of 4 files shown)


appetrosyan (dismissed):

s8sato (commented):
Please add to `tps/README.md`
- the usage, such as `cargo run --example tps-oneshot -- tracing.folded`
- the reference to how to visualize the result, such as [this](https://docs.rs/tracing-flame/0.2.0/tracing_flame/#generating-the-image)

Then I can approve after the checks passed

logger/src/lib.rs
```diff
@@ -56,6 +56,17 @@ pub fn init(configuration: &Configuration) -> Result<Option<Telemetries>> {
     Ok(Some(setup_logger(configuration)?))
 }
 
+/// Disables the logger by setting `LOGGER_SET` to true. Will fail
+/// if the logger has already been initialized. This function is
+/// required in order to generate flamegraphs and flamecharts.
+///
+/// Returns true on success.
+pub fn disable_logger() -> bool {
+    !LOGGER_SET
```
appetrosyan, line 65: This is a double negative. Better check if 
```suggestion
    LOGGER_SET.compare_exchange(false, true, ordering::AcqRel, Ordering::Relaxed).is_ok()
```

client/benches/tps/oneshot.rs
```diff
@@ -2,10 +2,41 @@
...
+    }
+
     let config = lib::Config::from_path("benches/tps/config.json").expect("Failed to configure");
     let tps = config.measure().expect("Failed to measure");
-    iroha_logger::info!(?config);
-    iroha_logger::info!(%tps);
+
+    match flush_guard {
+        Some(guard) => {
+            guard.flush().expect("flushed data without error");
```
outoftardis, line 32:
```suggestion
            guard.flush().expect("Flushed data without errors");
```

client/benches/tps/oneshot.rs
```diff
@@ -2,10 +2,41 @@
...
+            println!("Tracing data outputed to file: {}", &args[1]);
```
outoftardis, line 33:
```suggestion
            println!("Tracing data outputted to file: {}", &args[1]);
```

pesterev (changes requested):

client/benches/tps/oneshot.rs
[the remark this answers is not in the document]
appetrosyan: It's a benchmark, so it shouldn't make much difference, but I do prefer logging.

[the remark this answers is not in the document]
Erigara: We disable the logger in order to generate the flamegraph, which is why we use `println`, right?

client/benches/tps/oneshot.rs
```diff
@@ -2,10 +2,41 @@
...
+    }
+
     let config = lib::Config::from_path("benches/tps/config.json").expect("Failed to configure");
     let tps = config.measure().expect("Failed to measure");
-    iroha_logger::info!(?config);
-    iroha_logger::info!(%tps);
+
+    match flush_guard {
+        Some(guard) => {
+            guard.flush().expect("Flushed data without errors");
```
Erigara, line 32: We don't need to manually call `flush`, it will [call it on drop](https://docs.rs/tracing-flame/0.2.0/src/tracing_flame/lib.rs.html#363).

Erigara (approved):
An explicit `flush` suggestion may or may not be considered.

client/benches/tps/oneshot.rs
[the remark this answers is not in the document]
SamHSmith: @Erigara Yes that is correct.

client/benches/tps/oneshot.rs
SamHSmith: I think it's better to leave it explicit. That way when someone reads this code they are aware of what is happening.

Erigara: @SamHSmith, don't we call `flush` 2 times with this implementation?

pesterev (approved):

PR #2328, opened 2022-06-07
[refactor] #2204: Make Asset-related operations generic

### LINKED ISSUE #2204: Consider making Asset-related operations generic
Right now we have multiple very similar implementations of `impl Execute for Mint<Asset, u32 | u64 | u128>`, which have the potential of going out of sync with each other. 

We need to refactor our existing implementations to make the triples of `impl Mint`, `impl Burn` and `impl Transfer` a generic blanket implementation.

Erigara:

### Description of the Change

As mentioned by @appetrosyan in #2204 we have very similar implementations for asset-related operations (`Mint`, `Burn`, `Transfer`), purpose of this PR make them generic.

The approach i took was to extract parts where implementation diverge (arithmetic operations, conversation to `f64`, default values, expected value type is assertion) for different asset value types and encapsulate them in traits. 

- Add traits `InnerMint`, `InnerBurn`, `InnerTransfer` that contains generic blanket implementations;
- Add traits `CheckedAdd`/`CheckedSub` to encapsulate differences between asset value `checked_add`/`checked_sub`;
- Add trait `FloatMetricFrom` to encapsulate differences between asset value conversation to `f64`.
- Add trait `AssetInstructionInfo` to collect fields (asset value type, default asset value) related to asset instructions.

### Issue

- Closes #2204.

### Benefits

- Asset-related operations cannot be out of sync with each other.

### Possible Drawbacks

- One more call inside `excute` functions to inner `execute` blanket implementation increases overhead (most calls are inlined so overhead is small).

### Alternate Designs

Further improvements would be to change `metric` macro so that it can be parametrized in blanket implementations, with this change one could remove `InnerMint`, `InnerBurn`, `InnerTransfer` and create blanket implementations directly for `Mint<Asset, V>`, `Burn<Asset, V>`, `Transfer<Asset, V, Asset>`.

### DIFF
--- a/core/src/smartcontracts/isi/asset.rs
+++ b/core/src/smartcontracts/isi/asset.rs
@@ -1,7 +1,10 @@
 //! This module contains [`Asset`] structure, it's implementation and related traits and
 //! instructions implementations.
 
-use iroha_data_model::prelude::*;
+use iroha_data_model::{
+    prelude::*,
+    primitives::{CheckedOp, IntoMetric},
+};
 use iroha_telemetry::metrics;
 
 use super::prelude::*;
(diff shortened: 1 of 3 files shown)


core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +54,165 @@ pub mod isi {
...
+                ) -> Result<(), Self::Error> {
+                    <$ty as InnerMint>::execute(self, authority, wsv)
+                }
+            }
+        };
+    }
+
+    impl_mint!(u32, Error, "mint_qty");
+    impl_mint!(u128, Error, "mint_big_qty");
+    impl_mint!(Fixed, Error, "mint_fixed");
```
s8sato, line 214:
```suggestion
    impl_mint_asset!(u32, u128, Fixed);
```

I think you can go with the alternative design.
The metric names can be removed by having some map e.g. from `u32` to `"qty"`

```rust
impl Value {
    //
    fn metric_name(&self) -> &'static str {
        //
    }
}
```

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +54,165 @@ pub mod isi {
...
+        fn try_checked_add(self, rhs: Self) -> Result<Self, Error> {
+            Ok(self.checked_add(rhs).ok_or(MathError::Overflow)?)
+        }
+    }
 
-                Ok(AssetEvent::Added(asset_id.clone()))
-            })
+    impl TryCheckedAdd for Fixed {
+        fn try_checked_add(self, rhs: Self) -> Result<Self, Error> {
+            Ok(self.checked_add(rhs)?)
```
s8sato, line 102: I think we can unify this error handling with other cases. Then the new trait would not be necessary

```suggestion
            Ok(self.checked_add(rhs).ok_or(MathError::Overflow)?)
```

Erigara: Problem is that `Fixed::checked_add` returns `Result`, not `Option` like `u128::checked_add` or `u32::checked_add` and `ok_or` does not exists for `Result`

core/src/smartcontracts/isi/asset.rs
Erigara: Thank you for suggestion, do you have some kind of example for this? All my attempts to implement such design have failed, `#[metrics(+"metric_name")]` attribute requires literal value.

s8sato: I found that I haven't found any implementation for that. Have you considered changing the `metrics` attribute macro?

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +54,165 @@ pub mod isi {
         }
     }
 
-    impl<W: WorldTrait> Execute<W> for Mint<Asset, u32> {
-        type Error = Error;
+    /// Trait for asset values that could be converted to `f64` and observed by prometheus.
+    trait FloatMetricFrom {
```
appetrosyan, line 58: We need a separate trait because of the orphan rule, correct?

core/src/smartcontracts/isi/asset.rs
appetrosyan: That can be changed, to make the API for `Fixed` more in line with other numerical types.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +54,165 @@ pub mod isi {
...
+    }
 
-        #[metrics(+"mint_big_qty")]
-        fn execute(
-            self,
-            _authority: <Account as Identifiable>::Id,
-            wsv: &WorldStateView<W>,
-        ) -> Result<(), Self::Error> {
-            let asset_id = self.destination_id;
+    impl TryCheckedAdd for u32 {
```
appetrosyan, line 88: These traits could be moved to `iroha_data_model` or `iroha_data_primitives`. That would be more in line with our upcoming refactor of the data model.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +54,165 @@ pub mod isi {
...
 
                 Ok(AssetEvent::Added(asset_id.clone()))
-            })
+            })?;
+            Ok(())
         }
     }
 
+    macro_rules! impl_mint {
+        ($ty:ty, $error:ty, $metrics:literal) => {
```
appetrosyan, line 194:
```suggestion
        ($ty:ty, $metrics:literal) => {
```
Below you hard-code Error. The macro invocation repeats less.

core/src/smartcontracts/isi/asset.rs
appetrosyan: If you don't have any experience with `proc` macros, you can leave it as is and create a ticket to fix it. 

With that said, three lines vs. one line is still readable to me.

core/src/smartcontracts/isi/asset.rs
Erigara: Yeah, rust forbid to implement `From<u128>` for `f64`.

appetrosyan (commented):
Overall, I think this solution is on the right track, but could be simplified. 

Why did you choose to use macros and not blanket `impls`? You should also keep in mind that any non-uniformity between `Fixed` and `u32`/`u128` can be fixed in the `iroha_data_primitives`, so you can (and should) try to move the logic related to numerical types into that crate.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +54,165 @@ pub mod isi {
...
-        #[metrics(+"mint_qty")]
-        fn execute(
-            self,
-            _authority: <Account as Identifiable>::Id,
-            wsv: &WorldStateView<W>,
-        ) -> Result<(), Self::Error> {
-            let asset_id = self.destination_id;
+    impl FloatMetricFrom for u128 {
+        #[allow(clippy::cast_precision_loss)]
+        fn from(value: Self) -> f64 {
```
appetrosyan, line 64: inline

appetrosyan (commented):
Additionally, since most operations are primitive and you have no blanket implementations, you should be adding `inline` to most functions. This should avoid the extra overhead that you mention in the **Potential drawbacks**.

core/src/smartcontracts/isi/asset.rs
Erigara: But we still need to define `Checked(Add|Sub)` traits because we are not using crates like `num` or `num_taits` where they are defined for us (and Fixed is our custom type anyway), right?  Or i miss something?

I mean:

```rust
/// Trait for `checked_add` operation on iroha's primitive types.
pub trait CheckedAdd {
    fn checked_add(self, rhs: Self) -> Option<Self>
    where
        Self: Sized;
}

// impls for u32, u128

impl CheckedAdd for Fixed {
    #[inline]
    fn checked_add(self, rhs: Self) -> Option<Self> {
        self.checked_add(rhs).ok()
    }
}

fn test(a: impl CheckedAdd) -> Result<(), MathError> {
    // ...
    let c = a.checked_add(b).ok_or(MathError::Overflow)?
    Ok(())
}

```

appetrosyan:
RE blanket impls: You could technically change how the telemetry macro works to have them be blanket impls. But this is also good-enough. 

RE non-uniformity, You could change that behaviour in Fixed,though again, will approve regardless.

core/src/smartcontracts/isi/asset.rs
Erigara: I will create a ticket after merge.

data_model/primitives/src/lib.rs
```diff
@@ -15,3 +15,91 @@ extern crate alloc;
...
+    #[inline]
+    fn checked_add(self, rhs: Self) -> Option<Self> {
+        self.checked_add(rhs)
+    }
+}
+
+impl CheckedAdd for Fixed {
+    #[inline]
+    fn checked_add(self, rhs: Self) -> Option<Self> {
+        self.checked_add(rhs).ok()
```
Arjentix, line 74: Looks like it will be good to unify `Fixed::checked_add()` with types from `std`. Although it will stil require to use this trait...

data_model/primitives/src/lib.rs
```diff
@@ -15,3 +15,91 @@ extern crate alloc;
...
+    fn checked_add(self, rhs: Self) -> Option<Self>
+    where
+        Self: Sized;
+}
+
+impl CheckedAdd for u32 {
+    #[inline]
+    fn checked_add(self, rhs: Self) -> Option<Self> {
+        self.checked_add(rhs)
+    }
```
Arjentix, line 61: This works because `Self::checked_add()` has bigger priority than `<Self as CheckedAdd>::checked_add()`, yes?
This trick is also used in `tools/parity_scale_decoder/src/generate_map.rs`

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +57,93 @@ pub mod isi {
...
-                *quantity = quantity
-                    .checked_add(self.object)
-                    .ok_or(MathError::Overflow)?;
-                wsv.metrics.tx_amounts.observe(f64::from(*quantity));
+    /// Trait for collecting fields associated with asset isi instructions:
+    /// - asset value type
+    /// - default asset value
+    trait AssetInstructionInfo {
+        const EXPECTED_VALUE_TYPE: AssetValueType;
+        const DEFAULT_ASSET_VALUE: Self;
```
Arjentix, line 65: Interesting. Didn't think we can do that

data_model/primitives/src/lib.rs
```diff
@@ -15,3 +15,91 @@ extern crate alloc;
...
 pub mod fixed;
 pub mod small;
+
+use fixed::prelude::*;
+
+/// Trait for values that could be converted to `f64` and observed by prometheus.
+pub trait FloatMetricFrom {
+    /// Convert `value` to `f64`.
+    fn from(value: Self) -> f64;
+}
```
Arjentix, line 25: I'd prefer `IntoMetric` and `fn into(self) -> f64` more. But it's just my opinion

core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,89 +57,93 @@ pub mod isi {
...
+                ) -> Result<(), Self::Error> {
+                    <$ty as InnerMint>::execute(self, authority, wsv)
+                }
+            }
+        };
     }
 
+    impl_mint!(u32, "mint_qty");
+    impl_mint!(u128, "mint_big_qty");
+    impl_mint!(Fixed, "mint_fixed");
```
Arjentix, line 145: Hm, If we had a way to use metrics without macro `metrics` we would be able to implement mint without this macro

data_model/primitives/src/lib.rs
```diff
@@ -15,3 +15,91 @@ extern crate alloc;
...
+
+impl FloatMetricFrom for Fixed {
+    #[inline]
+    fn from(value: Self) -> f64 {
+        value.into()
+    }
+}
+
+/// Trait for `checked_add` operation on iroha's primitive types.
+pub trait CheckedAdd {
```
appetrosyan, line 50: I think checked operations need to be tied together into a (single) trait. It's unlike that we'll need only one of those two.

Arjentix (approved):

core/src/smartcontracts/isi/asset.rs
```diff
@@ -1,7 +1,10 @@
 //! This module contains [`Asset`] structure, it's implementation and related traits and
 //! instructions implementations.
 
-use iroha_data_model::prelude::*;
+use iroha_data_model::{
```
appetrosyan, line 4: Minor comments about this file's layout. You should probably move the macro invocations to the top of `mod isi`, this is what people will mainly be interested in. Then the `macro_rules` for mint, burn and transfer, with the assertions at the bottom. 

Alternatively, to keep things tidy, you could also put the implementations into `mod mint`, `mod burn` and `mod transfer`, so the business logic related to each operation is in one place. The previous comment holds, (impl invocations first, then macro, then asserts at bottom). 

This is not a hard requirement for approval, just a recommendation for how to organise code.

appetrosyan (dismissed):

data_model/primitives/src/lib.rs
Erigara: Yeah, i also think about this option, i will address your suggestion.

data_model/primitives/src/lib.rs
Erigara: Yes, we still need this trait, and `Fixed::checked_add` can fail for several reasons, so if we re-implement `checked_add` we will lose the source of the error.

data_model/primitives/src/lib.rs
Erigara: Yes, rust book has example of such behavior, [see](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name).

core/src/smartcontracts/isi/asset.rs
Erigara: I agree it will be better

appetrosyan (approved):

core/src/smartcontracts/isi/asset.rs
Arjentix: After that leave a link here, please

Arjentix (approved):

PR #2332, opened 2022-06-07
[feature] #2053: Add tests to 'private_blockchain'

### LINKED ISSUE #2053: Unit test the `private_blockchain` permissions validators
At present we don't have any unit tests for the `private_blockchain` use case. It would be important to include them in the test coverage of the integration tests.

pesterev:

### Description of the Change

I added tests for permissions related to account and some asset `queries` and also for permissions related to `register`. ~~But the `OnlyAccountsDomain` and `OnlyAccountsData` permissions need to be tested on many other queries.~~(will be in the next PRs) The tests are very different and it is very difficult to write a macro that would generate that tests.

- [x] register module
- [x] tests permissions with all account related and some asset related queries
- [x] tests permissions with other queries (resolved: will be in the next PRs)
- [x] weird impl `IsAllowed` for `ProhibitRegisterDomains`

### Issue
Partially resolves #2053

### Benefits

Improved coverage

### Possible Drawbacks

None

Using macros would reduce the number of tests, but the tests are too different.

Also we can just write tests only for queries with non-trivial logic.

### DIFF
--- a/permissions_validators/src/private_blockchain/mod.rs
+++ b/permissions_validators/src/private_blockchain/mod.rs
@@ -40,3 +40,625 @@ impl IsGrantAllowed for ProhibitGrant {
         Err("Granting at runtime is prohibited.".to_owned().into())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #![allow(clippy::restriction)]
+
+    use std::str::FromStr as _;
+
+    use super::*;
+
+    /// A test environment that contains test domains, accounts, and assets.
+    struct TestEnv {
+        /// Alice's Id that owns Gold and Bronze assets.
+        alice_id: AccountId,
+        /// Bob's Id that owns Silver asset.
+        bob_id: AccountId,
+        /// Carol's Id that owns Bronze asset.
+        carol_id: AccountId,
+        /// Gold asset's Id.
+        gold_asset_id: AssetId,
+        /// Gold asset definition's Id.
+        gold_asset_definition_id: AssetDefinitionId,
+        /// Silver asset's Id.
+        silver_asset_id: AssetId,
+        /// Silver asset definition's Id.
+        silver_asset_definition_id: AssetDefinitionId,
+        /// Bronze asset's Id.
+        bronze_asset_id: AssetId,
+        /// Bronze asset definition's Id.
+        bronze_asset_definition_id: AssetDefinitionId,
+        /// Wonderland is a domain that contains Alice and Bob
+        wonderland: (DomainId, Domain),
+        /// Denoland is a domain that contains Carol
+        denoland: (DomainId, Domain),
+        /// World state view contains wonderland and denoland domains.
+        wsv: WorldStateView,
+    }
+
+    impl TestEnv {
+        /// Creates a test environment
+        fn new() -> Self {
+            let alice_id = AccountId::from_str("alice@wonderland").expect("Valid");
+            let mut alice = Account::new(alice_id.clone(), []).build();
+
+            let gold_asset_definition_id =
+                AssetDefinitionId::from_str("gold#wonderland").expect("Valid");
+            let gold_asset_id = AssetId::new(gold_asset_definition_id.clone(), alice_id.clone());
+            let gold_asset = Asset::new(gold_asset_id.clone(), AssetValue::Quantity(100));
+
+            alice.add_asset(gold_asset);
+
+            let bob_id = AccountId::from_str("bob@wonderland").expect("Valid");
+            let mut bob = Account::new(bob_id.clone(), []).build();
+
+            let silver_asset_definition_id =
+                AssetDefinitionId::from_str("silver#wonderland").expect("Valid");
+            let silver_asset_id = AssetId::new(silver_asset_definition_id.clone(), bob_id.clone());
+            let silver_asset = Asset::new(silver_asset_id.clone(), AssetValue::Quantity(200));
+
+            bob.add_asset(silver_asset);
+
+            let carol_id = AccountId::from_str("carol@denoland").expect("Valid");
+            let mut carol = Account::new(carol_id.clone(), []).build();
+
+            let bronze_asset_definition_id =
+                AssetDefinitionId::from_str("bronze#denoland").expect("Valid");
+            let bronze_asset_id =
+                AssetId::new(bronze_asset_definition_id.clone(), carol_id.clone());
+            let bronze_asset = Asset::new(bronze_asset_id.clone(), AssetValue::Quantity(300));
+
+            carol.add_asset(bronze_asset.clone());
+
+            alice.add_asset(bronze_asset);
+
+            let wonderland_id = DomainId::from_str("wonderland").expect("Valid");
+            let mut wonderland = Domain::new(wonderland_id.clone()).build();
+
+            wonderland.add_account(alice);
+
+            wonderland.add_account(bob);
+
+            let denoland_id = DomainId::from_str("denoland").expect("Valid");
+            let mut denoland = Domain::new(denoland_id.clone()).build();
+
+            denoland.add_account(carol);
+
+            let wsv = WorldStateView::new(World::with(
+                [wonderland.clone(), denoland.clone()],
+                Vec::new(),
+            ));
+
+            Self {
+                alice_id,
+                bob_id,
+                carol_id,
+                gold_asset_id,
+                gold_asset_definition_id,
+                silver_asset_id,
+                silver_asset_definition_id,
+                bronze_asset_id,
+                bronze_asset_definition_id,
+                wonderland: (wonderland_id, wonderland),
+                denoland: (denoland_id, denoland),
+                wsv,
+            }
+        }
+    }
+
+    mod queries {
+        use super::*;
+
+        #[test]
+        fn find_all_accounts() {
+            let TestEnv {
+                alice_id,
+                bob_id,
+                carol_id,
+                wsv,
+                ..
+            } = TestEnv::new();
+
+            let op = QueryBox::FindAllAccounts(FindAllAccounts::new());
+
+            {
+                let only_accounts_domain: IsQueryAllowedBoxed = query::OnlyAccountsDomain.into();
+
+                assert!(only_accounts_domain.check(&alice_id, &op, &wsv).is_err());
+                assert!(only_accounts_domain.check(&bob_id, &op, &wsv).is_err());
+                assert!(only_accounts_domain.check(&carol_id, &op, &wsv).is_err());
+            }
+
+            {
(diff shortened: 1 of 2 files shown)


permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,421 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+        let instruction = Instruction::Register(RegisterBox::new(Account::new(
+            "bob@test".parse().expect("Valid"),
+            [],
+        )));
+
+        let wsv = WorldStateView::<World>::new(World::new());
+
+        assert!(register::ProhibitRegisterDomains
+            .check(&alice_id, &instruction, &wsv)
+            .is_err());
```
pesterev, line 406: By the way, I don't understand why this is forbidden to register account with `ProhibitRegisterDomains` permission. Looks like a bug.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,421 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+
+#[cfg(test)]
+mod tests {
+    #![allow(clippy::restriction)]
+
+    use std::str::FromStr as _;
+
+    use super::*;
+
+    type TestEnv = (
```
appetrosyan, line 50: Doesn't matter here, but this should be a `struct`. You can bind directly to names using Rust's pattern matching syntax.

permissions_validators/src/private_blockchain/mod.rs
appetrosyan: Maybe a good idea to ask in the chat.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,541 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+            carol_id,
+            wsv,
+            ..
+        } = TestEnv::new();
+
+        let op = QueryBox::FindAllAccounts(FindAllAccounts::new());
+
+        {
+            let only_accounts_domain: IsQueryAllowedBoxed<World> = query::OnlyAccountsDomain.into();
+
```
appetrosyan, line 163: There should be a negative test-case (with `is_ok()`).

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,541 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+        silver_asset_definition_id: AssetDefinitionId,
+        bronze_asset_id: AssetId,
+        bronze_asset_definition_id: AssetDefinitionId,
+        first_domain: (DomainId, Domain),
+        second_domain: (DomainId, Domain),
+        wsv: WorldStateView<World>,
+    }
+
+    impl TestEnv {
+        fn new() -> Self {
```
appetrosyan, line 66: With rust 1.62, we should be able to turn this into a `const` function.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,541 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+            // Second domain
+            let second_domain_id = DomainId::from_str("test1").expect("Valid");
+            let mut second_domain = Domain::new(second_domain_id.clone()).build();
+
+            // Second domain includes Carol
+            second_domain.add_account(carol);
+
+            let wsv = WorldStateView::<World>::new(World::with(
+                [first_domain.clone(), second_domain.clone()],
+                vec![],
```
appetrosyan, line 129: Style point: prefer `Vec::new()` for empty immutable vectors. This conveys to the reader that what you made empty is meant to remain empty.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,541 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+        let op2 = QueryBox::FindAssetDefinitionById(FindAssetDefinitionById::new(
+            silver_asset_definition_id,
+        ));
+        let op3 = QueryBox::FindAssetDefinitionById(FindAssetDefinitionById::new(
+            bronze_asset_definition_id,
+        ));
+
+        {
+            let only_accounts_domain: IsQueryAllowedBoxed<World> = query::OnlyAccountsDomain.into();
+
```
appetrosyan, line 443: This is an example of a good assertion block. You need to be able to demonstrate when it works and when it doesn't.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,541 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+            alice_id,
+            carol_id,
+            wsv,
+            gold_asset_id,
+            silver_asset_id,
+            bronze_asset_id,
+            ..
+        } = TestEnv::new();
+
+        let op1 = QueryBox::FindAssetById(FindAssetById::new(gold_asset_id));
```
appetrosyan, line 396: I'd give these operations more descriprtive names. e.g. `find_gold`, `find_bronze` etc.

permissions_validators/src/private_blockchain/mod.rs
appetrosyan: I'd remove the comments above blocks (but keep the grouping). Instead I'd add a doc-comment above the function, so people could hover and see both the assets, the domains and the relationships.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,541 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+
+        let wsv = WorldStateView::<World>::new(World::new());
+
+        assert!(register::ProhibitRegisterDomains
+            .check(&alice_id, &instruction, &wsv)
+            .is_ok());
+    }
+
+    #[test]
+    fn granted_allowed_register_domains_success() {
```
appetrosyan, line 530: I'd combine the test cases. I'd also give them better names: this one for example is closer to 
`add_register_domains_permission_allows_registering_domain`. You're documenting an invariant, in this case that it A) isn't possible to register the domain without this permission, B) that it becomes possible after adding the permission, C) that it becomes impossible again when you revoke it. 

All three should be checked in a single test.

permissions_validators/src/private_blockchain/mod.rs
appetrosyan: Yes. const new for some types is going to be stabilised soon, so it should be possible to propagate those changes to here.

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,633 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+        // Wonderland is a domain that contains Alice and Bob
+        wonderland: (DomainId, Domain),
+        // Denoland is a domain that contains Carol
+        denoland: (DomainId, Domain),
+        // World state view contains wonderland and denoland domains.
+        wsv: WorldStateView<World>,
+    }
+
+    impl TestEnv {
+        fn new() -> Self {
```
appetrosyan, line 78: If you have time, please add doc comments, which explain who has which assets to this associated function, or to the struct (preferably both, preferably using `#[doc(include...]`.

appetrosyan (approved):

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,633 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+
+        #[test]
+        fn find_all_accounts() {
+            let TestEnv {
+                alice_id,
+                bob_id,
+                carol_id,
+                wsv,
+                ..
+            } = TestEnv::new();
```
Arjentix, line 157: Interesting idea

permissions_validators/src/private_blockchain/mod.rs
Arjentix: Why not to use another *query* to check `is_ok()` variant?

Arjentix: Oh I see, it tests queries, not validators

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -38,3 +38,633 @@ impl<W: WorldTrait> IsGrantAllowed<W> for ProhibitGrant {
...
+                    .check(&carol_id, &find_bronze, &wsv)
+                    .is_err());
+                assert!(only_accounts_data
+                    .check(&alice_id, &find_bronze, &wsv)
+                    .is_err());
+            }
+        }
+    }
+
+    mod revoke_and_grant {
```
Arjentix, line 585: I'd suggest to add more negative tests scenarios in this module tests.
I.e. `add_register_domains_permission_allows_registering_domain_with_right_token()` checks that Alice can register new domainsm, but doesn't check if anyone without permission token can do it too.

Arjentix (changes requested):

permissions_validators/src/private_blockchain/mod.rs
pesterev: Added

permissions_validators/src/private_blockchain/register.rs
```diff
@@ -20,14 +20,19 @@ impl<W: WorldTrait> IsAllowed<W, Instruction> for ProhibitRegisterDomains {
...
         } else {
             return Ok(());
         };
-        Err("Domain registration is prohibited.".to_owned())
+
+        if let Ok(RegistrableBox::Domain(_)) = register.object.evaluate(wsv, &Context::new()) {
+            Err("Domain registration is prohibited.".to_owned())
+        } else {
+            Ok(())
+        }
```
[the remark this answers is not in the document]
pesterev: Updated

permissions_validators/src/private_blockchain/mod.rs
pesterev: I couldn't think of anything better than describing relationships in the comments to the fields of `TestEnv` structure. Comments are prompted when you hover over the field. I wanted to use something like `mermaid` but `rustdoc` doesn't support it. Creating a separate markdown file is probably too much.

appetrosyan: Sounds good.

appetrosyan (approved):
Excellent!

appetrosyan (approved):

Arjentix (approved):

Arjentix (approved):

appetrosyan (approved):

PR #2330, opened 2022-06-07
[fix] #2227: implement asset registration

mversic:
Signed-off-by: Marin Veršić <marin.versic101@gmail.com>

PR #2333, opened 2022-06-08
[fix] #2227: Implement Register and Unregister for Asset

### LINKED ISSUE #2227: Register Store asset: "Unsupported Register instruction"

QuentinI:
### Description of the Change

- Implemented Register and Unregister instructions for Asset
- Changed minting implementation to require registering the asset first
- Adjusted public validators accordingly

### Issue

Resolves #2227 

### Benefits

Consistency, no more "unsupported" Register/Unregister instructions

### Possible Drawbacks

Register asset and Mint should now be considered together when implementing permission systems.


core/src/smartcontracts/isi/asset.rs
```diff
@@ -54,6 +54,59 @@ pub mod isi {
         }
     }
 
+    #[allow(clippy::panic_in_result_fn)]
+    impl<W: WorldTrait> Execute<W> for Register<Asset> {
```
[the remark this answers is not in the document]
mversic: also, please compare with [this](https://github.com/hyperledger/iroha/pull/2330)

mversic (dismissed):

cli/src/torii/tests.rs
```diff
@@ -278,8 +278,12 @@ fn register_asset_definition(name: &str) -> Instruction {
     RegisterBox::new(AssetDefinition::quantity(asset_definition_id(name))).into()
 }
 
-fn mint_asset(quantity: u32, asset: &str, account: &str) -> Instruction {
-    MintBox::new(Value::U32(quantity), asset_id(asset, account)).into()
+fn register_asset(quantity: u32, asset: &str, account: &str) -> Instruction {
+    RegisterBox::new(Asset::new(
+        asset_id(asset, account),
+        AssetValue::Quantity(quantity),
```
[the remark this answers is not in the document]
QuentinI: Oh, thanks. I didn't know it was possible, I'll fix it everywhere.

configs/peer/genesis.json
```diff
@@ -62,6 +62,35 @@
...
+                      "Quantity": 0
+                    }
+                  }
+                }
+              }
+            }
+          }
+        },
         {
           "Mint": {
```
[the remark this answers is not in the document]
mversic: Ok, this was my dilema as well. I don't see a downside

We could still remove the mint instruction from genesis, though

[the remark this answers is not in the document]
mversic: The problem is it's implicit. It might be confusing to the user

permissions_validators/src/public_blockchain/burn.rs
```diff
@@ -5,7 +5,7 @@ use iroha_data_model::asset::DefinitionId;
 use super::*;
 
 declare_token!(
-    /// Can burn asset with the corresponding asset definition.
+    /// Can burn and unregister asset with the corresponding asset definition.
```
outoftardis, line 8:
```suggestion
    /// Can burn and unregister assets with the corresponding asset definition.
```

permissions_validators/src/public_blockchain/mint.rs
```diff
@@ -4,7 +4,7 @@ use iroha_data_model::asset::DefinitionId;
 use super::*;
 
 declare_token!(
-    /// Can mint asset with the corresponding asset definition.
+    /// Can register and mint asset with the corresponding asset definition.
```
outoftardis, line 7:
```suggestion
    /// Can register and mint assets with the corresponding asset definition.
```

core/src/wsv.rs
```diff
@@ -296,6 +296,7 @@ impl<W: WorldTrait> WorldStateView<W> {
     /// # Errors
     /// - No such [`Asset`]
     /// - No [`Account`] to which the [`Asset`] is associated.
+    /// - No [`Domain`] to which the [`Account`] is associated.
```
outoftardis, line 299:
```suggestion
    /// - No [`Account`] with which the [`Asset`] is associated.
    /// - No [`Domain`] with which the [`Account`] is associated.
```

permissions_validators/src/public_blockchain/burn.rs
```diff
@@ -70,22 +95,33 @@ impl<W: WorldTrait> HasToken<W> for GrantedByAssetCreator {
...
+        match instruction {
+            Instruction::Unregister(unregister) => {
+                if let IdBox::AssetId(asset_id) = unregister
+                    .object_id
+                    .evaluate(wsv, &Context::new())
+                    .map_err(|e| e.to_string())?
+                {
+                    Ok(CanBurnAssetWithDefinition::new(asset_id.definition_id).into())
+                } else {
+                    Err("Instruction is not unregister asset".to_owned())
```
outoftardis, line 107: here and in other similar error messages in this PR: I'd suggest saying what was expected, e.g. something like `Expected the unregister instruction` or `Expected the unregister asset instruction`

appetrosyan (commented):
This change is important for SDK developers. Need to update with #api_changes.

configs/peer/genesis.json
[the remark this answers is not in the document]
QuentinI: Sure, I'll create one.
I think I'll leave it as Mint + Register for now and will then change it depending on what we decide.

tools/kagami/src/main.rs
```diff
@@ -186,6 +186,14 @@ mod genesis {
             .with_asset("rose".parse()?, AssetValueType::Quantity)
             .finish_domain()
             .build();
+        let register = RegisterBox::new(iroha_core::tx::Asset::new(
```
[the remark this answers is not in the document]
QuentinI: Will do

configs/peer/genesis.json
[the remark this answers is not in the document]
QuentinI: Ah, I see it's already there.

client/tests/integration/asset.rs
```diff
@@ -9,6 +9,77 @@ use test_network::*;
...
+                && *asset.value() == AssetValue::Quantity(0)
+        })
+    })?;
+
+    let hash = test_client.submit(register_asset)?;
+
+    // But registering an asset to account already having one should fail
+    test_client.poll_request(client::transaction::by_hash(*hash), |tx| {
+        matches!(tx, TransactionValue::RejectedTransaction(_))
+    })?;
```
Arjentix, line 43:
```suggestion
    // But registering an asset to account already having one should fail
    assert!(test_client.submit_blocking(register_asset).is_err());
```

Arjentix: Should work the same but looks easier

core/src/smartcontracts/isi/account.rs
```diff
@@ -15,6 +15,59 @@ use crate::{ValidQuery, WorldStateView};
...
+        #[metrics(+"register_asset")]
+        fn execute(
+            self,
+            _authority: <Account as Identifiable>::Id,
+            wsv: &WorldStateView<W>,
+        ) -> Result<(), Self::Error> {
+            let asset_id = self.object.id().clone();
+
+            match wsv.asset(&asset_id) {
+                Err(e) if matches!(e, FindError::Asset(_)) => {
```
Arjentix, line 31:
```suggestion
                Err(FindError::Asset(_)) => {
```

core/src/smartcontracts/isi/account.rs
```diff
@@ -15,6 +15,59 @@ use crate::{ValidQuery, WorldStateView};
...
+                    let res = wsv.modify_account(&asset_id.account_id, |account| {
+                        // We are in an arm where asset doesn't exist
+                        assert!(account.add_asset(self.object).is_none());
```
Arjentix, line 34: What's the difference with `unwrap()`? Style? 🤔

core/src/smartcontracts/isi/account.rs
```diff
@@ -15,6 +15,59 @@ use crate::{ValidQuery, WorldStateView};
...
+        #[metrics(+"unregister_asset")]
+        fn execute(
+            self,
+            _authority: AccountId,
+            wsv: &WorldStateView<W>,
+        ) -> Result<(), Self::Error> {
+            let asset_id = self.object_id;
+            let account_id = asset_id.account_id.clone();
+
+            wsv.modify_account(&account_id, |acct| {
```
Arjentix, line 63: I'd like more clear name than more cryptographic one

core/src/smartcontracts/isi/mod.rs
```diff
@@ -263,11 +263,11 @@ impl<W: WorldTrait> Execute<W> for RegisterBox {
             RegistrableBox::AssetDefinition(asset_definition) => {
                 Register::<AssetDefinition>::new(*asset_definition).execute(authority, wsv)
             }
+            RegistrableBox::Asset(asset) => Register::<Asset>::new(*asset).execute(authority, wsv),
             RegistrableBox::Trigger(trigger) => {
                 Register::<Trigger<FilterBox>>::new(*trigger).execute(authority, wsv)
             }
             RegistrableBox::Role(role) => Register::<Role>::new(*role).execute(authority, wsv),
-            _ => Err(Error::Unsupported(InstructionType::Register)),
```
Arjentix, line 270: That's cool. Now we will have a compile-time error if forget to add branch for something new

client/tests/integration/unstable_network.rs
```diff
@@ -46,7 +46,7 @@ fn unstable_network_7_peers_1_fault() {
 }
 
 #[test]
-#[ignore = "This test does not guarantee to have positive outcome given a fixed time."]
+// #[ignore = "This test does not guarantee to have positive outcome given a fixed time."]
```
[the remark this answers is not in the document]
QuentinI: Oh, it's an accident

Arjentix:
I'll approve the when all tests will pass

core/src/smartcontracts/isi/account.rs
QuentinI: 🤦 
Right, thanks

QuentinI:
@Arjentix I don't think it's ready yet. Apart from there not being any consensus on #2339 , this PR as it is now introduces a loophole to mint non-mintable assets twice.

core/src/smartcontracts/isi/account.rs
QuentinI: It seems to me that it's more appropriate semantically, since it emphasizes that this should be unreachable state, not just unrecoverable. `unreachable!` would be even better semantically, but it would look awkward.

Arjentix:
Oh, yes, haven't think about it...

core/src/smartcontracts/isi/mod.rs
appetrosyan: Technically, most of our wildcards could be removed and our code would become more reliable.

appetrosyan:
Overall, I think we need to create an integration test for the double mint loophole. To me it seems that this could be fixed by pattern-matching on the initial amount, and registering the asset as `Mintable::Not` right after its initial registration. 

So if you want to defer the minting, you can still do it, but if all you want is to create a non-mintable asset with an initial quantity, you can do that too.

core/src/wsv.rs
```diff
@@ -295,7 +295,8 @@ impl<W: WorldTrait> WorldStateView<W> {
     ///
     /// # Errors
     /// - No such [`Asset`]
-    /// - No [`Account`] to which the [`Asset`] is associated.
+    /// - No [`Account`] with which the [`Asset`] is associated.
+    /// - No [`Domain`] with which the [`Account`] is associated.
```
outoftardis, line 299:
```suggestion
    /// - The [`Asset`] is not associated with any [`Account`].
    /// - The [`Account`] is not associated with any [`Domain`].
```

core/src/genesis.rs
```diff
@@ -476,6 +502,28 @@ impl RawGenesisDomainBuilder {
...
+    /// Finish this account and return to
+    /// genesis domain building.
+    pub fn finish_account(self) -> RawGenesisDomainBuilder {
+        RawGenesisDomainBuilder {
+            domain_id: self.account_id.domain_id,
+            transaction: self.transaction,
+        }
+    }
+
+    /// Add [`Asset`] to current account initialized with the given value
```
outoftardis, line 515:
```suggestion
    /// Add [`Asset`] to current account and initialize it with the given value
```

core/src/genesis.rs
```diff
@@ -476,6 +502,28 @@ impl RawGenesisDomainBuilder {
     }
 }
 
+impl RawGenesisAccountBuilder {
+    /// Finish this account and return to
```
outoftardis, line 506:
```suggestion
    /// Finish building this account and return to
```

core/src/genesis.rs
```diff
@@ -443,25 +450,44 @@ impl RawGenesisDomainBuilder {
...
+            .push(RegisterBox::new(account_id.clone()).into());
+        RawGenesisAccountBuilder {
+            account_id,
+            transaction: self.transaction,
+        }
     }
 
-    /// Add an account to this domain
+    /// Add an account to this domain and return account builder
+    /// which can be used to register assets to the account.
```
outoftardis, line 468:
```suggestion
    /// Add an account to this domain and return an account builder,
    /// which can be used to register assets to this account.
```

QuentinI:
Based on #2339 I removed requirement to register asset before minting and removed `Mintable::Once`.

I experimented with default values in asset definitions, but it ultimately resulted in convoluted API when you take NFTs into account, so I scrapped that. I still think asset registration semantics as implemented here is not ideal, so I would maybe like to try and revisit that after release.

QuentinI:
> Overall, I think we need to create an integration test for the double mint loophole. To me it seems that this could be fixed by pattern-matching on the initial amount, and registering the asset as `Mintable::Not` right after its initial registration.
> 
> So if you want to defer the minting, you can still do it, but if all you want is to create a non-mintable asset with an initial quantity, you can do that too.

@appetrosyan sorry, I missed your comment.
I think disallowing minting altogether and leaving registration as the only way to add non-mintable assets to accounts results in more straightforward API. Is there a particular reason you would want to support minting in this case?

core/src/wsv.rs
QuentinI: It's not that it's not associated with any account/domain, it's that specific account/domain it's supposed to be associated with doesn't exist.

Would "Assotiated X doesn't exist" be better?

client/tests/integration/non_mintable.rs
```diff
@@ -8,30 +8,27 @@ use iroha_data_model::{metadata::UnlimitedMetadata, prelude::*};
...
+fn non_mintable_asset_can_be_registered_but_cannot_be_minted() -> Result<()> {
     let (_rt, _peer, mut test_client) = <PeerBuilder>::new().start_with_runtime();
     wait_for_genesis_committed(&vec![test_client.clone()], 0);
 
     // Given
     let account_id = AccountId::from_str("alice@wonderland").expect("Valid");
     let asset_definition_id = AssetDefinitionId::from_str("xor#wonderland").expect("Valid");
     let create_asset =
-        RegisterBox::new(AssetDefinition::quantity(asset_definition_id.clone()).mintable_once());
+        RegisterBox::new(AssetDefinition::quantity(asset_definition_id.clone()).not_mintable());
```
appetrosyan, line 19: I don't see how `mintable_once` is problematic, in this context: these assets can really be minted once.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -16,6 +16,53 @@ use super::prelude::*;
...
+                Err(e) => Err(Error::Find(Box::new(e))),
+                Ok(_) => Err(Error::Repetition(
+                    InstructionType::Register,
+                    IdBox::AssetId(asset_id.clone()),
+                )),
+            }
+        }
+    }
+
+    impl Execute for Unregister<Asset> {
```
appetrosyan, line 50: This is an important point for SDK developers.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -318,15 +365,6 @@ pub mod isi {
         match definition.mintable() {
```
appetrosyan, line 365: At this point, we could revert back to a boolean. But I would recommend keeping the `Mintable::Once` variant. We want to offer people the choice to register an asset definition but to defer the minting.

appetrosyan (changes requested):

core/src/wsv.rs
appetrosyan:
```suggestion
    /// - The [`Account`] with which the [`Asset`] is associated doesn't exist. 
    /// - The [`Domain`] with which the [`Account`] is associated doesn't exist.
```

core/src/smartcontracts/isi/asset.rs
QuentinI: You can still register _asset definition_ without minting.
After this PR `Mintable::Not` behaves essentially the same as `Mintable::Once` currently, since you can register _asset to an account_ with whatever value, that's why I removed it.

> At this point, we could revert back to a boolean

Didn't do it based on [this](https://github.com/hyperledger/iroha/blob/916a2cc584ba110775eafeae69e9658157f704c9/data_model/src/asset.rs#L191)

client/tests/integration/non_mintable.rs
QuentinI: After this PR they become actually non-mintable, only registrable. See my other [comment](https://github.com/hyperledger/iroha/pull/2333#discussion_r900331226)

core/src/smartcontracts/isi/asset.rs
```diff
@@ -16,6 +16,55 @@ use super::prelude::*;
...
+            match wsv.asset(asset_id) {
+                Err(e) if matches!(e, FindError::Asset(_)) => {
+                    assert_can_register(&asset_id.definition_id, wsv, self.object.value())?;
+
+                    if wsv
+                        .asset_or_insert(asset_id, self.object.value().clone())
+                        .is_err()
+                    {
+                        unreachable!("Account is checked to exist")
+                    }
```
pesterev, line 40:
```suggestion
                    wsv.asset_or_insert(asset_id, self.object.value().clone()).expect("Account exists");
```
I think it should be replaced with `expect()` so that the panic matches the semantics and explains what is expected here.

pesterev (changes requested):

appetrosyan:
> I think disallowing minting altogether and leaving registration as the only way to add non-mintable assets to accounts results in more straightforward API. Is there a particular reason you would want to support minting in this case?

The APIs are comparably simple, but in different ways. 

If you retained the `Mintable::Once` variant, you can create an asset and can defer the amount that you wanted to mint. Why is this good? Consider that we made `Mint` variadic: it accepts multiple `(AccountId, Asset, Quantity)` tuples, and so if you want to create a non-mintable asset that is split evenly between e.g. shareholders of a company, you can do that in two instructions. If we don't allow it, we increase the load on the network, because to achieve the same result the person doing it has to split the single `Mint` ISI into a `Mint` and several transfers. 

Moreover, we will need to synchronise the behaviour of `Mint` with the implementation in `Execute for Register<AssetDefinition>`, so nominally you have fewer instructions but each of them is more complex. 

Both approaches are valid: in the CPU world what you propose is called CISC, while my stance is closer to RISC with SIMD. I personally would prefer that we kept the API changes in this PR to a minimum, but will approve given a good justification.

QuentinI:
Right, my stance on this was the way it was mostly because I confused AssetDefinition and Asset in regards to Mintable.
I reverted to having Mintable::Once and implemented your proposed solution for double-mint loophole:

> So  if Register has value 0, creates a Mintable::Once asset, while a register with any other non-zero value necessarily a Mintable::Not variant

I have one question in regards to that - in this implementation one can register Mintable::Once asset to any number of accounts as long as Register has value 0. Are we okay with that?

core/src/smartcontracts/isi/asset.rs
```diff
@@ -16,6 +16,50 @@ use super::prelude::*;
 pub mod isi {
     use super::*;
 
+    #[allow(clippy::expect_used, clippy::unwrap_in_result)]
+    impl Execute for Register<Asset> {
```
mversic, line 20: `Register<Asset>` and `Unregister<Asset>` should be placed in `smartcontracts/account.rs` since that seems to be the convention

core/src/smartcontracts/isi/asset.rs
```diff
@@ -319,12 +376,26 @@ pub mod isi {
             Mintable::Infinitely => Ok(()),
             Mintable::Not => Err(Error::Mintability(MintabilityError::MintUnmintable)),
             Mintable::Once => {
```
mversic, line 378: if we'll allow for initial value to be set in the asset registration request (as far as I see this is the case) then I don't see a need for `Mintable` to be enum, rather it can be a `bool` since there is no need for `Mintable::Once` any longer

SamHSmith (approved):
LGTM. But you should probably address marin's comment.

core/src/smartcontracts/isi/asset.rs
QuentinI: Right, I guess I messed up while rebasing and it ended back there.

appetrosyan (approved):

appetrosyan (approved):

Issue #2137, opened 2022-04-22
Prevent tests failing if they can't get access to a port


appetrosyan:
Our current solution of randomly mapping ports at run-time is not the best, because of how it can fail on restricted systems, but also fails to work corretly on e.g. nextest (#2136). 

My preferred solution would be to actively handle the OS error and re-assign the port in case of failure, so that random failed port assignments don't bring down the entire test harness.

PR #2336, opened 2022-06-08
[fix] #2137: Prepare tests for multiprocess context

ales-tsurko:

### DIFF
--- a/cli/src/torii/tests.rs
+++ b/cli/src/torii/tests.rs
@@ -42,7 +42,7 @@ async fn try_init_network(config: &mut Configuration) -> Result<IrohaNetwork, ir
         Broker::new(),
         config.torii.p2p_addr.clone(),
         config.public_key.clone(),
-        config.network.mailbox,
+        config.network.actor_channel_capacity,
     )
     .await
 }
--- a/core/test_network/Cargo.toml
+++ b/core/test_network/Cargo.toml
@@ -20,11 +20,13 @@ iroha_actor = { version = "=2.0.0-pre-rc.5", path = "../../actor" }
 iroha = { path = "../../cli", features = ["test-network"] }
 
 eyre = "0.6.5"
-tempfile = "3"
-unique_port = "0.1.0"
-tokio = { version = "1.6.0", features = ["rt", "rt-multi-thread", "macros"]}
-rand = "0.8"
 futures = { version = "0.3.17", default-features = false, features = ["std", "async-await"] }
+rand = "0.8"
+sysinfo = "0.24.3"
+tempfile = "3"
+thiserror = "1.0.28"
+tokio = { version = "1.6.0", features = ["rt", "rt-multi-thread", "macros"]}
+unique_port = "0.1.0"
 
 [dev-dependencies]
 
--- a/p2p/tests/integration/p2p.rs
+++ b/p2p/tests/integration/p2p.rs
@@ -209,6 +209,88 @@ async fn two_networks() {
     assert_eq!(connected_peers.peers.len(), 1);
 }
 
+async fn init_peers() -> Vec<(Addr<Network<TestMessage>>, String, Broker, KeyPair)> {
+    (0..10)
+        .map(|_| {
+            let broker = Broker::new();
+            let keypair = KeyPair::generate().unwrap();
+            let fut = {
+                let broker = broker.clone();
+                let keypair = keypair.clone();
+                async { (broker, keypair) }
+            };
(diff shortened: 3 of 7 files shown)


cli/src/torii/tests.rs
```diff
@@ -20,21 +20,35 @@ use iroha_core::{
 };
 use iroha_data_model::{account::GENESIS_ACCOUNT_NAME, prelude::*};
 use iroha_version::prelude::*;
-use tokio::time;
+use tokio::{time, time::Duration};
```
appetrosyan, line 23: Just FYI, This is a re-export of core::time::Duration.

cli/src/torii/tests.rs
```diff
@@ -20,21 +20,35 @@ use iroha_core::{
...
 
-async fn create_torii() -> (Torii<World>, KeyPair) {
-    let mut config = crate::samples::get_config(crate::samples::get_trusted_peers(None), None);
+async fn try_init_network(config: &mut Configuration) -> Result<IrohaNetwork, iroha_p2p::Error> {
     config.torii.p2p_addr = format!("127.0.0.1:{}", unique_port::get_unique_free_port().unwrap());
     config.torii.api_url = format!("127.0.0.1:{}", unique_port::get_unique_free_port().unwrap());
     config.torii.telemetry_url =
         format!("127.0.0.1:{}", unique_port::get_unique_free_port().unwrap());
+
+    time::sleep(Duration::from_millis(20)).await;
```
appetrosyan, line 39: How does this help? Can we reduce this number? When should we increase it?

cli/src/torii/tests.rs
```diff
@@ -20,21 +20,35 @@ use iroha_core::{
...
+        Broker::new(),
+        config.torii.p2p_addr.clone(),
+        config.public_key.clone(),
+        config.network.mailbox,
+    )
+    .await
+}
+
+async fn create_torii() -> (Torii<World>, KeyPair) {
+    let mut config = crate::samples::get_config(crate::samples::get_trusted_peers(None), None);
```
appetrosyan, line 51: use crate::samples;

cli/src/torii/tests.rs
```diff
@@ -56,16 +70,17 @@ async fn create_torii() -> (Torii<World>, KeyPair) {
...
-    )
-    .await
-    .expect("Failed to create network")
-    .start()
-    .await;
+
+    let network = loop {
+        // this loop is not infinite, because we call unwrap on unique_port::get_unique_free_port(),
+        // which leads to panic if all available ports are already in use
+        match try_init_network(&mut config).await {
```
appetrosyan, line 77: I'd be interested to know how many tries on average it takes for tests when run with nextest.

p2p/tests/integration/p2p.rs
```diff
@@ -42,6 +42,33 @@ fn setup_logger() {
...
+    broker: &Broker,
+    public_key: &iroha_crypto::PublicKey,
+) -> Result<(String, Network<TestMessage>), iroha_p2p::Error> {
+    let address = gen_address();
+
+    time::sleep(Duration::from_millis(20)).await;
+
+    Ok((
+        address.clone(),
+        Network::<TestMessage>::new(broker.clone(), address, public_key.clone(), 100).await?,
```
appetrosyan, line 55: How do we know that this isn't running `clone` in a loop? Would it make sense to replace clone with a different cheaper operation?

p2p/tests/integration/p2p.rs
```diff
@@ -51,18 +78,15 @@ async fn network_create() {
...
     )
     .unwrap();
-    let network =
-        Network::<TestMessage>::new(broker.clone(), address.clone(), public_key.clone(), 100)
-            .await
-            .unwrap();
-    network.start().await;
-    tokio::time::sleep(delay).await;
+
+    let (_, address) = init_network(broker.clone(), public_key.clone()).await;
```
appetrosyan, line 87: Perhaps you don't want to immediately deallocate the network?

core/test_network/Cargo.toml
```diff
@@ -20,11 +20,13 @@ iroha_actor = { version = "=2.0.0-pre-rc.5", path = "../../actor" }
 iroha = { path = "../../cli", features = ["test-network"] }
 
 eyre = "0.6.5"
+futures = { version = "0.3.17", default-features = false, features = ["std", "async-await"] }
```
appetrosyan, line 23: I'm assuming you sorted this alphabetically. It would be nice if we made it standard.

PR #2346, opened 2022-06-08
[fix] #2202: Fix total field in query response

### LINKED ISSUE #2202: Query pagination: `total` field is always greater than results count by 1
## Repro

- Up Iroha: `docker compose up`
- Query to it with e.g. `FindAllDomains` & some great pagination limit such as `15`

## Expected output

Expected that `total` will be less than limit and equals to results count.

## Actual output

In YAML format:

```yaml
pagination:
  page: 1
  page_size: 15
  total: 3
data:
  - id: wonderland
    accounts:
      - id: alice@wonderland
        assets:
          - account_id: alice@wonderland
            definition_id: "rose#wonderland"
            value:
              t: Quantity
              c: 13
        metadata: {}
        roles: []
    logo: ~
    metadata: {}
    asset_definitions:
      - id: "rose#wonderland"
        value_type: Quantity
        mintable: Infinitely
    triggers: 0
  - id: genesis
    accounts:
      - id: genesis@genesis
        assets: []
        metadata: {}
        roles: []
    logo: ~
    metadata: {}
    asset_definitions: []
    triggers: 0
```

Erigara:

### Description of the Change

The `total` is the number of values returned in the top-most level of an iterable container after applying filter and before pagination.

- Fix `total` field calculation in `handle_queries`;
- Extend test `torii_pagination` to check that `total` field value is unaffected by pagination. 

### Issue

Closes #2202.

### Benefits

The `total` field is now what the SDK developers expect.

### Possible Drawbacks

None.

### Usage Examples or Tests

```bash
cargo test --package iroha --lib -- torii::tests::torii_pagination --exact --nocapture
```

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -122,17 +122,19 @@ pub(crate) async fn handle_queries<W: WorldTrait>(
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
     let (valid_request, filter) = request.validate(&wsv, &query_validator)?;
     let original_result = valid_request.execute(&wsv)?;
-    let total: u64 = original_result
-        .len()
+    let result = filter.filter(original_result);
+    let (total, result) = if let Value::Vec(value) = result {
+        (
+            value.len(),
+            Value::Vec(value.into_iter().paginate(pagination).collect()),
+        )
+    } else {
+        (1, result)
+    };
+    let total = total
         .try_into()
         .map_err(|e: TryFromIntError| QueryError::Conversion(e.to_string()))?;
-
-    let result = filter.filter(original_result);
-    let result = QueryResult(if let Value::Vec(value) = result {
-        Value::Vec(value.into_iter().paginate(pagination).collect())
-    } else {
-        result
-    });
+    let result = QueryResult(result);
     let paginated_result = PaginatedQueryResult {
         result,
         pagination,
--- a/cli/src/torii/tests.rs
+++ b/cli/src/torii/tests.rs
@@ -114,22 +114,30 @@ async fn torii_pagination() {
         )
         .map(|result| {
             let Scale(query_result) = result.unwrap();
-            let VersionedPaginatedQueryResult::V1(PaginatedQueryResult { result, .. }) =
+            let VersionedPaginatedQueryResult::V1(PaginatedQueryResult { result, total, .. }) =
                 query_result;
 
             if let QueryResult(Value::Vec(domains)) = result {
-                domains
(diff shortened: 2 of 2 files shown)


cli/src/torii/tests.rs
```diff
@@ -114,22 +114,30 @@ async fn torii_pagination() {
...
             }
         })
     };
 
-    assert_eq!(get_domains(None, None).await.len(), 26);
-    assert_eq!(get_domains(Some(0), None).await.len(), 26);
-    assert_eq!(get_domains(Some(15), Some(5)).await.len(), 5);
-    assert_eq!(get_domains(None, Some(10)).await.len(), 10);
-    assert_eq!(get_domains(Some(1), Some(15)).await.len(), 15);
+    for (start, limit, len) in [
```
appetrosyan, line 128: This is an ingenious style. I think I'll use it more often.

cli/src/torii/routing.rs
```diff
@@ -122,17 +122,19 @@ pub(crate) async fn handle_queries<W: WorldTrait>(
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
     let (valid_request, filter) = request.validate(&wsv, &query_validator)?;
     let original_result = valid_request.execute(&wsv)?;
-    let total: u64 = original_result
-        .len()
-        .try_into()
-        .map_err(|e: TryFromIntError| QueryError::Conversion(e.to_string()))?;
-
     let result = filter.filter(original_result);
```
appetrosyan, line 125: @arndey Should we also include the number of filtered-out results in the total?

appetrosyan (approved):

cli/src/torii/routing.rs
```diff
@@ -122,17 +122,19 @@ pub(crate) async fn handle_queries<W: WorldTrait>(
...
-    let result = QueryResult(if let Value::Vec(value) = result {
-        Value::Vec(value.into_iter().paginate(pagination).collect())
+    let (total, result) = if let Value::Vec(value) = result {
+        (
+            value.len(),
+            Value::Vec(value.into_iter().paginate(pagination).collect()),
+        )
     } else {
```
s8sato, line 131: Practical enough, though not enough to deal with the case of nested `Value::Vec`

s8sato (approved):

cli/src/torii/routing.rs
Erigara: We don't need to deal with the nested structures, see this [comment](https://github.com/hyperledger/iroha/issues/2202#issuecomment-1149995883).

PR #2363, opened 2022-06-15
[fix] #2137: Prepare tests for multiprocess context

### LINKED ISSUE #2137: Prevent tests failing if they can't get access to a port

ales-tsurko:
Closes #2137 

### Description of the Change

Update [`unique_port`](https://crates.io/crates/unique_port) with setter for the port index from which `get_unique_free_port` starts returning free ports. And with macro `generate_unique_start_port`, which generates a port number using the function name (including its path) where it's called.

Also, macro `prepare_test_for_nextest` added to `iroha_core::test_network` to avoid effects on `cargo test`.

### Issue

We'd like to use [`cargo-nextest`](https://nexte.st/) for tests, but because it's multiprocessing instead of multithreading, `unique_port` uses the own state in each process. This leads to address collisions between processes.

### Benefits

This prepares tests to work with `cargo-nextest`, while not having any effects on `cargo test`.

### Possible Drawbacks

There are still possible collisions in case of small difference between generated starting ports.

Proposed by @appetrosyan procedural macro — a counter, which will be generated in compile time. It needs procedural macro expertise and a bit more work though.

Another possible solution is using the file system. We could write generated ports into files, to share the state between processes. To sync it, we could use a `lockfile`, which wouldn't allow IO while it exists. While it's working, it slows down the tests. You could review the implementation of this solution [here](https://github.com/hyperledger/iroha/pull/2336).


core/test_network/src/unique_port.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
+        let name = &name[..name.len() - 3];
+        let mut hasher = DefaultHasher::new();
+        name.hash(&mut hasher);
+        1000 + (hasher.finish() % ((u16::MAX - 1000) as u64)) as u16
```
appetrosyan, line 24: I assume that the casts here are intentional. It's not obvious even to me, why the numbers are the way they are. 

Why isn't it, for example:
```suggestion
        (hasher.finish() % (u16::MAX as u64) as u16
```
I would add comments.

core/test_network/src/unique_port.rs
```diff
@@ -0,0 +1,65 @@
...
+/// # Examples
+/// ```
+/// use unique_port::get_unique_free_port;
+///
+/// let port_1 = get_unique_free_port().unwrap();
+/// let port_2 = get_unique_free_port().unwrap();
+/// assert_ne!(port_1, port_2);
+/// ```
+pub fn get_unique_free_port() -> Result<u16, String> {
+    let mut port_idx = PORT_IDX
```
appetrosyan, line 49: Do we still need this to be a lock?

appetrosyan: Why is it a `Mutex`?

core/test_network/src/unique_port.rs
```diff
@@ -0,0 +1,65 @@
+#![allow(clippy::restriction)]
+
+use once_cell::sync::Lazy;
+use std::net::{Ipv4Addr, SocketAddrV4, TcpListener};
+use std::ops::Range;
+use std::sync::Mutex;
+
+static PORT_IDX: Lazy<Mutex<u16>> = Lazy::new(|| Mutex::new(1000));
```
appetrosyan, line 8: Have you considered that this might not be necessary? Or that perhaps you could use say... `AtomicU16`? 
https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU16.html

client/tests/integration/offline_peers.rs
```diff
@@ -2,11 +2,12 @@
 
 use iroha_client::client;
 use iroha_data_model::prelude::*;
-use test_network::*;
+use test_network::{generate_offset, *};
 use tokio::runtime::Runtime;
 
 #[test]
 fn genesis_block_is_commited_with_some_offline_peers() {
+    unique_port::set_offset(generate_offset!()).unwrap();
```
appetrosyan, line 10: OK, so instead of having it be done by adding an extra parameter and letting us control which port goes in, which would allow us to re-factor the way in which we generate the unique port with ease, instead everything is done by mutating the compile-time internal state of `unique_port`. 

Why is this a better idea than just adding a parameter to a function?

core/test_network/src/unique_port.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut hasher = DefaultHasher::new();
+        name.hash(&mut hasher);
+        1000 + (hasher.finish() % ((u16::MAX - 1000) as u64)) as u16
+    }};
+}
+
+pub fn set_offset(offset: u16) -> Result<(), String> {
+    let mut port_idx = PORT_IDX
+        .lock()
+        .map_err(|_| "Failed to aquire the lock".to_owned())?;
```
appetrosyan, line 31: If you used an atomic, this wouldn't be necessary. The function would be infallible.

core/test_network/src/unique_port.rs
ales-tsurko: Yeah, I'll document everything before I open the PR.

core/test_network/src/unique_port.rs
ales-tsurko: I think we need it to be thread-safe. Not only because it won't work with `cargo test` without it, but to safely generate ports in multi-threading context.

For the `Mutex`, as I've written in the description,  it's just a copypasta of [unique_port](https://github.com/soramitsu/iroha2-unique_port) implementation. And I'd rather provide a PR with these changes there.

core/test_network/src/unique_port.rs
ales-tsurko: Not yet.

client/tests/integration/offline_peers.rs
ales-tsurko: Do you mean `get_unique_free_port`? My changes are more like extension to `unique_port` crate (and I'd rather made a PR with these changes to the `unique_port`'s repo) - adding a parameter to the function would make it backward-incompatible. Also, I kept it because, per the implementation, it has a global state (I think you mentioned it). But adding a parameter to the function would be a lot better, of course.

core/test_network/Cargo.toml
```diff
@@ -10,17 +10,21 @@ edition = "2021"
...
 eyre = "0.6.5"
+futures = { version = "0.3.17", default-features = false, features = ["std", "async-await"] }
+once_cell = "1.8.0"
+rand = "0.8"
 tempfile = "3"
-unique_port = "0.1.0"
 tokio = { version = "1.6.0", features = ["rt", "rt-multi-thread", "macros"]}
-rand = "0.8"
-futures = { version = "0.3.17", default-features = false, features = ["std", "async-await"] }
+unique_port = { version = "0.2.1", git = "https://github.com/ales-tsurko/iroha2-unique_port", rev = "09e95bb3608626c995b8e7e51c64ba7486c88948" }
```
Erigara, line 27: Shouldn't the `unique_port` upstream be updated and used as a dependency?

appetrosyan: No. We need to verify that it works first, and updating it quickly (without a PR) is only possible if it uses a personal fork. Before merging this needs to be PR'd into `unique_port` and this can go back to being the actual soramitsu repo.

Erigara:
Also, i think we should not switch back to `Mutex` ([comment](https://github.com/soramitsu/iroha2-unique_port/pull/9#issuecomment-1164081247) on related PR) or i miss something?

cli/src/torii/tests.rs
```diff
@@ -83,6 +84,7 @@ async fn create_torii() -> (Torii, KeyPair) {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn create_and_start_torii() {
+    prepare_test_for_nextest!();
```
appetrosyan, line 87: This is a little verbose but it does the job.

appetrosyan (approved):

Erigara (approved):

appetrosyan (approved):

Cargo.lock
```diff
@@ -4021,9 +4022,8 @@ checksum = "957e51f3646910546462e67d5f7599b9e4fb8acdd304b087a6494730f9eebf04"
 
 [[package]]
 name = "unique_port"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d79d1fad39d9571050e827f7a33fde5a50952f50e3ccbc007a6566881b17df5d"
+version = "0.2.1"
+source = "git+https://github.com/ales-tsurko/iroha2-unique_port?rev=09e95bb3608626c995b8e7e51c64ba7486c88948#09e95bb3608626c995b8e7e51c64ba7486c88948"
```
s8sato, line 4026: Do you think we need to manually maintain this? I [nearly spoiled it](https://github.com/hyperledger/iroha/pull/2394#discussion_r906047312)

ales-tsurko: I think, if not automatically, but with `cargo update`, it will change the link to crates.io. Because git in manifest is supposed to be fallback in case if the specified version is not available yet.

PR #2365, opened 2022-06-16
[refactor] #1985: Reduce size of `Name` struct

### LINKED ISSUE #1985: Reduce size of `Name` struct
`Name` is a `repr(transparent)` struct wrapping `std::string::String`, which has a pointer, a length and a capacity. 

However, in most cases a name isn't modified, but instead, a new name is created. So in theory we could get rid of `capacity`, as we don't intend for the `Name` struct to be mutable (there isn't a single mutable borrow or mutable `Name` in the code-base). 

So we should investigate either using `std::str::str` or creating a custom structure for storing `Name`s

Erigara:
### Description of the Change

- Add `ConstString`: immutable, inlinable string, which uses `union` internally to store boxed and inlined variants in the same space;
- Replace `String` with `ConstString` in `Name` struct.

### Issue

Closes #1985.

### Benefits

All structs that uses `Name` take up less space.
We win at least one machine word for every instance of `Name` and up to 23 byte for strings 15 bytes long on 64-bit architecture, also strings shorter than 15 bytes are stored on the stack, which avoids allocation.

### Possible Drawbacks

`ConstString` uses `unsafe` code, which increases the risk of bugs and vulnerabilities.

### Usage Examples or Tests

Run unit tests:
```
cargo test --package iroha_data_primitives --lib -- conststr::tests --nocapture
```

Run unit tests with miri to find leaks and UB:
```
cargo +nightly-2022-04-20 miri test --package iroha_data_primitives --lib -- conststr::tests --nocapture
```

### Alternate Designs

It possible to use enum instead of union to avoid `unsafe`, but it take 24 bytes instead of 16, [see](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a28c4bfc22d99f5808638dcc82c46553).

### DIFF
--- a/data_model/primitives/src/conststr.rs
+++ b/data_model/primitives/src/conststr.rs
@@ -1,0 +1,545 @@
+//! Const-string related implementation and structs.
+#[cfg(not(feature = "std"))]
+use alloc::{boxed::Box, str::from_boxed_utf8_unchecked, string::String, vec::Vec};
+use core::{
+    cmp::{Eq, Ord, Ordering, PartialEq, PartialOrd},
+    convert::{AsRef, From, TryFrom},
+    fmt::{Debug, Display, Formatter, Result as FmtResult},
+    hash::{Hash, Hasher},
+    mem::{size_of, ManuallyDrop},
+    ops::Deref,
+};
+#[cfg(feature = "std")]
+use std::str::from_boxed_utf8_unchecked;
+
+use iroha_schema::{IntoSchema, MetaMap};
+use parity_scale_codec::{Encode, Output};
+use serde::{Serialize, Serializer};
+
+const MAX_INLINED_STRING_LEN: usize = 2 * size_of::<usize>() - 1;
+
+/// Immutable inlinable string.
+/// Strings shorter than 15/7/3 bytes (depending on architecture 64/32/16 bit pointer width) are inlined.
+#[derive(Clone)]
+pub struct ConstString(ConstStringData);
+
+impl ConstString {
+    /// Returns the length of this [`ConstString`], in bytes.
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns `true` if [`ConstString`] is empty.
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.0.len() == 0
+    }
+
+    /// Returns `true` if [`ConstString`] data is inlined.
+    #[inline]
+    pub const fn is_inlined(&self) -> bool {
+        self.0.is_inlined()
+    }
+
+    /// Construct empty [`ConstString`].
+    #[inline]
+    pub const fn new() -> Self {
+        Self(ConstStringData::new())
+    }
+}
+
+impl Deref for ConstString {
+    type Target = str;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        self.as_ref()
+    }
+}
+
+impl<T: ?Sized> AsRef<T> for ConstString
+where
+    ConstStringData: AsRef<T>,
+{
+    #[inline]
+    fn as_ref(&self) -> &T {
+        self.0.as_ref()
+    }
+}
+
+impl<T: Into<ConstStringData>> From<T> for ConstString {
+    #[inline]
+    fn from(value: T) -> Self {
+        Self(value.into())
+    }
+}
+
+impl Display for ConstString {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(f, "{}", &self[..])
+    }
+}
+
+impl Debug for ConstString {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(f, "{}", &self[..])
+    }
+}
+
+impl Hash for ConstString {
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        (**self).hash(state)
+    }
+}
+
+impl PartialOrd for ConstString {
+    #[inline]
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        PartialOrd::partial_cmp(&self[..], &other[..])
+    }
+}
+
+impl Ord for ConstString {
+    #[inline]
+    fn cmp(&self, other: &Self) -> Ordering {
+        Ord::cmp(&self[..], &other[..])
+    }
+}
+
+impl PartialEq for ConstString {
+    #[inline]
+    fn eq(&self, other: &ConstString) -> bool {
+        PartialEq::eq(&self[..], &other[..])
+    }
+}
+
+macro_rules! impl_eq {
+    ($($ty:ty,)*) => {
+        impl_eq!($($ty),*);
+    };
+    ($($ty:ty),*) => {$(
+        impl PartialEq<$ty> for ConstString {
+            #[allow(clippy::string_slice)]
+            #[inline]
+            fn eq(&self, other: &$ty) -> bool {
+                PartialEq::eq(&self[..], &other[..])
+            }
+        }
+
+        impl PartialEq<ConstString> for $ty {
+            #[allow(clippy::string_slice)]
+            #[inline]
+            fn eq(&self, other: &ConstString) -> bool {
+                PartialEq::eq(&self[..], &other[..])
+            }
+        }
+    )*};
+}
+
+impl_eq!(String, str, &str);
+
+impl Eq for ConstString {}
+
+impl Serialize for ConstString {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(self)
+    }
+}
+
+impl Encode for ConstString {
+    fn size_hint(&self) -> usize {
+        self.as_bytes().size_hint()
+    }
+
+    fn encode_to<W: Output + ?Sized>(&self, dest: &mut W) {
+        self.as_bytes().encode_to(dest)
+    }
+
+    fn encode(&self) -> Vec<u8> {
+        self.as_bytes().encode()
+    }
+
+    fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
+        self.as_bytes().using_encoded(f)
+    }
+}
+
+impl IntoSchema for ConstString {
+    fn type_name() -> String {
+        String::type_name()
+    }
+    fn schema(map: &mut MetaMap) {
+        String::schema(map);
+    }
+}
+
+/// Union representing const-string variants: inlined or boxed.
+/// Distinction between variants are achieved by setting least significant bit for inlined variant.
+/// Boxed variant should have have 4/3/2 (depending on architecture 64/32/16 bit pointer width) trailing zeros due to pointer alignment.
+#[repr(C)]
+union ConstStringData {
+    inlined: InlinedString,
+    boxed: ManuallyDrop<BoxedString>,
+}
+
+impl ConstStringData {
+    #[inline]
+    const fn is_inlined(&self) -> bool {
+        self.inlined().is_inlined()
+    }
+
+    #[allow(unsafe_code)]
+    #[inline]
+    const fn inlined(&self) -> &InlinedString {
+        // SAFETY: safe to access if `is_inlined` == `true`.
+        unsafe { &self.inlined }
+    }
+
+    #[allow(unsafe_code)]
+    #[inline]
+    const fn boxed(&self) -> &ManuallyDrop<BoxedString> {
+        // SAFETY: safe to access if `is_inlined` == `false`.
+        unsafe { &self.boxed }
+    }
+
+    #[inline]
+    fn len(&self) -> usize {
+        if self.is_inlined() {
+            self.inlined().len()
+        } else {
+            self.boxed().len()
+        }
+    }
+
+    #[inline]
+    const fn new() -> Self {
+        Self {
+            inlined: InlinedString::new(),
+        }
+    }
+}
+
+impl<T: ?Sized> AsRef<T> for ConstStringData
+where
+    InlinedString: AsRef<T>,
+    BoxedString: AsRef<T>,
+{
+    #[inline]
+    fn as_ref(&self) -> &T {
+        if self.is_inlined() {
+            self.inlined().as_ref()
+        } else {
+            self.boxed().as_ref()
+        }
+    }
+}
+
+impl<T> From<T> for ConstStringData
+where
+    T: TryInto<InlinedString>,
+    <T as TryInto<InlinedString>>::Error: Into<BoxedString>,
+{
+    #[inline]
+    fn from(value: T) -> Self {
+        match value.try_into() {
+            Ok(inlined) => Self { inlined },
+            Err(value) => Self {
+                boxed: core::mem::ManuallyDrop::new(value.into()),
+            },
+        }
+    }
+}
+
+impl Clone for ConstStringData {
+    fn clone(&self) -> Self {
+        if self.is_inlined() {
+            Self {
+                inlined: *self.inlined(),
+            }
+        } else {
+            Self {
+                boxed: self.boxed().clone(),
+            }
+        }
+    }
+}
+
+impl Drop for ConstStringData {
+    #[allow(unsafe_code)]
+    fn drop(&mut self) {
+        if !self.is_inlined() {
+            // SAFETY: safe because`is_inlined` == `false`.
+            unsafe {
+                ManuallyDrop::drop(&mut self.boxed);
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+struct BoxedString(Box<str>);
+
+impl BoxedString {
+    #[inline]
+    const fn len(&self) -> usize {
+        self.0.len()
+    }
+}
+
+impl AsRef<str> for BoxedString {
+    #[inline]
+    fn as_ref(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl From<&str> for BoxedString {
+    #[allow(unsafe_code)]
+    #[inline]
+    fn from(value: &str) -> Self {
+        let payload = value.as_bytes().to_vec().into_boxed_slice();
+        // SAFETY: correct string.
+        Self(unsafe { from_boxed_utf8_unchecked(payload) })
+    }
+}
+
+impl From<String> for BoxedString {
+    #[inline]
+    fn from(value: String) -> Self {
+        Self(value.into_boxed_str())
+    }
+}
+
+#[derive(Clone, Copy)]
+#[repr(C)]
+struct InlinedString {
+    #[cfg(target_endian = "big")]
+    payload: [u8; MAX_INLINED_STRING_LEN],
+    /// Least-significant bit is always 1 to distinguish inlined variant.  
+    len: u8,
+    #[cfg(target_endian = "little")]
+    payload: [u8; MAX_INLINED_STRING_LEN],
+}
+
+impl InlinedString {
+    #[inline]
+    const fn len(self) -> usize {
+        (self.len >> 1_u8) as usize
+    }
+
+    #[inline]
+    const fn is_inlined(self) -> bool {
+        self.len % 2 > 0
+    }
+
+    #[inline]
+    const fn new() -> Self {
+        Self {
+            payload: [0; MAX_INLINED_STRING_LEN],
+            // Set least-significant bit to mark inlined variant.
+            len: 1,
+        }
+    }
+}
+
+impl AsRef<str> for InlinedString {
+    #[allow(unsafe_code)]
+    #[inline]
+    fn as_ref(&self) -> &str {
+        // SAFETY: created from valid utf-8.
+        unsafe { core::str::from_utf8_unchecked(&self.payload[..self.len()]) }
+    }
+}
+
+impl<'value> TryFrom<&'value str> for InlinedString {
+    type Error = &'value str;
+
+    #[allow(clippy::cast_possible_truncation)]
+    #[inline]
+    fn try_from(value: &'value str) -> Result<Self, Self::Error> {
+        let len = value.len();
+        if len > MAX_INLINED_STRING_LEN {
+            return Err(value);
+        }
+        let mut inlined = Self::new();
+        inlined.payload.as_mut()[..len].copy_from_slice(value.as_bytes());
+        // Truncation won't happen because we checked that the length shorter than `MAX_INLINED_STRING_LEN`.
+        // Set least-significant bit to mark inlined variant.
+        inlined.len += (len << 1_usize) as u8;
+        Ok(inlined)
+    }
+}
+
+impl TryFrom<String> for InlinedString {
+    type Error = String;
+
+    #[inline]
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        match Self::try_from(value.as_str()) {
+            Ok(inlined) => Ok(inlined),
+            Err(_) => Err(value),
+        }
+    }
+}
+
+impl Debug for InlinedString {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        f.debug_struct("InlinedString")
+            .field("len", &self.len())
+            .field("payload", &self.as_ref())
+            .finish()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use core::mem::{align_of, size_of};
+    use std::collections::hash_map::DefaultHasher;
+
+    use super::*;
+
+    fn run_with_strings(f: impl Fn(String)) {
+        [0, 1, 7, 8, 15, 16, 30]
+            .into_iter()
+            // utf-8 encodes ascii characters in single byte
+            .map(|len| "x".repeat(len))
+            .for_each(f)
+    }
+
+    #[test]
+    fn const_string_layout() {
+        assert_eq!(size_of::<ConstString>(), size_of::<Box<str>>());
+        assert_eq!(align_of::<ConstString>(), align_of::<Box<str>>());
+    }
+
+    #[test]
+    fn const_string_new() {
+        let const_string = ConstString::new();
+        assert_eq!(const_string, "");
+        assert_eq!(const_string.len(), 0);
+    }
+
+    #[test]
+    fn const_string_is_inlined() {
+        run_with_strings(|string| {
+            let len = string.len();
+            let const_string = ConstString::from(string);
+            let is_inlined = len <= MAX_INLINED_STRING_LEN;
+            assert_eq!(const_string.is_inlined(), is_inlined, "with len {}", len);
+        });
+    }
+
+    #[test]
+    fn const_string_len() {
+        run_with_strings(|string| {
+            let len = string.len();
+            let const_string = ConstString::from(string);
+            assert_eq!(const_string.len(), len);
+        });
+    }
+
+    #[test]
+    fn const_string_deref() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.as_str());
+            assert_eq!(&*const_string, &*string);
+        });
+    }
+
+    #[test]
+    fn const_string_from_string() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.clone());
+            assert_eq!(const_string, string);
+        });
+    }
+
+    #[test]
+    fn const_string_from_str() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.as_str());
+            assert_eq!(const_string, string);
+        });
+    }
+
+    #[test]
+    fn const_string_clone() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string);
+            let const_string_clone = const_string.clone();
+            assert_eq!(const_string, const_string_clone);
+        });
+    }
+
+    #[test]
+    fn const_string_display() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.clone());
+            assert_eq!(format!("{const_string}"), string);
+        });
+    }
+
+    #[test]
+    fn const_string_hash() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.clone());
+            let mut string_hasher = DefaultHasher::new();
+            let mut const_string_hasher = DefaultHasher::new();
+            string.hash(&mut string_hasher);
+            const_string.hash(&mut const_string_hasher);
+            assert_eq!(const_string_hasher.finish(), string_hasher.finish());
+        });
+    }
+
+    #[test]
+    fn const_string_eq_string() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.as_str());
+            assert_eq!(const_string, string);
+            assert_eq!(string, const_string);
+        });
+    }
+
+    #[test]
+    fn const_string_eq_str() {
+        run_with_strings(|string| {
+            let const_string = ConstString::from(string.as_str());
+            assert_eq!(const_string, string.as_str());
+            assert_eq!(string.as_str(), const_string);
+        });
+    }
+
+    #[test]
+    fn const_string_eq_const_str() {
+        run_with_strings(|string| {
+            let const_string_1 = ConstString::from(string.as_str());
+            let const_string_2 = ConstString::from(string.as_str());
+            assert_eq!(const_string_1, const_string_2);
+            assert_eq!(const_string_2, const_string_1);
+        });
+    }
+
+    #[test]
+    fn const_string_cmp() {
+        run_with_strings(|string_1| {
+            run_with_strings(|string_2| {
+                let const_string_1 = ConstString::from(string_1.as_str());
+                let const_string_2 = ConstString::from(string_2.as_str());
+                assert!(
+                    ((const_string_1 <= const_string_2) && (string_1 <= string_2))
+                        || ((const_string_1 >= const_string_2) && (string_1 >= string_2))
+                );
+                assert!(
+                    ((const_string_2 >= const_string_1) && (string_2 >= string_1))
+                        || ((const_string_2 <= const_string_1) && (string_2 <= string_1))
+                );
+            });
+        });
+    }
+}
--- a/data_model/primitives/src/conststr.rs
+++ b/data_model/primitives/src/conststr.rs
@@ -1,195 +1,90 @@
 //! Const-string related implementation and structs.
 #[cfg(not(feature = "std"))]
-use alloc::{boxed::Box, str::from_boxed_utf8_unchecked, string::String, vec::Vec};
+use alloc::{
+    borrow::{Borrow, ToOwned},
+    boxed::Box,
+    str::from_utf8_unchecked,
+    string::String,
+};
 use core::{
     cmp::{Eq, Ord, Ordering, PartialEq, PartialOrd},
-    convert::{AsRef, From, TryFrom},
-    fmt::{Debug, Display, Formatter, Result as FmtResult},
+    convert::TryFrom,
+    fmt,
     hash::{Hash, Hasher},
     mem::{size_of, ManuallyDrop},
     ops::Deref,
+    ptr::NonNull,
+    slice::{from_raw_parts, from_raw_parts_mut},
 };
 #[cfg(feature = "std")]
-use std::str::from_boxed_utf8_unchecked;
+use std::{borrow::Borrow, str::from_utf8_unchecked};
 
+use derive_more::{DebugCustom, Display};
 use iroha_schema::{IntoSchema, MetaMap};
-use parity_scale_codec::{Encode, Output};
-use serde::{Serialize, Serializer};
+use parity_scale_codec::{WrapperTypeDecode, WrapperTypeEncode};
+use serde::{
+    de::{Deserialize, Deserializer, Error, Visitor},
+    ser::{Serialize, Serializer},
+};
 
 const MAX_INLINED_STRING_LEN: usize = 2 * size_of::<usize>() - 1;
 
 /// Immutable inlinable string.
-/// Strings shorter than 15/7/3 bytes (depending on architecture 64/32/16 bit pointer width) are inlined.
-#[derive(Clone)]
-pub struct ConstString(ConstStringData);
-
-impl ConstString {
-    /// Returns the length of this [`ConstString`], in bytes.
-    #[inline]
-    pub fn len(&self) -> usize {
-        self.0.len()
-    }
-
-    /// Returns `true` if [`ConstString`] is empty.
-    #[inline]
-    pub fn is_empty(&self) -> bool {
-        self.0.len() == 0
-    }
-
-    /// Returns `true` if [`ConstString`] data is inlined.
-    #[inline]
-    pub const fn is_inlined(&self) -> bool {
-        self.0.is_inlined()
-    }
-
-    /// Construct empty [`ConstString`].
-    #[inline]
-    pub const fn new() -> Self {
-        Self(ConstStringData::new())
-    }
-}
-
-impl Deref for ConstString {
-    type Target = str;
-
-    #[inline]
-    fn deref(&self) -> &Self::Target {
-        self.as_ref()
-    }
-}
-
-impl<T: ?Sized> AsRef<T> for ConstString
-where
-    ConstStringData: AsRef<T>,
-{
-    #[inline]
-    fn as_ref(&self) -> &T {
-        self.0.as_ref()
-    }
-}
-
-impl<T: Into<ConstStringData>> From<T> for ConstString {
-    #[inline]
-    fn from(value: T) -> Self {
-        Self(value.into())
-    }
-}
-
-impl Display for ConstString {
-    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
-        write!(f, "{}", &self[..])
-    }
-}
-
-impl Debug for ConstString {
-    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
-        write!(f, "{}", &self[..])
-    }
-}
-
-impl Hash for ConstString {
-    #[inline]
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        (**self).hash(state)
-    }
-}
-
-impl PartialOrd for ConstString {
-    #[inline]
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        PartialOrd::partial_cmp(&self[..], &other[..])
-    }
-}
-
-impl Ord for ConstString {
-    #[inline]
-    fn cmp(&self, other: &Self) -> Ordering {
-        Ord::cmp(&self[..], &other[..])
-    }
-}
-
-impl PartialEq for ConstString {
-    #[inline]
-    fn eq(&self, other: &ConstString) -> bool {
-        PartialEq::eq(&self[..], &other[..])
-    }
-}
-
-macro_rules! impl_eq {
-    ($($ty:ty,)*) => {
-        impl_eq!($($ty),*);
-    };
-    ($($ty:ty),*) => {$(
-        impl PartialEq<$ty> for ConstString {
-            #[allow(clippy::string_slice)]
-            #[inline]
-            fn eq(&self, other: &$ty) -> bool {
-                PartialEq::eq(&self[..], &other[..])
-            }
-        }
-
-        impl PartialEq<ConstString> for $ty {
-            #[allow(clippy::string_slice)]
-            #[inline]
-            fn eq(&self, other: &ConstString) -> bool {
-                PartialEq::eq(&self[..], &other[..])
-            }
-        }
-    )*};
-}
-
-impl_eq!(String, str, &str);
-
-impl Eq for ConstString {}
-
-impl Serialize for ConstString {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        serializer.serialize_str(self)
-    }
-}
-
-impl Encode for ConstString {
-    fn size_hint(&self) -> usize {
-        self.as_bytes().size_hint()
-    }
-
-    fn encode_to<W: Output + ?Sized>(&self, dest: &mut W) {
-        self.as_bytes().encode_to(dest)
-    }
-
-    fn encode(&self) -> Vec<u8> {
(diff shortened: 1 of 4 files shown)


data_model/src/lib.rs
```diff
@@ -84,11 +84,11 @@ impl ValidationError {
     DebugCustom, Display, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Serialize, IntoSchema,
 )]
 #[repr(transparent)]
-pub struct Name(String);
+pub struct Name(ConstString);
 
 impl Name {
     pub(crate) const fn empty() -> Self {
-        Self(String::new())
+        Self(ConstString::new())
```
mversic, line 91: if `Name` is immutable, and I think it is?, this can be a shared const value

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+
+    /// Returns `true` if [`ConstString`] data is inlined.
+    #[inline]
+    pub const fn is_inlined(&self) -> bool {
+        self.0.is_inlined()
+    }
+
+    /// Construct empty [`ConstString`].
+    #[inline]
+    pub const fn new() -> Self {
```
mversic, line 47: check out `derive_more::Constructor`

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+}
+
+impl<T: Into<ConstStringData>> From<T> for ConstString {
+    #[inline]
+    fn from(value: T) -> Self {
+        Self(value.into())
+    }
+}
+
+impl Display for ConstString {
```
mversic, line 78: check out `derive_more::Display`

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(f, "{}", &**self)
+    }
+}
+
+impl Debug for ConstString {
```
mversic, line 84: derive or check out `derive_more::Debug`

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+where
+    ConstStringData: AsRef<T>,
+{
+    #[inline]
+    fn as_ref(&self) -> &T {
+        self.0.as_ref()
+    }
+}
+
+impl<T: Into<ConstStringData>> From<T> for ConstString {
```
mversic, line 71: check out `derive_more::From`

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    /// Construct empty [`ConstString`].
+    #[inline]
+    pub const fn new() -> Self {
+        Self(ConstStringData::new())
+    }
+}
+
+impl Deref for ConstString {
```
mversic, line 52: check out `derive_more::Deref`

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        (**self).hash(state)
+    }
+}
+
+impl PartialOrd for ConstString {
+    #[inline]
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        PartialOrd::partial_cmp(&**self, &**other)
```
mversic, line 100:
```suggestion
        Some(self.cmp(other))
```

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+            fn eq(&self, other: &ConstString) -> bool {
+                PartialEq::eq(&self[..], &other[..])
+            }
+        }
+    )*};
+}
+
+impl_eq!(String, str, &str);
+
+impl Eq for ConstString {}
```
mversic, line 143: derive it

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+}
+
+impl IntoSchema for ConstString {
+    fn type_name() -> String {
+        String::type_name()
+    }
+    fn schema(map: &mut MetaMap) {
+        String::schema(map);
+    }
+}
```
mversic, line 179: are `String` and `ConstString` byte representations exactly the same when serialized to Scale or Json?

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+
+/// Union representing const-string variants: inlined or boxed.
+/// Distinction between variants are achieved by setting least significant bit for inlined variant.
+/// Boxed variant should have have 4/3/2 (depending on architecture 64/32/16 bit pointer width) trailing zeros due to pointer alignment.
+#[repr(C)]
```
mversic, line 184: why does it have to be `repr(C)`?

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+union ConstStringData {
+    inlined: InlinedString,
+    boxed: ManuallyDrop<BoxedString>,
+}
+
+impl ConstStringData {
+    #[inline]
+    const fn is_inlined(&self) -> bool {
+        self.inlined().is_inlined()
```
mversic, line 193: this doesn't look correct

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+            Ok(inlined) => Self { inlined },
+            Err(value) => Self {
+                boxed: core::mem::ManuallyDrop::new(value.into()),
+            },
+        }
+    }
+}
+
+impl Clone for ConstStringData {
+    fn clone(&self) -> Self {
```
mversic, line 259: does overriding `clone_from` makes sense?

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+
+impl From<String> for BoxedString {
+    #[inline]
+    fn from(value: String) -> Self {
+        Self(value.into_boxed_str())
+    }
+}
+
+#[derive(Clone, Copy)]
+#[repr(C)]
```
mversic, line 320: does it have to be `repr(C)`?

mversic: what is the alignment of this struct?

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+    fn type_name() -> String {
+        String::type_name()
+    }
+    fn schema(map: &mut MetaMap) {
+        String::schema(map);
+    }
+}
+
+/// Union representing const-string variants: inlined or boxed.
+/// Distinction between variants are achieved by setting least significant bit for inlined variant.
```
appetrosyan, line 182: I'm afraid that won't work. 

Have you checked that non-ASCII characters also work?

mversic (changes requested):

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+/// Boxed variant should have have 4/3/2 (depending on architecture 64/32/16 bit pointer width) trailing zeros due to pointer alignment.
```
outoftardis, line 183:
```suggestion
/// Due to pointer alignment, boxed variant should have 4/3/2 trailing zeros in 64/32/16-bit architecture.
```

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+use std::str::from_boxed_utf8_unchecked;
+
+use iroha_schema::{IntoSchema, MetaMap};
+use parity_scale_codec::{Encode, Output};
+use serde::{Serialize, Serializer};
+
+const MAX_INLINED_STRING_LEN: usize = 2 * size_of::<usize>() - 1;
+
+/// Immutable inlinable string.
+/// Strings shorter than 15/7/3 bytes (depending on architecture 64/32/16 bit pointer width) are inlined.
```
outoftardis, line 22:
```suggestion
/// Strings shorter than 15/7/3 bytes (in 64/32/16-bit architecture) are inlined.
```

data_model/primitives/src/conststr.rs
Erigara: Yes, i checked it, i will add test case for that.
In general, something like [proptest](https://github.com/altsysrq/proptest) would be helpful for testing.

data_model/primitives/src/conststr.rs
Erigara: rust requires all fields to derive `Eq` trait in that case

data_model/src/lib.rs
Erigara: I don't think about that! Thank you for suggestion.
Do we need empty `Id` in the first place?

data_model/primitives/src/conststr.rs
appetrosyan: `derive_more::DebugCustom`

data_model/primitives/src/conststr.rs
appetrosyan: Checking the LSB, so we can use a bit for  a tag, rather than a whole byte?

data_model/primitives/src/conststr.rs
appetrosyan: Re: proptest, It would be good for unit tests, if a bit expensive. We plan on using `hypothesis` from Python downstream. 

Still, there's this is an easy unit test to write and comment.

data_model/src/lib.rs
mversic: Yes, we have to support empty `Name`s at at least for now. There is an issue tracking the discussion on whether we'll allow or disallow empty identifiers

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use core::mem::{align_of, size_of};
+    use std::collections::hash_map::DefaultHasher;
+
+    use super::*;
+
+    fn run_with_strings(f: impl Fn(String)) {
```
appetrosyan, line 407: Like this function, since you can now add more strings to the iterator. I would suggest adding strings that contain UTF-8 characters here and seeing if the tests keep working. Consider the cases of 4-byte characters. many 2-bytes, ASCII all the way except last, odd number of UTF-8, wonky alignment…

data_model/primitives/src/conststr.rs
Erigara: Yes, because in both cases structs are serialized as `str` under the hood.
I will change code to use [`WrapperTypeEncode`](https://docs.rs/parity-scale-codec/2.3.1/parity_scale_codec/trait.WrapperTypeEncode.html) to make it more clear and add test to compare serialization of `ConstString`  and `String`.

data_model/primitives/src/conststr.rs
Erigara: When we check if the `ConstString` is inlined, we check the `len` field of the `InlinedString`, if the string is indeed inlined, then the last bit will be 1 and we will return true, if the string is actually boxed we will get the last byte of the pointer, the last bit of which should be 0 due to alignment in such a case, we return false.

data_model/primitives/src/conststr.rs
Erigara: yes, it has alignment of 1

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,545 @@
...
+        impl PartialEq<$ty> for ConstString {
+            #[allow(clippy::string_slice, clippy::deref_by_slicing)]
+            #[inline]
+            fn eq(&self, other: &$ty) -> bool {
+                PartialEq::eq(&self[..], &other[..])
+            }
+        }
+
+        impl PartialEq<ConstString> for $ty {
+            #[allow(clippy::string_slice, clippy::deref_by_slicing)]
```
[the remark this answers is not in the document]
Erigara: Without it, we cannot write a macro for `str`, `&str` and `String` in a uniform way.

data_model/primitives/src/conststr.rs
Erigara: `derive_more::Constructor` implements `new` without `const` and also uses struct fields as constructor arguments which we don't want in this case.

data_model/primitives/src/conststr.rs
Erigara: Thank you, we can derive `From` and `AsRef` using `#[as_ref/from(forward)]`.

data_model/primitives/src/conststr.rs
mversic: Not a big deal

data_model/primitives/src/conststr.rs
mversic: you're potentially creating a reference to an invalid object when calling inlined. This sounds like an UB

data_model/primitives/src/conststr.rs
mversic: from the looks of it, I don't think you need it for `ConstStringData` or `InlinedString` in the same way as you don't have it for `Box<str>` which I don't think has a defined repr

mversic: ok, InlinedString has to be repr(C), but I don't see why this union would have to be as well. The compiler will add appropriate offsets when casting union into InlinedString and you will be accessing LSB of that offset

I would note that you're accessing bit's directly and those aren't a concept in rust. There are no bitfields as in C. There are some bitfield crates though which would make these bit manipulation more obvious and readable. We might consider using them

mversic: ok, I get it. You are implicitly using tagged pointer when union variant is BoxedString. However, isn't it the case that the MSB, not LSB, is unset in every pointer? If tagged pointers is indeed used, another issue is that I think that the representation of boxed slice Box<str> is undefined because slices don't have a defined representation. If this is the case you should use raw pointers

I would note that you're accessing bit's directly and those aren't a concept in rust. There are no bitfields as in C. There are some bitfield crates though which would make these bit manipulation more obvious and readable. We might consider using them

data_model/primitives/src/conststr.rs
mversic: if this is UB, now this tricky because you're going to have to use raw pointers instead of references

data_model/src/lib.rs
appetrosyan: So the approach has to be one of two. Either we publish this as a separate crate that could be shared with other projects, in which case we need to allow empty values, or we tailor it directly to what our `Name` struct needs (in which case we could equally just merge the two and get rid of another layer of indirection. 

I don't mind either way, but I think that a `ConstString` could be useful in Sora.

data_model/primitives/src/conststr.rs
appetrosyan: So when you `derive` instead of implementing, you're communicating to the reader of the code that there's nothing interesting going on here. That yes, this object does implement this interface, but that there's nothing interesting in how it implements. 

In this particular case the `Deref` is straightforwardly the default. If you `derive(Deref)` you will communicate that there's nothing in there, even though the size of the code may not reduce, you will have a much clearer picture just looking at the struct definition.

data_model/primitives/src/conststr.rs
appetrosyan: If so, could you add a comment?

data_model/primitives/src/conststr.rs
appetrosyan: Strictly speaking there aren't bitfields on any CPU architecture regardless of programming language. The smallest addressable unit is always the byte, and bit-packing operations almost always get implemented with masks and binary logic operators. 

With that said, using the bitfield crates could make your  code more readable, as would better documentation (preferably both).

data_model/primitives/src/conststr.rs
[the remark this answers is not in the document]
appetrosyan: OK. Add these as comments (don't worry, they're stripped off during compilation).

data_model/primitives/src/conststr.rs
Erigara: sure

data_model/primitives/src/conststr.rs
Erigara: Deriving `Deref` requires inner union `ConstStringData` to be public, so i think we should leave it as is.

data_model/primitives/src/conststr.rs
Erigara: > You are implicitly using tagged pointer when union variant is BoxedString

Yeah, `BoxedString` is tagged with zero at the  LSB.

> However, isn't it the case that the MSB, not LSB, is unset in every pointer?

Both options are possible:
- in case of MSB it comes from fact that on x86 we have 48bit logical addressing space;
- in case of LSB it happens because of alignment. 

Useful link: [pointer packing](https://www.1024cores.net/home/lock-free-algorithms/tricks/pointer-packing).

> another issue is that I think that the representation of boxed slice Box<str> is undefined because slices don't have a defined representation. If this is the case you should use raw pointers

You are right, my bad, i assumed layout of `Box<str>` would be pointer and length when debugging and run simple **[test](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a0cb46112d554c70331d6dfd2c4cd802),** i'm going to switch to pointers in `BoxedString` struct, `Box::into_raw` should work (strore `NotNull<u8>` and length).

@appetrosyan, @mversic, I made wrong assumption, i think that if `*mut u8` has minimal alignment of 8 bytes, it means that pointer would have 3 last bits equal to zero, but in fact it means that reference to `*mut u8` would have 3 last bits equal to zero.

Tests passes because `libc::malloc` had minimal alignment of 16 bytes on `x86_64` ([MALLOC_ALIGNMENT](https://sourceware.org/glibc/wiki/MallocInternals).)

Windows `HeapAlloc` have the same alignment ([MEMORY_ALLOCATION_ALIGNMENT](https://docs.microsoft.com/en-us/windows/win32/api/heapapi/nf-heapapi-heapalloc)).

However i'm not sure about `wee_alloc`.

I have few options:
- Leave as is and document.
- Rely on MSB to be equal to 0;
- Alloc memory for `BoxedString` with at least `u16` alignment, so it would have LSB guaranteed to be 0 (most reliable).

appetrosyan: Could you try it with jemalloc? We might need to statically link an allocator, becauseof musl, so this might be a non-issue.

data_model/primitives/src/conststr.rs
Erigara: @appetrosyan, you mean create `ConstString` from `&str`?

data_model/primitives/src/conststr.rs
Erigara: I like your idea of storing the tag in the MSB of len, because with this approach we store the tag in the same way for the Boxed and Inlined variants.

Erigara: > Could you try it with jemalloc? We might need to statically link an allocator, becauseof musl, so this might be a non-issue.

It works with jemmalloc just fine.

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
+//! Const-string related implementation and structs.
+extern crate alloc;
```
mversic, line 2: I don't think you need this, it's inclded in the `data_model/lib.rs`

data_model/primitives/src/conststr.rs
appetrosyan: Good. We can add support for other allocators later if we need to.

data_model/primitives/src/conststr.rs
appetrosyan: > So I'm not sure if we need to implement this.

Rule of thumb. If we don't need it yet, and you're not sure we will, don't implement.

appetrosyan (approved):
Good job!

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+        }
+    )*};
+}
+
+impl_eq!(String, str, &str);
+
+/// Can't be derived.
+impl Eq for ConstString {}
+
+impl Serialize for ConstString {
```
mversic, line 128: how come there is no deserialze implementation?

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+const MAX_INLINED_STRING_LEN: usize = 2 * size_of::<usize>() - 1;
+
+/// Immutable inlinable string.
+/// Strings shorter than 15/7/3 bytes (in 64/32/16-bit architecture) are inlined.
+#[derive(Clone, Display, DebugCustom, From, AsRef)]
+#[from(forward)]
+#[as_ref(forward)]
+#[display(fmt = "{}", "&**self")]
+#[debug(fmt = "{:?}", "&**self")]
+pub struct ConstString(ConstStringData);
```
mversic, line 32: is this wrapper struct necessary? Wouldn't it be possible to implement everything directly on `ConstStringData`?

data_model/primitives/src/conststr.rs
appetrosyan: This is supposed to be a separate crate, so it might not be. @Erigara please check

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+unsafe impl Sync for BoxedString {}
+
+#[derive(Clone, Copy)]
+#[repr(C)]
+struct InlinedString {
+    #[cfg(target_endian = "little")]
+    payload: [u8; MAX_INLINED_STRING_LEN],
+    /// MSB is always 1 to distinguish inlined variant.  
+    len: u8,
+    #[cfg(target_endian = "big")]
```
mversic, line 394:
```suggestion
    payload: [u8; MAX_INLINED_STRING_LEN],
    /// MSB is always 1 to distinguish inlined variant.  
    len: u8,
```

I don't think we need to watch the endianess, just have the layouts of `InlinedString` and `BoxedString` be compatible

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+}
+
+impl InlinedString {
+    #[inline]
+    const fn len(self) -> usize {
+        (self.len - 128) as usize
+    }
+
+    #[inline]
+    const fn is_inlined(self) -> bool {
```
mversic, line 405: this should just return true or not exist at all.

The fact that you first cast into an `InlinedString` and then check for the first bit of len is a code smell to me because the underlying bit pattern might not be a valid string. Considering that the string is backed by byte slice all bit patterns might be allowed even though they may not be a valid string. 

There is a [difference](https://internals.rust-lang.org/t/two-kinds-of-invariants-safety-and-validity/8264/23) between unsafety and validity and I'm not sure. Try to find out what's appropriate here. You could cast the `ConstStringData` to a byte array and read the byte instead

mversic (changes requested):

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+        Self { ptr, len }
+    }
+}
+
+impl From<String> for BoxedString {
+    #[inline]
+    fn from(value: String) -> Self {
+        let len = value.len();
+        let boxed = value.into_boxed_str();
+        let ptr = Box::into_raw(boxed).cast::<u8>();
```
mversic, line 358:
```suggestion
        let boxed = value.into_bytes().into_boxed_slice();
        let ptr = Box::into_raw(boxed);
```

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+    }
+}
+
+impl From<&str> for BoxedString {
+    #[allow(unsafe_code)]
+    #[inline]
+    fn from(value: &str) -> Self {
+        let len = value.len();
+        let boxed = value.as_bytes().to_vec().into_boxed_slice();
+        let ptr = Box::into_raw(boxed).cast::<u8>();
```
mversic, line 348: the same, this cast is most likely a UB. I've checked the size of this pointer and it's a fat pointer you are casting to a raw pointer

data_model/primitives/src/conststr.rs
appetrosyan: I think that the endianness check could be useful. I'd keep it until it's proven safe.

data_model/primitives/src/conststr.rs
Erigara: @mversic, I get errors when deleting this line.

data_model/primitives/src/conststr.rs
Erigara: `Name` struct doesn't have `Deserialize` and `Decode`, so i decided not to implement them for now.
I can implement then if needed.

mversic: It has them implemented manually, no?

data_model/primitives/src/conststr.rs
Erigara: I run tests with `miri` and it doesn't find any UB (i understand that `miri` can't find all UBs).
Also despite the fact that `InlinedString` could be in invalid state from logical point of view (wrong `len`, non-utf8 characters), it's seems to be correct from data point of view (one byte field and one array of bytes).

Also i don't understand how switching to pointers could help in that case.

mversic: hm, this has me worried because I'm quite sure that fat to slim pointer casts that you've used are UB. How come miri didn't catch on that?

just to give you a feel of the problem, if you transmute `u8` (with a value other than 0 or 1) to `bool`, you're stepping into instant UB. This is because `bool` is a so-called non-robust type and has at least one trap representation. This may or may not be the case for strings, I'm not quite sure, I'll be looking it up some more

data_model/primitives/src/conststr.rs
Erigara: Actually i think endianness is important here, because it does not affect structs offsets.

Little-endian:

```text
+---------+-------+---------+----------+----------------+
| Bits    | 0..63 | 64..118 | 119..126 | 127            |
+---------+-------+---------+----------+----------------+
| Inlined | payload         | len      | tag (MSB)      |
+---------+-------+---------+----------+----------------+
| Box     | ptr   | len                | tag (MSB)      |
+---------+-------+--------------------+----------------+
```

Big-endian without changing layout (order of bytes for `BoxedString::len` would be reversed):

```text
+---------+-------+----------+-----------+---------+----------+-----------+
| Bits    | 0..63 | 64..70   | 71        | 72..118 | 119..126 | 127       |
+---------+-------+----------+-----------+---------+----------+-----------+
| Inlined | payload                                | len      | tag (MSB) |
+---------+-------+----------+-----------+---------+----------+-----------+
| Box     | ptr   | len...   | tag (MSB) | ...len                         |
+---------+-------+----------+-----------+--------------------------------+
```

data_model/primitives/src/conststr.rs
Erigara: oh, i missed them...
I will add `Decode` and `Deserialize` for `ConstString`!

data_model/primitives/src/conststr.rs
mversic: in fact if you try to use `transmute` the compiler will not allow it

data_model/primitives/src/conststr.rs
mversic: you have to be thinking about what is allowed in the type system. Let's have a look at [rust reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#dangling-pointers) or [nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html) for a list of UBs. These things are UBs because the compiler is allowed to assume these types will never have invalid representations and make optimizations based on that assumption. 

Even if you're not accessing the string in your code, the compiler is free to inject operations that read from it as long as it produces output semantically equivalent to your source code. This is at least one reason why conversions like `12_u8` -> `bool` or `null_ptr` -> `&T` are instant UB, not when they are used/accessed in the source code. 

Another example is storing null pointer into `NonNull<T>` even though it is backed by `*mut T` which can be a null pointer. This is quite a similar situation to what you have in your code

How interleaved `std::String` is with the compiler is hard to say, but I'd be very cautious about producing invalid String representations, otherwise `from_utf8_unchecked` wouldn't be marked unsafe.

data_model/primitives/src/conststr.rs
mversic: very good, you're correct, I forgot one is 4bytes and the other just 1byte

mversic: sorry, my bad, I'll resolve this conversation

data_model/primitives/src/conststr.rs
Erigara: @mversic, i'am not sure that is UB, in `std` we have [`as_ref(self) -> const *T`](https://doc.rust-lang.org/src/core/ptr/const_ptr.rs.html#1087) implemented for `*const [T]`, inside it is just `as` cast.

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,670 @@
...
+}
+
+impl Drop for BoxedString {
+    #[allow(unsafe_code)]
+    fn drop(&mut self) {
+        // SAFETY:
+        // - using global allocator
+        // - `BoxedString`s have the same layout varied only by it's length
+        unsafe {
+            alloc::alloc::dealloc(self.ptr as *mut u8, Self::layout(self.len));
```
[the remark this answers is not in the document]
Erigara: i like this style
maybe implement `BoxedString::clone` in the same way?

```rust
fn clone(&self) -> Self {
    let len = self.len;
    let boxed = unsafe {
        Box::<[_]>::from_raw(from_raw_parts_mut(self.ptr, self.len))
    };
    let clone = boxed.clone();
    let ptr = Box::into_raw(clone).cast::<u8>();
    forget(boxed);
    Self { ptr, len }
}
```

data_model/primitives/src/conststr.rs
mversic: let's consider this a little bit more before making a decision

data_model/primitives/src/conststr.rs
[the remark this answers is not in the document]
mversic: almost, but you don't need to take ownership:
```rs
fn clone(&self) -> Self {
    let len = self.len;
    let boxed = unsafe { from_raw_parts(self.ptr, self.len)) };
    let clone = boxed.clone();
    let ptr = Box::into_raw(clone).cast::<u8>();
    forget(boxed);
    Self { ptr, len }
}
```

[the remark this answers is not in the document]
Erigara: hmm, `from_raw_parts` will return `&[u8]` not `Box<[u8]>`, or i missed something?

data_model/primitives/src/conststr.rs
mversic: You shouldn't be getting any errors. At least add `#[cfg(not(feature = "std"))]`

If this'll be a separate crate there should be no `std` feature in it

data_model/primitives/src/conststr.rs
mversic: looking at [this PR](https://github.com/rust-lang/reference/pull/792/files) it may be the case that non-UTF strings don't cause UB

data_model/primitives/src/conststr.rs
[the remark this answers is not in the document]
mversic: yes, and `Clone::clone` takes a reference, or you can use `to_owned` on a slice as well

data_model/primitives/src/conststr.rs
mversic: I'd tentatively consider this a non-issue as far as I can conclude after researching into the matter

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+        String::schema(map);
+    }
+}
+
+#[derive(DebugCustom)]
+#[debug(fmt = "{:?}", "&**self")]
+#[repr(C)]
+struct BoxedString {
+    #[cfg(target_endian = "little")]
+    ptr: *mut u8,
```
mversic, line 277: use NonNull

data_model/primitives/src/conststr.rs
mversic: has the situation changed now?

data_model/src/lib.rs
```diff
@@ -110,20 +112,14 @@ impl Name {
...
 
-impl FromStr for Name {
-    type Err = ParseError;
-
-    fn from_str(candidate: &str) -> Result<Self, Self::Err> {
+    /// Checks if `candidate` string would be valid [`Name`].
+    ///
+    /// # Errors
+    /// Fails if not valid [`Name`].
+    pub fn validate_str(candidate: &str) -> Result<(), ParseError> {
```
mversic, line 120:
```suggestion
    fn validate_str(candidate: &str) -> Result<(), ParseError> {
```

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+impl AsRef<str> for InlinedString {
+    #[allow(unsafe_code)]
+    #[inline]
+    fn as_ref(&self) -> &str {
+        // SAFETY: created from valid utf-8.
+        unsafe { core::str::from_utf8_unchecked(&self.payload[..self.len()]) }
+    }
+}
+
+impl<'value> TryFrom<&'value str> for InlinedString {
```
mversic, line 408: I would usually implement `FromStr` instead

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+impl Clone for BoxedString {
+    /// Properly clone `BoxedString` into new allocation.
+    fn clone(&self) -> Self {
+        let ptr = Box::into_raw(self.as_bytes().to_owned().into_boxed_slice()).cast::<u8>();
+        let len = self.len;
+        Self { ptr, len }
+    }
+}
+
+impl From<&str> for BoxedString {
```
mversic, line 324: implement `FromStr` instead :)

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+    #[allow(unsafe_code)]
+    #[inline]
+    const fn inlined(&self) -> &InlinedString {
+        // SAFETY: safe to access if `is_inlined` == `true`.
+        unsafe { &self.inlined }
+    }
+
+    #[allow(unsafe_code)]
+    #[inline]
+    const fn boxed(&self) -> &ManuallyDrop<BoxedString> {
```
mversic, line 93:
```suggestion
    const fn boxed(&self) -> &BoxedString {
```

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+    }
+    fn schema(map: &mut MetaMap) {
+        String::schema(map);
+    }
+}
+
+#[derive(DebugCustom)]
+#[debug(fmt = "{:?}", "&**self")]
+#[repr(C)]
+struct BoxedString {
```
mversic, line 275: you should implement `Drop` for this structure

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+where
+    T: TryInto<InlinedString>,
+    <T as TryInto<InlinedString>>::Error: Into<BoxedString>,
+{
+    #[inline]
+    fn from(value: T) -> Self {
+        match value.try_into() {
+            Ok(inlined) => Self { inlined },
+            Err(value) => Self {
+                boxed: ManuallyDrop::new(value.into()),
```
mversic, line 192: I'm sensing a potential leak here. You're giving out to the user of the library an object which they should manually drop. That is not a good api

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+impl Deref for ConstString {
+    type Target = str;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        self.as_ref()
+    }
+}
+
+impl Hash for ConstString {
```
mversic, line 123: I would also consider implementing `Borrow`

data_model/primitives/src/conststr.rs
appetrosyan: Technically, this is an infallible conversion, so we should have both.

data_model/src/lib.rs
appetrosyan: I'd recommend keeping some version of this function public.

data_model/src/lib.rs
```diff
@@ -84,11 +84,13 @@ impl ValidationError {
     DebugCustom, Display, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Serialize, IntoSchema,
 )]
 #[repr(transparent)]
-pub struct Name(String);
+pub struct Name(ConstString);
 
 impl Name {
+    const EMPTY: Self = Self(ConstString::new());
+
     pub(crate) const fn empty() -> Self {
```
mversic, line 92: it's a const function, no need for `const EMPTY`

data_model/primitives/src/conststr.rs
Erigara: @mversic, i think `FromStr` exist in parsing contexts and also have notion of failure.
In our case it more of the conversion and with `From` + `TryForm` it's possible to create generic implementation `From` for `ConstString`.
Also creation of `BoxedString` from string can't fail.
However I could implement it in the same way as `String` [did](https://doc.rust-lang.org/src/alloc/string.rs.html#2351).

mversic: ok, I missed that it's infallible, in which case there shouldn't be an implementation of `From<&str>`. Rather one should turn `str` into `String` and call `From<String>`

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+    /// Construct empty [`ConstString`].
+    #[inline]
+    pub const fn new() -> Self {
+        Self {
+            inlined: InlinedString::new(),
+        }
+    }
+
+    #[inline]
+    const fn is_inlined(&self) -> bool {
```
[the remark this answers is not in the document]
appetrosyan: A pure const associated function is not the kind of thing that should be hidden.

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+#[display(fmt = "{}", "&**self")]
+#[debug(fmt = "{:?}", "&**self")]
+#[repr(C)]
+pub union ConstString {
+    inlined: InlinedString,
+    boxed: ManuallyDrop<BoxedString>,
+}
+
+impl ConstString {
+    /// Returns the length of this [`ConstString`], in bytes.
```
appetrosyan, line 54: In general, function docstrings should be written in imperative mood. I.e. 
```suggestion
    /// Return the length of [`Self`], in bytes.
```

appetrosyan: Using `Self` instead of `ConstString` is a good convention, it also saves you from having to check if you renamed the doc-comments.

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+            Box::<[_]>::from_raw(from_raw_parts_mut(self.ptr, self.len));
+        }
+    }
+}
+
+/// `BoxedString` is `Send` because the data they
+/// reference is unaliased. Aliasing invariant is enforced by
+/// creation of `BoxedString`.
+#[allow(unsafe_code)]
+unsafe impl Send for BoxedString {}
```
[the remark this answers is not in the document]
Erigara: If [`core::ptr::Unique`](https://docs.rs/rust-libcore/latest/core/ptr/struct.Unique.html) (which `Box` uses under the hood) was stable feature `BoxedString` could use them and will be automatically `Send` and `Sync`, but `Sync` and `Send` is not inplemented for pointers (and also `NonNull`), so we must implement them manually.
We can do that because every `BoxedString` instance has own unique pointer (there is no two `BoxedString`s pointing) to the same memory. 

Or what you mean by? 
> does this mean ConstString is Send and Sync transitively?

data_model/primitives/src/conststr.rs
Erigara: ok, but it won't be a `const` function anymore

data_model/src/lib.rs
Erigara: oh, okay, it was `const` function from the begging.

data_model/primitives/src/conststr.rs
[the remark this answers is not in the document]
appetrosyan: Are you going to remove the implementations once they are automatic, based off of the feature that you linked?

data_model/primitives/src/conststr.rs
Erigara: not realy, beacuse we still can't derive `Deref` implementation and need to do it manually

data_model/primitives/src/conststr.rs
Erigara: Std has [`impl<'_> From<&'_ str> for String`](https://doc.rust-lang.org/std/convert/trait.From.html#impl-From%3C%26%27_%20str%3E-3) , so i think it would be suitable to implement it for `BoxedString` (and `ConstString`) too. 

I we want to call `parse` we can implement `FromStr` on the top level for the `ConstString`.

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+
+/// `BoxedString` is `Send` because the data they
```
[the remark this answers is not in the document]
Erigara: ok, copy-paste error

data_model/src/lib.rs
Erigara: you suggested introduce shared const value for empty names

data_model/primitives/src/conststr.rs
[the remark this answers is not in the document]
mversic: my question was is `ConstString` both `Send` and `Sync`? Not about `BoxedString`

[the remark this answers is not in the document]
mversic: I think `ConsString` is both `Send` and `Sync`. Verify and resolve this

data_model/src/lib.rs
mversic: I remember suggesting that somewhere. If it was here, it was a bad suggestion

data_model/primitives/src/conststr.rs
Erigara: I think `TryFrom` would be better fit here.

Comparing descriptions of this traits.

[`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html):
> Parse a value from a string

[`TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html):
> Simple and safe type conversions that may fail in a controlled way under some circumstances. It is the reciprocal of [TryInto](https://doc.rust-lang.org/stable/std/convert/trait.TryInto.html).
> This is useful when you are doing a type conversion that may trivially succeed but may also need special handling. For example, there is no way to convert an [i64](https://doc.rust-lang.org/stable/std/primitive.i64.html) into an [i32](https://doc.rust-lang.org/stable/std/primitive.i32.html) using the [From](https://doc.rust-lang.org/stable/std/convert/trait.From.html) trait, because an [i64](https://doc.rust-lang.org/stable/std/primitive.i64.html) may contain a value that an [i32](https://doc.rust-lang.org/stable/std/primitive.i32.html) cannot represent and so the conversion would lose data. This might be handled by truncating the [i64](https://doc.rust-lang.org/stable/std/primitive.i64.html) to an [i32](https://doc.rust-lang.org/stable/std/primitive.i32.html) (essentially giving the [i64](https://doc.rust-lang.org/stable/std/primitive.i64.html)’s value modulo [i32::MAX](https://doc.rust-lang.org/stable/std/primitive.i32.html#associatedconstant.MAX)) or by simply returning [i32::MAX](https://doc.rust-lang.org/stable/std/primitive.i32.html#associatedconstant.MAX), or by some other method. The [From](https://doc.rust-lang.org/stable/std/convert/trait.From.html) trait is intended for perfect conversions, so the TryFrom trait informs the programmer when a type conversion could go bad and lets them decide how to handle it.

I think that the example with the `i64 -> i32` is indicative of our situation, we want to create `InlinedString` from `String` or `&str` but we can't due to exceeding length, so we have controlled fallback option to create `BoxedString`.

data_model/src/lib.rs
Erigara: ok, i will return to initial implementation)

data_model/primitives/src/conststr.rs
mversic: I just dislike implicit allocations

data_model/primitives/src/conststr.rs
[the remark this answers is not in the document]
Erigara: Both `BoxedString` and `InlinedString` are `Send` and `Sync` so yes `ConstString` is also `Send` and `Sync` transitively.

data_model/primitives/src/conststr.rs
Erigara: ok, resolve then

mversic (approved):

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,664 @@
...
+}
+
+/// `BoxedString` is `Send` because the data they
+/// reference is unaliased. Aliasing invariant is enforced by
+/// creation of `BoxedString`.
+#[allow(unsafe_code)]
+unsafe impl Send for BoxedString {}
+
+/// `BoxedString` is `Sync` because the data they
+/// reference is unaliased. Aliasing invariant is enforced by
```
Arjentix, line 376:
```suggestion
/// [`BoxedString`] is [`Sync`] because the data it
/// references is unaliased. Aliasing invariant is enforced by
```

appetrosyan (approved):

data_model/primitives/src/conststr.rs
```diff
@@ -0,0 +1,650 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        PartialEq::eq(&**self, &**other)
+    }
+}
+
+macro_rules! impl_eq {
+    ($($ty:ty,)*) => {
+        impl_eq!($($ty),*);
+    };
+    ($($ty:ty),*) => {$(
```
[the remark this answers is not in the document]
Erigara: Not really because with such approach 'impl_eq!(,)' would be valid macro invocation.

Issue #1658, opened 2021-12-01
Add documentation check to CI


appetrosyan:
`cargo doc -q` should produce a list of warnings that should be included in the `I2::Dev::StaticAnalysis` workflow and fail the workflow if any are left.

appetrosyan:
Could be done via a `bash` script.

PR #2367, opened 2022-06-16
[ci] #1658: Add documentation check

ales-tsurko:
Sorry for omitting the PR template here, but the change is very little and clear. It just adds  documentation check (using `cargo doc`) to CI.


pesterev (changes requested):
According to the CI report, a few comments need to be corrected to pass the CI checks.

ffi/src/lib.rs
```diff
@@ -30,7 +30,7 @@ pub enum FfiResult {
     Ok = 0_i32,
 }
 
-/// Implement [`Handle`] for given types with first argument as the initial handle id.
```
outoftardis, line 33: @ales-tsurko this is the only link that is still unresolved.
I think my knowledge of Rust is limited here to understand why all the syntax I'm trying is failing, so if you have any ideas how to fix this -- please do

ales-tsurko: The problem caused by the `Handle` trait defined inside macro, so there's no trait definition until the macro is used. We have two options: 
1. define the trait outside the macro;
2. don't link documentation for the trait.

@appetrosyan what do you think?

outoftardis: maybe we can omit the link altogether? the description would still be clear even without the link

appetrosyan: > don't link documentation for the trait.

Is fine by me. We should avoid defining non-private traits inside macros, though.

outoftardis (approved):

PR #2378, opened 2022-06-20
[feature] #2127: Add sanity check to ensure that all data decoded by `parity_scale_codec` is consumed

### LINKED ISSUE #2127: Add sanity check to ensure that all data decoded by `parity_scale_codec` are consumed
```Marin, [20.04.22 12:53]
I don't think this is an issue at all

Marin, [20.04.22 12:54]
the byte representations match for VersionedQueryResult and paginated version

Marin, [20.04.22 12:54]
to a point

Marin, [20.04.22 12:54]
if it was the other way around, it would fail. If you were encoding QueryResult and decoding PaginatedQueryResult

Marin, [20.04.22 12:55]
if there was something after if in the encoded form, it would fail as well

Aleksandr Petrosyan, [20.04.22 12:55]
[Forwarded from Aleksandr Petrosyan]
Thankfully it doesn't decode QueryResult, but the fact that it happily works with types containing VersionedQueryResult is a little worrying

Marin, [20.04.22 12:55]
but seems that scale happily disregards trailing bytes if it finished decoding an object

Aleksandr Petrosyan, [20.04.22 12:56]
[In reply to Marin]
It works more like a parser than a binary serialisation format.

Marin, [20.04.22 12:56]
we can inject a check I guess

Marin, [20.04.22 12:56]
to see if all bytes were consumed

Marin, [20.04.22 12:57]
check that parity_scale_codec::Input is depleted

Aleksandr Petrosyan, [20.04.22 12:58]
Letting unrecognised data into a blockchain is a ticking time-bomb.
```

Arjentix:
### Description of the Change

* Added `decode_all_versioned()` that works like  [`decode_all()`](https://docs.rs/parity-scale-codec/2.3.1/parity_scale_codec/trait.DecodeAll.html#tymethod.decode_all) from `parity_scale_codec`
* `decode_all()` and `decode_all_versioned()` were used in some places where `decode()` and `decode_versioned()` were used before. It wasn't replaced in all places, cause we have to do it only on high decoding level (i.e. receiving message from client, reading *kura* block and etc.)
* To not to introduce new errors for *SDK*-developres all usage of `_all*`-functions mentioned above check for error, print warning and call for not-`all` analogue

### Issue

* Closes #2127

### Benefits

* Checked that Iroha peers and client communicate without extra bytes in messages
* *SDK*-developers will get warning message if they send extra-bytes. In future we can return error

### Possible Drawbacks

* In the future there can be another place there decoding will be done. If someone will forget to use `_all*` version of function then it's only possible to track it on review. No code checks we can done with this approach
* I may missed some places where `decode()` (or `decode_versioned()`) should be replaced with `decode_all()` (or `decode_all_versioned()`)

### Usage Examples or Tests

I've inserted `panic!` in `decode_all_versioned()` impl in `impl_decode_versioned()` and in other places where just `parity_scale_codec::decode()` was used. After that I've ran our tests. Nothing has failed. Seems like our internal communication works fine 

### Alternate Designs

At first I was trying to implement my own procedural *derive* macro so that `decode` implementation would check if all bytes were consumed. I was thinking that it's better to make this check permanent and undependent from external usage. So that there will be no way to make a mistake in the future.
But another problem occurred: it only works for high level structures. If some structure can be decoded by it's own and at the same time as a part of another strucutre -- this approach fails. So I decided not no use this macro and find another way. That's why there is `2` in my branch name.

### DIFF
--- a/cli/src/torii/utils.rs
+++ b/cli/src/torii/utils.rs
@@ -80,8 +80,8 @@ pub mod body {
     {
         warp::body::bytes().and_then(|body: Bytes| async move {
             let mut res = DecodeVersioned::decode_all_versioned(body.as_ref());
-            if res.is_err() {
-                warn!("Can't decode body using all bytes");
+            if let Err(iroha_version::error::Error::ExtraBytesLeft(left)) = res {
+                warn!(left_bytes = %left, "Can't decode body, not all bytes were consumed");
                 res = DecodeVersioned::decode_versioned(body.as_ref());
             }
             res.map_err(warp::reject::custom)
--- a/cli/src/torii/utils.rs
+++ b/cli/src/torii/utils.rs
@@ -1,7 +1,7 @@
 use std::convert::Infallible;
 
 use iroha_cli_derive::generate_endpoints;
-use iroha_version::scale::DecodeVersioned;
+use iroha_version::prelude::*;
 use parity_scale_codec::Encode;
 use warp::{hyper::body::Bytes, reply::Response, Filter, Rejection, Reply};
 
@@ -58,12 +58,8 @@ pub mod body {
         type Error = WarpQueryError;
 
         fn try_from(body: &Bytes) -> Result<Self, Self::Error> {
-            let mut query_res = VersionedSignedQueryRequest::decode_all_versioned(body.as_ref());
-            if query_res.is_err() {
-                warn!("Can't decode query using all bytes");
-                query_res = VersionedSignedQueryRequest::decode_versioned(body.as_ref());
-            }
-            let query = query_res.map_err(|e| WarpQueryError(Box::new(e).into()))?;
+            let res = try_decode_all_or_just_decode!(VersionedSignedQueryRequest, body.as_ref());
+            let query = res.map_err(|e| WarpQueryError(Box::new(e).into()))?;
             let VersionedSignedQueryRequest::V1(query) = query;
             Ok(Self::try_from(query)?)
         }
@@ -79,12 +75,7 @@ pub mod body {
     pub fn versioned<T: DecodeVersioned>() -> impl Filter<Extract = (T,), Error = Rejection> + Copy
     {
         warp::body::bytes().and_then(|body: Bytes| async move {
-            let mut res = DecodeVersioned::decode_all_versioned(body.as_ref());
-            if let Err(iroha_version::error::Error::ExtraBytesLeft(left)) = res {
-                warn!(left_bytes = %left, "Can't decode body, not all bytes were consumed");
-                res = DecodeVersioned::decode_versioned(body.as_ref());
-            }
-            res.map_err(warp::reject::custom)
+            try_decode_all_or_just_decode!(T as "Body", body.as_ref()).map_err(warp::reject::custom)
         })
     }
 }
--- a/cli/src/torii/utils.rs
+++ b/cli/src/torii/utils.rs
@@ -39,6 +39,7 @@ macro_rules! add_state {
 pub mod body {
     use iroha_core::smartcontracts::query::Error as QueryError;
     use iroha_data_model::query::VersionedSignedQueryRequest;
+    use iroha_logger::warn;
 
     use super::*;
 
@@ -57,8 +58,12 @@ pub mod body {
         type Error = WarpQueryError;
 
         fn try_from(body: &Bytes) -> Result<Self, Self::Error> {
-            let query = VersionedSignedQueryRequest::decode_versioned(body.as_ref())
-                .map_err(|e| WarpQueryError(Box::new(e).into()))?;
+            let mut query_res = VersionedSignedQueryRequest::decode_all_versioned(body.as_ref());
+            if query_res.is_err() {
+                warn!("Can't decode query using all bytes");
+                query_res = VersionedSignedQueryRequest::decode_versioned(body.as_ref());
+            }
+            let query = query_res.map_err(|e| WarpQueryError(Box::new(e).into()))?;
             let VersionedSignedQueryRequest::V1(query) = query;
             Ok(Self::try_from(query)?)
         }
@@ -74,7 +79,12 @@ pub mod body {
     pub fn versioned<T: DecodeVersioned>() -> impl Filter<Extract = (T,), Error = Rejection> + Copy
     {
         warp::body::bytes().and_then(|body: Bytes| async move {
-            DecodeVersioned::decode_versioned(body.as_ref()).map_err(warp::reject::custom)
+            let mut res = DecodeVersioned::decode_all_versioned(body.as_ref());
+            if res.is_err() {
+                warn!("Can't decode body using all bytes");
+                res = DecodeVersioned::decode_versioned(body.as_ref());
+            }
+            res.map_err(warp::reject::custom)
         })
     }
 }
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -65,12 +65,12 @@ where
             match resp.status() {
                 StatusCode::OK => {
                     let mut res = VersionedPaginatedQueryResult::decode_all_versioned(resp.body());
-                    if res.is_err() {
-                        warn!("Can't decode query result using all bytes");
+                    if let Err(iroha_version::error::Error::ExtraBytesLeft(left)) = res {
+                        warn!(left_bytes = %left, "Can't decode query result, not all bytes were consumed");
                         res = VersionedPaginatedQueryResult::decode_versioned(resp.body());
                     }
                     res.wrap_err(
-                        "Failed to decode all response body as VersionedPaginatedQueryResult",
+                        "Failed to decode the whole response body as `VersionedPaginatedQueryResult`",
                     )
                     .map_err(Into::into)
                 }
@@ -80,10 +80,11 @@ where
                 | StatusCode::NOT_FOUND => {
                     let mut res = QueryError::decode_all(resp.body().as_ref());
                     if res.is_err() {
-                        warn!("Can't decode query error using all bytes");
+                        warn!("Can't decode query error, not all bytes were consumed");
                         res = QueryError::decode(&mut resp.body().as_ref());
                     }
-                    let err = res.wrap_err("Failed to decode all response body as QueryError")?;
+                    let err =
+                        res.wrap_err("Failed to decode the whole response body as `QueryError`")?;
                     Err(ClientQueryError::QueryError(err))
                 }
                 _ => Err(ResponseReport::with_msg("Unexpected query response", resp).into()),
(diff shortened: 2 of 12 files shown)


version/src/lib.rs
```diff
@@ -198,6 +201,12 @@ pub mod scale {
         /// # Errors
         /// Will return error if version is unsupported or if input won't have enough bytes for decoding.
         fn decode_versioned(input: &[u8]) -> Result<Self>;
```
mversic, line 203: why don't we remove this now?

version/src/lib.rs
```diff
@@ -53,7 +53,9 @@ pub mod error {
         /// Problem with parsing integers
         ParseInt,
         /// Input version unsupported
-        UnsupportedVersion(UnsupportedVersion),
+        UnsupportedVersion(Box<UnsupportedVersion>),
```
mversic, line 56: I'd like to hear what others think about boxing this. This variant has the size of 25 bytes while `ExtraBytesLeft` has the size of 8 bytes. This reduces the size by 17 bytes, but it's not as if there will be a lot of instances of this error

version/src/lib.rs
mversic: Why don't we use `decode_all` instead of `decode` everywhere?

version/derive/src/lib.rs
```diff
@@ -265,6 +301,8 @@ fn impl_decode_versioned(enum_name: &Ident) -> proc_macro2::TokenStream {
 }
 
 fn impl_json(enum_name: &Ident, version_field_name: &str) -> proc_macro2::TokenStream {
+    let box_path = get_box_path();
```
mversic, line 304: this is not good, because the macro expands at the call site. It works accidentally because you've enabled `no_std` for both `data_model` and `version` at the same time

mversic: you need `Box` in the path of the caller

version/src/lib.rs
Arjentix: Because it's impossible to use `decode_all()` if the structure is nested in the other type. There will be extra bytes for the next thing in the containing structure, but our type will give an error just because there are some bytes left. We can use `decode_all()` only for top-most structures

version/derive/src/lib.rs
mversic: ah, I guess you can do it like this

version/src/lib.rs
Arjentix: Yeah, I've done it just because clippy suggested. Can be removed easily by reverting some commits

p2p/src/peer.rs
```diff
@@ -546,7 +546,11 @@ where
                     }
                 }
             };
-            let decoded: Result<T, _> = Decode::decode(&mut data.as_slice());
+            let mut decoded: Result<T, _> = DecodeAll::decode_all(data.as_slice());
+            if decoded.is_err() {
```
mversic, line 550: can we check for a specific error or error message?

mversic: Why don't we just agree that buffers must be depleted? I would just return the error even if there are more bytes in the input. This seems like unnecessary complexity

version/src/lib.rs
```diff
@@ -198,6 +201,12 @@ pub mod scale {
         /// # Errors
         /// Will return error if version is unsupported or if input won't have enough bytes for decoding.
         fn decode_versioned(input: &[u8]) -> Result<Self>;
+
+        /// Use this function for versioned objects instead of `decode_all`.
+        ///
+        /// # Errors
+        /// Will return error if version is unsupported or if input won't have enough bytes for decoding.
```
mversic, line 208: or if it has too much bytes

p2p/src/peer.rs
appetrosyan: We need to deprecate the older approach, and our tests aren't the only thing that could break. I suggested to @Arjentix to leave this behaviour for the time being, so that we can find this in the Java/JavaScript SDKs.

mversic: I see, can we then open an issue about it?

client/src/client.rs
```diff
@@ -715,8 +728,12 @@ impl Client {
             .send()?;
 
             if response.status() == StatusCode::OK {
-                let pending_transactions =
-                    VersionedPendingTransactions::decode_versioned(response.body())?;
+                let mut res = VersionedPendingTransactions::decode_all_versioned(response.body());
+                if res.is_err() {
+                    warn!("Can't decode pending transactions using all bytes");
```
appetrosyan, line 733:
```suggestion
                    warn!("Failed to decode pending transactions, not all bytes were consumed");
```
Which should be the only reason for failure.

client/src/client.rs
```diff
@@ -919,6 +936,15 @@ pub mod events_api {
             }
         }
 
+        fn decode_publisher_message(bytes: &[u8]) -> Result<VersionedEventPublisherMessage> {
+            let mut res = VersionedEventPublisherMessage::decode_all_versioned(bytes);
+            if res.is_err() {
+                warn!("Can't decode event publisher message using all bytes");
```
appetrosyan, line 942:
```suggestion
                warn!("Can't decode event publisher message. Not all bytes were consumed");
```

core/src/kura.rs
```diff
@@ -404,7 +404,13 @@ impl<IO: DiskIO> BlockStore<IO> {
         #[allow(clippy::cast_possible_truncation)]
         buffer.resize(len as usize, 0);
         let _len = file_stream.read_exact(&mut buffer).await?;
-        Ok(Some(VersionedCommittedBlock::decode_versioned(&buffer)?))
+
+        let mut res = VersionedCommittedBlock::decode_all_versioned(&buffer);
+        if res.is_err() {
+            warn!("Can't decode block using all bytes");
```
appetrosyan, line 410:
```suggestion
            warn!("Can't decode block. Not all bytes were consumed");
```

p2p/src/peer.rs
appetrosyan: Be my guest.

version/src/lib.rs
appetrosyan: There should be no ExtraBytesLeft instances returned long-term.

Either leaving the box in, or adding `allow`, is fine. If you leave the box in, make sure to also leave a comment to remove the variant, once partial decoding is fully deprecated.

appetrosyan (approved):

client/src/client.rs
outoftardis: You are suggesting "Failed to..." on line 733, I'd say use the same structure here as well

version/src/lib.rs
```diff
@@ -79,20 +81,21 @@ pub mod error {
...
                 }
                 #[cfg(feature = "json")]
-                Self::Serde => "JSON (de)serialization issue",
+                Self::Serde => "JSON (de)serialization issue".to_owned(),
                 #[cfg(feature = "scale")]
-                Self::ParityScale => "Parity SCALE (de)serialization issue",
-                Self::ParseInt => "Problem with parsing integers",
-                Self::UnsupportedVersion(_) => "Input version unsupported",
+                Self::ParityScale => "Parity SCALE (de)serialization issue".to_owned(),
+                Self::ParseInt => "Problem with parsing integers".to_owned(),
```
outoftardis, line 96:
```suggestion
                Self::ParseInt => "Issue with parsing integers".to_owned(),
```

version/src/lib.rs
outoftardis, line 208: I'd suggest making this a list:

```suggestion
        /// Return error if:
        /// - version is not supported
        /// - input doesn't have enough bytes for decoding
```

core/src/kura.rs
outoftardis:
```suggestion
            warn!("Failed to decode block. Not all bytes were consumed");
```

p2p/src/peer.rs
Arjentix: Created #2384

version/derive/src/lib.rs
mversic: At first I missed that you've added `std` feature to the macro and thought this won't work

I don't know if this is better then the way I did it previously where I made the responsibility of the caller to add these things to the path.

I'd just like if we were to have one agreed way of doing things. If you want to go this route then I would appreciate you remove if there are unnecessary use statements from `data_model`. Those use statements are there only to satisfy macro expansion

We have the same situation for `iroha_schema`

client/src/client.rs
Arjentix: I'm agree with @outoftardis , using `.` or `:` in error messages could make message unreadable when errors are chained

version/derive/src/lib.rs
mversic: But we're getting rid of no_std, so it doesn't matter :)

Arjentix: Didn't noticed that you have used another aproach. Then yes, it's better to have only way to do it. I'll remove `std` feature and refactor like you suggested

cli/src/torii/utils.rs
```diff
@@ -57,8 +58,12 @@ pub mod body {
         type Error = WarpQueryError;
 
         fn try_from(body: &Bytes) -> Result<Self, Self::Error> {
-            let query = VersionedSignedQueryRequest::decode_versioned(body.as_ref())
-                .map_err(|e| WarpQueryError(Box::new(e).into()))?;
+            let mut query_res = VersionedSignedQueryRequest::decode_all_versioned(body.as_ref());
+            if query_res.is_err() {
```
appetrosyan, line 62: Can we add the error here as well?

core/src/kura.rs
```diff
@@ -404,7 +404,13 @@ impl<IO: DiskIO> BlockStore<IO> {
         #[allow(clippy::cast_possible_truncation)]
         buffer.resize(len as usize, 0);
         let _len = file_stream.read_exact(&mut buffer).await?;
-        Ok(Some(VersionedCommittedBlock::decode_versioned(&buffer)?))
+
+        let mut res = VersionedCommittedBlock::decode_all_versioned(&buffer);
+        if let Err(iroha_version::error::Error::ExtraBytesLeft(left)) = res {
```
appetrosyan, line 409: This is a frequently recurring pattern. I'd consider extracting it into a function.

version/src/lib.rs
appetrosyan: After looking at this `enum` a little, I'm surprised it didn't cause this warning earlier. I think that we crossed a threshold, so the box should stay permanently.

version/src/lib.rs
```diff
@@ -79,20 +81,21 @@ pub mod error {
...
                 #[cfg(feature = "json")]
-                Self::Serde => "JSON (de)serialization issue",
+                Self::Serde => "JSON (de)serialization issue".to_owned(),
                 #[cfg(feature = "scale")]
-                Self::ParityScale => "Parity SCALE (de)serialization issue",
-                Self::ParseInt => "Problem with parsing integers",
-                Self::UnsupportedVersion(_) => "Input version unsupported",
+                Self::ParityScale => "Parity SCALE (de)serialization issue".to_owned(),
+                Self::ParseInt => "Issue with parsing integers".to_owned(),
+                Self::UnsupportedVersion(_) => "Input version unsupported".to_owned(),
```
appetrosyan, line 97:
```suggestion
                Self::UnsupportedVersion(v) => format!("Input version {v} unsupported"),
```

appetrosyan (approved):

core/src/kura.rs
Arjentix: I'll use a macro. It will be more compact

mversic (approved):

Issue #1827, opened 2022-01-19
Parametrize hashes


mversic:
Currently, `iroha_crypto::Hash` length is hardcoded to 32 bytes. It's possible that different clients would like to have hashes of different length. As discussed this isn't required to be runtime modifiable, therefore it would be most appropriate to use const generics like:

```rs
    type HashLen = usize;
    iroha_crypto::Hash<const LEN: HashLen>
```

implemented naively, this would bring a lot of changes all around codebase so some thought should be put into how this will manifest around codebase. If this cannot be done immediately it should at least be done so that it's visible in the schemas so that clients don't make unwarranted assumptions about the size of the hash

ales-tsurko:
Some work done here: https://github.com/hyperledger/iroha/pull/2382

PR #2382, opened 2022-06-21
[feature] #1827: Parametrize hashes

ales-tsurko:
TBD


docs/source/references/config.md
```diff
@@ -1,858 +0,0 @@
-# Iroha Configuration reference
```
appetrosyan, line 1: OK, this file should definitely not be deleted!

appetrosyan (changes requested):

docs/source/references/config.md
ales-tsurko: Oops, I don't know how I did it 😅

appetrosyan:
Closed temporarily as issue is difficult to resolve directly.

Issue #2008, opened 2022-03-25
Interning of common immutable objects


mversic:
[Intering](https://en.m.wikipedia.org/wiki/Interning_(computer_science)) is a technique to reuse objects which are created often. This reduces memory footprint while the expected effect on performance should remain negligible.

In Iroha it would make sense to intern `Identifiable::Id`s like `DomainId`, `AccountId`, `Assetaid`, etc. If you know of other examples that would benefit from interning please name them in this issue.

There could be a library offering this functionality, if not, it can be implemented

appetrosyan:
https://dev.to/cad97/string-interners-in-rust-797

There's a bit more research that needs to be done. Since most of our operations are hash-table related unless we have many collisions, interning might hurt us more than help.

mversic:
just to be clear, I'm not proposing interning of all strings, just interning of object Ids.

mversic:
as far as the library is concerned [internment](https://github.com/droundy/internment) seems to be the best choice. There is also a fork of `internment` [arc-interner](https://github.com/ryzhyk/arc-interner) which doesn't seem to have the same momentum

s8sato:
The followings might have something to do with not having duplicate ids between layers:
https://wiki.hyperledger.org/display/iroha/Merkle-zation+World+State
https://docs.substrate.io/v3/advanced/storage/

appetrosyan:
During a discussion with @ilchu we concluded that after merging of #1985 interning of objects alone might not provide us with memory savings. 

For 95% of all strings we went from 24 bytes + actual string to 15 bytes + 1 byte in RAM usage. 

Interning can save us on allocation overhead (which we should benchmark), but that isn't much in the grand scheme of things. 

We should consider restricting the numbers of objects that we can have nested withing `Domain`, `Account`, etc. For example, if we require that there can only be 4bn accounts in a domain and 4bn domains, we can reduce the size of `<account as Identifiable>::Id` from `2*2*usize` to `usize`, plus the actual storage of the associated strings. As you can probably see, this only makes sense for things that are likely to be repeated.

ilchu:
Closing this and the related draft PR due to reasons stated above by @appetrosyan.

PR #2389, opened 2022-06-23
#2008: Interning of common objects

ilchu:

### Description of the Change
Tried adding internment inside `new()` methods of `data_model/src/account.rs` and `data_model/src/asset.rs`. Also added a new example to the `client` crate that would supposedly showcase interning better, with two domains having 500_000 accounts each. For note, the traces given in the last section were produced on a slightly smaller example of 300_000 accounts each.

### Issue

Work in progress towards #2008.

### Benefits
The only positive thing that I was able to track is a very slightly reduced number of calls to allocation functions, down from 1.22 billion to about 1.18 billion. The distribution of allocation sizes also changed slightly, with the internment case seeing an increase in mid-sized allocations (9 to 64 bytes), and a decrease in allocations <= 8B or within 65-256B range.

A smaller feature worthy of attention is that the memory consumed graph in the internment case has a small dip around 6 minute mark (probably connected to the fact that the two domains are created first, and only then they are populated with accounts), whereas no internment case just grows linearly from the start until both of them peak off around 1h40m and plunge (I suppose this marks hash verification stage).

### Possible Drawbacks
The actual performance seems to have degraded a bit, with a heap memory overhead max increased by about 4 Mb, from ~85 to 89 Mb.

It could also easily be the case that I'm just doing it wrong. The `internment` crate returns a `&'static T`, and as `Id` methods require ownership (otherwise I believe the refactoring would take up too many of the codebase, introducing lifetimes and drastically increasing complexity), there is inadvertent cloning involved even when the object is present in the intern, which definitely takes its toll. I couldn't come up with any strategy to circumvent that, so feedback would be appreciated.

### Usage Examples or Tests
The below traces were produced for use with [heaptrack](https://github.com/KDE/heaptrack), which gives flamegraphs to diagnose allocations or memory usage and provides a few other helpful visualizations compared to valgrind's massif. However, I've struggled with making actionable conclusions out of the differences produced between the cases. The links will die in 24 hours, so I'll update them after that should need arise.

Trace of `two_domains_million_accounts_tx.rs` with internment: [link](https://wormhole.app/lPAl1#B7wit4sFt_o7q06CT6zj1Q)
Same example, but without: [link](https://wormhole.app/ApZoO#X6iqIXJBk6sjO-b5Jp61LQ)

### Alternate Designs
Another crate allowing internment of any structures is [hashconsing](https://docs.rs/hashconsing/latest/hashconsing/index.html). It makes use of a lazy static `RWLock`-wrapped `HashSet` or `HashMap`, with the consigned terms each put into `Arc`. The `internment` crate also has other options for the strategy, such as arena or arc internment. Neither of those were tested properly yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,6 +72,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
+dependencies = [
+ "getrandom 0.2.6",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


appetrosyan:
As per our dicussion on 24-th of June, object interning might not win us a lot of space. Should we close both this PR and the issue?

Issue #2376, opened 2022-06-20
Simplify Kura


SamHSmith:
We do not spend much cpu time in Kura. It is however quite simple and self contained which makes it a decent first candidate for simplification. Especially since I already know roughly what I want to do with it. Consider it an example of what could be done to sumeragi and p2p.

SamHSmith:
Also according to @appetrosyan we should make kura it's own translation unit.

PR #2397, opened 2022-06-25
[fix] #2376: Simplified Kura, no async, two files

SamHSmith:
### Description of the Change

Removes most of the async from kura. Revamps the way blocks are stored on filesystem to a two file design. All communication *into* kura is composed of simple function calls.

### Issue

#2376 

### Benefits

The code and design  are a lot better. They will allow easier development.

### Possible Drawbacks


SamHSmith:
It still needs some more tests in kura.rs. And one of the integration tests fail, I think because kura having it's own thread causes hickups. This PR needs a proper shutdown proceedure.

actor/src/broker.rs
```diff
@@ -109,6 +109,43 @@ impl Broker {
             .fold(0, |p, (_, n)| p + if n.0.is_closed() { 0 } else { 1 })
     }
 
+    /// Synchronously send message via broker.
+    /// # Panics
+    /// Panics if called from within an async runtime.
```
appetrosyan, line 114: This isn't accurate. It panics if called as an asynchronous function, but it can be called sunchronously from within the run-time.

actor/src/lib.rs
```diff
@@ -153,7 +153,7 @@ impl<A: Actor> Addr<A> {
     /// Send a message without waiting for an answer.
     ///
     /// # Errors
-    /// Fails if queue is full or actor is disconnected
+    /// Fails if queue is full or actor is disconnected.
```
appetrosyan, line 156:
```suggestion
    /// - Queue is full;
    /// - Actor is disconnected. 
```

actor/src/lib.rs
```diff
@@ -166,6 +166,25 @@ impl<A: Actor> Addr<A> {
...
+    /// Panics when called from async code.
+    pub fn do_send_sync<M>(&self, message: M)
+    where
+        M: Message + Send + 'static,
+        M::Result: Send,
+        A: ContextHandler<M>,
+    {
+        let envelope = SyncEnvelopeProxy::pack(message, None);
+        if self.sender.blocking_send(envelope).is_err() {
+            iroha_logger::error!("Failed to send message to channel.");
```
appetrosyan, line 184: This is barely enough information. I'd add the `Debug` bound.

actor/src/lib.rs
```diff
@@ -259,6 +287,10 @@ where
         self.do_send(m).await
     }
 
+    fn send_sync(&self, m: M) {
```
appetrosyan, line 290: Not  a huge fan of `*_sync` nomenclature. Have you considered `send_blocking`? Namespacing the functions to a `synchronous` module?

cli/src/lib.rs
```diff
@@ -75,18 +74,17 @@ where
     /// Sumeragi consensus
     pub sumeragi: AlwaysAddr<S>,
     /// Kura - block storage
-    pub kura: AlwaysAddr<K>,
+    pub kura: Arc<Kura>,
```
appetrosyan, line 77: This is the main owning `Arc` as opposed to other `Weak` pointers, am I right?

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+            block_data.len() as u64,
+        )?;
+
+        Ok(())
+    }
+}
+
+/// An implementation of a block store for kura
+/// that uses rusts default file io. `std::fs`.
+pub struct StdFileBlockStore {
```
appetrosyan, line 354: I'd create an issue to find other potentially different `BlockStore` implementations.

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+    // Now for the actual printing
+    println!("Index file says there are {} blocks.", index_count);
+    println!(
+        "Printing blocks {}-{}...",
+        from_height + 1,
+        from_height + block_count
+    );
+
+    for i in 0..block_count {
+        let i_as_usize: usize = i.try_into().unwrap();
```
appetrosyan, line 103: This is an incredibly odd naming convention. Also, if this conversion is `unwrap`-ed, maybe just use `as`?

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
     }
 }
+
+#[allow(
+    clippy::print_stdout,
+    clippy::use_debug,
+    clippy::expect_used,
+    clippy::unwrap_used
+)]
+fn print_blockchain(block_store_path: &Path, from_height: Option<u64>, block_count: u64) {
```
appetrosyan, line 62:
```suggestion
fn print_blockchain(block_store_path: &Path, from_height: u64, block_count: u64) {
```

Explicitly asking the user to specify `u64::MAX` is better and shorter.

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+
+    let index_count = block_store
+        .read_index_count()
+        .expect("Failed to read index count from block store.");
+    assert!(
+        index_count != 0,
+        "The index file says there are no blocks in this block store."
+    );
+
+    let mut from_height = match from_height {
```
appetrosyan, line 73: Remove

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+)]
+fn print_blockchain(block_store_path: &Path, from_height: Option<u64>, block_count: u64) {
+    let block_store = StdFileBlockStore::new(block_store_path);
+
+    let index_count = block_store
+        .read_index_count()
+        .expect("Failed to read index count from block store.");
+    assert!(
+        index_count != 0,
+        "The index file says there are no blocks in this block store."
```
appetrosyan, line 70:
```suggestion
        "Index count is zero. This could be because there are no blocks in the store {}"
```

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+    clippy::use_debug,
+    clippy::expect_used,
+    clippy::unwrap_used
+)]
+fn print_blockchain(block_store_path: &Path, from_height: Option<u64>, block_count: u64) {
+    let block_store = StdFileBlockStore::new(block_store_path);
+
+    let index_count = block_store
+        .read_index_count()
+        .expect("Failed to read index count from block store.");
```
appetrosyan, line 67:
```suggestion
        .expect("Failed to read index count from block store {block_store:?}.");
```

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+    assert!(
+        index_count != 0,
+        "The index file says there are no blocks in this block store."
+    );
+
+    let mut from_height = match from_height {
+        Some(h) => h,
+        None => u64::MAX,
+    };
+    if from_height >= index_count {
```
appetrosyan, line 77: Remember that Rust is expression-based, so you can just have a single assignment.

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+            index_start,
+            index_len
+        );
+        let index_len_as_usize: usize = index_len.try_into().unwrap();
+        let mut block_buf = vec![0_u8; index_len_as_usize];
+        block_store
+            .read_block_data(index_start, &mut block_buf)
+            .expect("Failed to read block data.");
+        let block =
+            VersionedCommittedBlock::decode_versioned(&block_buf).expect("Failed to decode block");
```
appetrosyan, line 119:
```suggestion
            VersionedCommittedBlock::decode_versioned(&block_buf).expect(format!("Failed to decode block: № {}", index_index + 1);
```

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+            assert!(height != 0, "The genesis block has the height 1. Therefore the \"from height\" you specify must not be 0.");
+            // Kura starts counting blocks from 0 like an array while the outside world counts the first block as number 1.
+            Some(height - 1)
+        }
+        None => None,
+    };
 
-        match command {
-            Command::Print { length } => Config::Print(print::Config { from, length }),
+    match args.command {
```
appetrosyan, line 49: Consider turning this into an inherent impl.

appetrosyan: C. F. R. `iroha_client_cli`

tools/kura_inspector/Cargo.toml
```diff
@@ -8,11 +8,6 @@ edition = "2021"
...
-iroha_config = { path = "../../config" }
+iroha_version = { path = "../../version" }
 
-clap = { version = "3.1.10", features = ["derive", "cargo"] }
-futures-util = "0.3"
-tokio = { version = "1.6.0", features = ["rt"]}
-
-[dev-dependencies]
-tempfile = "3"
+clap = { version = "3.1.10", features = ["derive", "cargo"] }
```
appetrosyan, line 13: I don't mind not having a newline at the end of files, however, for consistency, I'd add one.

appetrosyan (dismissed):

core/src/sumeragi/fault.rs
```diff
@@ -41,10 +36,9 @@ impl FaultInjection for NoFault {
 }
 
 /// `Sumeragi` is the implementation of the consensus. This struct allows also to add fault injection for tests.
-pub struct SumeragiWithFault<G, K, F>
+pub struct SumeragiWithFault<G, F>
```
appetrosyan, line 39: Also somehow this is the same struct that is ran on the actual peer. We could just rename it to `Sumeragi` and keep the fault injection behind a feature flag.

actor/src/broker.rs
SamHSmith: Oh yea. That comment is old. Before I was creating a new runtime on the current thread to run async code and that would cause a panic if called from within an existing runtime. Will change the comment.

actor/src/lib.rs
SamHSmith: A module seems like it would be quite a hassle. I agree _sync isn't ideal, but blocking also implies some kind of wait when the point of this function is that it returns nearly instantly. Maybe in the future we could have it be just send? The async code should be able to use send_sync.

core/src/sumeragi/fault.rs
SamHSmith: Yea I think that's probably a good idea.

core/src/kura.rs
SamHSmith: What do you mean by find? Seems more like something you would pass down from above. "We want to run on this system."

tools/kura_inspector/src/main.rs
SamHSmith: I can't use `as` because that truncates it without panic. I can't put the try-unwrap inside the array index operation because the compiler won't know what version of try_into to use. It errors when you index with a u64, says it must be usize. But then it can't figure out that it has to try_into::<usize>. I think it's a compiler bug.

actor/src/lib.rs
appetrosyan: I'd prefer a future where async can live with sync. Feel free to resolve.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                    Ok(decoded_block) => {
+                        blocks.push(decoded_block);
+                    }
+                    Err(error) => {
+                        error!(%error, "Encountered malformed block. Not reading any blocks beyond this height.");
+                        break;
+                    }
+                },
+                Err(error) => {
+                    error!(%error, "Malformed block index or corrupted block data file. Not reading any blocks beyond this height.");
```
appetrosyan, line 126:
```suggestion
                    error!(?error, "Malformed block index or corrupted block data file. Not reading any blocks beyond this height.");
```

The debug representation contains the traceback. This will be useful once we have some more details to add to the error.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                    "TODO: handle allocation too large because block_len is corrupted."
+                )
+            ];
+            match block_store.read_block_data(block_start, &mut block_data_buffer) {
+                Ok(_) => match VersionedCommittedBlock::decode_versioned(&block_data_buffer) {
+                    Ok(decoded_block) => {
+                        blocks.push(decoded_block);
+                    }
+                    Err(error) => {
+                        error!(%error, "Encountered malformed block. Not reading any blocks beyond this height.");
```
appetrosyan, line 121: This nested error handling is the so-called tower of doom. Consider using map and map_err.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
-    /// Fails if reading from disk while initing fails
-    async fn new(
-        mode: Mode,
-        block_store_path: &Path,
-        blocks_per_file: NonZeroU64,
-        wsv: Arc<WorldStateView>,
-        broker: Broker,
-        actor_channel_capacity: u32,
-    ) -> Result<Self>;
+        let kura_weak_reference = Arc::downgrade(&kura);
```
appetrosyan, line 66: Needs comment.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                    break;
+                }
+            }
+        }
+
+        info!("Loaded {} blocks at init.", blocks.len());
+
+        let hash_array = blocks.iter().map(VersionedCommittedBlock::hash).collect();
+
+        let mut guard = self
```
appetrosyan, line 136: Add explicit scope around this. With Mutexex it's very important to release the lock as soon as possible. Rustc cannot possibly know that it can release the lock guard before returning.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+
+        Ok(blocks)
     }
-}
 
-#[async_trait::async_trait]
-impl<IO: DiskIO> Actor for KuraWithIO<IO> {
-    fn actor_channel_capacity(&self) -> u32 {
-        self.actor_channel_capacity
+    // I think we should rethink this. Kura should not be the
```
appetrosyan, line 145: Fully agree. Create issue.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                .lock()
+                .expect("able to lock kura recieve block mutex");
 
-        #[allow(clippy::panic)]
-        match self.init().await {
-            Ok(blocks) => {
-                #[cfg(feature = "telemetry")]
-                if let Some(block) = blocks.first() {
-                    iroha_logger::telemetry!(msg = iroha_telemetry::msg::SYSTEM_CONNECTED, genesis_hash = %block.hash());
+            match block_reciever_guard.try_recv() {
```
appetrosyan, line 178: Consider making this a trait and inherent impl. This pattern could come up more than once.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                Err(_) => {
+                    std::thread::sleep(std::time::Duration::from_millis(1));
```
appetrosyan, line 180: Keep in mind that thread sleep might release the OS thread, but not the run-time thread, which might need yield.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                    continue;
+                }
+                Ok(new_block) => {
+                    #[cfg(feature = "telemetry")]
+                    if new_block.header().height == 1 {
```
appetrosyan, line 185: extract into function.

appetrosyan (dismissed):

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
-                self.merkle_tree = blocks
-                    .iter()
-                    .map(VersionedCommittedBlock::hash)
-                    .collect::<MerkleTree<_>>();
-                Err(error)
-            }
+        start_location_in_data_file: u64,
+        block_data: &[u8],
+    ) -> Result<()> {
+        let mut data_file = std::fs::OpenOptions::new()
```
mversic, line 453: is this necessary considering that you have `create_files_if_they_do_not_exist`?

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+        self.block_sender
+            .send(block)
+            .await
+            .expect("Kura block channel is still open.");
+    }
+
+    /// Put a block in the queue to be stored by kura. If the queue is
+    /// full this function blocks until there is space.
+    #[allow(clippy::expect_used)]
+    pub fn store_block_blocking(&self, block: VersionedCommittedBlock) {
```
appetrosyan, line 242:
```suggestion
    pub fn blocking_store_block(&self, block: VersionedCommittedBlock) {
```

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+        self.block_sender
+            .blocking_send(block)
+            .expect("Kura block channel is still open.");
     }
 }
 
-#[async_trait::async_trait]
-impl<IO: DiskIO> Handler<StoreBlock> for KuraWithIO<IO> {
-    type Result = ();
+/// Kura is the block storage subsystem. It stores it's blocks
```
appetrosyan, line 249: @outoftardis

core/src/kura.rs
appetrosyan: I mean, in addition to `std::fs`, there could be other implementations. Some with fewer bells and whistles. We should create a meta-issue to gather those.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+        let start_location = start_block_height * 2 * 8;
+        let block_count = dest_buffer.len() as u64;
+
+        if start_location + (2 * 8) * block_count > index_file.metadata()?.len() {
+            return Err(Error::OutOfBoundsBlockRead(start_block_height, block_count));
+        }
+        index_file.seek(SeekFrom::Start(start_location))?;
+        // (start, length), (start,length) ...
+        for i in 0..block_count {
+            let index: usize = i.try_into().unwrap();
```
appetrosyan, line 391: So if this is an error that should come up more frequently, I'm not a huge fan of `unwrap`-ing. 
- `expect` also panics, but gives you an error message, which I think would be appropriate here. 
- This is a potentially recoverable error, which only occurs if `u64 > usize`, i.e. a 32-bit architecture, where we should have a workaround. That is, you can have another array of chunks, which can then be subscripted by the 32-bit sub-words of a single `u64`.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+        index_file.write_u64::<LittleEndian>(length)?;
+        Ok(())
+    }
+
+    #[allow(clippy::integer_division)]
+    fn read_index_count(&self) -> Result<u64> {
+        let index_file = std::fs::OpenOptions::new()
+            .read(true)
+            .open(self.path_to_blockchain.join("blocks.index"))?;
+        Ok(index_file.metadata()?.len() / (2 * 8)) // Each entry is 16 bytes.
```
appetrosyan, line 422:
```suggestion
        Ok(index_file.metadata()?.len() / (2 * std::mem::size_of::<u64>()))
```

core/src/kura.rs
```diff
@@ -305,224 +514,9 @@ pub enum Error {
     /// Inconsequential write
     #[error("Inconsequential block write.")]
```
appetrosyan, line 515: This error message is completely opaque. If we can, we should consider making it more obvious.

tools/kura_inspector/src/main.rs
mversic: there is also `try_from`

cli/src/lib.rs
```diff
@@ -75,18 +74,17 @@ where
     /// Sumeragi consensus
     pub sumeragi: AlwaysAddr<S>,
     /// Kura - block storage
-    pub kura: AlwaysAddr<K>,
```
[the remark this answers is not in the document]
SamHSmith: Yes.

core/src/kura.rs
SamHSmith: If the index fits in 32 bits then it doesn't fail. When `i >= 2^32` it will panic, and that makes sense because a lot of code has to change to load and parse a 4 billion block chain on a 32 bit machine. It can be done but then most of this code must be redone.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+            data_file.set_len(start_location_in_data_file + block_data.len() as u64)?;
         }
+        data_file.seek(SeekFrom::Start(start_location_in_data_file))?;
+        data_file.write_all(block_data)?;
+        Ok(())
+    }
+
+    fn create_files_if_they_do_not_exist(&mut self) -> Result<()> {
+        std::fs::create_dir_all(&self.path_to_blockchain)?;
+        let _index_file = std::fs::OpenOptions::new()
```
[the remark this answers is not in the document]
SamHSmith: I guess for readability? You are right, they are quite redundant.

core/src/kura.rs
SamHSmith: In this case I think it is warrented. There is such a thing as appropriate use.

core/src/kura.rs
SamHSmith: Yes, except this is in init and all the locks are available to kura only so we really don't need to care. The rest of kura cannot start running until this function returns.

core/src/kura.rs
SamHSmith: Seperate issue and PR in that case. We shouldn't prematurely compress. Once we have done similar simplification in other parts of iroha we can identify the common patterns and extract those. But we do that *after* we have multiple peices of code that do the same thing.

core/src/kura.rs
SamHSmith: Yes, rules are very different when in the runtime vs out of the runtime. Or if you are on the "blocking" tokio thread pool vs the normal one.

core/src/kura.rs
SamHSmith: I disagree. Things get to be functions when they have multiple users. This peice of code is used in one place and will only ever be used in one place since it does work specific to the store thread. You wouldn't extract a paragraph of an essay and put it at the end. Same thing applies. Extracting into a function hinders readability, this is normally fine because you get reuse, not the case here though.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+    fn read_block_data(
+        &self,
+        start_location_in_data_file: u64,
+        dest_buffer: &mut [u8],
+    ) -> Result<()>;
+    /// Write `block_data` into the data file starting at
+    /// `start_location_in_data_file`. Extend the file if
+    /// necessary.
+    /// # Errors
+    /// IO Error.
```
outoftardis, line 299:
```suggestion
    ///
    /// # Errors
    /// Fails if there is an IO Error.
```

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+        &mut self,
+        start_location_in_data_file: u64,
+        block_data: &[u8],
+    ) -> Result<()>;
+
+    /// Create the index and data files if they do not
+    /// already exist.
+    /// # Errors
+    /// Will fail if the any of the files don't exist
+    /// and couldn't be created.
```
outoftardis, line 310:
```suggestion
    ///
    /// # Errors
    /// Fails if any of the files don't exist
    /// and couldn't be created.
```

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+    fn create_files_if_they_do_not_exist(&mut self) -> Result<()>;
+
+    /// Above are the platform dependent functions.
+    /// Below are the platform independent functions that use the above functions.
```
outoftardis, line 314: I think these two lines should be non-doc comments (`//`). 

I'd suggest adding a doc comment to each function saying whether or not it is platform independent

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
     let args = Args::parse();
-    let mut output = DefaultOutput::new();
-    Config::from(args)
-        .run(&mut output)
-        .await
-        .unwrap_or_else(|e| eprintln!("{:?}", e))
+
+    let from_height = match args.from {
+        Some(height) => {
+            assert!(height != 0, "The genesis block has the height 1. Therefore the \"from height\" you specify must not be 0.");
```
outoftardis, line 42:
```suggestion
            assert!(height != 0, "The genesis block has the height 1. Therefore, the \"from height\" you specify must not be 0.");
```

core/src/kura.rs
```diff
@@ -305,224 +514,9 @@ pub enum Error {
...
-                format!("{}{}", dir_string, file_string.to_string_lossy()).parse::<NonZeroU64>();
-            if let Ok(num) = maybe_num {
-                bases.insert(num);
-            }
-        }
-    }
-
-    Ok(bases)
+    /// A read, which was initiated by the block store, went off the end of the block data file.
+    #[error("Block read out of bounds: {0}, {1}.")]
```
outoftardis, line 518: maybe something like this?

```suggestion
    #[error("Tried reading block data out of bounds: {0}, {1}.")]
```

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+
+    #[allow(clippy::integer_division)]
+    fn read_index_count(&self) -> Result<u64> {
+        let index_file = std::fs::OpenOptions::new()
+            .read(true)
+            .open(self.path_to_blockchain.join("blocks.index"))?;
+        Ok(index_file.metadata()?.len() / (2 * 8)) // Each entry is 16 bytes.
+    }
+
+    fn write_index_count(&mut self, new_count: u64) -> Result<()> {
```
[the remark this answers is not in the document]
SamHSmith: What do you think we should be fault tolerant against? I think it's safe to assume that the storage will be reliable. If we want to run on a platform where it is uncertain if we are able to read data, that should be handled inside the platform specific block store.

I think worrying about sudden powerloss is not our problem. That's why you buy a UPS for your servers. Or have your own powerplant for your data center. That leaves us with worrying about sudden shutdown due to an EMP. Which if that happens you have way bigger problems and the data is probably ruined anyway.

What we do need is testing for the case of corrupted data on disk. Maybe we screwed it up due to sudden shutdown or it was just bitrot. Doesn't matter. We should not load a blockchain unless it is in perfect order, or we can recover the data fully.

core/src/kura.rs
```diff
@@ -305,224 +514,9 @@ pub enum Error {
...
-            let maybe_num =
-                format!("{}{}", dir_string, file_string.to_string_lossy()).parse::<NonZeroU64>();
-            if let Ok(num) = maybe_num {
-                bases.insert(num);
-            }
-        }
-    }
-
-    Ok(bases)
+    /// A read, which was initiated by the block store, went off the end of the block data file.
```
outoftardis, line 517:
```suggestion
    /// Kura tried reading data beyond the end of the block data file.
```
```suggestion
    /// The block store tried reading data beyond the end of the block data file.
```

not sure if these are correct but I'd try simplifying this error description

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+            start_location_in_data_file,
+            block_data.len() as u64,
+        )?;
+
+        Ok(())
+    }
+}
+
+/// An implementation of a block store for kura
+/// that uses rusts default file io. `std::fs`.
```
outoftardis, line 353:
```suggestion
/// An implementation of a block store for Kura
/// that uses `std::fs`, the default IO file in Rust.
```

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+        self.read_block_indices(block_height, std::slice::from_mut(&mut index))?;
+        Ok(index)
+    }
+
+    /// Append `block_data` to this block store. First writing
+    /// the data to the data file and then creating a new index
+    /// for it.
+    /// # Errors
+    /// Fails if any of the required platform specific function operations
+    /// fail.
```
outoftardis, line 330:
```suggestion
    /// Append `block_data` to this block store: write
    /// the data to the data file and then create a new index
    /// for it.
    ///
    /// # Errors
    /// Fails if any of the required platform-specific function operations
    /// fail.
```

`function operations` -- maybe just `functions` or just `operations`?

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+    /// Will fail if the any of the files don't exist
+    /// and couldn't be created.
+    fn create_files_if_they_do_not_exist(&mut self) -> Result<()>;
+
+    /// Above are the platform dependent functions.
+    /// Below are the platform independent functions that use the above functions.
+
+    /// Short hand for calling `read_block_indices` with a buffer of one.
+    /// # Errors
+    /// Passes on any errors from `read_block_indices`.
```
outoftardis, line 318:
```suggestion
    /// Call [read_block_indices] with a buffer of one.
    ///
    /// # Errors
    /// Fails if there is an IO Error.
```

tools/kura_inspector/src/main.rs
SamHSmith: That might be the best solution

core/src/kura.rs
```diff
@@ -4,258 +4,487 @@
...
 
-    /// Methods consumes new validated block and atomically stores and caches it.
-    #[iroha_futures::telemetry_future]
-    #[log("INFO", skip(self, block))]
-    pub async fn store(
+    fn write_block_index(&mut self, block_height: u64, start: u64, length: u64) -> Result<()> {
+        let mut index_file = std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .open(self.path_to_blockchain.join("blocks.index"))?;
```
mversic, line 418: can we make filenames a constant?

core/src/kura.rs
```diff
@@ -4,258 +4,487 @@
...
+            .read(true)
+            .open(self.path_to_blockchain.join("blocks.index"))?;
+        Ok(index_file.metadata()?.len() / (2 * std::mem::size_of::<u64>() as u64)) // Each entry is 16 bytes.
+    }
+
+    fn write_index_count(&mut self, new_count: u64) -> Result<()> {
+        let index_file = std::fs::OpenOptions::new()
+            .write(true)
+            .open(self.path_to_blockchain.join("blocks.index"))?;
+        let new_byte_size = new_count * 2 * 8;
```
mversic, line 443: can we extract these magic numbers into some conversion function?

core/src/kura.rs
SamHSmith: I'm going with the second one

core/src/kura.rs
mversic: I think we've had an agreement to always prefer `expect` vs `unwrap`.

core/src/kura.rs
SamHSmith: good idea

SamHSmith: Well okay, they are all technically platform dependent from the "users" perspective. But those two comments should not be doc comments.

SamHSmith: They are comments for the programmers in the future implementing a block store.

core/src/kura.rs
SamHSmith: For these I prefer the short form.

core/src/kura.rs
```diff
@@ -4,258 +4,487 @@
...
+    ) -> Result<()>;
 
-    async fn handle(&mut self, StoreBlock(block): StoreBlock) {
-        #[cfg(feature = "telemetry")]
-        if block.header().height == 1 {
-            iroha_logger::telemetry!(msg = iroha_telemetry::msg::SYSTEM_CONNECTED, genesis_hash = %block.hash());
-        }
-        if let Err(error) = self.store(block).await {
-            error!(%error, "Failed to write block")
+    /// Write a single block index at the specified `block_height`.
```
outoftardis, line 277:
```suggestion
    /// Write the index of a single block at the specified `block_height`.
```

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+                            strong_kura_reference
+                                .block_hash_array
+                                .lock()
+                                .expect("lock on block hash array")
+                                .push(block_hash);
+                            strong_kura_reference.broker.issue_send_sync(&ContinueSync);
+                        }
+                        Err(error) => {
+                            error!(%error, "Failed to write block. Kura will now init again.");
+                            panic!("It is unclear what we should do here.");
```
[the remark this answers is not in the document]
SamHSmith: It's a work in progress. It being a bad message helps communicate to me at least that we must put actual code here before we ship.

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
-            mode,
-            block_store_path,
-            blocks_per_file,
-            wsv,
-            broker,
-            mailbox,
-            DefaultIO,
-        )
-        .await
+    /// After constructing [`Kura`] it should be initialized to be ready to work with it.
```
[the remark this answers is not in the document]
SamHSmith: 3 is best I think. In case someone forgets to call init, it's an extra lead.

core/src/kura.rs
SamHSmith: I'll change them to expects to honor such an agreement. But I think I end up repeating myself. The code already makes it obvious what has gone wrong and then I type it out again. Seems foolish.

core/src/kura.rs
SamHSmith: We could, but that would make it less readable in my opinion. It is true that we write this a couple of times. But the size of a function call vs this statement is the same. And with it this way I the reader can actually see what the value is without having to look it up.

mversic: seems like a difference in opinions. For me, as someone who hasn't been deep into `kura`, this is less readable than a function with a good descriptive name. Can we at least use constants instead of magic numbers?

core/src/kura.rs
```diff
@@ -4,258 +4,475 @@
...
+    /// Attempt to fill all of `dest_buffer`.
+    /// # Errors
+    /// IO Error.
+    fn read_block_indices(
+        &self,
+        start_block_height: u64,
+        dest_buffer: &mut [(u64, u64)],
+    ) -> Result<()>;
+    /// Write a single block index at the specified `block_height`.
+    /// If `block_height` is beyond the end of the index file attempt to
```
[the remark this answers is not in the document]
outoftardis: please add a comma after "file"

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
+    /// # Errors
+    /// IO Error.
+    fn read_block_indices(
+        &self,
+        start_block_height: u64,
+        dest_buffer: &mut [(u64, u64)],
+    ) -> Result<()>;
+
+    /// Write the index of a single block at the specified [`block_height`].    
+    /// If [`block_height`] is beyond the end of the index file attempt to
```
outoftardis, line 280:
```suggestion
    /// If [`block_height`] is beyond the end of the index file, attempt to
```

outoftardis (commented):
no other comments from my side

actor/src/lib.rs
```diff
@@ -166,6 +168,27 @@ impl<A: Actor> Addr<A> {
         }
     }
 
+    /// Send a message without waiting for an answer.
```
Erigara, line 171: @SamHSmith  are you sure that we would send without waiting?
From the docs and source code for [sender](https://docs.rs/tokio/latest/src/tokio/sync/mpsc/bounded.rs.html#654-656) it seems like code will block until receiving an answer.

core/src/kura.rs
SamHSmith: yes

actor/src/lib.rs
SamHSmith: It blocks if there is not space in the message pipe. In the normal case where everything is functioning it puts the message in the pipe and then returns directly.

Erigara: Ok, resolve than

core/src/kura.rs
```diff
@@ -4,258 +4,502 @@
...
+
+    /// Create the index and data files if they do not
+    /// already exist.
+    /// # Errors
+    /// Fails if the any of the files don't exist
+    /// and couldn't be created.
+    fn create_files_if_they_do_not_exist(&mut self) -> Result<()>;
+
+    // Above are the platform dependent functions.
+    // Below are the platform independent functions that use the above functions.
```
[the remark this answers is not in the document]
appetrosyan: To be perfectly honest, if you want to group functions you use modules. They're extremely readable, and if you're concerned about long names, you can always glob import.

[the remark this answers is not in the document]
appetrosyan: E.g. 
```
#[cfg(target = …]
pub mod platform_dependent {

}

pub mod platform_independent {
    use super::platform_depedent::*;
} 
```

tools/kura_inspector/src/main.rs
SamHSmith: Could you give a code example?

core/src/kura.rs
Arjentix: I disagree that functions make readability worse. Good function should do one semantic thing. Reading meaningful function calls is easier than just blocks of code

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
-#[async_trait::async_trait]
-impl<IO: DiskIO> Handler<GetBlockHash> for KuraWithIO<IO> {
-    type Result = Option<HashOf<VersionedCommittedBlock>>;
-    async fn handle(&mut self, GetBlockHash { height }: GetBlockHash) -> Self::Result {
-        if height == 0 {
+    /// Get the hash of the block at the provided height.
+    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
+    pub fn get_block_hash(&self, block_height: u64) -> Option<HashOf<VersionedCommittedBlock>> {
+        let hash_array_guard = self.block_hash_array.lock().expect("access hash array");
+        if block_height == 0 || block_height > hash_array_guard.len() as u64 {
```
Arjentix, line 233:
```suggestion
        if block_height == 0 || block_height > hash_array_guard.len().try_into().expect("`usize` should always fit in `u64`") {
```

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
+/// The interface for the **block store**, which is where [Kura],
+/// the block storage subsystem, stores its blocks.
+///
+/// The [`BlockStoreTrait`] defines and implements functionality used by Kura
+/// for every platform. The default implementation is `StdFileBlockStore`,
+/// which uses `std::fs`.
+pub trait BlockStoreTrait {
+    /// Read a series of block indices from the block index file and
+    /// attempt to fill all of `dest_buffer`.
+    /// # Errors
```
Arjentix, line 271:
```suggestion
    /// attempt to fill all of `dest_buffer`.
    ///
    /// # Errors
```

tools/kura_inspector/src/lib.rs
```diff
@@ -1,80 +0,0 @@
-//! General objects independent from executables.
```
Arjentix, line 1: So we can't use `kura_inspector` as a library now? Was it useless or what?

tools/kura_inspector/src/main.rs
```diff
@@ -21,27 +29,95 @@ enum Command {
...
+    );
+
+    for i in 0..block_count {
+        let i_as_usize: usize = i.try_into().unwrap();
+        let (index_start, index_len) = block_indices[i_as_usize];
+        let index_index = from_height + i;
+
+        println!(
+            "Block#{} starts at byte offset {} and is {} bytes long.",
+            index_index + 1,
```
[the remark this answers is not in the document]
SamHSmith: It is only used once. Doesn't deserve variable status.

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
-        )
-        .await
+    /// Initialize [`Kura`] after its construction to be able to work with it.
+    ///
+    /// # Errors
+    /// Fails if:
+    /// - file storage is unavailable
+    /// - data in file storage is invalid or corrupted
+    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
+    pub fn init(&self) -> Result<Vec<VersionedCommittedBlock>> {
```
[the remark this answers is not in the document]
SamHSmith: It's split up this way because we want the option of re-initializing kura. We may decide down the line that we don't ever need to do that and could then merge it with new. This aspect is inherited from the way kura was previously.

core/src/kura.rs
SamHSmith: I think this is pointless. There are no 128-bit platforms. This cast cannot fail, ever. So sacrificing readability for "correctness" that isn't more correct in the real world isn't productive.

tools/kura_inspector/src/lib.rs
SamHSmith: It was not used by anyone at all. It would have been necessary before because kura's storage format was so complicated you would want a library.

Now on the other hand, you don't need a kura_inspector library. Just write the code that inspects the blocks yourself using `BlockStoreTrait`. You get more power, flexibility and the code is still really small.

core/src/kura.rs
SamHSmith: I've now changed all of them.

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
+        for (block_start, block_len) in block_indices {
+            let mut block_data_buffer = Vec::new();
+            let block_data_buffer_len: usize = block_len
+                .try_into()
+                .expect("TODO: handle allocation too large because block_len is corrupted.");
+            block_data_buffer.try_reserve(block_data_buffer_len)?;
+            block_data_buffer.resize(block_data_buffer_len, 0_u8);
+
+            match block_store.read_block_data(block_start, &mut block_data_buffer) {
+                Ok(_) => match VersionedCommittedBlock::decode_versioned(&block_data_buffer) {
```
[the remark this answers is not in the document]
SamHSmith: Not really. It's written this way anticipaiting block specific error messages and potentially correcting correctable corrupting.

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
+        block_indices.try_reserve(block_index_count)?;
+        block_indices.resize(block_index_count, (0_u64, 0_u64));
+        block_store.read_block_indices(0, &mut block_indices)?;
+
+        for (block_start, block_len) in block_indices {
+            let mut block_data_buffer = Vec::new();
+            let block_data_buffer_len: usize = block_len
+                .try_into()
+                .expect("TODO: handle allocation too large because block_len is corrupted.");
+            block_data_buffer.try_reserve(block_data_buffer_len)?;
```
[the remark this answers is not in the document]
SamHSmith: Yes. But just using try_reserve would communicate to someone reading the code that it's an irrecoverable error. It's written this way to say to the code reader that the work here is not done.

core/src/kura.rs
```diff
@@ -4,258 +4,504 @@
...
-        blocks_per_file: NonZeroU64,
         wsv: Arc<WorldStateView>,
         broker: Broker,
-        actor_channel_capacity: u32,
-        io: IO,
-    ) -> Result<Self> {
-        Ok(Self {
+        block_channel_size: u32,
+    ) -> Result<Arc<Self>> {
+        let (block_sender, block_reciever) = channel(block_channel_size as usize);
```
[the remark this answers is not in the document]
SamHSmith: This as well cannot fail, _ever_, so there is no point in using `try_into`.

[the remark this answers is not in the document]
Arjentix: We have already discussed that some months ago. The idea is that it looks like a C-style cast. If you have any assumption like *"it will never fail"* then you can show this assumption in the code using `expect`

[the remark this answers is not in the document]
Arjentix: Consider that someone will refactor the type of  `block_channel_size` to `u64` using the auto-refactoring feature in the IDE. Then this `as` cast will be incorrect, but Rust won't say anything to you.
Using `expect` we can find this error quicker, but still only in runtime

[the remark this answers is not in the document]
SamHSmith: You make a good point. I will change it.

appetrosyan (approved):

appetrosyan (approved):

mversic (approved):

Arjentix (approved):

Issue #2000, opened 2022-03-23
Disallow empty `Identifiers`


mversic:
At the moment it is allowed to create empty `Name`, `DomainId`, `AccountId`, etc. Should we continue to allow empty string identifiers or should we forbid them. What about using `Option`? 

It should be noted that empty string is a well defined identifier

ales-tsurko:
@mversic @appetrosyan could you clarify this issue?

> It should be noted that empty string is a well defined identifier

So `Name` type is allowed to have empty string? But in which case `AccountId` can be empty? And is it possible, for example, to have `AccountId` like `""@""`? So we should require non-empty strings for **some** types? I mean, should we forbid, for example, using empty name and domainID for `AccountId`, or forbid name to be empty for `DomainId` (because it's the only field), but still allow `Name` to be empty?

appetrosyan:
To me it seems that the sole purpose of the `Name` struct is to encapsulate that some valid strings are invalid identifiers. As such an empty identifier is not a valid Name. Thus, we should allow an empty constString, but not an empty Name.

PR #2402, opened 2022-06-28
[feature] #2000: Disallow empty names

ales-tsurko:
This PR disallows to use empty strings for `iroha_data_model::Name`. Thus, it forbids empty `AccountId`, `DomainId` etc. In case of `Name`'s initialization from an empty string, an error is raised. For discussion see the linked issue (#2000).

### DIFF
--- a/data_model/src/account.rs
+++ b/data_model/src/account.rs
@@ -429,13 +429,6 @@ pub struct Id {
 }
 
 impl Id {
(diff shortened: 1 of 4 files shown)


ales-tsurko:
I've reopened it here: https://github.com/hyperledger/iroha/pull/2403 Because after changing the destination branch, something became wrong with CI (it wouldn't do iroha2 jobs).

PR #2403, opened 2022-06-28
[feature] #2000: Disallow empty names

ales-tsurko:
This PR disallows to use empty strings for `iroha_data_model::Name`. Thus, it forbids empty `AccountId`, `DomainId` etc. In case of `Name`'s initialization from an empty string, an error is raised. For discussion see the linked issue (#2000).

### DIFF
--- a/data_model/src/account.rs
+++ b/data_model/src/account.rs
@@ -455,7 +455,7 @@ impl FromStr for Id {
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         if string.is_empty() {
             return Err(ParseError {
-                reason: "`AccountId` can not be empty",
+                reason: "`AccountId` cannot be empty",
             });
         }
 
--- a/data_model/src/account.rs
+++ b/data_model/src/account.rs
@@ -429,13 +429,6 @@ pub struct Id {
 }
 
 impl Id {
-    pub(crate) const fn empty() -> Self {
-        Self {
-            name: Name::empty(),
-            domain_id: DomainId::empty(),
-        }
-    }
-
     /// Construct [`Id`] from an account `name` and a `domain_name` if
     /// these names are valid.
     #[inline]
@@ -461,7 +454,9 @@ impl FromStr for Id {
 
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         if string.is_empty() {
-            return Ok(Self::empty());
+            return Err(ParseError {
+                reason: "`AccountId` can not be empty",
+            });
         }
 
         let vector: Vec<&str> = string.split('@').collect();
--- a/data_model/src/asset.rs
+++ b/data_model/src/asset.rs
@@ -648,7 +648,7 @@ impl FromStr for DefinitionId {
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         if string.is_empty() {
             return Err(ParseError {
-                reason: "`DefinitionId` can not be empty",
+                reason: "`DefinitionId` cannot be empty",
             });
         }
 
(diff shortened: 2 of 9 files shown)


data_model/src/name.rs
```diff
@@ -22,9 +22,9 @@ use crate::{ParseError, ValidationError};
 pub struct Name(ConstString);
 
 impl Name {
-    pub(crate) const fn empty() -> Self {
-        Self(ConstString::new())
-    }
+    // pub(crate) const fn empty() -> Self {
```
Erigara, line 25: commented code

data_model/src/name.rs
```diff
@@ -80,12 +81,7 @@ impl FromStr for Name {
     type Err = ParseError;
 
     fn from_str(candidate: &str) -> Result<Self, Self::Err> {
-        Self::validate_str(candidate).map(|_| {
-            if candidate.is_empty() {
-                return Self::empty();
-            }
-            Self(ConstString::from(candidate))
-        })
+        Self::validate_str(candidate).map(|_| Self(ConstString::from(candidate)))
```
Erigara, line 84: looks much cleaner now without `is_empty` branch!

Erigara (changes requested):
It looks good in general, however I think we should remove the commented out code and also add a test case to check that trying to pass an empty string returns an error.

data_model/src/account.rs
```diff
@@ -461,7 +454,9 @@ impl FromStr for Id {
 
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         if string.is_empty() {
-            return Ok(Self::empty());
+            return Err(ParseError {
+                reason: "`AccountId` can not be empty",
```
outoftardis, line 458:
```suggestion
                reason: "`AccountId` cannot be empty",
```

data_model/src/asset.rs
```diff
@@ -654,7 +647,9 @@ impl FromStr for DefinitionId {
 
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         if string.is_empty() {
-            return Ok(Self::empty());
+            return Err(ParseError {
+                reason: "`DefinitionId` can not be empty",
```
outoftardis, line 651:
```suggestion
                reason: "`DefinitionId` cannot be empty",
```

data_model/src/name.rs
```diff
@@ -52,9 +52,10 @@ impl Name {
     /// Fails if not valid [`Name`].
     fn validate_str(candidate: &str) -> Result<(), ParseError> {
         if candidate.is_empty() {
-            return Ok(());
+            return Err(ParseError {
+                reason: "`Name` can not be empty",
```
outoftardis, line 56:
```suggestion
                reason: "`Name` cannot be empty",
```

outoftardis (commented):
`cannot` is just far more common compared to `can not`, so I suggest this change

data_model/src/name.rs
ales-tsurko: Ah, I just forgot to push it.

outoftardis (approved):

data_model/src/account.rs
```diff
@@ -461,7 +454,9 @@ impl FromStr for Id {
 
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         if string.is_empty() {
```
Erigara, line 456: I think this check is unnecessary because the next check for the correct format will always fail with an empty string, but it's useful in terms of error readability. 
Should we keep it? @ales-tsurko what do you think?

ales-tsurko: I think this check is unnecessary because the next check for the correct format will always fail with an empty string, but it's useful in terms of error readability.
Should we keep it? @ales-tsurko what do you think?

It has different `reason` description. The first is more specific and related to particular case (i.e. empty string).

Erigara: ok, resolve then

Erigara (approved):

PR #2417, opened 2022-06-30
[refactor] #2307: Make `events_sender` in `WorldStateView` non-optional

### LINKED ISSUE #2307: Refactor produce events
As of now `produce_events` in `core/src/wsv.rs`has a few worrying branches. While the confusing return statement was fixed, we still have the problem of checking for whether an event sender is present outside of a loop. 

Because of `impl Into<Event>` it's possible that the inlined implementation of the function could elide the repeated check, but it is far from guaranteed. It is preferable to use the borrow checker to precisely always have an `event_sender` and to not have to check an option; @Erigara referred to this as a state-machine. 

Another point is that the return statement contains an explicit `drop`, which is usually a code smell for something being better expressed as a scope. The function needs to be investigated further and possibly extended.

Erigara:

### Description of the Change

- Change `events_sender` in `WorldStateView` from `Option<EventsSender>` to `EventsSender`: this change is possible because previously `WorldStateView::with_events` was called single time during `Iroha` initialization just after `WorldStateView::from_configuration`. 
- Add `events_sender` parameter to `WorldStateView::from_configuration` constructor. 

### Issue

Closes #2307.

### Benefits

Avoid overhead of checking if `events_sender` is `Some` with every call to `produce_event`.

### Possible Drawbacks

In `WorldStateView::new` fake `events_sender` is created, but since `new` called only in tests it should be fine.

### DIFF
--- a/core/src/wsv.rs
+++ b/core/src/wsv.rs
@@ -14,7 +14,7 @@ use iroha_crypto::HashOf;
 use iroha_data_model::{prelude::*, small::SmallVec};
 use iroha_logger::prelude::*;
 use iroha_telemetry::metrics::Metrics;
-use tokio::task;
+use tokio::{sync::broadcast, task};
 
 use crate::{
     block::Chain,
@@ -86,7 +86,7 @@ pub struct WorldStateView {
     /// Notifies subscribers when new block is applied
     new_block_notifier: Arc<NewBlockNotificationSender>,
     /// Transmitter to broadcast [`WorldStateView`]-related events.
-    events_sender: Option<EventsSender>,
+    events_sender: EventsSender,
 }
 
 impl Default for WorldStateView {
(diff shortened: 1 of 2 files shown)


core/src/wsv.rs
```diff
@@ -117,14 +117,9 @@ impl WorldStateView {
...
     pub fn new(world: World) -> Self {
-        Self::from_configuration(Configuration::default(), world)
-    }
-
-    /// Add the ability of emitting events to [`WorldStateView`].
-    #[must_use]
-    pub fn with_events(mut self, events_sender: EventsSender) -> Self {
-        self.events_sender = Some(events_sender);
-        self
+        // Added to remain backward compatible with other code primary in tests
```
appetrosyan, line 120: Actually people normally don't want to do everything by hand, so this is a good design.

appetrosyan (approved):

s8sato (approved):

Issue #2302, opened 2022-06-01
Querying: get domain triggers


0x009922:
## The Problem

In the Iroha Explorer, we need to show domain triggers in the frontend. Only its count for now.

Backend fetches domain(s) with `FindAllDomains` and `FindDomainById`, but there is no any triggers data.

## Possible solutions

- Add domain triggers data to `FindAllDomains` and `FindDomainById`. Actually... it is more relevant to `FindDomainById` and may be omitted for `FindAllDomains`
- Add another query like `FindDomainTriggersByDomainId`

appetrosyan:
Since we don't actually support domain-scoped triggers (#1889 which is blocked by #2062 which is going to take us at least a month to stabilise), I propose introducing a stop-gap solution where the query returns an empty struct, or a struct with a length identically equal to `0`.

pesterev:
@0x009922 hey, we've implemented and merged domain-scoped triggers. Now the `FindTriggersByDomainId` query should return an actual list of triggers.

PR #2418, opened 2022-06-30
[feature] #1889: Add domain-scoped triggers (#2302)

### LINKED ISSUE #1889: Introduce Domain-scoped event-triggers
It should be possible to define event triggers which can only affect entities within the same domain. 

This will have important performance implications, improves correctness and security.

pesterev:

### Description of the Change

The PR adds domain-scoped triggers that interact with entities within the same domain.

- [x] Add API
- [ ] Add restrictions to interaction with entities.

### Issue

Resolves #1889 #2302

### Benefits

Allows to create domain-scoped triggers.

### Possible Drawbacks

Seems none.


core/src/smartcontracts/isi/triggers.rs
```diff
@@ -81,6 +91,9 @@ pub mod isi {
             wsv: &WorldStateView,
         ) -> Result<(), Self::Error> {
             let trigger_id = self.object_id.clone();
+
+            wsv.world.domain_scoped_triggers.remove(&trigger_id);
```
appetrosyan, line 95: This is an important point. regardless of domain, triggers have unique IDs. So no matter whether a trigger is global, or specific to a domain, no two triggers can have the same ID. So my question is, do we consider the domain part of the ID?

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -35,6 +35,16 @@ pub mod isi {
...
             }
 
+            if let Some(ref domain_id) = new_trigger.action.domain {
+                let domain = wsv.domain(domain_id)?;
+
+                let trigger_id = new_trigger.id.clone();
+
+                wsv.world
+                    .domain_scoped_triggers
+                    .insert(trigger_id, domain.id().clone());
```
Arjentix, line 45: Let's not use `world` from `WorldStateVeiw` directly. Be design all instructions should try to use only `modify_*` methods to modify something inside wsv.

Arjentix: Also I see, that your domain-scoped triggers are stored in World's `TriggerSet` as ordinary triggers, because of the old code below.
That won't help us to increase performance for events filtering

data_model/src/trigger/mod.rs
```diff
@@ -226,6 +230,8 @@ pub mod action {
         /// account must already exist in order for `Register<Trigger>` to
         /// work.
         pub technical_account: crate::account::Id,
+        /// Domain linked to this trigger.
+        pub domain: Option<crate::domain::Id>,
```
Arjentix, line 234: I like the idea proposed by @appetrosyan in the chat. Let's use *id* part of the trigger to identify if it's a domain-scoped trigger or not

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -226,9 +239,33 @@ pub mod query {
...
-            Ok(vec![])
+        fn execute(&self, wsv: &WorldStateView) -> eyre::Result<Self::Output, Error> {
+            let domain_id = &self
+                .domain_id
+                .evaluate(wsv, &Context::new())
+                .map_err(|e| Error::Evaluate(format!("Failed to evaluate domain id. {}", e)))?;
+
+            let mut triggers = Vec::<Trigger<FilterBox>>::new();
+
+            for pair in wsv.world.domain_scoped_triggers.iter() {
```
Arjentix, line 250:
```suggestion
            for (trigger_id, domain_id) in wsv.world.domain_scoped_triggers.iter() {
```

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -226,9 +239,33 @@ pub mod query {
...
+                let trigger_id = pair.key();
+
+                if pair.value() == domain_id {
+                    // Can't use just `ActionTrait::clone_and_box` cause this will trigger lifetime mismatch
+                    #[allow(clippy::redundant_closure_for_method_calls)]
+                    let action = wsv
+                        .world
+                        .triggers
```
Arjentix, line 258:
```suggestion
                    let action = wsv
                        .triggers()
```

core/src/smartcontracts/isi/triggers.rs
Arjentix: So filter of domain-event should not try to match events from different domains (like global triggers). But right now it's working like global trigger.
And, I think, it should be fixed not in filters but in the data flow

Arjentix (changes requested):

core/src/smartcontracts/isi/triggers.rs
pesterev: Yes, a trigger id is unique in one domain scope.

So your suggest is product two types `TriggerId = Id x Option<DomainId>`, and the TriggerId structure will look like this globally, right?
```
pub struct Id {
    /// Name given to trigger by its creator.
    pub name: Name,
    /// Refer to already exists domain
    pub domain_id: Option<DomainId>,
}
```

core/src/smartcontracts/isi/triggers.rs
pesterev: that's pattern doesn't work unfortunately

data_model/src/trigger/mod.rs
```diff
@@ -226,6 +230,8 @@ pub mod action {
         /// account must already exist in order for `Register<Trigger>` to
         /// work.
         pub technical_account: crate::account::Id,
+        /// Domain linked to this trigger.
```
outoftardis, line 233:
```suggestion
        /// The domain linked to this trigger.
```

core/src/lib.rs
```diff
@@ -34,6 +34,9 @@ pub type PeersIds = dashmap::DashSet<<Peer as Identifiable>::Id>;
 /// Provides an API to work with collection of key([`DomainId`]) - value([`Domain`]) pairs.
 pub type DomainsMap = dashmap::DashMap<<Domain as Identifiable>::Id, Domain>;
 
+/// [`Trigger`](TriggerId) to [`Domain`](DomainId) relationship.
```
outoftardis, line 37: what about this:

```suggestion
/// Domain-scoped triggers are used to define a [`Trigger`](TriggerId) for a given [`Domain`](DomainId).
```

core/src/smartcontracts/isi/triggers.rs
pesterev: @appetrosyan please reply to this comment.

appetrosyan: Yes, effectively. The option is not necessary in general, it's just useful until the point where we put the trigger into the `Set`.

core/src/smartcontracts/isi/triggers.rs
pesterev: @Arjentix I've added the "matching" trigger's domain to the event's domain before adding actions to "matched_ids". It will allow avoiding "blocking_write" and extrа iterations over "matched_ids". What do you think about this approach?

data_model/src/trigger/set.rs
```diff
@@ -263,19 +290,22 @@ impl Set {
...
-    ) where
+    fn handle_event<F, E>(&self, triggers: &DashMap<Id, Action<F>>, event: &E)
+    where
         F: Filter<EventType = E>,
+        E: GenericEvent,
     {
         for entry in triggers {
+            let id = entry.key();
+
+            if id.domain_id == event.scoped_domain_id() {
```
Arjentix, line 305: Ok, now I see what `GenericTrait` was introduced for

data_model/src/events/pipeline.rs
```diff
@@ -130,13 +134,24 @@ impl Event {
...
+impl GenericEvent for Event {
+    fn scoped_domain_id(&self) -> Option<<crate::prelude::Domain as crate::Identifiable>::Id> {
+        self.domain_id.clone()
+    }
+
+    fn event_type(&self) -> super::EventType {
+        EventType::Pipeline
+    }
+}
+
```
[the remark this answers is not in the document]
Arjentix: Ok, I've got why we need `GenericEvent` implemented for all events... But it still looks like a workaround...

Arjentix:
@pesterev , I've got your idea. Now it should not itereate over useless events for concrete trigger.
But to do this you introduced `GenericTrait` which makes other events like `TimeEvent` and  `PipelineEvent` think about *domain id* which is a completely unrelated thing to this events.

I have another idea. Let's do this *id* checking not inside `trigger::Set::hande_generic_event()` but inside `WorldStateView`. `wsv` knows if it's data event. Only data events can be domain-scoped.

data_model/src/events/pipeline.rs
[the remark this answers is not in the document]
pesterev: The alternative way is to keep the domain id inside `Filter` and `Event` and match them in the `match()` method. But it is difficult because `Filter` doesn't know about the domain and trigger.

Arjentix:
@pesterev , `DomainEvent` is a data event and it already contains domain id. I'm not fully understand what specific for `DomainEvent` you want to do

data_model/src/trigger/set.rs
```diff
@@ -221,15 +240,60 @@ impl Set {
...
+            _ => None,
+        };
+
+        for entry in self.data_triggers.iter() {
+            let id = entry.key();
+
+            // Check only if the trigger is domain scoped
+            if id.domain_id.is_some() && id.domain_id == data_domain_id {
+                continue;
+            }
```
Arjentix, line 259:
```suggestion
            // Check only if the trigger is domain scoped
            if id.domain_id.is_some() && id.domain_id != data_domain_id {
                continue;
            }
```

Shouln't it be like this?

data_model/src/trigger/set.rs
```diff
@@ -221,15 +240,60 @@ impl Set {
...
+
+            if let Repeats::Exactly(atomic) = &action.repeats {
+                if atomic.get() == 0 {
+                    continue;
+                }
+            }
+
+            task::block_in_place(|| self.matched_ids.blocking_write())
+                .push((EventType::Data, id.clone()))
```
Arjentix, line 273: Let's move this part in some function? It repeats for 3 times

pesterev:
@Arjentix, I finally got your idea to use `Identifiable` to retrieve `DomainId` from an event, and now changes look pretty good, thanks.

data_model/src/trigger/set.rs
```diff
@@ -120,14 +120,61 @@ impl Set {
...
+                    EventType::ExecuteTrigger => self
+                        .by_call_triggers
+                        .get(id)
+                        .map(|entry| ActionTrait::clone_and_box(entry.value()))
+                        .expect(
+                            "`Set::by_call_triggers` doesn't contain required id. This is a bug",
+                        ),
+                };
+
+                (trigger_domain_id == domain_id).then(|| f(id, action))
```
Arjentix, line 165: This looks elegant, but I think it's better to place this check before `match event_type` to not to search for trigger that won't be executed

data_model/src/trigger/set.rs
```diff
@@ -120,14 +120,61 @@ impl Set {
         self.ids.iter().map(|entry| entry.key().clone()).collect()
     }
 
+    /// Apply `f` to triggers that belong to the given [`DomainId`]
+    ///
+    /// Returns an empty list if [`Set`] doesn't contain any trigger belonging to [`DomainId`].
+    pub fn inspect_by_domain_id<F, R>(&self, domain_id: &DomainId, f: F) -> Vec<R>
+    where
+        F: Fn(&Id, Action<FilterBox>) -> R,
```
Arjentix, line 128:
```suggestion
        F: Fn(&Id, &dyn ActionTrait) -> R,
```

This is how it's done in `inspect_by_id()`. Why not to use the same technique here to prevent extra cloning

data_model/src/trigger/set.rs
```diff
@@ -258,38 +329,26 @@ impl Set {
...
     ///
-    /// Finds all actions, that are triggered by `event` and stores them.
-    /// This actions will be inspected ln the next [`Set::inspect_matched()`] call
-    pub fn handle_execute_trigger_event(&self, event: &ExecuteTriggerEvent) {
-        self.handle_event(&self.by_call_triggers, event, EventType::ExecuteTrigger)
-    }
-
-    /// Handle generic event
-    fn handle_event<F, E>(
+    /// Skips insertion if the action's filter doesn't match an event.
```
Arjentix, line 334:
```suggestion
    /// Skips insertion:
    /// - If the action's filter doesn't match an event
    /// - If the action's repeats count equals to 0
```

data_model/src/trigger/set.rs
pesterev: I thought about it but can't compile it due to lifetimes. That's why I added boxing. Actually, it will box and clone anyway eventually when calling this function.

data_model/src/trigger/mod.rs
```diff
@@ -165,6 +163,61 @@ impl Identifiable for Trigger<FilterBox> {
...
+        let vector: Vec<&str> = s.split('$').collect();
+
+        if vector.len() == 1 {
+            return Ok(Self {
+                name: Name::from_str(vector[0])?,
+                domain_id: None,
+            });
+        }
+
+        if vector.len() != 2 {
```
Erigara, line 210: IMO we could check `vector.len() > 2` and put this check above creation of `Id`.
This way we would have more clear logic we check errors first and then actually create `Id`.

data_model/src/trigger/set.rs
```diff
@@ -120,14 +120,65 @@ impl Set {
...
+                let id = pair.key();
+                let trigger_domain_id = id.domain_id.as_ref()?;
+
+                if trigger_domain_id != domain_id {
+                    return None;
+                }
+
+                let event_type = pair.value();
+
+                let result = match event_type {
```
Erigara, line 142: I think we could get rid of `result` variable and wrap `match` in `Some` directly.

Erigara (commented):
Looks good, fix doc comments and i will approve.

data_model/src/trigger/mod.rs
```diff
@@ -165,6 +163,61 @@ impl Identifiable for Trigger<FilterBox> {
...
+impl FromStr for Id {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.is_empty() {
+            return Err(ParseError {
+                reason: "Trigger ID cannot be empty",
+            });
+        }
+
```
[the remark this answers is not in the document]
pesterev: I will left check an empty string to give the user a more detailed error:
```rust
        if s.is_empty() {
            return Err(ParseError {
                reason: "Trigger ID cannot be empty",
            });
        }
        let mut split = s.split('$');
        let name = split.next();
        let domain = split.next();
        let tail = split.next();
        if tail.is_some() {
            return Err(ParseError {
                reason: "Trigger ID should have format `name` or `name$domain_id`",
            });
        }
        if let Some(name) = name {
            Ok(Self {
                name: Name::from_str(name)?,
                domain_id: domain.map(<Domain as Identifiable>::Id::from_str).transpose()?,
            })
        } else {
            Err(ParseError {
                reason: "Trigger ID cannot be empty",
            })
        }
    }
```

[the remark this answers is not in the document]
Erigara: if `name` is `None` it's automatically means that `s` is empty, no?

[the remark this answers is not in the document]
Erigara: Oh, sorry, you are right, split on empty string will always return empty string as first element.
I guess then last `else` could be `unreachable!`.

```rust
        if s.is_empty() {
            return Err(ParseError {
                reason: "Trigger ID cannot be empty",
            });
        }
        let mut split = s.split('$');
        let name = split.next();
        let domain = split.next();
        let tail = split.next();
        if tail.is_some() {
            return Err(ParseError {
                reason: "Trigger ID should have format `name` or `name$domain_id`",
            });
        }
        if let Some(name) = name {
            Ok(Self {
                name: Name::from_str(name)?,
                domain_id: domain.map(<Domain as Identifiable>::Id::from_str).transpose()?,
            })
        } else {
           unreachable!();
        }
    }
```

Erigara (approved):

data_model/src/trigger/mod.rs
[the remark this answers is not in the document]
pesterev: Eventually decided to use pattern matching.

Arjentix (approved):

PR #2421, opened 2022-06-30
[fix] #2420: Fix "Permission already exists" bug

### LINKED ISSUE #2420: Fix `Permission already exists` bug
We've encountered a bug: Iroha rejects transactions with permission granting, saying, that it already exists for given user, when it's actually not.

I've found, that the problem is in how `PermissionToken` is sorted in `BTreeSet`. More particularly the problem is in `PartialOrd` and `Ord` implementation for `PermissionToken`.

Need to fix it and write some tests to check if it works

Arjentix:
### Description of the Change

- `PermissonToken` now derives `PartialOrd` and `Ord`. Manual implementation that was using only token name used to cause collision and is the reason of the bug
- `PartialOrd` and `Ord` were derived for many other structs that `PermissionToken` depends on
- Added test to check this bug

### One more *dangerous* bug was found and fixed with the same changes

While fixing this bug I have also found another bug which this PR also fixes.
`PermissionToken` collision used to make it possible to i.e. have a permission to modify only one asset, but the same permission was enough to modify any other asset.
The test I introduced checks that too.

### Issue

- Closes #2420 

### Benefits

- Permission collision bug fixed
- One more test

### Possible Drawbacks

None

### DIFF
--- a/client/tests/integration/permissions.rs
+++ b/client/tests/integration/permissions.rs
@@ -266,13 +266,13 @@ fn permissions_differ_not_only_by_names() {
         .expect("Failed to modify Mouse's hats");
 
     // ... but not shoes
-    let mouse_shoes_id = <Asset as Identifiable>::Id::new(shoes_definition_id, mouse_id.clone());
+    let mouse_shoes_id = <Asset as Identifiable>::Id::new(shoes_definition_id, mouse_id);
     let set_shoes_color = SetKeyValueBox::new(
         mouse_shoes_id.clone(),
         Name::from_str("color").expect("Valid"),
         "yellow".to_owned(),
     );
-    let _ = client
+    let _err = client
         .submit_blocking(set_shoes_color.clone())
         .expect_err("Expected Alice to fail to modify Mouse's shoes");
 
--- a/client/tests/integration/permissions.rs
+++ b/client/tests/integration/permissions.rs
@@ -3,9 +3,17 @@
 use std::{str::FromStr as _, thread};
 
 use iroha_client::client::{self, Client};
-use iroha_core::smartcontracts::isi::permissions::combinators::DenyAll;
+use iroha_core::{
+    prelude::ValidatorBuilder,
+    smartcontracts::{
+        isi::permissions::combinators::DenyAll, permissions::combinators::ValidatorApplyOr as _,
+    },
+};
 use iroha_data_model::prelude::*;
-use iroha_permissions_validators::{private_blockchain, public_blockchain};
+use iroha_permissions_validators::{
+    private_blockchain,
+    public_blockchain::{self, key_value::CanSetKeyValueInUserAssets},
+};
 use test_network::{PeerBuilder, *};
 
 use super::Configuration;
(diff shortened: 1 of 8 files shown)


mversic (commented):
I recently removed `PartialOrd` and `Ord` from most of these structures. What does it mean that you can order `Equal` or `Subtract` or `SignatureConditionCheck` and alike? Also, once you make `Value` to be `Ord` it just forces every type it wraps to be as well. Does it need that requirement?

I don't think I touched the `Ord` implementation of `PermissionToken` - that one made sense.

data_model/src/lib.rs
```diff
@@ -286,7 +286,18 @@ pub type ValueBox = Box<Value>;
...
+    Clone,
+    PartialEq,
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    FromVariant,
+    IntoSchema,
+    PartialOrd,
```
appetrosyan, line 299: I'm not sure if total ordering of the catch-all Value type makes sense, but I'm guessing that we need this for another upstream type?

mversic:
my argument for removing them was that they are somewhat of a cheat/lie in many places even though they can, formally speaking, be ordered. Anyways, I don't mind if they stay

Arjentix:
Yeah, making value to be `Ord` was quite painful. But how can we properly implement `Ord` for `PermissionToken` without it?
I assume, we can use HashSet instead of BTreeSet. Then requirements will be different.
But I don't see a problem with `Ord` for now

data_model/src/permissions.rs
```diff
@@ -34,6 +34,8 @@ pub type Permissions = btree_set::BTreeSet<PermissionToken>;
     Deserialize,
     Serialize,
     IntoSchema,
+    PartialOrd,
```
Erigara, line 37: Crucial part, cmp not only by name, but also by params.

Arjentix:
Or we can force `PermissionToken` to store `String` or `Name` instead of `Value`

mversic (approved):

Erigara (approved):

Arjentix:
> my argument for removing them was that they are somewhat of a cheat/lie in many places even though they can, formally speaking, be ordered

I agree, there is a little smell... But it looks like the simplest solution. We can refactor it after the merge window closes. But I don't know what we can change here to make it working

Erigara:
> > my argument for removing them was that they are somewhat of a cheat/lie in many places even though they can, formally speaking, be ordered
> 
> I agree, there is a little smell... But it looks like the simplest solution. We can refactor it after the merge window closes. But I don't know what we can change here to make it working

same feeling, not really make sense to compare `Equal`s, but most straightforward solution

client/tests/integration/permissions.rs
```diff
@@ -201,3 +209,83 @@ fn permissions_checked_before_transaction_execution() {
 
     assert!(root_cause.contains("Account does not have the needed permission token"));
 }
+
+#[test]
+fn permissions_differ_not_only_by_names() {
```
appetrosyan, line 214: Good testing style: a scenario is more important to test than specific behaviour.

appetrosyan (approved):

PR #2424, opened 2022-07-01
[ci]: Fix docker build

### LINKED ISSUE #2345: Create new branch and add CI so that it can be used for LTS release-related activities.

appetrosyan:

### Description of the Change

- Remove cargo chef
- Add workflow for LTS release
- Add `cargo deny` configuration

### Issue

Closes #2345

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,56 +1,327 @@
 Changelog
 =========
 
-2.0.0-pre-rc.5-lts
-------------------
+
+2.0.0-pre.rc.6-lts
+---------
 
 Features
 ~~~~~~~~
+- 2395 Add panic if genesis cannot be applied.
+- 2000 Disallow empty names. #2000: Disallow empty names
+- 2127 Add sanity check to ensure that all data decoded by
+  `parity_scale_codec` is consumed.
+- 2360 Make `genesis.json` optional again.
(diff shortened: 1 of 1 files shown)


Erigara (approved):
Looks good.
One thing i noticed is that we are not pinning versions which go against docker best practices (our builds became not-reproducible).
Integration of docker linting tool like [hadolint](https://github.com/hadolint/hadolint) in our CI could improve quality and stability of builds.

appetrosyan:
> Looks good.


> One thing i noticed is that we are not pinning versions which go against docker best practices (our builds became not-reproducible).


Unfortunately, the non-pinning of versions is intentional. 
1) We prevent downstream users from using any version other than `dev`, `stable`, and `lts`. 
2) Dependencies cannot be pinned anyway, because we rely on the package manager, which can fetch a different version of a toolchain, or the toolchain may get removed. 
3) And this is the most important point, we shall be pinning dependencies in the new CI/CD pipeline. This is a stop-gap so that we don't need to merge an untested pipeline in the middle of a release. 

> Integration of docker linting tool like [hadolint](https://github.com/hadolint/hadolint) in our CI could improve quality and stability of builds.

@Erigara create an issue.

s8sato (approved):

PR #2428, opened 2022-07-01
[feature] #2228: Add Unauthorized variant to smartcontracts query error

### LINKED ISSUE #2228: Introduce Unauthorized `QueryError` when supplied Account ID is not found
## Description

Currently, if you try to make a query to Iroha with an account that is not found by Iroha, then `QueryError::FindError` is returned. `FindError` seems to be more about something *requested* is not found.

I'd suggest to introduce a new `QueryError` that will share semantics with HTTP 401 "Unauthorized" error.

ales-tsurko:
Closes #2228

### Description of the Change

Added `iroha_core::smartcontracts::query::Error::Unauthorized` to better mirror semantics of unauthorized request (i.e. without an account in this case). Also, in case of querying without an account, **HTTP 401 UNAUTHORIZED** status is returned.

### Issue

Previously, in case of query without an account `FindError::Account` returned. And the corresponding **HTTP 404 NOT_FOUND** status code returned. That appeared to be semantically incorrect.

### Benefits

Now `POST /query` in case of request without an account returns `Error::Unauthorized` and **HTTP 401 UNAUTHORIZED** status code.

At first glance, this can be moved to `Error::Permission` and the `Unauthorized` variant could be added to `DenialReason`, but it doesn't share semantics with **HTTP 401 UNAUTHORIZED** (it's **HTTP 403 FORBIDDEN** instead, [which is different](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403)).

### DIFF
--- a/core/src/smartcontracts/isi/account.rs
+++ b/core/src/smartcontracts/isi/account.rs
@@ -32,15 +32,11 @@ pub mod isi {
 
             match wsv.asset(asset_id) {
                 Err(err) => match err {
-                    QueryError::Find(ref find_err) => {
-                        if let FindError::Asset(_) = **find_err {
-                            assert_can_register(&asset_id.definition_id, wsv, self.object.value())?;
-                            wsv.asset_or_insert(asset_id, self.object.value().clone())
-                                .expect("Account exists");
-                            Ok(())
-                        } else {
-                            Err(err.into())
-                        }
+                    QueryError::Find(find_err) if matches!(*find_err, FindError::Asset(_)) => {
+                        assert_can_register(&asset_id.definition_id, wsv, self.object.value())?;
+                        wsv.asset_or_insert(asset_id, self.object.value().clone())
+                            .expect("Account exists");
+                        Ok(())
                     }
                     _ => Err(err.into()),
                 },
--- a/core/src/smartcontracts/isi/account.rs
+++ b/core/src/smartcontracts/isi/account.rs
@@ -13,7 +13,10 @@ use crate::{ValidQuery, WorldStateView};
 /// - grant permissions and roles
 /// - Revoke permissions or roles
 pub mod isi {
-    use super::{super::prelude::*, super::query::Error as QueryError, *};
+    use super::{
+        super::{prelude::*, query::Error as QueryError},
+        *,
+    };
 
     #[allow(clippy::expect_used, clippy::unwrap_in_result)]
     impl Execute for Register<Asset> {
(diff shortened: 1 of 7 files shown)


appetrosyan (dismissed):
After merge add it to #api_changes

core/src/wsv.rs
```diff
@@ -131,8 +134,9 @@ impl WorldStateView {
     ///
     /// # Errors
     /// Fails if there is no domain or account
-    pub fn account_assets(&self, id: &AccountId) -> Result<Vec<Asset>, FindError> {
+    pub fn account_assets(&self, id: &AccountId) -> Result<Vec<Asset>, QueryError> {
         self.map_account(id, |account| account.assets().cloned().collect())
+        // .map_err(Into::into)
```
Erigara, line 139: commented code

Erigara (changes requested):
Will approve after removing commented code

core/src/wsv.rs
```diff
@@ -560,11 +564,9 @@ impl WorldStateView {
         &self,
         id: &AccountId,
         f: impl FnOnce(&Account) -> T,
-    ) -> Result<T, FindError> {
+    ) -> Result<T, QueryError> {
         let domain = self.domain(&id.domain_id)?;
-        let account = domain
-            .account(id)
-            .ok_or_else(|| FindError::Account(id.clone()))?;
+        let account = domain.account(id).ok_or(QueryError::Unauthorized)?;
```
pesterev, line 569: it looks strange to me that not being able to find the account means the account is not authorized.

core/src/smartcontracts/isi/account.rs
```diff
@@ -28,13 +31,19 @@ pub mod isi {
...
+                    QueryError::Find(ref find_err) => {
+                        if let FindError::Asset(_) = **find_err {
+                            assert_can_register(&asset_id.definition_id, wsv, self.object.value())?;
+                            wsv.asset_or_insert(asset_id, self.object.value().clone())
+                                .expect("Account exists");
+                            Ok(())
+                        } else {
+                            Err(err.into())
+                        }
+                    }
```
pesterev, line 44:
```suggestion
                    QueryError::Find(find_err) if matches!(*find_err, FindError::Asset(_)) => {
                        assert_can_register(&asset_id.definition_id, wsv, self.object.value())?;
                        wsv.asset_or_insert(asset_id, self.object.value().clone())
                            .expect("Account exists");
                        Ok(())
                    }
```

pesterev (commented):
LGTM but left comment that suggests one minor improvement

Erigara (approved):

PR #2431, opened 2022-07-04
[feature] #2061: Derive macro for filters

### LINKED ISSUE #2061: Ensure all Event types have a correponding filter in the `data_model`
Could be implemented either by adding a test, or automatically generating filters for simple events.

ilchu:

### Description of the Change
To avoid having to update `Filter` structs manually each time `Events` are changed or vice versa, which is error-prone and tedious, have a derive procedural macro to do that instead, and also automatically generate `...Filter` structs, `...EventFilter` enums and their `impl Filter` methods.

### Issue
Closes #2061.

### Benefits
* Greatly simplifies the `data_model/src/events/data/filters.rs` file, leaving only common types and `EntityFilter` there.
* Removes the need for unit testing on event fields exhaustiveness.
* Makes a step toward domain-oriented hierarchy of the data model instead of the functional one, which is what was agreed upon in discussion of its structure.

### Possible Drawbacks
The macro is dependent on the conventions that have been adopted in our codebase so far, such as events being enums and having only tuple fields with one member, the latter being either some `Id` or `Event`. Thus it might become very brittle should any of those assumptions change.

### Usage Examples or Tests
Codegen can be checked by putting the `data_model/.../events.rs` file into crate's `tests` folder and running `cargo expand --test events` on it (after fixing some imports).

### Alternate Designs
The current design was selected upon comparison with the approach where filters would still be generated in the `filters.rs` file where they resided before. This option would require bringing the `Event` enum into scope somehow, and I found no other option besides reading the AST of the whole `events.rs` file, which would be *very* ugly. As @appetrosyan proposed, there could be some crates that would help with this, but in the end the current solution is cleaner, simpler and makes a lot of sense structurally.


data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    } = ast;
+
+    let event_filter_ident = format_ident!("{}Filter", event_ident);
+
+    let filter_ident = format_ident!(
+        "{}Filter",
+        event_ident
+            .to_string()
+            .strip_suffix("Event")
+            .expect("Events should follow the naming format")
```
appetrosyan, line 37: OK, so we need to call it `VariantEvent` and it becomes `VariantFilter`, right? This is a good convention. Some examples would be welcome, in the docs.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let filter_doc = format!("Filter for {} entity", filter_ident);
+    let new_doc = format!("Construct new {}", filter_ident);
+
+    quote! {
+        #[derive(
+            Clone,
```
appetrosyan, line 114: I'm in favour of enforcing invariants pre-emptively. Implementing as many traits as possible is also considered a good practice.

appetrosyan (dismissed):

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                    Some(quote! {
+                        #event_filter_variant_name (crate::prelude::FilterOpt<#inner_filter_ident>)
+                    })
+                } else {
+                    // Default case, make a unit variant with just the ident
+                    filter_variants_id_fields.push(event_filter_variant_name);
+                    None
+                }
+            }
+            else { None }
```
appetrosyan, line 76: [I'd consider a monadic approach. ](https://doc.rust-lang.org/std/primitive.bool.html#method.then)

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                // Only `...Event` containing fields are wrapped with `FilterOpt`
+                if var_ty_ident.to_string().contains("Event") {
+                    // Also store just the variant ident for later codegen
+                    filter_variants_idents_event_fields.push(event_filter_variant_name.clone());
+                    let inner_filter_ident = format_ident!(
+                        "{}Filter",
+                        var_ty_ident.to_string().strip_suffix("Event").expect("Checked in outer if"),
+                    );
+                    Some(quote! {
+                        #event_filter_variant_name (crate::prelude::FilterOpt<#inner_filter_ident>)
```
appetrosyan, line 68: On the one hand the prelude is not meant to be imported directly. 

On the other hand, this is the only fixed public API element, so having `crate::prelude::FilterOpt` is the most stable path.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    let filter_variants_event_fields = event_variants.iter().filter_map(|variant| {
+        let event_filter_variant_name = format_ident!("By{}", variant.ident);
+        if let syn::Fields::Unnamed(ref unnamed) = variant.fields {
+            let variant_type = &unnamed
+                .unnamed
+                .first()
+                .expect("Should have at least one type")
+                .ty;
+            if let syn::Type::Path(path) = variant_type {
+                let var_ty_ident = &path.path.segments[0].ident;
```
appetrosyan, line 58: sus

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                // Only `...Event` containing fields are wrapped with `FilterOpt`
+                if var_ty_ident.to_string().contains("Event") {
```
appetrosyan, line 60: That's not entirely correct. 

`SomethingEventFilter` also contains `Event` and a user might mistakenly add the derive macro, thinking that they `#[derive(Filter…)]` on a filter, instead of the `Event`. 

```suggestion
                if var_ty_ident.to_string().ends_with("Event") {
```

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                    // Also store just the variant ident for later codegen
+                    filter_variants_idents_event_fields.push(event_filter_variant_name.clone());
+                    let inner_filter_ident = format_ident!(
+                        "{}Filter",
+                        var_ty_ident.to_string().strip_suffix("Event").expect("Checked in outer if"),
```
appetrosyan, line 65:
```suggestion
                        var_ty_ident.to_string().strip_suffix("Event").expect("Variant name should have suffix `Event`"),
```

you could technically use `unreachable!` here, but if you're panicking anyway, might as well treat all error conditions as normal errors.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let filter_variants_event_fields = event_variants.iter().filter_map(|variant| {
+        let event_filter_variant_name = format_ident!("By{}", variant.ident);
+        if let syn::Fields::Unnamed(ref unnamed) = variant.fields {
+            let variant_type = &unnamed
+                .unnamed
+                .first()
+                .expect("Should have at least one type")
+                .ty;
+            if let syn::Type::Path(path) = variant_type {
```
appetrosyan, line 57: Good candidate to be extracted into a separate `fn`

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+/// imports. This macro is also dependent on the naming conventions adopted so far, such as that
+/// `Event` enums always have tuple variants with either some sort of `Id` or another `Event` inside
+/// of them.
+#[proc_macro_derive(Filter)]
+pub fn filter_derive(input: TokenStream) -> TokenStream {
+    let ast = syn::parse(input).expect("Failed to parse input TokenStream.");
+    impl_from_filter(&ast)
+}
+
+fn impl_from_filter(ast: &syn::DeriveInput) -> TokenStream {
```
appetrosyan, line 20: Overall, I think this function should be broken up into smaller pieces. More `structs` could help you with parsing, so that illegal states (Events without the `Event` suffix) are unrepresentable. It's good enough for a start.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+            else { None }
+        }
+        else { None }
+    }).collect::<Vec<_>>();
+
+    let filter_impl_variant_pairs_event_fields = filter_variants_idents_event_fields
+        .iter()
+        .zip(event_variants.iter())
+        .map(|(filter_variant_ident, event_var)| {
+            let event_var_ident = format_ident!("{}", &event_var.ident);
```
appetrosyan, line 85: OK. I normally find `format.*!("{}"` a little suspicious, always prefer writing what the macro expands to, which here is actually probably just equivalent to a `clone`.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        )
+        .map(|(filter_var, event_var)| {
+            let event_var_ident = format_ident!("{}", &event_var.ident);
+            quote! {
+                (Self::#filter_var, crate::prelude::#event_ident::#event_var_ident (_))
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let filter_doc = format!("Filter for {} entity", filter_ident);
```
appetrosyan, line 109: These probably should be expanded. Good enough for now.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        #[doc = #filter_doc]
+        #vis struct #filter_ident #generics {
+            id_filter: crate::prelude::FilterOpt<crate::prelude::IdFilter<<crate::prelude::#event_ident as Identifiable>::Id>>,
+            event_filter: crate::prelude::FilterOpt<#event_filter_ident>
+        }
+
+        impl #filter_ident {
+            #[doc = #new_doc]
+            pub const fn new(
+                id_filter: crate::prelude::FilterOpt<crate::prelude::IdFilter<<crate::prelude::#event_ident as Identifiable>::Id>>,
```
appetrosyan, line 136: Recommend using variables to hold long paths.

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
+//! Crate with `Filter` and `EventFilter` derive macro
+#![allow(clippy::expect_used, clippy::panic, clippy::too_many_lines)]
+
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+
+/// [`Filter`] is used for code generation of `...Filter` structs and `...EventFilter` enums, as well as
+/// implementing the `Filter` trait for both of them. This macro should only be attributed to `Event` enums.
+/// Due to name scoping, the macro currently only
+/// expands properly from within the `iroha_data_model` crate as it relies on a few of `crate::prelude`
```
outoftardis, line 10:
```suggestion
/// Due to name scoping, the macro currently properly
/// expands only from within the `iroha_data_model` crate as it relies on `crate::prelude`
```

data_model/derive/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+/// imports. This macro is also dependent on the naming conventions adopted so far, such as that
```
outoftardis, line 11:
```suggestion
/// imports. This macro also depends on the naming conventions adopted so far, such as that
```

data_model/derive/src/lib.rs
ilchu: Fiddled around with it a bit, but couldn't find a way to avoid if-else chains entirely (only some of it in the new extracted function), as destructuring with `if let` requires an `else` branch if we're returning this expression's results.

data_model/derive/src/lib.rs
ilchu: You mean this could be brittle?

Erigara (dismissed):
Good solution, i think readability could be slightly improved it will be easier to change this code in the future, but it's up to you.

Erigara (approved):

appetrosyan (approved):

PR #2433, opened 2022-07-04
[ci]: Update Dockerfiles & Codecov reports reuse & Concurrency

BAStos525:
Signed-off-by: Vasilii Zyabkin [vzyabkin@soramitsu.co.jp](mailto:vzyabkin@soramitsu.co.jp)

### Description of the Change
1. Update iroha2 Dockerfiles for ci and switch to musl architecture for building processes.
2. Update workflows for new Dockerfiles.
3. Update jobs for codecov coverage reports generation.
4. Add concurrency feature to stop previous running merged & PR workflows after pushing a new commit.

### Issue
1. [x] We need to update iroha2 ci process in order to improve it, make faster and get rid of some bottlenecks.
2. [ ] Codecov coverage reports takes so much time, so we have to try to reuse the same reports in the case if repository state is the same.
3. [x] Pushing a new commits into branch doesn't stop previously running workflows and jobs,

### Benefits
1. Reduce the size of images, switch to alpine and Arch base images, increase the speed of building process.
2. Implement the process of iroha-2 CI the reuse of codecov reports for the same state of the repository for which GH Actions workflow is running.
3. Add `concurrency` Actions feature to stop previous running outdated workflows.

### Possible Drawbacks
- Some possibility of failing compilation on any building stages is possible.
- Codecov coverage might not print coverage diff between PR and Merge reports.


.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -12,6 +12,10 @@ on:
       # Not part of the workspace
       - "!wasm/**"
 
+concurrency: 
+  group: I2::Dev::Static Analysis-${{ github.head_ref || github.run_id }}
```
s8sato, line 16: Since this workflow runs on `pull_request` and `github.head_ref` will always be defined, the expression will never fallback to `github.run_id` (whose granularity will be too small, by the way). And `github.head_ref` will be too large granularity because it represents the source branch.

I think your intention is that a runner should be single to the combination of the workflow and the pull request:

```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

The same for other `*pr.yml`

.github/workflows/iroha2-dev.yml
```diff
@@ -7,6 +7,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: I2::Dev::Deploy-${{ github.workflow }}-${{ github.ref }}
```
s8sato, line 11: `I2::Dev::Deploy` and `${{ github.workflow }}` are duplicates.

```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

The same for other `*.yml`

s8sato (changes requested):
Just a quick look into the 4th change (`concurrency`). I'll continue to review

appetrosyan:
The commits need to be signed off as well.

s8sato (commented):
Comments for the 3rd change (coverage)

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -44,24 +48,49 @@ jobs:
...
     steps:
-      # TODO Remove these steps #2165
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Find hash
+        id: hash
+        uses: theowenyoung/folder-hash@v2.0.1
+        with:
+          path: |
+            .
```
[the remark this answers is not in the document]
BAStos525: Should we remove all steps related to codecov reports files hash identification & comparison and artifacts uploading & downloading as well?

s8sato (changes requested):
Comments for the 1st (Dockerfiles) and the 2nd (workflows) changes. I've finished the review once

Cargo.toml
```diff
@@ -4,7 +4,6 @@ members = [
   "actor",
   "actor/derive",
   "cli",
-  "cli/derive",
```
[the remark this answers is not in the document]
appetrosyan: Why indeed.

Dockerfile
```diff
@@ -1,51 +1,21 @@
...
-        clang \
-        pkg-config \
-        llvm-dev; \
-    rm -rf /var/lib/apt/lists/*
+RUN  set -ex && \
+     apk --update add curl ca-certificates && \
+     adduser --disabled-password iroha --shell /bin/bash --home /app && \
+     mkdir -p $CONFIG_DIR && \
+     mkdir /chain && \
+     chown iroha:iroha /chain
```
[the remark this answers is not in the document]
appetrosyan: Storage sounds more explicit

docker-compose-single.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: iroha0:debug
     volumes:
       - './configs/peer:/config'
-      - './:/root/soramitsu/iroha'
+      - ./:/root/soramitsu/iroha
```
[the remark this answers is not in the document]
appetrosyan: No. I'd recommend wrapping more things in single quotes.

.github/workflows/iroha2-dev.yml
```diff
@@ -23,71 +27,48 @@ jobs:
...
-          tags: hyperledger/iroha2:kagami-dev
-          build-args: |
-            TARGET_DIR=release
-            PROFILE=--release
-            BIN=kagami
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: push iroha2:dev docker image
+        run: |
+          docker push hyperledger/iroha2:dev
```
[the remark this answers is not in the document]
appetrosyan: There's also non-negligible overhead. The actions assume that docker is not installed and not configured. This flies in the face of including the entire infrastructure at the level of the `iroha2:env` image.

Cargo.toml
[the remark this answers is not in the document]
BAStos525: Oh, it wasn't in [the base repo](https://github.com/soramitsu/soramitsu-iroha/blob/feature/DOPS-1789/Cargo.toml) that a forked

.github/workflows/iroha2-dev.yml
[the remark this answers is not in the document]
BAStos525: I suggest to move this issue to another new PR. So, to create a new ticket for DevOps team and a corresponding issue for this task. It will be combined with a new workflow file for `iroha2:env (base)` image building.

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
BAStos525: This approach is based on creating the unique codecov coverage report according to the repository state. This should allows to reuse codevov coverage report for the same repository (PR) state. So, confirm again if it's not needed.

Dockerfile.ci
```diff
@@ -1,67 +1,20 @@
...
+RUN rustup toolchain add nightly-2022-04-20-x86_64-unknown-linux-musl
+RUN rustup component add rust-src
+RUN rustup component add rust-src --toolchain nightly-2022-04-20-x86_64-unknown-linux-musl
+RUN rustup component add rustfmt --toolchain nightly-2022-04-20-x86_64-unknown-linux-musl
+RUN rustup component add rust-std
+# RUN rustup target add wasm32-unknown-unknown
+RUN rustup target add wasm32-unknown-unknown --toolchain nightly-2022-04-20
+RUN cargo install cargo-lints cargo-nono webassembly-test-runner
+RUN rustup target add x86_64-unknown-linux-musl
+RUN pacman -S musl rust-musl --noconfirm
```
pesterev, line 20: Is it OK that needs a separate RUN for each line here? RUN creates a new layer and increases the size of the image.

tools/parity_scale_decoder/Cargo.toml
```diff
@@ -30,4 +30,4 @@ iroha_schema_gen = { version = "=2.0.0-pre-rc.5", path = "../../schema/gen"}
 iroha_data_model = { version = "=2.0.0-pre-rc.5", path = "../../data_model", features = ["warp", "mutable_api"]}
 parity-scale-codec = { version = "2.3.1", default-features = false }
 serde_json = "1.0"
-serde = "1.0"
+serde = "1.0"
```
pesterev, line 33: needs a newline here

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
s8sato: Assuming the same repository state means the same folder hash, that means the `on.pull_request.paths` is not triggered, which means the entire workflow is skipped. If so, when will the report be reused?

docker-compose-single.yml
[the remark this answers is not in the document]
s8sato: Either is acceptable to me

.github/workflows/iroha2-dev.yml
```diff
@@ -23,71 +27,48 @@ jobs:
...
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - name: Build and push docker image
-        uses: docker/build-push-action@v2
+          driver: docker
+
+      - name: Build iroha2:env container
+        uses: docker/build-push-action@v3
         with:
+          load: true
+          tags: iroha2:env
```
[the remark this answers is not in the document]
s8sato: Also, since each Dockerfile corresponds to each build stage at the moment, `Dockerfile.base` instead of `Dockerfile.ci` would be more concise

.github/workflows/iroha2-dev.yml
```diff
@@ -19,75 +23,52 @@ jobs:
...
+          tags: iroha2:base
           context: .
-          push: true
-          tags: hyperledger/iroha2:dev
-          build-args: |
-            TARGET_DIR=release
-            PROFILE=--release
-      - name: Build and push docker image (iroha_client_cli)
-        uses: docker/build-push-action@v2
+          file: Dockerfile.ci
```
s8sato, line 38:
```suggestion
          file: Dockerfile.ci
```

The same for other 2 ymls

.github/workflows/iroha2-dev.yml
```diff
@@ -19,75 +23,52 @@ jobs:
...
-          tags: hyperledger/iroha2:client-cli-dev
-          build-args: |
-            TARGET_DIR=release
-            PROFILE=--release
-            BIN=iroha_client_cli
-      - name: Build and push docker image (kagami)
-        uses: docker/build-push-action@v2
+          file: Dockerfile.build
+
+      - name: build iroha2:dev docker image
```
s8sato, line 48:
```suggestion
      - name: Build iroha2:dev docker image
```

The same for `iroha2.yml`

.github/workflows/iroha2-dev.yml
```diff
@@ -19,75 +23,52 @@ jobs:
...
-          context: .
-          push: true
-          tags: hyperledger/iroha2:kagami-dev
-          build-args: |
-            TARGET_DIR=release
-            PROFILE=--release
-            BIN=kagami
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: push iroha2:dev docker image
```
s8sato, line 56:
```suggestion
      - name: Push iroha2:dev docker image
```

The same for `iroha2.yml`

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -45,23 +49,48 @@ jobs:
       image: 7272721/i2-ci:latest
     steps:
       # TODO Remove these steps #2165
```
s8sato, line 51:
```suggestion

```

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -45,23 +49,48 @@ jobs:
...
+        uses: andstor/file-existence-action@v1
+        with:
+          files: "lcov-${{ steps.hash.outputs.hash }}.info"
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
           override: true
           components: llvm-tools-preview
+        if: steps.check_files.outputs.files_exists == 'false'
       - uses: taiki-e/install-action@cargo-llvm-cov
```
s8sato, line 76:
```suggestion
      # TODO Remove these steps #2165
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: stable
          override: true
          components: llvm-tools-preview
        if: steps.check_files.outputs.files_exists == 'false'
      - uses: taiki-e/install-action@cargo-llvm-cov
      #
```

s8sato (changes requested):

Dockerfile.ci
BAStos525: Yeah, should be. We did single `RUN` in `Dockerfile.ci` that was on ubuntu base image, but I forgot to do it on Arch-based Dockerfile version.

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
BAStos525: It was useful for cross-workflows reuse until [this PR was merged](https://github.com/hyperledger/iroha/commit/219355e552eb3cb7ac9614994cf71f21ab6c9ebb). I'm not sure if this is still relevant.

rust-toolchain.toml
```diff
@@ -1,5 +0,0 @@
-[toolchain]
```
s8sato, line 1: Since the toolchain file helps us to develop in similar environments, I'd [override](https://rust-lang.github.io/rustup/overrides.html) instead of remove

.github/workflows/iroha2-dev.yml
```diff
@@ -19,75 +23,77 @@ jobs:
...
+      - uses: convictional/trigger-workflow-and-wait@v1.6.2
         with:
           owner: soramitsu
           repo: iroha2-longevity-load-rs
           github_token: ${{ secrets.G_ACCESS_TOKEN }}
           workflow_file_name: load-rs-push-from-dev.yaml
           ref: dev
 
-  build-ci-image:
+  upload_coverage:
```
s8sato, line 72: I think this is what we need to update the comparison base (iroha2-dev)

s8sato (changes requested):
We're almost to the merge

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
s8sato: Now that in fdac826 the push workflow reuse the coverage report, something like the repository hash will be required for identifying the report (I don't think `github.sha` works for this case)

.github/workflows/iroha2-dev.yml
```diff
@@ -7,6 +7,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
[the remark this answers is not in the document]
BAStos525: But does it stop previous ones workflows that are outdated?

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
BAStos525: "I don't think github.sha works for this case" : we don't use `github.sha`, but we try to calculate files & folders hash.

.github/workflows/iroha2-dev.yml
BAStos525: I didn't understand the essence of your thought.

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -44,24 +48,49 @@ jobs:
...
+        uses: dawidd6/action-download-artifact@v2.21.1
+        with:
+          workflow: iroha2-dev-pr.yml
+          check_artifacts:  true
+        continue-on-error: true
+      - name: Check if coverage file exists
+        id: check_files
+        uses: andstor/file-existence-action@v1
+        with:
+          files: "lcov-${{ steps.hash.outputs.hash }}.info"
```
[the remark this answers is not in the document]
BAStos525: If so, why we saw a cases when we had the same repository hash after committed to PR's branch and `iroha2-dev-pr` workflow was triggered to run?

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
s8sato: Yeah, I meant I agreed with the use of `theowenyoung/folder-hash`.

After researching, however, I found a way to avoid the hash calculation -- upload the report with `github.sha` on the PR workflow and download it with `github.event.pull_request.head.sha` on the ~~push~~ [merge workflow](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges):

`iroha2-dev-pr.yml`

```yml
jobs:
  # omit
  tests_with_coverage:
    # omit
    steps:
      # omit
      - name: Upload codecov report artifact
        uses: actions/upload-artifact@v3
        with:
          name: lcov-${{ github.sha }}.info
          path: lcov.info

```

`iroha2-dev.yml`

```yml
on:
  pull_request:
    branches: [iroha2-dev]
    types:
      - closed
# omit
jobs:
  # omit
  upload_coverage:
    if: github.event.pull_request.merged
    # omit
    steps:
      # omit
      - name: Download codecov report artifact
        uses: dawidd6/action-download-artifact@v2.21.1
        with:
          workflow: iroha2-dev-pr.yml
          name: lcov-${{ github.event.pull_request.head.sha }}.info

```

.github/workflows/iroha2-dev.yml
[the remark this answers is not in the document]
s8sato: In that case some commits may lack the corresponding published image. What do you think @appetrosyan

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
s8sato: That observation should change my assumption. https://github.com/hyperledger/iroha/pull/2433#discussion_r920741030 might help

.github/workflows/iroha2-dev.yml
s8sato: I mean I'll welcome the revival of this job, which was removed at 219355e5. I guess the absence of the job is keeping the comparison base to 6973f71, which is few commits ahead of 219355e5

image

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
appetrosyan: @BAStos525 I like the approach, the less we depend on external actions, the better.

.github/workflows/iroha2-dev.yml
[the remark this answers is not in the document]
appetrosyan: I don't think that we need concurrency on the push workflow. We only need it to stop the PR workflows automatically. I suggest doing this in a separate PR

rust-toolchain.toml
appetrosyan: I agree, this is what we're doing locally. I removed it during development, since we were tinkering with the environment image. We can reintroduce this file later in a separate PR.

.github/workflows/iroha2-dev.yml
appetrosyan: We removed the re-running of the job. The solution that we came up with @safinsaf is to cache the coverage reports as artifacts. Run tests once (in PR) and upload twice (PR and Push).

.github/workflows/iroha2-dev.yml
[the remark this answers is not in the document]
BAStos525: OK, let's remove concurrency for push workflows here,

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
BAStos525: I found out that `dawidd6/action-download-artifact@v2.21.1` Actioon fails on  `name: lcov-${{ github.event.pull_request.head.sha }}.info`
It works if just ti set the parameter to download the last  available artifact from previous workflow. Fortunately, it matches with created artifact on PR workflow with `name: lcov-${{ github.sha }}.info`
But probably expected `coverage diff` still doesn't work.

.github/workflows/iroha2-dev.yml
```diff
@@ -19,75 +21,67 @@ jobs:
...
         with:
           owner: soramitsu
           repo: iroha2-longevity-load-rs
           github_token: ${{ secrets.G_ACCESS_TOKEN }}
           workflow_file_name: load-rs-push-from-dev.yaml
           ref: dev
 
-  build-ci-image:
+  upload_coverage:
+    if: github.event.pull_request.merged
```
s8sato, line 71: image

This no reaction of workflows can imply some syntax errors. Could you try this:

```suggestion
    if: github.event.pull_request.merged == true
```

Also, this line should be inserted to other jobs in this workflow

I'll introduce [`actionlint`](https://github.com/rhysd/actionlint#actionlint) in my next PR. Even with `actionlint`, I don't know why the current expression can be invalid though

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -44,24 +48,28 @@ jobs:
...
     steps:
+      - uses: actions/checkout@v2
+
       # TODO Remove these steps #2165
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
           override: true
           components: llvm-tools-preview
       - uses: taiki-e/install-action@cargo-llvm-cov
```
s8sato, line 59: Please keep the range of steps clear in some way

```suggestion
      # TODO Remove these steps #2165
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: stable
          override: true
          components: llvm-tools-preview
      - uses: taiki-e/install-action@cargo-llvm-cov
      #
```

.github/workflows/iroha2-dev.yml
s8sato: Probably I was wrong with the cause of the no reaction. Maybe it is the conflicts

image

BAStos525: I don't think that there are syntax errors due all checks for this PR are running on fork.

Cargo.lock
```diff
@@ -4072,8 +4070,7 @@ checksum = "957e51f3646910546462e67d5f7599b9e4fb8acdd304b087a6494730f9eebf04"
 [[package]]
 name = "unique_port"
 version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d3f903132efabcb4242e0bbf16aedb2c3867526e42126b6ed3cb1b0bc4577708"
+source = "git+https://github.com/ales-tsurko/iroha2-unique_port?rev=09e95bb3608626c995b8e7e51c64ba7486c88948#09e95bb3608626c995b8e7e51c64ba7486c88948"
```
outoftardis, line 4073: I think this should point to https://github.com/soramitsu/iroha2-unique_port, not the fork

s8sato: From [this experience](https://github.com/hyperledger/iroha/pull/2363/files#r906053569) I don't think we need to take care of `Cargo.lock`. I guess @ales-tsurko 's work is now ahead of what crates.io has, thus this is temporary and acceptable, right?

outoftardis: it seems like Ales's change was already merged upstream though

BAStos525: IDK the exact reason, let's make it the same as in [the head branch](https://github.com/hyperledger/iroha/blob/5ff6b67d392a0f8238d483918ea36e82d9a22937/Cargo.lock#L4075)

s8sato: I'd leave as is. It will be automatically overwritten if necessary. We should not manually touch `Cargo.lock`

.github/workflows/iroha2-dev.yml
```diff
@@ -12,84 +13,78 @@ env:
...
-      - name: Build & push Docker container
-        uses: docker/build-push-action@v2
+          workflow: iroha2-dev-pr.yml
+      - name: Upload to codecov
+        uses: codecov/codecov-action@v3
         with:
-          file: Dockerfile.ci
-          tags: hyperledger/iroha2-ci:latest
-          push: true
+          file: lcov.info
```
s8sato, line 84: Now that you removed the renaming step, I don't think this corresponds to `lcov-${{ github.sha }}.info` uploaded by the PR workflow.

Regarding the download failure mentioned in https://github.com/hyperledger/iroha/pull/2433#discussion_r921399893,
have you tried unifying the artifact name to either `lcov-${{ github.sha }}.info` or `lcov-${{ github.event.pull_request.head.sha }}.info` between the upload on PR and the download on merge? 
On my second thought, since both workflows run on `pull_request` event, there was no reason to use different expressions

.github/workflows/iroha2-dev-pr.yml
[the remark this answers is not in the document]
s8sato: I hope https://github.com/hyperledger/iroha/pull/2433#discussion_r922151913 resolves this

.github/workflows/iroha2-dev.yml
BAStos525: Commit's SHA is changed on PR push and PR merge events. However, trying `{{ github.event.pull_request.head.sha }}` I got this on my fork
image

s8sato: So neither of these expressions can work for passing artifacts?

BAStos525: I guess `{{ github.event.pull_request.head.sha }}` could work to pass artifacts for [PR last push and PR merge commits](https://codecov.io/gh/BAStos525/soramitsu-iroha/commit/3edfd1b3969f0223adb0cecf05caf26d5a65a2e4/build).

s8sato: Ah, aren't the `267..` numbers the run ids? No wonder they are different

BAStos525: I think `267...` could be different. The main thing here that `commits.sha` for codecov coverage are matched.

s8sato: Yeah, exactly. I think `github.sha` is more appropriate since it points the merge commit:

> Note that `GITHUB_SHA` for this event is the last merge commit of the pull request merge branch. If you want to get the commit ID for the last commit to the head branch of the pull request, use `github.event.pull_request.head.sha` instead.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

s8sato: Nope, if "the pull request merge branch" means the base branch (`iroha2-dev`), its last merge commit will be useless. In that case, `${{ github.event.pull_request.head.sha }}` will be the only option

BAStos525: From practical and experimental side I can say that only `${{ github.event.pull_request.head.sha }}` worked for me. `${{ github.sha }}` showed different value on PR merge action to base branch `iroha2-dev` from PR last push action.

s8sato: Then I think we can resolve this conversation. I'll approve after the [format check](https://github.com/hyperledger/iroha/runs/7362025968?check_suite_focus=true)

s8sato: Please fetch [this commit to fix the format](https://github.com/s8sato/iroha/tree/adda15f37aca01fb444fbd315202f182471c9f4e) if you like

BAStos525: > Then I think we can resolve this conversation. I'll approve after passing the [format check](https://github.com/hyperledger/iroha/runs/7362025968?check_suite_focus=true)

Ok, I'll proceed with [remained tasks](https://github.com/hyperledger/iroha/issues/2461) then

s8sato (dismissed):

docker-compose.yml
```diff
@@ -16,11 +17,12 @@ services:
     volumes:
       - './configs/peer:/config'
     init: true
-    command: ./iroha --submit-genesis
+    command: iroha --submit-genesis
 
   iroha1:
     image: hyperledger/iroha2:dev
     environment:
+      KURA_BLOCK_STORE_PATH: /storage
```
pesterev, line 25: According to `docker-compose-local.yml`, perhaps needs to add `depends_on: - iroha0` here that peers who submit genesis will depend on a peer who submits genesis.

s8sato: Since in the real production the genesis peer should start without knowing the network status, I don't think the order of service start should be specified in general. The reason why it is specified in `docker-compose-local.yml` is for building the image, I think

.github/workflows/iroha2-dev.yml
```diff
@@ -12,84 +13,80 @@ env:
...
+          workflow: iroha2-dev-pr.yml
+          name: lcov-${{ github.event.pull_request.head.sha }}.info
+          search_artifacts: true
+      - name: Upload to codecov
+        uses: codecov/codecov-action@v3
         with:
-          file: Dockerfile.ci
-          tags: hyperledger/iroha2-ci:latest
-          push: true
+          file: lcov-${{ github.event.pull_request.head.sha }}.info
```
s8sato, line 86: One thing,
if we specify the commit ID in the filename for codecov, does codecov find relation between the different filenames and make proper comparisons?

docker-compose.yml
BAStos525: However, we don't have `depends_on:` block in docker-compose template for iroha2 deployment on our real stands (environments).

.github/workflows/iroha2-dev.yml
BAStos525: This is actually a good point. I didn't spend time to verify it, anyway, since `github.event.pull_request.head.sha` will be different for base branch lcov report and for merge branch correspondingly, we might unify the filenames for reports to `lcov.info`. Since files for `upload & download artifacts` should be the same, codecov has to join these run IDs reports in one coverage by PR head sha.

.github/workflows/iroha2-dev.yml
```diff
@@ -12,84 +13,80 @@ env:
...
+          workflow: iroha2-dev-pr.yml
+          name: lcov-${{ github.event.pull_request.head.sha }}.info
+          search_artifacts: true
+      - name: Upload to codecov
+        uses: codecov/codecov-action@v3
         with:
-          file: Dockerfile.ci
-          tags: hyperledger/iroha2-ci:latest
-          push: true
+          file: lcov.info
```
s8sato, line 86: According to [the doc](https://github.com/codecov/codecov-action#arguments),

```suggestion
          files: lcov.info
```

Also, renaming from `lcov-${{ github.event.pull_request.head.sha }}.info` will be needed

BAStos525: Right, it is better to follow official arguments.

pesterev (dismissed):

.github/workflows/iroha2-dev.yml
BAStos525: What renaming do you mean except [these ones](https://github.com/hyperledger/iroha/pull/2433/commits/e40617bd86203fa78cf38acd9d0a5da281ab9a76)?

s8sato: As of 35a3dc6 in the `upload_coverage` job the download step downloads `lcov-${{ github.event.pull_request.head.sha }}.info` and the upload step tries to upload `lcov.info`

BAStos525: Actually codecov action is able to detect any possible reports to be uploaded:
```
[2022-07-19T09:09:15.167Z] ['info'] Searching for coverage files...
[2022-07-19T09:09:15.[26](https://github.com/BAStos525/soramitsu-iroha/runs/7405964287?check_suite_focus=true#step:5:27)4Z] ['info'] => Found 1 possible coverage files:
  lcov.info
```
That's why it was not an error. But right, it's better to specify the filename.

.github/workflows/iroha2-dev.yml
```diff
@@ -12,84 +13,80 @@ env:
...
+          workflow: iroha2-dev-pr.yml
+          name: lcov-${{ github.event.pull_request.head.sha }}.info
+          search_artifacts: true
+      - name: Upload to codecov
+        uses: codecov/codecov-action@v3
         with:
-          file: Dockerfile.ci
-          tags: hyperledger/iroha2-ci:latest
-          push: true
+          files: lcov-${{ github.event.pull_request.head.sha }}.info
```
s8sato, line 86: https://github.com/hyperledger/iroha/pull/2433#discussion_r924142827 again.
In https://github.com/hyperledger/iroha/pull/2433#discussion_r924381751 I'm suggesting to revive the renaming step that you commented out in 6359b8bfccd4f674ae4ebf30beb52cd222410f3d

s8sato (dismissed):

pesterev (approved):

s8sato (approved):

pesterev (approved):

.github/workflows/iroha2-dev.yml
```diff
@@ -12,84 +13,82 @@ env:
...
-            BIN=iroha_client_cli
-      - name: Build and push docker image (kagami)
-        uses: docker/build-push-action@v2
+          file: Dockerfile.build
+
+      - name: Build iroha2:dev docker image
+        run: |
+          docker build -f Dockerfile -t hyperledger/iroha2:dev .
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
```
s8sato, line 50: I guess `docker/build-push-action@v3` requires login as a preliminary step, even if it does not push 
https://github.com/docker/build-push-action#usage

BAStos525: We can try to move `docker/login-action` up and set it before `docker/build-push-action@v3`. By the way, if the problem in that, I guess building from forks should has the same issue.

Issue #1966, opened 2022-03-10
`FindAllParameters` unusable


mversic:
Since `ValidQuery` is not implemented for `FindAllParameters` it's not possible to execute this query. Either this query should be removed entirely or it should be enlisted in the `QueryBox` enum

abhijeetjejurkar:
Hey, I am from Openforce 2022, I would like to work on this issue.

mversic:
@appetrosyan I guess we should support this query? not remove it

appetrosyan:
If neither Unkai nor Orillion use it, I would vote for removal. @Mingela do you think that such a query might be useful in the future?

appetrosyan:
@abhijeetjejurkar while you're free to choose what issue you work on, issues that are particularly easy to fix by people outside the core development team (and often higher priority are unassigned and marked with the "good first issue" tag. 

If you like, we could arrange a small introduction and give you an overview of the issues that would be higher priority for Iroha, and more self-contained.

abhijeetjejurkar:
Yes please give me overview of issues which can be solved by me, I would like to work with your community 😊. I will search and work for such high priority issues.

appetrosyan:
Thank you, it's much appreciated.

To get started, you should look at good first issues and on our board, issues in the Chores categories that aren't assigned to anyone. These are small self-contained modifications that do not interfere with our roadmap. 

Be sure to read our contributing guide, particularly to use the rebase workflow, and follow our commit conventions: square brackets, category, #issue number: description  

https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/board?repos=181739240

abhijeetjejurkar:
Sure👍, Thankyou 😊

mversic:
@Mingela any thoughts on this?

appetrosyan:
The `FindAllParameters` query is messed up. Instead of being tied to the configuration struct and returning a structure based off of configuration, it instead returns something known as `Parameter`. Instead of being a struct containing all of the publicly available fields, it instead is an enum. 

There needs to be more work done on this query before it can be made available publicly.

dmitrivenger:
Closing the issue now. The feature is relevant now. 
@appetrosyan and @astrokov7

Issue #2392, opened 2022-06-24
Refactor `data_model`


appetrosyan:
The `iroha_data_model` crate is used for all data that must be shared between the Iroha peer, and client. Unfortunately, this led to delocalisation of data, and any additional implementation can introduce cyclic dependencies. As such the crate needs to be broken down. 

### Proposed solution

We re-orient `iroha` into domain-specific hierarchy, with the `data_model` used for re-exporting specific crates, `client` handling the client communication and providing an example implementation. 

As such the main workspace should have the following top-level memebers
- `peer` 
- `client`
- `data_model`
- `system`
- `tools`

1. `peer` is the closest analogue of the current `core` crate. However, the smartcontract functionality is to be removed from `core`. It itself should consist of the following crates: 

- `sumeragi`
- `block_sync` (turned into a separate crate)
- `torii`
- `cli`
- `config` (which aggregates the configurations from domain-specific configs)

The core crate should only contain the consensus logic, and peer-specific implementations. 

2. `client` remains as is. The `iroha_client_cli` is moved into tools, as it's not meant to be used outside of the debugging context. 

3. `data_model`. This crate should contain the encapsulated and domain-first ordered crates for each type of entity that it deals with. 
- `world`
- `query`. 
- `instruction`
- `expression`
  - `predicate`
  - `arithmetic`
  - `fixed`
- `event`
  - `filter`
  - `account`
  - `domain`
  - …
- `permission`
- `version`
- `schema`
- `wasm`
- `domain`
  - `account`
  - `asset`
  - …
Because of the orphan rule, and contrary to the domain-first approach, we will need to encapsulate instructions completely. In other words, instruction should contain both the traits and implementations for usage in `sumeragi`. As such `impl Execute for Mint<Asset>` like before, should be implemented in the `instruction` crate, rather than the `asset` crate. However, the `instruction` crate should have a module for each domain-specific entity. 

4. `system`. This is the most straightforward modification. It should contain the following sub-crates
- `crypto`
- `actor`
- `logger`
- `ffi`
- `futures`
- `p2p`
- `macro`
- `config`
- `telemetry`
- `queue`
- `kura`
- primitives, such as `ConstString`. 

[...]

PR #2435, opened 2022-07-05
[fix] #2422: Hide private keys in configuration endpoint response

Erigara:
### Description of the Change

- Add configuration-related to `data_model` to be able to share configuration;
- Fix `GetConfiguration` endpoint to return sharable configuration without private keys;
- Add proc macro to derive conversation between inner and sharable config. 

### Issue

Closses #2422.

### Benefits

- No sensitive information are leaked;
- Opens possibility to share configuration between peers through Query (#1966);

### Possible Drawbacks

- Duplication configuration structs.
- Add dependency `iroha_data_model` to `telemetry` and `logger` crates.

### Alternate Designs

Simpler solution would be to mask or hide private keys during serialization but this would affect generated reference configuration.
Add it still won't be possible to send config through queries.


Erigara:
> Also why do we need _View-like_ configs in the `data_model`? 

If we want something to be queue result it must be in `data_model`.

> My suggestion is to remove it and generate _View_ in place using derive macro. If we really need them in the `data_model` then let's move basic _Configurations_ in the `data_model` and generate _View_ here

I'm not sure if it will be good thing to aggregate all configuration from different domains in one place.

config/derive/src/lib.rs
```diff
@@ -500,3 +500,131 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+                //  Important to write macro in uniform way.
+                #[allow(clippy::needless_update)]
+                Self {
+                    #(
+                        #field_idents: From::<_>::from(#field_idents),
+                    )*
+                    ..Self::default()
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Erigara: Explained my intentions [here](https://github.com/hyperledger/iroha/pull/2435#discussion_r915903105).

Arjentix:
> I'm not sure if it will be good thing to aggregate all configuration from different domains in one place.

If we will split this then we can't generate *view* from config. Then with updating of config we have to remember to update *view*. And event compiler won't say anything, because if we forgot then it will use `default()` for this field

Erigara:
> If we will split this then we can't generate view from config. Then with updating of config we have to remember to update view. And event compiler won't say anything, because if we forgot then it will use default() for this field. 

It won't compile because without `#[view(ignore)]` it will try to assign this field to view structure and view doesn't have this field.

config/src/lib.rs
```diff
@@ -58,6 +58,51 @@ pub mod derive {
...
+    /// #[view(View)]
+    /// struct Structure {
+    ///     a: u32,
+    ///     // `View` doesn't have field `b` so we must exclude it.
+    ///     #[view(ignore)]
+    ///     b: u32,
+    /// }
+    ///
+    /// let view = View { a: 32 };
+    /// let structure: Structure = view.into();
```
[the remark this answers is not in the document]
Erigara: Then we can load keys from environment variable or other places.

logger/src/config.rs
```diff
@@ -51,14 +55,27 @@ impl<T: Subscriber + Debug> ReloadMut<Level> for Handle<LevelFilter, T> {
     }
 }
 
+/// Wrapper around [`Level`] for runtime upgrades.
+#[derive(Clone, Debug, Serialize, Deserialize, Deref, DerefMut, Default)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub struct SyncLevel(handle::SyncValue<Level, handle::Singleton<Level>>);
```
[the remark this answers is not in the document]
Erigara: No sure, just add wrapper to be able to implement `From`.

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Configuration for `GenesisBlock`.
+    pub genesis: genesis::Configuration,
+    /// Configuration for `WorldStateView`.
+    pub wsv: wsv::Configuration,
+    /// Network configuration
+    pub network: network::Configuration,
+    /// Configuration for telemetry
+    #[cfg(feature = "telemetry")]
+    pub telemetry: telemetry::Configuration,
+}
```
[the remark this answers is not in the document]
Arjentix: What about to annotate such fields view `#[view(...)]` attribute?

config/src/lib.rs
```diff
@@ -58,6 +58,51 @@ pub mod derive {
     /// assert_eq!(ip.ip, Ipv4Addr::new(127, 0, 0, 1));
     /// ```
     pub use iroha_config_derive::Configurable;
+    /// Derive macro for conversation between type and it's view which contains only subset of type's fields.
```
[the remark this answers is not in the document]
Erigara: Oh, actually, it's `cargo format` apply such formatting, so i think we should leave it as is.

Erigara:
> If we really need them in the data_model then let's move basic Configurations in the data_model and generate View here

Also i think defining all configuration in one place go against upcoming refactoring #2392.

IMO we should gather opinions from other team members.

SamHSmith (approved):
Looks good to me.

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[serde(rename_all = "UPPERCASE")]
+    /// Configuration of the genesis block and the process of its submission.
+    pub struct Configuration {
+        /// The genesis account public key, should be supplied to all peers.
+        pub account_public_key: PublicKey,
+        /// Number of attempts to connect to peers, while waiting for them to submit genesis.
+        pub wait_for_peers_retry_count_limit: u64,
+        /// Period in milliseconds in which to retry connecting to peers, while waiting for them to submit genesis.
+        pub wait_for_peers_retry_period_ms: u64,
+        /// Delay before genesis block submission after minimum number of peers were discovered to be online.
```
outoftardis, line 230:
```suggestion
        /// The delay before genesis block submission after minimum number of peers were discovered to be online.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Module for genesis-related configuration and structs.
+pub mod genesis {
+    use super::*;
+
+    #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "UPPERCASE")]
+    /// Configuration of the genesis block and the process of its submission.
+    pub struct Configuration {
+        /// The genesis account public key, should be supplied to all peers.
```
outoftardis, line 224:
```suggestion
        /// The public key of the genesis account, should be supplied to all peers.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+        pub account_public_key: PublicKey,
+        /// Number of attempts to connect to peers, while waiting for them to submit genesis.
+        pub wait_for_peers_retry_count_limit: u64,
+        /// Period in milliseconds in which to retry connecting to peers, while waiting for them to submit genesis.
```
outoftardis, line 228:
```suggestion
        /// The period in milliseconds in which to retry connecting to peers while waiting for them to submit genesis.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+pub mod genesis {
+    use super::*;
+
+    #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "UPPERCASE")]
+    /// Configuration of the genesis block and the process of its submission.
+    pub struct Configuration {
+        /// The genesis account public key, should be supplied to all peers.
+        pub account_public_key: PublicKey,
+        /// Number of attempts to connect to peers, while waiting for them to submit genesis.
```
outoftardis, line 226:
```suggestion
        /// The number of attempts to connect to peers while waiting for them to submit genesis.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[derive(Clone, Deserialize, Serialize, Debug)]
+    #[serde(rename_all = "UPPERCASE")]
+    pub struct Configuration {
+        /// Maximum log level
+        pub max_log_level: Level,
+        /// Capacity (or batch size) for telemetry channel
+        pub telemetry_capacity: u32,
+        /// Compact mode (no spans from telemetry)
+        pub compact_mode: bool,
+        /// If provided, logs will be copied to said file in the
```
outoftardis, line 192:
```suggestion
        /// If provided, logs will be copied to the given file in the
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    /// Configuration for `Queue`.
+    #[derive(Copy, Clone, Deserialize, Serialize, Debug, PartialEq, Eq)]
+    #[serde(rename_all = "UPPERCASE")]
+    pub struct Configuration {
+        /// The upper limit of the number of transactions per block.
+        pub maximum_transactions_in_block: u32,
+        /// The upper limit of the number of transactions waiting in this queue.
+        pub maximum_transactions_in_queue: u32,
+        /// The transaction will be dropped after this time if it is still in a `Queue`.
```
outoftardis, line 171:
```suggestion
        /// The upper limit of the number of transactions waiting in the queue.
        pub maximum_transactions_in_queue: u32,
        /// The transaction will be dropped after this time if it is still in the queue.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Module for block_sync-related configuration and structs.
+pub mod block_sync {
+    use super::*;
+
+    /// Configuration for `BlockSynchronizer`.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "UPPERCASE")]
+    pub struct Configuration {
+        /// The time between sending requests for latest block.
```
outoftardis, line 149:
```suggestion
        /// The period of time to wait between sending requests for the latest block.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+        pub peers: std::collections::HashSet<crate::peer::Id>,
+    }
+}
+
+/// Module for torii-related configuration and structs.
+pub mod torii {
+    use super::*;
+    /// Structure that defines the configuration parameters of `Torii` which is the routing module.
+    /// For example the `p2p_addr`, which is used for consensus and block-synchronisation purposes,
+    /// as well as `max_transaction_size`.
```
outoftardis, line 124:
```suggestion
    /// Structure that defines the configuration parameters of `Torii`, the routing module.
    /// This includes parameters such as `p2p_addr`, which is used for consensus and block-synchronisation purposes,
    /// as well as `max_transaction_size`.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+        /// Buffer capacity of actor's MPSC channel
+        pub actor_channel_capacity: u32,
+        /// Maximum number of transactions in tx gossip batch message. While configuring this, attention should be payed to `p2p` max message size.
+        pub gossip_batch_size: u32,
+        /// Period in milliseconds for pending transaction gossiping between peers.
+        pub gossip_period_ms: u64,
+    }
+
+    /// `SumeragiConfiguration` provides an ability to define parameters
+    /// such as `BLOCK_TIME_MS` and list of `TRUSTED_PEERS`.
```
outoftardis, line 108:
```suggestion
    /// The list of trusted peers
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[derive(Debug, Clone, Default, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "UPPERCASE")]
+    #[serde(transparent)]
+    pub struct TrustedPeers {
+        /// Optional list of predefined trusted peers. Must contain unique
+        /// entries. Custom deserializer raises error if duplicates found.
```
outoftardis, line 114:
```suggestion
        /// Optional list of predefined trusted peers. Must contain unique
        /// entries.
        ///
        /// # Errors
        /// Custom deserializer raises an error if there are duplicates.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    pub struct Configuration {
+        /// Current Peer Identification.
+        pub peer_id: crate::peer::Id,
+        /// Amount of time peer waits for the `CreatedBlock` message after getting a `TransactionReceipt`
+        pub block_time_ms: u64,
+        /// Optional list of predefined trusted peers.
+        pub trusted_peers: TrustedPeers,
+        /// Amount of time Peer waits for CommitMessage from the proxy tail.
+        pub commit_time_limit_ms: u64,
+        /// Amount of time Peer waits for TxReceipt from the leader.
```
outoftardis, line 95:
```suggestion
        /// The amount of time a peer waits for `CommitMessage` from the proxy tail.
        pub commit_time_limit_ms: u64,
        /// The amount of time a peer waits for `TxReceipt` from the leader.
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+pub mod sumeragi {
+    use super::*;
+    /// `SumeragiConfiguration` provides an ability to define parameters such as `BLOCK_TIME_MS`
+    /// and list of `TRUSTED_PEERS`.
+    #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "UPPERCASE")]
+    pub struct Configuration {
+        /// Current Peer Identification.
+        pub peer_id: crate::peer::Id,
+        /// Amount of time peer waits for the `CreatedBlock` message after getting a `TransactionReceipt`
```
outoftardis, line 89:
```suggestion
        /// The amount of time a peer waits for the `CreatedBlock` message after getting a `TransactionReceipt`
```

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+        pub init_mode: Mode,
+        /// Path to the existing block store folder or path to create new folder.
+        pub block_store_path: String,
+        /// Maximum number of blocks to write into single storage file
+        pub blocks_per_storage_file: core::num::NonZeroU64,
+        /// Default buffer capacity of actor's MPSC channel
+        pub actor_channel_capacity: u32,
+    }
+
+    /// Kura work mode.
```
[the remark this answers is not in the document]
Erigara: yes

config/derive/src/lib.rs
```diff
@@ -500,3 +500,131 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+
+impl Parse for ViewIgnore {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        Ok(Self {
+            _ident: parse_const_ident(input, Self::IGNORE)?,
+        })
+    }
+}
+
+/// Derive conversation between type and it's view. Check other doc in `iroha_config` reexport.
```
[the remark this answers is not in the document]
Erigara: Doc comments for `View` reexport in `iroha_config::derive`.

data_model/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Module for wasm-related configuration and structs.
+pub mod wasm {
+    use super::*;
+    /// `WebAssembly Runtime` configuration.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "UPPERCASE")]
+    pub struct Configuration {
+        /// Every WASM instruction costs approximately 1 unit of fuel. See
+        /// [`wasmtime` reference](https://docs.rs/wasmtime/0.29.0/wasmtime/struct.Store.html#method.add_fuel)
```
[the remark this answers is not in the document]
Erigara: @outoftardis, maybe something like this will be better?

> Maximum amount of fuel provided to smart contract.
> Limits the number of instructions that can be executed during the execution of a smart contract. 
> Every WASM instruction costs approximately 1 unit of fuel.
> See [`wasmtime` reference](https://docs.rs/wasmtime/0.29.0/wasmtime/struct.Store.html#method.add_fuel).

Erigara:
After conversation with @Arjentix  and @appetrosyan was decided to change approach for solution. 

Convert PR to draft for now.

data_model/src/config.rs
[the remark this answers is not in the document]
outoftardis: much better, yes. 
you can make the first two sentences more concise, like this:

> The fuel limit determines the maximum number of instructions that can be executed during the execution of a smart contract.

data_model/src/config.rs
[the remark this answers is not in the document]
appetrosyan: The view attribute sounds good, but it might be difficult to implement.

config/src/lib.rs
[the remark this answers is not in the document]
Erigara: @appetrosyan, ok, i will make one way conversion `Config -> View` to avoid relying on `Default`, backward conversation could be achieved be using [proxy configuration objects](https://github.com/hyperledger/iroha/issues/2118#issuecomment-1121171520).

appetrosyan:
Just FYI, the commits are still not quite following the convention, so they need to be squashed.

Erigara:
> Just FYI, the commits are still not quite following the convention, so they need to be squashed.

Okay, that was my plan anyway

SamHSmith (dismissed):

config/src/sumeragi.rs
```diff
@@ -18,37 +18,42 @@ const DEFAULT_ACTOR_CHANNEL_CAPACITY: u32 = 100;
...
+    #[serde(rename_all = "UPPERCASE")]
+    #[config(env_prefix = "SUMERAGI_")]
+    pub struct Configuration {
+        /// Key pair of private and public keys.
+        #[serde(skip)]
+        #[view(ignore)]
+        pub key_pair: KeyPair,
+        /// Current Peer Identification.
+        pub peer_id: PeerId,
+        /// The amount of time a peer waits for the `CreatedBlock` message after getting a `TransactionReceipt`
```
outoftardis, line 37: same comment about changing `amount` to `period` as in another file

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use serde_json::Value;
+
+pub mod derive {
+    //! Modules with things related with deriving `Configurable`
+
+    use std::{error::Error as StdError, fmt};
+
+    use derive_more::Display;
+    /// Generate view for the type and implements conversion `Type -> View`.
```
outoftardis, line 12:
```suggestion
    /// Generate view for the type and implement conversion `Type -> View`.
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// struct Prefixed { a: String }
+    ///
+    /// std::env::set_var("PREFIXED_A", "B");
+    /// let mut prefixed = Prefixed { a: "a".to_owned() };
+    /// prefixed.load_environment();
+    /// assert_eq!(prefixed.a, "B");
+    /// ```
+    ///
+    /// ## `inner`
+    /// Tells macro that structure stores another config inside
```
outoftardis, line 86:
```suggestion
    /// Tells macro that the structure stores another config inside
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    ///
+    /// #[derive(serde::Deserialize, serde::Serialize, Configurable)]
+    /// struct Inner { b: String }
+    ///
+    /// let outer = Outer { inner: Inner { b: "a".to_owned() }};
+    /// assert_eq!(outer.get_recursive(["inner", "b"]).unwrap(), "a");
+    /// ```
+    ///
+    /// ## `serde_as_str`
+    /// Tells macro to deserialize from env variable as bare string:
```
outoftardis, line 101:
```suggestion
    /// Tells macro to deserialize from env variable as a bare string:
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+            Some(&self.error)
+        }
+    }
+
+    /// Derive `Configurable` error
+    #[derive(Debug)]
+    pub enum Error {
+        /// Got unknown field
+        UnknownField(Vec<String>),
+        /// Failed to deserialize or serialize field
```
outoftardis, line 137:
```suggestion
        /// Failed to deserialize or serialize a field
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+pub mod runtime_upgrades;
+
+/// Trait for dynamic and asynchronous configuration via maintenance endpoint for rust structures
+pub trait Configurable: Serialize + DeserializeOwned {
+    /// Error type returned by methods of trait
+    type Error;
+
+    /// Gets field of structure and returns as json-value
```
outoftardis, line 185:
```suggestion
    /// Return the JSON value of a given field
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// # Errors
+    /// Fails if field was unknown
+    fn get(&self, field: &'_ str) -> Result<Value, Self::Error> {
+        self.get_recursive([field])
+    }
+
+    /// Gets inner field of arbitrary inner depth and returns as json-value
```
outoftardis, line 192:
```suggestion
    /// Return the JSON value of a given inner field of arbitrary inner depth
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// # Errors
+    /// Fails if field was unknown
+    fn get_recursive<'tl, T>(&self, inner_field: T) -> Result<Value, Self::Error>
+    where
+        T: AsRef<[&'tl str]> + Send + 'tl;
+
+    /// Fails if fails to deserialize from environment
```
outoftardis, line 199:
```suggestion
    /// Load configuration from the environment
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// # Errors
+    /// Fails if field was unknown
+    fn get_doc(field: &str) -> Result<Option<String>, Self::Error> {
+        Self::get_doc_recursive([field])
+    }
+
+    /// Returns documentation for all fields in form of json object
+    fn get_docs() -> Value;
+
+    /// Gets inner docs for non-leaf fields
```
outoftardis, line 222:
```suggestion
    /// Get inner docs for non-leaf fields
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+        -> Result<Option<String>, Self::Error>;
+
+    /// Gets docs of field
+    /// # Errors
+    /// Fails if field was unknown
+    fn get_doc(field: &str) -> Result<Option<String>, Self::Error> {
+        Self::get_doc_recursive([field])
+    }
+
+    /// Returns documentation for all fields in form of json object
```
outoftardis, line 219:
```suggestion
    /// Return documentation for all fields in a form of a JSON object
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn load_environment(&mut self) -> Result<(), Self::Error>;
+
+    /// Gets docs of inner field of arbitrary depth
+    ///
+    /// # Errors
+    /// Fails if field was unknown
+    fn get_doc_recursive<'tl>(field: impl AsRef<[&'tl str]>)
+        -> Result<Option<String>, Self::Error>;
+
+    /// Gets docs of field
```
outoftardis, line 212:
```suggestion
    /// Get documentation of a given field
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    where
+        T: AsRef<[&'tl str]> + Send + 'tl;
+
+    /// Fails if fails to deserialize from environment
+    ///
+    /// # Errors
+    /// Fails if fails to deserialize from environment
+    fn load_environment(&mut self) -> Result<(), Self::Error>;
+
+    /// Gets docs of inner field of arbitrary depth
```
outoftardis, line 205:
```suggestion
    /// Get documentation of a given inner field of arbitrary depth
```

config/src/iroha.rs
```diff
@@ -0,0 +1,154 @@
...
+            iroha_data_model::peer::Id::new(&self.torii.p2p_addr, &self.public_key.clone());
+
+        Ok(())
+    }
+
+    /// Loads configuration from environment
+    ///
+    /// # Errors
+    /// If Configuration deserialize fails:
+    /// - Configuration `TrustedPeers` contains entries with duplicate public keys
```
outoftardis, line 121: are these two different errors or a single one?

config/src/iroha.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+impl Configuration {
+    /// Construct [`Self`] from a path-like object.
+    ///
+    /// # Errors
+    /// - File not found.
+    /// - File found, but peer configuration parsing failed.
+    /// - Length of the array in raw json representation is different
+    /// to the lenght of the array in
```
outoftardis, line 95:
```suggestion
    /// - The length of the array in raw JSON representation is different
    /// from the length of the array in
```

config/src/iroha.rs
Erigara: It's more like clarification. One of possible reasons why deserialization might fail.

outoftardis:
```suggestion
    /// Fails if Configuration deserialization fails (e.g. if `TrustedPeers` contains entries with duplicate public keys)
```

would this work?

config/src/sumeragi.rs
```diff
@@ -18,37 +18,42 @@ const DEFAULT_ACTOR_CHANNEL_CAPACITY: u32 = 100;
...
+        pub trusted_peers: TrustedPeers,
+        /// The amount of time a peer waits for `CommitMessage` from the proxy tail.
+        pub commit_time_limit_ms: u64,
+        /// The amount of time a peer waits for `TxReceipt` from the leader.
+        pub tx_receipt_time_limit_ms: u64,
+        /// Limits to which transactions must adhere
+        pub transaction_limits: TransactionLimits,
+        /// Buffer capacity of actor's MPSC channel
+        pub actor_channel_capacity: u32,
+        /// Maximum number of transactions in tx gossip batch message. While configuring this, attention should be payed to `p2p` max message size.
```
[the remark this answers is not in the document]
Erigara: It's hard to tell really, but my intuition is that we have limit on message size and by increasing amount of transactions in gossip message we could eventually hit this limit.

config/src/iroha.rs
Erigara: yer, sure

config/src/sumeragi.rs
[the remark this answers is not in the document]
outoftardis: ok
apply my suggestion above, plus if you can add an internal link to that other config param, that'd be great

config/base/derive/src/lib.rs
```diff
@@ -500,3 +500,277 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+
+    quote! {
+        impl #impl_generics Default for #view_ident #ty_generics #where_clause {
+            fn default() -> Self {
+                let #original_ident {
+                    #(
+                        #field_idents,
+                    )*
+                    ..
+                } =  #original_ident::default();
```
Arjentix, line 711:
```suggestion
                } =  <#original_ident as Default>::default();
```

This may be important if someone will write `default()` method for his structure not as `Default` impl

config/base/derive/src/lib.rs
```diff
@@ -500,3 +500,277 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+/// Check if [`Field`] has `#[view(ignore)]`
+fn is_view_field_ignored(field: &Field) -> bool {
+    field
+        .attrs
+        .iter()
+        .find_map(|attr| {
+            if !attr.path.is_ident("view") {
+                return None;
+            }
+            attr.parse_args::<ViewIgnore>().ok()
```
Arjentix, line 745:
```suggestion
            attr.path.is_ident("view").then(|| attr.parse_args::<ViewIgnore>().ok()).flatten()
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    //! Modules for `Configurable` entities
+
+    use std::{error::Error as StdError, fmt};
+
+    use derive_more::Display;
+    /// Generate view for the type and implement conversion `Type -> View`.
+    /// View contains a subset of the fields that the type has.
+    ///
+    /// Works only with structs.
+    /// Type must implement `Default`.
```
Arjentix, line 16:
```suggestion
    /// Type must implement [`Default`].
```

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    ///
+    /// ```rust
+    /// use iroha_config_base::derive::view;
+    ///
+    /// view! {
+    ///     #[derive(Default)]
+    ///     struct Structure {
+    ///         #[view(type = u64)]
+    ///         a: u32,
+    ///         // `View` doesn't have field `b` so we must exclude it.
```
Arjentix, line 37:
```suggestion
    ///         // `View` shouldn't have field `b` so we must exclude it.
```

I think this style is more suitable, because `View` does not exist. That's the user who creates it

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// //      fn from(value: Structure) -> Self {
+    /// //          let Structure {
+    /// //              a,
+    /// //              ..
+    /// //          } = value;
+    /// //          Self {
+    /// //              a: From::<_>::from(a),
+    /// //          }
+    /// //      }
+    /// // }
```
Arjentix, line 59: As I understood there will be also a `Default` impl, doesn't it?

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use serde_json::Value;
+
+pub mod derive {
+    //! Modules for `Configurable` entities
+
+    use std::{error::Error as StdError, fmt};
+
+    use derive_more::Display;
+    /// Generate view for the type and implement conversion `Type -> View`.
```
Arjentix, line 12: Let's add a new line here

config/base/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// //          }
+    /// //      }
+    /// // }
+    ///
+    /// let structure = Structure { a: 13, b: 37 };
+    /// let view: StructureView = structure.into();
+    /// assert_eq!(view.a, 13);
+    /// ```
+    pub use iroha_config_derive::view;
+    /// Derive macro for implementing [`iroha_config::Configurable`](`crate::Configurable`) for config structures.
```
Arjentix, line 66: Let's add a new line here

config/src/genesis.rs
```diff
@@ -0,0 +1,78 @@
...
+        #[serde(default = "default_wait_for_peers_retry_period_ms")]
+        pub wait_for_peers_retry_period_ms: u64,
+        /// The delay before genesis block submission after minimum number of peers were discovered to be online.
+        /// The delay between submissions, which is used to ensure that other peers had time to connect to each other.
+        #[serde(default = "default_genesis_submission_delay_ms")]
+        pub genesis_submission_delay_ms: u64,
+    }
+}
+
+#[allow(clippy::expect_used)]
```
Arjentix, line 37: Looks like it's outdated

config/src/iroha.rs
```diff
@@ -0,0 +1,153 @@
...
+        /// `Kura` configuration
+        #[config(inner)]
+        pub kura: kura::Configuration,
+        /// `Sumeragi` configuration
+        #[config(inner)]
+        #[view(into = sumeragi::ConfigurationView)]
+        pub sumeragi: sumeragi::Configuration,
+        /// `Torii` configuration
+        #[config(inner)]
+        pub torii: torii::Configuration,
```
Arjentix, line 38: Hm, I think it's quite dangerous. We have to remember which configuration contains private keys and replace them with their `View` analogue.
I have an idea how to bypass this, but it may be too complex and probably we can do it in another issue.

The idea is the next. When using `view!` let's implement a new marker trait, i.e. called `PrivateData`, for the original struct. Then in another `view!` call we can check if a type of a field implements such trait (there are ways to do that). Unfortunately at compile-time we know nothing about time except his name, so this check will be at runtime. I.e. in `From` implementation. If some filed's type implements `PrivateData` then panic.

What to you think?

config/src/wsv.rs
```diff
@@ -0,0 +1,42 @@
...
+#[serde(rename_all = "UPPERCASE", default)]
+pub struct Configuration {
+    /// [`MetadataLimits`] for every asset with store.
+    pub asset_metadata_limits: MetadataLimits,
+    /// [`MetadataLimits`] of any asset definition metadata.
+    pub asset_definition_metadata_limits: MetadataLimits,
+    /// [`MetadataLimits`] of any account metadata.
+    pub account_metadata_limits: MetadataLimits,
+    /// [`MetadataLimits`] of any domain metadata.
+    pub domain_metadata_limits: MetadataLimits,
```
Arjentix, line 24: What about

```rust
struct MetadataLimits<T> {
    // ...
    _phantom: PhantomData<T>,
}
```

Not for this PR, just an idea for a future

data_model/Cargo.toml
```diff
@@ -47,6 +47,7 @@ warp = { version = "0.3", default-features = false, optional = true }
 thiserror = { version = "1.0.28", optional = true }
 getset = "0.1.2"
 strum = { version = "0.24.0", default-features = false, features = ["derive"] }
+url = { version = "2.2.2", features = ["serde"], optional = true }
```
Arjentix, line 50: Where is it used?

Arjentix (commented):
Very cool. Macro is suprisingly easy to read

data_model/Cargo.toml
Erigara: outdated, good catch, i will remove it

config/base/src/lib.rs
Erigara: Unfortunately, `rustfmt` enforces such formatting.

config/base/src/lib.rs
Erigara: Unfortunately, `rustfmt` enforces such formatting.

config/base/src/lib.rs
Erigara: You are right, i will add it to the docs.

config/base/derive/src/lib.rs
Erigara: Thanks to your comment i rethink `Default` implementation, it could be as simple as:
```rust
fn default() -> Self {
    Self::from(<#original_ident as Default>::default())
}
```

config/base/derive/src/lib.rs
Erigara: like this style

config/base/derive/src/lib.rs
```diff
@@ -500,3 +500,277 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+            #vis struct #ident #generics {
+                #(#fields),*
+            }
+        };
+        tokens.extend(stream);
+    }
+}
+
+struct ViewIgnore {
+    _ident: Ident,
```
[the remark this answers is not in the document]
Erigara: Yeah, absolutely, don't know about this feature, much cleaner implementation now.

config/src/wsv.rs
Erigara: Strong typing is always great, we should consider this refactoring

config/src/iroha.rs
Erigara: I'm totally agree that converting to `View` automatically would be great feature to have.
However i'm not sure that proposed solution would work, problem is that we must provide `struct` field types at compile time, or i miss something?
Problem would be easily solved if only rust has [negative_impls](https://doc.rust-lang.org/beta/unstable-book/language-features/negative-impls.html).

config/base/src/lib.rs
Arjentix: Interesting

config/src/iroha.rs
Arjentix: Yes, we have to provide them in compile time and we can automatically find the error only in runtime. But it's better not to find error at all, I think.

config/src/wsv.rs
Arjentix: Created #2509

config/base/derive/src/lib.rs
```diff
@@ -694,41 +728,28 @@ fn gen_impl_default(original: &ViewInput, view: &ViewInput) -> proc_macro2::Toke
...
         generics,
         ident: view_ident,
-        fields,
         ..
     } = view;
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
-    let field_idents = extract_field_idents(fields);
 
     quote! {
         impl #impl_generics Default for #view_ident #ty_generics #where_clause {
```
Arjentix, line 736: Consider using fully quilified paths like `std::default::Default`. This is important for macros, cause with just `Default` user can cheat with `use my_default::Default`

Arjentix (dismissed):

config/src/iroha.rs
Erigara: oh, i get what you are suggesting, we could use `static_assertion` for this purpose

config/base/derive/src/lib.rs
```diff
@@ -500,3 +500,336 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+    }
+}
+
+fn gen_assertions(view: &ViewInput) -> proc_macro2::TokenStream {
+    let ViewInput { fields, .. } = view;
+    let field_types = extract_field_types(fields);
+    let messages: Vec<String> = extract_field_idents(fields)
+        .iter()
+        .map(|ident| {
+            format!("Field `{ident}` has it's own view consider adding attribute #[view(into = ViewType)]")
```
Arjentix, line 767:
```suggestion
            format!("Field `{ident}` has it's own view, consider adding attribute #[view(into = ViewType)]")
```

config/base/derive/src/lib.rs
```diff
@@ -500,3 +500,336 @@ fn impl_configurable(ast: &DeriveInput) -> TokenStream {
...
+        })
+        .collect();
+    quote! {
+        /// Assert that every field of 'View' doesn't implement `HasView` trait
+        const _: () = {
+            use iroha_config_base::view::NoView;
+            #(
+                const _: () = assert!(!iroha_config_base::view::IsHasView::<#field_types>::IS_HAS_VIEW, #messages);
```
Arjentix, line 775: Wait, is it basic [std assert](https://doc.rust-lang.org/std/macro.assert.html)? Then how does this work at compile time?

Arjentix: I think we can add tests for this, like we have for `version` macro. That could be another issue, this one is already pretty big

Arjentix (approved):

config/base/derive/src/lib.rs
Erigara: Yeah, we could use basic `assert!` in `const` context since [rust 1.57](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html).

mversic:
I'll just say I don't like moving all the configuration into one place. Also, `iroha_config` really shouldn't depend on `iroha_data_model`. I hope you have considered this while reviewing @Arjentix , @SamHSmith

Erigara:
> I'll just say I don't like moving all the configuration into one place.

Could you please provide some arguments why we shouldn't do that?
We agreed with @appetrosyan on such decision.

 > Also, `iroha_config` really shouldn't depend on `iroha_data_model`.
 
 I agree that we should get rid of `iroha_data_model` at some point.
 But duplicating all structs like `PeerId`, `AcountId`, `MetadataLimits` doesn't seems like nice solution.
 And extracting this structs into their own crate will add large amount of work to the current PR and stretched it's scope even further.    
 Also it wroth noticing that new `iroha_config` != old `iroha_config`, whats earlier called `iroha_config` now lives in `iroha_config_base`.
 
 All in all thanks for your comment, i open to any suggestion to make iroha better piece of software.

mversic:
> We agreed with @appetrosyan on such decision.

Ok then

>  I agree that we should get rid of `iroha_data_model` at some point.

If you agree, all good then. Let's not do it in this PR

SamHSmith (changes requested):
You must rebase onto iroha2-dev. And you must squash all your commits into one *fix* commit. That commit should be called `[fix] #2422: Hide private keys in configuration endpoint response`.

Erigara:
> You must rebase onto iroha2-dev. And you must squash all your commits into one _fix_ commit. That commit should be called `[fix] #2422: Hide private keys in configuration endpoint response`.

i will squash commits on merge

Arjentix (approved):

SamHSmith (approved):

PR #2436, opened 2022-07-05
[refactor] #2369: Refactor permission validators

### LINKED ISSUE #2369: Fix unreadable `Grant instruction validator` error message
# Problem

Take a look at `default_permissions()` at `permissions_validators/src/public_blockchain/mod.rs`. This configuration is used for checking instrucitons in Iroha 2 by default right now.

The first part of validators configuration is called `grant_instruction_validator`. If this part fails then it produces error message that says pretty nothing:

```
Error: Transaction was rejected

Caused by:
   0: Transaction rejected due to insufficient authorisation
   1: Action not permitted: None of the instructions succeeded in Any permission check block with name: Grant instruction validator.
```

All we can retrieve from this message is that something is wrong with some `Grant` instruction. We don't know which instruction and what is exactly wrong.
That was a problem while debugging #2081

# Suggested solution

I, personally, think that using `Result` doesn't fit well for validators, because we can't tell what (`Err` or `Ok`) should be returned from validator if it gets unexpected instruction. Right now `Grant` validators return `Err`, but `Instruction` validators return `Ok`. We need a unified way to do it. We need 3 variants: `Ok`, `Err` and `Skip`. We could use `Option<Result<...>>` but it looks ugly in my opinion.

So my suggestions is to introduce another return type for validators and refactor complex validators construction. This will allow us to tell which validators have failed in `any_should_succeed()` and etc. Altough I think that `any_should_succeed()` and `all_should_succeed()` will be less usable then and we can introduce `noone_should_fail()` which would meen *"All should return `Skip` or `Ok`"*. And maybe `at_least_one_success()`

Arjentix:
### Description of the Change

* Type returned by `IsAllowed` changed from `Result` to `ValidatorVerdict`
* Added `Judge` trait returning `Result`. Judge usually gather validators together to produce result based on verdicts from validators
* Removed `IsAllowedBoxed`. To be able to parse permissions from genesis we should introduce an enum`JudgeBoxed` in future. That's should be anough to deserialize.

### Issue

* Closes #2369 

### Benefits

* More readable error message
* More meaningfull validators system

### Possible Drawbacks

* Very detailed error message. Maybe we should use `Display` instead of `Debug` for it


permissions_validators/src/private_blockchain/query.rs
```diff
@@ -375,177 +321,144 @@ impl IsAllowed<QueryBox> for OnlyAccountsData {
...
-                    Err("Only access to the assets of your account is permitted.".to_owned().into())
+                    ValidatorVerdict::Deny("Only access to the assets of your account is permitted.".to_owned())
                 }
             FindAllRoles(_) | FindAllRoleIds(_) | FindRoleByRoleId(_) => {
-                Err("Only access to roles of the same account is permitted.".to_owned().into())
+                ValidatorVerdict::Deny("Only access to roles of the same account is permitted.".to_owned())
             },
             FindAllActiveTriggerIds(_) | FindTriggersByDomainId(_) => {
-                Err("Only access to the triggers of the same account is permitted.".to_owned().into())
+                ValidatorVerdict::Deny("Only access to the triggers of the same account is permitted.".to_owned())
```
outoftardis, line 330:
```suggestion
                ValidatorVerdict::Deny("Only the access to the triggers of the same account is permitted.".to_owned())
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -49,51 +66,49 @@ pub trait IsRevokeAllowed: Debug {
...
-        &self,
-        authority: &AccountId,
-        instruction: &RevokeBox,
-        wsv: &WorldStateView,
-    ) -> Result<()> {
-        IsRevokeAllowed::check(self.as_ref(), authority, instruction, wsv)
+    /// Convert this object to a type implementing [`IsAllowed`] trait
+    ///
+    /// Could not use `impl<R: IsGrantAllowed> IsAllowed for R`
+    /// because of conflicting trait implementations
```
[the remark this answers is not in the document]
Arjentix: Both options are acceptable to me

permissions_validators/src/lib.rs
```diff
@@ -15,83 +16,29 @@ use iroha_data_model::{isi::*, prelude::*};
...
-                Box::new(validator)
-            }
-        }
-
-        impl From<$ty> for IsInstructionAllowedBoxed {
-            fn from(validator: $ty) -> Self {
-                let validator: HasTokenBoxed = validator.into();
-                Box::new(validator)
-            }
+macro_rules! try_evaluate_or_deny {
```
Erigara, line 19: Nice macro used a lot through codebase!

client/tests/integration/permissions.rs
```diff
@@ -18,16 +13,6 @@ use test_network::{PeerBuilder, *};
 
 use super::Configuration;
 
-const BURN_REJECTION_REASON: &str = "Failed to pass first check with Can\'t burn assets from another account. \
```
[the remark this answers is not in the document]
Arjentix: Yeah. I don't know why it was here

data_model/src/lib.rs
```diff
@@ -338,7 +338,7 @@ pub enum Value {
     /// [`struct@Hash`]
     Hash(Hash),
     /// Block
-    Block(BlockValue),
+    Block(Box<BlockValue>),
```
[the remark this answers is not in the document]
Arjentix: Exactly

core/src/tx.rs
```diff
@@ -32,27 +34,23 @@ use crate::{
 #[derive(Clone)]
 pub struct TransactionValidator {
     transaction_limits: TransactionLimits,
-
-    is_instruction_allowed: Arc<IsInstructionAllowedBoxed>,
-    is_query_allowed: Arc<IsQueryAllowedBoxed>,
-
+    instruction_judge: Arc<dyn Judge<Operation = Instruction> + Send + Sync>,
```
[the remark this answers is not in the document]
Arjentix: Indeed

core/src/tx.rs
```diff
@@ -135,8 +133,8 @@ impl TransactionValidator {
                         check_instruction_permissions(
                             account_id,
                             instruction,
-                            &self.is_instruction_allowed,
-                            &self.is_query_allowed,
+                            self.instruction_judge.as_ref(),
```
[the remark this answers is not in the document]
Arjentix: Unfortunately, it doesn't

core/src/smartcontracts/isi/permissions/combinators.rs
```diff
@@ -180,398 +108,30 @@ impl IsAllowed<Instruction> for CheckNested {
...
-    fn check_type(&self, validator_type: ValidatorType) -> Result<()> {
-        if let Ok(self_type) = self.validator_type() {
-            if self_type != validator_type {
-                return Err(ValidatorTypeMismatch {
-                    expected: validator_type,
-                    actual: self_type,
+                .least_permissive_with(|| self.check(authority, &pair_box.right_instruction, wsv)),
+            Instruction::Sequence(sequence_box) => {
+                if sequence_box.instructions.is_empty() {
+                    ValidatorVerdict::Skip
```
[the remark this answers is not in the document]
Arjentix: Hm.... I think it makes more sense than others variants. Why should we allow empty sequence? And why should we deny it? We don't know what to do with it, so let's just make no decision and skip

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -2,14 +2,6 @@
...
-// Currently we simply can't do the following:
-// impl <T: IsGrantAllowed> PermissionsValidator for T {}
-// when we have
-// impl <T: HasToken> PermissionsValidator for T {}
-/// Boxed validator implementing [`IsGrantAllowed`] trait.
-pub type IsGrantAllowedBoxed = Box<dyn IsGrantAllowed + Send + Sync>;
-
 /// Checks the [`GrantBox`] instruction.
 pub trait IsGrantAllowed: Debug {
     /// Checks the [`GrantBox`] instruction.
```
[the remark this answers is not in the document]
Arjentix: Indeed old comment now is misleading. I'll change it to *"Denies"*

core/src/smartcontracts/isi/permissions/combinators.rs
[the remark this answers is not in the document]
Erigara: Ok, make sense. I just think that if we allow empty sequence we can eliminate branching.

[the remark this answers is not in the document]
Arjentix: Let's keep this discussion unresolved to see other people opinions

core/src/smartcontracts/wasm.rs
```diff
@@ -432,13 +432,13 @@ impl<'wrld> Runtime<'wrld> {
         account_id: &AccountId,
         bytes: impl AsRef<[u8]>,
         max_instruction_count: u64,
-        is_instruction_allowed: Arc<IsInstructionAllowedBoxed>,
-        is_query_allowed: Arc<IsQueryAllowedBoxed>,
+        instruction_judge: Arc<dyn Judge<Operation = Instruction> + Send + Sync>,
```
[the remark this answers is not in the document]
Arjentix: Looks like it was the last one

cli/src/torii/routing.rs
```diff
@@ -57,7 +54,7 @@ impl VerifiedQueryRequest {
     pub fn validate(
         self,
         wsv: &WorldStateView,
-        query_validator: &IsQueryAllowedBoxed,
+        query_judge: &dyn Judge<Operation = QueryBox>,
```
appetrosyan, line 57: Just FYI, even though this is written as though it's a fat pointer to a `dyn` object, MIR is smart enough to devirtualise the call.

client/tests/integration/roles.rs
```diff
@@ -4,7 +4,7 @@ use std::{str::FromStr as _, time::Duration};
 
 use eyre::{eyre, Result};
 use iroha_client::client::{self, Client};
-use iroha_core::prelude::*;
+use iroha_core::{prelude::*, smartcontracts::permissions::prelude::Judge};
```
appetrosyan, line 7: Glob import from the prelude. I know it's pointless now, but it's more consistent.

core/src/smartcontracts/isi/block.rs
```diff
@@ -1,17 +1,17 @@
...
-use iroha_data_model::block_value::BlockValue;
 use iroha_telemetry::metrics;
 
 use super::*;
 
 impl ValidQuery for FindAllBlocks {
     #[metrics(+"find_all_blocks")]
     fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, query::Error> {
-        let mut blocks: Vec<BlockValue> = wsv
+        let mut blocks: Vec<_> = wsv
```
appetrosyan, line 10: Actually, an explicit type here makes a bit more sense.

core/src/smartcontracts/isi/permissions/checks.rs
```diff
@@ -45,15 +33,13 @@ pub fn check_instruction_permissions(
 fn check_permissions_directly(
     account_id: &AccountId,
     instructions: &[Instruction],
-    is_instruction_allowed: &IsInstructionAllowedBoxed,
+    is_instruction_allowed: &dyn Judge<Operation = Instruction>,
```
appetrosyan, line 36: Quick question,  have you considered an opaque type? I don't think the distinction matters right now, but in the future it might contrast with how we distinguish between ISI and WASM. Static dispatch could work here just as well.

core/src/smartcontracts/isi/permissions/checks.rs
```diff
@@ -159,117 +163,131 @@ pub fn check_query_in_expression(
 ///
 /// # Errors
 /// If a user is not allowed to execute one of the inner queries,
-/// given the current `validator`.
+/// given the current `judge`.
```
appetrosyan, line 166:
```suggestion
/// given the current [`Judge`].
```

appetrosyan: We brought this up on retro, having as many hyperlinks as possible is a good idea.

core/src/smartcontracts/isi/permissions/combinators.rs
```diff
@@ -2,172 +2,103 @@
 
 use super::*;
```
appetrosyan, line 3: We need to get rid of this module in favour of a common combinatorial API.

permissions_validators/src/public_blockchain/unregister.rs
```diff
@@ -17,35 +17,33 @@ declare_token!(
...
+        let asset_definition_id: AssetDefinitionId = ok_or_skip!(object_id.try_into());
         let registered_by_signer_account = wsv
             .asset_definition_entry(&asset_definition_id)
             .map(|asset_definition_entry| asset_definition_entry.registered_by() == authority)
             .unwrap_or(false);
         if !registered_by_signer_account {
-            return Err("Can't unregister assets registered by other accounts."
-                .to_owned()
-                .into());
+            return ValidatorVerdict::Deny(
```
appetrosyan, line 42: Consider adding a `fn deny(&self, reason: &str)` as an associated `fn`

data_model/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: Consider adding an annotation for these cases. We might need to turn them into `cfg` directives when feature-complete.

core/src/tx.rs
```diff
@@ -32,27 +30,23 @@ use crate::{
...
 
 impl TransactionValidator {
     /// Construct [`TransactionValidator`]
     pub fn new(
         transaction_limits: TransactionLimits,
-
-        is_instruction_allowed: Arc<IsInstructionAllowedBoxed>,
-        is_query_allowed: Arc<IsQueryAllowedBoxed>,
-
+        instruction_judge: InstructionJudgeArc,
```
appetrosyan, line 42: What is the reason for creating `InstructionJudgeArc`? Why couldn't it be e.g. `Arc<Box<dyn Judge<Operation = Instruction>>>`?

core/src/smartcontracts/isi/permissions/combinators.rs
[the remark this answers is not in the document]
appetrosyan: I'm more concerned with how other people will understand the behaviour. `allow` and `deny` are obvious. `skip` is neither. When _should_ you skip? When is the validator going to `skip`? What should you do if your query is `skip`ped? 

Once these questions are answered it makes sense to make the API more complex to reduce branching

core/src/smartcontracts/isi/permissions/has_token.rs
```diff
@@ -15,38 +13,73 @@ pub trait HasToken: Debug {
...
         &self,
         authority: &AccountId,
         instruction: &Instruction,
         wsv: &WorldStateView,
     ) -> std::result::Result<PermissionToken, String>;
+
+    /// Convert this object to a type implementing [`IsAllowed`] trait
+    ///
+    /// Could not use `impl<H: HasToken> IsAllowed for H`
+    /// because of conflicting trait implementations
```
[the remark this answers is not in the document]
appetrosyan: Doc comment it is. I'd also consider making this function public.

core/src/smartcontracts/isi/permissions/roles.rs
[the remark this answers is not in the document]
appetrosyan: Doc comment +  make function public. There should be a point to hiding the internals.

permissions_validators/src/public_blockchain/unregister.rs
```diff
@@ -87,16 +83,15 @@ impl HasToken for GrantedByAssetCreator {
...
     fn check(
         &self,
         authority: &AccountId,
         instruction: &GrantBox,
         wsv: &WorldStateView,
-    ) -> Result<()> {
-        let token: CanUnregisterAssetWithDefinition = extract_specialized_token(instruction, wsv)?;
+    ) -> ValidatorVerdict {
+        let token: CanUnregisterAssetWithDefinition =
+            ok_or_skip!(extract_specialized_token(instruction, wsv));
```
appetrosyan, line 94: This could benefit from postfix calling convention.

permissions_validators/src/public_blockchain/transfer.rs
```diff
@@ -123,36 +114,41 @@ impl IsGrantAllowed for GrantMyAssetAccess {
...
+            }
+        };
         if params.is_empty() {
-            return Ok(());
+            return ValidatorVerdict::Allow;
         }
 
-        let period = retrieve_period(&params)?;
-        let count = retrieve_count(&params)?;
+        let period = ok_or_deny!(retrieve_period(&params));
```
appetrosyan, line 141: We should consider https://doc.rust-lang.org/std/ops/trait.FromResidual.html

permissions_validators/src/lib.rs
```diff
@@ -15,83 +16,29 @@ use iroha_data_model::{isi::*, prelude::*};
...
     };
 }
 
-macro_rules! try_into_or_exit {
-    ( $ident:ident ) => {
-        if let Ok(into) = $ident.try_into() {
-            into
-        } else {
-            return Ok(());
+macro_rules! ok_or_skip {
```
appetrosyan, line 37: Consider making this a trait once `FromResidual` is stable.

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -0,0 +1,459 @@
...
+pub type OperationJudgeBoxed<O> = Box<dyn Judge<Operation = O> + Send + Sync>;
+/// Boxed [`Instruction`] judge
+pub type InstructionJudgeBoxed = OperationJudgeBoxed<Instruction>;
+/// Boxed [`QueryBox`] judge
+pub type QueryJudgeBoxed = OperationJudgeBoxed<QueryBox>;
+/// Boxed [`Expression`] judge
+pub type ExpressionJudgeBoxed = OperationJudgeBoxed<Expression>;
+
+/// [`Arc`] with generic judge
+pub type OperationJudgeArc<O> = Arc<dyn Judge<Operation = O> + Send + Sync>;
```
appetrosyan, line 17: OK, that answers my question.

core/src/tx.rs
appetrosyan: Sorry, I just noticed that this is an alias.

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -0,0 +1,459 @@
...
+pub type ExpressionJudgeArc = OperationJudgeArc<Expression>;
+
+/// The judge that gives the final decision
+/// whether or not the `operation` should be accepted.
+///
+/// Unlike [`IsAllowed`], this trait requires the [`Result`] to be returned.
+///
+/// The judge accumulates [`verdicts`](ValidatorVerdict) from all validators,
+/// makes a decision, and returns the result.
+pub trait Judge: std::fmt::Debug {
```
appetrosyan, line 32: Debug is almost never a good supertrait. Instead consider a blanket impl if the underlying type `impl Debug`.

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -0,0 +1,459 @@
...
+                    messages.push(format!("Validator {validator:?} denied: {reason}"));
+                }
+                ValidatorVerdict::Skip => {
+                    messages.push(format!("Validator {validator:?} skipped"));
+                }
+            }
+        }
+
+        Err(format!(
+            "None of the validators has allowed operation {operation:?}: {messages:#?}",
```
appetrosyan, line 118: Good

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -2,105 +2,232 @@
...
+    /// Get the type of validator required to check the operation
+    ///
+    /// Accepts `self` because of the [`NeedsPermissionBox`]
+    fn required_validator_type(&self) -> ValidatorType;
+}
 
-impl NeedsPermission for Instruction {}
+impl NeedsPermission for Instruction {
+    fn required_validator_type(&self) -> ValidatorType {
+        ValidatorType::Instruction
```
appetrosyan, line 35: Could this be a constant instead?

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -2,105 +2,232 @@
...
+    #[inline]
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(Ord::cmp(self, other))
+    }
+}
+
+// Deny < Skip < Allow
+impl Ord for ValidatorVerdict {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.as_u8().cmp(&other.as_u8())
```
appetrosyan, line 135: I'd rewrite this using the above comment. You won't need the comment and will be guaranteed that the comparison works regardless of architecture.

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -2,105 +2,232 @@
...
+    #[must_use]
+    #[inline]
+    pub fn least_permissive(self, other: Self) -> Self {
+        std::cmp::min(self, other)
+    }
+
+    /// Similar to [`least_permissive`](Self::least_permissive)
+    /// but won't compute `f` if `self` is [`Deny`](ValidatorVerdict::Deny)
+    #[must_use]
+    pub fn least_permissive_with(self, f: impl FnOnce() -> Self) -> Self {
```
appetrosyan, line 170: Functional API. Like it.

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -2,105 +2,232 @@
...
+            self
+        } else {
+            self.most_permissive(f())
         }
     }
 
-    impl FromStr for DenialReason {
-        type Err = Infallible;
+    // Not implemented as some trait implementation cause it should stay private
+    fn as_u8(&self) -> u8 {
```
appetrosyan, line 199: I'd consider adding a standard trait as well. For example `Deref` might be a candidate.

cli/src/torii/routing.rs
Arjentix: Oh, that's cool

client/tests/integration/roles.rs
Arjentix: I think it was auto-imported. Don't know why `clippy` didn't warn about redundant import

core/src/smartcontracts/isi/permissions/checks.rs
Arjentix: Do you mean something like enum or to introduce generics in this function?

core/src/smartcontracts/isi/permissions/combinators.rs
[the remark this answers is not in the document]
Arjentix: For an external user query or other operation can't be skipped, because on the highest level there is a judge which returns `Ok` or `Err`.
The reasons of skipping you can find in `ValidatorVerdict::Skip` documentation.

But it still may vary from one implementation to another. So when constructing judge we should be sure how does every validator inside of it work: when it skips, when denies and when allows

core/src/tx.rs
Arjentix: And also `Arc<Box<...>>` looks like extra-fat `Arc<...>`. Or is there any pros of it?

core/src/smartcontracts/isi/permissions/combinators.rs
Arjentix: Well, maybe someday... I'll add a todo

Arjentix: Is there an issue for common combinatorial API?

permissions_validators/src/lib.rs
Arjentix: I'll add a TODO

core/src/smartcontracts/isi/permissions/has_token.rs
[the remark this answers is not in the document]
Arjentix: It's a trait function so it's already public

core/src/smartcontracts/isi/permissions/roles.rs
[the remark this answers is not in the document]
Arjentix: Already public, cause it's in a trait

core/src/smartcontracts/isi/permissions/judge.rs
Arjentix: Ok, it worked for `Judge`.
But it won't work for `IsAllowed` because we want to use `Debug` to construct an error message

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: Unfortunately functions can not be const in traits and traits impl

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: Okay, I'll remove `as_u8()` then

permissions_validators/src/public_blockchain/unregister.rs
Arjentix: I think it's better to import `ValidatorVerdict::*` here to make code less clogged

core/src/tx.rs
appetrosyan: Not really.

core/src/smartcontracts/isi/permissions/mod.rs
appetrosyan: Alright. The current implementation looks good.

permissions_validators/src/lib.rs
```diff
@@ -15,83 +17,33 @@ use iroha_data_model::{isi::*, prelude::*};
...
-            }
-        }
-
-        impl From<$ty> for IsInstructionAllowedBoxed {
-            fn from(validator: $ty) -> Self {
-                let validator: IsRevokeAllowedBoxed = validator.into();
-                Box::new(validator)
-            }
+/// TODO: Use [`FromResidual`](https://doc.rust-lang.org/std/ops/trait.FromResidual.html)
+/// once it becomes stable
```
outoftardis, line 30: if it's a TODO, I think it should be a regular comment, not a doc one

```suggestion
// TODO: Use [`FromResidual`](https://doc.rust-lang.org/std/ops/trait.FromResidual.html)
// once it becomes stable
```

permissions_validators/src/lib.rs
```diff
@@ -15,83 +17,33 @@ use iroha_data_model::{isi::*, prelude::*};
...
 }
 
-macro_rules! try_into_or_exit {
-    ( $ident:ident ) => {
-        if let Ok(into) = $ident.try_into() {
-            into
-        } else {
-            return Ok(());
+/// TODO: Use [`FromResidual`](https://doc.rust-lang.org/std/ops/trait.FromResidual.html)
+/// once it becomes stable
```
outoftardis, line 41:
```suggestion
// TODO: Use [`FromResidual`](https://doc.rust-lang.org/std/ops/trait.FromResidual.html)
// once it becomes stable
```

core/src/smartcontracts/isi/permissions/checks.rs
appetrosyan: `impl Judge<Operation = Instruction>`. Dynamic dispatch is fine, though.

permissions_validators/src/public_blockchain/unregister.rs
appetrosyan: Yep. Just make sure it doesn't pollute the global scope.

core/src/smartcontracts/isi/permissions/mod.rs
appetrosyan: I meant 
```rust 
const REQUIRED_VALIDATOR_TYPE: ValidatorType = ValidatorType::Instruction
``` 

and replace each `self.required_validator_type()` with `Self::REQUIRED_VALIDATOR_TYPE`.

permissions_validators/src/private_blockchain/query.rs
```diff
@@ -1,371 +1,306 @@
...
-                .to_owned()
-                .into()),
-            FindAllRoleIds(_) => Ok(()), // In case you need to debug the permissions.
-            FindRoleByRoleId(_) => Err("Only access to roles of the same domain is permitted."
-                .to_owned()
-                .into()),
-            FindAllPeers(_) => Ok(()), // Can be obtained in other ways, so why hide it.
-            FindAllActiveTriggerIds(_) => Ok(()),
+            FindAllDomains(_) => {
+                Deny("Only the access to the domain of the account is permitted.".to_owned())
```
appetrosyan, line 38:
```suggestion
                Deny("Only access to the domain of the account is permitted.".to_owned())
```

Similarly for other error messages.

appetrosyan (dismissed):

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: Hmm, but `NeedsPermissionBox` is useless right now... But may be usefull for deseriasation.
If I won't find `NeedsPermissionBox` useful while implementing #2199 I'll remove it and make this be const as you proposed

permissions_validators/src/lib.rs
```diff
@@ -20,7 +21,7 @@ macro_rules! try_evaluate_or_deny {
     ($e:expr, $wsv:ident) => {
         match $e.evaluate($wsv, &Context::new()) {
             Ok(value) => value,
-            Err(err) => return ValidatorVerdict::Deny(err.to_string().into()),
+            Err(err) => return ValidatorVerdict::Deny(err.to_string()),
```
Erigara, line 24: @Arjentix `ValidatorVerdict::Deny` and not just `Deny` because inside macro?

Erigara (approved):
looks good

appetrosyan (approved):

PR #2445, opened 2022-07-06
[fix] #1480: Shut down on panic

### LINKED ISSUE #1480: Stop Iroha on panic

ales-tsurko:
### Description of the Change

Currently, when panic occurs, only the corresponding thread shuts down. But we need the whole program to shut down if a panic appeared on any thread. A panic hook is provided, which uses [`tokio::sync::Notify`](https://docs.rs/tokio/1.19.2/tokio/sync/struct.Notify.html) to gracefully shut down the program.

### Issue

Fixes #1480 

### Benefits

The program will shut down gracefully if a panic occurs on any thread.

### Possible Drawbacks

As we use globals (i.e. panic hook), it can affect tests in some cases. But the test environment should be close or same as the production one. Following contributors should have it in mind.

[A cargo config](https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html) or `std::process::exit` can be used instead, but that could leave garbage.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -4,7 +4,7 @@
 //!
 //! `Iroha` is the main instance of the peer program. `Arguments`
 //! should be constructed externally: (see `main.rs`).
-use std::{panic, path::PathBuf, process, sync::Arc};
+use std::{panic, path::PathBuf, sync::Arc};
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
 use config::Configuration;
@@ -21,6 +21,7 @@ use iroha_core::{
     IrohaNetwork,
 };
 use iroha_data_model::prelude::*;
+use quit;
 use tokio::{
     signal,
     sync::{broadcast, Notify},
@@ -144,7 +145,7 @@ where
         let hook = panic::take_hook();
         panic::set_hook(Box::new(move |info| {
             hook(info);
-            process::exit(1);
+            quit::with_code(1);
         }));
     }
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -370,16 +370,16 @@ mod tests {
     use super::*;
 
     #[tokio::test]
+    #[allow(clippy::panic)]
     async fn iroha_should_notify_on_panic() {
         let notify = Arc::new(Notify::new());
         let hook = panic::take_hook();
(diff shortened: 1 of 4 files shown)


appetrosyan (approved):

cli/src/lib.rs
```diff
@@ -242,7 +250,9 @@ where
             Arc::clone(&notify_shutdown),
         );
 
-        Self::start_listening_signal(notify_shutdown)?;
+        Self::start_listening_signal(Arc::clone(&notify_shutdown))?;
```
QuentinI, line 253: Sounds a bit confusing to me, maybe we should rename it to something like `listen_for_signals`?

cli/src/lib.rs
```diff
@@ -138,6 +138,14 @@ where
         .await
     }
 
+    fn prepare_panic_hook(notify_shutdown: Arc<Notify>) {
+        let hook = panic::take_hook();
+        panic::set_hook(Box::new(move |info| {
+            hook(info);
+            notify_shutdown.notify_one();
```
ilchu, line 145: I might have missed some detail, but wouldn't it require
[`notify_waiters()`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notify_waiters) 
to send the signal to both the telemetry and API servers? In the same way it was done
in the `cli::start_listening_signal()`, for example.

ales-tsurko: Oh, thanks!

cli/src/lib.rs
QuentinI: > It's not related to the task

Sure, that's just an observation, I approved regardless.

cli/src/lib.rs
ales-tsurko: Hm, with `notify_waiters` it doesn't pass the test. The notification keeps being awaiting.

ilchu: Then I guess it's fine as it is. Maybe asking @pesterev to take a look would
make sense, as he did the other `Notify` PR, but I'll approve for now.

ilchu (approved):

PR #2447, opened 2022-07-06
[feature] #1889: Add a build script to the parity-scale-decoder tool

### LINKED ISSUE #1889: Introduce Domain-scoped event-triggers

pesterev:
### Description of the Change

Add a build script that automatically updates sample binaries from sources in the parity-scale-decoder.

### Issue

Partially resolves #1889 

### Benefits

Automatically updates sample binaries from sources.

### Possible Drawbacks

None.


Arjentix (commented):
I'd create a new issue for that instead of that PR, cause with #1889 in the git log history it will looks misleading.

Also I think we can improve the build script. Let's store *json* files in `samples`. So build scirpt will do the following:
1. Deserialise `Account` from `samples/account.json`
2. Encode `Account` to `samples/account.bin`
3. Do steps 1 and 2 for `Domain`
4. Do steps 1 and 2 for `Trigger`

Arjentix (approved):

Issue #2098, opened 2022-04-12
Add query for block header


appetrosyan:

ritikBhandari:
Hi,
Can I take up this issue??

appetrosyan:
@ritikBhandari, please, be my guest. Let me know if you'd like a guided tour of the code-base to help you with your task.

ritikBhandari:
@appetrosyan I'll appreciate that.

QuentinI:
Closed in #2453

PR #2453, opened 2022-07-07
[feature] #2098: Block header queries

QuentinI:
### Description of the Change

Added queries for block headers.

### Issue

#2098


tools/parity_scale_decoder/src/generate_map.rs
```diff
@@ -148,6 +148,8 @@ pub fn generate_map() -> DumpDecodedMap {
         FindAllAssets,
         FindAllAssetsDefinitions,
         FindAllBlocks,
+        FindAllBlockHeaders,
+        FindBlockHeaderByHash,
```
Arjentix, line 152: I'd like to keep this list in alphabetical order as it was before

data_model/src/query.rs
```diff
@@ -1479,19 +1483,23 @@ pub mod transaction {
 }
 
 pub mod block {
-    //! Queries related to `Transaction`.
+    //! Queries related to `Block`.
```
Arjentix, line 1486:
```suggestion
    //! Queries related to [`Block`].
```
Or we can't make it to be a link?

data_model/src/query.rs
```diff
@@ -1522,9 +1530,68 @@ pub mod block {
...
+        Deserialize,
+        Serialize,
+        IntoSchema,
+        PartialOrd,
+        Ord,
+    )]
+    /// `FindBlockHeaderByHash` Iroha Query will find block header by block hash
+    pub struct FindBlockHeaderByHash {
+        /// Block hash hash.
+        pub hash: EvaluatesTo<Hash>,
```
Arjentix, line 1578:
```suggestion
        pub hash: EvaluatesTo<HashOf<VersionedCommittedBlock>>,
```

What about this? As I can see this is widely used pattern in our codebase

core/src/smartcontracts/isi/block.rs
```diff
@@ -10,10 +13,43 @@ impl ValidQuery for FindAllBlocks {
...
+impl ValidQuery for FindAllBlockHeaders {
+    #[metrics(+"find_all_block_headers")]
+    fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, query::Error> {
+        let mut blocks: Vec<BlockHeaderValue> = wsv
+            .blocks()
+            .map(|block| block.clone())
+            .map(VersionedCommittedBlock::into_value)
+            .map(|block_value| block_value.header)
+            .collect();
+        blocks.reverse();
```
Arjentix, line 33: Please, add some documentation why do you reverse it. I'd prefer add comment here and a doc comment for `FindAllBlockHeaders` describing the block order in the output

core/src/smartcontracts/isi/block.rs
```diff
@@ -10,10 +13,43 @@ impl ValidQuery for FindAllBlocks {
...
+            .hash
+            .evaluate(wsv, &Context::default())
+            .wrap_err("Failed to evaluate hash")
+            .map_err(|e| query::Error::Evaluate(e.to_string()))?
+            .typed();
+
+        let block = wsv
+            .blocks()
+            .find(|block| block.hash() == hash)
+            .ok_or_else(|| query::Error::Find(Box::new(FindError::Block(hash))))?;
```
Arjentix, line 51: Hm, looks like very long operation but let's keep it now. We don't know real world scenarios and don't need preventual optimisations

Arjentix:
Could you also implement some tests for these new quires? Or create an issue for that?

client/src/client.rs
```diff
@@ -1109,6 +1109,28 @@ pub mod asset {
     }
 }
 
+pub mod block {
```
outoftardis, line 1112: @appetrosyan I'd need to add these queries to the tutorial, right?

data_model/src/query.rs
```diff
@@ -1522,9 +1530,68 @@ pub mod block {
...
+        Eq,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+        PartialOrd,
+        Ord,
+    )]
+    /// `FindAllBlockHeaders` Iroha Query will list all block headers
```
outoftardis, line 1548:
```suggestion
    /// `FindAllBlockHeaders` Iroha Query lists all block headers
```

data_model/src/query.rs
```diff
@@ -1522,9 +1530,68 @@ pub mod block {
...
+        Eq,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+        PartialOrd,
+        Ord,
+    )]
+    /// `FindBlockHeaderByHash` Iroha Query will find block header by block hash
```
outoftardis, line 1575:
```suggestion
    /// `FindBlockHeaderByHash` Iroha Query finds a block header by block hash
```

data_model/src/query.rs
```diff
@@ -1522,9 +1530,68 @@ pub mod block {
...
+    pub struct FindBlockHeaderByHash {
+        /// Block hash hash.
```
outoftardis, line 1577:
```suggestion
        /// The hash of the block hash.
```

Erigara (commented):
Agree with @Arjentix that we should have tests for new queries.

tools/parity_scale_decoder/src/generate_map.rs
QuentinI: Right, didn't notice it was sorted

data_model/src/query.rs
QuentinI: It doesn't actually involve `Block` directly, same with `transaction` module. I guess it would be better to just say "blocks" and "transactions" there.

data_model/src/query.rs
QuentinI: Not sure, I just did it by analogy with `FindTransactionByHash`. I'll have a look if having typed hash is disadvantageous somehow.

core/src/smartcontracts/isi/block.rs
QuentinI: Right, that makes sense.

data_model/src/query.rs
QuentinI: It's a typo actually, it's should be just one layer of hash. Thanks for catching.

core/src/smartcontracts/isi/block.rs
QuentinI: There's no faster way to find a block by hash anyway, we would need to rework how the `Chain` stores blocks to do it.

appetrosyan: Chain might not store all blocks in memory in the future. The solution has to involve a `MerkleTree`, which would allow you to find the object given hash in $O(\log n)$

client/src/client.rs
appetrosyan: One subtlety is that the tutorial now needs to serve either the up to date guide, or the guide that is appropriate for LTS. Right now we have no process for separating one from the other.

client/src/client.rs
```diff
@@ -1109,6 +1109,28 @@ pub mod asset {
...
+    //! Module with queries related to blocks
+    use iroha_crypto::Hash;
+
+    use super::*;
+
+    /// Get query to find all blocks
+    pub fn all() -> FindAllBlocks {
```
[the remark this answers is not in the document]
appetrosyan: Or a constant.

client/src/client.rs
```diff
@@ -1109,6 +1109,28 @@ pub mod asset {
...
+        FindAllBlocks::new()
+    }
+
+    /// Get query to find all block headers
+    pub fn all_headers() -> FindAllBlockHeaders {
```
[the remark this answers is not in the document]
appetrosyan: Or a constant

data_model/src/query.rs
QuentinI: Ah, of course. We can't use `VersionedCommitedBlock` here, since it's defined in core.

Arjentix: Indeed

Arjentix (approved):
LGTM

core/src/smartcontracts/isi/block.rs
```diff
@@ -9,11 +13,44 @@ impl ValidQuery for FindAllBlocks {
...
+impl ValidQuery for FindAllBlockHeaders {
+    #[metrics(+"find_all_block_headers")]
+    fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, query::Error> {
+        let mut blocks: Vec<BlockHeaderValue> = wsv
+            .blocks()
+            .map(|block| block.clone())
+            .map(VersionedCommittedBlock::into_value)
+            .map(|block_value| block_value.header)
+            .collect();
+        blocks.reverse(); // Sorted by height desc.
```
pesterev, line 34:
```suggestion
            .rev() // Sorted by height desc.
            .collect();
```

core/src/smartcontracts/isi/block.rs
```diff
@@ -9,11 +13,44 @@ impl ValidQuery for FindAllBlocks {
     fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, query::Error> {
         let mut blocks: Vec<Box<BlockValue>> = wsv
             .blocks()
-            .map(|blk| blk.clone())
+            .map(|block| block.clone())
             .map(VersionedCommittedBlock::into_value)
             .map(Box::new)
             .collect();
-        blocks.reverse();
+        blocks.reverse(); // Sorted by height desc.
```
pesterev, line 20: you can reverse by iterating
```rust
            .rev() // Sorted by height desc.
            .collect(); 
```

Erigara: good catch! `ChainIterator` implements`DoubleEndedIterator` so it's possible.

Erigara (approved):

Arjentix (approved):

Arjentix (approved):

Erigara (approved):

PR #2471, opened 2022-07-12
[feature] #1769: Optimize pub/sub messaging

### LINKED ISSUE #1769: Reconsider broker model RFC
> S Sato, [2021/12/28 23:58]
> The current broker model has a couple of channels:
> producer-broker and broker-consumer.
> I think the equivalent is achieved with single watch::channel.
> 
> Also planning to:
> - Introduce Producer and Consumer traits
> - Introduce an optional backtrace to an event to expect the result

EDIT:
"a couple of channels" was my misunderstanding -- there was only broker-subscriber channels from the beginning.
Still, another optimization could be to reduce access to the broker

s8sato:
### Description of the Change

WIP.

- Establish all the possible channels at the components initialization
- Remove the broker

### Issue

- Closes #1769

### Benefits

- No broker access

### Possible Drawbacks

- Readability, if the boilerplate were pushed into macros

### Usage Examples or Tests

- 4082d4d1 implies the current implementation is not properly working
- f8e4dc0d is a blueprint -- not working due to lack of implementations

### Alternate Designs

The broker stores a receiver of 1:n channel on the publish declaration and clones on the subscription request:

```rust
use tokio::sync::watch::Receiver;

pub struct Broker {
    receivers: Arc<DashSet<Receiver<Box<dyn Msg>>>>,
}
```

### DIFF
--- a/actor/examples/broker.rs
+++ b/actor/examples/broker.rs
@@ -1,0 +1,91 @@
+use iroha_actor::{broker::*, prelude::*};
+
+struct Alice(Broker);
+
+struct Bob(Broker);
+
+struct Carol(Broker);
+
+#[derive(Clone, Debug, Message)]
+#[message(result = "()")]
+struct MsgXA;
+
+#[derive(Clone, Debug, Message)]
+#[message(result = "()")]
+struct MsgAB;
+
+#[derive(Clone, Debug, Message)]
+#[message(result = "()")]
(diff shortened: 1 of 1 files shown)


appetrosyan (approved):
Brilliantly simple.

appetrosyan:
I suggest closing this PR as the point was a PoC.

Issue #2457, opened 2022-07-08
Fix test flakyness introduced by Panic hook


ales-tsurko:
Integration tests on CI fail randomly. For example, you can check log for re-running jobs [here](https://github.com/hyperledger/iroha/runs/7241479252?check_suite_focus=true). The picture is the same locally especially on M1.

It's probably introduced by #2445 as the global panic hook may affect other tests.

ales-tsurko:
I can confirm it's introduced by the panic hook from #2445

PR #2474, opened 2022-07-12
[fix] #2457: Fix tests flakiness related to shut down on panic

ales-tsurko:

### Description of the Change

Added configuration to enable Iroha shutting down on panic. This prevents the tests expecting panic to fail. Also, the test for the responsible panic hook is now sequential, to prevent side effects on parallel run, while the hook is enabled.

### Issue

We [introduced](https://github.com/hyperledger/iroha/pull/2445) a panic hook, which shuts down Iroha on panic (which just shutting down the thread by default). But the hook affects other tests and make them flaky.

### Benefits

Prevents tests flakiness. Allows users to choose the behavior on panic.

### Possible Drawbacks

The biggest one is the difference between the test and the production environment. Also, as the test case for the hook is running sequentially, the overall tests performance has a very little degradation.

It may seem that `#[cfg(not(test))]` can be used here, but it looks like integration tests ignore it.

Updating tests to work with the hook would be the best solution, but it's very hard to find out which tests should be updated, because random tests are failed and it's not clear which tests caused that.

Switching to [`cargo-nextest`](https://nexte.st/) can also fix the issue.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -63,4 +63,5 @@ tokio = { version = "1.6.0", features = ["sync", "time", "rt", "io-util", "rt-mu
 warp = "0.3"
 
 [dev-dependencies]
+serial_test = "0.8.0"
 test_network = { version = "=2.0.0-pre-rc.5", path = "../core/test_network" }
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -29,6 +29,8 @@ pub struct Configuration {
     pub private_key: PrivateKey,
     /// Disable coloring of the backtrace and error report on panic.
     pub disable_panic_terminal_colors: bool,
+    /// Iroha will shutdown on any panic, if this option is set to `true`.
+    pub enable_shutdown_on_panic: bool,
     /// `Kura` related configuration.
     #[config(inner)]
     pub kura: KuraConfiguration,
@@ -71,6 +73,7 @@ impl Default for Configuration {
             public_key,
             private_key,
             disable_panic_terminal_colors: bool::default(),
+            enable_shutdown_on_panic: Default::default(),
             kura: KuraConfiguration::default(),
             sumeragi: sumeragi_configuration,
             torii: ToriiConfiguration::default(),
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -30,7 +30,7 @@ pub struct Configuration {
     /// Disable coloring of the backtrace and error report on panic.
     pub disable_panic_terminal_colors: bool,
     /// Iroha will shutdown on any panic if this option is set to `true`.
-    pub enable_shutdown_on_panic: bool,
+    pub shutdown_on_panic: bool,
     /// `Kura` related configuration.
     #[config(inner)]
     pub kura: KuraConfiguration,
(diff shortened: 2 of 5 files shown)


cli/src/config.rs
```diff
@@ -29,6 +29,8 @@ pub struct Configuration {
     pub private_key: PrivateKey,
     /// Disable coloring of the backtrace and error report on panic.
     pub disable_panic_terminal_colors: bool,
+    /// Iroha will shutdown on any panic, if this option is set to `true`.
```
outoftardis, line 32:
```suggestion
    /// Iroha will shutdown on any panic if this option is set to `true`.
```
or
```suggestion
    /// If this option is set to `true`, Iroha will shutdown on any panic.
```

docs/source/references/config.md
```diff
@@ -170,6 +171,16 @@ Has type `bool`. Can be configured via environment variable `IROHA_DISABLE_PANIC
 false
 ```
 
+## `enable_shutdown_on_panic`
+
+Iroha will shutdown on any panic, if this option is set to `true`.
```
outoftardis, line 176: same as in `config.rs`

cli/src/config.rs
ales-tsurko: Thanks!

cli/src/config.rs
```diff
@@ -29,6 +29,8 @@ pub struct Configuration {
     pub private_key: PrivateKey,
     /// Disable coloring of the backtrace and error report on panic.
     pub disable_panic_terminal_colors: bool,
+    /// Iroha will shutdown on any panic if this option is set to `true`.
+    pub enable_shutdown_on_panic: bool,
```
Erigara, line 33: maybe just `shutdown_on_panic`?

Erigara (approved):
looks good

cli/src/config.rs
```diff
@@ -71,6 +73,7 @@ impl Default for Configuration {
             public_key,
             private_key,
             disable_panic_terminal_colors: bool::default(),
+            enable_shutdown_on_panic: Default::default(),
```
QuentinI, line 76: Why not just `false`? It's both shorter and more direct.

ales-tsurko: Less refactoring in case of type changing. But I'll change it to `false`, thanks.

cli/src/config.rs
ales-tsurko: Yeah, it's better. I just followed above convention (`disable_panic_terminal_colors`)

QuentinI (approved):

cli/Cargo.toml
```diff
@@ -63,4 +63,5 @@ tokio = { version = "1.6.0", features = ["sync", "time", "rt", "io-util", "rt-mu
 warp = "0.3"
 
 [dev-dependencies]
+serial_test = "0.8.0"
```
appetrosyan, line 66: Using nextest as standard is still the preferred solution. But this guarantees that the tests pass locally using just `cargo test`.

cli/src/config.rs
```diff
@@ -71,6 +73,7 @@ impl Default for Configuration {
             public_key,
             private_key,
             disable_panic_terminal_colors: bool::default(),
+            shutdown_on_panic: false,
```
appetrosyan, line 76: So we're effectively disabling the panic hook. This is useful as a stop-gap solution, but I would prefer resolving the problem.

appetrosyan (approved):

Issue #2437, opened 2022-07-06
Refactor `#[derive(Filter)]` for better parsing and readability


ilchu:
As mentioned in #2431 PR discussion, there are still a few improvements that could be made to the `Filter` derive macro in `data_model/derive`:
* More modular functions
* Custom structs with `Parse` to make invalid enums fail more predictably
* Better naming conventions
* Less nested mapping

0x009922:
I made some research about the `Filter` macro. I am confused by its documentation.

There is an example of how macro expands from this:

```rust
use iroha_data_model_derive::{Filter, IdEqOrdHash};
use iroha_data_model::prelude::{HasOrigin, Identifiable};
use serde::{Deserialize, Serialize};


#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Filter, Deserialize, Serialize)]
pub enum LayerEvent {
    SubLayer(SubLayerEvent),
    Created(LayerId),
}

pub enum SubLayerEvent {
    Created(SubLayerId),
}

pub struct LayerId {
    name: u32,
}

#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct SubLayerId {
    name: u32,
    parent_id: LayerId,
}

#[derive(Debug, Clone, IdEqOrdHash)]
pub struct Layer {
    id: <Self as Identifiable>::Id,
}

#[derive(Debug, Clone, IdEqOrdHash)]
pub struct SubLayer {
    id: <Self as Identifiable>::Id,
}

impl HasOrigin for LayerEvent {
    type Origin = Layer;

    fn origin_id(&self) -> &<Layer as Identifiable>::Id {
        match self {
            Self::SubLayer(sub_layer) => &sub_layer.origin_id().parent_id,
            Self::Created(id) => id,
        }
    }
}

impl HasOrigin for SubLayerEvent {
    type Origin = SubLayer;

    fn origin_id(&self) -> &<SubLayer as Identifiable>::Id {
        match self {
            Self::Created(id) => id,
        }
    }
}
```

To this:

```rust
#[doc = " Filter for LayerEvent entity"]
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, derive_more::Constructor, Decode, Encode, Deserialize, Serialize, IntoSchema)]
pub struct LayerFilter {
    origin_filter:
[...]

PR #2476, opened 2022-07-13
[feature] #1988, #2437: Derive macros for Identifiable, Eq, Hash, Ord

### LINKED ISSUE #1988: `PartialEq` and `PartialOrd` of `Identifiable` structures
Identifiable structures, such as `Domain`, derive their implementation of `PartialEq`, but have the following implementation of `PartialOrd`:
```rs
impl PartialOrd for Domain {
    #[inline]
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.id().cmp(&other.id))
    }
}
```

Would it make sense to have equivalent implementation of `PartialEq` instead of deriving it?
Other `Identifiable` structures should be considered as well

ilchu:

-------

### Description of the Change

* Automatic derivation of `Ord` / `Eq` / `Hash` for `Identifiable` entities.
* Derivation of `Identifiable` in a separate macro as well.
* More readable and modular code for `Filter` derive with custom structures (as mentioned in #2437).

### Issue

Closes #1988, #2437.

### Benefits
* Always compliant implementation of `Ord`, `PartialOrd`, `Eq`, `PartialEq` and `Hash`.
* Reduced repetitive code with `impl Identifiable` previously scattered throughout `data_model`.
* More readable `Filter` derive as compared to #2431.

### Possible Drawbacks
The `Identifiable` derivation might be a bit frail: so far it relies on there being exactly three possible types of entities that can derive it: `Event` enums, `IdBox` or almost any other `Identifiable` struct in `data_model`, with only the first option involving somewhat complex processing. If any type has a different impl for whatever reason than the rest of the structs, such as `NewRole`, which calls its `inner` field as well, the derive can't work and the trait has to be implemented manually.


mversic (changes requested):

data_model/src/lib.rs
```diff
@@ -191,13 +194,13 @@ pub enum IdBox {
...
+//     type Id = Self;
 
-    fn id(&self) -> &Self::Id {
-        self
-    }
-}
+//     fn id(&self) -> &Self::Id {
+//         self
+//     }
+// }
```
mversic, line 203: dead code

data_model/src/role.rs
```diff
@@ -49,18 +50,19 @@ pub struct Id {
     Debug,
     Display,
     Clone,
-    PartialEq,
-    Eq,
+    OrdEqHash,
```
mversic, line 53: how come some Identifiables don't have `OrdEqHash`? Shouldn't all have it?

data_model/derive/src/lib.rs
```diff
@@ -197,33 +419,178 @@ fn impl_from_filter(ast: &syn::DeriveInput) -> TokenStream {
...
+/// A derive macro for `Identifiable` trait. Currently supports derivations only for
+/// `IdBox`, `Event` enums and structs from the `data_model` crate such as `Domain`
+/// or `NewDomain`. Only `NewRole` is unsupported among the latter as it has a somewhat
+/// unconventional implementation of the trait.
+///
+/// As such, the macro introduces a new
+/// outer attribute `id` for the entities it is derived upon. This attribute should
+/// be supplied with the associated type to be used in `impl Identifiable`. The type
+/// should be supplied as a string literal. The string literal should constitute a
+/// legal Rust type path for that reason.
```
outoftardis, line 496:
```suggestion
/// should be supplied as a string literal that constitutes a
/// legal Rust type path.
```

data_model/src/account.rs
```diff
@@ -124,9 +125,19 @@ impl Default for SignatureCheckCondition {
 /// Builder which should be submitted in a transaction to create a new [`Account`]
 #[allow(clippy::multiple_inherent_impl)]
 #[derive(
-    Debug, Display, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema,
+    Debug,
+    Display,
+    Clone,
+    OrdEqHash,
```
Arjentix, line 131: I'd add the information, that it's working only with `id` field in the name of this derive. Something like `IdBasedComparision`

data_model/src/account.rs
```diff
@@ -207,8 +196,7 @@ impl NewAccount {
     Debug,
     Display,
     Clone,
-    PartialEq,
-    Eq,
+    OrdEqHash,
```
Arjentix, line 199: That's can be dangerous, we should think twice here. Account is used in a lot of quires, instrucitons and collections. I'm not sure if we want `AccountA` be equal to `AccountB` if they have the same id. This is not how it was before.

Arjentix: The same thing for other structs where previously `Eq` was using all fields and now only uses `id`. I don't think that's a good idea. If we want check for *id* equality we can just `account_a.id() == account_b.id()`.

Arjentix (commented):
A little bit hard to understand what's going on on procedural macros (that's their nature). That's why I proposed to introduce comments for private items on our retro session.
If you can add some simple comments it would be helpful

data_model/derive/src/lib.rs
```diff
@@ -1,8 +1,288 @@
-//! Crate with `Filter` and `EventFilter` derive macro
-#![allow(clippy::expect_used, clippy::panic, clippy::too_many_lines)]
+//! A crate containing various derive macros for `data_model`
+//!
+#![allow(
+    clippy::expect_used,
+    clippy::too_many_lines,
+    clippy::eval_order_dependence,
```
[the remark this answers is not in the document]
ilchu: If you're speaking about `eval_order_dependence`, that one is triggered by `let content;` lines, and their subsequent calls by `_brace_token: syn::braced!(content in input)`. Even though this comes from official doc examples for the `braced!` macro in `syn`, it triggers that clippy warning, so I muted it.

data_model/src/account.rs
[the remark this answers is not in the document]
mversic: Exactly

data_model/src/account.rs
mversic: Considering that account id contains domain id, why is this not a good idea?

data_model/src/account.rs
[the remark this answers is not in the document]
mversic: Also, I think that there are only two possible implementations; comapre by id vs. compare by all fields. So this could always be derived. Not sure if that is a good idea

data_model/src/role.rs
mversic: What about trigger?

data_model/src/account.rs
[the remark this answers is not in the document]
mversic: > Don't quite understand what you mean by that. Are you saying that instead of having a separate `OrdEqHash` derive, we could put it into `Identifiable` derivation?

My question is also. Why don't we always derive `Ord`?

data_model/src/account.rs
Arjentix: I don't understand your example... I just mean that we should probably use all fields to check accounts for equality, not just id

s8sato:
I've just realized, and don't think events should implement `Identifiable` just for convenience to retrieve the id of the entity that the event refers to. An event will not be identifiable in general

data_model/src/account.rs
mversic: I just wanted to know the reason why we should require structural vs nominal equality for structs that are uniquely identifiable, i.e. that implement `Identifiable`

ilchu: Shouldn't the id always be unique and thus sufficient? I see that the only case where id fields can be same, but some other fields different is when two accounts come from different domains -- but then they probably shouldn't be compared like that in the first place. So comparison of all fields would only give us a way to catch illegal duplicates, no?

Arjentix: Yes, I'm thinking about something like illegal duplicates. I just had a bad experience with PermissionToken when it had a bug in its comparison traits and now I'm afraid of the same thing here.
But ok, maybe it's not that dangerous. If all our tests will pass then I have no claims

ilchu:
> I've just realized, and don't think events should implement `Identifiable` just for convenience to retrieve the id of the entity that the event refers to. An event will not be identifiable in general

They had a manual implementation before. Do you think we should get rid of that altogether? That would actually make the proc macro code a lot easier as generating match arms was the most complex part.

data_model/src/account.rs
Arjentix: Not sure about this idea

data_model/src/account.rs
mversic: @Arjentix I get where you're coming from, though `PermissionToken` was never `Identifiable`. If a type is `Identifiable`, which means it's `Id` is unique to the blockchain, then there should be no reason why not to use this optimization.

The only moment that I can think of when you would produce a duplicate is for `Registrable`s (should they be `Identifiable` at all?), when trying to register a new `Identifiable` like `Domain` with the same `Id`. But we reject these transactions immediately

data_model/src/account.rs
ilchu: Changed it to `IdOrdEqHash` as the two macros have been merged.

data_model/src/lib.rs
```diff
@@ -164,10 +165,11 @@ pub enum Parameter {
...
     Clone,
-    PartialEq,
-    Eq,
-    PartialOrd,
-    Ord,
+    IdOrdEqHash,
+    // PartialEq,
+    // Eq,
+    // PartialOrd,
+    // Ord,
```
mversic, line 172: 1. commented code
2. we shouldn't need any of these impls on boxed enums

to give you the context, these enums are there just to be able to serialize/encode trait objects. They should be converted into concrete types as soon as deserialized/decoded, otherwise don't use them. Most likely, there is some trait bound that requires these bounds so you won't be able to remove `IdOrdEqHash`

data_model/src/domain.rs
```diff
@@ -147,21 +135,12 @@ impl NewDomain {
...
-}
-
 /// Named group of [`Account`] and [`Asset`](`crate::asset::Asset`) entities.
 #[derive(
     Debug,
     Display,
     Clone,
-    PartialEq,
-    Eq,
+    IdOrdEqHash,
```
mversic, line 143: I assume we settled on this name instead of `Identifiable`?

data_model/derive/src/lib.rs
```diff
@@ -14,216 +96,578 @@ use quote::{format_ident, quote};
...
+                Eq,
+                Debug,
+                Decode,
+                Encode,
+                Deserialize,
+                Serialize,
+                IntoSchema,
+                Hash,
+            )]
+            #[allow(clippy::enum_variant_names, missing_docs)]
```
mversic, line 655: please don't allow `missing_docs` on public items. You can store documentation into `EventEnum`

data_model/derive/src/lib.rs
```diff
@@ -1,8 +1,288 @@
-//! Crate with `Filter` and `EventFilter` derive macro
-#![allow(clippy::expect_used, clippy::panic, clippy::too_many_lines)]
+//! A crate containing various derive macros for `data_model`
```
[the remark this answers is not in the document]
mversic: I was thinking two files would be more readable :)

data_model/derive/src/lib.rs
```diff
@@ -14,216 +96,578 @@ use quote::{format_ident, quote};
...
+                    ident: input.parse()?,
+                    _generics: input.parse()?,
+                    _brace_token: syn::braced!(content in input),
+                    struct_fields: Some(content.parse_terminated(Field::parse_named)?),
+                    _enum_variants: None,
+                    id_type,
                 })
             } else {
-                None
+                // Only used in [`IdBox`] case
```
mversic, line 258: I get the feeling that implementing `Identifiable` on enums doesn't make sense? if you agree, I would rather make this implementation manual and remove the else branch

mversic (changes requested):

data_model/src/role.rs
mversic: you have full control over the macro. You could attach the `Filter` bound on the `Ord`, `Eq`, `Hash` impls for `Trigger`

data_model/src/domain.rs
```diff
@@ -71,16 +72,26 @@ impl From<GenesisDomain> for Domain {
...
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    Identifiable,
 )]
 #[allow(clippy::multiple_inherent_impl)]
 #[display(fmt = "[{id}]")]
+#[id(type = "<Domain as Identifiable>::Id")]
 pub struct NewDomain {
```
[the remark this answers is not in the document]
mversic: I'd be happier if `NewObject`s didn't implement `Identifiable` but I'm not sure that's possible in the current imlementation

data_model/derive/src/lib.rs
```diff
@@ -14,216 +96,578 @@ use quote::{format_ident, quote};
...
-                .map(|inner_filter_ident| {
-                    Some(quote! {
-                        #event_filter_variant_ident (#import_path::FilterOpt<#inner_filter_ident>)
-                    })
+mod id {
+    use super::*;
+
+    pub(super) struct IdInput {
+        _attrs: Vec<Attribute>,
+        _vis: Visibility,
```
mversic, line 225: why do you have `_` prefix?

data_model/derive/src/lib.rs
```diff
@@ -14,216 +96,578 @@ use quote::{format_ident, quote};
...
+///    pub enum DomainEvent {
+///        Account(AccountEvent),
+///        AssetDefinition(AssetDefinitionEvent),
+///        Created(DomainId),
+///        Deleted(DomainId),
+///        MetadataInserted(DomainId),
+///        MetadataRemoved(DomainId),
+///    }
+///
+/// // The following lengthy code will be derived
```
mversic, line 113:
```suggestion
///    // The following lengthy code will be derived
```

data_model/src/domain.rs
ilchu: I guess, if nobody minds. It is both succinct and distinct enough to my taste.

data_model/derive/src/lib.rs
ilchu: Even though the `attrs` were used for finding the attribute with `parse_id_attribute`, they weren't used in the structure per se. Removed that from the struct entirely.

mversic: And what about the other variables? I see this pattern is pervasive

ilchu: Done, just left it out of the `Parse`d structs

data_model/derive/src/lib.rs
ilchu: It's there to prevent lints screaming at no documentation for enum variants, such as `ByMetadataRemoved`. I believe they're quite self-descriptive, so it's hard to come up with useful docs, and the same allow attribute was in the original code as well.

mversic (approved):

data_model/src/role.rs
ilchu: Created an [issue](https://github.com/hyperledger/iroha/issues/2513).

data_model/derive/src/lib.rs
```diff
@@ -1,8 +1,80 @@
...
+///    /// The identification associated with the domain builder.
+///    id: <Domain as Identifiable>::Id,
+///    /// The (IPFS) link to the logo of this domain.
+///    logo: Option<IpfsPath>,
+///    /// Metadata associated with the domain builder.
+///    metadata: Metadata,
+/// }
+///
+/// // The following impls will be derived
+/// impl Identifiable for NewDomain {
```
mversic, line 43:
```suggestion
/// impl Identifiable for Domain {
```

data_model/derive/src/lib.rs
```diff
@@ -1,8 +1,80 @@
...
+/// anything concerning the trigger `Set` structure, or anywhere else where the specifics of `Ord`
+/// implementation become important, as subtle bugs may appear.
+///
+/// Example:
+/// ```rust
+///
+/// // For a struct decorated like this
+/// #[derive(IdOrdEqHash)]
+/// #[id(type = "<Domain as Identifiable>::Id")]
+/// pub struct NewDomain {
```
mversic, line 33:
```suggestion
/// pub struct Domain {
```

data_model/derive/src/lib.rs
```diff
@@ -1,8 +1,80 @@
...
+/// legal Rust type path.
+///
+/// As this macro also derives an implementation of `Ord`, `PartialOrd`, `Eq`, `PartialEq` and `Hash` traits that always
+/// conforms to the same implementation principles based on ids of the entities.
+/// Thus none of the entities that derive this macro should derive neither of the aforementioned traits,
+/// as they will be overridden.
+///
+/// Another warning: additional care should be exercised when using this macro for types that could be connected with
+/// anything concerning the trigger `Set` structure, or anywhere else where the specifics of `Ord`
+/// implementation become important, as subtle bugs may appear.
```
mversic, line 25: I don't like seeing this comment especially since you've opened the issue.
May I ask what's the big problem here with `TriggerSet`?

data_model/derive/src/id.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+pub(super) fn impl_id(ast: &IdInput) -> TokenStream {
+    let id = ast.id_type.clone();
+    let name = ast.ident.clone();
+
+    let ordeqhash = impl_ordeqhash(ast);
+
+    let body = if ast.ident.to_string().starts_with("NewRole") {
```
mversic, line 71: uh, I'm not happy about this hardcoding. Can't we do something about it?

data_model/derive/src/id.rs
```diff
@@ -0,0 +1,120 @@
...
+            fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+                self.id().hash(state);
+            }
+        }
+    }
+}
+
+pub(super) fn impl_id(ast: &IdInput) -> TokenStream {
+    let id = ast.id_type.clone();
+    let name = ast.ident.clone();
```
mversic, line 67:
```suggestion
    let id = &ast.id_type;
    let name = &ast.ident;
```

data_model/derive/src/id.rs
```diff
@@ -0,0 +1,120 @@
...
+        let content;
+        let _brace_token = syn::braced!(content in input);
+        let _struct_fields = content.parse_terminated::<Field, Token![,]>(Field::parse_named)?;
+
+        Ok(IdInput { ident, id_type })
+    }
+}
+
+fn impl_ordeqhash(ast: &IdInput) -> proc_macro2::TokenStream {
+    let name = ast.ident.clone();
```
mversic, line 33:
```suggestion
    let name = &ast.ident;
```

data_model/derive/src/lib.rs
ilchu: Added this as requested by @Arjentix in [this comment](https://github.com/hyperledger/iroha/pull/2476#discussion_r920046366). Might not be relevant anymore.

mversic: I see. I think it should be the other way around. There is a contract this derive and `Identifiable` offer and that is uniqueness of `Id`. If your structure can't be uniquely identifiable it should not in any circumstance implement `Identifiable` or this derive. That's the rule, there are no exceptions

ilchu: Changed the doc to reflect that.

data_model/derive/src/filter.rs
```diff
@@ -0,0 +1,316 @@
...
+            Hash,
+        )]
+        #[allow(clippy::enum_variant_names, missing_docs)]
+        #[doc = #event_filter_doc]
+        #vis enum #event_filter_ident #generics {
+            #(#id_variants),*,
+            #(#event_variants),*
+        }
+
+        impl Filter for #event_filter_ident {
```
Arjentix, line 305: What if `Filter` is not on the scope? Or another `Filter` trait is on the scope? I suggest to use fully quilified path

data_model/derive/src/lib.rs
```diff
@@ -1,8 +1,83 @@
...
+///     }
+/// }
+/// impl core::cmp::Eq for Domain {}
+/// impl core::hash::Hash for Domain {
+///     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+///         self.id().hash(state);
+///     }
+/// }
+/// ```
+#[proc_macro_derive(IdOrdEqHash, attributes(id))]
```
Arjentix, line 76: I like that naming

data_model/derive/src/id.rs
Arjentix: We can also introduce macro attribute so that it will look like:

```rust
#[derive(IdOrdEqHash)]
#[id(type = "Id")]
#[id(id = "inner.id")]
pub struct NewRole {
    inner: Role,
}
```

Arjentix: This can be another issue btw

data_model/src/domain.rs
Arjentix: Yeah, I like it

data_model/src/domain.rs
```diff
@@ -71,16 +72,17 @@ impl From<GenesisDomain> for Domain {
...
-    /// The identification associated to the domain builder.
+    /// The identification associated with the domain builder.
     id: <Domain as Identifiable>::Id,
     /// The (IPFS) link to the logo of this domain.
     logo: Option<IpfsPath>,
-    /// metadata associated to the domain builder.
+    /// Metadata associated with the domain builder.
     metadata: Metadata,
 }
```
Arjentix, line 87: Why can't we infer type from `id` field? Can `#[id(type = "...")]` be optional?

Arjentix (dismissed):

data_model/src/domain.rs
ilchu: Created an [issue](https://github.com/hyperledger/iroha/issues/2521).

data_model/derive/src/id.rs
ilchu: Created #2522.

mversic (dismissed):

mversic (dismissed):

mversic (approved):

mversic (approved):

Arjentix (approved):

PR #2484, opened 2022-07-14
[feature] #2252: Use errno for FFI error handling

ales-tsurko:

# TBD

### DIFF
--- a/data_model/derive/src/filter.rs
+++ b/data_model/derive/src/filter.rs
@@ -1,0 +1,316 @@
+#![allow(
+    clippy::expect_used,
+    clippy::eval_order_dependence,
+    clippy::unwrap_in_result
+)]
+
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{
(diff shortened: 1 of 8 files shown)


ffi/derive/src/bindgen.rs
```diff
@@ -37,7 +37,7 @@ pub fn gen_ffi_fn(fn_descriptor: &FnDescriptor) -> proc_macro2::TokenStream {
     quote! {
         #[doc = #ffi_fn_doc]
         #[no_mangle]
-        pub unsafe extern "C" fn #ffi_fn_name(#(#self_arg,)* #(#fn_args,)* #output_arg) -> iroha_ffi::FfiResult {
```
mversic, line 40: please keep returning `FfiResult` fieldless enum. How will the caller of the function know if something went wrong and whether they need to poll for error?

mversic: We don't want the user to poll for error on every FFI call

PR #2493, opened 2022-07-15
[fix] #2492: Fixes not all triggers being executed that match an event

### LINKED ISSUE #2492: [BUG] Trigger isn't executed on an event
### GIT commit hash

91da336a

### Minimum working example

I've written a test that reproduces a bug. [See](https://github.com/hyperledger/iroha/pull/2493/files#diff-d035124a94ee4403c63779770602a22ae84e47c6b0c71905605dedc46f2e88d0R9)

### Expected behaviour

Must execute all triggers that match an event.

### Actual behaviour

Executes triggers that match an event only until it meets a trigger that doesn't match. This skips some actual/matched triggers.

### Operating system

Any

### LibC type and version

Any

### Current environment

Source code build

### Logs in JSON format

<details>
  
  ```json
  ```
</details>

### Who can help?

@Arjentix

pesterev:

### Description of the Change

I've added a test that reproduces [this bug](https://github.com/hyperledger/iroha/issues/2492) and also changes that fix it. Also [see](https://github.com/hyperledger/iroha/pull/2495) these changes to `iroha2-lts` branch.

### Issue
Fixes #2492 

### Benefits
Fixes triggers execution

### Possible Drawbacks
None

### DIFF
--- a/client/tests/integration/triggers/data_trigger.rs
+++ b/client/tests/integration/triggers/data_trigger.rs
@@ -21,7 +21,7 @@ fn must_execute_both_triggers() -> Result<()> {
     let register_trigger = RegisterBox::new(Trigger::new(
         "mint_rose_1".parse()?,
         Action::new(
-            Executable::from(vec![instruction.into()]),
+            Executable::from(vec![instruction.clone().into()]),
             Repeats::Indefinitely,
             account_id.clone(),
             FilterBox::Data(BySome(DataEntityFilter::ByAccount(BySome(
@@ -31,7 +31,6 @@ fn must_execute_both_triggers() -> Result<()> {
     ));
     test_client.submit_blocking(register_trigger)?;
 
-    let instruction = MintBox::new(1_u32, asset_id.clone());
     let register_trigger = RegisterBox::new(Trigger::new(
         "mint_rose_2".parse()?,
         Action::new(
--- a/data_model/src/trigger/set.rs
+++ b/data_model/src/trigger/set.rs
@@ -320,7 +320,7 @@ impl Set {
                 count = min(atomic.get(), count);
             }
             if count == 0 {
-                return;
+                continue;
             }
 
             let ids = vec![


client/tests/integration/triggers/data_trigger.rs
```diff
@@ -0,0 +1,62 @@
...
+            account_id,
+            FilterBox::Data(BySome(DataEntityFilter::ByDomain(BySome(
+                DomainFilter::new(AcceptAll, BySome(DomainEventFilter::ByCreated)),
+            )))),
+        ),
+    ));
+    test_client.submit_blocking(register_trigger)?;
+
+    test_client.submit_blocking(RegisterBox::new(Account::new(
+        "buggy@wonderland".parse()?,
```
Arjentix, line 48:
```suggestion
        "bunny@wonderland".parse()?,
```

Arjentix (commented):
Wait... In your test you:
1. Register a trigger which should be executed on a new account creation
2. Register a trigger which should be executed on a new domain creation
3. Register new account
4. Expect both trigger to be executed? Why? No new domain was created, why the second trigger should be executed too? And why does this work...?

Upd: My bad, I've missed the line with new domain registration

Arjentix:
Also if it's the real bug we need the same PR (or without test) for `iroha2-lts`

Arjentix:
My bad, I've missed the line with new domain registration

Arjentix (approved):

Arjentix (approved):

Arjentix (approved):

mversic (dismissed):

Arjentix (approved):

mversic (commented):
you say that you've added a test but there is no test

pesterev:
@mversic, that's right, I've added the test (must_execute_both_triggers) to [this PR](https://github.com/hyperledger/iroha/pull/2418) and here. The PR about domain triggers was merged yesterday, and now it doesn't show that test in diff.

mversic (approved):

PR #2495, opened 2022-07-16
[fix] #2492: Fixes not all triggers being executed that match an event (lts)

pesterev:

### Description of the Change

It adds a test that reproduces [this bug](https://github.com/hyperledger/iroha/issues/2492) and also changes that fix it. Also [see](https://github.com/hyperledger/iroha/pull/2493) these changes to `iroha2-dev` branch.

### Issue

Fixes #2493 

### Benefits

Adds changes from #2493 directly into the LTS branch.

### Possible Drawbacks

None

### DIFF
--- a/client/tests/integration/triggers/data_trigger.rs
+++ b/client/tests/integration/triggers/data_trigger.rs
@@ -1,0 +1,61 @@
+#![allow(clippy::restriction)]
+
+use eyre::Result;
+use iroha_client::client;
+use iroha_data_model::prelude::*;
+use test_network::*;
+
+#[test]
+fn must_execute_both_triggers() -> Result<()> {
+    let (_rt, _peer, mut test_client) = <PeerBuilder>::new().start_with_runtime();
+    wait_for_genesis_committed(&vec![test_client.clone()], 0);
+
+    let account_id: AccountId = "alice@wonderland".parse()?;
+    let asset_definition_id = "rose#wonderland".parse()?;
+    let asset_id = AssetId::new(asset_definition_id, account_id.clone());
+
+    let prev_value = get_asset_value(&mut test_client, asset_id.clone())?;
+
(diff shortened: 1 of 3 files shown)


client/tests/integration/triggers/data_trigger.rs
```diff
@@ -0,0 +1,62 @@
...
+            Repeats::Indefinitely,
+            account_id.clone(),
+            FilterBox::Data(BySome(DataEntityFilter::ByAccount(BySome(
+                AccountFilter::new(AcceptAll, BySome(AccountEventFilter::ByCreated)),
+            )))),
+        ),
+    ));
+    test_client.submit_blocking(register_trigger)?;
+
+    let instruction = MintBox::new(1_u32, asset_id.clone());
```
Arjentix, line 33: You can remove this line, cause you already have it above. Just clone instruction in the second trigger

Arjentix (dismissed):

Arjentix (dismissed):

mversic (approved):

client/tests/integration/triggers/data_trigger.rs
```diff
@@ -0,0 +1,61 @@
...
+    )))?;
+    test_client.submit_blocking(RegisterBox::new(Domain::new("neverland".parse()?)))?;
+
+    let new_value = get_asset_value(&mut test_client, asset_id)?;
+    assert_eq!(new_value, prev_value + 2);
+
+    Ok(())
+}
+
+fn get_asset_value(client: &mut client::Client, asset_id: AssetId) -> Result<u32> {
```
Erigara, line 58: One small thing to consider (probably in different ticket).
We have `get_asset_value` defined in 4 different places, seems like we need to extract this function to the place with common integration test functionality.

Erigara (approved):

PR #2497, opened 2022-07-18
[refactor]: impl `HasOrigin` instead of `Identifiable` for the data events

s8sato:
### Description of the Change

The minimum change to [remove `Identifiable` implementation from events](https://github.com/hyperledger/iroha/pull/2476#issuecomment-1184107169)

### Issue

- Relates to #2476

### Benefits

- Less complexity to derive `Identifiable`
- Rationality that an event should not be identified by its origin

### Possible Drawbacks

None


data_model/src/events/data/filters.rs
```diff
@@ -94,50 +94,60 @@ impl Filter for EntityFilter {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+}
+
+impl<T: Origin> core::hash::Hash for OriginFilter<T> {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+    }
+}
```
s8sato, line 131: These implementations should be consistent each other. See [derive_hash_xor_eq](https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq)

data_model/src/events/data/filters.rs
```diff
@@ -94,50 +94,60 @@ impl Filter for EntityFilter {
...
+    <<T as Origin>::Origin as Identifiable>::Id: Deserialize<'de>,
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let origin_id = <T::Origin as Identifiable>::Id::deserialize(deserializer)?;
+        Ok(Self::new(origin_id))
     }
 }
```
s8sato, line 144: This is an alternative to forcing `Identifiable::Id` to be `Deserialize`, otherwise it would bring larger change

data_model/src/lib.rs
```diff
@@ -720,7 +712,20 @@ where
...
+        + PartialEq
+        + Eq
+        + PartialOrd
+        + Ord
+        + core::hash::Hash
+        + Decode
+        + Encode
+        // + Deserialize<'a>
+        + Serialize
+        + IntoSchema;
```
s8sato, line 728: here

data_model/src/lib.rs
```diff
@@ -193,14 +193,6 @@ pub enum IdBox {
     RoleId(<role::Role as Identifiable>::Id),
 }
 
-impl Identifiable for IdBox {
-    type Id = Self;
-
-    fn id(&self) -> &Self::Id {
-        self
-    }
-}
```
s8sato, line 202: Deleted as not in use. Can be implemented again as needed

data_model/src/events/data/filters.rs
ilchu: Do you mean this in terms of the amount of generated code, or would it require manual impls?

ilchu (approved):

data_model/src/lib.rs
mversic: `PartialEq` and `PartialOrd` are implied

mversic: I wouldn't place `IntoSchema` bound on the type if we can do without it. It's somewhat of an internal detail

mversic (dismissed):

data_model/src/events/data/filters.rs
```diff
@@ -94,50 +94,60 @@ impl Filter for EntityFilter {
...
-    Serialize,
-    IntoSchema,
-    Hash,
-)]
-/// Filter for identifiers
-///
-/// Passes id trough filter, if it equals to the one provided in construction
-pub struct IdFilter<Id: Eq>(Id);
+#[derive(Clone, PartialOrd, Ord, Eq, Debug, Decode, Encode, Serialize, IntoSchema)]
+/// Filter that accepts an data event whose origin matches what this filter specifies.
```
outoftardis, line 98:
```suggestion
/// Filter that accepts a data event with the matching origin.
```

data_model/src/events/data/events.rs
```diff
@@ -236,11 +320,31 @@ mod trigger {
     }
 }
 
+/// Has an origin whose change caused this event.
+pub trait Origin {
+    /// Type of the origin whose change caused this event.
+    type Origin: Identifiable;
```
mversic, line 326: How about?

```suggestion
    type Origin: Identifiable::Id;
```

data_model/src/events/data/filters.rs
s8sato: I mean manual changes caused by the lifetime parameter. The idea is to minimize changes so as not to disturb #2476

data_model/src/events/data/events.rs
s8sato: Assuming your suggestion is literally

```suggestion
    type Origin = Identifiable::Id;
```
, _associated type defaults are unstable_

mversic: I see, sorry

data_model/src/events/data/filters.rs
```diff
@@ -95,10 +95,15 @@ impl Filter for EntityFilter {
 }
 
 #[derive(Clone, PartialOrd, Ord, Eq, Debug, Decode, Encode, Serialize, IntoSchema)]
-/// Filter that accepts an data event whose origin matches what this filter specifies.
-pub struct OriginFilter<T: Origin>(<T::Origin as Identifiable>::Id);
+/// Filter that accepts a data event with the matching origin.
+pub struct OriginFilter<T: HasOrigin>(<T::Origin as Identifiable>::Id)
+where
+    <T::Origin as Identifiable>::Id: IntoSchema;
```
s8sato, line 101: This where clause may be introduced, but I don't see the value of leaving room for not implementing `IntoSchema` for `Id`

data_model/src/lib.rs
s8sato: https://github.com/hyperledger/iroha/pull/2497#discussion_r923307138

mversic: I would say that one should specify bounds on associated types only if the bound is used from within the trait. Otherwise, the the bound requirement is external to the trait and should be controlled from outside with `where` clause. This is similar to the fact that defining bounds on a struct definition is considered a bad practice. Rather you specify bounds on impl blocks.

This is probably the case for all of the bounds on `type Id` here, but I didn't want to insist on it

mversic (approved):

mversic (approved):

data_model/src/lib.rs
```diff
@@ -712,16 +712,7 @@ where
...
-        + fmt::Display
-        + fmt::Debug
-        + Clone
-        + Eq
-        + Ord
-        + core::hash::Hash
-        + Decode
-        + Encode
-        + Serialize;
+    type Id;
```
s8sato, line 715: With https://github.com/hyperledger/iroha/pull/2497#discussion_r923344278 I've changed my heart. These bounds on `Id` should not be involved in the `Identifiable` trait declaration

data_model/src/lib.rs
s8sato: https://github.com/hyperledger/iroha/pull/2497#discussion_r923494428

data_model/src/events/data/filters.rs
s8sato: https://github.com/hyperledger/iroha/pull/2497#discussion_r923494428

data_model/src/events/data/filters.rs
```diff
@@ -94,50 +96,81 @@ impl Filter for EntityFilter {
...
-impl<Id: Eq> IdFilter<Id> {
-    /// Construct new `IdFilter`
-    pub fn new(id: Id) -> Self {
-        Self(id)
+#[derive(Clone, PartialOrd, Ord, Eq, Debug, Decode, Encode, Serialize, IntoSchema)]
+/// Filter that accepts a data event with the matching origin.
+pub struct OriginFilter<T: HasOrigin>(<T::Origin as Identifiable>::Id)
+where
+    <T::Origin as Identifiable>::Id:
+        Debug + Clone + Eq + Ord + Hash + Decode + Encode + Serialize + IntoSchema;
```
mversic, line 104: now that you've decided in favor of this, it's likely you don't need all these bounds on a struct definition?

ilchu (approved):

Issue #2499, opened 2022-07-18
[documentation] Incorrect error text in client_cli


6r1d:
There are two incorrect values for error text I'd prefer to fix, copying the account-related error.

This value should be "`Failed to get all assets`":

https://github.com/hyperledger/iroha/blob/5ff6b67d392a0f8238d483918ea36e82d9a22937/client_cli/src/main.rs#L674-L675

And this one should be "`Failed to get all domains`":

https://github.com/hyperledger/iroha/blob/5ff6b67d392a0f8238d483918ea36e82d9a22937/client_cli/src/main.rs#L288-L289

mversic:
@6r1d we can close this issue now?

PR #2500, opened 2022-07-18
[documentation] #2499 corrects some client_cli error messages

6r1d:
### Description of the Change

A tiny correction for the error messages, which seem to be copied previously; the error message ''Failed to get all **accounts**" was used for:

```rust
.request(client::domain::all())
```

and 

```rust
.request(client::asset::all())
```

### Issue

I can open one if needed, but this is nothing more than a typo.

### Benefits

Error messages being displayed will be corrected.

### Possible Drawbacks

None


outoftardis:
`corrects some client_cli error messages` -> `Fix client_cli error messages`

sorry :) it's just that we create changelog from these messages, that's why it's important

6r1d:
> sorry :) it's just that we create changelog from these messages, that's why it's important

All good. I'm just often confused by this format. For example, I thought it's based on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), more or less, so I wrote "docs" instead of the "documentation".

It's also a bit confusing with error messages. I was thinking if this small edit was a "documentation" or a "fix"? For example, what if some small test relied on that string?

PR #2503, opened 2022-07-18
[refactor] #2253: Add `Registrable` trait to `data_model`

### LINKED ISSUE #2253: `data_model` builder trait
In our `data_model` we have a lot of builders (though only for the `Identifiable` structures) offering the `build` method as was noted [here](https://github.com/hyperledger/iroha/pull/2211#discussion_r877841198). Currently, we have an `Identifiable::RegisteredWith` which connects the builder and the structure to build. Most likely we can have both

Erigara:

### Description of the Change

Introduce `Registrable` trait to create unified interface for building objects.

### Issue

Closes #2253.

### Benefits

Unified interface.

### Possible Drawbacks

None.

### DIFF
--- a/client/tests/integration/queries/role.rs
+++ b/client/tests/integration/queries/role.rs
@@ -91,7 +91,7 @@ fn find_role_by_id() -> Result<()> {
 
     let found_role = test_client.request(client::role::by_id(role_id))?;
 
-    assert_eq!(found_role.id(), new_role.id());
+    assert_eq!(found_role.id(), new_role.build().id());
     assert!(found_role.permissions().next().is_none());
 
     Ok(())
--- a/data_model/src/asset.rs
+++ b/data_model/src/asset.rs
@@ -416,11 +416,19 @@ pub struct NewAssetDefinition {
     metadata: Metadata,
 }
 
-impl Identifiable for NewAssetDefinition {
-    type Id = <AssetDefinition as Identifiable>::Id;
+#[cfg(feature = "mutable_api")]
+impl crate::Registrable for NewAssetDefinition {
+    type Target = AssetDefinition;
 
-    fn id(&self) -> &Self::Id {
-        &self.id
+    #[must_use]
+    #[inline]
+    fn build(self) -> Self::Target {
+        Self::Target {
+            id: self.id,
+            value_type: self.value_type,
+            mintable: self.mintable,
+            metadata: self.metadata,
+        }
     }
 }
 
@@ -455,17 +463,10 @@ impl NewAssetDefinition {
         }
     }
 
-    /// Construct [`AssetDefinition`]
+    /// Identification
     #[inline]
-    #[must_use]
-    #[cfg(feature = "mutable_api")]
-    pub fn build(self) -> AssetDefinition {
-        AssetDefinition {
-            id: self.id,
-            value_type: self.value_type,
-            mintable: self.mintable,
-            metadata: self.metadata,
-        }
+    pub fn id(&self) -> &<AssetDefinition as Identifiable>::Id {
+        &self.id
     }
 }
 
--- a/data_model/src/domain.rs
+++ b/data_model/src/domain.rs
@@ -84,6 +84,23 @@ pub struct NewDomain {
     metadata: Metadata,
 }
 
+#[cfg(feature = "mutable_api")]
+impl crate::Registrable for NewDomain {
+    type Target = Domain;
+
+    #[must_use]
+    #[inline]
+    fn build(self) -> Self::Target {
+        Self::Target {
+            id: self.id,
+            accounts: AccountsMap::default(),
+            asset_definitions: AssetDefinitionsMap::default(),
+            metadata: self.metadata,
+            logo: self.logo,
+        }
+    }
+}
+
 impl HasMetadata for NewDomain {
     #[inline]
     fn metadata(&self) -> &crate::metadata::Metadata {
@@ -116,17 +133,9 @@ impl NewDomain {
         }
     }
 
-    /// Construct [`Domain`]
-    #[must_use]
-    #[cfg(feature = "mutable_api")]
-    pub fn build(self) -> Domain {
-        Domain {
-            id: self.id,
-            accounts: AccountsMap::default(),
-            asset_definitions: AssetDefinitionsMap::default(),
-            metadata: self.metadata,
-            logo: self.logo,
-        }
+    /// Identification
+    pub fn id(&self) -> &<Domain as Identifiable>::Id {
+        &self.id
     }
 }
 
@@ -144,14 +153,6 @@ impl NewDomain {
     pub fn with_metadata(mut self, metadata: Metadata) -> Self {
         self.metadata = metadata;
         self
-    }
-}
-
-impl Identifiable for NewDomain {
-    type Id = <Domain as Identifiable>::Id;
-
-    fn id(&self) -> &Self::Id {
-        &self.id
     }
 }
 
--- a/data_model/src/lib.rs
+++ b/data_model/src/lib.rs
@@ -741,6 +741,16 @@ pub trait Registered: Identifiable {
     type With: Into<RegistrableBox>;
 }
 
+/// Trait for proxy objects used for registration.
+#[cfg(feature = "mutable_api")]
+pub trait Registrable {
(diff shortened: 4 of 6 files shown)


data_model/src/asset.rs
```diff
@@ -416,8 +416,24 @@ pub struct NewAssetDefinition {
     metadata: Metadata,
 }
 
+impl Builder for NewAssetDefinition {
+    type Target = AssetDefinition;
+
+    #[cfg(feature = "mutable_api")]
```
mversic, line 422: I think you should place it above the impl block

Erigara: My intention to capture relation between builder and target type. 
To be able to express smt like this:
```rust
impl Identifiable for NewRole {
    type Id =  <<Self as Builder>::Target as Identifiable>::Id;
}
```
If we make `Builder` trait available only under "mutable_api" feature we lose this ability.
I can do it both ways, what do you think?

mversic:
```rs
impl Identifiable for NewRole { 
    type Id = <<Self as Builder>::Target as Identifiable>::Id; 
}
```
is essentially the same thing as
```rs
impl Identifiable for NewRole { 
    type Id = <Role as Identifiable>::Id; 
}
```
after the compiler fills in the types

Erigara: Ok, i will make `Builder` only available under "mutable_api", thank you for the suggestion.

mversic (dismissed):

mversic (dismissed):

data_model/src/lib.rs
```diff
@@ -745,6 +745,16 @@ pub trait Registered: Identifiable {
...
 
+/// Trait for objects used to create target objects
+#[cfg(feature = "mutable_api")]
+pub trait Builder {
+    /// Constructed type
+    type Target;
+
+    /// Construct [`Self::Target`]
+    fn build(self) -> Self::Target;
+}
```
s8sato, line 756: What do you think about this:

```suggestion
pub trait Registrable {
    /// Constructed type
    type Target: Identifiable;

    /// Identification of [`Self::Target`]
    fn target_id(&self) -> <Self::Target as Identifiable>::Id;

    /// Construct [`Self::Target`]
    fn build(self) -> Self::Target;
}
```

data_model/src/domain.rs
```diff
@@ -77,13 +77,30 @@ impl From<GenesisDomain> for Domain {
 #[display(fmt = "[{id}]")]
 pub struct NewDomain {
     /// The identification associated to the domain builder.
-    id: <Domain as Identifiable>::Id,
+    id: <Self as Identifiable>::Id,
```
s8sato, line 80: and this:

```suggestion
    id: <<Self as Registrable>::Target as Identifiable>::Id,
```

s8sato (dismissed):

data_model/src/lib.rs
mversic: it does make sense to combine them into one trait. `Builder` and `Register` always go together, at least at the moment. I don't think we need `fn target_id` though

Erigara: As i find out there some cases where type implement `Registered<With = Self>` (Trigger, Peer, Account).
In this case having single trait `Builder + Registered` seems inconvenient.
However @s8sato suggestion is still applicable.

client/tests/integration/queries/role.rs
```diff
@@ -91,7 +91,7 @@ fn find_role_by_id() -> Result<()> {
 
     let found_role = test_client.request(client::role::by_id(role_id))?;
 
-    assert_eq!(found_role.id(), new_role.id());
+    assert_eq!(found_role.id(), new_role.target_id());
```
mversic, line 94:
```suggestion
    assert_eq!(found_role.id(), new_role.build().id());
```

data_model/src/lib.rs
```diff
@@ -741,6 +741,19 @@ pub trait Registered: Identifiable {
     type With: Into<RegistrableBox>;
 }
 
+/// Trait for proxy objects used for registration.
+pub trait Registrable {
+    /// Constructed type
+    type Target: Identifiable + Registered<With = Self>;
+
+    /// Identification of [`Self::Target`]
+    fn target_id(&self) -> &<Self::Target as Identifiable>::Id;
```
mversic, line 750: this method is redundant. You will never read the `id` of a builder, i.e. `NewXXX` structure. You always know which id you stored into the builder. I gave you a [suggestion](https://github.com/hyperledger/iroha/pull/2503/files#r926764681) in one place that it's used

mversic, line 750: this method is redundant. You will never read the id of a builder, i.e. NewXXX structure. You always know which id you stored into the builder. Here is a [suggestion](https://github.com/hyperledger/iroha/pull/2503/files#r926764681) in one place that it's used outside this crate

data_model/src/lib.rs
```diff
@@ -261,10 +261,10 @@ pub enum IdentifiableBox {
...
     fn id_box(&self) -> IdBox {
         match self {
-            IdentifiableBox::NewDomain(a) => a.id().clone().into(),
-            IdentifiableBox::NewAccount(a) => a.id().clone().into(),
-            IdentifiableBox::NewAssetDefinition(a) => a.id().clone().into(),
-            IdentifiableBox::NewRole(a) => a.id().clone().into(),
+            IdentifiableBox::NewDomain(a) => a.target_id().clone().into(),
+            IdentifiableBox::NewAccount(a) => a.target_id().clone().into(),
+            IdentifiableBox::NewAssetDefinition(a) => a.target_id().clone().into(),
+            IdentifiableBox::NewRole(a) => a.target_id().clone().into(),
```
mversic, line 267: you can access the field directly with if you make it `pub(crate)` which is ok

data_model/src/lib.rs
```diff
@@ -261,10 +261,10 @@ pub enum IdentifiableBox {
 impl IdentifiableBox {
     fn id_box(&self) -> IdBox {
         match self {
-            IdentifiableBox::NewDomain(a) => a.id().clone().into(),
-            IdentifiableBox::NewAccount(a) => a.id().clone().into(),
-            IdentifiableBox::NewAssetDefinition(a) => a.id().clone().into(),
-            IdentifiableBox::NewRole(a) => a.id().clone().into(),
+            IdentifiableBox::NewDomain(a) => a.id.clone().into(),
```
mversic, line 264: this was my initial suggestion, I would have preferred inherent method, eh. :)

mversic (dismissed):

data_model/src/role.rs
```diff
@@ -126,15 +126,17 @@ impl Registered for Role {
     IntoSchema,
 )]
 pub struct NewRole {
-    inner: Role,
+    pub(crate) inner: Role,
 }
 
-impl Identifiable for NewRole {
-    type Id = <Role as Identifiable>::Id;
+impl Registrable for NewRole {
```
mversic, line 132: I think entire trait can be `mutable_api`

mversic: it just increases API surface for the client without any benefit

mversic (approved):

data_model/src/lib.rs
```diff
@@ -741,6 +741,16 @@ pub trait Registered: Identifiable {
     type With: Into<RegistrableBox>;
 }
 
+/// Trait for proxy objects used for registration.
+#[cfg(feature = "mutable_api")]
+pub trait Registrable {
+    /// Constructed type
+    type Target: Identifiable + Registered<With = Self>;
```
mversic, line 748:
```suggestion
    type Target: Registered<With = Self>;
```

Seems that it's implied by `Registered`

mversic: or should we remove the bound from `Registered`?

Erigara: I think `Registered` is unnecessary, so we can remove it.

Erigara: Actually we can omit any trait bounds on `Target`, since they are not used.

mversic: Are you getting rid of `Registered`?

Erigara: Yes, in `Target` trait bounds.

s8sato (approved):

PR #2515, opened 2022-07-21
[fix] #2501: Fixes the order of assets in queries with pagination

### LINKED ISSUE #2501: [FEATURE] Pagination in the order of insertion
### GIT commit hash

LTS

### Minimum working example

```rust
fn main() {
     // Steps to reproduce …
     // Steps to reproduce …
}
```

### Expected behaviour

For example: if we have 10 assets and want to paginate it by 5 assets per page -> I'll get `01234 56789`. Then if we add 1 asset I expect to get (where x is a new asset) `x0123 45678 9`

### Actual behaviour

I got `x0123 56478 9` and next time I was running this test I got `01234 x5678 9`

This behavior makes pagination unusable as far as I can see

### Operating system

macos

### LibC type and version

_No response_

### Current environment

Docker Hub

### Logs in JSON format

<details>
  
  ```json
  ```
</details>

### Who can help?

_No response_

pesterev:

### Description of the Change

Replaces the lexicographical order of assets with a timestamp-based order in `AssetsMap`.

For example:

Before: `1 (1658432537)`, `10 (1658432546)`, `2 (1658432538)`, `3 (1658432539)`, `4 (1658432540)`, `5 (1658432541)`, `6 (1658432542)`, `7 (1658432543)`, `8 (1658432544)`, `9 (1658432545)`.

Now: `1 (1658432537)`, `2 (1658432538)`, `3 (1658432539)`, `4 (1658432540)`, `5 (1658432541)`, `6 (1658432542)`, `7 (1658432543)`, `8 (1658432544)`, `9 (1658432545)`, `10 (1658432546)`.

Where `N` is Id and `T` is an insertion timestamp in `N (T)` format.

### Issue

Resolves #2501 

### Benefits

Order now depends on insertion timestamp.

### Possible Drawbacks

The queries that retrieve an asset will execute a bit slower because $O(\log n)$ (current implementation) and $O (\log n + \log n)$ (twice b-tree lookup) (impl in this PR).


data_model/src/asset.rs
```diff
@@ -585,20 +661,6 @@ impl DefinitionId {
...
-    pub const fn new(
-        definition_id: <AssetDefinition as Identifiable>::Id,
-        account_id: <Account as Identifiable>::Id,
-    ) -> Self {
-        Self {
-            definition_id,
-            account_id,
-        }
-    }
-}
```
mversic, line 600: please don't move this around because you'll end up in conflict with someone else. There is a lot of parallel work on `data_model`

data_model/src/asset.rs
```diff
@@ -20,9 +20,71 @@ use crate::{
...
 };
 
-/// [`AssetsMap`] provides an API to work with collection of key ([`Id`]) - value
+#[derive(Debug, Clone, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+/// Key of [`AssetsMap`]
+pub struct AssetsMapKey {
+    /// [`Id`] of inserted [`Asset`]
+    pub asset_id: <Asset as Identifiable>::Id,
+    /// Insertion timestamp
+    pub timestamp: i64,
```
mversic, line 29: I don't think you don't need a timestamp. Just one `usize` int tracking the insertion order

mversic (commented):
isn't the problem that we have inconsistent ordering, i.e. every query returns assets in a different order? Or is the problem that assets are sorted alphabetically and not chronologically?

data_model/src/asset.rs
mversic: also have you considered [this](https://github.com/bluss/indexmap). Might be an overkill for this purpose only

pesterev: > Preserves insertion order as long as you don't call .remove().

I don't think that it fits our requirements.

pesterev:
Convert this to a draft because the current decision affects memory related optimizations.

mversic:
@pesterev I believe we can close this in favor of #2575

Issue #2052, opened 2022-04-04
Refactor permission tokens API


Arjentix:
Right now we have to identify `PermissionTokens` using some constant string. It makes sense for external API, but inside Iroha it could be strong typed.

If someone has something to add, feel free to leave a comment

QuentinI:
After some consideration and failed experiments I can't come up with a satisfactory design that would allow us to both create permission tokens dynamically and have them as completely independent types, as opposed to specialized wrappers around stringly-typed `PermissionToken`.

If we don't have/expect to have any cases where one would like to create and use new permission tokens via e.g. smartcontracts, we could make them strongly-typed and predefined per iroha build.

In the other case, I would propose bringing permission tokens more in line with other dynamically created entities: make them Identifiable and Registrable and require registering a token before granting it, like is the case with Roles. That would allow to easily query for all existing permission tokens and make them more consistent with everything else.

I'm more in favor of the second approach, but I think it depends mostly on whether we have any use-cases for defining new permission tokens at run-time, if there are none, it would make sense to go with the first one.

PR #2517, opened 2022-07-21
[refactor] #2052: make permission tokens registrable with definition

QuentinI:
Signed-off-by: Artemii Gerasimovich <gerasimovich@soramitsu.co.jp>

### Description of the Change

Make PermissionToken Ids Registrable (world-scoped by analogy to roles) and require registering PT id before granting it.
Also add query to find all registered PermissionToken ids.

### Issue

#2052 

### Benefits

More consistent API, ability to query for all permission tokens.

### Possible Drawbacks

A bit more cumbersome to work with.

### Alternate Designs

As described in #2052, disallow creating PermissionTokens dynamically and parametrize iroha with strongly-typed list of available permission tokens.


macro/derive/src/lib.rs
```diff
@@ -211,3 +211,45 @@ fn impl_from_variant(ast: &syn::DeriveInput) -> TokenStream {
...
+/// assert_eq!(MyEnum::variant_count(), 3)
+/// ```
+///
+/// # Panics
+/// When derive attribute target is not an enum
+//
+// TODO: remove when https://github.com/rust-lang/rust/issues/73662
+// or alternative stabilizes
+#[proc_macro_derive(VariantCount)]
+pub fn variant_count_derive(input: TokenStream) -> TokenStream {
```
appetrosyan, line 237: I recommend using an existing crate: https://crates.io/crates/variant_count

permissions_validators/src/lib.rs
```diff
@@ -57,7 +57,7 @@ macro_rules! declare_token {
                 $param_name:ident ($param_string:literal): $param_typ:ty
              ),* $(,)? // allow trailing comma
         },
-        $string:tt // Token name
+        $string:tt // Token id
```
appetrosyan, line 60: Good!

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -387,7 +385,9 @@ mod tests {
         let mut domain = Domain::new(DomainId::from_str("test").expect("Valid")).build();
         let mut bob_account = Account::new(bob_id.clone(), []).build();
         assert!(bob_account.add_permission(PermissionToken::new(
-            Name::from_str("token").expect("Valid")
+            "token"
+                .parse::<<PermissionToken as Identifiable>::Id>()
```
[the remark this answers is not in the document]
QuentinI: Yes, asked me to specify the type for some reason.

core/src/smartcontracts/isi/world.rs
```diff
@@ -184,6 +216,71 @@ pub mod isi {
...
+            let mut unregister_roles = vec![];
+
+            for entry in wsv.roles().iter() {
+                let (role_id, role) = entry.pair();
+                if role.permissions().any(|token| *token.id() == token_id) {
+                    unregister_roles.push(Unregister::<Role>::new(role_id.clone()))
+                }
+            }
+
+            for isi in unregister_roles {
```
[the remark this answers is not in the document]
QuentinI: That would make more sense, right. You wouldn't want to re-register all roles if you just want to unregister one token.
I would even say also leave the empty roles in place for consistency's sake. Empty role existing doesn't hurt, and it's one less corner-case to worry about when writing some generic permission management gizmo.

data_model/src/permissions.rs
```diff
@@ -11,19 +11,54 @@ use alloc::{
...
+    Hash,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+)]
+pub struct Id {
+    /// Permission token name
+    pub name: Name,
```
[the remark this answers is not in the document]
QuentinI: It's what we will register in domain, so it needs a wrapper to implement `Registered`.
Alternative would be to implement `Registered` with `With = Name` for `PermissionToken`, but I think it doesn't fit `Registered` semantics well, since we aren't registering `PermissionToken` itself.
Another alternative would be something like `PermissionTokenDefinition` analogous to `AssetDefinition`.

client/src/client.rs
```diff
@@ -1192,6 +1192,27 @@ pub mod trigger {
     }
 }
 
+pub mod permissions {
+    //! Module with queries for permission tokens
+    use super::*;
+
+    /// Get query to get all registered [`PermissionToken`]
```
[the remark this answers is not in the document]
QuentinI: I think "Query" may sound like it it does the actual querying part, while it just returns a query. Maybe "Construct a query"?

core/src/smartcontracts/isi/world.rs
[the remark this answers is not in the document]
QuentinI: Also, maybe we should implement a way to "dry-run" ISIs to check your transaction does what you think it does?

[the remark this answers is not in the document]
appetrosyan: The dry run is a feature we might need to implement in a different context. I'd keep the idea in mind for next week.

data_model/src/permissions.rs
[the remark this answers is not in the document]
appetrosyan: > Another alternative would be something like PermissionTokenDefinition analogous to AssetDefinition.

Sounds good to me. This also opens the possibility of registering a custom validator for that permission token.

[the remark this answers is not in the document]
QuentinI: > Sounds good to me

Sure, I could re-implement it that way.

macro/derive/src/lib.rs
QuentinI: Actually `vairant_count` doesn't generate docs for derived constant and I don't think there's a way to suppress a lint in derived code without making it module-level.

configs/peer/genesis.json
```diff
@@ -62,6 +62,162 @@
...
+              "Raw": {
+                "Identifiable": {
+                  "PermissionTokenId": {
+                    "name": "can_transfer_only_fixed_number_of_times_per_period"
+                  }
+                }
+              }
+            }
+          }
+        },
```
mversic, line 220: now we have to explicitly register all the permissions in the genesis instead of having them implicitly defined?

mversic: could these permission registrations be replaced with one role registration? What happens if user registers a role containing permissions that weren't registered previously?

data_model/src/events/data/events.rs
```diff
@@ -170,6 +170,54 @@ mod role {
...
+        type Origin = PermissionTokenId;
+
+        fn origin_id(&self) -> &<Self::Origin as Identifiable>::Id {
+            match self {
+                PermissionTokenEvent::Created(id) | PermissionTokenEvent::Deleted(id) => id,
+            }
+        }
+    }
+
+    impl Identifiable for PermissionTokenEvent {
```
mversic, line 210: I don't think we implement `Identifiable` by hand anymore after #2476 was merged. Also, I think it was agreed that events will implement `HashOrigin` instead of `Identifiable`. `HashOrigin` was specifically introduced to handle events

data_model/src/permissions.rs
```diff
@@ -43,21 +81,29 @@ pub type Permissions = btree_set::BTreeSet<PermissionToken>;
...
 pub struct PermissionToken {
     /// Name of the permission rule given to account.
-    name: Name,
+    id: <Self as Identifiable>::Id,
     /// Params identifying how this rule applies.
     #[getset(skip)]
     params: btree_map::BTreeMap<Name, Value>,
 }
 
+impl Identifiable for PermissionToken {
```
mversic, line 93: this warrants a discussion. `PermissionToken` wasn't `Identifiable` on purpose as far as I remember. Please contact @Arjentix for his work on #2421

mversic: also, never implement `Identifiable` by hand without a specific reason now that we have the macro

data_model/src/permissions.rs
```diff
@@ -11,19 +11,57 @@ use alloc::{
...
+)]
+#[cfg_attr(feature = "ffi", derive(IntoFfi, TryFromFfi))]
+#[cfg_attr(feature = "ffi", ffi_export)]
+#[display(fmt = "{name}")]
+pub struct Id {
+    /// Permission token name
+    pub name: Name,
+}
+
+impl Identifiable for Id {
```
mversic, line 53: I don't see other `Id`s implementing `Identifiable`? This seems redundant

permissions_validators/src/private_blockchain/tests/revoke_and_grant.rs
```diff
@@ -64,7 +64,7 @@ fn add_register_domains_permission_denies_registering_domain_with_wrong_token()
 
     let mut alice = Account::new(alice_id.clone(), []).build();
     alice.add_permission(PermissionToken::new(
-        Name::from_str("incorrecttoken").expect("Valid"),
+        "incorrecttoken".parse().expect("Valid"),
```
mversic, line 67: IMO `PermissionToken` could implement `FromStr` directly in some cases which would reduce the code to:
```rs
alice.add_permission(PermissionToken::from_str("incorrect_token"));
```

data_model/src/permissions.rs
mversic: I think that you can have `PermissionToken` of the same name in two different accounts or two different roles. Maybe you can come up with an appropriate `Id` for `PermissionTokens`, but in any case it can't be identified only with name

mversic: considering that we store permissions in both accounts and roles located in wsv, I can propose something like this for you:
```rs
    struct Id {
        name: Name,
        scope: Either<AccountId, RoleId>
    }
```

data_model/src/permissions.rs
appetrosyan: It's not only redundant, something is identifiable if it **has** an Id. This **is** an Id.

configs/peer/genesis.json
appetrosyan: This in and of itself isn't a problem, but we need to cut down on the noise. Here we have a long sequence of essentially redundant information which is completely unreadable. 

Our RISC + Vector approach would reduce this down to 

```
Register": {
             "object": {
               "Raw": {
                 "Identifiable": {
                   "PermissionTokenId": {
                     "names": [
                                "can_transfer_only_fixed_number_of_times_per_period", 
                                "can_transfer_user_assets"
                   }
                 }
               }
             }
```

This compression would also reduce the on-disk and in memory size of the permission tokens.

permissions_validators/src/private_blockchain/tests/revoke_and_grant.rs
appetrosyan: Or 
```rust
PermissionToken::new("incorrect_token").expect("valid");
```

Rust dislikes hidden allocations, but constructors are allocating by definition.

core/src/smartcontracts/isi/world.rs
```diff
@@ -241,4 +338,15 @@ pub mod query {
...
     }
+
+    impl ValidQuery for FindAllPermissionTokenIds {
+        #[metrics("find_all_token_ids")]
+        fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, Error> {
+            Ok(wsv
+                .permission_token_ids()
+                .iter()
+                .map(|id_ref| id_ref.clone())
+                .collect())
```
Arjentix, line 349:
```suggestion
            Ok(wsv
                .permission_token_ids()
                .iter()
                .cloned()
                .collect())
```

core/src/wsv.rs
```diff
@@ -487,6 +489,12 @@ impl WorldStateView {
         &self.world.roles
     }
 
+    /// Get all permission token ids
+    #[inline]
+    pub fn permission_token_ids(&self) -> &crate::PermissionTokenIds {
+        &self.world.permission_token_ids
+    }
+
```
Arjentix, line 497: I understand, that this looks reasonable, but I just don't like that anyone can modify `PermissionTokenIds` by non-mutable reference (because it's the `DashSet`).

@appetrosyan , @mversic ,  what do you think? Is this okay? Or maybe someday we can refactor `WorldStateView` to remove things like that using iterators or *inspect*-like approach?

core/src/smartcontracts/isi/world.rs
```diff
@@ -114,6 +114,30 @@ pub mod isi {
         }
     }
 
+    impl Execute for Register<PermissionTokenId> {
```
Arjentix, line 117: Hmmm.... By design all permission tokens with this *id* should have the same parameters keys... Maybe it makes sense to pass expected parameters on `Register` step and then check if they exist on `Grant` step?

It's like dynamic typing

data_model/src/permissions.rs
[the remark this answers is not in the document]
Arjentix: Yeah, I also like the `PermissionTokenDefinition` idea. This will allow to specify required parameters

data_model/src/permissions.rs
```diff
@@ -11,19 +11,57 @@ use alloc::{
...
+    type Id = Self;
+
+    fn id(&self) -> &Self::Id {
+        self
+    }
+}
+
+impl Registered for Id {
+    type With = Self;
+}
```
Arjentix, line 63: Also redundant as `impl Identifiable for Id`

data_model/src/permissions.rs
Arjentix: `PermissionTokens` differ not only by their *ids* but also by their *parameter values*. `PermissionTokenDefinition` can be differed only by *id*, but it's not true for `PermissionToken` it-self. And I don't think we need this

macro/derive/src/lib.rs
Arjentix: Anyway I like the idea of `associated constant` instead of `const fn`

permissions_validators/src/private_blockchain/tests/revoke_and_grant.rs
Arjentix: I'm not sure why to introduce a constructor like that. What if we already have `Name`? We can't move it to the `PermissionToken`, we should always deref it and clone

permissions_validators/src/public_blockchain/mod.rs
```diff
@@ -698,4 +720,11 @@ mod tests {
             .check(&bob_id, &set, &wsv)
             .is_deny());
     }
+
+    #[test]
+    fn default_permission_tokens_are_unique() {
+        let mut permissions = default_permission_token_ids();
+        permissions.sort();
+        assert!(permissions.windows(2).all(|wnd| wnd[0] != wnd[1]));
+    }
```
Arjentix, line 729:
```suggestion
    #[test]
    fn default_permission_tokens_are_unique() {
        let permissions_set = HashSet::from(default_permission_token_ids());
        assert_eq!(permissions_set.len(), default_permission_token_ids().len());
    }
```

This should work faster and looks better, IMO.

Your approach has `O(NlogN) + 2*O(N)` complexity and mine has `O(N) * Amortized_O(1)`

permissions_validators/src/private_blockchain/tests/revoke_and_grant.rs
mversic: > I'm not sure why to introduce a constructor like that. What if we already have Name? We can't move it to the PermissionToken, we should always deref it and clone

you are correct, this is the convention we use in `data_model`

data_model/src/permissions.rs
QuentinI: Right, it's an awkward construction. `PermissionTokenDefinition` will fix it.

core/src/smartcontracts/isi/account.rs
```diff
@@ -195,6 +195,10 @@ pub mod isi {
             let account_id = self.destination_id;
             let permission = self.object;
 
+            wsv.permission_token_ids()
+                .get(permission.id())
+                .ok_or_else(|| FindError::PermissionToken(permission.id().clone()))?;
+
```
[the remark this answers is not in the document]
QuentinI: If there's no such token registered, it shouldn't exist in the user account anyway, since they are revoked on unregistration.
It would make sense to log such cases though, since it indicates inconsistent blockchain state.

core/src/smartcontracts/isi/world.rs
QuentinI: It's a [RefMulti](https://docs.rs/dashmap/5.3.4/dashmap/setref/multiple/struct.RefMulti.html), not an actual reference, so `cloned` doesn't work here.

configs/peer/genesis.json
QuentinI: @mversic 

> now we have to explicitly register all the permissions in the genesis instead of having them implicitly defined?

Yes, that's mostly the point of this PR.

> could these permission registrations be replaced with one role registration?

No, because...

> What happens if user registers a role containing permissions that weren't registered previously?

...`FindError`

data_model/src/permissions.rs
QuentinI: Right, that was a blunder.
I don't think I'll need it with `PermissionTokenDefinition` anyway.

mversic: > `PermissionTokens` differ not only by their _ids_ but also by their _parameter values_. `PermissionTokenDefinition` can be differed only by _id_, but it's not true for `PermissionToken` it-self. And I don't think we need this

is this really the case? You're saying that account could have two permissions of the same name with different parameters?

mversic: > `PermissionTokens` differ not only by their _ids_ but also by their _parameter values_. `PermissionTokenDefinition` can be differed only by _id_, but it's not true for `PermissionToken` it-self. And I don't think we need this

yeah, I was wrong there

core/src/smartcontracts/isi/world.rs
QuentinI: It would make sense to add some schema-like thing to PermissionTokenDefinition, not sure if it's in scope for this PR thought.

core/src/wsv.rs
appetrosyan: Unfortunately any concurrent (high-performance) `Set` implementation will have to pass a non-`mut` reference, and so anyone holding it can modify. The only way of preventing others from modifying it, is to expose a subset of its functionality either via a wrapper `struct`, or by completely hiding the `DashSet` and instead using `inspect`. 

I don't think that this is such a huge problem, though. In e.g. Java the point is that you want to be Open-Closed. In rust the same approach doesn't work, because you have to statically link against the other libraries' source code, and we're writing open source software. 

Thus, we shouldn't try to defend against malicious modification (we have the Sumeragi algorithm for that), but instead focus on preventing user error. In this case, modifying the `permission_token_ids` from outside the blockchain is a mistake, which can be fixed by returning a `PermissionTokenIds` wrapper, which can be read as a `DashMap` but not modified. Ideally, for FFI, it might even be good to write it as a `Vec` of tuples.

core/src/smartcontracts/isi/world.rs
appetrosyan: > It's like dynamic typing

It is the lisp variety of dynamic typing, to be more precise. 

> It would make sense to add some schema-like thing to PermissionTokenDefinition, not sure if it's in scope for this PR thought.

Can we generate the aforementioned schema? To me it looks like we shouldn't need to hard-code anything and instead move permission validators which depend on these parameters into run-time dependencies.

permissions_validators/src/public_blockchain/mod.rs
```diff
@@ -64,6 +66,26 @@ impl From<PredefinedPermissionToken> for PermissionToken {
...
+        key_value::CanRemoveKeyValueInUserAssets::definition(),
+        key_value::CanSetKeyValueInUserMetadata::definition(),
+        key_value::CanRemoveKeyValueInUserMetadata::definition(),
+        key_value::CanSetKeyValueInAssetDefinition::definition(),
+        key_value::CanRemoveKeyValueInAssetDefinition::definition(),
+        mint::CanMintUserAssetDefinitions::definition(),
+        transfer::CanTransferUserAssets::definition(),
+        transfer::CanTransferOnlyFixedNumberOfTimesPerPeriod::definition(),
+    ]
+}
```
Arjentix, line 87: Theoretically this can be macro-generated as well. I don't insist just an idea how to reduce the amount of work (and possible errors) when someone will be adding a new variant

core/src/wsv.rs
Arjentix: @appetrosyan , I understand, but I don't mean external users. I mean developers and accidental mistakes. I just want to improve `WSV` design so that modification won't be possible if the method is not designed for it

core/src/smartcontracts/isi/world.rs
Arjentix: @QuentinI , if you want this to be in the separate PR, please, create an issue for it. Seems like we decided that it worths implementation

data_model/src/role.rs
```diff
@@ -77,6 +77,20 @@ impl Role {
...
+    // https://github.com/rust-lang/rust/issues/70530
+    pub fn remove_permissions(&mut self, definition_id: &crate::permissions::Id) {
+        let new_permissions = self
+            .permissions
+            .iter()
+            .filter(|token| token.definition_id() != definition_id)
+            .cloned()
+            .collect();
+
+        self.permissions = new_permissions;
```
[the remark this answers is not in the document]
QuentinI: `drain_filter` isn't stabilized for BTreeSet yet. Didn't think of `retain`, thanks

client/tests/integration/permissions.rs
```diff
@@ -19,6 +19,40 @@ fn get_assets(iroha_client: &mut Client, id: &AccountId) -> Vec<Asset> {
         .expect("Failed to execute request.")
 }
 
+#[test]
+fn permissions_require_registration_before_grant() {
+    let (_rt, _peer, iroha_client) = <PeerBuilder>::new()
+        .with_instruction_judge(public_blockchain::default_permissions())
```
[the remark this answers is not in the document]
QuentinI: I don't think it is, `new()` just forwards to `default()` which has [None](https://github.com/hyperledger/iroha/blob/33ab59a1c6c3f4e85a842df3fbbdfbdcfe38fdf6/core/test_network/src/lib.rs#L728) for instruction_judge

core/src/smartcontracts/isi/world.rs
```diff
@@ -184,6 +226,84 @@ pub mod isi {
             })
         }
     }
+
+    impl Execute for Unregister<PermissionTokenDefinition> {
```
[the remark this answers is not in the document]
QuentinI: Right, that makes sense.

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
Arjentix: Hm... I was confused at a moment. Yes, builder has `None` by default, but then it creates `Peer` it will is `default_permissions()` if user hasn't provided any. Please, look at `start_with_peer()` method and where it is used

core/src/smartcontracts/isi/world.rs
```diff
@@ -184,6 +226,84 @@ pub mod isi {
...
+            for role_id in roles_containing_token {
+                wsv.modify_world(|world| match world.roles.get_mut(&role_id) {
+                    Some(mut role) => {
+                        role.remove_permissions(&definition_id);
+                        Ok(RoleEvent::Modified(role_id.clone()).into())
+                    }
+                    None => {
+                        error!(%role_id, "role not found - this is a bug");
+                        Err(FindError::Role(role_id.clone()).into())
+                    }
```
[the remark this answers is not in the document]
QuentinI: I don't have a strong opinion either way, here it's by analogy with how similar case in `Unregister<Role>` was handled.

permissions_validators/src/public_blockchain/mod.rs
```diff
@@ -64,6 +66,26 @@ impl From<PredefinedPermissionToken> for PermissionToken {
     }
 }
 
+/// Lists ids of all predefined permission tokens, e.g. for easier
```
outoftardis, line 69:
```suggestion
/// List ids of all predefined permission tokens, e.g. for easier
```

permissions_validators/src/private_blockchain/mod.rs
```diff
@@ -7,6 +7,12 @@ pub mod register;
 #[cfg(test)]
 mod tests;
 
+/// Lists ids of all predefined permission tokens, e.g. for easier
```
outoftardis, line 10:
```suggestion
/// List ids of all predefined permission tokens, e.g. for easier
```

data_model/src/permissions.rs
```diff
@@ -15,17 +15,72 @@ use std::{
...
+        id: Id,
+    }
+}
+
+impl Registered for PermissionTokenDefinition {
+    type With = Self;
+}
+
+impl PermissionTokenDefinition {
+    /// Constructs new `PermissionTokenDefinition`
```
outoftardis, line 78:
```suggestion
    /// Construct new `PermissionTokenDefinition`
```

core/src/smartcontracts/isi/world.rs
```diff
@@ -184,6 +226,102 @@ pub mod isi {
...
+                    error!(%role_id, "role not found - this is a bug");
+                    Err(FindError::Role(role_id.clone()).into())
+                }
+            })?;
+        }
+
+        Ok(())
+    }
+
+    /// Removes all tokens with specified definition id from all accounts in all domains
```
outoftardis, line 290:
```suggestion
    /// Remove all tokens with specified definition id from all accounts in all domains
```

core/src/smartcontracts/isi/world.rs
```diff
@@ -184,6 +226,102 @@ pub mod isi {
...
+                    }
+                    None => Err(FindError::PermissionTokenDefinition(definition_id).into()),
+                }
+            })?;
+
+            Ok(())
+        }
+    }
+
+    /// Removes all tokens with specified definition id from all registered roles
```
outoftardis, line 257:
```suggestion
    /// Remove all tokens with specified definition id from all registered roles
```

core/src/smartcontracts/isi/query.rs
```diff
@@ -98,6 +98,7 @@ impl ValidQuery for QueryBox {
             FindTransactionsByAccountId(query) => query.execute_into_value(wsv),
             FindTransactionByHash(query) => query.execute_into_value(wsv),
             FindPermissionTokensByAccountId(query) => query.execute_into_value(wsv),
+            FindAllPermissionTokenDefinitions(query) => query.execute_into_value(wsv),
```
outoftardis, line 101: a comment to myself: add this to the list of queries in the tutorial

core/src/lib.rs
```diff
@@ -29,13 +30,17 @@ pub const TX_RETRIEVAL_INTERVAL: Duration = Duration::from_millis(100);
...
 /// Provides an API to work with collection of key([`DomainId`]) - value([`Domain`]) pairs.
-pub type DomainsMap = dashmap::DashMap<<Domain as Identifiable>::Id, Domain>;
+pub type DomainsMap = DashMap<<Domain as Identifiable>::Id, Domain>;
 
-/// `RolesMap` provides an API to work with collection of key(`PeerId`) - value(`Role`) pairs.
-pub type RolesMap = dashmap::DashMap<<Role as Identifiable>::Id, Role>;
+/// `RolesMap` provides an API to work with collection of key(`RoleId`) - value(`Role`) pairs.
+pub type RolesMap = DashMap<<Role as Identifiable>::Id, Role>;
+
+/// `PermissionTokenDefinitionsMap` provides an API to work with collection of key(`PermissionTokenDefinitionId`) - value(`PermissionTokenDefinition`) pairs.
```
outoftardis, line 41:
```suggestion
/// `PermissionTokenDefinitionsMap` provides an API to work with a collection of key(`PermissionTokenDefinitionId`) - value(`PermissionTokenDefinition`) pairs.
```

core/src/lib.rs
```diff
@@ -29,13 +30,17 @@ pub const TX_RETRIEVAL_INTERVAL: Duration = Duration::from_millis(100);
...
-pub type PeersIds = dashmap::DashSet<<Peer as Identifiable>::Id>;
+pub type PeersIds = DashSet<<Peer as Identifiable>::Id>;
 
 /// Provides an API to work with collection of key([`DomainId`]) - value([`Domain`]) pairs.
-pub type DomainsMap = dashmap::DashMap<<Domain as Identifiable>::Id, Domain>;
+pub type DomainsMap = DashMap<<Domain as Identifiable>::Id, Domain>;
 
-/// `RolesMap` provides an API to work with collection of key(`PeerId`) - value(`Role`) pairs.
-pub type RolesMap = dashmap::DashMap<<Role as Identifiable>::Id, Role>;
+/// `RolesMap` provides an API to work with collection of key(`RoleId`) - value(`Role`) pairs.
```
outoftardis, line 38:
```suggestion
/// `RolesMap` provides an API to work with a collection of key(`RoleId`) - value(`Role`) pairs.
```

core/src/smartcontracts/isi/query.rs
outoftardis: https://github.com/hyperledger/iroha-2-docs/issues/107

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
appetrosyan: I think that being explicit here makes it far more readable.

core/src/smartcontracts/isi/world.rs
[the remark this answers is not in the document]
appetrosyan: We can still panic on this error. What forwarding it thoguth the call stack allows us to do is attach more diagnostics. Personally panics should always be executed at the top level.

permissions_validators/src/public_blockchain/mod.rs
appetrosyan: I strongly think that we should make a proc-macro that generates the information.

appetrosyan: However this is best left as a chore.

core/src/smartcontracts/isi/world.rs
```diff
@@ -184,6 +226,102 @@ pub mod isi {
...
+            {
+                roles_containing_token.push(role_id.clone())
+            }
+        }
+
+        for role_id in roles_containing_token {
+            wsv.modify_world(|world| match world.roles.get_mut(&role_id) {
+                Some(mut role) => {
+                    role.remove_permissions(target_definition_id);
+                    Ok(RoleEvent::Modified(role_id.clone()).into())
```
Arjentix, line 278: I see, that it's used only here. Let's rename it to `PermissionRemoved` and add a permission token inside this event. May be it should look like this:

```rust
pub enum RoleEvent {
    // ...
    PermissionRemoved(PermissionRemoved)
}

pub struct PermissionRemoved {
    role_id: RoleId,
    permission_definition_id: PermissionDefinitionId,
}
```

I'm thinking about this, because now with [parametrised triggers](https://github.com/hyperledger/iroha/pull/2569) it may be useful for smart contract writers to know wat exactly have been changed

data_model/src/events/data/events.rs
```diff
@@ -157,14 +157,54 @@ mod role {
...
             match self {
-                Self::Created(id) | Self::Deleted(id) => id,
+                Self::Created(id) | Self::Deleted(id) | Self::Modified(id) => id,
+            }
+        }
+    }
+}
+
+mod permission_token {
+    //! This module contains `PermissionTokenEvent` and its impls
```
Arjentix, line 175:
```suggestion
    //! This module contains [`PermissionTokenEvent`] and its impls
```

Arjentix (approved):
LGTM

Arjentix:
@appetrosyan has mentioned somewhere, that our tests coverage has decreased. Can you please, create one or more issues for the new functional you have added so that in future, then someone will want to write some tests, it will be easier to find a task for it

appetrosyan (approved):

Arjentix (dismissed):

appetrosyan (dismissed):

appetrosyan (approved):

Arjentix (approved):

Issue #2334, opened 2022-06-08
Display for expression


appetrosyan:
`Expression` variants should all have a `Display` implementation.

Erigara:
@appetrosyan, i believe that this issue is resolved now.

PR #2523, opened 2022-07-22
[refactor] #2468: Remove debug supertrait from permission validators

### LINKED ISSUE #2468: Remove `Debug` as a supertrait.
Ok, it worked for `Judge`.
But it won't work for `IsAllowed` because we want to use `Debug` to construct an error message

_Originally posted by @Arjentix in https://github.com/hyperledger/iroha/pull/2436#discussion_r916695464_

Arjentix:
### Description of the Change

* Removed as much output-like supertraits from validators as possible
* `Debug` was replaced with `Display`
* `Display` implemented for permissions validators
* `Display` was also implemented for instructions, queries and expressions

This gives us much more readable error output when something is wrong with permissions. See examples below.

### Issue

* Closes #2468
* Impls #2334. Why it was already closed I have no idea

### Benefits

* Better permission error messages
* More things can be `Display`ed now
* No more `Debug` supertrait for validators

### Possible Drawbacks

Can't see any

#### Example 1

<details>

**iroha2-dev**:

```
Action not permitted: Validator CheckNested { validator: JudgeAsValidator { judge: NoDenies { validators: [IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetAccess }, IsGrantAllowedAsValidator { is_grant_allowed: GrantRegisteredByMeAccess }, IsGrantAllowedAsValidator { is_grant_allowed: GrantRegisteredByMeAccess }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetAccess }, IsGrantAllowedAsValidator { is_grant_allowed: GrantRegisteredByMeAccess }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetAccessRemove }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetAccessSet }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyMetadataAccessSet }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyMetadataAccessRemove }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetDefinitionSet }, IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetDefinitionRemove }] } } } denied operation Grant(GrantBox { object: EvaluatesTo { expression: Raw(PermissionToken(PermissionToken { name: "can_set_key_value_in_asset_definition", params: {"asset_definition_id": Id(AssetDefinitionId(DefinitionId { name: "rose", domain_id: Id { name: "wonderland" } }))} })), _value_type: PhantomData }, destination_id: EvaluatesTo { expression: Raw(Id(AccountId(Id { name: "alice", domain_id: Id { name: "wonderland" } }))), _value_type: PhantomData } }): Validator IsGrantAllowedAsValidator { is_grant_allowed: GrantMyAssetDefinitionSet } denied operation Grant(GrantBox { object: EvaluatesTo { expression: Raw(PermissionToken(PermissionToken { name: "can_set_key_value_in_asset_definition", params: {"asset_definition_id": Id(AssetDefinitionId(DefinitionId { name: "rose", domain_id: Id { name: "wonderland" } }))} })), _value_type: PhantomData }, destination_id: EvaluatesTo { expression: Raw(Id(AccountId(Id { name: "alice", domain_id: Id { name: "wonderland" } }))), _value_type: PhantomData } }): Cannot grant access for assets registered by another account.
```

**this branch**:

```
Action not permitted: Validator ``Grant validator` with nested checking` denied the operation `GRANT `can_set_key_value_in_asset_definition` TO `alice@wonderland`` by `mouse@wonderland`: Validator `Allow to grant `can_set_key_value_in_asset_definition` permission token if `the signer is the asset creator`` denied the operation: Cannot grant access for assets registered by another account.
```

</details>

#### Example 2

<details>

**iroha2-dev**:

```
Action not permitted: Validator CheckNested { validator: Or { first: AccountSetOnlyForSignerAccount, second: HasTokenAsValidator { has_token: SetGrantedByAccountOwner }, _phantom_operation: PhantomData } } denied operation SetKeyValue(SetKeyValueBox { object_id: EvaluatesTo { expression: Raw(Id(AccountId(Id { name: "mouse", domain_id: Id { name: "wonderland" } }))), _value_type: PhantomData }, key: EvaluatesTo { expression: Raw(Name("key")), _value_type: PhantomData }, value: EvaluatesTo { expression: Raw(String("value")), _value_type: PhantomData } }): Nor first validator AccountSetOnlyForSignerAccount succeed: Cannot set values to the account store of another account., nor second validator HasTokenAsValidator { has_token: SetGrantedByAccountOwner } succeed: Account does not have the needed permission token: PermissionToken { name: "can_set_key_value_in_user_metadata", params: {"account_id": Id(AccountId(Id { name: "mouse", domain_id: Id { name: "wonderland" } }))} }.
```

**this branch**:

```
Action not permitted: Validator ```Allow to set account metadata keys only for signer` or `Allow if signer has the corresponding `can_set_key_value_in_user_metadata` permission token`` with nested checking` denied the operation `SET `key` = `value` IN `mouse@wonderland`` by `alice@wonderland`: Nor first validator `Allow to set account metadata keys only for signer` succeed: Cannot set values to the account store of another account, nor second validator `Allow if signer has the corresponding `can_set_key_value_in_user_metadata` permission token` succeed: Account does not have the needed permission token: PermissionToken { name: "can_set_key_value_in_user_metadata", params: {"account_id": Id(AccountId(Id { name: "mouse", domain_id: Id { name: "wonderland" } }))} }
```

</details>


core/src/smartcontracts/isi/permissions/combinators.rs
```diff
@@ -21,7 +21,8 @@ impl<O: NeedsPermission, F: IsAllowed<Operation = O>> ValidatorApplyOr<O> for F
 /// *Or*-combinator for two validators
 ///
 /// `check` succeeds if either `first` or `second` validator succeeds
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Display)]
+#[display(fmt = "`{}` or `{}`", first, second)]
```
appetrosyan, line 25:
```suggestion
#[display(fmt = "`{}` || `{}`", first, second)]
```

More in line with Java and Rust, less so with Python. This also conveys that the short-circuiting logic is in place.

core/src/smartcontracts/isi/permissions/combinators.rs
```diff
@@ -65,8 +69,8 @@ impl<O: NeedsPermission, F: IsAllowed<Operation = O>, S: IsAllowed<Operation = O
             (&first_verdict, &second_verdict)
         {
             return ValidatorVerdict::Deny(format!(
-                "Nor first validator {:?} succeed: {first_reason}, \
-                 nor second validator {:?} succeed: {second_reason}",
+                "Nor first validator `{}` succeed: {first_reason}, \
+                 nor second validator `{}` succeed: {second_reason}",
```
appetrosyan, line 73:
```suggestion
                "Neither the first validator {} succeeded: {first_reason}, \
                 nor the second validator `{}`: {second_reason}",
```

core/src/smartcontracts/isi/permissions/has_token.rs
```diff
@@ -42,6 +45,16 @@ pub struct HasTokenAsValidator<H: HasToken> {
     has_token: H,
 }
 
+impl<H: HasToken> Display for HasTokenAsValidator<H> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Allow if signer has the corresponding `{}` permission token",
```
appetrosyan, line 52:
```suggestion
            "Allow if signer has the `{}` permission token",
```

core/src/smartcontracts/isi/permissions/has_token.rs
```diff
@@ -78,8 +90,7 @@ impl<H: HasToken> IsAllowed for HasTokenAsValidator<H> {
             ValidatorVerdict::Allow
         } else {
             ValidatorVerdict::Deny(format!(
-                "Account does not have the needed permission token: {:?}.",
-                permission_token
+                "Account does not have the needed permission token: {permission_token:?}",
```
appetrosyan, line 93: Shouldn't this be `Display`?

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -45,6 +45,12 @@ pub trait Judge {
         wsv: &WorldStateView,
     ) -> Result<()>;
 
+    /// Disable operation description in validation error message
+    ///
+    /// Useful when one judge is nested in the other judge and we don't want to
+    /// print operation description twice in the error message
+    fn no_display_operation(&mut self);
```
appetrosyan, line 52:
```suggestion
    /// Disable showing the operation description in the validation error message
    ///
    /// Use this when you have one [`Judge`] nested inside another, as it will prevent 
    /// printing the same error message twice. 
    fn disable_display_of_operation_on_error(&mut self);
```

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -64,11 +70,36 @@ pub trait Judge {
...
+        self.name = Some(name);
+        self
+    }
 }
 
-impl<O: NeedsPermission, J: Judge<Operation = O> + std::fmt::Debug> IsAllowed
-    for JudgeAsValidator<O, J>
-{
+impl<O: NeedsPermission, J: Judge<Operation = O> + Display> Display for JudgeAsValidator<O, J> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
appetrosyan, line 93: inline.

data_model/src/expression.rs
```diff
@@ -668,8 +833,20 @@ impl From<If> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` contains `{}`", collection, element)]
```
appetrosyan, line 849:
```suggestion
#[display(fmt = "{}.contains({})", "collection.parenthesise(Operation::negation)", element)]
```
Parenthesise everything except single objects for collection. Use familiar syntax for Rust developers. Other libraries will present this differently.

data_model/src/expression.rs
```diff
@@ -704,8 +881,21 @@ impl From<Contains> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` contains all `{}`", collection, elements)]
```
appetrosyan, line 897:
```suggestion
#[display(fmt = "{}.contains_all({})", "collection.parenthesise(Operation::negation)", elements)]
```

data_model/src/expression.rs
```diff
@@ -740,8 +930,20 @@ impl From<ContainsAll> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` contains any `{}`", collection, elements)]
```
appetrosyan, line 946: Same as above.

data_model/src/expression.rs
```diff
@@ -775,8 +977,20 @@ impl From<ContainsAny> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` == `{}`", left, right)]
```
appetrosyan, line 993:
```suggestion
#[display(fmt = "{} == {}", left, right)]
```

In general terms on both sides of `equals` are not parenthesised.

data_model/src/expression.rs
```diff
@@ -858,6 +1072,23 @@ pub struct Where {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "`{} where: [", self.expression)?;
+
+        let mut first = true;
+        for (key, value) in &self.values {
+            if !first {
+                write!(f, ", ")?;
+            }
+            first = false;
+            write!(f, "`{}` = `{}`", key, value)?;
```
appetrosyan, line 1085:
```suggestion
            write!(f, "`{}` : `{}`", key, value)?;
```
This isn't assignment or equals. This is association.

data_model/src/query.rs
```diff
@@ -252,16 +256,18 @@ pub mod role {
         PartialOrd,
         Ord,
     )]
+    #[display(fmt = "Find all roles")]
```
appetrosyan, line 259: Consider making this the Upper-camel-case name of struct.

core/src/smartcontracts/isi/permissions/has_token.rs
Arjentix: I'd like to keep `corresponging` word here, because *permission token* name does not fully identify it. *Signer* can have a permission token with the same name but with different parameters and that won't work

core/src/smartcontracts/isi/permissions/has_token.rs
Arjentix: I've choosed to use `Debug` because someone has already decided that `Display` for `PermissionToken` should not expose parameters.

But I can see, that this decision is pretty confusing, so I'll change `impl Display for PermissionToken` and will use `Display` here

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -64,11 +70,36 @@ pub trait Judge {
 #[derive(Debug)]
 pub struct JudgeAsValidator<O: NeedsPermission, J: Judge<Operation = O>> {
     judge: J,
+    name: Option<String>,
+}
+
+impl<O: NeedsPermission, J: Judge<Operation = O>> JudgeAsValidator<O, J> {
+    /// Create new [`JudgeAsValidator`] with given `judge`
+    #[inline]
+    fn new(judge: J) -> Self {
```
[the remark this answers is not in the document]
Arjentix: Unfortunately, no:

```
trait bounds other than `Sized` on const fn parameters are unstable
see issue #93706 <https://github.com/rust-lang/rust/issues/93706> for more information
```

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -64,11 +70,36 @@ pub trait Judge {
...
+    pub fn display_as(mut self, name: String) -> Self {
+        self.name = Some(name);
+        self
+    }
 }
 
-impl<O: NeedsPermission, J: Judge<Operation = O> + std::fmt::Debug> IsAllowed
-    for JudgeAsValidator<O, J>
-{
+impl<O: NeedsPermission, J: Judge<Operation = O> + Display> Display for JudgeAsValidator<O, J> {
```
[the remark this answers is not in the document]
Arjentix: Hmm, interesting, ok

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -222,6 +223,29 @@ impl From<Result<()>> for ValidatorVerdict {
 /// Reason for denying the execution of a particular instruction.
 pub type DenialReason = String;
 
+/// Trait for hard-coded strongly-typed permission tokens.
+pub trait PermissionTokenTrait:
+    Into<PermissionToken> + TryFrom<PermissionToken, Error = PredefinedTokenConversionError>
+{
+    /// Name of the permission token.
+    fn name() -> &'static Name;
```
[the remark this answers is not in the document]
Arjentix: Yes, I can't use asociated constant because we can't use `Name` in `const` context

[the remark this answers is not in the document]
Arjentix: I'll mention this in the comments for `name()` function

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -88,12 +119,39 @@ impl<O: NeedsPermission, J: Judge<Operation = O> + std::fmt::Debug> IsAllowed
...
+        }
+    }
 }
 
-impl<O: NeedsPermission> Judge for AtLeastOneAllow<O> {
+impl<O: NeedsPermission> Display for AtLeastOneAllow<O> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.write_str("At least one allow in: [")?;
+
+        let mut first = true;
```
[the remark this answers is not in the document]
Arjentix: I understand what you suggest. But I think [appetrosyan suggestion](https://github.com/hyperledger/iroha/pull/2523#discussion_r929058448) will be better here

core/src/smartcontracts/wasm.rs
```diff
@@ -6,7 +6,6 @@
 use config::Configuration;
 use eyre::WrapErr;
 use iroha_data_model::{prelude::*, ParseError};
-use iroha_logger::prelude::*;
```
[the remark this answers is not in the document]
Arjentix: Created #2542

permissions_validators/src/public_blockchain/mod.rs
```diff
@@ -80,9 +80,12 @@ pub fn default_permissions() -> InstructionJudgeBoxed {
             .with_validator(key_value::GrantMyAssetDefinitionSet.into_validator())
             .with_validator(key_value::GrantMyAssetDefinitionRemove.into_validator())
             .no_denies()
-            .build();
+            .no_display_operation()
+            .build()
+            .into_validator()
+            .display_as("Grant validator".to_owned());
```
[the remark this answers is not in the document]
Arjentix: I don't like implicit allocations, but in this place I can even do it to be `&'static str`, because this name is given at compile-time and won't be changed

data_model/src/expression.rs
```diff
@@ -211,8 +226,20 @@ impl From<ContextValue> for ExpressionBox {
 /// Evaluates to the multiplication of right and left expressions.
 /// Works only for `Value::U32`
```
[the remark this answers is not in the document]
Arjentix: Created #2543

data_model/src/query.rs
Arjentix: I like how it is now. Because `UpperCamelCase` won't look good with other displays like:

```rust
#[display(fmt = "Find `{}` role", id)]
```

data_model/src/expression.rs
```diff
@@ -211,8 +226,20 @@ impl From<ContextValue> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` * `{}`", left, right)]
```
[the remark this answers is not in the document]
Arjentix: Great idea, I like it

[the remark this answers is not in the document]
Arjentix: I will use [Python operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence)

data_model/src/expression.rs
```diff
@@ -343,8 +406,20 @@ impl From<RaiseTo> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` + `{}`", left, right)]
```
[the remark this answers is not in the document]
Arjentix: If we are going to use [Python operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence) then it's not the lowest priority supported operation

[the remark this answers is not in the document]
Arjentix: From PEP 8 you provided [in this comment](https://github.com/hyperledger/iroha/pull/2523#discussion_r929088850):

> If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies)

Although `Add` is not the lowest priority operation it's the lowest priority *math* operation, so let's keep spaces here

data_model/src/expression.rs
```diff
@@ -409,8 +497,20 @@ impl From<Subtract> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "`{}` > `{}`", left, right)]
```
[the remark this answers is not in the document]
Arjentix: Let's keep using style like that:

```rust
#[display(
    fmt = "{} > {}", // Keep spaces
    "left.parenthesise(Operation::Greater)",
    "right.parenthesise(Operation::Greater)"
)]
```

It has more code, but with that we can easily manupulate the order of operations and they will re-render as we want

data_model/src/expression.rs
```diff
@@ -475,8 +587,20 @@ impl From<Less> for ExpressionBox {
...
+    Eq,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoSchema,
+    PartialOrd,
+    Ord,
 )]
+#[display(fmt = "!`{}`", expression)]
```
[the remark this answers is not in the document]
Arjentix: Yeah. Actually this is true for all operations. We need parenthesis only if operand has more than once operand

data_model/src/expression.rs
Arjentix: I think, it's better to use `MethodCall` operation here instead of `Negation`

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -287,20 +313,48 @@ mod tests {
...
+            static NAME: once_cell::sync::Lazy<Name> =
+                once_cell::sync::Lazy::new(|| "some_token".parse().expect("Valid"));
+            &NAME
+        }
+    }
+
+    impl From<SomeToken> for PermissionToken {
+        fn from(_: SomeToken) -> Self {
+            PermissionToken::new(SomeToken::name().clone())
+        }
```
pesterev, line 329: I think cloning is unnecessary.
```rust
        fn from(token: SomeToken) -> Self {
            PermissionToken::new(token)
        }
```

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -287,20 +313,48 @@ mod tests {
     }
 
     #[derive(Debug, Clone, Serialize)]
-    struct GrantedToken;
+    struct SomeToken;
+
+    impl PermissionTokenTrait for SomeToken {
+        fn name() -> &'static Name {
```
pesterev, line 319: It seems that you can add #[inline] here.

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: `PermissionToken` requires `Name` by value, but `SomeToken::name()` returns a static reference

pesterev: Ah got you

pesterev (approved):

permissions_validators/src/public_blockchain/key_value.rs
```diff
@@ -143,7 +146,8 @@ impl IsGrantAllowed for GrantMyAssetAccessSet {
 }
 
 /// Checks that account can set keys only the for signer account.
-#[derive(Debug, Copy, Clone, Serialize)]
+#[derive(Debug, Display, Copy, Clone, Serialize)]
+#[display(fmt = "Allow to set account metadata keys only for signer")]
```
outoftardis, line 150:
```suggestion
#[display(fmt = "Allow only the signer to set account metadata keys")]
```

permissions_validators/src/public_blockchain/key_value.rs
```diff
@@ -194,37 +200,39 @@ impl HasToken for SetGrantedByAccountOwner {
...
             );
         }
         Allow
     }
 }
 
 /// Checks that account can remove keys for assets only the for signer account.
-#[derive(Debug, Copy, Clone, Serialize)]
+#[derive(Debug, Display, Copy, Clone, Serialize)]
+#[display(fmt = "Allow to remove asset metadata keys only for signer")]
```
outoftardis, line 235:
```suggestion
#[display(fmt = "Allow only the signer to remove asset metadata keys")]
```

permissions_validators/src/public_blockchain/key_value.rs
```diff
@@ -274,38 +284,39 @@ impl HasToken for RemoveGrantedByAssetOwner {
...
             );
         }
         Allow
     }
 }
 
 /// Checks that account can remove keys only the for signer account.
-#[derive(Debug, Copy, Clone, Serialize)]
+#[derive(Debug, Display, Copy, Clone, Serialize)]
+#[display(fmt = "Allow to remove account metadata keys only for the account owner")]
```
outoftardis, line 319:
```suggestion
#[display(fmt = "Allow only the account owner to remove account metadata keys")]
```

permissions_validators/src/public_blockchain/key_value.rs
```diff
@@ -356,76 +369,77 @@ impl HasToken for RemoveGrantedByAccountOwner {
...
-                "The signer does not own the account specified in the permission token.".to_owned(),
+                "The signer does not own the account specified in the permission token".to_owned(),
             );
         }
         Allow
     }
 }
 
-/// Validator that checks Grant instruction so that the access is granted to the assets defintion
+/// Validator that checks Grant instruction so that the access is granted to the assets definition
```
outoftardis, line 402:
```suggestion
/// Validator that checks Grant instruction so that the access to the asset definition is granted
```

permissions_validators/src/public_blockchain/key_value.rs
```diff
@@ -356,76 +369,77 @@ impl HasToken for RemoveGrantedByAccountOwner {
...
-            ok_or_skip!(extract_specialized_token(instruction, wsv));
-
         check_asset_creator_for_asset_definition(&token.asset_definition_id, authority, wsv)
     }
 }
 
 /// Checks that account can set keys for asset definitions only registered by the signer account.
-#[derive(Debug, Copy, Clone, Serialize)]
+#[derive(Debug, Display, Copy, Clone, Serialize)]
+#[display(fmt = "Allow to set asset definition metadata keys only for the asset creator")]
```
outoftardis, line 442:
```suggestion
#[display(fmt = "Allow only the asset creator to set asset definition metadata keys")]
```

permissions_validators/src/public_blockchain/key_value.rs
```diff
@@ -461,7 +475,8 @@ impl IsAllowed for AssetDefinitionSetOnlyForSignerAccount {
 }
 
 /// Checks that account can set keys for asset definitions only registered by the signer account.
-#[derive(Debug, Copy, Clone, Serialize)]
+#[derive(Debug, Display, Copy, Clone, Serialize)]
+#[display(fmt = "Allow to remove asset definition metadata keys only for the asset creator")]
```
outoftardis, line 479:
```suggestion
#[display(fmt = "Allow only the asset creator to remove asset definition metadata keys")]
```

data_model/src/query.rs
```diff
@@ -1340,8 +1421,9 @@ pub mod trigger {
         PartialOrd,
         Ord,
     )]
+    #[display(fmt = "Find trigger under `{}` domain", domain_id)]
     pub struct FindTriggersByDomainId {
-        /// `DomainId` under which triggers should be found.
+        /// [`DomainId`] under which triggers should be found.
```
outoftardis, line 1426:
```suggestion
        /// [`DomainId`] specifies the domain in which to search for triggers.
```

data_model/src/query.rs
```diff
@@ -1416,10 +1501,11 @@ pub mod transaction {
         }
     }
 
-    /// `FindTransactionsByAccountId` Iroha Query finds all transaction included in blockchain
+    /// [`FindTransactionsByAccountId`] Iroha Query finds all transaction included in blockchain
```
outoftardis, line 1504:
```suggestion
    /// [`FindTransactionsByAccountId`] Iroha Query finds all transactions included in a blockchain
```

data_model/src/query.rs
```diff
@@ -1387,9 +1470,11 @@ pub mod transaction {
         transaction::{TransactionQueryResult, TransactionValue},
     };
 
+    /// [`FindAllTransactions`] Iroha Query lists all transactions included in blockchain
```
outoftardis, line 1473:
```suggestion
    /// [`FindAllTransactions`] Iroha Query lists all transactions included in a blockchain
```

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -259,6 +399,25 @@ impl<O: NeedsPermission> Judge for DenyAll<O> {
...
     }
+
+    fn disable_display_of_operation_on_error(&mut self) {
+        // [`DenyAll`] never displays operation, hence no-op.
+    }
+}
+
+/// Create string with operation description and
+/// **leading** space if `display_operation` is `true`.
+/// Returns empty string if `display_operation` is `false`.
```
outoftardis, line 410:
```suggestion
/// Create a string with an operation description and a
/// **leading** space if `display_operation` is `true`.
/// Return empty string if `display_operation` is `false`.
```

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -222,6 +223,85 @@ impl From<Result<()>> for ValidatorVerdict {
...
+///         Err(PredefinedTokenConversionError::Param(&PARAM_NAME))
+///     }
+/// }
+/// ```
+pub trait PermissionTokenTrait:
+    Into<PermissionToken> + TryFrom<PermissionToken, Error = PredefinedTokenConversionError>
+{
+    /// Name of the permission token.
+    ///
+    /// Can't use an associated constant because [`Name`] can't be created at *const* context.
```
outoftardis, line 286:
```suggestion
    /// Cannot use an associated constant because [`Name`] cannot be created at *const* context.
```

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -222,6 +223,85 @@ impl From<Result<()>> for ValidatorVerdict {
...
+    fn name() -> &'static Name;
+}
+
+/// Represents error when converting specialized permission tokens
+/// to generic `[PermissionToken]`
```
outoftardis, line 291:
```suggestion
/// Errors that may appear when converting specialized permission tokens
/// to generic `[PermissionToken]`
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -3,15 +3,18 @@
 use super::{super::Evaluate, *};
 
 /// Checks the [`GrantBox`] instruction.
-pub trait IsGrantAllowed: Debug {
-    /// Checks the [`GrantBox`] instruction.
+pub trait IsGrantAllowed: Display {
+    /// Type of token to check.
+    type Token: PermissionTokenTrait;
+
+    /// Check if authority can grant the permission token.
```
outoftardis, line 10:
```suggestion
    /// Check if the authority can grant the permission token.
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -47,24 +61,29 @@ impl<G: IsGrantAllowed> IsAllowed for IsGrantAllowedAsValidator<G> {
...
 }
 
 /// Checks the [`RevokeBox`] instruction.
-pub trait IsRevokeAllowed: Debug {
-    /// Checks the [`RevokeBox`] instruction.
+pub trait IsRevokeAllowed {
+    /// Type of token to check.
+    type Token: PermissionTokenTrait;
+
+    /// Check if authority can Revoke the permission token.
```
outoftardis, line 79:
```suggestion
    /// Check if the authority can Revoke the permission token.
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -172,3 +203,79 @@ pub fn unpack_if_role_revoke(
...
+            }
+            _ => Err(format!(
+                "Provided `{}` instruction contains unsupported object type",
+                stringify!($isi_type)
+            )),
+        }
+    }};
+}
+
+/// Extracts specialized token from [`GrantBox`]
```
outoftardis, line 247:
```suggestion
/// Extract the specialized token from [`GrantBox`]
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -172,3 +203,79 @@ pub fn unpack_if_role_revoke(
...
+///
+/// # Errors
+/// - Cannot evaluate `instruction`
+/// - `instruction` doesn't evaluate to [`RoleId`] or [`PermissionToken`]
+/// - There is no such role
+/// - Role doesn't contain requested specialized token
+/// - Generic [`PermissionToken`] can't be converted to requested specialized token.
```
outoftardis, line 254:
```suggestion
/// - Generic [`PermissionToken`] cannot be converted to requested specialized token.
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -172,3 +203,79 @@ pub fn unpack_if_role_revoke(
...
+                stringify!($isi_type)
+            )),
+        }
+    }};
+}
+
+/// Extracts specialized token from [`GrantBox`]
+///
+/// # Errors
+/// - Cannot evaluate `instruction`
```
outoftardis, line 250:
```suggestion
/// - `instruction` cannot be evaluated.
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -172,3 +203,79 @@ pub fn unpack_if_role_revoke(
...
+}
+
+/// Extracts specialized token from [`RevokeBox`]
+///
+/// # Errors
+/// - Cannot evaluate `instruction`
+/// - `instruction` doesn't evaluate to [`RoleId`] or [`PermissionToken`]
+/// - There is no such role
+/// - Role doesn't contain requested specialized token
+/// - Generic [`PermissionToken`] can't be converted to requested specialized token.
```
outoftardis, line 272:
```suggestion
/// Extract the specialized token from [`RevokeBox`]
///
/// # Errors
/// - `instruction` cannot be evaluated.
/// - `instruction` doesn't evaluate to [`RoleId`] or [`PermissionToken`].
/// - There is no such role.
/// - Role doesn't contain requested specialized token.
/// - Generic [`PermissionToken`] cannot be converted to requested specialized token.
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -172,3 +203,79 @@ pub fn unpack_if_role_revoke(
...
+    }};
+}
+
+/// Extracts specialized token from [`GrantBox`]
+///
+/// # Errors
+/// - Cannot evaluate `instruction`
+/// - `instruction` doesn't evaluate to [`RoleId`] or [`PermissionToken`]
+/// - There is no such role
+/// - Role doesn't contain requested specialized token
```
outoftardis, line 253:
```suggestion
/// - `instruction` doesn't evaluate to [`RoleId`] or [`PermissionToken`].
/// - There is no such role.
/// - Role doesn't contain requested specialized token.
```

core/src/smartcontracts/isi/permissions/roles.rs
```diff
@@ -34,11 +37,22 @@ pub trait IsGrantAllowed: Debug {
 /// Implements [`IsAllowed`] trait so that
 /// it's possible to use it in [`JudgeBuilder`](super::judge::builder::Builder)
 #[derive(Debug)]
-pub struct IsGrantAllowedAsValidator<G: IsGrantAllowed> {
+pub struct IsGrantAllowedAsValidator<G: IsGrantAllowed + Display> {
     is_grant_allowed: G,
 }
 
-impl<G: IsGrantAllowed> IsAllowed for IsGrantAllowedAsValidator<G> {
+impl<R: IsGrantAllowed + Display> Display for IsGrantAllowedAsValidator<R> {
```
Erigara, line 44: What about deriving `Display`?

data_model/src/expression/operation.rs
```diff
@@ -0,0 +1,81 @@
...
+    Other,
+}
+
+/// Priority of operation.
+///
+/// [`First`](Operation::First) is the highest priority
+/// and [`Eight`](Operation::Eight) is the lowest.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Priority {
+    First = 1,
```
Erigara, line 28: You can explicitly set discriminant only to the first element, check this [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f327be47cc951e08ab9b155ea0ebfac).
We can even omit setting discriminant at all rust automatically start from 0 and increment with every element.

data_model/src/expression/operation.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl Operation {
+    /// Get the priority of the operation.
+    ///
+    /// Ordering is the same as in Python code.
+    /// See [`here`](https://docs.python.org/3/reference/expressions.html#operator-precedence)
+    /// for more details.
+    pub fn priority(self) -> Priority {
+        use Operation::*;
+
```
Erigara, line 47: Maybe add also?
```suggestion
use Priority::*;
```

core/src/smartcontracts/isi/permissions/has_token.rs
appetrosyan: That's a good point. Should we allow permission tokens with the same name?

data_model/src/expression.rs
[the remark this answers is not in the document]
appetrosyan: I'm afraid I have to insist: 
1) This is the lowest priority mathematical operation. 
2) In Python there are the following lower priority operations `==`, `>=` (etc. which we all support), bitwise operations (which we don't have), logical operations (whose whitespace is already handled), `if { } else { } ` which appear differently, and lambdas which are effectively the same as WASM. 

All of those cases would not change the reason why whitespace IMO should be removed.

data_model/src/expression.rs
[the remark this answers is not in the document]
appetrosyan: Good point. Also more uniform.

data_model/src/expression.rs
appetrosyan: Yeah, because we have more than just contains.

data_model/src/expression/tests.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn complex_expr_should_use_parentheses() {
+        let left = Add::new(Value::U32(12), Value::U32(8));
+        let right = Value::U32(10);
+        let div = Divide::new(left, right);
+
+        assert_eq!(div.to_string(), "(12 + 8)/10");
+    }
+
+    #[test]
+    fn raise_to_expr_should_use_stars() {
```
Erigara, line 25: This and next test seems a bit strange to me because we are testing trivial behavior.

Idea not necessary for this PR, we could try to evaluate resulting string and compare result with expected result, e.g.:
```rust
  let left = Add::new(Value::U32(12), Value::U32(8));
  let right = Value::U32(10);
  let div = Divide::new(left, right);

  assert_eq!(eval(div.to_string()), (12 + 8) / 10);
```

data_model/src/query.rs
appetrosyan: How about 
```rust
#[display(fmt = "FindRoleById {{ id: {} }}" id)]
```

In fairness this looks more like `Debug`, so another alternative would be 
```rust 
#[display(fmt = "FindRoleById: `{}`", id)]
```

data_model/src/isi.rs
```diff
@@ -181,22 +181,30 @@ pub struct Pair {
...
-#[display(fmt = "{instructions:?}")] // TODO: map to Display.
+#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
 pub struct SequenceBox {
     /// Sequence of Iroha Special Instructions to execute.
     pub instructions: Vec<Instruction>,
 }
 
+impl core::fmt::Display for SequenceBox {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "SEQUENCE [")?;
```
Erigara, line 192: Would be grate to reuse `format_comma_separated`, can't we place it in common crate for `core` and `data_model`?

core/src/smartcontracts/isi/permissions/has_token.rs
appetrosyan: > I've choosed to use Debug because someone has already decided that Display for PermissionToken should not expose parameters.

that's actually why using `Display` would be a good idea. We might not want to automatically let everyone know what the parameters are. As an option we might enable showing parameters (e.g. using `Debug`) for `#[cfg(debug_assertions)]`.

data_model/src/permissions.rs
```diff
@@ -52,6 +49,23 @@ pub struct PermissionToken {
     params: btree_map::BTreeMap<Name, Value>,
 }
 
+impl core::fmt::Display for PermissionToken {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}: {{", self.name)?;
```
Erigara, line 54: if we deligate writing brackets to outer code than we could do something like this:

```rust
let comma_separated_params = format_comma_separated(self.params.iter().map(|(name, value)| format!("`{}`: `{}`", name, value)))
write!(f, "{}: {{{}}}", self.name, comma_separated_params)
```

core/src/smartcontracts/isi/permissions/has_token.rs
Arjentix: That is also related to #2517. I think duplicating names is okay. It's like asset definition id. We can replace **name** with `PermissionTokenDefinitionId` someday

core/src/smartcontracts/isi/permissions/has_token.rs
Arjentix: So let's keep `Display` showing parameters for now? If we want just to show the permission token name we can write `name()` method and print it

data_model/src/query.rs
Arjentix: Stil unclear what to do with i.e. `FindAccountKeyValueByIdAndKey`, which has 2 fields.
Yes, the first version looks more like `Debug` and I wanted to make an error message more human-redable

permissions_validators/src/public_blockchain/burn.rs
```diff
@@ -213,18 +219,19 @@ impl HasToken for GrantedByAssetOwner {
 
 /// Validator that checks Grant instruction so that the access is granted to the assets
 /// of the signer account.
-#[derive(Debug, Copy, Clone, Serialize)]
+#[derive(Debug, Display, Copy, Clone, Serialize)]
+#[display(fmt = "the signer is the asset owner")]
```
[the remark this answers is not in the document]
Arjentix: Initially it was done by purpose. It's implicit, but all this messages for types implementing `IsGrantAllowed` (or `IsRevokeAllowed`) will be printed with the message from `IsGrantAllowedAsValidator` (or `IsRevokeAllowedAsValidator`). So this will look like this:

```
Allow to grant `CanBurnUserAssets` permission token if `the signer is the asset owner`
```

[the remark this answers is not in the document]
Arjentix: So inside quotes the *variables* are placed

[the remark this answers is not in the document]
Arjentix: Or you think it still will be better to start with capital letter?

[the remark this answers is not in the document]
Arjentix: Also, take a look at examples I've provided in the PR description

core/src/smartcontracts/isi/permissions/roles.rs
Arjentix: As this is the enumeration I'd like to use `;`

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -45,6 +45,12 @@ pub trait Judge {
         wsv: &WorldStateView,
     ) -> Result<()>;
 
+    /// Disable showing the operation description in the validation error message
+    ///
+    /// Use this when you have one [`Judge`] nested inside another, as it will prevent
```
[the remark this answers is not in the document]
Arjentix: It's a trait, so it's public to anyone. However the end-user will probably use the function with the same name of `Builder` type.
I need to keep this function in trait to be able to call this from generic `Builder`

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -81,19 +115,55 @@ impl<O: NeedsPermission, J: Judge<Operation = O> + std::fmt::Debug> IsAllowed
...
 
+fn format_comma_separated<T: Display>(
+    input: impl Iterator<Item = T>,
+    f: &mut core::fmt::Formatter<'_>,
+) -> core::fmt::Result {
+    f.write_str("[")?;
+
+    let mut first = true;
+    for item in input {
+        if !first {
```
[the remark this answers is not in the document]
Arjentix: Good, but you forget to skip first element on the loop

data_model/src/expression/operation.rs
Arjentix: Oh, didn't know rust automatically starts from 0, good

data_model/src/expression/tests.rs
Arjentix: Yeah, it's trivial, but that's what someone else can chagne. If he changes this he will get failing tests, which means, that this display behaviour is important and was done for a reason

Arjentix: About evaluation tests I can't say if it's really worth it. @appetrosyan ?

data_model/src/isi.rs
Arjentix: Oh, I could do that, but that sounds more like for `utils` crate rather than `data_model`. Well, let's keep in `data_model` for now

data_model/src/expression/tests.rs
Erigara: I think it's probably overkill

data_model/src/expression.rs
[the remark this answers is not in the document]
Arjentix: Alright let's do it without spaces

data_model/src/query.rs
appetrosyan: We can revisit this later.

data_model/src/expression/tests.rs
appetrosyan: These tests don't test anything. `Display` is meant for humans to see, and you can achieve the same coverage by adding `iroha_logger::debug` statements.

Arjentix: Okay, I'll rollback the commit which introduced this tests

Erigara (approved):

PR #2535, opened 2022-07-25
[refactor] #2419: Remove explicit `drop`s

### LINKED ISSUE #2419: Remove explicit `drop`s
As discussed in #2307 explicit `drop`s usually code smell and we can express then with proper scope.

Erigara:

### Description of the Change

Remove explicit `drop` call through the codebase. 
There were 2 use cases for `drop`:
- Ignore `Result` and pass `clippy` linting: replaced with `let _result = some_func_call(...)`. IMO ideally we should handle errors not ignoring them;
- Actually free some resource: replaced with proper scope.

### Issue

Closes #2419.

### Benefits

No code smell constructions.

### Possible Drawbacks

None.

### DIFF
--- a/actor/src/broker.rs
+++ b/actor/src/broker.rs
@@ -260,36 +260,38 @@ mod tests {
         let broker = Broker::new();
         Actor1(broker.clone()).start().await;
         Actor1(broker.clone()).start().await;
-        let mut rec = broker.subscribe_with_channel::<Message1>();
 
-        time::sleep(Duration::from_millis(100)).await;
-        assert_eq!(
-            (
-                broker.subscribers::<Message1>(),
-                broker.subscribers::<Stop>()
-            ),
-            (3, 2)
-        );
+        {
+            let mut rec = broker.subscribe_with_channel::<Message1>();
 
-        broker.issue_send(Message1).await;
-        time::sleep(Duration::from_millis(100)).await;
+            time::sleep(Duration::from_millis(100)).await;
+            assert_eq!(
+                (
+                    broker.subscribers::<Message1>(),
+                    broker.subscribers::<Stop>()
+                ),
+                (3, 2)
+            );
 
-        broker.issue_send(Stop).await;
-        time::sleep(Duration::from_millis(100)).await;
+            broker.issue_send(Message1).await;
+            time::sleep(Duration::from_millis(100)).await;
 
-        assert_eq!(
-            (
-                broker.subscribers::<Message1>(),
-                broker.subscribers::<Stop>()
-            ),
-            (1, 0)
-        );
+            broker.issue_send(Stop).await;
+            time::sleep(Duration::from_millis(100)).await;
 
-        tokio::time::timeout(Duration::from_millis(10), rec.recv())
-            .await
-            .unwrap()
-            .unwrap();
-        drop(rec);
+            assert_eq!(
+                (
+                    broker.subscribers::<Message1>(),
+                    broker.subscribers::<Stop>()
+                ),
+                (1, 0)
+            );
+
+            tokio::time::timeout(Duration::from_millis(10), rec.recv())
+                .await
+                .unwrap()
+                .unwrap();
+        }
 
         assert_eq!(
             (
--- a/actor/src/lib.rs
+++ b/actor/src/lib.rs
@@ -491,7 +491,7 @@ where
     M::Result: Send,
 {
     async fn handle(self, sender: oneshot::Sender<M::Result>) {
-        drop(sender.send(self));
+        let _send_event_result = sender.send(self);
     }
 }
 
--- a/client/tests/integration/restart_peer.rs
+++ b/client/tests/integration/restart_peer.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::restriction)]
 
-use std::{str::FromStr, sync::Arc, thread, time::Duration};
+use std::{str::FromStr, sync::Arc};
 
 use eyre::Result;
 use iroha_client::client;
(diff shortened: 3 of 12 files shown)


actor/src/broker.rs
```diff
@@ -260,36 +260,38 @@ mod tests {
...
-            ),
-            (1, 0)
-        );
-
-        tokio::time::timeout(Duration::from_millis(10), rec.recv())
-            .await
-            .unwrap()
-            .unwrap();
-        drop(rec);
+        {
```
appetrosyan, line 264: Fine as is, consider extracting into `fn`.

actor/src/lib.rs
```diff
@@ -491,7 +491,7 @@ where
     M::Result: Send,
 {
     async fn handle(self, sender: oneshot::Sender<M::Result>) {
-        drop(sender.send(self));
+        let _send_event_result = sender.send(self);
```
appetrosyan, line 494: I agree that ignoring a result is not the right approach. We did this because we don't want deadlock the actor, but considering work done by @s8sato and ongoing work by you and @SamHSmith I think we might add an error handling mechanism that doesn't block.

client/tests/integration/restart_peer.rs
```diff
@@ -21,74 +21,68 @@ fn restarted_peer_should_have_the_same_asset_amount() -> Result<()> {
...
-                .any(|asset| asset.id().definition_id == asset_definition_id)
-        })
-        .expect("Valid")
-        .into_iter()
-        .find(|asset| asset.id().definition_id == asset_definition_id)
-        .unwrap();
+    let mut iroha_client = client::Client::test(&peer.api_address, &peer.telemetry_address);
 
-    assert_eq!(AssetValue::Quantity(quantity), *account_asset.value());
+    {
```
appetrosyan, line 31: Excellent use of scoping. Extracting `fn` is also still an option, but less useful in this case.

client/tests/integration/unstable_network.rs
```diff
@@ -58,7 +58,7 @@ fn unstable_network(
     polling_max_attempts: u32,
     polling_period: Duration,
 ) {
-    drop(iroha_logger::install_panic_hook());
+    let _install_hook_result = iroha_logger::install_panic_hook();
```
appetrosyan, line 61:
```suggestion
    if let Err(err) = iroha_logger::install_panic_hook() {
        eprintln!("Installing panic hook failed: {err}");
    }   
```

appetrosyan: Then you're not ignoring the error anymore. Could also add a helpful message on how to avoid it, or what might cause it.

core/src/sumeragi/fault.rs
```diff
@@ -642,7 +642,7 @@ impl<G: GenesisNetworkTrait, F: FaultInjection> SumeragiWithFault<G, F> {
         );
         for event in Vec::<Event>::from(&block) {
             trace!(?event);
-            drop(self.events_sender.send(event));
+            let _send_event_result = self.events_sender.send(event);
```
appetrosyan, line 645: Same as above. Try printing some feedback, using e.g. `iroha_logger::debug!`.

data_model/src/trigger/set.rs
```diff
@@ -411,52 +411,54 @@ impl Set {
...
-                    .map(|entry| apply_f(id, entry.value())),
+            // Cloning and clearing `self.ids_write` so that `handle_` call won't deadlock
+            let matched_ids = {
+                let mut ids_write = self.matched_ids.write().await;
+                let ids_clone = ids_write.clone();
+                ids_write.clear();
+                ids_clone
             };
+            for (event_type, id) in matched_ids {
+                // Ignoring `None` variant cause this means that action was deleted after `handle_`
```
appetrosyan, line 437:
```suggestion
                // Ignoring `None` variant because this means that action was deleted after `handle_*()`
```

data_model/src/trigger/set.rs
```diff
@@ -411,52 +411,54 @@ impl Set {
 
         let succeed_clone = Arc::clone(&succeed);
         let errors_clone = Arc::clone(&errors);
-        let apply_f = move |id: Id, action: &dyn ActionTrait| {
-            if let Repeats::Exactly(atomic) = action.repeats() {
-                if atomic.get() == 0 {
-                    return;
+
+        {
```
appetrosyan, line 415: Consider extracting into a `fn`

p2p/tests/integration/p2p.rs
```diff
@@ -206,7 +206,7 @@ async fn multiple_networks() {
         compact_mode: false,
         ..Configuration::default()
     };
-    drop(iroha_logger::init(&log_config));
+    let _logger_init_result = iroha_logger::init(&log_config);
```
appetrosyan, line 209: consider `eprintln!("{err}")`

p2p/tests/integration/p2p.rs
```diff
@@ -269,7 +269,7 @@ async fn start_network(
         broker: broker.clone(),
         messages,
     };
-    drop(actor.start().await);
+    let _actor_addr = actor.start().await;
```
appetrosyan, line 272: This is an important error message getting lost. Should have some feedback.

telemetry/src/ws.rs
```diff
@@ -439,68 +439,70 @@ mod tests {
...
-            assert!(payload.contains_key("implementation"));
-            assert!(payload.contains_key("version"));
-            assert!(payload.contains_key("config"));
-            assert!(payload.contains_key("authority"));
-            assert!(payload.contains_key("startup_time"));
-            assert!(payload.contains_key("network_id"));
-        }
+            // Move `telemetry_sender` into inner scope
+            let telemetry_sender = telemetry_sender;
+            // The first message is initialization
```
appetrosyan, line 445:
```suggestion
            // The first message is `initialization`
```

telemetry/src/ws.rs
```diff
@@ -439,68 +439,70 @@ mod tests {
...
-            assert_eq!(map.get("id"), Some(&Value::Number(0_i32.into())));
-            assert!(map.contains_key("ts"));
-            assert!(map.contains_key("payload"));
-            let payload = map.get("payload").unwrap().as_object().unwrap();
-            assert_eq!(
-                payload.get("msg"),
-                Some(&Value::String("system.interval".to_owned()))
-            );
-            assert_eq!(payload.get("peers"), Some(&Value::Number(2_i32.into())));
+            // The second message is update
```
appetrosyan, line 481:
```suggestion
            // The second message is `update`
```

p2p/tests/integration/p2p.rs
Erigara: Gave wrong impression in PR description: `actor.start().await` always return `Addr` not `Result`.
So we can safely just write:
```rust
actor.start().await;
```

core/src/sumeragi/fault.rs
Erigara: @QuentinI, you are right about failure reasons, [docs](https://docs.rs/tokio/latest/tokio/sync/broadcast/error/struct.SendError.html) says: 
> A send operation can only fail if there are no active receivers, implying that the message could never be received. The error contains the message being sent as a payload so it can be recovered.

> I'm guessing isn't a problem here

I agree that it's not strictly speaking problem, however i think we should at least log that send operation failed, it could be useful during debugging.

NOTE: it could be a problem because we are losing events

telemetry/src/ws.rs
```diff
@@ -561,35 +565,37 @@ mod tests {
...
+            // Fail sending the first message after reconnect
+            fail_send.store(true, Ordering::Release);
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            assert!(message_receiver.try_next().is_err());
+
+            // The message is sent
+            fail_send.store(false, Ordering::Release);
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            assert!(message_receiver.try_next().is_ok());
+        }
```
mversic, line 598: is this really better than previous solution?

actor/src/lib.rs
Erigara: Ok, leave it as is for now.

client/tests/integration/restart_peer.rs
Erigara: Decide to stick with scope here.

telemetry/src/ws.rs
mversic: if you were to extract it into a function I wouldn't mind

data_model/src/trigger/set.rs
Erigara: Simplified function to avoid using scope or `drop`.

telemetry/src/ws.rs
Erigara: I will extract it into function.

core/src/wsv.rs
```diff
@@ -276,7 +276,9 @@ impl WorldStateView {
 
     /// Send [`Event`]s to known subscribers.
     fn produce_event(&self, event: impl Into<Event>) {
-        let _result = self.events_sender.send(event.into());
+        if let Err(error) = self.events_sender.send(event.into()) {
+            debug!(%error, event = ?error.0, "Failed send event");
+        }
```
mversic, line 281: the pattern where you send event and log an error if it occurs 4 times in this PR and possibly in other places. You think you could extract this functionality into some `produce_event` function or wrap the sender so that it's `send` method logs on fail?

On the other hand, I also find it good as is

actor/src/lib.rs
mversic: I don't understand what is gained with this variable binding?

core/src/wsv.rs
Erigara: Good point

actor/src/lib.rs
Erigara: Don't have strong opinion on this, but i think by this binding we indicate that we don't care about `Result` returned from `sender.send(self)`.
We can't just write because of [unused_must_use](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/unused/static.UNUSED_MUST_USE.html):
```rust
sender.send(self);
```

actor/src/broker.rs
Erigara: Not sure about extracting into `fn`, because it's `test` and used only in one place.

mversic (approved):

appetrosyan (approved):

Issue #2465, opened 2022-07-11
Singlethreaded but unoptimized sumeragi able to perform consensus


SamHSmith:

SamHSmith:
All other tasks in the sumeragi refactor depend on this being completed. It's a big task because it breaks a lot of existing code.

PR #2556, opened 2022-07-28
[fix] #2465: Reimplement sumeragi node as singlethreaded state machine

SamHSmith:

### Description of the Change

Sumeragi now runs on it's own thread and is a relatively simple statemachine. To the eye it might seem more complicated because it is laid out flat. But that is to make explicit everything that is done. The final form of the code is not this. But it is a very malleable intermediate stage. Which is why it's going on the staging branch.

### Issue

#2465

### Benefits

It's actually very possible to debug this implementation as opposed to the previous version.

### Possible Drawbacks
None
The best way to get an understanding for the code is to run
`cargo test integration::unstable_network::unstable_network_4_peers_1_fault -- --nocapture`.
I have left in prints to stdout that should help you follow the execution.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -210,12 +210,6 @@ impl Iroha {
         let mut wsv_mutable =
             WorldStateView::from_configuration(config.wsv, world, events_sender.clone());
 
-        wsv_mutable.init(kura.init()?);
-
-        let wsv = wsv_mutable;
-        let latest_block_hash = wsv.latest_block_hash();
-        let latest_block_height = wsv.height();
-
         let query_validator = Arc::new(query_validator);
 
         let transaction_validator = TransactionValidator::new(
@@ -227,9 +221,15 @@ impl Iroha {
         // Validate every transaction in genesis block
         if let Some(ref genesis) = genesis {
             transaction_validator
-                .validate_every(&***genesis, &wsv)
+                .validate_every(&***genesis, &wsv_mutable)
                 .wrap_err("Transaction validation failed in genesis block")?;
         }
+
+        wsv_mutable.init(kura.init()?);
+
+        let wsv = wsv_mutable;
+        let latest_block_hash = wsv.latest_block_hash();
+        let latest_block_height = wsv.height();
 
         let notify_shutdown = Arc::new(Notify::new());
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -72,6 +72,48 @@ pub struct Iroha {
     pub block_sync: AlwaysAddr<BlockSynchronizer>,
     /// Torii web server
     pub torii: Option<Torii>,
+    sumeragi_relay: AlwaysAddr<FromNetworkBaseRelay>,
+}
+
+use iroha_p2p::network::NetworkBaseRelayOnlinePeers;
+struct FromNetworkBaseRelay {
+    sumeragi: Arc<Sumeragi>,
+    broker: Broker,
+}
+
+#[async_trait::async_trait]
+impl Actor for FromNetworkBaseRelay {
+    async fn on_start(&mut self, ctx: &mut iroha_actor::prelude::Context<Self>) {
+        // to start connections
+        self.broker.subscribe::<NetworkBaseRelayOnlinePeers, _>(ctx);
+        self.broker.subscribe::<iroha_core::NetworkMessage, _>(ctx);
+    }
+}
+
+#[async_trait::async_trait]
+impl Handler<NetworkBaseRelayOnlinePeers> for FromNetworkBaseRelay {
+    type Result = ();
+
+    async fn handle(&mut self, msg: NetworkBaseRelayOnlinePeers) {
+        self.sumeragi.update_online_peers(msg.online_peers);
+    }
+}
+
+#[async_trait::async_trait]
+impl Handler<iroha_core::NetworkMessage> for FromNetworkBaseRelay {
+    type Result = ();
+
+    async fn handle(&mut self, msg: iroha_core::NetworkMessage) -> Self::Result {
+        use iroha_core::NetworkMessage::*;
+
+        match msg {
+            SumeragiMessage(data) => {
+                self.sumeragi.incoming_message(data.into_v1());
+            }
+            BlockSync(data) => self.broker.issue_send(data.into_v1()).await,
+            Health => {}
+        }
+    }
 }
 
 impl Iroha {
@@ -205,10 +247,18 @@ impl Iroha {
                 Arc::clone(&queue),
                 broker.clone(),
                 Arc::clone(&kura),
-                // network_addr.clone(),
+                network_addr.clone(),
             )
             .wrap_err("Failed to initialize Sumeragi.")?,
         );
+
+        let sumeragi_relay = FromNetworkBaseRelay {
+            sumeragi: sumeragi.clone(),
+            broker: broker.clone(),
+        }
+        .start()
+        .await
+        .expect_running();
 
         Sumeragi::initialize_and_start_thread(
             sumeragi.clone(),
@@ -245,6 +295,7 @@ impl Iroha {
             kura,
             block_sync,
             torii,
+            sumeragi_relay,
         })
     }
 
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -99,7 +99,7 @@ pub(crate) async fn handle_instructions(
     sumeragi: Arc<Sumeragi>,
     transaction: VersionedTransaction,
 ) -> Result<Empty> {
-    let wsv = sumeragi.get_clone_of_world_state_view(); // This probably means this function doesn't work
+    let wsv = sumeragi.wsv_clone();
     let transaction: Transaction = transaction.into_v1();
     let transaction = VersionedAcceptedTransaction::from_transaction(
         transaction,
@@ -124,7 +124,7 @@ pub(crate) async fn handle_queries(
     pagination: Pagination,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
-    let wsv = sumeragi.get_clone_of_world_state_view();
+    let wsv = sumeragi.wsv_clone();
     let (valid_request, filter) = request.validate(&wsv, &query_validator)?;
     let original_result = valid_request.execute(&wsv)?;
     let result = filter.filter(original_result);
@@ -172,7 +172,7 @@ async fn handle_pending_transactions(
     sumeragi: Arc<Sumeragi>,
     pagination: Pagination,
 ) -> Result<Scale<VersionedPendingTransactions>> {
-    let wsv = sumeragi.get_clone_of_world_state_view();
+    let wsv = sumeragi.wsv_clone();
     Ok(Scale(
         queue
             .all_transactions(&wsv)
@@ -387,13 +387,10 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) ->
 
 #[cfg(feature = "telemetry")]
 async fn handle_status(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) -> Result<Json> {
-    /*
-    update_metrics(&wsv, network).await?;
-    let status = Status::from(&wsv.metrics);
+    sumeragi.update_metrics(network);
+    let status = Status::from(&sumeragi.wsv_clone().metrics);
+    println!("{:?}", status);
     Ok(reply::json(&status))
-     */
-
-    Ok(warp::reply::json(b"Not implemented"))
 }
 
 /*
--- a/core/src/sumeragi/fault.rs
+++ b/core/src/sumeragi/fault.rs
@@ -156,7 +156,6 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
         self.broadcast_msg(TransactionGossip::new(txs), topology);
     }
 
-
     /// Connects or disconnects peers according to the current network topology.
     pub fn connect_peers(&self, topology: &Topology) {
         trace!("Connecting peers...");
--- a/core/src/sumeragi/fault.rs
+++ b/core/src/sumeragi/fault.rs
@@ -2,6 +2,8 @@
 //! should be reserved for testing, and only [`NoFault`], should be
 //! used in code.
 
+use std::sync::{mpsc, Mutex};
+
 use super::{config::SumeragiConfiguration, *};
 use crate::genesis::GenesisNetwork;
 
@@ -32,6 +34,17 @@ impl FaultInjection for NoFault {
 }
 
 /// `Sumeragi` is the implementation of the consensus. This struct allows also to add fault injection for tests.
+///
+///
+/// sumeragi_state_machine_data is a Mutex instead of a RWLock because
+/// it communicates more clearly the correct use of the lock. The most
+/// frequent action on this lock is the main loop writing to it. This
+/// means that if anyone holds this lock they are blocking the sumeragi
+/// thread. A RWLock will tempt someone to hold a read lock because
+/// they think they are being smart, whilst a Mutex screams *DO NOT
+/// HOLD ME*. That is why the SumeragiStateMachineData is wrapped in
+/// a mutex, it's more self documenting.
+
 pub struct SumeragiWithFault<F>
 where
     F: FaultInjection,
@@ -39,12 +52,8 @@ where
     pub(crate) key_pair: KeyPair,
     /// Address of queue
     pub queue: Arc<Queue>,
-    /// The current topology of the peer to peer network.
-    pub topology: Topology,
     /// The peer id of myself.
     pub peer_id: PeerId,
-    /// The block in discussion this round, received from a leader.
-    pub(crate) voting_block: Option<VotingBlock>,
     /// This field is used to count votes when the peer is a proxy tail role.
     pub(crate) votes_for_blocks: BTreeMap<HashOf<VersionedValidBlock>, VersionedValidBlock>,
     pub(crate) events_sender: EventsSender,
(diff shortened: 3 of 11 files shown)


cli/src/lib.rs
```diff
@@ -245,6 +295,7 @@ impl Iroha {
             kura,
             block_sync,
             torii,
+            sumeragi_relay,
         })
     }
```
s8sato, line 301: I'm commenting in the code so that this conversation is threaded, resolved, and collapsed.

- You consider `iroha2-staging` as a staging branch for `iroha2-dev`, right?
- You could enable CI tests by, for example, renaming `iroha2-dev` in `I2::Dev::Tests` workflow
- `iroha2-staging` is outdated relative to `iroha2-dev`. When do you plan to update?

appetrosyan: 1. Tests should be enabled in CI, because we want more than one person to contribute to this branch. 
2. We want to merge at least once every two weeks from `iroha2-dev`. Fortunately nothing concerns the new functionality, but you need to be aware which PRs might, so occasionally have a look.

p2p/src/network.rs
```diff
@@ -198,13 +198,17 @@ where
     type Result = ();
 
     async fn handle(&mut self, msg: ConnectPeer) {
+        if self.peers.contains_key(&msg.peer.public_key) {
```
appetrosyan, line 201: Style point, I usually prefer having a matching `else`.

core/src/sumeragi/network_topology.rs
```diff
@@ -228,13 +228,16 @@ impl Topology {
     /// Apply new committed block hash.
     #[allow(clippy::expect_used)]
     pub fn apply_block(&mut self, block: HashOf<VersionedCommittedBlock>) {
+        /*
+        nocheckin
```
appetrosyan, line 232: So is this not supposed to be checked into `stating` or `dev`?

p2p/src/network.rs
```diff
@@ -317,6 +326,15 @@ where
                 self.broker.issue_send(*msg).await;
             }
         };
+        let mut online_peer_keys = Vec::new();
```
appetrosyan, line 329: Good!

p2p/src/network.rs
```diff
@@ -317,6 +326,15 @@ where
...
+        for peer_key in self.peers.keys() {
```
appetrosyan, line 330:
```suggestion
        self.peers.keys().iter().cloned().collect();
```
The `iter()` is sometimes redundant, but I like to keep it.

cli/src/torii/routing.rs
```diff
@@ -99,7 +99,7 @@ pub(crate) async fn handle_instructions(
     sumeragi: Arc<Sumeragi>,
     transaction: VersionedTransaction,
 ) -> Result<Empty> {
-    let wsv = sumeragi.get_clone_of_world_state_view(); // This probably means this function doesn't work
+    let wsv = sumeragi.get_clone_of_world_state_view();
```
appetrosyan, line 102: Does this definitely mean that the function definitely works?

cli/src/torii/routing.rs
```diff
@@ -387,13 +387,10 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) ->
 
 #[cfg(feature = "telemetry")]
 async fn handle_status(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) -> Result<Json> {
-    /*
-    update_metrics(&wsv, network).await?;
-    let status = Status::from(&wsv.metrics);
+    sumeragi.update_metrics(network);
+    let status = Status::from(&sumeragi.get_clone_of_world_state_view().metrics);
```
appetrosyan, line 391: Naming convention

```suggestion
    let status = Status::from(&sumeragi.wsv_clone().metrics);
```

Or even 


```suggestion
    let status = Status::from(&sumeragi.wsv().clone().metrics);
```

And even better, following the example of our FFI function, 
```suggestion
let status = sumeragi.update_metrics(network).clone().into()
```
i.e.  `update_metrics` returns `&Metrics`, which you can clone.

cli/src/torii/routing.rs
```diff
@@ -387,13 +387,10 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) ->
...
+    println!("{:?}", status);
```
appetrosyan, line 392:
```suggestion
    println!("{status:?}");
```

core/src/sumeragi/mod.rs
```diff
@@ -156,6 +213,22 @@ impl Sumeragi {
             )
         });
     }
+
+    pub fn update_online_peers(&self, online_peers: Vec<PublicKey>) {
+        *self
+            .internal
+            .current_online_peers_by_public_key
+            .lock()
+            .unwrap() = online_peers;
```
appetrosyan, line 222:
```suggestion
            .expect("Lock on update online peers") = online_peers;
```

core/src/sumeragi/mod.rs
```diff
@@ -93,25 +104,71 @@ impl Sumeragi {
...
         self.internal.wsv.lock().unwrap().latest_block_hash()
     }
 
     pub fn get_network_topology(&self, header: &BlockHeader) -> Topology {
-        self.internal.get_network_topology(header)
+        // TODO: make use of block header
+        self.internal
+            .atomic_sumeragi_state
+            .lock()
+            .expect("take lock")
```
appetrosyan, line 169:
```suggestion
            .expect("Get network topology lock")
```

core/src/sumeragi/mod.rs
```diff
@@ -93,25 +104,71 @@ impl Sumeragi {
...
     }
 
+    pub fn update_metrics(&self, network: Addr<IrohaNetwork>) -> Result<()> {
+        use eyre::WrapErr;
+        use thiserror::Error;
+        let online_peers = self
+            .internal
+            .current_online_peers_by_public_key
+            .lock()
+            .unwrap();
```
appetrosyan, line 131: Consider returning as a recoverable error.

core/src/sumeragi/mod.rs
```diff
@@ -93,25 +104,71 @@ impl Sumeragi {
...
                 broker,
                 kura,
-                // network,
+                network,
                 actor_channel_capacity: configuration.actor_channel_capacity,
                 fault_injection: PhantomData,
                 gossip_batch_size: configuration.gossip_batch_size,
                 gossip_period: Duration::from_millis(configuration.gossip_period_ms),
+
+                atomic_sumeragi_state: Mutex::new(atomic_sumeragi_state),
```
appetrosyan, line 116: From experience, whenever a Rust programmer sees a `Mutex`, they are itching to optimise it to an `RwLock`. Given they see `atomic` they might get rid of the `Mutex`. 

If it's a bad idea, you should explain it in a comment.

core/src/sumeragi/mod.rs
```diff
@@ -93,25 +104,71 @@ impl Sumeragi {
...
+                current_online_peers_by_public_key: Mutex::new(Vec::new()),
+                incomming_message_sender: Mutex::new(incomming_message_sender),
+                incomming_message_receiver: Mutex::new(incomming_message_receiver),
```
appetrosyan, line 119:
```suggestion
                incoming_message_sender: Mutex::new(incoming_message_sender),
                incoming_message_receiver: Mutex::new(incoming_message_receiver),
```

appetrosyan (changes requested):

core/src/sumeragi/mod.rs
```diff
@@ -66,19 +70,26 @@ impl Sumeragi {
...
             .build()?;
 
+        let atomic_sumeragi_state = AtomicSumeragiState {
+            genesis_network,
+            latest_block_hash: Hash::zeroed().typed(),
+            latest_block_height: 0,
+            current_topology: network_topology,
+        };
+
+        let (incomming_message_sender, incomming_message_receiver) = std::sync::mpsc::channel();
```
appetrosyan, line 87:
```suggestion
        let (incoming_message_sender, incoming_message_receiver) = std::sync::mpsc::channel();
```

core/src/sumeragi/fault.rs
```diff
@@ -68,194 +66,55 @@ where
...
-            broker,
-            kura,
-            // network,
-            actor_channel_capacity: configuration.actor_channel_capacity,
-            fault_injection: PhantomData,
-            gossip_batch_size: configuration.gossip_batch_size,
-            gossip_period: Duration::from_millis(configuration.gossip_period_ms),
-        })
-    }
+pub struct AtomicSumeragiState {
```
appetrosyan, line 88: It has the word `Atomic` in it. But nothing so far suggests to me that it is `Atomic`. Moreover, a struct made up of atomic units isn't itself atomic. I know that this is WIP, but we need 0day good naming conventions.

core/src/sumeragi/fault.rs
```diff
@@ -68,194 +66,55 @@ where
...
     /// Buffer capacity of actor's MPSC channel
     pub actor_channel_capacity: u32,
     pub(crate) fault_injection: PhantomData<F>,
     pub(crate) gossip_batch_size: u32,
     pub(crate) gossip_period: Duration,
+
+    pub atomic_sumeragi_state: Mutex<AtomicSumeragiState>,
+    pub current_online_peers_by_public_key: Mutex<Vec<PublicKey>>,
+
+    pub incomming_message_sender: Mutex<std::sync::mpsc::Sender<Message>>,
```
appetrosyan, line 84:
```rust 
use std::sync::mpsc
```
```suggestion
    pub incomming_message_sender: Mutex<mpsc::Sender<Message>>,
```

core/src/sumeragi/fault.rs
```diff
@@ -68,194 +66,55 @@ where
...
+}
 
-    /// Initializes sumeragi with the `latest_block_hash` and `block_height` after Kura loads the blocks.
-    pub fn init(&mut self, latest_block: HashOf<VersionedCommittedBlock>, block_height: u64) {
-        self.block_height = block_height;
-        self.topology.apply_block(latest_block);
+impl<F: FaultInjection> SumeragiWithFault<F> {
+    pub fn get_online_peer_keys(&self) -> Vec<PublicKey> {
+        self.current_online_peers_by_public_key
+            .lock()
```
appetrosyan, line 98: Why do we need this, and not just an `RwLock`, if all we're going to do is `clone` the `Vec`?

appetrosyan (changes requested):

core/src/sumeragi/network_topology.rs
SamHSmith: Not to dev. Originally it was a note to me to remember to reimplement view changes. But instead I decided that this can live on the staging branch and there is a issue tracking it. https://github.com/hyperledger/iroha/issues/2561

cli/src/torii/routing.rs
SamHSmith: Yes. Because the function is only reading and the broadcast queue's and arc's get cloned properly as well. If this function didn't work then we wouldn't have 90+ % pass rate on the integration tests.

core/src/sumeragi/mod.rs
SamHSmith: good idea

core/src/sumeragi/fault.rs
SamHSmith: Because holding locks is more complicated then not doing so. `current_online_peers_by_public_key` is not a big array so this is not a problem. Cloning all the time causes performance problems. Sharing data all the time also causes performance problems and *slows down development* because there is more to think about. My guiding principle is to avoid holding locks when I don't have to. Makes things easier.

core/src/sumeragi/message.rs
```diff
@@ -192,6 +195,8 @@ impl BlockCreated {
     /// Fails if there is an issue with block signing.
 
     pub fn handle<F: FaultInjection>(&self, sumeragi: &mut SumeragiWithFault<F>) -> Result<()> {
+        println!("Received Block");
```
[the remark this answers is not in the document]
SamHSmith: I forgot about these since nothing in message.rs is ever called. Will remove them.

core/src/sumeragi/message.rs
```diff
@@ -110,6 +110,7 @@ impl Message {
     #[log(skip(self, sumeragi))]
 
     pub fn handle<F: FaultInjection>(self, sumeragi: &mut SumeragiWithFault<F>) -> Result<()> {
+        println!("Hello there you");
```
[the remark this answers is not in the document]
SamHSmith: I'm using println because logging output is disable for some reason in the test `integration::unstable_network::unstable_network_4_peers_1_fault`. Of course they will be removed or replaced by logging before we merge into iroha2-dev. But for now they stay.

core/src/sumeragi/mod.rs
SamHSmith: yes

core/src/sumeragi/mod.rs
SamHSmith: Turns out, actually not a good idea. Mutex poison errors can't be passed between threads safely.

core/src/sumeragi/mod.rs
SamHSmith: thx

p2p/src/network.rs
SamHSmith: I like it like this when it is a pure early return. As opposed to a different code path.

p2p/src/network.rs
SamHSmith: yes

cli/src/torii/routing.rs
SamHSmith: using suggestion 1

cli/src/torii/routing.rs
SamHSmith: that seems redundant

core/src/sumeragi/fault.rs
```diff
@@ -630,21 +280,1363 @@ pub fn run_sumeragi_main_loop<F>(
...
+                                    }
+
+                                    let valid_signatures = network_topology
+                                        .filter_signatures_by_roles(
+                                            &[Role::ValidatingPeer, Role::Leader],
+                                            block.verified_signatures(),
+                                        );
+
+                                    let mut number_of_votes = 0;
+                                    for i in 0..validating_peers.len() {
```
Erigara, line 650: `zip` could be used here.

core/src/sumeragi/fault.rs
```diff
@@ -630,21 +280,1363 @@ pub fn run_sumeragi_main_loop<F>(
...
+                                        required_number_of_votes = network_topology.min_votes_for_commit() - 1,
+                                        "Received a vote for block",
+                                    );
+
+                                    if number_of_votes < network_topology.min_votes_for_commit() - 1
+                                    {
+                                        continue;
+                                    }
+
+                                    let mut signatures = Vec::new();
```
Erigara, line 674: I think `filter_map` would be convinient here.

Erigara (commented):
I found current state good starting point for further improvements.

core/src/sumeragi/fault.rs
```diff
@@ -630,21 +280,1363 @@ pub fn run_sumeragi_main_loop<F>(
...
+                                sumeragi.broadcast_msg(
+                                    BlockCommitted::from(signed_block.clone()),
+                                    &atomic_sumeragi_guard.current_topology,
+                                );
+
+                                let block = signed_block.commit();
+                                let block_hash = block.hash();
+
+                                if let Err(error) = wsv_guard.apply(block.clone()) {
+                                    panic!("Failed to apply block on WSV. This is absolutely not acceptable.");
```
outoftardis, line 379: I'd remove the `absolutely not acceptable` remark, or at least rephrase it somehow

core/src/sumeragi/network_topology.rs
SamHSmith: yes

core/src/sumeragi/fault.rs
SamHSmith: sure

core/src/sumeragi/fault.rs
SamHSmith: I've changed it to SumeragiStateMachineData. Good of you two for pushing me on this.

core/src/sumeragi/fault.rs
SamHSmith: Yes. And allowing concurrent reading access makes programmers think it's okay to hold the lock. Because it avoids a clone and "that's good right?". In this case it is very important that the data is able to be updated. I'd rather avoid the bug of "sometimes the online peers won't update" that got introduced because someone thought they were being smart. You guys being so resistant to the mutex proves why it is good. It communicates very strongly. DO NOT HOLD THIS LOCK. DOING SO WILL BE BAD. A RWLock communicates a more seducive "Look at how smart I'm being by allowing multiple readers on the same data."

A codebase where one is very hesitant to hold locks leads to a more healthy view on multithreading. It's fast when threads don't communicate. When they have independent work to do. And the only reason for communicating between threads is to arrange for more work to be done in solitude.

cli/src/lib.rs
```diff
@@ -72,6 +72,48 @@ pub struct Iroha {
     pub block_sync: AlwaysAddr<BlockSynchronizer>,
     /// Torii web server
     pub torii: Option<Torii>,
+    sumeragi_relay: AlwaysAddr<FromNetworkBaseRelay>,
+}
+
+use iroha_p2p::network::NetworkBaseRelayOnlinePeers;
```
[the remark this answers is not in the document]
SamHSmith: Rust supports imports inside functions. The reason being that you want to keep your imports as close to the place they are used as possible, if they are only used by that one piece of code.

There has to be a relay, otherwise there's a cyclical dependency between p2p and sumeragi. My instinct was to merge the two modules but since that would have been very destructive I chose this method instead.

core/src/sumeragi/mod.rs
```diff
@@ -93,25 +104,71 @@ impl Sumeragi {
...
+                incomming_message_sender: Mutex::new(incomming_message_sender),
+                incomming_message_receiver: Mutex::new(incomming_message_receiver),
             },
         })
     }
 
+    pub fn update_metrics(&self, network: Addr<IrohaNetwork>) -> Result<()> {
+        use eyre::WrapErr;
+        use thiserror::Error;
+        let online_peers = self
```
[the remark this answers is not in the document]
SamHSmith: sure

core/src/sumeragi/fault.rs
```diff
@@ -266,360 +125,151 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
-                    | view_change::Reason::BlockCreationTimeout(_) => None,
-                };
-                self.change_view(merged_proof.clone(), invalidated_block_hash);
-            }
-        }
-    }
-
-    fn broadcast_msg(&self, msg: impl Into<Message> + Send) {
+    fn broadcast_msg(&self, msg: impl Into<Message> + Send, topology: &Topology) {
         let msg = VersionedMessage::from(msg.into());
```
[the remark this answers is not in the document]
SamHSmith: The inlining seems premature but otherwise yes.

core/src/sumeragi/fault.rs
```diff
@@ -266,360 +125,151 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+                prev_view_changes_count = topology.view_change_proofs().len(),
                 new_view_changes_count = suggested_chain.len(),
-                latest_block = ?self.latest_block_hash(),
+                latest_block = ?latest_block_hash,
                 "Swapping view change proof chain."
             );
-            self.topology = self
-                .topology
+            topology
                 .clone()
```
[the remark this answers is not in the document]
SamHSmith: It isn't.

core/src/sumeragi/fault.rs
```diff
@@ -630,21 +280,1363 @@ pub fn run_sumeragi_main_loop<F>(
...
+                            "Created a block",
+                        );
+                        for event in Vec::<Event>::from(&block) {
+                            trace!(?event);
+                            drop(sumeragi.events_sender.send(event));
+                        }
+                        let signed_block = block
+                            .sign(sumeragi.key_pair.clone())
+                            .expect("Sign genesis block.");
+                        {
```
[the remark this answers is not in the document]
SamHSmith: We don't *need* all these scopes. But it was constructed this way because the scopes loosely reflect what would previously been function calls. These should be refined over time. But they communicate things about what is connected to what.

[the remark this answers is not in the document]
Erigara: Ok, thanks you, for clarification!

cli/src/lib.rs
s8sato, line 301: Failed tests as of 0390bd29 compared to 03ef82b1:

- [ ] `integration::asset_propagation::client_add_asset_quantity_to_existing_asset_should_increase_asset_amount_on_another_peer`
- [ ] `integration::connected_peers::connected_peers_with_f_1_0_1`
- [ ] `integration::connected_peers::connected_peers_with_f_2_1_2`
- [ ] `integration::events::pipeline::transaction_with_fail_instruction_should_be_rejected`
- [ ] `integration::events::pipeline::transaction_with_no_instructions_should_be_committed`
- [ ] `integration::multisignature_transaction::multisignature_transactions_should_wait_for_all_signatures`
- [ ] `integration::unregister_peer::unstable_network_stable_after_add_and_after_remove_peer`

If it's intentional, I'd like an explanation

s8sato (commented):
Let me take a closer look at the main loop

cli/src/lib.rs
SamHSmith: In order to allow multiple people to work on the reimplementation, this PR has things that are broken. Then we will fix these with more PR's into iroha2-staging. We only merge into iroha2-dev when the reimplementation is complete and everything works. Aleksandr for instance is working on #2561 which fixes most of the tests you mention. And if some are still failing after that we will fix them.

core/src/sumeragi/mod.rs
SamHSmith: I will do so.

core/src/sumeragi/fault.rs
```diff
@@ -266,360 +125,151 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
-        {
-            info!(%peer_to_be_connected.address, "Connecting peer");
+        for peer_to_be_connected in peers_expected.iter() {
             self.broker.issue_send_sync(&ConnectPeer {
-                address: peer_to_be_connected.address.clone(),
+                peer: peer_to_be_connected.clone(),
             })
         }
+
+        /* TODO FIX
```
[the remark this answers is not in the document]
SamHSmith: There doesn't seem to be an issue currently. It was implemented though in the previous implementation. This is a, "we'll see if we end up needing this", sort of situation.

core/src/sumeragi/fault.rs
SamHSmith: yea

core/src/sumeragi/fault.rs
SamHSmith: It was indeed useful.

appetrosyan (approved):

s8sato (approved):

Issue #2231, opened 2022-05-19
FFI wrapper API


mversic:
Create wrapper over FFI function calls so that the smart contract API reflects the client API

mversic:
wrapper API for opaque types will be finished in a separate ticket

PR #2557, opened 2022-07-28
[feature] #2536: Introduce `ffi_import` for FFI clients

### LINKED ISSUE #2536: FFI import flag for clients
Introduce FFI `import` flag to differentiate between libraries exposing FFI bindings vs libraries linking dynamically against previous libraries that exposed the bindings. This means that `iroha_core` will not set the `ffi_import` flag while `iroha_wasm` smartcontract will. For `iroha_wasm` this will result in `data_model` replaced by it's FFI equivalent.

For example, when `ffi_import` is enabled, this struct:
```rs
ffi! {
    pub struct NewDomain {
        /// The identification associated with the domain builder.
        id: <Domain as Identifiable>::Id,
        /// The (IPFS) link to the logo of this domain.
        logo: Option<IpfsPath>,
        /// Metadata associated with the domain builder.
        metadata: Metadata,
    }
}

#[ffi_import]
impl NewDomain {
    #[must_use]
    pub fn with_logo(mut self, logo: IpfsPath) -> Self {
        self.logo = Some(logo);
        self
    }
}
```

will be replaced with something akin to the following:
```rs
    pub struct NewDomain {
        ptr: [u8; 0],
    }

    impl NewDomain {
        #[must_use]
        pub fn with_logo(mut self, logo: IpfsPath) -> Self {
            let handle = self.into_ffi();
            let logo = logo.into_ffi();
            let _out_ptr = MaybeUninit::uninit();
            let output = NewDomain__with_logo(handle, logo, _out_ptr);
            _out_ptr.try_from_repr_c(&mut ())
        }
    }
```

mversic:

### Description of the Change

* introduce mutually exclusive `ffi_import` and `ffi_export` features for `iroha_data_model` and `iroha_crypto`
* introduce additional `ffi!` macro that convert the wrapped struct into opaque pointer
* rename `TryFromFfi` -> `TryFromReprC`
* make `ffi` crate `no_std`
* refactor of FFI crate - reduce duplicated code
* closes #2536 

# NOTE
To make the change manageable, this PR only introduces the `ffi_import` flag. The flag does not expand to completely valid code yet and is commented out. This is expected to be completed with #2231 . I'm trying to get this merged so that @Erigara and I don't step on each other's toes anymore


crypto/src/lib.rs
```diff
@@ -164,15 +164,16 @@ impl KeyGenConfiguration {
...
-/// Pair of Public and Private keys.
-#[cfg_attr(feature = "ffi", derive(IntoFfi, TryFromFfi))]
-#[derive(Debug, Clone, PartialEq, Eq, Getters, Serialize)]
-#[getset(get = "pub")]
-pub struct KeyPair {
-    /// Public Key.
-    public_key: PublicKey,
-    /// Private Key.
-    private_key: PrivateKey,
+ffi! {
```
mversic, line 167: idea is that this macro will replace the structure it wraps with
```rs
pub struct KeyPair {
    ptr: [u8; 0]
}
```

in the wasm smartcontract

crypto/Cargo.toml
```diff
@@ -13,12 +13,15 @@ default = ["std"]
 std = ["ursa"]
 # Force static linking
 vendored = ["openssl-sys"]
-# Generate extern functions callable via FFI
-ffi = ["iroha_ffi"]
+## Replace structures and methods with FFI equivalents. Facilitates dynamic linkage
+#ffi = ["iroha_ffi/client"]
+
+# Expose FFI API for dynamic linking (Internal use only)
+ffi_api = ["std"]
```
mversic, line 20: activated by `iroha_core`

crypto/Cargo.toml
```diff
@@ -13,12 +13,15 @@ default = ["std"]
 std = ["ursa"]
 # Force static linking
 vendored = ["openssl-sys"]
-# Generate extern functions callable via FFI
-ffi = ["iroha_ffi"]
+## Replace structures and methods with FFI equivalents. Facilitates dynamic linkage
+#ffi = ["iroha_ffi/client"]
```
mversic, line 17: activated by `wasm` smartcontract

ffi/derive/src/wrapper.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+fn gen_return_stmt(arg: &ReturnArg) -> TokenStream {
+    let (arg_name, output_arg_conversion) = (arg.name(), gen_arg_ffi_to_src(arg, true));
+
+    quote! {
+        #output_arg_conversion
+
+        if __output == iroha_dev::FfiResult::Ok {
```
Arjentix, line 113: `iroha_dev`?

ffi/derive/tests/ui.rs
```diff
@@ -3,6 +3,7 @@ use trybuild::TestCases;
 #[test]
 fn ui() {
     let test_cases = TestCases::new();
-    test_cases.pass("tests/ui_pass/*.rs");
+    test_cases.pass("tests/ui_pass/valid.rs");
+    #[cfg(not(feature = "client"))]
```
Arjentix, line 7: Is it temporary?

ffi/derive/tests/ui_fail/derive_skip_field.stderr
```diff
@@ -1,17 +1,17 @@
 error[E0425]: cannot find function, tuple struct or tuple variant `FfiStruct__b` in this scope
```
Arjentix, line 1: Errors in this file dont help to understand that `#[getset(skip)]` should not be used. Maybe this can be improved? At least in the other issue?

ffi/derive/tests/ui_fail/derive_skip_struct.rs
```diff
@@ -1,12 +1,14 @@
...
 
 use getset::{MutGetters, Setters};
-use iroha_ffi::{ffi_export, IntoFfi, TryFromFfi, TryFromReprC};
+use iroha_ffi::{ffi_export, IntoFfi, TryFromReprC};
 
-#[derive(Clone, Setters, MutGetters, IntoFfi, TryFromFfi)]
-#[ffi_export]
+/// FfiStruct
+#[derive(Clone, Setters, MutGetters, IntoFfi, TryFromReprC)]
 #[getset(skip)]
```
Arjentix, line 8: What does `#[getset(skip)]` mean for a struct and why is this bad? Quickly looking at [getset doc](https://docs.rs/getset/latest/getset/index.html) I can't find this info

ffi/derive/tests/ui_pass/valid.rs
```diff
@@ -1,21 +1,21 @@
-use std::{collections::BTreeMap, mem::MaybeUninit};
+use std::collections::BTreeMap;
```
Arjentix, line 1: I think the filename can be improved. `valid` gives nothing new to me, because I already know that it should be valid by looking at `ui_pass` folder name

ffi/src/handle.rs
```diff
@@ -172,3 +173,74 @@ macro_rules! gen_ffi_impl {
...
+            /// # Safety
+            ///
+            /// All of the given pointers must be valid and the given handle id must match the expected
+            /// pointer type
+            #[no_mangle]
+            $vis fn __clone(
+                handle_id: $crate::handle::Id,
+                handle_ptr: *const core::ffi::c_void,
+                output_ptr: *mut *mut core::ffi::c_void
+            ) -> $crate::FfiResult;
```
Arjentix, line 194: So this is just a function declaration, but there is the actual definition/implementation? Macro name and doc says that there should be an implementation

data_model/src/name.rs
```diff
@@ -18,11 +17,21 @@ use crate::{ParseError, ValidationError};
...
+    Display,
+    Clone,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Encode,
+    Serialize,
+    IntoFfi,
```
Arjentix, line 30: Also maybe I've missed that in the code, but is `FromFfi` has blanked implementation for `T: IntoFfi`?
If yes, then I think it's more convinient to derive `FromFfi` and have blanked implementation for `IntoFfi` -- just like std has `impl<T, U: From<T>> Into<U> for T`

crypto/src/lib.rs
```diff
@@ -12,16 +12,17 @@ mod signature;
...
 use core::{fmt, str::FromStr};
 
 #[cfg(feature = "base64")]
 pub use base64;
 use derive_more::{DebugCustom, Display};
 use getset::Getters;
 pub use hash::*;
-#[cfg(feature = "ffi")]
-use iroha_ffi::{ffi_export, IntoFfi, TryFromFfi};
+#[cfg(any(feature = "ffi_api", feature = "ffi"))]
```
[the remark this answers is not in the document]
mversic: `ffi` and `ffi_api` are mutually exclusive since they do different transformations. `ffi_api` is actived in `iroha_core`, and `ffi` in `iroha_wasm`. When it becomes possible to remove the `std` feature I believe should also be able to remove the `ffi_api` feature

data_model/src/name.rs
```diff
@@ -91,9 +100,9 @@ impl FromStr for Name {
 ///
 /// All of the given pointers must be valid
 #[no_mangle]
-#[cfg(feature = "ffi")]
+#[cfg(feature = "ffi_api")]
 #[allow(non_snake_case, unsafe_code)]
-pub unsafe extern "C" fn Name__from_str<'itm>(
+unsafe extern "C" fn Name__from_str<'itm>(
```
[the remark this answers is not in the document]
mversic: as far as I understand extern functions will be visible to the linker even if private. This way the public api isn't affected

data_model/src/trigger/mod.rs
```diff
@@ -32,10 +32,10 @@ impl Registered for Trigger<FilterBox> {
     type With = Self;
 }
 
-impl<F: Filter + PartialEq> PartialOrd for Trigger<F> {
+impl<F: Filter + Eq> PartialOrd for Trigger<F> {
```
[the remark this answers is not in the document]
mversic: it could be if `PartialEq` and `Eq` were derived for every filter, but they are not

ffi/derive/src/convert.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+fn derive_try_from_ffi_for_opaque_item(name: &Ident) -> TokenStream2 {
+    #[cfg(not(feature = "client"))]
+    let owned_try_from_repr_c = quote! {
+        impl<'itm> iroha_ffi::TryFromReprC<'itm> for #name {
+            type Source = *mut Self;
+            type Store = ();
+
+            unsafe fn try_from_repr_c(source: Self::Source, _: &mut <Self as iroha_ffi::TryFromReprC<'itm>>::Store) -> Result<Self, iroha_ffi::FfiResult> {
```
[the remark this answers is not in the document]
mversic: yes, I was lazy. Corrected

ffi/derive/tests/ui_fail/derive_skip_struct.rs
mversic: it's not bad. You can use `getset(skip)`. This test confirms that if you skipped the derivation of getter for a field the corresponding extern function wasn't generated for it

ffi/derive/tests/ui_fail/derive_skip_field.stderr
mversic: this is not what this test is about. See [this comment](https://github.com/hyperledger/iroha/pull/2557#discussion_r934033768)

data_model/src/name.rs
mversic: The situation is somewhat different. `FromFfi` does not exist, there is `TryFromReprC` which has an additional associated type. I haven't found a way to support such blanket impl at the moment. I'll refer you to my previous PR [description](https://github.com/hyperledger/iroha/pull/2314#issue-1259646200)

ffi/src/handle.rs
mversic: fixed the docs

ffi/derive/tests/ui_pass/valid.rs
mversic: it holds a variety of test all of which should compile. I learnt recently that passing compile tests actually get executed so I might move them to integration tests

ffi/derive/tests/ui.rs
mversic: most likely, at least there should be equivalent for `client` feature

crypto/src/lib.rs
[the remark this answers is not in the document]
mversic: mutual exclusiveness could be enforced in the `build.rs`

data_model/src/account.rs
```diff
@@ -126,13 +129,22 @@ impl Default for SignatureCheckCondition {
...
+    Serialize,
+    IntoFfi,
+    TryFromReprC,
+    IntoSchema,
 )]
-#[cfg_attr(feature = "ffi", derive(IntoFfi, TryFromFfi))]
-#[display(fmt = "[{id}]")]
 #[id(type = "<Account as Identifiable>::Id")]
+#[allow(clippy::multiple_inherent_impl)]
+#[display(fmt = "[{id}]")]
```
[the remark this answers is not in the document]
mversic: What's the problem here? I haven't added this function

data_model/src/account.rs
```diff
@@ -126,13 +129,22 @@ impl Default for SignatureCheckCondition {
...
+    Debug,
+    Display,
+    Clone,
+    IdOrdEqHash,
+    Decode,
+    Encode,
+    Deserialize,
+    Serialize,
+    IntoFfi,
+    TryFromReprC,
```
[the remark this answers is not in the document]
mversic: they could be, I found it to look easier to read if they weren't guarded by the `ffi_api` or `ffi` feature. Would you like me to return the guard? Also, my intention is to remove the `no_std` requirement and the `ffi` feature long term in which case we'll get back to the current state

data_model/src/trigger/mod.rs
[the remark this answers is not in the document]
mversic: I restored it the way it was

crypto/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: Mutual exlusiveness should be documented. The names don't immediately tell me that I can have only one or the other. If they were called `ffi_std` and `ffi_no_std` it would be a little clearer (though the connection isn't strictly to `std`). We could call them 
- `ffi_std`/`ffi_no_std`
- `ffi_core`/`ffi_client`
- `ffi_core`/`ffi_contract`
- `ffi_no_wasm`/`ffi_wasm`

data_model/src/account.rs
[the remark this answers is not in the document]
appetrosyan: A `Display` impl never hurts.

crypto/Cargo.toml
appetrosyan: Again, name is a little confusing. Docstring isn't usually displayed for features.

data_model/Cargo.toml
```diff
@@ -22,9 +22,11 @@ default = ["std"]
 # Disabled for WASM interoperability, to reduce the binary size.
 # Please refer to https://docs.rust-embedded.org/book/intro/no-std.html
 std = ["iroha_macro/std", "iroha_version/std", "iroha_version/warp", "iroha_crypto/std", "iroha_primitives/std", "thiserror", "strum/std", "dashmap", "tokio"]
-# Generate extern functions callable via FFI
-ffi = ["iroha_crypto/ffi", "iroha_ffi"]
+## Replace structures and methods with FFI equivalents. Facilitates dynamic linkage
```
appetrosyan, line 25:
```suggestion
## Replace structures and methods with FFI equivalents. Facilitates dynamic linkage in smartcontracts
```

crypto/src/lib.rs
```diff
@@ -494,21 +496,22 @@ impl Decode for PublicKey {
...
+    #[derive(DebugCustom, Clone, PartialEq, Eq, Display, Serialize, IntoFfi, TryFromReprC)]
+    #[debug(fmt = "{{ digest: {digest_function}, payload: {:X?}}}", payload)]
+    #[display(fmt = "{}", "hex::encode(payload)")]
+    #[allow(clippy::multiple_inherent_impl)]
+    pub struct PrivateKey {
+        /// Digest function
+        digest_function: ConstString,
+        /// key payload. WARNING! Do not use `"string".as_bytes()` to obtain the key.
+        #[serde(with = "hex::serde")]
+        payload: Vec<u8>,
```
appetrosyan, line 510: We might be able to reduce the size of this as well. 

The `payload` is effectively fixed size. So what we can have is a generic wrapped by a `union`.

data_model/src/account.rs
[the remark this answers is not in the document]
appetrosyan: > Would you like me to return the guard?

We have a pattern of making things private unless we need to use them. The guard is effectively making the item private, and since we're already used to reading complex `cfg_attr` directives, it makes sense to keep them. 

This is a personal preference only, so unless it's trivial to bring back I wouldn't bother.

ffi/src/lib.rs
```diff
@@ -13,6 +16,9 @@ pub mod owned;
 mod primitives;
 pub mod slice;
 
+/// A specialized `Result` type for FFI operations
+pub type Result<T> = core::result::Result<T, FfiReturn>;
```
appetrosyan, line 20: I'm guessing that this is why you renamed `FfiResult` to `FfiReturn`.

ffi/src/handle.rs
```diff
@@ -156,19 +157,90 @@ macro_rules! gen_ffi_impl {
...
                 Ok(())
             })
         }
     };
 }
+
+/// Generate FFI equivalent declaration of the requested trait method (e.g. Clone, Eq, Ord)
+#[macro_export]
+#[cfg(feature = "client")]
+macro_rules! ffi_fn {
```
appetrosyan, line 180: Can we do better in terms of naming. @Arjentix @s8sato @Erigara ?

ffi/derive/tests/ui_pass/valid.rs
appetrosyan: > I learnt recently that passing compile tests actually get executed so I might move them to integration tests

Good idea, since they'll be considered examples by downstream users.

ffi/derive/src/convert.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+fn derive_into_ffi_for_opaque_item(name: &Ident) -> TokenStream2 {
+    #[cfg(not(feature = "client"))]
+    let owned_into_ffi = quote! {
+        impl iroha_ffi::IntoFfi for #name {
+            type Target = *mut Self;
+
+            fn into_ffi(self) -> Self::Target {
+                Box::into_raw(Box::new(self))
```
appetrosyan, line 236: Is `Box`-ing and then immediately unboxing the best or fastest way to implement it?

appetrosyan: In other words, should we set up a ticket to make this better?

ffi/derive/src/ffi_fn.rs
```diff
@@ -0,0 +1,223 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::Ident;
+
+use crate::impl_visitor::{ffi_output_arg, Arg, FnDescriptor};
+
+pub fn generate(fn_descriptor: &FnDescriptor) -> TokenStream {
```
appetrosyan, line 7: Consider adding `mod gen` and using it for namespacing.

ffi/derive/src/impl_visitor.rs
```diff
@@ -11,33 +11,33 @@ pub trait Arg {
 }
 
 #[derive(Constructor)]
-pub struct Receiver<'ast> {
```
appetrosyan, line 14: Nice.

ffi/derive/tests/ui.rs
```diff
@@ -3,6 +3,6 @@ use trybuild::TestCases;
 #[test]
 fn ui() {
     let test_cases = TestCases::new();
-    test_cases.pass("tests/ui_pass/*.rs");
+    test_cases.pass("tests/ui_pass/valid.rs");
```
appetrosyan, line 6: I'm with @Arjentix on this one.

ffi/derive/src/convert.rs
mversic: not sure what you mean. This is where ownership is transfered across the FFI. The structure needs to be placed on the heap to escape local function context.

data_model/src/name.rs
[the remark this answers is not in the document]
mversic: > We should also check with many linkers. At the very least we need ld, lld and mold.

this is not related to linkers. It's that `rustc` will not remove extern functions, even if they are private, when building libraries. LTO on a `cdylib` might affect this behavior. I can't find the issue discussion about it now, but this cannot go unseen so I'll definitely address this in the future.

data_model/src/account.rs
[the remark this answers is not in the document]
Arjentix: Well, a lot of other ffi-related code is wrapped into guards, and this derives are not anymore, that's why I asked.

I can't really say, which optnion is better:
1) Keep as is
2) Bring guards back
3) Make derives smart with guards inside generated code

crypto/src/lib.rs
mversic: very true, and worth doing. I'll open an issue

ffi/derive/tests/ui.rs
Arjentix: Let's add a TODO then

crypto/src/lib.rs
mversic: #2573

data_model/src/name.rs
```diff
@@ -91,21 +100,21 @@ impl FromStr for Name {
...
     candidate: <&'itm str as iroha_ffi::TryFromReprC<'itm>>::Source,
     out_ptr: <<Name as iroha_ffi::IntoFfi>::Target as iroha_ffi::Output>::OutPtr,
-) -> iroha_ffi::FfiResult {
+) -> iroha_ffi::FfiReturn {
```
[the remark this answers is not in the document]
mversic: to clearly disambiguate from `iroha_ffi::Result` I introduced

ffi/src/lib.rs
mversic: yeah

data_model/src/name.rs
```diff
@@ -91,21 +100,21 @@ impl FromStr for Name {
...
         // False positive - doesn't compile otherwise
         #[allow(clippy::let_unit_value)]
         let fn_body = || {
             let mut store = Default::default();
             let candidate: &str = iroha_ffi::TryFromReprC::try_from_repr_c(candidate, &mut store)?;
-            let method_res = Name::from_str(candidate)
-                .map_err(|_e| iroha_ffi::FfiResult::ExecutionFail)?
-                .into_ffi();
+            let method_res = iroha_ffi::IntoFfi::into_ffi(
+                Name::from_str(candidate).map_err(|_e| iroha_ffi::FfiReturn::ExecutionFail)?,
```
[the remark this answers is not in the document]
mversic: yeah, should be handled in #2252

data_model/src/lib.rs
```diff
@@ -909,57 +909,54 @@ pub fn current_time() -> core::time::Duration {
...
-    gen_ffi_impl! { pub Drop:
-        account::Account,
-        asset::Asset,
-        domain::Domain,
-        metadata::Metadata,
-        permissions::PermissionToken,
-        role::Role,
-        Name,
-    }
+iroha_ffi::handles! {0,
```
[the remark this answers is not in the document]
mversic: restored

crypto/src/lib.rs
```diff
@@ -574,18 +577,15 @@ impl<'de> Deserialize<'de> for PrivateKey {
...
-    use super::{KeyPair, PrivateKey, PublicKey};
-
-    handles! {0, KeyPair, PublicKey, PrivateKey}
+    use super::*;
 
-    gen_ffi_impl! { Clone: KeyPair, PublicKey, PrivateKey}
-    gen_ffi_impl! { Eq: KeyPair, PublicKey, PrivateKey}
-    gen_ffi_impl! { Ord: PublicKey }
-    gen_ffi_impl! { Drop: KeyPair, PublicKey, PrivateKey}
+    iroha_ffi::handles! {0, KeyPair, PublicKey, PrivateKey}
```
Arjentix, line 584: This syntax is still confusing me... What is `0` here? Maybe rename the macro or update the syntax?

data_model/src/lib.rs
```diff
@@ -23,10 +23,9 @@ use derive_more::Into;
 use derive_more::{AsRef, Deref, Display, From};
 use events::FilterBox;
 use iroha_crypto::{Hash, PublicKey};
-#[cfg(feature = "ffi")]
-use iroha_ffi::{IntoFfi, TryFromFfi};
+use iroha_ffi::{IntoFfi, TryFromReprC};
 use iroha_macro::{error::ErrorTryFromEnum, FromVariant};
-use iroha_primitives::{fixed, small, small::SmallVec};
+use iroha_primitives::{fixed, small};
```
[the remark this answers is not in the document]
mversic: in the code there were both `small::SmallVec` and `SmallVec` used. So I choose one. `small::Array` is also used

ffi/derive/src/convert.rs
```diff
@@ -0,0 +1,315 @@
...
+                acc
+            },
+        );
+
+    quote! {
+        impl<'itm> iroha_ffi::TryFromReprC<'itm> for #enum_name {
+            type Source = #ffi_type;
+            type Store = ();
+
+            unsafe fn try_from_repr_c(source: <Self as iroha_ffi::TryFromReprC<'itm>>::Source, _: &mut <Self as iroha_ffi::TryFromReprC<'itm>>::Store) -> iroha_ffi::Result<Self> {
```
Arjentix, line 174:
```suggestion
            unsafe fn try_from_repr_c(
                source: <Self as iroha_ffi::TryFromReprC<'itm>>::Source,
                _: &mut <Self as iroha_ffi::TryFromReprC<'itm>>::Store
            ) -> iroha_ffi::Result<Self> {
```

Arjentix: Same for other lines like that

Arjentix: Sad, that `cargo fmt` does not format macros although, I believe, there are reasons for that

data_model/src/name.rs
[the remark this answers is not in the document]
mversic: > This If should be turned into a test. Trybuild is normally there for macros, but we can use it to check if this really affects the linker.

ok, but why test compiler behavior?

> We should also check with many linkers. At the very least we need `ld`, `lld` and `mold`.

IMO this is about `rustc` not removing `extern` functions, even if they are private. `LTO` on a `cdylib` might affect this behavior. I can't find the issue discussion about it now, but this cannot go unseen so I'll definitely address this in the future. I've opened an issue

ffi/src/handle.rs
mversic: I'll take suggestions. If `client` feature is not active `ffi_fn` will produce `extern` function implementations, if it's active it'll produce declarations

ffi/derive/src/lib.rs
```diff
@@ -1,47 +1,176 @@
...
+/// }
+/// extern "C" fn Foo__bar(handle: *const Foo, output: OutSliceRef<u8>) -> FfiReturn {
+///     /* function implementation */
+///     FfiReturn::Ok
+/// }
+/// extern "C" fn Foo__id(handle: *const Foo, output: *mut u8) -> FfiReturn {
+///     /* function implementation */
+///     FfiReturn::Ok
+/// }
+/// ```
```
Arjentix, line 135: One moment. How does you macro can generate `Foo__bar` if bar was skipped with `getset(skip)` and was implemented inside `impl Foo`? You macro simply cannot obtain information about `impl Foo` block...

ffi/derive/src/util.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    quote! {
+        #ffi_conversion
+        // NOTE: `AsReprCRef` prevents ownerhip transfer over FFI
+        let #arg_name = iroha_ffi::AsReprCRef::as_ref(&#arg_name);
+    }
+}
+
+/// Parses getset attributes to find out which methods it derives
```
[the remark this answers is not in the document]
mversic: `getset` is the name of the library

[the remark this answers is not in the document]
outoftardis: oh okay

```suggestion
/// Parse `getset` attributes to find out which methods it derives
```

crypto/Cargo.toml
mversic: I was inspired by `mutable_api` feature name. It expands the API by exposing ffi functions

crypto/src/lib.rs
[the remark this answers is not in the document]
mversic: As for the `ffi_api`, I was inspired by mutable_api feature name since it expands the API by exposing ffi functions.

ffi/derive/tests/ui.rs
Arjentix: It can be `tests/ui_pass.rs` at least

data_model/src/account.rs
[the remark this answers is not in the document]
mversic: > Incidentally, how long do you reckon until we get rid of no_std?

When wasm dynamic linking is fully operational because those two heavily conflict. I would expect it removed during the next release cycle

ffi/derive/src/lib.rs
mversic: should be fixed, it was the intention to derive from the `impl` block

crypto/src/lib.rs
mversic: I've updated the description

ffi/derive/src/ffi_fn.rs
mversic: I don't understand

data_model/src/lib.rs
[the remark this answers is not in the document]
Erigara: Possible option `small::{self, SmallVec}`.

crypto/Cargo.toml
appetrosyan: `mutable_api` works, because you have the context of "this is API for the `data_model`, this is the API that allows you to mutate the objects". Here, the distinction between the `ffi` and the `ffi_api` is very murky. It won't stop me from approving, just something we should ponder before it comes to merging.

data_model/src/name.rs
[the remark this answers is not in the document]
appetrosyan: > ok, but why test compiler behavior?

Because in this instance we're relying on it. It's not 100% clear what to do with a private `extern` function. `rustc` might keep them, might change its mind to optimise binary size, and `mold` might start doing some thin LTO by default, which we would notice by a test failing. 

The preferred alternative is, of course, to enforce the inclusion by making the function public.

data_model/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: Or 
```
use small::{SmallVec, Array as SmallArray}
```

ffi/src/handle.rs
appetrosyan: That's silently changing the behaviour, and can lead to subtle bugs. I would instead prefer if 
- no `client` => `define_ffi_fn` is defined and `declare_ffi_fn` is not. 
- `client` active => `declare_ffi_fn` and `define_ffi_fn` is not.

ffi/derive/src/ffi_fn.rs
appetrosyan: What I meant to say was, 

There's a lot of functions with the `gen_` prefix. It might be more readable if they were all defined in a module named `gen`, so calls are `gen::docs` rather than `gen_docs`.

ffi/derive/src/convert.rs
appetrosyan: Rust has insignificant whitespace. Proc. macros in particular could have significant whitespace. So altering code might cause a problem. You can format region in most text editors though, so it should be done selectively.

ffi/derive/src/convert.rs
Erigara: Unboxing seems to be rather cheap operation, all we are doing is leaking `Box` and obtaining resulting pointer.

ffi/derive/tests/ui_pass/valid.rs
```diff
@@ -110,3 +116,18 @@ fn main() {
         assert_eq!(result.get_param(&Name("Nomen")), Some(&Value("Omen")));
     }
 }
+
+#[cfg(feature = "client")]
+fn main() {
+    let name = Name("X");
+
+    let mut ffi_struct: FfiStruct::new(name);
```
Erigara, line 124:
```suggestion
    let mut ffi_struct = FfiStruct::new(name);
```

ffi/derive/src/ffi_fn.rs
```diff
@@ -0,0 +1,223 @@
...
+
+fn gen_out_ptr_arg(arg: &impl Arg) -> TokenStream {
+    let arg_name = arg.name();
+    let arg_type = arg.ffi_type_resolved();
+
+    quote! { #arg_name: <#arg_type as iroha_ffi::Output>::OutPtr }
+}
+
+#[cfg(not(feature = "client"))]
+fn gen_input_conversion_stmts(fn_descriptor: &FnDescriptor) -> TokenStream {
```
Erigara, line 142: Why do we need guards here?
Seems like it's private function.

ffi/derive/src/wrapper.rs
```diff
@@ -0,0 +1,119 @@
...
+        syn::Data::Struct(_) => quote! {struct},
+        syn::Data::Union(_) => quote! {union},
+    };
+
+    match &input.data {
+        syn::Data::Enum(_) | syn::Data::Struct(_) => {
+            quote! {
+                #(#attrs)*
+                #vis #item_type #ident{
+                    ptr: [u8; 0],
```
Erigara, line 27: Why do we have type `[u8; 0]` for pointer?

ffi/derive/src/convert.rs
```diff
@@ -0,0 +1,345 @@
...
+            unsafe fn try_from_repr_c(
+                source: <Self as iroha_ffi::TryFromReprC<'itm>>::Source,
+                _: &mut <Self as iroha_ffi::TryFromReprC<'itm>>::Store
+            ) -> iroha_ffi::Result<Self> {
+                if source.is_null() {
+                    return Err(iroha_ffi::FfiReturn::ArgIsNull);
+                }
+
+                // TODO: Casting from non opaque to opaque.
+                Ok(Self(source.cast()))
```
Erigara, line 100: Is it going to work? I think opaque types is `Self { ptr, _marker }`

Erigara (commented):
IMO we should have distinctive macroses which behavior don't depend on feature flags. 
I think about this as `ffi_export` and `ffi_import`.

ffi/derive/src/ffi_fn.rs
mversic: because it activates the lint `dead_code`

ffi/derive/src/convert.rs
appetrosyan: And if we needed more control where it's allocated, with e.g. a custom allocator?

Erigara: `allocator_api` is unstable for now.

crypto/src/lib.rs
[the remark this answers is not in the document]
mversic: I introduced mutually exclusive features `ffi_import` and `ffi_export`. I hope the naming makes it more clear now

data_model/src/name.rs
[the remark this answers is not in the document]
mversic: > The preferred alternative is, of course, to enforce the inclusion by making the function public.

restored it to public visibility

ffi/tests/gen_shared_fns.rs
```diff
@@ -1,26 +1,27 @@
...
+        name: String,
+    }
 }
 
 handles! {0, FfiStruct1, FfiStruct2}
-gen_ffi_impl! {Drop: FfiStruct1, FfiStruct2}
-gen_ffi_impl! {Clone: FfiStruct1, FfiStruct2}
-gen_ffi_impl! {Eq: FfiStruct1, FfiStruct2}
-gen_ffi_impl! {Ord: FfiStruct1, FfiStruct2}
+ffi_fn! {Drop: FfiStruct1, FfiStruct2}
```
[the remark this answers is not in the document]
mversic: there are now two macros - `def_ffi_fn` and `decl_ffi_fn`

ffi/src/handle.rs
mversic: I think I resolved all of the concerns here

ffi/src/handle.rs
```diff
@@ -3,9 +3,43 @@
...
+/// assigned incrementally to every type in the macro invocation. Check the following example:
+///
+/// ```rust
+/// struct Foo1;
+/// struct Foo2;
+/// struct Bar1;
+/// struct Bar2;
+///
+/// handles! {0, Foo, Bar}
+///
```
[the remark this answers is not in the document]
mversic: thank you, I was trying to figure out what to do about this

ffi/derive/src/lib.rs
```diff
@@ -1,47 +1,178 @@
...
+
+    if !matches!(item.vis, syn::Visibility::Public(_)) {
+        abort!(item.vis, "Only public items are supported");
+    }
+
+    if !item.generics.params.is_empty() {
+        abort!(item.generics, "Generics are not supported");
+    }
+
+    quote! { #try_from_ffi_derive }.into()
```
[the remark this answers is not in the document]
mversic: the difference is `proc_macro2::TokenStream` vs `proc_macro::TokenStream`

ffi/derive/src/wrapper.rs
mversic: I believe I fixed it now

data_model/Cargo.toml
```diff
@@ -22,9 +22,11 @@ default = ["std"]
 # Disabled for WASM interoperability, to reduce the binary size.
 # Please refer to https://docs.rust-embedded.org/book/intro/no-std.html
 std = ["iroha_macro/std", "iroha_version/std", "iroha_version/warp", "iroha_crypto/std", "iroha_primitives/std", "thiserror", "strum/std", "dashmap", "tokio"]
-# Generate extern functions callable via FFI
-ffi = ["iroha_crypto/ffi", "iroha_ffi"]
+# Replace structures and methods with FFI equivalents. Facilitates dynamic linkage (mainly used in smartcontracts)
```
outoftardis, line 25:
```suggestion
# Replace structures and methods with FFI equivalents to facilitate dynamic linkage (mainly used in smartcontracts)
```

data_model/src/account.rs
```diff
@@ -195,43 +212,46 @@ impl NewAccount {
...
+        TryFromReprC,
+        IntoSchema,
+    )]
+    #[allow(clippy::multiple_inherent_impl)]
+    #[display(fmt = "({id})")] // TODO: Add more?
+    #[id(type = "Id")]
+    pub struct Account {
+        /// An Identification of the [`Account`].
+        id: <Self as Identifiable>::Id,
+        /// Asset's in this [`Account`].
```
outoftardis, line 239:
```suggestion
        /// Assets in this [`Account`].
```

data_model/src/account.rs
```diff
@@ -195,43 +212,46 @@ impl NewAccount {
...
+        assets: AssetsMap,
+        /// [`Account`]'s signatories.
+        signatories: Signatories,
+        /// Permissions tokens of this account
```
outoftardis, line 243:
```suggestion
        /// Permission tokens of this account
```

ffi/src/lib.rs
```diff
@@ -237,6 +240,23 @@ where
     type OutPtr = *mut Self;
 }
 
+/// Wrapper around struct/enum opaque pointer. When wrapped with the [`ffi`] macro in the
+/// crate linking dynamically to some`cdylib`, it replaces struct/enum body definition
```
outoftardis, line 244:
```suggestion
/// crate linking dynamically to some `cdylib`, it replaces struct/enum body definition
```

ffi/derive/src/util.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    quote! {
+        #ffi_conversion
+        // NOTE: `AsReprCRef` prevents ownerhip transfer over FFI
+        let #arg_name = iroha_ffi::AsReprCRef::as_ref(&#arg_name);
+    }
+}
+
+/// Parses `getset` attributes to find out which methods it derives
```
outoftardis, line 131:
```suggestion
/// Parse `getset` attributes to find which methods it derives
```

ffi/derive/src/lib.rs
```diff
@@ -1,47 +1,153 @@
...
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        let items = &self.0;
+        tokens.extend(quote! {#(#items)*})
+    }
+}
+
+/// Replace struct/enum definition with opaque pointer. This applies to structs/enums that
+/// are converted to an opaque pointer when sent across FFI but does not affect any other
+/// item wrapped with this macro (e.g. fieldless enums). This is so that most of the time
+/// user can safely wrap all of their structs with this macro and not be concerned with the
```
outoftardis, line 43:
```suggestion
/// users can safely wrap all of their structs with this macro and not be concerned with the
```

crypto/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: Much clearer, thank you!

ffi/derive/src/convert.rs
appetrosyan: I'll paraphrase: if we do it via a direct allocation, using the `unsafe` keyword and requesting direct access to the allocator we 
1. Clearly communicate that this is where the programmer modifying the code needs to be careful
2. Have more freedom in where and how we want the memory to be allocated (we know that this slab can be reused via a custom allocator)
3. Are more direct about our intention. In more detail: 

`Box::new` is a move with allocation. `Box::into_raw_parts` consumes the box, but retains (leaks) the allocated memory. A user doesn't see `unsafe` assumes that it's trivial and assumes that memory isn't leaked, when in reality it is.

crypto/build.rs
```diff
@@ -0,0 +1,14 @@
+//! Build script
+//!
+//! Warn if `ffi_import` and `ffi_export` features are active at the same time
+
+fn main() {
+    let ffi_import = std::env::var_os("CARGO_FEATURE_FFI_IMPORT").is_some();
```
appetrosyan, line 6: I'm with @Erigara on this one. It should be done with a `compile_error`.

appetrosyan (changes requested):

crypto/build.rs
mversic: `compile_error` would cause the compilation to fail. The check in the build script only produces warning. I can move the warning from build script to the code if you would prefer that.

Is there some other way that I am not aware of to implement feature precedence?

Erigara: @appetrosyan, currently, as @mversic said in chat, our ci uses `--all-features` which will always fail in case of `compile_error`.
Cool that `cargo-all-features` could skip incompatible sets of features.

ffi/src/handle.rs
```diff
@@ -3,9 +3,43 @@
...
+/// will produce: /*
```
Erigara, line 17:
```suggestion
/// /* will produce
```

ffi/derive/src/convert.rs
```diff
@@ -343,3 +363,21 @@ fn derive_into_ffi_for_fieldless_enum(enum_name: &Ident, repr: &[syn::NestedMeta
         }
     }
 }
+
+fn is_opaque_wrapper(input: &DeriveInput) -> bool {
```
Erigara, line 367: Good!
I'm thinking about small trick we could do when we would implement wrapper.
Inside `ffi!` macro we can put attribute-like macro `#[opaque]` at the bottom of the item attribute list and then inside `is_opaque_wrapper` we could check presence of this attribute.
Macro `#[opaque]` will just return item as is and remove itself.

Erigara: Small [proof of concept](https://github.com/Erigara/transform).

ffi/derive/src/impl_visitor.rs
```diff
@@ -3,27 +3,8 @@ use proc_macro2::Span;
...
-}
-
-pub struct InputArg {
-    self_ty: Option<syn::Path>,
-    name: Ident,
-    type_: Type,
-}
-
-pub struct ReturnArg {
+pub struct Arg {
```
Erigara, line 7: Nice, that you collapsed all this structs into one `Arg`.

ffi/src/lib.rs
```diff
@@ -237,6 +240,23 @@ where
...
+#[derive(Debug, Clone, Copy)]
+#[repr(transparent)]
+pub struct Opaque {
+    __data: [u8; 0],
```
Erigara, line 248: What is the purpose of this field?

ffi/src/handle.rs
```diff
@@ -3,9 +3,43 @@
...
+/// Implement [`crate::Handle`] for given types with the given initial handle id. Ids are
+/// assigned incrementally to every type in the macro invocation. Check the following example:
+///
+/// ```rust
+/// struct Foo1;
+/// struct Foo2;
+/// struct Bar1;
+/// struct Bar2;
+///
+/// handles! {0, Foo, Bar}
```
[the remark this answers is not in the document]
mversic: thanks

ffi/derive/src/convert.rs
mversic: you seem to have a good track record of surprising me. I was considering something like this, but I thought it wouldn't work. I see no downsides to this approach. I'll name the attribute `opaque_wrapper` though. It's not enough to indicate that it's an opaque type, it's also client side.

mversic: you seem to have a good track record of surprising me. I was considering something like this, but I thought it wouldn't work. I should name the attribute `opaque_wrapper` though. It's not enough to indicate that it's an opaque type, it's also client side.

The only downside is that `opaque_wrapper` attribute will be listed in the public API

mversic: it can also be listed as a helper attribute in the `IntoFfi` and `TryFromReprC` derive macros

ffi/src/lib.rs
mversic: this is rust idiom to represent `extern` type

ffi/derive/src/convert.rs
mversic: I'll make it a derive macro attribute, I see no difference there

crypto/build.rs
mversic: Features are not incompatible. `build.rs` only warns that precedence is taking place. How would I print the warning otherwise?

ffi/derive/src/convert.rs
mversic: better now?

crypto/build.rs
mversic: @appetrosyan are we using `cargo-all-features` yet? because in the case of precedence we will never test one of the features with `--all-features`

ffi/derive/src/convert.rs
Erigara: > I'll make it a derive macro attribute, I see no difference there

even better this way

data_model/src/lib.rs
```diff
@@ -9,24 +9,29 @@ extern crate alloc;
 
 #[cfg(not(feature = "std"))]
 use alloc::{
+    alloc::alloc,
```
Erigara, line 12: Why do we have this `alloc` import seems not to be used, no?

ffi/derive/src/convert.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    quote! {
+        impl iroha_ffi::IntoFfi for #name {
+            type Target = *mut Self;
+
+            fn into_ffi(self) -> Self::Target {
+                let layout = core::alloc::Layout::for_value(&self);
+
+                unsafe {
+                    let ptr: Self::Target = alloc(layout).cast();
```
Erigara, line 367: Because of this we had to import `alloc` in every crate that uses `IntoFfi`, right?
And we can't depend on anykind of conditional compilation here becasue we can't know feature that turns on `std` beforehand.

ffi/src/lib.rs
Erigara: Ok, it's ZST, compitable with `#[repr(C)]` and we can't construct `Opaque` because of private fields.

Erigara (approved):
LGTM, it's a bit inconvenient that we have to import `alloc` everywhere, but I think we could live with that.

crypto/build.rs
appetrosyan: Cargo all features is orthogonal to `cargo test --all-features`, with good reason. We only need to check feature combinations once every release. Making all tests take longer is a bad idea. 

Resolving.

ffi/derive/src/convert.rs
appetrosyan: Sadly yes. Fortunately, this is going to get fixed very soon.

appetrosyan (approved):

Issue #2461, opened 2022-07-08
[suggestion] [ci] improve iroha2 build


BAStos525:
### Feature request

After merging [this PR](https://github.com/hyperledger/iroha/pull/2433), Soramitsu DevOps team has to improve a couple of things in the iroha2 CI:
-  separate building of `iroha2:base` docker image into a dedicated workflow which runs on `Dockerfile.ci` change, and replace 7272721/i2-ci:latest in <job>.container.image with hyperledger/iroha2:env
- the CI task `Build and push iroha2:dev docker image` for `iroha2:dev, iroha2:stable, iroha2:release` images works only as `bash` command to build and push the corresponding docker image. For some reason, GH Action `docker/build-push-action@v3` fails with following error:
   `buildx failed with: error: failed to solve: failed to compute cache key: failed to calculate checksum of ref x8jvn8e81it8h6cbjf205gxgt::u6fgfckzgk3t5lxzms465fgd1: failed to walk /tmp/buildkit-mount878509508/release: lstat /tmp/buildkit-mount878509508/release: no such file or directory`
 DevOps team could try to resolve this issue if it is possible.

### Motivation

To make iroha2 CI faster, more effective and reliable.

### Who can help?

@appetrosyan @s8sato

appetrosyan:
A few corrections, we should **never** push a non-iroha image into a separate tag, instead we fought hard and won an image called `iroha2-ci` which we can use to pin the version of Iroha to the container that was used to test and build it (so LTS has its own, while the normal image has a separate image).

BAStos525:
Also we have to decide what to do with `rust-toolchain.toml` file according to [this comment](https://github.com/hyperledger/iroha/pull/2433#discussion_r919960840)

s8sato:
Check if the coverage behaves as expected after #2433 @s8sato

s8sato:
![image](https://user-images.githubusercontent.com/49983831/180492898-6fab9c5e-90fd-4e79-87cf-b9a3aae43d3b.png)

https://github.com/hyperledger/iroha/runs/7463816572

s8sato:
![image](https://user-images.githubusercontent.com/49983831/180498266-47718b17-02c6-4ab9-b931-504f2c2a1f34.png)

https://github.com/hyperledger/iroha/runs/7463817017

BAStos525:
[...]

PR #2565, opened 2022-07-29
[ci] #2461: Move coverage uploading to a new workflow

BAStos525:

### Description of the Change
1. Move `upload_coverage` job to a separate workflow.
2. Revert `I2::Dev::Deploy` workflow to `branch: push` trigger.
3. Come back to `docker/login-action@v2`.

### Issue
We can not use secrets in `I2::Dev::Deploy` as it was changed to `pull_request: closed` trigger because of `upload_coverage` job. So, it means that we had failed step to login to Docker Hub.
- Resolves partially #2461

### Benefits
1. Secrets should be accessible for `I2::Dev::Deploy` now.
2. `docker/login-action@v2` should work now.

### Possible Drawbacks
Some unexpected fails on CI process could be.

### DIFF
--- a/.github/workflows/iroha2-dev-upload-coverage.yml
+++ b/.github/workflows/iroha2-dev-upload-coverage.yml
@@ -1,0 +1,25 @@
+name: I2::Dev::Upload Coverage
+
+on:
+  pull_request:
+    branches: [iroha2-dev]
+    types: [closed]
+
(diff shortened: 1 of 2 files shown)


s8sato (approved):
Perfect

pesterev (approved):
LGTM

PR #2569, opened 2022-07-29
[feature] #2407: Parametrise triggers

### LINKED ISSUE #2407: Parametric smartcontracts
It should be possible to forward the event that triggered the execution of a trigger into the body of the trigger, at least in the WASM format.

Arjentix:
### Description of the Change

* Added event storing in `trigger::Set`
* `WorldStateView` now passes event from `trigger::Set` to `wasm::Runtime`
* `wasm::Runtime` was a little bit refactored and got a new function `execute_trigger()`
* Added `trigger_entrypoint` macro to `iroha_wasm` which accepts function with `event` parameter
* `mint_nft_for_every_user_smartcontract` was updated to use `trigger_entrypoint`

### Issue

* Closes #2407
* After merging this PR we should discuss possibility to make event strongly typed

### Benefits

* Now smartcontract writers can handle an event, that triggered thier smartcontract execution

### Possible Drawbacks

Events are not strong-typed. A top-level enum `Event` is used for now, which is not very handy.
I have some ideas, how to fix ths, but it needs further disscussion

### Usage Examples or Tests 

1. Comment line **173** (`#[ignore = "Only on nightly"]`) in `client/tests/integration/triggers/time_triggers.rs`
2. Run
    ```bash
    cargo +nightly-2022-04-20 test --package iroha_client --test mod --integration::triggers::time_trigger::mint_nft_for_every_user_every_1_sec --exact --nocapture
    ```


Arjentix:
@appetrosyan , @mversic , I'm hopping to discuss with you event strong-typing later. I have some ideas

data_model/src/trigger/set.rs
```diff
@@ -333,18 +337,17 @@ impl Set {
         }
     }
 
-    /// Matches and insert an [`TriggerId`] into the matched ids set.
+    /// Match and insert a [`TriggerId`] into the matched ids set.
```
outoftardis, line 340:
```suggestion
    /// Match and insert a [`TriggerId`] into the set of matched ids.
```

wasm/derive/src/lib.rs
```diff
@@ -5,85 +5,64 @@
...
 use proc_macro_error::{abort, proc_macro_error};
 use quote::quote;
-use syn::{parse_macro_input, parse_quote, ItemFn, Path, ReturnType, Signature, Type};
+use syn::{parse_macro_input, parse_quote};
 
-/// Used to annotate user-defined function which starts the execution of smartcontract
+mod smartcontract;
+mod trigger;
+
+/// Used to annotate user-defined function which starts the execution of smartcontract.
```
outoftardis, line 13:
```suggestion
/// Used to annotate the user-defined function that starts the execution of a smart contract.
```

wasm/derive/src/lib.rs
```diff
@@ -5,85 +5,64 @@
...
-    if sig.inputs.len() != 1 {
-        abort!(
-            sig.inputs,
-            "Exported function must have exactly 1 input argument of type `Account::Id`"
-        );
-    }
+/// Used to annotate user-defined function which starts the execution of trigger.
+///
+/// Should be used for trigger smartcontracts.
+/// For just smartcontract (i.e. for transactions) see [`entrypoint`].
```
outoftardis, line 28:
```suggestion
/// Used to annotate the user-defined function that starts the execution of a trigger.
///
/// Should be used for trigger smart contracts.
/// For just smart contract (for transactions), see [`entrypoint`].
```

wasm/derive/src/lib.rs
```diff
@@ -5,85 +5,64 @@
...
+use syn::{parse_macro_input, parse_quote};
 
-/// Used to annotate user-defined function which starts the execution of smartcontract
+mod smartcontract;
+mod trigger;
+
+/// Used to annotate user-defined function which starts the execution of smartcontract.
+///
+/// Should be used for smartcontracts i.e. inside transactions.
+/// For triggers see [`trigger_entrypoint`].
```
outoftardis, line 16:
```suggestion
/// Should be used for smart contracts (inside transactions).
/// For triggers, see [`trigger_entrypoint`].
```

core/src/smartcontracts/wasm.rs
```diff
@@ -465,55 +541,73 @@ impl<'wrld> Runtime<'wrld> {
...
+    ) -> Result<T, Trap> {
+        // Accessing memory as a byte slice to avoid the use of unsafe
+        let mem_range = offset as usize..(offset + len) as usize;
+        let mut bytes = &memory.data(&caller)[mem_range];
+        T::decode(&mut bytes).map_err(|error| Trap::new(error.to_string()))
+    }
+
+    /// Encode `obj` to the given `memory` with the given `alloc_fn` and `store`
+    ///
+    /// Returns the offset and length of the encoded object
```
outoftardis, line 582:
```suggestion
    /// Return the offset and length of the encoded object
```

core/src/smartcontracts/wasm.rs
```diff
@@ -441,6 +475,48 @@ impl<'wrld> Runtime<'wrld> {
         self.execute_with_state(account_id, bytes, state)
     }
 
+    /// TODO
+    ///
+    /// # Errors
+    /// TODO
```
outoftardis, line 481: will you add the description before merging?

core/src/smartcontracts/wasm.rs
```diff
@@ -22,18 +23,29 @@ use crate::{
...
+}
+
+pub mod imported {
+    //! Module functions names imported from iroha to wasm
+
+    /// Name of the imported function to execute instructions
+    pub const EXECUTE_ISI_FN_NAME: &str = "execute_instruction";
+    /// Name of the imported function to execute queries
+    pub const EXECUTE_QUERY_FN_NAME: &str = "execute_query";
+    /// Name of the imported function to debug print object
```
outoftardis, line 46:
```suggestion
    /// Name of the imported function to debug print objects
```

core/src/smartcontracts/wasm.rs
```diff
@@ -22,18 +23,29 @@ use crate::{
...
+pub mod exported {
+    //! Module functions names exported from wasm to iroha
+
+    /// Exported function to allocate memory
+    pub const WASM_ALLOC_FN: &str = "_iroha_wasm_alloc";
+    /// Name of the exported memory
+    pub const WASM_MEMORY_NAME: &str = "memory";
+    /// Name of the exported entry to smartcontract (not trigger) execution
+    pub const WASM_MAIN_FN_NAME: &str = "_iroha_wasm_main";
+    /// Name of the exported entry to trigger execution
```
outoftardis, line 35:
```suggestion
    /// Name of the exported entry for trigger execution
```

core/src/smartcontracts/wasm.rs
```diff
@@ -22,18 +23,29 @@ use crate::{
...
-/// Name of the imported function to debug print object
-pub const DBG_FN_NAME: &str = "dbg";
+pub mod exported {
+    //! Module functions names exported from wasm to iroha
+
+    /// Exported function to allocate memory
+    pub const WASM_ALLOC_FN: &str = "_iroha_wasm_alloc";
+    /// Name of the exported memory
+    pub const WASM_MEMORY_NAME: &str = "memory";
+    /// Name of the exported entry to smartcontract (not trigger) execution
```
outoftardis, line 33:
```suggestion
    /// Name of the exported entry for smartcontract (not trigger) execution
```

core/src/smartcontracts/wasm.rs
Arjentix: Oh, completely forgot about this, thanks

wasm/derive/src/lib.rs
```diff
@@ -5,85 +5,64 @@
...
-/// Used to annotate user-defined function which starts the execution of smartcontract
+mod smartcontract;
+mod trigger;
+
+/// Used to annotate the user-defined function that starts the execution of a smart contract.
+///
+/// Should be used for smart contracts (inside transactions).
+/// For triggers, see [`trigger_entrypoint`].
+//
+// TODO: Should it be renamed to something like `smartcontract_entrypoint`?
```
mversic, line 18: I think it's good as is. I would remove the comment

client/tests/integration/create_nft_for_every_user_smartcontract/src/lib.rs
```diff
@@ -14,8 +14,8 @@ use core::str::FromStr;
 
 use iroha_wasm::{data_model::prelude::*, DebugUnwrapExt, Execute};
 
-#[iroha_wasm::entrypoint]
-fn smartcontract_entry_point(_account_id: AccountId) {
+#[iroha_wasm::trigger_entrypoint]
+fn trigger_entrypoint(_account_id: AccountId, _event: Event) {
```
mversic, line 18: another possibility would have been to let smartcontract ask wsv about the context of it's execution. Here this would mean to query which event triggered it's execution. The same could be done with `AccountId`. That's pull vs push architecture. I'm not strongly advocating one vs the other, I'm just thinking what would be easier for the writer to comprehend. Especially, if the number of args in the input will grow in the future and if they'll be rarely used

core/src/smartcontracts/wasm.rs
```diff
@@ -22,18 +23,29 @@ use crate::{
...
+    pub const WASM_ALLOC_FN: &str = "_iroha_wasm_alloc";
+    /// Name of the exported memory
+    pub const WASM_MEMORY_NAME: &str = "memory";
+    /// Name of the exported entry for smart contract (not trigger) execution
+    pub const WASM_MAIN_FN_NAME: &str = "_iroha_wasm_main";
+    /// Name of the exported entry for trigger execution
+    pub const TRIGGER_MAIN_FN_NAME: &str = "_iroha_trigger_main";
+}
+
+pub mod imported {
```
mversic, line 39: I would name it `import` and `export`

core/src/smartcontracts/wasm.rs
```diff
@@ -285,11 +300,7 @@ impl<'wrld> Runtime<'wrld> {
         let alloc_fn = Self::get_alloc_fn(&mut caller)?;
         let memory = Self::get_memory(&mut caller)?;
 
-        // Accessing memory as a byte slice to avoid the use of unsafe
-        let query_mem_range = offset as usize..(offset + len) as usize;
-        let mut query_bytes = &memory.data(&caller)[query_mem_range];
-        let query =
-            QueryBox::decode(&mut query_bytes).map_err(|error| Trap::new(error.to_string()))?;
+        let query = Self::decode_from_memory(&memory, &mut caller, offset, len)?;
```
mversic, line 303:
```suggestion
        let query = Self::decode_from_memory(&mut caller, offset, len)?;
```

seems to me that you could do it this way

core/src/smartcontracts/wasm.rs
```diff
@@ -465,55 +547,73 @@ impl<'wrld> Runtime<'wrld> {
...
+        // Accessing memory as a byte slice to avoid the use of unsafe
+        let mem_range = offset as usize..(offset + len) as usize;
+        let mut bytes = &memory.data(&caller)[mem_range];
+        T::decode(&mut bytes).map_err(|error| Trap::new(error.to_string()))
+    }
+
+    /// Encode `obj` to the given `memory` with the given `alloc_fn` and `store`
+    ///
+    /// Return the offset and length of the encoded object
+    fn encode_to_memory<T: Encode>(
```
mversic, line 589:
```suggestion
    fn encode_into_memory<T: Encode>(
```

core/src/smartcontracts/wasm.rs
mversic: also, can you use `encode_to_memory` when writing into memory inside this function?

core/src/smartcontracts/wasm.rs
```diff
@@ -465,55 +547,73 @@ impl<'wrld> Runtime<'wrld> {
...
-                        "{}: export not found or not a memory",
-                        WASM_MEMORY_NAME
-                    )))
-                })?
-                .write(&mut store, acc_offset as usize, &account_bytes)
-                .map_err(|error| Trap::new(error.to_string()))?;
+            let memory = Self::get_smartcontract_memory(&smart_contract, &mut store)?;
 
-            acc_offset
+            Self::encode_to_memory(account_id, &memory, &alloc_fn, &mut store)?
```
mversic, line 560:
```suggestion
            Self::encode_to_memory(account_id, &alloc_fn, &mut store)?
```
also seems to me that you can get memory from inside the function

mversic: although it's a little bit weird that you get memory from caller when decoding and from store when encoding. As if something's not quite right with abstraction. Idk, if it can be made better

mversic (changes requested):

client/tests/integration/create_nft_for_every_user_smartcontract/src/lib.rs
appetrosyan: Personally I think that we can do a bit of both. The number of parameters of this function can be different, and we can use the macros to generate the query of the WSV as if it were a parameter, and only add the event to the smartcontract if we need it (which we can tell from the macro by either not having the parameter at all, or having it unused).

core/src/smartcontracts/wasm.rs
appetrosyan: This is a code smell, but I doubt that it can leak memory. 

New ticket?

client/tests/integration/create_nft_for_every_user_smartcontract/src/lib.rs
appetrosyan: As a bonus you now have only one `entrypoint` macro

data_model/src/trigger/set.rs
```diff
@@ -271,51 +271,55 @@ impl Set {
...
     /// This actions will be inspected in the next [`Set::handle_data_event()`] call
-    pub fn handle_data_event(&self, event: &DataEvent) {
+    // Passing by value to follow other `handle_` methods interface
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn handle_data_event(&self, event: DataEvent) {
         for entry in self.data_triggers.iter() {
             let id = entry.key();
 
-            // Check only if the trigger is domain scoped
+            // Skip if trigger is domain scoped and event is from another domain
```
appetrosyan, line 280: This comment just repeats the if. Could just remove it.

wasm/derive/src/lib.rs
```diff
@@ -5,85 +5,64 @@
...
-        abort!(sig.output, "Exported function must not have a return type");
-    }
-
-    if sig.inputs.len() != 1 {
-        abort!(
-            sig.inputs,
-            "Exported function must have exactly 1 input argument of type `Account::Id`"
-        );
-    }
+/// Used to annotate the user-defined function that starts the execution of a trigger.
```
appetrosyan, line 25:
```suggestion
/// Use to annotate the user-defined function that starts the execution of a trigger.
```

wasm/derive/src/lib.rs
```diff
@@ -5,85 +5,64 @@
...
+            abort!(fn_arg, "Exported function must not take `self` argument");
         }
     }
 }
 
-fn type_is_account_id(account_id_ty: &Type) -> bool {
-    if *account_id_ty == parse_quote!(<Account as Identifiable>::Id) {
-        return true;
+fn check_type(ty: &syn::Type, expected_type: &str) {
+    if *ty == parse_quote!(<Account as Identifiable>::Id) {
```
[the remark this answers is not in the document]
appetrosyan: Add a Ui test. See if this check can be simplified (removed).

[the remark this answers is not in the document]
appetrosyan: Also, we're mixing levels of abstraction here. If we allow a path as a type specification, we need to make sure that the entire path agrees, not just the last segment. That way you can't confuse `my::AccountId` with the expected type. 

Doing this properly is a lot of work. 

Instead consider offloading the work onto the built-in type checker. Remove this check and ensure that the type is enforced in the generated code.

wasm/derive/src/smartcontract.rs
```diff
@@ -0,0 +1,60 @@
...
+        #[no_mangle]
+        pub unsafe extern "C" fn _iroha_wasm_main(ptr: *const u8, len: usize) {
+            #fn_name(
+                iroha_wasm::_decode_from_raw::<
+                    <iroha_wasm::data_model::account::Account as iroha_wasm::data_model::Identifiable>::Id>
+                (ptr, len)
+            )
+        }
+
+        #[allow(clippy::needless_pass_by_value)]
```
appetrosyan, line 39: This doesn't look right.

wasm/derive/src/lib.rs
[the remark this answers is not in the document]
Arjentix: @mversic , you are right, it's just a peace of old code I forgot to update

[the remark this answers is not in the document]
Arjentix: > Instead consider offloading the work onto the built-in type checker. Remove this check and ensure that the type is enforced in the generated code.

Good point, I'll do that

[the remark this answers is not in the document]
Arjentix: Although this will make an error message less descriptive

client/tests/integration/create_nft_for_every_user_smartcontract/src/lib.rs
Arjentix: So do you suggest to make entrypoint function parameter-less? And query all needed parameters from fn body? To me it sounds like a step away from strong-typing from user point of view. I think an *event* which have triggered smartcontract execution will be usefull pretty always

core/src/smartcontracts/wasm.rs
Arjentix: > seems to me that you could do it this way

To retrieve `memory` from `caller` we need to call `Self::get_memory()` which will try to find the exported memory. If I'm gonna to do that in `decode_from_memory()` I have to call `get_memory()`. And if some function wants to decode two objects from the same `memory` then there will be **2** calls for memory exporting instead of one. Exporting looks like a big job, so I decided to add a `memory` parameter.

core/src/smartcontracts/wasm.rs
Arjentix: > also seems to me that you can get memory from inside the function

See [this comment](https://github.com/hyperledger/iroha/pull/2569#discussion_r936034813).
Also as an example of writing two times see `execute_trigger()`

Arjentix: > although it's a little bit weird that you get memory from caller when decoding and from store when encoding. As if something's not quite right with abstraction. Idk, if it can be made better

Hm... All I've done is just refactoring by looking and the previous existed code and trying to avoid repeating it by encapsulating in functions when I was writing the `execute_trigger()` function. I'm not really sure what does `store` means here. If you have some ideas, as Aleksandr suggested, let's create a new issue. I'm just have no idea how to describe this issue

wasm/derive/src/smartcontract.rs
Arjentix: Fixed. I just moved that from the old code

Arjentix: From the smartcontract writer point of view it makes sense to pass parameters by values. It may use them or may not

core/src/smartcontracts/wasm.rs
mversic: I think it's ok

client/tests/integration/create_nft_for_every_user_smartcontract/src/lib.rs
appetrosyan: This is relatively similar to how Rust handles CLI arguments. Most other languages accept `int argc, char** argv`, but Rust has a parameter-less entrypoint, so what we're doing wouldn't be surprising to seasoned Rust engineers.

appetrosyan: But we can also make a concession to `Event` as an argument. 

Most executables are parameter-less, so for them the `iroha_wasm::entrypoint` behaviour doesn't change. 

Some of the trigger executables need the `Event`, but most don't, which is why we didn't have it as part of the original implementation. So `iroha_wasm::trigger_entrypoint` could parse the extra argument, and simply be syntactic sugar around the query in the `fn` body. That way, if you don't need an event, you don't waste time moving it from the outers scope.

core/src/smartcontracts/wasm.rs
appetrosyan: I think this will go away if instead of pushing the event into triggers unconditionally, we instead wrap syntactic sugar around querying.

client/tests/integration/create_nft_for_every_user_smartcontract/src/lib.rs
mversic: > To me it sounds like a step away from strong-typing from user point of view.

I don' think this would affect typing

> Fortunately, the number of parameters of the entrypoint function can vary without changing the number of parameters in the generated function.

I would take this suggestion which leaves the smartcontract API pretty much unaffected. In this way user would never manually execute queries like `get_account_id` or `get_event`, instead the macro would guess the queries from the function signature. Also, there should be no need for `needless_pass_by_value` anymore

Arjentix:
After discussion we decided to implement *pull* architecture. I'm converting this PR into draft for now

Arjentix:
While updating this PR I also have tried to implement some tests for my procedural macro using `trybuild` crate.
I've spent a lot of time and realised that I simply can't do that because of problems with target architecture desired by `iroha_wasm` and problems with `std` and `no_std` depenencies.

I think, someday we will need tests for `iroha_wasm` and `iroha_wasm_derive` crates, because they are an important part of our API for external users. We need to do a research and implement it somehow.

Interesting note: `trybuild` does not activate `cfg(test)` for your macro

mversic (changes requested):

wasm/derive/src/lib.rs
```diff
@@ -1,89 +1,229 @@
...
+            return Ok(Attr::Empty);
+        }
+
+        Ok(Attr::Params(input.parse()?))
+    }
+}
+
+/// Attribute with expected parameters for smart contract entrypoint function
+struct ParamsAttr {
+    _params_kw: kw::params,
```
mversic, line 36: why the underscores in front?

wasm/derive/src/lib.rs
```diff
@@ -1,89 +1,229 @@
...
+                    }
+                }
+                ParamType::TriggeringEvent => {
+                    parse_quote! {{
+                        use ::iroha_wasm::debug::DebugExpectExt as _;
 
-    false
+                        let top_event = ::iroha_wasm::query_triggering_event();
+                        ::core::convert::TryInto::try_into(top_event)
+                            .dbg_expect("Failed to convert top-level event to the concrete one")
```
mversic, line 223: would it be possible to do the conversion before calling the smart contract, not inside of it. I wouldn't mind too much, had it not made you mess up the dbg macro. Also, it's better to move out of smart contract as much code as possible

core/src/smartcontracts/wasm.rs
```diff
@@ -438,82 +476,141 @@ impl<'wrld> Runtime<'wrld> {
...
+        let state = State::new(wsv, account_id.clone(), self.config).with_triggering_event(event);
+        let mut store = self.create_store(state)?;
+        let smart_contract = self.create_smart_contract(&mut store, bytes)?;
+
+        let main_fn = smart_contract
+            .get_typed_func(&mut store, export::TRIGGER_MAIN_FN_NAME)
+            .map_err(Error::ExportNotFound)?;
+
+        // NOTE: This function takes ownership of the pointer
+        main_fn.call(&mut store, ()).map_err(Error::ExportFnCall)
```
mversic, line 507:
```suggestion
        self.execute_with_state(bytes, state)
```

maybe this?

mversic, line 507:
```suggestion
        self.execute_with_state(bytes, state)
```

wasm/derive/src/lib.rs
mversic: what you're saying is that executing a smart contract is valid only for a certain type of event. If you registered this smart contract for some other event the execution would fail. 

My questions are:
1. Why is it not ok for a smart contract to be able to be triggered on multiple events?
2. Is this covered in the validation step? I wouldn't like to see that we've successfully registered a trigger but when the actual event occurs the execution of smart contract fails (I'm not sure how trigger validation works).

wasm/derive/src/lib.rs
Arjentix: Because this field is not used anywhere and clippy complains about it. As far as I remember it's a common practice for structures like that (i.e. `actor/derive/` and `config/derive/`)

wasm/derive/src/lib.rs
Arjentix: > Is this covered in the validation step? I wouldn't like to see that we've successfully registered a trigger but when the actual event occurs the execution of smart contract fails (I'm not sure how trigger validation works).

No yet covered. But yes, I agree, I'd like to get the error about incompatible event type at `Register` step. This is a subject for the next PR, as we discussed with @appetrosyan.

> Why is it not ok for a smart contract to be able to be triggered on multiple events?

There are two notes:
1. Currently we provide `Filter` at trigger registration step. `Filter` can work only with one concrete type of events. That's why our trigger can't receive different event types for now. Maybe this will be changed once we have better filters. That's probably a question for @appetrosyan .
2. You still can use top-level event, because there is always `impl TryInto<T> for T`.

wasm/src/debug.rs
```diff
@@ -61,11 +78,74 @@ impl<T> DebugUnwrapExt for Option<T> {
...
+
+    /// Just like `expect()` but prints error message before panic
+    fn dbg_expect(self, msg: &str) -> Self::Output;
+}
+
+impl<T, E: Debug> DebugExpectExt for Result<T, E> {
+    type Output = T;
+
+    #[allow(clippy::panic)]
+    fn dbg_expect(self, msg: &str) -> Self::Output {
```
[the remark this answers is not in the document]
Arjentix: Hm..., I'm looking at it more like [`colored crate`](https://crates.io/crates/colored) rather than like `proc_macro_error`.

With `colored` you can write code like `println!("foo".blue());`. This will display `foo` using blue color. But if you enable feature `no-color`, then this `blue()` will do nothing. This is usefull, because you can have beautiful output for normal terminals and have optional feature `dumb_terminal = ["colored/no-color"]` for terminals without color support.

The same thing here. I'd like the users to keep their code the same and only manipulate with feature flag to enable/disable debugging.

`expect_or_abort` has different meaning -- it will always abort on `Err/None` variant.

core/src/smartcontracts/wasm.rs
```diff
@@ -438,82 +476,141 @@ impl<'wrld> Runtime<'wrld> {
...
+    /// using [`encode_with_length_and_prefix()`]
+    ///
+    /// Return the offset of the encoded object
+    fn encode_with_length_prefix_into_memory<T: Encode>(
+        obj: &T,
+        memory: &wasmtime::Memory,
+        alloc_fn: &wasmtime::TypedFunc<WasmUsize, WasmUsize>,
+        mut context: impl wasmtime::AsContextMut,
+    ) -> Result<WasmUsize, Trap> {
+        let res_bytes = Self::encode_with_length_prefix(obj)?;
```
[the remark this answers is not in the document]
Arjentix: Good idea

core/src/smartcontracts/wasm.rs
```diff
@@ -22,18 +21,33 @@ use crate::{
...
-pub const DBG_FN_NAME: &str = "dbg";
+pub mod export {
+    //! Module functions names exported from wasm to iroha
+
+    /// Exported function to allocate memory
+    pub const WASM_ALLOC_FN: &str = "_iroha_wasm_alloc";
+    /// Name of the exported memory
+    pub const WASM_MEMORY_NAME: &str = "memory";
+    /// Name of the exported entry for smart contract (not trigger) execution
+    pub const WASM_MAIN_FN_NAME: &str = "_iroha_wasm_main";
```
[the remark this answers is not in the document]
Arjentix: Indeed, that's a bug. I'll use this one and remove `_iroha_trigger_main`

core/src/smartcontracts/wasm.rs
Arjentix: You are right, didn't mentioned that before

wasm/derive/src/lib.rs
mversic: Alternative would be to support only `Event` as `entrypoint` argument and leave it to the writer of smart contract to make the conversion. Not sure if I would prefer it, but it kinda makes sense because you are essentially faking strong typing in the `entrypoint` by implicitly doing the conversion

Arjentix: Well, I think, all strong-typing with WASM here actually would be faking. You can always send something different instead of i.e. `AccountId` and you won't notice that until runtime.
Here is the same thing. How can this typing be stronger (especially with *pull*-architecture).
In terms of implicit conversion. I think, your approach will allow only one way -- use top-level event. And with my approach user can use any compatible event, including top-level

Arjentix: Let's hear more voices here. Democracy!

wasm/src/debug.rs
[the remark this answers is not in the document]
Arjentix: Let's summon more voices here as well

mversic (dismissed):

wasm/derive/src/lib.rs
appetrosyan: Guys, the typing arguments are all well and good, but we should agree on a few things:

1. We need this soon. ASAP. 
2. If we enforce top-level events first and then keeping the API the same, change the underlying behaviour, we can have a smooth transition. 
3. Allowing anything but the top-level events is UB. We can define that behaviour later.

Arjentix: I don't fully understand, what do you mean, but okay. If we need this ASAP, when we need just one more approval here

appetrosyan (dismissed):

mversic (approved):

appetrosyan (approved):

PR #2575, opened 2022-08-01
[feature] #2553: Add sorting to queries

### LINKED ISSUE #2553: [suggestion] Sort query results by metadata key
### Feature request

Solving #2501 via #2515 rerquires us to store a timestamp for all assets. This doubles the memory overhead and increases the time complexity of all queries that can return multiple `Asset`s or `AssetId`s by $O ( \log n)$, where $n$ is the (usually large) number of assets over which we're paginating. 

### Motivation

Instead, a better solution would be to allow sorting of arbitrary query results based off of the metadata in each element. We still get the near-doubling of RAM usage, and time complexity for the queries which request sorting is also affected, however: 
1) Providing the timestamp is optional. If you don't need it in an asset, you don't have it. RAM impact is minimal. 
2) Sorting is optional. Most queries with pagination won't need it. 
3) Sorting is more flexible: you can sort by timestamp or by other metadata keys. 
4) The instability described in #2501 can be resolved separately without introducing a timestamp, instead by making pagination a resumable operation, that takes a snapshot of the data in the blockchain. Currently all results that aren't part of the pagination are re-computed during each query, and data is often re-created. 
5) #2515 requires minimal modifications to work. 

### Who can help?

@pesterev @anrdey

pesterev:

### Description of the Change

I've added sorting parameter into ISI queries.

The sorting will apply only to entities that have metadata.

[More about here](https://github.com/hyperledger/iroha/issues/2553).

- [x] Minimal representation of this feature.
- [x] Implement for all queries.
- [x] More tests.

### Issue

Resolves #2553 

### Benefits

Allows to sort the result of queries

### Possible Drawbacks

None.

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -119,19 +119,22 @@ pub(crate) async fn handle_queries(
     wsv: Arc<WorldStateView>,
     query_judge: QueryJudgeArc,
     pagination: Pagination,
+    sorting: Sorting,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
     let (valid_request, filter) = request.validate(&wsv, query_judge.as_ref())?;
     let original_result = valid_request.execute(&wsv)?;
     let result = filter.filter(original_result);
-    let (total, result) = if let Value::Vec(value) = result {
-        (
-            value.len(),
-            Value::Vec(value.into_iter().paginate(pagination).collect()),
-        )
+
+    let (total, result) = if let Value::Vec(vec_of_val) = result {
+        let len = vec_of_val.len();
+        let vec_of_val = apply_sorting_and_pagination(vec_of_val, &sorting, pagination);
+
+        (len, Value::Vec(vec_of_val))
     } else {
         (1, result)
     };
+
     let total = total
         .try_into()
         .map_err(|e: TryFromIntError| QueryError::Conversion(e.to_string()))?;
@@ -143,6 +146,35 @@ pub(crate) async fn handle_queries(
         total,
     };
     Ok(Scale(paginated_result.into()))
+}
+
+fn apply_sorting_and_pagination(
+    mut vec_of_val: Vec<Value>,
+    sorting: &Sorting,
+    pagination: Pagination,
+) -> Vec<Value> {
+    if let Some(ref key) = sorting.sort_by_metadata_key {
+        let f = |value1: &Value| {
+            if let Value::U128(num) = value1 {
+                *num
+            } else {
+                0
+            }
+        };
+
+        vec_of_val.sort_by_key(|value0| match value0 {
+            Value::Identifiable(IdentifiableBox::Asset(asset)) => match asset.value() {
+                AssetValue::Store(store) => store.get(key).map_or(0, f),
+                _ => 0,
+            },
+            Value::Identifiable(v) => TryInto::<&dyn HasMetadata>::try_into(v)
+                .map(|has_metadata| has_metadata.metadata().get(key).map_or(0, f))
+                .unwrap_or(0),
+            _ => 0,
+        });
+    }
+
+    vec_of_val.into_iter().paginate(pagination).collect()
 }
 
 #[derive(serde::Serialize)]
@@ -482,11 +514,12 @@ impl Torii {
                 ))
                 .and(body::versioned()),
         )
-        .or(endpoint4(
+        .or(endpoint5(
             handle_queries,
             warp::path(uri::QUERY)
                 .and(add_state!(self.wsv, self.query_judge))
                 .and(paginate())
+                .and(sorting())
                 .and(body::query()),
         ))
         .or(endpoint2(
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -611,6 +611,7 @@ impl Client {
         &self,
         request: R,
         pagination: Pagination,
+        sorting: Sorting,
         filter: PredicateBox,
     ) -> Result<(B, QueryResponseHandler<R>)>
     where
@@ -619,6 +620,7 @@ impl Client {
         B: RequestBuilder,
     {
         let pagination: Vec<_> = pagination.into();
+        let sorting: Vec<_> = sorting.into();
         let request = QueryRequest::new(request.into(), self.account_id.clone(), filter);
         let request: VersionedSignedQueryRequest = self.sign_query(request)?.into();
 
@@ -628,16 +630,62 @@ impl Client {
                 format!("{}/{}", &self.torii_url, uri::QUERY),
             )
             .params(pagination)
+            .params(sorting)
             .headers(self.headers.clone())
             .body(request.encode_versioned()),
             QueryResponseHandler::default(),
         ))
     }
 
-    /// Create a request with pagination and add the filter.
+    /// Create a request with pagination, sorting and add the filter.
     ///
     /// # Errors
-    /// Forwards from [`Self::prepare_query_request`].
+    /// Fails if sending request fails
+    pub fn request_with_pagination_and_filter_and_sorting<R>(
+        &self,
+        request: R,
+        pagination: Pagination,
+        sorting: Sorting,
+        filter: PredicateBox,
+    ) -> QueryHandlerResult<ClientQueryOutput<R>>
+    where
+        R: Query + Into<QueryBox> + Debug,
+        <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>, // Seems redundant
+    {
+        iroha_logger::trace!(?request, %pagination, ?sorting, ?filter);
+        let (req, resp_handler) = self.prepare_query_request::<R, DefaultRequestBuilder>(
+            request, pagination, sorting, filter,
+        )?;
+        let response = req.build()?.send()?;
+        resp_handler.handle(response)
+    }
+
+    /// Create a request with pagination and sorting.
+    ///
+    /// # Errors
+    /// Fails if sending request fails
+    pub fn request_with_pagination_and_sorting<R>(
+        &self,
+        request: R,
+        pagination: Pagination,
+        sorting: Sorting,
+    ) -> QueryHandlerResult<ClientQueryOutput<R>>
+    where
+        R: Query + Into<QueryBox> + Debug,
+        <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>,
+    {
+        self.request_with_pagination_and_filter_and_sorting(
+            request,
+            pagination,
+            sorting,
+            PredicateBox::default(),
+        )
+    }
+
+    /// Create a request with pagination, sorting, and the given filter.
+    ///
+    /// # Errors
+    /// Fails if sending request fails
     pub fn request_with_pagination_and_filter<R>(
         &self,
         request: R,
@@ -648,11 +696,12 @@ impl Client {
         R: Query + Into<QueryBox> + Debug,
         <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>, // Seems redundant
     {
-        iroha_logger::trace!(?request, %pagination, ?filter);
-        let (req, resp_handler) =
-            self.prepare_query_request::<R, DefaultRequestBuilder>(request, pagination, filter)?;
-        let response = req.build()?.send()?;
-        resp_handler.handle(response)
+        self.request_with_pagination_and_filter_and_sorting(
+            request,
+            pagination,
+            Sorting::default(),
+            filter,
+        )
     }
 
     /// Query API entry point. Requests queries from `Iroha` peers with pagination.
@@ -672,6 +721,22 @@ impl Client {
         <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>,
     {
         self.request_with_pagination_and_filter(request, pagination, PredicateBox::default())
+    }
+
+    /// Query API entry point. Requests queries from `Iroha` peers with sorting.
+    ///
+    /// # Errors
+    /// Fails if sending request fails
+    pub fn request_with_sorting<R>(
+        &self,
+        request: R,
+        sorting: Sorting,
+    ) -> QueryHandlerResult<ClientQueryOutput<R>>
+    where
+        R: Query + Into<QueryBox> + Debug,
+        <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>,
+    {
+        self.request_with_pagination_and_sorting(request, Pagination::default(), sorting)
     }
 
     /// Query API entry point. Requests queries from `Iroha` peers.
--- a/data_model/src/lib.rs
+++ b/data_model/src/lib.rs
@@ -57,6 +57,7 @@ pub mod permissions;
 pub mod predicate;
 pub mod query;
 pub mod role;
+pub mod sorting;
 pub mod transaction;
 pub mod trigger;
 
@@ -400,6 +401,24 @@ impl IdentifiableBox {
             IdentifiableBox::Trigger(a) => a.id().clone().into(),
             IdentifiableBox::Role(a) => a.id().clone().into(),
             IdentifiableBox::PermissionTokenDefinition(a) => a.id().clone().into(),
+        }
+    }
+}
+
+impl<'idbox> TryFrom<&'idbox IdentifiableBox> for &'idbox dyn HasMetadata {
+    type Error = ();
+
+    fn try_from(
+        v: &'idbox IdentifiableBox,
+    ) -> Result<&'idbox (dyn HasMetadata + 'idbox), Self::Error> {
+        match v {
+            IdentifiableBox::NewDomain(v) => Ok(v.as_ref()),
+            IdentifiableBox::NewAccount(v) => Ok(v.as_ref()),
+            IdentifiableBox::NewAssetDefinition(v) => Ok(v.as_ref()),
+            IdentifiableBox::Domain(v) => Ok(v.as_ref()),
+            IdentifiableBox::Account(v) => Ok(v.as_ref()),
+            IdentifiableBox::AssetDefinition(v) => Ok(v.as_ref()),
+            _ => Err(()),
         }
     }
 }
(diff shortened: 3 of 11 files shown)


data_model/src/query.rs
```diff
@@ -683,7 +682,10 @@ pub mod asset {
         Ord,
     )]
     #[display(fmt = "Find all assets")]
-    pub struct FindAllAssets;
+    pub struct FindAllAssets {
+        /// [`Name`] of key in metadata of [`Asset`]s by which sort.
```
mversic, line 686:
```suggestion
        /// [`Name`] of the key in [`Asset`]'s metadata used to order query result.
```

data_model/src/query.rs
```diff
@@ -989,8 +993,15 @@ pub mod asset {
...
         /// Construct [`FindAllAssets`].
-        pub const fn new() -> Self {
-            FindAllAssets
+        pub fn new() -> Self {
+            Self { sort_by_key: None }
+        }
+
+        /// Construct [`FindAllAssets`] with sort by key.
+        #[must_use]
+        pub fn with_sort(mut self, sort_by_key: impl Into<EvaluatesTo<Name>>) -> Self {
```
mversic, line 1002: I think that just one constructor would be enough

data_model/src/query.rs
```diff
@@ -810,6 +812,8 @@ pub mod asset {
     pub struct FindAssetsByAccountId {
         /// [`AccountId`] under which assets should be found.
         pub account_id: EvaluatesTo<AccountId>,
+        /// [`Name`] of key in metadata of [`Asset`]s by which sort.
+        pub sort_by_key: Option<EvaluatesTo<Name>>,
```
mversic, line 816:
```suggestion
        pub sort_by_metadata_key: Option<EvaluatesTo<Name>>,
```

Maybe rename to this?

client/src/client.rs
```diff
@@ -1093,10 +1093,15 @@ pub mod asset {
     use super::*;
 
     /// Get query to get all assets
-    pub const fn all() -> FindAllAssets {
+    pub fn all() -> FindAllAssets {
         FindAllAssets::new()
     }
 
+    /// Get query to get all assets with specified sorting parameter
+    pub fn all_with_sort(sort_by_key: impl Into<EvaluatesTo<Name>>) -> FindAllAssets {
```
mversic, line 1101:
```suggestion
    pub fn all_sorted(sort_key: impl Into<EvaluatesTo<Name>>) -> FindAllAssets {
```

client/src/client.rs
```diff
@@ -1114,6 +1119,14 @@ pub mod asset {
         FindAssetsByAccountId::new(account_id)
     }
 
+    /// Get query to get all assets by account id and sort parameter
+    pub fn by_account_id_with_sort(
+        account_id: impl Into<EvaluatesTo<AccountId>>,
+        sort_by_key: impl Into<EvaluatesTo<Name>>,
```
mversic, line 1125:
```suggestion
    pub fn by_account_sorted(
        account_id: impl Into<EvaluatesTo<AccountId>>,
        sort_key: impl Into<EvaluatesTo<Name>>,
```

core/src/smartcontracts/isi/asset.rs
```diff
@@ -451,8 +494,51 @@ pub mod query {
...
+                    };
+
+                    let right = match right {
+                        Value::U128(value) => value,
+                        _ => &0,
+                    };
+
+                    left.cmp(right)
+                });
+            }
```
mversic, line 539: comparison functions are the same, please extract them into a freestanding function

mversic (changes requested):
1. What are our thoughts on supporting sorting of other queries, not just asset related?
2. Maybe we can generalize this to support smart contract like server side query processing in the future?

pesterev:
@mversic 
1. Depends on prioritization. Currently, it seems we don't have the resources for that.
2. I suppose that the right decision is to add a middleware that will sort the results of queries by specific parameters and cache them. The current decision is temporary and makes it possible to use the sorting feature.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -361,14 +361,57 @@ pub mod query {
     impl ValidQuery for FindAllAssets {
         #[metrics(+"find_all_assets")]
         fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, Error> {
+            let sort_by_key = match self.sort_by_key.as_ref() {
```
appetrosyan, line 364: Move sorting to where the `pagination` and `filter` are applied in the query endpoint in `torii`. 

Sorting should be applied before all other operations. 

That way, you don't need to add sorting to all queries and all queries are sorted the same way.

appetrosyan (changes requested):

core/src/smartcontracts/isi/asset.rs
appetrosyan: Better yet, remove them into the query endpoing module, so they work independently of the query.

data_model/src/query.rs
appetrosyan: Don't mind having two. 

`#[inline]` both.

pesterev: I think no because, for a large number of cases, clients will have to specify disabled sorting like this:
```rust
FindAllAssets::new(None)
```

pesterev:
@appetrosyan I've moved the sorting to torii handler but it sorts only assets-related queries.

data_model/src/sorting.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use crate::prelude::*;
+
+const SORT_BY_KEY: &str = "sort_by_key";
+
+/// Structure for sorting requests
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct Sorting {
+    /// [`Name`] of the key in [`Asset`]'s metadata used to order query result.
+    pub sort_by_key: Option<Name>,
```
mversic, line 27: I don't think this should be optional. If additional sorting possibilities are introduced, this should be turned into an enum. Maybe you should make it enum even now

client/src/client.rs
```diff
@@ -628,33 +630,80 @@ impl Client {
...
+        )?;
         let response = req.build()?.send()?;
         resp_handler.handle(response)
     }
 
+    /// Create a request with pagination and sorting.
+    ///
+    /// # Errors
+    /// Fails if sending request fails
+    pub fn request_with_pagination_and_sorting<R>(
```
mversic, line 667: feels as if we should introduce a request builder in the client. This is getting unwieldy. Maybe open an issue?

data_model/src/sorting.rs
```diff
@@ -0,0 +1,68 @@
...
+        if let Some(key) = sorting.sort_by_key {
+            vec.push((SORT_BY_KEY, key.to_string()));
+        }
+        vec
+    }
+}
+
+#[cfg(feature = "warp")]
+/// Filter for warp which extracts sorting
+pub fn sorting() -> impl Filter<Extract = (Sorting,), Error = Rejection> + Copy {
```
mversic, line 61: shouldn't we define this in torii?

data_model/src/sorting.rs
mversic: also, I think `sort_by_metadata_key` would be more appropriate naming to make the intended use more clear

mversic (dismissed):

data_model/src/sorting.rs
mversic: you could do it like:
```rs
pub struct Sorting {
    NoSorting,
    ByKey(Name),
}
```

data_model/src/sorting.rs
pesterev: I think so but `Pagination` defined also in data_model so I decided to define `Sorting` here. It looks weird but in my opinion it should be refactored in another issue.

data_model/src/sorting.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl Default for Sorting {
+    fn default() -> Self {
+        Self::NoSorting
+    }
+}
+
+impl Sorting {
+    ///
```
outoftardis, line 39: missing description

data_model/src/sorting.rs
```diff
@@ -0,0 +1,74 @@
...
+use crate::prelude::*;
+
+const SORT_BY_KEY: &str = "sort_by_key";
+
+/// Enum for sorting requests
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Sorting {
+    /// Doesn't apply any sort.
+    NoSorting,
+    /// [`Name`] of the key in [`Asset`]'s metadata used to order query result.
```
outoftardis, line 28:
```suggestion
    /// Do not sort query result.
    NoSorting,
    /// Sort query results using [`Name`] of the key in [`Asset`]'s metadata.
```

client/src/client.rs
```diff
@@ -628,33 +630,80 @@ impl Client {
...
             .params(pagination)
+            .params(sorting)
             .headers(self.headers.clone())
             .body(request.encode_versioned()),
             QueryResponseHandler::default(),
         ))
     }
 
-    /// Create a request with pagination and add the filter.
+    /// Create a request with pagination, sorting and add the filter.
```
outoftardis, line 640:
```suggestion
    /// Create a request with pagination and sorting, add the filter.
```

or maybe 

```suggestion
    /// Create a request with pagination, sorting, and the given filter.
```

cli/src/torii/routing.rs
```diff
@@ -119,12 +119,29 @@ pub(crate) async fn handle_queries(
...
     query_judge: QueryJudgeArc,
     pagination: Pagination,
+    sorting: Sorting,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
     let (valid_request, filter) = request.validate(&wsv, query_judge.as_ref())?;
     let original_result = valid_request.execute(&wsv)?;
     let result = filter.filter(original_result);
-    let (total, result) = if let Value::Vec(value) = result {
+    let (total, result) = if let Value::Vec(mut value) = result {
```
appetrosyan, line 128: This isn't quite general-enough to work in all cases. I'd recommend using the `HasMetadata` trait and relying on the fact that most if not all things returned by queries have metadata.

client/tests/integration/asset.rs
```diff
@@ -235,3 +235,103 @@ fn client_add_asset_with_name_length_more_than_limit_should_not_commit_transacti
 
     Ok(())
 }
+
+#[test]
+fn correct_pagination_assets_after_creating_new_one() -> Result<()> {
```
appetrosyan, line 240: Excellent test. I would recommend breaking it up into smaller functions that just get called, but it's readable like this also.

data_model/src/sorting.rs
```diff
@@ -0,0 +1,74 @@
...
+    ByKey(Name),
+}
+
+impl Default for Sorting {
```
appetrosyan, line 32: Either `derive_more` or https://blog.rust-lang.org/2022/06/30/Rust-1.62.0.html#default-enum-variants

data_model/src/sorting.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl Sorting {
+    ///
+    pub fn by_key(key: Name) -> Self {
+        Self::ByKey(key)
+    }
+}
+
+impl From<Sorting> for btree_map::BTreeMap<String, String> {
```
appetrosyan, line 45: This seems very quaint. I don't think that we need this.

data_model/src/sorting.rs
appetrosyan: These should be extracted into their own `query` crate: Pagination, Predicate, Sorting and the queries themselves.

appetrosyan (changes requested):

cli/src/torii/routing.rs
pesterev: I realized that `Asset` doesn't implement `HasMetadata`. Probably because an asset does not always have a metadata. We have to rethink `HasMetadata` so that it returns `Option<Metadata>`:
```rust
trait HasMetadata {
  fn metadata(&self) -> Option<Metadata>;
}
```

pesterev:
@appetrosyan 

> These should be extracted into their own query crate: Pagination, Predicate, Sorting and the queries themselves.

The suggestion to move `queries` to a separate crate looks non-trivial, and I think it will be done in a new issue.

cli/src/torii/routing.rs
```diff
@@ -119,15 +119,37 @@ pub(crate) async fn handle_queries(
...
+                    0
+                }
+            };
+
+            vec.sort_by_key(|value0| match value0 {
+                Value::Identifiable(IdentifiableBox::Asset(asset)) => match asset.value() {
+                    AssetValue::Store(store) => store.get(key).map_or(0, f),
+                    _ => 0,
+                },
+                Value::Identifiable(v) => TryInto::<Box<&dyn HasMetadata>>::try_into(v)
```
mversic, line 143: does it have to be boxed? Maybe you could have written: `TryInto::<&dyn HasMetadata>::try_into(v)`

mversic: also there is `TryAsRef` in the `data_model` that would make sense to make use of here

data_model/src/sorting.rs
```diff
@@ -0,0 +1,52 @@
...
+impl Sorting {
+    /// Creates a sorting by [`Name`] of the key.
+    pub fn by_metadata_key(key: Name) -> Self {
+        Self {
+            sort_by_metadata_key: Some(key),
+        }
+    }
+}
+
+impl From<Sorting> for Vec<(&'static str, String)> {
```
mversic, line 33: what is the purpose of this?

pesterev: It's used in `prepare_query_request` to convert into parameters for warp [here](https://github.com/hyperledger/iroha/pull/2575/files#diff-f74337de83c9232a0cc949a8d6d7390cf6e46b011b79383d9666ede90a902526R623).

cli/src/torii/routing.rs
pesterev: To be honest, I don't have any idea how to implement `TryFrom` for raw `&dyn HasMetadata`. [See](https://github.com/hyperledger/iroha/pull/2575/files#diff-6af9c278c26aa09da5007e70f12f7210072f252c41cab1e09aef22932ac8856dR401)

pesterev: Unfortunately, `TryAsRef` implemented only for `AssetValue` not for `IdentifiableBox`.

data_model/src/lib.rs
```diff
@@ -397,6 +398,24 @@ impl IdentifiableBox {
...
+            IdentifiableBox::NewDomain(v) => Ok(Box::new(v.as_ref())),
+            IdentifiableBox::NewAccount(v) => Ok(Box::new(v.as_ref())),
+            IdentifiableBox::NewAssetDefinition(v) => Ok(Box::new(v.as_ref())),
+            IdentifiableBox::Domain(v) => Ok(Box::new(v.as_ref())),
+            IdentifiableBox::Account(v) => Ok(Box::new(v.as_ref())),
+            IdentifiableBox::AssetDefinition(v) => Ok(Box::new(v.as_ref())),
+            _ => Err(()),
+        }
+    }
+}
```
mversic, line 417:
```suggestion
impl<'live> TryFrom<&'live IdentifiableBox> for &'live dyn HasMetadata {
    type Error = ();

    fn try_from(
        v: &'live IdentifiableBox,
    ) -> Result<&'live (dyn HasMetadata + 'live), Self::Error> {
        match v {
            IdentifiableBox::NewDomain(v) => Ok(v.as_ref()),
            IdentifiableBox::NewAccount(v) => Ok(v.as_ref()),
            IdentifiableBox::NewAssetDefinition(v) => Ok(v.as_ref()),
            IdentifiableBox::Domain(v) => Ok(v.as_ref()),
            IdentifiableBox::Account(v) => Ok(v.as_ref()),
            IdentifiableBox::AssetDefinition(v) => Ok(v.as_ref()),
            _ => Err(()),
        }
    }
}
```

this should be sufficient

cli/src/torii/routing.rs
mversic: > To be honest, I don't have any idea how to implement TryFrom for raw &dyn HasMetadata. See

[see](https://github.com/hyperledger/iroha/pull/2575/files#r941015465)

> Unfortunately, TryAsRef implemented only for AssetValue not for IdentifiableBox.

I meant to say that it might be possible to implement that trait for `IdentifiableBox`

pesterev: > [see](https://github.com/hyperledger/iroha/pull/2575/files#r941015465)
damn, it works...

> I meant to say that it might be possible to implement that trait for IdentifiableBox. I don't know why do we have that trait when TryFrom would do as fine when implemented for references
Ah, I got you, absolutely agree.

mversic (dismissed):

mversic (approved):

cli/src/torii/routing.rs
```diff
@@ -119,15 +119,37 @@ pub(crate) async fn handle_queries(
...
+                    .map(|has_metadata| has_metadata.metadata().get(key).map_or(0, f))
+                    .unwrap_or(0),
+                _ => 0,
+            });
+        }
+
         (
-            value.len(),
-            Value::Vec(value.into_iter().paginate(pagination).collect()),
+            vec.len(),
```
appetrosyan, line 151: Please revert. `vec` shadows the macro. Call it `vec_of_val` if you like.

appetrosyan (changes requested):

cli/src/torii/routing.rs
```diff
@@ -119,15 +119,37 @@ pub(crate) async fn handle_queries(
...
     query_judge: QueryJudgeArc,
     pagination: Pagination,
+    sorting: Sorting,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
     let (valid_request, filter) = request.validate(&wsv, query_judge.as_ref())?;
     let original_result = valid_request.execute(&wsv)?;
     let result = filter.filter(original_result);
-    let (total, result) = if let Value::Vec(value) = result {
+    let (total, result) = if let Value::Vec(mut vec) = result {
```
appetrosyan, line 128: Consider extract into function.

data_model/src/lib.rs
```diff
@@ -397,6 +398,24 @@ impl IdentifiableBox {
     }
 }
 
+impl<'live> TryFrom<&'live IdentifiableBox> for &'live dyn HasMetadata {
```
appetrosyan, line 401:
```suggestion
impl<'idbox> TryFrom<&'idbox IdentifiableBox> for &'idbox dyn HasMetadata {
```

appetrosyan: Are the lifetimes absolutely necessary here?

client/src/client.rs
appetrosyan, line 667: I think at some point it might be better to write this as a macro, and get rid of these functions.

mversic: I suggested request builder previously. Something like:
```rs
client.query_request(request).with_pagination(pagination).with_filter(filter)
```

pesterev: I support the idea of a builder.

data_model/src/lib.rs
pesterev: yes, it's necessary

cli/src/torii/routing.rs
pesterev: I don't want to hide this if-else expression inside a function, but I can extract the part about applying the sorting and pagination into a function.

client/src/client.rs
pesterev: Anyway, I would like to do it in another PR.

data_model/src/lib.rs
mversic: while they are needed in trait definition, they are most likely not need on the method implementation below

mversic (approved):

appetrosyan (approved):

PR #2427, opened 2022-07-01
[ci] hyperledger#2309: Re-enable doc tests in CI

s8sato:
#### Features

1. Enable doc tests in `tests_with_coverage` job
2. Enable `--no-fail-fast` in `tests_with_coverage` job
3. Add tests with `--no-default-features` mainly for `no_std`
4. Consolidate `Dockerfile.build` and `Dockerfile`

#### Refactors

- Consolidate `iroha2-stable*.yml` and `iroha2-lts*.yml`
- Rename image `iroha2-ci` to `iroha2-base`

#### Fixes

- Fix doc tests

### Issue

- Closes #2309
- Closes #2640
- Opens #2604
- Opens #2623

### Benefits

1. CI will test the examples in the documentation, which have not been tested since #2223
2. CI will run all tests regardless of failure
3. CI will test the codebase with no default features, which has not been tested before
4. No need for separate build of prior images

### Possible Drawbacks

1. Takes more CI time. Allows doc tests to use nightly features
2. None
3. Consumes 2 more runners
4. None

Issue #2309, opened 2022-06-02
Re-enable doc tests in CI


s8sato:
CI has been skipping doc tests since #2223

s8sato:
Additional task:

> S. Sato, [2022/07/22 23:30]
> [In reply to Roman Shanin]
> This should be an effect of the removal of the toolchain file (https://github.com/hyperledger/iroha/pull/2433#discussion_r919960840) in #2433. I'll address in my next CI PR (https://github.com/hyperledger/iroha/pull/2427#issuecomment-1175955669)

PR #2576, opened 2022-08-01
[ci] #2461: Improve iroha2 CI

BAStos525:

### Description of the Change

1. Move `iroha2:base` image build to the separate workflow and publish it on Dockerhub in order to make it reusable.
2. Distribute `iroha2:base` and `iroha2:build` images usage among branches.
3. Update references of `building load-rs script` jobs to the [iroha2-longevity-load-rs repository](https://github.com/soramitsu/iroha2-longevity-load-rs.git).
4. Switch to `actions/checkout@v3` where it hasn't been done so far.
5. Add step to trigger workflow of `Cargo.toml` dependencies updating.

### Issue
1. This PR resolves the major part of #2461 issue.
2. It also resolves [this issue](https://github.com/soramitsu/iroha2-longevity-load-rs/issues/18) partially.

### Benefits
1. Make iroha2 CI faster and more effective.
2. Make `iroha2-ci:<tag>` and `iroha2:build-<tag>` images be available for calling from other branches.
3. Follow to the last versions of third-party actions.
6. Update workflows calling for `longevity-load-rs` repo.

### Possible Drawbacks
1. A fresh feature: CI might be failed on any of `release` workflows or deploying `iroha2-ci` image.
2. It will not work until the corresponding `Dockerfile.base` and `Dockerfile.base` will be added to `iroha2` and `iroha2-lts` branches.
3. This [note](https://github.com/hyperledger/iroha/issues/2461#issuecomment-1184150212) hasn't still resolved (@s8sato @appetrosyan). [Original link to issue](https://github.com/hyperledger/iroha/pull/2433#discussion_r919960840).
4. `hyperledger/iroha2-ci` image doesn't compatible to be used for any of iroha2 tests workflows and jobs instead of `7272721/i2-ci:latest`. I spent some time to adopt it but it was unsuccessful due to a bunch of possible reasons. So, It is not the time for that yet.
5. Firstly we should merge [this PR](https://github.com/soramitsu/iroha2-longevity-load-rs/pull/23).


s8sato:
> 3. This [note](https://github.com/hyperledger/iroha/issues/2461#issuecomment-1184150212) hasn't still resolved (@s8sato @appetrosyan). [Original link to issue](https://github.com/hyperledger/iroha/pull/2433#discussion_r919960840).

I'll address this in https://github.com/hyperledger/iroha/issues/2309#issuecomment-1192732181

BAStos525:
One thing: since you are going to use a special `iroha2-ci:stable` and `iroha2-ci:lts` images sometimes, it means that `Dockerfile.base` and `Dockerfile.build` will be different for iroha2-dev, iroha2 and iroha-lts branches. I guess it's not convenient and it will cause a conflicts while merging branches.

.github/workflows/iroha2-deploy-base-image.yml
```diff
@@ -0,0 +1,81 @@
+name: I2::Base Image Deploy
+
```
appetrosyan, line 2: We should have a way of pushing the iroha2-ci:dev image before merging into iroha2-dev.

appetrosyan: created issue

appetrosyan (approved):
LGTM.

appetrosyan:
> One thing: since you are going to use a special iroha2-ci:stable and iroha2-ci:lts images sometimes, it means that Dockerfile.base and Dockerfile.build will be different for iroha2-dev, iroha2 and iroha-lts branches. I guess it's not convenient and it will cause a conflicts while merging branches.

We can work around the inconvenience if we query the context for the branch name. 

So on iroha2 it resolves to iroha2:stable and so on.

.github/workflows/iroha2-deploy-base-image.yml
BAStos525: DevOps team got creds for the user that has write access to Hyperledger DockerHub account.
[iroha2-ci:dev image](https://hub.docker.com/layers/iroha2-ci/hyperledger/iroha2-ci/dev/images/sha256-776f52d9c1247250c84c3c00b8f2465146e65571fe6fc76177db572f6d1bde30?context=explore)

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     container:
       image: 7272721/i2-ci:latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
s8sato, line 28: The same for other 16 lines in 7 files

.github/workflows/iroha2-deploy-base-image.yml
```diff
@@ -0,0 +1,81 @@
...
+on:
+  push:
```
s8sato, line 4:
```suggestion
on:
  workflow_dispatch:
  push:
```
Just in case `archlinux:base-devel` is silently outdated

.github/workflows/iroha2-deploy-base-image.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push iroha2-ci:lts image
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: hyperledger/iroha2-ci:lts
+          context: ${{ env.REPOSITORY_URL }}#iroha2-lts
+          file: Dockerfile.base
```
s8sato, line 81:
```suggestion
jobs:
  base_image_deploy:
    strategy:
      matrix:
        tags: [dev, stable, lts]
    if: github.ref == 'refs/heads/iroha2-${{ matrix.tags }}'
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Set up buildx
        uses: docker/setup-buildx-action@v2
        id: buildx
        with:
          install: true
      - name: Login to DockerHub
        uses: docker/login-action@v2
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      - name: Build and push iroha2-ci:${{ matrix.tags }} image
        uses: docker/build-push-action@v3
        with:
          push: true
          tags: hyperledger/iroha2-ci:${{ matrix.tags }}
          context: ${{ env.REPOSITORY_URL }}#iroha2-${{ matrix.tags }}
          file: Dockerfile.base
          build-args: TAG=${{ matrix.tags }}
```

To apply this, we need to unify `iroha2` and `iroha2-stable` with `iroha2-stable`. What do you think @appetrosyan

Dockerfile.build
```diff
@@ -1,4 +1,4 @@
-FROM    iroha2:base
+FROM    hyperledger/iroha2-ci:dev
```
s8sato, line 1: Since this docker file is also used for other tags than `dev`, it would better to remove the `dev` designation:

```suggestion
ARG TAG
FROM    hyperledger/iroha2-ci:$TAG
```

.github/workflows/iroha2-deploy-base-image.yml
appetrosyan: Technically the `stable` `lts` and `dev` are all tags, not branches. Unification might be a good idea.

Dockerfile.build
appetrosyan: Excellent!

.github/workflows/iroha2-lts.yml
```diff
@@ -6,49 +6,35 @@ on:
...
 jobs:
   deploy:
     runs-on: [self-hosted, Linux]
     container:
       image: 7272721/i2-ci:latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v1
-      - name: set up buildx
```
appetrosyan, line 18: Good riddance

.github/workflows/iroha2-lts.yml
```diff
@@ -6,49 +6,35 @@ on:
...
-          context: .
-          file: Dockerfile.base
-
       - name: Build iroha2:build docker image
         uses: docker/build-push-action@v3
         with:
           load: true
           tags: iroha2:build
-          context: .
+          context: ${{ env.REPOSITORY_URL }}#iroha2-lts
```
appetrosyan, line 24: Does this result in an extra `git clone`?

.github/workflows/iroha2.yml
```diff
@@ -6,49 +6,35 @@ on:
```
s8sato, line 6: If we synchronize tags and branches and unify the `iroha2` and `iroha2-stable` __branches__ as I mentioned in https://github.com/hyperledger/iroha/pull/2576#discussion_r937453231,

```suggestion
on:
  push:
    branches: [iroha2-stable]
```

Also, the workflow filenames will be renamed to `iroha2-stable.yml`, etc.

.github/workflows/iroha2-deploy-base-image.yml
s8sato: Let me make a clarification. I'm suggesting to synchronize tags and branches.

image
image

So these branches will be `iroha2-stable`

s8sato: I think this is a question of which tag we prefer, implicit `latest` or explicit `stable`

BAStos525: I'll rename `iroha2` to `iroha2-stable` branch if so

.github/workflows/iroha2-lts.yml
BAStos525: It seems [like that](https://docs.docker.com/engine/reference/commandline/build/)

s8sato: I think `.` implies `${{ env.REPOSITORY_URL }}#branch`, thanks to `actions/checkout`

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       - "8180:8180"
     volumes:
       - './configs/peer:/config'
-    init: true
+  #  init: true
```
s8sato, line 19: What is the reason to comment out? This reverts the result of #2311

BAStos525: I apologize, it's was caused by playing #2594 (I forgot to change the branch).

s8sato (approved):

appetrosyan (approved):

.github/workflows/iroha2-lts.yml
s8sato: Taken over by https://github.com/hyperledger/iroha/commit/3796924f02191981ff7cef3ceec4a6343874a95f in #2427

PR #2582, opened 2022-08-02
[refactor] #2001: `EvaluatesTo` static type checking

### LINKED ISSUE #2001: Static type checking of `EvaluatesTo` not working
I've noticed that our type checking of `EvaluatesTo` is not giving out compile errors when it is expected to do so.

The following test should not compile, yet it does:
```rs
#[test]
    // Given
    let rt = Runtime::test();
    let (network, mut iroha_client) = rt.block_on(<Network>::start_test_with_offline(4, 1, 1));
    wait_for_genesis_committed(&network.clients(), 1);

    //When
    // NOTE: I have used `AssetDefinitionId` instead of `AccountId` for demonstration purpose
    let alice_id: AssetDefinitionId = "alice#wonderland".parse().expect("Valid");					
    let alice_has_roses = 13;

    //Then
    let assets = iroha_client
        // NOTE: This should not compile because `client::asset::by_account_id`
        // should only accept expressions that evaluate to `AccountId`!
        .request(client::asset::by_account_id(alice_id))
        .expect("Failed to execute request.");
    let asset = assets
        .iter()
        .find(|asset| asset.id().definition_id == "rose#wonderland".parse().expect("Valid"))
        .unwrap();
    assert_eq!(AssetValue::Quantity(alice_has_roses), *asset.value());
}
```

Notice that `client::asset::by_account_id` as it's argument accepts `impl Into<EvaluatesTo<AccountId>>`, yet when given `AssetDefinitionId` it still compiles as if there is an implementation `AssetDefinitionId: Into<AccountId>`.

### LINKED ISSUE #2580: [bug] Signature condition checking result is ignored in `Queue`
The `bool` inside the returning `Result`  from `Transaction::check_signature_condition()` should be checked

Arjentix:
#### EvaluatesTo (#2001)

* `Into EvaluatesTo<V>` conversion now has stricter bounds which provides static type checking
* For some cases there is no possibility to make a compile-time check (i.e. `Where`, `ContextValue` expressions and all queries) so we need a way to explicitly construct `EvaluatesTo<V>` omitting the type checking. That's why I've created `EvaluatedTo::<V>::new_unchecked()`
* Since all expressions and queries return `impl TryFrom<Value>` it's always safe to construct `EvaluatesTo<Value>`. That's why there is `EvaluatesTo::<Value>::new_evaluates_to_value()`
* Added ui test which ensures that there is no way to implicitly construct `EvaluatesTo` from uncompatible type

While doing that I've found a bug in `Queue` and fixed it. See next section.

#### Tx signature condition checking bug (#2580)

* Added `MustUse` type wrapper to `primitives`
* `MustUse` used for a checking function
* Added test which ensures that you get `unused_must_use` warning will be generated if `MustUse` type is unused
* Fixed actual bug

### Issue

* Closes #2001 
* Closes #2580 

After succesfull review I'll squash my commits into two -- one per resolved issue

#### EvaluatesTo (#2001)

* Now it's easier to write correct tests
* Found the important bug

#### Tx signature condition checking bug (#2580)

* Added helpful `MustUse` wrapper type
* Important bug fixed

### Possible Drawbacks

Harder to construct `EvaluatesTo` from types which require runtime check

### Usage Examples or Tests

You can run existing test, because a bunch of them where updated.

Also there are two new tests with `trybuild`:

#### `EvaluatesTo`

```bash
cargo test --package iroha_data_model --test ui -- ui --exact --nocapture 
```

#### `MustUse`

```bash
cargo test --package iroha_primitives --test ui -- ui --exact --nocapture
```

### Alternate Designs

@appetrosyan suggested to add `ok_or(err)` function to `MustUse` which should work something like this:

```rust
impl MustUse<bool> {
    pub fn ok_or<E>(err: E) -> Result<(), E> {
        if self.0 {
            Ok(())
        } else {
            Err(E)
        }
    }
}
```

But I have next the thoughts:

1. It's not that usable as it may look
2. [`std ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or) does not force you to return `()` as `Ok`
3. It does not follow the *Unix rule* and *Single-responsibility rule*

### DIFF
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -330,7 +330,7 @@ mod tests {
             let mut domain = Domain::new(domain_id.clone()).build();
             let account_id = AccountId::from_str("alice@wonderland").expect("Valid");
             let mut account = Account::new(account_id, [public_key]).build();
-            account.set_signature_check_condition(SignatureCheckCondition(0_u32.into()));
+            account.set_signature_check_condition(SignatureCheckCondition(false.into()));
             assert!(domain.add_account(account).is_none());
 
             Arc::new(WorldStateView::new(World::with([domain], PeersIds::new())))
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -5,7 +5,7 @@ use std::{sync::Arc, time::Duration};
 
 use crossbeam_queue::ArrayQueue;
 use dashmap::{mapref::entry::Entry, DashMap};
-use eyre::{Report, Result};
+use eyre::{eyre, Report, Result};
 use iroha_config::queue::Configuration;
 use iroha_crypto::HashOf;
 use iroha_data_model::transaction::prelude::*;
@@ -100,8 +100,13 @@ impl Queue {
             return Err(Error::InBlockchain);
         }
 
-        tx.check_signature_condition(&self.wsv)?;
-        Ok(())
+        if tx.check_signature_condition(&self.wsv)? {
+            Ok(())
+        } else {
+            Err(Error::SignatureCondition(eyre!(
+                "Signature condition check failed"
+            )))
+        }
     }
 
     /// Pushes transaction into queue.
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -367,7 +367,6 @@ mod tests {
         ));
     }
 
-    // #[ignore = "Multisignature is not working for now. See #2595"]
     #[test]
     fn push_multisignature_tx() {
         let key_pairs = [KeyPair::generate().unwrap(), KeyPair::generate().unwrap()];
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -240,12 +240,8 @@ mod tests {
     fn accepted_tx(
         account_id: &str,
         proposed_ttl_ms: u64,
-        key: Option<&KeyPair>,
+        key: KeyPair,
     ) -> VersionedAcceptedTransaction {
-        let key = key
-            .cloned()
-            .unwrap_or_else(|| KeyPair::generate().expect("Failed to generate keypair."));
-
         let message = std::iter::repeat_with(rand::random::<char>)
             .take(16)
             .collect();
@@ -265,19 +261,23 @@ mod tests {
             .expect("Failed to accept Transaction.")
     }
 
-    pub fn world_with_test_domains(public_key: PublicKey) -> World {
+    pub fn world_with_test_domains(
+        signatures: impl IntoIterator<Item = iroha_crypto::PublicKey>,
+    ) -> World {
         let domain_id = DomainId::from_str("wonderland").expect("Valid");
         let account_id = AccountId::from_str("alice@wonderland").expect("Valid");
         let mut domain = Domain::new(domain_id).build();
-        let account = Account::new(account_id, [public_key]).build();
+        let account = Account::new(account_id, signatures).build();
         assert!(domain.add_account(account).is_none());
         World::with([domain], PeersIds::new())
     }
 
     #[test]
     fn push_tx() {
-        let (public_key, _) = KeyPair::generate().unwrap().into();
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(public_key)));
+        let key_pair = KeyPair::generate().unwrap();
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([key_pair
+            .public_key()
+            .clone()])));
 
         let queue = Queue::from_configuration(
             &Configuration {
@@ -290,7 +290,7 @@ mod tests {
         );
 
         queue
-            .push(accepted_tx("alice@wonderland", 100_000, None))
+            .push(accepted_tx("alice@wonderland", 100_000, key_pair))
             .expect("Failed to push tx into queue");
     }
 
@@ -298,8 +298,10 @@ mod tests {
     fn push_tx_overflow() {
         let max_txs_in_queue = 10;
 
-        let (public_key, _) = KeyPair::generate().unwrap().into();
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(public_key)));
+        let key_pair = KeyPair::generate().unwrap();
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([key_pair
+            .public_key()
+            .clone()])));
 
         let queue = Queue::from_configuration(
             &Configuration {
@@ -313,13 +315,13 @@ mod tests {
 
         for _ in 0..max_txs_in_queue {
             queue
-                .push(accepted_tx("alice@wonderland", 100_000, None))
+                .push(accepted_tx("alice@wonderland", 100_000, key_pair.clone()))
                 .expect("Failed to push tx into queue");
             thread::sleep(Duration::from_millis(10));
         }
 
         assert!(matches!(
-            queue.push(accepted_tx("alice@wonderland", 100_000, None)),
+            queue.push(accepted_tx("alice@wonderland", 100_000, key_pair)),
             Err((_, Error::Full))
         ));
     }
@@ -327,13 +329,13 @@ mod tests {
     #[test]
     fn push_tx_signature_condition_failure() {
         let max_txs_in_queue = 10;
+        let key_pair = KeyPair::generate().unwrap();
 
         let wsv = {
-            let (public_key, _) = KeyPair::generate().unwrap().into();
             let domain_id = DomainId::from_str("wonderland").expect("Valid");
             let mut domain = Domain::new(domain_id.clone()).build();
             let account_id = AccountId::from_str("alice@wonderland").expect("Valid");
-            let mut account = Account::new(account_id, [public_key]).build();
+            let mut account = Account::new(account_id, [key_pair.public_key().clone()]).build();
             account.set_signature_check_condition(SignatureCheckCondition(false.into()));
             assert!(domain.add_account(account).is_none());
 
@@ -351,15 +353,21 @@ mod tests {
         );
 
         assert!(matches!(
-            queue.push(accepted_tx("alice@wonderland", 100_000, None)),
+            queue.push(accepted_tx("alice@wonderland", 100_000, key_pair)),
             Err((_, Error::SignatureCondition(_)))
         ));
     }
 
+    // #[ignore = "Multisignature is not working for now. See #2595"]
     #[test]
     fn push_multisignature_tx() {
-        let (public_key, _) = KeyPair::generate().unwrap().into();
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(public_key)));
+        let key_pairs = [KeyPair::generate().unwrap(), KeyPair::generate().unwrap()];
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
+            key_pairs
+                .iter()
+                .map(|key_pair| key_pair.public_key())
+                .cloned(),
+        )));
 
         let queue = Queue::from_configuration(
             &Configuration {
@@ -375,22 +383,21 @@ mod tests {
             Vec::<Instruction>::new().into(),
             100_000,
         );
-        let get_tx = || {
+        let get_tx = |key_pair| {
             let tx_limits = TransactionLimits {
                 max_instruction_number: 4096,
                 max_wasm_size_bytes: 0,
             };
             VersionedAcceptedTransaction::from_transaction(
-                tx.clone()
-                    .sign(KeyPair::generate().expect("Failed to generate keypair."))
-                    .expect("Failed to sign."),
+                tx.clone().sign(key_pair).expect("Failed to sign."),
                 &tx_limits,
             )
             .expect("Failed to accept Transaction.")
         };
 
-        queue.push(get_tx()).unwrap();
-        queue.push(get_tx()).unwrap();
+        for key_pair in key_pairs {
+            queue.push(get_tx(key_pair)).unwrap();
+        }
 
         assert_eq!(queue.queue.len(), 1);
         let signature_count = queue
@@ -407,9 +414,9 @@ mod tests {
     fn get_available_txs() {
         let max_block_tx = 2;
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
         let queue = Queue::from_configuration(
             &Configuration {
                 maximum_transactions_in_block: max_block_tx,
@@ -421,7 +428,7 @@ mod tests {
         );
         for _ in 0..5 {
             queue
-                .push(accepted_tx("alice@wonderland", 100_000, Some(&alice_key)))
+                .push(accepted_tx("alice@wonderland", 100_000, alice_key.clone()))
                 .expect("Failed to push tx into queue");
             thread::sleep(Duration::from_millis(10));
         }
@@ -434,10 +441,10 @@ mod tests {
     fn push_tx_already_in_blockchain() {
         let max_block_tx = 2;
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
-        let tx = accepted_tx("alice@wonderland", 100_000, Some(&alice_key));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
+        let tx = accepted_tx("alice@wonderland", 100_000, alice_key);
         wsv.transactions.insert(tx.hash());
         let queue = Queue::from_configuration(
             &Configuration {
@@ -456,10 +463,10 @@ mod tests {
     fn get_tx_drop_if_in_blockchain() {
         let max_block_tx = 2;
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
-        let tx = accepted_tx("alice@wonderland", 100_000, Some(&alice_key));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
+        let tx = accepted_tx("alice@wonderland", 100_000, alice_key);
         let queue = Queue::from_configuration(
             &Configuration {
                 maximum_transactions_in_block: max_block_tx,
@@ -479,9 +486,9 @@ mod tests {
     fn get_available_txs_with_timeout() {
         let max_block_tx = 6;
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
         let queue = Queue::from_configuration(
             &Configuration {
                 maximum_transactions_in_block: max_block_tx,
@@ -493,19 +500,19 @@ mod tests {
         );
         for _ in 0..(max_block_tx - 1) {
             queue
-                .push(accepted_tx("alice@wonderland", 100, Some(&alice_key)))
+                .push(accepted_tx("alice@wonderland", 100, alice_key.clone()))
                 .expect("Failed to push tx into queue");
             thread::sleep(Duration::from_millis(10));
         }
 
         queue
-            .push(accepted_tx("alice@wonderland", 200, Some(&alice_key)))
+            .push(accepted_tx("alice@wonderland", 200, alice_key.clone()))
             .expect("Failed to push tx into queue");
         std::thread::sleep(Duration::from_millis(101));
         assert_eq!(queue.get_transactions_for_block().len(), 1);
 
         queue
-            .push(accepted_tx("alice@wonderland", 300, Some(&alice_key)))
+            .push(accepted_tx("alice@wonderland", 300, alice_key))
             .expect("Failed to push tx into queue");
         std::thread::sleep(Duration::from_millis(210));
         assert_eq!(queue.get_transactions_for_block().len(), 0);
@@ -516,9 +523,9 @@ mod tests {
     #[test]
     fn transactions_available_after_pop() {
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
         let queue = Queue::from_configuration(
             &Configuration {
                 maximum_transactions_in_block: 2,
@@ -529,7 +536,7 @@ mod tests {
             wsv,
         );
         queue
-            .push(accepted_tx("alice@wonderland", 100_000, Some(&alice_key)))
+            .push(accepted_tx("alice@wonderland", 100_000, alice_key))
             .expect("Failed to push tx into queue");
 
         let a = queue
@@ -550,9 +557,9 @@ mod tests {
     fn concurrent_stress_test() {
         let max_block_tx = 10;
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
         let wsv_clone = Arc::clone(&wsv);
         let queue = Arc::new(Queue::from_configuration(
             &Configuration {
@@ -573,7 +580,7 @@ mod tests {
         // Spawn a thread where we push transactions
         let push_txs_handle = thread::spawn(move || {
             while start_time.elapsed() < run_for {
-                let tx = accepted_tx("alice@wonderland", 100_000, Some(&alice_key));
+                let tx = accepted_tx("alice@wonderland", 100_000, alice_key.clone());
                 match queue_arc_clone_1.push(tx) {
                     Ok(()) => (),
                     Err((_, Error::Full)) => (),
@@ -616,9 +623,9 @@ mod tests {
         let future_threshold_ms = 1000;
 
         let alice_key = KeyPair::generate().expect("Failed to generate keypair.");
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(
-            alice_key.public_key().clone(),
-        )));
+        let wsv = Arc::new(WorldStateView::new(world_with_test_domains([alice_key
+            .public_key()
+            .clone()])));
 
         let queue = Queue::from_configuration(
             &Configuration {
@@ -628,7 +635,7 @@ mod tests {
             wsv,
         );
 
-        let mut tx = accepted_tx("alice@wonderland", 100_000, Some(&alice_key));
+        let mut tx = accepted_tx("alice@wonderland", 100_000, alice_key);
         assert!(queue.push(tx.clone()).is_ok());
         // tamper timestamp
         tx.as_mut_v1().payload.creation_time += 2 * future_threshold_ms;
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -49,16 +49,13 @@ pub enum Error {
     #[error("Transaction is already applied")]
     InBlockchain,
     /// Signature condition check failed
-    #[error("{0}")]
-    SignatureCondition(#[from] SignatureConditionError),
-}
-
-/// Signature condition check error
-#[derive(Error, Debug)]
-#[error("Failure during signature condition execution, tx hash: {tx_hash}, reason: {reason}")]
-pub struct SignatureConditionError {
-    tx_hash: HashOf<VersionedTransaction>,
-    reason: Report,
+    #[error("Failure during signature condition execution, tx hash: {tx_hash}, reason: {reason}")]
+    SignatureCondition {
+        /// Transaction hash
+        tx_hash: HashOf<VersionedTransaction>,
+        /// Failure reason
+        reason: Report,
+    },
 }
 
 impl Queue {
@@ -114,12 +111,9 @@ impl Queue {
                     .then(|| ())
                     .ok_or_else(|| eyre!("Signature condition check failed"))
             })
-            .map_err(|reason| {
-                SignatureConditionError {
-                    tx_hash: tx.hash(),
-                    reason,
-                }
-                .into()
+            .map_err(|reason| Error::SignatureCondition {
+                tx_hash: tx.hash(),
+                reason,
             })
     }
 
@@ -369,7 +363,7 @@ mod tests {
 
         assert!(matches!(
             queue.push(accepted_tx("alice@wonderland", 100_000, key_pair)),
-            Err((_, Error::SignatureCondition(_)))
+            Err((_, Error::SignatureCondition { .. }))
         ));
     }
 
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -100,7 +100,7 @@ impl Queue {
             return Err(Error::InBlockchain);
         }
 
-        if tx.check_signature_condition(&self.wsv)? {
+        if *tx.check_signature_condition(&self.wsv)? {
             Ok(())
         } else {
             Err(Error::SignatureCondition(eyre!(
@@ -184,7 +184,7 @@ impl Queue {
             }
 
             seen.push(hash);
-            if entry
+            if *entry
                 .get()
                 .check_signature_condition(&self.wsv)
                 .expect("Checked in `check_tx` just above")
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -99,14 +99,13 @@ impl Queue {
         if tx.is_in_blockchain(&self.wsv) {
             return Err(Error::InBlockchain);
         }
-
-        if *tx.check_signature_condition(&self.wsv)? {
-            Ok(())
-        } else {
-            Err(Error::SignatureCondition(eyre!(
+        if !*tx.check_signature_condition(&self.wsv)? {
+            return Err(Error::SignatureCondition(eyre!(
                 "Signature condition check failed"
-            )))
+            )));
         }
+
+        Ok(())
     }
 
     /// Pushes transaction into queue.
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -49,8 +49,16 @@ pub enum Error {
     #[error("Transaction is already applied")]
     InBlockchain,
     /// Signature condition check failed
-    #[error("Failure during signature condition execution")]
-    SignatureCondition(#[from] Report),
+    #[error("{0}")]
+    SignatureCondition(#[from] SignatureConditionError),
+}
+
+/// Signature condition check error
+#[derive(Error, Debug)]
+#[error("Failure during signature condition execution, tx hash: {tx_hash}, reason: {reason}")]
+pub struct SignatureConditionError {
+    tx_hash: HashOf<VersionedTransaction>,
+    reason: Report,
 }
 
 impl Queue {
@@ -99,13 +107,20 @@ impl Queue {
         if tx.is_in_blockchain(&self.wsv) {
             return Err(Error::InBlockchain);
         }
-        if !*tx.check_signature_condition(&self.wsv)? {
-            return Err(Error::SignatureCondition(eyre!(
-                "Signature condition check failed"
-            )));
-        }
-
-        Ok(())
+        tx.check_signature_condition(&self.wsv)
+            .and_then(|success| {
+                success
+                    .into_inner()
+                    .then(|| ())
+                    .ok_or_else(|| eyre!("Signature condition check failed"))
+            })
+            .map_err(|reason| {
+                SignatureConditionError {
+                    tx_hash: tx.hash(),
+                    reason,
+                }
+                .into()
+            })
     }
 
     /// Pushes transaction into queue.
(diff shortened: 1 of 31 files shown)


core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -513,16 +513,23 @@ mod tests {
...
-                    Expression::Query(
-                        FindAssetQuantityById::new(AssetId::new(
-                            AssetDefinitionId::from_str("btc2eth_rate#exchange").expect("Valid"),
-                            AccountId::from_str("dex@exchange").expect("Valid"),
-                        ))
-                        .into(),
-                    ),
-                    10_u32,
-                )),
+                EvaluatesTo::new_evaluates_to_value(
```
appetrosyan, line 516: Reminds me of the tower of doom.

core/src/queue.rs
```diff
@@ -330,7 +335,7 @@ mod tests {
             let mut domain = Domain::new(domain_id.clone()).build();
             let account_id = AccountId::from_str("alice@wonderland").expect("Valid");
             let mut account = Account::new(account_id, [public_key]).build();
-            account.set_signature_check_condition(SignatureCheckCondition(0_u32.into()));
+            account.set_signature_check_condition(SignatureCheckCondition(false.into()));
```
appetrosyan, line 338: So we didn't notice this, because signature checking didn't work?

appetrosyan (approved):

core/src/queue.rs
Arjentix: Yeah. Actually this is the test where I've found signature condition checking bug, because at first it was passing check successfully with `false` as condition

data_model/src/expression.rs
```diff
@@ -65,6 +62,21 @@ impl<V: TryFrom<Value>> EvaluatesTo<V> {
         self.expression.len()
     }
 
+    /// Construct new [`EvaluatesTo`] from [`ExpressionBox`] without type checking.
+    ///
+    /// # Warning
+    /// Prefer using [`Into`] conversions rather than this method,
+    /// cause it does not check type at compile-time.
+    ///
+    /// Exists mainly to test unsupported expressions.
```
outoftardis, line 71:
```suggestion
    /// Prefer using [`Into`] conversions rather than this method,
    /// because it does not check the value type at compile-time.
    ///
    /// This method is mainly to test unsupported expressions.
```

data_model/src/account.rs
```diff
@@ -111,14 +111,15 @@ impl From<EvaluatesTo<bool>> for SignatureCheckCondition {
     }
 }
 
-/// Default signature condition check for accounts. Returns true if any of the signatories have signed a transaction.
+/// Default signature condition check for accounts.
+/// Returns true if any of the signatories have signed a transaction.
```
outoftardis, line 115:
```suggestion
/// Returns true if any of the signatories have signed the transaction.
```

appetrosyan:
I'd like to defend my proposal. 

First, I proposed introducing a trait, not an inherent `impl`, which in this case can work transitively: 

```rust
pub trait ValueIsError<T, E: Error> {
  const VALUE: Self;

  fn ok_or(&self, err: E) -> Result<T, E> {
    if *self == VALUE {
      err
    } else {
      Ok(())
    }
  }
} 

impl<E: Error> ValueIsError<(), E> for bool {
  const VALUE: Self = false;

  fn ok_or(
} 
```


1. `bool` has the function `then` which converts the `true` value into an option. It is the equivalent of `Result::map`. We need a similar pattern, except we want to convert a boolean into a short-circuit with an error message. 

The point of this isn't to reduce the typing by orders of magnitude, but to allow monadic chaining of such conversions. It might seem stupid to use `map` and `or_else` in most cases, but it allows you to flatten the decision trees. 

2. If we want to convert it to something else, we can use `map`. We can also use `then`, followed by `ok_or(Err)`, in case we want to also have a value other than `Ok(())`. 

3. If you're referring to "do one thing and do it well", then neither does `then`.

data_model/src/expression.rs
```diff
@@ -392,8 +416,14 @@ impl From<Multiply> for ExpressionBox {
     }
 }
 
+impl From<Multiply> for EvaluatesTo<u32> {
+    fn from(expression: Multiply) -> Self {
```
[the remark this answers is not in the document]
Arjentix: I'll add more `inline`s in this file

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: I'm not in the context of this game)

But I agree, this looks not so good, but I can't find a better name

Arjentix:
> Failing test

Yes, it seems like we have contradictory requirements for transaction `Queue` in our tests:

* `iroha_core::queue::tests::push_tx_signature_condition_failure` — requires to deny transaction if signature condition failied
* `iroha_client::mod::integration::multisignature_transaction::multisignature_transactions_should_wait_for_all_signatures` — requires to allow transactions from multisig account so that it can wait for other signatures

We didn't catch that before because the first test was wrong

Arjentix:
@appetrosyan , alright, now the idea of `ok_or` seems more reasonable to me.
However I stil have one question: there is `T` in trait definition and it's used as `ok_or` `Ok` value, but you return `Ok(())`. So maybe remove `T` at all?

core/src/sumeragi/fault.rs
```diff
@@ -526,7 +526,10 @@ impl<G: GenesisNetworkTrait, F: FaultInjection> SumeragiWithFault<G, F> {
             "Forwarding tx to leader"
         );
         // Don't require leader to submit receipts and therefore create blocks if the tx is still waiting for more signatures.
-        if let Ok(true) = tx.check_signature_condition(&self.wsv) {
+        if matches!(
```
Erigara, line 529: I think following looks cleaner:
```
if let Ok(MustUse(true)) = tx.check_signature_condition(&self.wsv) {
    ...
}
```

primitives/src/must_use.rs
```diff
@@ -0,0 +1,67 @@
...
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    AsRef,
+    AsMut,
+    Deref,
+)]
+#[must_use]
+pub struct MustUse<T>(T);
```
Erigara, line 48:
```suggestion
pub struct MustUse<T>(pub T);
```

This will allow us to use `MustUse` in pattern matchin and destructuring.
I see no downsides in making field public.

core/src/smartcontracts/isi/permissions/mod.rs
Erigara: Also known as [pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)).

Erigara: Does SDK developers need to wrap there queries into all this `EvaluatesTo` now or it only for internal use?

Erigara (commented):
Solution looks good in general, left few suggestions.

Some thoughts: i think when user need to wrpar expression into `EvaluatesTo::new_unchecked` it will be reason for double check, however current solution starts to get wordy when it comes nested expressions.

data_model/src/lib.rs
```diff
@@ -392,6 +392,16 @@ impl IdentifiableBox {
     }
 }
 
+/// Create a [`Vec`] containing the arguments, which should satisfy `Into<Value>` bound.
+///
+/// Syntax is the same as [`vec`](macro@vec)
```
outoftardis, line 397:
```suggestion
/// Syntax is the same as in [`vec`](macro@vec)
```

data_model/src/expression.rs
```diff
@@ -387,13 +415,21 @@ impl Multiply {
 }
 
 impl From<Multiply> for ExpressionBox {
+    #[inline]
     fn from(expression: Multiply) -> Self {
         Expression::Multiply(expression).into()
     }
 }
 
+impl From<Multiply> for EvaluatesTo<u32> {
```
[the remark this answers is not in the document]
Arjentix: Good idea. I've tryed it and there are some conflicting implementations caused by `impl<V: TryFrom<Value>, E: Into<ExpressionBox> + Into<V>> From<E> for EvaluatesTo<V>` which is also very important.
So can't do it there

core/src/tx.rs
```diff
@@ -180,8 +181,13 @@ impl TransactionValidator {
         }
 
         let option_reason = match tx.check_signature_condition(&self.wsv) {
-            Ok(true) => None,
-            Ok(false) => Some("Signature condition not satisfied.".to_owned()),
+            Ok(must_use_bool) => {
```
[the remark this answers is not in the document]
Arjentix: Because the internal field is private. I don't know if it's a good idea to make it public... I thing it's a thing that deserves voting

core/src/sumeragi/fault.rs
Arjentix: The same problem with private field. I'm not sure why is it bad to make it public, but in the same time not sure why is it good. Looks like better matching is the only benefit

primitives/src/must_use.rs
Arjentix: Ok, let's do it

core/src/sumeragi/fault.rs
Arjentix: Ok, decided to [implement it](https://github.com/hyperledger/iroha/pull/2582#discussion_r940535189)

core/src/tx.rs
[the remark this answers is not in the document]
Arjentix: Ok, decided to [implement it](https://github.com/hyperledger/iroha/pull/2582#discussion_r940535189)

core/src/smartcontracts/isi/permissions/mod.rs
Arjentix: Only internal usage. We got input from SDK using some (de-)serialization. Stong-typing affects only internal usage

Arjentix: > Also known as [pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)).
Oh, now I got it

Arjentix:
@appetrosyan and I have decided to ignore multisignature test for now. That's will be fixed in #2595

Arjentix:
@appetrosyan , I've created an issue for `ValueIsError`: #2598

Arjentix:
CI is broken :(

data_model/src/expression.rs
```diff
@@ -325,11 +348,13 @@ pub struct ContextValue {
 
 impl ContextValue {
     /// Number of underneath expressions.
+    #[inline]
```
appetrosyan, line 351: Not usually necessary for `const` functions, **but** because this function is likely to also be called in run-time, this might not be a bad idea.

data_model/src/expression.rs
```diff
@@ -325,11 +348,13 @@ pub struct ContextValue {
...
     pub const fn len(&self) -> usize {
         1
     }
 
     /// Constructs `ContextValue`.
+    #[inline]
```
appetrosyan, line 357: Doesn't work because `String::from` is allocating. This function would never be inlined.

data_model/src/expression.rs
```diff
@@ -338,6 +363,7 @@ impl ContextValue {
 }
 
 impl From<ContextValue> for ExpressionBox {
+    #[inline]
```
appetrosyan, line 366: `<Expression as Into>::into` can be arbitrarily complex, so this is not a great candidate.

data_model/src/expression.rs
```diff
@@ -373,11 +399,13 @@ pub struct Multiply {
 
 impl Multiply {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Multiply` expression.
+    #[inline]
```
appetrosyan, line 408: Bad candidate. `::into` is complex.

data_model/src/expression.rs
```diff
@@ -387,13 +415,21 @@ impl Multiply {
 }
 
 impl From<Multiply> for ExpressionBox {
+    #[inline]
```
appetrosyan, line 418: Bad candidate.

data_model/src/expression.rs
```diff
@@ -387,13 +415,21 @@ impl Multiply {
...
     fn from(expression: Multiply) -> Self {
         Expression::Multiply(expression).into()
     }
 }
 
+impl From<Multiply> for EvaluatesTo<u32> {
+    #[inline]
```
appetrosyan, line 425: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -422,11 +458,13 @@ pub struct Divide {
 
 impl Divide {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Multiply` expression.
+    #[inline]
```
appetrosyan, line 467: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -436,13 +474,21 @@ impl Divide {
 }
 
 impl From<Divide> for ExpressionBox {
+    #[inline]
```
appetrosyan, line 477: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -436,13 +474,21 @@ impl Divide {
...
     fn from(expression: Divide) -> Self {
         Expression::Divide(expression).into()
     }
 }
 
+impl From<Divide> for EvaluatesTo<u32> {
+    #[inline]
```
appetrosyan, line 484: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -471,11 +517,13 @@ pub struct Mod {
 
 impl Mod {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Mod` expression.
+    #[inline]
```
appetrosyan, line 526: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -485,13 +533,21 @@ impl Mod {
...
 
 impl From<Mod> for ExpressionBox {
+    #[inline]
     fn from(expression: Mod) -> Self {
         Expression::Mod(expression).into()
     }
 }
 
+impl From<Mod> for EvaluatesTo<u32> {
+    #[inline]
```
appetrosyan, line 543: Bad candidate. Into

data_model/src/expression.rs
```diff
@@ -520,11 +576,13 @@ pub struct RaiseTo {
 
 impl RaiseTo {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `RaiseTo` expression.
+    #[inline]
```
appetrosyan, line 585: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -569,11 +635,13 @@ pub struct Add {
 
 impl Add {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Add` expression.
+    #[inline]
```
appetrosyan, line 644: Bad candidate. `Into::into`

data_model/src/expression.rs
```diff
@@ -583,13 +651,21 @@ impl Add {
...
 
 impl From<Add> for ExpressionBox {
+    #[inline]
     fn from(expression: Add) -> Self {
         Expression::Add(expression).into()
     }
 }
 
+impl From<Add> for EvaluatesTo<u32> {
+    #[inline]
```
appetrosyan, line 661: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -619,11 +695,13 @@ pub struct Subtract {
 
 impl Subtract {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Subtract` expression.
+    #[inline]
```
appetrosyan, line 704: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -633,13 +711,21 @@ impl Subtract {
...
 
 impl From<Subtract> for ExpressionBox {
+    #[inline]
     fn from(expression: Subtract) -> Self {
         Expression::Subtract(expression).into()
     }
 }
 
+impl From<Subtract> for EvaluatesTo<u32> {
+    #[inline]
```
appetrosyan, line 721: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -668,11 +754,13 @@ pub struct Greater {
 
 impl Greater {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Greater` expression.
+    #[inline]
```
appetrosyan, line 763: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -682,13 +770,21 @@ impl Greater {
...
 
 impl From<Greater> for ExpressionBox {
+    #[inline]
     fn from(expression: Greater) -> Self {
         Expression::Greater(expression).into()
     }
 }
 
+impl From<Greater> for EvaluatesTo<bool> {
+    #[inline]
```
appetrosyan, line 780: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -717,11 +813,13 @@ pub struct Less {
 
 impl Less {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Less` expression.
+    #[inline]
```
appetrosyan, line 822: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -731,11 +829,19 @@ impl Less {
...
 
 impl From<Less> for ExpressionBox {
+    #[inline]
     fn from(expression: Less) -> Self {
         Expression::Less(expression).into()
     }
 }
 
+impl From<Less> for EvaluatesTo<bool> {
+    #[inline]
```
appetrosyan, line 839: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -773,11 +881,19 @@ impl Not {
 }
 
 impl From<Not> for ExpressionBox {
+    #[inline]
```
appetrosyan, line 884: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -807,11 +923,13 @@ pub struct And {
 
 impl And {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `And` expression.
+    #[inline]
```
appetrosyan, line 932: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -49,12 +49,9 @@ pub struct EvaluatesTo<V: TryFrom<Value>> {
     _value_type: PhantomData<V>,
```
appetrosyan, line 49: Consider adding a trait (or derive macro) for generating the imeplementations for binary operations. There's a lot of repeated boilerplate.

data_model/src/expression.rs
```diff
@@ -962,11 +1101,13 @@ pub struct If {
 
 impl If {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.condition.len() + self.then_expression.len() + self.else_expression.len() + 1
     }
 
     /// Constructs `If` expression.
+    #[inline]
```
appetrosyan, line 1110: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -889,6 +1025,7 @@ pub struct IfBuilder {
 
 impl IfBuilder {
     ///Sets the `condition`.
+    #[inline]
```
appetrosyan, line 1028: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -898,6 +1035,7 @@ impl IfBuilder {
     }
 
     /// Sets `then_expression`.
+    #[inline]
```
appetrosyan, line 1038: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -906,6 +1044,7 @@ impl IfBuilder {
     }
 
     /// Sets `else_expression`.
+    #[inline]
```
appetrosyan, line 1047: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -1020,11 +1161,13 @@ pub struct Contains {
 
 impl Contains {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.collection.len() + self.element.len() + 1
     }
 
     /// Constructs `Contains` expression.
+    #[inline]
```
appetrosyan, line 1170: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -1037,11 +1180,19 @@ impl Contains {
...
 
 impl From<Contains> for ExpressionBox {
+    #[inline]
     fn from(expression: Contains) -> Self {
         Expression::Contains(expression).into()
     }
 }
 
+impl From<Contains> for EvaluatesTo<bool> {
+    #[inline]
```
appetrosyan, line 1190: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -1073,11 +1224,13 @@ pub struct ContainsAll {
 
 impl ContainsAll {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.collection.len() + self.elements.len() + 1
     }
 
     /// Constructs `Contains` expression.
+    #[inline]
```
appetrosyan, line 1233: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -1090,11 +1243,19 @@ impl ContainsAll {
...
 
 impl From<ContainsAll> for ExpressionBox {
+    #[inline]
     fn from(expression: ContainsAll) -> Self {
         Expression::ContainsAll(expression).into()
     }
 }
 
+impl From<ContainsAll> for EvaluatesTo<bool> {
+    #[inline]
```
appetrosyan, line 1253: Bad candidate. `into`.

data_model/src/expression.rs
```diff
@@ -1125,11 +1286,13 @@ pub struct ContainsAny {
 
 impl ContainsAny {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.collection.len() + self.elements.len() + 1
     }
 
     /// Constructs `Contains` expression.
+    #[inline]
```
appetrosyan, line 1295: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -1142,11 +1305,19 @@ impl ContainsAny {
...
 
 impl From<ContainsAny> for ExpressionBox {
+    #[inline]
     fn from(expression: ContainsAny) -> Self {
         Expression::ContainsAny(expression).into()
     }
 }
 
+impl From<ContainsAny> for EvaluatesTo<bool> {
+    #[inline]
```
appetrosyan, line 1315: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -1176,11 +1347,13 @@ pub struct Equal {
 
 impl Equal {
     /// Number of underneath expressions.
+    #[inline]
     pub fn len(&self) -> usize {
         self.left.len() + self.right.len() + 1
     }
 
     /// Constructs `Or` expression.
+    #[inline]
```
appetrosyan, line 1356: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -1193,11 +1366,19 @@ impl Equal {
...
 
 impl From<Equal> for ExpressionBox {
+    #[inline]
     fn from(equal: Equal) -> Self {
         Expression::Equal(equal).into()
     }
 }
 
+impl From<Equal> for EvaluatesTo<bool> {
+    #[inline]
```
appetrosyan, line 1376: Bad candidate. `into`

data_model/src/expression.rs
```diff
@@ -1209,6 +1390,7 @@ pub struct WhereBuilder {
 
 impl WhereBuilder {
     /// Sets the `expression` to be evaluated.
+    #[inline]
```
appetrosyan, line 1393: Horrible candidate. `into`

data_model/src/expression.rs
```diff
@@ -1218,6 +1400,7 @@ impl WhereBuilder {
     }
 
     /// Binds `expression` result to a `value_name`, by which it will be reachable from the main expression.
+    #[inline]
```
appetrosyan, line 1403: Horrible candidate.

data_model/src/expression.rs
```diff
@@ -1275,6 +1458,7 @@ impl Where {
 
     /// Constructs `Or` expression.
     #[must_use]
+    #[inline]
```
appetrosyan, line 1461: Horrible candidate.

data_model/src/expression.rs
```diff
@@ -1287,19 +1471,22 @@ impl Where {
 }
 
 impl From<Where> for ExpressionBox {
+    #[inline]
```
appetrosyan, line 1474: Horrible candidate.

data_model/src/expression.rs
```diff
@@ -1287,19 +1471,22 @@ impl Where {
...
 impl QueryBox {
     /// Number of underneath expressions.
+    #[inline]
     pub const fn len(&self) -> usize {
         1
     }
 }
 
 impl From<QueryBox> for ExpressionBox {
+    #[inline]
```
appetrosyan, line 1489: Horrible candidate.

data_model/src/lib.rs
```diff
@@ -397,6 +397,16 @@ impl IdentifiableBox {
     }
 }
 
+/// Create a [`Vec`] containing the arguments, which should satisfy `Into<Value>` bound.
+///
+/// Syntax is the same as in [`vec`](macro@vec)
+#[macro_export]
+macro_rules! val_vec {
```
appetrosyan, line 404: Good!

core/src/queue.rs
```diff
@@ -261,19 +261,23 @@ mod tests {
...
+        let account = Account::new(account_id, signatures).build();
         assert!(domain.add_account(account).is_none());
         World::with([domain], PeersIds::new())
     }
 
     #[test]
     fn push_tx() {
-        let (public_key, _) = KeyPair::generate().unwrap().into();
-        let wsv = Arc::new(WorldStateView::new(world_with_test_domains(public_key)));
+        let key_pair = KeyPair::generate().unwrap();
```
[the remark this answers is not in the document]
Arjentix: What do you mean? `key_pair`? I just need this as `KeyPair` object, not as (`PublicKey`, `PrivateKey`). This object is consumed later in this test

data_model/src/expression.rs
```diff
@@ -65,6 +62,19 @@ impl<V: TryFrom<Value>> EvaluatesTo<V> {
         self.expression.len()
     }
 
+    /// Construct new [`EvaluatesTo`] from [`ExpressionBox`] without type checking.
+    ///
+    /// # Warning
+    /// Prefer using [`Into`] conversions rather than this method,
```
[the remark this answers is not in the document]
Arjentix: No, it shouldn't be `TryInto`.
All things, that guarnatees their type have something like: `impl From<OurExpression> for EvaluatesTo<GuaranteedType>`.
Like `And` has `impl From<And> for EvaluatesTo<bool>`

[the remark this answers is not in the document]
Arjentix: All other things can't check the type they are evaluated to until runtime

data_model/src/expression.rs
```diff
@@ -101,10 +111,21 @@ impl<V: TryFrom<Value>> EvaluatesTo<V> {
...
+    /// Construct `EvaluatesTo<Value>` from any `expression`
+    /// because all of them evaluate to [`Value`].
+    #[inline]
+    pub fn new_evaluates_to_value(expression: ExpressionBox) -> Self {
+        Self::new_unchecked(expression)
+    }
+}
+
 impl<V: IntoSchema + TryFrom<Value>> IntoSchema for EvaluatesTo<V> {
+    #[inline]
```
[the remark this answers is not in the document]
Arjentix: Good to know about implicit `inline`, thanks. Although this is a trait impl, so I can't use `const` here

data_model/src/expression.rs
Arjentix: Good idea. I've found a way to do it not only for binary operations

core/src/queue.rs
```diff
@@ -49,8 +49,16 @@ pub enum Error {
     #[error("Transaction is already applied")]
     InBlockchain,
     /// Signature condition check failed
-    #[error("Failure during signature condition execution")]
-    SignatureCondition(#[from] Report),
+    #[error("{0}")]
+    SignatureCondition(#[from] SignatureConditionError),
+}
+
+/// Signature condition check error
```
appetrosyan, line 56: Very good. Now we have a lot more info.

appetrosyan: One thing you can consider (and not necessarily implement) is a struct-like variant. 

We don't use them in IntoSchema structs, because it's not something that other languages support, but in this case, it might save you creating a struct, if all you need is two named fields.

data_model/src/expression.rs
```diff
@@ -24,6 +24,142 @@ use serde::{Deserialize, Serialize};
 
 use super::{query::QueryBox, Value, ValueBox};
 
+/// Generate expression structure and basic impls for it.
```
appetrosyan, line 27: The syntax is much better now. We can probably improve the syntax slightly more by having a derive macro, but that is well beyond the scope of this PR.

appetrosyan (dismissed):

core/src/queue.rs
Arjentix: Done

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
-    /// Right operand.
-    pub right: EvaluatesTo<u32>,
+        "self.left.parenthesise(Operation::Multiply)",
+        "self.right.parenthesise(Operation::Multiply)"
+    )]
+    pub Multiply(u32, u32) -> u32
+}
+
+gen_expr_and_impls! {
+    /// Evaluates to the division of right and left expressions.
```
outoftardis, line 534: I think you should specify that the left is divided by the right. E.g. you can say which expression is the dividend and which is the divisor

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
+    #[display(
+        fmt = "{}+{}",
+        "self.left.parenthesise(Operation::Add)",
+        "self.right.parenthesise(Operation::Add)"
+    )]
+    pub Add(u32, u32) -> u32
+}
+
+gen_expr_and_impls! {
+    /// Evaluates to the difference of right and left expressions.
```
outoftardis, line 634: You need to say that the right is subtracted from the left

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
-    /// Number of underneath expressions.
-    pub fn len(&self) -> usize {
-        self.left.len() + self.right.len() + 1
-    }
+/// Builder for [`If`] expression.
+#[derive(Debug)]
+#[must_use = ".build() not used"]
+pub struct IfBuilder {
+    /// Condition expression, which should evaluate to `Value::Bool`.
+    /// If it is `true` then the evaluated `then_expression` will be returned, else - evaluated `else_expression`.
```
outoftardis, line 782:
```suggestion
    /// If it is `true`, then the evaluated `then_expression` is returned. Otherwise, the evaluated `else_expression` is returned.
```

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
+    pub else_expression: Option<EvaluatesTo<Value>>,
+}
 
-    /// Constructs `Multiply` expression.
-    pub fn new(left: impl Into<EvaluatesTo<u32>>, right: impl Into<EvaluatesTo<u32>>) -> Self {
-        Self {
-            left: left.into(),
-            right: right.into(),
+impl IfBuilder {
+    ///Sets the `condition`.
```
outoftardis, line 791:
```suggestion
    /// Sets the `condition`.
```

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
+            ..self
+        }
     }
 
-    /// Constructs `Multiply` expression.
-    pub fn new(left: impl Into<EvaluatesTo<u32>>, right: impl Into<EvaluatesTo<u32>>) -> Self {
-        Self {
-            left: left.into(),
-            right: right.into(),
+    /// Returns [`If`] expression, if all the fields are filled.
```
outoftardis, line 816:
```suggestion
    /// Returns [`If`] expression if all the fields are filled.
```

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
+    ///
+    /// # Errors
+    ///
+    /// Fails if some of fields are not filled.
```
outoftardis, line 820:
```suggestion
    /// Fails if some of the fields are not filled.
```

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
-    "right.parenthesise(Operation::Mod)"
-)]
-pub struct Mod {
-    /// Left operand.
-    pub left: EvaluatesTo<u32>,
-    /// Right operand.
-    pub right: EvaluatesTo<u32>,
+gen_expr_and_impls! {
+    /// If expression. Returns either a result of `then_expression`, or a result of `else_expression`
+    /// based on the `condition`.
```
outoftardis, line 834:
```suggestion
    /// If expression. Based on the `condition`, returns the result of
    /// either `then_expression` or `else_expression`.
```

data_model/src/expression.rs
```diff
@@ -338,962 +499,561 @@ impl ContextValue {
...
+    #[must_use]
+    pub fn build(self) -> Where {
+        Where::new(self.expression, self.values)
     }
 }
 
-/// Evaluates to the right expression in power of left expressions.
-/// Works only for `Value::U32`
+/// Adds a local context of `values` for the `expression`.
+/// It is similar to *Haskell's where syntax* although, evaluated eagerly.
```
outoftardis, line 997:
```suggestion
/// It is similar to **where** syntax in Haskell.
```

Not sure what you want to say with `evaluated eagerly`

Arjentix: I think, this comment was first introduced by @appetrosyan. It's a terminology in computer science that applies to the expression evaluation order

appetrosyan: Actually It was introduced by @i1i1r, but the gist is that the `where` block is evaluated as soon as possible, and not when it's actually used.

core/src/queue.rs
```diff
@@ -261,19 +261,23 @@ mod tests {
             .expect("Failed to accept Transaction.")
     }
 
-    pub fn world_with_test_domains(public_key: PublicKey) -> World {
+    pub fn world_with_test_domains(
+        signatures: impl IntoIterator<Item = iroha_crypto::PublicKey>,
```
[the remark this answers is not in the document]
appetrosyan: > But this naming is used in Account::new(). Where should it be renamed then? Everywhere?

At some point, perhaps.

appetrosyan (approved):

appetrosyan (approved):

data_model/src/expression.rs
```diff
@@ -24,6 +24,150 @@ use serde::{Deserialize, Serialize};
...
+        }
+
+        impl From<$i> for ExpressionBox {
+            fn from(expression: $i) -> Self {
+                Expression::$i(expression).into()
+            }
+        }
+    };
+    // Internal usage: generate extra `From` impl for expressions with known result type
+    (impl_extra_convert $i:ident $result_type:ty) => {
```
Erigara, line 162:
```suggestion
    (impl_extra_convert $i:ident ?) => {
    };
    (impl_extra_convert $i:ident $result_type:ty) => {
        impl From<$i> for EvaluatesTo<$result_type> {
            fn from(expression: $i) -> Self {
                EvaluatesTo::new_unchecked(expression.into())
            }
        }
    };
```
```

File: data_model/src/expression.rs [hunk 67]
```diff
@@ -24,6 +24,150 @@ use serde::{Deserialize, Serialize};
...
+/// }
+///
+/// impl From<Add> for EvaluatesTo<u32> {
+///     fn from(expression: Add) -> Self {
+///         EvaluatesTo::new_unchecked(expression.into())
+///     }
+/// }
+/// ```
+macro_rules! gen_expr_and_impls {
+    // Case: one unnamed parameter with known result type
```
Erigara, line 103: I believe you could decrease number of variants from 6 to 3 + 1 additional internal rule by utilize [`incremental tt-matcher`](https://veykril.github.io/tlborm/decl-macros/patterns/tt-muncher.html).

```suggestion
    // Case: one unnamed parameter
    ($(#[$me:meta])* $v:vis $i:ident($first_type:ty $(,)?) -> $($tt:tt)*) => {
        gen_expr_and_impls!(impl_basic $(#[$me])* $v $i(expression: $first_type));
        gen_expr_and_impls!(impl_extra_convert $i $($tt)*);
    };
```

core/src/queue.rs
```diff
@@ -347,15 +368,21 @@ mod tests {
         );
 
         assert!(matches!(
-            queue.push(accepted_tx("alice@wonderland", 100_000, None)),
+            queue.push(accepted_tx("alice@wonderland", 100_000, key_pair)),
             Err((_, Error::SignatureCondition(_)))
         ));
     }
 
+    // #[ignore = "Multisignature is not working for now. See #2595"]
```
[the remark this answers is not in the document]
Arjentix: No, I forgot to delete this. Surprisingly this test works fine

data_model/src/expression.rs
Arjentix: Cool idea!

Arjentix: Actually I can also forward cases like *"one unnamed parameter"* and *"two unnamed parameters"* into generic case *"any number of named parameters"*

Erigara (approved):
Good job

appetrosyan (approved):

PR #2583, opened 2022-08-02
[refactor] #2118: WIP: `Default` config

ilchu:

### Description of the Change

- [x] A new trait `Combine` in `iroha_config_base::proxy` and its corresponding error
- [x] Refactor `iroha_config_base::derive` crate into modules
- [x] `Proxy` derive macro
- [ ] Deeper refactor of `Configurable` trait to make it more readable
- [ ] Make `Proxy` completely conform to `Configurable` and implement all its methods
- [x] Placeholder trait `DocsDefault` to make `kagami` work for the time being 

All of these changes won't probably go into a final real PR (quite likely nothing in its current form besides the crate refactor), so instead this draft may serve as a ground for comments that can become guidelines for my future work. Moreover, the code doesn't even compile at the time of writing due to the unfinished point 5 from the list above.

### Issue

Was part of work towards #2118, but in a discussion with @appetrosyan it was decided to make it an epic and split it into more smaller ones to make the progress more trackable. Will update this part after most of them are created.

### Benefits
Ideally, the end result should not only allow flexible loading of configuration, but also give informative messages/suggestions in case of failure. E.g. if no keys were provided, suggest a command to generate them (or have a flag for generation on the fly). Moreover, it wouldn't hurt if at start Iroha logs explicitly, which parts of configs are loaded as default values due to them not being provided. 

### Possible Drawbacks
Current design of `Configurable` trait suffers from interface multiplicity, that is, it serves two functions at once which should eventually be decoupled: things pertaining to configuration loading and initialization, and all the things related to querying it or modifying parameters at runtime.

### Alternate Designs
Not necessarily alternate designs, but a few things mentioned in previous discussion:
* To save space, some of the values in config proxies could be stored as `ConstStr`.
* Typechecking in procedural macros. This is more of a general issue relating to all our macro work. Instead of trying to naively check some AST node ident against a string, it would make sense to have a small function that's part of the macro expansion doing this job.
* Despite all of the fields in `ConfigProxy` being wrapped in `Option` to signify the possibility of them being uninitialized, there are some fields which have to be initialized eventually, and those that don't and can instead rely on the default value. This distinction could be marked by having a separate struct for the non-defaultable option fields.


config/base/src/lib.rs
```diff
@@ -256,3 +271,59 @@ pub mod view {
...
+        /// Construct [`Self`] from a path-like object.
+        ///
+        /// # Errors
+        /// - File not found.
+        /// - File found, but peer configuration parsing failed.
+        /// - The length of the array in raw JSON representation is different
+        /// from the length of the array in
+        /// [`self.sumeragi.trusted_peers.peers`], most likely due to two
+        /// (or more) peers having the same public key.
+        fn from_path<P: AsRef<Path> + Debug + Clone>(path: P) -> eyre::Result<Self, eyre::Error> {
```
ilchu, line 306: Don't know how robust it is to try to make a default `from_path` method for all proxy structs, but there definitely is some space for improvement in existing code, as we have 3 or 4 implementations of almost the same function scattered as `impl Configuration` here and there.

tools/kagami/src/main.rs
```diff
@@ -209,7 +210,7 @@ mod docs {
 
     use super::*;
 
-    impl<E: Debug, C: Configurable<Error = E> + Send + Sync + Default> PrintDocs for C {}
+    impl<E: Debug, C: Configurable<Error = E> + Send + Sync + DocsDefault> PrintDocs for C {}
```
ilchu, line 213: This seemed to work as a stopgap when we take away the `#[serde(default)]` on config structs, but we need something more sophisticated here. That would be done at a later stage.

config/base/src/lib.rs
```diff
@@ -256,3 +271,59 @@ pub mod view {
...
+            let file =
+                File::open(&path).wrap_err(format!("Failed to open the config file {:?}", path))?;
+            let reader = BufReader::new(file);
+            serde_json::from_reader(reader)
+                .wrap_err(format!("Failed to deserialize json {:?} from reader", path))
+        }
+
+        // fn finalize(&mut self) -> Result<()> {
```
ilchu, line 314: Left commented code here as it marks an important point: even with the public/private keys being most important non-defaultable config parameters, they probably don't need to be specified in all the overlapping structures. I.e. it could always be enough to specify it either in `torii` or `iroha` config only, and then propagate it into the empty one. Something evolved from this func could be a mechanism to help that.

config/base/derive/src/config.rs
```diff
@@ -0,0 +1,477 @@
...
+                quote! {
+                    #l_value = serde_json::from_value(var.into())
+                        .map_err(|e| iroha_config_base::derive::Error::field_error(stringify!(#ident), e))?
+                }
+            } else if is_option {
+                quote! {
+                    #l_value = Some(serde_json::from_value(var.into())
+                        .map_err(|e| iroha_config_base::derive::Error::field_error(stringify!(#ident), e))?)
+                }
+            }
```
ilchu, line 93: This code path should get run when deriving `Configurable` for our `Proxy` structs.

config/base/derive/src/proxy.rs
```diff
@@ -0,0 +1,217 @@
...
+    field_attrs: &[&[Attribute]],
+) -> proc_macro2::TokenStream {
+    let checked_fields = field_idents
+        .iter()
+        .zip(field_attrs.iter())
+        .map(|(ident, attrs)| {
+            quote! {
+                // #(#attrs)*
+                #ident: self.#ident.ok_or(iroha_config_base::derive::Error::ProxyError(ident.to_string()))?
+            }
```
ilchu, line 212: This is how fields would be checked for presence at build time.

config/base/derive/src/config.rs
```diff
@@ -0,0 +1,477 @@
...
+        .zip(option.iter())
+        .map(|((((set_field, l_value), field_env), &inner_thing), &is_option)| {
+            let inner_thing2 = if inner_thing && is_option {
+                quote! {
+                    // <#inner>::load_environment()?;
+                    #l_value.unwrap().load_environment()?;
+                    // if let Ok(config) = <#inner>::load_environment() {
+                    //     #l_value = Some(config)
+                    // };
+                }
```
ilchu, line 113: An unfinished mess here as the whole `Configurable` trait is not implemented yet for proxy and I wasn't able to test this piece. What gets done to load the inner configuration of fields wrapped inside `Option` still needs to be thought through.

cli/src/lib.rs
```diff
@@ -8,7 +8,8 @@ use std::{panic, path::PathBuf, sync::Arc};
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
 use iroha_actor::{broker::*, prelude::*};
-use iroha_config::iroha::Configuration;
+use iroha_config::base::proxy::DocsDefault;
+use iroha_config::iroha::{Configuration, ConfigurationProxy};
```
appetrosyan, line 12:
```suggestion
use iroha_config::iroha::{Configuration, ConfigurableProxy};
```

tools/kagami/src/main.rs
appetrosyan: Create a ticket, reference in `// TODO`.

config/src/sumeragi.rs
```diff
@@ -53,7 +54,7 @@ view! {
     }
 }
 
-impl Default for Configuration {
+impl DocsDefault for Configuration {
```
appetrosyan, line 57: Not a huge fan of `DocsDefault` for a trait name. Maybe `ReferenceDefault`? `ExampleDefault`?

appetrosyan: Also, the point is that we shouldn't allow people anything resembling a `Default` trait. The point is to prevent this kind of usage, which we can achieve by 
1. enforcing that the default is either implemented only for the proxy structure, and no concrete `Configuration` `impl Default` which we can check. 
2. By encoding the default values in place where they're used. So there isn't a general `Default` trait, just an example struct instance.

config/base/src/lib.rs
```diff
@@ -256,3 +271,59 @@ pub mod view {
...
     }
 }
+
+pub mod proxy {
+    //! Module for configuration proxies' traits
+
+    use super::*;
+
+    /// Pseudo-default trait only used for doc generation
+    pub trait DocsDefault {
```
appetrosyan, line 281: Which is very much the exact same thing as `Default`, can be used in the same way, and people will probably abuse in the same way. 

We need to get rid not only of `core::Default` but also of any trait that can be used like it. Adding a `DocsDefault` trait in the exact same spot doesn't solve anything.

config/base/src/lib.rs
```diff
@@ -256,3 +271,59 @@ pub mod view {
...
+        fn default() -> Self;
+    }
+
+    /// Trait used to convert configs from file and env
+    pub trait Combine: Sized + Serialize + DeserializeOwned {
+        /// Which type of [`Configuration`] it builds into
+        type Target;
+
+        /// Build the config, do the necessary checks
+        fn build(self) -> Result<Self::Target, derive::Error>;
```
appetrosyan, line 291: Need a separate `Builder` trait.

config/base/src/lib.rs
```diff
@@ -256,3 +271,59 @@ pub mod view {
...
+
+        /// If any of the fields in [`other`] are filled, they
+        /// override the values of the fields in [`self`].
+        fn combine(self, other: Self) -> eyre::Result<Self, eyre::Error>;
```
appetrosyan, line 295: Should return `iroha_config::base::proxy::Error`, which enumerates all possible failure modes. Eyre is only useful for quick and dirty front-end stuff, not buried deep inside a lbrary.

config/base/src/lib.rs
appetrosyan: Should be a separate trait. I'd say `LoadFromDisk` and make it a little richer.

config/base/src/lib.rs
appetrosyan: Good point. I'd add this to the front of the PR and to the description of the ticket that fixes it. 

Also consider making it a doc-test.

appetrosyan, line 314:
```suggestion
        // fn finish(&mut self) -> Result<()> {
```

"-ize" in names can confuse users. Rust uses the "-ize" in all of its docs. Our compny policy is to use "-ise". To avoid confusion, use a word that has neither.

config/base/src/lib.rs
```diff
@@ -256,3 +271,59 @@ pub mod view {
...
+        //         iroha_data_model::peer::Id::new(&self.torii.p2p_addr, &self.public_key.clone());
+
+        //     Ok(())
+        // }
+
+        /// Load configuration from the environment
+        ///
+        /// # Errors
+        /// Fails if Configuration deserialization fails (e.g. if `TrustedPeers` contains entries with duplicate public keys)
+        fn load_environment(&mut self) -> core::result::Result<(), derive::Error>;
```
appetrosyan, line 327: Again, separate trait. These traits should be pre-requisites for something to be configurable, not part of the API. 

Reason being that we might want to have stuff that we can load from the ENV, but not have it part of config.

config/src/client.rs
```diff
@@ -137,6 +137,7 @@ impl Configuration {
         AccountId::from_str("alice@wonderland").expect("Account ID not valid")
     }
 
+    // TODO: Delete this?
```
appetrosyan, line 140: Just remember _to do_ it.

config/base/derive/src/config.rs
```diff
@@ -0,0 +1,477 @@
...
+        })
+        .zip(field_environment.iter())
+        .zip(inner.iter())
+        .zip(option.iter())
+        .map(|((((set_field, l_value), field_env), &inner_thing), &is_option)| {
+            let inner_thing2 = if inner_thing && is_option {
+                quote! {
+                    // <#inner>::load_environment()?;
+                    #l_value.unwrap().load_environment()?;
+                    // if let Ok(config) = <#inner>::load_environment() {
```
Erigara, line 110:
```
#l_value = <#inner>::load_enviroment().ok()
```

config/base/derive/src/proxy.rs
```diff
@@ -0,0 +1,217 @@
...
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_quote, Attribute, Data, DataStruct, DeriveInput, Field, Fields, Ident, Token, Type,
+};
+
+use super::{config, utils};
+
+// Take struct with named fields as input
+#[derive(Debug, Clone)]
+struct ProxyInput {
```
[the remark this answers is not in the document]
ilchu: Yeah, that's more or less part of a big refactoring plan. Could greatly improve `Configurable` readability.

config/base/derive/src/config.rs
```diff
@@ -0,0 +1,477 @@
...
+
+use super::utils;
+
+pub(crate) struct EnvPrefix {
+    _ident: Ident,
+    _eq: Token![=],
+    pub(crate) prefix: LitStr,
+}
+
+mod attrs {
```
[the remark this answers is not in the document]
ilchu: Good point, not my code though. Will try to refactor using that.

config/base/derive/src/proxy.rs
```diff
@@ -0,0 +1,217 @@
...
+    field_idents: &[&Ident],
+    field_attrs: &[&[Attribute]],
+    field_ty: &[Type],
+) -> proc_macro2::TokenStream {
+    let combined_fields = field_idents
+        .iter()
+        .zip(field_attrs.iter())
+        .zip(field_ty.iter())
+        .map(|((ident, attrs), ty)| {
+            if utils::is_option_type(ty) {
```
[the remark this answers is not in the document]
ilchu: Actually there is only one case in configs like that, private key in genesis, so the only purpose of this piece is to flatten that. Might need to think of something more elegant for this, true.

config/base/derive/src/proxy.rs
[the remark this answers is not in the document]
Erigara: Great! it's really hard to read right now

config/base/derive/src/proxy.rs
ilchu: Aha, so this is the case against option flattening. So no special handling is needed, just as you mentioned [here](https://github.com/hyperledger/iroha/pull/2583#discussion_r937729839). Scrapping that part for sure then. Thanks for a good catch!

appetrosyan:
I suggest closing this PR in favour of the real PRs.

PR #2590, opened 2022-08-03
[feature] #2511: Restrict FFI types on wasm

### LINKED ISSUE #2511: Reduce primitives type set for Wasm in FFI
Wasm allows only a subset of C ABI compatbile datatypes. The types supported are 32-bit and 64-bit sized types. This means that a flag is needed to convert `u8`/`i8` and `u16`/`i16` to `u32`/`i32` when building FFI API for wasm. This affects integers as function arguments or inside structures as well as `repr(int)` enums. A special care should be taken in handling `bool` type conversion

Erigara:

### Description of the Change

- Add `wasm` feature to `iroha_ffi` crate;
- Add `PrimitiveRepr` trait to substitute unsupported types with supported ones;
- Refactor `Vec`  transfer through ff-boundry.  

### Issue

Closes #2511.

### Benefits

Protect from various errors (when user can pass u32 where u8 supposed to be).

### Possible Drawbacks

More complex implementation.

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -113,7 +113,7 @@ macro_rules! def_ffi_fn {
             handle_id: $crate::handle::Id,
             left_handle_ptr: *const core::ffi::c_void,
             right_handle_ptr: *const core::ffi::c_void,
-            output_ptr: *mut u8,
+            output_ptr: *mut <bool as $crate::IntoFfi>::Target,
         ) -> $crate::FfiReturn {
             $crate::def_ffi_fn!(@catch_unwind {
                 use core::borrow::Borrow;
@@ -152,7 +152,7 @@ macro_rules! def_ffi_fn {
             handle_id: $crate::handle::Id,
             left_handle_ptr: *const core::ffi::c_void,
             right_handle_ptr: *const core::ffi::c_void,
-            output_ptr: *mut i8,
+            output_ptr: *mut <core::cmp::Ordering as $crate::IntoFfi>::Target,
         ) -> $crate::FfiReturn {
             $crate::def_ffi_fn!(@catch_unwind {
                 use core::borrow::Borrow;
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -113,6 +113,7 @@ macro_rules! def_ffi_fn {
             handle_id: $crate::handle::Id,
             left_handle_ptr: *const core::ffi::c_void,
             right_handle_ptr: *const core::ffi::c_void,
+            // Pointer to FFI-safe representation of `bool` (u8 or u32 if `wasm` feature is active)
             output_ptr: *mut <bool as $crate::IntoFfi>::Target,
         ) -> $crate::FfiReturn {
             $crate::def_ffi_fn!(@catch_unwind {
@@ -152,6 +153,7 @@ macro_rules! def_ffi_fn {
             handle_id: $crate::handle::Id,
             left_handle_ptr: *const core::ffi::c_void,
             right_handle_ptr: *const core::ffi::c_void,
+            // Pointer to FFI-safe representation of `Ordering` (i8 or i32 if `wasm` feature is active)
             output_ptr: *mut <core::cmp::Ordering as $crate::IntoFfi>::Target,
         ) -> $crate::FfiReturn {
             $crate::def_ffi_fn!(@catch_unwind {
--- a/ffi/src/primitives.rs
+++ b/ffi/src/primitives.rs
@@ -173,39 +173,47 @@ impl<'itm> IntoFfiSliceRef<'itm> for core::cmp::Ordering {
     }
 }
 
-/// Trait for replacing unsupported types with supported ones when crossing FFI-boundary (for example in case of wasm)
-pub trait PrimitiveRepr {
+/// Trait for replacing unsupported types with supported when crossing WASM FFI-boundary
+#[cfg(feature = "wasm")]
+pub trait WasmRepr {
     /// Type used to represent [`Self`] when crossing FFI-boundry
-    type Repr;
+    type Repr: ReprC;
 }
 
-macro_rules! primitive_repr_impls {
-    ($($source:ty => $target:ty),+ $(,)?) => {$(
-        impl PrimitiveRepr for $source {
-            type Repr = $target;
+#[cfg(feature = "wasm")]
+macro_rules! wasm_repr_impls {
+    ($src:ty = $dst:ty, $($tail:tt)+) => {
+        wasm_repr_impls! { $src = $dst }
+        wasm_repr_impls! { $($tail)+ }
+    };
+    ($src:ty, $($tail:tt)+) => {
+        wasm_repr_impls! { $src }
+        wasm_repr_impls! { $($tail)+ }
+    };
+    ($src:ty = $dst:ty) => {
+        impl WasmRepr for $src {
+            type Repr = $dst;
         }
-    )+};
-}
-macro_rules! primitive_repr_self_impls {
-    ($($source:ty),+ $(,)?) => {
-        primitive_repr_impls! {
-            $($source => $source),*
+    };
+    ($src:ty) => {
+        impl WasmRepr for $src {
+            type Repr = Self;
         }
     };
 }
 
 #[cfg(feature = "wasm")]
-primitive_repr_impls! {u8 => u32, u16 => u32, i8 => i32, i16 => i32}
-#[cfg(not(feature = "wasm"))]
-primitive_repr_self_impls! {u8, u16, i8, i16}
-primitive_repr_self_impls! {u32, u64, u128, i32, i64, i128}
+wasm_repr_impls! {u8 = u32, u16 = u32, i8 = i32, i16 = i32, u32, u64, i32, i64}
 
 macro_rules! primitive_impls {
     ( $( $ty:ty ),+ $(,)? ) => { $(
         unsafe impl ReprC for $ty {}
 
         impl TryFromReprC<'_> for $ty {
-            type Source = <$ty as PrimitiveRepr>::Repr;
+            #[cfg(feature = "wasm")]
+            type Source = <$ty as WasmRepr>::Repr;
+            #[cfg(not(feature = "wasm"))]
+            type Source = Self;
             type Store = ();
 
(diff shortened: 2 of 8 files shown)


ffi/tests/ffi_export.rs
```diff
@@ -285,12 +285,12 @@ fn return_result() {
     unsafe {
         assert_eq!(
             FfiResult::ExecutionFail,
-            FfiStruct__fallible_int_output(u8::from(false), output.as_mut_ptr())
+            FfiStruct__fallible_int_output(From::from(false), output.as_mut_ptr())
```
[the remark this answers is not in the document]
Erigara: When `wasm` feature is active ffi representation of `bool` would be `u32` not `u8`, so I did this so that the test worked in both cases.

ffi/src/primitives.rs
```diff
@@ -169,16 +171,43 @@ impl<'itm> IntoFfiSliceRef<'itm> for core::cmp::Ordering {
     }
 }
 
+/// Trait for replacing unsupported types with supported ones when crossing FFI-boundary (for example in case of wasm)
+pub trait PrimitiveRepr {
+    /// Type used to represent [`Self`] when crossing FFI-boundry
+    type Repr;
```
mversic, line 177:
```suggestion
    type Repr: ReprC;
```

no?

ffi/src/primitives.rs
```diff
@@ -169,16 +171,43 @@ impl<'itm> IntoFfiSliceRef<'itm> for core::cmp::Ordering {
...
+            $($source => $source),*
+        }
+    };
+}
+
+#[cfg(feature = "wasm")]
+primitive_repr_impls! {u8 => u32, u16 => u32, i8 => i32, i16 => i32}
+#[cfg(not(feature = "wasm"))]
+primitive_repr_self_impls! {u8, u16, i8, i16}
+primitive_repr_self_impls! {u32, u64, u128, i32, i64, i128}
```
mversic, line 199: `i128`/`u128` is not ffi safe, I would remove it

ffi/src/primitives.rs
```diff
@@ -169,16 +171,43 @@ impl<'itm> IntoFfiSliceRef<'itm> for core::cmp::Ordering {
...
         unsafe impl ReprC for $ty {}
 
         impl TryFromReprC<'_> for $ty {
-            type Source = Self;
+            type Source = <$ty as PrimitiveRepr>::Repr;
             type Store = ();
 
             unsafe fn try_from_repr_c(source: Self::Source, _: &mut Self::Store) -> Result<Self, FfiResult> {
-                Ok(source)
+                source.try_into().map_err(|_| FfiResult::TrapRepresentation)
```
mversic, line 210: this is not strictly true, it's not a trap representation, it's a conversion error

mversic: I'm not sure if this is worth introducing a new variant for `FfiReturn`

ffi/src/primitives.rs
```diff
@@ -201,10 +230,10 @@ macro_rules! primitive_impls {
         }
 
         impl IntoFfi for $ty {
-            type Target = Self;
+            type Target = <$ty as PrimitiveRepr>::Repr;
```
mversic, line 233:
```suggestion
            #'
            ;
            type Target = <$ty as PrimitiveRepr>::Repr;
```

ffi/src/primitives.rs
```diff
@@ -221,8 +250,29 @@ macro_rules! primitive_impls {
...
+
+        impl<'itm> TryFromReprCVec<'itm> for $ty {
+            type Source = SliceRef<'itm, $ty>;
+            type Store = ();
+
+            unsafe fn try_from_repr_c(
+                source: Self::Source,
+                _: &'itm mut Self::Store,
+            ) -> Result<Vec<Self>, FfiResult> {
+                source.into_rust().ok_or(FfiResult::ArgIsNull).map(std::borrow::ToOwned::to_owned)
```
mversic, line 272:
```suggestion
                source.into_rust().ok_or(FfiResult::ArgIsNull).map(core::borrow::ToOwned::to_owned)
```

ffi/src/primitives.rs
mversic, line 233:
```suggestion
            #[cfg(feature = "wasm")]
            type Target = <$ty as WasmRepr>::Repr;
            #[cfg(not(feature = "wasm"))]
            type Target = Self;
```

ffi/src/primitives.rs
```diff
@@ -169,16 +171,43 @@ impl<'itm> IntoFfiSliceRef<'itm> for core::cmp::Ordering {
...
+primitive_repr_self_impls! {u8, u16, i8, i16}
+primitive_repr_self_impls! {u32, u64, u128, i32, i64, i128}
+
 macro_rules! primitive_impls {
     ( $( $ty:ty ),+ $(,)? ) => { $(
         unsafe impl ReprC for $ty {}
 
         impl TryFromReprC<'_> for $ty {
-            type Source = Self;
+            type Source = <$ty as PrimitiveRepr>::Repr;
```
mversic, line 206:
```suggestion
            #[cfg(feature = "wasm")]
            type Source = <$ty as WasmRepr>::Repr;
            #[cfg(not(feature = "wasm"))]
            type Source = Self;
```

here is what I suggest. `try_into` will work in both cases and compiler should elide `try_into` when `Source = Self`

ffi/src/primitives.rs
mversic: same as previous. `into` will work in both cases

ffi/src/primitives.rs
mversic, line 199: now you can remove the macros if you follow the previous suggestion

ffi/src/primitives.rs
```diff
@@ -169,16 +171,43 @@ impl<'itm> IntoFfiSliceRef<'itm> for core::cmp::Ordering {
     }
 }
 
+/// Trait for replacing unsupported types with supported ones when crossing FFI-boundary (for example in case of wasm)
+pub trait PrimitiveRepr {
```
mversic, line 175:
```suggestion
#[cfg(feature = "wasm")]
pub trait WasmRepr {
```

why not this naming?

mversic (changes requested):

ffi/src/primitives.rs
Erigara: ok, i should remove `u128` from `primitive_impls!` too, right?

mversic: yes

ffi/src/primitives.rs
Erigara: > this is not strictly true, it's not a trap representation, it's a conversion error 

I agree, let's add `FfiResult::ConversionFailed` to avoid confusion.

ffi/src/primitives.rs
Erigara: Oh, i see, you already introduced it in your PR.

mversic: FYI crate needs to be no_std for the time being

ffi/src/primitives.rs
Erigara: I'm not quite understand how your suggestion will help to remove macro.
We still need to implement `WasmRepr` for every supported primitive type, no?
Or you mean that we could remove macro invocation when `wasm` feature is turned off?

ffi/src/primitives.rs
Erigara: Ok, this naming, will be more precise.
Initially i choose `PrimitiveRepr` because i want to minimize places where we have conditional compilation.

ffi/src/handle.rs
```diff
@@ -118,7 +118,7 @@ macro_rules! gen_ffi_impl {
             handle_id: $crate::handle::Id,
             left_handle_ptr: *const core::ffi::c_void,
             right_handle_ptr: *const core::ffi::c_void,
-            output_ptr: *mut i8,
+            output_ptr: *mut <core::cmp::Ordering as $crate::IntoFfi>::Target,
```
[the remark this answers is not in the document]
Erigara: sure

ffi/derive/src/lib.rs
```diff
@@ -299,6 +307,25 @@ fn derive_try_from_ffi_for_opaque_item(name: &Ident) -> TokenStream2 {
...
+
+        impl<'itm> iroha_ffi::owned::TryFromReprCVec<'itm> for #name {
+            type Source = iroha_ffi::slice::SliceRef<'itm, <Self as iroha_ffi::TryFromReprC<'itm>>::Source>;
+            type Store = ();
+
+            unsafe fn try_from_repr_c(
+                source: Self::Source,
+                _: &'itm mut <Self as iroha_ffi::owned::TryFromReprCVec<'itm>>::Store,
+            ) -> Result<Vec<Self>, iroha_ffi::FfiResult> {
+                let slice = source.into_rust().ok_or(iroha_ffi::FfiResult::ArgIsNull)?;
```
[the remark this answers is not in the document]
Erigara: I have added `FfiResult::ConversionFailed`, however for this particular case  `FfiResult::ArgIsNull` is fine, because only case in which `slice.into_rust()` return `None` is when `slice` is `null`.

ffi/src/primitives.rs
mversic: > Or you mean that we could remove macro invocation when wasm feature is turned off?

Yeah, this is what I meant

mversic (approved):

ffi/tests/ffi_export.rs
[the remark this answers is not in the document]
appetrosyan: `u8` for `bool` is problematic. Every operation involving booleans would be extremely slow (which is why it's almost always `size_t`)

ffi/src/primitives.rs
```diff
@@ -221,8 +263,29 @@ macro_rules! primitive_impls {
...
+                source: Self::Source,
+                _: &'itm mut Self::Store,
+            ) -> Result<Vec<Self>> {
+                source.into_rust().ok_or(FfiReturn::ArgIsNull).map(alloc::borrow::ToOwned::to_owned)
+            }
+        }
+    )+};
 }
 
-primitive_impls! {u8, u16, u32, u64, u128, i8, i16, i32, i64}
```
appetrosyan, line 228: So what do we do when we actually have a `u128`?

appetrosyan (approved):

Issue #2158, opened 2022-04-28
Update to Parity Scale codec 3.2.1


appetrosyan:
Updating all of the libraries except the Parity Scale Codec can be done at will. 

The update to the codec needs to be synchronised with all of the SDK developers: 
@arndey @0x009922 and @posplaw

PR #2591, opened 2022-08-04
[feature] #2158: Update `parity_scale_codec`

appetrosyan:

### Description of the Change

Add changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,146 +2,115 @@ Changelog
 =========
 
 
-2.0.0-pre.rc.6-lts
+2.0.0-pre.rc.7
 ---------
 
 Features
 ~~~~~~~~
-- 2395 Add panic if genesis cannot be applied.
-- 2000 Disallow empty names. #2000: Disallow empty names
-- 2127 Add sanity check to ensure that all data decoded by
-  `parity_scale_codec` is consumed.
-- 2360 Make `genesis.json` optional again.
-- 2053 Add tests to all remaining queries in private blockchain.
-- 2381 Unify `Role` registration.
-- 2053 Add tests to the asset-related queries in private blockchain.
-- 2053 Add tests to 'private_blockchain'
-- 2302 Add 'FindTriggersByDomainId' stub-query.
-- 1998 Add filters to queries.
-- 2276 Include current Block hash into BlockHeaderValue.
-- 2161 Handle id and shared FFI fns.
+- hyperledger#2338 Add `cargo-all-features` instrumentation.
+- hyperledger#2564 Kagami tool algorithm options.
+- hyperledger#2490 Implement ffi_export for freestanding functions.
+- hyperledger#1891 Validate trigger execution.
+- hyperledger#1988 Derive macros for Identifiable, Eq, Hash, Ord.
+- hyperledger#2434 FFI bindgen library.
+- hyperledger#2073 Prefer ConstString over String for types in blockchain.
+- hyperledger#1889 Add domain-scoped triggers.
+- hyperledger#2098 Block header queries. #2098: add block header queries
+- hyperledger#2467 Add account grant subcommand into iroha_client_cli.
+- hyperledger#2301 Add transaction's block hash when querying it.
+- hyperledger#2454 Add a build script to the parity-scale-decoder tool.
+- hyperledger#2061 Derive macro for filters.
+- hyperledger#2228 Add Unauthorized variant to smartcontracts query  error.
+- hyperledger#2395 Add panic if genesis cannot be applied.
+- hyperledger#2000 Disallow empty names. #2000: Disallow empty names
+- hyperledger#2127 Add sanity check to ensure that all data decoded by  `parity_scale_codec` is consumed.
+- hyperledger#2360 Make `genesis.json` optional again.
+- hyperledger#2053 Add tests to all remaining queries in private  blockchain.
+- hyperledger#2381 Unify `Role` registration.
+- hyperledger#2053 Add tests to the asset-related queries in private  blockchain.
+- hyperledger#2053 Add tests to 'private_blockchain'
+- hyperledger#2302 Add 'FindTriggersByDomainId' stub-query.
+- hyperledger#1998 Add filters to queries.
+- hyperledger#2276 Include current Block hash into BlockHeaderValue.
+- hyperledger#2161 Handle id and shared FFI fns.
   * add handle id and implement FFI equivalents of shared traits - Clone,Eq,Ord
-- 1638 `configuration` return doc subtree.
-- 2132 Add `endpointN` proc macro.
-- 2257 Revoke<Role> emits RoleRevoked event.
-- 2125 Add FindAssetDefinitionById query.
-- 1926 Add signal handling and graceful shutdown.
-- 2161 generate FFI functions for `data_model`
-- 1149 Block file count does not exceed 1000000 per directory.
-- 1413 Add API version endpoint.
-- 2103 support querying for blocks and transactions. add FindAllTransactions query
-- 2186 Add transfer ISI for `BigQuantity` and `Fixed`.
-- 2056 Add a derive proc macro crate for AssetValueType enum.
-- 2100 Add query to find all accounts with asset.
-- 2179 Optimise trigger execution.
-- 1883 Remove embedded configuration files.
-- 2004 Forbid `isize` and `usize` from becoming `IntoSchema`.
-- 2105 handle query errors in client.
-- 2050 Add role-related queries.
-- #1572: Specialized permission tokens.
-- 2121 Check keypair is valid when constructed.
-- 2099 Add WASM integration test based on Orillion use-case.
-- 2003 Introduce Parity Scale Decoder tool.
-- 1952 Add a TPS benchmark as a standard for optimizations.
-- 2040 Add integration test with transaction execution limit.
-- 1890 Introduce integration tests based on Orillion use-cases.
-- 2048 Add toolchain file.
-- 2100 Add query to find all accounts with asset.
-- 2179 Optimise trigger execution.
-- 1883 Remove embedded configuration files.
-- 2004 Forbid `isize` and `usize` from becoming `IntoSchema`.
-- 2105 handle query errors in client.
-- 2050 Add role-related queries.
-- #1572: Specialized permission tokens.
-- 2121 Check keypair is valid when constructed.
-- 2099 Add WASM integration test based on Orillion use-case.
-- 2003 Introduce Parity Scale Decoder tool.
-- 1952 Add a TPS benchmark as a standard for optimizations.
-- 2040 Add integration test with transaction execution limit.
-- 1890 Introduce integration tests based on Orillion use-cases.
-- 2048 Add toolchain file.
-- 2037 Introduce Pre-commit Triggers.
-- 1621 Introduce By Call Triggers.
-- 1970 Add optional schema endpoint.
-- 1620 Introduce time based triggers.
-- 1918 Implement basic authentication for `client`
-- 1726 Implement a release PR workflow.
-- 1815 Make query responses more type-structured.
-- 1928 implement changelog generation using `gitchangelog`
-- 1902 Bare metal 4-peer setup script.
-- 1619 Introduce event-based triggers.
-- 1195 Close a websocket connection cleanly.
-- 1606 Add ipfs link to domain logo in Domain structure.
-- 1767 restrict linear memory usage for wasm smartcontracts.
-- 1766 Wasm permission validation.
+- hyperledger#1638 `configuration` return doc subtree.
+- hyperledger#2132 Add `endpointN` proc macro.
+- hyperledger#2257 Revoke<Role> emits RoleRevoked event.
+- hyperledger#2125 Add FindAssetDefinitionById query.
+- hyperledger#1926 Add signal handling and graceful shutdown.
+- hyperledger#2161 generate FFI functions for `data_model`
+- hyperledger#1149 Block file count does not exceed 1000000 per directory.
+- hyperledger#1413 Add API version endpoint.
+- hyperledger#2103 support querying for blocks and transactions.  add FindAllTransactions query
+- hyperledger#2186 Add transfer ISI for `BigQuantity` and `Fixed`.
+- hyperledger#2056 Add a derive proc macro crate for AssetValueType  enum.
+- hyperledger#2100 Add query to find all accounts with asset.
+- hyperledger#2179 Optimise trigger execution.
+- hyperledger#1883 Remove embedded configuration files.
+- hyperledger#2004 Forbid `isize` and `usize` from becoming  `IntoSchema`.
+- hyperledger#2105 handle query errors in client.
+- hyperledger#2050 Add role-related queries.
+- hyperledger#1572: Specialized permission tokens.
+- hyperledger#2121 Check keypair is valid when constructed.
+- hyperledger#2099 Add WASM integration test based on Orillion use-case.
+- hyperledger#2003 Introduce Parity Scale Decoder tool.
+- hyperledger#1952 Add a TPS benchmark as a standard for optimizations.
+- hyperledger#2040 Add integration test with transaction execution  limit.
+- hyperledger#1890 Introduce integration tests based on Orillion use-  cases.
+- hyperledger#2048 Add toolchain file.
+- hyperledger#2100 Add query to find all accounts with asset.
+- hyperledger#2179 Optimise trigger execution.
+- hyperledger#1883 Remove embedded configuration files.
+- hyperledger#2004 Forbid `isize` and `usize` from becoming  `IntoSchema`.
+- hyperledger#2105 handle query errors in client.
+- hyperledger#2050 Add role-related queries.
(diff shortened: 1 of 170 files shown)


CHANGELOG.rst
```diff
@@ -2,11 +2,25 @@ Changelog
...
+- 1988 Derive macros for Identifiable, Eq, Hash, Ord. #1988, #2437: Derive macros for Identifiable, Eq, Hash, Ord
+- 2434 FFI bindgen library.
+- 2073 Prefer ConstString over String for types in blockchain. #2073: Prefer ConstString over String for data_model
+- 1889 Add domain-scoped triggers.
+- 2098 Block header queries. #2098: add block header queries
+- 2467 Add account grant subcommand into iroha_client_cli.
+- 2301 Add transaction's block hash when querying it.
+- 2454 Add a build script to the parity-scale-decoder tool.
+- 2061 Derive macro for filters.
+- 2228 Add Unauthorized variant to smartcontracts query error. #2228: Add Unauthorized query error variant
```
outoftardis, line 23:
```suggestion
- 2564 Add Kagami tool algorithm options.
- 2490 Implement ffi_export for freestanding functions.
- 1891 Validate trigger execution.
- 1988, 2437 Derive macros for Identifiable, Eq, Hash, Ord.
- 2434 Add FFI bindgen library.
- 2073 Prefer ConstString over String for types in blockchain.
- 1889 Add domain-scoped triggers.
- 2098 Add block header queries.
- 2467 Add account grant subcommand into iroha_client_cli.
- 2301 Add transaction's block hash when querying it.
- 2454 Add a build script to the parity-scale-decoder tool.
- 2061 Derive macro for filters.
- 2228 Add Unauthorized variant to smartcontracts query error.
```

CHANGELOG.rst
```diff
@@ -154,6 +166,16 @@ Features
...
+- 2422 Hide private keys in configuration endpoint response.
+- Hyperledger#2492: Fixes not all triggers being executed that match an
+  event.
+- 2504 Fix failing tps benchmark.
+- 2477 Fix bug when permissions from roles weren't counted.
+- 2416 Fix lints on macOS arm.
+- 2457 Fix tests flakiness related to shut down on panic. #2457: Add shut down on panic configuration
+- 2473 parse rustc --version instead of RUSTUP_TOOLCHAIN.
+- 1480 Shut down on panic. #1480: Add panic hook to exit program on panic
+- 2376 Simplified Kura, no async, two files.
```
outoftardis, line 178:
```suggestion
- 2422 Hide private keys in configuration endpoint response.
- 2492: Fix not all triggers being executed that match an
  event.
- 2504 Fix failing tps benchmark.
- 2477 Fix bug when permissions from roles weren't counted.
- 2416 Fix lints on macOS arm.
- 2457 Fix tests flakiness related to shutting down on panic.
- 2473 Parse rustc --version instead of RUSTUP_TOOLCHAIN.
- 1480 Add panic hook to exit programs on panic.
- 2376 Simplify Kura: no async, two files.
```

CHANGELOG.rst
```diff
@@ -255,6 +277,14 @@ Fixes
 
 Refactor
 ~~~~~~~~
+- Ensure all feature combinations work.
+- 2468 Remove debug supertrait from permission validators.
+- 2419 Remove explicit `drop`s.
+- 2253 Add `Registrable` trait to `data_model`
+- Impl `Origin` instead of `Identifiable` for the data events.
+ : impl `Origin` instead of `Identifiable` for the data events
```
outoftardis, line 285:
```suggestion
- Use Impl `Origin` instead of `Identifiable` for the data events.
```

CHANGELOG.rst
```diff
@@ -321,6 +351,10 @@ Refactor
 
 Documentation
 ~~~~~~~~~~~~~
+- Flakyness report for Jul 30.
```
outoftardis, line 354:
```suggestion
- Update flakyness report for Jul 30.
```

outoftardis:
Please check older instances in changlelog: they were updated after generation, and now you are changing the phrasing back to the original. 
Since changelog usually requires some minor clean-up, I think it makes sense to generate changelog updates with commit hashes as parameters. This way you can specify which commits to include in this update and not alter the changelog for earlier releases

appetrosyan:
Good point.

pesterev (dismissed):

CHANGELOG.rst
```diff
@@ -2,11 +2,25 @@ Changelog
 =========
 
 
-2.0.0-pre.rc.6-lts
+2.0.0-pre.rc.7
 ---------
 
 Features
 ~~~~~~~~
+- 2338 Add `cargo-all-features` instrumentation.
```
s8sato, line 10: Please consider prefixing a number with `hyperledger#`. It enables linking to issues and PRs

image

appetrosyan:
PSA, this PR is to be squashed, and not rebased.

CHANGELOG.rst
appetrosyan: Done that. We'll nee to be a bit careful with older changelog entries.

cli/src/lib.rs
```diff
@@ -333,6 +333,8 @@ where
         Ok(false)
     }
 
+    // Which raises the question: does it make sense to enable `nursery` lints?
```
mversic, line 336: should we write a TODO here or remove the comment? Personally, I wouldn't enable nursery lints

actor/src/deadlock.rs
```diff
@@ -1,6 +1,10 @@
 //! Module which is used for deadlock detection of actors
 
-#![allow(clippy::expect_used, clippy::panic)]
+#![allow(
+    clippy::expect_used,
+    clippy::panic,
+    clippy::declare_interior_mutable_const
```
mversic, line 6: may I ask what activates this lint?

cli/src/torii/mod.rs
```diff
@@ -137,7 +142,7 @@ impl Error {
         let mut idx = 0_i32;
         let mut err_opt = Some(err);
         while let Some(e) = err_opt {
-            s += &format!("    {}: {}\n", idx, &e.to_string());
+            write!(s, "    {idx}: {}", &e.to_string()).expect("Valid");
```
mversic, line 145: it doesn't ring true to me that this write would necessarily always succeed

ffi/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(unsafe_code)]
+#![allow(unsafe_code, clippy::extra_unused_lifetimes)]
```
mversic, line 1: finally we got this lint, I was always surprised this is valid rust code or at least not deny by default lint

mversic: you could have fixed this directly in the code instead of allowing the lint, there is only one occurrence triggering this lint in the `impl_tuple`. It's ok to remove the lifetime declaration

ffi/src/lib.rs
```diff
@@ -161,20 +159,15 @@ impl<'itm, T: ReprC> TryFromReprC<'itm> for &'itm mut T {
     }
 }
 
-impl<T: ReprC + Copy> IntoFfi for &T
-where
-    T: IntoFfi<Target = T>,
-{
+impl<T: ReprC + Copy + IntoFfi<Target = T>> IntoFfi for &T {
```
mversic, line 162: ok, let's say defining bounds in one place is more readable. Should we strive to place bounds in the `where` clause or type declaration or as the case requires? I've seen both approaches in this PR. Not that you have to change this now

ffi/src/owned.rs
```diff
@@ -1,5 +1,6 @@
 //! Logic related to the conversion of structures with ownership. Ownership is never transferred
 //! across FFI. This means that contents of these structures are copied into provided containers
+#![allow(clippy::extra_unused_lifetimes)]
```
mversic, line 3: you can remove the lifetime declaration in the code, and not allow this lint

p2p/Cargo.toml
```diff
@@ -1,4 +1,4 @@
-[package]
+    [package]
```
mversic, line 1: seems like bad formatting

lints.toml
```diff
@@ -1,4 +1,4 @@
-# For all clippy lints please visit: https://rust-lang.github.io/rust-clippy/master/
+    # For all clippy lints please visit: https://rust-lang.github.io/rust-clippy/master/
```
mversic, line 1: seems like bad formatting

mversic (changes requested):

p2p/src/handshake.rs
```diff
@@ -0,0 +1,309 @@
...
+        K: KeyExchangeScheme + Send + 'static,
+        E: Encryptor + Send + 'static;
+
+    /// Peer in disconnected state.
+    #[derive(Debug)]
+    pub struct Disconnected(peer::Id);
+
+    /// Peer in broken state.
+    #[derive(Debug)]
+    pub struct Broken(peer::Id, crate::Error);
```
mversic, line 292: should I care to mention that you're implementing `Identifiable` by hand? You could have done this and have it done for you:

```suggestion
    pub struct Broken(peer, crate::Error);
```

actor/src/deadlock.rs
appetrosyan: Essentially, anyhting that defines a const with interior mutability. Normally I'd say fix it. But in this particular case it comes from code generated by a macro.

cli/src/torii/mod.rs
appetrosyan: In each case in which it would reasonably fail (OOM, non-UTF characters in Display), panicking is sensible. Also, unlike the previous code, this is explicitly causing an error.

ffi/src/lib.rs
appetrosyan: This lint has had a lot of false-positives, which is why it wasn't enabled by default, and why I disabled it here.

ffi/src/lib.rs
appetrosyan: In this case, visually. Here having them all in one place is still under 80 columns.

p2p/src/handshake.rs
appetrosyan: Temporarily. As you see from the draft status this PR is WIP

mversic: ah, pardon me

ffi/src/lib.rs
appetrosyan: Speaking of which, we should consider this for our contributing guide and style guide

cli/src/lib.rs
appetrosyan: Created issue.

p2p/src/peer.rs
```diff
@@ -29,39 +32,38 @@ use tokio::{
...
-    Error, Message, MessageResult,
+    CryptographicError, Error, HandshakeError, Message, MessageResult,
 };
 
 /// Max message length in bytes.
-const MAX_MESSAGE_LENGTH: usize = 16 * 1024 * 1024;
-const MAX_HANDSHAKE_LENGTH: usize = 255;
+pub const MAX_MESSAGE_LENGTH: usize = 16 * 1024 * 1024;
+/// Max length of message in bytes.
+pub const MAX_HANDSHAKE_LENGTH: usize = 255;
```
ilchu, line 41: Shouldn't this say "length of handshake"?

p2p/src/handshake.rs
ilchu: Do we need an issue for this?

p2p/src/peer.rs
appetrosyan: Not really. The handshake length could be shorter.

version/derive/src/lib.rs
```diff
@@ -2,7 +2,9 @@
     clippy::module_name_repetitions,
     missing_docs,
     clippy::shadow_reuse,
-    clippy::str_to_string
+    clippy::str_to_string,
+    clippy::arithmetic,
+    clippy::std_instead_of_core,
```
mversic, line 7: I think we should disable this lint. I don't see what it brings us. If we use `no_std` then the crate will not compile and this is covered in CI

p2p/src/lib.rs
```diff
@@ -11,60 +11,80 @@ pub use network::{ConnectPeer, DisconnectPeer, NetworkBase, Post};
...
 /// The main type to use for secure communication.
 pub type Network<T> = NetworkBase<T, X25519Sha256, ChaCha20Poly1305>;
 
 /// Errors used in [`crate`].
-#[derive(Clone, Debug, iroha_macro::FromVariant, Error, iroha_actor::Message)]
+#[derive(Debug, Error, iroha_actor::Message)]
 pub enum Error {
     /// Failed to read or write
     #[error("Failed IO operation.")]
     Io(#[source] std::sync::Arc<io::Error>),
```
mversic, line 26: out of scope here, but it feels weird to return `Arc` in an error

mversic (changes requested):

ilchu (dismissed):

version/derive/src/lib.rs
appetrosyan: If I add that to `lints.toml` it will stop `cargo lints` from working. The lint is present in the compatible nightly, but not in  the latest stable.

core/src/sumeragi/message.rs
```diff
@@ -1,5 +1,7 @@
 //! Contains message structures for p2p communication during consensus.
 #![allow(clippy::module_name_repetitions)]
+#![allow(clippy::arithmetic, clippy::std_instead_of_core, clippy::std_instead_of_alloc)]
+#![allow(clippy::significant_drop_in_scrutinee)]
```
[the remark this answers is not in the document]
appetrosyan: 1. `Significant drop` is just a _notification_ that somebody thought was a good idea to include in `restriction` rather than `nursery`. We might want  to revisit it once we have pruned the lints, but for now the benefits are not worth the effort. 

2. `arithmetic` is a lint that I intend to enable globally, once `cargo lints` is updated to ignore unknown lints. We already use `checked_add` and so forth in cases where it matters, everything else is going to add overhead and make the API more cumbersome. 

3. `std_instead_of_core` and `std_instead_of_alloc` as you said should be disabled globally. Using `core` is just a matter of changing the imports, while using `alloc` is much more involved, and in this case, completely unnecessary. While it's possible that our downstream users might be using Iroha in an embedded context I doubt that `no_std` can help if even a modest blockchain takes up Gigabytes of RAM.

core/src/sumeragi/fault.rs
```diff
@@ -373,7 +373,7 @@ impl<G: GenesisNetworkTrait, F: FaultInjection> SumeragiWithFault<G, F> {
 
     /// Updates network topology by taking the actual list of peers from `WorldStateView`.
     /// Updates it only if there is a change in WSV peers, otherwise leaves the order unchanged.
-    #[allow(clippy::expect_used)]
+    #[allow(clippy::expect_used, clippy::unused_async)]
```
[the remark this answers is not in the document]
appetrosyan: The function is `async` if it's annotated with `async`. Given that you need them to be `async` for API purposes, this is an instance where gentrifying over `async`-ness would be beneficial, both in trait definitions and in code.

core/src/kura.rs
```diff
@@ -2,7 +2,13 @@
 //! logic.  [`Kura`] is the main entity which should be used to store
 //! new [`Block`](`crate::block::VersionedCommittedBlock`)s on the
 //! blockchain.
-
+#![allow(
+    clippy::module_name_repetitions,
+    clippy::std_instead_of_core,
+    clippy::std_instead_of_alloc,
+    clippy::arithmetic,
+    clippy::significant_drop_in_scrutinee
```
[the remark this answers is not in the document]
appetrosyan: `significant_drop_in_scrutinee` is at best a notification and only useful for people who don't know the lifetimes of objects inside a match statement. We could fix this lint, but only in a separate PR, and this is a low priority.

p2p/src/lib.rs
appetrosyan: This is a bit of a hacky workaround for bad API decisions in `std`. `io::Error` holds a `dyn Error` object, meaning that you can't clone it (even though in theory you should be able to). Given that we need to store the error that caused the `Peer<T, K, E>` to enter the broken state, and raise an error up a call stack, this was the least painful way to handle the situation. 

The real solution is to convert an `std::io::Error` into a `crate::IoError` enum. As an added bonus, we would have more specific error messages, with instructions on how to fix them.

p2p/src/handshake.rs
appetrosyan: #2620

core/src/sumeragi/message.rs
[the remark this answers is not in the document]
mversic: > While it's possible that our downstream users might be using Iroha in an embedded context I doubt that no_std can help if even a modest blockchain takes up Gigabytes of RAM.

No way they would host blockchain on an embedded device, but they might want to run `Iroha` client on embedded device at some point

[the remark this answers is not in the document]
appetrosyan: >  Iroha client on embedded device at some point

Certainly. At that point we should consider reducing the memory footprint by other means as well.

mversic (approved):

ilchu (approved):

mversic (approved):

PR #2601, opened 2022-08-09
[feature] #2488: Add support for trait impls in `ffi_export`

### LINKED ISSUE #2488: support `ffi_export` on trait implementations
`ffi_export` attribute can be attached on inherent impls. This is not a hard limitation, this attribute could be made to support trait impls as well. This is not a very difficult thing to do but requires the macro library to support parsing of associated types.

If you take a look at `data_model::name` you can see that `Name__from_str` is implemented manually and is essentially a wrapper around `FromStr` implementation for `Name`. Impls such as this could be generated with `ffi_export` as well like this:

```rs
#[ffi_export]
impl FromStr for Name {
    type Err = ParseError;

    fn from_str(candidate: &str) -> Result<Self, Self::Err> {
        Self::validate_str(candidate).map(|_| Self(ConstString::from(candidate)))
    }
}
```

which would generate, once macro is expanded, something like:

```rs
extern fn Name__from_str(candidate: <<&str> as TryFromReprC>::Source) -> iroha_ffi::FfiResult {
    // ... generated code ...
}
```

Erigara:
### Description of the Change

- Add support for trait impls (methods are generated with following name `Type__Trait__method`)
- Remove requirements for items to have doc string

### Issue

Closes #2488.

### Benefits

No need to manually implement trait methods. 

### Possible Drawbacks

None.

```
cargo +nightly-2022-04-20 miri test --package iroha_ffi
```


data_model/src/name.rs
```diff
@@ -207,7 +168,7 @@ mod tests {
 
             assert_eq!(
                 iroha_ffi::FfiReturn::Ok,
-                Name__from_str(candidate.into_ffi(), name.as_mut_ptr())
+                Name__FromStr__from_str(candidate.into_ffi(), name.as_mut_ptr())
```
mversic, line 171: this test is now useless, you can remove it altogether

ffi/tests/ffi_export.rs
```diff
@@ -296,3 +311,20 @@ fn return_result() {
         assert_eq!(42, output.assume_init());
     }
 }
+
+#[test]
+fn invoke_trait_method() {
+    let ffi_struct = get_new_struct_with_params();
+    let mut output = MaybeUninit::uninit();
```
mversic, line 318:
```suggestion
    let mut output = MaybeUninit::<Name>::new();
```

Please try to use exact types where ever possible so that we are sure this works correctly. If you can use exact types in other types as well. Initially, I did this is `valid.rs` but that file will most likely be removed because it's just a duplicate of integration tests.

Also, don't use `uninit` in tests, instead use `new`. This is just a precaution to ensure that if the feature tested is not implemented correctly it does not introduce UB because UB could make the test pass

ffi/derive/src/impl_visitor.rs
```diff
@@ -287,10 +303,6 @@ impl<'ast> Visit<'ast> for FnVisitor<'ast> {
     fn visit_impl_item_method(&mut self, node: &'ast syn::ImplItemMethod) {
         self.doc = find_doc_attr(&node.attrs).cloned();
 
-        if !matches!(node.vis, syn::Visibility::Public(_)) {
```
mversic, line 290: can we keep the check for associated methods? or do you think it's ok to allow export of private methods?

mversic (changes requested):

ffi/derive/src/ffi_fn.rs
```diff
@@ -61,11 +61,17 @@ pub fn gen_definition(fn_descriptor: &FnDescriptor) -> TokenStream {
...
+    let self_ty_name = fn_descriptor
+        .self_ty_name()
+        .map_or_else(Default::default, ToString::to_string);
+    let trait_name = fn_descriptor
+        .trait_name()
+        .map_or_else(Default::default, |trait_name| format!("__{trait_name}"));
 
     Ident::new(
-        &format!("{}__{}", self_ty_name, method_name),
+        &format!("{}{}__{}", self_ty_name, trait_name, method_name),
```
mversic, line 74:
```suggestion
        &format!("{}__{}__{}", self_ty_name, trait_name, method_name),
```

aren't you missing underscores?

ffi/derive/src/ffi_fn.rs
```diff
@@ -189,13 +195,19 @@ fn gen_output_assignment_stmts(fn_descriptor: &FnDescriptor) -> TokenStream {
             let src_type = receiver.src_type();
 
             if matches!(src_type, syn::Type::Path(_)) {
-                return quote! {
-                    if __out_ptr.is_null() {
-                        return Err(iroha_ffi::FfiReturn::ArgIsNull);
+                match output_arg.src_type() {
+                    // NOTE: case like fn(mut self, ...) -> Self (builder pattern)
+                    syn::Type::Path(path) if path.path.is_ident("Self") => {
```
mversic, line 200: I'm not sure, but this looks like you should be using `SelfResolver`. Also, what if the type is used as input argument?

ffi/derive/src/impl_visitor.rs
```diff
@@ -154,13 +169,9 @@ impl FnDescriptor {
     pub fn self_ty_name(&self) -> Option<&Ident> {
         self.self_ty.as_ref().map(get_ident)
     }
-}
-impl From<&syn::ItemFn> for FnDescriptor {
```
mversic, line 158: thank you, I like it much better now

ffi/derive/src/impl_visitor.rs
```diff
@@ -131,18 +134,30 @@ impl<'ast> ImplDescriptor<'ast> {
...
-        let mut visitor = FnVisitor::new(Some(self_ty));
+    pub(crate) fn from_impl_method(
+        self_ty: &syn::Path,
+        trait_name: Option<&syn::Path>,
+        node: &syn::ImplItemMethod,
+    ) -> Self {
+        let mut visitor = FnVisitor::new(Some(self_ty), trait_name);
 
         visitor.visit_impl_item_method(node);
         FnDescriptor::from_visitor(visitor)
```
mversic, line 145:
```suggestion
        Self::from_visitor(visitor)
```

ffi/tests/ffi_export.rs
Erigara: > does not introduce UB because UB could make the test pass
good point, thanks

ffi/derive/src/ffi_fn.rs
Erigara: > Also, what if the type is used as input argument?

Outer `if` is responsible for handling such case (match `self` and `mut self`).

```rust
if matches!(src_type, syn::Type::Path(_))
```

This change is not strictly relevant for the adding support for trait impls, but i find out that we currently can't convert to ffi functions like `fn(self) -> T`, so this change fix this.

ffi/derive/src/impl_visitor.rs
Erigara: You're right, we should keep this check.

data_model/src/name.rs
```diff
@@ -88,6 +90,7 @@ impl AsRef<str> for Name {
     }
 }
 
+#[cfg_attr(all(feature = "ffi_export", not(feature = "ffi_import")), ffi_export)]
```
mversic, line 93: If possible I would encourage you to add the `ffi_import` as well. The issue wasn't created with it in mind because `ffi_import` didn't exist at that time

ffi/derive/src/ffi_fn.rs
mversic: you probably mean methods, functions cannot have self receiver. So what about [this](https://github.com/hyperledger/iroha/blob/iroha2-dev/ffi/tests/ffi_export.rs#L50) test or [this](https://github.com/hyperledger/iroha/blob/iroha2-dev/ffi/tests/ffi_export.rs#L57)?

mversic: ah, I think I get you now. Please follow up with a test case for this

mversic: yes, if self is not returned from the ffi function memory will be leaked because in the generated body of the function `core::ptr::read` is used. PM me if you need more detailed explanation

Erigara: > I'm not sure, but this looks like you should be using SelfResolver
 
i think you are right, current solution would fail if instead of `fn(self) -> Self` someone would write `fn(self) -> Type` where `Self == Type`.

ffi/tests/ffi_export.rs
mversic: > I would try to use exact types where ever possible so that we are sure this works correctly.

This advice was ill given actually because you'll always be comparing to some actual value in `assert_eq`

Erigara: > > I would try to use exact types where ever possible so that we are sure this works correctly.
> 
> This advice was ill given actually because you'll always be comparing to some actual value in `assert_eq`

You want to rollback this change?

mversic: No, I don't care

ffi/derive/src/ffi_fn.rs
```diff
@@ -61,21 +49,42 @@ pub fn gen_definition(fn_descriptor: &FnDescriptor) -> TokenStream {
...
+        .as_ref()
+        .and_then(syn::Path::get_ident);
 
     let path = self_type.map_or_else(
         || method_name.to_string(),
-        |self_ty| format!("{}::{}", self_ty, method_name),
+        |self_ty| {
+            trait_name.map_or_else(
+                || format!("{}::{}", self_ty, method_name),
+                // NOTE: Fully-qualified syntax currently not supported
```
mversic, line 84: you mean like `<Self as Trait>::method_name`?. I think we don't have to support that for the docs

ffi/derive/src/ffi_fn.rs
```diff
@@ -153,20 +164,29 @@ fn gen_input_conversion_stmts(fn_descriptor: &FnDescriptor) -> TokenStream {
 fn gen_method_call_stmt(fn_descriptor: &FnDescriptor) -> TokenStream {
     let ident = &fn_descriptor.sig.ident;
     let self_type = &fn_descriptor.self_ty;
+    let trait_name = &fn_descriptor.trait_name;
 
     let receiver = fn_descriptor.receiver.as_ref();
     let self_arg_name = receiver.map_or_else(Vec::new, |arg| {
-        if matches!(arg.src_type(), syn::Type::Path(_)) {
+        if matches!(arg.src_type(), syn::Type::Path(_))
+            && Some(arg.name()) == fn_descriptor.output_arg.as_ref().map(Arg::name)
```
mversic, line 172:
```suggestion
        if Some(arg.name()) == fn_descriptor.output_arg.as_ref().map(Arg::name)
```

is this sufficient? It should not be possible to have two types with the same name, unless it's a special case for `fn(self) -> Self`.

ffi/derive/src/ffi_fn.rs
```diff
@@ -185,10 +205,11 @@ fn gen_method_call_stmt(fn_descriptor: &FnDescriptor) -> TokenStream {
 fn gen_output_assignment_stmts(fn_descriptor: &FnDescriptor) -> TokenStream {
     if let Some(output_arg) = &fn_descriptor.output_arg {
         if let Some(receiver) = &fn_descriptor.receiver {
-            let arg_name = receiver.name();
-            let src_type = receiver.src_type();
-
-            if matches!(src_type, syn::Type::Path(_)) {
+            // NOTE: case like fn(self, ...) -> Self (builder pattern)
+            if matches!(receiver.src_type(), syn::Type::Path(_))
+                && receiver.name() == output_arg.name()
```
mversic, line 210:
```suggestion
            if receiver.name() == output_arg.name()
```

again, this should be sufficient. Would you extract it into a function and add a description, this is very succint. Both here and earlier

mversic: maybe you can take out this whole if nesting into a function

mversic (changes requested):

ffi/tests/ffi_export.rs
```diff
@@ -54,7 +54,7 @@ impl FfiStruct {
 
     /// With params
     #[must_use]
-    pub fn with_params(mut self, params: impl IntoIterator<Item = (Name, Value)>) -> Self {
+    pub fn with_params(mut self, params: impl IntoIterator<Item = (Name, Value)>) -> FfiStruct {
```
mversic, line 57:
```suggestion
    pub fn with_params(mut self, params: impl IntoIterator<Item = (Name, Value)>) -> Self {
```

why the change? I don't mind if it doesn't affect anything

ffi/tests/unambiguous.rs
```diff
@@ -0,0 +1,77 @@
...
+/// FfiStruct
+#[derive(Clone, Copy, IntoFfi, TryFromReprC)]
+#[ffi_export]
+pub struct FfiStruct {}
+
+#[ffi_export]
+impl FfiStruct {
+    /// Ambiguous method
+    pub fn ambiguous() -> Ambiguous {
+        Ambiguous::Inherit
```
mversic, line 25:
```suggestion
        Ambiguous::Inherent
```

mversic: otherwise, this is an awesome test

ffi/tests/ffi_export.rs
Erigara: @mversic, i added this change to showcase that such style of writing signature are supported now

ffi/derive/src/ffi_fn.rs
Erigara: Yeah, there is [issue](https://github.com/rust-lang/rust/issues/74563) for that.
For now i stick with link to `Trait::method_name`.

ffi/derive/src/ffi_fn.rs
Erigara: Yes, good point, `receiver` and `output_arg` will have same name only in this case.

ffi/derive/src/ffi_fn.rs
Erigara: @mversic, fixed, rewrite function a bit to avoid nesting

mversic (dismissed):

ffi/derive/src/ffi_fn.rs
```diff
@@ -61,21 +49,42 @@ pub fn gen_definition(fn_descriptor: &FnDescriptor) -> TokenStream {
...
+    let self_ty_name = fn_descriptor
+        .self_ty_name()
+        .map_or_else(Default::default, ToString::to_string);
+    let trait_name = fn_descriptor
+        .trait_name()
+        .map_or_else(Default::default, |trait_name| format!("__{trait_name}"));
 
     Ident::new(
-        &format!("{}__{}", self_ty_name, method_name),
+        &format!("{}{}__{}", self_ty_name, trait_name, method_name),
```
appetrosyan, line 62:
```suggestion
        &format!("{}_{}__{}", self_ty_name, trait_name, method_name),
```
Imagine `NameFromStr__from_str` vs `Name_FromStr__from_str`

ffi/derive/src/ffi_fn.rs
appetrosyan: @Erigara The underscores aren't missing, but it's very difficult to understand that. Please format the `trait_name` without the underscores, and use the underscores uniformly in the final format statement.

ffi/derive/src/ffi_fn.rs
appetrosyan: Low priority.

ffi/derive/src/ffi_fn.rs
```diff
@@ -183,30 +198,31 @@ fn gen_method_call_stmt(fn_descriptor: &FnDescriptor) -> TokenStream {
...
+                #output_arg_conversion
+                iroha_ffi::OutPtrOf::<#arg_type>::write(__out_ptr, #arg_name)?;
+            }
+        }
+        None => quote! {},
     }
+}
 
-    quote! {}
+/// Check if fn signature is like `fn(self, ...) -> Self`
```
appetrosyan, line 224: Honestly, this is a rare case of a function being named so-well that this doc is redundant.

ffi/derive/src/impl_visitor.rs
```diff
@@ -66,13 +59,21 @@ fn resolve_src_type(self_type: Option<&syn::Path>, mut arg_type: Type) -> Type {
 
     arg_type
 }
+pub struct ImplDescriptor<'ast> {
```
appetrosyan, line 62: Add newline before, please

ffi/derive/src/impl_visitor.rs
```diff
@@ -131,18 +134,30 @@ impl<'ast> ImplDescriptor<'ast> {
 }
 
 impl FnDescriptor {
-    fn from_impl_method(self_ty: &syn::Path, node: &syn::ImplItemMethod) -> Self {
-        let mut visitor = FnVisitor::new(Some(self_ty));
+    pub(crate) fn from_impl_method(
```
appetrosyan, line 137:
```suggestion
    pub fn from_impl_method(
```

**Explanation:** This function is defined in a private module in a proc-macro crate. If you wanted to import it, the only place where you would be able to do so is yet another proc-macro crate.

ffi/derive/src/impl_visitor.rs
```diff
@@ -131,18 +134,30 @@ impl<'ast> ImplDescriptor<'ast> {
...
+        trait_name: Option<&syn::Path>,
+        node: &syn::ImplItemMethod,
+    ) -> Self {
+        let mut visitor = FnVisitor::new(Some(self_ty), trait_name);
 
         visitor.visit_impl_item_method(node);
         FnDescriptor::from_visitor(visitor)
     }
 
+    pub(crate) fn from_fn(node: &syn::ItemFn) -> Self {
```
appetrosyan, line 148:
```suggestion
    pub fn from_fn(node: &syn::ItemFn) -> Self {
```
Same as above.

ffi/derive/src/impl_visitor.rs
```diff
@@ -131,18 +134,30 @@ impl<'ast> ImplDescriptor<'ast> {
...
+        Self::from_visitor(visitor)
+    }
+
     fn from_visitor(visitor: FnVisitor) -> Self {
         Self {
             self_ty: visitor.self_ty.map(Clone::clone),
+            trait_name: visitor.trait_name.map(Clone::clone),
 
-            doc: visitor.doc.expect_or_abort("Missing doc"),
+            doc: visitor.doc,
```
appetrosyan, line 160: We need a way to handle docs properly. Most trait implementations don't have doc-comments, true, but they need to inherit the docs from the trait. Since we carried the docs for ordinary functions through, I suggest doing this in another ticket.

ffi/derive/src/impl_visitor.rs
```diff
@@ -188,9 +199,13 @@ impl<'ast> ImplVisitor<'ast> {
 }
 
 impl<'ast> FnVisitor<'ast> {
-    pub const fn new(self_ty: Option<&'ast syn::Path>) -> Self {
+    pub const fn new(
+        self_ty: Option<&'ast syn::Path>,
```
appetrosyan, line 203: Prefer passing the value directly, unwrap options before use.

ffi/derive/src/impl_visitor.rs
```diff
@@ -287,8 +295,12 @@ impl<'ast> Visit<'ast> for FnVisitor<'ast> {
     fn visit_impl_item_method(&mut self, node: &'ast syn::ImplItemMethod) {
         self.doc = find_doc_attr(&node.attrs).cloned();
 
-        if !matches!(node.vis, syn::Visibility::Public(_)) {
-            abort!(node.vis, "Methods defined in the impl block must be public");
+        // NOTE: only public method in inherit impls are allowed
+        if self.trait_name.is_none() && !matches!(node.vis, syn::Visibility::Public(_)) {
+            abort!(
+                node.vis,
+                "Methods defined in the inherit impl block must be public"
```
appetrosyan, line 302:
```suggestion
                "Methods defined in the inherent `impl` block must be public"
```

On another note, I'd actually prefer if you just skipped them, instead of failing to compile. Notify via https://docs.rs/proc-macro-error/latest/proc_macro_error/macro.emit_warning.html

ffi/derive/src/lib.rs
```diff
@@ -173,7 +173,7 @@ pub fn ffi_export(_attr: TokenStream, item: TokenStream) -> TokenStream {
                 abort!(item.sig.generics, "Generics are not supported");
             }
 
-            let fn_descriptor = FnDescriptor::from(&item);
+            let fn_descriptor = FnDescriptor::from_fn(&item);
```
appetrosyan, line 176:
```suggestion
            let fn_descriptor = FnDescriptor::fn_from(&item);
```

`from_fn` => From Function. 
`fn_from` => The function `from`.

ffi/derive/tests/ui_fail/private_method_inside_impl.rs
```diff
@@ -0,0 +1,19 @@
+use std::alloc::alloc;
```
appetrosyan, line 1: Like I said previously, just skip private functions and (optionally) warn about them. I'm quite sure that `cargo lints` turns all warnings into errors, so notifying of private associated functions might not be possible.

ffi/tests/ffi_export.rs
appetrosyan: Add comment in code?

ffi/derive/src/impl_visitor.rs
Erigara: Thank you for explanation!

ffi/derive/src/lib.rs
Erigara: In this case we want `from_fn`, because we are creating `FnDescriptor` from freestanding function, likewise we have `from_impl_method` when we creating descriptor from method inside `impl` block.

ffi/derive/src/impl_visitor.rs
Erigara: We don't carry docs from functions, we replace docs to something like: 
```rust
/// FFI function equivalent of [`Struct::method`] (or [`Trait::method`])
```

IMO having link to original method is enough in this case.

My suggestion is that @mversic add docs here to later use them when generating opaque wrapper structs.

ffi/derive/src/impl_visitor.rs
Erigara: I tried your suggestion and i think resulting code is less convenient.
In case of freestanding functions `self_ty` is `None` and this is intended.

ffi/derive/src/ffi_fn.rs
Erigara: Format uniformly would require to introduce branching, because method name is `Struct[__Trait]__method`.
Another option would be to use `"{}{}{}"` as template and underscores would be added before where needed.

ffi/derive/src/impl_visitor.rs
Erigara: We are `abort`ing execution all over the place (i mean `ffi_export/import` macros) where we don't support such item. 
What benefits will we get if we skip private functions?

mversic (approved):

ffi/derive/src/ffi_fn.rs
appetrosyan: Let's go with `{}{}{}`.

ffi/derive/src/impl_visitor.rs
appetrosyan: Basically, you wouldn't force the user to add the function elsewhere. A private function is perfectly valid code, while it shouldn't be reflected in the ffi export, forcing people to have multiple inherent impls is not good design.

ffi/derive/src/lib.rs
appetrosyan: In that case `desc_from_fn`?

ffi/derive/src/impl_visitor.rs
Erigara: Ok, in this case we should make `impl` blocks more permissive in general, skip consts, types, ... instead of aborting execution.

ffi/derive/src/lib.rs
Erigara: I don't mind but don't full invocation `FnDescriptor::from_fn` is descriptive enough?

ffi/derive/src/impl_visitor.rs
appetrosyan: exactly.

Erigara: Now private functions are skipped, however i left this abortion, to indicate that something going wrong if we try to export private function.

appetrosyan (approved):

ffi/derive/src/impl_visitor.rs
```diff
@@ -261,34 +262,39 @@ impl<'ast> Visit<'ast> for ImplVisitor<'ast> {
...
     fn visit_impl_item_method(&mut self, node: &'ast syn::ImplItemMethod) {
         self.doc = find_doc_attr(&node.attrs).cloned();
 
-        if !matches!(node.vis, syn::Visibility::Public(_)) {
-            abort!(node.vis, "Methods defined in the impl block must be public");
+        if self.trait_name.is_none() && !matches!(node.vis, syn::Visibility::Public(_)) {
+            abort!(
+                node.vis,
+                "Private methods defined in an inherent `impl` block should not be exported, this is a bug in the library",
+            );
```
mversic, line 297: shouldn't this be removed now?

mversic: or is this a freestanding function and therefore it makes no sense to attach `ffi_export` on it which this message is about

mversic (approved):

Issue #2544, opened 2022-07-26
Doctests and CI to add them directly to docs repo as examples


appetrosyan:
We want to keep a set of tests which specifically showcase basic operations and types. 

We want to include them as examples in the in the docs repo and whenever there are changes to these examples, we want to automatically create a PR into the docs repo. 

### Benefits

Automagical maintenance of docs in an up-to-date state. 

We will know that the code snippets in docs are up to date, because they have been tested in our repo when they were introduced and passed CI checks. 

### Drawbacks 

Time consuming and needs to be done for all major supported SDKs. 
- [x] Rust
- [x] Java
- [x] Kotlin
- [x] JavaScript
- [x] Python

s8sato:
As for Rust, #2309 needs to be resolved

appetrosyan:
@timofeevmd Please verify that the tests are all running in the docs repo and if not, create relevant issues, before closing this one.

appetrosyan:
@timofeevmd asked to clarify:

We need to point to the documentation files. We need to ensure that every example that **can** be moved into a working piece of checked code in the repos is extracted into a separate test, even a doctest. 

This ticket is not about updating the docs, it's about starting issues for when tests need to be written and the docs need to be updated.

timofeevmd:
I checked the documentation for our SDKs, below is a list of tasks

 Rust https://github.com/hyperledger/iroha-2-docs/issues/361
 Java/Kotlin   https://github.com/hyperledger/iroha-2-docs/issues/364
 JavaScript https://github.com/hyperledger/iroha-2-docs/issues/363
 Python https://github.com/hyperledger/iroha-2-docs/issues/362

PR #2603, opened 2022-08-09
[documentation] #2544: Tutorial doctests 

appetrosyan:
Signed-off-by: 6r1d <vic.6r1d@gmail.com>

### Description of the Change

Adds example that mirrors tutorial examples, so that changes in the API can be propagated by directly taking them from GitHub. 

### Issue

Addresses #2544 

### Benefits

Direct connection to tutorial

### Possible Drawbacks

More code to test

```bash
cargo run --example tutorial
```

As test

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -35,8 +35,8 @@ pub trait ResponseHandler<T = Vec<u8>> {
     fn handle(self, response: Response<T>) -> Self::Output;
 }
 
-/// Phantom struct that handles responses of Query API.
-/// Depending on input query struct, transforms a response into appropriate output.
+/// Phantom struct that handles responses of Query API.  Depending on
+/// input query struct, transforms a response into appropriate output.
 #[derive(Clone, Copy)]
 pub struct QueryResponseHandler<R>(PhantomData<R>);
 
@@ -95,8 +95,9 @@ where
 
 /// Different errors as a result of query response handling
 #[derive(Debug, thiserror::Error)]
-// `QueryError` variant is too large (32 bytes), but I think that this enum is not
-// very frequently constructed, so boxing here is unnecessary.
+// `QueryError` variant is too large (32 bytes), but I think that this
+// enum is not very frequently constructed, so boxing here is
+// unnecessary.
 #[allow(variant_size_differences)]
 pub enum ClientQueryError {
     /// Certain Iroha query error
@@ -152,7 +153,8 @@ impl ResponseHandler for StatusResponseHandler {
     }
 }
 
-/// Private structure to incapsulate error reporting for HTTP response.
+/// Private structure to incapsulate error reporting for HTTP
+/// response.
 struct ResponseReport(eyre::Report);
 
 impl ResponseReport {
@@ -176,9 +178,10 @@ impl From<ResponseReport> for eyre::Report {
     }
 }
 
-/// More convenient version of [`iroha_data_model::prelude::PaginatedQueryResult`].
-/// The only difference is that this struct has `output` field extracted from the result
-/// accordingly to the source query.
+/// More convenient version of
+/// [`iroha_data_model::prelude::PaginatedQueryResult`]. The only
+/// difference is that this struct has `output` field extracted from
+/// the result accordingly to the source query.
 #[derive(Clone, Debug)]
 pub struct ClientQueryOutput<R>
 where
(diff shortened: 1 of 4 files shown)


6r1d (approved):

client/src/http.rs
```diff
@@ -10,11 +10,20 @@ pub use http::{Method, Response, StatusCode};
...
+{
     #[must_use]
+    /// Entrypoint - create a new builder with specified method and URL.
     fn new(method: Method, url: impl AsRef<str>) -> Self;
 
+    /// Entrypoint - create a new builder with specified method and URL.
+    ///
+    /// # Errors
+    /// If either the `method` or `url` are invalid.
+    fn try_new(method: Method, url: impl AsRef<str> + core::fmt::Display) -> eyre::Result<Self>;
```
0x009922, line 25: What is the point of `try_new` method? How should the `RequestBuilder` user decide whether to use `new` or `try_new` method to create a builder?

0x009922: Also, `try_new` method should have the `#[must_use]` attr as well, I guess.

6r1d (dismissed):

client/src/http.rs
0x009922: I see. LGTM, but I'd persist on adding `#[deprecated]` for `new` method and `#[must_use]` for `try_new`.

0x009922 (dismissed):

client/src/http.rs
appetrosyan: Certainly.

6r1d:
TO-DOs:

 - [ ] [Add a deprecation warning](https://t.me/c/1705484471/1616) (`#[deprecated]`) into `http.rs`'s `new`, since it is being superseded by `try_new`. 
 - [ ] `#[must_use]` is redundant for `try_new`.
 - [ ] There's some code that should be added to the examples

6r1d:
Documentation check [currently stops](https://github.com/hyperledger/iroha/runs/7788103547?check_suite_focus=true), because it sees the `^` character, caused by the example warning:

```bash
if cargo doc -q 2>&1 | grep '^'; then exit 1; fi
```

My first attempt at fixing the code failed, but I'll try more.

client/src/client.rs
```diff
@@ -152,7 +153,8 @@ impl ResponseHandler for StatusResponseHandler {
     }
 }
 
-/// Private structure to incapsulate error reporting for HTTP response.
+/// Private structure to incapsulate error reporting for HTTP
```
appetrosyan, line 156:
```suggestion
/// Private structure to encapsulate error reporting for HTTP
```

6r1d:
I have to extend `expect` with actual error messages.

client/src/http.rs
```diff
@@ -10,11 +10,21 @@ pub use http::{Method, Response, StatusCode};
...
+pub trait RequestBuilder
+where
+    Self: Sized,
+{
     #[must_use]
+    #[deprecated]
+    /// Entrypoint - create a new builder with specified method and URL.
     fn new(method: Method, url: impl AsRef<str>) -> Self;
 
+    /// Entrypoint - create a new builder with specified method and URL.
```
outoftardis, line 22:
```suggestion
    /// Entrypoint: create a new builder with specified method and URL.
    fn new(method: Method, url: impl AsRef<str>) -> Self;

    /// Entrypoint: create a new builder with specified method and URL.
```

client/src/client.rs
```diff
@@ -424,16 +432,18 @@ impl Client {
         let response = req
             .build()?
             .send()
-            .wrap_err_with(|| format!("Failed to send transaction with hash {:?}", hash))?;
+            .wrap_err_with(|| format!("Failed to send transaction with hash {hash}"))?;
```
0x009922, line 435:
```suggestion
            .wrap_err_with(|| format!("Failed to send transaction with hash {hash:?}"))?;
```

I guess `Debug` output is more suitable in an error branch.

client/src/http.rs
```diff
@@ -10,11 +10,21 @@ pub use http::{Method, Response, StatusCode};
...
+    ///
+    /// # Errors
+    /// If either the `method` or `url` are invalid.
```
0x009922, line 25: I am not sure whether `method` could be invalid in any case because of `Method` type which follows the "parse, not validate" pattern. It cannot be constructed without validation.

ilchu: Maybe what's meant here is that e.g. the server may or may not support methods such as `CONNECT`, and in the latter case calling the function with it would be invalid (or rather disallowed).

PR #2610, opened 2022-08-11
[ci]: Fix `iroha2:build` image build

BAStos525:

### Description of the Change
Add build-args corresponding tags values for workflows.

### Issue
The step of `iroha2:build` image building can not identify TAG value in `Dockerfile.base` ARG pointer,

### Benefits

Make CI step of `iroha2:base` image build workable.

### Possible Drawbacks

Should be none.


pesterev (approved):

s8sato (approved):
Good catch. Previously, we mistakenly specified `build-args` for `Dockerfile.base`. What requires the args is `Dockerfile.build`

Issue #2560, opened 2022-07-28
Fix late joiner bug


SamHSmith:
Currently on iroha2-staging the following test fails.
```
integration::unregister_peer::unstable_network_stable_after_add_and_after_remove_peer
```
The reason being that it requires a peer to connect to the network late, having missed the genesis round and all following rounds. What should happen is that block_sync is able to fetch the genesis and later blocks, Sumeragi should then start participating in the latest rounds.

I'm not sure how difficult this issue is. It is probably better to do it after view changes are properly implemented. This is because without them, the new peer might not be able to join consensus properly.

PR #2618, opened 2022-08-12
[fix] #2560: Added back in block_sync and peer disconnecting.

SamHSmith:

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -389,7 +389,6 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) ->
 async fn handle_status(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) -> Result<Json> {
     sumeragi.update_metrics(network);
     let status = Status::from(&sumeragi.wsv_clone().metrics);
-    println!("{:?}", status);
     Ok(reply::json(&status))
 }
 
--- a/core/src/block_sync.rs
+++ b/core/src/block_sync.rs
@@ -2,37 +2,31 @@
 
 use std::{fmt::Debug, sync::Arc, time::Duration};
 
-use iroha_actor::{broker::*, prelude::*, Context};
-use iroha_crypto::SignatureOf;
+use iroha_actor::{
+    broker::{Broker, *},
+    prelude::*,
+    Context,
+};
+use iroha_crypto::{SignatureOf, *};
 use iroha_data_model::prelude::*;
 use iroha_logger::prelude::*;
+use iroha_macro::*;
+use iroha_p2p::Post;
+use iroha_version::prelude::*;
+use parity_scale_codec::{Decode, Encode};
 
-use self::{
-    config::BlockSyncConfiguration,
-    message::{Message, *},
-};
+use self::config::BlockSyncConfiguration;
 use crate::{
     prelude::*,
     sumeragi::{network_topology::Role, Sumeragi},
-    VersionedCommittedBlock,
+    NetworkMessage, VersionedCommittedBlock,
 };
-
-/// The state of `BlockSynchronizer`.
-#[derive(Clone, Debug)]
-enum State {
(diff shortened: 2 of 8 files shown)


core/src/block_sync.rs
```diff
@@ -129,193 +101,140 @@ impl BlockSynchronizer {
...
-pub mod message {
-    use iroha_actor::broker::Broker;
-    use iroha_crypto::*;
-    use iroha_data_model::prelude::*;
-    use iroha_logger::prelude::*;
-    use iroha_macro::*;
-    use iroha_p2p::Post;
-    use iroha_version::prelude::*;
-    use parity_scale_codec::{Decode, Encode};
+use iroha_actor::broker::Broker;
```
appetrosyan, line 106: Please move the imports to the top, or use a wrapping module.

core/src/block_sync.rs
```diff
@@ -129,193 +101,140 @@ impl BlockSynchronizer {
...
-                        warn!("Error: not sending any blocks as batch_size is equal to zero.");
-                        return;
-                    }
-                    if *hash == block_sync.sumeragi.latest_block_hash() {
-                        return;
-                    }
+impl Message {
+    /// Handles the incoming message.
+    #[iroha_futures::telemetry_future]
+    pub async fn do_message(&self, block_sync: &mut BlockSynchronizer) {
```
appetrosyan, line 187: `do` isn't very informative. Please consider renaming.

core/src/sumeragi/fault.rs
```diff
@@ -278,7 +287,7 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
                 return;
             }
 
-            let reconnect_in_ms = 100;
+            let reconnect_in_ms = 250;
```
appetrosyan, line 290: Any particular reason for this change?

core/src/sumeragi/fault.rs
```diff
@@ -267,7 +276,7 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
 
     pub fn wait_for_peers(&self, network_topology: &Topology) {
         iroha_logger::info!("Waiting for active peers",);
-        for i in 0..80 {
+        for i in 0..120 {
```
appetrosyan, line 279: Explain, please.

core/src/block_sync.rs
SamHSmith: It's not that different from "handle" but sure.

core/src/sumeragi/fault.rs
SamHSmith: The timings needed to be changed for the test to pass.

core/src/sumeragi/fault.rs
SamHSmith: Same as above. You want the test to fail in a reasonable time but wait long enough so that it always passes if the code works.

cli/src/torii/routing.rs
```diff
@@ -389,7 +389,6 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) ->
 async fn handle_status(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) -> Result<Json> {
     sumeragi.update_metrics(network);
     let status = Status::from(&sumeragi.wsv_clone().metrics);
-    println!("{:?}", status);
     Ok(reply::json(&status))
 }
```
s8sato, line 394: > It is probably better to do it after view changes are properly implemented. 

So you have determined that the order of implementation does not matter, right?

core/src/block_sync.rs
SamHSmith: has been done

appetrosyan (approved):

cli/src/torii/routing.rs
SamHSmith: Yes. I initially thought I that this would depend on having view changes implemented. But all it required was to have block_sync connected to sumeragi again.

s8sato (approved):

PR #2625, opened 2022-08-15
[feature] #2491: Enum support in FFi

### LINKED ISSUE #2491: Support data-carrying enum serialization in FFI
Example of such an enum is:
```rs
pub enum Enumeration {
    A(u32),
    B(PublicKey),
    C,
}
```

These enums (and, recursively, their variants) have to be converted into their `repr(c)` equivalent. Something like the following:
```rs
#[repr(C)]
#[derive(Clone, Copy)]
pub struct FfiEnumeration {
    tag: u8,
    payload: FfiEnumerationPayload,
}

#[repr(C)]
#[derive(Clone, Copy)]
union FfiEnumerationPayload {
    A: <u32 as TryFromReprC>::Source,
    B: <PublicKey as TryFromReprC>::Source,
}
```

where tag is calculated from the number of variants and doesn't have to be defined on the enum

### LINKED ISSUE #2622: `u128`/`i128` support in FFI
The implementation after #2625 is outright wrong. `u128`/`i128` is not FFI safe and cannot be sent across without serialization. I propose it to be serialized as `FfiTuple2(u64, u64)` or `[u64; 2]`

mversic:
### Description of the Change

* removed `wasm` feature flag for `target_type`. After some discussion with @Erigara this was decided to be preferable
* replaced `IntoFfi` and `TryFromReprC` with `FfiType` and `FfiConvert` traits. Previous separation couldn't be worked with mostly due to conflicting requirements on lifetimes on data carrying enums. Maybe GATs would have been of help here, not sure anymore. If needed `FfiConvert` can be split into `TryFromFfi`/`IntoFfi`
* removed lifetimes from all `ReprC` types (namely `SliceRef`/`SliceMut`). It was too noisy and difficult to deal with but I think it can (should?) be reintroduced now
* added `non_robust_ref_mut` feature flag. Might be useful to disallow mutation for non-robust types
* improved conversion system so that it can now convert any recursive type
* added implementation for data-carrying enums
* prevent double-free in function return via `NonLocal`

### Type challenges
There is no specialization or HKTs in `Rust`, so it's very difficult to properly delineate type categories. But it can be emulated in the type system. This is what the `Ir` trait is for. `Rust` type is first wrapped in the correct `Ir` type after which static dispatch on IR marker types provides an optimal(zero-copy when possible) implementation of `FFIType`. 

### Issue

Closes #2491 #2629

### Benefits

optimal conversion, enums support

### Possible Drawbacks

I'm not sure if it's more complex than before of just a different perspective

This is what a conversion for data-carrying enum would produce (it is not even required it is `repr(C)`:
```rs
pub enum DataCarryingEnum {
    A(String);
    B
}
pub struct ReprCDataCarryingEnum {
    tag: u8,
    payload: DataCarryingEnumPayload,
}
union DataCarryingEnumPayload {
    A: ManuallyDrop<<String as FfiType>::ReprC>,
    B: ()
}
```

There was a suggestion to use a different enum representation. It can be modified. At the moment I found this easier to implement

I've tried some but all failed in expresivity


ffi/src/primitives.rs
```diff
@@ -288,4 +288,4 @@ macro_rules! primitive_impls {
     )+};
 }
 
-primitive_impls! {u8, u16, u32, u64, i8, i16, i32, i64}
+primitive_impls! {u8, u16, u32, u64, i8, i16, i32, i64, i128, u128}
```
mversic, line 291: [this](/https://github.com/rust-lang/rust/issues/54341#issuecomment-428606606) is not ffi safe as it seems. What I propose is to use `FfiTuple2(u64, u64)`

ffi/tests/ffi_export.rs
```diff
@@ -308,28 +338,100 @@ fn return_result() {
...
+
+#[test]
+fn data_carrying_enum_conversion() {
+    let data_carrying_enum = DataCarryingEnum::A(get_new_struct());
+    let mut output = MaybeUninit::new(DataCarryingEnum::B(42));
+
+    unsafe {
+        assert_eq!(
+            FfiReturn::Ok,
+            __freestanding_with_data_carrying_enum(data_carrying_enum.into_ffi(), output.as_mut_ptr())
```
mversic, line 396: this is a problem because `into_ffi` produces `__iroha_ffi__ReprCDataCarryingEnumIntoFfiTarget` but `__freestanding_with_data_carrying_enum` takes `__iroha_ffi__ReprCDataCarryingEnumTryFromReprCSource`

I believe this could be resolved if we made `IntoFfi::into_ffi` also take `Store` as does `TryFromReprC`

ffi/src/primitives.rs
appetrosyan: The link is broken.

ffi/src/primitives.rs
```diff
@@ -195,7 +195,7 @@ macro_rules! wasm_repr_impls {
             type Repr = $dst;
         }
     };
-    ($src:ty) => {
+    ($src:ty) = {
```
Erigara, line 198: typo

ffi/src/primitives.rs
Erigara: I checked `struct U128([u64; 2]);` and it seems like it passed by value (as `i64 i64`) and returned by reference.

```rust
#[repr(transparent)]
pub struct U128Pair([u64; 2]);

#[no_mangle]
pub unsafe extern "C" fn simple(value: U128Pair) -> U128Pair {
    value
}
```

Produced wasm function: 
```wasm
  (func $simple (type 1) (param i32 i64 i64)
    local.get 0
    local.get 2
    i64.store offset=8
    local.get 0
    local.get 1
    i64.store)
```

It is wroth noticing that rustc produce exactly the same code if we just use `u128`: 

```rust
#[no_mangle]
pub unsafe extern "C" fn simple_plain(value: u128) -> u128 {
    value
}
```

```wasm
  (func $simple_plain (type 1) (param i32 i64 i64)
    local.get 0
    local.get 2
    i64.store offset=8
    local.get 0
    local.get 1
    i64.store)
```

mversic: ok tnx. So here is the thing. Arrays in Rust are `improper_ctypes`, i.e. they have defined layout, but cannot be used in `extern` functions binding to some C code because in C arrays decay to pointers. To interface C bindings they can either be sent by pointer or wrapped into a struct in which case they'll be passed by value.

Having said this, our library always produces rust FFI code from some rust code and I believe that we shouldn't worry about this. However, if the user wishes to generate c bindings with `cbindgen`, they currently [don't support](https://github.com/eqrion/cbindgen/issues/528) this and offer no workaround

mversic: To paraphrase, this is something that should be worried about when you have a C library and are generating rust bindings to match it's ABI. If you have a Rust library and are generating C bindings to match it's ABI you can match it on the C side by wrapping an array in a struct. However, this is not something that `cbindgen` supports as of yet. I've opened [a discussion](https://github.com/eqrion/cbindgen/issues/777) to see if this is something that can be supported

Also, note that `u128`/`i128` [have](https://doc.rust-lang.org/reference/type-layout.html#primitive-data-layout) defined layouts

mversic: I might have been wrong here, I didn't think about calling convention.
Let's move the discussion to #2622 and #2629

ffi/derive/src/convert.rs
```diff
@@ -44,449 +45,763 @@ pub fn derive_into_ffi(input: &DeriveInput) -> TokenStream2 {
...
+                    }
+                }
+                syn::Fields::Unnamed(syn::FieldsUnnamed { unnamed, .. }) => {
+                    abort!(unnamed, "Only 1-sized variants are supported")
+                }
+                syn::Fields::Named(syn::FieldsNamed { named, .. }) => {
+                    abort!(named, "Named variants are not supported")
+                }
+                syn::Fields::Unit => {
+                    quote! { #enum_name::#variant_name => #repr_c_enum_name { tag: #idx }, }
```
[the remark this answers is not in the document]
Erigara: Another approach could be to store `tag` inside union in the same place for every variant ([as proposed here](https://github.com/rust-lang/rfcs/pull/2195)), this could lead to more compact memory layout.

ffi/tests/ffi_export.rs
appetrosyan: Explicit ownership over memory, right?

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
appetrosyan: Prefer the other approach. This is less surprising to people who are only familiar with Rust.

data_model/src/events/data/filters.rs
```diff
@@ -2,11 +2,45 @@
...
+    type Store = ();
+
+    unsafe fn try_from_repr_c(
+        source: Self::Source,
+        store: &'itm mut Self::Store,
+    ) -> iroha_ffi::Result<Self> {
+        unimplemented!()
+    }
+}
+impl IntoFfi for EventFilter {
```
appetrosyan, line 36: newline

data_model/src/events/data/filters.rs
```diff
@@ -2,11 +2,45 @@
...
+pub struct ReprCEventFilter {
+    tag: u8,
+    //payload: <EntityFilter as TryFromReprC<'itm>>::Source,
+}
+
+impl<'itm> iroha_ffi::Output for ReprCEventFilter {
+    type OutPtr = *mut Self;
+}
+
+// TODO: Temporary manual implementation
```
appetrosyan, line 23: Create a ticket to fix.

data_model/src/peer.rs
```diff
@@ -1,42 +1,59 @@
...
-#[derive(
-    Debug, Display, Clone, IdOrdEqHash, Decode, Encode, Deserialize, Serialize, IntoSchema,
-)]
-#[display(fmt = "@@{}", "id.address")]
-#[id(type = "Id")]
-pub struct Peer {
-    /// Peer Identification.
-    pub id: <Self as Identifiable>::Id,
+ffi::ffi_item! {
+    /// Peer represents Iroha instance.
```
appetrosyan, line 22:
```suggestion
    /// Representation of other Iroha Peer instances running in separate processes. 
```

ffi/derive/src/convert.rs
```diff
@@ -1,9 +1,9 @@
 use proc_macro2::TokenStream as TokenStream2;
 use proc_macro_error::abort;
 use quote::quote;
-use syn::{parse_quote, DeriveInput, Ident};
```
appetrosyan, line 4: This is the meat of the PR.

ffi/src/ir.rs
```diff
@@ -0,0 +1,235 @@
...
+        unimplemented!()
+    }
+}
+
+unsafe impl<'itm> Transmute for &'itm str {
+    type ReprC = SliceRef<u8>;
+
+    unsafe fn is_valid(source: Self::ReprC) -> FfiReturn {
+        if let Some(slice) = source.into_rust() {
+            if matches!(core::str::from_utf8(slice), Err(_)) {
```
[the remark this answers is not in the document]
mversic: done, will update later

ffi/src/repr_c.rs
```diff
@@ -0,0 +1,191 @@
...
+    type RustStore = Vec<Transmutable<T, T::ReprC>>;
+    type FfiStore = ();
+
+    fn into_repr_c(self, store: &mut Self::RustStore) -> Self::ReprC {
+        *store = self;
+
+        let (ptr, len) = (store.as_ptr(), store.len());
+        SliceRef::from_raw_parts(ptr.cast(), len)
+    }
+    unsafe fn try_from_repr_c(source: Self::ReprC, _: &mut ()) -> Result<Self> {
```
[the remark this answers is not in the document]
mversic: that's true, but I'd rather not reallocate

ffi/src/primitives.rs
```diff
@@ -3,288 +3,329 @@
...
-
-        if !is_valid_bool(source) {
-            return Err(FfiReturn::TrapRepresentation);
+#[cfg(target_family = "wasm")]
+use crate::{FfiType, NonTransmute};
+
+// TODO: It might be better to make it a proc macro and have it be called from this crate
+// in which case it wouldn't be in the public API of this crate as the current macro is.
+// NOTE: This code is literaly the same as is in the `iroha_ffi_derive` crate for deriving enums
+macro_rules! fieldless_enum_derive {
```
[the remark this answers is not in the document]
mversic: name of the macro is quite correct, `Ordering` is enum, `bool` can also be considered an enum. You can use it to derive for other fieldless enums in std

ffi/src/primitives.rs
```diff
@@ -3,288 +3,329 @@
...
+            unsafe fn is_valid(source: Self::ReprC) -> FfiReturn {
+                FfiReturn::Ok
             }
         }
-        unsafe impl<'slice> IntoFfiSliceMut<'slice> for $ty {
-            type Target = SliceMut<'slice, $ty>;
 
-            fn into_ffi(source: &mut [Self]) -> Self::Target {
-                SliceMut::from_slice(source)
+        // TODO: How should wasm types be transfered as *mut $src or *mut $dst?
```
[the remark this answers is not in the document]
mversic: ah yes, this is actually what is being done, I left this comment by accident

ffi/src/primitives.rs
```diff
@@ -3,288 +3,329 @@
...
+        unsafe impl crate::Output for $ty {
+            type OutPtr = *mut Self;
+        }
+
+        impl Ir for $ty {
+            type Type = Transmutable<Self, Self>;
```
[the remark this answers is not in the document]
mversic: because the struct is parametrized with the input and the output type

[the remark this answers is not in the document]
mversic: maybe I can take the output type from the `Transmute` trait

ffi/src/primitives.rs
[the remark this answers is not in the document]
Erigara: ok, sounds reasonable

ffi/src/repr_c.rs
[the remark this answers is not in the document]
Erigara: Uh, in this case i think calling `shrink_to_fit` is not suitable, because it call [`Allocator::shrink`](https://doc.rust-lang.org/src/core/alloc/mod.rs.html#319) inside which reallocates memory.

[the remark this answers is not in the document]
mversic: Maybe I could go with `into_boxed_slice`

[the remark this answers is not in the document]
Erigara: The same problem actually ([source](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#1032))

ffi/src/ir.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn into_ir(slice_ref: &[Self]) -> &[Self::Type];
+    fn from_ir(source: &[Self::Type]) -> &[Self];
+}
+pub trait IrSliceMut: Ir {
+    fn into_ir(slice_mut: &mut [Self]) -> &mut [Self::Type];
+    fn from_ir(source: &mut [Self::Type]) -> &mut [Self];
+}
+
+#[repr(C)]
+pub union Transmutable<T: Transmute, U: ReprC> {
```
Erigara, line 64: Do we assume that `T` and `U` have the same memory layout?

ffi/derive/src/lib.rs
```diff
@@ -134,12 +125,16 @@ pub fn try_from_repr_c_derive(input: TokenStream) -> TokenStream {
...
-pub fn ffi_export(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn ffi_export(attr: TokenStream, item: TokenStream) -> TokenStream {
     match parse_macro_input!(item) {
         Item::Impl(item) => {
             let impl_descriptor = ImplDescriptor::from_impl(&item);
             let ffi_fns = impl_descriptor.fns.iter().map(ffi_fn::gen_definition);
 
+            if !attr.is_empty() {
+                abort!(item, "Unknown tokens in the attribute");
+            }
```
Erigara, line 136: Are you planning to use `attr` in future?

ffi/derive/src/convert.rs
```diff
@@ -431,472 +488,380 @@ fn derive_try_from_repr_c_for_fieldless_enum(
...
     let name = &input.ident;
-    let mut generics = input.generics.clone();
-    let lifetime: syn::Lifetime = syn::parse_quote!('__iroha_ffi_slice);
-    let lifetimes = &[lifetime.clone()];
-    let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, &[], &[]);
+    let inner = match &input.data {
+        // TODO: Do we check it's not ZST?
+        syn::Data::Enum(_) => {
+            unimplemented!()
```
Erigara, line 672: i'm not sure if we sould support transparent enums, they are sound like rare case

ffi/derive/src/convert.rs
```diff
@@ -431,472 +488,380 @@ fn derive_try_from_repr_c_for_fieldless_enum(
...
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+
     let name = &input.ident;
-    let mut generics = input.generics.clone();
-    let lifetime: syn::Lifetime = syn::parse_quote!('__iroha_ffi_slice);
-    let lifetimes = &[lifetime.clone()];
-    let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, &[], &[]);
+    let inner = match &input.data {
+        // TODO: Do we check it's not ZST?
```
Erigara, line 670: we could add such check, but it would produce akward errors in case if `inner` is generic.

```
const _: () = assert!(core::mem::size_of::<#inner>() > 0, "inner field is ZST");
```

ffi/derive/src/convert.rs
```diff
@@ -431,472 +488,380 @@ fn derive_try_from_repr_c_for_fieldless_enum(
...
-            fn into_ffi(self) -> Self::Target {
-                let inner = <Self as iroha_ffi::Transparent>::into_inner(self);
-                iroha_ffi::IntoFfi::into_ffi(inner)
-            }
-        }
+    input
+        .generics
+        .make_where_clause()
+        .predicates
+        .push(transparency_bound);
```
Erigara, line 719:
```suggestion
    let mut where_clause = input.generics.make_where_clause();
    where_clause.predicates.push(transparency_bound);

    let mut owned_where_clause = where_clause.expect("Missing where clause").clone();
    let mut ref_where_clause = where_clause.expect("Missing where clause").clone();
    let mut mut_ref_where_clause = where_clause.expect("Missing where clause").clone();
    let mut vec_where_clause = where_clause.expect("Missing where clause").clone();
    let mut slice_ref_where_clause = where_clause.expect("Missing where clause").clone();
    let mut slice_mut_where_clause = where_clause.expect("Missing where clause").clone();
```

ffi/derive/src/convert.rs
```diff
@@ -431,472 +488,380 @@ fn derive_try_from_repr_c_for_fieldless_enum(
...
-    let self_trait_bounds = &[syn::parse_quote!(
-        iroha_ffi::TransparentVec<Inner = __iroha_ffi_T>
-    )];
     let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, types, self_trait_bounds);
-
-    quote! {
-        impl<#generic_params> iroha_ffi::slice::IntoFfiSliceRef<#lifetime> for #name #ty_generics #where_clause {
-            type Target = <<Self as iroha_ffi::TransparentSlice>::Inner as iroha_ffi::slice::IntoFfiSliceRef<#lifetime>>::Target;
+        split_for_impl_with_type_params(&mut input.generics, types);
```
Erigara, line 729:
```suggestion
    let (generic_params, ty_generics, _) =
        split_for_impl_with_type_params(&mut input.generics, types);
```

ffi/src/ir.rs
```diff
@@ -0,0 +1,249 @@
...
+    // NOTE: Deriving `into_inner` is always valid because T::Inner is a superset of T
+    fn into_inner(outer: Self) -> Self::Inner;
+
+    /// Recover transparent struct from its non-ZST field
+    // FIXME: We should force the transparent type to have `from_inner` and derive implementation
+    // only if requested specifically by the user (via macro attribute). The reason for this is
+    // that even though two types are equal on the byte level, they may have different semantics
+    // that cannot be proven when deriving `from_inner` (e.g. `NonZeroU8` is a wrapper for `u8`)
+    // Deriving implementation of this method cannot be guaranteed to be correct
+    unsafe fn from_inner(inner: Self::Inner) -> Self;
```
Erigara, line 25: In this case should't this function be `try_from_inner`?

ffi/derive/src/convert.rs
```diff
@@ -3,426 +3,483 @@ use proc_macro_error::{abort, OptionExt};
...
+
+    pub fn derive_ffi_type_for_fieldless_enum(
+        enum_name: &Ident,
+        enum_: &syn::DataEnum,
+        repr: &[syn::NestedMeta],
+    ) -> TokenStream2 {
+        let (ffi_type, largest_discriminant, is_wasm_primitive) = enum_size(enum_name, repr);
+        let (discriminants, discriminant_decls) = gen_discriminants(enum_name, enum_, &ffi_type);
+
+        // TODO: most of the times niche trait can be implemented and in edge cases robust trait
```
Erigara, line 201: So only case when enum does't have niche is when it has number of variants equal to `ffi_type::MAX - ffi_type::MIN`, we could check this condition during macro expansion.

Then we could do [the following](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b1fcf019c2454cff92e41b63f9d659fc) to find niche value.

ffi/derive/src/convert.rs
```diff
@@ -3,426 +3,483 @@ use proc_macro_error::{abort, OptionExt};
...
 }
 
-pub fn derive_into_ffi(input: &DeriveInput) -> TokenStream2 {
-    if !matches!(input.vis, syn::Visibility::Public(_)) {
-        abort!(input.vis, "Only public items are supported");
+fn derive_ffi_type_for_struct(_: &Ident) -> TokenStream2 {
+    quote! {
+        // TODO:
     }
+}
```
Erigara, line 45: I suppose we should move this into `struct_` mod.

ffi/derive/src/convert.rs
```diff
@@ -3,426 +3,483 @@ use proc_macro_error::{abort, OptionExt};
...
-        );
+                fn into_ir(self) -> Self::Type {
+                    self
+                }
+                fn from_ir(source: Self::Type) -> Self {
+                    source
+                }
+            }
+            impl<#generic_params> iroha_ffi::ir::Ir for Box<#name #ty_generics> #where_clause {
+                type Type = Box<#name #ty_generics>;
```
Erigara, line 80:
```suggestion
                type Type = Self;
```

ffi/derive/src/convert.rs
```diff
@@ -3,426 +3,483 @@ use proc_macro_error::{abort, OptionExt};
...
-        };
+                fn into_ir(self) -> Self::Type {
+                    unsafe {Box::from_raw(Box::into_raw(self).cast())}
+                }
+                fn from_ir(source: Self::Type) -> Self {
+                    unsafe {Box::from_raw(Box::into_raw(source).cast())}
+                }
+            }
+            impl<#lifetime, #generic_params> iroha_ffi::ir::Ir for &#lifetime #name #ty_generics #where_clause {
+                type Type = iroha_ffi::ir::Transmutable<&#lifetime #name #ty_generics, *const #name #ty_generics>;
```
Erigara, line 90:
```suggestion
                type Type = iroha_ffi::ir::Transmutable<Self, *const #name #ty_generics>;
```

ffi/derive/src/convert.rs
```diff
@@ -3,426 +3,483 @@ use proc_macro_error::{abort, OptionExt};
...
-}
+                fn into_ir(self) -> Self::Type {
+                    From::from(self)
+                }
+                fn from_ir(source: Self::Type) -> Self {
+                    unsafe {source.into_inner_ir()}
+                }
+            }
+            impl<#lifetime, #generic_params> iroha_ffi::ir::Ir for &#lifetime mut #name #ty_generics #where_clause {
+                type Type = iroha_ffi::ir::Transmutable<&#lifetime mut #name #ty_generics, *mut #name #ty_generics>;
```
Erigara, line 100:
```suggestion
                type Type = iroha_ffi::ir::Transmutable<Self, *mut #name #ty_generics>;
```

ffi/derive/src/convert.rs
```diff
@@ -431,472 +488,380 @@ fn derive_try_from_repr_c_for_fieldless_enum(
...
-        // SAFETY: `Self` and `Self::Inner` have the same size because from macro generation `Self::Inner` is the inner field of transparent `Self` with the same size.
-        unsafe impl <#generic_params> iroha_ffi::Transparent for #name #ty_generics #where_clause {
-            type Inner = #inner;
-
-            fn into_inner(outer: Self) -> Self::Inner {
-                // SAFETY: `Self` and `Self::Inner` have the same memory layout
-                unsafe { core::mem::transmute(outer) }
+            fn into_ir(self) -> Self::Type {
+                iroha_ffi::ir::Ir::into_ir(unsafe {
+                    &*(self as *const #name #ty_generics as *const _)
```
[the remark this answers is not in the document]
mversic: casting is safe, dereferencing isn't

ffi/derive/src/convert.rs
mversic: we should make an issue to track this

ffi/src/ir.rs
mversic: yes, this is an invariant enforced in the `Transmute` trait

ffi/src/primitives.rs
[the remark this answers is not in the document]
mversic: yes

ffi/src/ir.rs
mversic: true, I resorted to returning `Option<Self>`

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
Erigara: sure, i mean that we dereferencing instantly here

ffi/derive/src/util.rs
```diff
@@ -39,42 +39,41 @@ pub fn gen_derived_methods(item: &syn::ItemStruct) -> Vec<FnDescriptor> {
...
-            let mut store = ();
-            let #arg_name: #src_type = iroha_ffi::TryFromReprC::try_from_repr_c(#arg_name, &mut store)?;
+            let mut #store_name = ();
+            let #arg_name: #src_type = iroha_ffi::FfiType::try_from_ffi(#arg_name, &mut #store_name)?;
         };
 
         if let Type::Reference(ref_type) = &src_type {
             let elem = &ref_type.elem;
 
+            // TODO: Seems that this hack can be removed now?
```
Erigara, line 54: Could you provide more context?

ffi/derive/tests/ui_fail/nested_owned_structure.rs
```diff
@@ -0,0 +1,21 @@
+use std::{alloc::alloc, mem::MaybeUninit};
```
Erigara, line 1: Is't `.stderr` file is missing?

ffi/src/lib.rs
```diff
@@ -95,23 +87,18 @@ pub trait OutPtrOf<T>: ReprC {
     ///
     /// # Safety
     ///
-    /// All conversions from a pointer must ensure pointer validity beforehand
+    /// * All conversions from a pointer must ensure pointer validity beforehand
+    // TODO: It could return bool for successful vs not?
```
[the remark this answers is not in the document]
mversic: yes, that's the question

ffi/derive/tests/ui_fail/nested_owned_structure.rs
mversic: could be, I haven't fixed compile tests yet

ffi/Cargo.toml
```diff
@@ -5,7 +5,31 @@ authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
...
+
+# TODO:
+# Whether or not to perform validity checks when converting from C type to Rust type.
+# This is useful when the side you are receiving data from can be considered trusted.
+# Most notably this feature can be used in smart contracts that interact with `Iroha`
+# no_checks = []
+
+# Enables mutable references of non-robust transmutable types in FFI, non-transmutable are always disallowed.
+# When handing out non-robust mutable references across FFI, it's possible that the other
+# side will modify the given reference and store type's trap representation inside it.
```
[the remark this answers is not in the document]
mversic: it's valid to modify `*mut T` and invalid to modify `*const T`

ffi/Cargo.toml
```diff
@@ -5,7 +5,31 @@ authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
...
+# send references as is across FFI they should enable this feature in their crate
+#
+# # Example
+#
+# When `by_ref` is active:
+# &T -> *const T
+#
+# When `by_ref` is not active:
+# &T -> T
+by_ref = []
```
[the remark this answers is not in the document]
mversic: I would say that we always want to activate this. Not activating this feature removes a family of UBs where data behind `*const T` is modified but this comes at a high cost of cloning all the data

appetrosyan:
Just an observation, rather than a code-related comment. The FFI API is changing quite a bit. We probably need to have a discussion before the implementations are set in stone. We need to agree on names, such that we don't need to change them ever.

ffi/derive/src/util.rs
mversic: it's not relevant anymore, I've updated the code with attached comment

mversic:
> Just an observation, rather than a code-related comment. The FFI API is changing quite a bit. We probably need to have a discussion before the implementations are set in stone. We need to agree on names, such that we don't need to change them ever.

totally agree. Now is the about the right time to do it

Erigara:
> There was a suggestion to use a different enum representation. It can be modified. At the moment I found this easier to implement

I think current approach is fine, we can change it later if we really need to.

ffi/derive/src/convert.rs
```diff
@@ -1,871 +1,497 @@
...
-            1 + #discriminant
-        };
-
-        acc
-    })
-}
-
-fn derive_try_from_repr_c_for_opaque_item_wrapper(input: &DeriveInput) -> TokenStream2 {
+fn derive_ffi_type_for_transparent_item(input: &syn::DeriveInput) -> TokenStream2 {
+    // TODO: We don't check to find which field is not a ZST.
```
Erigara, line 44: I'm not sure that this possible to do during macro expansion.

ffi/derive/src/convert.rs
```diff
@@ -1,871 +1,497 @@
...
-    let mut generics = generics.clone();
-    let lifetime: syn::Lifetime = syn::parse_quote!('__iroha_ffi_itm);
-    let lifetimes = &[lifetime.clone()];
-    let types = &[syn::parse_quote!(__iroha_ffi_T: iroha_ffi::TryFromReprC<#lifetime>)];
-    let self_trait_bounds = &[syn::parse_quote!(
-        iroha_ffi::Transparent<Inner = __iroha_ffi_T>
-    )];
-    let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, types, self_trait_bounds);
+    //struct OpaqueStruct(*mut Extern);
```
Erigara, line 132: commented code.

ffi/src/ir.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+}
+
+// SAFETY: Trivially transmutable
+unsafe impl<T: Transmute> Transmute for &T {
+    type Target = *const T::Target;
+
+    #[inline]
+    unsafe fn is_valid(source: &*const T::Target) -> bool {
+        if let Some(item) = source.as_ref() {
```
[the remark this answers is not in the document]
mversic: This is for the caller to ensure. The best we can do is check it isn't null

ffi/src/ir.rs
```diff
@@ -0,0 +1,326 @@
...
+impl<'itm, T: Ir> Ir for &'itm mut [T] {
+    type Type = IrSliceMut<'itm, T, T::Type>;
+}
+impl<T: Ir> Ir for Vec<T> {
+    type Type = IrVec<T, T::Type>;
+}
+
+// NOTE: This can be contested as it is nowhere documented that String is
+// actually transmutable into Vec<u8>, but implicitly it should be
+// SAFETY: Vec<String> type should be transmutable into Vec<u8>
```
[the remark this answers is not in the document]
mversic: Yes, as far as I was able to find out it is a fair assumption that `String` is a wrapper for `Vec<u8>`. Explicitly it isn't transparent, but implicitly it should be. This conclusion stems from the API of a `String`. This is something to have in mind so I left a note. We should reconsider the String conversion at some point to rely only on guaranteed representations

ffi/derive/src/convert.rs
mversic: I guess you are right. I suppose it could be controlled with derive macro helper attribute. Using the first field is good enough. ZSTs cannot be supported in FFI so the conversion should fail if the first  field is ZST

ffi/src/ir.rs
```diff
@@ -0,0 +1,326 @@
...
+        Self(source)
+    }
+}
+
+impl<T: Transmute> Transparent<T> {
+    /// Convert from [`Self`] into [`T::Inner`].
+    /// Type returned is guaranteed to be valid
+    pub(crate) fn into_inner(self) -> T::Target {
+        #[repr(C)]
+        union TransmuteHelper<T: Transmute> {
```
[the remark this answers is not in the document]
mversic: yeah, for `transmute` the compiler makes a check that types are of same size but cannot do so for generics types

ffi/src/repr_c.rs
```diff
@@ -0,0 +1,690 @@
...
+        source: Self::ReprC,
+        store: &'itm mut Self::FfiStore,
+    ) -> Result<Self> {
+        if source.as_mut().is_some() {
+            return Err(FfiReturn::ArgIsNull);
+        }
+
+        U::try_from_repr_c(source.read(), store)
+            .map(ManuallyDrop::new)
+            .map(|item| (*item).clone())
```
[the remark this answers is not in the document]
mversic: Because there is no ownership transfer:
1. a pointer `*mut T::Type::ReprC` is received
2. you read from the pointer to obtain `T::Type::ReprC`
3. you convert into an owned type with `try_from_repr_c` which gives you `T::Type`
4. `T::Type` is owned so you have to quickly remove the "ownership" by deciding to drop it manually (so that panic doesn't occur in the meantime)
5. Now you can clone it, which is all you wanted since you cannot take ownership over FFI

I can try explain with an example if this isn't clear

crypto/Cargo.toml
```diff
@@ -14,7 +14,7 @@ std = ["ursa"]
 # Force static linking
 vendored = ["openssl-sys"]
 # Replace structures and methods with FFI equivalents to facilitate dynamic linkage (mainly used in smartcontracts)
-ffi_import = []
+#ffi_import = []
```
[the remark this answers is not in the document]
mversic: because it's not finished

ffi/src/ir.rs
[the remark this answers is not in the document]
mversic: I didn't want to worry about special handling `String` impls when the assumption that it is transmutable into `Vec<u8>` seems quite reasonable. This can be dealt with in a separate issue if deemed necessary

crypto/Cargo.toml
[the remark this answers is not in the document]
Erigara: ok, do we have issue for this?

Erigara (dismissed):
lgtm

crypto/Cargo.toml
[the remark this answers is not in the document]
mversic: don't worry, this cannot go unseen. #2231 should be a good starting point

ffi/derive/src/convert.rs
appetrosyan: This warrants a separate issue.

mversic: opened #2794

ffi/src/ir.rs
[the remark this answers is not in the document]
appetrosyan: We shouldn't rely on API features of the standard lib. Given that `String` is given no special treatment syntactically, we could very well suggest people just use `ConstString` and nothing else.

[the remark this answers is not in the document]
mversic: ok, I opened #2795

ffi/Cargo.toml
```diff
@@ -5,7 +5,16 @@ authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
 edition = "2021"
 
 [features]
-wasm = []
+# TODO:
+# Whether or not to perform validity checks when converting from C type to Rust type.
+# This is useful when the side you are receiving data from can be considered trusted.
+# Most notably this feature can be used in smart contracts that interact with `Iroha`
+# no_checks = []
```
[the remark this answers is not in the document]
mversic: The assumption here was that `Iroha` can be trusted to be without bugs and will always send valid values. For instance, it will never send `3` when `core::cmp::Ordering` is expected (valid values are `-1`, `0`, `1`) or `null` for a pointer, etc. Given this assumption smart contract can shave of a few bytes by not including code for doing these checks

macro/derive/tests/ui_pass/variant_count.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(unused)]
```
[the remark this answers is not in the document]
mversic: hm, I guess that since the calculated const value is not bound to a variable it fires the `unused` lint. However, compiler always verifies that the compile time code is valid which still makes this test operational

ffi/Cargo.toml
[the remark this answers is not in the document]
mversic: I'll open a ticket for this and remove the comment

data_model/src/asset.rs
```diff
@@ -3,17 +3,15 @@
 #![allow(clippy::std_instead_of_alloc)]
 
 #[cfg(not(feature = "std"))]
-use alloc::{alloc::alloc, boxed::Box, collections::btree_map, format, string::String, vec::Vec};
+use alloc::{boxed::Box, collections::btree_map, format, string::String, vec::Vec};
```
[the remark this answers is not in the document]
mversic: except for one `vec`, these imports are not used by the ffi macros

ffi/Cargo.toml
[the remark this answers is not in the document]
mversic: #2799

ffi/Cargo.toml
```diff
@@ -5,7 +5,16 @@ authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
...
+
+# Enables mutable references of non-robust transmutable types in FFI, non-transmutable are always disallowed.
+# When handing out non-robust mutable references across FFI, it's possible that the other
+# side will modify the given reference and store type's trap representation inside it.
+non_robust_ref_mut = []
```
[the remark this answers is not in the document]
mversic: I forgot about this. Now I see it's not used at all. I opened #2800 to track it and assigned myself

Erigara (approved):

appetrosyan (approved):

PR #2630, opened 2022-08-16
[fix] #2457: Forcibly shut down kura in tests

QuentinI:
Signed-off-by: Artemii Gerasimovich <gerasimovich@soramitsu.co.jp>

### Description of the Change

So I’ve been debugging #2457, and the problem there is as follows:

When we start a peer for tests, we [create](https://github.com/hyperledger/iroha/blob/a2b373d7fb21dc24f27cf83bf1c8cb9a6b187b98/core/test_network/src/lib.rs#L673) a [TempDir](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html) for kura storage. It is then [held](https://github.com/hyperledger/iroha/blob/a2b373d7fb21dc24f27cf83bf1c8cb9a6b187b98/core/test_network/src/lib.rs#L502) until the peer is dropped to prevent its deletion. But kura thread receiving new blocks actually [lives on its own](https://github.com/hyperledger/iroha/blob/a2b373d7fb21dc24f27cf83bf1c8cb9a6b187b98/core/src/kura.rs#L77) and periodically [checks](https://github.com/hyperledger/iroha/blob/a2b373d7fb21dc24f27cf83bf1c8cb9a6b187b98/core/src/kura.rs#L182) if all other references to kura have been dropped and then returns. 

What happens is essentially a race condition, where after a test is finished, peer is dropped, dropping the tempdir, but kura is in process of writing a new block, so it's working directory is pulled from underneath it. This causes kura thread to panic, which wasn't a problem on its own, but is now caught by panic handler and leads to failure.

This PR is my shot of solving it by providing an ability to forcibly shut kura down and doing just that while dropping a test peer.

### Issue

#2457

### Benefits

Kura doesn't panic in tests.

### Possible Drawbacks

Not a particularly elegant solution.

### Alternate Designs

We can't hold tempdir for long enough, since there's no scope that's guaranteed to exceed kura threads lifetime.

We also can't provide a method to just _wait_ for kura to shut down without forcing it to with how it's currently implemented, since kura shuts down only when all strong references to it are dropped, and there is one in the same task that holds the tempdir.

### DIFF
--- a/config/src/iroha.rs
+++ b/config/src/iroha.rs
@@ -25,8 +25,6 @@ view! {
         pub private_key: PrivateKey,
         /// Disable coloring of the backtrace and error report on panic
         pub disable_panic_terminal_colors: bool,
-        /// Iroha will shutdown on any panic if this option is set to `true`.
-        pub shutdown_on_panic: bool,
         /// `Kura` configuration
         #[config(inner)]
         pub kura: kura::Configuration,
@@ -71,7 +69,6 @@ impl Default for Configuration {
             public_key,
             private_key,
             disable_panic_terminal_colors: bool::default(),
-            shutdown_on_panic: true,
             kura: kura::Configuration::default(),
             sumeragi: sumeragi_configuration,
             torii: torii::Configuration::default(),
--- a/config/src/iroha.rs
+++ b/config/src/iroha.rs
@@ -71,7 +71,7 @@ impl Default for Configuration {
             public_key,
             private_key,
             disable_panic_terminal_colors: bool::default(),
-            shutdown_on_panic: false,
+            shutdown_on_panic: true,
             kura: kura::Configuration::default(),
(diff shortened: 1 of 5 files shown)


config/src/iroha.rs
```diff
@@ -70,7 +70,7 @@ impl Default for Configuration {
             public_key,
             private_key,
             disable_panic_terminal_colors: bool::default(),
-            shutdown_on_panic: false,
+            shutdown_on_panic: true,
```
appetrosyan, line 73: Excellent. I'd remove this option. And approve.

appetrosyan (approved):

appetrosyan (approved):

ilchu (approved):
I'd rather have `expect()` messages telling which condition was violated than what it is that we're currently doing, as it's going to be the panic message on fail, but otherwise LGTM.

PR #2633, opened 2022-08-17
[refactor]: Use context in IntoFfi

mversic:


ffi/src/lib.rs
```diff
@@ -81,8 +76,13 @@ pub trait IntoFfi: Sized {
...
     type Target: ReprC;
 
+    /// Type into which state can be stored during conversion. Useful for returning
+    /// non-owning types but performing some conversion which requires allocation.
+    /// Serves similar purpose as does context in a closure
+    type Store;
+
     /// Convert from [`Self`] into [`Self::Target`]
-    fn into_ffi(self) -> Self::Target;
+    fn into_ffi(self, store: &mut core::mem::MaybeUninit<Self::Store>) -> Self::Target;
```
mversic, line 85: At first I thought that I'll have to use `MaybeUninit`, but I don't, it's just that I'd have to use `Option<T>` in places where T doesn't implement `Default`. So, using `MaybeUninit` is only added as an optimization but it introduces a risk of causing UB if handled improperly. Although I think that such bugs would surface promptly

ffi/src/owned.rs
```diff
@@ -209,26 +212,43 @@ impl<'itm> TryFromReprC<'itm> for &'itm str {
...
-    type Target = LocalSlice<T::Target>;
-
-    fn into_ffi(self) -> Self::Target {
-        self.into_iter().map(IntoFfi::into_ffi).collect()
+impl<'itm, T: IntoFfi, const N: usize> IntoFfi for [T; N] {
+    type Target = *mut T::Target;
+    type Store = ([T::Target; N], [MaybeUninit<T::Store>; N]);
+
+    fn into_ffi(self, store: &mut MaybeUninit<Self::Store>) -> Self::Target {
+        let mut data_store: [MaybeUninit<_>; N] = unsafe { MaybeUninit::uninit().assume_init() };
```
Erigara, line 237: Instant UB

ffi/src/owned.rs
```diff
@@ -209,26 +212,43 @@ impl<'itm> TryFromReprC<'itm> for &'itm str {
...
+        let mut store_store: [MaybeUninit<_>; N] = unsafe { MaybeUninit::uninit().assume_init() };
+
+        self.into_iter().enumerate().for_each(|(i, item)| {
```
Erigara, line 240: Maybe use `zip` here?

ffi/src/owned.rs
```diff
@@ -161,18 +146,36 @@ impl<'itm, T> IntoFfiVec for &'itm T
...
+            core::iter::repeat_with(MaybeUninit::uninit)
+                .take(source.len())
+                .collect(),
+        ));
 
-    fn into_ffi(source: Vec<Self>) -> Self::Target {
-        source.into_iter().map(IntoFfi::into_ffi).collect()
+        store.0 = source
+            .into_iter()
+            .enumerate()
```
Erigara, line 165: I would suggest `zip` if possible.

ffi/src/primitives.rs
```diff
@@ -21,8 +22,8 @@ impl<'itm> TryFromReprC<'itm> for bool {
     type Source = <u8 as TryFromReprC<'itm>>::Source;
     type Store = ();
 
-    unsafe fn try_from_repr_c(source: Self::Source, _: &mut ()) -> Result<Self> {
-        let source: u8 = TryFromReprC::try_from_repr_c(source, &mut ())?;
+    unsafe fn try_from_repr_c(source: Self::Source, store: &mut ()) -> Result<Self> {
```
Erigara, line 25: In this case should we specify `type Store = <u8 as TryFromReprC<'itm>>::Store`?

In other places as well.

ffi/src/lib.rs
```diff
@@ -41,13 +42,8 @@ pub unsafe trait Handle {
...
 pub unsafe trait ReprC: Sized {}
 
-/// Used to do a cheap reference-to-[`ReprC`]-reference conversion
-pub trait AsReprCRef<'itm> {
-    /// Robust C ABI compliant representation of &[`Self`]
-    type Target: ReprC + 'itm;
-
-    /// Convert from &[`Self`] into [`Self::Target`].
-    fn as_ref(&'itm self) -> Self::Target;
+pub trait FfiType {
```
Erigara, line 45: I think if possible we should do:

```rust
trait IntoFfi: FfiType {
    type Target: <Self as FfiType>::ReprC
}

trait TryFromReprC: FfiType {
    type Source: <Self as FfiType>::ReprC
}
```

ffi/src/owned.rs
```diff
@@ -161,18 +146,36 @@ impl<'itm, T> IntoFfiVec for &'itm T
...
+            .map(|(i, item)| IntoFfi::into_ffi(item, &mut store.1[i]))
+            .collect();
+
+        SliceRef::from_slice(&store.0)
```
Erigara, line 169: This mean that `Target` will be valid only for lifetime of `Store`, right?

So for example user can't safely return `Target` from function call if previously allocate `Store` on stack.

mversic: exactly, you're on it. Only types where `Target = ()` are ever allowed to be returned

ffi/src/owned.rs
mversic: no, this is literally from the [documentation](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element)

ffi/src/primitives.rs
mversic: no, why would we? the ownership is returned from the function

ffi/src/owned.rs
Erigara: oh, great, i was confused by: [this examples](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initialization-invariant)

ffi/src/primitives.rs
Erigara: Because we recovering `bool` from `u8` so i think it is reasonable to explicitly state this for `Store` as well since we reuse store from function argument.

ffi/src/lib.rs
appetrosyan: I don't really see how that can be viewed as an optimisation. If it is, I'd like to see the performance impact before signing off.

ffi/src/lib.rs
mversic: this is not valid because `<Self as FfiType>::ReprC` is a type and not a trait that you can use as a bound to `type Target`

Erigara: right, maybe then use this type in fn signature of `into_ffi` and `try_from_repr_c`?

mversic: check it out now

ffi/src/owned.rs
```diff
@@ -45,193 +49,115 @@ pub trait TryFromReprCVec<'slice>: Sized {
...
+
+        self.into_iter().enumerate().for_each(|(i, item)| {
+            data_store[i].write(IntoFfi::into_ffi(item, &mut store.1[i]));
+        });
 
-    fn into_ffi(self) -> Self::Target {
-        self.into_iter().map(IntoFfi::into_ffi).collect()
+        store
+            .0
+            .insert(unsafe { data_store.as_ptr().cast::<[T::ReprC; N]>().read() })
```
mversic, line 160: this should be fine?

ffi/src/lib.rs
```diff
@@ -39,28 +38,42 @@ pub unsafe trait Handle {
...
+/// Indicates that the type can be converted into an FFI compatible representation
+pub trait FfiType: Sized {
+    /// Corresponding FFI-compatible type
+    type ReprC: ReprC;
+}
 
-    /// Convert from &[`Self`] into [`Self::Target`].
-    fn as_ref(&'itm self) -> Self::Target;
+/// Type that can be returned from an FFI function via out-pointer function argument
+pub trait Output: FfiType {
```
mversic, line 50: This trait is tricky, and might be too permissive, I should add tests as a guard

ffi/src/primitives.rs
mversic: isn't `<u8 as TryFromReprC<'itm>>::Store` == `()`. Otherwise, this wouldn't compile

ffi/src/owned.rs
```diff
@@ -45,193 +49,115 @@ pub trait TryFromReprCVec<'slice>: Sized {
...
-// NOTE: Arrays must be converted into pointers to be repr(C)
-unsafe impl<T: ReprC, const N: usize> ReprC for [T; N] {}
-impl<T: IntoFfi, const N: usize> IntoFfi for [T; N] {
-    type Target = LocalSlice<T::Target>;
+impl<T: FfiType, const N: usize> FfiType for [T; N] {
+    type ReprC = *mut T::ReprC;
+}
+
+// NOTE: Arrays must be converted into pointers
+impl<T: IntoFfi, const N: usize> IntoFfi for [T; N]
```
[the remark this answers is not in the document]
mversic: if you still think this is an issue, I'd open an issue

data_model/src/name.rs
```diff
@@ -96,7 +96,7 @@ impl AsRef<str> for Name {
 impl FromStr for Name {
     type Err = ParseError;
 
-    fn from_str(candidate: &str) -> Result<Self, Self::Err> {
+    fn from_str(candidate: &str) -> Result<Self, <Self as FromStr>::Err> {
```
appetrosyan, line 99: Why?

ffi/derive/src/ffi_fn.rs
```diff
@@ -133,9 +133,7 @@ fn gen_body(fn_descriptor: &FnDescriptor) -> syn::Block {
 }
 
 fn gen_out_ptr_arg(arg: &Arg) -> TokenStream {
-    let arg_name = arg.name();
-    let arg_type = arg.ffi_type_resolved(true);
-
+    let (arg_name, arg_type) = (arg.name(), arg.ffi_type_resolved(true));
```
appetrosyan, line 136: This doesn't really make it more readable (in fact, less).

ffi/src/owned.rs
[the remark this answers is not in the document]
appetrosyan: I second opening an issue.

ffi/src/owned.rs
```diff
@@ -40,198 +51,167 @@ pub trait TryFromReprCVec<'slice>: Sized {
...
+//        store.0 = source
+//            .into_iter()
+//            .enumerate()
+//            .map(|(i, item)| IntoFfiVec::into_ffi(item, &mut store.1[i]).0)
+//            .collect();
+//
+//        Local(SliceRef::from_slice(&store.0))
+//    }
+//}
+
```
appetrosyan, line 129: Add issue if commented code is to be merged. Add it to the epic.

ffi/src/lib.rs
```diff
@@ -259,109 +257,143 @@ pub struct Opaque {
...
-                Local::new($ffi_ty($( <$ty as IntoFfi>::into_ffi($ty),)+))
-            }
-        }
+        //        let mut i = 0;
+        //        while let Some((first, rest)) = substore.split_first_mut() {
+        //            let elem: &$ffi_ty<$($ty::ReprC,)*> = &slice[i];
+        //            res.push(Clone::clone(TryFromReprC::try_from_repr_c(elem, first)?));
+        //            substore = rest;
+        //            i += 1;
+        //        }
```
appetrosyan, line 317: Either remove or create issue.

ffi/src/lib.rs
```diff
@@ -372,13 +404,13 @@ macro_rules! impl_tuple {
...
+//impl_tuple! {(A, B, C) -> FfiTuple3}
+//impl_tuple! {(A, B, C, D) -> FfiTuple4}
+//impl_tuple! {(A, B, C, D, E) -> FfiTuple5}
+//impl_tuple! {(A, B, C, D, E, F) -> FfiTuple6}
+//impl_tuple! {(A, B, C, D, E, F, G) -> FfiTuple7}
+//impl_tuple! {(A, B, C, D, E, F, G, H) -> FfiTuple8}
+//impl_tuple! {(A, B, C, D, E, F, G, H, I) -> FfiTuple9}
+//impl_tuple! {(A, B, C, D, E, F, G, H, I, J) -> FfiTuple10}
+//impl_tuple! {(A, B, C, D, E, F, G, H, I, J, K) -> FfiTuple11}
+//impl_tuple! {(A, B, C, D, E, F, G, H, I, J, K, L) -> FfiTuple12}
```
appetrosyan, line 416: No longer needed? It would be nice if we can automate this process.

mversic:
closed in favor of #2625

PR #2634, opened 2022-08-18
[feature] #2611: Derive `TryFromReprC` and `IntoFfi` for generic opaque structs

### LINKED ISSUE #2611: generic structs parametrized with concrete types in FFI
If we take a look at `crypto/src/signature.rs` we can see the `SignatureOf<T>` which is used in our `data_model` API (namely in transactions API). For instance, `ValidTransaction` is defined as:
```rs
pub struct ValidTransaction {
    /// The [`Transaction`]'s payload.
    pub payload: Payload,
    /// [`SignatureOf`] [`Payload`].
    pub signatures: SignaturesOf<Payload>,
}
```

and would have to be turned into:
```rs
#[derive(Getters, IntoFfi, TryFromReprC)]
#[getset(get = "pub")]
#[cfg_attr(
    all(feature = "ffi_export", not(feature = "ffi_import")),
    iroha_ffi::ffi_export
)]
#[cfg_attr(feature = "ffi_import", iroha_ffi::ffi_import)]
pub struct ValidTransaction {
    /// The [`Transaction`]'s payload.
    payload: Payload,
    /// [`SignatureOf`] [`Payload`].
    signatures: SignaturesOf<Payload>,
}
```

which, once the macro is expanded, will generate ffi function with a signature equivalent to the following:
```rs
pub extern "C" fn ValidTransaction__signatures(*const ValidTransaction, output: *mut *const SignaturesOf<Payload>) -> FfiReturn {
    /// fn body
}
```
because unlike `SignaturesOf<T>`, `SignaturesOf<Payload>` is allowed in `extern` functions

While the code for generating the previous FFI function is mostly in place already, it remains to be investigated how to implement `IntoFfi` and `TryFromReprC` derive macros on `SignaturesOf<T>` and alike. The fact that `SignaturesOf<T>` is turned into an opaque pointer might make it possible to derive both of the traits for generic `SignaturesOf<T>`

Erigara:
### Description of the Change

Derive `TryFromReprC` and `IntoFfi` for generic opaque structs.

### Issue

Resolves #2611.

### Benefits

Make possible to define ffi functions with generic struct parametrized with concrete type as argument or return value.

### Possible Drawbacks

None. 

### Usage Examples or Tests

```
cargo test --package iroha_ffi --test generics -- return_generic --exact --nocapture
```

### DIFF
--- a/ffi/derive/src/convert.rs
+++ b/ffi/derive/src/convert.rs
@@ -9,16 +9,13 @@ pub fn derive_try_from_repr_c(input: &DeriveInput) -> TokenStream2 {
     if !matches!(input.vis, syn::Visibility::Public(_)) {
         abort!(input.vis, "Only public items are supported");
     }
-    if !input.generics.params.is_empty() {
-        abort!(input.generics, "Generics are not supported");
-    }
 
     if is_opaque(input) {
         if is_opaque_wrapper(input) {
-            return derive_try_from_repr_c_for_opaque_item_wrapper(&input.ident);
+            return derive_try_from_repr_c_for_opaque_item_wrapper(input);
         }
 
-        return derive_try_from_repr_c_for_opaque_item(&input.ident);
+        return derive_try_from_repr_c_for_opaque_item(input);
     }
 
     match &input.data {
@@ -40,16 +37,13 @@ pub fn derive_into_ffi(input: &DeriveInput) -> TokenStream2 {
     if !matches!(input.vis, syn::Visibility::Public(_)) {
         abort!(input.vis, "Only public items are supported");
     }
-    if !input.generics.params.is_empty() {
-        abort!(input.generics, "Generics are not supported");
-    }
 
     if is_opaque(input) {
         if is_opaque_wrapper(input) {
-            return derive_into_ffi_for_opaque_item_wrapper(&input.ident);
+            return derive_into_ffi_for_opaque_item_wrapper(input);
         }
 
-        return derive_into_ffi_for_opaque_item(&input.ident);
+        return derive_into_ffi_for_opaque_item(input);
     }
 
     match &input.data {
@@ -85,19 +79,26 @@ fn variant_discriminants(enum_: &syn::DataEnum) -> Vec<syn::Expr> {
     })
 }
 
-fn derive_try_from_repr_c_for_opaque_item_wrapper(name: &Ident) -> TokenStream2 {
+fn derive_try_from_repr_c_for_opaque_item_wrapper(input: &DeriveInput) -> TokenStream2 {
+    let name = &input.ident;
+    let mut generics = input.generics.clone();
+    let lifetime: syn::Lifetime = syn::parse_quote!('itm);
+    let lifetimes = &[lifetime.clone()];
+    let (generic_params, ty_generics, where_clause) =
+        add_bounds_and_split_generics(&mut generics, lifetimes, &[]);
+
     let opaque_item_slice_try_from_repr_c_derive =
-        derive_try_from_repr_c_for_opaque_item_slice(name);
-    let opaque_item_vec_try_from_repr_c_derive = derive_try_from_repr_c_for_opaque_item_vec(name);
+        derive_try_from_repr_c_for_opaque_item_slice(input);
+    let opaque_item_vec_try_from_repr_c_derive = derive_try_from_repr_c_for_opaque_item_vec(input);
 
     quote! {
-        impl<'itm> iroha_ffi::TryFromReprC<'itm> for #name {
+        impl #generic_params iroha_ffi::TryFromReprC<#lifetime> for #name #ty_generics #where_clause {
             type Source = *mut iroha_ffi::Opaque;
             type Store = ();
 
             unsafe fn try_from_repr_c(
-                source: <Self as iroha_ffi::TryFromReprC<'itm>>::Source,
-                _: &mut <Self as iroha_ffi::TryFromReprC<'itm>>::Store
+                source: <Self as iroha_ffi::TryFromReprC<#lifetime>>::Source,
+                _: &mut <Self as iroha_ffi::TryFromReprC<#lifetime>>::Store
             ) -> iroha_ffi::Result<Self> {
                 if source.is_null() {
                     return Err(iroha_ffi::FfiReturn::ArgIsNull);
@@ -107,13 +108,13 @@ fn derive_try_from_repr_c_for_opaque_item_wrapper(name: &Ident) -> TokenStream2 
             }
         }
 
-        impl<'itm> iroha_ffi::TryFromReprC<'itm> for &'itm #name {
+        impl #generic_params iroha_ffi::TryFromReprC<#lifetime> for &#lifetime #name #ty_generics #where_clause {
             type Source = *const iroha_ffi::Opaque;
             type Store = Option<#name>;
 
             unsafe fn try_from_repr_c(
-                source: <Self as iroha_ffi::TryFromReprC<'itm>>::Source,
-                store: &'itm mut <Self as iroha_ffi::TryFromReprC<'itm>>::Store
+                source: <Self as iroha_ffi::TryFromReprC<#lifetime>>::Source,
+                store: &#lifetime mut <Self as iroha_ffi::TryFromReprC<#lifetime>>::Store
             ) -> iroha_ffi::Result<Self> {
                 if source.is_null() {
                     return iroha_ffi::FfiReturn::ArgIsNull;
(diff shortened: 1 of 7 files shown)


ffi/derive/src/convert.rs
```diff
@@ -197,16 +205,24 @@ fn derive_try_from_repr_c_for_opaque_item(name: &Ident) -> TokenStream2 {
     }
 }
 
-fn derive_try_from_repr_c_for_opaque_item_slice(name: &Ident) -> TokenStream2 {
+fn derive_try_from_repr_c_for_opaque_item_slice(input: &DeriveInput) -> TokenStream2 {
+    let name = &input.ident;
+    let mut generics = input.generics.clone();
+    let lifetime: syn::Lifetime = syn::parse_quote!('slice);
+    let lifetimes = &[lifetime.clone()];
+    let trait_bounds = &[syn::parse_quote!(core::clone::Clone)];
```
Erigara, line 213: Required because of cloning inside this method.

I was surprised to find out that rust accept smt like `T: Clone + Clone` [just fine](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b7455998d82a3e368db1c731831f8776).

So we shouldn't worry about this.

My another concern is lifetime name collisions so maybe we should give our lifetimes more obfuscated name.

ffi/tests/generics.rs
```diff
@@ -0,0 +1,32 @@
...
+use iroha_ffi::{ffi_export, IntoFfi, TryFromReprC};
+
+#[derive(Clone, Copy, IntoFfi, TryFromReprC, PartialEq, Eq, Debug)]
+pub struct GenericFfiStruct<T>(T);
+
+#[derive(IntoFfi, TryFromReprC, Clone, Copy, Getters)]
+#[getset(get = "pub")]
+#[ffi_export]
+pub struct FfiStruct {
+    inner: GenericFfiStruct<bool>,
```
mversic, line 15: can you add a test freestanding function as well. Like: 
```rs
fn freestanding(input: GenericFfiStruct<bool>) -> GenericFfiStruct<bool>) {
    input
}
```

ffi/tests/generics.rs
```diff
@@ -0,0 +1,51 @@
...
+        FfiStruct__inner(ffi_struct.into_ffi(), output.as_mut_ptr());
+        assert_eq!(
+            TryFromReprC::try_from_repr_c(output.assume_init(), &mut ()),
+            Ok(&ffi_struct.inner)
+        );
+    }
+}
+
+#[test]
+fn freestanding_accept_and_return_generic() {
```
mversic, line 40: while you're at it, could you add a compile test confirming that generics are not supported on freestanding functions or methods?

mversic: if you think this is something compiler will always refuse, you could say we don't need this test

ffi/derive/src/convert.rs
```diff
@@ -490,3 +529,30 @@ fn is_opaque_wrapper(input: &DeriveInput) -> bool {
...
+        .lifetimes_mut()
+        .for_each(|lt| lt.bounds.extend(lifetimes.iter().cloned()));
+    generics.type_params_mut().for_each(|param| {
+        param
+            .bounds
+            .extend(lifetimes.iter().cloned().map(syn::TypeParamBound::from));
+        param
+            .bounds
+            .extend(trait_bounds.iter().cloned().map(syn::TypeParamBound::from));
+    });
```
mversic, line 552: so you need the `<'itm>` lifetime bound on every generic parameter?

ffi/derive/src/convert.rs
```diff
@@ -490,3 +529,30 @@ fn is_opaque_wrapper(input: &DeriveInput) -> bool {
...
+    syn::TypeGenerics<'generics>,
+    Option<&'generics syn::WhereClause>,
+) {
+    generics
+        .lifetimes_mut()
+        .for_each(|lt| lt.bounds.extend(lifetimes.iter().cloned()));
+    generics.type_params_mut().for_each(|param| {
+        param
+            .bounds
+            .extend(lifetimes.iter().cloned().map(syn::TypeParamBound::from));
```
mversic, line 548:
```suggestion
        param.bounds.extend(parse_quote! {#lifetimes});
```

if you could do it this way, I think it'll be more readable

mversic (approved):

ffi/derive/src/convert.rs
Erigara: @mversic, I think so, `TryFromReprC<'itm>: Sized + 'itm` so if `Type: TryFromRepr<'itm>` then `Type: 'itm` which should imply that every generic parameter of `Type` should live for at least `'itm`.

Small [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0a3d2a2ee833a5d8992033054171de4d).

mversic: yes, it's good as is, you can also set a bound on the whole type and the bound will work recursively on inner types. Let me save you the trouble [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9feab412c1eb20f0842c9b8e184125d)

ffi/derive/src/convert.rs
mversic: > My another concern is lifetime name collisions so maybe we should give our lifetimes more obfuscated name.

Definitely. Something like `__iroha_ffi_itm`

mversic (approved):

ffi/derive/src/convert.rs
Erigara: wow, cool! i don't know that this will work :) this allow write function in really clear way

mversic (approved):

ffi/derive/src/convert.rs
```diff
@@ -85,19 +79,26 @@ fn variant_discriminants(enum_: &syn::DataEnum) -> Vec<syn::Expr> {
...
 
-fn derive_try_from_repr_c_for_opaque_item_wrapper(name: &Ident) -> TokenStream2 {
+fn derive_try_from_repr_c_for_opaque_item_wrapper(input: &DeriveInput) -> TokenStream2 {
+    let name = &input.ident;
+    let mut generics = input.generics.clone();
+    let lifetime: syn::Lifetime = syn::parse_quote!('itm);
+    let lifetimes = &[lifetime.clone()];
+    let (generic_params, ty_generics, where_clause) =
+        add_bounds_and_split_generics(&mut generics, lifetimes, &[]);
+
```
ilchu, line 89: Seeing that we have so many repeating code makes me itch to refactor `input` to be some custom struct instead of `DeriveInput` just like I did in `config`, but I understand that it's definitely outside of scope here.

ffi/derive/src/util.rs
```diff
@@ -231,13 +231,13 @@ fn gen_derived_method_sig(field: &syn::Field, derive: Derive) -> syn::Signature
 
     match derive {
         Derive::Setter => parse_quote! {
-            #field_vis fn #method_name(&mut self, #field)
+            fn #method_name(&mut self, #field)
```
ilchu, line 234: So I see that visibility is not present in `syn::Signature` and thus it's not needed, but I wonder how it typechecked before

ilchu:
Would probably make sense to document somehow that `#[ffi_export]` can accept generics, but monomorphized only, but I'm not yet aware where such description could fit.

ilchu (approved):
LGTM overall, let's merge

ffi/derive/src/util.rs
Erigara: So as we discussed with @mversic in pm, before we get lucky because all fields with getters/setters was private and thus there `#field_vis` was `None`.
A accidentally noticed that when wrote test with `pub` field and getter.

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
Erigara: I agree with you that at some point we might consider such refactoring, but problem is that we work with broader set of items here.
NOTE: while it's more convenient to code with the approach you took in the configuration, I've noticed one downside of `syn` panic messages are pretty bad and give little or no information about what went wrong.

Erigara:
> Would probably make sense to document somehow that `#[ffi_export]` can accept generics and monomorphized only, but I'm not yet aware where such description could fit.

@ilchu, i agree that good documentation won't hurt, however work on ffi conversion lib is kinda in progress, so maybe we can document supported items when lib would be more stable. 
What do you think?

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
ilchu: > @ilchu, i agree that good documentation won't hurt, however work on ffi conversion lib is kinda in progress, so maybe we can document supported items when lib would be more stable. What do you think?
Certainly, I'm not pushing for it now, just thinking for posterity.

ffi/derive/src/util.rs
ilchu: Maybe something like property testing could have prevented that from happening in the first place. As @appetrosyan mentioned, we're in high need of better testing harnesses overall and for macros especially.

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
ilchu: Yeah, and there's always the overhead of thinking about spans and macro hygiene too. At some point the cognitive load in macro land gets really taxing.

ilchu:
> @ilchu, i agree that good documentation won't hurt, however work on ffi conversion lib is kinda in progress, so maybe we can document supported items when lib would be more stable. What do you think?

Certainly, I'm in no way pushing for it right now, just thinking towards posterity.

ilchu (approved):

ffi/derive/src/util.rs
appetrosyan: Proptest can't test macros directly. Unfortunately it would have been of little use here.

appetrosyan (approved):

Issue #2561, opened 2022-07-28
Reintroduce viewchanges to consensus


SamHSmith:
This is what `network_topology::apply_block` looks like.
```
    pub fn apply_block(&mut self, block: HashOf<VersionedCommittedBlock>) {
        /*
        nocheckin
        *self = self
            .clone()
            .into_builder()
            .at_block(block)
            .with_view_changes(ViewChangeProofs::empty())
            .build()
        .expect("Given a valid Topology, it is impossible to have error here.")
        */
    }
```
This was done in order to be able to produce many blocks without supporting view changes yet. The things to implement are view change after a commit timeout. And a view change if the leader is offline or is ignoring transactions. Transaction forwarding reciepts also need implementing. These are things that exist in the old sumeragi implentation. Use that as a guide. Start by uncommenting the above code and then get to work at unbreaking the tests by implemting the above.

SamHSmith:
I suspect the best way to implement this is to add a new "view change requested" message. Essentially build a parallel system that runs in the same event loop to coordinate view changes. If there is consensus that a view change should be initiated, reset all the state to how it was when entering the current round, but with a shuffled network topology.

Issue #2548, opened 2022-07-27
[BUG]: Zombie Genesis


appetrosyan:
Under high load conditions, and given high packet loss, genesis application might not finish, or go through partially. 

In this state Iroha does not have a valid genesis. It looks like a `default` or rather empty genesis is at play. 

We need to reduce the incidence of these situations, and panic when they occur

Issue #2429, opened 2022-07-04
Kura IO fails, what should be done?


SamHSmith:
What assumptions should we make about the storage medium Kura uses? Is it reliable, is it always online? These are important questions that impact the design of kura and by extension the whole program.

In my opinion the correct and simplest way forward is this. We assume that IO is always available and reliable. If that is detected to not be the case, aka any read or write fails, we shutdown gracefully. This way our code becomes vastly simpler because we don't have to handle a vast number of cases. We only need to handle two cases, IO works as expected or it doesn't. If it doesn't we can't run iroha.

The other alternative is that we can try to handle these errors. In the case that someone unmounts the drive containing the block store, we don't simply crash. Instead we start buffering new blocks in memory hoping that the drive will get remounted and our writes will start succeeding again. This is just one example of the things we would have to do and the complications that ensue.

I think this is a premature optimization of sorts because almost all usecases of iroha can assume correctly functioning hardware. Our time is better spent elsewhere than handling hypothetical errors.

appetrosyan:
Do full error handling.

Issue #1170, opened 2021-07-15
SUMERAGI: Handle Soft Forks


takemiyamakoto:
Separate finalized WSV with pending WSV.

Un-apply blocks that are replaced with proofs of cancellation.

takemiyamakoto:
Soft forks can happen when the proxy tail sends a fully signed block to only one peer, for example, then that peer goes offline while other peers do a view change and replace the already committed local block. Then when the peer comes back online, it should replace their committed block with a newly created block that supercedes it.

takemiyamakoto:
NOTE: because we cancel out overridden block hashes, we know what block(s) should be replaced. Finalization happens after 1 block, so max depth of soft fork is just 1 block.

PR #2636, opened 2022-08-18
[fix] #2561: Reintroduce viewchanges to consensus.

SamHSmith:
### Description of the Change

Viewchanges are reintroduced without transaction receipts. This method does diverge slightly from the whitepaper, but I think it is warrented as no malicious node would ever send you a proof of the fact that they are censoring you. And in the common case on a real world network, most nodes have transactions to test against the leader, so if a block is not created they can all very quickly coordinate a view change because they will all be suspicious of the leader.

### Issue

#2561 

### Benefits
This version is simpler, faster, and more robust.

### Possible Drawbacks
In this implementation we do not track who to blame for a failed round. This can be added though.
The test for time triggers fails on my machine, I don't know why though. Probably has to do with this implementation producing blocks at a different pace.

I've enabled trace level logging on the tests. Use the following to observe execution and build an understanding. 
```
cargo test integration::unstable_network::unstable_network_4_peers_1_fault -- --nocapture | grep sumeragi
```


takemiyamakoto:
The use was with the receipts I had in mind was not for other nodes but for clients to have a proof of blame that they could submit to other nodes if their tx is not going through

SamHSmith:
In order to trigger a view change or to avoid censorship? My thinking was this. We change leader each round in order to avoid long term censorship. f+1 nodes are required in order to suspect the leader and cause a view change. It seems quite likely that this will always be the case in an active network. If it's an inactive network where less than n+1 nodes have transactions, transaction gossip will soon enough cause n+1 nodes to have transactions and a view change will be triggered.

It seems unlikely to me that there would ever exist a node that *would* send you a receipt but *not* create a block. It's seems like the game theory would suggest that nodes either send you receipts and create blocks or completely ignore your messages.

takemiyamakoto:
If a node is proven to be bad, it should be moved to set B and not used unless set A doesn't have enough nodes, in which case set B gets put into order from B's head to tail into set A. That is why we prove a give node is bad because otherwise we cannot move it to set B.

SamHSmith:
Neither this implentation or the old one did any sort of intelligent choosing of nodes in set a. They just shuffle the sets around repeatedly until a block passes consensus. We can still prove a given node is bad without receipts. And we can add the system you describe down the line.

appetrosyan:
I'm afraid I'll have to intrude into this discussion. 

1. Regarding the use of receipts for bad nodes, I believe that the original implementation was not particulalry faithful. This implementaion is also not exactly following the letter of the whitepaper either, but it can be modified into one that uses the information. I believe that this is what @SamHSmith is proposing. 

2. We have had discussions previously about both a gossiping reputation system and a separate voting reputation system, both of which are meant to improve the average case performance (reduce the amount of gossiping instances and the number of reshuffling steps that result in a valid block). 

It seems that we're all in agreement that such a system needs to be designed and discussed separately, carefully, and proven to work as well if not better than the original implementation, before being upstreamed. 

I'm personally in favour of a monte-carlo simulation to benchmark the different approaches, since asymptotic behaviour is often misleading. 

3. This Pull request is very much a work in progress to be merged into our staging branch and subjected to rigorous testing. This is far from a final review of the consensus modifications, and is merely meant to raise such questions, not necessarily answer them.

takemiyamakoto:
> Neither this implentation or the old one did any sort of intelligent choosing of nodes in set a. They just shuffle the sets around repeatedly until a block passes consensus. We can still prove a given node is bad without receipts. And we can add the system you describe down the line.

Set A should just be shuffled around. But what is necessary is to find bad nodes to push out into set B.

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
+        let signature = Signature::new(key_pair, self.signature_payload().as_ref())?;
+        self.signatures.push(signature);
+        Ok(())
     }
 
-    /// Adds verified signatures of `other` to self.
-    pub fn merge_signatures(&mut self, other: SignaturesOf<Proof>) {
-        self.signatures
-            .extend(other.into_verified_by_hash(&self.hash()))
+    /// Verifies and adds signatures of other to this proof.
```
outoftardis, line 53:
```suggestion
    /// Verify the signatures of `other` and add them to this proof.
```

core/src/sumeragi/fault.rs
```diff
@@ -1033,6 +973,121 @@ pub fn run_sumeragi_main_loop<F>(
...
+                            if !tx_maybe.is_ok() {
+                                continue;
+                            }
+                            let tx = tx_maybe.unwrap();
+                            match sumeragi.queue.push(tx, &wsv_guard) {
+                                Err((_, crate::queue::Error::InBlockchain)) | Ok(()) => {}
+                                Err((_, err)) => {
+                                    warn!(
+                                        ?err,
+                                        "Failed to push #[inline]to queue gossiped transaction."
```
outoftardis, line 1001:
```suggestion
                                        "Failed to push #[inline] to queue gossiped transaction."
```

cli/src/torii/routing.rs
```diff
@@ -361,28 +357,28 @@ mod subscription {
...
-                .last()
-                .expect("At least genesis should always exist")
-                .value()
-                .version()
-                .to_string(),
-    )
-         */
-    warp::reply::json(b"Not implemented")
+
+    let wsv = sumeragi.wsv_clone();
```
appetrosyan, line 361: There's a small discussion to be had about how `wsv.clone()` actually lets you do the same. Interior mutability of `wsv` means that we might as well make `wsv` a public member, given that any getter can use this to clone and to modify.

cli/src/torii/routing.rs
```diff
@@ -361,28 +357,28 @@ mod subscription {
...
+
+    #[allow(clippy::expect_used)]
+    let string = wsv
+        .blocks()
+        .last()
+        .expect("At least genesis should always exist")
```
appetrosyan, line 367: Not necessarily true #2548

cli/src/torii/routing.rs
```diff
@@ -361,28 +357,28 @@ mod subscription {
...
 }
 
 #[cfg(feature = "telemetry")]
 async fn handle_metrics(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) -> Result<String> {
-    /*
-        update_metrics(&wsv, network).await?;
-        wsv.metrics.try_to_string().map_err(Error::Prometheus)
-    */
-    Ok("".to_owned())
+    sumeragi.update_metrics(network);
```
appetrosyan, line 376: This is no longer `async`. Have you benchmarked with and without?

core/src/kura.rs
```diff
@@ -168,6 +171,12 @@ impl Kura {
                 }
             };
 
+            if failed_to_store_block.is_some() {
+                let error = failed_to_store_block_reason.unwrap();
```
appetrosyan, line 175:
```suggestion
                let error = failed_to_store_block_reason.expect("Already know that failure occurred");
```

Which is also why we might want to reconsider the structure. We're checking the failure condition twice.

core/src/kura.rs
```diff
@@ -168,6 +171,12 @@ impl Kura {
...
+                error!(%error, "Failed to write block. Kura will now init again.");
```
appetrosyan, line 176: OK, if we really are going to retry the `init` step, then you can end up in a boot loop, which is terrible. Rather I'd like to have a constant that sets the retry count, tries 1-2 times and panics with an error message.

core/src/kura.rs
```diff
@@ -201,8 +210,16 @@ impl Kura {
                                 .push(block_hash);
                         }
                         Err(error) => {
-                            error!(%error, "Failed to write block. Kura will now init again.");
-                            panic!("It is unclear what we should do here.");
+                            failed_to_store_block = Some(serialized_block);
+                            failed_to_store_block_reason = Some(error);
+                            std::thread::sleep(std::time::Duration::from_millis(250));
```
appetrosyan, line 215: This is extremely weird, and we should explore better options.

core/src/kura.rs
```diff
@@ -201,8 +210,16 @@ impl Kura {
...
+                            // Why is this sleep needed?
+                            // Well in many of our tests we have kura operate within temp directories.
+                            // When Iroha is shutdown very fast these temp directories get free'd before
+                            // this thread can exit, causing a panic due to faulty I/O. This pushing of
+                            // the pushing to the next round makes sure that we don't panic if we were going
+                            // to exit anyway. The sleep is also required to make sure that the exit code
+                            // has finished running before the next iteration of the kura thread loop.
```
appetrosyan, line 222: This seems to be a problem with how we handle shutdown, and to me it seems that a fixed timeout isn't going to help much. 

Firstly, our tests no longer continue running on panic. Have a look at #2630. This is probably what we should do here more-less. 

Secondly, instead of expecting the exit code to finish running in a fixed amount of time, there should be a signal that the exit code finished running, not an expectation that by this time it would have.

core/src/sumeragi/message.rs
```diff
@@ -17,7 +17,7 @@ use tokio::task;
 
 use super::{
     fault::{FaultInjection, SumeragiWithFault},
-    view_change::{self, Proof},
+    view_change::{self, Proof, ProofChain},
```
appetrosyan, line 20:
```suggestion
view_change::{self, Proof, ProofChain as _},
```

core/src/sumeragi/message.rs
```diff
@@ -99,18 +97,13 @@ pub enum Message {
...
 }
 
 impl ViewChangeSuggested {
     /// Constructor.
-    pub const fn new(
-        proof: view_change::Proof,
-        chain: view_change::ProofChain,
-    ) -> ViewChangeSuggested {
-        Self { proof, chain }
+    pub const fn new(proofs: Vec<view_change::Proof>) -> ViewChangeSuggested {
```
appetrosyan, line 105: inline.

core/src/sumeragi/network_topology.rs
```diff
@@ -163,38 +159,31 @@ impl Builder {
...
-        self
-    }
-
     /// Build and get topology.
     ///
     /// # Errors
     /// 1. Required field is omitted.
     /// 2. No peer exists.
-    pub fn build(self) -> Result<Topology> {
+    pub fn build(self, number_of_view_changes_this_round: u64) -> Result<Topology> {
```
appetrosyan, line 167: When you do something like this, you're emphasising that the information is important to the network topology. Keep it up.

core/src/sumeragi/network_topology.rs
```diff
@@ -221,35 +208,27 @@ impl Topology {
...
         *self = self
             .clone()
             .into_builder()
             .at_block(block)
-            .with_view_changes(ViewChangeProofs::empty())
-            .build()
-        .expect("Given a valid Topology, it is impossible to have error here.")
-        */
+            .build(0)
+            .expect("Given a valid Topology, it is impossible to have error here.");
```
appetrosyan, line 222:
```suggestion
            .expect("Unreachable. Valid topology implies no error.");
```

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
-    ) -> Result<Self> {
-        let payload = ProofPayload {
-            reason: Reason::CommitTimeout(CommitTimeout { hash }),
-            previous_proof,
-            latest_block,
-        };
-        Self::from_payload(payload, key_pair)
-    }
+    fn signature_payload(&self) -> Hash {
+        use byteorder::{ByteOrder, LittleEndian};
```
appetrosyan, line 26: Please avoid having `use` statements inside functions. If the thing that you're using is local, you can rely on explicit namespacing for the thing, and if it's used in more than one function, it should be an import at the top.

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
-            reason: Reason::NoTransactionReceiptReceived(NoTransactionReceiptReceived),
-            previous_proof,
-            latest_block,
-        };
-        Self::from_payload(payload, key_pair)
+        let mut buf = [0_u8; Hash::LENGTH + std::mem::size_of::<u64>()];
+        let prev_block_hash = self.latest_block_hash.as_ref();
+        for i in 0..Hash::LENGTH {
+            buf[i] = prev_block_hash[i];
+        }
```
appetrosyan, line 32: All of this can be done without `mut`, if you use `iter()`, `cloned()` and `take(Hash::LENGTH + std::mem::size_of::<u64>()`.

appetrosyan: The added benefit is that you can then use something like `rayon` to do it using SIMD (the Rust compiler isn't smart-enough to do it that way every time).

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
+    /// Verifies and adds signatures of other to this proof.
+    pub fn merge_signatures(&mut self, other: &Vec<Signature>) {
+        let signature_payload = self.signature_payload();
+        for signature in other {
+            if signature.verify(signature_payload.as_ref()).is_ok()
+                && !self.signatures.contains(signature)
+            {
+                self.signatures.push(signature.clone());
+            }
+        }
```
appetrosyan, line 62: To me it seems that this is a library function.

core/src/sumeragi/view_change.rs
```diff
@@ -108,249 +68,87 @@ impl Proof {
...
-impl ProofChain {
-    /// Initialize an empty proof chain.
-    #[must_use]
-    pub const fn empty() -> Self {
-        Self { proofs: Vec::new() }
-    }
-
-    /// Verify the view change proof chain.
-    pub fn verify_with_state(
+impl ProofChain for Vec<Proof> {
```
appetrosyan, line 108: OK. I see you found your legs w.r.t. traits.

cli/src/torii/routing.rs
```diff
@@ -251,8 +249,6 @@ async fn stream_blocks(
     wsv: &WorldStateView,
     stream: &mut WebSocket,
 ) -> eyre::Result<()> {
-    /*
-    TODO REDO
```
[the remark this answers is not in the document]
SamHSmith: You are correct.

cli/src/torii/routing.rs
```diff
@@ -223,26 +223,24 @@ async fn handle_post_configuration(
...
-        stream_blocks(&mut from_height, &wsv, &mut stream).await?;
+    let mut rx = wsv.subscribe_to_new_block_notifications();
+    stream_blocks(&mut from_height, &wsv, &mut stream).await?;
 
-        loop {
-            rx.changed().await?;
-            stream_blocks(&mut from_height, wsv, &mut stream).await?;
+    loop {
+        rx.changed().await?;
+        stream_blocks(&mut from_height, &wsv, &mut stream).await?;
```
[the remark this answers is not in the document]
SamHSmith: Yes. The World State View is the single largest collection of data in iroha2. Improving how we manipulate it is essential for producing more efficient software. It's the first D in Data Oriented Design.

cli/src/torii/routing.rs
SamHSmith: It's the metrics. What is there to benchmark? How many metrics per second we can produce?

core/src/sumeragi/network_topology.rs
SamHSmith: good suggestion.

core/src/kura.rs
SamHSmith: change accepted

core/src/kura.rs
SamHSmith: This is a problem with RAII and our tests. The temporary filesystem is destroyed at the same time as we shutdown iroha. In production this is not an issue. If we want to solve this properly there should be a iroha.shutdown() method that tells all the threads to exit and waits their termination. Then each test can call this at the end. Shall I create an issue for it in the "Polishing Kura" epic?

core/src/sumeragi/message.rs
SamHSmith: Why is this better?

core/src/sumeragi/view_change.rs
SamHSmith: Here's what when through my head when I wrote this code.
1. I need a buffer containing the hash, followed by the view change index in an endian specific order.
2. I will declare a buffer. Copy the hash in to it. Then write the index in little endian at the end.
3. Now write the dumbest code that does exactly that, a loop. Because dumb code is often fast code.

I will change to `copy_from_slice`. Because it will indeed get optimized into a memcpy. But notice what the implentation of `copy_from_slice` looks like.
```Rust
    default fn spec_clone_from(&mut self, src: &[T]) {
        assert!(self.len() == src.len(), "destination and source slices have different lengths");
        // NOTE: We need to explicitly slice them to the same length
        // to make it easier for the optimizer to elide bounds checking.
        // But since it can't be relied on we also have an explicit specialization for T: Copy.
        let len = self.len();
        let src = &src[..len];
        for i in 0..len {
            self[i].clone_from(&src[i]);
        }
    }
```
It's control flow and basic data ops all the way down. Even though the function calls often disguise that fact.

core/src/sumeragi/view_change.rs
SamHSmith: Is it really that bad to have an array of something?

core/src/sumeragi/view_change.rs
SamHSmith: yes

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
     /// This way peers vote for changing the view - changing the roles of peers.
     ///
     /// # Errors
     /// Can fail during creation of signature
-    pub fn sign(mut self, key_pair: KeyPair) -> Result<Self> {
-        let signature = SignatureOf::new(key_pair, &self.payload)?.transmute();
-        self.signatures.insert(signature);
-        Ok(self)
+    pub fn sign(&mut self, key_pair: KeyPair) -> Result<()> {
+        let signature = Signature::new(key_pair, self.signature_payload().as_ref())?;
```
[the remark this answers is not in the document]
SamHSmith: And why is it good to keep `Signature::new` private? It fits my usage better so there is certainly reason for it to be public.

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
-        };
-        Self::from_payload(payload, key_pair)
+        let mut buf = [0_u8; Hash::LENGTH + std::mem::size_of::<u64>()];
+        let prev_block_hash = self.latest_block_hash.as_ref();
+        for i in 0..Hash::LENGTH {
+            buf[i] = prev_block_hash[i];
+        }
+        {
+            let mut final_buf_segment = &mut buf[Hash::LENGTH..];
+            LittleEndian::write_u64(final_buf_segment, self.view_change_index);
```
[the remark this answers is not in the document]
SamHSmith: good point, good suggestion. Less dependencies is better.

core/src/sumeragi/view_change.rs
```diff
@@ -16,90 +16,50 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
+        }
+        // Now we hash the buffer to produce a payload that is completely
+        // different between view change proofs in the same sumeragi round.
+        Hash::new(buf)
```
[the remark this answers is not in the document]
SamHSmith: No. I think that confuses the concepts very much. The hash of a datastructure should be the hash of the actual data in that datastructure.

core/src/kura.rs
```diff
@@ -201,8 +210,16 @@ impl Kura {
                                 .push(block_hash);
                         }
                         Err(error) => {
-                            error!(%error, "Failed to write block. Kura will now init again.");
-                            panic!("It is unclear what we should do here.");
+                            failed_to_store_block = Some(serialized_block);
```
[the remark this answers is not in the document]
SamHSmith: tracking issue for this is #2429

core/src/sumeragi/network_topology.rs
SamHSmith: You say "every time" like our jobs are mostly about typing. You want to write the code in such a way that maximises the performance of engineers who have to work on this code later. Forcing the concept that view-changes and topology are linked concepts is therefore good.

core/src/sumeragi/message.rs
SamHSmith: noted

core/src/block.rs
```diff
@@ -212,8 +212,7 @@ impl PendingBlock {
         self,
         height: u64,
         previous_block_hash: HashOf<VersionedCommittedBlock>,
-        view_change_proofs: ViewChangeProofs,
-        invalidated_blocks_hashes: Vec<HashOf<VersionedValidBlock>>,
```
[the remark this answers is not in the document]
SamHSmith: Because they aren't used. It might be so that I have to reintroduce it when adding handling for soft-forks, #1170. I didn't want to assume an a priori solution. You could for example allow the latest block to be swapped for another valid block then lean on block sync to bring lost peers back on to the main chain.

core/src/kura.rs
```diff
@@ -156,6 +156,9 @@ impl Kura {
 
     #[allow(clippy::expect_used, clippy::cognitive_complexity, clippy::panic)]
     fn kura_recieve_blocks_thread(weak_kura_reference: &Weak<Kura>) {
+        let mut failed_to_store_block = None;
+        let mut failed_to_store_block_reason = None;
+
```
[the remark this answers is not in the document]
SamHSmith: I will change it.

core/src/sumeragi/network_topology.rs
Erigara: I'm not really get your point.

core/src/block.rs
```diff
@@ -752,7 +735,7 @@ impl VersionedCommittedBlock {
             previous_block_hash: *previous_block_hash,
             transactions_hash,
             rejected_transactions_hash,
-            invalidated_blocks_hashes: invalidated_blocks_hashes.into_iter().map(|h| *h).collect(),
+            invalidated_blocks_hashes: Vec::new(),
```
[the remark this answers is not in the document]
SamHSmith: I didn't want to make changes in the data model. It might be so that we start using this again. It has to do with soft-forks.

core/src/sumeragi/view_change.rs
```diff
@@ -108,249 +68,87 @@ impl Proof {
...
-                latest_block,
-                key_pair_1.clone(),
-            )?
-            .sign(key_pair_2.clone())?;
-            proof_chain.push(proof);
+        let next_unfinished_view_change = self.verify_with_state(peers, max_faults, latest_block);
+        if new_proof.view_change_index != (next_unfinished_view_change as u64) {
+            return; // We only care about the current view change that may or may not happen.
+        }
+        self.truncate(next_unfinished_view_change + 1);
```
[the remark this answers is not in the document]
SamHSmith: Truncate does that internally. And I've used it more as an assertion. "From now on self.len() is <= next_unfinished_view_change + 1"

core/src/sumeragi/fault.rs
```diff
@@ -1033,6 +973,121 @@ pub fn run_sumeragi_main_loop<F>(
                 },
             };
 
+            if maybe_incoming_message.is_some() {
```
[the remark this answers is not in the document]
SamHSmith: good suggestion.

core/src/sumeragi/fault.rs
```diff
@@ -1033,6 +973,121 @@ pub fn run_sumeragi_main_loop<F>(
...
+                let peer_list = state_machine_guard
+                    .current_topology
+                    .sorted_peers()
+                    .iter()
+                    .cloned()
+                    .collect();
+                match stolen_message {
+                    Message::TransactionGossip(tx_gossip) => {
+                        for transaction in tx_gossip.txs {
+                            let tx_maybe = VersionedAcceptedTransaction::from_transaction(
```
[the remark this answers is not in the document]
SamHSmith: This one I think is a bit different because I think it's best to handle the error case first. A early return becomes confusing if it's written at the end in my opinion.

Issue #2616, opened 2022-08-12
Modify `LoadFromEnv` to work on `_Proxy` types


ilchu:
For compatibility, currently the whole `Configurable` derive and all the related trait impls are put onto the parent `Configuration` struct instead of the automatically generated `_Proxy` one, and that needs to be changed.

PR #2639, opened 2022-08-20
[feature] #2586, #2616: `Builder` and `LoadFromEnv` for proxy structs

### LINKED ISSUE #2586: `Builder` trait for proxy config
Verify that `_Proxy` types build properly into normal `Configuration` types via the `Builder` trait, failing when expected. Would probably need some tests too.

ilchu:

-------

### Description of the Change
Proxy structs in the `Combine` derive macro now also derive `LoadFromEnv`, as well
as have their `Builder` implementation fixed.

### Issue
Resolves #2586 and #2616.

### Benefits
Getting closer to the aims of the corresponding epic.

### Possible Drawbacks
Implementations still untested. Though they typecheck, thorough foolproofing
is a matter of future work.

### DIFF
--- a/config/base/derive/src/proxy.rs
+++ b/config/base/derive/src/proxy.rs
@@ -1,48 +1,40 @@
 use proc_macro::TokenStream;
 use quote::{format_ident, quote};
-use syn::{Ident, Type, TypePath};
+use syn::{parse_quote, Type, TypePath};
 
 use super::utils::{get_inner_type, StructWithFields};
+use crate::utils;
 
-pub fn impl_combine(ast: &StructWithFields) -> TokenStream {
-    let StructWithFields {
-        ident: parent_name,
-        vis,
-        generics,
-        ..
-    } = ast;
+pub fn impl_combine(ast: StructWithFields) -> TokenStream {
+    let parent_load_env_impl = impl_load_env(&ast);
+    let parent_name = &ast.ident.clone();
 
-    let proxy_name = format_ident!("{}Proxy", parent_name);
-
-    let load_env_impl = impl_load_env(parent_name, ast);
-    let disk_impl = impl_load_from_disk(&proxy_name);
-    let proxy_fields = gen_proxy_struct_fields(ast);
-    let build_impl = impl_build(&proxy_name, ast);
+    let proxy_struct = gen_proxy_struct(ast);
+    let build_impl = impl_build(parent_name, &proxy_struct);
+    let disk_impl = impl_load_from_disk(&proxy_struct);
+    let proxy_load_env_impl = impl_load_env(&proxy_struct);
 
     quote! {
         /// Proxy configuration structure to be used as an intermediate
         /// for config loading
         #[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
-        #vis struct #proxy_name #generics {
-            #proxy_fields
-        }
+        #proxy_struct
 
-        #load_env_impl
+        #parent_load_env_impl
+        #proxy_load_env_impl
         #disk_impl
         #build_impl
     }
     .into()
 }
 
-fn impl_load_env(struct_name: &Ident, ast: &StructWithFields) -> proc_macro2::TokenStream {
+fn impl_load_env(ast: &StructWithFields) -> proc_macro2::TokenStream {
     let set_field = ast.fields
         .iter()
         .map(|field| {
             let ty = &field.ty;
             let l_value = &field.lvalue_write;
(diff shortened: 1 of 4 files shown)


config/base/derive/src/view.rs
```diff
@@ -83,6 +83,7 @@ mod gen {
                 }
             });
         remove_attr_struct(&mut ast, "view");
+        remove_attr_struct(&mut ast, "config");
```
Erigara, line 86: why do we need to clear `config` attribute here?

config/base/derive/src/utils.rs
```diff
@@ -197,6 +197,7 @@ pub struct StructWithFields {
     pub env_prefix: String,
     pub vis: syn::Visibility,
     _struct_token: Token![struct],
+    pub parent_ident: Option<Ident>,
```
Erigara, line 200: I think we should keep this struct single purpose (smt we receive as macro input).

config/base/derive/src/proxy.rs
```diff
@@ -140,30 +116,76 @@ fn impl_load_from_disk(proxy_name: &Ident) -> proc_macro2::TokenStream {
...
+    utils::remove_attr_struct(&mut ast, "config");
+    ast
+}
+
+// REVIEW: need lvalue here as well?
+fn impl_build(ast: &StructWithFields) -> proc_macro2::TokenStream {
+    let checked_fields = gen_none_fields_check(ast);
+    let proxy_name = &ast.ident;
+    #[allow(clippy::expect_used)]
+    let parent_name = ast
```
Erigara, line 152: I would suggest to add `derive(Builder)` macro with macro attribute like `#[builer(parent = ...)]`, this way we avoid adding new fields to `StructWithFields`.

It should be pretty simple change since `impl_build` already recieves full `StructWithField`.

config/base/derive/src/proxy.rs
```diff
@@ -140,30 +116,76 @@ fn impl_load_from_disk(proxy_name: &Ident) -> proc_macro2::TokenStream {
...
+        // as it breaks proxy deserialization
+        utils::remove_attr(&mut field.attrs, "serde");
     });
+    ast.parent_ident = Some(ast.ident.clone());
+    ast.ident = format_ident!("{}Proxy", ast.ident);
+    utils::remove_attr_struct(&mut ast, "config");
+    ast
+}
+
+// REVIEW: need lvalue here as well?
```
Erigara, line 147: Could you explain what you mean here?

config/base/derive/src/utils.rs
ilchu: Fair enough, maybe I overoptimised thoughtlessly.

config/base/derive/src/proxy.rs
ilchu: Or instead just do it the old way and pass `impl_build` the parent name together with the struct. Will do that split in the next PR as I was going to anyways.

config/base/derive/src/view.rs
ilchu: We don't actually need to, as the program still compiles with it present, but I guess it doesn't make sense to have `config` on `view`s. Due to this attribute is in charge of documentation loading and querying, we don't need it on views currently as none of those actions are being performed on them. Of course this might change later. I remember a [comment](https://github.com/hyperledger/iroha/issues/2118#issuecomment-1181471140) you made some time ago how `Combine` trait could help loading config on the fly, but enabling that functionality will require a separate line of thinking. I guess I'll create an issue to keep track of this.

config/base/derive/src/proxy.rs
ilchu: The `config_derive` crate still has logic related to types inside `Arc<RwLock>`s, such as calling `read()` and awaiting or calling `write()` on the field. It was there initially and I only refactored it away into `utils` mod. Even though not a single component is wrapped inside that in our codebase now, @appetrosyan mentioned it might be prudent to still have it -- for better generality, I suppose. And these `lvalue`s are currently used in e.g. `LoadFromEnv` impl. Thus what I mean here is that probably `gen_none_fields_check` needs to incorporate that as well.

Erigara: Ok, i get it.
Our proxy configs are used only for obtaining original configs from different sources as i understand, so it's hard to imagine that their fields would be wrapped into something like `Arc<RwLock>`.

Erigara (approved):

appetrosyan (approved):

PR #2641, opened 2022-08-20
[feature] #2596: Wasm validators

### LINKED ISSUE #2596: WASM permission validators

Arjentix:
### Description of the Change

* Added `Validator` and related things to the *data_model*
* Added `Chain` of validators to the `World`. With that we can check operations using runtime validators.
* Added instructions to modify this chain. Such as `Register<Validator>` and `Unregister<Validator>`
* Added entrypoint macro for validators
* Added support of validators into `wasm.rs` module
* Added test with sample validator

There are also **a lot** of **TODO** in this PR. This **TODO**s are the questions to the reviewers and my suggestions for a future.

### Issue

* Closes #2596 

### Benefits

* Runtime validators

### Possible Drawbacks

There are two main drawbacks (which are mentioned in the code too):

1. Malicious validator probably can break Iroha peer because right now Iroha believes that *validator* provides valid memory with `Verdict` sruct inside. If not, then UB. 
2. I really don't know why, but I can't prevent memory leak with this returned `Verdict`. Then I'm trying to deallocate it on Iroha side I get: `pointer being freed was not allocated` error. See `wasm.rs` and `wasm/derive/src/validator.rs` for details (EDIT: Fixed).

I think this 2 problems can be fixed with one shot. We should limit `Verdict` struct size (I mean denial reason string should be limited e.g. 256 bytes). Then we can allocate memory on Iroha side and provide it to the validator.
Remember that right now we need to allocate not just `size_of::<Verdict>()` but unpredictable number of bytes, because we use `parity_scale_codec::Encode` which encodes string with denial reason as well, and that string can have any size.

### Usage Examples or Tests

Comment `ignore` attribute on `deny_new_validators()` test.

Then run:

```bash
cargo +nightly-2022-08-15 test --package iroha_client --test mod -- integration::permissions::runtime_validators::deny_new_validators --exact --nocapture
```


data_model/src/permission/validator.rs
```diff
@@ -0,0 +1,231 @@
...
+    account::Account,
+    expression::Expression,
+    isi::Instruction,
+    query::QueryBox,
+    transaction::{Transaction, WasmSmartContract},
+    ParseError,
+};
+
+ffi_item! {
+    /// Permission validator that checks if some operation satisfies some conditions.
```
outoftardis, line 22:
```suggestion
    /// Permission validator that checks if an operation satisfies some conditions.
```

data_model/src/permission/validator.rs
```diff
@@ -0,0 +1,231 @@
...
+        #[getset(get = "pub")]
+        wasm: WasmSmartContract,
+    }
+}
+
+impl Registered for Validator {
+    type With = Self;
+}
+
+/// Identification of an [`Validator`].
```
outoftardis, line 62:
```suggestion
/// Identification of a [`Validator`].
```

data_model/src/permission/validator.rs
```diff
@@ -0,0 +1,231 @@
...
+    Clone,
+    derive_more::From,
+    derive_more::TryInto,
+    Encode,
+    Decode,
+    Serialize,
+    Deserialize,
+)]
+pub enum NeedsPermissionBox {
+    /// [`Transaction`] application
```
outoftardis, line 181: not `operation`?

wasm/derive/src/validator.rs
```diff
@@ -0,0 +1,63 @@
...
+        /// This function transfers the ownership of allocated
+        /// [`Verdict`](::iroha_wasm::data_model::permission::validator::Verdict)
+        #[no_mangle]
+        pub unsafe extern "C" fn _iroha_validator_main()
+            -> *const u8
+        {
+            use ::iroha_wasm::DebugExpectExt as _;
+
+            let verdict: ::iroha_wasm::data_model::permission::validator::Verdict = #fn_name(#arg);
+            let bytes_box =
```
Erigara, line 52: I'm not really sure that we could pass ownership this way, i suspect that smartcontract memory is not the same memory used by host iroha, this probaly explains error you recieve, when you try to wrap pointer into `Box` on the other end, you can't deallocate 
memory that was allocated by other allocator.
IMO you actually could use `Box::leak` here to recieve pointer, you would anyway free memory after droping runtime, however this should be investigated.

wasm/src/lib.rs
```diff
@@ -86,6 +97,44 @@ impl Execute for data_model::query::QueryBox {
...
+    type Value;
+    /// Type of error
+    type Error;
+
+    /// Calculate result on the host side.
+    ///
+    /// # Errors
+    ///
+    /// Depends on the implementation.
+    fn evaluate_on_host(&self) -> Result<Self::Value, Self::Error>;
```
Erigara, line 112: Smart trick!

wasm/src/lib.rs
```diff
@@ -228,6 +321,27 @@ unsafe fn encode_and_execute<T: Encode, O>(
...
+pub fn encode_with_length_prefix<T: Encode>(val: &T) -> Vec<u8> {
+    let len_size_bytes = core::mem::size_of::<usize>();
+
+    let mut r = Vec::with_capacity(len_size_bytes + val.size_hint());
+
+    // Reserve space for length
+    r.resize(len_size_bytes, 0);
+    val.encode_to(&mut r);
+
+    // Store length of the whole vector as byte array at the beginning of the vec
```
Erigara, line 336: I would suggest [`copy_from_slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice).

wasm/src/lib.rs
```diff
@@ -246,23 +360,7 @@ mod tests {
     use super::*;
```
Erigara, line 361: Hmm, could we use different allocator here, to emulate what happens between host and smartcontract? Just a thought.

client/build.rs
```diff
@@ -5,15 +5,29 @@
...
     // Rerunning if smart contract was changed
-    println!("cargo:rerun-if-changed={SMARTCONTRACT_LOCAL_PATH}");
+    for path in &TEST_SMARTCONTRACTS {
+        println!("cargo:rerun-if-changed={TEST_SMARTCONTRACTS_DIR}/{path}");
+    }
+    // TODO: check if this was causing the recursive loop.
+    // println!("cargo:rerun-if-changed=..");
 
     // Build and format the smartcontract if and only if the tests are
     // invoked with the nightly toolchain. We should not force our
```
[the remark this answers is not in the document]
mversic: because smart contracts can only be built on nightly

core/src/smartcontracts/isi/world.rs
```diff
@@ -325,6 +325,55 @@ pub mod isi {
...
+        fn execute(
+            self,
+            _authority: <Account as Identifiable>::Id,
+            wsv: &WorldStateView,
+        ) -> Result<(), Self::Error> {
+            let validator_id = self.object_id;
+
+            wsv.modify_validators(|validator_chain| {
+                if !validator_chain.remove_validator(&validator_id) {
+                    return Err(Error::Repetition(
```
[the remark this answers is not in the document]
Arjentix: Repetition error is returned every time user tries to remove someting, that doesn't exist yet. That is how it's done for accounts, assets, triggers and etc. I'm just doing the same thing here.
Although I agree, that it's strange and should be refactored someday

core/src/smartcontracts/wasm.rs
```diff
@@ -524,18 +616,68 @@ impl<'wrld> Runtime<'wrld> {
...
+        let len = u32::from_le_bytes(
+            memory.data(&mut context)[offset as usize..(offset + len_size_bytes) as usize]
+                .try_into()
+                .expect("Prefix length size(bytes) incorrect"),
+        );
+
+        // TODO: Should be wrapped into Box, but that causes:
+        // mod-da9c11e78642b700(13541,0x171f2f000) malloc: *** error for object 0x30018dfd4: pointer being freed was not allocated
+        let bytes = core::slice::from_raw_parts_mut(
+            memory.data_mut(context)[offset.try_into().expect(U32_TO_USIZE_ERROR_MES)
```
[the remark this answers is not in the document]
Arjentix: Yeah, I increase it by `len` on the next line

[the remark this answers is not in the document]
Arjentix: Yeah, `len` is the part of encoding

core/src/smartcontracts/wasm.rs
```diff
@@ -524,18 +616,68 @@ impl<'wrld> Runtime<'wrld> {
...
+            })?;
+        let len = u32::from_le_bytes(
+            memory.data(&mut context)[offset as usize..(offset + len_size_bytes) as usize]
+                .try_into()
+                .expect("Prefix length size(bytes) incorrect"),
+        );
+
+        // TODO: Should be wrapped into Box, but that causes:
+        // mod-da9c11e78642b700(13541,0x171f2f000) malloc: *** error for object 0x30018dfd4: pointer being freed was not allocated
+        let bytes = core::slice::from_raw_parts_mut(
```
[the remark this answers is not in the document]
Arjentix: Indeed, it works

core/src/tx.rs
```diff
@@ -128,6 +125,50 @@ impl TransactionValidator {
             }));
         }
 
+        // WSV is cloned here so that instructions don't get applied to the blockchain
+        // Therefore, this instruction execution validates before actually executing
+        let wsv_for_builtin_validators = WorldStateView::clone(&self.wsv);
+        self.validate_with_builtin_validators(&tx, &wsv_for_builtin_validators, is_genesis)?;
+
+        // Making a new clone so that instructions applied in the previous step won't break validation
```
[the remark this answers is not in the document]
Arjentix: It's already a significant performance hit, because we are already clonning wsv for checking. I'm just doing it once more.
In future we will remove builtin validators and we will have one clonning as before

data_model/src/transaction.rs
```diff
@@ -140,12 +140,18 @@ impl<T: IntoIterator<Item = Instruction>> From<T> for Executable {
...
 /// implementations for from byte vector can be split off from
 /// [`IntoIterator<Item = Instruction>`].
-#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+#[derive(Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
 pub struct WasmSmartContract {
     /// Raw wasm blob.
     pub raw_data: Vec<u8>,
 }
 
+impl core::fmt::Debug for WasmSmartContract {
```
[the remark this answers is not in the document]
Arjentix: Didn't know it has `DebugCustom`, cool

wasm/derive/src/entrypoint.rs
```diff
@@ -0,0 +1,160 @@
...
+        sig,
+        mut block,
+    } = parse_macro_input!(item);
+
+    assert!(
+        syn::ReturnType::Default == sig.output,
+        "Exported function must not have a return type"
+    );
+
+    let args = match syn::parse_macro_input!(attr as Attr) {
```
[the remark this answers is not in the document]
Arjentix: I don't think so. We can't check types of a function signature. Using string comparision instead of type checking is bad. I think, the approach with attributes is pretty ok

wasm/derive/src/validator.rs
Arjentix: I understand your suspection. But we are already using one allocator to allocate memory and another allocator to deallocate. It's used when Iroha allocates memory for i.e. query result and when WASM deallocates it. But somehow it's not woking on the contrary.
I've provided a possible solution in the PR description: allocate and deallocate memory for `Verdict` only on Iroha side

Arjentix: `Box::leak()` leads to the same thing as `ManuallyDrop` here. I just use the approach that was used in another places, see `wasm/src/lib.rs`

wasm/src/lib.rs
Arjentix: Hm... I don't know... Probably @mversic could help

data_model/src/permission/validator.rs
```diff
@@ -0,0 +1,231 @@
+//! Structures, traits and impls related to *runtime* `Validator`s.
+//!
+//! # Note
+//!
+//! Currently Iroha 2 has only builtin validators (see `core/src/smartcontracts/permissions`).
+//! They are partly using API from this module.
+//! In the future they will be replaced with *runtime validators* using WASM.
```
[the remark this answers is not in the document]
Arjentix: Oh, yes

data_model/src/permission/validator.rs
Arjentix: Right, I'll fix comments for this enum

core/src/genesis.rs
```diff
@@ -170,6 +170,8 @@ impl GenesisNetworkTrait for GenesisNetwork {
     fn from_configuration(
         submit_genesis: bool,
         raw_block: RawGenesisBlock,
+        // TODO: Replace with `Option<&Configuration>`,
+        // see https://www.reddit.com/r/rust/comments/wqsxk2/is_it_better_to_pass_optiont_or_optiont/
```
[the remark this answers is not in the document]
Arjentix: Created #2667

wasm/derive/src/entrypoint.rs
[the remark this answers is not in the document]
Erigara: > I think, the approach with attributes is pretty ok
 
 Yes, sure

wasm/derive/src/validator.rs
Erigara: > It's used when Iroha allocates memory for i.e. query result and when WASM deallocates it.

@Arjentix could you please point me where this happens?

Arjentix: Oh, I was wrong. On Iroha side we allocate with exported `_iroha_wasm_alloc()` (look at `encode_bytes_into_memory()`)

Arjentix: Looks like I need to create `_iroha_wasm_dealloc()` export and use it

wasm/src/lib.rs
```diff
@@ -58,7 +58,20 @@ fn oom(layout: ::core::alloc::Layout) -> ! {
...
 #[no_mangle]
 extern "C" fn _iroha_wasm_alloc(len: usize) -> *const u8 {
-    core::mem::ManuallyDrop::new(Vec::<u8>::with_capacity(len)).as_mut_ptr()
+    if len == 0 {
+        debug::dbg_panic("Cannot allocate 0 bytes");
+    }
+    let layout = alloc::alloc::Layout::array::<u8>(len).dbg_expect("Cannot allocate layout");
+    // Safety: safe until `layout` has non-zero size
+    unsafe { alloc::alloc::alloc_zeroed(layout) }
+}
```
Arjentix, line 67: Using `Vec` to allocate memory and then just `Box` to destroy this memory doesn't look right.
So I decided to replace it with this. This should take less place too.
See [here](https://www.reddit.com/r/rust/comments/jzwwqb/about_creating_a_boxed_slice/) for explanation

wasm/derive/src/validator.rs
Arjentix: Done

core/src/smartcontracts/wasm.rs
```diff
@@ -469,37 +529,72 @@ impl<'wrld> Runtime<'wrld> {
...
         bytes: impl AsRef<[u8]>,
         event: Event,
     ) -> Result<(), Error> {
-        let state = State::new(wsv, account_id.clone(), self.config).with_triggering_event(event);
-        self.execute_with_state(bytes, state)
+        let state = State::new(wsv, account_id, self.config).with_triggering_event(event);
+        self.execute_main_with_state(bytes, state)
+    }
+
+    /// Executes the given wasm runtime permission validator
```
outoftardis, line 553:
```suggestion
    /// Execute the given wasm runtime permission validator
```

wasm/derive/src/entrypoint.rs
```diff
@@ -0,0 +1,162 @@
...
+        let params = syn::parse_str(&params_str.value())?;
+        Ok(ParamsAttr {
+            _params_kw: params_kw,
+            _equal: equal,
+            params,
+        })
+    }
+}
+
+/// Collection of types of parameters that smart contract entrypoint function is expecting
```
outoftardis, line 57:
```suggestion
/// Collection of parameter types that the smart contract entrypoint function is expecting
```

wasm/src/lib.rs
mversic: > Hmm, could we use different allocator here, to emulate what happens between host and smartcontract? Not a suggestion, just a thought.

can you paraphrase?

core/src/tx.rs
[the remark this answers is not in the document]
Erigara: Uh, seems like we haven't any workaround for now.
> I'm just doing it once more.

If validation is in the hot path it could significant effect, however without bench-marking it's just speculations.
Could you try to run `tps` benchmark with and without addition cloning?

wasm/src/lib.rs
Erigara: @mversic, sure, @Arjentix, faced bug when he tried to deallocate inside iroha memory allocated by smartcontract. 
And our tests can't identify this because mocks are using the same global allocator as the tests themself i suppose.

Arjentix: @Erigara , so you want to introduce a test that will check for allocation/deallocation problems with Iroha and WASM?

Erigara: No, i talking about using custom allocator inside mocks.

wasm/src/lib.rs
```diff
@@ -58,7 +58,20 @@ fn oom(layout: ::core::alloc::Layout) -> ! {
 
 #[no_mangle]
 extern "C" fn _iroha_wasm_alloc(len: usize) -> *const u8 {
-    core::mem::ManuallyDrop::new(Vec::<u8>::with_capacity(len)).as_mut_ptr()
+    if len == 0 {
+        debug::dbg_panic("Cannot allocate 0 bytes");
+    }
+    let layout = alloc::alloc::Layout::array::<u8>(len).dbg_expect("Cannot allocate layout");
+    // Safety: safe until `layout` has non-zero size
```
Erigara, line 65: I think in `Safety` section we should specify, why using unsafe is safe in this case e.g.

```rust
// Safety: safe because `layout` has non-zero size
```

So if we write `unsafe` trait or function we write what invariants should hold and in `unsafe` block we write why this invariants hold in this particular case.

wasm/src/lib.rs
Erigara: wow, significant difference in size
do we need overhead of zeroing memory?

Erigara (approved):
LGTM

Erigara:
> Malicious validator probably can break Iroha peer because right now Iroha believes that validator provides valid memory with Verdict sruct inside. If not, then UB. 

But we provide encoded verdict, no? In this case it seems like UB couldn't happen.

core/src/smartcontracts/wasm.rs
```diff
@@ -524,18 +623,69 @@ impl<'wrld> Runtime<'wrld> {
...
+    /// # Warning
+    ///
+    /// This method takes ownership of the given pointer.
+    ///
+    /// # Safety
+    ///
+    /// It's safe to call this function as long as it's safe to construct, from the given
+    /// pointer, a byte array of prefix length and `Box<[u8]>` containing the encoded object
+    #[allow(unsafe_code, clippy::expect_used, clippy::unwrap_in_result)]
+    unsafe fn decode_with_length_prefix_from_memory<
```
Erigara, line 654: Seems like we don't have any unsafe operation inside.

core/src/tx.rs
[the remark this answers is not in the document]
Arjentix: After running on `iroha2-dev` and then on my branch benchmark reported that **no changes detected**

wasm/src/lib.rs
Arjentix: I'm not sure. But `Vec` does this. I just wanted to keep API the same. That's probably another question to @mversic

wasm/src/lib.rs
Arjentix: Right

Arjentix:
> But we provide encoded by parity scale Verdict, no? In this case it seems like UB couldn't happen.
Hm... Yes, you are right. Iroha expects only bytes, which should be always valid, I think

Then looks like we don't need such a trick with `Verdict` static size, that I proposed. Good

client/tests/integration/permissions/runtime_validators.rs
```diff
@@ -0,0 +1,52 @@
...
+    transaction::WasmSmartContract,
+};
+use test_network::*;
+
+#[test]
+// TODO: Maybe activate with a feature,
+// which will be enabled in the build-script when `nightly` is enabled?
+// Same for `mint_nft_for_every_use_every_1_sec()` test.
+#[ignore = "Only on nightly"]
+fn deny_new_validators() -> Result<()> {
```
Arjentix, line 18: @appetrosyan , I need your opinion on this

core/src/smartcontracts/wasm.rs
```diff
@@ -650,8 +646,8 @@ impl<'wrld> Runtime<'wrld> {
     ///
     /// It's safe to call this function as long as it's safe to construct, from the given
     /// pointer, a byte array of prefix length and `Box<[u8]>` containing the encoded object
-    #[allow(unsafe_code, clippy::expect_used, clippy::unwrap_in_result)]
-    unsafe fn decode_with_length_prefix_from_memory<
+    #[allow(clippy::expect_used, clippy::unwrap_in_result)]
+    fn decode_with_length_prefix_from_memory<
```
Erigara, line 650: Safety section in safe function.

Erigara (dismissed):

wasm/derive/src/entrypoint.rs
```diff
@@ -0,0 +1,162 @@
...
+        let params = syn::parse_str(&params_str.value())?;
+        Ok(ParamsAttr {
+            _params_kw: params_kw,
+            _equal: equal,
+            params,
+        })
+    }
+}
+
+/// Collection of parameters types that the smart contract entrypoint function is expecting
```
outoftardis, line 57:
```suggestion
/// Collection of parameter types that the smart contract entrypoint function is expecting
```

wasm/src/lib.rs
```diff
@@ -47,7 +58,20 @@ fn oom(layout: ::core::alloc::Layout) -> ! {
...
+    let layout = core::alloc::Layout::array::<u8>(len).dbg_expect("Cannot allocate layout");
+    // Safety: safe because `layout` is guaranteed to have non-zero size
+    unsafe { alloc::alloc::alloc_zeroed(layout) }
+}
+
+/// # Safety
+/// - `offset` is a pointer to a `[u8; len]` which is allocated in the WASM memory.
+/// - This function can't call destructor of the encoded object.
+#[no_mangle]
+unsafe extern "C" fn _iroha_wasm_dealloc(offset: *const u8, len: usize) {
```
mversic, line 73: yeah, previously there was no `dealloc` function because ownership was transferred back to smart contract

mversic (changes requested):

wasm/src/lib.rs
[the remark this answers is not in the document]
mversic: yeah, previously there was no dealloc function because ownership was transferred back to smart contract

wasm/src/debug.rs
```diff
@@ -37,6 +37,18 @@ pub fn dbg<T: Debug + ?Sized>(_obj: &T) {
     }
 }
 
+/// Print `mes` and call [`panic!`].
+///
+/// Only call [`panic!`] if `debug` feature is not specified.
```
mversic, line 42: and how is this maintained?

core/src/smartcontracts/isi/world.rs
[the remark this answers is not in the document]
Arjentix: Hm, now I see: `Repetition` is used for `Register` errors, not for `Unregister`. I'll replace with `FindError`

wasm/src/lib.rs
Arjentix: > I'm quite sure it doesn't. Stack part is initialized, but stuff on heap is uninitialized

Hm... When why does [this program](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1e98efc7cd5e04565db471462ce9ce2) print slice of zeros?

client/tests/integration/permissions/runtime_validators.rs
Arjentix: Yeah, I didn't know about nightly toolchain by default until yesterday.
So we will keep it ignored until we fully start using nigthly?

core/src/tx.rs
```diff
@@ -91,30 +91,27 @@ impl TransactionValidator {
     /// Fails if validation of any transaction fails
     pub fn validate_every(
         &self,
-        txs: &[VersionedAcceptedTransaction],
+        txs: impl IntoIterator<Item = VersionedAcceptedTransaction>,
```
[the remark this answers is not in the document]
Arjentix: Yes, but now we need to capture `txs` by value, because every transaction will be moved into runtime validator for validation (which will be stored in the `State`)

client/tests/integration/permissions/runtime_validators.rs
```diff
@@ -0,0 +1,52 @@
...
+        validator::Type::Instruction,
+        WasmSmartContract {
+            raw_data: wasm.clone(),
+        },
+    );
+    test_client.submit_blocking(RegisterBox::new(validator))?;
+
+    // Trying to register the validator again
+    let validator_2 = Validator::new(
+        "deny_new_validators_2%alice@wonderland".parse().unwrap(),
```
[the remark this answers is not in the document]
Arjentix: Yeah, but I'm open for suggestions

wasm/src/lib.rs
mversic: I may be wrong but UB can print anything

mversic (changes requested):

core/src/smartcontracts/wasm.rs
```diff
@@ -377,6 +388,41 @@ impl<'wrld> Runtime<'wrld> {
...
+        let state = caller.data();
+        let operation = state
+            .operation_to_validate
+            .as_ref()
+            .ok_or_else(|| Trap::new("There is no operation to validate".to_owned()))?;
+
+        let bytes = Self::encode_with_length_prefix(operation)?;
+        let operation_offset =
+            Self::encode_bytes_into_memory(&bytes, &memory, &alloc_fn, &mut caller)?;
+        Ok(operation_offset)
```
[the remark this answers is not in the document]
Arjentix: Well, yes, it is transferred. It works like in the other such functions in this file. 

- Operation is encoded into allocated bytes on Iroha side
- This bytes are transfered to WASM
- Operation is decoded on WASM side
- Bytes are deallocated on WASM side

wasm/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > yeah, previously there was no dealloc function because ownership was transferred back to smart contract

 Yeah, in never was transfered from smart contract to Iroha

wasm/derive/src/validator.rs
```diff
@@ -0,0 +1,63 @@
...
+    } = parse_macro_input!(item);
+
+    let fn_name = &sig.ident;
+    assert!(
+        matches!(sig.output, syn::ReturnType::Type(_, _)),
+        "Validator entrypoint must have `Verdict` return type"
+    );
+
+    let arg: syn::Expr = parse_quote! {{
+        let top_operation = ::iroha_wasm::query_operation_to_validate();
```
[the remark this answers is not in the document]
Arjentix: It means general enum of things, that may require validation. This is `NeedsPermissionBox`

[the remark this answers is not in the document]
Arjentix: Well, I see, that may look misleading. I'll rephrase

wasm/derive/src/lib.rs
```diff
@@ -162,68 +76,34 @@ impl syn::parse::Parse for ParamType {
...
+/// - If the function does not have a return type
+///
+/// # Example
+///
+// `ignore` because this macro idiomatically should be imported from `iroha_wasm` crate.
+//
+/// ```ignore
+/// use iroha_wasm::validator::prelude::*;
+///
+/// #[entrypoint]
```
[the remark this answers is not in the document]
Arjentix: No. Look at `wasm/src/lib.rs` `validator` mod and its reexports. I actually think, that it may be a good idea, to do such modules with preludes for all kind of smart contracts

wasm/src/debug.rs
Arjentix: Take a look at `dbg()` fn in this file

data_model/src/permission/validator.rs
```diff
@@ -0,0 +1,230 @@
...
+/// incorrect interpretation of validators system. All operations are allowed by default.
+/// Validators are checking for operation **incorrectness**, not for operation correctness.
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, Deserialize, Serialize, IntoSchema)]
+pub enum Verdict {
+    /// Operation is approved to pass to the next validator
+    /// or to be executed if there are no more validators
+    Pass,
+    /// Operation is denied
+    Deny(DenialReason),
+}
```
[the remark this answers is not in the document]
Arjentix: The reason is the same as for "*Why `Pass` instead of `Allow`*".
`Allow` and `Result` have a wrong connotation and suggest an incorrect interpretation of validators system.
All operations are allowed by default. Validators are checking for operation **incorrectness**, not for operation correctness.

I'll write a note about `Result`.

It's my vision, but if you are not agree, we can discuss this

core/src/validator.rs
```diff
@@ -0,0 +1,171 @@
...
+/// Works similarly to the
+/// [`Chain of responsibility`](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern).
+/// The validation of an operation is forwarded to all
+/// validators in the chain which have the required type.
+/// The validation stops at the first
+/// [`Deny`](iroha_data_model::permission::validator::Verdict::Deny) verdict.
+#[derive(Debug, Default, Clone)]
+pub struct Chain {
+    all_validators: DashMap<Id, Validator>,
+    concrete_type_validators: DashMap<Type, Vec<Id>>,
```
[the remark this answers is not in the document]
Arjentix: Yes, it shouldn't be modifying frequently... Does it worth rewriting to `RwLock`? Should it be faster than `DashMap`? Looks like `DashMap` suits well for both *rare* and *frequent* modifying. Or is `RwLock` just a more simple way we should considered if we haven't already used `DashMap` a lot?

wasm/src/lib.rs
mversic: > Hm... When why does [this program](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1e98efc7cd5e04565db471462ce9ce2) print slice of zeros?

the fact that this program prints all 0s doesn't mean that data is initialized to 0. Uninitialized data can be 0 as well

wasm/derive/src/validator.rs
```diff
@@ -0,0 +1,63 @@
...
+        pub unsafe extern "C" fn _iroha_validator_main()
+            -> *const u8
+        {
+            use ::iroha_wasm::DebugExpectExt as _;
+
+            let verdict: ::iroha_wasm::data_model::permission::validator::Verdict = #fn_name(#arg);
+            let bytes_box =
+                ::core::mem::ManuallyDrop::new(::iroha_wasm::encode_with_length_prefix(&verdict).into_boxed_slice());
+
+            bytes_box.as_ptr()
```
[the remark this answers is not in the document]
Arjentix: 1. Need to use as `Box::into_inner()`, because it's an associated function, not a method
2.
    ```
    the trait `Sized` is not implemented for `[u8]`
    note: required by a bound in `Box::<T, A>::into_inner`
    ```

core/src/tx.rs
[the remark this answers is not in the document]
Arjentix: Ok, we can do a research for that. I kind off agree with you. But right now `tx` has do be converted to another state of `Transaction`. That makes things complicated.
The easiest way for now: add a TODO and create an issue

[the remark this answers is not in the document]
Arjentix: Created #2742

mversic (approved):

wasm/src/lib.rs
Erigara: Int

> > I'm quite sure it doesn't. Stack part is initialized, but stuff on heap is uninitialized
> 
> Hm... When why does [this program](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1e98efc7cd5e04565db471462ce9ce2) print slice of zeros?

Internally `Vec` rely on [`Allocator::allocate`](https://doc.rust-lang.org/std/alloc/trait.Allocator.html#tymethod.allocate) function which do not guarante that data would be zero-initialized.

core/src/tx.rs
```diff
@@ -169,32 +214,40 @@ impl TransactionValidator {
...
-            Ok(MustUse(true)) => None,
-            Ok(MustUse(false)) => Some("Signature condition not satisfied.".to_owned()),
-            Err(reason) => Some(reason.to_string()),
-        }
-        .map(|reason| UnsatisfiedSignatureConditionFail { reason })
-        .map(TransactionRejectionReason::UnsatisfiedSignatureCondition);
+        let pure_tx = Transaction {
+            payload,
+            signatures,
+        };
```
Erigara, line 235: I just realized that this seems to go against out restrictions about not using `Transaction`.
We have discussion about this in #2704 recently.
And imo seems reasonable that validator get `Accepted` transaction, that pass basic checks. 
Problem is that `AcceptedTransaction` currently defined in `core` not in `data_model`.
What do you think?

Arjentix: I agree with you. But I don't know how we can fix this problem.

Maybe we could move `AcceptedTransaction` inside `data_model` and use it. Then we will have its methods inside `core` implemented with a triat.

Another question: *Why is Transaction considered Accepeted, when it wasn't checked by Validators yet?*. Maybe we should introduce a new state? Or let Validators receive Transaction and use it as `AcceptedTransaction` only after passing all validators?

Arjentix: We need this PR merged ASAP. So I've created #2744 for this

mversic (approved):

Erigara (approved):

PR #2644, opened 2022-08-23
[feature] #2628: Add support for `repr(transparent)` in FFI

### LINKED ISSUE #2628: `repr(transparent)` in FFI
`iroha_ffi` doesn't support delegating struct representations to `repr(transparent)`, rather it converts these structs into opaque pointers when crossing FFI. It would be a useful optimization to consider `repr(transparent)` when deriving `IntoFfi` or `TryFromReprC` and reduce number of indirections used

Erigara:
### Description of the Change

- Before transparent structs cross ffi as opaque pointers, now conversion to ffi delegated to inner field;
- Remove special handling of `fn(T) -> T` because this case was created under assumption that every struct passed as opaque pointer (implementation use `read/write/is_null` methods) which not hold in general for transparent and non-opaque structs.

### Issue

Closes #2628.

### Benefits

Reduce number of indirections.

### Possible Drawbacks

Case `fn(T) -> T` now uses 2 allocations instead of 1 and a bit less convenient to write.

### DIFF
--- a/ffi/derive/src/convert.rs
+++ b/ffi/derive/src/convert.rs
@@ -1,9 +1,9 @@
 use proc_macro2::TokenStream as TokenStream2;
-use proc_macro_error::abort;
+use proc_macro_error::{abort, OptionExt};
 use quote::quote;
 use syn::{parse_quote, DeriveInput, Ident};
 
-use crate::{enum_size, find_attr, is_fieldless_enum, is_opaque};
+use crate::{enum_size, find_attr, is_fieldless_enum, is_opaque, is_repr_attr};
 
 pub fn derive_try_from_repr_c(input: &DeriveInput) -> TokenStream2 {
     if !matches!(input.vis, syn::Visibility::Public(_)) {
@@ -28,7 +28,13 @@ pub fn derive_try_from_repr_c(input: &DeriveInput) -> TokenStream2 {
                 derive_try_from_repr_c_for_item(&input.ident)
             }
         }
-        syn::Data::Struct(_) => derive_try_from_repr_c_for_item(&input.ident),
+        syn::Data::Struct(item) => {
+            if is_transparent(input) {
+                derive_try_from_repr_c_for_transparent_item(&input.ident, &input.generics, item)
+            } else {
+                derive_try_from_repr_c_for_item(&input.ident)
+            }
+        }
         syn::Data::Union(item) => abort!(item.union_token, "Unions are not supported"),
     }
 }
@@ -56,7 +62,13 @@ pub fn derive_into_ffi(input: &DeriveInput) -> TokenStream2 {
                 derive_into_ffi_for_item(&input.ident)
             }
         }
-        syn::Data::Struct(_) => derive_into_ffi_for_item(&input.ident),
+        syn::Data::Struct(item) => {
+            if is_transparent(input) {
+                derive_into_ffi_for_transparent_item(&input.ident, &input.generics, item)
+            } else {
+                derive_into_ffi_for_item(&input.ident)
+            }
+        }
         syn::Data::Union(item) => abort!(item.union_token, "Unions are not supported"),
     }
 }
@@ -85,7 +97,7 @@ fn derive_try_from_repr_c_for_opaque_item_wrapper(input: &DeriveInput) -> TokenS
     let lifetime: syn::Lifetime = syn::parse_quote!('__iroha_ffi_itm);
     let lifetimes = &[lifetime.clone()];
     let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, &[]);
+        add_bounds_and_split_generics(&mut generics, lifetimes, &[], &[]);
 
     let opaque_item_slice_try_from_repr_c_derive =
         derive_try_from_repr_c_for_opaque_item_slice(input);
@@ -153,7 +165,7 @@ fn derive_try_from_repr_c_for_opaque_item(input: &DeriveInput) -> TokenStream2 {
     let lifetime: syn::Lifetime = syn::parse_quote!('__iroha_ffi_itm);
     let lifetimes = &[lifetime.clone()];
     let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, &[]);
+        add_bounds_and_split_generics(&mut generics, lifetimes, &[], &[]);
 
     let opaque_item_slice_try_from_repr_c_derive =
         derive_try_from_repr_c_for_opaque_item_slice(input);
@@ -212,7 +224,7 @@ fn derive_try_from_repr_c_for_opaque_item_slice(input: &DeriveInput) -> TokenStr
     let lifetimes = &[lifetime.clone()];
     let trait_bounds = &[syn::parse_quote!(core::clone::Clone)];
     let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, trait_bounds);
+        add_bounds_and_split_generics(&mut generics, lifetimes, &[], trait_bounds);
 
     quote! {
         impl<#generic_params> iroha_ffi::slice::TryFromReprCSliceRef<#lifetime> for #name #ty_generics #where_clause {
@@ -241,7 +253,7 @@ fn derive_try_from_repr_c_for_opaque_item_vec(input: &DeriveInput) -> TokenStrea
     let lifetime: syn::Lifetime = syn::parse_quote!('__iroha_ffi_itm);
     let lifetimes = &[lifetime.clone()];
     let (generic_params, ty_generics, where_clause) =
-        add_bounds_and_split_generics(&mut generics, lifetimes, &[]);
+        add_bounds_and_split_generics(&mut generics, lifetimes, &[], &[]);
 
     quote! {
         impl<#generic_params> iroha_ffi::owned::TryFromReprCVec<#lifetime> for #name #ty_generics #where_clause {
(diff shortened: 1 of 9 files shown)


ffi/derive/src/convert.rs
```diff
@@ -525,29 +762,130 @@ fn derive_into_ffi_for_fieldless_enum(enum_name: &Ident, repr: &[syn::NestedMeta
...
+    let (generic_params, ty_generics, where_clause) =
+        add_bounds_and_split_generics(&mut generics, &[], &[], &[]);
+
+    quote! {
+        // Note: if `#inner` is generic or doesn't have the same layout with `#name` this won't compile
+        impl <#generic_params> iroha_ffi::Transparent for #name #ty_generics #where_clause {
+            type Inner = #inner;
+
+            fn into_inner(outer: Self) -> Self::Inner {
+                unsafe { core::mem::transmute(outer) }
```
mversic, line 788: why do you have to transmute? Couldn't you do something like `outer.0` or `outer.first_field_name`?

ffi/derive/src/convert.rs
```diff
@@ -525,29 +762,130 @@ fn derive_into_ffi_for_fieldless_enum(enum_name: &Ident, repr: &[syn::NestedMeta
...
+            type Inner = &#lifetime #inner;
+
+            fn into_inner(outer: Self) -> Self::Inner {
+                unsafe { & *(outer as *const _ as *const _) }
+            }
+
+            fn from_inner(inner: Self::Inner) -> Self {
+                unsafe { & *(inner as *const _ as *const _) }
+            }
+        }
```
mversic, line 806: I don't like approach with `Transparent` traits because it gives user the option to implement different `Inner` type then the type that actually inside the transparent struct. Especially, when all of the information to derive correct behavior is given in the type definition and you could just delegate to the inner type.

Could you derive `IntoFfi` and `TryFromReprC` directly without the `Transparent` trait? Something like:
```rs
        impl<#generic_params> iroha_ffi::IntoFfi for #name #ty_generics #where_clause {
            type Target = <#inner_type as iroha_ffi::IntoFfi>::Target;

            fn into_ffi(self) -> Self::Target {
                iroha_ffi::IntoFfi::into_ffi(self.0)
            }
```

Also, why do you need to special case implementations for references as in `&#name`, etc?

ffi/derive/src/convert.rs
Erigara: I think i could do that quite easily in case of `IntoFfi`, but other way around would require additional assumptions like type has constructor function `new` or that every other field has default value.

ffi/derive/src/convert.rs
Erigara: I will try to explain my intentions here.
Maybe you noticed kinda strange signature that we have:
```rust
impl<T: IntoFfi> IntoFfi for Type where Self: Transparent<Inner = T> {
     ...
}
```
Problem is following: suppose we have some kind of newtype `Newtype(Inner)`. By just looking at `Inner` we can't know if `Inner`, `&Inner`, `&mut Inner` implement `IntoFfi`.
So by introducing generic parameter we can move error invocation from `impl` block to trying actually call unsupported conversion.

For example currently we don't implement `IntoFfi` for `&String/&mut String`, so:
```rust
// Without generic parameter expansion would always fail 
// because we can't know that we need to derive only `Name: IntoFfi` but not `&Name/&mut Name: IntoFfi`.
// However with generic param everything would generate without problem and ... 
#[derive(IntoFfi, TryFromReprC)]
#[repr(transparent)]
struct Name(String);

fn main() {
    /// only trying to call unimplemented method would fail 
    /// because there no such type `T` that both `&Name: Transparent<Inner = &T>` and `&T: IntoFfi`.
    let name_ref_ffi = IntoFfi::into_ffi(&name);
}
```

I think without `Tranparent` trait we can't actually express this i believe. 
I agree with you that it is not great to expose kinda inner trait to public interface.

One small notice: `transmute` are kind of guard here :) because it's checks that size of `Name` and `String` are the same. However i understand that `transmute` is really unsafe and try to use other conversions where possible.

ffi/derive/src/convert.rs
```diff
@@ -265,6 +277,136 @@ fn derive_try_from_repr_c_for_opaque_item_vec(input: &DeriveInput) -> TokenStrea
     }
 }
 
+fn derive_try_from_repr_c_for_transparent_item(
```
outoftardis, line 280: just noticed this file doesn't have any doc comments at all

ffi/src/lib.rs
```diff
@@ -256,6 +256,42 @@ pub struct Opaque {
     __marker: core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
 }
 
+/// Trait for `#[repr(transparent)]` structs to convert between [`Self`] and [`Self::Inner`]
+pub trait Transparent: Sized {
+    /// non-ZST field of transparent struct
```
outoftardis, line 261:
```suggestion
    /// Non-ZST field of transparent struct
```

ffi/src/lib.rs
```diff
@@ -256,6 +256,42 @@ pub struct Opaque {
...
+    /// Convert transparent struct into its non-ZST field
+    fn into_inner(outer: &[Self]) -> &[Self::Inner];
+
+    /// Recover transparent struct from its non-ZST field
+    fn from_inner(inner: &[Self::Inner]) -> &[Self];
+}
+
+/// Trait for `#[repr(transparent)]` structs to convert between vec of [`Self`] and vec of [`Self::Inner`]
+pub trait TransparentVec: Sized {
+    /// non-ZST field of transparent struct
```
outoftardis, line 285:
```suggestion
    /// Non-ZST field of transparent struct
```

ffi/src/lib.rs
```diff
@@ -256,6 +256,42 @@ pub struct Opaque {
...
+    /// Convert transparent struct into its non-ZST field
+    fn into_inner(outer: Self) -> Self::Inner;
+
+    /// Recover transparent struct from its non-ZST field
+    fn from_inner(inner: Self::Inner) -> Self;
+}
+
+/// Trait for `#[repr(transparent)]` structs to convert between slice of [`Self`] and slice of [`Self::Inner`]
+pub trait TransparentSlice: Sized {
+    /// non-ZST field of transparent struct
```
outoftardis, line 273:
```suggestion
    /// Non-ZST field of transparent struct
```

ffi/derive/src/convert.rs
mversic: >  I agree with you that it is not great to expose inner trait to public interface.

my main concern is that it's getting complicated and offering forms of extensibility that will never get to be used.

> Problem is following: suppose we have some kind of newtype `Newtype(Inner)`. By just looking at Inner we can't know if `Inner`, `&Inner`, `&mut Inner` implement `IntoFfi`.
For example currently we don't implement IntoFfi for &String/&mut String

Ok, I get you. While we could require that `&T: IntoFfi`, in many cases it's not possible to give a sound `&mut T` implementation (`String` is a good example)

mversic: > I think without Tranparent trait we can't actually express this i believe.

After some thought, I have to agree

> What we can do is to mark this trait and every it's method as unsafe to warn user to be extra careful.

unsafe on traits is a marker for those implementing the trait that they have to obey certain invariants when implementing the trait while unsafe on methods is a marker for those calling the method that they have to uphold certain invariants when calling the method. From the looks of it, I think that trait should be marked unsafe, but not the methods

Erigara: > From the looks of it, I think that trait should be marked unsafe, but not the methods

Yes, i agree with you on that.

Erigara: > After some thought, I have to agree. The only alternative would be to offload the decision to the user via derive macro helper attributes

I was thinking about helper attributes but in different way: make `derive(Transparent)` macro with helper attribute on the inner field. 
Also this macro could check that struct indeed have `repr(Transparent)` attribute and place kind some kind of helper attribute like `#[transparent]` on it to check it from `IntoFfi/TryFromFfi`.

This would allow user to decide how them what to pass transparent struct by unwrapping inner field or by opaque pointer e.g. despite `Name` is transparent right now we cant derive `IntoFfi/TryFromFfi` on it because `ConstString` is union and unions is not supported right now. 

However this whole thing seems to be overkill.

ffi/src/lib.rs
```diff
@@ -256,6 +256,42 @@ pub struct Opaque {
...
+
+    /// Convert transparent struct into its non-ZST field
+    fn into_inner(outer: Self) -> Self::Inner;
+
+    /// Recover transparent struct from its non-ZST field
+    fn from_inner(inner: Self::Inner) -> Self;
+}
+
+/// Trait for `#[repr(transparent)]` structs to convert between slice of [`Self`] and slice of [`Self::Inner`]
+pub trait TransparentSlice: Sized {
```
mversic, line 272: can't we replace this with:
```rs
impl<T: Transparent> iroha_ffi::Transparent for &[T] {
    type Inner = &[T::Inner];
}
```

or

```rs
impl<T: Transparent> iroha_ffi::Transparent for [T] {
    type Inner = [T::Inner];
}
```

depending on whether sizedness is required?

mversic: because if `T` is transparent, doesn't that mean that `[T]` is also transparent?

Erigara: Answer for [T] i think yes, but for `Vec<T>` it's not correct because we doesn't have defined layout for the vector.
Thank you for the idea i will try to create such blanket impls.

mversic: Ok, neither `Vec<T>` nor `&[T]` have defined layouts, but you can make blanket impls via pointer casts

mversic: On 2nd thought I would implement `Transparent` only for `T`/`[T]` and add the following impls:

```rs
impl<T: ? Sized + Transparent<Inner = U>, U: ?Sized> IntoFfi for &T where &U: IntoFfi {
    type Target = <&U as IntoFfi>::Target:

    fn into_ffi(self) -> Self::Target {
        // do the transmutation
    }
}
```

but this might conflict with current blanket impl for `&T`

I can't verify this works right now, so it might be a blunt

ffi/tests/ffi_export.rs
```diff
@@ -130,15 +130,21 @@ fn get_new_struct() -> FfiStruct {
 
 #[allow(trivial_casts)]
 fn get_new_struct_with_params() -> FfiStruct {
-    let mut ffi_struct = get_new_struct();
+    let ffi_struct = get_new_struct();
     let params = get_default_params();
 
+    let mut output = MaybeUninit::new(core::ptr::null_mut());
```
appetrosyan, line 136: Why are we using `MaybeUninit` in this instance?

ffi/src/lib.rs
appetrosyan: I'd recommend opening a ticket for it and adding the change later.

ffi/derive/src/convert.rs
appetrosyan: It doesn't have to. Anything that stays in the `derive` crate is almost completely for internal use only.

ffi/derive/src/convert.rs
appetrosyan: This sounds like an `X` / `Y` problem. All we need is to be able to use transparent structures. A manual `impl` for `Name` at this time is very much not an issue, and given how it's implemented, definitely not an issue of code size. I recommend postponing this disucssion after dynamic linking is at least somewhat functional. We've spent a lot of time doing the FFI, we need to show something for it.

ffi/derive/src/convert.rs
appetrosyan: Not too keen on having `transmute` be generated by a macro. If it triggers UB, we need to be able to see it. 

As for the assumptions, I think that given this is done inside a macro, the exceptions that require transmute can be handled differently. At the very least it could be an argument to the derive macro, I'd prefer if there was a blanket `impl` and proper traits.

appetrosyan (approved):

ffi/derive/src/convert.rs
mversic: > Not too keen on having transmute be generated by a macro. If it triggers UB, we need to be able to see it.

That's why `Transmute` trait must be made `unsafe` as I mentioned [here](https://github.com/hyperledger/iroha/pull/2644#discussion_r955754068)

ffi/src/lib.rs
Erigara: > I think we should aim for this but this might conflict with current blanket impl for `&T`

Sadly there is conflict with `&T`.

Erigara: > I'd recommend opening a ticket for it and adding the change later.

Here it is #2668

ffi/derive/src/convert.rs
Erigara: Made `Transparent` trait `unsafe`.

ffi/derive/src/convert.rs
Erigara: Ok, i will resolve this discussion than, i have created #2669.

Erigara:
> I would recommend that we still do this for opaque structs which will be the prevalent case. This is obviously not appropriate for non-opaque structs. As for the transparent structures I have a feeling that it could be handled appropriately depending on whether the inner struct is opaque or not, but that should be left for future improvements.

So from `impl` block we have no way to determine with which kind of item we are dealing with. Maybe i miss something?

What we can do in theory is to transform signature `fn(self) -> Self` into `fn(&mut self)` however as i remember you said that ffi of `&mut T` is unsound in most cases.   

> Did you think about how this affects deallocation of opaque structures?

I don't find any leaks or you are talking about something else?

ffi/tests/ffi_export.rs
Erigara: We could avoid `MaybeUninit` here, i just use it for convenience here as we use `MaybeUninit` everywhere in ffi tests.

mversic:
> So from `impl` block we have no way to determine with which kind of item we are dealing with. Maybe i miss something?

uh, you're right unfortunately. So, there isn't much that can be done here?

> What we can do in theory is to transform signature `fn(self) -> Self` into `fn(&mut self)` however as i remember you said that ffi of `&mut T` is unsound in most cases.

yeah, like when transmuting `&mut bool` into `&mut u8`. FFI side can manipulate the `u8` in such a way that it causes UB on the Rust side. Unless it's opaque pointer or is already a robust `repr(C)` type like `u8` the shared reference cannot be sent safely across FFI

ffi/tests/ffi_export.rs
mversic: just to demonstrate how one can emulate out-pointer. User can, of course, use other means to define the pointer

mversic (approved):

Erigara:
> > So from `impl` block we have no way to determine with which kind of item we are dealing with. Maybe i miss something?
> 
> uh, you're right unfortunately. So, there isn't much that can be done here at the moment?

I think so.

mversic:
> > > So from `impl` block we have no way to determine with which kind of item we are dealing with. Maybe i miss something?
> > 
> > 
> > uh, you're right unfortunately. So, there isn't much that can be done here at the moment?
> 
> I think so.

Please open an issue, so we can discuss it there

Erigara:
> Please open an issue, so we can discuss it there

Created #2676.

appetrosyan (approved):

PR #2655, opened 2022-08-25
[fix] #0: Modify `test_env.sh` to allocate storage for each peer

s8sato:
### Description of the Change

- Modify `test_env.sh` to allocate storage for each peer

### Issue

- Nothing special before

### Benefits

- Closer to the actual situation about storages

### Possible Drawbacks

- None

### Usage Examples or Tests

```bash
bash scripts/test_env.sh setup
bash scripts/tests/register_mint_quantity.sh
```

### DIFF
--- a/scripts/test_env.sh
+++ b/scripts/test_env.sh
@@ -73,6 +73,7 @@ function set_up_peers_common {
 }
 
 function bulk_export {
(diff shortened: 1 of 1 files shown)


appetrosyan (approved):

6r1d (approved):
LGTM!

PR #2658, opened 2022-08-26
[refactor] #2482: Store `Account` permission tokens in `Wsv`

### LINKED ISSUE #2482: Optimize how `PermissionToken`s are stored
Right now `Account` stores permission tokens and roles. Roles are stored as ids, their permission tokens aren't stored at `Account`, instead they are stored in `WorldStateView`.
The problem is that it's misleading and leads to bugs, i.e. #2477. Programmer without knowing full context thinks that just using `Account::contains_permission()` he can check if this account can perform action or not, although this account can also have permissions granted by roles, so that programmer should check it using `WorldStateView::account_permission_tokens()`, which returns *all* tokens inclidung tokens from roles.

We should consider refactoring that. There is no perfect solution, but here are some ideas:
1. Remove `Grant<Account, PermissionToken>` and use only `Grant<Account, Role>` (same for `Revoke`). This will break the *API*
2. Store permission tokens from roles inside the account. Leads to more usage of memory and time
 2.1. (optional) Use interning for permission tokens. Less memory usage but can be too complex

Erigara:

### Description of the Change

- Move permission token set from `Account` into `Wsv`.
- Store accounts permission tokens in hash map: `Account::Id -> Set<PermissionToken>`

I this PR i tried to preserve previous iroha behavior.

### Issue

Closes #2482.

### Benefits

- Bugs described in related issue should not happen.
- `Account` struct is more light-weighted now. 

### Possible Drawbacks

- Increase size of `Wsv`, already large struct.


core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
...
+            .account_permission_tokens
+            .get_mut(account)
+            .map_or(false, |permissions| permissions.contains(token))
+    }
+
+    /// Add [`permission`](PermissionToken) into the [`Account`].
+    ///
+    /// If `Account` did not have this permission present, `true` is returned.
+    /// If `Account` did have this permission present, `false` is returned.
+    pub fn add_account_permission(
```
Erigara, line 193: I don't check if account with such id exist, but this seems unnecessary, because we use `add/remove_account_permission` in such a way that this check happened before (except tests).

And such check will introduce additional branching and more complicated api.

core/src/wsv.rs
```diff
@@ -52,6 +52,8 @@ pub struct World {
     pub(crate) domains: DomainsMap,
     /// Roles. [`Role`] pairs.
     pub(crate) roles: crate::RolesMap,
+    /// Account permission tokens.
```
outoftardis, line 55:
```suggestion
    /// Permission tokens of an account.
```

core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
...
+        self.world
+            .account_permission_tokens
+            .get(account.id())
+            .map_or_else(Default::default, |permissions_ref| {
+                permissions_ref.value().clone()
+            })
+            .into_iter()
+    }
+
+    /// Return `true` if `Account` contains permission token not accociated with any role
```
outoftardis, line 176:
```suggestion
    /// Return `true` if `Account` contains a permission token not associated with any role.
```

core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
         tokens
     }
 
+    /// Returns a set of permission tokens granted to this account separately.
```
outoftardis, line 162:
```suggestion
    /// Return a set of permission tokens granted to this account separately.
```

by `separately` you mean not part of any role or what?

core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
...
+                self.world
+                    .account_permission_tokens
+                    .insert(account.clone(), permissions);
+                true
+            }
+            Some(mut permissions) => permissions.insert(token),
+        }
+    }
+
+    /// Remove a [`permission`](PermissionToken) from the [`Account`] and return whether the permission was present in the [`Account`]
```
outoftardis, line 212:
```suggestion
    /// Remove a [`permission`](PermissionToken) from the [`Account`] if the account has this permission. Return a Boolean value indicating whether the [`Account`] had this permission.
```

core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
...
+        self.world
+            .account_permission_tokens
+            .get_mut(account)
+            .map_or(false, |permissions| permissions.contains(token))
+    }
+
+    /// Add [`permission`](PermissionToken) into the [`Account`].
+    ///
+    /// If `Account` did not have this permission present, `true` is returned.
+    /// If `Account` did have this permission present, `false` is returned.
```
outoftardis, line 192:
```suggestion
    /// Add [`permission`](PermissionToken) to the [`Account`] if the account does not have this permission yet.
    ///
    /// Return a Boolean value indicating whether or not the  [`Account`] already had this permission.
```

core/src/wsv.rs
Erigara: yes

Erigara:
```
/// Return a set of permission tokens granted to this account individually.
```
Maybe?

outoftardis: I'd suggest mentioning the role explicitly 

```suggestion
    /// Return a set of permission tokens granted to this account not as part of any role.
```

core/src/wsv.rs
```diff
@@ -143,12 +145,12 @@ impl WorldStateView {
         self.map_account(id, |account| account.assets().cloned().collect())
     }
 
-    /// Returns a set of permission tokens granted to this account as part of roles and separately.
+    /// Return a set of permission tokens granted to this account as part of roles or individually.
```
outoftardis, line 148: why not this;
```suggestion
    /// Return a set of all permission tokens granted to this account.
```

Erigara: Thank you for all of your suggestions!

core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
         tokens
     }
 
+    /// Return a set of permission tokens granted to this account not as part of any role.
+    pub fn account_own_permission_tokens(
```
appetrosyan, line 163:
```suggestion
    pub fn account_explicit_permission_tokens(
```

core/src/wsv.rs
```diff
@@ -157,6 +159,68 @@ impl WorldStateView {
...
+            .get(account.id())
+            .map_or_else(Default::default, |permissions_ref| {
+                permissions_ref.value().clone()
+            })
+            .into_iter()
+    }
+
+    /// Return `true` if [`Account`] contains a permission token not associated with any role.
+    #[inline]
+    pub fn account_contains_own_permission(
```
appetrosyan, line 178:
```suggestion
    pub fn account_contains_explicit_permission(
```

appetrosyan (approved):

appetrosyan:
LGTM. I noticed that the binary `Account` structure has reduced in size by a single byte.

core/src/wsv.rs
ilchu: Or maybe "inherent" instead of "explicit" would also work here, to signify that the permission is tied to account only?

ilchu (approved):
LGTM

s8sato (approved):
Now both two kinds of permission tokens are stored in their respective maps in WSV and united on `wsv.account_permission_tokens()` call. More concise

appetrosyan (approved):

s8sato (approved):

core/src/wsv.rs
s8sato: Such checks should be done outside of this method, just like `wsv.modify_account()` does at the moment

Issue #2607, opened 2022-08-11
Compress Code in sumeragi main loop


SamHSmith:

SamHSmith:
As of now the code in the sumeragi main loop has been very explicit. This was done in order to get a clear view of the actual operations that were being done. Now that we have a working reimplementation it is time to recompress commonly done operations into their own functions.

PR #2659, opened 2022-08-26
[fix] #2561, #2607: Fully operational new-sumeragi implentation.

SamHSmith:
### Description of the Change

Readded view-changes, solved several robustness bugs and refactored for reduced code size.

### Benefits

Higher TPS, more robust code.


core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
+
+                            let block = block.validate(
+                                &sumeragi.transaction_validator,
+                                &state_machine_guard.wsv,
+                            );
+                            block
+                        };
 
                         for event in Vec::<Event>::from(&block) {
                             trace!(?event);
```
Erigara, line 867: We had discussion about avoiding calling `drop`, i think maybe this one pop up after rebase.

core/src/sumeragi/fault.rs
```diff
@@ -1305,500 +872,511 @@ pub fn run_sumeragi_main_loop<F>(
...
-                            sumeragi.kura.blocking_store_block(block);
-                            SumeragiWithFault::<F>::update_network_topology(
-                                &mut state_machine_guard.current_topology,
-                                &wsv_guard,
-                            );
+                            block_commit(sumeragi, signed_block, &mut state_machine_guard);
                             has_sent_transactions = false;
                             voting_block_option = None;
+                            old_view_change_index = 0;
+                            view_change_proof_chain.clear();
```
Erigara, line 884: Maybe introduce function which would introduce state clean up?

core/src/sumeragi/fault.rs
```diff
@@ -1305,500 +872,511 @@ pub fn run_sumeragi_main_loop<F>(
...
+                        instant_at_which_we_should_have_committed =
+                            Instant::now() + sumeragi.commit_time;
+                    }
+                    Message::BlockSigned(block_signed) => {
+                        let block = block_signed.block;
+                        let block_hash = block.hash();
 
-                            let network_topology = &state_machine_guard.current_topology;
+                        if voting_block_option.is_some()
+                            && block_hash != voting_block_option.as_ref().unwrap().block.hash()
```
Erigara, line 1149: Maybe?

```rust
if Some(block_hash) != voting_block_option.map(|block| block.block.hash()) {
    ...
}
```

core/src/sumeragi/fault.rs
```diff
@@ -1305,500 +872,511 @@ pub fn run_sumeragi_main_loop<F>(
...
-                    // if there is no message sleep
-                    should_sleep = true;
+                    _ => {
+                        trace!("Not handling message {:?}", incoming_message);
+                    }
                 }
+            } else {
+                // if there is no message sleep
+                should_sleep = true;
+            }
```
Erigara, line 1173: We have this branch in many places could we extract it somehow in single place?

core/src/sumeragi/fault.rs
```diff
@@ -1305,500 +872,511 @@ pub fn run_sumeragi_main_loop<F>(
...
+        );
+    }
+    {
+        *sumeragi
+            .latest_block_hash_for_use_by_block_sync
+            .lock()
+            .expect("push hash to block sync for genesis") = Hash::zeroed().typed();
+    }
+
+    loop {
```
Erigara, line 1332: Oh, i was thinking about simmilar approah: instead of just `finite state machine` introduce [`hierarchical finite state machine`](https://web.stanford.edu/class/cs123/lectures/CS123_lec08_HFSM_BT.pdf).

Erigara (commented):
I would suggest to find places where `is_some/is_ok` is followed by `unwrap` and replace it with `if let`.

core/src/sumeragi/mod.rs
```diff
@@ -188,42 +178,50 @@ impl Sumeragi {
...
-            .choose(rng)
-            .cloned()
+            .expect("lock on online peers for get random peer")
+            .iter()
+            .map(|peer| Peer::new((*peer).clone()))
+            .collect::<Vec<Peer>>();
+        if vec.is_empty() {
+            return None;
+        }
+        vec.sort();
```
appetrosyan, line 202:
```suggestion
        if vec.is_empty() {
            None
        } else {
             vec.sort();
             Some(...)
       }
```

You're mixing styles. If you're short circuiting, it's not immediately obvious that things outside of the `if` are alternatives to the `if`. This is obvious to a C programmer (who would consider this an anti-pattern), but everyone else would find the `else` style more explicit.

appetrosyan: Also consider sorting after the `iter()` call.

core/src/sumeragi/mod.rs
```diff
@@ -85,48 +85,44 @@ impl Sumeragi {
...
 
+            wsv: wsv.clone(),
+            transaction_cache: Vec::new(),
+
             sumeragi_thread_should_exit: false,
         };
 
-        let (incoming_message_sender, incoming_message_receiver) = std::sync::mpsc::channel();
+        let (incoming_message_sender, incoming_message_receiver) =
+            std::sync::mpsc::sync_channel(50); // TODO change to constant
```
appetrosyan, line 103: Or change it to a configuration variable. Or a compile-time configuration variable. The smaller the struct the more cache-friendly it is. We want to be able to profile it.

core/src/sumeragi/mod.rs
```diff
@@ -61,7 +61,7 @@ trait Consensus {
 /// `Sumeragi` is the implementation of the consensus.
 #[derive(Debug)]
 pub struct Sumeragi {
-    internal: SumeragiWithFault<NoFault>,
+    pub internal: SumeragiWithFault<NoFault>,
```
appetrosyan, line 64: @Arjentix We should eventually get rid of these. I personally wouldn't mind getting rid of it after merging this.

cli/src/torii/routing.rs
```diff
@@ -124,7 +125,7 @@ pub(crate) async fn handle_queries(
     sorting: Sorting,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
-    let wsv = sumeragi.wsv_clone();
+    let wsv = sumeragi.wsv_mutex_access();
```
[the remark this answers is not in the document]
SamHSmith: I would rather use an explict drop.
```rust
    let wsv = sumeragi.wsv_mutex_access();
    let (valid_request, filter) = request.validate(&wsv, query_judge.as_ref())?;
    drop(wsv);
```
Let's see what Aleksandr says.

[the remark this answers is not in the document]
appetrosyan:
```suggestion
    let result = {
      let wsv = sumeragi.wsv_mutex_access();
      let (valid_request, filter) = request.validate(&wsv, query_judge.as_ref())?;
      let original_result = valid_request.execute(&wsv)?;
      filter.filter(original_result);
    }
```

cli/src/torii/routing.rs
```diff
@@ -256,26 +256,24 @@ async fn handle_post_configuration(
...
+            BlockPublisherMessage::SubscriptionAccepted,
+        ))
+        .await?;
 
-        stream
-            .send(VersionedBlockPublisherMessage::from(
-                BlockPublisherMessage::SubscriptionAccepted,
-            ))
-            .await?;
+    let wsv = sumeragi.wsv_mutex_access().clone();
```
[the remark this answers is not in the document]
appetrosyan: Actually, this isn't so much of a bug, as an API choice, I think that @Erigara is right. We need to reacquire the lock during the loop. Streaming the block is a long process, and we want to propagate the changes made to the later blocks. This is the definition of streaming.

appetrosyan (changes requested):

cli/src/torii/routing.rs
```diff
@@ -394,68 +390,36 @@ mod subscription {
...
-    )
-         */
-    warp::reply::json(b"Not implemented")
+
+    #[allow(clippy::expect_used)]
+    let string = sumeragi
+        .wsv_mutex_access()
+        .blocks()
+        .last()
+        .expect("At least genesis should always exist")
```
appetrosyan, line 399: Actually, sometimes it doesn't. 

```suggestion
        .expect("Genesis not applied. Nothing we can do. Solve the issue and rerun.")
```

client/benches/tps/config.json
```diff
@@ -1,7 +1,7 @@
 {
     "peers": 4,
-    "interval_us_per_tx": 80000,
-    "max_txs_per_block": 8192,
-    "blocks": 5,
+    "interval_us_per_tx": 10000,
+    "max_txs_per_block": 512,
```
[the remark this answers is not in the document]
appetrosyan: Actually this means that we reduce the worst-case latency, which effectively increases the short-run average TPS. It might seem that because it takes non-zero time to vote on a new block, that we'd reduce the long-run average TPS, but you're shuffling less data. The overhead is asymptotically irrelevant.

client/benches/tps/lib.rs
```diff
@@ -113,7 +114,9 @@ impl Config {
         let tps = txs_accepted as f64 / elapsed_secs;
         iroha_logger::info!(%tps, %txs_accepted, %elapsed_secs, %txs_rejected);
         if txs_rejected > 0 {
-            return Err(eyre::eyre!("{txs_rejected} transactions were rejected"));
+            // There will be rejected transactions since we submit more then the
```
appetrosyan, line 117:
```suggestion
            // There will be rejected transactions since we submit more than the
```

client/tests/integration/unstable_network.rs
```diff
@@ -65,7 +65,7 @@ fn unstable_network(
     let (network, mut iroha_client) = rt.block_on(async {
         let mut configuration = Configuration::test();
         configuration.queue.maximum_transactions_in_block = MAXIMUM_TRANSACTIONS_IN_BLOCK;
-        configuration.logger.max_log_level = Level::ERROR.into();
+        configuration.logger.max_log_level = Level::TRACE.into();
```
appetrosyan, line 68: This is a safe default for most tests. If it's not written to `stdout` it gets dropped, so it shouldn't make things much slower.

core/src/block_sync.rs
```diff
@@ -198,6 +202,7 @@ impl Message {
                 if blocks.is_empty() {
                     warn!(%hash, "Block hash not found");
                 } else {
+                    info!("Sharing blocks after hash: {}", hash);
```
appetrosyan, line 205:
```suggestion
                    trace!("Sharing blocks after hash: {}", hash);
```

core/src/queue.rs
```diff
@@ -31,9 +31,9 @@ pub struct Queue {
     ///
     /// DashMap right now just iterates over itself and calculates its length like this:
     /// self.txs.iter().len()
-    txs_in_block: usize,
+    pub txs_in_block: usize,
     max_txs: usize,
-    ttl: Duration,
+    pub ttl: Duration,
```
appetrosyan, line 36:
```suggestion
    pub tx_time_to_live: Duration,
```

core/src/sumeragi/fault.rs
```diff
@@ -294,8 +262,133 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+
+    // Update WSV copy that is public facing
+    {
+        let mut wsv_for_public_use_guard = sumeragi.wsv_for_public_use.lock().unwrap();
+        *wsv_for_public_use_guard = WorldStateView::clone(&state_machine_guard.wsv);
+    }
+
+    for event in Vec::<Event>::from(&block) {
+        trace!(?event);
+        drop(sumeragi.events_sender.send(event));
```
appetrosyan, line 294: These drops need a bit of explaining. I don't understand why they can't just go out of scope.

core/src/sumeragi/fault.rs
```diff
@@ -294,8 +262,133 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+
+    // Transaction Cache
+    {
+        let mut transaction_cache = &mut state_machine_guard.transaction_cache;
+        // We prune transactions now because now is the only time they may have been
+        // invalidated by a new block being applied to the world state view.
+        //
+        // We also check for if the transaction has expired because we might as well.
+        let mut read_index = 0;
+        let mut write_index = 0;
```
appetrosyan, line 332: These mutable counters are becoming an anti-pattern. But in this _particular_ case the simple functional rewrite actually makes the thing less efficient.

core/src/sumeragi/fault.rs
```diff
@@ -294,8 +262,133 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+    F: FaultInjection,
+{
+    let mut suspect_proof = Proof {
+        latest_block_hash: state_machine_guard.latest_block_hash,
+        view_change_index: current_view_change_index,
+        signatures: Vec::new(),
+    };
+    suspect_proof
+        .sign(sumeragi.key_pair.clone())
+        .expect("must be able to perform signing");
```
appetrosyan, line 368:
```suggestion
    let suspect_proof = Proof {
        latest_block_hash: state_machine_guard.latest_block_hash,
        view_change_index: current_view_change_index,
        signatures: Vec::new(),
    }.sign(sumeragi.key_pair.clone())
     .expect("signing of suspect proof failed");
```

core/src/sumeragi/fault.rs
```diff
@@ -294,8 +262,133 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+
+    sumeragi.broadcast_msg(
+        Message::ViewChangeSuggested(ViewChangeSuggested::new(view_change_proof_chain.clone())),
+        &state_machine_guard.current_topology,
+    );
 }
 
+use tracing::{span, Level};
+
+use crate::sumeragi::tracing::instrument;
```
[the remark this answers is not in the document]
appetrosyan: `rustfmt` should do that.

core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
+                .lock()
+                .expect("take lock");
+            state_machine_guard.genesis_network.take()
+        };
 
-            state_machine_guard.current_topology = sumeragi
-                .try_get_online_topology(&state_machine_guard.current_topology)
-                .expect("enough peers to pass genesis");
+        if let Some(genesis_network) = maybe_genesis_network {
+            // Submit genesis
```
appetrosyan, line 440: These comments don't really explain anything.

core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
+            while read_index < state_machine_guard.transaction_cache.len() {
+                if state_machine_guard.transaction_cache[read_index].is_some() {
+                    let tx = state_machine_guard.transaction_cache[read_index]
+                        .take()
+                        .unwrap();
+                    if tx.is_expired(sumeragi.queue.ttl) {
+                        read_index += 1;
+                        continue;
                     }
+                    state_machine_guard.transaction_cache[write_index] = Some(tx);
```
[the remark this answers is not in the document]
appetrosyan: This can be done after merging into `iroha2-dev`

core/src/sumeragi/mod.rs
Arjentix: @appetrosyan, I'm not in the context. Do you mean `FaultInjectionTrait` or what?

appetrosyan: Yes. Specifically if we want to do fault injection, we should do it via a separate binary, not the way we do it now.

cli/src/torii/routing.rs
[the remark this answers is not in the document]
SamHSmith: I will scope it.

cli/src/torii/routing.rs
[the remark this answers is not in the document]
SamHSmith: will change this.

client/benches/tps/config.json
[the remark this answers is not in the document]
SamHSmith: But in practice it is relevant. We wouldn't set `max_txs_per_block` to `1` because that will reduce TPS. The ping time around the network cause rounds in the real world to take time. This means that for maximum throughput you want the largest possible block size. It should be obvious why you need to constrain it the other way though, infinity sized blocks take infinite time to process at each node and so no consensus is ever performed.

The general guide is that the `max_txs_per_block` should be as large as possible while still being able to be processed and pass consensus. My testing suggested 512 was a safe number.

data_model/src/trigger/set.rs
```diff
@@ -296,10 +296,10 @@ impl Set {
     pub fn handle_execute_trigger_event(&mut self, event: ExecuteTriggerEvent) {
         let pair;
         {
-        let entry = match self.by_call_triggers.get(&event.trigger_id.clone()) {
-            Some(entry) => entry,
-            None => return,
-        };
+            let entry = match self.by_call_triggers.get(&event.trigger_id.clone()) {
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/kura.rs
```diff
@@ -237,22 +234,48 @@ impl Kura {
...
-                    match strong_kura_reference
+                    let io_error = strong_kura_reference
                         .block_store
                         .lock()
                         .expect("lock on block store")
                         .append_block_to_chain(&serialized_block)
-                    {
-                        Ok(()) => {
+                        .err();
+                    match io_error {
```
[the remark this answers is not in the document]
SamHSmith: Because the borrow checker doesn't allow me to drop `strong_kura_reference` while matching on a value derived from it.

core/src/kura.rs
```diff
@@ -237,22 +234,48 @@ impl Kura {
...
                                 .lock()
                                 .expect("lock on block hash array")
                                 .push(block_hash);
                         }
-                        Err(error) => {
-                            error!(%error, "Failed to write block. Kura will now init again.");
-                            panic!("It is unclear what we should do here.");
+                        Some(error) => {
+                            let failed_to_store_block = serialized_block;
+                            let failed_to_store_block_reason = error;
```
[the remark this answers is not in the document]
SamHSmith: The error I will print when we panic. The `failed_to_store_block` could be used to try recover.

core/src/queue.rs
```diff
@@ -104,6 +104,28 @@ impl Queue {
             )
     }
 
+    pub fn prune(&self) {
```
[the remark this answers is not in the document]
SamHSmith: will address.

core/src/queue.rs
```diff
@@ -104,6 +104,28 @@ impl Queue {
...
+        let count = self.queue.len();
+        let mut holder = Vec::new();
+        holder.reserve(count);
```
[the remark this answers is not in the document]
SamHSmith: yep

core/src/queue.rs
```diff
@@ -104,6 +104,28 @@ impl Queue {
...
+        for i in 0..count {
+            match self.queue.pop() {
+                Some(tx_hash) => {
+                    let tx = self.txs.get(&tx_hash).unwrap().clone();
```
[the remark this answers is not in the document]
SamHSmith: I remove the prune function because I had stopped using it.

core/src/queue.rs
```diff
@@ -104,6 +104,28 @@ impl Queue {
             )
     }
 
+    pub fn prune(&self) {
+        let count = self.queue.len();
+        let mut holder = Vec::new();
+        holder.reserve(count);
+        for i in 0..count {
+            match self.queue.pop() {
```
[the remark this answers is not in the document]
SamHSmith: good suggestion.

core/src/sumeragi/mod.rs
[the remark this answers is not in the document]
SamHSmith: Yes. I also like it because by being specific about what operations you are performing it's easier to modify this later. Say you wanted to select a group of 4 unique peers. The path from what I've written to this hypothetical version is shorter than from I had used `choose`. The more basic the language features you are using, the more things they can be used for. Specific functions like `choose` have only one very specific function, and if you wanted to continue that pattern you would also need `choose_multiple_overlapping` and `choose_multiple_unique` et cetera. Or what if you decided you wanted to try out gathering a list of 8 peers where a peer can be repeated at most 3 times. You get the point.

core/src/sumeragi/view_change.rs
```diff
@@ -112,249 +64,103 @@ impl Proof {
...
-        let mut previous_proof = EmptyChainHash::default().into();
-        for proof in &self.proofs {
-            if proof.has_same_state(latest_block, &previous_proof)
-                && proof.verify(peers, max_faults)
-            {
-                previous_proof = proof.hash();
-            } else {
-                return false;
+    ) -> usize {
+        let mut valid_view_change_counter = 0;
```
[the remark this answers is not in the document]
SamHSmith: yea okay

core/src/sumeragi/fault.rs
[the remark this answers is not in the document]
SamHSmith: maybe it should, but it doesn't. I move them though.

core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
+        let _enter_for_sumeragi_cycle = span_for_sumeragi_cycle.enter();
 
-                    println!("Genesis being submitted by {}", sumeragi.peer_id.public_key);
-                    println!("Online peers are {:?}", sumeragi.get_online_peer_keys());
+        let mut state_machine_guard = {
+            let span_for_sumeragi_guard_lock = span!(
+                Level::TRACE,
+                "Sumeragi Main Thread Acquire state machine data lock."
+            );
+            let _enter_for_sumeragi_guard_lock = span_for_sumeragi_guard_lock.enter();
```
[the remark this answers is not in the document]
SamHSmith: No, we are descriping the thing that the span is tracking. We want to know how much congestion there is for this lock.

core/src/sumeragi/fault.rs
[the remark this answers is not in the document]
SamHSmith: Yes. Although I would only be for such a change if it was as fast as this. But I'm sure that's the case of what you are describing.

core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
-                            }
-                        }
+        // Transaction Cache
+        {
+            // We prune expired transactions. We do not check if they are in the blockchain, it would be a waste.
+            let mut read_index = 0;
+            let mut write_index = 0;
+            while read_index < state_machine_guard.transaction_cache.len() {
+                if state_machine_guard.transaction_cache[read_index].is_some() {
+                    let tx = state_machine_guard.transaction_cache[read_index]
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
+                    maybe_incoming_message = Some(Message::BlockCommitted(block_committed));
+                }
+                other => {
+                    maybe_incoming_message = Some(other);
+                }
             }
-            let mut state_machine_guard = sumeragi.sumeragi_state_machine_data.lock().unwrap();
-            if state_machine_guard.sumeragi_thread_should_exit {
-                return;
+            match foreign_proof_chain {
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/queue.rs
```diff
@@ -215,6 +237,36 @@ impl Queue {
         }
     }
 
+    /// Pops single transaction.
```
outoftardis, line 240:
```suggestion
    /// Pops a single transaction.
```

config/src/sumeragi.rs
```diff
@@ -8,10 +8,9 @@ use iroha_crypto::prelude::*;
 use iroha_data_model::{prelude::*, transaction};
 use serde::{Deserialize, Serialize};
 
-/// Default Amount of time peer waits for the `CreatedBlock` message
-/// after getting a `TransactionReceipt`.
+/// Default Amount of time peer waits for transactions before creating a block.
 pub const DEFAULT_BLOCK_TIME_MS: u64 = 1000;
-/// Default amount of time Peer waits for `CommitMessage` from the proxy tail.
+/// Default amount of time allocated for voting on a block before a peer asking for a view change.
```
outoftardis, line 13:
```suggestion
/// Default amount of time allocated for voting on a block before a peer can ask for a view change.
```

core/src/sumeragi/view_change.rs
```diff
@@ -20,90 +20,42 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
+    fn signature_payload(&self) -> Hash {
+        let mut buf = [0_u8; Hash::LENGTH + std::mem::size_of::<u64>()];
+        buf[..Hash::LENGTH].copy_from_slice(self.latest_block_hash.as_ref());
+        buf[Hash::LENGTH..].copy_from_slice(&self.view_change_index.to_le_bytes());
+        // Now we hash the buffer to produce a payload that is completely
+        // different between view change proofs in the same sumeragi round.
+        Hash::new(buf)
     }
 
     /// Signs this message with the peer's public and private key.
```
outoftardis, line 38:
```suggestion
    /// Sign this message with the peer's public and private key.
```

core/src/sumeragi/view_change.rs
```diff
@@ -20,90 +20,42 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
+        let signature = Signature::new(key_pair, self.signature_payload().as_ref())?;
+        self.signatures.push(signature);
+        Ok(())
     }
 
-    /// Adds verified signatures of `other` to self.
-    pub fn merge_signatures(&mut self, other: SignaturesOf<Proof>) {
-        self.signatures
-            .extend(other.into_verified_by_hash(&self.hash()))
+    /// Verifies and adds signatures of other to this proof.
```
outoftardis, line 49:
```suggestion
    /// Verify the signatures of `other` and add them to this proof.
```

core/src/sumeragi/fault.rs
```diff
@@ -310,991 +403,465 @@ pub fn run_sumeragi_main_loop<F>(
...
+                Message::TransactionGossip(tx_gossip) => {
+                    for transaction in tx_gossip.txs {
+                        let tx_maybe = VersionedAcceptedTransaction::from_transaction(
+                            transaction.into_v1(),
+                            &sumeragi.transaction_limits,
+                        );
+                        if !tx_maybe.is_ok() {
+                            continue;
+                        }
+                        let tx = tx_maybe.unwrap();
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/sumeragi/fault.rs
SamHSmith: yes

core/src/sumeragi/fault.rs
SamHSmith: I tried that but didn't find it helped readability.

core/src/sumeragi/fault.rs
SamHSmith: This one I think is a bit too much. It's too *clever*.

core/src/sumeragi/fault.rs
SamHSmith: Not really because then you would have to link to that single place from all the places this bool is currently set. The problem is the problem, there is nothing we can do that makes the problem go away or become simpler than it actually is.

core/src/sumeragi/view_change.rs
```diff
@@ -20,90 +20,42 @@ use crate::block::{EmptyChainHash, VersionedCommittedBlock, VersionedValidBlock}
...
-        latest_block: HashOf<VersionedCommittedBlock>,
-        key_pair: KeyPair,
-    ) -> Result<Self> {
-        let payload = ProofPayload {
-            reason: Reason::NoTransactionReceiptReceived(NoTransactionReceiptReceived),
-            previous_proof,
-            latest_block,
-        };
-        Self::from_payload(payload, key_pair)
+    fn signature_payload(&self) -> Hash {
```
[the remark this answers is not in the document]
SamHSmith: pub it is

core/src/sumeragi/view_change.rs
```diff
@@ -112,249 +64,103 @@ impl Proof {
...
             .collect();
-        let n_signatures = self
-            .signatures
-            .verified_by_hash(self.hash())
-            .filter(|signature| peer_public_keys.contains(signature.public_key()))
-            .count();
+
+        let mut valid_count = 0;
+        let signature_payload = self.signature_payload();
+        for signature in &self.signatures {
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/sumeragi/view_change.rs
```diff
@@ -112,249 +64,103 @@ impl Proof {
...
-                Hash::prehashed([i; Hash::LENGTH]).typed(),
-                proof_chain.latest_hash(),
-                latest_block,
-                key_pair_1.clone(),
-            )?
-            .sign(key_pair_2.clone())?;
-            proof_chain.push(proof);
+    fn prune(&mut self, latest_block: &HashOf<VersionedCommittedBlock>) {
+        let mut valid_count = 0;
+        for (i, proof) in self.iter().enumerate() {
```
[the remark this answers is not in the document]
SamHSmith: I'm doing it anyway

core/src/sumeragi/view_change.rs
```diff
@@ -112,249 +64,103 @@ impl Proof {
...
-            )?
-            .sign(key_pair_2.clone())?;
-            proof_chain.push(proof);
+    fn insert_proof(
+        &mut self,
+        peers: &HashSet<PeerId>,
+        max_faults: usize,
+        latest_block: &HashOf<VersionedCommittedBlock>,
+        new_proof: &Proof,
+    ) {
```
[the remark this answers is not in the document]
SamHSmith: I will add it. And you shouldn't think I worry about 30 more bytes in the `.text` of the program. It's negligable.

core/src/sumeragi/network_topology.rs
```diff
@@ -226,35 +213,27 @@ impl Topology {
...
         *self = self
             .clone()
             .into_builder()
             .at_block(block)
-            .with_view_changes(ViewChangeProofs::empty())
-            .build()
-        .expect("Given a valid Topology, it is impossible to have error here.")
-        */
+            .build(0)
+            .expect("Given a valid Topology, it is impossible to have error here.");
```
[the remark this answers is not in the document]
SamHSmith: sure

core/src/sumeragi/mod.rs
```diff
@@ -234,20 +232,25 @@ impl Sumeragi {
...
+        if self
+            .internal
             .incoming_message_sender
             .lock()
             .expect("Lock on sender")
-            .send(msg);
+            .try_send(msg)
+            .is_err()
+        {
+            error!("Sumeragi is so slow that incomming messages have to be dropped. This peer is in practice faulty.");
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/sumeragi/mod.rs
```diff
@@ -188,42 +178,50 @@ impl Sumeragi {
...
     }
 
     pub fn get_random_peer_for_block_sync(&self) -> Option<Peer> {
-        use rand::{prelude::SliceRandom, SeedableRng};
+        use rand::{prelude::SliceRandom, RngCore, SeedableRng};
 
         let rng = &mut rand::rngs::StdRng::from_entropy();
-        self.internal
-            .wsv
+        let mut vec = self
```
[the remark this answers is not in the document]
SamHSmith: I will take your suggestion. But the reason I tend to split things up is because I'm dividing up the problem in my head. Long chains of method calls on top of each other are harder to debug and harder to understand than discrete imperitive steps. One is using `object.method1().method2().method3()...` as a replacement for
```
procedure1(object);
procedure2(object);
procedure3(object);
```
Both have mutability, except in the former it is hidden and never declared openly. I think the later method is better because it's more powerful and allows you to insert code to inspect the state of `object` in between the operations.

core/src/sumeragi/mod.rs
SamHSmith: sure

core/src/sumeragi/mod.rs
SamHSmith: I disagree. After I added the transaction cache this is basically a non issue. I will bump it up to a slightly higher value and then leave it because playing around with this value is not worth our time.

cli/src/torii/routing.rs
SamHSmith: yes

client/benches/tps/lib.rs
SamHSmith: thx

core/src/block_sync.rs
SamHSmith: yup

core/src/queue.rs
SamHSmith: sure

core/src/sumeragi/fault.rs
SamHSmith: They were inherited. Turns out the drops could just be removed.

core/src/sumeragi/mod.rs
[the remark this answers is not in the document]
SamHSmith: "I don't see why sorting couldn't have been done before collect()". Could you provide code of such a thing? I applied your suggestion without thinking much and I don't think it is possible. You must have all the data to be able to sort it. You can't sort on an iterator.

core/src/sumeragi/fault.rs
SamHSmith: Yes. The others have mutable counters because I start off by doing the most flexible thing because I'm not exactly sure what it is I'm making. (This probably means I should do a second pass on my code more often.) In this instance it was a very particular choice in order to be fast.

core/src/sumeragi/fault.rs
SamHSmith: Sorry. The API does not work that way. `sign` inserts signatures into an already existing proof. That is how the proof "votes" are accumulated across peers. And converting it to a functional style would make the borrow checker go off.

core/src/sumeragi/fault.rs
SamHSmith: true

config/src/sumeragi.rs
SamHSmith: thx

core/src/sumeragi/mod.rs
```diff
@@ -188,42 +178,60 @@ impl Sumeragi {
...
-        });
+        std::thread::Builder::new()
+            .name("sumeragi thread".to_string())
+            .spawn(move || {
+                fault::run_sumeragi_main_loop(
+                    &sumeragi.internal,
+                    latest_block_hash,
+                    latest_block_height,
+                );
+                info!("Sumeragi Thread has Shutdown");
```
outoftardis, line 233:
```suggestion
                info!("Sumeragi Thread Shutdown");
```
```suggestion
                info!("Sumeragi Thread has shut down");
```

Shutdown is a noun, if you want a verb, you should use "shut down"

Erigara (commented):
I would suggest fix linter warnings: 
```
cargo +nightly-2022-08-15 lints clippy --workspace --tests --benches --examples --all-features
```

core/src/sumeragi/fault.rs
```diff
@@ -310,1495 +400,970 @@ pub fn run_sumeragi_main_loop<F>(
...
+                    &mut view_change_proof_chain,
+                    current_view_change_index,
+                );
+                sent_transaction_time = Instant::now();
+            }
+        }
+
+        if state_machine_guard.current_topology.role(&sumeragi.peer_id) == Role::ObservingPeer {
+            if maybe_incoming_message.is_some() {
+                let incoming_message = maybe_incoming_message.take().unwrap();
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/sumeragi/fault.rs
```diff
@@ -310,1495 +400,970 @@ pub fn run_sumeragi_main_loop<F>(
...
-                                    Ok(()) => (),
-                                    Err((_, crate::queue::Error::InBlockchain)) | Ok(()) => (),
-                                    Err((_, err)) => {
-                                        error!(%err, "Error while pushing transaction into queue?");
-                                    }
+        } else if state_machine_guard.current_topology.role(&sumeragi.peer_id) == Role::Leader {
+            if maybe_incoming_message.is_some() {
+                use crate::sumeragi::Message::TransactionForwarded;
+
+                let msg = maybe_incoming_message.take().unwrap();
```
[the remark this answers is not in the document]
SamHSmith: yes

core/src/sumeragi/fault.rs
```diff
@@ -310,1495 +400,970 @@ pub fn run_sumeragi_main_loop<F>(
...
+            } else {
+                // if there is no message sleep
+                should_sleep = true;
+            }
 
-                if voting_block_option.is_some() {
-                    // count votes
-
-                    let validating_peers = state_machine_guard.current_topology.peers_set_a();
+            if voting_block_option.is_some() {
```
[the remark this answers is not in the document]
SamHSmith: sure thing

core/src/sumeragi/fault.rs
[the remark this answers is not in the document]
SamHSmith: Well no actually, non of the changes to `if let` you suggest actually work because they get killed by the borrow checker. It is written `is_some` followed by an unwrap because that doesn't consume the value.

SamHSmith:
> I would suggest linter warnings:
> 
> ```
> cargo +nightly-2022-08-15 lints clippy --workspace --tests --benches --examples --all-features
> ```

This is to be done in a seperate commit. The lints don't need to be fixed for it to go into staging. They need to be addressed before going into dev, although Aleksandr said that dev won't pass the new lints either so maybe we fix them all in one pass.

core/src/sumeragi/mod.rs
SamHSmith: thank you

core/src/sumeragi/fault.rs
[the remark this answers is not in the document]
Erigara: I think `Option::take` would help in this case if value is `None` this is no-op anyway.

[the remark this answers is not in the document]
SamHSmith: That might be so. How it is finally written isn't really that important to me. You could go and change all these things to the way you prefer after we've merged. What matters to me is getting things done faster. `if a.is_some() { a.unwrap() ...` and `if let` do the exact same thing. So changing from one to the other is shuffling papers around on the desk without actually changing anything for the end user.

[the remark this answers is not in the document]
Erigara: So i agree with you that this 2 approaches do the same thing.
However i think that such stylistic choices go against our guidelines and because of this we patch then with `allow` attributes.

>  without actually changing anything for the end user

Beside end users there developers who actually read and write code. 
I believe that such concessions tend to accumulate and it easier to fix them sooner than later.

I think we need third opinion on this one: @appetrosyan what do you think?

core/src/wsv.rs
```diff
@@ -577,14 +569,12 @@ impl WorldStateView {
     }
 
     /// Returns iterator over blockchain blocks starting with the block of the given `height`
-    pub fn blocks_from_height(
-        &self,
-        height: usize,
-    ) -> impl Iterator<Item = VersionedCommittedBlock> + '_ {
+    pub fn blocks_from_height(&self, height: usize) -> Vec<VersionedCommittedBlock> {
```
appetrosyan, line 572: `impl Iterator` seems better. You can postpone collecting if you need to.

appetrosyan (approved):

core/src/sumeragi/fault.rs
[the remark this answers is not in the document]
appetrosyan: @SamHSmith 's style needs to be reworked anyway. I'm thinking that this is better done after we have a safe-to-work-with Sumeragi in staging.

Erigara (approved):

core/src/wsv.rs
SamHSmith: This can be explored in the future. I made this choice to avoid contention between threads since the wsv is currently very global.

Issue #2646, opened 2022-08-24
[ci] Fix `telemetry` job in `I2::Dev::Publish` workflow


s8sato:
https://github.com/hyperledger/iroha/issues/2461#issuecomment-1201441039

PR #2662, opened 2022-08-27
[ci]: Fix: Make publishing workflows sequential, and complements to #2427

s8sato:
### Description of the Change

1. Make publishing workflows sequential:
    Whenever `iroha2-base:<tag>` updates, `iroha2:<tag>` also updates on the base image. If `Dockerfile.base` has no change, only `iroha2:<tag>` updates
2. Exclude other `.yml` files than workflows from the PR workflow triggers
3. Refactor workflows, etc

### Issue

- None

### Benefits

1. `iroha2-base:<tag>` never diverges from the history of `iroha2:<tag>`
2. Changes to docker-compose files never trigger PR workflows

### Possible Drawbacks

- None


.github/workflows/iroha2-release-pr.yml
```diff
@@ -65,6 +65,7 @@ jobs:
         with:
           context: .
           push: false
+          load: true
           tags: hyperledger/iroha2:dev
```
s8sato, line 65: Making the locally built `hyperledger/iroha2:dev` image available for subsequent steps

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -18,10 +18,9 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  CARGO_INCREMENTAL: 0
```
s8sato, line 21: `CARGO_INCREMENTAL=0` is implied by [`Swatinem/rust-cache`](https://github.com/Swatinem/rust-cache)

.github/workflows/iroha2-dev-nightly.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Login to DockerHub
-        uses: docker/login-action@v2
+      - uses: docker/setup-buildx-action@v2
```
appetrosyan, line 10: This is a bug. We **should have** buildx set up inside the container.

.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -17,25 +17,28 @@ concurrency:
...
         if: always()
-      - name: no_std compatibility check
-        run: cargo nono check --package iroha_data_model --no-default-features
+        run: cargo +nightly-2022-08-15 doc --no-deps --quiet --no-default-features
+      - name: Documentation with all features enabled
         if: always()
-      - name: Documentation check
-        run: if cargo doc -q 2>&1 | grep '^'; then exit 1; fi
+        run: cargo +nightly-2022-08-15 doc --no-deps --quiet --all-features
+      - name: no_std compatibility
```
appetrosyan, line 42: These shouldn't be necessary. We are using conditional compilation where we should instead be using LTO.

.github/workflows/iroha2-dev-pr-title.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           regex: '^\[feature\] #\d+(, #\d+)*: .+$'
       # - name: Add feature label
       #   uses: actions-ecosystem/action-add-labels@v1
-      #   if: ${{ steps.feature-match.outputs.match != '' }}
+      #   if: steps.feature-match.outputs.match != ''
```
appetrosyan, line 21: I don't see why this change was necessary, but OK.

.github/workflows/iroha2-dev-pr-unstable.yml
```diff
@@ -1,4 +1,4 @@
-name: I2::Dev::Tests (Unstable)
```
appetrosyan, line 1: Actually, let's get rid of this workflow, and move it as a separate `job` inside `I2::Dev::Tests`.

appetrosyan: The original reason we had this as a separate workflow was because we couldn't restart the failed jobs, just failed workflows. Now that we can re-run failed jobs, we should merge it back.

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -18,10 +18,9 @@ concurrency:
...
 
 jobs:
-  check:
+  checks:
```
appetrosyan, line 23:
```suggestion
  check:
```

.github/workflows/iroha2-dev-update_base_coverage.yml
```diff
@@ -1,4 +1,4 @@
-name: I2::Dev::Update Base Coverage
+name: I2::Dev::UpdateBaseCoverage
```
appetrosyan, line 1:
```suggestion
name: I2::Dev::BaseCoverage
```

.github/workflows/iroha2-dev.yml
```diff
@@ -91,6 +95,8 @@ jobs:
           path: target/schema
 
   telemetry:
+    # FIXME #2646
```
appetrosyan, line 98: Noted

appetrosyan: #2646

.github/workflows/iroha2-release.yml
```diff
@@ -1,40 +1,52 @@
...
+    steps:
+      - id: tag
+        if: github.ref == 'refs/heads/iroha2-${{ matrix.tag }}'
+        run: echo "::set-output name=tag::${{ matrix.tag }}"
+
+  dockerhub:
+    needs: tag
+    runs-on: ubuntu-latest
+    container:
+      image: 7272721/i2-ci:nightly
```
appetrosyan, line 29: We're getting a little overreliant on `matrix`, which contradicts both the simplicity and the performance gains from using a custom image.

telemetry/src/ws.rs
```diff
@@ -2,7 +2,7 @@
 #![allow(clippy::std_instead_of_core, clippy::std_instead_of_alloc)]
 use std::time::Duration;
 
-use chrono::Utc;
+use chrono::Local;
```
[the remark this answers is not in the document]
appetrosyan: Good!

appetrosyan (commented):
Overall LGTM. We should have a brief discussion and then merge.

.github/workflows/iroha2-dev-pr-static.yml
s8sato: Are you suggesting to remove `Documentation with all features enabled` and `no_std compatibility` checks?

.github/workflows/iroha2-dev-pr-title.yml
s8sato: I just intended to unify to either `if: expression` or `if: ${{ expression }}`

.github/workflows/iroha2-release.yml
s8sato: I agree. It would run as many jobs as the number of combinations in the matrix.
Previously I assumed that when used in `<job_id>.if` it could filter out unrelated jobs, but it turned out the matrix values were not evaluated at that point.
I'll take another way

.github/workflows/iroha2-dev-nightly.yml
s8sato: I added `docker/setup-buildx-action` not because I saw errors but just because it seemed to be the best practice. I'll just remove it then

.github/workflows/iroha2-dev-pr-static.yml
appetrosyan: Just the `--all-features` documentation.

.github/workflows/iroha2-dev-pr-title.yml
appetrosyan: Got it. Makes sense.

.github/workflows/iroha2-dev-pr-static.yml
appetrosyan: Can be done in a separate PR.

appetrosyan (dismissed):

.github/workflows/iroha2-dev-pr-static.yml
s8sato: #2663

s8sato: Done in this PR

pesterev (approved):

appetrosyan (approved):

.github/workflows/iroha2-dev.yml
```diff
@@ -1,7 +1,9 @@
 name: I2::Dev::Publish
 
 on:
-  push:
+  workflow_run:
+    workflows: [I2::Base::Publish]
+    types: [completed]
     branches: [iroha2-dev]
```
s8sato, line 7: This is mis-implementation. The workflow was not triggered after this PR merge

s8sato: Problems:

- [ ] `workflow_run` event lacks `paths` filters
- [ ] Not sure if `completed` condition reacts to the preceding workflow skipped
- [ ] Use of [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) needs to be committed to the default branch
	image

s8sato: Solution:
- Publish the base image not as a workflow but simply as an additional step to each workflow

Issue #2667, opened 2022-08-29
[refactor] Replace `&Option<T>` with `Option<&T>`


Arjentix:
We are using `&Option<T>` i.e. in `GenesisNetworkTrait`.
But [it looks like](https://www.reddit.com/r/rust/comments/wqsxk2/is_it_better_to_pass_optiont_or_optiont/) it's better to use `Option<&T>` instead cause it's easier to create and manipulate with.

Probably someday it will be a new lint in `clippy` or in our `lints`. But for now let's just search for all usages of `&Option<T>` and replace it with `Option<&T>`.

PR #2670, opened 2022-08-29
[fix] #2667: Refactor &Option<T> to Option<&T>

JonathanPlasse:
Signed-off-by: Jonathan Plasse <13716151+JonathanPlasse@users.noreply.github.com>

### Description of the Change

Replace `&Option<T>` by `Option<&T>` as described in  #2667

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -123,7 +123,7 @@ where
             G::from_configuration(
                 args.submit_genesis,
                 RawGenesisBlock::from_path(genesis_path)?,
-                &Some(config.genesis.clone()),
+                Some(&config.genesis),
                 &config.sumeragi.transaction_limits,
             )
             .wrap_err("Failed to initialize genesis.")?
--- a/crypto/src/merkle.rs
+++ b/crypto/src/merkle.rs
@@ -91,7 +91,7 @@ impl<T> FromIterator<HashOf<T>> for MerkleTree<T> {
         let mut tree = Vec::with_capacity(2_usize.pow(height + 1));
         while let Some(r_node) = queue.pop_back() {
             if let Some(l_node) = queue.pop_back() {
-                queue.push_front(Self::nodes_pair_hash(&l_node, &r_node));
+                queue.push_front(Self::nodes_pair_hash(l_node.as_ref(), r_node.as_ref()));
                 tree.push(r_node);
                 tree.push(l_node);
             } else {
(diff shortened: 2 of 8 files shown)


SamHSmith:
You missed these two:
```
client/benches/torii.rs:36
client/benches/torii.rs:135
```
And you need to rename both the PR title and the commit to `[fix] #2667: Refactor &Option<T> to Option<&T>`
Then we can merge your contribution.

JonathanPlasse:
@SamHSmith, thank you for your help.
I do not understand how I missed it.
I ran `cargo check`, `cargo build`, `cargo test` and it did not show me an error.

cli/src/lib.rs
```diff
@@ -123,7 +123,7 @@ where
             G::from_configuration(
                 args.submit_genesis,
                 RawGenesisBlock::from_path(genesis_path)?,
-                &Some(config.genesis.clone()),
+                Some(&config.genesis.clone()),
```
Arjentix, line 126: Could it be just:

```rust
Some(&config.genesis)
```

?

Like in the other places

crypto/src/merkle.rs
```diff
@@ -212,8 +212,8 @@ impl<T> MerkleTree<T> {
 
     #[cfg(feature = "std")]
     fn nodes_pair_hash(
-        l_node: &Option<HashOf<T>>,
-        r_node: &Option<HashOf<T>>,
+        l_node: Option<&HashOf<T>>,
+        r_node: Option<&HashOf<T>>,
```
Arjentix, line 216: Please, also take a look at **line 75** in this file. It also contains `&Option<...>` instead of `Option<&...>`

Arjentix (changes requested):
First of all, thank you for your contribution!

We will try to make more *good first issue* tasks.

We would be happy to see other PRs from you in the future!

cli/src/lib.rs
JonathanPlasse: Yes

crypto/src/merkle.rs
JonathanPlasse: It breaks the trait implementation of `CompleteBTree<T>::get()` if we use `Option<Option<&...>>` instead of `Option<&Option<...>>`.
So I do not think it is possible to change it.

SamHSmith:
> @SamHSmith, thank you for your help. I do not understand how I missed it. I ran `cargo check`, `cargo build`, `cargo test` and it did not show me an error.

We have this tool `cargo lints`. This is the command the CI ran to find those errors. `cargo +nightly-2022-08-15 lints clippy --workspace --benches --tests --examples --quiet --all-features`

SamHSmith (commented):
LGTM

SamHSmith (approved):

Arjentix (approved):

PR #2663, opened 2022-08-27
[ci]: Feature: commit hash in image, toolchain file again, UI isolation, schema tracking

s8sato:
### Description of the Change

1. Make a published image have the reference to Git commit
2. Reintroduce a toolchain file removed at 162d8e86
3. Isolate UI tests from regular coverage tests
4. Put `schema.json` under version control

### Issue

- None

### Benefits

1. Mapping between commits and images
2. Consistent dev environments
3. Separate run units for `trybuild` and other tests
4. Tracking of schema changes

### Possible Drawbacks

- None

#### Get the Git commit hash of `IMAGE`

```bash
docker inspect --format='{{.Config.Labels.commit}}' $IMAGE
```

#### List the images built at `COMMIT`

```bash
docker images --filter "label=commit=$COMMIT"
```

#### Update `schema.json` on commit

```diff
# .git/hooks/pre-commit
+ cargo run --bin kagami -- schema >docs/source/references/schema.json
+ git add docs/source/references/schema.json
```

PR #2680, opened 2022-08-31
[ci]: Enable sequential publish of images, after #2662

s8sato:
### Description of the Change

1. Publish `iroha2-base` and `iroha2` images in sequence
2. Modify trigger condition of `trybuild` tests

### Issue

1. https://github.com/hyperledger/iroha/pull/2662#discussion_r959346170
2. https://github.com/hyperledger/iroha/pull/2663#discussion_r959424599

### Benefits

1. `iroha2-base:<tag>` never diverges from the history of `iroha2:<tag>`

### Possible Drawbacks

- None


.github/workflows/iroha2-dev.yml
```diff
@@ -19,10 +17,22 @@ jobs:
       image: 7272721/i2-ci:nightly
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
```
s8sato, line 21: This is a requirement from [`tj-actions/changed-files`](https://github.com/tj-actions/changed-files#usage) for `push` workflows

.github/workflows/iroha2-dev.yml
```diff
@@ -19,10 +17,22 @@ jobs:
...
+      - uses: tj-actions/changed-files@v29.0.2
```
BAStos525, line 22:
```
Run tj-actions/changed-files@v29.0.2
changed-files-diff-sha
  Verifying git version...
  Valid git version found: (2.34.1)
  Error: Process completed with exit code 128.
```

s8sato: Could you provide a link to the error log?

BAStos525: I tried to test [on my fork](https://github.com/BAStos525/soramitsu-iroha/runs/8130280446?check_suite_focus=true)

s8sato:
Converted to draft, waiting for #2691

.github/workflows/iroha2-dev.yml
s8sato: Do you have any alternative ideas to filter out the base image publishing @BAStos525  ?
I've on my fork tried `workflow_call` and `workflow_run`, both of them did not work

.github/workflows/iroha2-dev.yml
```diff
@@ -19,10 +17,22 @@ jobs:
...
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push iroha2-base image
+        if: contains(steps.changed-files.outputs.modified_files, 'Dockerfile.base')
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: hyperledger/iroha2-base:dev
+          labels: commit=${{ github.sha }}
+          file: Dockerfile.base
```
BAStos525, line 35:
```suggestion
      - uses: dorny/paths-filter@v2.10.2
        id: filter
        with:
          filters: |
            base-image:
              - 'Dockerfile.base'
      - uses: docker/login-action@v2
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      - name: Build and push iroha2-base image
        if: steps.filter.outputs.base-image == 'true'
        uses: docker/build-push-action@v3
        with:
          push: true
          tags: hyperledger/iroha2-base:dev
          labels: commit=${{ github.sha }}
          file: Dockerfile.base
```

But I have another one error with this action that we don't have with `load-rs` repo:
```
Run dorny/paths-filter@v2.10.2
/usr/bin/docker exec  70593ac1b69ec16d10ebedea742c0cf84665d9971872fa7e56001cb05bbb612d sh -c "cat /etc/*release | grep ^ID"
Get current git ref
  /usr/bin/git branch --show-current
  fatal: detected dubious ownership in repository at '/__w/soramitsu-iroha/soramitsu-iroha'
  To add an exception for this directory, call:
  
  	git config --global --add safe.directory /__w/soramitsu-iroha/soramitsu-iroha
Error: The process '/usr/bin/git' failed with exit code [12](https://github.com/BAStos525/soramitsu-iroha/runs/8169021883?check_suite_focus=true#step:4:13)8
```
Could you try @s8sato

.github/workflows/iroha2-dev-pr-ui.yml
```diff
@@ -2,7 +2,7 @@ name: I2::Dev::Tests::UI
 
 on:
   pull_request:
-    branches: [iroha2-dev]
+    branches: [iroha2-stable, iroha2-lts]
```
appetrosyan, line 5: I'd add `dev` at least temporarily. If it's the only failing workflow we can live without it.

Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM hyperledger/iroha2-base:$TAG AS builder
 
 WORKDIR /iroha
 COPY . .
+RUN  rm -f rust-toolchain.toml
```
appetrosyan, line 7: This is such a controversial addition to the repo that I'm inclined to remove and ban it. If we can't agree on having it, we should agree on getting rid of it permanently. If we need to keep this information somewhere, it could be kept away from the main code, just like `hooks`.

.github/workflows/iroha2-release.yml
```diff
@@ -16,6 +14,10 @@ jobs:
       image: 7272721/i2-ci:nightly
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - uses: tj-actions/changed-files@v29.0.2
+        id: changed-files
```
appetrosyan, line 20: I'd be careful. It's not necessarily true that all branches are in a consistent state, because we don't normally update with rebase automatically.

s8sato:
Rebased to 4 commits. The first 3 are the same with what you already reviewed @BAStos525  @appetrosyan

.github/workflows/iroha2-dev.yml
s8sato: I'd leave as is and merge, because this can be a problem specific to forks, as experiments on my fork report that both [`tj-actions/changed-files`](https://github.com/s8sato/iroha/actions/runs/3002076947) and [`dorny/paths-filter`](https://github.com/s8sato/iroha/actions/runs/3002856835) failed with exit code 128, which seems to be authentication problem

.github/workflows/iroha2-dev.yml
s8sato: https://github.com/hyperledger/iroha/pull/2680#discussion_r964408423

Dockerfile
s8sato: Left to #2690, including inspection whether the toolchain file relates or not

.github/workflows/iroha2-release.yml
s8sato: Could you rephrase it? I don't think our merge strategy affects the changes detection

.github/workflows/iroha2-dev.yml
BAStos525: Interesting thing here that [iroha2 longevity load-rs repository](https://github.com/soramitsu/iroha2-longevity-load-rs) is also [fork](https://github.com/e-ivkov/iroha2-longevity-load-rs). But `dorny/paths-filter` Actions works as expected.

s8sato: That would be the difference whether secrets are set properly or not

BAStos525: Assume that `tj-actions/changed-files` will work inside `hyperledger` repository, does it mean that we will get the workflows run fails of forks wherever this particular Action is specified?
If third-part Actions  doesn't work for us, we could try something with a [classical bash commands](https://www.meziantou.net/executing-github-actions-jobs-or-steps-only-when-specific-files-change.htm).

s8sato: What do you think about actually merging and observing the results, because:
- This is in accordance with the usage
- The exit code 128 implies an authentication error

BAStos525: Well, let's get a chance for it and see.

BAStos525 (dismissed):

.github/workflows/iroha2-dev-pr-label.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
```
s8sato, line 17: This is a fix for #2708 after observing an [error](https://github.com/hyperledger/iroha/runs/8228305240?check_suite_focus=true#step:3:428)

appetrosyan (approved):

.github/workflows/iroha2-dev-pr-label.yml
BAStos525: Did you try it? It still seems as the same error.

.github/workflows/iroha2-release.yml
appetrosyan: This check should not fail if the target branch has more recent commits that you don't have (i.e. the update with rebase button exists).

.github/workflows/iroha2-dev-pr-label.yml
s8sato: https://github.com/hyperledger/iroha/pull/2680#discussion_r964673683

BAStos525 (approved):

.github/workflows/iroha2-release.yml
s8sato: Your concern would be that if the diverging branches are identical with respect to `Dockerfile.base` it will detect no changes and fail to update the base image. In that case, there is actually no need to update the base image

s8sato:
> https://github.com/hyperledger/iroha/pull/2680#discussion_r964740904

Ok, `tj-actions/changed-files` does not help for some reason
https://github.com/hyperledger/iroha/runs/8302565687?check_suite_focus=true

PR #2693, opened 2022-09-02
[refactor] #2650: Add `ThreadHandler` to shutdown iroha submodules

### LINKED ISSUE #2650: Formalized thread shutdown
This is a problem with RAII and our tests. The temporary filesystem is destroyed at the same time as we shutdown iroha. In production this is not an issue. If we want to solve this properly there should be a iroha.shutdown() method that tells all the threads to exit and waits their termination. Then each test can call this at the end. Shall I create an issue for it in the "Polishing Kura" epic?

_Originally posted by @SamHSmith in https://github.com/hyperledger/iroha/pull/2636#discussion_r953608731_

Erigara:

### Description of the Change

- Add `ThreadHandler` struct to shutdown iroha submodules threads on `Iroha` shutdown (RAII).
- Rewrite `Kura` block writing loop: on shutdown `Kura` writes remaining blocks and then exits.
- Add temporary directory to `Peer` struct to delete directory after `Iroha` (and thus `Kura` thread).  

### Issue

Closes #2650.

### Benefits

Unified mechanism for stopping iroha submodules threads.

### Possible Drawbacks

None.

### Alternate Designs

Possibly better name for `ThreadHandler` :)

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -17,6 +17,7 @@ use iroha_config::iroha::Configuration;
 use iroha_core::{
     block_sync::{BlockSynchronizer, BlockSynchronizerTrait},
     genesis::{GenesisNetwork, GenesisNetworkTrait, RawGenesisBlock},
+    handler::ThreadHandler,
     kura::Kura,
     prelude::{World, WorldStateView},
     queue::Queue,
@@ -83,6 +84,20 @@ where
     pub block_sync: AlwaysAddr<B>,
     /// Torii web server
     pub torii: Option<Torii>,
+    /// Thread handlers
+    thread_handlers: Vec<ThreadHandler>,
+}
+
+impl<G, S, B> Drop for Iroha<G, S, B>
+where
+    G: GenesisNetworkTrait,
+    S: SumeragiTrait<GenesisNetwork = G>,
+    B: BlockSynchronizerTrait<Sumeragi = S>,
+{
+    fn drop(&mut self) {
+        // Drop thread handles first
+        let _thread_handles = core::mem::take(&mut self.thread_handlers);
+    }
 }
 
 impl<G, S, B> Iroha<G, S, B>
@@ -233,6 +248,8 @@ where
         let telemetry_started = Self::start_telemetry(telemetry, &config).await?;
         let kura = Kura::from_configuration(&config.kura, Arc::clone(&wsv), broker.clone())?;
 
+        let kura_thread_handler = Kura::start(Arc::clone(&kura));
+
         let sumeragi: AlwaysAddr<_> = S::from_configuration(
             &config.sumeragi,
             events_sender.clone(),
@@ -284,6 +301,7 @@ where
             kura,
             block_sync,
             torii,
+            thread_handlers: vec![kura_thread_handler],
         })
     }
 
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -13,7 +13,7 @@ use std::{
     fmt::Debug,
     io::{Read, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
-    sync::{Arc, Mutex, Weak},
+    sync::{Arc, Mutex},
 };
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
@@ -22,9 +22,12 @@ use iroha_config::kura::{Configuration, Mode};
 use iroha_crypto::HashOf;
 use iroha_logger::prelude::*;
 use iroha_version::scale::{DecodeVersioned, EncodeVersioned};
-use tokio::sync::mpsc::{channel, Receiver, Sender};
+use tokio::sync::mpsc::{channel, error::SendError, Receiver, Sender};
 
-use crate::{block::VersionedCommittedBlock, block_sync::ContinueSync, prelude::*, sumeragi};
+use crate::{
+    block::VersionedCommittedBlock, block_sync::ContinueSync, handler::ThreadHandler, prelude::*,
+    sumeragi,
+};
 
 /// The interface of Kura subsystem
 pub struct Kura {
@@ -37,13 +40,6 @@ pub struct Kura {
     broker: Broker,
     block_reciever: Mutex<Receiver<VersionedCommittedBlock>>,
     block_sender: Sender<VersionedCommittedBlock>,
-    block_thread_handle: Mutex<Option<BlockThreadHandle>>,
-}
-
-/// Handle to block receiver thread with ability to stop it
-struct BlockThreadHandle {
-    thread_handle: std::thread::JoinHandle<()>,
-    shutdown_sender: tokio::sync::oneshot::Sender<()>,
 }
 
 impl Kura {
@@ -79,48 +75,25 @@ impl Kura {
             broker,
             block_reciever: Mutex::new(block_reciever),
             block_sender,
-            block_thread_handle: Mutex::new(None),
-        });
-
-        // How do we avoid the spawned thread taking on a
-        // life of it's own, aka keeping kura alive and running
-        // long after it has been dropped by all who needed it.
-        // By using a weak reference, see `kura_recieve_blocks_thread`
-        // for details on how this works.
-        let kura_weak_reference = Arc::downgrade(&kura);
-
-        // Oneshot channel to allow forcefully stopping the thread.
-        let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel();
-
-        let thread_handle = std::thread::spawn(move || {
-            Self::kura_recieve_blocks_thread(&kura_weak_reference, shutdown_receiver);
-        });
-
-        *kura.block_thread_handle.lock().expect("able to lock") = Some(BlockThreadHandle {
-            thread_handle,
-            shutdown_sender,
         });
 
         Ok(kura)
     }
 
-    /// Forcefully shut down the Kura thread.
-    /// Warning: any blocks currently in queue to be stored will be lost.
-    #[allow(clippy::expect_used)]
-    pub fn force_shutdown(&self) {
-        if let Some(BlockThreadHandle {
-            thread_handle,
-            shutdown_sender,
-        }) = self
-            .block_thread_handle
-            .lock()
-            .expect("lock block thread handle")
-            .take()
-        {
-            // Error here indicates the thread is already stopped, which is fine.
-            let _res = shutdown_sender.send(());
-            thread_handle.join().expect("able to join");
-        }
+    /// Start the Kura thread
+    pub fn start(kura: Arc<Self>) -> ThreadHandler {
+        // Oneshot channel to allow forcefully stopping the thread.
+        let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel();
+
+        let thread_handle = std::thread::spawn(move || {
+            Self::kura_recieve_blocks_loop(&kura, shutdown_receiver);
+        });
+
+        let shutdown = move || {
+            let _result = shutdown_sender.send(());
+        };
+
+        ThreadHandler::new(Box::new(shutdown), thread_handle)
     }
 
     /// Loads kura from configuration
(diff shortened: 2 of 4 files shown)


core/src/kura.rs
```diff
@@ -79,15 +75,19 @@ impl Kura {
...
+
+    /// Start kura thread
+    pub fn start(kura: &Arc<Self>) -> ThreadHandler {
         // How do we avoid the spawned thread taking on a
         // life of it's own, aka keeping kura alive and running
         // long after it has been dropped by all who needed it.
         // By using a weak reference, see `kura_recieve_blocks_thread`
         // for details on how this works.
-        let kura_weak_reference = Arc::downgrade(&kura);
+        let kura_weak_reference = Arc::downgrade(kura);
```
SamHSmith, line 90: The weak reference was the original shutdown mechanism. Since you are replacing it we can revert to just using an Arc. This simplifies the code inside the kura thread loop since we don't need to upgrade the reference.

core/src/kura.rs
```diff
@@ -224,32 +204,27 @@ impl Kura {
         weak_kura_reference: &Weak<Kura>,
```
SamHSmith, line 204: So this can be changed to `Arc<Kura>`

core/src/kura.rs
```diff
@@ -224,32 +204,27 @@ impl Kura {
...
-            if shutdown_receiver.try_recv().is_ok() {
-                info!("Kura block thread is being forcibly shut down");
-                return;
-            }
-
+        let mut is_closed = false;
+        // Here we essentially check if the thread should keep running.
+        // If all other references to `Kura` have been dropped, then
+        // the thread associated with that `Kura` instance should exit.
+        while let Some(strong_kura_reference) = weak_kura_reference.upgrade() {
```
SamHSmith, line 211: And this can become `loop {`

core/src/kura.rs
```diff
@@ -298,21 +274,28 @@ impl Kura {
...
+    /// Return error if kura thread was closed previously
     #[allow(clippy::expect_used)]
-    pub fn blocking_store_block(&self, block: VersionedCommittedBlock) {
-        self.block_sender
-            .blocking_send(block)
-            .expect("Kura block channel is still open.");
+    pub fn blocking_store_block(
+        &self,
+        block: VersionedCommittedBlock,
+    ) -> Result<(), SendError<VersionedCommittedBlock>> {
```
SamHSmith, line 297: This would be a leak of system responsibility. I/O is separated into it's own subsystem so that the rest of the codebase can be free from I/O code. Producing blocks and consensus is sumeragi's responsibility, once the blocks exist they are fully kura's responsibility and the API should reflect that. If you want to handle the case where the kura thread has been shutdown and the channel closed, do it in this function. Emit an error with `error!();`, but sumeragi should not be aware that this has happened. Maybe you could signal to iroha to shutdown, which is what the expect was doing, although improperly.

What cases could such an error occur? Either we are shutting down iroha and the sumeragi thread will be shutdown as well so missing a block does not matter. Or the kura thread has crashed at which point our ability to reason about the problem and write code that acts properly goes to zero. In both cases the best we can do is produce an error message and exit the program.

SamHSmith (changes requested):
I think 2 things should be changed. Otherwise it looks very good.

core/src/kura.rs
appetrosyan: A little verbose but I agree.

Erigara: Good point, thanks.

core/src/lib.rs
```diff
@@ -74,6 +74,42 @@ pub trait IsInBlockchain {
...
+        clippy::expect_used,
+        clippy::std_instead_of_alloc,
+        clippy::panic,
+        clippy::print_stdout
+    )]
+    use std::thread::JoinHandle;
+
+    /// Call cancellation function and join thread on drop
+    pub struct ThreadHandler {
+        /// Shutdown function: after calling it thread must terminate in finite amount of time
```
outoftardis, line 89:
```suggestion
        /// Shutdown function: after calling it, the thread must terminate in finite amount of time
```

cli/src/lib.rs
```diff
@@ -83,6 +84,20 @@ where
...
+
+impl<G, S, B> Drop for Iroha<G, S, B>
+where
+    G: GenesisNetworkTrait,
+    S: SumeragiTrait<GenesisNetwork = G>,
+    B: BlockSynchronizerTrait<Sumeragi = S>,
+{
+    fn drop(&mut self) {
+        // Drop thread handles first
+        let _thread_handles = core::mem::take(&mut self.thread_handlers);
```
Arjentix, line 99: You can remove this explicit earlier drop and just place `thread_handlers` at the top of the `Iroha` structure. See order of drops [here](https://doc.rust-lang.org/reference/destructors.html).

But I'dd add a description comment, that noone should place something higher than `thread_handlers`.

Both options with *Please don't modify* comment and `mem::take()` looks like a crutch for me. Not sure, which one is better.

cli/src/lib.rs
```diff
@@ -284,6 +301,7 @@ where
             kura,
             block_sync,
             torii,
+            thread_handlers: vec![kura_thread_handler],
```
Arjentix, line 304: So what do we need `Vec` for? For a future?

core/src/kura.rs
```diff
@@ -220,36 +193,28 @@ impl Kura {
...
-            } else {
-                info!("Kura block thread is shutting down");
-                return;
-            };
+            let mut block_reciever_guard = kura
+                .block_reciever
+                .lock()
+                .expect("able to lock kura recieve block mutex");
 
+            // If kura receive shutdown than close block channel and write remaining blocks to the storage
```
Arjentix, line 207: You mean `then`?

core/src/kura.rs
```diff
@@ -220,36 +193,28 @@ impl Kura {
...
             if shutdown_receiver.try_recv().is_ok() {
                 info!("Kura block thread is being forcibly shut down");
```
Arjentix, line 209: Why is it *fordcibly*? Is there another way to shutdown?

Arjentix: Without panic case, of course

core/src/lib.rs
```diff
@@ -74,6 +74,42 @@ pub trait IsInBlockchain {
     fn is_in_blockchain(&self, wsv: &WorldStateView) -> bool;
 }
 
+/// Module constaining [`Starter`]
```
Arjentix, line 77: No more `Starter` in this module

core/src/lib.rs
```diff
@@ -74,6 +74,42 @@ pub trait IsInBlockchain {
...
+        clippy::std_instead_of_alloc,
+        clippy::panic,
+        clippy::print_stdout
+    )]
+    use std::thread::JoinHandle;
+
+    /// Call cancellation function and join thread on drop
+    pub struct ThreadHandler {
+        /// Shutdown function: after calling it thread must terminate in finite amount of time
+        shutdown: Option<Box<dyn FnOnce() + Send>>,
```
Arjentix, line 90: What about to use [`fn pointers`](https://doc.rust-lang.org/reference/types/function-pointer.html)? They are safe and can be used here for a polymorphic behaviour.

core/test_network/src/lib.rs
```diff
@@ -404,6 +404,9 @@ where
     shutdown: Option<JoinHandle<()>>,
     /// Iroha itself
     pub iroha: Option<Iroha<G, S, B>>,
+    /// Temp dir
+    // Note: last field to be dropped after Iroha (struct fields drops in FIFO RFC 1857)
```
Arjentix, line 408: Oh, here you use the approach with fields drop order

Arjentix:
Is there a way to ensure, that it works?
For example then *"I run test on `iroha2-dev` I see some error/warning in log, but the I run it on `Erigara::iroha_shutdown` branch I see no this error/warning"*?

cli/src/lib.rs
```diff
@@ -83,6 +84,20 @@ where
     pub block_sync: AlwaysAddr<B>,
     /// Torii web server
     pub torii: Option<Torii>,
+    /// Thread handlers
+    thread_handlers: Vec<ThreadHandler>,
```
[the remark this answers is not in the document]
SamHSmith: I think @appetrosyan assumes ThreadHandler to handle multiple threads. I think the name is fine.

cli/src/lib.rs
SamHSmith: You seem to assume *implicit* is somehow inherently better than being *explicit*. If anything I'd argue that the opposite is true.

Code that is *explicit* is **less clever**. Code that is *less clever* is **easier to understand**. Code that is *easier to understand* is more **debuggable** and more **maintainable**.

I think the use of `mem::take` is very tasteful and nothing ought to be changed.

cli/src/lib.rs
SamHSmith: Yes. There will be more threads in the future, such as the `sumeragi_thread` that exists on `iroha2-staging`.

core/src/kura.rs
SamHSmith: There was until very recently so this should probably just say `Kura block thread is being shut down`.

core/src/lib.rs
SamHSmith: Can you motivate why they would be better in this situation? Why is a polymorphic behaviour desired?

core/src/kura.rs
```diff
@@ -220,36 +193,28 @@ impl Kura {
...
-            let strong_kura_reference = if let Some(kura_arc) = weak_kura_reference.upgrade() {
-                kura_arc
-            } else {
-                info!("Kura block thread is shutting down");
-                return;
-            };
+            let mut block_reciever_guard = kura
+                .block_reciever
+                .lock()
+                .expect("able to lock kura recieve block mutex");
```
SamHSmith, line 205: The `block_reciever` lock was acquired every loop iteration as a consequence of the `weak_kura_referenc`. Since you now have an owned reference now you can pull this statement out of the loop.

SamHSmith: It's a bit pedantic of a suggestion but you seem to care about being optimal.

core/src/kura.rs
```diff
@@ -220,36 +193,28 @@ impl Kura {
...
-
             match block_reciever_guard.try_recv() {
-                Err(_) => {
-                    std::thread::sleep(std::time::Duration::from_millis(1));
-                    continue;
-                }
+                // No new blocks would be received
+                Err(_) if is_closed => break,
+                Err(_) => std::thread::sleep(std::time::Duration::from_millis(1)),
                 Ok(new_block) => {
```
SamHSmith, line 218: Very elegant way of draining the channel before exciting. Many users will be have a better experience due to this. Well done.

SamHSmith (approved):
You have my approval. This is a very good solution in my book. Well done.

core/src/lib.rs
Erigara: > What about to use [`fn pointers`](https://doc.rust-lang.org/reference/types/function-pointer.html)? They are safe and can be used here for a polymorphic behaviour.

As we discussed in DM, we want closure to capture objects (e.g. sender part of channel to send shutdown signal) unfortunately fn pointers incapable of this.

core/test_network/src/lib.rs
Erigara: Yes, because we can't use `ManuallyDrop/Option/...` to drop field last one without wrapping other fields too.

Erigara:
> Is there a way to ensure, that it works?
For example then "I run test on iroha2-dev I see some error/warning in log, but the I run it on Erigara::iroha_shutdown branch I see no this error/warning"?

You can run integration tests, it works previously too, purpose of this change is to create more general  mechanism.

cli/src/lib.rs
Erigara: I read through issue regarding stabilization of drop order and it seems like it was created because too many users of language relied on unspecified behavior. 
So i want to be explicit on this one and decide to implement `Drop` manually.

core/src/kura.rs
Erigara: Good point, i overlooked that the thread is the only user of receiver so we can freely take the lock.

SamHSmith (approved):

core/src/lib.rs
Arjentix: > Can you motivate why they would be better in this situation? Why is a polymorphic behaviour desired?

After discussion in DM with @Erigara I'm agree, that using trait object is a good choice here.
However one pros of using fn pointers is to remove needless allocation. But with that approach we can't use objects, which implements `FnOnce` including move closures. And here in `Kura` we use such a closure.

cli/src/lib.rs
Arjentix: @SamHSmith, just my note about *implicit* and *explicit*.

It's not an absolute. We do not choose between *all should be implicit* and *all should be explicit*. There is a balance.

If you want everything to be explicit -- there is the assembly language. But we do not operate with it. It's **too** explicit for our purposes. We operate with some abstractions.

Same is here. There are a lot of abstractions and agreements in **Rust**. One agreement is -- the order of drops in structure is the same as the order of field declarations. This is how strucutres in Rust work. If you are working with Rust you should know it. But with `core::mem::take()` we are breaking this agreement. We are breaking the agreement of the tool we use. The agreement that everyone expects looking at structure declaration. We are not using things that were designed for that, we are doing it manually (which can be errorprone sometimes; less code -- better code).

We are placing `Default` value into the field and then let Rust drop it in the defined order. That's why I think it's a **crutch**. 

This `Drop` implementation is hard to understand without the helpful comment. Without that comment this implementation makes a lot of questions like *Why? For what do we call `take()` here?*

Although I don't insist on my approach. Just wanted to respond to you

Arjentix (approved):

Issue #2678, opened 2022-08-31
Fix tests on staging branch


SamHSmith:

Tohanich:
Description required:
What the tests
What the branch
Steps to reproduce

SamHSmith:
```
branch: iroha2-staging
steps to reproduce: cargo test
```

SamHSmith:
I have a partial fix.

SamHSmith:
Here is the partial fix: https://github.com/hyperledger/iroha/pull/2682
But this issue isn't really closed. The PR only makes it easier to debug because the tests don't abort.

PR #2695, opened 2022-09-05
[fix] #2678: Fix tests on staging branch.

SamHSmith:

### DIFF
--- a/client/tests/integration/triggers/time_trigger.rs
+++ b/client/tests/integration/triggers/time_trigger.rs
@@ -22,9 +22,9 @@ macro_rules! const_assert {
 /// But in general it works well and this test demonstrates it
 #[test]
 #[allow(clippy::cast_precision_loss)]
-fn time_trigger_execution_count_error_should_be_less_than_10_percent() -> Result<()> {
+fn time_trigger_execution_count_error_should_be_less_than_15_percent() -> Result<()> {
     const PERIOD_MS: u64 = 100;
-    const ACCEPTABLE_ERROR_PERCENT: u8 = 10;
+    const ACCEPTABLE_ERROR_PERCENT: u8 = 15;
     const_assert!(PERIOD_MS < DEFAULT_CONSENSUS_ESTIMATION_MS);
     const_assert!(ACCEPTABLE_ERROR_PERCENT <= 100);
 
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -14,9 +14,7 @@ use std::{collections::BTreeSet, error::Error, iter, marker::PhantomData};
 
 use dashmap::{mapref::one::Ref as MapRef, DashMap};
 use eyre::{eyre, Context, Result};
-use iroha_config::sumeragi::{
-    DEFAULT_BLOCK_TIME_MS, DEFAULT_COMMIT_TIME_LIMIT_MS, DEFAULT_TX_RECEIPT_TIME_LIMIT_MS,
-};
+use iroha_config::sumeragi::{DEFAULT_BLOCK_TIME_MS, DEFAULT_COMMIT_TIME_LIMIT_MS};
 use iroha_crypto::{HashOf, KeyPair, MerkleTree, SignatureOf, SignaturesOf};
 use iroha_data_model::{
     block_value::{BlockHeaderValue, BlockValue},
@@ -37,13 +35,12 @@ use crate::{
     tx::{TransactionValidator, VersionedAcceptedTransaction},
 };
 
-const PIPELINE_TIME_MS: u64 =
-    DEFAULT_BLOCK_TIME_MS + DEFAULT_COMMIT_TIME_LIMIT_MS + DEFAULT_TX_RECEIPT_TIME_LIMIT_MS;
+const PIPELINE_TIME_MS: u64 = DEFAULT_BLOCK_TIME_MS + DEFAULT_COMMIT_TIME_LIMIT_MS;
 
 /// Default estimation of consensus duration
 #[allow(clippy::integer_division)]
 pub const DEFAULT_CONSENSUS_ESTIMATION_MS: u64 =
-    (DEFAULT_COMMIT_TIME_LIMIT_MS + PIPELINE_TIME_MS) / 2;
+    DEFAULT_BLOCK_TIME_MS + (DEFAULT_COMMIT_TIME_LIMIT_MS / 2);
 
 /// The chain of the previous block hash. If there is no previous
 /// block - the blockchain is empty.
--- a/core/src/smartcontracts/isi/account.rs
+++ b/core/src/smartcontracts/isi/account.rs
@@ -201,24 +201,24 @@ pub mod isi {
 
             let wsv_clone = wsv.clone();
 
-            wsv_clone.permission_token_definitions()
+            wsv_clone
+                .permission_token_definitions()
                 .get(permission.definition_id())
                 .ok_or_else(|| {
                     FindError::PermissionTokenDefinition(permission.definition_id().clone())
                 })?;
 
-            let temporary_error = wsv.modify_account(&account_id, |account| {
+            wsv.modify_account(&account_id, |account| {
                 let id = account.id();
                 if wsv_clone.account_contains_inherent_permission(id, &permission) {
                     return Err(ValidationError::new("Permission already exists").into());
                 }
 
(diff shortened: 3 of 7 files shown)


client/tests/integration/triggers/time_trigger.rs
```diff
@@ -22,9 +22,9 @@ macro_rules! const_assert {
 /// But in general it works well and this test demonstrates it
 #[test]
 #[allow(clippy::cast_precision_loss)]
-fn time_trigger_execution_count_error_should_be_less_than_10_percent() -> Result<()> {
+fn time_trigger_execution_count_error_should_be_less_than_15_percent() -> Result<()> {
```
Erigara, line 25: Why do we need to relax requirement here?

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -174,18 +172,16 @@ pub mod isi {
...
                     } else {
                         Err(ValidationError::new("Unauthorized trigger execution").into())
                     }
                 })
-                .ok_or_else(|| Error::Find(Box::new(FindError::Trigger(id))))?;
-            let new_array = acc.lock().unwrap().clone();
-            for (id, authority) in new_array {
+                .ok_or_else(|| Error::Find(Box::new(FindError::Trigger(id.clone()))))?;
+            if temporary_error.is_ok() {
                 wsv.execute_trigger(id, authority);
```
Erigara, line 182: Maybe instead of `if` chain with `.and_then` this way you don't need bind variable at all. 

```suggestion
                .and_then(|_| wsv.execute_trigger(id, authority))
```

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -174,18 +172,16 @@ pub mod isi {
...
                     if allow_execute {
-                        acc.lock().unwrap().push((id.clone(), authority.clone()));
                         Ok(())
                     } else {
                         Err(ValidationError::new("Unauthorized trigger execution").into())
                     }
                 })
-                .ok_or_else(|| Error::Find(Box::new(FindError::Trigger(id))))?;
-            let new_array = acc.lock().unwrap().clone();
-            for (id, authority) in new_array {
```
Erigara, line 185: This function had kinda strange implementaion previously, it seems like `execute_trigger` could be called more then once, however `acc` is empty or contain only one item.

client/tests/integration/triggers/time_trigger.rs
SamHSmith: Because I didn't manage to get it within 10 % without eyeballing other constants for my machine. New-Sumeragi's block production time = a constant + cpu bound processing time + network latency. My understanding is that @Arjentix previously picked a block time estimate constant that had an error rate within 10 %.

Essentially, the "< 10 %" is not a guarantee made to the end customer/user. So I changed it to 15 % so that it would pass, and the end customer/user can get a faster consensus.

core/src/smartcontracts/isi/triggers.rs
SamHSmith: Yes I should change it to an `Option`. This design should change long term though.

SamHSmith: > This function had kinda strange implementaion previously, it seems like `execute_trigger` could be called more then once, however `acc` is empty or contain only one item.

It's what I did when rebasing. I didn't know if the closure got called more then once. It's one of the many reasons I don't like closures.

core/src/smartcontracts/isi/triggers.rs
SamHSmith: good suggestion

core/src/smartcontracts/isi/triggers.rs
Erigara: > closure got called more then once
@SamHSmith, i think `rust` do us good service here, it provides `FnOnce` closure type which might be called only once, however in this paticular case  inside `inspect_by_id` `Fn` is used which is misleading.

Erigara (changes requested):
Run formatter and i will approve.

Erigara (approved):
LGTM

client/tests/integration/triggers/time_trigger.rs
Arjentix: Yeah, 10% was something that worked for me and for CI.
Although it's sad, that we need to relax the gap. Does Sumeragi works slower now?

core/src/smartcontracts/isi/account.rs
```diff
@@ -201,24 +201,25 @@ pub mod isi {
...
                 Ok(AccountEvent::PermissionAdded(id.clone()))
-            });
-            if temporary_error.is_ok() {
+            })
+            .and_then(|_| {
                 wsv.add_account_permission(&account_id, permission);
-            }
-            temporary_error
+                Ok(())
+            })
```
Arjentix, line 222: I think [`map`](https://doc.rust-lang.org/core/result/enum.Result.html#method.map) is what you was looking for

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -174,18 +172,16 @@ pub mod isi {
...
-                wsv.execute_trigger(id, authority);
-            }
-            Ok(())
+                .ok_or_else(|| Error::Find(Box::new(FindError::Trigger(id.clone()))))?
+                .and_then(|_| {
+                    wsv.execute_trigger(id, authority);
+                    Ok(())
+                })
```
Arjentix, line 184: Same suggestion with `map`

client/tests/integration/triggers/time_trigger.rs
SamHSmith: No. It's actually a lot faster. But I think the policy is slightly different. In the new Sumeragi, the leader waits for *block time* milliseconds before starting to create a block if the transaction cache isn't full. Then the whole networks policy is to commit the block as fast as possible. This means that when unloaded the networks total block time = [BLOCK_TIME_MS] + [leader: block validation runtime] + [worst latency to validating peer] + [validating peer: block validation runtime] + [worst latency to proxy tail] + [worst latency to nodes] + [wsv apply runtime]. This is supposed to be less than [BLOCK_TIME_MS] + [COMMIT_TIME_MS] otherwise the network is considered faulty.

And if the network is loaded, then we produce blocks as fast as possible, which means that the constant is inevitably wrong.

There is a fix however. We could in the future add code that calculates an estimated block time based on previous block times right? A rolling average of the previous X block times for instance.

core/src/block.rs
```diff
@@ -37,13 +35,12 @@ use crate::{
...
 
-const PIPELINE_TIME_MS: u64 =
-    DEFAULT_BLOCK_TIME_MS + DEFAULT_COMMIT_TIME_LIMIT_MS + DEFAULT_TX_RECEIPT_TIME_LIMIT_MS;
+const PIPELINE_TIME_MS: u64 = DEFAULT_BLOCK_TIME_MS + DEFAULT_COMMIT_TIME_LIMIT_MS;
 
 /// Default estimation of consensus duration
 #[allow(clippy::integer_division)]
 pub const DEFAULT_CONSENSUS_ESTIMATION_MS: u64 =
-    (DEFAULT_COMMIT_TIME_LIMIT_MS + PIPELINE_TIME_MS) / 2;
+    DEFAULT_BLOCK_TIME_MS + (DEFAULT_COMMIT_TIME_LIMIT_MS / 2);
```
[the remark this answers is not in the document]
SamHSmith: I tried many different values for this constant and was unable to decrease the error. You could probably bring it down since you understand time triggers better. When sumeragi is unloaded, the shortest possible block time is simply the block time. The longest allowed block time is *block time* + *commit time*. The average of these two is *block time* + *commit time* / *2*.

core/src/smartcontracts/isi/account.rs
SamHSmith: good suggestion

core/src/block.rs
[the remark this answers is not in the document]
Arjentix: Ok, got it

client/tests/integration/triggers/time_trigger.rs
Arjentix: Yeah, @appetrosyan has also told about analysing previous blocks to predict the next block timestamp. That's our plan for a future

Arjentix (approved):

Erigara (approved):

PR #2701, opened 2022-09-06
[fix] #2700: Export `KURA_BLOCK_STORE_PATH` in docker images

### LINKED ISSUE #2700: Inconsistent block storage configuration in `Dockerfile`
### GIT commit hash

908f09f0

### Minimum working example

```rust
fn main() {
}
```

### Expected behaviour

The block storage is available from docker containers without any special configuration

### Actual behaviour

Since `Dockerfile` has a hardcoded `./storage`, `./blocks` as `DEFAULT_BLOCK_STORE_PATH` needs to be overwritten by `KURA_BLOCK_STORE_PATH` in `config.json` or environment variables, which already confused DevOps. This can be considered as a bug

### Operating system

x86_64 GNU/Linux

### LibC type and version

_No response_

### Current environment

Docker Hub

### Logs in JSON format

<details>
  
  ```json
  ```
</details>

### Who can help?

@s8sato

s8sato:
### Description of the Change

1. Export `KURA_BLOCK_STORE_PATH` in docker images
	- Additionally, make it configurable as a build arg
2. Rename `DEFAULT_BLOCK_STORE_PATH` from `./blocks` to `./storage`

### Issue

1. Closes #2700

### Benefits

1. The block storage is available from docker containers without any special configuration
2. Have a default storage name regardless of deployment methods

### Possible Drawbacks

2. The rename can affect documentation. Could you check @outoftardis


appetrosyan (approved):

outoftardis:
I found one place in the tutorial where we would need to make the update.

Erigara (approved):

PR #2704, opened 2022-09-06
[refactor] #2478: Add `SignedTransaction`

### LINKED ISSUE #2478: [refactor] Add state machine for signed `Transaction`
### Feature request

Add strong-typed transition from `Transaction` to `SignedTransaction` when it was signed

### Motivation

Introducing a state machine pattern for `Transaction` when it was signed would help to understand some code using it and will decrease the amount of runtime errors when working with `Transaction` type

### Who can help?

@arjentix

Erigara:

### Description of the Change

- Renamed `Transaction/VersionedTransaction` to `SignedTransaction/VersionedTransaction`;
- Add `Transaction`: which represent transaction which do not receive sign yet; 
- Add `SignableTransaction` trait with method `sign` to create generic function accepting ether `Transaction` or `SignedTransaction` (multisig).

**For reviewers**: most important changes in `data_model/src/transaction.rs, client/src/client.rs`.

### Issue

Closes #2478.

### Benefits

Allow to identify buggy behavior on type level.  

### Possible Drawbacks

None.

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -323,7 +323,7 @@ impl Client {
         &self,
         instructions: Executable,
         metadata: UnlimitedMetadata,
-    ) -> Result<Transaction> {
+    ) -> Result<SignedTransaction> {
         let transaction = Transaction::new(
             self.account_id.clone(),
             instructions,
@@ -345,7 +345,7 @@ impl Client {
     ///
     /// # Errors
     /// Fails if generating signature fails
-    pub fn sign_transaction(&self, transaction: Transaction) -> Result<Transaction> {
+    pub fn sign_transaction<Tx: Sign>(&self, transaction: Tx) -> Result<SignedTransaction> {
         transaction
             .sign(self.key_pair.clone())
             .wrap_err("Failed to sign transaction")
@@ -369,7 +369,7 @@ impl Client {
     pub fn submit(
         &self,
         instruction: impl Into<Instruction> + Debug,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         let isi = instruction.into();
         self.submit_all([isi])
     }
@@ -382,7 +382,7 @@ impl Client {
     pub fn submit_all(
         &self,
         instructions: impl IntoIterator<Item = Instruction>,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         self.submit_all_with_metadata(instructions, UnlimitedMetadata::new())
     }
 
@@ -396,7 +396,7 @@ impl Client {
         &self,
         instruction: Instruction,
         metadata: UnlimitedMetadata,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         self.submit_all_with_metadata([instruction], metadata)
     }
 
@@ -410,7 +410,7 @@ impl Client {
         &self,
         instructions: impl IntoIterator<Item = Instruction>,
         metadata: UnlimitedMetadata,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         self.submit_transaction(self.build_transaction(instructions.into(), metadata)?)
     }
 
@@ -421,8 +421,8 @@ impl Client {
     /// Fails if sending transaction to peer fails or if it response with error
     pub fn submit_transaction(
         &self,
-        transaction: Transaction,
-    ) -> Result<HashOf<VersionedTransaction>> {
+        transaction: SignedTransaction,
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         iroha_logger::trace!(tx=?transaction);
         let (req, hash, resp_handler) =
             self.prepare_transaction_request::<DefaultRequestBuilder>(transaction)?;
@@ -441,8 +441,8 @@ impl Client {
     /// Fails if sending a transaction to a peer fails or there is an error in the response
     pub fn submit_transaction_blocking(
         &self,
-        transaction: Transaction,
-    ) -> Result<HashOf<VersionedTransaction>> {
+        transaction: SignedTransaction,
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         let client = self.clone();
         let (init_sender, init_receiver) = mpsc::channel();
         let hash = transaction.hash();
@@ -476,9 +476,9 @@ impl Client {
     // TODO: introduce timeout
     fn listen_for_tx_confirmation(
         &self,
-        hash: HashOf<Transaction>,
+        hash: HashOf<SignedTransaction>,
         init_sender: &mpsc::Sender<()>,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         let event_iterator = self
             .listen_for_events(PipelineEventFilter::new().hash(hash.into()).into())
             .wrap_err("Failed to establish event listener connection")?;
@@ -515,10 +515,14 @@ impl Client {
     /// Fails if transaction check fails
     pub fn prepare_transaction_request<B: RequestBuilder>(
         &self,
-        transaction: Transaction,
-    ) -> Result<(B, HashOf<VersionedTransaction>, TransactionResponseHandler)> {
+        transaction: SignedTransaction,
+    ) -> Result<(
+        B,
+        HashOf<VersionedSignedTransaction>,
+        TransactionResponseHandler,
+    )> {
         transaction.check_limits(&self.transaction_limits)?;
-        let transaction: VersionedTransaction = transaction.into();
+        let transaction: VersionedSignedTransaction = transaction.into();
         let hash = transaction.hash();
         let transaction_bytes: Vec<u8> = transaction.encode_versioned();
 
@@ -542,7 +546,7 @@ impl Client {
     pub fn submit_blocking(
         &self,
         instruction: impl Into<Instruction>,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         self.submit_all_blocking(vec![instruction.into()])
     }
 
@@ -554,7 +558,7 @@ impl Client {
     pub fn submit_all_blocking(
         &self,
         instructions: impl IntoIterator<Item = Instruction>,
-    ) -> Result<HashOf<VersionedTransaction>> {
+    ) -> Result<HashOf<VersionedSignedTransaction>> {
         self.submit_all_blocking_with_metadata(instructions, UnlimitedMetadata::new())
     }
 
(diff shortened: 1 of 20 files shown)


data_model/src/transaction.rs
```diff
@@ -287,12 +287,20 @@ impl From<VersionedValidTransaction> for VersionedTransaction {
...
+    /// # Errors
+    ///
+    /// Fails if signature creation fails
+    fn sign(
+        self,
+        key_pair: iroha_crypto::KeyPair,
+    ) -> Result<SignedTransaction, iroha_crypto::Error>;
+}
+
+/// Structure represents transaction in non-trusted form which not received any signatures yet.
```
Arjentix, line 303: May be we don't need `signatures` field in this struct anymore?

data_model/src/transaction.rs
```diff
@@ -342,34 +350,53 @@ impl Transaction {
...
-    ) -> Result<Transaction, iroha_crypto::Error> {
-        let signature = SignatureOf::new(key_pair, &self.payload)?;
-        self.signatures.insert(signature);
-
-        Ok(Self {
-            payload: self.payload,
-            signatures: self.signatures,
-        })
-    }
+/// Structure represents transaction in non-trusted form.
```
Arjentix, line 355: I think, this form is trusted

data_model/src/transaction.rs
```diff
@@ -342,34 +350,53 @@ impl Transaction {
...
+///
+/// `Iroha` and its' clients use [`SignedTransaction`] to send transactions
```
Arjentix, line 357:
```suggestion
/// `Iroha` and its clients use [`SignedTransaction`] to send transactions
```

data_model/src/transaction.rs
```diff
@@ -342,34 +350,53 @@ impl Transaction {
...
+/// via network.  Direct usage in business logic is strongly
+/// prohibited. Before any interactions `accept`.
```
Arjentix, line 359: I don't get it with `accept`. What do you mean?

data_model/src/transaction.rs
```diff
@@ -342,34 +350,53 @@ impl Transaction {
...
-        &self.payload
-    }
+            #[inline]
+            fn payload(&self) -> &Payload {
+                &self.payload
+            }
+        }
+
+        #[cfg(feature = "std")]
+        impl SignableTransaction for $ty {
```
Arjentix, line 381: What do you think about renaming to just `Sign`? It's more convinient in Rust.

We also know, that it is related to `Transaction` because it's inside `transaction` module

data_model/src/transaction.rs
```diff
@@ -342,34 +350,53 @@ impl Transaction {
...
+                mut self,
+                key_pair: iroha_crypto::KeyPair,
+            ) -> Result<SignedTransaction, iroha_crypto::Error> {
+                let signature = SignatureOf::new(key_pair, &self.payload)?;
+                self.signatures.insert(signature);
+
+                Ok(SignedTransaction {
+                    payload: self.payload,
+                    signatures: self.signatures,
+                })
```
Arjentix, line 392: I think, you can use new signature collection instead of using the one from `self`. Then this will not confuse peaple, then they read that *Transaction without signatures*, but struct actually stores `signatures`

Arjentix: Oh, I see, you need to store `signatures` in `Transaction` to be able to use this macro with `Transaction` and `SignedTransaction` at the same time....

Arjentix: Well, in my opinion it's better to write this impl by hand and not to store meaningless `signatures` in `Transaction`.

If you are not agree, we can wait for other opinions

Arjentix: Or you can upgrade your macro to accept default `signatures` state.
Some examples as I see it:

```rust
impl_tx! {
    Transaction<Singatures=BTreeSet::new()>,
    SignedTransaction<Signatures=self.signatures>
}

impl_tx! {
    Transaction where Singatures: BTreeSet::new(),
    SignedTransaction where Signatures: self.signatures
}

impl_tx! {
    Transaction with BTreeSet::new() as signatures,
    SignedTransaction with self.signatures as signatures
}
```

client/src/client.rs
```diff
@@ -345,7 +345,10 @@ impl Client {
     ///
     /// # Errors
     /// Fails if generating signature fails
-    pub fn sign_transaction(&self, transaction: Transaction) -> Result<Transaction> {
```
Arjentix, line 348: This is the place, where I decided to create this issue. Finally it's fixed, thanks!

tools/parity_scale_decoder/src/generate_map.rs
```diff
@@ -407,14 +407,22 @@ mod tests {
...
-        let exceptions = HashSet::from([
-            "Vec<iroha_core::genesis::GenesisTransaction>",
-            "iroha_core::genesis::GenesisTransaction",
-            "iroha_core::genesis::RawGenesisBlock",
-            "iroha_crypto::merkle::MerkleTree<iroha_data_model::transaction::VersionedTransaction>",
+        let exceptions = HashSet::from(type_names_arr![
+            Vec<iroha_core::genesis::GenesisTransaction>,
+            iroha_core::genesis::GenesisTransaction,
+            iroha_core::genesis::RawGenesisBlock,
+            iroha_crypto::MerkleTree<iroha_data_model::transaction::VersionedSignedTransaction>,
```
Arjentix, line 425: Good

data_model/src/transaction.rs
Erigara: It's not my comment, but i suppose that `iroha` don't trust `SignedTransaction` before verifying signature and convert transaction into `AcceptedTrasaction`.

outoftardis: I think this comment should be rephrased

data_model/src/transaction.rs
outoftardis:
```suggestion
/// Structure representing transactions in a non-trusted form.
```
or
```suggestion
/// Structure that represents transactions in a non-trusted form.
```
(or `in a trusted form` if Daniil is right)

data_model/src/transaction.rs
outoftardis, line 303:
```suggestion
/// Structure that represents a transaction in a non-trusted form before the transaction receives any signatures.
```

data_model/src/transaction.rs
```diff
@@ -287,12 +287,20 @@ impl From<VersionedValidTransaction> for VersionedTransaction {
     }
 }
 
-/// This structure represents transaction in non-trusted form.
-///
-/// `Iroha` and its' clients use [`Transaction`] to send transactions
-/// via network.  Direct usage in business logic is strongly
-/// prohibited. Before any interactions `accept`.
-#[version(n = 1, versioned = "VersionedTransaction")]
+/// Trait for signing transaction
```
outoftardis, line 290:
```suggestion
/// Trait for signing transactions
```

data_model/src/transaction.rs
Erigara: i think manual implementation would be enough in this case.

data_model/src/transaction.rs
```diff
@@ -342,28 +347,64 @@ impl Transaction {
...
+    fn payload(&self) -> &Payload {
+        &self.payload
+    }
+}
+
+/// Structure that represents the second state of the transaction after receiving at least one signature.
+///
+/// `Iroha` and its clients use [`SignedTransaction`] to send transactions over the network.
+/// Before further processing,
+/// the transaction must be accepted by the `Iroha` peer: verify the signatures and check the limits.
```
outoftardis, line 381:
```suggestion
/// After a transaction is signed and before it can be processed any further,
/// the transaction must be accepted by the `Iroha` peer. The peer verifies the signatures and checks the limits.
```

Arjentix (dismissed):

tools/parity_scale_decoder/src/generate_map.rs
```diff
@@ -407,14 +407,22 @@ mod tests {
...
+        ($($ty:ty),+$(,)?) => {
+            [$(
+                <$ty as IntoSchema>::type_name(),
+            )+]
+        }
+    }
+
     #[test]
     fn schemas_types_is_a_subset_of_map_types() {
         // Exceptions which does not implement `Decode` so that they can't be decoded by this tool
```
appetrosyan, line 420:
```suggestion
        // These types **shouldn't** implement `Decode`. As such we need to make an exception. 
```

tools/parity_scale_decoder/src/generate_map.rs
```diff
@@ -407,14 +407,22 @@ mod tests {
 
     use super::*;
 
+    macro_rules! type_names_arr {
+        ($($ty:ty),+$(,)?) => {
+            [$(
+                <$ty as IntoSchema>::type_name(),
+            )+]
+        }
+    }
```
appetrosyan, line 416: Good

data_model/src/transaction.rs
```diff
@@ -287,22 +287,28 @@ impl From<VersionedValidTransaction> for VersionedTransaction {
...
+/// Trait for signing transactions
+pub trait Sign {
+    /// Sign transaction with provided key pair.
+    ///
+    /// # Errors
+    ///
+    /// Fails if signature creation fails
+    fn sign(
+        self,
+        key_pair: iroha_crypto::KeyPair,
```
appetrosyan, line 299: Just an architectural question. Do we **need** to provide the public key inside the key pair. Knowing the algorithm we can derive it.

Erigara: Yes, we could implement `From<PrivateKey> for PublicKey` and recover `PublicKey` on demand.
In such case we don't need `KeyPair` at all.

```rust
impl From<PrivateKey> for PublicKey {
    #[allow(clippy::expect_used)]
    fn from(private_key: PrivateKey) -> Self {
        let digest_function = private_key.digest_function();
        let key_gen_option = Some(UrsaKeyGenOption::FromSecretKey(UrsaPrivateKey(
            private_key.payload,
        )));
        let (mut public_key, _) = match digest_function {
            Algorithm::Ed25519 => Ed25519Sha512.keypair(key_gen_option),
            Algorithm::Secp256k1 => EcdsaSecp256k1Sha256::new().keypair(key_gen_option),
            Algorithm::BlsNormal => BlsNormal::new().keypair(key_gen_option),
            Algorithm::BlsSmall => BlsSmall::new().keypair(key_gen_option),
        }
        .expect("can't fail for valid `PrivateKey`");
        PublicKey {
            digest_function: private_key.digest_function,
            payload: core::mem::take(&mut public_key.0),
        }
    }
}
```

outoftardis (approved):
doc comments are ok

data_model/src/transaction.rs
Arjentix: User can provide invalid private key and I don't think we need to panic in that case. Better to implement `TryFrom`, I believe

Arjentix (dismissed):

data_model/src/transaction.rs
Erigara: `TryFrom` is good option, i just think that it may be redundant to return error in case if invariants for private keys are checked before.

appetrosyan (approved):

data_model/src/transaction.rs
Arjentix: Hm, if this invariant is guaranteed, then I agree

Arjentix (approved):

Issue #2614, opened 2022-08-12
Make entrypoint use `Proxy` config


ilchu:
When all implementation is nearly done, replace the config entrypoint in `cli/src/lib.rs` with proxy if it's needed, and go through all the initialization / building steps.

Issue #2589, opened 2022-08-03
Kagami docgen without defaults


ilchu:
See how `kagami` would generate docs without `Default` impls on configs. Probably find a smarter way to diversify which fields could have a `Default` value, and which would never have one, and specify that automatically during docgen.

PR #2706, opened 2022-09-06
[feature] #2627: Config finalisation

### LINKED ISSUE #2627: Finalize `Config` if needed
See [comment](https://github.com/hyperledger/iroha/pull/2583#discussion_r935760671).

ilchu:

### Description of the Change
* Only `..Proxy` structs have `Default` impls
* `Documented` trait works on `..Proxy` as well for the sake of `kagami` gen
* `finalize()` methods reworked to be more comprehensive
* `placeholder..()` methods ditched

### Issue
Closes #2627, #2589, #2614.

### Benefits
Almost the final steps in the config epic, tests notwithstanding.

### Possible Drawbacks
Still work needs to be done with the test network, as some old tests keep failing

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -13,7 +13,10 @@ use std::{panic, path::PathBuf, sync::Arc};
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
 use iroha_actor::{broker::*, prelude::*};
-use iroha_config::iroha::Configuration;
+use iroha_config::{
+    base::proxy::{LoadFromDisk, LoadFromEnv},
+    iroha::{Configuration, ConfigurationProxy},
+};
 use iroha_core::{
     block_sync::{BlockSynchronizer, BlockSynchronizerTrait},
     genesis::{GenesisNetwork, GenesisNetworkTrait, RawGenesisBlock},
@@ -124,11 +127,9 @@ where
         query_judge: QueryJudgeBoxed,
     ) -> Result<Self> {
         let broker = Broker::new();
-        let mut config = match Configuration::from_path(&args.config_path) {
-            Ok(config) => config,
-            Err(_) => Configuration::default(),
-        };
-        config.load_environment()?;
+        let mut proxy = ConfigurationProxy::from_path(&args.config_path)?;
+        proxy.load_environment()?;
+        let config = proxy.build()?;
 
         let telemetry = iroha_logger::init(&config.logger)?;
         iroha_logger::info!("Hyperledgerいろは2にようこそ！");
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1362,7 +1362,11 @@ mod tests {
     #![allow(clippy::restriction)]
     use std::str::FromStr;
 
-    use iroha_config::client::{BasicAuth, WebLogin};
+    use iroha_config::{
+        base::proxy::Builder,
+        client::{BasicAuth, ConfigurationProxy, WebLogin},
+        torii::{uri::DEFAULT_API_URL, DEFAULT_TORII_TELEMETRY_URL},
+    };
 
     use super::*;
 
@@ -1375,12 +1379,21 @@ mod tests {
     fn txs_same_except_for_nonce_have_different_hashes() {
         let (public_key, private_key) = KeyPair::generate().unwrap().into();
 
-        let cfg = Configuration {
-            public_key,
-            private_key,
-            add_transaction_nonce: true,
-            ..Configuration::default()
-        };
+        let cfg = ConfigurationProxy {
+            public_key: Some(public_key),
+            private_key: Some(private_key),
+            account_id: Some(
+                "alice@wonderland"
+                    .parse()
+                    .expect("This account ID should be valid"),
+            ),
+            torii_api_url: Some(SmallStr::from_str(DEFAULT_API_URL)),
+            torii_telemetry_url: Some(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL)),
+            add_transaction_nonce: Some(true),
+            ..ConfigurationProxy::default()
+        }
+        .build()
+        .expect("Client config should build as all required fields were provided");
         let client = Client::new(&cfg).expect("Invalid client configuration");
 
         let build_transaction = || {
@@ -1404,10 +1417,28 @@ mod tests {
             password: SmallStr::from_str(PASSWORD),
         };
 
-        let cfg = Configuration {
-            basic_auth: Some(basic_auth),
-            ..Configuration::default()
-        };
+        let cfg = ConfigurationProxy {
+            public_key: Some(
+                "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
+                    .parse()
+                    .expect("Public key not in mulithash format"),
+            ),
+            private_key: Some(iroha_crypto::PrivateKey::from_hex(
+            iroha_crypto::Algorithm::Ed25519,
+            "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
+            ).expect("Private key not hex encoded")),
+            account_id: Some(
+                "alice@wonderland"
+                    .parse()
+                    .expect("This account ID should be valid"),
+            ),
+            torii_api_url: Some(SmallStr::from_str(DEFAULT_API_URL)),
+            torii_telemetry_url: Some(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL)),
+            basic_auth: Some(Some(basic_auth)),
+            ..ConfigurationProxy::default()
+        }
+        .build()
+        .expect("Client config should build as all required fields were provided");
         let client = Client::new(&cfg).expect("Invalid client configuration");
 
         let value = client
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -1,7 +1,5 @@
 //! Crate contains iroha client which talks to iroha network via http
 
-pub use iroha_config::client::Configuration;
-
 /// Module with iroha client itself
 pub mod client;
 /// Module with general communication primitives like an HTTP request builder.
@@ -10,21 +8,31 @@ mod http_default;
 
 /// Module containing sample configurations for tests and benchmarks.
 pub mod samples {
-    use iroha_config::torii::uri;
+    use iroha_config::{
+        base::proxy::Builder,
+        client::{Configuration, ConfigurationProxy},
+        torii::{uri::DEFAULT_API_URL, DEFAULT_TORII_TELEMETRY_URL},
+    };
     use iroha_crypto::KeyPair;
-
-    use super::Configuration;
+    use iroha_primitives::small::SmallStr;
 
     /// Get sample client configuration.
     #[allow(clippy::expect_used)]
     pub fn get_client_config(key_pair: &KeyPair) -> Configuration {
         let (public_key, private_key) = key_pair.clone().into();
-        Configuration {
-            public_key,
-            private_key,
-            account_id: "alice@wonderland".parse().expect("Should not fail."),
-            torii_api_url: iroha_primitives::small::SmallStr::from_str(uri::DEFAULT_API_URL),
-            ..Configuration::default()
+        ConfigurationProxy {
+            public_key: Some(public_key),
+            private_key: Some(private_key),
+            account_id: Some(
+                "alice@wonderland"
+                    .parse()
+                    .expect("This account ID should be valid"),
+            ),
+            torii_api_url: Some(SmallStr::from_str(DEFAULT_API_URL)),
+            torii_telemetry_url: Some(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL)),
+            ..ConfigurationProxy::default()
         }
+        .build()
+        .expect("Client config should build as all required fields were provided")
     }
 }
--- a/client/tests/integration/config.rs
+++ b/client/tests/integration/config.rs
@@ -2,7 +2,7 @@
 
 use test_network::*;
 
-use super::Configuration;
+use super::{Builder, Configuration, ConfigurationProxy};
 
 #[test]
 fn get_config() {
@@ -13,10 +13,16 @@ fn get_config() {
     let field = test_client.get_config_docs(&["torii"]).unwrap().unwrap();
     assert!(field.contains("IROHA_TORII"));
 
-    let cfg: Configuration =
-        serde_json::from_value(test_client.get_config_value().unwrap()).unwrap();
     let test = Configuration::test();
-    assert_eq!(cfg.block_sync, test.block_sync);
-    assert_eq!(cfg.network, test.network);
-    assert_eq!(cfg.telemetry, test.telemetry);
+    let cfg_proxy: ConfigurationProxy =
+        serde_json::from_value(test_client.get_config_value().unwrap()).unwrap();
+    assert_eq!(
+        cfg_proxy.block_sync.unwrap().build().unwrap(),
+        test.block_sync
+    );
+    assert_eq!(cfg_proxy.network.unwrap().build().unwrap(), test.network);
+    assert_eq!(
+        cfg_proxy.telemetry.unwrap().build().unwrap(),
+        test.telemetry
+    );
 }
--- a/config/base/derive/src/documented.rs
+++ b/config/base/derive/src/documented.rs
@@ -1,9 +1,9 @@
 use proc_macro::TokenStream;
 use proc_macro2::Span;
 use quote::quote;
-use syn::{Lit, LitStr, Meta};
+use syn::{parse_quote, Lit, LitStr, Meta, Path};
 
-use super::utils::StructWithFields;
+use super::utils::{get_inner_type, StructWithFields};
 
 pub fn impl_documented(ast: &StructWithFields) -> TokenStream {
     let name = &ast.ident;
@@ -46,18 +46,29 @@ fn impl_get_doc_recursive(docs: Vec<LitStr>, ast: &StructWithFields) -> proc_mac
         .iter()
         .zip(docs)
         .map(|(field, documentation)| {
-            let inner_thing = field.has_inner;
             let ty = &field.ty;
             let ident = &field.ident;
-            if inner_thing {
+            let documented_trait: Path = parse_quote! { iroha_config_base::proxy::Documented };
+            if field.has_inner && field.has_option {
+                let inner_ty = get_inner_type("Option", &field.ty);
                 quote! {
                     [stringify!(#ident)] => {
                         let curr_doc = #documentation;
-                        let inner_docs = <#ty as ::iroha_config_base::proxy::Documented>::get_inner_docs();
+                        let inner_docs = <#inner_ty as #documented_trait>::get_inner_docs();
                         let total_docs = format!("{}\n\nHas following fields:\n\n{}\n", curr_doc, inner_docs);
                         Some(total_docs)
                     },
-                    [stringify!(#ident), rest @ ..] => <#ty as ::iroha_config_base::proxy::Documented>::get_doc_recursive(rest)?,
+                    [stringify!(#ident), rest @ ..] => <#inner_ty as #documented_trait>::get_doc_recursive(rest)?,
+                }
+            } else if field.has_inner {
+                quote! {
+                    [stringify!(#ident)] => {
+                        let curr_doc = #documentation;
+                        let inner_docs = <#ty as #documented_trait>::get_inner_docs();
+                        let total_docs = format!("{}\n\nHas following fields:\n\n{}\n", curr_doc, inner_docs);
+                        Some(total_docs)
+                    },
+                    [stringify!(#ident), rest @ ..] => <#ty as #documented_trait>::get_doc_recursive(rest)?,
                 }
             } else {
                 quote! { [stringify!(#ident)] => Some(#documentation.to_owned()), }
@@ -83,11 +94,16 @@ fn impl_get_doc_recursive(docs: Vec<LitStr>, ast: &StructWithFields) -> proc_mac
 
 fn impl_get_inner_docs(docs: Vec<LitStr>, ast: &StructWithFields) -> proc_macro2::TokenStream {
     let inserts = ast.fields.iter().zip(docs).map(|(field, documentation)| {
-        let inner_thing = field.has_inner;
         let ty = &field.ty;
         let ident = &field.ident;
-        let doc = if inner_thing {
-            quote! { <#ty as ::iroha_config_base::proxy::Documented>::get_inner_docs().as_str() }
+        let documented_trait: Path = parse_quote! { ::iroha_config_base::proxy::Documented };
+        let doc = if field.has_inner && field.has_option {
+            let inner_ty = get_inner_type("Option", &field.ty);
+            quote! {
+                <#inner_ty as #documented_trait>::get_inner_docs().as_str()
+            }
+        } else if field.has_inner {
+            quote! { <#ty as #documented_trait>::get_inner_docs().as_str() }
         } else {
             quote! { #documentation.into() }
         };
@@ -113,9 +129,12 @@ fn impl_get_docs(docs: Vec<LitStr>, ast: &StructWithFields) -> proc_macro2::Toke
     let inserts = ast.fields.iter().zip(docs).map(|(field, documentation)| {
         let ident = &field.ident;
         let ty = &field.ty;
-        let inner_thing = field.has_inner;
-        let doc = if inner_thing {
-            quote! { <#ty as ::iroha_config_base::proxy::Documented>::get_docs().into() }
+        let documented_trait: Path = parse_quote! { iroha_config_base::proxy::Documented };
+        let doc = if field.has_inner && field.has_option {
+            let inner_ty = get_inner_type("Option", &field.ty);
+            quote! { <#inner_ty as #documented_trait>::get_docs().into() }
+        } else if field.has_inner {
+            quote! { <#ty as #documented_trait>::get_docs().into() }
         } else {
             quote! { #documentation.into() }
         };
@@ -152,10 +171,17 @@ fn impl_get_recursive(ast: &StructWithFields) -> proc_macro2::TokenStream {
     let variants = ast.fields
         .iter()
         .map(|field | {
-            let inner_thing = field.has_inner;
             let ident = &field.ident;
             let l_value = &field.lvalue_read;
-            let inner_thing2 = if inner_thing {
+            let inner_thing2 = if field.has_inner && field.has_option {
+                let inner_ty = get_inner_type("Option", &field.ty);
+                let documented_trait: Path = parse_quote! { iroha_config_base::proxy::Documented };
+                quote! {
+                    [stringify!(#ident), rest @ ..] => {
+                        <#inner_ty as #documented_trait>::get_recursive(#l_value.as_ref().expect("Should be instantiated"), rest)?
+                    },
+                }
+            } else if field.has_inner {
                 quote! {
                     [stringify!(#ident), rest @ ..] => {
                         #l_value.get_recursive(rest)?
@@ -216,7 +242,7 @@ pub fn gen_docs(ast: &StructWithFields) -> Vec<LitStr> {
                 });
             let real_doc = real_doc.map(|doc| doc.value() + "\n\n").unwrap_or_default();
             let docs = format!(
-                "{}Has type `{}`. Can be configured via environment variable `{}`",
+                "{}Has type `{}`. Can be configured via environment variable `{}`. Refer to [configuration types](#configuration-types) for details.",
                 real_doc,
                 quote! { #field_ty }.to_string().replace(' ', ""),
                 env
--- a/config/base/derive/src/proxy.rs
+++ b/config/base/derive/src/proxy.rs
@@ -7,23 +7,27 @@ use crate::utils;
 
 pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
     // somewhat awkward conversion, could it be better?
-    let parent_name = ast.ident.clone();
+    let parent_name = &ast.ident;
     let parent_ty: Type = parse_quote! { #parent_name };
     let proxy_struct = gen_proxy_struct(ast);
     let loadenv_derive = quote! { ::iroha_config_base::derive::LoadFromEnv };
     let disk_derive = quote! { ::iroha_config_base::derive::LoadFromDisk };
     let builder_derive = quote! { ::iroha_config_base::derive::Builder };
     let combine_derive = quote! { ::iroha_config_base::derive::Combine };
+    let documented_derive = quote! { ::iroha_config_base::derive::Documented };
     quote! {
         /// Proxy configuration structure to be used as an intermediate
         /// for config loading
-        #[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize,
+        #[derive(Debug, Clone, serde::Serialize, serde::Deserialize,
                  #builder_derive,
                  #loadenv_derive,
                  #disk_derive,
-                 #combine_derive)]
+                 #combine_derive,
+                 #documented_derive
+        )]
         #[builder(parent = #parent_ty)]
         #proxy_struct
+
     }
     .into()
 }
@@ -128,9 +132,12 @@ pub fn impl_load_from_disk(ast: &StructWithFields) -> TokenStream {
         impl #disk_trait for #proxy_name {
             type Error = #error_ty;
             fn from_path<P: AsRef<std::path::Path> + std::fmt::Debug + Clone>(path: P) -> Result<Self, Self::Error> {
-                let file = std::fs::File::open(path)?;
-                let reader = std::io::BufReader::new(file);
-                serde_json::from_reader(reader)?
+                let mut file = std::fs::File::open(path)?;
+                // String has better parsing speed, see [issue](https://github.com/serde-rs/json/issues/160#issuecomment-253446892)
+                let mut s = String::new();
+                std::io::Read::read_to_string(&mut file, &mut s)?;
+                let res: Self = serde_json::from_str(&s)?;
+                Ok(res)
             }
         }
     }.into()
@@ -154,13 +161,20 @@ fn gen_proxy_struct(mut ast: StructWithFields) -> StructWithFields {
         field.ty = parse_quote! {
             Option<#ty>
         };
+        // Fields that already wrap an option should have a
+        // custom deserializer so that json `null` becomes
+        // `Some(None)` and not just `None`
+        if field.has_option {
+            let de_helper = stringify! { ::iroha_config_base::proxy::some_option };
+            let serde_attr: syn::Attribute =
+                parse_quote! { #[serde(default, deserialize_with = #de_helper)] };
+            field.attrs.push(serde_attr);
+        }
         field.has_option = true;
-        // Also remove `#[serde(default = ..)]` if present
-        // as it breaks proxy deserialization
-        utils::remove_attr(&mut field.attrs, "serde");
     });
     ast.ident = format_ident!("{}Proxy", ast.ident);
-    // Removing as `..Proxy` has its own doc
+    // Removing struct-level docs as `..Proxy` has its own doc,
+    // but not the field documentation as they stay the same
     utils::remove_attr(&mut ast.attrs, "doc");
     ast
 }
--- a/config/base/derive/src/utils.rs
+++ b/config/base/derive/src/utils.rs
@@ -244,7 +244,6 @@ impl Parse for StructWithFields {
                 .map(|field| StructField::from_ast(field, &env_prefix))
                 .collect(),
             env_prefix,
-            // parent_ty,
             _semi_token: input.parse()?,
         })
     }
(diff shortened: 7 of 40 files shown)


config/base/derive/src/documented.rs
```diff
@@ -47,17 +47,30 @@ fn impl_get_doc_recursive(docs: Vec<LitStr>, ast: &StructWithFields) -> proc_mac
         .zip(docs)
         .map(|(field, documentation)| {
             let inner_thing = field.has_inner;
```
Erigara, line 49: `option_thing` and `inner_thing` non-informative to me, in this case `has_inner`/`has_option` would be more descriptive.

config/base/derive/src/documented.rs
```diff
@@ -47,17 +47,30 @@ fn impl_get_doc_recursive(docs: Vec<LitStr>, ast: &StructWithFields) -> proc_mac
...
+            let option_thing = field.has_option;
             let ty = &field.ty;
             let ident = &field.ident;
-            if inner_thing {
+            let documented_trait = quote! { iroha_config_base::proxy::Documented };
```
Erigara, line 53: i would prefer strong typing here

```suggestion
            let documented_trait: syn::Path = syn::parse_quote!{ iroha_config_base::proxy::Documented };
```

config/src/sumeragi.rs
```diff
@@ -24,13 +24,12 @@ view! {
...
     /// and a list of `TRUSTED_PEERS`.
-    #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Documented, Proxy, LoadFromEnv)]
-    #[serde(default)]
+    #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Proxy, Documented, LoadFromEnv)]
     #[serde(rename_all = "UPPERCASE")]
     #[config(env_prefix = "SUMERAGI_")]
     pub struct Configuration {
         /// The key pair consisting of a private and a public key.
-        #[serde(skip)]
+        //TODO: Why skip was here?
```
Erigara, line 32: Maybe because keys was substituted for sumeragi in `finalization` stage?

config/src/iroha.rs
```diff
@@ -12,8 +15,8 @@ use super::*;
 // Generate `ConfigurationView` without the private key
 view! {
     /// Configuration parameters for a peer
-    #[derive(Debug, Clone, Deserialize, Serialize, Proxy, LoadFromEnv, Documented)]
-    #[serde(default)]
+    #[derive(Debug, Clone, Deserialize, Serialize, Proxy, Documented, LoadFromEnv)]
+    //#[serde(default)]
```
Erigara, line 19: commented code

config/src/iroha.rs
```diff
@@ -60,29 +63,76 @@ view! {
...
+            network: Some(network::ConfigurationProxy::default()),
+            telemetry: Some(telemetry::ConfigurationProxy::default()),
         }
     }
 }
 
+impl ConfigurationProxy {
+    pub fn finalize(&mut self) -> Result<()> {
+        if let Some(sumeragi_proxy) = &mut self.sumeragi {
+            // First, iroha public/private key and sumeragi keypair are interchangeable, but at least
```
Erigara, line 89: I would suggest to prohibit providing keys to sumeragi at all (it would be sumeragi private field). Imo there should be only one way to configurate keys.

config/src/iroha.rs
```diff
@@ -60,29 +63,76 @@ view! {
...
+            } else if let Some(ref key_pair) = sumeragi_proxy.key_pair {
+                self.public_key = Some(key_pair.public_key().clone());
+                self.private_key = Some(key_pair.private_key().clone());
+            } else {
+                eyre::bail!(ConfigError::ProxyBuildError(
+                    "Neither Iroha public and private key nor sumeragi keypair supplied".to_owned()
+                ))
+            }
+
+            // Second, torii gateway and sumeragi peer id are interchangeable too
```
Erigara, line 103: Same as above.

config/src/iroha.rs
ilchu: Makes sense. I wanted to leave it more flexible, but that actually complicates finalization and such, so your approach should allow for a leaner implementation.

config/base/derive/src/proxy.rs
```diff
@@ -7,23 +7,27 @@ use crate::utils;
...
         /// Proxy configuration structure to be used as an intermediate
         /// for config loading
-        #[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize,
+        #[derive(Debug, Clone, serde::Serialize, serde::Deserialize,
                  #builder_derive,
                  #loadenv_derive,
                  #disk_derive,
-                 #combine_derive)]
+                 #combine_derive,
+                 #documented_derive
```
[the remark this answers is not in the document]
ilchu: Doing it this way for the sake of potential hot-swapping if the paths to traits ever change.

cli/src/lib.rs
```diff
@@ -124,11 +127,9 @@ where
         query_judge: QueryJudgeBoxed,
     ) -> Result<Self> {
         let broker = Broker::new();
-        let mut config = match Configuration::from_path(&args.config_path) {
-            Ok(config) => config,
-            Err(_) => Configuration::default(),
-        };
-        config.load_environment()?;
+        let mut proxy = ConfigurationProxy::from_path(&args.config_path)?;
+        proxy.load_environment()?;
```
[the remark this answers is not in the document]
ilchu: That's the old API of loading from the environment that we had, and I didn't budge much. I suppose @appetrosyan would agree with you here. As I see, the added benefit of having two separate env and path proxy would help with inspection of the values and thus debugging. Will fix it in a later commit.

config/base/derive/src/view.rs
```diff
@@ -88,6 +87,7 @@ mod gen {
         remove_attr_struct(&mut ast, "view");
         remove_attr_struct(&mut ast, "config");
         remove_attr_struct(&mut ast, "builder");
+        remove_attr_struct(&mut ast, "proxy");
```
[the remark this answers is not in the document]
ilchu: Yeah, don't see a way to escape that yet, as view is a proc macro with the whole struct inside of it, we have to do this manual plucking.

client/src/lib.rs
```diff
@@ -10,21 +8,24 @@ mod http_default;
...
     pub fn get_client_config(key_pair: &KeyPair) -> Configuration {
         let (public_key, private_key) = key_pair.clone().into();
-        Configuration {
-            public_key,
-            private_key,
-            account_id: "alice@wonderland".parse().expect("Should not fail."),
-            torii_api_url: iroha_primitives::small::SmallStr::from_str(uri::DEFAULT_API_URL),
-            ..Configuration::default()
-        }
+        let mut proxy = ConfigurationProxy::default();
```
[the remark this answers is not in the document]
ilchu: True, my mind slipped a bit as I tried to make it work. Old style is defo more readable.

p2p/tests/integration/p2p.rs
```diff
@@ -203,7 +206,9 @@ async fn multiple_networks() {
     let log_config = Configuration {
         max_log_level: Level::TRACE.into(),
         compact_mode: false,
-        ..Configuration::default()
+        ..ConfigurationProxy::default()
+            .build()
+            .expect("Default logger config should always build")
```
appetrosyan, line 211:
```suggestion
            .expect("Default logger config failed to build. This is a programmer error")
```

logger/tests/telemetry.rs
```diff
@@ -2,12 +2,19 @@
...
+use iroha_logger::{info, init, ConfigurationProxy, Telemetry, TelemetryFields};
 use tokio::time;
 
 #[tokio::test]
 async fn telemetry_separation_default() {
-    let (mut receiver, _) = init(&Configuration::default()).unwrap().unwrap();
+    let (mut receiver, _) = init(
+        &ConfigurationProxy::default()
+            .build()
+            .expect("Default logger should always build"),
```
appetrosyan, line 14:
```suggestion
            .unwrap(),
```

futures/Cargo.toml
```diff
@@ -14,6 +14,7 @@ telemetry = ["iroha_futures_derive/telemetry"]
 [dependencies]
 iroha_futures_derive = { version = "=2.0.0-pre-rc.7", path = "derive" }
 iroha_logger = { version = "=2.0.0-pre-rc.7", path = "../logger" }
+iroha_config = { version = "=2.0.0-pre-rc.7", path = "../config" }
```
appetrosyan, line 17: Increased coupling. I thought we would actually decrease it.

appetrosyan: Could we move the tests out of futures?

config/src/telemetry.rs
```diff
@@ -17,36 +16,26 @@ pub struct Configuration {
...
     /// The maximum exponent of 2 that is used for increasing delay between reconnections
-    #[serde(default = "default_max_retry_delay_exponent")]
     pub max_retry_delay_exponent: u8,
     /// The filepath that to write dev-telemetry to
     #[config(serde_as_str)]
     pub file: Option<std::path::PathBuf>,
 }
 
-impl Default for Configuration {
+impl Default for ConfigurationProxy {
```
appetrosyan, line 27: Since we don't actually use this outside the compile-time context, we might consider using the `ConstDefault` trait:

```rust
pub trait ConstDefault {
     const DEFAULT: Self;
}

File: config/src/sumeragi.rs (hunk 3 above)
Comment (appetrosyan) [reply config/src/sumeragi.rs:32]: Keys must be substituted later. So deserializing them here makes no sense.

File: config/src/iroha.rs (hunk 5 above)
Comment (appetrosyan) [reply config/src/iroha.rs:95]: >  I wanted to leave it more flexible, but that actually complicates finalization and such

Correct assessment. But I'd be more specific. Configs need to have certain fields in sync. The easiest (and most correct) way to achieve that is by enforcing that synchronised config types are created together. 

You can prevent accidental desynchronised configs from being created by hiding necessary fields which should be synced. This is what @Erigara is proposing, which IMO is the better approach.

File: config/base/derive/src/proxy.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (appetrosyan) [reply config/base/derive/src/proxy.rs:26]: Good point. Avoid hard-coding if possible.

File: config/base/derive/src/view.rs (hunk 9 above)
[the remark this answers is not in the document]
Comment (appetrosyan) [reply config/base/derive/src/view.rs:90]: The first step to automate it, i identifying what needs plucking, why and leaving it as a comment for someone else to fix later.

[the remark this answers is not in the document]
Comment (ilchu) [reply config/base/derive/src/view.rs:90]: Leaving it as a comment in that place for now.

File: logger/tests/telemetry.rs (hunk 12 above)
Comment (ilchu) [reply logger/tests/telemetry.rs:14]: So the difference between two messages signifies that one is an integration test, and the other a unit one, I see

File: futures/Cargo.toml (hunk 13 above)
Comment (ilchu) [reply futures/Cargo.toml:17]: Created #2745.

File: config/src/telemetry.rs (hunk 14 above)
Comment (ilchu) [reply config/src/telemetry.rs:27]: To be dealt in #2746.

File: config/src/sumeragi.rs (hunk 3 above)
Comment (ilchu) [reply config/src/sumeragi.rs:32]: Still it seems that line is unneeded on the `Configuration` now (and trying to put it there complains about a lack of `Default` impl), but might need to be put on proxy specifically. Also as a side question for sake of clarification: should we ditch the `Deserialize` derive on `Configuration` altogether? It feels like that responsibility is completely superceded by the `Proxy`, but I might be missing some other points, so it needs checking.

File: config/src/iroha.rs (hunk 5 above)
Comment (ilchu) [reply config/src/iroha.rs]: Opted for just overwriting whatever was in the `key_pair` field for now. Would that be sufficient or maybe it'll be more prudent to also slap a `#[serde(skip)]` on the relevant proxy field? It would be a bit hacky as that has to be done inside the derive macro itself.

File: docs/source/references/config.md [hunk 15]
```diff
@@ -8,11 +8,8 @@ The following is the default configuration used by Iroha.
 
 ```json
 {
-  "PUBLIC_KEY": "ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b",
-  "PRIVATE_KEY": {
-    "digest_function": "ed25519",
-    "payload": "282ed9f3cf92811c3818dbc4ae594ed59dc1a2f78e4241e31924e101d6b1fb831c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b"
-  },
+  "PUBLIC_KEY": null,
+  "PRIVATE_KEY": null,
```
Erigara, line 12: In the beginning of the section we say that this is default iroha configuration, but this config is non working, because it has missing fields. Is this ok?

p2p/tests/integration/p2p.rs
```diff
@@ -36,7 +37,9 @@ fn setup_logger() {
         let log_config = Configuration {
             max_log_level: Level::TRACE.into(),
             compact_mode: false,
-            ..Configuration::default()
+            ..ConfigurationProxy::default()
+                .build()
+                .expect("Default logger config failed to build. This is a programmer error")
```
Erigara, line 42: We have combination of `default -> build -> expect` all over the place.
I wonder if we can add method like `ConfigurationProxy::build_default() -> Configuration` which would return `Configuration` object directly.

config/src/iroha.rs
Erigara: I think overwriting is ok for now, but in the future i think we should exclude such fields from de/serialization, because there is no point in having fields that will be overwritten in any case.

config/base/derive/src/utils.rs
```diff
@@ -214,6 +214,7 @@ impl ToTokens for StructField {
 pub struct StructWithFields {
     pub attrs: Vec<Attribute>,
     pub env_prefix: String,
+    // pub parent_ty: Type,
```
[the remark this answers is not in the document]
ilchu: Weird, deleted that, but as I squashed commits locally, it got undone. Thanks for noticing.

config/src/iroha.rs
ilchu: Agree, making an issue: #2715.

config/src/iroha.rs
```diff
@@ -60,26 +62,82 @@ view! {
...
+        }
+    }
+}
+
+impl ConfigurationProxy {
+    /// Finalisation of Iroha config, i.e. in case where only some of mutually equivalent fields
+    /// were provided, instantiation of all the missing fields happens.
+    ///
+    /// # Errors
+    /// Bails early if none of the fields from the corresponding mutually equivalent set were
```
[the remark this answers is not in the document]
ilchu: Followed the name of `eyre`'s [macro](https://docs.rs/eyre/latest/eyre/macro.bail.html), but I guess the meaning is quite narrow indeed.

config/src/iroha.rs
```diff
@@ -60,26 +62,82 @@ view! {
...
+                            "Iroha `public_key` should have been initialized above at the latest",
+                        ),
+                    ));
+                }
+            }
+            // Finally, if trusted peers were not supplied, we can fall back to inserting itself as
+            // the only trusted one
+            if sumeragi_proxy.trusted_peers.is_none() {
+                sumeragi_proxy.insert_self_as_trusted_peers()
+            }
```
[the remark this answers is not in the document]
ilchu: Technically the `Configuration` would fail at build stage, as the `torii.p2p_addr` is `None` by default. We could bail early here, e.g. adding `else` after `if let Some(torii_proxy) = ..` in finalization, but I guess it would already be informative enough when one of the `expect()` calls fails or `build()` crashes.

config/src/iroha.rs
```diff
@@ -128,16 +186,20 @@ mod tests {
...
-        let configuration = Configuration::from_path(CONFIGURATION_PATH)
+        let cfg_proxy = ConfigurationProxy::from_path(CONFIGURATION_PATH)
             .wrap_err("Failed to read configuration from example config")?;
-        assert_eq!("127.0.0.1:1337", configuration.torii.p2p_addr);
-        assert_eq!(1000, configuration.sumeragi.block_time_ms);
+        assert_eq!("127.0.0.1:1337", cfg_proxy.torii.unwrap().p2p_addr.unwrap());
+        assert_eq!(
+            10000,
+            cfg_proxy.block_sync.unwrap().gossip_period_ms.unwrap()
+        );
```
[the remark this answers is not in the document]
ilchu: Thing is, the default json won't build as it doesn't currently fall back on default impl at all, and thus requires that *all* fields be provided, which is terribly unergonomic. A good catch again. Would probably implement [this](https://github.com/hyperledger/iroha/pull/2706#discussion_r971965855) and incorporate it into build logic.

docs/source/references/config.md
ilchu: There's still quite some work that needs to be done on kagami docs, and there are quite a few places I don't like about the new `config.md`, such as mentioning the types of `ConfigurationProxy`. At the very least, it needs a small explanation at the start that any fields shown as `null` should be user provided. At best, come up with a new workaround for `kagami` instead of deriving `Documented` on `Proxy`. But as for latter, I'm a bit at my wit's end for now.

p2p/tests/integration/p2p.rs
ilchu: Like this one, a readability boost. Will do.

client/tests/integration/config.rs
```diff
@@ -13,10 +13,16 @@ fn get_config() {
...
+        serde_json::from_value(test_client.get_config_value().unwrap()).unwrap();
+    assert_eq!(
+        cfg_proxy.block_sync.unwrap().build().unwrap(),
+        test.block_sync
+    );
+    assert_eq!(cfg_proxy.network.unwrap().build().unwrap(), test.network);
+    assert_eq!(
+        cfg_proxy.telemetry.unwrap().build().unwrap(),
+        test.telemetry
+    );
```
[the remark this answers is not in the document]
ilchu: Ah, my bad. As the routing fetches a `ConfigurationView` on `GetConfiguration` request, this is only to be expected that the `private_key` is hidden, as that's the whole purpose. Which means that there's no way to build the proxy here unfortunately, so it should stay as is.

p2p/tests/integration/p2p.rs
ilchu: On second thought, not really feasible in any of the places. E.g. logger configs in tests are supplied traces externally, and only some of the fields are default; queue configs also overload some of its default values in all of their tests.

config/src/iroha.rs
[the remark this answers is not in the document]
ilchu: I suggest leaving it as is for now as testing will be dealt with in the next PR, and that's when `config.json` will be changed appropriately too.

docs/source/references/config.md
ilchu: Just adding a note at the top of kagami docgen in the last push. Surely a makeshift measure.

config/src/iroha.rs
```diff
@@ -60,26 +62,83 @@ view! {
...
+                sumeragi_proxy.key_pair =
+                    Some(KeyPair::new(public_key.clone(), private_key.clone())?);
+            } else {
+                eyre::bail!(ConfigError::ProxyBuildError(
+                    "Iroha public and private key not supplied, instantiating sumeragi keypair is impossible"
+                        .to_owned()
+                ))
+            }
+            // Second, torii gateway and sumeragi peer id are interchangeable too
+            if let Some(torii_proxy) = &mut self.torii {
```
Erigara, line 107: I want emphasize on this idea one more time. 

If we _can_ construct `sumeragi.peer_id` from `public_key` and `torii.p2p_addr` then we probably _should_ construct it from this values. It's invalid state to have `peer_id.addr != p2p_addr` or `peer.public_key != public_key`.

I believe that we avoid a lot pain in the ass if we would restrict configuration to eliminate possible divergence of configuration parameters.

Erigara (approved):
Looks good

config/src/iroha.rs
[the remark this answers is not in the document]
ilchu: Just encountered that sort of error during property testing. I think the freshly bumped version of this func should do the trick now.

config/src/sumeragi.rs
Erigara: WIth `#[serde(skip)]` on proxy field you could just don't check `sumeragi_proxy.key_pair.is_some()` because you know that it will be `None`.

Erigara (dismissed):

config/src/client.rs
```diff
@@ -1,18 +1,15 @@
 //! Module for client-related configuration and structs
 #![allow(clippy::std_instead_of_core, clippy::std_instead_of_alloc)]
-use std::{fmt, fs::File, io::BufReader, path::Path, str::FromStr};
+use std::str::FromStr;
```
appetrosyan, line 3: You can now also do 
```suggestion
use core::str::FromStr;
```

config/base/derive/src/view.rs
```diff
@@ -46,45 +44,38 @@ mod gen {
...
+                            true
+                        })
+                        .collect();
+                    *attr = syn::parse_quote!(
+                        #[derive(#(#items),*)]
+                    )
                 }
             });
+        // TODO: Find a way to make this more ergonomic. As `View` struct
+        // are formed inside a proc macro, we have to remove unrelated attributes from `Configuration` here.
```
appetrosyan, line 78: Not a high priority problem. Create a good first issue and delegate to a FOSS contributor.

config/base/derive/src/proxy.rs
```diff
@@ -128,9 +132,12 @@ pub fn impl_load_from_disk(ast: &StructWithFields) -> TokenStream {
         impl #disk_trait for #proxy_name {
             type Error = #error_ty;
             fn from_path<P: AsRef<std::path::Path> + std::fmt::Debug + Clone>(path: P) -> Result<Self, Self::Error> {
-                let file = std::fs::File::open(path)?;
-                let reader = std::io::BufReader::new(file);
-                serde_json::from_reader(reader)?
+                let mut file = std::fs::File::open(path)?;
+                // String has better parsing speed, see [issue](https://github.com/serde-rs/json/issues/160#issuecomment-253446892)
```
appetrosyan, line 136: This is a micro-optimisation, and I'd normally be against it, however, in this case we also have a few benefits, namely that the file can be close very quickly, or that we can use another method of obtaining the config later on.

docs/source/references/config.md
```diff
@@ -120,7 +107,7 @@ The following is the default configuration used by Iroha.
 
 `BlockSynchronizer` configuration
 
-Has type `block_sync::Configuration`. Can be configured via environment variable `IROHA_BLOCK_SYNC`
+Has type `Option<block_sync::ConfigurationProxy>`. Can be configured via environment variable `IROHA_BLOCK_SYNC`
```
appetrosyan, line 110: We should add an explanatory note to the header, saying that the `Option<bool>` config variables will fall back to a default value if not provided, while the `bool` value must be specified explicitly. Ideally all `Option<…>` segments should point to that (foot)note.

appetrosyan (approved):

config/base/derive/src/view.rs
ilchu: Created #2773.

docs/source/references/config.md
ilchu: Done a very extensive one.

appetrosyan (dismissed):

appetrosyan (dismissed):

config/base/derive/src/documented.rs
```diff
@@ -152,10 +171,18 @@ fn impl_get_recursive(ast: &StructWithFields) -> proc_macro2::TokenStream {
     let variants = ast.fields
         .iter()
         .map(|field | {
-            let inner_thing = field.has_inner;
+            // if ast.ident == "ConfigurationProxy" {println!("FIELD: {}, INNER: {}, OPTION: {}", field.ident, field.has_inner, field.has_option);}
```
Erigara, line 174: commented code

Erigara (dismissed):

Erigara (approved):

docs/source/references/config.md
```diff
@@ -1,18 +1,15 @@
 # Iroha Configuration reference
 
-In this document we provide a reference and detailed descriptions of Iroha's configuration options.
+In this document we provide a reference and detailed descriptions of Iroha's configuration options. Also make sure to read thoroughly an important note[^1] for an explanation of underlying types and default values as this should clear up any confusion about them.
```
outoftardis, line 3: if the info in this note is so important, it shouldn't be a note. I'd put this info in the main body of the text, not in a footnote

docs/source/references/config.md
```diff
@@ -856,3 +813,29 @@ Has type `wasm::Configuration`. Can be configured via environment variable `WSV_
 }
 ```
 
+#### `wsv.wasm_runtime_config.fuel_limit`
+
+The fuel limit determines the maximum number of instructions that can be executed during the execution of a smart contract.
```
outoftardis, line 818: maybe 

```suggestion
The fuel limit determines the maximum number of instructions that can be executed within a smart contract.
```

docs/source/references/config.md
```diff
@@ -856,3 +813,29 @@ Has type `wasm::Configuration`. Can be configured via environment variable `WSV_
...
+
+Has type `Option<u64>`[^1]. Can be configured via environment variable `WASM_FUEL_LIMIT`
+
+```json
+1000000
+```
+
+#### `wsv.wasm_runtime_config.max_memory`
+
+Maximum amount of linear memory a given smartcontract can allocate
```
outoftardis, line 828:
```suggestion
Maximum amount of linear memory a given smart contract can allocate.
```

docs/source/references/config.md
```diff
@@ -856,3 +813,29 @@ Has type `wasm::Configuration`. Can be configured via environment variable `WSV_
...
+
+Has type `Option<u32>`[^1]. Can be configured via environment variable `WASM_MAX_MEMORY`
+
+```json
+524288000
+```
+
+[^1]: Types wrapped in a single `Option<..>` signify that there is a fallback value for them specified in the corresponding `json` block, but it only serves as a reference. A default `null` for such types means that there is no meaningful fallback for this particular value available. The default peer `config.json` file can serve as a starting point, but there are only two final sources of truth for configuration: all the variables should either be provided via environment or in a file, and no other overloading happens. Failure to provide them in any of the above two ways results in an error.
```
outoftardis, line 836: not sure I understood you correctly, so correct me if I am wrong, but I think this kind of explanation will be more clear to the users:

```suggestion
[^1]: A type wrapped in a single `Option<..>` signifies that in the corresponding `json` block there is a fallback value for this type, and that it only serves as a reference. The default `null` for such types means that there is no meaningful fallback for this particular value available. Note that all variables should either be provided via environment or in a file. Failure to provide them in any of the above two ways results in an error. This means that for configuration options that have types wrapped in a single `Option<..>` with `null` as the default, you *have to* to use environment variables.
```

docs/source/references/config.md
```diff
@@ -856,3 +813,29 @@ Has type `wasm::Configuration`. Can be configured via environment variable `WSV_
...
+
+	`Option<Option<..>>` types should be distinguished from types wrapped in singular `Option<..>`. Only the double option ones are allowed to stay `null`, meaning that *not* providing them in an environment variable or a file will not result in an error. Thus only these types are optional in the mundane sense of the word.
```
outoftardis, line 838:
```suggestion
	`Option<Option<..>>` types should be distinguished from types wrapped in a single `Option<..>`. Only the double option ones are allowed to stay `null`, meaning that *not* providing them in an environment variable or a file will not result in an error. Thus only these types are optional in the mundane sense of the word.
```

tools/kagami/src/main.rs
```diff
@@ -252,14 +251,32 @@ mod docs {
...
             writeln!(writer, "## Default configuration\n")?;
             writeln!(
                 writer,
                 "The following is the default configuration used by Iroha.\n"
             )?;
             writeln!(writer, "```json\n{}\n```\n", defaults)?;
             Self::get_markdown_with_depth(writer, &docs, &mut vec, 2)?;
+            writeln!(
+                writer,
+                "[^1]: Types wrapped in a single `Option<..>` signify that there is a fallback value for them specified in the corresponding \
```
outoftardis, line 266: see my comments for the config.md

docs/source/references/config.md
ilchu: The point is, it is a note so that it's present both in the text and linked to by the relevant types, and footnote's the only viable way I know that can do it. I can move the right after the above paragraph, though.

outoftardis: can't it be a separate section of the document that we link to?

outoftardis: the phrasing of the sentence should be changed in any case:

```suggestion
In this document we provide a reference and detailed descriptions of Iroha's configuration options. The options have different underlying types, which are denoted in code as types wrapped in a single `Option<..>` or in a double `Option<Option<..>>`. For the detailed explanation, please refer to the note at the end of this document[^1].
```

I think this also works:

```suggestion
In this document we provide a reference and detailed descriptions of Iroha's configuration options. The options have different underlying types, which are denoted in code as types wrapped in a single `Option<..>` or in a double `Option<Option<..>>`[^1].
```

client/src/client.rs
```diff
@@ -1375,12 +1379,21 @@ mod tests {
...
-            add_transaction_nonce: true,
-            ..Configuration::default()
-        };
+        let cfg = ConfigurationProxy {
+            public_key: Some(public_key),
+            private_key: Some(private_key),
+            account_id: Some(
+                "alice@wonderland"
+                    .parse()
+                    .expect("Account ID should not fail"),
```
outoftardis, line 1388:
```suggestion
                    .expect("This account ID should be valid"),
```

client/src/client.rs
```diff
@@ -1404,10 +1417,28 @@ mod tests {
...
+                    .expect("Public key not in mulithash format"),
+            ),
+            private_key: Some(iroha_crypto::PrivateKey::from_hex(
+            iroha_crypto::Algorithm::Ed25519,
+            "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
+            ).expect("Private key not hex encoded")),
+            account_id: Some(
+                "alice@wonderland"
+                    .parse()
+                    .expect("Account ID should not fail"),
```
outoftardis, line 1433:
```suggestion
                    .expect("This account ID should be valid"),
```

config/base/derive/src/documented.rs
```diff
@@ -216,7 +242,7 @@ pub fn gen_docs(ast: &StructWithFields) -> Vec<LitStr> {
                 });
             let real_doc = real_doc.map(|doc| doc.value() + "\n\n").unwrap_or_default();
             let docs = format!(
-                "{}Has type `{}`. Can be configured via environment variable `{}`",
+                "{}Has type `{}`. Can be configured via environment variable `{}`. More explanation in [this section](#configuration-types).",
```
outoftardis, line 245:
```suggestion
                "{}Has type `{}`. Can be configured via environment variable `{}`. Check [configuration types](#configuration-types) for more details.",
```

docs/source/references/config.md
```diff
@@ -120,7 +115,7 @@ The following is the default configuration used by Iroha.
 
 `BlockSynchronizer` configuration
 
-Has type `block_sync::Configuration`. Can be configured via environment variable `IROHA_BLOCK_SYNC`
+Has type `Option<block_sync::ConfigurationProxy>`. Can be configured via environment variable `IROHA_BLOCK_SYNC`. More explanation in [this section](#configuration-types).
```
outoftardis, line 118:
```suggestion
Has type `Option<block_sync::ConfigurationProxy>`. Can be configured via environment variable `IROHA_BLOCK_SYNC`. Refer to [configuration types](#configuration-types) for details.
```

same in other places in this file

docs/source/references/config.md
```diff
@@ -856,3 +821,23 @@ Has type `wasm::Configuration`. Can be configured via environment variable `WSV_
 }
 ```
 
+#### `wsv.wasm_runtime_config.fuel_limit`
+
+The fuel limit determines the maximum number of instructions that can be executed during within a smart contract.
```
outoftardis, line 826:
```suggestion
The fuel limit determines the maximum number of instructions that can be executed within a smart contract.
```

docs/source/references/config.md
```diff
@@ -1,18 +1,23 @@
 # Iroha Configuration reference
 
-In this document we provide a reference and detailed descriptions of Iroha's configuration options.
+In this document we provide a reference and detailed descriptions of Iroha's configuration options. Also make sure to read thoroughly the following important [section](#configuration-types) for an explanation of underlying types and default values as this should clear up any confusion about them.
```
outoftardis, line 3: see https://github.com/hyperledger/iroha/pull/2706#discussion_r977425829

docs/source/references/config.md
```diff
@@ -1,18 +1,23 @@
...
+
+## Configuration types
+
+A type wrapped in a single `Option<..>` signifies that in the corresponding `json` block there is a fallback value for this type, and that it only serves as a reference. If a default for such a type has a `null` value, it means that there is no meaningful fallback available for this particular value. All these default values can be freely obtained from a provided [sample configuration file](../../../configs/peer/config.json), but it should only serve as a starting point. If left unchanged, the sample configuration file as is would still fail to build due to it having `null` in place of [public](#public_key) and [private](#private_key) keys as well as [endpoint](#torii.api_url) [URLs](#torii.telemetry_url). These should be provided either by modifying the sample config file or as environment variables. No other overloading of configuration values happens besides reading them from a file and capturing the environment variables. For both types of configuration options wrapped in a single `Option<..>` (i.e. both those that have meaningful defaults and those that have `null`), failure to provide them in any of the above two ways results in an error.
```
outoftardis, line 7:
```suggestion

### `Option<..>` 
A type wrapped in a single `Option<..>` signifies that in the corresponding `json` block there is a fallback value for this type, and that it only serves as a reference. If a default for such a type has a `null` value, it means that there is no meaningful fallback available for this particular value.

All the default values can be freely obtained from a provided [sample configuration file](../../../configs/peer/config.json), but it should only serve as a starting point. If left unchanged, the sample configuration file would still fail to build due to it having `null` in place of [public](#public_key) and [private](#private_key) keys as well as [endpoint](#torii.api_url) [URLs](#torii.telemetry_url). These should be provided either by modifying the sample config file or as environment variables. No other overloading of configuration values happens besides reading them from a file and capturing the environment variables.

For both types of configuration options wrapped in a single `Option<..>` (i.e. both those that have meaningful defaults and those that have `null`), failure to provide them in any of the above two ways results in an error.
```

docs/source/references/config.md
```diff
@@ -1,18 +1,23 @@
...
+
+`Option<Option<..>>` types should be distinguished from types wrapped in a single `Option<..>`. Only the double option ones are allowed to stay `null`, meaning that **not** providing them in an environment variable or a file will **not** result in an error. Thus only these types are truly optional in the mundane sense of the word. An example of this distinction is genesis [public](#genesis.account_public_key) and [private](#genesis.account_private_key) key. While the first one is a single `Option<..>` wrapped type, the latter is wrapped in `Option<Option<..>>`. This means that the genesis *public* key should always be provided by the user, be it via a file config or an environment variable, whereas the *private* key is only needed for the peer that submits the genesis block, and can be omitted for all others. The same logic goes for other double option fields such as logger file path.
```
outoftardis, line 9:
```suggestion
### `Option<Option<..>>`
`Option<Option<..>>` types should be distinguished from types wrapped in a single `Option<..>`. Only the double option ones are allowed to stay `null`, meaning that **not** providing them in an environment variable or a file will **not** result in an error.

Thus, only these types are truly optional in the mundane sense of the word. An example of this distinction is genesis [public](#genesis.account_public_key) and [private](#genesis.account_private_key) key. While the first one is a single `Option<..>` wrapped type, the latter is wrapped in `Option<Option<..>>`. This means that the genesis *public* key should always be provided by the user, be it via a file config or an environment variable, whereas the *private* key is only needed for the peer that submits the genesis block, and can be omitted for all others. The same logic goes for other double option fields such as logger file path.
```

docs/source/references/config.md
```diff
@@ -1,18 +1,23 @@
...
+
+Finally, special note about sumeragi fields with `null` as default: only the [`trusted_peers`](#sumeragi.trusted_peers) field out of the three can be initialized via a provided file or an environment variable. The other two, namely [`key_pair`](#sumeragi.key_pair) and [`peer_id`](#sumeragi.peer_id), go through a process of finalization where their values are derived from the corresponding ones in the uppermost Iroha config (for `key_pair`) or the Torii config (in case of [`p2p_addr`](#torii.p2p_addr)). This ensures that these linked fields stay in sync, and prevents the programmer error when different values are provided to these field pairs. Providing either `sumeragi.key_pair` or `sumeragi.peer_id` by hand will result in an error, as it should never be done directly.
```
outoftardis, line 11:
```suggestion
### Sumeragi: default `null` values

A special note about Sumeragi fields with `null` as default: only the [`trusted_peers`](#sumeragi.trusted_peers) field out of the three can be initialized via a provided file or an environment variable.

The other two fields, [`key_pair`](#sumeragi.key_pair) and [`peer_id`](#sumeragi.peer_id), go through a process of finalization where their values are derived from the corresponding ones in the uppermost Iroha config (for `key_pair`) or the Torii config (in case of [`p2p_addr`](#torii.p2p_addr)). This ensures that these linked fields stay in sync, and prevents the programmer error when different values are provided to these field pairs. Providing either `sumeragi.key_pair` or `sumeragi.peer_id` by hand will result in an error, as it should never be done directly.
```

docs/source/references/config.md
ilchu: As all this is generated from the above one, choosing this wording

docs/source/references/config.md
ilchu: Agreed, sounds quite harsh when I think about it. Thank you.

Erigara (approved):

outoftardis (approved):
lgtm

Issue #2426, opened 2022-07-01
Simplify Sumeragi Execution


SamHSmith:
Sumeragi will probably be easier to work on if it easier to see how it executes.

Issue #2078, opened 2022-04-08
Transactions becomes Rejected under high load conditions


s8sato:
## Expected Behavior
Ideally, transaction results are not affected by transaction load or other network status

## Actual Behavior
As the transaction interval narrows, the number of rejected transactions increases (and eventually the program crashes).

## Steps to Reproduce

> commit dbe8a901c330454ed44446923ea462a29763879e
> 
> [feature] #1952: Add a TPS benchmark as a standard for optimizations (#1963)

1. Configure `interval_us_per_tx` in `client/benches/tps/config.json`

2. Simulate frequent transactions:

    ```bash
    cd client
    cargo run --example tps-oneshot
    ```

3. See the `txs_rejected` log appear at INFO level
    - After several local trials @s8sato can see `(80000, 0)` and `(50000, 8)` of `(interval_us_per_tx, txs_rejected)`

### Hint?
- All the rejection reason seem to be `FindError`
- One of sample instructions contains a query request
- It seems always `txs_rejected=0` when one of the two instructions in `tps_oneshot::lib::MeasurerUnit::instructions` is disabled

appetrosyan:
Related to #1753

SamHSmith:
This is fixed by new-sumeragi. Make sure to run the tps-oneshot in release though, otherwise the nodes can never get round to reporting new blocks because they are so overwhelmed with transactions.

PR #2711, opened 2022-09-08
Merge New-Sumeragi into iroha2-dev

SamHSmith:
### Description of the Change

Rewrite the sumeragi consensus algorithm as a single threaded operation.

### Issue
#2426 

### Benefits

Testing needs to qualify this, but it is probably already a lot faster and there is a lot of headroom for optimization still. Many bugs in the consensus may be fixed due to the rewrite.

### Possible Drawbacks

There may be new bugs introduced.


data_model/src/trigger/set.rs
```diff
@@ -403,69 +415,67 @@ impl Set {
...
     where
         F: Fn(&dyn ActionTrait, Event) -> std::result::Result<(), E> + Send + Copy,
         E: Send + Sync,
     {
         let mut succeed = Vec::new();
         let mut errors = Vec::new();
 
-        let apply_f = move |action: &dyn ActionTrait, event: Event| {
+        let apply_f = move |succeed_clone: &mut Vec<_>,
+                            errors_clone: &mut Vec<_>,
```
appetrosyan, line 427: these names are incredibly confusing.

data_model/src/trigger/set.rs
```diff
@@ -304,17 +312,21 @@ impl Set {
     /// This actions will be inspected in the next [`Set::inspect_matched()`] call
     // Passing by value to follow other `handle_` methods interface
     #[allow(clippy::needless_pass_by_value)]
-    pub fn handle_pipeline_event(&self, event: PipelineEvent) {
+    pub fn handle_pipeline_event(&mut self, event: PipelineEvent) {
+        let mut acc = Vec::new();
```
appetrosyan, line 316: Consider using a smallvec. Stack space is relatively cheap and we avoid quite a bit of indirection.

data_model/src/trigger/set.rs
```diff
@@ -273,29 +273,37 @@ impl Set {
...
-            None => return,
-        };
+    pub fn handle_execute_trigger_event(&mut self, event: ExecuteTriggerEvent) {
+        let pair;
+        {
+            let entry = match self.by_call_triggers.get(&event.trigger_id) {
+                Some(entry) => entry,
+                None => return,
+            };
+            pair = (entry.key().clone(), entry.value().clone());
```
appetrosyan, line 303:
```suggestion
        let pair = {
            let entry = match self.by_call_triggers.get(&event.trigger_id) {
                Some(entry) => entry,
                None => return,
            };
            (entry.key().clone(), entry.value().clone())
```

appetrosyan: In rust everything is an expression so you can do the assignment without duplicating the name.

data_model/src/trigger/set.rs
```diff
@@ -273,29 +273,37 @@ impl Set {
     /// This actions will be inspected in the next [`Set::handle_data_event()`] call
     // Passing by value to follow other `handle_` methods interface
     #[allow(clippy::needless_pass_by_value)]
-    pub fn handle_data_event(&self, event: DataEvent) {
+    pub fn handle_data_event(&mut self, event: DataEvent) {
+        let mut acc = Vec::new();
```
appetrosyan, line 277: Conisder using a smallvec.

core/test_network/src/lib.rs
```diff
@@ -355,7 +344,7 @@ where
 /// When unsuccessful after `MAX_RETRIES`.
 pub fn wait_for_genesis_committed(clients: &[Client], offline_peers: u32) {
     const POLL_PERIOD: Duration = Duration::from_millis(1000);
-    const MAX_RETRIES: u32 = 60 * 3; // 3 minutes
+    const MAX_RETRIES: u32 = 10; // 3 minutes
```
appetrosyan, line 347: And the comment is still valid, because we're retrying every 18 seconds.

core/test_network/src/lib.rs
```diff
@@ -221,6 +209,7 @@ where
     ) -> (Self, Client) {
         let mut configuration = Configuration::test();
         configuration.queue.maximum_transactions_in_block = max_txs_in_block;
+        configuration.logger.max_log_level = iroha_logger::Level::INFO.into();
```
appetrosyan, line 212: Really interested to see if we can set this somewhere else, so it's easier to refactor.

appetrosyan: Maybe `impl Default`

core/src/wsv.rs
```diff
@@ -452,15 +583,20 @@ impl WorldStateView {
...
+    ) -> Vec<VersionedCommittedBlock> {
+        let mut iterator = self.blocks.iter();
+        if Hash::zeroed() != hash.into() {
+            for block in &mut iterator {
+                if block.hash() == hash {
+                    break;
+                }
+            }
+        }
+        return iterator.map(|b| b.clone()).collect();
```
appetrosyan, line 599: This block has a lot of weird things going on. Why do we have a mutable borrow if we don't use it?

core/src/wsv.rs
```diff
@@ -342,15 +474,14 @@ impl WorldStateView {
...
-    async fn execute_transactions(&self, block: &CommittedBlock) -> Result<()> {
+    fn execute_transactions(&mut self, block: &CommittedBlock) -> Result<()> {
         // TODO: Should this block panic instead?
         for tx in &block.transactions {
             self.process_executable(
                 &tx.as_v1().payload.instructions,
                 tx.payload().account_id.clone(),
             )?;
             self.transactions.insert(tx.hash());
-            task::yield_now().await;
```
appetrosyan, line 353: Good catch. I would personally be too afraid to check if this is actually useful.

core/src/wsv.rs
```diff
@@ -636,11 +770,10 @@ impl WorldStateView {
...
 
     /// Initializes WSV with the blocks from block storage.
-    #[iroha_futures::telemetry_future]
-    pub async fn init(&self, blocks: Vec<VersionedCommittedBlock>) {
+    pub fn init(&mut self, blocks: Vec<VersionedCommittedBlock>) {
         for block in blocks {
             #[allow(clippy::panic)]
-            if let Err(error) = self.apply(block).await {
+            if let Err(error) = self.apply(block) {
                 error!(%error, "Initialization of WSV failed");
```
appetrosyan, line 777: Do we really need both? What happens if we just `.expect`.

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
...
+                .remove_zeros(&self.world.triggers.data_triggers);
+            self.world
+                .triggers
+                .remove_zeros(&self.world.triggers.pipeline_triggers);
+            self.world
+                .triggers
+                .remove_zeros(&self.world.triggers.time_triggers);
+            self.world
+                .triggers
+                .remove_zeros(&self.world.triggers.by_call_triggers);
```
appetrosyan, line 426:
```suggestion
let triggers = self.world.triggers;
triggers.remove_zeros(&triggers.data_triggers); 
// etc.
```

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
...
+                            })
+                        }
+                    };
+                }
+
+                if errors.is_empty() {
+                    (succeed, Ok(()))
+                } else {
+                    (succeed, Err(errors))
+                }
```
appetrosyan, line 400:
```suggestion
(succeed, if errors.is_empty() {Ok(())} else {Err(errors)})
```
Though if we think about the function it could be done even more compactly.

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
...
+            };
+
+            for id in &succeed {
+                // Ignoring error if trigger has not `Repeats::Exact(_)` but something else
+                let _mod_repeats_res = self.world.triggers.mod_repeats(id, |n| {
+                    if n == 0 {
+                        // Possible i.e. if one trigger burned it-self or another trigger, cause we
+                        // decrease the number of execution after successful execution
```
appetrosyan, line 408: *because.

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
...
+                    (succeed, Ok(()))
+                } else {
+                    (succeed, Err(errors))
+                }
+            };
+
+            for id in &succeed {
+                // Ignoring error if trigger has not `Repeats::Exact(_)` but something else
+                let _mod_repeats_res = self.world.triggers.mod_repeats(id, |n| {
+                    if n == 0 {
```
appetrosyan, line 406: Rust is expression based, this if can be used as a ternary operator.

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
     /// you likely have data corruption.
     /// - If trigger execution fails
     /// - If timestamp conversion to `u64` fails
-    #[iroha_futures::telemetry_future]
-    #[log(skip(self, block))]
     #[allow(clippy::expect_used)]
-    pub async fn apply(&self, block: VersionedCommittedBlock) -> Result<()> {
+    pub fn apply(&mut self, block: VersionedCommittedBlock) -> Result<()> {
```
appetrosyan, line 283: This function at the very least needs to be broken up, it's too long to be practical.

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
...
+                            })
+                        }
+                        Event::Time(_) => {
+                            let dangerous_reference = unsafe {
+                                let ptr = &self.world.triggers as *const TriggerSet; // TODO: REALLY FIX THIS
+                                &*ptr
+                            };
+
+                            dangerous_reference.time_triggers.get(&id).map(|entry| {
+                                let action = entry.value();
```
appetrosyan, line 360: Each of these is repeated. You can compress this with either a generic or a macro.

core/src/wsv.rs
```diff
@@ -279,23 +279,155 @@ impl WorldStateView {
...
+
+                                match self.process_trigger(action, event_type) {
+                                    Ok(()) => succeed.push(id),
+                                    Err(err) => errors.push(err),
+                                }
+                            })
+                        }
+                        Event::Pipeline(_) => {
+                            let dangerous_reference = unsafe {
+                                let ptr = &self.world.triggers as *const TriggerSet; // TODO: REALLY FIX THIS
```
appetrosyan, line 335: OK. This comment tells us that something is wrong and the fact that it' bad can be undestood from `unsafe`. Imagine that you're not the one fixing it. An issue number would be just as effective and way more informative.

cli/src/torii/routing.rs
```diff
@@ -250,36 +257,31 @@ async fn handle_post_configuration(
...
 
-    let mut rx = wsv.subscribe_to_new_block_notifications();
-    stream_blocks(&mut from_height, wsv, &mut stream).await?;
-
     loop {
-        rx.changed().await?;
-        stream_blocks(&mut from_height, wsv, &mut stream).await?;
+        let blocks = sumeragi.blocks_from_height(from_height.try_into().unwrap());
+        stream_blocks(&mut from_height, blocks, &mut stream).await?;
+        std::thread::sleep(std::time::Duration::from_millis(10));
```
appetrosyan, line 273: This is a code smell. I'd add a comment and revisit.

cli/src/torii/tests.rs
```diff
@@ -30,6 +30,8 @@ use crate::{
     stream::{Sink, Stream},
 };
 
+/* NEEDS TO TALK TO A SUMERAGI
```
appetrosyan, line 33: Attach issue. Anything that's commented out must either be removed or re-enabled.

client/benches/tps/lib.rs
```diff
@@ -92,14 +92,15 @@ impl Config {
         // Sleep to let the blocks to be committed on all peers
         thread::sleep(core::time::Duration::from_secs(5));
         let blocks_out_of_measure = 1 + MeasurerUnit::PREPARATION_BLOCKS_NUMBER * self.peers;
-        let mut blocks = network
+        let mut blocks_wsv = network
```
appetrosyan, line 95:
```suggestion
        let mut blocks =  { 
           network.
           // etc. 
           .blocks()
           .skip(blocks_out_of_measure as usize);
        };
```

client/benches/tps/lib.rs
```diff
@@ -113,7 +114,9 @@ impl Config {
         let tps = txs_accepted as f64 / elapsed_secs;
         iroha_logger::info!(%tps, %txs_accepted, %elapsed_secs, %txs_rejected);
         if txs_rejected > 0 {
-            return Err(eyre::eyre!("{txs_rejected} transactions were rejected"));
+            // There will be rejected transactions since we submit more than the
+            // network can process.
+            println!("txs_rejected: {}", txs_rejected);
```
appetrosyan, line 119: Debug statement left in. Please convert it to `debug!()`

core/src/block.rs
```diff
@@ -37,13 +35,12 @@ use crate::{
...
 
-const PIPELINE_TIME_MS: u64 =
-    DEFAULT_BLOCK_TIME_MS + DEFAULT_COMMIT_TIME_LIMIT_MS + DEFAULT_TX_RECEIPT_TIME_LIMIT_MS;
+const PIPELINE_TIME_MS: u64 = DEFAULT_BLOCK_TIME_MS + DEFAULT_COMMIT_TIME_LIMIT_MS;
 
 /// Default estimation of consensus duration
 #[allow(clippy::integer_division)]
 pub const DEFAULT_CONSENSUS_ESTIMATION_MS: u64 =
-    (DEFAULT_COMMIT_TIME_LIMIT_MS + PIPELINE_TIME_MS) / 2;
+    DEFAULT_BLOCK_TIME_MS + (DEFAULT_COMMIT_TIME_LIMIT_MS / 2);
```
appetrosyan, line 43: Interesting. @Arjentix thoughts?

core/src/block.rs
```diff
@@ -218,8 +215,7 @@ impl PendingBlock {
         self,
         height: u64,
         previous_block_hash: HashOf<VersionedCommittedBlock>,
-        view_change_proofs: ViewChangeProofs,
-        invalidated_blocks_hashes: Vec<HashOf<VersionedValidBlock>>,
+        view_change_proofs: Vec<Proof>,
```
appetrosyan, line 218: Consider a `SmallVec` this probably won't have more than `32` at a time.

core/src/kura.rs
```diff
@@ -119,7 +116,12 @@ impl Kura {
         {
             // Error here indicates the thread is already stopped, which is fine.
             let _res = shutdown_sender.send(());
-            thread_handle.join().expect("able to join");
+            if let Err(error) = thread_handle.join() {
+                error!(
+                    "Kura force shutdown routine: Kura thread panicked with error: {:?}",
+                    error
+                );
```
appetrosyan, line 123: Are we ignoring the error here?

core/src/sumeragi/network_topology.rs
```diff
@@ -168,38 +164,31 @@ impl Builder {
...
+    pub fn build(self, number_of_view_changes_this_round: u64) -> Result<Topology> {
         let peers = field_is_some_or_err!(self.peers)?;
         if peers.is_empty() {
             return Err(eyre!("There must be at least one peer in the network."));
         }
         let at_block = field_is_some_or_err!(self.at_block)?;
         let peers: Vec<_> = peers.into_iter().collect();
-        let n_view_changes = self.view_change_proofs.len();
-        let since_last_shuffle = n_view_changes % peers.len();
+        let n_view_changes = number_of_view_changes_this_round;
```
appetrosyan, line 179: Probably just remove. Or rename.

core/src/sumeragi/mod.rs
```diff
@@ -53,42 +59,209 @@ trait Consensus {
...
+            .lock()
+            .expect("Failed to lock on update online peers.") = online_peers;
+    }
+
+    pub fn incoming_message(&self, msg: Message) {
+        if self
+            .internal
+            .incoming_message_sender
+            .lock()
+            .expect("Lock on sender")
```
appetrosyan, line 258: Please use consistent `expect` message formats.

core/src/smartcontracts/isi/world.rs
```diff
@@ -292,16 +292,19 @@ pub mod isi {
...
     /// Remove all tokens with specified definition id from all accounts in all domains
     fn remove_token_from_accounts(
-        wsv: &WorldStateView,
+        wsv: &mut WorldStateView,
         target_definition_id: &<PermissionTokenDefinition as Identifiable>::Id,
     ) -> Result<(), Error> {
         let mut accounts_with_token = std::collections::HashMap::new();
 
-        for domain in wsv.domains().iter() {
+        let wsv_clone = wsv.clone();
```
appetrosyan, line 300: Cloning the `wsv` is an expensive operation. We need to discuss how to optimise it.

core/src/kura.rs
Erigara: This part would probably be deleted after rebasing.

cli/src/torii/routing.rs
```diff
@@ -250,36 +257,31 @@ async fn handle_post_configuration(
...
         ))
         .await?;
 
-    let mut rx = wsv.subscribe_to_new_block_notifications();
-    stream_blocks(&mut from_height, wsv, &mut stream).await?;
-
     loop {
-        rx.changed().await?;
-        stream_blocks(&mut from_height, wsv, &mut stream).await?;
+        let blocks = sumeragi.blocks_from_height(from_height.try_into().unwrap());
```
Erigara, line 271: Hmm, maybe we should update `from_height` inside loop to return only new blocks?
Currently seems like we would return blocks starting from `from_height` on each cycle.

core/src/block.rs
```diff
@@ -870,10 +846,7 @@ impl From<&CommittedBlock> for Vec<Event> {
             .into(),
         );
 
-        tx.chain(rejected_tx)
-            .chain(invalid_blocks)
```
Erigara, line 874: Why do we exclude `invalid_blocks`?

core/src/block_sync.rs
```diff
@@ -8,75 +8,52 @@ use std::{fmt::Debug, sync::Arc, time::Duration};
...
     broker: Broker,
     actor_channel_capacity: u32,
 }
 
 /// Block synchronizer
-pub trait BlockSynchronizerTrait: Actor + Handler<ContinueSync> + Handler<Message> {
-    /// Requires sumeragi for sending direct messages to it
-    type Sumeragi: SumeragiTrait;
-
+pub trait BlockSynchronizerTrait: Actor + Handler<Message> {
```
Erigara, line 37: What is the purpose of this trait? 
I understand that this is not your change but trait seems to be redundant here.

cli/src/torii/routing.rs
SamHSmith: `from_height` is mutated by `stream_blocks`.

cli/src/lib.rs
```diff
@@ -65,32 +65,68 @@ impl Default for Arguments {
...
-    S: SumeragiTrait<GenesisNetwork = G>,
-    B: BlockSynchronizerTrait<Sumeragi = S>,
-{
+impl Drop for Iroha {
+    fn drop(&mut self) {
+        self.sumeragi.stop_thread();
+    }
+}
+
+use iroha_p2p::network::NetworkBaseRelayOnlinePeers;
```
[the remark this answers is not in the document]
SamHSmith: fixed

cli/src/lib.rs
```diff
@@ -200,60 +236,74 @@ where
         .wrap_err("Unable to start P2P-network")?;
         let network_addr = network.start().await;
 
-        let (events_sender, _) = broadcast::channel(100);
+        let (events_sender, _) = broadcast::channel(10000);
```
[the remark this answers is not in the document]
SamHSmith: Not only that I think we need to split up the event queue into more specific use cases. Currently this is used as a catch all, not very good when performance is concerned.

cli/src/torii/routing.rs
Erigara: oh, right, mutable reference should give me a sign.

data_model/src/trigger/set.rs
SamHSmith: yes

core/src/smartcontracts/isi/domain.rs
```diff
@@ -126,18 +128,22 @@ pub mod isi {
...
             }
 
+            for (key_id, account_id) in array {
+                wsv.modify_account(&account_id, |account_mut| {
+                    if account_mut.remove_asset(&key_id).is_none() {
+                        error!(%key_id, "asset not found - this is a bug");
+                    }
+
+                    Ok(AccountEvent::Asset(AssetEvent::Deleted(key_id)))
+                })?;
```
Erigara, line 144: Why do we need this change?

core/src/smartcontracts/isi/domain.rs
```diff
@@ -218,13 +224,13 @@ pub mod isi {
...
             self,
             _authority: <Account as Identifiable>::Id,
-            wsv: &WorldStateView,
+            wsv: &mut WorldStateView,
         ) -> Result<(), Self::Error> {
             let domain_id = self.object_id;
 
-            wsv.modify_domain(&domain_id.clone(), |domain| {
-                let limits = wsv.config.domain_metadata_limits;
+            let limits = wsv.config.domain_metadata_limits.clone();
```
Erigara, line 231: `Limits` implement `Copy` so should be fine:

```
            let limits = wsv.config.domain_metadata_limits;
```

cli/src/torii/routing.rs
```diff
@@ -250,36 +257,31 @@ async fn handle_post_configuration(
...
+        stream_blocks(&mut from_height, blocks, &mut stream).await?;
+        std::thread::sleep(std::time::Duration::from_millis(10));
     }
+    Ok(())
```
Erigara, line 275: Also this expression is unreachable.

data_model/src/trigger/set.rs
SamHSmith: To be fair I think this has to be overhauled anyway. Once we have merged new-sumeragi my attention will be turned towards the WorldStateView + ISI/Wasm execution + triggers.

data_model/src/trigger/set.rs
SamHSmith: yes, I think I wrote this a while ago before I had gotten used to the rust style.

core/test_network/src/lib.rs
SamHSmith: Comment was not still valid, so I removed the comment.

core/src/wsv.rs
SamHSmith: Because *iterating* mutates the iterator. And we want to use the iterator a second time after the loop has finished.

SamHSmith: It's the rough equivilant of this C code.
```C
int index = 0;
for(; index < len; index++)
{
    if(check_condition(index))
    { break; }
}
memcpy(src + index, dest, len - index);

File: client/tests/integration/triggers/time_trigger.rs [hunk 35]
```diff
@@ -22,9 +22,9 @@ macro_rules! const_assert {
 /// But in general it works well and this test demonstrates it
 #[test]
 #[allow(clippy::cast_precision_loss)]
-fn time_trigger_execution_count_error_should_be_less_than_10_percent() -> Result<()> {
+fn time_trigger_execution_count_error_should_be_less_than_15_percent() -> Result<()> {
```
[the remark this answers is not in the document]
Arjentix: It's my naming originaly. I don't understand what do you mean here. Do you suggest renaming to `time_trigger_execution_count_error_is_less_then_15_percent()`?

core/src/smartcontracts/isi/world.rs
Erigara: I would suggeest the following tweek to avoid clone:
```suggestion
        for domain in wsv.domains().iter() {
            let account_ids = domain.accounts().map(|account| {
                (
                    account.id().clone(),
                    wsv
                        .account_inherent_permission_tokens(account)
                        .filter(|token| token.definition_id() == target_definition_id)
                        .collect::<Vec<_>>(),
                )
            });

            accounts_with_token.extend(account_ids);
        }

        for (account_id, tokens) in accounts_with_token {
            for token in tokens {
                // Most important part move condition out of closure
                if !wsv.remove_account_permission(&account_id, &token) {
                    error!(%token, "token not found - this is a bug");
                }
                wsv.modify_account(&account_id, |account| {
                    let id = account.id();
                    Ok(AccountEvent::PermissionRemoved(id.clone()))
                })?;
            }
        }
```

core/src/wsv.rs
SamHSmith: You are right this should be expect.

client/tests/integration/triggers/time_trigger.rs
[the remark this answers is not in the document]
Erigara: @Arjentix, i mean the following: imo name suggest that we promised that trigger execution errors less than 15%.
When actually we don't make such promise as i understand after discussion with @SamHSmith in one of the earlier PRs.
Correct me if i wrong.

core/src/wsv.rs
SamHSmith: I think this change is less readable, even if it is more compact. The function ought to be rewritted at some point though. I think the whole 'inspect with function' pattern is very harmful.

core/src/wsv.rs
SamHSmith: sure

core/src/wsv.rs
SamHSmith: Yes but not now. This function was created by unifying many 'inspect with closure' calls in order to satisfy the borrow checker. We will have to rewrite this when we move to a patch based way of doing things.

core/src/wsv.rs
SamHSmith: Not worth it for code that it transitory. Also in my experience this type of pattern tends to diverge and so you'll most likely have to remove the macro later. It's not like you have to write the code more than once so the savings by a macro are not large.

core/src/wsv.rs
SamHSmith: I will create the issue.

core/src/smartcontracts/wasm.rs
```diff
@@ -340,12 +344,13 @@ impl<'wrld> Runtime<'wrld> {
         let account_id = caller.data().account_id.clone();
         if let Some(validator) = &mut caller.data_mut().validator {
```
Erigara, line 345: I believe that desctructuring could help with borrow checker and avoid cloning. 

```suggestion
        let State {
            wsv,
            validator,
            ..
        } = caller.data_mut();
```

I have applied suggestion and it seems to compile without errors.

core/src/wsv.rs
SamHSmith: https://github.com/hyperledger/iroha/issues/2757

cli/src/torii/routing.rs
SamHSmith: I think it's fine.

cli/src/torii/tests.rs
SamHSmith: Tests were removed in favor of integration tests.

core/src/wsv.rs
Erigara: I think we should change approach before merging with dev branch.

Using `mem::{take, replace}` could be way to go. 

```rust
// Take triggers
let triggers = core::mem::take(&mut self.world.triggers);
for (event_type, id) in matched_ids {
    // ...
}
// Return triggers onto its place
core::mem::replace(&mut self.world.triggers, triggers);
```

data_model/src/trigger/set.rs
Arjentix: @SamHSmith , what exactly do you think we should revise about triggers?

core/src/wsv.rs
mversic: you are correct, this is definitely UB. `mem::take` or `mem::replace` should be used

mversic: what you end up with is something like the following
```rs
WorldStateView::process_trigger(WorldStateView *restrict self, *dyn ActionTrait action, Event event_type);
```

but you break the contract of the `*restrict` pointer because your two pointers alias (`action` is a pointer into `self`). This is the definition of UB

mversic: I would highly recommend this is fixed before merging

Erigara: Small showcase of what happens ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a7d973f922e6e3e7d06e73edd01e7f12)), you could run `miri` to see that it complains on UB.

core/src/sumeragi/mod.rs
```diff
@@ -53,42 +59,209 @@ trait Consensus {
...
+        use rand::{prelude::SliceRandom, RngCore, SeedableRng};
+
+        let rng = &mut rand::rngs::StdRng::from_entropy();
+        let peers = self
+            .internal
+            .current_online_peers
+            .lock()
+            .expect("lock on online peers for get random peer")
+            .iter()
+            .map(|peer| Peer::new((*peer).clone()))
```
Erigara, line 205:
```suggestion
            .map(|peer| Peer::new(peer.clone()))
```

core/src/sumeragi/mod.rs
```diff
@@ -53,42 +59,209 @@ trait Consensus {
...
+            .collect::<Vec<Peer>>();
+        if peers.is_empty() {
+            return None;
+        } else {
+            let mut sorted_peers = peers;
+            sorted_peers.sort();
+            Some(sorted_peers[rng.next_u32() as usize % sorted_peers.len()].clone())
```
Erigara, line 212: I would suggest to sort `Vec<PeerId>` and only create `Peer` as last step this way we would remove redundant cloning.

core/src/sumeragi/mod.rs
```diff
@@ -53,42 +59,209 @@ trait Consensus {
...
+
+    pub fn stop_thread(&self) {
+        self.internal
+            .sumeragi_state_machine_data
+            .lock()
+            .expect("lock to stop sumeragi thread")
+            .sumeragi_thread_should_exit = true;
+    }
+
+    pub fn update_online_peers(&self, online_peers: Vec<PeerId>) {
```
Erigara, line 245: At some point we could change this to patch update approach, not to replace whole vector every time.

core/src/sumeragi/network_topology.rs
```diff
@@ -168,38 +164,31 @@ impl Builder {
...
+        let since_last_shuffle = n_view_changes % (peers.len() as u64);
         let is_full_circle = since_last_shuffle == 0;
         let sorted_peers = if is_full_circle {
-            sort_peers_by_hash_and_counter(peers, &at_block, n_view_changes as u64)
+            sort_peers_by_hash_and_counter(peers, &at_block, n_view_changes)
```
Erigara, line 183: Small note on `sort_peers_by_hash_and_coundter`:
- allocation could be removed
- sorting seems to be redundant and degrade performance (or what purpose of sorting here?)
- name probably should be `shuffle_peers_by_hash_and_counter`

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+
+    pub fn pipeline_time(&self) -> Duration {
+        self.block_time + self.commit_time
     }
 }
 
-#[async_trait::async_trait]
-impl<G: GenesisNetworkTrait, F: FaultInjection> Handler<CheckCommitTimeout>
-    for SumeragiWithFault<G, F>
+fn block_commit<F>(
```
Erigara, line 272: Why not method? the same question for other functions with signature `fn(&SumeragiWithFault<F>, ...)`

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+                    .wsv
+                    .peers()
+                    .iter()
+                    .map(|peer| peer.id().clone())
+                    .collect(),
+            )
+            .build(0)
+            .expect("Should be able to reconstruct topology from wsv");
+    } else {
+        // We need to perform a round of some form
```
Erigara, line 426: maybe? 

```suggestion
        // We need to perform genesis round
```

Erigara:
`clippy` lints should be fixed before merging into dev

client/benches/tps/lib.rs
Arjentix: Should be possible without `clone()`

client/benches/tps/lib.rs
Arjentix: Previously it was an error... Your comment about *more tx then network can process* makes sense. But I still want to know @s8sato 's  opinion

client/tests/integration/triggers/time_trigger.rs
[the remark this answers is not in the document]
Arjentix: Well, our triggers are *best effort*. The best we can do... How can we rename it then?

client/tests/integration/unstable_network.rs
```diff
@@ -65,7 +65,7 @@ fn unstable_network(
     let (network, mut iroha_client) = rt.block_on(async {
         let mut configuration = Configuration::test();
         configuration.queue.maximum_transactions_in_block = MAXIMUM_TRANSACTIONS_IN_BLOCK;
-        configuration.logger.max_log_level = Level::ERROR.into();
+        configuration.logger.max_log_level = Level::TRACE.into();
```
Arjentix, line 68: Looks like you did it for debugging

core/src/block.rs
Arjentix: I've already reviewed that previously in another PR. I had a discussion with Sam and we decided, that it's okay

core/src/smartcontracts/wasm.rs
Erigara: Yes, it return reference.

core/src/wsv.rs
appetrosyan: UB is unacceptable.

appetrosyan:
To have a quick turnaround, please commit changes directly to https://github.com/appetrosyan/iroha/tree/iroha2-staging

We need this fixed and merged ASAP, even if we need to do some work in `iroha2-dev` later.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+                    sumeragi,
+                    &mut state_machine_guard,
+                    &mut view_change_proof_chain,
+                    current_view_change_index,
+                );
+                instant_at_which_we_should_have_committed += sumeragi.commit_time;
             }
         }
+    }
+}
```
Arjentix, line 1236: I'm unable to read this function. This is the **main** part of the whole *Iroha*. This part should be audited by other companies and programmers.
This is not *ready for lts* state. We need a great refactor here

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
 
-        self.topology.clone()
+fn sumeragi_init_commit_genesis<F>(sumeragi: &SumeragiWithFault<F>, genesis_network: GenesisNetwork)
+where
+    F: FaultInjection,
+{
+    std::thread::sleep(Duration::from_millis(250));
```
Arjentix, line 1242: I think every such sleep should be documented

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+
+    trace!("Start listen for genesis.");
+    let mut state_machine_guard = sumeragi
+        .sumeragi_state_machine_data
+        .lock()
+        .expect("take lock");
+
+    if !state_machine_guard.current_topology.is_consensus_required() {
+        panic!(
+            "How am I supposed to receive a genesis block if I am the only peer in the network?"
```
Arjentix, line 1309: Ok, that's funny

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
-            self.init(last_block, height);
-        } else if let Some(genesis_network) = self.genesis_network.take() {
-            let addr = self.network.clone();
-            if let Err(error) = genesis_network.submit_transactions(self, addr, ctx).await {
-                error!(%error, "Failed to submit genesis transactions")
+        // we must connect to peers so that our block_sync can find us the genesis block.
+        match incoming_message_receiver.try_recv() {
+            Ok(msg) => {
+                match msg {
+                    Message::BlockCommitted(block_committed) => {
```
Arjentix, line 1331:
```suggestion
            Ok(Message::BlockCommitted(block_committed)) => {
```

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+
+                        block_commit(sumeragi, block, &mut state_machine_guard);
+                        info!("Genesis block received and committed.");
+                        return;
+                    }
+                    _ => {
+                        trace!("Not handling message, waiting genesis. : {:?}", msg);
+                    }
+                }
+            }
```
Arjentix, line 1358:
```suggestion
            Ok(_) => {
                trace!("Not handling message, waiting genesis. : {:?}", msg);
            },
```

core/src/sumeragi/message.rs
```diff
@@ -470,47 +151,6 @@ pub struct BlockCommitted {
...
-            &[Role::ValidatingPeer, Role::Leader, Role::ProxyTail],
-            &verified_signatures,
-        );
-        let proxy_tail_signatures =
-            network_topology.filter_signatures_by_roles(&[Role::ProxyTail], &verified_signatures);
-        if valid_signatures.len() >= network_topology.min_votes_for_commit()
-            && proxy_tail_signatures.len() == 1
-            && sumeragi.latest_block_hash() == &self.block.header().previous_block_hash
-        {
-            let mut block = self.block.clone();
```
Arjentix, line 502: Did you just move it to the Sumeragi it-self? Am I right?

cli/src/torii/routing.rs
```diff
@@ -116,15 +119,18 @@ pub(crate) async fn handle_instructions(
...
     query_judge: QueryJudgeArc,
     pagination: Pagination,
     sorting: Sorting,
     request: VerifiedQueryRequest,
 ) -> Result<Scale<VersionedPaginatedQueryResult>> {
-    let (valid_request, filter) = request.validate(&wsv, query_judge.as_ref())?;
-    let original_result = valid_request.execute(&wsv)?;
-    let result = filter.filter(original_result);
+    let (result, filter) = {
+        let wsv = sumeragi.wsv_mutex_access();
```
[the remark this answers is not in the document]
SamHSmith: We could cover those edge cases later. I wanted to start simple.

cli/src/torii/routing.rs
SamHSmith: This has been fixed.

cli/src/lib.rs
```diff
@@ -65,32 +65,68 @@ impl Default for Arguments {
...
-    B: BlockSynchronizerTrait<Sumeragi = S>,
-{
+impl Drop for Iroha {
+    fn drop(&mut self) {
+        self.sumeragi.stop_thread();
+    }
+}
+
+use iroha_p2p::network::NetworkBaseRelayOnlinePeers;
+struct FromNetworkBaseRelay {
```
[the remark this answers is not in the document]
SamHSmith: I didn't know a better place. If we move block synchronization off the actor system it would be used for that as well.

core/src/wsv.rs
SamHSmith: Because mutating the thing which you are inspecting is technically a immutable and mutable borrow at the same time. It works with interior mutability, but that means that you have locks causing it to be a inherently single threaded task. I have no problem with the 'inspect with function' as long as we don't have any interior mutability. That way you can collect a set of changes to make and then apply them after you've inspected. This makes more sense in the context of a patch based world state view.

core/src/wsv.rs
SamHSmith: > WTH is going on in this function?? Why entire implementation from `trigger::Set` is here now?

I found it difficult to understand what was going on when it was split up in many small functions. It can probably be broken up again though.

SamHSmith: As for the UB. I would like to say that the actual operations performed on the world state view are the exact same using the exact same data structures. As far as I know the code doesn't do anything different just because you said one of the references was mut. It enables you to do UB things, but we don't currently. The code is the exact same as if it were 2 immutable references.

core/src/kura.rs
SamHSmith: I have indeed removed this in the rebased branch I gave to aleksandr.

core/src/smartcontracts/isi/world.rs
SamHSmith: I have a plan for removing nearly all the clones. It's a big change though since we would have to change to a patch based model for updating the world state view. But that way you hold a immutable reference to the world state and accumulate patches as you perform modifications. Then at the end you apply all the patches to the world state. You can also add caching to this model for more speedup.

core/src/block.rs
SamHSmith: I removed block invalidation because they serve no purpose until we add partial rolebacks as a handling for softforks. https://github.com/hyperledger/iroha/issues/1170 I have a solution in mind for this, so the invalidated blocks will be added back in later.

core/src/block_sync.rs
SamHSmith: Yes I think we should remove this one as well.

core/src/smartcontracts/isi/domain.rs
SamHSmith: Borrow checker. With a patch based world state view we would be "postponing" all modifications, but checking against the unapplied changes at the same time.

core/src/wsv.rs
```diff
@@ -190,7 +190,7 @@ impl WorldStateView {
     ///
     /// Return a Boolean value indicating whether or not the  [`Account`] already had this permission.
     pub fn add_account_permission(
-        &self,
+        &mut self,
```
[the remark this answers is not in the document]
SamHSmith: Yea, I didn't want to change the implementation in case I broke something. But I have a short term plan for how we can remove interior mutability. It's not that far.

core/src/wsv.rs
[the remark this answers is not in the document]
SamHSmith: Because you can't call async code from sync code without a runtime. For more stability now and more performance later new-sumeragi is manually threaded. This means that it runs outside any tokio runtime and all functions it calls has to be sync.

data_model/src/trigger/set.rs
```diff
@@ -40,19 +40,19 @@ pub struct RepeatsOverflowError;
...
-    by_call_triggers: DashMap<Id, Action<ExecuteTriggerEventFilter>>,
+    pub by_call_triggers: DashMap<Id, Action<ExecuteTriggerEventFilter>>,
     /// Trigger ids with type of events they process
     ids: DashMap<Id, EventType>,
     /// List of actions that should be triggered by events provided by `handle_*` methods.
     /// Vector is used to save the exact triggers order.
     /// Not being cloned
-    matched_ids: RwLock<Vec<(Event, Id)>>,
+    pub matched_ids: Vec<(Event, Id)>,
 }
```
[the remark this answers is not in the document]
SamHSmith: We can make them private again once the implementation has moved back inside. I admit this is quite ugly since it is kind of half way towards the new/better way of doing things, aka with patches.

core/src/sumeragi/fault.rs
```diff
@@ -2,17 +2,18 @@
 //! should be reserved for testing, and only [`NoFault`], should be
 //! used in code.
```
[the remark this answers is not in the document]
SamHSmith: I agree.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+///
+///
+/// sumeragi_state_machine_data is a Mutex instead of a RWLock because
+/// it communicates more clearly the correct use of the lock. The most
+/// frequent action on this lock is the main loop writing to it. This
+/// means that if anyone holds this lock they are blocking the sumeragi
+/// thread. A RWLock will tempt someone to hold a read lock because
+/// they think they are being smart, whilst a Mutex screams *DO NOT
+/// HOLD ME*. That is why the SumeragiStateMachineData is wrapped in
+/// a mutex, it's more self documenting.
```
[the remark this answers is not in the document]
SamHSmith: Most often we need writing. Mutex represents that best. It's a way of not fooling yourself.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+            .finish()
     }
 }
 
-#[async_trait::async_trait]
-impl<G: GenesisNetworkTrait, F: FaultInjection> Actor for SumeragiWithFault<G, F> {
-    fn actor_channel_capacity(&self) -> u32 {
-        self.actor_channel_capacity
-    }
+pub struct SumeragiStateMachineData {
```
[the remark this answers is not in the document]
SamHSmith: Sumeragi is implemented as a state machine. This is it's data.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
 
-    async fn handle(&mut self, ctx: &mut Context<Self>, msg: Message) {
-        trace!(peer_role=?self.topology.role(&self.peer_id), ?msg);
-        if let Err(error) = msg.handle(self, ctx).await {
-            error!(%error, "Failed to handle message");
+    /// Updates network topology by taking the actual list of peers from `WorldStateView`.
+    /// Updates it only if there is a change in WSV peers, otherwise leaves the order unchanged.
+    #[allow(clippy::expect_used)]
+    pub fn update_network_topology(topology: &mut Topology, wsv: &WorldStateView) {
+        let wsv_peers: HashSet<_> = wsv.trusted_peers_ids().clone().into_iter().collect();
```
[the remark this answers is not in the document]
SamHSmith: probably.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
-#[async_trait::async_trait]
-impl<G: GenesisNetworkTrait, F: FaultInjection> Handler<IsLeader> for SumeragiWithFault<G, F> {
-    type Result = bool;
 
-    async fn handle(&mut self, IsLeader: IsLeader) -> Self::Result {
-        self.is_leader()
+    // Update WSV copy that is public facing
+    {
+        let mut wsv_for_public_use_guard = sumeragi.wsv_for_public_use.lock().unwrap();
+        *wsv_for_public_use_guard = WorldStateView::clone(&state_machine.wsv);
```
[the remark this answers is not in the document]
SamHSmith: yup.

core/src/sumeragi/fault.rs
SamHSmith: To me it's about frame of mind. This function is not associated with some object. It just does stuff, in fact it does things to all the things passed in. It commits the block, it communicates with the outside world through `sumeragi` and it modifies the state machine data. All of these three are equal participants and so to me it doesn't make sense to make this function method when it could be a method on all three.

core/src/sumeragi/fault.rs
SamHSmith: I said of some form because the genesis round can have already passed and you just recieve the block.

core/src/smartcontracts/wasm.rs
```diff
@@ -130,29 +130,34 @@ impl Validator<'_> {
...
 }
 
 struct State<'wrld> {
     account_id: AccountId,
     /// Ensures smartcontract adheres to limits
-    validator: Option<Validator<'wrld>>,
+    validator: Option<Validator>,
     store_limits: StoreLimits,
-    wsv: &'wrld WorldStateView,
+    wsv: &'wrld mut WorldStateView,
```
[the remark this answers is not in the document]
SamHSmith: As I have said in other places. &mut is a steping stone to patches. & with internal mutability -> &mut -> & with patches.

core/test_network/src/lib.rs
```diff
@@ -55,17 +55,11 @@ macro_rules! prepare_test_for_nextest {
...
 
 /// Network of peers
-pub struct Network<G = GenesisNetwork, S = Sumeragi<G>, B = BlockSynchronizer<S>>
-where
-    G: GenesisNetworkTrait,
-
-    S: SumeragiTrait<GenesisNetwork = G>,
-    B: BlockSynchronizerTrait<Sumeragi = S>,
-{
+pub struct Network {
```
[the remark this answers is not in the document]
SamHSmith: yes

client/benches/tps/config.json
```diff
@@ -1,7 +1,7 @@
 {
     "peers": 4,
-    "interval_us_per_tx": 80000,
-    "max_txs_per_block": 8192,
-    "blocks": 5,
+    "interval_us_per_tx": 10000,
+    "max_txs_per_block": 512,
+    "blocks": 15,
```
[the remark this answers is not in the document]
SamHSmith: The low tx's per block does help the old implementation as well. You are right though, it is important both are testing with the same parameters.

cli/src/torii/routing.rs
```diff
@@ -380,81 +382,57 @@ mod subscription {
...
-    update_metrics(&wsv, network).await?;
-    let status = Status::from(&wsv.metrics);
+async fn handle_status(sumeragi: Arc<Sumeragi>, network: Addr<IrohaNetwork>) -> Result<Json> {
+    sumeragi.update_metrics(network);
+    let status = Status::from(&sumeragi.wsv_mutex_access().metrics);
     Ok(reply::json(&status))
 }
 
-#[cfg(feature = "telemetry")]
-async fn update_metrics(wsv: &WorldStateView, network: Addr<IrohaNetwork>) -> Result<()> {
```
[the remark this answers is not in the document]
SamHSmith: It was moved into sumeragi for data ownership reasons.

cli/src/torii/routing.rs
```diff
@@ -380,81 +382,57 @@ mod subscription {
...
     /// Construct `Torii` from `ToriiConfiguration`.
     pub fn from_configuration(
         iroha_cfg: Configuration,
-        wsv: Arc<WorldStateView>,
         queue: Arc<Queue>,
         query_judge: QueryJudgeArc,
         events: EventsSender,
         network: Addr<IrohaNetwork>,
         notify_shutdown: Arc<Notify>,
+        sumeragi: Arc<Sumeragi>,
```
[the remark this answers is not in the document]
SamHSmith: An arc is a reference though. A reference we need in order to be able to call functions on it.

core/src/block_sync.rs
```diff
@@ -85,253 +62,174 @@ impl<S: SumeragiTrait> BlockSynchronizerTrait for BlockSynchronizer<S> {
...
-                .await;
-        } else {
-            warn!(block_hash = %this_block.hash(), "Failed to commit a block received via synchronization request - validation failed");
-            self.state = State::Idle;
-        }
-    }
 }
 
-/// The module for block synchronization related peer to peer messages.
-pub mod message {
```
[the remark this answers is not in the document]
SamHSmith: It was causing some weird issues. I genuinly couldn't get it to build until I removed it.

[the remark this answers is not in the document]
SamHSmith: Could probably be added back in though.

core/src/queue.rs
```diff
@@ -192,15 +199,45 @@ impl Queue {
...
+            };
+            if self.check_tx(entry.get(), wsv).is_err() {
+                entry.remove_entry();
+                continue;
+            }
+
+            if *entry
+                .get()
+                .check_signature_condition(wsv)
                 .expect("Checked in `check_tx` just above")
```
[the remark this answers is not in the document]
SamHSmith: I think the code requires it. I would be for changing that though. I was just trying to get stuff done when I wrote this.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
-}
 
-#[async_trait::async_trait]
-impl<G: GenesisNetworkTrait, F: FaultInjection> Handler<CheckCreationTimeout>
-    for SumeragiWithFault<G, F>
-{
-    type Result = ();
+    pub fn wait_for_peers(&self, network_topology: &Topology) {
+        iroha_logger::info!("Waiting for active peers",);
+        for i in 0..120 {
```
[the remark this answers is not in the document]
SamHSmith: sure? Seems pretty trivial though.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+        transaction_cache.truncate(write_index);
     }
 }
 
-#[async_trait::async_trait]
-impl<G: GenesisNetworkTrait, F: FaultInjection> Handler<NetworkMessage>
-    for SumeragiWithFault<G, F>
+fn request_view_change<F>(
+    sumeragi: &SumeragiWithFault<F>,
+    state_machine_guard: &mut SumeragiStateMachineData,
```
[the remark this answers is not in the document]
SamHSmith: You are right. This has been changed.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
+            .with_peers(
+                state_machine_guard
+                    .wsv
+                    .peers()
+                    .iter()
+                    .map(|peer| peer.id().clone())
+                    .collect(),
+            )
+            .build(0)
+            .expect("Should be able to reconstruct topology from wsv");
```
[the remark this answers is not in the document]
SamHSmith: How does it help to have more functions if it only increases indirection? You've read long books before, if things read top to bottom they are actually very easy to navigate. And what happens before what correlates with location in the file.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
-            self.topology = self.topology
-                .clone()
-                .into_builder()
-                .with_peers(wsv_peers)
-                .build()
-                // TODO: Check it during instruction execution.
-                .expect("The safety of changing the number of peers should have been checked at Instruction execution stage.");
-        }
-    }
+    // do normal rounds
```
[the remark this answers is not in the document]
SamHSmith: Aren't the purpose of comments to inform you about the code? Plus if it's a function you loose information because you don't know where it is called from. Here it is obvious.

core/src/sumeragi/fault.rs
```diff
@@ -39,865 +37,1333 @@ impl FaultInjection for NoFault {
...
-                PendingBlock::new(transactions, Vec::new())
-                    .chain_first_with_genesis_topology(genesis_topology),
-                ctx,
-            )
-            .await
-        }
-    }
+    let mut view_change_proof_chain = Vec::new();
+    let mut old_view_change_index = 0;
+    let mut old_latest_block_height = 0;
```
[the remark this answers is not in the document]
SamHSmith: The problem is the problem. I can't wish it to go away. There is mutable state in the solution to this problem, I can't change that.

core/src/sumeragi/fault.rs
SamHSmith: It is the main part. And it's only 1000 lines, that's remarkably small in my opinion. Adding more indirection and increasing sloc count doesn't make it easier to understand in my opinion. I wrote this code following principles laid out here: http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html

core/src/wsv.rs
Erigara: Problem with this code is that it violates some compiler invariants and because of this it is UB.

core/src/sumeragi/message.rs
SamHSmith: Yes

client/tests/integration/triggers/time_trigger.rs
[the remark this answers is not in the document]
Erigara: Not wroth to spend time on this right now.

core/src/smartcontracts/isi/world.rs
Erigara: resolved with rollback

core/src/smartcontracts/isi/account.rs
```diff
@@ -190,49 +194,63 @@ pub mod isi {
...
             let account_id = self.destination_id;
             let permission = self.object;
 
+            let wsv_clone = wsv.clone();
+
+            let did_remove = wsv.remove_account_permission(&account_id, &permission);
+
             wsv.modify_account(&account_id, |account| {
-                if !wsv
+                if !wsv_clone
```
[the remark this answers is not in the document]
Erigara: Resolved with return to interior mutability.

client/benches/tps/lib.rs
s8sato: I think we can be flexible with this line. Returning error always detects #2078 bug, but on the other hand, it makes the benchmark unmeasurable as long as this bug exists

core/src/wsv.rs
Erigara: Resolved with rollback.

core/src/wsv.rs
Erigara: Resolved with rollback.

core/src/block_sync.rs
Erigara: Removed

appetrosyan:
Closing as the actual merge is happening in #2762

PR #2715, opened 2022-09-08
[feature] #2502: Store GIT hash of build

### LINKED ISSUE #2502: [suggestion] Storing a GIT hash while building Iroha for versioning
### Feature request

Currently, issue templates require people to write GIT commit hash so it's easier to understand what may be broken.
@arjentix pointed an inconvenience with those fields: Docker builds do not provide that information and it's only possible to write an approximation or "not available".

I don't want to remove debug information and I'm thinking of a command-line argument to get version info, including the GIT commit. It's not too hard to implement in Cargo, because we can define [compile-time environment variables](https://stackoverflow.com/a/44407625/703462).

### Motivation

Having a GIT hash for any builds would allow to quickly get it in Docker with a command like this: `docker exec iroha_iroha1_1 /iroha --version`. The exact commit would then help to make debugging quicker.

### Who can help?

@s8sato @appetrosyan

Erigara:

### Description of the Change

Add `build.rs` to extract git hash of build.

### Issue

Closes #2502.

### Benefits

Easier to debug.

### Possible Drawbacks

None.

### Usage Examples or Tests

```
cargo run --bin iroha_client_cli  -- --version
cargo run --bin iroha -- --version
```

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -15,6 +15,11 @@ async fn main() -> Result<(), color_eyre::Report> {
         return Ok(());
     }
 
+    if std::env::args().any(|a| is_version(&a)) {
+        print_version();
+        return Ok(());
+    }
+
     if std::env::args().any(|a| is_submit(&a)) {
         args.submit_genesis = true;
         if let Ok(genesis_path) = std::env::var("IROHA2_GENESIS_PATH") {
@@ -68,11 +73,16 @@ fn is_submit(arg: &str) -> bool {
     ["--submit-genesis", "-s"].contains(&arg)
 }
 
+fn is_version(arg: &str) -> bool {
+    ["--version", "-V"].contains(&arg)
+}
+
 #[allow(clippy::print_stdout)]
 fn print_help() {
     println!("Iroha 2");
     println!("pass `--help` or `-h` for this message");
     println!("pass `--submit-genesis` or `-s` to submit genesis from this peer");
+    println!("pass `--version` or `-V` to print version information");
     println!();
     println!("Iroha 2 is configured via environment variables:");
     println!("    IROHA2_CONFIG_PATH is the location of your `config.json`");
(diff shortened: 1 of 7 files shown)


Erigara:
> I suppose you forget to add "build" = "build.rs" to Cargo.toml files, no?

hmm, everything works fine without it, but i can add it, no problem

pesterev:
> > I suppose you forget to add "build" = "build.rs" to Cargo.toml files, no?
> 
> hmm, everything works fine without it, but i can add it, no problem

Okay, I've just checked doc, it seems `build = "build.rs"` now by default starting from some version.

pesterev (approved):

cli/src/main.rs
```diff
@@ -92,3 +102,12 @@ as follows:"
     println!("    IROHA_GENESIS is the genesis block config");
     println!("Examples of these variables can be found in the default `configs/peer/config.json`.")
 }
+
+#[allow(clippy::print_stdout)]
+fn print_version() {
+    println!(
+        "iroha2 {} ({})",
```
QuentinI, line 109:
```suggestion
        "iroha {} ({})",
```

Version will already include 2, so maybe just "iroha"?

QuentinI (commented):
Would also make sense to output enabled cargo features.

cli/src/main.rs
appetrosyan: Very good point

pesterev (approved):

client_cli/Cargo.toml
```diff
@@ -26,3 +26,7 @@ color-eyre = "0.6.2"
 clap = { version = "3.2.16", features = ["derive"] }
 dialoguer = { version = "0.10.2", default-features = false }
 serde_json = "1.0.83"
+
+[build-dependencies]
+anyhow = "1.0.60"
+vergen = { version = "7.3.2", default-features = false, features = ["git"] }
```
mversic, line 32: this is the first time I see `vergen` used. Could we use it in other build scripts?

QuentinI (approved):

client_cli/Cargo.toml
Erigara: @mversic, actually we already use `vergen` in `telemetry` (`telemetry/build.rs`).

mversic: not very relevant, but do you think `client/build.rs` would benefit from it?

Erigara: currently we are using `client/build.rs` for building smartcontracts, where do you want to include version inside `client`?

PR #2724, opened 2022-09-11
[fix] #2703: Fix orillion dev env bugs (dev)

### LINKED ISSUE #2703: Can't register trigger in Orillion dev-environment
@arndey reports that he is unable to register trigger on Orillion dev-environment.
Probably to debug it we will need acceess to this environment.
There is also a problem debugging it using mac M1, because of #2690

Some logs from Iroha

<details>
<summary>logs</summary>

```
2022-09-05T13:59:52.228172Z DEBUG request{method=POST path=/transaction version=HTTP/1.1 remote.addr=10.1.3.20:60700}: warp::filters::trace: received request
2022-09-05T13:59:52.228329Z  INFO request{method=POST path=/transaction version=HTTP/1.1 remote.addr=10.1.3.20:60700}: warp::filters::trace: processing request
2022-09-05T13:59:52.231334Z DEBUG hyper::proto::h1::conn: incoming body completed
2022-09-05T13:59:52.234042Z  INFO request{method=POST path=/transaction version=HTTP/1.1 remote.addr=10.1.3.20:60700}: warp::filters::trace: finished processing with success status=200
2022-09-05T13:59:52.234287Z DEBUG hyper::proto::h1::io: flushed 75 bytes
2022-09-05T13:59:52.236712Z DEBUG hyper::proto::h1::conn: read eof
2022-09-05T13:59:52.407345Z  INFO iroha_core::sumeragi::fault: Forwarding tx to leader peer_addr=0.0.0.0:1337 peer_role=ObservingPeer leader_addr=orillion-iroha2-peer-1.orillion-dev.svc.cluster.local:1337 tx_hash=8794a2133054ed3443228b202c0fbb5db774872c58d67d5e5d84660835baffbe
2022-09-05T13:59:52.441277Z DEBUG tracing_bunyan_formatter::formatting_layer: msg is a reserved field in the bunyan log format. Skipping it.
2022-09-05T13:59:52.494603Z DEBUG tracing_bunyan_formatter::formatting_layer: msg is a reserved field in the bunyan log format. Skipping it.
2022-09-05T13:59:52.532188Z DEBUG tracing_bunyan_formatter::formatting_layer: msg is a reserved field in the bunyan log format. Skipping it.
2022-09-05T13:59:53.845552Z ERROR iroha::torii::routing: Failed to subscribe someone error=Event receiving error: channel lagged by 142
2022-09-05T13:59:53.845831Z DEBUG hyper::proto::h1::io: parsed 4 headers
2022-09-05T13:59:53.845874Z DEBUG hyper::proto::h1::conn: incoming body is empty
```

</details>

# Who can help?
@Arjentix , @arndey

Arjentix:
### Description of the Change

- Add TLS support for websockets
- Fix closing WebSocket bug

See also #2728   -- lts version of this PR

### Issue

* Closes #2703 
* Opens #2725
* Opens #2726 
* Opens #2727

### Benefits

Two bugs fixed and no more annoying `ERROR` log in Iroha saying about wrong web-socket closing

### Possible Drawbacks

I had to remove timeout in client when listening to events. To bring it back properly we need more work to be done

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -32,7 +32,7 @@ http = "0.2.8"
 rand = "0.8.5"
 serde = { version = "1.0.142", features = ["derive"] }
 serde_json = "1.0.83"
-tungstenite = "0.16"
+tungstenite = { version = "0.16", features = ["native-tls"] }
 base64 = "0.13.0"
 thiserror = "1.0.32"
 derive_more = "0.99.17"
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -37,6 +37,7 @@ base64 = "0.13.0"
 thiserror = "1.0.32"
 derive_more = "0.99.17"
 parity-scale-codec = { version = "3.1.5", default-features = false, features = ["derive"] }
+crossbeam = "0.8.2"
 
 
 [dev-dependencies]
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -443,44 +443,64 @@ impl Client {
         &self,
         transaction: Transaction,
     ) -> Result<HashOf<VersionedTransaction>> {
-        struct EventListenerInitialized;
-
         let client = self.clone();
-        let (event_sender, event_receiver) = mpsc::channel();
         let (init_sender, init_receiver) = mpsc::channel();
         let hash = transaction.hash();
-        let _handle = thread::spawn(move || -> eyre::Result<()> {
-            let event_iterator = client
-                .listen_for_events(PipelineEventFilter::new().hash(hash.into()).into())
-                .wrap_err("Failed to establish event listener connection.")?;
-            init_sender
-                .send(EventListenerInitialized)
-                .wrap_err("Failed to send init message through init channel.")?;
-            for event in event_iterator.flatten() {
-                if let Event::Pipeline(this_event) = event {
-                    match this_event.status {
-                        PipelineStatus::Validating => {}
-                        PipelineStatus::Rejected(reason) => event_sender
-                            .send(Err(reason))
-                            .wrap_err("Failed to send the transaction through event channel.")?,
-                        PipelineStatus::Committed => event_sender
-                            .send(Ok(hash.transmute()))
-                            .wrap_err("Failed to send the transaction through event channel.")?,
+
+        // TODO: use `std::thread::scope()` after update to Rust 1.63
+        let scope_res = crossbeam::scope(|spawner| {
+            let submitter_handle = spawner.spawn(move |_| -> Result<()> {
+                init_receiver
+                    .recv()
+                    .wrap_err("Failed to receive init message.")?;
+
+                self.submit_transaction(transaction)?;
+                Ok(())
+            });
+
+            let confirmation_res = client.listen_for_tx_confirmation(hash, &init_sender);
+
+            match submitter_handle.join() {
+                Ok(Ok(())) => confirmation_res,
+                Ok(Err(e)) => Err(e).wrap_err("Transaction submitter thread exited with error"),
+                Err(_) => Err(eyre!("Transaction submitter thread panicked")),
+            }
+        });
+
+        match scope_res {
+            Ok(res) => res,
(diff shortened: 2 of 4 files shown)


client/Cargo.toml
```diff
@@ -32,11 +32,12 @@ http = "0.2.8"
 rand = "0.8.5"
 serde = { version = "1.0.142", features = ["derive"] }
 serde_json = "1.0.83"
-tungstenite = "0.16"
+tungstenite = { version = "0.16", features = ["native-tls"] }
```
Arjentix, line 35: Without that tungstenite won't work with wss protocol at all.
The other option is to use rust-tls.
I don't know which is better, but native-tls is more popular and, I believe, should be faster

client/src/client.rs
```diff
@@ -443,44 +443,64 @@ impl Client {
         &self,
         transaction: Transaction,
     ) -> Result<HashOf<VersionedTransaction>> {
-        struct EventListenerInitialized;
-
         let client = self.clone();
-        let (event_sender, event_receiver) = mpsc::channel();
         let (init_sender, init_receiver) = mpsc::channel();
         let hash = transaction.hash();
-        let _handle = thread::spawn(move || -> eyre::Result<()> {
```
Arjentix, line 452: Previously this thread was detached and values weren't being dropped.
WebSocket closing handshake is done inside event_iterator Drop impl. So connection was never closed properly

client/Cargo.toml
appetrosyan: We already have `native-tls` enabled, (thanks to workspaces resolving features in a weird way), but making it explicit helps.

client/src/client.rs
appetrosyan: #1899 needs to be mentioned.

client/src/client.rs
```diff
@@ -443,44 +443,64 @@ impl Client {
...
+            match submitter_handle.join() {
+                Ok(Ok(())) => confirmation_res,
+                Ok(Err(e)) => Err(e).wrap_err("Transaction submitter thread exited with error"),
+                Err(_) => Err(eyre!("Transaction submitter thread panicked")),
+            }
+        });
+
+        match scope_res {
+            Ok(res) => res,
+            Err(_) => Err(eyre!("Transaction submitter thread panicked")),
```
appetrosyan, line 472: Why not `map` and `wrap_err`? This way we'd have a traceback for the panic.

client/src/client.rs
```diff
@@ -443,44 +443,64 @@ impl Client {
...
+                self.submit_transaction(transaction)?;
+                Ok(())
+            });
+
+            let confirmation_res = client.listen_for_tx_confirmation(hash, &init_sender);
+
+            match submitter_handle.join() {
+                Ok(Ok(())) => confirmation_res,
+                Ok(Err(e)) => Err(e).wrap_err("Transaction submitter thread exited with error"),
+                Err(_) => Err(eyre!("Transaction submitter thread panicked")),
```
appetrosyan, line 466: Why discard the error? Why not include it in `wrap_err`.

client/Cargo.toml
Arjentix: Do you mean `iroha2-dev` specifically? Because it didn't work on lts branch until I enabled this feature

client/src/client.rs
Arjentix: Yes, but this is another kind of error. This is the error indicating thread panic, and it does not implement `Error` trait, so `map` and `wrap_err` don't  work. All we know about this error is that it's `Box<dyn Any + Send + 'static>`. See [here](https://doc.rust-lang.org/std/thread/type.Result.html)

client/src/client.rs
Arjentix: Same answer as above

appetrosyan (approved):

client/Cargo.toml
appetrosyan: Nope. Linux x86.

s8sato (approved):

PR #2762, opened 2022-09-20
[feat] #2713: New sumeragi into `iroha2-dev`

### LINKED ISSUE #2713: Fix lints on staging branch
We may also need to discuss some changes to `cargo lints`

appetrosyan:
### Description of the Change

- [x] Updated the new consensus. 
- [x] Fixed lints
- [ ] Ran tests
- [ ] Test flakyness documented
- [ ] Re-reviewed changes to `iroha2-dev`

### Issue

Closes #2713 
### Possible Drawbacks

Readability and maintainability.


core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
-        }
-        let txs = self.queue.get_transactions_for_block();
-        // TODO: This should properly process triggers
-        let event_recommendations = Vec::new();
-        if let Err(error) = self.round(txs, event_recommendations, ctx).await {
-            error!(%error, "Round failed");
+    /// Updates network topology by taking the actual list of peers from `WorldStateView`.
+    /// Updates it only if there is a change in WSV peers, otherwise leaves the order unchanged.
+    #[allow(clippy::expect_used)]
+    pub fn update_network_topology(topology: &mut Topology, wsv: &WorldStateView) {
```
Erigara, line 169: Could be removed, seems to pop up after merge.

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
-    async fn on_start(&mut self, ctx: &mut Context<Self>) {
-        self.broker.subscribe::<Init, _>(ctx);
-        self.broker.subscribe::<Message, _>(ctx);
-        self.broker.subscribe::<CommitBlock, _>(ctx);
-        self.broker.subscribe::<NetworkMessage, _>(ctx);
-    }
+// TODO: In general naming things after the programming patterns is
+// considered a bad practice. We need a better name.
+/// Internal structure that retains the state.
+pub struct SumeragiStateMachineData {
```
Erigara, line 113: `SumeragiState`?

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
+    F: FaultInjection,
+{
+    let suspect_proof = {
+        let mut proof = Proof {
+            latest_block_hash: state_machine.latest_block_hash,
+            view_change_index: current_view_change_index,
+            signatures: Vec::new(),
+        };
+        proof
+            .sign(sumeragi.key_pair.clone())
```
Erigara, line 360: Just note for further improvement: could be chained is `sign` would return `Proof`.

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
+            .expect("Suspect proof must be able to perform signing");
+        proof
+    };
+
+    let _ = view_change_proof_chain.insert_proof(
```
Erigara, line 365: We are ignoring error here and then broadcast probably wrong `Vec<Proof>` to other peers, is it ok?

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
-    /// Get peers as a hash set of their ids.
-    pub fn peers(&self) -> HashSet<PeerId> {
-        self.topology.sorted_peers().iter().cloned().collect()
-    }
+#[instrument(skip(sumeragi, state_machine_guard))]
+#[allow(clippy::expect_used)]
+/// Execute the main loop of [`SumeragiWithFault`]
+pub fn run_sumeragi_main_loop<F>(
+    sumeragi: &SumeragiWithFault<F>,
+    mut state_machine_guard: SumeragiStateMachineData,
```
Erigara, line 391: Not guard anymore.

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
+        {
+            // Normal startup
+            // Aka, we don't have to do anything.
         } else {
-            self.validate_and_publish_created_block(
-                PendingBlock::new(transactions, Vec::new())
-                    .chain_first_with_genesis_topology(genesis_topology),
-                ctx,
+            // We don't have the genesis block.
+            // We need to perform a round of some form.
```
Erigara, line 410: Does this comment true? because seems that we actually have genesis block

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
-                "Block height should be 0 for genesis round. But it is: {}",
-                self.block_height
-            ))
+    // Initialization code
+    {
+        if state_machine_guard.latest_block_height != 0
+            && state_machine_guard.latest_block_hash != Hash::zeroed().typed()
+        {
+            // Normal startup
+            // Aka, we don't have to do anything.
```
Erigara, line 407: Empty branch seems a bit strange, i suppose we could just write comment `Otherwise do nothing (normal startup)`.

core/src/sumeragi/fault.rs
```diff
@@ -38,866 +37,1337 @@ impl FaultInjection for NoFault {
...
-                let count = proof_votes.signatures().len();
-                proof_votes.merge_signatures(proof.signatures().clone());
-
-                if proof_votes.signatures().len() > count {
-                    (true, proof_votes.clone())
-                } else {
-                    (false, proof_votes.clone())
+        {
+            update_transaction_cache(sumeragi, &mut state_machine_guard);
+            fn update_transaction_cache<F>(
```
Erigara, line 475: I believe this is merge conflict.

core/src/sumeragi/fault.rs
SamHSmith: Originally I didn't have a return type because the code can't 'fail'. Aleksandr wanted to have a return type to know what was going on inside. `insert_proof` is written such that you can insert any/all possible proofs and only the correct ones will stick. It internally combines the proof with the existing proofs where possible and removes redundant signatures et cetera.

core/src/sumeragi/fault.rs
SamHSmith: Yea, I didn't rename it yet because I was making other changes.

core/src/sumeragi/fault.rs
SamHSmith: Maybe it's more accurate to say we don't have a committed genesis block.

core/src/sumeragi/fault.rs
appetrosyan: I'd keep it.

core/src/sumeragi/fault.rs
appetrosyan: `State` is enough. It's no more a state machine than other things we don't have a state machine inside of. Moreover, now the fully qualified path is `sumeragi::main_loop::State`

core/src/sumeragi/fault.rs
appetrosyan: Created #2767

core/src/sumeragi/fault.rs
appetrosyan: I.e. we need to check for the error message, and hope that the designers of the language were not idiots and marked the error path as the cold path.

core/src/sumeragi/fault.rs
appetrosyan: Removed. 

Comment is not accurate and not necessary. 

Comments are **only** there to explain things which are time consuming to find out without them.

core/src/sumeragi/fault.rs
appetrosyan: Another anti-pattern is scope-around-scope. A function call cannot leak internal state. If you don't have a top-level let-binding the scope does nothing.

appetrosyan: Also, this is the one case, where a short-circuiting `if` statement is the preferred solution.

core/src/sumeragi/fault.rs
Erigara: But why? it's now `Topology` method, this function don't called anywhere, you could check this by removing `pub`.

core/src/smartcontracts/isi/domain.rs
```diff
@@ -126,18 +128,22 @@ pub mod isi {
...
+            for (key_id, account_id) in array {
+                wsv.modify_account(&account_id, |account_mut| {
+                    if account_mut.remove_asset(&key_id).is_none() {
+                        error!(%key_id, "asset not found - this is a bug");
+                    }
+
+                    Ok(AccountEvent::Asset(AssetEvent::Deleted(key_id)))
+                })?;
+            }
+
```
Arjentix, line 146: Can be moved back to the first loop as it was before

core/src/wsv.rs
```diff
@@ -454,15 +447,20 @@ impl WorldStateView {
...
+    /// Returns a vector of blockchain blocks after the block with the given `hash`
     pub fn blocks_after_hash(
         &self,
         hash: HashOf<VersionedCommittedBlock>,
-    ) -> impl Iterator<Item = VersionedCommittedBlock> + '_ {
-        self.blocks
-            .iter()
-            .skip_while(move |block_entry| block_entry.value().header().previous_block_hash != hash)
-            .map(|block_entry| block_entry.value().clone())
+    ) -> Vec<VersionedCommittedBlock> {
```
Arjentix, line 454: I like `Iterator` approach more. Why changed?

core/src/wsv.rs
```diff
@@ -499,14 +497,12 @@ impl WorldStateView {
     }
 
     /// Returns iterator over blockchain blocks starting with the block of the given `height`
-    pub fn blocks_from_height(
-        &self,
-        height: usize,
-    ) -> impl Iterator<Item = VersionedCommittedBlock> + '_ {
+    pub fn blocks_from_height(&self, height: usize) -> Vec<VersionedCommittedBlock> {
```
Arjentix, line 500: Same question. Also outdated doc comment

crypto/src/signature.rs
```diff
@@ -67,7 +67,7 @@ impl Signature {
     /// # Errors
     /// Fails if signing fails
     #[cfg(feature = "std")]
-    fn new(key_pair: KeyPair, payload: &[u8]) -> Result<Self, Error> {
+    pub fn new(key_pair: KeyPair, payload: &[u8]) -> Result<Self, Error> {
```
Arjentix, line 70: I still think that we can do things without this change

appetrosyan: Hiding a constructor method usually means that the constructor is unnecessary, and that the type should have been private.

core/src/wsv.rs
appetrosyan: Reverted

core/src/sumeragi/main_loop.rs
```diff
@@ -0,0 +1,1340 @@
...
+    old_latest_block_height: &mut u64,
+    // below is the state that gets reset.
+    current_topology: &mut Topology,
+    voting_block_option: &mut Option<VotingBlock>,
+    block_signature_acc: &mut Vec<(
+        HashOf<VersionedValidBlock>,
+        SignatureOf<VersionedValidBlock>,
+    )>,
+    has_sent_transactions: &mut bool,
+    instant_when_we_should_create_a_block: &mut Instant,
```
Arjentix, line 530: 😧

core/src/sumeragi/main_loop.rs
```diff
@@ -0,0 +1,1340 @@
...
+    /// subsystem.
+    ///
+    /// This transaction cache was therefore introduced so that only
+    /// interact with the transaction queue as necessary. If this were
+    /// written in C it would have just been an array of
+    /// `VersionedAcceptedTransaction`, but because of Rust's RAII
+    /// enforcement it has to be an array of options.  Otherwise we
+    /// would need to use unsafe to implement the fast pruning on the
+    /// array.
+    pub transaction_cache: Vec<Option<VersionedAcceptedTransaction>>,
```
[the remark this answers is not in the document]
Arjentix: Especially then AFAIK another good approaches without `Option` and `unsafe` were disscussed

core/src/sumeragi/main_loop.rs
Arjentix: Okay, I understand, that it's just the best we can do so simplify the main loop for now.

But I don't think it's okay for the future

appetrosyan: So, there's two things, on the one hand, you kinda have to concede that the new approach while not conducive to modification (which should happen rarely), is conducive to debugging (which might happen often, since we need to rule out consensus first in most cases). 

Normally I wouldn't have accepted this change at all, but we can work our way around it. Given that I or @SamHSmith have no problem navigating the main loop and no-one else is supposed to anyway, I'm all for earmarking a refactor, but doing it in October/November.

mversic (changes requested):

docs/source/iroha_2_whitepaper.md
```diff
@@ -1,4 +1,4 @@
-# Iroha v2.0
```
s8sato, line 1: I want to conclude that I'm not qualified to be a reviewer for now. Let me decline my review. I'd like to contribute by adding tests later

cli/src/torii/routing.rs
```diff
@@ -254,36 +261,34 @@ async fn handle_post_configuration(
...
-
     loop {
-        rx.changed().await?;
-        stream_blocks(&mut from_height, wsv, &mut stream).await?;
+        let blocks = sumeragi.blocks_from_height(from_height.try_into().wrap_err("Failed to convert `from_height` into `usize`. You should consider upgrading to a 64-bit arch")?);
+        if blocks.is_empty() {
+            break;
+        }
+        stream_blocks(&mut from_height, blocks, &mut stream).await?;
+        std::thread::sleep(std::time::Duration::from_millis(10));
```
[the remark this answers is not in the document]
appetrosyan: Ideally yes. But this can be done in a separate pass.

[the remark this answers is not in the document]
appetrosyan: #2784

cli/src/torii/routing.rs
```diff
@@ -254,36 +261,34 @@ async fn handle_post_configuration(
...
     }
+    Ok(())
 }
 
 async fn stream_blocks(
     from_height: &mut u64,
-    wsv: &WorldStateView,
+    blocks: Vec<VersionedCommittedBlock>,
```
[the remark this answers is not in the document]
appetrosyan: Being explicit about the type has advantages in terms of compilation time and code size. Unless we need to, use explicit type.

cli/src/torii/routing.rs
```diff
@@ -384,81 +389,65 @@ mod subscription {
...
 }
 
 #[cfg(feature = "telemetry")]
-async fn handle_status(wsv: Arc<WorldStateView>, network: Addr<IrohaNetwork>) -> Result<Json> {
-    update_metrics(&wsv, network).await?;
-    let status = Status::from(&wsv.metrics);
-    Ok(reply::json(&status))
+#[allow(clippy::unused_async)] // TODO: remove?
+async fn handle_metrics(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -> Result<String> {
+    // TODO: Remove network.
```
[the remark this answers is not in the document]
appetrosyan: No. 

Network is an unused argument, and `async` is unused. We can fix this after the `iroha2-dev` merge.

cli/src/torii/routing.rs
```diff
@@ -384,81 +389,65 @@ mod subscription {
...
-    for domain in domains {
-        wsv.metrics
-            .accounts
-            .get_metric_with_label_values(&[domain.id().name.as_ref()])
-            .wrap_err("Failed to compose domains")
-            .map_err(Error::Prometheus)?
-            .set(domain.accounts().len() as u64);
+#[allow(clippy::unused_async)] // TODO: remove?
+async fn handle_status(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -> Result<Json> {
+    // TODO: remove network
```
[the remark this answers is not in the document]
appetrosyan: See previous comment.

core/src/wsv.rs
```diff
@@ -281,22 +281,19 @@ impl WorldStateView {
     /// you likely have data corruption.
     /// - If trigger execution fails
     /// - If timestamp conversion to `u64` fails
-    #[iroha_futures::telemetry_future]
-    #[log(skip(self, block))]
-    #[allow(clippy::expect_used)]
-    pub async fn apply(&self, block: VersionedCommittedBlock) -> Result<()> {
+    #[allow(clippy::too_many_lines)]
```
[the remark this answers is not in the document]
appetrosyan: Nope. See previous comment.

core/src/wsv.rs
```diff
@@ -638,14 +630,12 @@ impl WorldStateView {
     }
 
     /// Initializes WSV with the blocks from block storage.
-    #[iroha_futures::telemetry_future]
-    pub async fn init(&self, blocks: Vec<VersionedCommittedBlock>) {
+    #[allow(clippy::expect_used)]
+    pub fn init(&self, blocks: Vec<VersionedCommittedBlock>) {
```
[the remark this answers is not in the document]
appetrosyan: See previous comment for explanation.

core/src/sumeragi/message.rs
```diff
@@ -7,279 +7,77 @@
 )]
 #![allow(clippy::significant_drop_in_scrutinee)]
```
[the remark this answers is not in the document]
appetrosyan: Not anymore.

core/src/sumeragi/mod.rs
```diff
@@ -53,42 +56,253 @@ trait Consensus {
...
+        peers.choose(rng).map(|id| Peer::new(id.clone()))
+    }
+
+    /// Access the world state view object in a locking fashion.
+    /// If you intend to do anything substantial you should clone
+    /// and release the lock. This is because no blocks can be produced
+    /// while this lock is held.
+    // TODO: Return result.
+    #[allow(clippy::expect_used)]
+    pub fn wsv_mutex_access(&self) -> std::sync::MutexGuard<WorldStateView> {
```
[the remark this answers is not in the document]
appetrosyan: True.

core/src/block_sync.rs
```diff
@@ -288,35 +170,42 @@ pub mod message {
...
                 Message::GetBlocksAfter(GetBlocksAfter { hash, peer_id }) => {
                     if block_sync.block_batch_size == 0 {
                         warn!("Error: not sending any blocks as batch_size is equal to zero.");
                         return;
                     }
-                    if *hash == block_sync.wsv.latest_block_hash() {
+                    if *hash
+                        == block_sync
+                            .sumeragi
+                            .latest_block_hash_for_use_by_block_sync()
```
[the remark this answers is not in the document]
appetrosyan: No. This is a bit of an anti-pattern.

core/src/block_sync.rs
```diff
@@ -288,35 +170,42 @@ pub mod message {
...
                         return;
                     }
 
-                    let blocks: Vec<_> = block_sync
-                        .wsv
-                        .blocks_after_hash(*hash)
-                        .take(block_sync.block_batch_size as usize)
-                        .collect();
+                    let mut blocks = block_sync.sumeragi.blocks_after_hash(*hash);
+                    blocks.truncate(block_sync.block_batch_size as usize);
```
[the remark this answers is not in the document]
appetrosyan: True. 

And we might need to revisit this at a later date.

core/src/kura.rs
```diff
@@ -35,8 +31,7 @@ pub struct Kura {
     mode: Mode,
     block_store: Mutex<Box<dyn BlockStoreTrait + Send>>,
     block_hash_array: Mutex<Vec<HashOf<VersionedCommittedBlock>>>,
-    wsv: Arc<WorldStateView>,
-    broker: Broker,
+    // broker: Broker,
```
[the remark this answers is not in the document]
appetrosyan: Dead code.

core/src/queue.rs
```diff
@@ -66,61 +70,68 @@ pub enum Error {
 
 impl Queue {
     /// Makes queue from configuration
-    pub fn from_configuration(cfg: &Configuration, wsv: Arc<WorldStateView>) -> Self {
+    pub fn from_configuration(cfg: &Configuration) -> Self {
```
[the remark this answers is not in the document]
appetrosyan: Actually, here I'd have to agree with @SamHSmith's design. This doesn't have to be a reference to **the** world state view, it could be a working copy. This might become useful later when the world state view is a sparse view structure that lazily computes the data.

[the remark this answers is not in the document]
mversic: ok, I see

cli/src/torii/routing.rs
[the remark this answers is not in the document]
mversic: but the difference is that you may get away without allocating a vector across a function boundary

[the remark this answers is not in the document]
appetrosyan: Might be worth it at some point, I would also not be surprised if LLVM elided the vector allocation without `impl IntoIterator`.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
appetrosyan: This comment actually has a point. In C, this approach would have largely worked, except in some cases it would have led to an occasional segfault. More importantly, this doesn't actually take up more space, (remember the niche optimisation).

core/src/smartcontracts/isi/asset.rs
```diff
@@ -250,6 +254,7 @@ pub mod isi {
 
                 Ok(AssetEvent::Removed(transfer.source_id.clone()))
             })?;
+            let metrics_arc = Arc::clone(&wsv.metrics);
```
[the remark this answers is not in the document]
appetrosyan: Was left out of a refactor.

core/src/sumeragi/mod.rs
```diff
@@ -53,42 +56,253 @@ trait Consensus {
...
+    ) -> Self {
+        let (incoming_message_sender, incoming_message_receiver) =
+            std::sync::mpsc::sync_channel(250);
+
+        Self {
+            internal: SumeragiWithFault::<NoFault> {
+                key_pair: configuration.key_pair.clone(),
+                peer_id: configuration.peer_id.clone(),
+                events_sender,
+                wsv: std::sync::Mutex::new(wsv),
```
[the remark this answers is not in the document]
appetrosyan: Not quite, but it makes sense that the consensus has access to *a* world state view. In reality we want to make these be generated on-demand from a world, given specific conditions.

cli/src/torii/routing.rs
```diff
@@ -108,7 +109,9 @@ pub(crate) async fn handle_instructions(
     )
     .map_err(Error::AcceptTransaction)?;
     #[allow(clippy::map_err_ignore)]
-    let push_result = queue.push(transaction).map_err(|(_, err)| err);
+    let push_result = queue
+        .push(transaction, &sumeragi.wsv_mutex_access())
```
mversic, line 113: at the moment the only way to get access to `wsv` outside of sumeragi is through a mutex?

SamHSmith: Yes. It is only used by torii for queries on the node from the outside world. The wsv that is access through the mutex is however not the one that sumeragi operates on. It is the 'public facing' wsv. So queries can take place while the wsv is being updated internally to sumeragi.

core/src/queue.rs
```diff
@@ -192,15 +204,45 @@ impl Queue {
...
+        }
+    }
+
+    /// Pops a single transaction.
+    #[allow(
+        clippy::expect_used,
+        clippy::unwrap_in_result,
+        clippy::cognitive_complexity
+    )]
+    pub fn pop_without_seen(&self, wsv: &WorldStateView) -> Option<VersionedAcceptedTransaction> {
```
ilchu, line 229: This is minor, but it feels like just adding a `bool` flag to normal `pop()` would remove the need for this function.

core/src/sumeragi/message.rs
```diff
@@ -1,285 +1,82 @@
...
-/// Get [`PeerId`] of leader in current topology.
-#[derive(Clone, Copy, Debug, iroha_actor::Message)]
-#[message(result = "PeerId")]
-pub struct GetLeader;
-
-declare_versioned_with_scale!(VersionedMessage 1..2, Debug, Clone, iroha_macro::FromVariant, iroha_actor::Message);
-
-impl VersionedMessage {
+impl VersionedPacket {
     /// Converts from `&VersionedMessage` to V1 reference
```
ilchu, line 20:
```suggestion
    /// Converts from `&VersionedPacket` to V1 reference
```

ilchu (approved):

cli/src/lib.rs
```diff
@@ -69,46 +70,71 @@ impl Default for Arguments {
...
     pub kura: Arc<Kura>,
     /// Block synchronization actor
-    pub block_sync: AlwaysAddr<B>,
+    pub block_sync: AlwaysAddr<BlockSynchronizer>,
     /// Torii web server
     pub torii: Option<Torii>,
     /// Thread handlers
     thread_handlers: Vec<ThreadHandler>,
+    /// Relay that redirects messages from the network subsystem to core subsystems.
+    _sumeragi_relay: AlwaysAddr<FromNetworkBaseRelay>,
```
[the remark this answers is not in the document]
appetrosyan: Added comment.

client/benches/tps/lib.rs
```diff
@@ -112,10 +113,6 @@ impl Config {
         #[allow(clippy::float_arithmetic, clippy::cast_precision_loss)]
         let tps = txs_accepted as f64 / elapsed_secs;
         iroha_logger::info!(%tps, %txs_accepted, %elapsed_secs, %txs_rejected);
-        if txs_rejected > 0 {
-            return Err(eyre::eyre!("{txs_rejected} transactions were rejected"));
-        }
-
```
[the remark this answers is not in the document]
appetrosyan: Yes. We want to know the raw overall performance.

core/src/queue.rs
appetrosyan: I actually prefer it this way. `pop_without_seen` tells me volumes: 
1. There's a function `pop` whose default behaviour has to do with `seen`. 
2. Using this function is unambiguously not going to record things. Compared to that it's ambiguous which: `pop(true)` or `pop(false)` records txn. 
3. You see this as an option in intellisense (which I don't use, but most of you guys do). 
4. You don't have to change all instances of `pop` into `pop(false)`. 

This approach is an anti-pattern for most Python code I've seen. Thankfully we're not writing Python.

ilchu (approved):

SamHSmith (approved):
I want to see the CI run to completion before merging.

SamHSmith:
Seems good enough.

Issue #2698, opened 2022-09-05
 Unclear error Failed to decode the whole response body as `VersionedPaginatedQueryResult`


astrokov7:
### Current environment

MacBook M1
On metal
**OS:** MacOS
**branch:** iroha2-dev

### Description of actions

1. Clone and run first project on metal

- `git clone https://github.com/hyperledger/iroha.git --branch iroha2-dev`
- `cd iroha`
- `cargo build -p iroha_client_cli`
- `scripts/test_env.sh setup`
- `mkdir -p test_docker`
- `echo '{"comment":{"String": "Hello Meta!"}}' > test_docker/metadata.json`
- `cp ./target/debug/iroha_client_cli test_docker/`
- `cd ~/Git/iroha/test_docker`
- `./iroha_client_cli`
- `/iroha_client_cli --config ./test_docker/config.json/`
- `./iroha_client_cli domain list all`

2. Clone and configure second project in another directory, but on the same metal
- `git clone https://github.com/hyperledger/iroha.git --branch iroha2-dev`
- `cd iroha`
- `cargo build -p iroha_client_cli`
- `scripts/test_env.sh setup`
- `mkdir -p test_docker`
- `echo '{"comment":{"String": "Hello Meta!"}}' > test_docker/metadata.json`
- `cp ./target/debug/iroha_client_cli test_docker/`
- `cd ~/Git/iroha/test_docker`
- `./iroha_client_cli`
- `/iroha_client_cli --config ./test_docker/config.json/`

3. Run `./iroha_client_cli domain list all`

### Actual behaviour
`User: alice@wonderland
[...]

PR #2763, opened 2022-09-20
[feature] #2698: Fix unclear error message in `iroha_client`

### LINKED ISSUE #2689: [feature] Add parameters in `PermissionTokenDefinition`
Right now we have `PermissionTokenDefinition` which should be registered before any `PermissionToken` with corresponding defition can be granted.
But definition for now contains only *id*, when actual tokens can also have some parameters (i.e. which concrete asset can be burned). We need to put parameters in `PermissionTokenDefinition` too, because they should be fixed.

Without that user can forget to provide parameters but he would know that not on `Grant` step but when some permisison validator will check this token.

We need to insert parameters inside `PermissionTokenDefinition` and check that every granted `PermissionToken` with this definition contains the same parameters (and ideally check parameter types somehow too)

Arjentix:
### Description of the Change

* Reprhased two error messages

### Issue

* Closes #2689

### Benefits

* More helpful error message for the end users

### Possible Drawbacks

None

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -72,7 +72,8 @@ where
                         try_decode_all_or_just_decode!(VersionedPaginatedQueryResult, resp.body());
                     res.wrap_err(
                         "Failed to decode response from Iroha. \
-                             Do you use compatible client version?",
+                         You are likely using a version of the client library \
+                         that is incompatible with the version of the peer software",
                     )
                     .map_err(Into::into)
                 }
@@ -87,7 +88,8 @@ where
                     }
                     let err = res.wrap_err(
                         "Failed to decode error-response from Iroha. \
-                         Do you use compatible client version?",
+                         You are likely using a version of the client library \
+                         that is incompatible with the version of the peer software",
                     )?;
                     Err(ClientQueryError::QueryError(err))
                 }
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -71,7 +71,8 @@ where
                     let res =
                         try_decode_all_or_just_decode!(VersionedPaginatedQueryResult, resp.body());
                     res.wrap_err(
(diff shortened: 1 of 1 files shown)


client/src/client.rs
```diff
@@ -71,7 +71,8 @@ where
                     let res =
                         try_decode_all_or_just_decode!(VersionedPaginatedQueryResult, resp.body());
                     res.wrap_err(
-                        "Failed to decode the whole response body as `VersionedPaginatedQueryResult`",
+                        "Failed to decode response from Iroha. \
+                             Do you use compatible client version?",
```
appetrosyan, line 75:
```suggestion
                             You are likely using a version of the client library that is incompatible with the version of the peer software?",
```

client/src/client.rs
```diff
@@ -84,8 +85,10 @@ where
                         warn!("Can't decode query error, not all bytes were consumed");
                         res = QueryError::decode(&mut resp.body().as_ref());
                     }
-                    let err =
-                        res.wrap_err("Failed to decode the whole response body as `QueryError`")?;
+                    let err = res.wrap_err(
+                        "Failed to decode error-response from Iroha. \
+                         Do you use compatible client version?",
```
appetrosyan, line 90:
```suggestion
                          You are likely using a version of the client library that is incompatible with the version of the peer software?",
```

appetrosyan (approved):

client/src/client.rs
Arjentix: @appetrosyan , may be remove `?` then?

appetrosyan: yes. remove. my mistake.

appetrosyan (approved):

6r1d (dismissed):
LGTM

appetrosyan (approved):

6r1d (approved):

PR #2770, opened 2022-09-21
[feature] #2712: Config proptests

### LINKED ISSUE #2712: Write proptests to check `Proxy` invariants
Probably something along [these](https://altsysrq.github.io/proptest-book/proptest/tutorial/recursive.html) lines.

ilchu:
### Description of the Change
* Added property testing to the Iroha and Iroha client `ConfigurationProxy` structs
* Expanded default `config.json` to allow them to be built from proxies immediately upon reading
* Updates some docs, error messages, trait derives 

### Issue
Closes #2712.

### Benefits
More confidence in current proxy implementation's robustness.

### Possible Drawbacks
I've struggled a bit with a generation strategy that would be compact yet showing the intended invariant, so I hope nothing's missing.

### Usage Examples or Tests 
`cargo test --package iroha_config --lib -- iroha::tests client::tests --nocapture`

### DIFF
--- a/config/iroha_test_config.json
+++ b/config/iroha_test_config.json
@@ -1,0 +1,115 @@
+{
+  "PUBLIC_KEY": "ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b",
+  "PRIVATE_KEY": {
+    "digest_function": "ed25519",
+    "payload": "282ed9f3cf92811c3818dbc4ae594ed59dc1a2f78e4241e31924e101d6b1fb831c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b"
+  },
+  "DISABLE_PANIC_TERMINAL_COLORS": false,
+  "KURA": {
+    "INIT_MODE": "strict",
+    "BLOCK_STORE_PATH": "./storage",
+    "BLOCKS_PER_STORAGE_FILE": 1000,
+    "ACTOR_CHANNEL_CAPACITY": 100
+  },
+  "SUMERAGI": {
+    "BLOCK_TIME_MS": 1000,
+    "TRUSTED_PEERS": [
+      {
+        "address": "127.0.0.1:1337",
+        "public_key": "ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b"
+      },
+      {
+        "address": "127.0.0.1:1338",
+        "public_key": "ed0120cc25624d62896d3a0bfd8940f928dc2abf27cc57cefeb442aa96d9081aae58a1"
+      },
+      {
+        "address": "127.0.0.1:1339",
+        "public_key": "ed0120faca9e8aa83225cb4d16d67f27dd4f93fc30ffa11adc1f5c88fd5495ecc91020"
+      },
+      {
+        "address": "127.0.0.1:1340",
+        "public_key": "ed01208e351a70b6a603ed285d666b8d689b680865913ba03ce29fb7d13a166c4e7f1f"
+      }
+    ],
+    "COMMIT_TIME_LIMIT_MS": 2000,
+    "TX_RECEIPT_TIME_LIMIT_MS": 500,
+    "TRANSACTION_LIMITS": {
+      "max_instruction_number": 4096,
+      "max_wasm_size_bytes": 4194304
+    },
+    "ACTOR_CHANNEL_CAPACITY": 100,
+    "GOSSIP_BATCH_SIZE": 500,
+    "GOSSIP_PERIOD_MS": 1000
+  },
+  "TORII": {
+    "P2P_ADDR": "127.0.0.1:1337",
+    "API_URL": "127.0.0.1:8080",
+    "TELEMETRY_URL": "127.0.0.1:8180",
+    "MAX_TRANSACTION_SIZE": 32768,
+    "MAX_CONTENT_LEN": 16384000
+  },
+  "BLOCK_SYNC": {
+    "GOSSIP_PERIOD_MS": 10000,
+    "BLOCK_BATCH_SIZE": 4,
+    "ACTOR_CHANNEL_CAPACITY": 100
+  },
+  "QUEUE": {
+    "MAXIMUM_TRANSACTIONS_IN_BLOCK": 8192,
+    "MAXIMUM_TRANSACTIONS_IN_QUEUE": 65536,
+    "TRANSACTION_TIME_TO_LIVE_MS": 86400000,
+    "FUTURE_THRESHOLD_MS": 1000
+  },
+  "LOGGER": {
+    "MAX_LOG_LEVEL": "INFO",
+    "TELEMETRY_CAPACITY": 1000,
+    "COMPACT_MODE": false,
+    "LOG_FILE_PATH": null,
+    "TERMINAL_COLORS": true
+  },
+  "GENESIS": {
+    "ACCOUNT_PUBLIC_KEY": "ed01204cffd0ee429b1bdd36b3910ec570852b8bb63f18750341772fb46bc856c5caaf",
+    "ACCOUNT_PRIVATE_KEY": {
+      "digest_function": "ed25519",
+      "payload": "d748e18ce60cb30dea3e73c9019b7af45a8d465e3d71bcc9a5ef99a008205e534cffd0ee429b1bdd36b3910ec570852b8bb63f18750341772fb46bc856c5caaf"
+    },
+    "WAIT_FOR_PEERS_RETRY_COUNT_LIMIT": 100,
+    "WAIT_FOR_PEERS_RETRY_PERIOD_MS": 500,
+    "GENESIS_SUBMISSION_DELAY_MS": 1000
+  },
+  "WSV": {
+    "ASSET_METADATA_LIMITS": {
+      "max_len": 1048576,
+      "max_entry_byte_size": 4096
+    },
+    "ASSET_DEFINITION_METADATA_LIMITS": {
+      "max_len": 1048576,
+      "max_entry_byte_size": 4096
+    },
+    "ACCOUNT_METADATA_LIMITS": {
+      "max_len": 1048576,
+      "max_entry_byte_size": 4096
+    },
+    "DOMAIN_METADATA_LIMITS": {
+      "max_len": 1048576,
+      "max_entry_byte_size": 4096
+    },
+    "IDENT_LENGTH_LIMITS": {
+      "min": 1,
+      "max": 128
+    },
+    "WASM_RUNTIME_CONFIG": {
+      "FUEL_LIMIT": 1000000,
+      "MAX_MEMORY": 524288000
+    }
+  },
+  "NETWORK": {
+    "ACTOR_CHANNEL_CAPACITY": 100
+  },
+  "TELEMETRY": {
+    "NAME": null,
+    "URL": null,
+    "MIN_RETRY_PERIOD": 1,
+    "MAX_RETRY_DELAY_EXPONENT": 4,
+    "FILE": null
+  }
+}
--- a/config/src/client.rs
+++ b/config/src/client.rs
@@ -3,8 +3,8 @@
 use core::str::FromStr;
 
 use derive_more::Display;
-use eyre::{eyre, Result};
-use iroha_config_base::derive::{Documented, LoadFromEnv, Proxy};
+use eyre::{eyre, Result, WrapErr};
+use iroha_config_base::derive::{Documented, Error as ConfigError, LoadFromEnv, Proxy};
 use iroha_crypto::prelude::*;
 use iroha_data_model::{prelude::*, transaction};
 use iroha_primitives::small::SmallStr;
@@ -15,7 +15,7 @@ const DEFAULT_TRANSACTION_STATUS_TIMEOUT_MS: u64 = 15_000;
 const DEFAULT_ADD_TRANSACTION_NONCE: bool = false;
 
 /// Wrapper over `SmallStr` to provide basic auth login checking
-#[derive(Debug, Display, Clone, Serialize)]
+#[derive(Debug, Display, Clone, Serialize, PartialEq, Eq)]
 pub struct WebLogin(SmallStr);
 
 impl WebLogin {
@@ -51,7 +51,7 @@ impl<'de> Deserialize<'de> for WebLogin {
 }
 
 /// Basic Authentication credentials
-#[derive(Clone, Deserialize, Serialize, Debug, Documented)]
+#[derive(Clone, Deserialize, Serialize, Debug, Documented, PartialEq, Eq)]
 pub struct BasicAuth {
     /// Login for Basic Authentication
     pub web_login: WebLogin,
@@ -60,7 +60,7 @@ pub struct BasicAuth {
 }
 
 /// `Configuration` provides an ability to define client parameters such as `TORII_URL`.
-#[derive(Debug, Clone, Deserialize, Serialize, Proxy, LoadFromEnv, Documented)]
+#[derive(Debug, Clone, Deserialize, Serialize, Proxy, LoadFromEnv, Documented, PartialEq, Eq)]
 #[serde(rename_all = "UPPERCASE")]
 #[config(env_prefix = "IROHA_")]
 pub struct Configuration {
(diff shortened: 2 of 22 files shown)


primitives/src/small.rs
```diff
@@ -21,7 +21,7 @@ pub const SMALL_SIZE: usize = 8_usize;
 mod small_string {
     use super::*;
 
-    #[derive(Debug, Clone, derive_more::Display, Deserialize, Serialize)]
+    #[derive(Debug, Clone, derive_more::Display, Deserialize, Serialize, PartialEq, Eq)]
```
appetrosyan, line 24: In general, the more we derive the better.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::torii::{uri::DEFAULT_API_URL, DEFAULT_TORII_TELEMETRY_URL};
+
+    const CONFIGURATION_PATH: &str = "../configs/client_cli/config.json";
+
+    #[allow(clippy::expect_used)]
+    fn placeholder_keypair() -> KeyPair {
+        let public_key = "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
```
appetrosyan, line 122: Style choice, but I'd prefer this be derived from the private key.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+
+        KeyPair::new(public_key, private_key).expect("Key pair mismatch")
+    }
+
+    #[allow(clippy::expect_used)]
+    fn placeholder_account() -> <Account as Identifiable>::Id {
+        AccountId::from_str("alice@wonderland").expect("Account ID not valid")
+    }
+
+    fn arb_keys() -> BoxedStrategy<(Option<PublicKey>, Option<PrivateKey>)> {
```
appetrosyan, line 138: I'm not 100% sure about this. Most keys are completely random so a proptest does literally nothing. Could we run tests with edge cases (one missing key, other missing key, both present but don't match), and leave this out of the proptest?

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+            .boxed()
+    }
+
+    prop_compose! {
+        fn arb_proxy()
+            (
+                (public_key, private_key) in arb_keys(),
+                account_id in prop::option::of(Just(placeholder_account())),
+                basic_auth in prop::option::of(Just(Some(BasicAuth{web_login:WebLogin(SmallStr::from_str("mad_hatter")),password:SmallStr::from_str("ilovetea")}))),
+                torii_api_url in prop::option::of(Just(SmallStr::from_str(DEFAULT_API_URL))),
```
appetrosyan, line 153: Actually here it'd be nice if we could also test different API URLs (e.g. one that's malformed, `http`, `ftp`, `ws` and so on. Things which are obviously wrong should be caught early, so someone who put in an extra `https` knows what went wrong before the peer is up.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+                torii_telemetry_url in prop::option::of(Just(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL))),
```
appetrosyan, line 154: Same here, except we know that this also has to provide a port, because the one case where the port was optional is the previous case. If the URLs are not identical, this warrants a warning in the logs, because it **might be** but isn't necessarily a mistake.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+        )
+            .boxed()
+    }
+
+    prop_compose! {
+        fn arb_proxy()
+            (
+                (public_key, private_key) in arb_keys(),
+                account_id in prop::option::of(Just(placeholder_account())),
+                basic_auth in prop::option::of(Just(Some(BasicAuth{web_login:WebLogin(SmallStr::from_str("mad_hatter")),password:SmallStr::from_str("ilovetea")}))),
```
appetrosyan, line 152: I wouldn't test this at all. Basic auth is completely optional.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+                torii_api_url in prop::option::of(Just(SmallStr::from_str(DEFAULT_API_URL))),
+                torii_telemetry_url in prop::option::of(Just(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL))),
+                transaction_time_to_live_ms in prop::option::of(Just(DEFAULT_TRANSACTION_TIME_TO_LIVE_MS)),
```
appetrosyan, line 155: Probably going to be fine if we test edge-cases only. Although I'd probably throw a warning when the TTL is like a few milliseconds.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+                transaction_status_timeout_ms in prop::option::of(Just(DEFAULT_TRANSACTION_STATUS_TIMEOUT_MS)),
```
appetrosyan, line 156: Check if this is shorter than TTL. Again, can be done via edge cases.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+                transaction_limits in prop::option::of(Just(TransactionLimits {
```
appetrosyan, line 157: Probably warrant a warning if below certain threshold. `1` instruction per transaction is probably not a problem, but WASM has a certain minimum size. If the max is set too low, we should notify the user that they're doing something silly.

config/src/genesis.rs
```diff
@@ -43,3 +43,48 @@ impl Default for ConfigurationProxy {
...
+            prop::option::of(Just(None)),
+        )
+            .boxed()
+    }
+
+    prop_compose! {
+        pub fn arb_proxy()
+            (
+                (account_public_key, account_private_key) in arb_keys(),
+                wait_for_peers_retry_count_limit in prop::option::of(Just(DEFAULT_WAIT_FOR_PEERS_RETRY_COUNT_LIMIT)),
```
appetrosyan, line 82: Just like above, I'd consider adding warnings (but in a separate ticket).

config/src/client.rs
Erigara: This would require adding `From<PrivateKey>`, we have this [discussion](https://github.com/hyperledger/iroha/pull/2704#discussion_r970818247) in one of previous PR.
Maybe this PR is good opportunity to add this.

config/src/client.rs
```diff
@@ -106,3 +106,75 @@ impl Default for ConfigurationProxy {
...
+    use iroha_config_base::proxy::{Builder, LoadFromDisk};
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::torii::{uri::DEFAULT_API_URL, DEFAULT_TORII_TELEMETRY_URL};
+
+    const CONFIGURATION_PATH: &str = "../configs/client_cli/config_test.json";
+
+    #[allow(clippy::expect_used)]
+    fn placeholder_keypair() -> KeyPair {
```
Erigara, line 121: Another option would be to generate `KeyPair` with seed and seed would be determined by proptest.

config/src/client.rs
```diff
@@ -60,7 +60,7 @@ pub struct BasicAuth {
 }
 
 /// `Configuration` provides an ability to define client parameters such as `TORII_URL`.
-#[derive(Debug, Clone, Deserialize, Serialize, Proxy, LoadFromEnv, Documented)]
+#[derive(Debug, Clone, Deserialize, Serialize, Proxy, LoadFromEnv, Documented, PartialEq, Eq)]
 #[serde(rename_all = "UPPERCASE")]
 #[config(env_prefix = "IROHA_")]
 pub struct Configuration {
```
Erigara, line 66: I think we could add here `finalize` method as well and in finalize we would check configuration invaraints, log warnings and so on.

Erigara (commented):
I like the first steps in direction of property based testing. 
IMO if done right (especially something like [stateful-testsing](https://hypothesis.readthedocs.io/en/latest/stateful.html)) it could help us to find bugs that would be hard to find otherwise.
John Hughes has great talks on this topic.

config/src/torii.rs
```diff
@@ -76,3 +76,24 @@ pub mod uri {
...
 }
+
+#[cfg(test)]
+pub mod tests {
+    use proptest::prelude::*;
+
+    use super::*;
+
+    prop_compose! {
+        pub fn arb_proxy()
```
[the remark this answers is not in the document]
ilchu: Arbitrary (as in randomly generated) proxy. Just following the naming convention from the `proptest` docs.

config/src/sumeragi.rs
```diff
@@ -90,7 +90,15 @@ impl ConfigurationProxy {
         self.trusted_peers = Some(TrustedPeers { peers });
     }
 }
+
 impl Configuration {
+    /// Set `trusted_peers` configuration parameter. Will overwrite
+    /// existing `trusted_peers` but does not check for duplication.
```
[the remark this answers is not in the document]
ilchu: It's an old method that's used nowhere, so maybe it would be more correct to just delete it? But what the doc meant I suppose is that it doesn't warn the user if some of the provided peers were identical.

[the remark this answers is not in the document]
Erigara: I don't mind deleting this method if it doesn't used anywhere.

config/src/client.rs
ilchu: I wouldn't say it's actually being tested, as that default value is wrapped in an option, and only the option is being flipped. But yeah, `None` looks and reads better.

config/src/client.rs
ilchu: I'll choose generating from a private key hex, as having that additional `From` impl might be beneficial for other parts of the code in the future.

appetrosyan:
Just a vocabulary note. The word `finalize` has the `-ize` ending, the spelling of which cannot be agreed upon. If we write `ise` we'll annoy all American contributors, while writing `-ize` annoys British English speakers. 

Can we come up with a different word? For example: `finish`, `commit`, `complete`, hell even `consummate`?

config/src/genesis.rs
ilchu: Created https://github.com/hyperledger/iroha/issues/2801

config/src/client.rs
ilchu: Update: I naively thought that any randomly generated 512-bit hex would properly decompress into a key, which turned out not to be the case. Thus going for generation from seed, but still adding that `From` impl.

config/src/client.rs
```diff
@@ -106,3 +106,166 @@ impl Default for ConfigurationProxy {
...
+        }
+        if let Some(tx_limits) = self.transaction_limits {
+            // 1 KiB
+            if tx_limits.max_wasm_size_bytes < 2_u64.pow(10) {
+                eyre::bail!(ConfigError::ProxyBuildError("`TRANSACTION_LIMITS` parameter's `max_wasm_size` field really small, consider making it bigger".to_owned()));
+            }
+        }
+        if let Some(api_url) = &self.torii_api_url {
+            let api_url = api_url.clone().to_string();
+            let split_api_url = api_url.split("://").collect::<Vec<_>>();
```
Erigara, line 144: Could be done without allocation, but i wouldn't insist on this. Since we execute this method only once and i remember you suggestion about strong typing.

config/src/iroha.rs
```diff
@@ -151,79 +148,101 @@ impl ConfigurationProxy {
...
 
     #[test]
     fn parse_example_json() -> Result<()> {
         let cfg_proxy = ConfigurationProxy::from_path(CONFIGURATION_PATH)
             .wrap_err("Failed to read configuration from example config")?;
-        assert_eq!(
-            "./storage",
-            cfg_proxy.kura.unwrap().block_store_path.unwrap()
-        );
+        assert_eq!("127.0.0.1:1337", cfg_proxy.torii.unwrap().p2p_addr.unwrap());
```
Erigara, line 230: Why do we switch this line back and forth?

config/src/iroha.rs
```diff
@@ -151,79 +148,101 @@ impl ConfigurationProxy {
...
+
     use super::*;
     use crate::{base::proxy::LoadFromDisk, sumeragi::TrustedPeers};
 
-    const CONFIGURATION_PATH: &str = "../configs/peer/config.json";
+    const CONFIGURATION_PATH: &str = "../configs/peer/config_test.json";
+
+    /// Key-pair used for proptests generation
+    #[allow(clippy::expect_used)]
+    pub fn placeholder_keypair() -> KeyPair {
```
Erigara, line 173: Why not generate keys here as well?

config/base/tests/simple.rs
```diff
@@ -6,7 +6,7 @@ use iroha_config_base::{
 };
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Deserialize, Serialize, Documented, LoadFromEnv)]
+#[derive(Clone, Debug, Deserialize, Serialize, Documented, LoadFromEnv, PartialEq, Eq)]
```
appetrosyan, line 9: In general, if some trait can be derived, it should be derived in library crates. Not so in binary crates.

config/src/client.rs
appetrosyan: I propose a different naming convention, one that avoids the `ize`/`ise` suffices.

config/src/client.rs
appetrosyan: I prefer it this way. The allocation is only a problem when you do several times in a row. Almost everything related to configuration is run once, so flexibility is more important.

crypto/src/lib.rs
```diff
@@ -439,6 +439,27 @@ impl From<Multihash> for PublicKey {
...
+        let key_gen_option = Some(UrsaKeyGenOption::FromSecretKey(UrsaPrivateKey(
+            private_key.payload,
+        )));
+        let (mut public_key, _) = match digest_function {
+            Algorithm::Ed25519 => Ed25519Sha512.keypair(key_gen_option),
+            Algorithm::Secp256k1 => EcdsaSecp256k1Sha256::new().keypair(key_gen_option),
+            Algorithm::BlsNormal => BlsNormal::new().keypair(key_gen_option),
+            Algorithm::BlsSmall => BlsSmall::new().keypair(key_gen_option),
+        }
+        .expect("can't fail for valid `PrivateKey`");
```
appetrosyan, line 455: We should actually fix this upstream in hyperledger/ursa

configs/peer/config_test.json
```diff
@@ -0,0 +1,115 @@
+{
```
appetrosyan, line 1: If you need this file, it needs to be placed elsewhere.

appetrosyan: Given that this is used for testing of the `config` crate, I'd suggest putting it there locally.

appetrosyan (changes requested):

config/src/iroha.rs
appetrosyan: Formatting. Small changes lead to it being either written on separate lines, or on the same line.

config/src/iroha.rs
appetrosyan: This is going to add stochasticity to the (already flaky) tests. We need to remove as many variables as we can, unless they are part of the proptest. In that case, the keys will need to be generated from a seed that's provide by the proptest framework.

crypto/src/lib.rs
ilchu: If you mean filing a PR with this impl into their repo, then I'm a bit skeptical about the speed of their PR acceptance process, but I could try to.

config/src/client.rs
```diff
@@ -106,3 +106,166 @@ impl Default for ConfigurationProxy {
         }
     }
 }
+
+impl ConfigurationProxy {
+    /// Finalise Iroha config proxy by instantiating mutually equivalent fields
```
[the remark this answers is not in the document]
ilchu: Missed updating it, thanks!

config/src/iroha.rs
ilchu: Actually it's not about formatting here, but due to me mindlessly rewriting back changes introduced by [this](https://github.com/hyperledger/iroha/pull/2782#discussion_r980902572). Will revert it back.

crypto/src/lib.rs
ilchu: On further inspection, ursa has nothing but opaque bytearrays in place of private / public keys with no associated digests, and changing the structs to a product type would break FFI, macros, and who knows what else. So I guess the cost is not worth it.

config/iroha_test_config.json
```diff
@@ -0,0 +1,115 @@
+{
```
appetrosyan, line 1: Much better.

appetrosyan (dismissed):

Erigara (approved):

config/src/client.rs
```diff
@@ -106,3 +106,165 @@ impl Default for ConfigurationProxy {
         }
     }
 }
+
+impl ConfigurationProxy {
+    /// Finalise Iroha client config proxy by checking that certain fields identify reasonable limits or
+    /// or are well formatted.
```
outoftardis, line 112:
```suggestion
    /// are well formatted.
```

config/src/client.rs
```diff
@@ -106,3 +106,165 @@ impl Default for ConfigurationProxy {
...
+            }
+        }
+        Ok(())
+    }
+
+    /// The wrapper around the client `ConfigurationProxy` that performs
+    /// finalisation prior to building `Configuration`. Just like
+    /// Iroha peer config, its `<Self as iroha_config_base::proxy::Builder>::build()`
+    /// method should never be used directly, as only this wrapper ensures final
+    /// coherence and issues important warnings.
```
outoftardis, line 180: maybe
```suggestion
    /// coherence and warns if there are any issues.
```

appetrosyan (dismissed):

SamHSmith (approved):
LGTM

Erigara (approved):

PR #2771, opened 2022-09-21
[feature] #2765: Generate synthetic genesis in `kagami`

### LINKED ISSUE #2765: [suggestion]: Generate syntatic genesis
### Feature request

Add functionality to `kagami` to generate syntactic genesis. 

### Motivation

Could be helpful for load testing and blockchain explorer.

### Who can help?

_No response_

Erigara:

### Description of the Change

Add option to `kagami` to generate synthetic genesis with specified number of domains, accounts, assets.  

### Issue

Closes #2765.

### Benefits

Possible to generate large genesis for load testing and blockchain explorer.

### Possible Drawbacks

None.

```bash
cargo run --bin kagami genesis --synthetic --domains 1 --accounts-per-domain 1 --assets-per-domain 1
```

Could be separate script.

### DIFF
--- a/tools/kagami/src/main.rs
+++ b/tools/kagami/src/main.rs
@@ -175,16 +175,39 @@ mod genesis {
     use super::*;
 
     #[derive(StructOpt, Debug, Clone, Copy)]
-    pub struct Args;
+    pub struct Args {
+        /// If set generate synthetic genesis with specified number of domains, accounts and assets.
+        #[clap(short, long, action)]
+        synthetic: bool,
+        /// Number of domains in synthetic genesis.
+        /// This parameter is ignored if `synthetic` is `false`.
+        #[clap(long, default_value_t = 0)]
+        domains: u64,
+        /// Number of accounts per domains in synthetic genesis.
+        /// Total number of  accounts would be `domains * assets_per_domain`.
+        /// This parameter is ignored if `synthetic` is `false`.
+        #[clap(long, default_value_t = 0)]
+        accounts_per_domain: u64,
+        /// Number of assets per domains in synthetic genesis.
+        /// Total number of assets would be `domains * assets_per_domain`.
+        /// This parameter is ignored if `synthetic` is `false`.
+        #[clap(long, default_value_t = 0)]
+        assets_per_domain: u64,
+    }
 
     impl<T: Write> RunArgs<T> for Args {
         fn run(self, writer: &mut BufWriter<T>) -> Outcome {
-            writeln!(
-                writer,
-                "{}",
-                serde_json::to_string_pretty(&generate_default()?)?
-            )
-            .wrap_err("Failed to write.")
+            let genesis = if self.synthetic {
+                generate_synthetic(
+                    self.domains,
+                    self.accounts_per_domain,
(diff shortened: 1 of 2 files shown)


tools/kagami/src/main.rs
```diff
@@ -175,16 +175,39 @@ mod genesis {
     use super::*;
 
     #[derive(StructOpt, Debug, Clone, Copy)]
-    pub struct Args;
+    pub struct Args {
+        /// If set generate synthetic genesis with specified number of domains, accounts and assets.
+        #[clap(short, long, action)]
```
appetrosyan, line 180: I'd imagine that it's much better to phrase this as an `enum`. So `genesis --default` is the default value, `genesis` becomes `genesis --default` and `genesis --synthetic` accepts the three args. 

Just a suggestion. Fix at will.

tools/kagami/src/main.rs
```diff
@@ -211,6 +234,63 @@ mod genesis {
...
+                            ),
+                        );
+                        acc.push(mint);
+                    }
+                }
+            }
+            acc
+        }
+        .into_iter()
+        .map(Into::into);
```
appetrosyan, line 284: Good-enough for a start, but not quite representative of a real blockchain.

appetrosyan (approved):

tools/kagami/src/main.rs
```diff
@@ -211,6 +234,63 @@ mod genesis {
...
+        assets_per_domain: u64,
+    ) -> color_eyre::Result<RawGenesisBlock> {
+        let mut builder = RawGenesisBlockBuilder::new();
+        for domain in 0..domains {
+            let mut domain_builder = builder.domain(format!("domain_{domain}").parse()?);
+
+            for account in 0..accounts_per_domain {
+                domain_builder = domain_builder.with_account(
+                    format!("account_{account}").parse()?,
+                    DEFAULT_PUBLIC_KEY.parse()?,
```
s8sato, line 250: It depends on the use of `synthetic`, but generating keys would be closer to the actual situation

s8sato (approved):
Did you intentionally renamed from `syntactic` to `synthetic`?

tools/kagami/src/main.rs
appetrosyan: I agree. Please generate different keys. In the new account scheme two accounts shall not have the same primary key.

Erigara:
> Did you intentionally renamed from `syntactic` to `synthetic`?

Yes, like [artificial](https://dictionary.cambridge.org/dictionary/english/synthetic).

tools/kagami/src/main.rs
```diff
@@ -211,6 +234,63 @@ mod genesis {
...
+            for asset in 0..assets_per_domain {
+                domain_builder = domain_builder
+                    .with_asset(format!("asset_{asset}").parse()?, AssetValueType::Quantity);
+            }
+
+            builder = domain_builder.finish_domain();
+        }
+        let mut genesis = builder.build();
+
+        let mints = {
```
appetrosyan, line 263: These aren't strictly necessary, but since we have them, we should consider adding `Transfer`s in a separate issue.

PR #2772, opened 2022-09-21
[feature] #2756: Less verbose json representation

### LINKED ISSUE #2756: [suggesion] Make structs json representation less verbose
Right now we have big json representation of some structs. I.e. in genesis users should write:

```json
{
  "account_id": {
    "name": "mad_hatter",
    "domain_id": {
      "name": "looking_glass"
    }
  }
}
```

Instead of just:

```json
{
  "account_id": "mad_hatter@looking_glass",
}
```

The same is true for a lot of other structs. This a pain to read and write. Let's fix that

Arjentix:
### Description of the Change

See commits

If you know something else, which json representation should be improved -- leave a comment

### Issue

- Closes #2756

### Benefits

- Much more clean json (genesis.json `-100` lines!)

### Possible Drawbacks

- A little bit more code


Erigara (approved):
Cool, `json` looks much more readable now

data_model/src/asset.rs
```diff
@@ -384,6 +385,27 @@ pub struct DefinitionId {
     pub domain_id: <Domain as Identifiable>::Id,
 }
 
+/// Asset Definition Identification is represented by `name#domain_name` string.
+impl FromStr for DefinitionId {
+    type Err = ParseError;
```
appetrosyan, line 390: I don't like our `ParseError` it contains no useful information, because it has to contain a `'static str`. We need to come up with a different better alternative,

data_model/src/asset.rs
```diff
@@ -384,6 +385,27 @@ pub struct DefinitionId {
...
+        match (split.next(), split.next(), split.next()) {
+            (Some(""), _, _) => Err(ParseError {
+                reason: "Asset Definition ID cannot be empty",
+            }),
+            (Some(name), Some(domain_id), None) if !domain_id.is_empty() => Ok(Self {
+                name: name.parse()?,
+                domain_id: domain_id.parse()?,
+            }),
+            _ => Err(ParseError {
+                reason: "Asset Definition ID should have format `asset#domain`",
```
appetrosyan, line 403: Ideally we want to retain the candidate `String` as well. Haskell has a lot of resources on the benefits of functional error handling (where you don't just throw away the partial candidate). 

Ideally this should allow people to fix their `String` representations.

data_model/src/events/data/filters.rs
```diff
@@ -21,18 +26,50 @@ pub type EventFilter = FilterOpt<EntityFilter>;
     IntoSchema,
     Hash,
 )]
-/// Optional filter. May pass all items or may filter them by `F`
-///
-/// It's better than `Optional<F>` because `Optional` already has its own `filter` method and it
-/// would be ugly to use fully qualified syntax to call `Filter::filter()` method on it.
-/// Also `FilterOpt` variant names look better for filter needs
+#[serde(untagged)]
```
appetrosyan, line 29: Fascinating.

data_model/src/expression.rs
```diff
@@ -457,6 +490,51 @@ impl<T: Into<Value>> From<T> for ExpressionBox {
     }
 }
 
+/// Deserialize [`Expression`] with possibility to omit [`Raw`](Expression::Raw) tag
+/// as it is the most popular variant.
+///
+/// First will try to deserialize as `Raw` and if it fails will try to deserialize as `Expression`.
```
appetrosyan, line 496: This is normally an anti-pattern, but in this case, fail-fast parsing is the preferred solution.

appetrosyan (approved):

Issue #2775, opened 2022-09-22
Generate different keys in `kagami --synthetic`


appetrosyan:
It depends on the use of `synthetic`, but generating keys would be closer to the actual situation

_Originally posted by @s8sato in https://github.com/hyperledger/iroha/pull/2771#discussion_r976770513_

PR #2777, opened 2022-09-22
[feature] #2774, #2775: `kagami` genesis improvments

### LINKED ISSUE #2774: Improve kagami synthetic genesis UX
I'd imagine that it's much better to phrase this as an `enum`. So `genesis --default` is the default value, `genesis` becomes `genesis --default` and `genesis --synthetic` accepts the three args. 

Just a suggestion. Fix at will.

_Originally posted by @appetrosyan in https://github.com/hyperledger/iroha/pull/2771#discussion_r976746045_

Erigara:
### Description of the Change

- Generate key for every `Account`;
- Extend synthetic genesis with default domain (`wonderland`) and account (`alice`) to be able to perform query;
- Make genesis generation mode subcommand instead of flag.

### Issue

Closes #2774, #2775.

### Benefits

Improved UX and generation.

### Possible Drawbacks

None.

Generation of default genesis left unchanged:

```bash
cargo run --bin kagami -- genesis
```

Generation of synthetic genesis:

```bash
cargo run --bin kagami -- genesis synthetic --domians 100 --accounts-per-domain 10 --assets-per-domain 10
```

### DIFF
--- a/tools/kagami/src/main.rs
+++ b/tools/kagami/src/main.rs
@@ -165,6 +165,7 @@ mod schema {
 }
 
 mod genesis {
+    use clap::{Parser, Subcommand};
     use iroha_core::{
         genesis::{RawGenesisBlock, RawGenesisBlockBuilder},
         tx::{AssetValueType, MintBox, RegisterBox},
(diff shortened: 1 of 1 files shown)


appetrosyan (approved):

ilchu (approved):
LGTM. As I understand, the synthetic accounts won't really be usable, and are only there
to make a richer genesis to explore. I wonder if there could be a usecase with SDK devs for autogenerated
accounts that would also return their full keys (maybe to make bootstrapping a bit easier).

PR #2779, opened 2022-09-22
[documentation] #2544: Tutorial doctests, a version without API changes

6r1d:
### Description of the Change

Same as [2603](https://github.com/hyperledger/iroha/pull/2603) except the API changes.
Adds example that mirrors tutorial examples, so that changes in the API can be propagated by directly taking them from GitHub.

### Issue

Addresses https://github.com/hyperledger/iroha/issues/2544

### Benefits

Direct connection to tutorial

### Possible Drawbacks

More code to test

`cargo run --example tutorial`

### Alternate Designs

As a test

### DIFF
--- a/client/examples/tutorial.rs
+++ b/client/examples/tutorial.rs
@@ -1,0 +1,186 @@
+//! This file contains examples from the Rust tutorial.
+//! <https://hyperledger.github.io/iroha-2-docs/guide/rust.html#_2-configuring-iroha-2>
+//!
+#![allow(clippy::restriction, clippy::needless_borrow)]
+
+// use eyre::{Error, WrapErr};
+// BEGIN FRAGMENT: rust_config_example
+// use iroha_config::client::Configuration;
+
+fn main() {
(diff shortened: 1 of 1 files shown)


appetrosyan (dismissed):

client/examples/tutorial.rs
```diff
@@ -0,0 +1,186 @@
...
+    // account_registration_test(&config)
+    //     .expect("Account registration example is expected to work correctly");
+    // asset_registration_test(&config)
+    //     .expect("Asset registration example is expected to work correctly");
+    println!("Success!");
+    // END ESCAPE
+}
+// END FRAGMENT
+
+// fn json_config_client_test(config: &Configuration) -> Result<(), Error> {
```
appetrosyan, line 31: The entire point of these tests is that we need to not have them commented. Uncomment please. If they fail we need to fix them.

appetrosyan (changes requested):

PR #2782, opened 2022-09-23
[ci] #2778: Client config check

### LINKED ISSUE #2778: [ci] Add a way to check client config.json
After a merge of #2756 I noticed that we broke `configs/client/config.json`. And our ci didn't catch that.
I suggest to add client config generation in kagami and add check with that into our `ci` as well as we do for `genesis.json` and etc.

Arjentix:
### Description of the Change

* Added kagami sub-command to generate valid client config
* Added ci check
* New config was tested with `iroha_client_cli` and it works. Current config on `iroha2-dev` is not working

### Issue

* Closes #2778 

### Benefits

* CI check for corrent client config
* Automatic way to generate valid config

### Possible Drawbacks

None

### Usage Examples or Tests

```bash
scripts/check.sh client
```

### DIFF
--- a/config/src/iroha.rs
+++ b/config/src/iroha.rs
@@ -213,7 +213,10 @@ mod tests {
     fn parse_example_json() -> Result<()> {
         let cfg_proxy = ConfigurationProxy::from_path(CONFIGURATION_PATH)
             .wrap_err("Failed to read configuration from example config")?;
-        assert_eq!("127.0.0.1:1337", cfg_proxy.torii.unwrap().p2p_addr.unwrap());
+        assert_eq!(
+            "./storage",
+            cfg_proxy.kura.unwrap().block_store_path.unwrap()
+        );
         assert_eq!(
             10000,
             cfg_proxy.block_sync.unwrap().gossip_period_ms.unwrap()
--- a/configs/client_cli/config.json
+++ b/configs/client_cli/config.json
@@ -1,18 +1,21 @@
 {
-  "TORII_API_URL": "http://127.0.0.1:8080",
-  "ACCOUNT_ID": {
-    "name": "alice",
-    "domain_id": {
-      "name": "wonderland"
-    }
-  },
-  "BASIC_AUTH": {
-    "web_login": "mad_hatter",
-    "password": "ilovetea"
-  },
   "PUBLIC_KEY": "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0",
   "PRIVATE_KEY": {
     "digest_function": "ed25519",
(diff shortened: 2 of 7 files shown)


tools/kagami/src/main.rs
```diff
@@ -44,8 +44,10 @@ pub enum Args {
     Crypto(Box<crypto::Args>),
     /// Generate the schema used for code generation in Iroha SDKs
     Schema(schema::Args),
-    /// Generate the default genesis block that is used in tests
+    /// Generate the genesis block that is used in tests
     Genesis(genesis::Args),
+    /// Generate the default client configuration
+    Client(client::Args),
```
Erigara, line 50: As follow up task we could generate configuration for peer as well, in this case we could use the following design: `kagami config client/peer`.

Erigara (approved):

config/src/iroha.rs
```diff
@@ -213,7 +213,10 @@ mod tests {
     fn parse_example_json() -> Result<()> {
         let cfg_proxy = ConfigurationProxy::from_path(CONFIGURATION_PATH)
             .wrap_err("Failed to read configuration from example config")?;
-        assert_eq!("127.0.0.1:1337", cfg_proxy.torii.unwrap().p2p_addr.unwrap());
+        assert_eq!(
+            "./storage",
+            cfg_proxy.kura.unwrap().block_store_path.unwrap()
+        );
```
ilchu, line 219: What's the purpose of changing this assert? Also I don't understand how this test is necessary. All it does is check that the example file has certain fields, but we're just checking two of them here seemingly at random.

ilchu (approved):

config/src/iroha.rs
Arjentix: I agree that this test looks strange, but

1. I changed it because it was failing (on iroha2-dev too)
2. Right now this is the only way we check peer config correctness and it wasn't introduced by me. So I decided to fix it somehow to make it working

configs/client_cli/config.json
```diff
@@ -1,18 +1,21 @@
...
-  }
+  "TORII_API_URL": "http://127.0.0.1:8080",
+  "TORII_TELEMETRY_URL": "http://127.0.0.1:8180",
+  "TRANSACTION_TIME_TO_LIVE_MS": 100000,
+  "TRANSACTION_STATUS_TIMEOUT_MS": 10000,
+  "TRANSACTION_LIMITS": {
+    "max_instruction_number": 4096,
+    "max_wasm_size_bytes": 4194304
+  },
+  "ADD_TRANSACTION_NONCE": false
```
outoftardis, line 20: update config.json example in docs

Issue #2780, opened 2022-09-23
Fix `I2::Dev::Tests`


s8sato:
https://github.com/hyperledger/iroha/actions/runs/3099465203/jobs/5018635965

s8sato:
Results of [the failing test command](https://github.com/hyperledger/iroha/blob/88c5290cf34d8a35915d966a5737161bfdfb4546/.github/workflows/iroha2-dev-pr.yml#L66
) in my local at 88c5290c:

- When
    ```bash
    cargo llvm-cov --no-report --all-features --workspace --no-fail-fast
    ```
    , the error is reproducible.

- When
    ```bash
    cargo test --bins --lib --all-features --workspace --no-fail-fast
    ```
    , the error is NOT reproducible. It fails for another reason -- there are some failed unit tests.

Thus, it is likely a problem related to [`cargo-llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov)

adaagava:
Fixed here: https://github.com/hyperledger/iroha/actions/runs/3626857776/jobs/6116209060

PR #2785, opened 2022-09-26
[ci]: move jobs to github runners

BAStos525:
### Description of the Change
Replace self-hosted runners inside `tests` workflows with the default `ubuntu-latest` Gihub Runners.

### Issue
Some long tests are failed on self-hosted runners due to undetected reason.

### Benefits
Long tests should not be failed when they are run on the deault GH Runners.

### Possible Drawbacks
Should be none, but we have to re-implement self-hosted runners.

### DIFF
--- a/.github/workflows/iroha2-dev-pr-wasm.yaml
+++ b/.github/workflows/iroha2-dev-pr-wasm.yaml
@@ -41,7 +41,7 @@ jobs:
         run: cargo doc --no-deps --quiet
 
   tests:
-    runs-on: [self-hosted, Linux]
+    runs-on: ubuntu-latest #[self-hosted, Linux]
     container:
       image: 7272721/i2-ci:nightly
     steps:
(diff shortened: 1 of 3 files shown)


BAStos525:
After some tests, I think that the problem isn't in runners.

BAStos525:
> Isn't the issue this PR is targeting #2780 @BAStos525 ? If so, it seems to be related to `cargo-llvm-cov`

Perhaps you are right. I can't say for sure what is the exact reason. I can just tell that it's not related to compute resources lack: I tried to run this job on 1 x Intel(R) Xeon(R) E-2278G CPU @ 3.40GHz & 32GB RAM and it was the same.
But this workflow is successfully finished on [this state](https://github.com/BAStos525/soramitsu-iroha/actions/runs/3103527122).

appetrosyan:
Didn't seem to help much. We should revisit this later. For now, not having coverage is fine.

BAStos525:
This PR is actual now: our colleague destroyed all AWS ephemeral self-hosted runners, so we can use only GH Actions default runners. We can create a new self-hosted runners in other cloud platform if they will be required in the feature.

appetrosyan (approved):

pesterev (approved):

PR #2788, opened 2022-09-26
[fix] #2787: Notify every listener to shutdown on panic

### LINKED ISSUE #2787: [bug]: Iroha panic handler not working correctly
Currently `iroha` on panic notify only single listener about shutdown.

Erigara:

### Description of the Change

Replaced `.notify_one()` with `.notify_waiters()` in order to notify every waiter about shutdown.

### Issue

Closes #2787.

### Benefits

Iroha actually shutdown. 

### Possible Drawbacks

There is still possibility of deadlock if `.notified().await` would be called after `.notify_waiters()` was called, but this seems to be very unlikely case , because either start warp (in this case waiters would be ready) or warp not started yet (int this case nothing will prevent iroha from shutdown).

```
cargo test --package iroha --lib -- tests::iroha_should_notify_on_panic --exact --nocapture
```

- Replace `Notify` channel with different sync primitive.
- Every `waiter` must call `.notify_one()` after receiving notification: more verbose, but solve case of possible deadlock.


cli/src/lib.rs
```diff
@@ -390,6 +391,7 @@ where
                 _ = sigterm.recv() => {},
             }
 
+            // NOTE: shutdown all currently listening waiters
```
appetrosyan, line 394: Do we **really** need this comment?

appetrosyan (approved):

cli/src/lib.rs
Erigara: no, i can remove it if you want

appetrosyan: Please remove.

ilchu (approved):

PR #2807, opened 2022-09-29
[fix]: Fix expression serialization bug

Arjentix:
### Description of the Change

`Expression` can cause stack overflow when serializing non-Raw variant. I missed it in my PR previously.
@Erigara has mentioned it and I fixed.

### Issue

No

### Benefits

- No recursion bug when serializing `Expression`

### Possible Drawbacks

- None

### Usage Examples or Tests

Probably we should add tests for it?


Erigara (dismissed):

appetrosyan (approved):

appetrosyan (approved):

data_model/src/expression.rs
```diff
@@ -400,6 +399,28 @@ macro_rules! expression_serde_internal_repr {
...
                 }
             }
+
+            pub mod reference {
+                use super::*;
+
+                #[derive(::serde::Serialize)]
+                pub enum $i <'re> {
+                    $(
+                        $(#[$var_meta])*
```
Erigara, line 409: Field attributes without enum attributes could lead to problems.

Erigara (changes requested):

data_model/src/expression.rs
Arjentix: Indeed. Then I brign back meta for the enum

Arjentix: Or better remove field attributes, because they cause some problems with lifetimes

Erigara (dismissed):

Arjentix:
CI fails in tests I couldn't touch by any chance, I believe

Erigara (approved):

appetrosyan (approved):

PR #2808, opened 2022-09-30
[refactor]: Add extra examples to `genesis.json`

appetrosyan:

### Description of the Change

Add extra examples which provide more context for SDK developers in case `genesis.json` syntax changes once again. 

### Benefits

Better examples

### Possible Drawbacks
None

``` cargo run --bin kagami -- genesis ```

### DIFF
--- a/core/src/genesis.rs
+++ b/core/src/genesis.rs
@@ -15,7 +15,10 @@ use eyre::{eyre, Result, WrapErr};
 use iroha_actor::Addr;
 use iroha_config::genesis::Configuration;
 use iroha_crypto::{KeyPair, PublicKey};
-use iroha_data_model::{asset::AssetDefinition, prelude::*};
+use iroha_data_model::{
+    asset::AssetDefinition,
+    prelude::{Metadata, *},
+};
 use iroha_primitives::small::{smallvec, SmallVec};
 use iroha_schema::prelude::*;
 use serde::{Deserialize, Serialize};
@@ -274,7 +277,7 @@ pub struct GenesisTransaction {
 }
 
 impl GenesisTransaction {
-    /// Convert `GenesisTransaction` into `AcceptedTransaction` with signature
+    /// Convert [`GenesisTransaction`] into [`AcceptedTransaction`] with signature
     ///
     /// # Errors
     /// Fails if signing or accepting fails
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: core/src/genesis.rs [hunk 1]
```diff
@@ -352,11 +352,20 @@ impl RawGenesisBlockBuilder {
...
     }
+
     /// Create a domain and return a domain builder which can
     /// be used to create assets and accounts.
-    pub fn domain(mut self, domain_name: Name) -> RawGenesisDomainBuilder {
+    pub fn domain(self, domain_name: Name) -> RawGenesisDomainBuilder {
+        self.domain_meta(domain_name, Default::default())
+    }
+
+    pub fn domain_meta(
```
mversic, line 362:
```suggestion
    pub fn domain_with_meta(
```

core/src/genesis.rs
```diff
@@ -395,9 +405,19 @@ impl RawGenesisDomainBuilder {
 
     /// Add an account to this domain
     #[must_use]
-    pub fn with_account(mut self, account_name: Name, public_key: PublicKey) -> Self {
+    pub fn with_account(self, account_name: Name, public_key: PublicKey) -> Self {
+        self.with_account_meta(account_name, public_key, Default::default())
+    }
+
+    pub fn with_account_meta(
```
mversic, line 412:
```suggestion
    pub fn account_with_meta(
```

I would name it like this, the previous name make me think we're adding metadata to an existing account instead of registering an account with metadata in one instruction

mversic (approved):

core/src/genesis.rs
Erigara, line 362: Missing docs

core/src/genesis.rs
```diff
@@ -352,11 +352,20 @@ impl RawGenesisBlockBuilder {
...
     /// be used to create assets and accounts.
-    pub fn domain(mut self, domain_name: Name) -> RawGenesisDomainBuilder {
+    pub fn domain(self, domain_name: Name) -> RawGenesisDomainBuilder {
+        self.domain_meta(domain_name, Default::default())
+    }
+
+    pub fn domain_meta(
+        mut self,
+        domain_name: Name,
+        metadata: iroha_data_model::prelude::Metadata,
```
Erigara, line 365: full path is redundant here because we already imported prelude on top of the module

core/src/genesis.rs
Erigara, line 412: should we add `#[must_use]` here as well and same about missing docs and Metadata path

appetrosyan: Better to add `[must_use]` to the type.

mversic (dismissed):

core/src/genesis.rs
```diff
@@ -384,8 +404,7 @@ impl RawGenesisDomainBuilder {
 
     /// Add an account to this domain without a public key.
     /// Should only be used for testing.
-    #[must_use]
-    pub fn with_account_without_public_key(mut self, account_name: Name) -> Self {
+    pub fn account_without_public_key(mut self, account_name: Name) -> Self {
```
mversic, line 407:
```suggestion
    #[cfg(test)]
    pub fn account_without_public_key(mut self, account_name: Name) -> Self {
```

Erigara (dismissed):

mversic (dismissed):

SamHSmith (approved):
LGTM

mversic (approved):

PR #2810, opened 2022-09-30
[fix] #2548: Warn on large genesis file

### LINKED ISSUE #2548: [BUG]: Zombie Genesis

Erigara:

### Description of the Change

Put warning in logs if genesis file is larger or equal 1Gb.

### Issue

Partially closes #2548. 

### Benefits

Notify uses.

### Possible Drawbacks

None.

Run iroha with genesis above 1Gb.

I take simplest possible approach, other designs could be considered.

### DIFF
--- a/core/src/genesis.rs
+++ b/core/src/genesis.rs
@@ -231,6 +231,8 @@ pub struct RawGenesisBlock {
 }
 
 impl RawGenesisBlock {
+    const WARN_ON_GENESIS_GTE: u64 = 1024 * 1024 * 1024; // 1Gb
+
     /// Construct a genesis block from a `.json` file at the specified
     /// path-like object.
     ///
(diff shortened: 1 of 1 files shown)


core/src/genesis.rs
```diff
@@ -260,13 +260,22 @@ pub struct RawGenesisBlock {
...
     ///
     /// # Errors
     /// If file not found or deserialization from file fails.
     pub fn from_path<P: AsRef<Path> + Debug>(path: P) -> Result<Self> {
         let file = File::open(&path).wrap_err(format!("Failed to open {:?}", &path))?;
+        let size = file
+            .metadata()
+            .wrap_err("Unable to access genesis file metadata")?
+            .len();
+        if size >= Self::WARN_ON_GENESIS_GTE {
```
mversic, line 276: are we checking file size or deserialized block size?

Erigara: file size

mversic (approved):

appetrosyan (approved):

PR #2822, opened 2022-10-04
[feature] #2797: Enrich events

### LINKED ISSUE #2797: [suggestion]: Enrich events with additional data
### Feature request

Enrich events to provide more detailed information inside  events.

### Motivation

This would be helpful to consumer of events to have more clear picture of whats doing on inside iroha.
E.g. currently there is no permission token id inside `PermissionAdded/Removed` event only `AccountId`.

### Who can help?

_No response_

Erigara:
### Description of the Change

Added struct that additional data to events.

### Issue

Closes #2797

### Benefits

More data provided to any consumer of event.

### Possible Drawbacks

Additional memory consumption.


data_model/src/events/data/events.rs
```diff
@@ -5,35 +5,38 @@ use iroha_primitives::small::SmallVec;
...
+/// Generic [`MetadataChanged`] struct.
+/// Depending on wrapping event could mean inserted or removed metadata `(key, value)` pair.
+#[derive(
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Decode, Encode, Deserialize, Serialize, IntoSchema,
+)]
+#[allow(missing_docs)]
+pub struct MetadataChanged<ID> {
+    pub origin_id: ID,
+    pub key: Name,
+    pub value: Box<Value>,
```
Erigara, line 17: Box to make enum variants equally sized.

Removing value could be considered since we could query this value, however removed value couldn't be observed this way.

I took approach by representing opposite events by single struct and distinguish what actually happens by wrapping event to reduce code duplication.

data_model/src/events/data/events.rs
```diff
@@ -5,35 +5,38 @@ use iroha_primitives::small::SmallVec;
...
+}
+
 mod asset {
     //! This module contains `AssetEvent`, `AssetDefinitionEvent` and its impls
 
     use super::*;
 
+    type AssetMetadataChanged = MetadataChanged<AssetId>;
```
Erigara, line 25: Required by `Filter` macro.

data_model/derive/src/filter.rs
```diff
@@ -205,8 +205,6 @@ pub fn impl_filter(event: &EventEnum) -> TokenStream {
         #[derive(
             Clone,
             PartialEq,
-            PartialOrd,
```
Erigara, line 208: We don't depend on this traits anywhere in code.

data_model/src/events/data/events.rs
```diff
@@ -5,35 +5,38 @@ use iroha_primitives::small::SmallVec;
 
 use super::*;
 
+/// Generic [`MetadataChanged`] struct.
+/// Depending on wrapping event could mean inserted or removed metadata `(key, value)` pair.
```
outoftardis, line 9:
```suggestion
/// Contains the changed metadata (`(key, value)` pair), either inserted or removed, which is determined by the wrapping event.
```

data_model/src/events/data/events.rs
```diff
@@ -87,11 +89,19 @@ mod asset {
...
                 | Self::MintabilityChanged(id)
-                | Self::MetadataInserted(id)
-                | Self::MetadataRemoved(id) => id,
+                | Self::MetadataInserted(MetadataChanged { origin_id: id, .. })
+                | Self::MetadataRemoved(MetadataChanged { origin_id: id, .. }) => id,
             }
         }
     }
+
+    /// [`AssetChangedBy`] depending on wrapping event represents added/removed asset quantity
```
outoftardis, line 98:
```suggestion
    /// Depending on the wrapping event, [`AssetChangedBy`] represents the added or removed asset quantity.
```

data_model/src/events/data/events.rs
```diff
@@ -316,23 +316,38 @@ mod account {
...
             }
         }
     }
-}
-
-mod domain {
-    //! This module contains `DomainEvent` and its impls
 
-    use super::*;
+    /// [`AccountPermissionChanged`] role depending on wrapping event represents added/removed account role
```
outoftardis, line 329:
```suggestion
    /// Depending on the wrapping event, [`AccountPermissionChanged`] role represents the added or removed account role
```

data_model/src/events/data/events.rs
```diff
@@ -316,23 +316,38 @@ mod account {
...
+        IntoSchema,
+    )]
+    #[allow(missing_docs)]
+    pub struct AccountPermissionChanged {
+        pub account_id: AccountId,
+        pub permission_id: PermissionTokenId,
+    }
 
-    /// Domain Event
+    /// [`AccountRoleChanged`] depending on wrapping event represent granted/revoked role
```
outoftardis, line 350:
```suggestion
    /// Depending on the wrapping event, [`AccountRoleChanged`] represents the granted or revoked role
```

data_model/src/events/data/events.rs
```diff
@@ -412,11 +445,32 @@ mod trigger {
...
                 | Self::Deleted(id)
-                | Self::Extended(id)
-                | Self::Shortened(id) => id,
+                | Self::Extended(TriggerNumberOfExecutionsChangedBy { trigger_id: id, .. })
+                | Self::Shortened(TriggerNumberOfExecutionsChangedBy { trigger_id: id, .. }) => id,
             }
         }
     }
+
+    /// [`TriggerNumberOfExecutionsChangedBy`] depending on wrapping event represents increased/decreased number of event executions
```
outoftardis, line 454:
```suggestion
    /// Depending on the wrapping event, [`TriggerNumberOfExecutionsChangedBy`] represents the increased or decreased number of event executions.
```

data_model/src/events/data/events.rs
```diff
@@ -5,35 +5,38 @@ use iroha_primitives::small::SmallVec;
 
 use super::*;
 
+/// Generic [`MetadataChanged`] struct.
```
outoftardis, line 8: note to self: add this to docs

core/src/smartcontracts/isi/account.rs
```diff
@@ -59,7 +59,12 @@ pub mod isi {
             wsv.modify_account(&account_id, |account| {
                 account
                     .remove_asset(&asset_id)
-                    .map(|asset| AccountEvent::Asset(AssetEvent::Removed(asset.id().clone())))
+                    .map(|asset| {
+                        AccountEvent::Asset(AssetEvent::Removed(AssetChangedBy {
+                            asset_id: asset.id().clone(),
+                            by: asset.value().clone(),
```
appetrosyan, line 65: `by` is ambiguous. I don't know off the top of my head what it means and need to dig into the source code to find out.

core/src/smartcontracts/isi/account.rs
```diff
@@ -153,12 +158,16 @@ pub mod isi {
...
-                    self.key,
-                    self.value,
+                    self.key.clone(),
+                    self.value.clone(),
                     account_metadata_limits,
                 )?;
 
-                Ok(AccountEvent::MetadataInserted(account_id.clone()))
+                Ok(AccountEvent::MetadataInserted(MetadataChanged {
+                    origin_id: account_id.clone(),
```
appetrosyan, line 167: `origin_id` being the account whose metadata is being modified, or the account which modified the metadata? It's not obvious from the name, so I suggest you rename it.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -174,7 +186,10 @@ pub mod isi {
                     .ok_or(MathError::Overflow)?;
                 metrics_arc.tx_amounts.observe((*quantity).into_metric());
 
-                Ok(AssetEvent::Added(asset_id.clone()))
+                Ok(AssetEvent::Added(AssetChangedBy {
+                    asset_id: asset_id.clone(),
+                    by: mint.object.into(),
```
appetrosyan, line 191: `by` is even more ambiguous in this context. Recommend `amount`.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -124,7 +124,10 @@ pub mod isi {
                     n.checked_add(self.object)
                         .ok_or(trigger::set::RepeatsOverflowError)
                 })?;
-                Ok(TriggerEvent::Extended(id))
+                Ok(TriggerEvent::Extended(TriggerNumberOfExecutionsChangedBy {
+                    trigger_id: id,
+                    by: self.object,
```
appetrosyan, line 129: Here `by` is not ambiguous.

data_model/derive/src/filter.rs
appetrosyan: Revert. 

In library crates if you *can* implement a trait, you should.

data_model/src/events/data/events.rs
appetrosyan: What `enum`? If this boxing is needed for a structure that houses this object, you should at least mention this object in the comments. Moreover, why isn't this entire comment a literal comment in-code?

data_model/src/events/data/events.rs
appetrosyan: Should be an inline comment.

data_model/src/events/data/events.rs
```diff
@@ -5,35 +5,38 @@ use iroha_primitives::small::SmallVec;
...
     #[allow(missing_docs)]
     pub enum AssetEvent {
         Created(AssetId),
         Deleted(AssetId),
-        Added(AssetId),
-        Removed(AssetId),
-        MetadataInserted(AssetId),
-        MetadataRemoved(AssetId),
+        Added(AssetChangedBy),
+        Removed(AssetChangedBy),
```
appetrosyan, line 37: `-By` suffix unnecessary.

data_model/src/events/data/events.rs
appetrosyan, line 98:
```suggestion
    /// [`Self`] depending on wrapping event represents added/removed asset quantity
```

data_model/src/events/data/events.rs
```diff
@@ -412,11 +445,32 @@ mod trigger {
...
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    #[allow(missing_docs)]
+    pub struct TriggerNumberOfExecutionsChangedBy {
+        pub trigger_id: TriggerId,
+        pub by: u32,
```
appetrosyan, line 472: Either rename the `by` or remove the `By` suffix. I know what you wanted to do, but `trigger_number_of_executions_changed_by.by` isn't pretty.

appetrosyan (commented):
Overall comment: 

Mostly cosmetic changes. Needs to be held back until after the RC9 release window closes.

data_model/src/events/data/events.rs
```diff
@@ -346,17 +361,35 @@ mod domain {
...
+
+    type DomainMetadataChanged = MetadataChanged<DomainId>;
+
+    /// Domain Event
+    #[derive(
+        Clone, PartialEq, Eq, Debug, Decode, Encode, Deserialize, Serialize, IntoSchema, Filter,
+    )]
+    #[non_exhaustive]
+    // TODO: fix variant size differences
+    #[allow(missing_docs, variant_size_differences)]
```
mversic, line 385: what about size differences?

data_model/src/events/data/events.rs
```diff
@@ -5,35 +5,38 @@ use iroha_primitives::small::SmallVec;
 
 use super::*;
 
+/// Generic [`MetadataChanged`] struct.
+/// Depending on wrapping event could mean inserted or removed metadata `(key, value)` pair.
+#[derive(
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Decode, Encode, Deserialize, Serialize, IntoSchema,
+)]
+#[allow(missing_docs)]
+pub struct MetadataChanged<ID> {
```
mversic, line 14:
```suggestion
pub struct MetadataChanged<ID: Identifiable> {
```

maybe?

data_model/src/events/data/events.rs
mversic: why do we care that they are equally sized? also we actually have a `ValueBox` type

data_model/src/events/data/events.rs
appetrosyan: If the trait bound isn't necessary it's better to have the structure as generic as possible.

Erigara: Initially it is what i do, but such approach poorly work with derive macroses, because it require us to manually put trait bounds onto `<ID as Identifiable>::Id`.

data_model/src/events/data/events.rs
Erigara: @mversic what do you mean?

core/src/smartcontracts/isi/account.rs
Erigara: @appetrosyan i adopted naming used in `HasOrigin` trait, origin in this context mean account whose metadata is being modified. 
So `origin_id` is id of entity which is modified during isi execution.

data_model/derive/src/filter.rs
Erigara: Reverted.

core/src/smartcontracts/isi/account.rs
appetrosyan: My comment stands. In this case a `target_account_id` is unambiguous, `origin_id` actually looks like the account that is doing the insertion. `HasOrigin`'s naming conventions can stay, because in that context, the origin makes sense (given that we're talking about events in general), while in this case, the naming should be clarified.

data_model/src/events/data/events.rs
appetrosyan: Boxing seems to be the most sensible option. `WorldEvent`s are released quite frequently, meaning that if we don't box, we'd be wasting a lot of extra space.

data_model/src/events/data/events.rs
appetrosyan: I think, he means that we should fix the variant size differences before merging.

core/src/smartcontracts/isi/account.rs
Erigara: I don't think that `target_accound_id` would fit because `MetadataInserted` is generic and used in domains, accounts, assets, ...
So maybe just `target_id`? What do you think?

data_model/src/events/data/events.rs
Erigara: Fixed

core/src/smartcontracts/isi/account.rs
Erigara: Decided to stick with `target_id`.

mversic (approved):

data_model/src/events/data/events.rs
```diff
@@ -87,11 +114,32 @@ mod asset {
...
+        Hash,
+        Debug,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    #[allow(missing_docs)]
+    pub struct AssetChanged {
```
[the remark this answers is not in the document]
Erigara: The difference is that this structure represents not total amount of assets someone has but difference.

mversic (approved):

appetrosyan (approved):

Issue #2379, opened 2022-06-21
Add plain-text block option to kura


appetrosyan:
Add option to save blocks in JSON/YAML. For debugging.

PR #2823, opened 2022-10-04
[fix] #2379: Ability to dump new kura blocks to blocks.txt.

SamHSmith:

### DIFF
--- a/config/src/kura.rs
+++ b/config/src/kura.rs
@@ -23,6 +23,8 @@ pub struct Configuration {
     pub blocks_per_storage_file: NonZeroU64,
     /// Default buffer capacity of actor's MPSC channel.
     pub actor_channel_capacity: u32,
+    /// Whether or not new blocks be outputted to a file called blocks.json.
+    pub debug_output_new_blocks: bool,
 }
 
 impl Default for ConfigurationProxy {
@@ -36,6 +38,7 @@ impl Default for ConfigurationProxy {
                     .expect("BLOCKS_PER_STORAGE cannot be set to a non-positive value."),
             ),
             actor_channel_capacity: Some(DEFAULT_ACTOR_CHANNEL_CAPACITY),
+            debug_output_new_blocks: Some(false),
         }
     }
 }
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -38,6 +38,8 @@ pub struct Kura {
     block_reciever: Mutex<Receiver<VersionedCommittedBlock>>,
     /// The sending actor for the [`VersionedCommittedBlock`]
     block_sender: Sender<VersionedCommittedBlock>,
+    /// Path to file for plain text blocks.
+    block_plain_text_path: Option<PathBuf>,
 }
 
 impl Kura {
@@ -54,6 +56,7 @@ impl Kura {
         block_store_path: &Path,
         // broker: Broker,
         block_channel_size: u32,
+        debug_output_new_blocks: bool,
     ) -> Result<Arc<Self>> {
         let (block_sender, block_reciever) = channel(
             block_channel_size
@@ -64,6 +67,12 @@ impl Kura {
         let mut block_store = StdFileBlockStore::new(block_store_path);
         block_store.create_files_if_they_do_not_exist()?;
 
+        let block_plain_text_path = debug_output_new_blocks.then(|| {
+            let mut path_buf = block_store_path.to_path_buf();
+            path_buf.push("blocks.json");
+            path_buf
+        });
+
         let kura = Arc::new(Self {
             mode,
             block_store: Mutex::new(Box::new(block_store)),
@@ -71,6 +80,7 @@ impl Kura {
             // broker,
             block_reciever: Mutex::new(block_reciever),
             block_sender,
+            block_plain_text_path,
         });
 
         Ok(kura)
@@ -104,6 +114,7 @@ impl Kura {
             Path::new(&configuration.block_store_path),
             // broker,
             configuration.actor_channel_capacity,
+            configuration.debug_output_new_blocks,
         )
     }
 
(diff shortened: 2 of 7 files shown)


core/src/kura.rs
```diff
@@ -137,15 +147,27 @@ impl Kura {
...
-                        blocks.push(decoded_block);
-                    }
-                    Err(error) => {
-                        error!(?error, "Encountered malformed block. Not reading any blocks beyond this height.");
-                        break;
-                    }
-                },
+                Ok(_) => {
+                    let decoded_block = if block_stored_as_plain_text {
+                        match serde_json::from_slice::<crate::CommittedBlock>(&block_data_buffer) {
```
mversic, line 152:
```suggestion
                        match serde_json::from_slice::<crate::VersionedCommittedBlock>(&block_data_buffer) {
```

shouldn't we store blocks as versioned? as we do for the binary format. What's the reasoning here?

mversic: as far as I see you're serializing it as versioned

core/src/kura.rs
```diff
@@ -340,8 +379,8 @@ pub trait BlockStoreTrait {
     ///
     /// # Errors
     /// IO Error.
-    fn read_block_index(&self, block_height: u64) -> Result<(u64, u64)> {
-        let mut index = (0, 0);
+    fn read_block_index(&self, block_height: u64) -> Result<(u64, u64, u64)> {
+        let mut index = (0, 0, 0);
```
mversic, line 383: you could be a smartass here and use the first bit of the block start to store the information about the format type. That is unless we'll support more serialization formats that we can change mid stream while blockchain is up

mversic (approved):

mversic (changes requested):
some remarks

config/src/kura.rs
```diff
@@ -23,6 +23,8 @@ pub struct Configuration {
     pub blocks_per_storage_file: NonZeroU64,
     /// Default buffer capacity of actor's MPSC channel.
     pub actor_channel_capacity: u32,
+    /// Store new blocks as plain text.
+    pub new_blocks_are_plain_text: bool,
```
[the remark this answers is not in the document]
SamHSmith: This is a bit deceiving. The Kura struct is mutable only during initialization and after that is naked inside an Arc. So it can't change while iroha is running. As for it being able to change between runs on the same block store, that I think is important. You may run in binary mode, but when you have an issue want to switch it to plain text.

core/src/kura.rs
SamHSmith: Very good of you to catch me on this. For some reason I thought it would be harder, but of course I just had to an extra derive.

SamHSmith: Switching to always storing and loading versioned.

core/src/kura.rs
SamHSmith: I considered that. But I wanted to preserve this ability, if you only ever store plain text blocks the `blocks.data` file is a continuous text file without any funny characters or invalid unicode. There is space in the block index now for additional formats and information. All 63 bits per block.

mversic (dismissed):

appetrosyan (dismissed):

Erigara:
Maybe with flag set we would put plaintext blocks in separate flag for debugging purposes?
Mixing binary and plaintext in single file seems to complicate solution unnecessary.
And we have kura inspector, so why do we want to have this feature at all?

appetrosyan:
> Maybe with flag set we would put plaintext blocks in separate file for debugging purposes? 

Agreed. That was the original intention. 

> And we have kura inspector, so why do we want to have this feature at all?

Nobody's using it on the SDK side. Frankly the UX leaves much to be desired.

core/src/kura.rs
```diff
@@ -30,6 +30,8 @@ pub struct Kura {
     #[allow(dead_code)]
     /// The mode of initialisation of [`Kura`].
     mode: Mode,
+    /// Should store new blocks as plain text.
```
outoftardis, line 33:
```suggestion
    /// Whether or not to store new blocks as plain text.
```

SamHSmith:
> Maybe with flag set we would put plaintext blocks in separate file for debugging purposes? Mixing binary and plaintext in single file seems to complicate solution unnecessary.

Having a seperate plaintext file that some of the blocks are in is more complicated in my opinion. At least when you consider someone reading it. The way I wrote it in this PR there are never doubts about the data you are reading being that which iroha is also reading.

SamHSmith:
I do as well think it is a good compliment to kura_inspector. Instead of running the blockchain and then printing the output to a file to search through, you can just have the data be json.

SamHSmith:
I retract my comments as it is apparently required that it be implemented the way you guys suggest.

config/src/kura.rs
```diff
@@ -23,6 +23,8 @@ pub struct Configuration {
     pub blocks_per_storage_file: NonZeroU64,
     /// Default buffer capacity of actor's MPSC channel.
     pub actor_channel_capacity: u32,
+    /// Whether or not new blocks be outputted to a file called blocks.txt.
```
appetrosyan, line 26:
```suggestion
    /// Whether or not new blocks be outputted to a file called blocks.json.
```

core/src/kura.rs
```diff
@@ -64,13 +67,22 @@ impl Kura {
         let mut block_store = StdFileBlockStore::new(block_store_path);
         block_store.create_files_if_they_do_not_exist()?;
 
+        let block_plain_text_path = if debug_output_new_blocks {
+            let mut path_buf = block_store_path.to_path_buf();
+            path_buf.push("blocks.txt");
```
appetrosyan, line 72:
```suggestion
            path_buf.push("blocks.json");
```

appetrosyan (approved):

mversic (approved):

mversic (approved):

appetrosyan (approved):

core/src/kura.rs
```diff
@@ -194,6 +205,17 @@ impl Kura {
...
                     }
 
+                    if let Some(path) = kura.block_plain_text_path.as_ref() {
+                        let mut plain_text_file = std::fs::OpenOptions::new()
+                            .create(true)
+                            .append(true)
+                            .open(path)
+                            .expect("Couldn't create file for plain text blocks.");
+
+                        serde_json::to_writer_pretty(&mut plain_text_file, &new_block)
```
Erigara, line 215: I want to notice one thing: despite this file called `blocks.json` it is not valid json file, because it has the following structure:
```
{ BLOCK 1}
{ BLOCK 2}
{ ...... }
{ BLOCK N}
```

However converting it to valid `json` is rather straightforward postprocessing: 
```
[
{ BLOCK 1},
{ BLOCK 2},
{ ...... },
{ BLOCK N}
]
```

PR #2839, opened 2022-10-06
[feature] #2725: Reintroduce timeout into `submit_transaction_blocking`

### LINKED ISSUE #2725: [feature] Client: bring back timeout when listening for events
After merge of #2724 timeout will be removed. We need to bring it back. This may require using non-blocking `tokio::TcpStream` or something like this. See [this discussion](https://github.com/snapview/tungstenite-rs/issues/14) for more info

Erigara:

### Issue

Closes #2725

### Benefits

- Timeout working again;
- No deadlock. 

### Possible Drawbacks

Some transactions can fail.

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -5,13 +5,12 @@
     clippy::std_instead_of_core,
     clippy::std_instead_of_alloc
 )]
-use std::{
-    collections::HashMap, fmt::Debug, marker::PhantomData, sync::mpsc, thread, time::Duration,
-};
+use std::{collections::HashMap, fmt::Debug, marker::PhantomData, thread, time::Duration};
 
 use derive_more::{DebugCustom, Display};
 use eyre::{eyre, Result, WrapErr};
-use http_default::WebSocketStream;
+use futures_util::StreamExt;
+use http_default::{AsyncWebSocketStream, WebSocketStream};
 use iroha_config::{client::Configuration, torii::uri, GetConfiguration, PostConfiguration};
 use iroha_core::smartcontracts::isi::query::Error as QueryError;
 use iroha_crypto::{HashOf, KeyPair};
@@ -24,6 +23,7 @@ use parity_scale_codec::DecodeAll;
 use rand::Rng;
 use serde::de::DeserializeOwned;
 
+use self::events_api::AsyncEventStream;
 use crate::{
     http::{Method as HttpMethod, RequestBuilder, Response, StatusCode},
     http_default::{self, DefaultRequestBuilder, WebSocketError, WebSocketMessage},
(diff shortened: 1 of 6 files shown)


client/src/client.rs
```diff
@@ -480,18 +493,9 @@ impl Client {
...
         hash: HashOf<SignedTransaction>,
-        init_sender: &mpsc::Sender<()>,
     ) -> Result<HashOf<VersionedSignedTransaction>> {
-        let event_iterator = self
-            .listen_for_events(PipelineEventFilter::new().hash(hash.into()).into())
-            .wrap_err("Failed to establish event listener connection")?;
-
-        init_sender
-            .send(())
-            .wrap_err("Failed to send init message through init channel")?;
```
Erigara, line 493: This place cause deadlock previously. 
On error in `event_iterator` thread hangs indefinitely waiting for message and then we `join` this thread.

client/src/client.rs
```diff
@@ -480,18 +493,9 @@ impl Client {
...
-
         for event in event_iterator.flatten() {
```
Erigara, line 499: This place could be considered harmful too, because we are ignoring all errors and possibly stucked in infinite loop.

SamHSmith (approved):
Looks fine.

client/src/client.rs
```diff
@@ -1105,6 +1147,95 @@ pub mod events_api {
...
+                    }
+                    Err(err) => return Err(err.into()),
+                }
+            };
+            debug!("`EventIterator` created successfully");
+            Ok(Self { stream, handler })
+        }
+
+        /// Return next event
+        pub async fn next(&mut self) -> Option<Result<Event>> {
```
pesterev, line 1192: It seems to me it would be great to implement a more general [Stream](https://docs.rs/tokio/latest/tokio/stream/index.html) instead of a specific Iterator.

client/src/client.rs
```diff
@@ -452,22 +451,23 @@ impl Client {
         &self,
         transaction: SignedTransaction,
     ) -> Result<HashOf<VersionedSignedTransaction>> {
-        let client = self.clone();
-        let (init_sender, init_receiver) = mpsc::channel();
+        let (init_sender, init_receiver) = tokio::sync::oneshot::channel();
         let hash = transaction.hash();
 
         // TODO: use `std::thread::scope()` after update to Rust 1.63
```
ilchu, line 457: How about introducing this change here, as long as we're already touching that bit of code? Or rather, how much do we need to actually switch it to `std`?

client/src/client.rs
```diff
@@ -817,6 +845,18 @@ impl Client {
...
 
+    /// Connects through `WebSocket` to async listen for `Iroha` pipeline and data events.
+    ///
+    /// # Errors
+    /// Fails if subscribing to websocket fails
+    pub async fn listen_for_events_async(
+        &self,
+        event_filter: FilterBox,
+    ) -> Result<AsyncEventIterator> {
+        iroha_logger::trace!(?event_filter);
```
ilchu, line 856:
```suggestion
        iroha_logger::trace!(?event_filter, "Async listening with");
```

client/src/client.rs
```diff
@@ -482,26 +482,54 @@ impl Client {
         }
     }
 
-    // TODO: introduce timeout
     fn listen_for_tx_confirmation(
         &self,
+        init_sender: tokio::sync::oneshot::Sender<bool>,
```
ilchu, line 487: I'm a bit ambiguous on this wrapping a `bool`. From a type standpoint, the older variant of sending unit made more sense, as we don't distinguish true and false scenarios here.

Erigara: @ilchu, maybe i understand you wrong, but we handle this scenarios differently.

```rust
let _send_result = init_sender.send(event_iterator_result.is_ok());
```

In case error happened during initialization we send false to channel and don't submit transaction.

client/src/client.rs
Erigara: Sure, i will introduce this change.

> Or rather, how much do we need to actually switch it to std?

We will be able to remove crossbeam dependency from client.

client/src/client.rs
ilchu: Right, my bad, then 100% fine

client/src/client.rs
Erigara: I implemented `Stream` for `AsyncEventStream` as you suggested.

client/src/client.rs
```diff
@@ -1105,6 +1142,93 @@ pub mod events_api {
...
+    pub struct AsyncEventStream {
+        stream: AsyncWebSocketStream,
+        handler: flow::Events,
+    }
+
+    impl AsyncEventStream {
+        /// Constructs [`AsyncEventStream`] and sends the subscription request.
+        ///
+        /// # Errors
+        /// Fails if connecting and sending subscription to web socket fails
```
appetrosyan, line 1156:
```suggestion
        /// - Request failed to build
        /// - `connect_async` failed
        /// - sending failed
        /// - Message not received in stream during connection or subscription
        /// - Message is an error        
```

client/src/client.rs
```diff
@@ -1105,6 +1142,93 @@ pub mod events_api {
...
+                    .await
+                    .ok_or(eyre!("Expect to receive message"))?;
+                debug!(?mes, "Received message");
+                match mes {
+                    Ok(WebSocketMessage::Binary(message)) => break handler.message(message)?,
+                    Ok(_) => continue,
+                    Err(err) => return Err(err.into()),
+                }
+            };
+            debug!("`AsyncEventStream` created successfully");
```
appetrosyan, line 1179:
```suggestion
            trace!("`AsyncEventStream` created successfully");
```

client/src/client.rs
```diff
@@ -1105,6 +1142,93 @@ pub mod events_api {
...
+
+            let mut stream = req.build()?.connect_async().await?;
+            stream.send(WebSocketMessage::Binary(first_message)).await?;
+
+            let handler = loop {
+                let mes = stream
+                    .next()
+                    .await
+                    .ok_or(eyre!("Expect to receive message"))?;
+                debug!(?mes, "Received message");
```
appetrosyan, line 1172:
```suggestion
                trace!(?mes, "Received message");
```

client/src/client.rs
```diff
@@ -1105,6 +1142,93 @@ pub mod events_api {
...
+                    Err(err) => return Err(err.into()),
+                }
+            };
+            debug!("`AsyncEventStream` created successfully");
+            Ok(Self { stream, handler })
+        }
+
+        /// Close websocket
+        /// # Errors
+        /// Fail if closing underlying websocket fails
```
appetrosyan, line 1185:
```suggestion
        /// - Server fails to send `Close` message
        /// - Closing the websocket connection itself fails.
```

client/src/client.rs
```diff
@@ -817,6 +840,18 @@ impl Client {
         events_api::EventIterator::new(self.events_handler(event_filter)?)
     }
 
+    /// Connects through `WebSocket` to async listen for `Iroha` pipeline and data events.
```
appetrosyan, line 843:
```suggestion
    /// Connect (through `WebSocket`) to `async` listen for `Iroha` `pipeline` and `data` events.
```

client/src/client.rs
```diff
@@ -817,6 +840,18 @@ impl Client {
...
+    ///
+    /// # Errors
+    /// Fails if subscribing to websocket fails
```
appetrosyan, line 846:
```suggestion
    /// - Forwards from [`Self::events_handler`]
```

appetrosyan: When you do that, the person interested in knowing why it could fail, can just look at the implementation and see the full list of error conditions.

client/src/client.rs
```diff
@@ -452,56 +452,79 @@ impl Client {
...
-        init_sender
-            .send(())
-            .wrap_err("Failed to send init message through init channel")?;
+            let result = tokio::time::timeout_at(
+                deadline,
+                Self::listen_for_tx_confirmation_loop(&mut event_iterator, hash),
+            )
+            .await
+            .map_err(Into::into)
+            .and_then(std::convert::identity);
```
appetrosyan, line 512: Odd construction.

client/src/client.rs
```diff
@@ -452,56 +452,79 @@ impl Client {
...
 
-        // TODO: use `std::thread::scope()` after update to Rust 1.63
-        let scope_res = crossbeam::scope(|spawner| {
-            let submitter_handle = spawner.spawn(move |_| -> Result<()> {
-                init_receiver
-                    .recv()
-                    .wrap_err("Failed to receive init message.")?;
-
-                self.submit_transaction(transaction)?;
+        std::thread::scope(|spawner| {
```
appetrosyan, line 458: Minor point: `std::thread` is already imported.

appetrosyan (commented):
Overall comment on logging. `trace!` should be used, when we want to delineate when a specific process or part of the process is initiated. Starting a handshake, receiving data, and so on can be considered a `trace` event. 

`debug!` is used for information that is used for debugging. Don't use it before a fallible operation, especially if you can just add the data in `fmt::Debug` format to the actual error message. Instead use `debug!` to print the internal state, where looking at the internal state teaches you something. For example, the number of peers that are connected at this time.

client/src/client.rs
Erigara: Intention here is to flatten `Result`.
1. initially we have `Result<Result<T, Report>, E>`
2. by `.map_err(Into::into)` we convert `E` into `Report`
3. by `.and_then(std::convert::identity)` we flatten result from `Result<Result<T, Report>, Report>` into `Result<T, Report>` (i take inspiration from nightly api [Result::flatten](https://doc.rust-lang.org/src/core/result.rs.html#1769))

appetrosyan (approved):

ilchu (dismissed):

appetrosyan (approved):

PR #2843, opened 2022-10-06
[fix] #2819: Move non-essential members out of WSV.

### LINKED ISSUE #2819: Move WSV members into Sumeragi

SamHSmith:
### Description of the Change

Consolidate structure of `WorldStateView` object. Remove unnecessary duplication and allow things to be accessed from a more appropriate object. 

### Issue

Closes #2819 

### Benefits

Smaller memory footprint. 

### Possible Drawbacks

None

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -413,8 +413,7 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -
         iroha_logger::error!(%error, "Error while calling sumeragi::update_metrics.");
     }
     sumeragi
-        .wsv_mutex_access()
-        .metrics
+        .metrics_mutex_access()
         .try_to_string()
         .map_err(Error::Prometheus)
 }
@@ -426,7 +425,7 @@ async fn handle_status(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) ->
     if let Err(error) = sumeragi.update_metrics() {
         iroha_logger::error!(%error, "Error while calling `sumeragi::update_metrics`.");
     }
-    let status = Status::from(&sumeragi.wsv_mutex_access().metrics);
+    let status = Status::from(&sumeragi.metrics_mutex_access());
     Ok(reply::json(&status))
 }
 
--- a/core/src/smartcontracts/isi/asset.rs
+++ b/core/src/smartcontracts/isi/asset.rs
@@ -12,7 +12,6 @@ use super::prelude::*;
 /// - update metadata
 /// - transfer, etc.
 pub mod isi {
-    use std::sync::Arc;
 
     use super::*;
 
@@ -163,7 +162,6 @@ pub mod isi {
                 &asset_id,
                 <Self as AssetInstructionInfo>::DEFAULT_ASSET_VALUE,
             )?;
-            let metrics_arc = Arc::clone(&wsv.metrics);
             wsv.modify_asset(&asset_id, |asset| {
                 let quantity: &mut Self = asset
                     .try_as_mut()
@@ -172,7 +170,11 @@ pub mod isi {
                 *quantity = quantity
                     .checked_add(mint.object)
                     .ok_or(MathError::Overflow)?;
-                metrics_arc.tx_amounts.observe((*quantity).into_metric());
+                #[allow(clippy::float_arithmetic)]
+                wsv.metric_tx_amounts
+                    .set(wsv.metric_tx_amounts.get() + (*quantity).into_metric());
+                wsv.metric_tx_amounts_counter
+                    .set(wsv.metric_tx_amounts_counter.get() + 1);
 
                 Ok(AssetEvent::Added(asset_id.clone()))
             })?;
@@ -201,7 +203,6 @@ pub mod isi {
                 wsv,
                 <Self as AssetInstructionInfo>::EXPECTED_VALUE_TYPE,
             )?;
-            let metrics_arc = Arc::clone(&wsv.metrics);
             wsv.modify_asset(&asset_id, |asset| {
                 let quantity: &mut Self = asset
                     .try_as_mut()
@@ -210,7 +211,11 @@ pub mod isi {
                 *quantity = quantity
                     .checked_sub(burn.object)
                     .ok_or(MathError::NotEnoughQuantity)?;
-                metrics_arc.tx_amounts.observe((*quantity).into_metric());
+                #[allow(clippy::float_arithmetic)]
+                wsv.metric_tx_amounts
+                    .set(wsv.metric_tx_amounts.get() + (*quantity).into_metric());
+                wsv.metric_tx_amounts_counter
+                    .set(wsv.metric_tx_amounts_counter.get() + 1);
 
                 Ok(AssetEvent::Removed(asset_id.clone()))
             })?;
@@ -262,7 +267,11 @@ pub mod isi {
                 *quantity = quantity
                     .checked_add(transfer.object)
                     .ok_or(MathError::Overflow)?;
-                wsv.metrics.tx_amounts.observe((*quantity).into_metric());
+                #[allow(clippy::float_arithmetic)]
+                wsv.metric_tx_amounts
+                    .set(wsv.metric_tx_amounts.get() + (*quantity).into_metric());
+                wsv.metric_tx_amounts_counter
+                    .set(wsv.metric_tx_amounts_counter.get() + 1);
 
                 Ok(AssetEvent::Added(transfer.destination_id.clone()))
             })?;
--- a/core/src/smartcontracts/isi/world.rs
+++ b/core/src/smartcontracts/isi/world.rs
@@ -85,7 +85,6 @@ pub mod isi {
                 Ok(DomainEvent::Created(domain_id).into())
             })?;
 
-            wsv.metrics.domains.inc();
             Ok(())
         }
     }
@@ -109,7 +108,6 @@ pub mod isi {
                 Ok(DomainEvent::Deleted(domain_id).into())
             })?;
 
-            wsv.metrics.domains.dec();
             Ok(())
         }
     }
--- a/core/src/sumeragi/mod.rs
+++ b/core/src/sumeragi/mod.rs
@@ -21,6 +21,7 @@ use iroha_crypto::{HashOf, KeyPair, SignatureOf};
 use iroha_data_model::prelude::*;
 use iroha_logger::prelude::*;
 use iroha_p2p::{ConnectPeer, DisconnectPeer};
+use iroha_telemetry::metrics::Metrics;
 use network_topology::{Role, Topology};
 
 use crate::{genesis::GenesisNetwork, handler::ThreadHandler};
@@ -55,16 +56,30 @@ trait Consensus {
     ) -> Option<VersionedPendingBlock>;
 }
 
+/*
+The values in the following struct are not atomics because the code that
+operates on them assumes their values does not change during the course of
+the function.
+*/
+#[derive(Debug)]
+struct LastUpdateMetricsData {
+    block_height: u64,
+    metric_tx_amounts: f64,
+    metric_tx_amounts_counter: u64,
+}
+
 /// `Sumeragi` is the implementation of the consensus.
 #[derive(Debug)]
 pub struct Sumeragi {
     internal: SumeragiWithFault<NoFault>,
     config: Configuration,
+    metrics_mutex: Mutex<Metrics>,
+    last_update_metrics_mutex: Mutex<LastUpdateMetricsData>,
 }
 
 impl Sumeragi {
     /// Construct [`Sumeragi`].
-    #[allow(clippy::too_many_arguments)]
+    #[allow(clippy::too_many_arguments, clippy::mutex_integer)]
     pub fn new(
         configuration: &Configuration,
         events_sender: EventsSender,
@@ -102,6 +117,12 @@ impl Sumeragi {
                 message_receiver: Mutex::new(incoming_message_receiver),
             },
             config: configuration.clone(),
+            metrics_mutex: Mutex::new(Metrics::default()),
+            last_update_metrics_mutex: Mutex::new(LastUpdateMetricsData {
+                block_height: 0,
+                metric_tx_amounts: 0.0_f64,
+                metric_tx_amounts_counter: 0,
+            }),
         }
     }
 
@@ -112,7 +133,13 @@ impl Sumeragi {
     ///
     /// # Panics
     /// - If either mutex is poisoned
-    #[allow(clippy::expect_used, clippy::unwrap_in_result)]
+    #[allow(
+        clippy::expect_used,
+        clippy::unwrap_in_result,
+        clippy::cast_precision_loss,
+        clippy::float_arithmetic,
+        clippy::mutex_integer
+    )]
     pub fn update_metrics(&self) -> Result<()> {
         let online_peers_count: u64 = self
             .internal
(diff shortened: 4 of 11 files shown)


client/benches/tps/lib.rs
```diff
@@ -203,7 +203,9 @@ impl MeasurerUnit {
         thread::spawn(move || -> Result<()> {
             for instruction in instructions {
                 submitter.submit(instruction)?;
-                thread::sleep(core::time::Duration::from_micros(interval_us_per_tx));
+                if interval_us_per_tx > 0 {
+                    thread::sleep(core::time::Duration::from_micros(interval_us_per_tx));
+                }
```
Erigara, line 208: From the docs to [`thread::sleep`](https://doc.rust-lang.org/stable/std/thread/fn.sleep.html):

> Currently, specifying a zero duration on Unix platforms returns immediately without invoking the underlying [nanosleep](https://linux.die.net/man/2/nanosleep) syscall, whereas on Windows platforms the underlying [Sleep](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep) syscall is always invoked.

Since we are mostly develop and run CI on unix systems seems that we could remove branching freely.

core/src/sumeragi/mod.rs
```diff
@@ -56,15 +57,20 @@ trait Consensus {
...
 /// `Sumeragi` is the implementation of the consensus.
+#[allow(clippy::mutex_integer)]
 #[derive(Debug)]
 pub struct Sumeragi {
     internal: SumeragiWithFault<NoFault>,
     config: Configuration,
+    metrics_mutex: Mutex<Metrics>,
+    last_block_height: Mutex<u64>,
+    last_metric_tx_amounts: Mutex<f64>,
+    last_metric_tx_amounts_counter: Mutex<u64>,
```
Erigara, line 68: Why not atomic?

core/src/sumeragi/mod.rs
```diff
@@ -102,6 +108,10 @@ impl Sumeragi {
                 message_receiver: Mutex::new(incoming_message_receiver),
             },
             config: configuration.clone(),
+            metrics_mutex: Mutex::new(Metrics::default()),
+            last_block_height: Mutex::new(0),
+            last_metric_tx_amounts            : Mutex::new(0.0_f64),
+            last_metric_tx_amounts_counter            : Mutex::new(0),
```
Erigara, line 114: Weird formatting.

core/src/wsv.rs
```diff
@@ -487,7 +442,7 @@ impl WorldStateView {
 
         for event in data_events {
             self.world.triggers.handle_data_event(event.clone());
-            self.produce_event(event);
+            self.events_buffer.borrow_mut().push(event.into());
```
Erigara, line 445: Maybe we could do this with single `borrow_mut().extend()`?

core/src/smartcontracts/isi/world.rs
```diff
@@ -85,7 +85,6 @@ pub mod isi {
                 Ok(DomainEvent::Created(domain_id).into())
             })?;
 
-            wsv.metrics.domains.inc();
```
Erigara, line 88: We are tracking this somewhere else?

client/benches/tps/lib.rs
SamHSmith: You say that. But my testing suggested otherwise.

core/src/sumeragi/mod.rs
SamHSmith: oh yea sorry.

core/src/wsv.rs
SamHSmith: Maybe? But doesn't that make the `.into()` harder?

core/src/smartcontracts/isi/world.rs
SamHSmith: `core/src/sumeragi/mod.rs:184`

core/src/sumeragi/mod.rs
SamHSmith: To me, this type of problem is not what "atomics are for". If you see that it's a mutex you directly understand the type of interaction that is going on. Maybe these three should be consolidated into one struct that sits inside the mutex instead.

client/benches/tps/lib.rs
Erigara: It's not me saying this, i take this directly from the official docs.

core/src/wsv.rs
Erigara: .map(Into::into) should do the job

core/src/sumeragi/mod.rs
Erigara: Ok, lets wait for other reviewers opinions on this one
Initially [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#arithmetic) brought my attention to this one.

appetrosyan: > To me, this type of problem is not what "atomics are for"
I'm afraid this is precisely what atomics are for, synchronised interior mutability of shared data. 

All you do by wrapping these objects into individual mutexes is confuse the reader. 

- If you had a structure behind a mutex like you suggested, the reader knows that having partially updated the variables inside the structure is inconsistent state. `RWLock` is the exact same thing, but more complex and useful if you read frequently, and write infrequently. 

- If you have three atomics, it means just that: you need to mutate the numbers, and you share the data between different threads. Data is completely unrelated. 

- If you have a RefCell, that means that data races are impossible. All you want is to make the borrow checker happy.

appetrosyan: So please @SamHSmith consider using atomics and using comments to convey the intent, instead of $6 \times$ the amount of space that's actually needed.

cli/src/torii/routing.rs
```diff
@@ -426,7 +425,7 @@ async fn handle_status(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) ->
     if let Err(error) = sumeragi.update_metrics() {
         iroha_logger::error!(%error, "Error while calling `sumeragi::update_metrics`.");
     }
-    let status = Status::from(&sumeragi.wsv_mutex_access().metrics);
+    let status = Status::from(&sumeragi.metrics_mutex_access());
```
appetrosyan, line 428: Please consider using atomics.

client/benches/tps/lib.rs
appetrosyan, line 208: I wonder if both this and `thread::sleep(0_duration)` optimises to the same thing.

core/src/sumeragi/mod.rs
```diff
@@ -151,6 +221,14 @@ impl Sumeragi {
         Ok(())
     }
 
+    /// Access node metrics.
+    #[allow(clippy::expect_used)]
+    pub fn metrics_mutex_access(&self) -> std::sync::MutexGuard<Metrics> {
+        self.metrics_mutex
+            .lock()
+            .expect("Couldn't lock mutex in `metrics_mutex_access`")
```
appetrosyan, line 229:
```suggestion
            .expect("`Mutex` in `metrics_mutex_access` poisoned. This should not happen, given that panics should stop Iroha. ")
```

client/benches/tps/lib.rs
appetrosyan: Nope. https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:10,endLineNumber:3,positionColumn:10,positionLineNumber:3,selectionStartColumn:10,selectionStartLineNumber:3,startColumn:10,startLineNumber:3),source:'//+Type+your+code+here,+or+load+an+example.%0Apub+fn+sleep_if(num:+u64)+-%3E+i32+%7B%0A++++if+num+%3E+0+%7B%0A++++++++std::thread::sleep(core::time::Duration::from_micros(num))%0A++++%7D%3B%0A++++0%0A%7D%0A%0A//+Type+your+code+here,+or+load+an+example.%0Apub+fn+sleep_no_if(num:+u64)+-%3E+i32+%7B%0A++++std::thread::sleep(core::time::Duration::from_micros(num))%3B%0A++++0%0A%7D%0A%0A//+If+you+use+%60main()%60,+declare+it+as+%60pub%60+to+see+it+in+the+output:%0A//+pub+fn+main()+%7B+...+%7D%0A'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1640,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-O',selection:(endColumn:1,endLineNumber:15,positionColumn:1,positionLineNumber:15,selectionStartColumn:1,selectionStartLineNumber:14,startColumn:1,startLineNumber:14),source:1,tree:'1'),l:'5',n:'0',o:'rustc+1.64.0+(Rust,+Editor+%231,+Compiler+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'rustc+1.64.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.64.0+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

appetrosyan: Please remove the conditional. As per [the official docs ](https://doc.rust-lang.org/std/thread/fn.sleep.html), the behaviour is exactly the same, and the call never gets inlined. 

> Currently, specifying a zero duration on Unix platforms returns immediately without invoking the underlying [nanosleep](https://linux.die.net/man/2/nanosleep) syscall, whereas on Windows platforms the underlying [Sleep](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep) syscall is always invoked. If the intention is to yield the current time-slice you may want to use [yield_now](https://doc.rust-lang.org/std/thread/fn.yield_now.html) instead.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -172,7 +170,11 @@ pub mod isi {
                 *quantity = quantity
                     .checked_add(mint.object)
                     .ok_or(MathError::Overflow)?;
-                metrics_arc.tx_amounts.observe((*quantity).into_metric());
+                #[allow(clippy::float_arithmetic)]
```
appetrosyan, line 173: This doesn't look particularly efficient. How is this better than what we had before?

core/src/sumeragi/mod.rs
```diff
@@ -127,22 +143,76 @@ impl Sumeragi {
...
+                metrics_guard
+                    .txs
+                    .with_label_values(&["total"])
+                    .inc_by(block_txs_accepted + block_txs_rejected);
+                metrics_guard.block_height.inc();
+            }
+            *last_block_height_guard = wsv_guard.height();
+        }
+
+        metrics_guard.domains.set(wsv_guard.domains().len() as u64);
```
appetrosyan, line 184: I'd prefer this be computed on-demand in response to a query.

core/src/wsv.rs
```diff
@@ -101,12 +93,12 @@ pub struct WorldStateView {
     blocks: Arc<Chain>,
     /// Hashes of transactions
     pub transactions: DashSet<HashOf<VersionedSignedTransaction>>,
-    /// Metrics for prometheus endpoint.
-    pub metrics: Arc<Metrics>,
-    /// Notifies subscribers when new block is applied
-    new_block_notifier: Arc<NewBlockNotificationSender>,
-    /// Transmitter to broadcast [`WorldStateView`]-related events.
-    events_sender: EventsSender,
+    /// Temporary buffer containing events generated during `WorldStateView::apply`.
```
ilchu, line 96: I feel saying that it's temporary is a bit confusing, and stating when it's updated would be beneficial.
```suggestion
    /// Buffer containing events generated during `WorldStateView::apply`. Renewed on every block commit.
```

core/src/sumeragi/mod.rs
```diff
@@ -127,22 +143,76 @@ impl Sumeragi {
...
+            .last_metric_tx_amounts
+            .lock()
+            .expect("Failed to lock on `last_metric_tx_amounts`. Mutex poisoned");
+        let mut last_metric_tx_amounts_counter_guard = self.last_metric_tx_amounts_counter.lock().expect("Failed to lock on `last_total_tx_amounts_counter_when_updating_metrics`. Mutex poisoned");
+
+        let diff_count =
+            wsv_guard.metric_tx_amounts_counter.get() - *last_metric_tx_amounts_counter_guard;
+        let diff_amount_per_count = (wsv_guard.metric_tx_amounts.get()
+            - *last_metric_tx_amounts_guard)
+            / (diff_count as f64);
```
ilchu, line 196: Here you're explicitly casting `diff_count`, but `wsv.metric_tx_amounts` is already `f64`, even though here's the only place where it's needed. I'd rather have it `u64` like everything else and cast it here as this field cannot be fractional by itself AFAICT.

core/src/sumeragi/mod.rs
```diff
@@ -127,22 +143,76 @@ impl Sumeragi {
...
+        for _ in 0..diff_count {
+            *last_metric_tx_amounts_counter_guard += 1;
+            *last_metric_tx_amounts_guard += diff_amount_per_count;
+
+            metrics_guard.tx_amounts.observe(diff_amount_per_count);
```
ilchu, line 201: Also a bit confused about the reasoning for this refactor here, and it seems that we're losing a bit of precision? Because instead of directly adding observations to the `Histogram` on each mint/burn/transfer like before, we aggregate them and then average out the `tx_amounts` difference across the whole window, instead of recording it precisely. Could you explain the purpose here?

client/benches/tps/lib.rs
SamHSmith: So I originally put the if there because of some benchmarks I ran. I was now unable to replicate them so I will remove the conditional.

core/src/sumeragi/mod.rs
SamHSmith: It is. `update_metrics` is called when there is a query.

core/src/sumeragi/mod.rs
```diff
@@ -127,22 +143,76 @@ impl Sumeragi {
...
+            *last_block_height_guard = wsv_guard.height();
+        }
+
+        metrics_guard.domains.set(wsv_guard.domains().len() as u64);
+
+        let mut last_metric_tx_amounts_guard = self
+            .last_metric_tx_amounts
+            .lock()
+            .expect("Failed to lock on `last_metric_tx_amounts`. Mutex poisoned");
+        let mut last_metric_tx_amounts_counter_guard = self.last_metric_tx_amounts_counter.lock().expect("Failed to lock on `last_total_tx_amounts_counter_when_updating_metrics`. Mutex poisoned");
```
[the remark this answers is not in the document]
SamHSmith: good catch

core/src/wsv.rs
SamHSmith: Good suggestion.

core/src/sumeragi/mod.rs
SamHSmith: The cast to f64 is required because otherwise the operation is `f64 / u64` which is not legal in rust.

core/src/sumeragi/mod.rs
SamHSmith: Internally `Histogram` does the same thing I am doing. It stores a total amount and the amount of observations. Except for some precision lost in the floating point maths it is the exact same value.

I would like to improve the precision of the gathered metrics but that was outside the scope of this PR. Since this PR is simply about decoupling the metrics from the WSV.

core/src/smartcontracts/isi/asset.rs
SamHSmith: Because previously you could poll the metrics and recieve information that includes a partially applied block. Doing it this way keeps the data local to a thread and it updates the metrics at a predictable time.

cli/src/torii/routing.rs
SamHSmith: I quite strongly disagree. You can't lock atomics. They can be changed from elsewhere while you are reading them.

core/src/sumeragi/mod.rs
SamHSmith: sure

core/src/sumeragi/mod.rs
SamHSmith: "A mutex muddies the intent. If you need just the interior mutability in a single-threaded context, use RefCell. If you need just the thread-safe separate modifications of the objects use a atomics. If you need them to be modified at the same time, and the values need to be in sync with each other, you need a structure."

I would prefer to have a struct inside a RefCell. But the struct lives inside an Arc so rust says no.

cli/src/torii/routing.rs
SamHSmith: You don't want another query to partially update the metrics while you are reading from them.

core/src/smartcontracts/isi/asset.rs
appetrosyan: Why couldn't you update the information when the data is requested instead? Lazy evaluation, allows you not to pay for metrics, when you don't use them, at the cost of doing more work when you do.

core/src/sumeragi/mod.rs
```diff
@@ -55,16 +56,31 @@ trait Consensus {
...
+*/
+#[derive(Debug)]
+struct LastUpdateMetricsData {
+    block_height: u64,
+    metric_tx_amounts: f64,
+    metric_tx_amounts_counter: u64,
+}
+
 /// `Sumeragi` is the implementation of the consensus.
+#[allow(clippy::mutex_integer)]
```
appetrosyan, line 72: 1. This should be enabled module-wide using `#![allow(…)`, given that you allow it almost everywhere it's used. 
2. This lint warns you very precisely about not using a `Mutex` wrapping an integer. I don't see an integer being wrapped, given that both `Metrics` and `LastUpdateMetricsData` is a `struct`. If this is a false-positive, reporting it would be nice.

core/src/sumeragi/mod.rs
```diff
@@ -102,6 +118,12 @@ impl Sumeragi {
                 message_receiver: Mutex::new(incoming_message_receiver),
             },
             config: configuration.clone(),
+            metrics_mutex: Mutex::new(Metrics::default()),
+            last_update_metrics_mutex: Mutex::new(LastUpdateMetricsData {
+                block_height: 0,
```
appetrosyan, line 122:
```suggestion
                block_height: 0_u64,
```

core/src/sumeragi/mod.rs
```diff
@@ -102,6 +118,12 @@ impl Sumeragi {
...
+                metric_tx_amounts: 0.0_f64,
+                metric_tx_amounts_counter: 0,
```
appetrosyan, line 124:
```suggestion
                metric_tx_amounts_counter: 0_u64,
```

core/src/sumeragi/mod.rs
SamHSmith: Oh yea sorry, I forgot to remove this one. Are there ways to detect unneeded allow statements?

core/src/wsv.rs
```diff
@@ -596,21 +555,15 @@ impl WorldStateView {
...
             world,
             config,
             transactions: DashSet::new(),
             blocks: Arc::new(Chain::new()),
-            metrics: Arc::new(Metrics::default()),
-            new_block_notifier: Arc::new(new_block_notifier),
-            events_sender,
+            events_buffer: std::cell::RefCell::new(Vec::new()),
+            metric_tx_amounts: std::cell::Cell::new(0.0_f64),
+            metric_tx_amounts_counter: std::cell::Cell::new(0),
```
appetrosyan, line 566: Doesn't look like we fully decoupled the behaviours. But I think it's OK for this PR.

core/src/sumeragi/mod.rs
appetrosyan: > Internally Histogram does the same thing I am doing. It stores a total amount and the amount of observations. Except for some precision lost in the floating point maths it is the exact same value.

Histograms always do that. The more precise approximation is called a kernel-density estimate, and we have neither the need, nor the time to implement it in Rust. An even more precise approach would be to return all of the transaction amounts given an immutable view of a blockchain, but I don't think I need to explain why this is a bad idea. 

> I would like to improve the precision of the gathered metrics but that was outside the scope of this PR. Since this PR is simply about decoupling the metrics from the WSV.

I don't think that's necessary for the time being. We might revisit this, once metrics are used in full force.

appetrosyan (approved):

core/src/smartcontracts/isi/asset.rs
appetrosyan: Thread locality sounds good. Given that we aggregate data, should be fine in general.

core/src/sumeragi/mod.rs
appetrosyan: > I would prefer to have a struct inside a RefCell. But the struct lives inside an Arc so rust says no.

It should be possible to remove the `Arc` and instead spawn a single async task to scan the `isi` and update the metrics, so not even a `RefCell` would be needed.

ilchu (dismissed):

core/src/wsv.rs
Erigara: Just idea but maybe metrics could be updated through events? Such approach seems to decouple metrics.

SamHSmith: No. That would basically the same as we had before. It's not strange that you can't fully decouple metrics from the thing they measure. The point is making the contents of the WSV less global in nature and more just data.

core/src/sumeragi/mod.rs
appetrosyan: I made that mistake too. Rust is strictly typed.

appetrosyan (approved):

core/src/wsv.rs
Erigara: > That would basically the same as we had before. 
In which way?

PR #2848, opened 2022-10-07
[ci]: Disable slow integration tests

6r1d:


SamHSmith (approved):
LGTM, but aleksandr might care about the missing newline at the end of `data_model/derive/Cargo.toml`.

client/tests/integration/roles.rs
```diff
@@ -12,6 +12,7 @@ use iroha_permissions_validators::public_blockchain::{
 };
 use test_network::*;
 
+#[ignore]
```
appetrosyan, line 15: Missed a spot

client/tests/integration/events/pipeline.rs
```diff
@@ -8,22 +8,26 @@ use test_network::*;
 
 use super::Configuration;
 
+#[allow(dead_code)]
```
appetrosyan, line 11: So the dead code shouldn't be removed, given that it should also be possible to re-enable these tests without re-compiling.

appetrosyan (dismissed):

appetrosyan:
> LGTM, but aleksandr might care about the missing newline at the end of data_model/derive/Cargo.toml.

Actually, newline at the end of file is an anti-pattern that I don't much care for.

client/tests/integration/roles.rs
appetrosyan:
```suggestion
#[ignore = "ignore, more in #2851"]
```

client/tests/integration/events/pipeline.rs
appetrosyan:
```suggestion
// Needed to re-enable ignored tests. 
#[allow(dead_code)]
```

PR #2861, opened 2022-10-12
Blocks on kura

SamHSmith:

### DIFF
--- a/cli/src/torii/mod.rs
+++ b/cli/src/torii/mod.rs
@@ -15,6 +15,7 @@ use iroha_core::{
     prelude::*,
     queue::{self, Queue},
     sumeragi::Sumeragi,
+    kura::Kura,
     EventsSender, IrohaNetwork,
 };
 use thiserror::Error;
@@ -40,6 +41,7 @@ pub struct Torii {
     network: iroha_actor::Addr<IrohaNetwork>,
     notify_shutdown: Arc<Notify>,
     sumeragi: Arc<Sumeragi>,
+    kura: Arc<Kura>,
 }
 
 /// Torii errors.
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -262,7 +262,7 @@ async fn handle_post_configuration(
 }
 
 #[iroha_futures::telemetry_future]
-async fn handle_blocks_stream(sumeragi: Arc<Sumeragi>, mut stream: WebSocket) -> eyre::Result<()> {
+async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::Result<()> {
     let subscription_request: VersionedBlockSubscriberMessage = stream.recv().await?;
     let mut from_height: u64 = subscription_request.into_v1().try_into()?;
 
@@ -273,37 +273,25 @@ async fn handle_blocks_stream(sumeragi: Arc<Sumeragi>, mut stream: WebSocket) ->
         .await?;
 
     loop {
-        let blocks = sumeragi.blocks_from_height(from_height.try_into().wrap_err("Failed to convert `from_height` into `usize`. You should consider upgrading to a 64-bit arch")?);
-        if blocks.is_empty() {
-            break;
+        match kura.get_block_by_height(from_height) {
+            None => break,
+            Some(block) => {
+                stream
+                    .send(VersionedBlockPublisherMessage::from(
+                        BlockPublisherMessage::from(VersionedCommittedBlock::clone(&block)), // TODO: Remove unnecessary clone.
+                    ))
+                    .await?;
+
+                let message: VersionedBlockSubscriberMessage = stream.recv().await?;
+                if let BlockSubscriberMessage::BlockReceived = message.into_v1() {
+                    from_height += 1;
+                } else {
+                    return Err(eyre!("Expected `BlockReceived` message"));
+                }
+            }
         }
-        stream_blocks(&mut from_height, blocks, &mut stream).await?;
         std::thread::sleep(std::time::Duration::from_millis(10));
     }
-    Ok(())
-}
-
-async fn stream_blocks(
-    from_height: &mut u64,
-    blocks: Vec<VersionedCommittedBlock>,
-    stream: &mut WebSocket,
-) -> eyre::Result<()> {
-    #[allow(clippy::expect_used)]
-    for block in blocks {
-        stream
-            .send(VersionedBlockPublisherMessage::from(
-                BlockPublisherMessage::from(block),
-            ))
-            .await?;
-
-        let message: VersionedBlockSubscriberMessage = stream.recv().await?;
-        if let BlockSubscriberMessage::BlockReceived = message.into_v1() {
-            *from_height += 1;
-        } else {
-            return Err(eyre!("Expected `BlockReceived` message"));
-        }
-    }
-
     Ok(())
 }
 
@@ -396,10 +384,8 @@ async fn handle_version(sumeragi: Arc<Sumeragi>) -> Json {
     #[allow(clippy::expect_used)]
     let string = sumeragi
         .wsv_mutex_access()
-        .blocks()
-        .last()
+        .latest_block_ref()
         .expect("Genesis not applied. Nothing we can do. Solve the issue and rerun.")
-        .value()
         .version()
         .to_string();
     reply::json(&string)
@@ -439,6 +425,7 @@ impl Torii {
         network: Addr<IrohaNetwork>,
         notify_shutdown: Arc<Notify>,
         sumeragi: Arc<Sumeragi>,
+        kura: Arc<Kura>,
     ) -> Self {
         Self {
             iroha_cfg,
@@ -448,6 +435,7 @@ impl Torii {
             network,
             notify_shutdown,
             sumeragi,
+            kura,
         }
     }
 
@@ -545,7 +533,7 @@ impl Torii {
             });
 
         let blocks_ws_router = block_ws_router_path
-            .and(add_state!(self.sumeragi))
+            .and(add_state!(self.kura))
             .and(warp::ws())
             .map(|sumeragi: Arc<_>, ws: Ws| {
                 ws.on_upgrade(|this_ws| async move {
--- a/client/benches/tps/lib.rs
+++ b/client/benches/tps/lib.rs
@@ -100,7 +100,7 @@ impl Config {
             .sumeragi
             .wsv_mutex_access()
             .clone();
-        let mut blocks = blocks_wsv.blocks().skip(blocks_out_of_measure as usize);
+        let mut blocks = blocks_wsv.all_blocks_by_value().into_iter().skip(blocks_out_of_measure as usize);
         let (txs_accepted, txs_rejected) = (0..self.blocks)
             .map(|_| {
                 let block = blocks
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -66,8 +66,9 @@ derive_more = "0.99.17"
 getset = "0.1.2"
 itertools = "0.10.3"
 
-[dev-dependencies]
 tempfile = "3.3.0"
+
+[dev-dependencies]
 criterion = "0.3.6"
 hex = "0.4.3"
 byte-unit = "4.0.14"
--- a/core/src/block_sync.rs
+++ b/core/src/block_sync.rs
@@ -16,12 +16,13 @@ use iroha_p2p::Post;
 use iroha_version::prelude::*;
 use parity_scale_codec::{Decode, Encode};
 
-use crate::{sumeragi::Sumeragi, NetworkMessage, VersionedCommittedBlock};
+use crate::{sumeragi::Sumeragi, kura::Kura, NetworkMessage, VersionedCommittedBlock};
 
 /// Structure responsible for block synchronization between peers.
 #[derive(Debug)]
 pub struct BlockSynchronizer {
     sumeragi: Arc<Sumeragi>,
+    kura: Arc<Kura>,
     peer_id: PeerId,
     gossip_period: Duration,
     block_batch_size: u32,
@@ -75,12 +76,14 @@ impl BlockSynchronizer {
     pub fn from_configuration(
         config: &Configuration,
         sumeragi: Arc<Sumeragi>,
+        kura: Arc<Kura>,
         peer_id: PeerId,
         broker: Broker,
     ) -> Self {
         Self {
             peer_id,
             sumeragi,
+            kura,
             gossip_period: Duration::from_millis(config.gossip_period_ms),
             block_batch_size: config.block_batch_size,
             broker,
@@ -181,11 +184,34 @@ pub mod message {
                         return;
                     }
 
-                    let mut blocks = block_sync.sumeragi.blocks_after_hash(*hash);
-                    blocks.truncate(block_sync.block_batch_size as usize);
+                    let mut h = 1;
+                    loop {
+                        match block_sync.kura.get_block_hash(h) {
+                            None => break,
+                            Some(hash) => {
+                                println!("{} : {}", h, hash);
+                                if h == 3 {
+                                    println!("POTATO");
+                                }
+                                h += 1;
+                            }
+                        }
+                    }
+                    let start_height = match block_sync.kura.get_block_height_by_hash(hash) {
+                        None => { warn!(%hash, "Block hash not found"); return; }
+                        Some(height) => height + 1, // It's get blocks *after*, so we add 1.
+                    };
+
+                    let mut blocks = Vec::new();
+                    for i in 0..(block_sync.block_batch_size as u64) {
+                        match block_sync.kura.get_block_by_height(start_height + i) {
+                            Some(block) => blocks.push(VersionedCommittedBlock::clone(&block)), // TODO: avoid this clone.
+                            None => break,
+                        }
+                    }
 
                     if blocks.is_empty() {
-                        warn!(%hash, "Block hash not found");
+                        error!(%hash, "Blocks array is empty but shouldn't be.");
                     } else {
                         trace!("Sharing blocks after hash: {}", hash);
                         Message::ShareBlocks(ShareBlocks::new(blocks, block_sync.peer_id.clone()))
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -17,7 +17,7 @@ use std::{
 };
 
 use iroha_config::kura::{Configuration, Mode};
-use iroha_crypto::HashOf;
+use iroha_crypto::{Hash, HashOf};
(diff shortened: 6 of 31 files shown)


core/src/wsv.rs
```diff
@@ -459,12 +461,8 @@ impl WorldStateView {
     }
 
     /// Returns iterator over blockchain blocks starting with the block of the given `height`
```
Erigara, line 463: Docs require update

permissions_validators/src/public_blockchain/transfer.rs
```diff
@@ -217,7 +217,8 @@ fn retrieve_count(params: &BTreeMap<Name, Value>) -> Result<u32> {
 fn count_executions(wsv: &WorldStateView, authority: &AccountId, period: Duration) -> usize {
     let period_start_ms = current_time().saturating_sub(period).as_millis();
 
-    wsv.blocks()
+    wsv.all_blocks_by_value()
+        .iter()                 // nocheckin, can be done better
```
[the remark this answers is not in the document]
appetrosyan: We don't have a `nocheckin` rule, but `todo!` will be linted, and `// TODO:` is universal language for "this is not the final version, I might need to revisit it later". This way, you also allow us to decide whether or not we merge as is.

cli/src/torii/routing.rs
```diff
@@ -262,7 +262,7 @@ async fn handle_post_configuration(
 }
 
 #[iroha_futures::telemetry_future]
-async fn handle_blocks_stream(sumeragi: Arc<Sumeragi>, mut stream: WebSocket) -> eyre::Result<()> {
+async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::Result<()> {
```
appetrosyan, line 265: You're taking an `Arc` by value. Meaning that you atomically increment the reference counter and then decrement it. Accepting a `Weak` might require you to call an `expect` once, but it removes code from the hot path. In reality, I'd argue that given that there's always precisely one `Kura`, it could also be a global static with lazy initialisation, and interior mutability. What do you think @Erigara ? 

Also, regarding function signature, consider accepting an `impl Deref<Target = Kura>` instead. This will also allow you to switch it out for a `Weak<Kura>`, or a reference to a static.

cli/src/torii/routing.rs
```diff
@@ -273,40 +273,28 @@ async fn handle_blocks_stream(sumeragi: Arc<Sumeragi>, mut stream: WebSocket) ->
         .await?;
 
     loop {
```
appetrosyan, line 275: Consider a `while let` loop. 

https://doc.rust-lang.org/rust-by-example/flow_control/while_let.html

cli/src/torii/routing.rs
```diff
@@ -273,40 +273,28 @@ async fn handle_blocks_stream(sumeragi: Arc<Sumeragi>, mut stream: WebSocket) ->
...
+                    .send(VersionedBlockPublisherMessage::from(
+                        BlockPublisherMessage::from(VersionedCommittedBlock::clone(&block)), // TODO: Remove unnecessary clone.
+                    ))
+                    .await?;
+
+                let message: VersionedBlockSubscriberMessage = stream.recv().await?;
+                if let BlockSubscriberMessage::BlockReceived = message.into_v1() {
+                    from_height += 1;
+                } else {
+                    return Err(eyre!("Expected `BlockReceived` message"));
```
appetrosyan, line 289: Rewrite `if let` as a match. Match onto `BlockReceived` in one block, name the other `message_as_v1` and 

```suggestion
                    eyre::bail!("Expected `BlockReceived` message, got {}", message_as_v1)));
```

appetrosyan: We need as much information as possible. Given that errors shouldn't happen frequently, adding a bit of extra info doesn't hurt.

cli/src/torii/routing.rs
```diff
@@ -396,10 +384,8 @@ async fn handle_version(sumeragi: Arc<Sumeragi>) -> Json {
     #[allow(clippy::expect_used)]
     let string = sumeragi
         .wsv_mutex_access()
-        .blocks()
-        .last()
+        .latest_block_ref()
```
appetrosyan, line 387: Such granularity is fine here, but in general if you're writing a library (and all crates outside tools and `client_cli` are), the previous syntax is preferable. The compiler is smart enough to treat both appropriately.

client/benches/tps/lib.rs
```diff
@@ -100,7 +100,7 @@ impl Config {
             .sumeragi
             .wsv_mutex_access()
             .clone();
-        let mut blocks = blocks_wsv.blocks().skip(blocks_out_of_measure as usize);
+        let mut blocks = blocks_wsv.all_blocks_by_value().into_iter().skip(blocks_out_of_measure as usize);
```
appetrosyan, line 103:
```suggestion
        let mut blocks = blocks_wsv.blocks().into_iter().skip(blocks_out_of_measure as usize);
```

The `by_value` part is implied by `into_iter()`, and the fact that there are `all` blocks is implied by the fact that you apply `filter`, `skip` and other `iterator` methods after.

core/Cargo.toml
```diff
@@ -66,8 +66,9 @@ derive_more = "0.99.17"
 getset = "0.1.2"
 itertools = "0.10.3"
 
-[dev-dependencies]
 tempfile = "3.3.0"
```
appetrosyan, line 69: We need `tempfile` on a regular basis now?

core/src/block_sync.rs
```diff
@@ -181,11 +184,34 @@ pub mod message {
                         return;
                     }
 
-                    let mut blocks = block_sync.sumeragi.blocks_after_hash(*hash);
-                    blocks.truncate(block_sync.block_batch_size as usize);
+                    let mut h = 1;
```
appetrosyan, line 187: This is very much unreadable. First of all, `h` is completely opaque. Secondly, unless the person reading this sees the `println` statement, they wouldn't know that `h` stands for `height`.

core/src/block_sync.rs
```diff
@@ -181,11 +184,34 @@ pub mod message {
...
+                    loop {
```
appetrosyan, line 188: Consider separating it out. This looks either as a `while let` loop, but I'm not sure (and that's the problem), or a `for` loop over `h`, with a few premature breaks. 

In either case, splitting this function into smaller chunks, with a lower level of nesting would help.

core/src/block_sync.rs
```diff
@@ -181,11 +184,34 @@ pub mod message {
...
+                                println!("{} : {}", h, hash);
+                                if h == 3 {
+                                    println!("POTATO");
+                                }
+                                h += 1;
+                            }
+                        }
+                    }
+                    let start_height = match block_sync.kura.get_block_height_by_hash(hash) {
+                        None => { warn!(%hash, "Block hash not found"); return; }
```
appetrosyan, line 201:
```suggestion
                        None => { error!(%hash, "Block hash not found"); return; }
```

appetrosyan (changes requested):

core/src/kura.rs
```diff
@@ -34,6 +35,7 @@ pub struct Kura {
     block_store: Mutex<Box<dyn BlockStoreTrait + Send>>,
     /// The array of block hashes. This is normally recovered from the index file.
     block_hash_array: Mutex<Vec<HashOf<VersionedCommittedBlock>>>,
```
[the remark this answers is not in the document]
appetrosyan: They're basically already hashed, so I wonder if we could use the identity function as the hash function?

core/src/kura.rs
```diff
@@ -77,7 +78,7 @@ impl Kura {
             mode,
             block_store: Mutex::new(Box::new(block_store)),
             block_hash_array: Mutex::new(Vec::new()),
-            // broker,
+            block_arc_array: Mutex::new(Vec::new()),
```
appetrosyan, line 81: Consider `blocks` and `block_hashes`, given that the implementation detail of the backing storage is irrelevant.

core/src/kura.rs
```diff
@@ -86,6 +87,14 @@ impl Kura {
         Ok(kura)
     }
 
+    #[allow(clippy::unwrap)]
```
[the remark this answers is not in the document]
appetrosyan: And return `tempdir` to the status of a `dev-dependency`.

core/src/kura.rs
```diff
@@ -86,6 +87,14 @@ impl Kura {
...
+    pub fn blank_kura_for_testing() -> (Arc<Kura>, ThreadHandler, tempfile::TempDir) {
+        let dir = tempfile::tempdir().unwrap();
+        let kura = Self::new(Mode::Strict, dir.path(), 100, false).unwrap();
+        let th = Self::start(kura.clone());
```
appetrosyan, line 94: While I can see it from the type, you should try and make the naming of things as clear as possible. `thread_handle` is sufficient.

core/src/kura.rs
```diff
@@ -166,13 +174,23 @@ impl Kura {
...
+                .block_arc_array
+                .lock()
+                .expect("lock on block arc array");
 
-        let mut guard = self
-            .block_hash_array
-            .lock()
-            .expect("lock on block hash array");
-        *guard = hash_array;
+            *guard = vec![None; blocks.len()];
```
appetrosyan, line 192: This value is kind of thing that should be documented with comments. Or reconsidered…

core/src/kura.rs
```diff
@@ -255,6 +277,59 @@ impl Kura {
...
+        if *hash == Hash::zeroed().typed() {
+            None
+        } else {
+            match hash_array_guard.iter().position(|&x| x == *hash) {
+                None => None,
+                Some(index) => {
+                    Some(index as u64 + 1)
+                }
+            }
+        }
```
appetrosyan, line 292: I advocate for a monadic error handling model. 
```suggestion
        (*hash == Hash::zeroed().typed())
        .then_some(())
        .and(hash_array_guard.iter().position(|&x| x == *hash)
                .map(|index| index as u64 + 1))
```

appetrosyan: Given that the `hash` is unlikely to be `zero`, the order of logical checks should be reversed
```suggestion
       hash_array_guard.iter()
           .position(|&x| x == *hash)
           .map(|index| index as u64 + 1)
           .filter(|_| *hash == Hash::zeroed().typed())
```

core/src/kura.rs
```diff
@@ -255,6 +277,59 @@ impl Kura {
...
+            block_store
+                .read_block_data(block_start, &mut block_buf)
+                .expect("Failed to read block data.");
+            let block = VersionedCommittedBlock::decode_versioned(&block_buf)
+                .expect("Failed to decode block");
+
+            let block_arc = Arc::new(block);
+            arc_array_guard[block_number] = Some(Arc::clone(&block_arc));
+            Some(block_arc)
+        }
```
appetrosyan, line 319: Technically, this is a `then_some` branch of the conditional. If you use `then_some` you get rid of the two `Some` instances, plus the `return None` line.

core/src/kura.rs
```diff
@@ -255,6 +277,59 @@ impl Kura {
...
+
+    pub fn get_block_by_hash(&self, block_hash: &HashOf<VersionedCommittedBlock>) -> Option<Arc<VersionedCommittedBlock>> {
+        let index = {
+            let hash_array_guard = self.block_hash_array.lock().expect("access hash array");
+            hash_array_guard.iter().position(|&b| b == *block_hash)
+        };
+        match index {
+            None => None,
+            Some(index) => self.get_block_by_height(index as u64 + 1),
+        }
```
appetrosyan, line 330:
```suggestion
            self.block_hash_array
                .lock()
                .expect("access hash array")
                .iter()
                .position(|&b| b == *block_hash)
                .map(|idx| self.get_block_by_height(idx as u64 + 1);
```

core/src/kura.rs
```diff
@@ -278,7 +353,7 @@ impl Kura {
 /// The [`BlockStoreTrait`] defines and implements functionality used by Kura
 /// for every platform. The default implementation is `StdFileBlockStore`,
 /// which uses `std::fs`.
-pub trait BlockStoreTrait {
+pub trait BlockStoreTrait : Debug {
```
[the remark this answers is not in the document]
appetrosyan: Technically this doesn't hurt, but I find that this anti-pattern of requiring `Debug` is only needed when you want to print something with a trai-bound `T: BlockStoreTrait`. 

In general this trait should go, given that we don't really have any implementation other than Kura, and rely on the actor system to the extent to which alternative implementations would be difficult.

core/src/wsv.rs
```diff
@@ -411,59 +397,32 @@ impl WorldStateView {
...
         &self,
         hash: HashOf<VersionedCommittedBlock>,
-    ) -> Vec<VersionedCommittedBlock> {
-        self.blocks
-            .iter()
-            .skip_while(move |block_entry| block_entry.value().header().previous_block_hash != hash)
-            .map(|block_entry| block_entry.value().clone())
-            .collect()
+    ) -> Vec<HashOf<VersionedCommittedBlock>> {
+        if hash == Hash::zeroed().typed() {
```
appetrosyan, line 416: Most hashes are non-zeroed. We also store a lot of them. I'd consider declaring a niche, and turning the few places where hashes can also be zeroed into an `Option<Hash>`.

core/src/wsv.rs
appetrosyan: A `std::vec::Vec` _is_ an iterator.

core/src/block_sync.rs
```diff
@@ -181,11 +184,34 @@ pub mod message {
...
+                            None => break,
+                            Some(hash) => {
+                                println!("{} : {}", h, hash);
+                                if h == 3 {
+                                    println!("POTATO");
+                                }
+                                h += 1;
+                            }
+                        }
+                    }
```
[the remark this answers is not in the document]
SamHSmith: Haha, yea I should have put a `//nocheckin` there. I was debugging.

core/src/kura.rs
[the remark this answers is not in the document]
SamHSmith: I tried this first. But it didn't work since when tests in other modules are run the test flag is not active for the core crate.

core/src/kura.rs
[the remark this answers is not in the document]
SamHSmith: The reason we store them linearly is not for the case of finding one. The most common operations are on sequences of blocks. Therefore they are stored linearly. "What blocks come after this one?" et cetera. If most of the blocks we ask for are recent then reverse linear search will be better. But if we ask for truly random blocks then maybe a separate hashmap/acceleration structure could help. Seems very premature of an optimization though to me.

core/src/kura.rs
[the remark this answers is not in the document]
SamHSmith: It's because other objects have derive. They need a reference to the Kura subsystem so they need an Arc<Kura>. This requires Kura to also implement debug and therefore BlockStoreTrait as well. We should probably do a custom implementation that doesn't print anything.

core/src/wsv.rs
```diff
@@ -83,14 +83,13 @@ impl World {
...
 /// Current state of the blockchain aligned with `Iroha` module.
-#[derive(Debug)]
 pub struct WorldStateView {
     /// The world - contains `domains`, `triggers`, etc..
     pub world: World,
     /// Configuration of World State View.
     pub config: Configuration,
     /// Blockchain.
-    blocks: Arc<Chain>,
+    pub block_hashes: std::cell::RefCell<Vec<HashOf<VersionedCommittedBlock>>>,
```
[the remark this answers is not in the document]
SamHSmith: Because the hashes are inexpensive to store. If we go to kura we are engaging in cross thread communication. Also in the case of rollbacks there can be a situation where kura is aware of more blocks then the world state is.

core/src/kura.rs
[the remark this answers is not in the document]
appetrosyan: LRU cache?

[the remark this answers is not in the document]
appetrosyan: Adding it would be trivial if we need it.

cli/src/torii/mod.rs
```diff
@@ -40,6 +41,7 @@ pub struct Torii {
     network: iroha_actor::Addr<IrohaNetwork>,
     notify_shutdown: Arc<Notify>,
     sumeragi: Arc<Sumeragi>,
+    kura: Arc<Kura>,
```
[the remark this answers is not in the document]
SamHSmith: If you want to stream blocks that this peer knows about. That has nothing to do with sumeragi. So we shouldn't talk to it.

permissions_validators/src/public_blockchain/transfer.rs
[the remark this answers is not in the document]
SamHSmith: I have a nocheckin rule. They are all gone before the PR is marked non-draft.

cli/src/torii/routing.rs
SamHSmith: Ideally all our subsystems would be statics. But that breaks our tests because there would be just one peer per process.

client/benches/tps/lib.rs
SamHSmith: The `by_value` is referring to how destructive the function is to performance. Perhaps the normal reader does not consider it insane to get "*all the blocks by* **value**". How would you name this in such a way that it shouts "really don't use this, it is terrible for performance".

core/Cargo.toml
SamHSmith: Yes because tests in other crates need it to be present.

core/src/kura.rs
SamHSmith: `block_arcs` and `block_hashes` then, because the `Arc` bit is significant.

core/src/kura.rs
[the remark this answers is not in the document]
appetrosyan: This doesn't have to be an inherent `impl`, you could try adding it to a trait that's only available under `cfg(test)`.

core/src/wsv.rs
Erigara: Yeah, but we are no longer return blocks, but blocks hashes.

cli/src/torii/routing.rs
Erigara: Maybe i miss something but i don't see how `Weak` would help, to access `Kura` we must call `upgrade` which will return `Arc<Kura>` and thus do increment/decrement.

core/src/block_sync.rs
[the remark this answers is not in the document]
appetrosyan: `gdb` should be the preferred approach. When we can inspect objects using `Debug` that is.

SamHSmith:
This work will be resumed at a later date.

PR #2871, opened 2022-10-17
[feature] #2280: Produce permission events when role is granted/revoked

### LINKED ISSUE #2280: Notify event listners when a role is Revoked
## Problem
As noted in [PR #2264](https://github.com/hyperledger/iroha/pull/2264) an account can lose a permission that's included in a role and event listeners that look for the permission being removed will not be notified.
## Solution
 It would be nice if  event listeners that look for `PermissionAdded` / `PermissionRemoved` to also be notified if a role that contains this permission is being revoked.

Erigara:

### Description of the Change

- Add mechanism to return multiple events from `modify_*` functions;
- Fixed error when revoked role does not exist;
- Revoke/Grant role also produce `PermissionAdded/Removed` events.

### Issue

Closes #2280.

### Benefits

More detailed events.

### Possible Drawbacks

None.

### DIFF
--- a/core/src/wsv.rs
+++ b/core/src/wsv.rs
@@ -109,7 +109,6 @@ impl Default for WorldStateView {
 }
 
 impl Clone for WorldStateView {
-    #[allow(clippy::expect_used)]
     fn clone(&self) -> Self {
         Self {
             world: Clone::clone(&self.world),
@@ -146,9 +145,7 @@ impl WorldStateView {
     }
 
     /// Return a set of all permission tokens granted to this account.
-    #[allow(clippy::unused_self)]
     pub fn account_permission_tokens(&self, account: &Account) -> Vec<PermissionToken> {
-        #[allow(unused_mut)]
         let mut tokens: Vec<PermissionToken> =
             self.account_inherent_permission_tokens(account).collect();
         for role_id in account.roles() {
@@ -279,7 +276,6 @@ impl WorldStateView {
     /// you likely have data corruption.
     /// - If trigger execution fails
     /// - If timestamp conversion to `u64` fails
-    #[allow(clippy::too_many_lines)]
     pub fn apply(&self, block: VersionedCommittedBlock) -> Result<()> {
         let time_event = self.create_time_event(block.as_v1())?;
         self.events_buffer
@@ -423,24 +419,34 @@ impl WorldStateView {
             .collect()
     }
 
-    /// Get `World` and pass it to closure to modify it
-    ///
-    /// Produces events in the `WSV` that are produced by `f` during execution.
-    /// Events are produced in the order of expanding scope: from specific to general.
-    /// Example: account events before domain events.
+    /// The same as [`Self::modify_world_multiple_events`] except closure `f` returns a single [`WorldEvent`].
     ///
     /// # Errors
-    /// Fails if `f` fails
-    ///
-    /// # Panics
-    /// (Rare) Panics if can't lock `self.events` for writing
-    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
+    /// Forward errors from [`Self::modify_world_multiple_events`]
     pub fn modify_world(
         &self,
         f: impl FnOnce(&World) -> Result<WorldEvent, Error>,
     ) -> Result<(), Error> {
-        let world_event = f(&self.world)?;
-        let data_events: SmallVec<[DataEvent; 3]> = world_event.into();
+        self.modify_world_multiple_events(move |world| f(world).map(std::iter::once))
+    }
+
+    /// Get [`World`] and pass it to `closure` to modify it.
+    ///
+    /// The function puts events produced by `f` into `events_buffer`.
+    /// Events should be produced in the order of expanding scope: from specific to general.
+    /// Example: account events before domain events.
+    ///
+    /// # Errors
+    /// Forward errors from `f`
+    pub fn modify_world_multiple_events<I: IntoIterator<Item = WorldEvent>>(
+        &self,
+        f: impl FnOnce(&World) -> Result<I, Error>,
+    ) -> Result<(), Error> {
+        let world_events = f(&self.world)?;
+        let data_events: SmallVec<[DataEvent; 3]> = world_events
+            .into_iter()
+            .flat_map(WorldEvent::flatten)
+            .collect();
 
         for event in data_events.iter() {
             self.world.triggers.handle_data_event(event.clone());
@@ -523,21 +529,34 @@ impl WorldStateView {
         Ok(value)
     }
 
-    /// Get `Domain` and pass it to closure to modify it
+    /// The same as [`Self::modify_domain_multiple_events`] except closure `f` returns a single [`DomainEvent`].
     ///
     /// # Errors
-    /// Fails if there is no domain
+    /// Forward errors from [`Self::modify_domain_multiple_events`]
     pub fn modify_domain(
         &self,
         id: &<Domain as Identifiable>::Id,
         f: impl FnOnce(&mut Domain) -> Result<DomainEvent, Error>,
     ) -> Result<(), Error> {
-        self.modify_world(|world| {
+        self.modify_domain_multiple_events(id, move |domain| f(domain).map(std::iter::once))
+    }
+
+    /// Get [`Domain`] and pass it to `closure` to modify it
+    ///
+    /// # Errors
+    /// - If there is no domain
+    /// - Forward errors from `f`
+    pub fn modify_domain_multiple_events<I: IntoIterator<Item = DomainEvent>>(
+        &self,
+        id: &<Domain as Identifiable>::Id,
+        f: impl FnOnce(&mut Domain) -> Result<I, Error>,
+    ) -> Result<(), Error> {
+        self.modify_world_multiple_events(|world| {
             let mut domain = world
                 .domains
                 .get_mut(id)
                 .ok_or_else(|| FindError::Domain(id.clone()))?;
-            f(domain.value_mut()).map(Into::into)
+            f(domain.value_mut()).map(|events| events.into_iter().map(Into::into))
         })
     }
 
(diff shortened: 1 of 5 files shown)


SamHSmith (approved):
Approved. You may want to consider adding a test. Otherwise LGTM.

QuentinI (commented):
Roles are also revoked when unregistering a role, so if we want to cover all such cases, we should generate token events [there](https://github.com/hyperledger/iroha/blob/iroha2-dev/core/src/smartcontracts/isi/world.rs#L178).

core/src/wsv.rs
```diff
@@ -430,17 +430,31 @@ impl WorldStateView {
...
         f: impl FnOnce(&World) -> Result<WorldEvent, Error>,
     ) -> Result<(), Error> {
-        let world_event = f(&self.world)?;
-        let data_events: SmallVec<[DataEvent; 3]> = world_event.into();
+        self.modify_world_multiple_events(move |world| f(world).map(std::iter::once))
+    }
+
+    /// Get `World` and pass it to closure to modify it
+    ///
+    /// Produces events in the `WSV` that are produced by `f` during execution.
```
outoftardis, line 443: "produces events ... that are produced" -- this should be rephrased

maybe like this (if I understood you correctly): "The function takes events from `WSV` that were produced during the execution of `f` and modifies them."?

core/src/wsv.rs
```diff
@@ -625,58 +652,103 @@ impl WorldStateView {
...
     pub fn modify_asset(
         &self,
         id: &<Asset as Identifiable>::Id,
         f: impl FnOnce(&mut Asset) -> Result<AssetEvent, Error>,
     ) -> Result<(), Error> {
-        self.modify_account(&id.account_id, |account| {
+        self.modify_asset_multiple_events(id, move |asset| f(asset).map(std::iter::once))
+    }
+
+    /// Get `Asset` by its id
```
outoftardis, line 695: function names start with `modify` but function descriptions are a bit different:

1) `and modify it with f`
2) `with an ability to modify it`
3) `and pass it to closure to modify it`
4) nothing

I'm trying to understand whether this is intentional: do all four cases mean different things? if not, then the descriptions should not be different

core/src/wsv.rs
```diff
@@ -430,17 +430,31 @@ impl WorldStateView {
     /// Example: account events before domain events.
     ///
     /// # Errors
-    /// Fails if `f` fails
-    ///
-    /// # Panics
```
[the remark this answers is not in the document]
Erigara: This section is outdated, as you see there is no usage of `self.events` at all.

core/src/wsv.rs
```diff
@@ -430,17 +430,31 @@ impl WorldStateView {
...
+    /// # Errors
+    /// Forward errors from `f`
+    pub fn modify_world_multiple_events<I: IntoIterator<Item = WorldEvent>>(
+        &self,
+        f: impl FnOnce(&World) -> Result<I, Error>,
+    ) -> Result<(), Error> {
+        let world_events = f(&self.world)?;
+        let data_events: SmallVec<[DataEvent; 3]> = world_events
+            .into_iter()
+            .flat_map(SmallVec::<[DataEvent; 3]>::from)
```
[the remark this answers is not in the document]
Erigara: Good point!

core/src/wsv.rs
Erigara: Use case of all this functions is the same.
We are passing object (world/asset/account/...) into closure `f` and receive bunch of events or error from it. 
We cant know what `f` is actually do with provided object (`f` can modify it or leave it as is).

core/src/wsv.rs
```diff
@@ -625,58 +652,103 @@ impl WorldStateView {
...
-        self.modify_account(&id.account_id, |account| {
+        self.modify_asset_multiple_events(id, move |asset| f(asset).map(std::iter::once))
+    }
+
+    /// Get `Asset` by its id
+    ///
+    /// # Errors
+    /// - If there are no such asset or account
+    /// - Forward errors from `f`
+    #[allow(clippy::missing_panics_doc)]
```
Erigara, line 700: Not relevant any more

Erigara:
> Roles are also revoked when unregistering a role, so if we want to cover all such cases, we should generate token events [there](https://github.com/hyperledger/iroha/blob/iroha2-dev/core/src/smartcontracts/isi/world.rs#L178).

Good catch, thanks

core/src/wsv.rs
Erigara: > "produces events ... that are produced" -- this should be rephrased

Yeah, this sound confusing.

I would slightly change you suggestion: "The function put events produced during `f` execution into `event_buffer`"

core/src/wsv.rs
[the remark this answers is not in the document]
Erigara: I have added function `WorldEvent::flatten(self)`.

core/src/wsv.rs
```diff
@@ -523,21 +529,34 @@ impl WorldStateView {
         Ok(value)
     }
 
-    /// Get `Domain` and pass it to closure to modify it
+    /// The same as [`Self::modify_domain_multiple_events`] except closure `f` return single [`DomainEvent`].
```
outoftardis, line 532:
```suggestion
    /// The same as [`Self::modify_domain_multiple_events`] except closure `f` returns a single [`DomainEvent`].
```

core/src/wsv.rs
```diff
@@ -622,61 +641,108 @@ impl WorldStateView {
         Ok(f(account))
     }
 
-    /// Get `Account` and pass it to closure to modify it
+    /// The same as [`Self::modify_account_multiple_events`] except closure `f` return single [`AccountEvent`].
```
outoftardis, line 644:
```suggestion
    /// The same as [`Self::modify_account_multiple_events`] except closure `f` returns a single [`AccountEvent`].
```

core/src/wsv.rs
```diff
@@ -622,61 +641,108 @@ impl WorldStateView {
...
             let account = domain
                 .account_mut(id)
                 .ok_or_else(|| FindError::Account(id.clone()))?;
-            f(account).map(DomainEvent::Account)
+            f(account).map(|events| events.into_iter().map(DomainEvent::Account))
         })
     }
 
-    /// Get `Asset` by its id
+    /// The same as [`Self::modify_asset_multiple_events`] except closure `f` return single [`AssetEvent`].
```
outoftardis, line 674:
```suggestion
    /// The same as [`Self::modify_asset_multiple_events`] except closure `f` returns a single [`AssetEvent`].
```

core/src/wsv.rs
```diff
@@ -622,61 +641,108 @@ impl WorldStateView {
...
         &self,
         id: &<Asset as Identifiable>::Id,
         f: impl FnOnce(&mut Asset) -> Result<AssetEvent, Error>,
     ) -> Result<(), Error> {
-        self.modify_account(&id.account_id, |account| {
+        self.modify_asset_multiple_events(id, move |asset| f(asset).map(std::iter::once))
+    }
+
+    /// Get [`Asset`] and pass it to `closure` to modify it
+    /// If asset value hits 0 after modification asset is removed from the [`Account`]
```
outoftardis, line 687:
```suggestion
    /// Get [`Asset`] and pass it to `closure` to modify it.
    /// If asset value hits 0 after modification, asset is removed from the [`Account`].
```

core/src/wsv.rs
```diff
@@ -622,61 +641,108 @@ impl WorldStateView {
...
                 assert!(account.remove_asset(id).is_some());
             }
 
-            event_result.map(AccountEvent::Asset)
+            events_result.map(|events| events.into_iter().map(AccountEvent::Asset))
         })
     }
 
-    /// Get `AssetDefinitionEntry` with an ability to modify it.
+    /// The same as [`Self::modify_asset_definition_multiple_events`] except closure `f` return single [`AssetDefinitionEvent`].
```
outoftardis, line 714:
```suggestion
    /// The same as [`Self::modify_asset_definition_multiple_events`] except closure `f` returns a single [`AssetDefinitionEvent`].
```

core/src/wsv.rs
```diff
@@ -851,17 +929,30 @@ impl WorldStateView {
         self.events_buffer.borrow_mut().push(event.into());
     }
 
-    /// Get chain of validators and modify it with `f`
-    ///
-    /// Produces [`PermissionValidatorEvent`] from `f`
+    /// The same as [`Self::modify_validator_multiple_events`] except closure `f` return single [`PermissionValidatorEvent`].
```
outoftardis, line 932:
```suggestion
    /// The same as [`Self::modify_validator_multiple_events`] except closure `f` returns a single [`PermissionValidatorEvent`].
```

data_model/src/events/data/events.rs
```diff
@@ -560,6 +560,49 @@ pub enum WorldEvent {
     PermissionValidator(permission::PermissionValidatorEvent),
 }
 
+impl WorldEvent {
+    /// Unfold [`Self`] and return vector of [`Event`]s in expanding scope order: from specific to general.
```
outoftardis, line 564:
```suggestion
    /// Unfold [`Self`] and return vector of [`Event`]s in the expanding scope order: from specific to general.
```

core/src/wsv.rs
```diff
@@ -423,24 +419,34 @@ impl WorldStateView {
...
     pub fn modify_world(
         &self,
         f: impl FnOnce(&World) -> Result<WorldEvent, Error>,
     ) -> Result<(), Error> {
-        let world_event = f(&self.world)?;
-        let data_events: SmallVec<[DataEvent; 3]> = world_event.into();
+        self.modify_world_multiple_events(move |world| f(world).map(std::iter::once))
+    }
+
+    /// Get [`World`] and pass it to `closure` to modify it
```
outoftardis, line 433:
```suggestion
    /// Get [`World`] and pass it to `closure` to modify it.
```

core/src/wsv.rs
```diff
@@ -423,24 +419,34 @@ impl WorldStateView {
...
+    ///
+    /// The function puts events produced by `f` into `events_buffer`
```
outoftardis, line 435:
```suggestion
    /// The function puts events produced by `f` into `events_buffer`.
```

outoftardis (approved):
two minor things, otherwise doc comments are ok

core/src/wsv.rs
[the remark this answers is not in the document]
QuentinI: I think this referred to [this](https://github.com/hyperledger/iroha/pull/2871/files#diff-b210796f83124da8c546d1f8ad43c2be94f98fd576a9e7018404a2760775dc92R455) access, which may still panic, although to my understanding that would indicate a bug and thus doesn't warrant a "Panics" section.

QuentinI (commented):
LGTM

QuentinI (approved):

appetrosyan (approved):

Issue #2858, opened 2022-10-12
Replace all uses of `std::sync::Mutex` with one that doesn't poisen.


SamHSmith:

appetrosyan:
The `in-house` instances should be used to plug in and replace the `std::sync::Mutex` and compare performance. We need to retain this ability late into development, so direct replacement with the underlying type is less preferable.

SamHSmith:
Fix for this has been merged already.

PR #2873, opened 2022-10-18
[fix] #2858: Introduce Iroha in-house `Mutex`.

SamHSmith:

### DIFF
--- a/core/src/sumeragi/mod.rs
+++ b/core/src/sumeragi/mod.rs
@@ -31,8 +31,7 @@ pub mod message;
 pub mod network_topology;
 pub mod view_change;
 
-use std::sync::Mutex;
-
+use iroha_primitives::sync::{Mutex, MutexGuard};
 use main_loop::State;
 
 use self::{
@@ -98,7 +97,7 @@ impl Sumeragi {
                 key_pair: configuration.key_pair.clone(),
                 peer_id: configuration.peer_id.clone(),
                 events_sender,
-                wsv: std::sync::Mutex::new(wsv),
+                wsv: Mutex::new(wsv),
                 commit_time: Duration::from_millis(configuration.commit_time_limit_ms),
                 block_time: Duration::from_millis(configuration.block_time_ms),
                 transaction_limits: configuration.transaction_limits,
@@ -145,26 +144,15 @@ impl Sumeragi {
             .internal
             .current_online_peers
             .lock()
-            .expect("Failed to lock `current_online_peers` for `update_metrics`")
             .len()
             .try_into()
             .expect("casting usize to u64");
 
-        let wsv_guard = self
-            .internal
-            .wsv
-            .lock()
-            .expect("Failed to lock on `wsv`. Mutex poisoned");
+        let wsv_guard = self.internal.wsv.lock();
 
-        let metrics_guard = self
-            .metrics_mutex
-            .lock()
-            .expect("Failed to lock on `metrics`. Mutex poisoned");
+        let metrics_guard = self.metrics_mutex.lock();
 
-        let mut last_guard = self
-            .last_update_metrics_mutex
-            .lock()
-            .expect("Failed to lock on `last_update_metrics`. Mutex poisoned");
+        let mut last_guard = self.last_update_metrics_mutex.lock_mut();
 
         let start_index = last_guard.block_height;
         {
@@ -228,20 +216,14 @@ impl Sumeragi {
 
     /// Access node metrics.
     #[allow(clippy::expect_used)]
-    pub fn metrics_mutex_access(&self) -> std::sync::MutexGuard<Metrics> {
-        self.metrics_mutex
-            .lock()
-            .expect("`Mutex` in `metrics_mutex_access` poisoned. This should not happen, given that panics should stop Iroha.")
+    pub fn metrics_mutex_access(&self) -> MutexGuard<Metrics> {
+        self.metrics_mutex.lock()
     }
 
     /// Get latest block hash for use by the block synchronization subsystem.
     #[allow(clippy::expect_used)]
     pub fn latest_block_hash(&self) -> HashOf<VersionedCommittedBlock> {
-        *self
-            .internal
-            .latest_block_hash
-            .lock()
-            .expect("Mutex on internal WSV poisoned in `latest_block_hash`")
+        *self.internal.latest_block_hash.lock()
     }
 
     /// Get an array of blocks after the block identified by `block_hash`. Returns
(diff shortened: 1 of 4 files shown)


core/src/sumeragi/mod.rs
```diff
@@ -228,20 +216,14 @@ impl Sumeragi {
 
     /// Access node metrics.
     #[allow(clippy::expect_used)]
-    pub fn metrics_mutex_access(&self) -> std::sync::MutexGuard<Metrics> {
-        self.metrics_mutex
-            .lock()
-            .expect("`Mutex` in `metrics_mutex_access` poisoned. This should not happen, given that panics should stop Iroha.")
+    pub fn metrics_mutex_access(&self) -> MutexGuard<Metrics> {
```
appetrosyan, line 219: Normally you want to return something which `impl Deref<Metrics>` so that you're not confined to a single implementation of a Guard. For example, once it's no longer a `Mutex` guard,

appetrosyan: Another option is to return a short-lived reference. Then you don't need to leak the `MutexGuard` into the public API of another crate.

core/src/sumeragi/mod.rs
```diff
@@ -276,11 +256,8 @@ impl Sumeragi {
     /// while this lock is held.
     // TODO: Return result.
     #[allow(clippy::expect_used)]
-    pub fn wsv_mutex_access(&self) -> std::sync::MutexGuard<WorldStateView> {
-        self.internal
-            .wsv
-            .lock()
-            .expect("World state view Mutex poisoned. You have had a panic somewhere else in the process, which didn't shut down the peer.")
+    pub fn wsv_mutex_access(&self) -> MutexGuard<WorldStateView> {
```
appetrosyan, line 259: Same here.

core/src/sumeragi/mod.rs
```diff
@@ -348,24 +325,13 @@ impl Sumeragi {
...
     pub fn incoming_message(&self, msg: MessagePacket) {
-        if self
-            .internal
-            .message_sender
-            .lock()
-            .expect("Lock on sender")
-            .try_send(msg)
-            .is_err()
-        {
+        if self.internal.message_sender.lock().try_send(msg).is_err() {
```
appetrosyan, line 334: The `if` on an error variant discards the error message. If you do an `if let Err(err)` you can also print the error message in `debug!`, so that we have a fuller picture.

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
+use core::ops::{Deref, DerefMut};
+use std::{
+    sync::atomic::{AtomicU32, Ordering},
+    time::{Duration, Instant},
+};
+
+const TIME_SPINNING_MICRO_SECONDS: u128 = 500;
```
appetrosyan, line 7: This is a completely arbitrary number.

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
...
+const SLEEP_TIME: Duration = Duration::from_micros(100);
```
appetrosyan, line 8: Another arbitrary number.

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
...
+
+pub struct MutexGuard<'a, T: ?Sized> {
+    mutex: &'a Mutex<T>,
+}
+
+pub struct Mutex<T: ?Sized> {
+    counter: AtomicU32,
+    interior: T,
+}
+
```
appetrosyan, line 22: This looks like you're implementing a custom `Mutex`. When in reality you should be implementing a thin wrapper for a pre-existing Mutex implementation.

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn lock_mut(&self) -> MutexMutGuard<T> {
+        let instant_entered = Instant::now();
+        loop {
+            match AtomicU32::compare_exchange_weak(
+                &self.counter,
+                0,
+                1,
+                Ordering::AcqRel,
+                Ordering::Relaxed,
```
appetrosyan, line 45: CAREFUL! These are not the right orderings!

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
+use core::ops::{Deref, DerefMut};
```
appetrosyan, line 1: Consider treating this crate as a means to easily swap out standard implementations of `Mutex` rather than copying an existing implementation. 

DRY

primitives/src/sync.rs
Erigara: `parking_lot` offers [`Mutex`](https://docs.rs/parking_lot/latest/parking_lot/type.Mutex.html#) without poisoning.

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
...
+            match AtomicU32::compare_exchange_weak(
+                &self.counter,
+                0,
+                1,
+                Ordering::AcqRel,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => break,
+                Err(prev) => debug_assert!(prev == 1),
+            }
```
mversic, line 69:
```suggestion
        while let Err(prev) = AtomicU32::compare_exchange_weak(
                &self.counter, 0, 1, Ordering::AcqRel, Ordering::Relaxed,
        ) {
            debug_assert!(prev == 1);
        }
```

mversic: this is one of the rare cases for the use of weak CAS operation

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
...
+            if instant_entered.elapsed().as_micros() > TIME_SPINNING_MICRO_SECONDS {
+                std::thread::sleep(SLEEP_TIME);
+            }
+        }
```
mversic, line 73: `compare_exchange_weak` can fail spuriously. Therefore, I believe that the `prev` might also be `0` and the assertion `prev == 1` you make is incorrect. 
If this is true then I suggest the following:

```suggestion
        while let Err(_) = self.counte.compare_exchange_weak(
                0, 1, Ordering::AcqRel, Ordering::Relaxed,
        ) {
            if instant_entered.elapsed().as_micros() > TIME_SPINNING_MICRO_SECONDS {
                std::thread::sleep(SLEEP_TIME);
            }
        }
```

primitives/src/sync.rs
mversic: I think that the success case is ok, sequential consistency is not required in the current implementation. Not sure about the failure case

mversic (changes requested):

primitives/src/sync.rs
mversic: I take it back. Success should be `Acquire` and failure can be `Relaxed` ordering

primitives/src/sync.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl<T> Mutex<T> {
+    pub fn new(interior: T) -> Self {
+        Self {
+            counter: AtomicU32::new(0),
+            interior,
+        }
+    }
+
+    pub fn lock_mut(&self) -> MutexMutGuard<T> {
```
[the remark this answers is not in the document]
appetrosyan: Actually this is a good point. There is a difference with an `RWLock` but the two should be differentiated.

PR #2887, opened 2022-10-19
[fix] #2870: Reduce amount of ConnectPeer messages sent to p2p (lts)

SamHSmith:

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -564,6 +564,8 @@ pub fn run<F>(
         state.current_topology.role(&sumeragi.peer_id),
     );
 
+    let mut last_connect_peers_instant = Instant::now();
+
     // do normal rounds
     let mut voting_block_option = None;
     let mut block_signature_acc = Vec::new();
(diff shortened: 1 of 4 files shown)


appetrosyan (approved):

outoftardis:
same comments as in #2888

core/src/sumeragi/main_loop.rs
```diff
@@ -592,7 +594,10 @@ pub fn run<F>(
         let span_for_sumeragi_cycle = span!(Level::TRACE, "Sumeragi Main Thread Cycle");
         let _enter_for_sumeragi_cycle = span_for_sumeragi_cycle.enter();
 
-        sumeragi.connect_peers(&state.current_topology);
+        if last_connect_peers_instant.elapsed().as_millis() > 1000 {
+            sumeragi.connect_peers(&state.current_topology);
+            last_connect_peers_instant = Instant::now();
+        }
```
Erigara, line 600: Can we instead of trying to connect peers in every loop iteration do it when `current_topology` is changed? (Seems to be after `refresh_at_new_block`)

Erigara (approved):

core/src/sumeragi/main_loop.rs
SamHSmith: No because you need the peers to connect to each other even when nothing is happening.

Erigara: @SamHSmith, what do you mean by that?
As i see it with `connect_peers` we are sending bunch of `ConnectPeer`, `DisconnectPeer` messages and after processing on p2p side peers will be connected and we can send messages to them.
Why we need to restarting connection over and over again without any changes in topology?

PR #2888, opened 2022-10-19
[fix] #2870: Reduce amount of ConnectPeer messages sent to p2p.

SamHSmith:

### DIFF
--- a/core/src/sumeragi/mod.rs
+++ b/core/src/sumeragi/mod.rs
@@ -258,16 +258,14 @@ impl Sumeragi {
         self.wsv_mutex_access().blocks_from_height(block_height)
     }
 
-    /// Get a random online peer for use in block synchronization.
+    /// Get list of peer id's for use in block synchronization.
     #[allow(clippy::expect_used, clippy::unwrap_in_result)]
-    pub fn get_random_peer_for_block_sync(&self) -> Option<Peer> {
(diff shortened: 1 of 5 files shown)


core/src/sumeragi/mod.rs
```diff
@@ -258,16 +258,14 @@ impl Sumeragi {
         self.wsv_mutex_access().blocks_from_height(block_height)
     }
 
-    /// Get a random online peer for use in block synchronization.
+    /// Get list of peer id's for use in block synchronization.
```
outoftardis, line 261:
```suggestion
    /// Get the list of peer ids for block synchronization.
```

p2p/src/network.rs
```diff
@@ -59,6 +59,8 @@ where
     pub new_peers: HashMap<ConnectionId, Addr<Peer<T, K, E>>>,
     /// Current [`Peer`]s in [`Peer::Ready`] state.
     pub peers: HashMap<PublicKey, RefPeer<T, K, E>>,
+    /// Connecting to peer id's.
```
outoftardis, line 62:
```suggestion
    /// The ids of peers to connect to
```
```suggestion
    /// The ids of connecting peers
```
also: why so we say `id` if it's a vector of public keys?
```suggestion
    /// The public keys of connecting peers
```

SamHSmith:
Closed in favour of more robust fix.

Issue #2773, opened 2022-09-21
Decouple `ConfigurationView` from `ConfigurationProxy` structs


ilchu:
In `iroha_config::base::derive`, the `view!` proc macro has to remove attributes unrelated to itself, such as `#[config]` or `#[builder]`. Ideally, this should be out of that macro's concern. As the number of such attributes might increase, it would make remembering to remove them unwieldy and macro debugging tough.

Issue #2632, opened 2022-08-17
Link to the original struct in `Proxy` doc


ilchu:
>I think we could add link to the original struct in docs, because we would have more than one proxy objects.
_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/2602#discussion_r947701400_

Determine if it's going to be needed in the final design, as proxies may end up being not exposed to the user at all and thus also not requiring any documentation.

Issue #2617, opened 2022-08-12
Remove all unneede parent attributes on `_Proxy` structs


ilchu:

appetrosyan:
Should we open the discussion here?

ilchu:
> Imagine that the struct has an attr from derive_more, which defines some behaviour. Would they not lead to a compile error given that the derive_more::From isn't part of the derive() directives?

It definitely would, and in the case of `derive_more` we wouldn't need most of this functionality on the proxy struct (such as display formatting etc.), as it's only an intermediary that gets thrown away after bootstrapping.

My doubts initially concerned attributes such as `#[config(inner)]`, as I figured they would be needed for trait impls, but upon further look it's perfectly sufficient to only have them on parent struct as we do now.

So I guess if everyone agrees, and I don't stumble against unforeseen roadblocks, the attributes can actually be ditched.

ilchu:
Currently the only present attributes, `#[view]` and `#[serde]` are being deleted, but a small change might be required to make sure to delete *all* irrelevant attributes to make it future proof.

Erigara:
IMO we should keep scope of `Proxy` structs narrow, so remove unnecessary attributes.

PR #2900, opened 2022-10-21
[refactor] #2747: Change `LoadFromEnv` API

### LINKED ISSUE #2747: Change `LoadFromEnv` API
As per [comment](https://github.com/hyperledger/iroha/pull/2706#discussion_r964492532). Also currently the entrypoint in `cli/src/lib.rs` doesn't use the combine trait, but that will be changed when `LoadFromEnv` is put into desired shape.

ilchu:

### Description of the Change
* `LoadFromEnv` and `LoadFromDisk` updated to return proxy configs in any case, without failure
* `Combine` trait now properly loads inner config values, updating all the fields
* Documentation for all proxy traits in `config/base/lib.rs` updated
* Small refactor of `config/base/derive/src/view.rs`, mostly in connection with #2773 

### Issue

Closes #2747, #2773, #2632, #2617

### Benefits
Expected behaviour (e.g. bootstrapping a peer with just the env variables) is back

### Possible Drawbacks
None I could think of

### Usage Examples or Tests
Some new tests added to `config/base/tests/simple.rs`, reworked `config/base/lib.rs` documentation examples as well


cli/src/samples.rs
```diff
@@ -103,3 +103,65 @@ pub fn get_config(trusted_peers: HashSet<PeerId>, key_pair: Option<KeyPair>) ->
     .build()
     .expect("Iroha config should build as all required fields were provided")
 }
+
+#[allow(clippy::implicit_hasher)]
+/// Get a sample Iroha configuration proxy. Trusted peers must either be
```
outoftardis, line 108:
```suggestion
/// Get a sample Iroha configuration proxy. Trusted peers must be
```

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
+        if field.has_inner {
+            let inner_ty = get_inner_type("Option", &field.ty);
+            quote! {
+                if let Some(other_field) = other.#field_name {
+                    if let Some(this_field) = self.#field_name {
+                        self.#field_name = Some(<#inner_ty as #combine_trait>::combine(this_field, other_field))
+                    } else {
+                        self.#field_name = Some(other_field)
+                    }
+                }
```
Erigara, line 51: This is for the inner configurations, right?

cli/src/samples.rs
```diff
@@ -103,3 +103,65 @@ pub fn get_config(trusted_peers: HashSet<PeerId>, key_pair: Option<KeyPair>) ->
...
+/// specified in this function, including the current peer. Use [`get_trusted_peers`]
+/// to populate `trusted_peers` if in doubt. Almost equivalent to the [`get_config`]
+/// function, except the fact that the proxy is left unbuilt.
```
outoftardis, line 111:
```suggestion
/// function, except the proxy is left unbuilt.
```

cli/src/samples.rs
```diff
@@ -103,3 +103,65 @@ pub fn get_config(trusted_peers: HashSet<PeerId>, key_pair: Option<KeyPair>) ->
...
+///
+/// # Panics
+/// when [`KeyPair`] generation fails (rare).
```
outoftardis, line 114:
```suggestion
/// - when [`KeyPair`] generation fails (rare case)
```

config/base/src/lib.rs
```diff
@@ -83,78 +154,147 @@ pub mod derive {
...
+    ///
+    /// Meant to be used on proxy types only, for
+    /// details see [`iroha_config::base::derive::Proxy`](`crate::derive::Proxy`).
+    ///
+    /// The trait's only method, `from_path`,
+    /// deserializes a JSON config at the provided path into the parent proxy structure,
+    /// leaving it empty in case of any error.
+    ///
+    /// The `ReturnValue` associated type can be
+    /// swapped for anything suitable, returning just the proxy structure itself currently.
```
outoftardis, line 186:
```suggestion
    /// swapped for anything suitable, returning just the proxy structure itself.
```

config/base/src/lib.rs
```diff
@@ -83,78 +154,147 @@ pub mod derive {
...
-    /// [`iroha_config::base::derive::LoadFromEnv`](`crate::proxy::LoadFromEnv`)
+    /// [`iroha_config::base::proxy::LoadFromDisk`](`crate::proxy::LoadFromDisk`)
     /// trait for config structures.
     ///
-    /// Has several attributes:
+    /// Meant to be used on proxy types only, for
+    /// details see [`iroha_config::base::derive::Proxy`](`crate::derive::Proxy`).
+    ///
+    /// The `ReturnValue` associated type can be
+    /// swapped for anything suitable, returning just the proxy structure itself currently.
```
outoftardis, line 196:
```suggestion
    /// swapped for anything suitable, returning just the proxy structure itself.
```

config/base/src/lib.rs
```diff
@@ -83,78 +154,147 @@ pub mod derive {
...
     pub use iroha_config_derive::LoadFromEnv;
-    // TODO: new docs -- probably better left until last steps
-    pub use iroha_config_derive::{Builder, Combine, LoadFromDisk, Proxy};
+    /// Derive macro for implementing the corresponding proxy type
+    /// for config structures. Most of the other traits in the
+    /// [`iroha_config_base::proxy`](`crate::proxy`) module are
+    /// best derived indirectly via this macro. Proxy types serve
+    /// as a stand-in for flexible configuration loading either
+    /// from environment variables or configuration files, while
+    /// providing methods to build the initial parent type from them
```
outoftardis, line 270:
```suggestion
    /// from environment variables or configuration files. Proxy types also
    /// provide methods to build the initial parent type from them
```

config/base/src/lib.rs
```diff
@@ -296,28 +433,30 @@ pub mod proxy {
 
     /// Trait for configuration loading and deserialization from
     /// the environment
-    pub trait LoadFromEnv {
-        /// Error type returned by methods of this trait
-        type Error;
+    pub trait LoadFromEnv: Sized {
+        /// The return type. Could be target `Configuration`,
+        /// some `Result`, `Option` as users see fit.
```
outoftardis, line 438:
```suggestion
        /// The return type. Could be target `Configuration`,
        /// `Result`, or `Option`.
```

config/base/src/lib.rs
```diff
@@ -296,28 +433,30 @@ pub mod proxy {
...
-        fn load_environment(&mut self) -> Result<(), Self::Error>;
+        fn from_env() -> Self::ReturnValue;
     }
 
     /// Trait for configuration loading and deserialization from disk
     pub trait LoadFromDisk: Sized {
-        /// Error type returned by methods of this trait
-        type Error;
+        /// The return type. Could be target `Configuration`,
+        /// some `Result`, `Option` as users see fit.
```
outoftardis, line 451:
```suggestion
        /// The return type. Could be target `Configuration`,
        /// `Result`, or `Option`.
```

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
-                }
-            } else if field.has_inner {
-                quote! {
-                    #l_value.load_environment()?;
+                if field.has_inner {
+                    quote! {
+                        let #ident = if let Ok(var) = std::env::var(#field_env) {
+                            let inner: Result<#inner_ty, #err_ty> = #inner;
+                            inner.ok()
+                        } else { None };
```
Erigara, line 108: Consider (just personal stylistic preferences):

```rust
let #ident = std::env::var(#field_env)
    .ok()
    .and_then(|var| {
        let inner: Result<#inner_ty, #err_ty> = #inner;
        inner.ok()
    })
```

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
+                        let inner_proxy = <#inner_ty as ::iroha_config_base::proxy::LoadFromEnv>::from_env();
+                        let #ident = if #ident.is_some() {
```
Erigara, line 110: While not `if let Some(...)`? instead of `unwrap`.

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
+                            Some(inner_proxy)
+                        };
+                    }
+                } else {
+                    quote! {
+                        let #ident = if let Ok(var) = std::env::var(#field_env) {
+                            let inner: Result<#inner_ty, #err_ty> = #inner;
+                            inner.ok()
+                        } else { None };
+                    }
```
Erigara, line 122: i wonder could we deduplicate code here like: 
```rust
let set_field = {
    let mut set_field = quote! {
        let #ident = if let Ok(var) = std::env::var(#field_env) {
            let inner: Result<#inner_ty, #err_ty> = #inner;
            inner.ok()
        } else { None };
    };
    if field.has_inner {
       set_field.extend(
           quote! {
           // inner proxy logic...
           }
       )
    }
    set_field
} 
```

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
+    let env_trait = quote! { ::iroha_config_base::proxy::LoadFromEnv };
     quote! {
-        impl ::iroha_config_base::proxy::LoadFromEnv for #name {
-            type Error = ::iroha_config_base::derive::Error;
-            fn load_environment(
-                &'_ mut self
-            ) -> core::result::Result<(), ::iroha_config_base::derive::Error> {
+        impl #env_trait for #name {
+            type ReturnValue = Self;
+            fn from_env() -> Self::ReturnValue {
```
Erigara, line 144: If we are always return `Self` does it make sense to simplify `LoadFromEnv` by removing `ReturnValue`?

config/base/derive/src/proxy.rs
```diff
@@ -125,16 +156,27 @@ pub fn impl_load_from_disk(ast: &StructWithFields) -> TokenStream {
...
-                Ok(res)
+                let res = file
+                    .and_then(|mut f| {
+                        ::std::io::Read::read_to_string(&mut f, &mut s).map_err(#disk_err_variant)
+                    })
+                    .and_then(
+                        |_| -> ::core::result::Result<Self, #error_ty> {
+                            serde_json::from_str(&s).map_err(#serde_err_variant)
+                        },
+                    )
```
Erigara, line 177: I think it is possible to avoid referencing outer scope inside closure.

```suggestion
                    .and_then(|mut f| {
                        let mut s = String::new();
                        ::std::io::Read::read_to_string(&mut f, &mut s).map(move |_| s).map_err(#disk_err_variant)
                    })
                    .and_then(
                        |s| -> ::core::result::Result<Self, #error_ty> {
                            serde_json::from_str(&s).map_err(#serde_err_variant)
                        },
                    )
```

cli/src/samples.rs
```diff
@@ -103,3 +103,65 @@ pub fn get_config(trusted_peers: HashSet<PeerId>, key_pair: Option<KeyPair>) ->
...
+        sumeragi: Some(iroha_config::sumeragi::ConfigurationProxy {
+            key_pair: None,
+            peer_id: None,
+            trusted_peers: Some(TrustedPeers {
+                peers: trusted_peers,
+            }),
+            ..iroha_config::sumeragi::ConfigurationProxy::default()
+        }),
+        torii: Some(iroha_config::torii::ConfigurationProxy {
+            p2p_addr: Some(DEFAULT_TORII_P2P_ADDR.to_owned()),
```
appetrosyan, line 138: Why couldn't this entire struct be `DEFAULT_TORII_CONFIG_PROXY`?

config/base/derive/src/proxy.rs
appetrosyan, line 51: I'd argue that a `match` statement would make it much more readable. Consider
```suggestion
                match (self.#field_name, other.#field_name) {
                      (Some(this), Some(other)) => Some(<#inner_ty as #combine_trait>::combine(this, other)), 
                      (this, None) => this, 
                      (None, other) => other,
                }
```

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
+            quote! {
+                if let Some(other_field) = other.#field_name {
+                    if let Some(this_field) = self.#field_name {
+                        self.#field_name = Some(<#inner_ty as #combine_trait>::combine(this_field, other_field))
+                    } else {
+                        self.#field_name = Some(other_field)
+                    }
+                }
+            }
+        } else {
```
appetrosyan, line 53: If field doesn't have inner, does that automatically mean that we cna just ignore the `this`?

config/base/derive/src/proxy.rs
```diff
@@ -35,86 +37,115 @@ pub fn impl_proxy(ast: StructWithFields) -> TokenStream {
...
             let as_str_attr = field.has_as_str;
             let ident = &field.ident;
             let field_env = &field.env_str;
 
             let set_field = {
-                let inner_ty = field.has_option.then(|| get_inner_type("Option", ty)).unwrap_or(ty);
+                let inner_ty = if field.has_option {
+                    get_inner_type("Option", ty)
+                } else {
+                    abort!(ast, "This macro should only be used on `..Proxy` types")
```
appetrosyan, line 87:
```suggestion
                    abort!(ast, "This macro should only be used on Configuration Proxy types, i.e. types which represent a partially finalised configuration (with some required fields omitted and to be read from other sources). These types have the `Option` type everywhere that's applicable. ")
```

config/base/derive/src/proxy.rs
appetrosyan: A little bit of Sam's influence.

config/base/derive/src/proxy.rs
appetrosyan: What if we're not? Moreover, this is just code fluff that gets optimised away after macro expansion...

config/base/derive/src/proxy.rs
```diff
@@ -223,3 +269,20 @@ fn gen_none_fields_check(ast: &StructWithFields) -> proc_macro2::TokenStream {
         #(#checked_fields),*
     }
 }
+
+/// Helper function to be used as an empty fallback for [`impl LoadFromEnv`] or [`impl LoadFromDisk`].
+/// Only meant for proxy types usage.
```
appetrosyan, line 274: Appreciate the comment.

core/test_network/src/lib.rs
```diff
@@ -834,20 +841,18 @@ impl TestRuntime for Runtime {
...
+            iroha::samples::get_config_proxy(HashSet::new(), Some(get_key_pair()));
+        let env_proxy = ConfigurationProxy::from_env();
         let (public_key, private_key) = KeyPair::generate().unwrap().into();
-        configuration.public_key = public_key;
-        configuration.private_key = private_key;
-        configuration
+        sample_proxy.public_key = Some(public_key);
+        sample_proxy.private_key = Some(private_key);
+        sample_proxy
+            .combine(env_proxy)
```
appetrosyan, line 853: Shouldn't it be the other way around? Specifically, shouldn't the environment override the sample?

core/test_network/src/lib.rs
```diff
@@ -834,20 +841,18 @@ impl TestRuntime for Runtime {
...
+            .build()
+            .expect("Test Iroha config should build as all required fields have been provided")
```
appetrosyan, line 855:
```suggestion
            .expect("Test Iroha config failed to build. This is either a programmer error or a compiler bug. ")
```

config/src/wsv.rs
```diff
@@ -11,9 +11,7 @@ const DEFAULT_METADATA_LIMITS: MetadataLimits = MetadataLimits::new(2_u32.pow(20
 const DEFAULT_IDENT_LENGTH_LIMITS: LengthLimits = LengthLimits::new(1, 2_u32.pow(7));
 
 /// `WorldStateView` configuration.
-#[derive(
-    Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, Proxy, LoadFromEnv, Documented,
-)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, Proxy, Documented)]
```
appetrosyan, line 14: You should consider re-exporting `Documented` and `Proxy` from `iroha_config_base`. As a result, you could set up something like either `iroha_config_base::Proxy`, or `config::Proxy`, so that the namespacing disambiguates the `Proxy` and `Documented` terms. 

`ConfigProxy` is also a viable option, but it's usually better to use namespaces, in case someone wants to shorten the naming.

config/src/iroha.rs
```diff
@@ -240,13 +238,11 @@ mod tests {
...
         );
-        Ok(())
     }
 
     #[test]
     fn example_json_proxy_builds() -> Result<()> {
-        let cfg_proxy = ConfigurationProxy::from_path(CONFIGURATION_PATH)
-            .wrap_err("Failed to read configuration from example config")?;
+        let cfg_proxy = ConfigurationProxy::from_path(CONFIGURATION_PATH);
         cfg_proxy.build()?;
```
appetrosyan, line 246:
```suggestion
        ConfigurationProxy::from_path(CONFIGURATION_PATH).build()?;
```

appetrosyan:
```suggestion
        ConfigurationProxy::from_path(CONFIGURATION_PATH).build().expect()
```
And remove `Ok(())` to conform to the previous test.

config/src/client.rs
```diff
@@ -258,8 +258,7 @@ mod tests {
         fn client_proxy_build_fails_on_none(proxy in arb_proxy()) {
             let cfg = proxy.build();
             if cfg.is_ok() {
-                let example_cfg = ConfigurationProxy::from_path(CONFIGURATION_PATH)
-                    .expect("Failed to read example config file").build().expect("Failed to build example Iroha config");
+                let example_cfg = ConfigurationProxy::from_path(CONFIGURATION_PATH).build().expect("Failed to build example Iroha config");
                 let arb_cfg = cfg.unwrap();
```
appetrosyan, line 262:
```suggestion
                let arb_cfg = cfg.expect("<Invariant> got violated. <Instructions for where to look for problems>");
```

config/base/src/lib.rs
```diff
@@ -284,10 +424,7 @@ pub mod proxy {
     }
 
     /// Trait for configuration loading and deserialization
-    pub trait Combine: Serialize + DeserializeOwned + Sized + LoadFromEnv + LoadFromDisk {
-        // /// Error type returned by methods of this trait
-        // type Error;
-
+    pub trait Combine: Serialize + DeserializeOwned + Sized {
```
appetrosyan, line 427: Excellent. This decouples the traits, so we can use them for other stuff.

config/base/src/lib.rs
ilchu: What I'm trying to say here is that return value could be changed, but as of now it's set to be the proxy struct.

config/base/src/lib.rs
ilchu: Same comment as above.

config/base/src/lib.rs
ilchu: How about
```suggestion
        /// The return type. Could be target `Configuration`,
        /// some `Result`, `Option`, or any other reasonable alternative.
```
Just to signify that these three are the most likely options, but it could be anything else if desired.

cli/src/lib.rs
```diff
@@ -154,9 +154,9 @@ impl Iroha {
         query_judge: QueryJudgeBoxed,
     ) -> Result<Self> {
         let broker = Broker::new();
-        let mut proxy = ConfigurationProxy::from_path(&args.config_path)?;
-        proxy.load_environment()?;
-        let config = proxy.build()?;
+        let file_proxy = ConfigurationProxy::from_path(&args.config_path);
+        let env_proxy = ConfigurationProxy::from_env();
+        let config = file_proxy.combine(env_proxy).build()?;
```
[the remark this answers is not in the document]
ilchu: Yeah finally making use of things I write heh.

config/base/derive/src/proxy.rs
ilchu: As @appetrosyan mentioned a long time ago, keeping an associated type allows to have more flexible impls in the future, if we ever need it to be done for something else.

cli/src/samples.rs
```diff
@@ -103,3 +103,65 @@ pub fn get_config(trusted_peers: HashSet<PeerId>, key_pair: Option<KeyPair>) ->
...
+
+#[allow(clippy::implicit_hasher)]
+/// Get a sample Iroha configuration proxy. Trusted peers must either be
+/// specified in this function, including the current peer. Use [`get_trusted_peers`]
+/// to populate `trusted_peers` if in doubt. Almost equivalent to the [`get_config`]
+/// function, except the fact that the proxy is left unbuilt.
+///
+/// # Panics
+/// when [`KeyPair`] generation fails (rare).
+pub fn get_config_proxy(
```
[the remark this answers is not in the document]
ilchu: As I slapped on the `..proxy` counterpart post-hoc, didn't really notice this alternative, thanks.

core/test_network/src/lib.rs
ilchu: That's exactly what was happening, so I'm renaming trait and method to make it more clear as you proposed.

config/base/derive/src/proxy.rs
ilchu: I guess it does lead to a somewhat smaller cognitive load, yeah. Commiting.

config/base/derive/src/proxy.rs
ilchu: Yeah, and as a side note, same logic in `LoadFromEnv` now made it loosely coupled to the `Combine` trait, as the same method gets called on inner proxies. That should be fine though AFAICT.

config/base/derive/src/proxy.rs
ilchu: Fields in `other` are meant to override fields in `self`, so I guess yeah?

cli/src/samples.rs
```diff
@@ -103,3 +103,65 @@ pub fn get_config(trusted_peers: HashSet<PeerId>, key_pair: Option<KeyPair>) ->
...
+        Some(key_pair) => key_pair.into(),
+        None => KeyPair::generate()
+            .expect("Key pair generation failed")
+            .into(),
+    };
+    iroha_logger::info!(?public_key);
+    ConfigurationProxy {
+        public_key: Some(public_key.clone()),
+        private_key: Some(private_key.clone()),
+        sumeragi: Some(iroha_config::sumeragi::ConfigurationProxy {
```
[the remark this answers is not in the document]
ilchu: Could ditch keypair and id, yeah.

cli/src/samples.rs
ilchu: Because technically we're not having these addresses set currently in `torii`'s `Default` peer config, thus requiring them to be provided. So those constants are pretty much only used here and for client config.

config/src/wsv.rs
ilchu: You mean bringing these derives up to `iroha_config` level outside of `iroha_config_base`? Or disambiguating the derive and its trait? If it's the latter, they are already namespaced by belonging to `iroha_config_base::derive::Documented` and `iroha_config_base::proxy::Documented` respectively.

config/base/src/lib.rs
outoftardis: hm, maybe then say that by default it returns the proxy struct? something like this:

```suggestion
    /// swapped for anything suitable. Currently, the proxy structure is returned by default.
```

config/base/src/lib.rs
outoftardis: can you define `reasonable alternative`? what's reasonable and what is not?

config/base/src/lib.rs
```diff
@@ -283,41 +423,40 @@ pub mod proxy {
...
-        // /// Error type returned by methods of this trait
-        // type Error;
-
+    /// Trait for combining two configuration instances
+    pub trait Override: Serialize + DeserializeOwned + Sized {
         /// If any of the fields in `other` are filled, they
         /// override the values of the fields in [`self`].
         #[must_use]
-        fn combine(self, other: Self) -> Self;
+        fn override_with(self, other: Self) -> Self;
```
Erigara, line 431: Like this name, because we have clear precedence order

Erigara (approved):
LGTM

Erigara (approved):

config/src/wsv.rs
appetrosyan: The former.

appetrosyan (approved):

Issue #2458, opened 2022-07-08
[suggestion] Common combinators for Filters, Permission validators and predicates


appetrosyan:
### Feature request

Need to introduce a common combinator API that leverages `iroha_data_model::predicate`. 

### Motivation

Need to make sure that combinatorial objects are combined in a consistent fashion. 

### Who can help?

@appetrosyan

PR #2907, opened 2022-10-25
[refactor] #2458: implement generic combinator API

QuentinI:


data_model/src/predicate.rs
```diff
@@ -46,10 +74,83 @@ impl PredicateBox {
             Self::And(preds) => Self::Or(preds.into_iter().map(Self::negate).collect()),
             Self::Or(preds) => Self::And(preds.into_iter().map(Self::negate).collect()),
             Self::Not(pred) => *pred, // TODO: should we recursively simplify?
-            Self::Raw(pred) => Self::Not(Box::new(PredicateBox::Raw(pred))),
+            Self::Raw(pred) => Self::Not(Box::new(Self::Raw(pred))),
+        }
+    }
+}
+
+pub trait PredicateTraitExt<I>: Sized {
```
appetrosyan, line 82:
```suggestion
pub trait LogicalCombinator<I>: Sized + core::ops::Neg {
```

Might also want to replace the `negate` with an implementation of `core::ops::Neg` as a trait bound. At the time, given that `&&` and `||` were not overloadable, I thought it made sense to have these in a separate trait. Now it seems that two different implementations of `neg` are redundant.

data_model/src/predicate.rs
```diff
@@ -3,38 +3,66 @@
...
     /// Negate the result of applying the predicate.
-    Not(Box<PredicateBox>),
+    Not(Box<Self>),
     /// The raw predicate that must be applied.
-    Raw(value::Predicate),
+    Raw(P),
 }
 
-impl PredicateBox {
+impl<T> Display for GenericPredicateBox<T>
```
appetrosyan, line 24: Please omit this for now.

data_model/src/predicate.rs
```diff
@@ -3,38 +3,66 @@
...
 
+use core::fmt::Display;
+
 use super::*;
 use crate::{IdBox, Name, Value};
 
 /// Predicate combinator enum.
 #[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema)]
-pub enum PredicateBox {
+pub enum GenericPredicateBox<P> {
```
appetrosyan, line 13:
```suggestion
pub enum GenericPredicateBox<P> {
```

Should we enforce a trait bound?

data_model/src/predicate.rs
appetrosyan: Note that the standard library doesn't support `core::fmt::Display` for `alloc::vec::Vec`. This is intentional.

data_model/src/predicate.rs
```diff
@@ -46,10 +74,83 @@ impl PredicateBox {
...
+    T: PredicateTrait<I>,
+    I: Copy,
+{
+}
+
+impl<T> From<T> for GenericPredicateBox<T> {
+    fn from(pred: T) -> Self {
+        GenericPredicateBox::new(pred)
+    }
+}
```
appetrosyan, line 107: Very bold assumption. I'd omit it unless it's necessary.

data_model/src/predicate.rs
```diff
@@ -46,10 +74,83 @@ impl PredicateBox {
...
+
+impl<T, I> PredicateTrait<I> for GenericPredicateBox<T>
```
appetrosyan, line 109: These types can have longer names, and in this case **should**. I'd call `I` `TrivialType`. I'd rename `T` with `P`, for `Predicate<TrivialType>`.

data_model/src/predicate.rs
```diff
@@ -46,10 +74,83 @@ impl PredicateBox {
...
+            Self::And(predicates) => {
+                let mut operands = predicates.iter().map(|predicate| predicate.applies(input));
+                let mut value = operands.next().unwrap();
+                for operand in operands {
+                    if operand.is_truthful() {
+                        value = value.and(operand);
+                    } else {
+                        value = operand;
+                        break;
+                    }
```
appetrosyan, line 129: I'd add a comment: 
```rust
// This implements the short-circuiting logic. We need this because…
```

data_model/src/predicate.rs
```diff
@@ -46,10 +74,83 @@ impl PredicateBox {
...
+    #[inline] // This is not a simple function, but it allows you to inline the logic and optimise away the logical operations.
+    fn applies(&self, input: I) -> Self::EvaluatesTo {
+        match self {
+            Self::Raw(predicate) => predicate.applies(input),
+            Self::And(predicates) => {
+                let mut operands = predicates.iter().map(|predicate| predicate.applies(input));
+                let mut value = operands.next().unwrap();
+                for operand in operands {
+                    if operand.is_truthful() {
+                        value = value.and(operand);
```
appetrosyan, line 125: Not sure we need this specifically. It looks to me that we're getting a little too caught up with not knowing whether we're dealing with ternary or binary logic.

appetrosyan (approved):

core/src/smartcontracts/isi/permissions/judge.rs
```diff
@@ -467,7 +467,7 @@ pub mod builder {
         #[inline]
         fn new<V: IsAllowed<Operation = O> + Send + Sync + 'static>(validator: V) -> Self {
             Self {
-                validators: vec![Box::new(validator)],
+                validators: vec![validator.boxed()],
```
appetrosyan, line 470:
```suggestion
                validators: vec![validator.box()],
```

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -72,6 +125,38 @@ pub enum ValidatorVerdict {
     Allow,
 }
 
+impl PredicateSymbol for ValidatorVerdict {
```
appetrosyan, line 128: this is getting into the realm of ternary logic and a rookie mistake. 

We can't apply boolean arithmetic to ternary constructs, but thankfully (thanks to some abstract logicians) you can decouple ternary logic into binary logic plus extensions. 

### True + False + Skip = (True * Skip) + (False * Skip) 

You want to have two extra methods: `false()`, and `skip()`. So your `is_truthful()` ⇒ 
```rust 
fn true(&self) -> Option<bool> {
	match self {
           Self::Accept(_) => Some(true), 
	   Self::Deny(_) => Some(false), 
           _ => None,
       }
}
```

And then you use the monadic methods on `Option` to track the necessary skipping behaviour. You get rid of the associated `EvaluatesTo` type and **ordinary** rust logic applies. This follows the design of `PartialEq`. 

So in places where you'd use `is_truthful` which is more accurately `not_false`, you'd use `true().unwrap_or(false)`, and similarly for the other method. 

Then you get the short-circuiting and ignoring behaviour for basically free.

appetrosyan (changes requested):

data_model/src/predicate.rs
QuentinI: I assume you meant `core::ops::Not`. Implemented.

data_model/src/predicate.rs
QuentinI: That is unfortunately needed because of [this](https://github.com/hyperledger/iroha/blob/b2769df18bce2295cf39192ee2f8b3949946a4e6/core/src/smartcontracts/isi/permissions/mod.rs#L35).
I would argue that requiring `Display` isn't the optimal choice there either, but that warrants its own discussion and is out of scope for this PR.

data_model/src/predicate.rs
QuentinI: We can't without polluting everything downstream with explicit lifetimes, I think, since `PredicateTrait` is generic over a reference to input type, which is `&Value` for filter predicates. If we change `PredicateTrait` to be generic over input type directly (`Value`), this problem pops up for permission validators, which accept a tuple of references. And we can't make input associated type instead, since some predicates are applicable over all types implementing certain trait.

Constructor though can and should've enforce this, which I added.
In addition to that we could make the enum private and wrap it into an opaque struct, which would deny every possible way of instantiating `GenericPredicateBox` with a type not implementing `PredicateTrait`. I find access to enum variants potentially useful, so I guess I would be weakly opposed to such a solution.

data_model/src/predicate.rs
QuentinI: It was indeed unnecessary, forgot to remove it. Thanks!

data_model/src/predicate.rs
QuentinI: I updated this module with more explicit generic names, see if you think it fits.

core/src/smartcontracts/isi/permissions/judge.rs
QuentinI: Sorry, not sure I follow. Do you propose renaming `boxed` to `box`?

core/src/smartcontracts/isi/permissions/mod.rs
QuentinI: We cannot get rid of `EvaluatesTo` entirely, since `ValidatorVerdict` contains state we want to preserve.
I've rewritten the logic more clearly with `ControlFlow`, see if it looks better this way.

core/src/smartcontracts/isi/permissions/judge.rs
appetrosyan: Yes. Minor point, but `box` can be both treated as a noun (`box` containing this object) and a verb (put this thing into a `Box`).

core/src/smartcontracts/isi/permissions/mod.rs
appetrosyan: Like I said, the problem is the ternary logic. If you insist on doing it this way, please add tests, to demonstrate that it is indeed a compliant implenentation of https://www.wikidoc.org/index.php/Ternary_logic

data_model/src/predicate.rs
appetrosyan: OK. I'd revisit this later. Leave a comment to that effect.

data_model/src/predicate.rs
```diff
@@ -46,10 +77,45 @@ impl PredicateBox {
...
+}
+
+impl<Pred, Input> PredicateTrait<Input> for GenericPredicateBox<Pred>
+where
+    Input: ?Sized + Copy,
+    Pred: PredicateTrait<Input>,
+{
+    type EvaluatesTo = Pred::EvaluatesTo;
 
+    #[inline] // This is not a simple function, but it allows you to inline the logic and optimise away the logical operations.
```
appetrosyan, line 92: Good point. Though, you'd have to recognise that it is in fact a simple function once compiled. Most of the text is type noise, just like accessor methods.

core/src/smartcontracts/isi/permissions/mod.rs
QuentinI: Ah, okay. I didn't quite understand your suggestion.
`ValidatorVerdict` does not and should not conform to *the* ternary logic as per link you provided, since that would e.g. mean that `Skip or Deny = Skip`, which would be a security issue. I'm not sure if `(Validators, ValidatorVerdicts, Operations, Operators)` conforms to definition of _a_ first-order logic or propositional calculus, but I don't think it is really relevant here; what is relevant is that definitions of `and`, `or` and `not` make sense under the circumstances and conform to De Morgan's laws, which I think they do.
That said, I absolutely agree that it's complex enough logic that it warrants being tested.

core/src/smartcontracts/isi/permissions/judge.rs
QuentinI: Just realized `box` is a keyword. What's our stance on `r#`-escaped identifiers? Alternatively this could be called `into_boxed()`

appetrosyan: Boxed is fine then.

data_model/src/predicate.rs
```diff
@@ -2,54 +2,155 @@
...
 
-impl PredicateBox {
+impl<P> Display for GenericPredicateBox<P>
+where
+    P: Display,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            GenericPredicateBox::And(predicates) => {
+                write!(f, "OR(")?;
```
Erigara, line 51: Is this correct? we match `And` and print `OR`.

data_model/src/predicate.rs
```diff
@@ -2,54 +2,155 @@
...
+    {
         Self::Raw(pred.into())
     }
 
     /// Construct [`PredicateBox::And`] variant.
     #[inline]
-    pub fn and(left: impl Into<PredicateBox>, right: impl Into<PredicateBox>) -> Self {
-        Self::And(vec![left.into(), right.into()])
+    pub fn and(left: impl Into<Self>, right: impl Into<Self>) -> Self {
+        Self::And(nonempty![left.into(), right.into()])
```
Erigara, line 86: Why do we need `NonEmpty` if `And` and `Or` is binary operations?

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -72,6 +140,62 @@ pub enum ValidatorVerdict {
...
+/// |------|-------|-------|-------|
+/// |Allow | Allow | Allow | Allow |
+/// |Skip  | Allow | Skip  | Deny  |
+/// |Deny  | Allow | Deny  | Deny  |
+impl PredicateSymbol for ValidatorVerdict {
+    fn and(self, other: Self) -> ControlFlow<Self, Self> {
+        match (&self, &other) {
+            (Self::Deny(_), _) => ControlFlow::Break(self),
+            (Self::Skip, Self::Skip) => ControlFlow::Continue(Self::Skip),
+            _ => ControlFlow::Continue(other),
```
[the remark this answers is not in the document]
QuentinI: Right, thanks! This definitely needs tests, I'll add them.

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -6,15 +6,22 @@
 )]
 #![allow(clippy::module_name_repetitions)]
 
-use core::{fmt::Display, marker::PhantomData, ops::Deref};
+use core::{
+    fmt::Display,
+    marker::PhantomData,
+    ops::{Deref, Not},
+};
+use std::ops::ControlFlow;
```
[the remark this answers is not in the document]
QuentinI: Good catch, thanks

data_model/src/predicate.rs
QuentinI: Incorrect indeed, thanks

data_model/src/predicate.rs
QuentinI: `And` and `Or` allow any number of operands, without NonEmpty it would be possible to manually construct empty `And` variant, which we can't evaluate since `PredicateSymbol` doesn't require `Default` implementation (and I think shouldn't, since it isn't really well-defined for e.g. ValidatorVerdict, and for boolean we would want different defaults in different contexts)

Erigara: > `And` and `Or` allow any number of operands

But `and/or` accepts only two arguments, so we always constructing tree-like structure `add(a, add(b, c))`.

Is user supposed to manually create multi-argument `And/Or`? 

My suggestion is that we can inspect `left` and if it is `And` for `fn and` (`Or` for `fn or`) we can append `right` to argument list.

Otherwise we can restrict `And` as `And(Self, Self)`.

data_model/src/predicate.rs
```diff
@@ -2,54 +2,155 @@
...
 use alloc::vec;
+use core::{fmt::Display, ops::Not};
 
 use super::*;
 use crate::{IdBox, Name, Value};
 
+/// Struct representing sequence of non-zero length
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema)]
+pub struct NonEmpty<T> {
+    head: Box<T>,
```
Erigara, line 13: Why do we want `Box<T>` and not just `T`?

data_model/src/predicate.rs
QuentinI: > But and/or accepts only two arguments, so we always constructing tree-like structure add(a, add(b, c)).

That is true, I was worried about user manually constructing invalid PredicateBox
> My suggestion is that we can inspect left and if it is And for fn and (Or for fn or) we can append right to argument list.

Agreed, I'll implement it

data_model/src/predicate.rs
QuentinI: `GenericPredicateBox` becomes self-referential otherwise

core/src/smartcontracts/isi/permissions/has_token.rs
```diff
@@ -64,11 +64,7 @@ impl<H: HasToken> IsAllowed for HasTokenAsValidator<H> {
     ) -> ValidatorVerdict {
         let permission_token = match self.has_token.token(authority, instruction, wsv) {
             Ok(concrete_token) => concrete_token.into(),
-            Err(err) => {
-                return ValidatorVerdict::Deny(format!(
-                    "Unable to identify the corresponding permission token: {err}",
-                ));
-            }
+            Err(_) => return ValidatorVerdict::Skip,
```
QuentinI, line 67: **IMPORTANT**
This here were, I think, two bugs cancelling each other: validator replied `Deny` instead of `Skip` for inapplicable instructions, but they weren't checked.
I _think_ `HasToken::token` doesn't produce errors that should make validator reply `Deny`, but I am not sure, since it's `Error` doesn't bear any additional context. I think if we merge this, we should open an issue to refactor `HasToken::token` return type and check this line again.

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -72,6 +139,78 @@ pub enum ValidatorVerdict {
...
+/// | Or   | Allow | Skip  | Deny  |
+/// |------|-------|-------|-------|
+/// |Allow | Allow | Allow | Allow |
+/// |Skip  | Allow | Skip  | Deny  |
+/// |Deny  | Allow | Deny  | Deny  |
+///
+impl PredicateSymbol for ValidatorVerdict {
+    #[allow(clippy::unnested_or_patterns)] // More verbose, but more readable here
+    fn and(self, other: Self) -> ControlFlow<Self, Self> {
+        match (self, other) {
```
Erigara, line 166: I think some branches could be condensed, but it's fine this way too.

core/src/smartcontracts/isi/permissions/mod.rs
```diff
@@ -386,6 +524,73 @@ mod tests {
         ))
     }
 
+    // Convenience macro to concisely test truth tables
+    macro_rules! assert_maps {
```
Erigara, line 528: Cool macros!

data_model/src/predicate.rs
```diff
@@ -2,54 +2,177 @@
...
-        Self::And(vec![left.into(), right.into()])
+    pub fn and(left: impl Into<Self>, right: impl Into<Self>) -> Self {
+        match (left.into(), right.into()) {
+            (Self::And(mut left), Self::And(right)) => {
+                left.tail.push(*right.head);
+                left.tail.extend(right.tail);
+                Self::And(left)
+            }
+            (Self::And(mut and), other) | (other, Self::And(mut and)) => {
+                and.tail.push(other);
```
Erigara, line 93: We should be careful with reordering of predicates.
Operation is commutative so result would be the same but we can degrade early exit accidentally.

Erigara (approved):

core/src/smartcontracts/isi/permissions/mod.rs
QuentinI: They could, I just find it more readable this way. I don't think it impacts performance much if at all.

data_model/src/predicate.rs
QuentinI: True, another level of pointer indirection should be a better trade-off than degraded early exit. I removed out-of-order matches here.

core/src/smartcontracts/isi/permissions/mod.rs
Erigara: Ok, let's leave as is.

Erigara (approved):

core/src/smartcontracts/isi/permissions/has_token.rs
mversic: did you write a test for confirmation?

QuentinI: @mversic All permission tests pass with this PR, but so they do on current master, so our test suite is incomplete either way. I'm investigating the issue currently, won't merge this until I at least know for sure where the bug is.

data_model/src/predicate.rs
```diff
@@ -204,13 +204,181 @@ impl Default for PredicateBox {
...
+        $(assert!($eq(
+            &<$typ as PredicateSymbol>::unwrapped_or(
+                $sym.clone(),
+                $syms.clone()
+            ),
+            &<$typ as PredicateSymbol>::unwrapped_or(
+                $syms.clone(),
+                $sym.clone()
+            )
+        ));)*
```
Erigara, line 318: I'm not sure that we are testing `associativity` here, because associativity is `(a && b) && c == a && (b && c)`.
And we are testing `a && b == b && a` which is `commutativity`.

QuentinI: Oh right 🤦‍♂️
I think I mixed them up basically everywhere I used the term, I'll fix it

Erigara: Also i think this testing is not exhaustive because we don't test **every** pair of elements first element is always used.

QuentinI: We recur without the first symbol at the end

Erigara: We need tests for `associativity` because when we adding thing to `NonEmpty` we are relying on `(a && b) && c == a && b && c`.

QuentinI: Hmm, true enough
I'll add them as well

Erigara: > We recur without the first symbol at the end

Indeed, my bad, i missed recursion.

QuentinI: @Erigara fixed both issues.

<details> 
<summary> Also I realized there was absolutely no need to make this a macro.</summary>

</details>

data_model/src/lib.rs
```diff
@@ -1005,19 +1006,214 @@ impl From<LengthLimits> for RangeInclusive<usize> {
...
+    ///    let good = Check::Good;
+    ///    let bad = Check::Bad("example".to_owned());
+    ///    // Would fail some assertions, since derived PartialEq is "deep"
+    ///    // PredicateSymbol::test_conformity(vec![good, bad]);
+    ///
+    ///    // Works as expected
+    ///    PredicateSymbol::test_conformity_with_eq(vec![good, bad], shallow_eq);
+    /// }
+    /// ```
+    fn test_conformity_with_eq(values: Vec<Self>, shallow_eq: impl FnMut(&Self, &Self) -> bool)
```
Erigara, line 1128: `shallow_eq` could be `impl Fn(&Self, &Self) -> bool` or even `fn(&Self, &Self) -> bool` to me it's seems unlikely that we ever want to capture context with `shallow_eq` function. 

But i will approve regardless.

Erigara (dismissed):

data_model/src/lib.rs
QuentinI: Usually I just choose the bare minimum I need/will plausibly need not to add any additional restrictions on the caller.
Are there any benefits in further restricting `shallow_eq`?

Erigara: > Usually I just choose the bare minimum I need/will plausibly need not to add any additional restrictions on the caller.

Sounds like reassuring strategy. 

> Are there any benefits in further restricting `shallow_eq`?

Compile time if we are talking about fn pointers.

data_model/src/predicate.rs
appetrosyan: Yes. Typo

data_model/src/predicate.rs
appetrosyan: > I would argue that requiring Display isn't the optimal choice there either, but that warrants its own discussion and is out of scope for this PR.

Agree. In fact supertraits seem to be used as shorthands for trait combinations a whole lot in Iroha. We should address this in the style guide.

data_model/src/lib.rs
```diff
@@ -1135,10 +1141,243 @@ impl From<LengthLimits> for RangeInclusive<usize> {
...
+    }
+
+    fn or(self, other: Self) -> ControlFlow<Self, Self> {
+        if self || other {
+            ControlFlow::Break(true)
+        } else {
+            ControlFlow::Continue(false)
+        }
+    }
+}
```
appetrosyan, line 1372: While normally the teranry ⇒ binary monadic correspondence is written with an `Option` or a monoid like `Result`, I like this representation. It's more vivid and more explicit about the interaction of different behaviours. I'm glad you decided to rewrite it this way.

core/src/smartcontracts/isi/permissions/mod.rs
appetrosyan: > it's an overkill, but I think it's important, since iroha security fundamentally depends on Validators.

No. It's not overkill, and it's good that it's written as a test. Ideally  these things are written using a formal verification harness, but in this particular case, we need any violation of the logic to be detectable.

core/src/smartcontracts/isi/permissions/has_token.rs
appetrosyan: I'd address this in a separate PR, as is, the code here is more important to merge sooner rather than later.

appetrosyan (dismissed):

Erigara (approved):

data_model/src/predicate.rs
```diff
@@ -2,54 +2,184 @@
...
 
 use super::*;
 use crate::{IdBox, Name, Value};
 
+/// Struct representing sequence of non-zero length
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, IntoSchema)]
+pub struct NonEmpty<T> {
+    head: Box<T>,
+    tail: Vec<T>,
+}
```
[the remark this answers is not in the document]
QuentinI: Right, I didn't think of it as a hot path for some reason. I'll follow up with an adequate implementation.

appetrosyan (approved):

Erigara (approved):

PR #2910, opened 2022-10-25
Iroha2 dev

appetrosyan:
### Description of the Change
Remove dead code and improve coverage by excluding binary entrypoints which should be tested with UI/UX

### Issue
None

### Benefits

More coverage

### Possible Drawbacks

Mocking is harder to implement if we decide to backtrack on it.


core/src/sumeragi/main_loop.rs
```diff
@@ -1007,11 +1000,12 @@ pub fn run<F>(
                         }
 
                         trace!("I, the proxy tail, have received a block.");
-                        for event in Vec::<Event>::from(&block) {
-                            trace!(?event);
-                            sumeragi.events_sender.send(event).unwrap_or(0);
+                        if sumeragi.events_sender.receiver_count() > 0 {
```
mversic, line 1003: ah, this is what we needed. The pattern is pervasive, see if you can extract it into a function

appetrosyan: I'm going to fix this separately in #2969.

appetrosyan: And backport a fix to LTS.

client/build.rs
```diff
@@ -114,7 +114,7 @@ fn build_smartcontract(smartcontract_path: &Path, out_dir: &OsStr) -> Result<()>
         .env_remove("CARGO_ENCODED_RUSTFLAGS")
         .env("CARGO_TARGET_DIR", out_dir)
         .current_dir(smartcontract_path)
-        .args(&[
+        .args([
             "+nightly-2022-08-15",
```
mversic, line 118: is this a good time to bump the version? We haven't bumped it in a while and we'd finally get stable GATs

mversic, line 118: is this a good time to bump the version? We haven't bumped it in a while and we'd finally get stable GATs

rust-toolchain.toml
```diff
@@ -1,5 +0,0 @@
-[toolchain]
-channel = "1.63"
```
mversic, line 2: I thought we got rid of this file already. I noticed we didn't since there were some improvements in the `1.64` borrow checker I depend on but I see the code compiling on `nightly-2022-08-15`. So either bumping the version or removing the file works for me :)

config/base/src/runtime_upgrades.rs
```diff
@@ -208,7 +196,7 @@ pub mod handle {
 
     impl<T: Send + Sync + Debug> Reload<T> for Singleton<T> {
         fn reload(&self, item: T) -> Result<()> {
-            match &mut *self.inner.lock()? {
+            match &mut *self.inner.lock().expect("Valid") {
```
mversic, line 199: then why don't we use non poisoning mutex as in sumeragi?

telemetry/derive/src/lib.rs
```diff
@@ -66,6 +66,7 @@ fn arg_metrics(input: &Punctuated<FnArg, Comma>) -> Result<syn::Ident, &Punctuat
         .ok_or(input)
 }
 
+#[allow(unused_tuple_struct_fields)]
```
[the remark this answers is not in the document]
appetrosyan: No. It 100% is dead code, just not something that we should immediately rip out.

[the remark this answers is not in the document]
appetrosyan: Come to think of it, this is something that the comment should say

config/base/src/runtime_upgrades.rs
appetrosyan: We should. It's like this, because this PR was started before we had `parking_lot::Mutex`

Issue #2690, opened 2022-09-02
[BUG]: Local `docker build` fails on M1 Mac


s8sato:
### GIT commit hash

02283d16

### Minimum working example

```bash
COMMIT=02283d16
```
```bash
git checkout $COMMIT
```
```bash
docker build -t 2690:base-$COMMIT -t iroha2:base -f Dockerfile.base .
```
```bash
docker build -t 2690:build-$COMMIT -t iroha2:build -f Dockerfile.build .
```
```bash
docker build -t 2690:final-$COMMIT .
```

### Expected behaviour

Succeeds, regardless of the host platform (confirmed by e.g. `uname -io`)

### Actual behaviour

Fails on M1 Mac

### Operating system

x86_64 GNU/Linux (Succeeds to build)

### Current environment

Source code build

### Logs in JSON format

<details>
  <summary>Log contents</summary>
  
  ```json
  Replace this text with a JSON log,
  so it doesn't grow too large and has highlighting.
  ```
</details>

### Who can help?

@s8sato

s8sato:
```bash
bash 2690-inspect.sh before >before.log
```
```bash
bash 2690-inspect.sh fixme >fixme.log
```

<details>
<summary>2690-inspect.sh</summary>

```bash
#!/bin/bash
set -eu

# configure

ISSUE=2690
TOOLCHAINS=()

case $1 in

before)
    COMMIT='70e782bb'
    STAGE_BASE='rust-base'
    STAGE_BUILD='builder'
    TOOLCHAINS+=('stable')
    TOOLCHAINS+=('nightly-2022-04-20')
    ;;

fixme)
    COMMIT='02283d16'
    TAG_BASE='iroha2:base'
    TAG_BUILD='iroha2:build'
    TOOLCHAINS+=('stable')
    TOOLCHAINS+=('nightly-2022-04-20')
    TOOLCHAINS+=('nightly-2022-04-20-x86_64-unknown-linux-musl')
    ;;

*)
    exit 1
    ;;

esac

# build

git checkout $COMMIT

case $1 in

before)
    docker build -t $ISSUE:base-$COMMIT --target $STAGE_BASE . -q
    docker build -t $ISSUE:build-$COMMIT --target $STAGE_BUILD . -q
    # docker build -t $ISSUE:final-$COMMIT . -q
    ;;

fixme)
    docker build -t $ISSUE:base-$COMMIT -t iroha2:base -f Dockerfile.base . -q
[...]

PR #2911, opened 2022-10-25
[ci] #2690: Add cargo chef caching

BAStos525:
### Description of the Change
1. Enable Cargo cache the dependencies with `cargo chef`
2. Substitute `Codecov` by `Coverall` coverage solution

### Issue
1. No caching for cargo crates: issue #2690 
2. Codecov coverage `diff` doesn't work correctly: issue #2781 

### Benefits
1. Speed up the iroha2 CI
2. Apply a possible way to fix Codecov coverage `diff` by replacing it by `Coveralls`

### Possible Drawbacks
1. Within the current PR, telemetry couldn't be compiled:
```
#20 933.9    Compiling iroha_client_cli v2.0.0-pre-rc.9 (/client_cli)
#20 934.2 error: failed to run custom build command for `iroha_telemetry v2.0.0-pre-rc.9 (/telemetry)`
#20 934.2 
#20 934.2 Caused by:
#20 934.2   process didn't exit successfully: `/target/release/build/iroha_telemetry-aee0cffb91c378a9/build-script-build` (exit status: 1)
#20 934.2   --- stderr
#20 934.2   Error: could not find repository from '/telemetry'; class=Repository (6); code=NotFound (-3)
```
Could be a reason in the `Cargo.lock` and `deploy profile`? @s8sato @appetrosyan 

2. `Coverall` coverage solution is not really tested.


appetrosyan (dismissed):

appetrosyan:
We've tried something like this and it didn't work out. This was actually the first step in optimising the CI. 

If you try to cache with `cache@v3` you'll end up discarding the layer cache each time `Cargo.lock` changes, which isn't what you want, given that `Cargo.lock` can change for reasons not related to dependencies. 

What I suggest we do instead, is continue our work on using a custom CI image. If we bake the compiled dependencies into the CI image and trigger a re-creation of the CI image each time we have a change in `Cargo.lock` **after** the PR is merged, we can make use of a partially compiled set of dependencies (`cargo` is smart enough to find them when it needs to), improve the compile times for regular workflows, not just the part where we publish the docker image, and win on PR latency.

appetrosyan:
As for the coverage diff… Myself and @6r1d came to the conclusion that codecov.io has a lot of fancy features that we can't use, and is far too unstable. We want to try using Coveralls.

BAStos525:
What if we will revert to keeping ci-image as dedicated docker image that is compiled only when its corresponding `Dockerfile` has been changed: perhaps from iroha building on M1 perspective it doesn't matter either builder is presented as cache or as docker image that should be pulled from DockerHub - it should not be compiled on M1 in any case. Along with this, we can try an attempt to cache `cargo` dependencies. It will also help do not have an over limit of Github cache size.

BAStos525:
> As for the coverage diff… Myself and @6r1d came to the conclusion that codecov.io has a lot of fancy features that we can't use, and is far too unstable. We want to try using Coveralls.

If it works with RUST language correctly... Just didn't find RUST in the list of supported languages on the Coveralls site.

appetrosyan:
> If it works with RUST language correctly... Just didn't find RUST in the list of supported languages on the Coveralls site.

Good point. It is supported, but we'd need to run some tests before we make the switch, particularly to know if it accurately shows coverage changes. 

https://github.com/corebreaker/coveralls-rust

Dockerfile.planner
```diff
@@ -0,0 +1,8 @@
+#planner stage
```
BAStos525, line 1: Should we keep image with cargo cached dependencies separately from Linux Arch based `ci-image`?

Dockerfile
```diff
@@ -1,23 +1,7 @@
...
-    rustup +$NIGHTLY target add x86_64-unknown-linux-musl wasm32-unknown-unknown && \
-    rustup +$NIGHTLY component add rust-src llvm-tools-preview && \
-    # cargo install
-    cargo install cargo-lints webassembly-test-runner && \
-    :
-
-# builder stage
-WORKDIR /iroha
 COPY . .
 RUN  rm -f rust-toolchain.toml
```
BAStos525, line 6: Should we delete `rust-toolchain.toml` file before `cargo build` command or before `cargo chef prepare` && `cargo chef cook` commands on the `planner` stage?

appetrosyan: We should delete it from the repository. Add it to `.gitignore` too.

Dockerfile.planner
appetrosyan: Unfortunately, this takes us back to the architecture that we moved away from. I don't think that doing things this way is tenable. It's preferable for us to compile the binary separately and simply copy it into the container and produce an image.

BAStos525: Do you mean to combine Arch based ci-image and `cargo chef` in one Dockerfile.base file? Or to join `cargo chef` and `cargo build` together? In this case, we have to pay attention to `Cargp.lock` change trigger and where to use it properly.

appetrosyan: > Do you mean to combine Arch based ci-image and cargo chef in one Dockerfile.base file?

Yes. 

> In this case, we have to pay attention to Cargo.lock change trigger and where to use it properly.

We should just have a manual workflow trigger.  This is  needed anyway so that we can update the image when there's a new toolchain. 

Cargo lock changes far more often than is necessary to re-run this operation. If one line changed, it makes the compilation process take 1/600th longer, if that. Re-caching the dependencies only makes sense if we have a large rewrite and version bump all of the external dependencies. Plus if `Cargo.lock` is slightly outdated, the compilation will take only slightly longer.

Dockerfile
BAStos525: Without `rust-toolchain.toml` we got this error while trying to compile ci-image: `error: rustup could not choose a version of rustup to run, because one wasn't specified explicitly, and no default is configured.`

Dockerfile.planner
BAStos525: > > Do you mean to combine Arch based ci-image and cargo chef in one Dockerfile.base file?
> 
> Yes.
> 
It was combined in one `Dockerfile.base`

PR #2912, opened 2022-10-25
[documentation]: Extend tutorial with asset-related docs

6r1d:
Signed-off-by: 6r1d <vic.6r1d@gmail.com>

### Description of the Change

Addresses https://github.com/hyperledger/iroha-2-docs/issues/207 and https://github.com/hyperledger/iroha-2-docs/issues/208

### Benefits

Documentation examples need to cover more topics.

### Possible Drawbacks

None

### Usage Examples or Tests

Those are the code examples to be used in the documentation

### DIFF
--- a/client/examples/tutorial.rs
+++ b/client/examples/tutorial.rs
@@ -28,6 +28,8 @@ fn main() {
         .expect("Account registration example is expected to work correctly");
     asset_registration_test(&config)
         .expect("Asset registration example is expected to work correctly");
+    asset_minting_test(&config).expect("Asset minting example is expected to work correctly");
+    asset_burning_test(&config).expect("Asset burning example is expected to work correctly");
     println!("Success!");
     // END ESCAPE
 }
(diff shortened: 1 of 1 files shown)


client/examples/tutorial.rs
```diff
@@ -185,3 +189,56 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
     // Finish the test successfully
     Ok(())
 }
+
+fn asset_minting_test() -> Result<(), Error> {
```
mversic, line 193:
```suggestion
fn asset_minting_test() {
```
you're never returning error in the test

mversic: or you could use `?` instead of `expect` which I don't think is a good idea in this case

6r1d: Thanks!

client/examples/tutorial.rs
```diff
@@ -185,3 +189,56 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
...
+    // BEGIN FRAGMENT: mint_asset
+    use std::str::FromStr;
+    use iroha_data_model::Value;
+    use iroha_data_model::IdBox;
+    use iroha_data_model::prelude::MintBox;
+    use iroha_data_model::prelude::AssetId;
+    use iroha_data_model::prelude::AssetDefinitionId;
+    use iroha_data_model::prelude::AccountId;
```
Erigara, line 201: you can combine imports:

```suggestion
use iroha_data_model::{
    IdBox, Value, prelude::{
        AccountId, AssetId, AssetDefinitionId, MintBox, 
    }
 }
```

Also you can import prelude all at once `prelude::*`.

client/examples/tutorial.rs
```diff
@@ -185,3 +189,56 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
...
+        Value::U32(42),
+        IdBox::AssetId(AssetId::new(roses.clone(), alice.clone())),
+    );
+    // END FRAGMENT
+
+    // Finish the test successfully
+    Ok(())
+}
+
+fn asset_burning_test() -> Result<(), Error> {
```
Erigara, line 218: Is it intentional that you do nothing with `MintBox/BurnBox`?
We need to submit instruction in order to execute it.

6r1d: Yes, I believe we should test the data structures here, not the whole Iroha. It allows to run tests both locally and not, and this will be used in the doctests. I was thinking about modifying the `dst-parser`, so some comments are uncommented and the submit instructions are added, too. I am not sure yet.

client/examples/tutorial.rs
6r1d: Thank you.

client/examples/tutorial.rs
```diff
@@ -185,3 +189,55 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
...
+    // BEGIN FRAGMENT: burn_asset
+    use std::str::FromStr;
+    use iroha_data_model::Value;
+    use iroha_data_model::IdBox;
+    use iroha_data_model::prelude::BurnBox;
+    use iroha_data_model::prelude::AssetId;
+    use iroha_data_model::prelude::AssetDefinitionId;
+    use iroha_data_model::prelude::AccountId;
```
Erigara, line 225: Same thing about imports

```suggestion
    use iroha_data_model::{
        IdBox, Value, prelude::{
            AccountId, AssetId, AssetDefinitionId, MintBox, 
        }
     }
```

client/examples/tutorial.rs
```diff
@@ -185,3 +189,55 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
     // Finish the test successfully
     Ok(())
 }
+
+fn asset_minting_test() {
+    // BEGIN FRAGMENT: mint_asset
+    use std::str::FromStr;
+use iroha_data_model::{
```
Erigara, line 196: Something with formatting here.

client/examples/tutorial.rs
6r1d: UPD: now that I've re-read the code, I have noticed the request are sent in the other parts and I will update it accordingly.

6r1d:
Returning to the previous discussion on the exceptions and error propagation, I am leaving ` -> Result<(), Error> {}` as-is and using `.expect("Message")`. This is the intended behavior for the library code we've discussed with @appetrosyan and it'll be marked as such in the documentation.

client/examples/tutorial.rs
```diff
@@ -185,3 +187,55 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
...
+    };
+    // Create an Iroha client
+    let iroha_client: Client = Client::new(&config).unwrap();
+    // Define the instances of an Asset and Account
+    let roses = AssetDefinitionId::from_str("rose#wonderland")
+        .expect("Valid, because the string contains no whitespace, has a single '#' character and is not empty after");
+    let alice: AccountId = "alice@wonderland".parse()
+        .expect("Valid, because the string contains no whitespace, has a single '@' character and is not empty after");
+    // Mint the Asset instance
+    let mint_roses = MintBox::new(Value::U32(42), IdBox::AssetId(AssetId::new(roses, alice)));
```
appetrosyan, line 208: Add section with alternative syntax: 
```rust
let mint_roses_again = MintBox::new(Value::U32(24), IdBox::AssetId("rose##alice@wonderland".parse()?));
// The syntax is `asset_name#asset_domain#account_name@account_domain`, or `roses.to_string() + "#" + alice.to_string()`.
// The `##` is a short-hand for the rose `which belongs to the same domain as the account to which it belongs to. 
```

client/examples/tutorial.rs
```diff
@@ -185,3 +187,55 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
...
+    iroha_client
+        .submit(mint_roses)
+        .wrap_err("Failed to submit transaction")?;
+    // END FRAGMENT
+    // Finish the test successfully
```
appetrosyan, line 213:
```suggestion

```

client/examples/tutorial.rs
```diff
@@ -185,3 +187,55 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
...
+        .expect("Valid, because the string contains no whitespace, has a single '#' character and is not empty after");
+    let alice: AccountId = "alice@wonderland".parse()
+        .expect("Valid, because the string contains no whitespace, has a single '@' character and is not empty after");
+    // Burn the Asset instance
+    let burn_roses = BurnBox::new(Value::U32(10), IdBox::AssetId(AssetId::new(roses, alice)));
+    iroha_client
+        .submit(burn_roses)
+        .wrap_err("Failed to submit transaction")?;
+    // END FRAGMENT
+    // Finish the test successfully
```
appetrosyan, line 239:
```suggestion

```

client/examples/tutorial.rs
appetrosyan: You could do this in the `Burn` test, though.

6r1d: Understood, will apply to both in the morning!

Erigara (approved):

appetrosyan (approved):

PR #2913, opened 2022-10-26
Rebuild p2p

SamHSmith:


core/Cargo.toml
```diff
@@ -59,6 +58,8 @@ tokio = { version = "1.20.1", features = ["sync", "time", "rt", "io-util", "rt-m
 crossbeam-queue = "0.3.6"
 thiserror = "1.0.32"
 wasmtime = "0.39.1"
+aead = "0.3.2"
```
appetrosyan, line 61: Careful

client/tests/integration/connected_peers.rs
```diff
@@ -33,7 +33,7 @@ fn connected_peers_with_f(faults: u64) {
     let mut status = genesis_client.get_status().unwrap();
     assert_eq!(status.peers, n_peers - 1);
     assert_eq!(status.blocks, 1);
-
+println!("GOT HERE 1");
```
appetrosyan, line 36: GDB?

cli/src/torii/routing.rs
```diff
@@ -427,7 +427,7 @@ async fn handle_version(sumeragi: Arc<Sumeragi>) -> Json {
 
 #[cfg(feature = "telemetry")]
 #[allow(clippy::unused_async)] // TODO: remove?
-async fn handle_metrics(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -> Result<String> {
+async fn handle_metrics(sumeragi: Arc<Sumeragi>, _: usize) -> Result<String> {
```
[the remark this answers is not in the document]
SamHSmith: The macros broke and I don't have the skillset to investigate why.

core/src/block_sync.rs
```diff
@@ -25,66 +26,100 @@ pub struct BlockSynchronizer {
...
+    let mut message_receiver = block_sync.message_receiver.lock().unwrap();
+    let mut last_requested_blocks = Instant::now();
+    let mut request_blocks_peer_index = 0_usize;
+    loop {
+        // We have no obligations to network delivery so we simply exit on shutdown signal.
+        if shutdown_receiver.try_recv().is_ok() {
+            info!("P2P thread is being shut down");
+            return;
+        }
+        std::thread::sleep(Duration::from_millis(10));
```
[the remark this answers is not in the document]
appetrosyan: Sleep is necessary to yield the OS thread, though frankly the way we're doing it: rigidly, with arbitrary values, is error-prone. What happens if we have another problem on a weaker machine on which the sleep interval must be lower/higher? 

The only viable option I can think of at this hour is coroutines and callback functions. 

Given that we need `async` anyway, the implementation is likely going to be easier than it looks.

cli/Cargo.toml
```diff
@@ -47,7 +47,6 @@ iroha_telemetry = { version = "=2.0.0-pre-rc.9", path = "../telemetry", optional
 iroha_version = { version = "=2.0.0-pre-rc.9", path = "../version", features = ["warp"] }
 iroha_config = { version = "=2.0.0-pre-rc.9", path = "../config" }
 iroha_crypto = { version = "=2.0.0-pre-rc.9", path = "../crypto" }
-iroha_p2p = { version = "=2.0.0-pre-rc.9", path = "../p2p" }
```
appetrosyan, line 50: I'm not against this shift, just trying to understand if this is an æsthetic preference or a hard requirement given the tighter coupling.

cli/src/lib.rs
```diff
@@ -22,15 +22,14 @@ use iroha_core::{
     genesis::{GenesisNetwork, GenesisNetworkTrait, RawGenesisBlock},
     handler::ThreadHandler,
     kura::Kura,
+    p2p::P2PSystem,
```
appetrosyan, line 25: Needs a better name.

cli/src/lib.rs
```diff
@@ -79,13 +78,11 @@ pub struct Iroha {
     /// Kura — block storage
     pub kura: Arc<Kura>,
     /// Block synchronization actor
-    pub block_sync: AlwaysAddr<BlockSynchronizer>,
+    pub block_sync: Arc<BlockSynchronizer>,
```
appetrosyan, line 81: There's only one of those things, and it's pretty much static. Should we try and make it static?

cli/src/lib.rs
```diff
@@ -337,8 +277,13 @@ impl Iroha {
...
             block_sync,
             torii,
-            thread_handlers: vec![sumeragi_thread_handler, kura_thread_handler],
-            _sumeragi_relay: sumeragi_relay,
+            thread_handlers: vec![
+                sumeragi_thread_handler,
+                block_sync_thread_handler,
+                kura_thread_handler,
+                p2p_thread_handler1,
+                p2p_thread_handler2,
```
appetrosyan, line 285: We have thing number 1 and thing number 2. Are there really no distinguishing features between the two?

cli/src/torii/routing.rs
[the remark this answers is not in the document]
appetrosyan: You need to use an `endpoint` with a lower number of arguments in place of this one. Look at other routing functions with one argument to see how it's done.

cli/src/torii/routing.rs
```diff
@@ -440,7 +440,7 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -
 
 #[cfg(feature = "telemetry")]
 #[allow(clippy::unused_async)] // TODO: remove?
-async fn handle_status(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -> Result<Json> {
+async fn handle_status(sumeragi: Arc<Sumeragi>, _: usize) -> Result<Json> {
```
appetrosyan, line 443: Same here.

cli/src/torii/routing.rs
```diff
@@ -478,11 +476,11 @@ impl Torii {
     ) -> impl warp::Filter<Extract = impl warp::Reply> + Clone + Send {
         let get_router_status = endpoint2(
```
appetrosyan, line 477: This is the culprit. The `2` here is the number of arguments. You need to replace it with the inline code that would have been generated for `1` (we don't have an `endpoint1` macro).

client/tests/integration/connected_peers.rs
```diff
@@ -46,7 +46,7 @@ fn connected_peers_with_f(faults: u64) {
     assert_eq!(status.blocks, 2);
     status = peer_client.get_status().unwrap();
     assert_eq!(status.peers, 0);
-    assert_eq!(status.blocks, 2);
+    assert_eq!(status.blocks, 1);
```
appetrosyan, line 49: Interesting. Have you investigated why the number of blocks changed?

client/tests/integration/restart_peer.rs
```diff
@@ -57,7 +57,9 @@ fn restarted_peer_should_have_the_same_asset_amount() -> Result<()> {
             .expect("Asset not found");
         assert_eq!(AssetValue::Quantity(quantity), *asset.value());
     }
-
+    peer.iroha.take();
+    // Make sure previous peer has time to shutdown.
+    std::thread::sleep(std::time::Duration::from_secs(1));
```
appetrosyan, line 62: This is fragile. 

We need some form of message passing so that this peer can wait for the other one to shutdown. 

Consider using an `async` function which blocks until the other peer has shutdown.

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -10,6 +10,8 @@ use test_network::*;
 
 const TRIGGER_NAME: &str = "mint_rose";
 
+// nocheckin why it not work
```
appetrosyan, line 13: While you're at it, see if adding extra diagnostics to this path makes a difference.

core/src/block_sync.rs
```diff
@@ -25,66 +26,100 @@ pub struct BlockSynchronizer {
...
-    broker: Broker,
-    actor_channel_capacity: u32,
-}
-
-#[async_trait::async_trait]
-impl Actor for BlockSynchronizer {
-    fn actor_channel_capacity(&self) -> u32 {
-        self.actor_channel_capacity
-    }
+    p2p: Arc<P2PSystem>,
```
appetrosyan, line 29: I'm pretty sure that this is also a static object.

core/src/block_sync.rs
```diff
@@ -25,66 +26,100 @@ pub struct BlockSynchronizer {
...
+    /// Sender channel
+    pub message_sender: Mutex<mpsc::SyncSender<message::VersionedMessage>>,
```
appetrosyan, line 31: Yet another pair of `Mutex`. Why not just have a single `Mutex` with a tuple of two channels?

appetrosyan: Better yet, consider alternatives which don't need to be wrapped in a `Mutex`.

core/test_network/src/lib.rs
```diff
@@ -844,6 +844,7 @@ impl TestConfiguration for Configuration {
         configuration
             .load_environment()
             .expect("Failed to load configuration from environment");
+        configuration.logger.max_log_level = iroha_logger::Level::TRACE.into();
```
appetrosyan, line 847: BAD!

appetrosyan: You have the `configuration.load_environment()` use that to set the log level to `TRACE`.

core/src/sumeragi/mod.rs
```diff
@@ -143,9 +138,8 @@ impl Sumeragi {
     pub fn update_metrics(&self) -> Result<()> {
         let online_peers_count: u64 = self
             .internal
-            .current_online_peers
-            .lock()
-            .expect("Failed to lock `current_online_peers` for `update_metrics`")
+            .p2p
+            .get_connected_to_peer_keys()
```
appetrosyan, line 142:
```suggestion
            .keys_of_peer_connected_to()
```

core/src/sumeragi/main_loop.rs
```diff
@@ -495,6 +453,7 @@ fn handle_role_agnostic_messages<F>(
 #[allow(clippy::too_many_arguments)]
 fn compare_view_change_index_and_block_height_to_old<F>(
     sumeragi: &SumeragiWithFault<F>,
+    view_change_proof_chain_for_sharing: &Vec<Proof>,
```
appetrosyan, line 456:
```suggestion
    view_change_proofs: &Vec<Proof>,    
```

core/src/sumeragi/main_loop.rs
```diff
@@ -523,6 +482,16 @@ fn compare_view_change_index_and_block_height_to_old<F>(
         current_topology.rebuild_with_new_view_change_count(current_view_change_index);
 
         // there has been a view change, we must reset state for the next round.
+        if current_view_change_index > 0 {
+            // Notify others.
```
appetrosyan, line 486:
```suggestion

```

appetrosyan (changes requested):

core/src/p2p.rs
```diff
@@ -0,0 +1,571 @@
...
+            return;
+        }
+        std::thread::sleep(Duration::from_micros(2500));
+
+        // For protocol violators.
+        let mut to_disconnect_keys = Vec::<PublicKey>::new();
+
+        let target = p2p.connect_peer_target.lock().unwrap().clone();
+        let mut connected_to_peers = p2p.connected_to_peers.lock().unwrap();
+        for (public_key, (stream, crypto)) in connected_to_peers.iter_mut() {
```
appetrosyan, line 202: Notice that in this *function*, the mantra: 
```rust 
if something.is_err() {
	to_disconnect_keys.push(public_key.clone();
	continue;
}
``` 
is repeated. The overwhelming redundancy of this statement should signal to you that in fact a custom error type would have been helpful. 

A much better solution would have been to put the entire body of the `for` loop inside a function. Forward the errors up that function with the `?` operator, and then do a single `if let Err(_)` at the top level. In the only two cases where you short-circuit without adding the key to the disconnected keys you should return `Ok`, and wrap the `else` block of the `if let` wrap the remainder of the function.

core/src/p2p.rs
```diff
@@ -0,0 +1,571 @@
...
+) {
+    let listener = TcpListener::bind(&p2p.listen_addr).expect("Could not bind p2p tcp listener.");
+    listener
+        .set_nonblocking(true)
+        .expect("P2P subsystem could not enable nonblocking on listening tcp port.");
+
+    let time_slice = Duration::from_millis(5);
+    
+    let mut countdown_to_outgoing = 20;
+    loop {
```
appetrosyan, line 295: Same pattern, same problems.

core/src/p2p.rs
```diff
@@ -0,0 +1,571 @@
...
+
+    let shutdown = move || {
+        let _result = shutdown_sender.send(());
+    };
+
+    ThreadHandler::new(Box::new(shutdown), thread_handle)
+}
+
+// nocheckin replace println's with trace's.
+fn p2p_listen_loop(
```
appetrosyan, line 283: Prefixing the name of the function with its common namespace, and describing the function by its shape, rather than purpose, is an anti-pattern common in OOP. 

More importantly it is **not at all obvious** what's the difference between a `listen` and a `read` loop.

appetrosyan (changes requested):
Overall, I have mixed feelings about this PR as it is. 

On the one hand, the threading model seems a much more natural fit for the `p2p` system, and the explicit `loop` statements with a shutdown handle using a `onseshot` channel are beautifully simple. As a `mod` rather than a full-blown `crate` it's also much more at home in the only crate that uses it. In some ways the refactor away from type-state makes it more readable. 

However, the same old anti-patterns are present. There was no need to have a huge function  with a `main_loop` inside. There are easily stages of the handshake process that, if extracted into functions, would have simplified the error handling code greatly. The comments keep saying **what** is being done, the thing that can be inferred by looking at the code, as opposed to **why** something is done the way it's done. Monadic patterns: `map` and short-circuiting with the `?` operator are ignored completely. Those are meant to remedy the exact anti-patterns and code-duplication which causes the greatest readability issues. 

It's difficult to tell if there are errors. This makes any performance advantage this could have completely irrelevant. If it's impossible to tell if the code does the right thing it's not good code.

core/src/block_sync.rs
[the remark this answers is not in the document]
Erigara: > Sleep is necessary to yield the OS thread

Yes, i understand that, my concern is about taken architectural approach. 

We are burning cycles in a loop every 10ms when 1) shutdown is one time event 2) gossip period is 10000ms in default configuration and 3) and i highly doubt that we will receive messages from `message_reciever` with such frequency.

Tasks here (and in other places) seems to be about waiting a lot and there is mechanism to do this more effectively.

core/src/p2p.rs
```diff
@@ -0,0 +1,571 @@
...
+    public_key: PublicKey,
+    connect_peer_target: Mutex<Vec<PeerId>>,
+    connected_to_peers: Mutex<HashMap<PublicKey, (TcpStream, Cryptographer)>>,
+}
+
+impl std::fmt::Debug for P2PSystem {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::result::Result<(), std::fmt::Error> {
+        write!(f, "P2PSystem")
+    }
+}
```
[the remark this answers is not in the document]
Erigara: Could be derived with `derive_more`

cli/Cargo.toml
SamHSmith: It is a hard requirement.

cli/src/lib.rs
SamHSmith: We should make them all static. After we have rewritten the tests so that they can support the change.

core/src/block_sync.rs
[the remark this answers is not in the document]
appetrosyan: > Yes, i understand that, my concern is about taken architectural approach.

Sorry, the reason I mentioned this, is because one of Iroha's targets is an OS-less deployment on embedded devices, which might not have a decent scheduler.

cli/src/lib.rs
appetrosyan: Created #2919 resolving.

core/src/p2p.rs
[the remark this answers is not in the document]
appetrosyan: Make sure you `#[skip]` the problematic fields @SamHSmith

core/src/block_sync.rs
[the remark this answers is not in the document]
Erigara: > one of Iroha's targets is an OS-less deployment on embedded devices, which might not have a decent scheduler

Insteresting, it's our plans for the future to add support for such target?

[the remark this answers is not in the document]
appetrosyan: Yes. Long-term plan.

core/src/p2p.rs
```diff
@@ -0,0 +1,694 @@
...
+        }
+
+
+        connected_to_peers.insert(other_public_key, (stream, crypto));
+    }
+}
+
+const ESTABLISH_CONNECTION_TIME_SLICE: Duration = Duration::from_millis(25);
+
+unsafe fn establish_new_connection(p2p: &P2PSystem, listener: &TcpListener) -> Option<TcpStream> {
```
Erigara, line 429: Why `unsafe`???

core/src/p2p.rs
```diff
@@ -0,0 +1,694 @@
...
+            stream.set_nonblocking(false).unwrap();
+            return Some(None);
+        };
+        if byte_count == 0 {
+            return Some(None);
+        }
+        let mut packet_size_buf = [0_u8; 4];
+        stream.read_exact(&mut packet_size_buf).ok()?;
+        packet_size = u32::from_le_bytes(packet_size_buf);
+    }
```
Erigara, line 347: A bit odd construction.

core/src/p2p.rs
```diff
@@ -0,0 +1,694 @@
...
+
+    let mut buf = vec![0_u8; packet_size as usize];
+    stream.read_exact(&mut buf).ok()?;
```
Erigara, line 350: We are hiding errors.

Issue #2914, opened 2022-10-26
[suggestion]: Extend `FindAssetQuantityById` to return `NumericValue`


Erigara:
### Feature request

I'd like to be able to query for asset quantity not only for `u32` type, but for other types as well (`u128`, `Fixed`).

PR #2915, opened 2022-10-26
[feature] #2096: Query for total amount of asset

### LINKED ISSUE #2096: Query for total amount of asset
In the blockchain explorer we often need to consider the total number of an asset in a network. 
Ideally, we only need a query and some caching so that expensive computations aren't done more than once, and only updated when they need to be (e.g. when an asset is minted).

Erigara:
### Description of the Change

- Add `NumericValue` to `Value`;
- Add macroses `PartiallyTaggedSerialize`, `PartaillyTaggedDeserialize`;
- Fix deadlock when unregistering `AssetDefinition`;
- Extend `Domain` with total assets quantity map;
- Add query `FindAssetTotalQuantitiyBeAssetDefinitionId`;
- Change return type of `FindAssetQuantityById` from `u32` to `NumericValue`;
- `Burn`/`Transfer`/`Mint` instructions now accepts `NumericValue`.

### Issue

Closes #2096, #2914.

### Benefits

New query + extended `FindAssetQuantityById` query.

### Possible Drawbacks

None.

### Usage Examples or Tests

```
cargo test --package iroha_client --test mod -- integration::queries::asset::find_asset_total_quantity --exact --nocapture 
```


data_model/src/query.rs
```diff
@@ -975,7 +977,34 @@ pub mod asset {
...
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+        PartialOrd,
+        Ord,
+        Hash,
+    )]
+    #[display(fmt = "Find quantity of the `{}` asset", id)]
+    pub struct FindTotalAssetQuantityByAssetDefinitionId {
```
outoftardis, line 1001: comment to self: add this to docs

data_model/src/query.rs
```diff
@@ -975,7 +977,34 @@ pub mod asset {
     }
 
     impl Query for FindAssetQuantityById {
-        type Output = u32;
+        type Output = AssetValue;
```
outoftardis, line 980: note to sefl: update output value in docs

mversic: who is sefl?

data_model/src/lib.rs
```diff
@@ -532,6 +532,8 @@ pub enum Value {
     Ipv4Addr(iroha_primitives::addr::Ipv4Addr),
     /// IP Version 6 address.
     Ipv6Addr(iroha_primitives::addr::Ipv6Addr),
+    /// AssetValue
+    AssetValue(AssetValue),
```
mversic, line 536: now we seem to have some redundancy here because the variants of the `AssetValue` are also variants of the `Value`

data_model/src/events/data/events.rs
```diff
@@ -140,6 +142,27 @@ mod asset {
...
+        Hash,
+        Debug,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    #[allow(missing_docs)]
+    pub struct AssetDefinitionTotalQuantityChanged {
```
mversic, line 162: do we need this event? it should be implied by events emitted when asset is minted/burned

data_model/src/domain.rs
```diff
@@ -179,6 +182,8 @@ declare_item! {
         accounts: AccountsMap,
         /// [`Asset`](AssetDefinition)s defined of the `Domain`.
         asset_definitions: AssetDefinitionsMap,
+        /// Total amount of [`Asset`].
+        asset_total_quantities: AssetTotalQuantityMap,
```
mversic, line 186: so you're storing total assets by domain, not by blockchain? Is it possible to mint asset for an account in another domain? like mint `asset#domain1` for `account@domain2`. I think it is possible

core/src/smartcontracts/isi/mod.rs
```diff
@@ -419,11 +425,15 @@ impl Execute for TransferBox {
...
-            Value::Fixed(quantity) => {
+            Value::AssetValue(AssetValue::Quantity(quantity)) | Value::U32(quantity) => {
+                Transfer::new(source_asset_id, quantity, destination_asset_id)
+                    .execute(authority, wsv)
+            }
+            Value::AssetValue(AssetValue::BigQuantity(quantity)) | Value::U128(quantity) => {
+                Transfer::new(source_asset_id, quantity, destination_asset_id)
+                    .execute(authority, wsv)
+            }
+            Value::AssetValue(AssetValue::Fixed(quantity)) | Value::Fixed(quantity) => {
```
mversic, line 436: this redundancy is exactly what I was referring to in the comment on `Value`

data_model/src/lib.rs
Erigara: Agree with you, maybe we can introduce something like this? 

```rust
enum NumericValue {
   U32(u32),
   U128(u128),
   Fixed(Fixed),
}

enum Value  {
    // ...
    Numeric(NumericValue),
    // ...
    // delete U32, U128 and Fixed variants
}

// we can still accept `u32`, `u128` and `Fixed` by implementing `TryFrom<T> for Value`.

enum AssetValue {
     Numeric(NumericValue),
     Store(Metadata),
}
```

core/src/smartcontracts/isi/asset.rs
```diff
@@ -585,10 +639,28 @@ pub mod query {
...
+                .id
+                .evaluate(wsv, &Context::default())
+                .wrap_err("Failed to get asset definition id")
+                .map_err(|e| Error::Evaluate(e.to_string()))?;
+            iroha_logger::trace!(%id);
+
+            let asset_definition_entry = wsv.asset_definition_entry(&id)?;
+            let asset_definition = asset_definition_entry.definition();
+            if let AssetValueType::Store = asset_definition.value_type() {
+                return Err(Error::Conversion(String::from("`AssetValueType::Store`")));
```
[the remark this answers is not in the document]
Erigara: We can have counter to track quantity of `Store` assets. 

> number of elements in the store across all accounts 

This sound too artificial for me.

data_model/src/domain.rs
Erigara: No, we are storing total amount through blockchain. I just thought to store total assets would be convenient to store inside `Domain` because we already store `AssetsDefinition` this way. 

> I think it is possible

Yeap, this is very possible, i cover such scenario in [test case](https://github.com/hyperledger/iroha/blob/67bdfb030cffe3cee00a90e5034a53174cbe01c9/client/tests/integration/queries/asset.rs#L29) (`white_rabbit@looking_glass` belongs to different domain).

mversic: Excuse my mistake

data_model/src/asset.rs
```diff
@@ -519,7 +524,7 @@ impl HasMetadata for NewAssetDefinition {
 #[cfg_attr(feature = "ffi_import", iroha_ffi::ffi_import)]
 impl NewAssetDefinition {
     /// Create a [`NewAssetDefinition`], reserved for internal use.
-    fn new(id: <AssetDefinition as Identifiable>::Id, value_type: AssetValueType) -> Self {
+    pub fn new(id: <AssetDefinition as Identifiable>::Id, value_type: AssetValueType) -> Self {
```
mversic, line 527: this shouldn't be a public fn. You have at your disposal `AssetDefinition::quantity`, `AssetDefinition::store`, etc

Erigara: Why not? It is not possible to create any kind of inconsistent asset definition using `NewAssetDefinition::new`.

I made this function public because i want to create different types of asset definitions inside loop.
I can use function pointers instead, but it sounds a bit more complicated.

mversic: it's a matter of consistent API. We don't expose any methods on builders. You can add this function to `AssetDefinition` and use `AssetDefinition::new(definition_id.clone(), asset_value_type)`:
```rs
    pub fn new(id: <Self as Identifiable>::Id, value_type: AssetValueType) -> <Self as Registered>::With {
        <Self as Registered>::With::new(id, value_type)
    }
```

mversic (dismissed):

data_model/src/events/data/events.rs
appetrosyan: Recommend keeping it.

appetrosyan (approved):

tools/parity_scale_decoder/samples/trigger.json
```diff
@@ -6,7 +6,9 @@
                 {
                     "Mint": {
                         "object": {
-                            "U32": 1
+                            "Numeric": {
+                                "U32": 1
+                            }
```
appetrosyan, line 11: This is problematic. Please make sure that this serialises and deserialises the way it used to.

appetrosyan (changes requested):

client/tests/integration/sorting.rs
```diff
@@ -75,8 +75,8 @@ fn correct_pagination_assets_after_creating_new_one() {
     new_asset_metadata
         .insert_with_limits(
             sort_by_metadata_key,
-            Value::U128(10),
-            MetadataLimits::new(10, 22),
+            Value::from(10_u128),
```
appetrosyan, line 78: `10_u128.to_value()`.

configs/peer/genesis.json
```diff
@@ -237,8 +237,8 @@
               "PermissionTokenDefinition": {
                 "id": "can_transfer_only_fixed_number_of_times_per_period",
                 "params": {
-                  "count": "U32",
-                  "period": "U128"
+                  "count": "Numeric",
+                  "period": "Numeric"
```
appetrosyan, line 241: This is good, please keep it

configs/peer/genesis.json
```diff
@@ -247,7 +247,9 @@
         {
           "Mint": {
             "object": {
-              "U32": 13
+              "Numeric": {
+                "U32": 13
+              }
```
appetrosyan, line 252: These are redundant, and only pad the length.

core/src/smartcontracts/isi/asset.rs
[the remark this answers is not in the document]
appetrosyan: I don't mind the artificial approach. The number of entries is rarely confidential.

core/src/smartcontracts/isi/domain.rs
```diff
@@ -93,6 +93,21 @@ pub mod isi {
...
+                        domain.add_asset_total_quantity(asset_definition_id.clone(), Fixed::ZERO);
+                    }
+                    AssetValueType::Quantity => {
+                        domain.add_asset_total_quantity(asset_definition_id.clone(), u32::MIN);
+                    }
+                    AssetValueType::BigQuantity => {
+                        domain.add_asset_total_quantity(asset_definition_id.clone(), u128::MIN);
+                    }
+                    AssetValueType::Store => {
+                        // `Store` asset value type doesn't have total quantity
```
appetrosyan, line 107: Or it does as per @mversic's comment.

docs/source/references/schema.json
```diff
@@ -47,6 +47,13 @@
       "sorted_by_key": true
     }
   },
+  "Map<iroha_data_model::asset::DefinitionId, iroha_data_model::NumericValue>": {
+    "Map": {
+      "key": "iroha_data_model::asset::DefinitionId",
+      "value": "iroha_data_model::NumericValue",
+      "sorted_by_key": true
+    }
+  },
```
appetrosyan, line 56: Lots of changes in schema. Could cause problems for SDKs.

core/src/smartcontracts/isi/asset.rs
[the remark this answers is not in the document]
Erigara: @appetrosyan @mversic i want to clarify this a bit.
You want to store number of registered assets of type `Store` for particular `AssetDefinition` or total number of key value pairs across all assets for this `AssetDefinition`?

client/examples/tutorial.rs
```diff
@@ -175,7 +176,7 @@ fn asset_registration_test(config: &Configuration) -> Result<(), Error> {
 
     // Create a MintBox using a previous asset and account
     let mint = MintBox::new(
-        Value::Fixed(12.34_f64.try_into()?),
+        Value::Numeric(NumericValue::Fixed(12.34_f64.try_into()?)),
```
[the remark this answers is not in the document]
Erigara: We can introduce `ToValue` and `TryToValue` because `u128` always can be converted to `Value` and for `f64` this is not always possible.

data_model/src/domain.rs
```diff
@@ -232,6 +236,15 @@ impl Domain {
         self.asset_definitions.get(asset_definition_id)
     }
 
+    /// Return a reference to the asset definition corresponding to the asset definition id
+    #[inline]
+    pub fn asset_total_quantity(
+        &self,
+        asset_definition_id: &<AssetDefinition as Identifiable>::Id,
+    ) -> Option<&NumericValue> {
```
ilchu, line 244: With the change to accept `AssetValueType::Store` for total amount, this one would also become non-optional and more ergonomic IMO.

core/src/smartcontracts/isi/asset.rs
```diff
@@ -150,6 +150,7 @@ pub mod isi {
     impl_transfer!(Fixed, "transfer_fixed");
 
     /// Trait for blanket mint implementation.
+    #[allow(clippy::trait_duplication_in_bounds)] // False positive
```
ilchu, line 153: Maybe file an issue to clippy? They had something [similar](https://github.com/rust-lang/rust-clippy/issues/9076) fixed recently with this lint, though.

client/tests/integration/queries/asset.rs
```diff
@@ -0,0 +1,168 @@
...
+            "fixed#wonderland",
+            AssetValueType::Fixed,
+            Value::Numeric(NumericValue::Fixed(Fixed::try_from(10.0)?)),
+            Value::Numeric(NumericValue::Fixed(Fixed::try_from(5.0)?)),
+            NumericValue::Fixed(Fixed::try_from(25.0)?),
+        ),
+    ] {
+        // Registering new asset definition
+        let definition_id: <AssetDefinition as Identifiable>::Id =
+            definition.parse().expect("Valid");
```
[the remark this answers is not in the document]
Erigara: Sure!

core/src/smartcontracts/isi/asset.rs
Erigara: I checked this bug for rust 1.63-65, error message is gone on rust 1.65.

data_model/src/domain.rs
Erigara: @ilchu, you mean that we can `.expect` inside function body and not outside?

core/src/smartcontracts/isi/asset.rs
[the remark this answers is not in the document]
Erigara: I have decided to stick with the first approach.

client/tests/integration/queries/asset.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        let burn_asset = assets
+            .iter()
+            .cloned()
+            .map(|asset_id| BurnBox::new(to_burn.clone(), asset_id).into())
+            .collect::<Vec<_>>();
+        test_client.submit_all_blocking(burn_asset)?;
+
+        // Assert that total asset quantity is equal to: `sum(initial_values) + sum(mints) - sum(burns)`
+        let total_asset_quantity = test_client.request(
```
ilchu, line 145: This was a bit confusing to me on first read. Took a while to realize that those initial values are repeated 5 times for each respective account.

data_model/src/expression.rs
```diff
@@ -373,115 +374,65 @@ mod operation {
...
+    Not(Not),
+    /// And expression.
+    And(And),
+    /// Or expression.
+    Or(Or),
+    /// If expression.
+    If(If),
+    /// Raw value.
+    #[serde_partially_tagged(untagged)]
+    Raw(ValueBox),
```
ilchu, line 423: AFAICT the new proc macro is a complete replacement for the declarative macro we had here, and also more general? Good news again.

Erigara: > AFAICT the new proc macro is a complete replacement for the declarative macro we had here, and also more general?

Yes

data_model/src/events/data/events.rs
```diff
@@ -115,7 +113,11 @@ mod asset {
                 | Self::Deleted(id)
                 | Self::MintabilityChanged(id)
                 | Self::MetadataInserted(MetadataChanged { target_id: id, .. })
-                | Self::MetadataRemoved(MetadataChanged { target_id: id, .. }) => id,
+                | Self::MetadataRemoved(MetadataChanged { target_id: id, .. })
+                | Self::TotalQuantityChanged(AssetDefinitionTotalQuantityChanged {
+                    asset_definition_id: id,
+                    ..
+                }) => id,
```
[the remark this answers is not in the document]
Erigara: Sure, there is plenty room for improvements, feel free to create ticket.

core/src/smartcontracts/isi/mod.rs
```diff
@@ -623,25 +627,25 @@ mod tests {
     use super::*;
     use crate::{wsv::World, PeersIds};
 
-    fn world_with_test_domains() -> Result<World> {
-        let mut domain = Domain::new(DomainId::from_str("wonderland")?).build();
+    fn wsv_with_test_domains() -> Result<WorldStateView> {
```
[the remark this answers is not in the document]
Erigara: Created ticket #2942

client/tests/integration/queries/asset.rs
```diff
@@ -0,0 +1,260 @@
...
+        test_client.submit_all_blocking(mint_asset)?;
+
+        let burn_asset = assets
+            .iter()
+            .cloned()
+            .map(|asset_id| BurnBox::new(to_burn.clone(), asset_id).into())
+            .collect::<Vec<_>>();
+        test_client.submit_all_blocking(burn_asset)?;
+
+        // Assert that total asset quantity is equal to: `sum(initial_values) + sum(mints) - sum(burns)`
```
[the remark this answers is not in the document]
Erigara: I rephrased this as: 

```rust
        // Assert that total asset quantity is equal to: `n_accounts * (initial_value + to_mint - to_burn)`
``` 

Should be more readable.

client/tests/integration/queries/asset.rs
```diff
@@ -0,0 +1,260 @@
...
+        let result = test_client.request(FindTotalAssetQuantityByAssetDefinitionId::new(
+            definition_id.clone(),
+        ));
+        assert!(matches!(
+            result,
+            Err(ClientQueryError::QueryError(QueryError::Find(_)))
+        ));
+    }
+
+    // Test for `Store` asset value type
```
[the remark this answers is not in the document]
Erigara: I want this to be one test case because it would be faster to execute, because we this way we setup peers and wait for genesis only one time.

data_model/src/permission/token.rs
```diff
@@ -95,12 +95,17 @@ pub trait ValueTrait: Into<Value> {
...
+            const TYPE: ValueKind = $kind;
+        }
+    )+}
 }
 
-impl ValueTrait for u128 {
-    const TYPE: ValueKind = ValueKind::U128;
+impl_value_trait! {
+    u32: ValueKind::Numeric,
+    u128: ValueKind::Numeric
```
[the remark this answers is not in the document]
Erigara: As i understand this trait means that this type could be used as parameter to in `PermissionTokenDefinition` params map. 

I don't want to change behavior too much so decided to leave this as is.
However adding support would be pretty trivial if we need to.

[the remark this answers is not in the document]
ilchu: Fair enough, just felt it's an outlier of sorts.

data_model/src/events/data/events.rs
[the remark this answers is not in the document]
ilchu: Created #2943.

ilchu (approved):

core/src/smartcontracts/isi/account.rs
```diff
@@ -56,6 +79,21 @@ pub mod isi {
             let asset_id = self.object_id;
             let account_id = asset_id.account_id.clone();
 
+            match wsv.asset(&asset_id)?.value() {
+                AssetValue::Quantity(increment) => {
```
appetrosyan, line 83: Technically it's 
```suggestion
                AssetValue::Quantity(decrement) => {
```

docs/source/references/schema.json
appetrosyan: This warrants a #dev_update

client/tests/integration/queries/asset.rs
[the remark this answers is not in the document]
appetrosyan: While that is certainly true, splitting it off into functions is possible if in the future we want to refactor it down.

appetrosyan (approved):

PR #2916, opened 2022-10-26
[refactor] #2909: Hardcode ports for nextest

### LINKED ISSUE #2909: [BUG] Failing integration test: client_register_asset_should_add_asset_once_but_not_twice
### OS and Environment

[Artix Linux](https://artixlinux.org/), kernel `5.19.10`

### GIT commit hash

64078706a9ef27763a826916f38b942d9b8dfbdf

Test the open ports with:

```bash
# A)
lsof -Pn | grep LISTEN
# B)
netstat -plnt
```

Port `1024` should not be in use at the time.

```bash
$ cargo nextest run --no-fail-fast -p iroha_client
```

### Actual result

```
--- STDOUT:              iroha_client::mod integration::asset::client_register_asset_should_add_asset_once_but_not_twice ---

running 1 test
2022-10-25T11:09:05.044713Z  INFO iroha: Starting peer listen_addr=127.0.0.1:1024
2022-10-25T11:09:05.044824Z  INFO iroha_p2p::network: Binding listener listen_addr=127.0.0.1:1024
2022-10-25T11:09:05.045705Z  INFO test_network: Stopping peer p2p_addr=127.0.0.1:1024 api_addr=127.0.0.1:1025
test integration::asset::client_register_asset_should_add_asset_once_but_not_twice ... FAILED

failures:

failures:
    integration::asset::client_register_asset_should_add_asset_once_but_not_twice

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 77 filtered out; finished in 0.02s


--- STDERR:              iroha_client::mod integration::asset::client_register_asset_should_add_asset_once_but_not_twice ---
thread 'tokio-runtime-worker' panicked at 'Failed to start iroha: Unable to start P2P-network

Caused by:
   0: Failed IO operation.
   1: Address already in use (os error 98)

Location:
    /home/grid/git/iroha/cli/src/lib.rs:244:10', core/test_network/src/lib.rs:483:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'integration::asset::client_register_asset_should_add_asset_once_but_not_twice' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError', core/test_network/src/lib.rs:491:43
```

### Expected result

Tests should run normally

@ilchu @appetrosyan

ilchu:

### Description of the Change
For each test that is run by nextest in `iroha_client`, `iroha_data_model` and `iroha_p2p`, replace usage of `prepare_for_nextest!` macro with hardcoded ports. 

### Issue
Closes #2909.

### Benefits
Nextest now doesn't fail on either of those packages and finishes successfully. Full manual control.

### Possible Drawbacks
If any new tests are added, the programmer will have to keep track of ports already in use and maybe do some simple arithmetic. Also might have some debugging involved in case of failing to get it right at first. Also some code duplication introduced with parallel methods that just allow specifying a port.

### Usage Examples or Tests
`cargo nextest run -p iroha_client` or any other of the mentioned crates

### Alternate Designs
Making that `unique_port` crate work by making an inhouse solution with other primitives?..

### DIFF
--- a/client/tests/integration/multisignature_account.rs
+++ b/client/tests/integration/multisignature_account.rs
@@ -12,7 +12,7 @@ use super::Configuration;
 
 #[test]
 fn transaction_signed_by_new_signatory_of_account_should_pass() -> Result<()> {
-    let (_rt, peer, iroha_client) = <PeerBuilder>::new().start_with_runtime();
+    let (_rt, peer, iroha_client) = <PeerBuilder>::new().with_port(10_605).start_with_runtime();
     wait_for_genesis_committed(&vec![iroha_client.clone()], 0);
     let pipeline_time = Configuration::pipeline_time();
 
--- a/client/tests/integration/offline_peers.rs
+++ b/client/tests/integration/offline_peers.rs
@@ -7,11 +7,15 @@ use tokio::runtime::Runtime;
 
 #[test]
 fn genesis_block_is_commited_with_some_offline_peers() {
-    prepare_test_for_nextest!();
     // Given
     let rt = Runtime::test();
 
-    let (network, iroha_client) = rt.block_on(<Network>::start_test_with_offline(4, 1, 1));
+    let (network, iroha_client) = rt.block_on(<Network>::start_test_with_offline_and_set_n_shifts(
+        4,
+        1,
+        1,
+        Some(10_560),
+    ));
     wait_for_genesis_committed(&network.clients(), 1);
 
     //When
--- a/core/test_network/src/lib.rs
+++ b/core/test_network/src/lib.rs
@@ -44,18 +44,6 @@ use tokio::{
 };
 pub use unique_port;
 
-/// Prevent port collisions in `unique_port`, when using `cargo nextest`.
-#[macro_export]
-macro_rules! prepare_test_for_nextest {
-    () => {{
-        if std::env::var("NEXTEST").is_ok() {
-            use $crate::unique_port::{generate_unique_start_port, set_port_index};
-            set_port_index(generate_unique_start_port!())
-                .expect("Can't set port index for unique_port");
-        }
-    }};
-}
-
 #[derive(Debug, Clone, Copy)]
 struct ShutdownRuntime;
 
@@ -140,8 +128,6 @@ impl TestGenesis for GenesisNetwork {
             .into(),
         );
 
-        configure_world();
-
         GenesisNetwork::from_configuration(
             submit_genesis,
             genesis,
@@ -151,8 +137,6 @@ impl TestGenesis for GenesisNetwork {
         .expect("Failed to init genesis")
     }
 }
-
-fn configure_world() {}
 
 impl Network {
     /// Send message to an actor instance on peers.
@@ -191,17 +175,26 @@ impl Network {
     /// Starts network with peers with default configuration and
     /// specified options in a new async runtime.  Returns its info
     /// and client for connecting to it.
-    pub fn start_test_with_runtime(n_peers: u32, max_txs_in_block: u32) -> (Runtime, Self, Client) {
+    pub fn start_test_with_runtime(
+        n_peers: u32,
+        max_txs_in_block: u32,
+        start_port: Option<u16>,
+    ) -> (Runtime, Self, Client) {
         let rt = Runtime::test();
-        let (network, client) = rt.block_on(Self::start_test(n_peers, max_txs_in_block));
+        let (network, client) =
+            rt.block_on(Self::start_test(n_peers, max_txs_in_block, start_port));
         (rt, network, client)
     }
 
     /// Starts network with peers with default configuration and
     /// specified options.  Returns its info and client for connecting
     /// to it.
-    pub async fn start_test(n_peers: u32, max_txs_in_block: u32) -> (Self, Client) {
-        Self::start_test_with_offline(n_peers, max_txs_in_block, 0).await
+    pub async fn start_test(
+        n_peers: u32,
+        max_txs_in_block: u32,
+        start_port: Option<u16>,
+    ) -> (Self, Client) {
+        Self::start_test_with_offline(n_peers, max_txs_in_block, 0, start_port).await
     }
 
     /// Starts network with peers with default configuration and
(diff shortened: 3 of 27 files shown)


core/test_network/src/lib.rs
```diff
@@ -656,6 +772,13 @@ impl PeerBuilder {
         peer
     }
 
+    /// Create and start a peer with preapplied arguments at the specified port.
+    pub async fn start_with_port(self, port: u16) -> Peer {
```
pesterev, line 776: It seems you can add `port` into `PeerBuilder` like this:
```rust
pub struct PeerBuilder {
    configuration: Option<Configuration>,
    genesis: WithGenesis,
    // ...
    port: Option<u16>, // <===
}
```
and also add separate setter `with_port(self, port: u16) -> Self` into `PeerBuilder`

client/tests/integration/multisignature_account.rs
```diff
@@ -12,7 +12,7 @@ use super::Configuration;
 
 #[test]
 fn transaction_signed_by_new_signatory_of_account_should_pass() -> Result<()> {
-    let (_rt, peer, iroha_client) = <PeerBuilder>::new().start_with_runtime();
+    let (_rt, peer, iroha_client) = <PeerBuilder>::new().start_with_runtime_and_port(10_605);
```
appetrosyan, line 15: We could have a non-sanitary declarative macro like this:
```rust
test_network! { port: 10_605, ..defaults };
```
that expands to the same and where some things **could** be customised. That way we can get rid of the methods in `test_network` that just take up space.

client/tests/integration/offline_peers.rs
```diff
@@ -7,11 +7,11 @@ use tokio::runtime::Runtime;
 
 #[test]
 fn genesis_block_is_commited_with_some_offline_peers() {
-    prepare_test_for_nextest!();
     // Given
     let rt = Runtime::test();
 
-    let (network, iroha_client) = rt.block_on(<Network>::start_test_with_offline(4, 1, 1));
+    let (network, iroha_client) =
+        rt.block_on(<Network>::start_test_with_offline_and_port_and_set_n_shifts(4, 1, 1, 10_560));
```
appetrosyan, line 14: This could become 

```suggestion
test_network! {
	offline_peers: 1, 
	total_peers: 4, 
	n_shifts: 1, 
	port: 10_560
}
```
and give us `network` and `iroha_client` in scope as a side-effect.

Erigara: is it possible due to hygiene? ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=565508d01125e46c65c04bb1614be50b))

appetrosyan: https://gist.github.com/Kestrer/8c05ebd4e0e9347eb05f265dfb7252e1#procedural-macros-use-mixed_site-where-possible

Erigara (commented):
IMO it will be inconvenient to add new tests, because you are not only had to look for all occupied ports in tests, but also add additional offsets (for api port, telemetry, ...).

Since we don't check if this ports occupied by other programs in system, couldn't we use some kind of atomic counter `Atomic<u16>` with step of 10 that will assign ports.

client/tests/integration/offline_peers.rs
Erigara: Sure, i was confused by declarative macro.

core/test_network/src/lib.rs
appetrosyan: Much prefer the syntax `port(self, port: u16) -> Self`.

appetrosyan:
> IMO it will be inconvenient to add new tests, because you are not only had to look for all occupied ports in tests, but also add additional offsets (for api port, telemetry, ...).

Good. We have 77 which is at least 30 too many. 

But more importantly, if you do need to add a new test, you could add the test with a single `sleep` statement. If any other test uses the same port, you'll see a collision. It will tell you which test is using that particular port. 

Also you have `git grep`, which can be run in the tests dir, and see if that numerical pattern occurs anywhere else in the code. 
Or better yet, to bring all the occurrences in a human readable format in a nice table. 

> Since we don't check if this ports occupied by other programs in system, couldn't we use some kind of static atomic counter AtomicU16 with step of 10 that will assign ports.
EDIT: problem is that test would run in different processes and counter is shared only between threads of single process.

If we could, then Aleś solution would have done the job. 

But the idea with a global counter can be re-used. We can rely on the order of macro expansion to produce the counters via proc_macro. 

Finally, making the tests not fail if the ports are occupied is also a good option.

Erigara:
> But the idea with a global counter can be re-used. We can rely on the order of macro expansion to produce the counters via proc_macro.

I have tried to implement such approach, but seems like macro expands to the same value for different files.

```rust
const PORT_INCREMENT: u16 = 10;
static CURRENT_PORT: AtomicU16 = AtomicU16::new(20_000);

#[proc_macro]
pub fn unique_port(_: TokenStream) -> TokenStream {
    let port: u16 = CURRENT_PORT.fetch_add(PORT_INCREMENT, Ordering::SeqCst);
    quote!(#port).into()
}
```

Or what do you mean?

core/test_network/src/lib.rs
```diff
@@ -280,45 +283,83 @@ impl Network {
...
+                        }
+                    })
+                    .collect::<Vec<_>>()
+            },
+        );
+        let mut peers = builders
+            .iter_mut()
+            .map(PeerBuilder::build)
+            .collect::<Result<Vec<_>>>()?;
+        // builders.remove(0)
```
Erigara, line 325: Commented code.

core/test_network/src/lib.rs
```diff
@@ -280,45 +283,83 @@ impl Network {
...
+                                .with_into_genesis(GenesisNetwork::test(true))
+                        } else {
+                            PeerBuilder::new()
+                                .with_port(port + offset)
+                                .with_into_genesis(GenesisNetwork::test(false))
+                        }
+                    })
+                    .collect::<Vec<_>>()
+            },
+        );
```
Erigara, line 320: Consider:

```rust
let mut builders = core::iter::repeat_with(PeerBuilder::new)
    .enumerate()
    .map(|(n, builder)| {
        if let Some(port) = start_port {
            (n, builder.with_port(port + (n * 5)))
        } else {
            (n, builder)
        }
    })
    .map(|(n, builder)| builder.with_into_genesis(GenesisNetwork::test(n == 0)))
    .take(n_peers)
    .collect::<Vec<_>>();
```

Erigara (commented):
Does i understand correctly that we fallback to unique port if port is not provided manually?

core/test_network/src/lib.rs
ilchu: This is really neat, thanks!

ilchu:
> Does i understand correctly that we fallback to unique port if port is not provided manually?

Yes, I suppose that's the intended behaviour here. I could just make it use some predetermined port and throw out `unique_port` crate altogether, though.

core/test_network/src/lib.rs
ilchu: It felt to me that `with_port` goes more in line with the other builder setters.

client/tests/integration/offline_peers.rs
ilchu: Not going this route as we've discussed that the solution is good as it is, not worth a separate new crate with a non-hygenic macro for now.

Erigara:
I don't mind using `unique_port` as fallback, we can throw it out later.

Erigara (approved):

appetrosyan (approved):

pesterev (approved):

Issue #2163, opened 2022-04-28
Detailed logging of Error conditions


appetrosyan:
Every time any function inside Iroha returns an error, we should be able to find that failure in the logs (probably with a backtrace). 

As such it may be a good idea to enable a custom trait that logs all/some error conditions but only in Debug builds and is otherwise a no-op.

mversic:
especially this is relevant to logging errors comming from `wasm::Error`. For some reason only the top level error is displayed in the log and it's impossible to figure out what was the cause

PR #2917, opened 2022-10-26
[feature] #2163: Add trait to handle log-and-ignore pattern

appetrosyan:

### Description of the Change

The log-and-ignore pattern is all too common in our code base. I've added a trait that makes it much easier to enforce "best practices" for such errors. Specifically, errors now need to be explicitly ignored using an `expect`-like `ignored` function. 

For most errors their kind determines the severity. So instead of writing 
```rust
if let Err(e) = some_fallible_operation() {
    warn!(%e, "Some erorr occurred")
}
```

We write 
```rust
some_fallible_operation()
	.logged()
	.ignored("If `some_fallible_operation` failed, you probably have something wrong with your ports");
```
Less code, more information, the message in the `ignored` doubles as a comment for why we aren't handling the error. 

Additionally the code generated here will make use of the fact that `eyre::Report`'s `core::fmt::Debug` implementation prints the traceback, so as to print the entire traceback in the logs for the **debug** build, while reducing the verbosity in the **release** build. 

### Issue

#2163 

### Benefits

Less boilerplate, more diagnostic information when you need it. 

### Possible Drawbacks

For all errors handled this way, the `error` span is attached to the implementation for this error type and not the call-site where the error "first" occurred. 

Resolution pending 
https://discord.com/channels/500028886025895936/627649734592561152/1034897182333091850

A (postfix) macro based solution could do just as well. 

A prefix macro with a mini-syntax could resolve the issue of spans, but be more cumbersome to use. 

An attribute-like macro is also possible with the following syntax:
```rust 
#[log_on_err]
fn function_calling_fallible(… ) -> … {
	some_fallible_operation().logged().ignored("Helpful message");
	another_fallible_op()
		.map(do_something)
		.map_err(do_something_else)
		.logged!() // Expands in-place. 
		.map_err(one_last_time)
		.ignored("With this error"); 
	for e in results {
		e.is_err().break() now; // `.break()` returns `ControlFlow::Break` which then gets applied by the `now` keyword
	} // functions as `if let Err(e)` break. 
	result.if_error().short_circuit()
}
```
which corrects the span information and adds functionality and posprocesses the methods via keywords, but complicates the system and removes the flexibility of the trait implementation using the full-fat type level information.

### DIFF
--- a/logger/src/error.rs
+++ b/logger/src/error.rs
@@ -1,0 +1,159 @@
+//! Ergonomic log-and-ignore syntax.
+
+/// Trait used to pipeline the common log-and-ignore pattern by
+/// providing better handling of the process, and a clearer more
+/// expressive syntax.
+pub trait Logged {
+    /// Log the provided value. Useful for logging objects in a chain
+    /// of events, or before the `?` operator.
+    #[must_use]
+    fn logged(self) -> Self;
+
+    /// Ignore the provided error.
+    ///
+    /// # Motivation
+    ///
+    /// Sometimes errors cannot be sensibly handled, they can only be
+    /// reported to the user. If the error isn't handled further this
+    /// allows one to cheaply explain to the user of the debug build
+    /// **why** a particular error cannot be ignored.  More
+    /// importantly when one sees an explicit
+    /// ```rust
+    /// err.logged().ignore("I can't handle this error because of X")
+    /// ```
+    /// the reader of the code doesn't need to be provided
+    /// with a comment as to why the error message was ignored
+    #[inline]
+    fn ignored(self, reason: &str)
+    where
+        Self: Sized,
+    {
+        #[cfg(debug_assertions)]
+        crate::debug!(reason, "Ignored");
+    }
+
+    /// Promote the error from a lower log level to `ERROR`.
+    ///
+    /// # Motivation
+    ///
+    /// The implementation of `Logged` assumes that most errors have a
+    /// log level that's strictly between `DEBUG` and `ERROR`, in most
+    /// cases, but in some contexts, while they can still only be
+    /// logged and ignored, they signal a configuration issue and call
+    /// for the user's attention.
+    #[inline]
+    fn promoted(self) -> ImportantError<Self>
+    where
+        Self: Sized + core::fmt::Debug,
+    {
+        ImportantError(self)
+    }
+}
+
+impl<T, E: Logged> Logged for core::result::Result<T, E> {
+    #[inline]
+    fn logged(self) -> Self {
+        self.map_err(E::logged)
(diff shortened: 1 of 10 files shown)


logger/src/error.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// importantly when one sees an explicit
+    /// ```rust
+    /// err.logged().ignore("I can't handle this error because of X")
+    /// ```
+    /// the reader of the code doesn't need to be provided
+    /// with a comment as to why the error message was ignored
+    #[inline]
+    fn ignored(self, reason: &str)
+    where
+        Self: Sized,
```
mversic, line 29: you should make `Sized` a supertrait of `Logged`, it's not going to be possible to make it a trait object anyhow

logger/src/error.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// allows one to cheaply explain to the user of the debug build
+    /// **why** a particular error cannot be ignored.  More
+    /// importantly when one sees an explicit
+    /// ```rust
+    /// err.logged().ignore("I can't handle this error because of X")
+    /// ```
+    /// the reader of the code doesn't need to be provided
+    /// with a comment as to why the error message was ignored
+    #[inline]
+    fn ignored(self, reason: &str)
```
Erigara, line 27: What about passing log level inside? Instead of selecting one when deriving.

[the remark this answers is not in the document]
appetrosyan: Suppose you had an `eyre::Report` returned from `torii`. Those are not supposed to be logged server side, at least not with the `<Report as Logged>::logged`. If you *just* call `ignored`, you can still signal that something went wrong that way, but only in the `Trace` level, not the native `Warn`. 

This makes it much cleaner to state that the failure to execute something is perfectly normal and feedback is sent directly to the user, so you don't have to see it server-side.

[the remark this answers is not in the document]
appetrosyan: It's actually one of the main reasons why I chose to keep the two separate. That and having the ability to add `logged` to a chain of operations.

logger/src/error.rs
appetrosyan: I'm not sure we gain much. I wonder if this would allow us to work with references separately from the type, which can be interesting.

logger/src/error.rs
appetrosyan: I suppose I could add 
```rust
fn logged_as(self, level: tracing::Level) -> Self {
	event!(level, "{self}");
}

fn ignored_at(self, level: …) {
	…
```
but it doesn't seem to be useful in many cases. It makes the call much more cumbersome. Moreover, there's a `promote` method.

PR #2918, opened 2022-10-27
[feature] #2899: Add multi-instructions subcommand into `client_cli`

### LINKED ISSUE #2899: [suggestion] Make a subcommand that executes multi-instructions from JSON in `client_cli`
### Feature request

Currently, `client_cli` can submit a single transaction but has no way to submit multi-instructions using a single command. It would be more convenient because all instructions will be parsed from a file or std input in JSON format.

The suggestion is to reuse of already existing syntax that is used to generate genesis [here](https://github.com/hyperledger/iroha/blob/iroha2-dev/configs/peer/genesis.json) because that's an easier way and simpler to keep compatible.

### Motivation

`client_cli` will be able to execute multi-instructions passed as JSON or read from a JSON file.

### Who can help?

_No response_

pesterev:

### Description of the Change

More details about the suggestion you can find [here.](https://github.com/hyperledger/iroha/issues/2899)

I spent some time trying to reuse `GenesisTransaction`s format (`configs/peer/genesis.json`) but it's a bad way cause demand depends on `iroha_core` and a lot of boilerplate code being converted. But I noticed that we can just use `Serialize/Deserialize` which already implemented for `Instruction`s and use those traits to represent instructions as JSON. As a result, I got a way to make a more convenient and simple implementation of this command.

### Issue

Closes #2899 

### Benefits

Submit multiple instructions using one command

### Possible Drawbacks

None

### Usage Examples or Tests

Save these instructions into a file:
```json
[
    {
        "Register": {
            "Identifiable": {
                "NewDomain": {
                    "id": "neverland",
                    "logo": null,
                    "metadata": {}
                }
            }
        }
    },
    {
        "Register": {
            "Identifiable": {
                "NewAccount": {
                    "id": "steve@neverland",
                    "signatories": [
                        "ed0120cc25624d62896d3a0bfd8940f928dc2abf27cc57cefeb442aa96d9081aae58a1"
                    ],
                    "metadata": {}
                }
            }
        }
    },
    {
        "Register": {
            "Identifiable": {
                "NewAssetDefinition": {
                    "id": "sapporo#neverland",
                    "value_type": "Quantity",
                    "mintable": "Infinitely",
                    "metadata": {}
                }
            }
        }
    },
    {
        "Mint": {
            "object": {
                "U32": 1010
            },
            "destination_id": {
                "Id": {
                    "AssetId": "sapporo##steve@neverland"
                }
            }
        }
    }
]
```

And then pass to `client_cli` like this:
```
cat ./test.json | ./iroha_client_cli json
```

### DIFF
--- a/client_cli/src/main.rs
+++ b/client_cli/src/main.rs
@@ -106,6 +106,8 @@ pub enum Subcommand {
     Wasm(wasm::Args),
     /// The subcommand related to block streaming
     Blocks(blocks::Args),
+    /// The subcommand related to multi-instructions as Json
+    Json(json::Args),
 }
 
 /// Runs subcommand
@@ -128,7 +130,7 @@ macro_rules! match_run_all {
 impl RunArgs for Subcommand {
     fn run(self, cfg: &ClientConfiguration) -> Result<()> {
         use Subcommand::*;
-        match_run_all!((self, cfg), { Domain, Account, Asset, Peer, Events, Wasm, Blocks })
+        match_run_all!((self, cfg), { Domain, Account, Asset, Peer, Events, Wasm, Blocks, Json })
     }
 }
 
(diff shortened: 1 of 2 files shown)


appetrosyan (approved):

appetrosyan:
Short and sweet. Excellent.

Erigara (approved):
Cool! Probably need to update docs too.

appetrosyan (approved):

Erigara (approved):

client_cli/src/main.rs
```diff
@@ -106,6 +106,8 @@ pub enum Subcommand {
     Wasm(wasm::Args),
     /// The subcommand related to block streaming
     Blocks(blocks::Args),
+    /// The subcommand related to multi-instructions as Json
+    Json(json::Args),
```
outoftardis, line 110: note to self: add to the docs

PR #2929, opened 2022-11-01
[feature] #2905: Extend arithmetic operations to support `NumericValue` 

### LINKED ISSUE #2905: Ensure `Add`, `Subtract`, `Multiply` and `Power` instructions are strictly applied to `Quantity`, `BigQuantity` and `Fixed`

Erigara:
### Description of the Change

Extend arithmetic operations to support `NumericValue`  instead of only `u32`.

### Issue

Closes #2905.

### Benefits

Arithmetic instructions support more types.

### Possible Drawbacks

None.

### DIFF
--- a/core/src/smartcontracts/isi/expression.rs
+++ b/core/src/smartcontracts/isi/expression.rs
@@ -95,9 +95,28 @@ impl Evaluate for Add {
         wsv: &WorldStateView,
         context: &Context,
     ) -> Result<Self::Value, Self::Error> {
+        use NumericValue::*;
         let left = self.left.evaluate(wsv, context)?;
         let right = self.right.evaluate(wsv, context)?;
-        Ok((left + right).into())
+        match (left, right) {
+            (U32(left), U32(right)) => left
+                .checked_add(right)
+                .ok_or(Error::Math(MathError::Overflow))
+                .map(NumericValue::from),
+            (U128(left), U128(right)) => left
+                .checked_add(right)
+                .ok_or(Error::Math(MathError::Overflow))
+                .map(NumericValue::from),
+            (Fixed(left), Fixed(right)) => left
+                .checked_add(right)
+                .map(NumericValue::from)
+                .map_err(Error::from),
+            (left, right) => Err(MathError::BinaryOpIncompatibleNumericValueTypes(
+                left, right,
+            ))
+            .map_err(Error::from),
+        }
+        .map(Value::from)
     }
 }
 
@@ -110,9 +129,28 @@ impl Evaluate for Subtract {
         wsv: &WorldStateView,
         context: &Context,
     ) -> Result<Self::Value, Self::Error> {
+        use NumericValue::*;
         let left = self.left.evaluate(wsv, context)?;
         let right = self.right.evaluate(wsv, context)?;
-        Ok((left - right).into())
+        match (left, right) {
+            (U32(left), U32(right)) => left
+                .checked_sub(right)
+                .ok_or(Error::Math(MathError::NotEnoughQuantity))
+                .map(NumericValue::from),
+            (U128(left), U128(right)) => left
+                .checked_sub(right)
+                .ok_or(Error::Math(MathError::NotEnoughQuantity))
+                .map(NumericValue::from),
+            (Fixed(left), Fixed(right)) => left
+                .checked_sub(right)
+                .map(NumericValue::from)
+                .map_err(Error::from),
+            (left, right) => Err(MathError::BinaryOpIncompatibleNumericValueTypes(
+                left, right,
+            ))
+            .map_err(Error::from),
+        }
+        .map(Value::from)
     }
 }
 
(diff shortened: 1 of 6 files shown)


core/src/smartcontracts/isi/expression.rs
```diff
@@ -328,9 +405,25 @@ impl Evaluate for RaiseTo {
...
-        let left = self.left.evaluate(wsv, context)?;
-        let right = self.right.evaluate(wsv, context)?;
-        Ok(left.pow(right).into())
+        use NumericValue::*;
+        let value = self.left.evaluate(wsv, context)?;
+        let exp = self.right.evaluate(wsv, context)?;
+        match (value, exp) {
+            (U32(value), U32(exp)) => value
+                .checked_pow(exp)
+                .ok_or(Error::Math(MathError::DivideByZero))
```
QuentinI, line 414: You probably  meant `Overflow`

core/src/smartcontracts/isi/expression.rs
```diff
@@ -328,9 +405,25 @@ impl Evaluate for RaiseTo {
...
+                .map(NumericValue::from),
+            (U128(value), U32(exp)) => value
+                .checked_pow(exp)
+                .ok_or(Error::Math(MathError::DivideByZero))
```
QuentinI, line 418: Same here

core/src/smartcontracts/isi/expression.rs
```diff
@@ -360,22 +470,31 @@ impl Evaluate for Mod {
         wsv: &WorldStateView,
         context: &Context,
     ) -> Result<Self::Value, Self::Error> {
+        use NumericValue::*;
         let left = self.left.evaluate(wsv, context)?;
         let right = self.right.evaluate(wsv, context)?;
-        Ok((left % right).into())
+        match (left, right) {
+            (U32(left), U32(right)) => Ok(left % right).map(NumericValue::from),
```
QuentinI, line 477: `%` also [can panic](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_rem)

data_model/src/expression.rs
```diff
@@ -597,12 +595,11 @@ gen_expr_and_impls! {
         "self.left.parenthesise(Operation::Mod)",
         "self.right.parenthesise(Operation::Mod)"
     )]
-    pub Mod(u32, u32) -> u32
+    pub Mod(NumericValue, NumericValue) -> NumericValue
 }
 
 gen_expr_and_impls! {
     /// Evaluates to the left expression in the power of right expression.
-    /// Works only for [`NumericValue::U32`]
```
QuentinI, line 605: We should probably mention here that it's NYI for `Fixed`

core/src/smartcontracts/isi/mod.rs
```diff
@@ -225,6 +228,14 @@ pub mod error {
             }
         }
     }
+
+    /// Expression error, which occurs during expression evaluation
+    #[derive(Debug, Clone, Copy, Error, PartialEq, Eq)]
+    pub enum ExpressionError {
+        /// Encountered incompatible type of arguments
+        #[error("Expression does not support provided combination of arguments ({0}, {1})")]
+        BinaryOpIncompatibleNumericValueTypes(NumericValue, NumericValue),
```
QuentinI, line 237: Is this semantically different from `MathError::DomainViolation`?

core/src/smartcontracts/isi/expression.rs
Erigara: Sure, copypast error.

core/src/smartcontracts/isi/expression.rs
Erigara: good catch!

core/src/smartcontracts/isi/mod.rs
Erigara: From what i found, this error originates from `fixnum` crate:
```
    /// When someone tries to use operand out of the set of departure of the function.
    /// E.g.: when you try to compute the square root of a negative number.
```
So, it's kinda similar, but not entirely. 
Also i want to provide useful feedback for user and `DomainViolation` does not help with that.

QuentinI: Re: feedback for the user - totally makes sense, I didn't catch it has to conform to what (none) information `fixnum` provides.
I would still say we should move this error into `MathError`. Although it could be argued that it's not strictly speaking a math error, `MathError` is what we return in all other cases of expressions failing, so IMO that would be less confusing than having another enum dedicated solely to this type of error.

QuentinI (dismissed):

mversic (dismissed):

mversic (approved):

core/src/smartcontracts/isi/expression.rs
```diff
@@ -328,9 +405,25 @@ impl Evaluate for RaiseTo {
...
-        Ok(left.pow(right).into())
+        use NumericValue::*;
+        let value = self.left.evaluate(wsv, context)?;
+        let exp = self.right.evaluate(wsv, context)?;
+        match (value, exp) {
+            (U32(value), U32(exp)) => value
+                .checked_pow(exp)
+                .ok_or(Error::Math(MathError::Overflow))
+                .map(NumericValue::from),
+            (U128(value), U32(exp)) => value
```
pesterev, line 416: there is u128 and u32, is it ok?
```suggestion
            (U128(value), U128(exp)) => value
```

pesterev (commented):
looks fine to me

core/src/smartcontracts/isi/expression.rs
Erigara: Yea, this is intentional you can look at [u128::checked_pow](https://doc.rust-lang.org/std/primitive.u128.html#method.checked_pow). `u32` is more then enough to cover all possible exponent  values that do not cause overflow.

pesterev (approved):

Issue #2885, opened 2022-10-19
Add pipeline event to report execution of smartcontract


appetrosyan:
It is extremely useful to know if triggered smartcontracts get executed or not. As such there must be a pipeline event signalling (un)successful execution of a trigger. 

However, such an event should **NOT** have a corresponding trigger. There must be mechanisms preventing adding such a trigger in the future.

PR #2939, opened 2022-11-09
[feature] #2885: Define `TriggerExecuted` event

s8sato:
### Description of the Change
- Add `TriggerExecuted` as a variant of `Event`
  - Not a variant of `PipelineEvent` because trigger executions does not seem to fit `PipelineEvent` structure

### Issue
- Partial solution to #2885

### Benefits
- Data structure for notification of triggers execution

### Possible Drawbacks
- This event is assumed not to chain trigger executions at the moment. Developers can break this assumption and reach `unreachable!()`


data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+)]
+pub enum Status {
+    /// The trigger execution was succeeded.
+    Succeeded,
+    /// The trigger execution was failed due to [`FailReason`].
+    Failed(FailReason),
+}
+
+/// Details of why the trigger execution failed.
+pub type FailReason = Metadata;
```
s8sato, line 122: Defined as the most general-purpose data structure

mversic (commented):
correct me if I'm wrong, but I don't see `TriggerExecuted` event emitted anywhere

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
+//! Event that notifies some trigger was somehow executed.
```
outoftardis, line 1:
```suggestion
//! Event that notifies that a trigger was executed.
```

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+    Encode,
+    IntoSchema,
+    Hash,
+    Serialize,
+    Deserialize,
+)]
+pub enum StatusKind {
+    /// The trigger execution was succeeded.
+    Succeeded,
+    /// The trigger execution was failed.
```
outoftardis, line 143:
```suggestion
    /// The trigger execution succeeded.
    Succeeded,
    /// The trigger execution failed.
```

```suggestion
    /// The trigger execution was successful.
    Succeeded,
    /// The trigger execution failed.
```

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+    Serialize,
+    Deserialize,
+    FromVariant,
+    IntoSchema,
+)]
+pub enum Status {
+    /// The trigger execution was succeeded.
+    Succeeded,
+    /// The trigger execution was failed due to [`FailReason`].
+    Failed(FailReason),
```
outoftardis, line 118:
```suggestion
    /// The trigger execution succeeded.
    Succeeded,
    /// The trigger execution failed due to [`FailReason`].
    Failed(FailReason),
```

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+        [
+            Self::field_matches(self.id.as_ref(), &event.id),
+            Self::field_matches(self.status_kind.as_ref(), &event.status.kind()),
+        ]
+        .into_iter()
+        .all(core::convert::identity)
+    }
+}
+
+/// Notification that some trigger was somehow executed.
```
outoftardis, line 84:
```suggestion
/// Notification that a trigger was executed.
```

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+pub struct Event {
+    /// Which trigger was executed.
```
outoftardis, line 87:
```suggestion
    /// The id of the executed trigger.
```

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub id: TriggerId,
+    /// How the trigger execution is going on, or resulted in.
```
outoftardis, line 89:
```suggestion
    /// The status of the trigger execution.
```

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+impl Event {
+    /// Construct [`Event`].
+    pub const fn new(id: TriggerId, status: Status) -> Self {
+        Event { id, status }
+    }
+}
+
+/// How the trigger execution is going on, or resulted in.
```
outoftardis, line 100:
```suggestion
/// The status of the trigger execution.
```

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -61,6 +61,8 @@ pub mod isi {
                             .try_into()
                             .map_err(|e: &str| Self::Error::Conversion(e.to_owned()))?,
                     ),
+                    // `TriggerExecuted` event does not chain trigger executions at the moment
+                    FilterBox::TriggerExecuted(_) => unreachable!(),
```
QuentinI, line 65: This is definitely reachable:
```rust
let trigger = Trigger::new(
    "test".parse()?,
    Action::new(
        vec![],
        Repeats::Indefinitely,
        "alice@wonderland".parse()?,
        TriggerExecutedEventFilter::new().into(),
    ),
);
client.submit_blocking(RegisterBox::new(trigger))?;
```

Maybe we should return an error instead indicating that chaining triggers isn't supported?

data_model/src/trigger/mod.rs
```diff
@@ -428,6 +428,8 @@ mod tests {
                         .map(|_| ())
                         .unwrap()
                 }
+                // `TriggerExecuted` event does not chain trigger executions at the moment
+                FilterBox::TriggerExecuted(_) => unreachable!(),
```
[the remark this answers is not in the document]
s8sato: I thought these `unreachable`s might be temporary i.e. `TriggerExecuted` event might chain trigger executions.
If not, we could separate `Event` enum to avoid `unreachable`

> I don't see `TriggerExecuted` event emitted anywhere

You are right. I only defined the data structure in this PR

core/src/smartcontracts/isi/triggers.rs
s8sato: Yes. That's exactly my concern mentioned in Possible Drawbacks

QuentinI: Ah, sorry, I assumed "iroha developers" there, not "trigger developers".

I would say having a way for malicious client to induce panic in iroha node is a security issue and we shouldn't allow it

s8sato: Makes sense

ilchu (commented):
I wonder if it's possible to make these events part of `pipeline::EventFilter` -- as it is a pipeline event technically. From what I see, it would require extending `PipelineEntityKind` to include triggers, and then we could use `PipelineStatusKind` to represent `FailReason` instead of `Metadata`. This would require some (quite a few?) checks to be inserted to prevent creating a trigger from this variant of pipeline events, though, and that could actually be difficult.
 
Is there some other limitation to this approach I didn't notice?

data_model/src/events/trigger_executed.rs
```diff
@@ -0,0 +1,255 @@
...
+    Hash,
+    Serialize,
+    Deserialize,
+    FfiType,
+)]
+pub struct EventFilter {
+    /// If `Some`, filters by the [`TriggerId`]. Otherwise accepts any [`TriggerId`].
+    pub id: Option<TriggerId>,
+    /// If `Some`, filters by the [`StatusKind`]. Otherwise accepts any [`StatusKind`].
+    pub status_kind: Option<StatusKind>,
```
mversic, line 36: We could go without `StatusKind` if event `TriggerExecuted` is only be emitted if trigger was successfully executed. We don't seem to emit failed events(only successful) for any other type of fallible event. Consider following this convention

s8sato: The `Failed` variant would correspond to `Rejected` of pipeline events. Otherwise what will be a notification of trigger execution failures?

data_model/src/events/trigger_executed.rs
appetrosyan: For now, yes. We need to create an Epic to actually produce appropriate error messages.

PR #2947, opened 2022-11-15
[refactor]: `p2p` with guard rails

appetrosyan:
### Description of the Change

Add extra guard rails to fix problems of multigenerational Iroha networks.
### Issue

None (TODO)

### Benefits

Better UX

### Possible Drawbacks

More coupled code


appetrosyan:
Conflict resolution failed

PR #2949, opened 2022-11-15
[fix] #2506: Fix the block validation

### LINKED ISSUE #2506: [PERFORMANCE BUG] When sending ~ 5000 assets in a row then Iroha 2 down without information in logs.
### GIT commit hash

e3cd89bd

### Minimum working example

```rust
fn main() {
     // Steps to reproduce …
     // Steps to reproduce …
}
```

### Expected behaviour

I have a script which creates ~650 assets and then query findAllAssets to print total size. After like ~30k assets script starts to print same size after every new iteration (because findAllAssets returns old data). Btw all requests are sync. But after some  iterations data will be updated and Iroha returns all new assets. But after ~30k assets it doesn't update after every new bunch of assets.

### Actual behaviour

👆👆👆

### Operating system

macos

### LibC type and version

_No response_

### Current environment

Docker Hub

### Logs in JSON format

<details>
  
  ```json
  ```
</details>

### Who can help?

_No response_

pesterev:

### Description of the Change

This PR fixes the incorrect block validation - applies WSV cloning on every block, not on every transaction.

### Issue

Closes #2506

### Benefits

It fixes crashes on the Sumeragi main loop due to a lack of correct validation

### Possible Drawbacks

None

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -268,8 +268,10 @@ impl Iroha {
 
         // Validate every transaction in genesis block
         if let Some(ref genesis) = genesis {
+            let wsv_clone = wsv.clone();
+
             transaction_validator
-                .validate_every(genesis.iter().cloned(), &wsv)
+                .validate_every(genesis.iter().cloned(), &wsv_clone)
                 .wrap_err("Transaction validation failed in genesis block")?;
         }
 
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -183,14 +183,10 @@ impl ChainedBlock {
     ) -> ValidBlock {
         let mut txs = Vec::new();
         let mut rejected = Vec::new();
-        let wsv_clone = WorldStateView::clone(wsv);
-        let wsv = if self.header.is_genesis() {
-            wsv
-        } else {
-            &wsv_clone
-        };
+
+        let wsv = wsv.clone();
         for tx in self.transactions {
-            match transaction_validator.validate(tx.into_v1(), self.header.is_genesis(), wsv) {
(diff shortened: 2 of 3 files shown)


Erigara (approved):

mversic (approved):

cli/src/lib.rs
```diff
@@ -268,8 +268,10 @@ impl Iroha {
 
         // Validate every transaction in genesis block
         if let Some(ref genesis) = genesis {
+            let wsv_clone = wsv.clone();
+
             transaction_validator
-                .validate_every(genesis.iter().cloned(), &wsv)
+                .validate_every(genesis.iter().cloned(), &wsv_clone)
```
mversic, line 274: this one was causing you problems

core/src/block.rs
```diff
@@ -616,7 +618,7 @@ impl CandidateBlock {
             .map(|accepted_tx| {
                 accepted_tx.and_then(|tx| {
                     transaction_validator
```
mversic, line 620: this pattern and the fact that you had a problem finding a bug makes me think we should have `block_validator` instead of `transaction_validator` which would always clone wsv internally. I would appreciate if you can make this PR include this change

pesterev: Let's make it in a separate pull request. It doesn't seem easy.

pesterev:
I'm going to add a new integration test to the Iroha repo that will reproduce that problem. I feel it can break anytime, and the test will help detect it.

mversic (approved):

Erigara (approved):

Issue #2955, opened 2022-11-17
[suggestion]: Lightweight consensus messages


Erigara:
### Feature request

I'd like Iroha2 to omit sending full block on `BlockSigned`, `BlockCommited` messages. 
Seems like peers could preserve block received through `BlockCreated` message.

There is corner case however: 
If consensus does not required (less then 4 peers network) or it is genesis round `Leader` peer sends block directly through `BlockCommited` message. 

### Motivation

Lighter messages would improve consensus speed due to lower network costs especially for heavy blocks. 

### Who can help?

@appetrosyan @Erigara @SamHSmith

PR #2956, opened 2022-11-18
[refactor] #1868: Prevent validated transactions from being sent between peers

### LINKED ISSUE #1868: Prevent validated transactions and blocks from being sent between peers
Transactions and blocks have lifecycles like:
`Transaction` -> `AcceptedTransaction` -> `ValidTransaction`
`PendingBlock` -> `ChainedBlock` -> `ValidBlock`

only `Transaction` and `PendingBlock` should be allowed to be sent between peers because invariants that are expected to hold true in later parts of the lifecycle cannot be guaranteed (there can be malicious nodes pretending to have performed validation). This means that transaction/block should traverse it's lifecycle only in the context of one peer

Currently, we do this by developers diligence but it would be better if we were to disable these programs from compiling. If there is a proposal or an idea on how to do this it should be discussed before implementation

Erigara:

### Description of the Change

Add `CandidateBlock` which is send between peers, `revalidation` required to obtain `ValidBlock` for `CandidateBlock`.

### Issue

Closes #1868.

### Benefits

Don't rely on programmer to do revalidation and make it unavoidable through strong typing.  

### Possible Drawbacks

Currently, there may be a performance penalty due to the fact that revalidation now occurs in places where it is not really necessary (would be resolved with #2955).


core/src/sumeragi/message.rs
```diff
@@ -110,11 +130,21 @@ impl From<VersionedValidBlock> for BlockSigned {
 #[non_exhaustive]
 pub struct BlockCommitted {
     /// The corresponding block.
-    pub block: VersionedValidBlock,
+    pub block: VersionedCandidateBlock,
```
pesterev, line 133: It seems we can make this field private and add an impl like this:
```rust
impl From<BlockCommitted> for VersionedCandidateBlock {
    fn from(block: BlockCommitted) -> Self {
        Self { block: block.block }
    }
}
```
or add the `inner()` method

```rust
impl BlockCommitted {
// ...
    fn inner(self) -> VersionedCandidateBlock {
        self.block
    }
}
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+
+    /// Performs the conversion from `VersionedCandidateBlock` to V1
+    #[inline]
+    pub fn into_v1(self) -> CandidateBlock {
+        match self {
+            Self::V1(v1) => v1,
+        }
+    }
+
+    /// Returns header of valid block
```
outoftardis, line 523:
```suggestion
    /// Returns the header of a valid block
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+    pub const fn header(&self) -> &BlockHeader {
+        &self.as_v1().header
+    }
+
+    /// Calculate hash of the current block.
```
outoftardis, line 528:
```suggestion
    /// Calculate the hash of the current block.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+    pub fn hash(&self) -> HashOf<Self> {
+        self.as_v1().hash().transmute()
+    }
+
+    /// Signatures that are verified with the `hash` of this block as `payload`.
```
outoftardis, line 533:
```suggestion
    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+        wsv: &WorldStateView,
+        latest_block: &HashOf<VersionedCommittedBlock>,
+        block_height: u64,
+    ) -> Result<ValidBlock, eyre::Report> {
+        self.into_v1()
+            .revalidate(transaction_validator, wsv, latest_block, block_height)
+    }
+}
+
+/// Block send through network requires revalidation to be transformed back into [`ValidBlock`]
```
outoftardis, line 556:
```suggestion
/// Revalidate the block that was sent through the network by transforming it back into [`ValidBlock`]
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+    /// Array of transactions.
+    pub transactions: Vec<VersionedSignedTransaction>,
+    /// Signatures of peers which approved this block.
+    pub signatures: BTreeSet<SignatureOf<Self>>,
+    /// Event recommendations.
+    pub event_recommendations: Vec<Event>,
+}
+
+impl CandidateBlock {
+    /// Calculate hash of the current block.
```
outoftardis, line 573:
```suggestion
    /// Calculate the hash of the current block.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+    pub fn hash(&self) -> HashOf<Self> {
+        HashOf::new(&self.header).transmute()
+    }
+
+    /// Signatures that are verified with the `hash` of this block as `payload`.
```
outoftardis, line 578:
```suggestion
    /// Return signatures (as `payload`) that are verified with the `hash` of this block.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+            .iter()
+            .filter(move |sign| sign.verify_hash(&hash).is_ok())
+    }
+
+    /// Checks if there are no transactions in this block.
+    pub fn is_empty(&self) -> bool {
+        self.transactions.is_empty() && self.rejected_transactions.is_empty()
+    }
+
+    /// Checks if block has transactions that are already in blockchain.
```
outoftardis, line 591:
```suggestion
    /// Checks if a block has transactions that are already in the blockchain.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
     }
 }
 
+declare_versioned_with_scale!(VersionedCandidateBlock 1..2, Debug, Clone, iroha_macro::FromVariant, IntoSchema);
+
+impl VersionedCandidateBlock {
+    /// Converts from `&VersionedCandidateBlock` to V1 reference
```
outoftardis, line 499:
```suggestion
    /// Convert from `&VersionedCandidateBlock` to V1 reference
```

the verbs should be in imperative (without `s`)

applies to other descriptions as well

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+
+/// Block send through network requires revalidation to be transformed back into [`ValidBlock`]
+#[version_with_scale(n = 1, versioned = "VersionedCandidateBlock")]
+#[derive(Debug, Clone, Decode, Encode, IntoSchema)]
+pub struct CandidateBlock {
+    /// Header
+    pub header: BlockHeader,
+    /// Array of rejected transactions.
+    pub rejected_transactions: Vec<VersionedSignedTransaction>,
+    /// Array of transactions.
```
outoftardis, line 564:
```suggestion
    /// Array of all transactions in this block.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+        self.transactions
+            .iter()
+            .any(|transaction| transaction.is_in_blockchain(wsv))
+            || self
+                .rejected_transactions
+                .iter()
+                .any(|transaction| transaction.is_in_blockchain(wsv))
+    }
+
+    /// Revalidate block against current state of the world.
```
outoftardis, line 602:
```suggestion
    /// Revalidate a block against the current state of the world.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+    }
+
+    /// Signatures that are verified with the `hash` of this block as `payload`.
+    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
+        self.as_v1()
+            .verified_signatures()
+            .map(SignatureOf::transmute_ref)
+    }
+
+    /// Validate block transactions against current state of the world.
```
outoftardis, line 540:
```suggestion
    /// Validate block transactions against the current state of the world.
```

core/src/block.rs
```diff
@@ -659,6 +493,269 @@ impl From<&ValidBlock> for Vec<Event> {
...
+                .any(|transaction| transaction.is_in_blockchain(wsv))
+    }
+
+    /// Revalidate block against current state of the world.
+    ///
+    /// # Errors
+    /// - Block is empty
+    /// - Block has committed transactions
+    /// - There is mismatch between candidate block height and actual blockchain height
+    /// - There is mismatch between candidate block previous block hash and actual latest block hash
```
outoftardis, line 608:
```suggestion
    /// - There is a mismatch between candidate block height and actual blockchain height
    /// - There is a mismatch between candidate block, previous block hash, and actual latest block hash
```

core/src/sumeragi/message.rs
Erigara: I have plans to change `BlockCommitted` message to have only hash and signatures, so not sure if we should add this change. 
What benifits suggested approach have compared to plain field access.

core/src/sumeragi/main_loop.rs
```diff
@@ -980,6 +977,26 @@ pub fn run<F>(
                         if valid_signatures.len() >= state.current_topology.min_votes_for_commit()
                             && state.latest_block_hash == block.header().previous_block_hash
                         {
+                            // TODO: don't need to revalidate here instead send only hash and signatures in `BlockCommitted` message (#2955)
+                            let block = {
+                                let span = span!(
+                                    Level::TRACE,
+                                    "Sumeragi Validating Peer revalidate block."
```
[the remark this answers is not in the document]
Erigara: @outoftardis, so this message indicates that peer with particular role (`ValidatingPeer`) now in process of revalidation. 
Any suggestions how we can rephrase this?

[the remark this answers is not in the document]
outoftardis:
```suggestion
                                    "Sumeragi Validating Peer is revalidating the block."
```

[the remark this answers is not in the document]
outoftardis:
```suggestion
                                    "Revalidating the block by Sumeragi Validating Peer"
```
might also work

core/src/sumeragi/message.rs
pesterev: > What benefits suggested approach have compared to plain field access?

In my opinion, this approach encapsulates the inner field

Erigara: I will try to explain my point of view on this.
Imo this structure is not some kind of class where we need to held invariants it's just plain data which we use to transfer data through network, so there is no point in hiding any of its fields.

pesterev: I proceeded from the [use cases](https://github.com/hyperledger/iroha/blob/iroha2-dev/core/src/sumeragi/main_loop.rs#L718) of the field in Sumeragi. In all cases, the block is moved to a separate `VersionedValidBlock` variable, and the `BlockCommitted` is unwrapped and removed.

But yes, I understand your point. Let's leave it.

pesterev (approved):

mversic (approved):

Issue #1916, opened 2022-02-17
Make `SignaturesOf` use a BTreeSet rather than BTreeMap


appetrosyan:
```rust
// TODO: Serialize/Encode as BTreeSet?
pub struct SignaturesOf<T> {
    // This structure is backed by map because only one signature is allowed per public key.
    // In the case of Iroha this means that each peer can sign the payload at most once.
    //
    // TODO: If uniqueness of public key in this collection would be upheld by other means or
    // if it were true that `Signature: Borrow<PublicKey>` then set could be used instead of map
    signatures: btree_map::BTreeMap<PublicKey, SignatureOf<T>>,
}
```

SamHSmith:
This has already been done by @Erigara .

PR #2966, opened 2022-11-24
[refactor] #2955: Remove `Block` from `BlockSigned` message

### LINKED ISSUE #2955: [suggestion]: Lightweight consensus messages

Erigara:
### Description of the Change

- Store `BTreeSet` instead of `BTreeMap` inside `SignaturesOf`;
- Add additional state to block `ValidSignedBlock` to represent valid block that received at least one signature;
-  Store hash and signatures inside of `BlockSigned` message instead of full block.

### Issue

Partially closes #2955. 

### Benefits

- Less space occupied by`SignaturesOf`
- Smaller `BlockSigned` messages
- Additional strong-typing check that block is signed before sending to other peers

### Possible Drawbacks

None.

`BlockCommitted` can be optimized too, but it not as straightforward as `BlockSigned` because involve block synchronization and genesis round. So the changes will come later.


core/src/block.rs
```diff
@@ -363,30 +361,50 @@ pub struct ValidBlock {
...
     pub transactions: Vec<VersionedValidTransaction>,
-    /// Signatures of peers which approved this block.
-    pub signatures: BTreeSet<SignatureOf<Self>>,
     /// Event recommendations.
     pub event_recommendations: Vec<Event>,
 }
 
 impl ValidBlock {
+    /// Calculate hash of the current block.
+    pub fn hash(&self) -> HashOf<Self> {
```
appetrosyan, line 370: `#[inline]`. The hashing function isn't inlineable, but the transmute is, so you remove the overhead of one extra function call.

core/src/block.rs
```diff
@@ -363,30 +361,50 @@ pub struct ValidBlock {
...
-        self.rejected_transactions
-            .iter()
-            .map(|tx| tx.check_limits(tx_limits))
-            .collect::<Result<Vec<_>, _>>()
-            .map(drop)?;
-        Ok(())
+    /// Fails if generating signature fails
+    pub fn sign(self, key_pair: KeyPair) -> Result<ValidSignedBlock> {
+        let signature = SignatureOf::from_hash(key_pair, &self.hash().transmute())
+            .wrap_err("Failed to sign block")?;
```
appetrosyan, line 380: I'd like more context in the error message. Maybe which block with which hash?

core/src/block.rs
```diff
@@ -397,11 +415,6 @@ impl ValidBlock {
...
         } = self;
 
-        #[allow(clippy::expect_used)]
-        let signatures: SignaturesOf<ValidBlock> = signatures
-            .try_into()
-            .expect("Expected at least one signature");
-
         CommittedBlock {
             event_recommendations,
             header,
```
appetrosyan, line 420: Can we replace this with `..self` instead?

core/src/block.rs
```diff
@@ -429,20 +442,17 @@ impl ValidBlock {
 
     /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
     pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
```
appetrosyan, line 444: This function is two function calls. It could be `#[inline]`.

core/src/block.rs
```diff
@@ -363,30 +361,50 @@ pub struct ValidBlock {
...
+    pub transactions: Vec<VersionedValidTransaction>,
+    /// Signatures of peers which approved this block.
+    pub signatures: SignaturesOf<Self>,
+    /// Event recommendations.
+    pub event_recommendations: Vec<Event>,
+}
+
+impl ValidSignedBlock {
     /// Commit block to the store.
     pub fn commit(self) -> CommittedBlock {
```
appetrosyan, line 409: If we don't use this function outside this crate, I suggest making it `pub(crate)`. Alternatively, if we want to keep it in the public API, consider making it `#[inline]`.

core/src/block.rs
```diff
@@ -867,7 +873,7 @@ impl CommittedBlock {
     }
 }
 
-impl From<CommittedBlock> for ValidBlock {
+impl From<CommittedBlock> for ValidSignedBlock {
     fn from(
```
appetrosyan, line 877: This function is public, please add `#[inline]`

crypto/src/signature.rs
```diff
@@ -263,10 +263,71 @@ impl<T: Encode> SignatureOf<T> {
     }
 }
 
+/// Wrapper around [`SignatureOf`] used to reimplement [`Eq`], [`Ord`]
+/// to compare signatures only by there [`PublicKey`].
```
appetrosyan, line 267:
```suggestion
/// Wrapper around [`SignatureOf`] used to reimplement [`Eq`], [`Ord`]
/// to compare signatures only by their [`PublicKey`].
```

crypto/src/signature.rs
```diff
@@ -263,10 +263,71 @@ impl<T: Encode> SignatureOf<T> {
...
+#[allow(clippy::derive_hash_xor_eq)]
```
appetrosyan, line 268: Clippy has a point. To me it's not obvious why comparing by public key in `PartialEq` is compatible with the default `hash` implementation which also includes the `payload`. This will lead to inconsistencies in behaviour in any `HashMap`, so I'd rather lay it out explicitly via a manual implementation or a unit test.

crypto/src/signature.rs
```diff
@@ -263,10 +263,71 @@ impl<T: Encode> SignatureOf<T> {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl<T> Clone for SignatureWrapperOf<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
```
appetrosyan, line 314: Derive?

crypto/src/signature.rs
```diff
@@ -263,10 +263,71 @@ impl<T: Encode> SignatureOf<T> {
...
+
+// No need to show this type in the schema
```
appetrosyan, line 316:
```suggestion
// Wrapper is transparent, so forward to the inner type and exclude from schema. 
```

crypto/src/signature.rs
appetrosyan: Specifically, if we want to `impl Hash`, we need to explain why the usage as is, will not cause any problems.

appetrosyan: The unit test should verify that the difference in behaviour between `PartialEq` and `Hash` is there, and explain why it's OK.

crypto/src/signature.rs
```diff
@@ -263,10 +263,71 @@ impl<T: Encode> SignatureOf<T> {
...
+        self.0.public_key.partial_cmp(&other.0.public_key)
+    }
+}
+impl<T> Ord for SignatureWrapperOf<T> {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.0.public_key.cmp(&other.0.public_key)
+    }
+}
+
+impl<T> fmt::Debug for SignatureWrapperOf<T> {
```
appetrosyan, line 304: I'd rather this was derived.

crypto/src/signature.rs
Erigara: Derive would add `T: Clone` bound and we don't need this.

crypto/src/signature.rs
Erigara: Derive would add `T: Debug` bound and we don't need this.

crypto/src/signature.rs
appetrosyan: Because it's a phantom. If `#[derive(` weren't a macro, it'd be able to see that.

crypto/src/signature.rs
appetrosyan: In principle, we don't, but in this case, everything that could be signed, might fail to get signed, and we might want to print it out. 

Up to you. I'd rather we had this bound, but this could entail a sweeping modification of the code base. If you don't think it's worth the effort, resolve.

core/src/block.rs
```diff
@@ -363,30 +361,50 @@ pub struct ValidBlock {
...
+}
 
+/// After receiving first signature `ValidBlock` can transform into `ValidSignedBlock`.
+#[derive(Debug, Clone)]
+pub struct ValidSignedBlock {
+    /// Header
+    pub header: BlockHeader,
+    /// Array of rejected transactions.
+    pub rejected_transactions: Vec<VersionedRejectedTransaction>,
+    /// Array of transactions.
```
outoftardis, line 399:
```suggestion
    /// Array of all transactions in this block.
```

core/src/block.rs
```diff
@@ -363,30 +361,50 @@ pub struct ValidBlock {
...
+            header: self.header,
+            rejected_transactions: self.rejected_transactions,
+            transactions: self.transactions,
+            event_recommendations: self.event_recommendations,
+            signatures,
+        })
     }
+}
 
+/// After receiving first signature `ValidBlock` can transform into `ValidSignedBlock`.
```
outoftardis, line 392:
```suggestion
/// After receiving first signature, `ValidBlock` can transform into `ValidSignedBlock`.
```

core/src/block.rs
```diff
@@ -363,30 +361,50 @@ pub struct ValidBlock {
...
+#[derive(Debug, Clone)]
+pub struct ValidSignedBlock {
+    /// Header
```
outoftardis, line 395:
```suggestion
    /// Block header
```

mversic:
Closes #1916

crypto/src/signature.rs
```diff
@@ -366,34 +426,39 @@ impl<A> Extend<SignatureOf<A>> for SignaturesOf<A> {
...
     fn from_iter<T: IntoIterator<Item = SignatureOf<A>>>(iter: T) -> Self {
-        let signatures: btree_set::BTreeSet<_> = iter.into_iter().collect();
-        signatures.try_into()
+        let mut iter = iter.into_iter();
+        iter.next()
+            .ok_or_else(|| {
+                Error::Other(format!(
+                    "{} must contain at least one signature",
+                    core::any::type_name::<Self>()
+                ))
```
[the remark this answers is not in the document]
Erigara: I will add additional variant to `Error`.

core/src/block.rs
Erigara: Sadly same types are required for this.

core/src/block.rs
Erigara: Made `inline`.

crypto/src/signature.rs
Erigara: I would leave this as is, since we kinda mimic behavior of `SignatureOf`.

crypto/src/signature.rs
Erigara: Added manual implementation and covered it with test.

crypto/src/signature.rs
```diff
@@ -263,10 +263,71 @@ impl<T: Encode> SignatureOf<T> {
...
+    }
+
+    fn type_name() -> String {
+        SignatureOf::<T>::type_name()
+    }
+}
+
 /// Container for multiple signatures, each corresponding to a different public key.
 ///
 /// If signature is added which conflicts with a signature already present in the
```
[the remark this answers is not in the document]
Erigara: I wouldn't say that they are `incompatible`. 
User can try to add signature which public key already present in the set, in this case old signature with equal public key will be replaced by new one.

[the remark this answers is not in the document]
outoftardis: then it should probably say exactly that: that if the public key is already in the set, the associated signature will be replaced with the new one

appetrosyan (dismissed):

SamHSmith (approved):
LGTM

SamHSmith:
Me and Roman discussed on telegram how Signatures might better be handled. But we agreed that it should be a seperate ticket.

Erigara:
Created #2976.

appetrosyan (approved):

Issue #2967, opened 2022-11-24
[feature-request]: Add execution start time to wasm trigger entrypoint


mversic:
There is no way of retrieving current time in `no_std` environment which every wasm smartcontract is. Dynamic linking would also facilitate this but it's not yet finished.

In the scope of this issue add `unix_timestamp` to entrypoint arguments so that trigger code has at it's disposal the time of execution

PR #2968, opened 2022-11-24
[feature] #2967: add start time to trigger entrypoint

mversic:


core/src/smartcontracts/wasm.rs
```diff
@@ -539,11 +562,12 @@ impl<'wrld> Runtime<'wrld> {
...
     ) -> Result<(), Error> {
-        let state = State::new(wsv, account_id.clone(), self.config).with_validator(
-            max_instruction_count,
-            instruction_judge,
-            query_judge,
-        );
+        let curr_time_millis = current_time()
+            .as_millis()
+            .try_into()
+            .expect("Timestamp millis should fit into u64");
```
appetrosyan, line 568: This might be controversial, but why can't this just be `u128`? It solves a problem that we're unlikely to encounter, but still an unnecessary panic.

appetrosyan: We already have `Numeric::BigQuantity` which is `u128` and crossing the FFI boundary, so why not make everything work without conversions?

mversic: This cannot fail(for ~500million years) unless system clock is tampered with. I made this choice because `u64` is natively supported by `WebAssembly`, for `u128` it gets more hairy. I don't think FFI library can be used here because this is communication between wasm module and wasm runtime

a counter argument - do you think trigger can leverage the subsecond precision?

appetrosyan: > a counter argument - do you think trigger can leverage the subsecond precision?

Erigara: I think we should be extra careful with passing current time inside trigger. 
Suppose that peer try to reconnect to the network network it will commits blocks at different time that they are committed originally and it could lead to divergence. 
Maybe we can pass inside block related timestamp?

Arjentix:
I'm not sure that I should review this, because it looks very outdated. As @appetrosyan said in the chat, we still want this PR.

@mversic , could you update it?

core/src/smartcontracts/wasm.rs
```diff
@@ -177,9 +181,16 @@ struct State<'wrld> {
 
 impl<'wrld> State<'wrld> {
     fn new(wsv: &'wrld WorldStateView, account_id: AccountId, config: Configuration) -> Self {
+        #[allow(clippy::expect_used)]
+        let unix_timestamp_millis = current_time()
+            .as_millis()
+            .try_into()
+            .expect("Timestamp millis should fit into u64");
```
appetrosyan, line 188:
```suggestion
            .unwrap_or(0);
```
The odds of this happening naturally are 0. Plus we get a signal that the timestamp for one reason or another didn't fit.

appetrosyan (approved):

mversic:
due to the difficulties of reproducibility, we'll consider introducing Oracle in a separate issue

Issue #2875, opened 2022-10-18
Rewrite p2p to not use actor system.


SamHSmith:

SamHSmith:
As mentioned here https://github.com/hyperledger/iroha/pull/2969#issuecomment-1342252339 this issue is now closed.

PR #2969, opened 2022-11-25
[fix] #2875: Rewrite of p2p system and remove of actor system.

SamHSmith:
Co-authored-by: Aleksandr Petrosyan <a-p-petrosyan@yandex.ru>


core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
     /// Create [`Self`] from [`Configuration`]
     pub fn from_configuration(
         config: &Configuration,
         sumeragi: Arc<Sumeragi>,
+        p2p: Arc<P2PSystem>,
         peer_id: PeerId,
-        broker: Broker,
-    ) -> Self {
-        Self {
+    ) -> Arc<Self> {
```
appetrosyan, line 45: Maybe it's a good idea to construct the `Arc` explicitly, and remove it from the signature.

core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
+}
+
+/// Initiate the read thread with the main loop `block_sync_read_loop`
+///
+/// # Panics
+///
+/// If thread wasn't spawned
+#[allow(clippy::expect_used)]
+pub fn start_read_loop(block_sync: Arc<BlockSynchronizer>) -> ThreadHandler {
+    // Oneshot channel to allow forcefully stopping the thread.
```
appetrosyan, line 63:
```suggestion
```

It's obvious from the code.

core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
+}
+
+fn block_sync_read_loop(
+    block_sync: &BlockSynchronizer,
+    mut shutdown_receiver: tokio::sync::oneshot::Receiver<()>,
+) {
+    let mut last_requested_blocks = Instant::now();
+    let mut request_blocks_peer_index = 0_usize;
+    loop {
+        // We have no obligations to network delivery so we simply exit on shutdown signal.
```
appetrosyan, line 87:
```suggestion
```

core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
+        if shutdown_receiver.try_recv().is_ok() {
+            iroha_logger::debug!("P2P thread is being shut down");
```
appetrosyan, line 89:
```suggestion
            iroha_logger::debug!("P2P thread is shutting down. All obligations fulfilled. ");
```

core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
+            .poll_network_for_packet(NetworkMessageVariant::BlockSync)
+            .map_or_else(
+                || {
+                    std::thread::sleep(Duration::from_millis(10));
+                },
+                |message| match message {
+                    NetworkMessage::BlockSync(message) => {
+                        message.into_v1().handle_message(block_sync);
+                    }
+                    #[allow(clippy::restriction)]
```
appetrosyan, line 103:
```suggestion
                    #[allow(clippy::panic)]
```

core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
+            return;
+        }
+        block_sync
+            .p2p
+            .poll_network_for_packet(NetworkMessageVariant::BlockSync)
+            .map_or_else(
+                || {
+                    std::thread::sleep(Duration::from_millis(10));
+                },
+                |message| match message {
```
appetrosyan, line 99: We could rewrite this idiomatically with an `if let NetworkMessage::BlockSync(message)`

core/src/block_sync.rs
```diff
@@ -168,55 +225,40 @@ pub mod message {
     }
 
     impl Message {
-        /// Handles the incoming message.
-        #[iroha_futures::telemetry_future]
-        pub async fn handle_message(&self, block_sync: &mut BlockSynchronizer) {
+        #[allow(clippy::expect_used)]
```
appetrosyan, line 228:
```suggestion
```

Mutex no longer gets poisoned.

core/test_network/src/lib.rs
```diff
@@ -551,7 +515,7 @@ impl Peer {
             telemetry_address,
             shutdown,
             iroha: None,
-            broker: Broker::new(),
+            // broker: Broker::new(),
```
appetrosyan, line 518:
```suggestion
```

core/test_network/src/lib.rs
```diff
@@ -409,8 +375,8 @@ pub struct Peer {
     pub telemetry_address: String,
     /// The key-pair for the peer
     pub key_pair: KeyPair,
-    /// Broker
-    pub broker: Broker,
+    // /// Broker
+    // pub broker: Broker,
```
appetrosyan, line 379:
```suggestion
```

core/test_network/src/lib.rs
```diff
@@ -371,10 +337,11 @@ impl Network {
 /// # Panics
 /// When unsuccessful after `MAX_RETRIES`.
 pub fn wait_for_genesis_committed(clients: &[Client], offline_peers: u32) {
-    const POLL_PERIOD: Duration = Duration::from_millis(1000);
-    const MAX_RETRIES: u32 = 10;
+    const POLL_PERIOD: Duration = Duration::from_millis(5000);
+    const MAX_RETRIES: u32 = 6;
```
appetrosyan, line 341: These numbers seem arbitrary. We need to have a clear idea on how to set them, and how to measure the timeouts for the types of deployment. Perhaps they should even be extracted into configuration files.

core/src/lib.rs
```diff
@@ -50,14 +48,20 @@ pub type PermissionTokenDefinitionsMap =
...
+#[derive(Clone, Debug, Encode, Decode)]
 pub enum NetworkMessage {
     /// Blockchain message
     SumeragiPacket(Box<SumeragiPacket>),
     /// Block sync message
     BlockSync(Box<BlockSyncMessage>),
     /// Health check message
     Health,
+    /// Connection check message variant. It contains the size of the
+    /// message.
```
appetrosyan, line 60:
```suggestion
    /// message. **DO NOT TURN INTO UNIT VARIANT**. The size is needed for `parity_scale_codec`
```

core/src/lib.rs
```diff
@@ -50,14 +48,20 @@ pub type PermissionTokenDefinitionsMap =
...
+    ConnectionCheck(u64),
+    /// Connection check acknowledgement message. Contains the size of
+    /// the message that was suppsoed to be
```
appetrosyan, line 63:
```suggestion
    /// the message that was supposed to be acknowledged. 
```

core/src/sumeragi/mod.rs
```diff
@@ -135,19 +129,20 @@ impl Sumeragi {
     #[allow(
         clippy::expect_used,
         clippy::unwrap_in_result,
-        clippy::cast_precision_loss,
         clippy::float_arithmetic,
         clippy::mutex_integer
```
appetrosyan, line 133:
```suggestion
```

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+    HandshakeLength(usize),
+    /// Parity Scale codec error
+    #[error("Parity Scale codec error")]
+    ParityScale(#[from] parity_scale_codec::Error),
+    /// Failed to create keys
+    #[error("Failed to create session key")]
+    Keys(#[from] CryptographicError),
+    /// Failed to resolve to any socket address
+    #[error("Failed to resolve to any socket address")]
+    Resolve,
```
appetrosyan, line 49: This is just a plug. We actually need some context for what we were trying to resolve.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+    /// Decryption failed
+    #[error("Decryption failed")]
+    Decrypt(aead::Error),
+    /// Encryption failed
+    #[error("Encryption failed")]
+    Encrypt(aead::Error),
+    /// Ursa Cryptography error
+    #[error("Ursa Cryptography error")]
+    Ursa(CryptoError),
+}
```
appetrosyan, line 64: When I originally added this type, it was used to encapsulate the `Cryptographer` type and make all associated functions return that type of error. Now it seems that we're better off merging this with the Error enum.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+            if let Some(PeerConnection { stream, crypto, .. }) =
+                connected_to_peers.get_mut(public_key)
+            {
+                let encrypted = crypto
+                    .encrypt(message.encode().clone())
+                    .expect("We should always be able to encrypt.");
+                let write1 = stream.write_all(&(u32::try_from(encrypted.len())).expect("Encrypted length exceeds 32 bits. This will cause issues with Parity Scale codec. Aborting. ").to_le_bytes());
+                let write2 = stream.write_all(&encrypted);
+
+                if write1.is_err() || write2.is_err() {
```
appetrosyan, line 182: This should be refactored. We have a good use for the `?` operator. 

This block is a function that we can neatly add to a `map` on `connected_to_peers.get_mut`, pipelined from `recipients.iter()` and collected into `to_disconnect_keys`. 

I'm not sure if we should remove the `expect`, or match on the returned result. We **do** have an `Error::Keys(CryptographicError::Encrypt(err))` variant, though I doubt we can meaningfully handle it inside an `iter()` block.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+        } = *self
+            .read_thread_data
+            .lock();
+
+        let mut insert_index = 0;
+        for i in 0..packet_cache.len() {
+            if let Some(content) = packet_cache[i].take() {
+                match content {
+                    SumeragiPacket(internal) => {
+                        if matches!(variant, NetworkMessageVariant::Sumeragi) {
```
appetrosyan, line 219: Rewrite using an `if let`. This does the exact same as `matches!`, which is introduced to help writing `assert!` statements and not for general use.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+    let mut packet_size_buf = [0_u8; 4];
+    stream.read_exact(&mut packet_size_buf)?;
+    let packet_size = u32::from_le_bytes(packet_size_buf);
+    let mut buf = vec![0_u8; packet_size as usize];
+    stream.read_exact(&mut buf)?;
+    let data = crypto.decrypt(buf)?;
+    if let ConnectionCheck(_) = Decode::decode(&mut data.as_slice())? {}
+    else {
+        error!("Actually not established connection to peer {other_public_key}."); // Rust errors are dumb, TODO invent an error type for this case.
+    }
```
appetrosyan, line 543: Actually, that's precisely why the Rust Errors are the way they are: you're supposed to produce new types for new kinds of events. And this would let you catch these problems with a neat syntax.

core/src/kura.rs
```diff
@@ -88,13 +87,17 @@ impl Kura {
...
 
-        let thread_handle = std::thread::spawn(move || {
-            Self::kura_recieve_blocks_loop(&kura, shutdown_receiver);
-        });
+        let thread_handle = std::thread::Builder::new()
+            .name("Kura Thread".to_owned())
+            .spawn(move || {
+                Self::kura_recieve_blocks_loop(&kura, shutdown_receiver);
+            })
+            .expect("Failed to start Kura thread. Please resolve the OS level issue, and re-start Iroha. No point in continuing.");
```
outoftardis, line 100:
```suggestion
            .expect("Failed to start Kura thread. Please resolve the OS level issue and re-start Iroha.");
```

core/src/block_sync.rs
```diff
@@ -25,66 +32,103 @@ pub struct BlockSynchronizer {
...
+                || {
+                    std::thread::sleep(Duration::from_millis(10));
+                },
+                |message| match message {
+                    NetworkMessage::BlockSync(message) => {
+                        message.into_v1().handle_message(block_sync);
+                    }
+                    #[allow(clippy::restriction)]
+                    _ => {
+                        panic!("Above call to poll_network_for_packet returned wrong packet type.")
```
outoftardis, line 105: can we say what type was expected?

core/src/block_sync.rs
```diff
@@ -135,9 +182,26 @@ pub mod message {
     }
 
     impl GetBlocksAfter {
-        /// Construct [`GetBlocksAfter`].
-        pub const fn new(hash: HashOf<VersionedCommittedBlock>, peer_id: PeerId) -> Self {
-            Self { hash, peer_id }
+        fn handle(&self, block_sync: &BlockSynchronizer) {
+            if block_sync.block_batch_size == 0 {
+                warn!("Error: not sending any blocks as batch_size is equal to zero.");
```
outoftardis, line 187: suggestions:
```suggestion
                warn!("Not sending any blocks. Reason: batch_size is zero.");
```
or
```suggestion
                warn!("Batch size is zero. Not sending any blocks.");
```

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
+//! handshake and connection logic.
```
outoftardis, line 1:
```suggestion
//! Handshake and connection logic.
```

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+            // to_socket_addrs enables dns look-ups.
+            let maybe_addr = address_candidate
+                .to_socket_addrs()
+                .unwrap_or_else(|error| {
+                    debug!(?error, "Socket addr");
+                    Vec::new().into_iter()
+                })
+                .next();
+            if maybe_addr.is_none() {
+                error!(%address_candidate, "Error can't produce address from str. ");
```
outoftardis, line 160:
```suggestion
                error!(%address_candidate, "Cannot produce address from str. ");
```

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+            {
+                let encrypted = crypto
+                    .encrypt(message.encode().clone())
+                    .expect("We should always be able to encrypt.");
+                let write1 = stream.write_all(&(u32::try_from(encrypted.len())).expect("Encrypted length exceeds 32 bits. This will cause issues with Parity Scale codec. Aborting. ").to_le_bytes());
+                let write2 = stream.write_all(&encrypted);
+
+                if write1.is_err() || write2.is_err() {
+                    error!(
+                        "Disconecting, cause: Could not post message to peer {}.",
```
outoftardis, line 184:
```suggestion
                        "Disconnecting. Cause: could not post message to peer {}.",
```

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+                .lock();
+
+            if connected_to_peers.keys().any(|key| *key == other) {
+                trace!(%other, "Dropping because already connected.");
+            } else if let Err(e) =
+                finish_connection(crypto, stream, &mut connected_to_peers, &other)
+            {
+                error!(?e,
+                    %other,
+                    "Connecting to peer failed in final step."
```
outoftardis, line 381:
```suggestion
                    "Connection to peer failed in the final step."
```

what is the final step btw? can we be more specific why the connection failed?

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+            {
+                trace!(%other, "Dropping because not in target.");
+                continue;
+            }
+
+            let mut connected_to_peers = self
+                .connected_to_peers
+                .lock();
+
+            if connected_to_peers.keys().any(|key| *key == other) {
```
Erigara, line 374: We are using `HashMap`, so [contains_key](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.contains_key) is available.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+                trace!(%other, "Dropping because already connected.");
+            } else if let Err(e) =
+                finish_connection(crypto, stream, &mut connected_to_peers, &other)
+            {
+                error!(?e,
+                    %other,
+                    "Connecting to peer failed in final step."
+                );
+            }
```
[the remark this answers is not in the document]
Erigara: Agree seems that we can express this with `and_then` pretty well.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+    fn disconnect_peers(&self, to_disconnect_keys: Vec<PublicKey>) {
+        let mut connected_to_peers = self.connected_to_peers.lock();
+        for key in to_disconnect_keys {
+            info!("Disconnecting from: {}", &key);
+            let PeerConnection { stream, .. } = connected_to_peers
+                .remove(&key)
+                .expect("Peer no longer present. This is memory corruption");
+            stream
+                .shutdown(Shutdown::Both)
+                .unwrap_or_else(|e| debug!(?e, "Error shutting down stream"));
```
outoftardis, line 413:
```suggestion
                .unwrap_or_else(|e| debug!(?e, "Shutting down the stream"));
```

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+                                .set_write_timeout(Some(P2P_TCP_TIMEOUT))
+                                .expect("Could not set write timeout on socket.");
+                            stream
+                        })
+                        .map_err(Error::Io)
+                }),
+        }
+    }
+
+    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
```
Erigara, line 479: Seems to be redundant.

cli/src/torii/routing.rs
```diff
@@ -465,11 +460,11 @@ impl Torii {
     ) -> impl warp::Filter<Extract = impl warp::Reply> + Clone + Send {
         let get_router_status = endpoint2(
             handle_status,
-            warp::path(uri::STATUS).and(add_state!(self.sumeragi, self.network)),
+            warp::path(uri::STATUS).and(add_state!(self.sumeragi, 0)),
```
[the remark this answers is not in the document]
appetrosyan: It was one of the earliest fixes. To be fair, the syntax for a single endpoint is a little confusing, so I adjusted it a bit.

client/tests/integration/connected_peers.rs
```diff
@@ -46,12 +46,12 @@ fn connected_peers_with_f(faults: u64, port: u16) {
     let unregister_peer = UnregisterBox::new(IdBox::PeerId(peer.id.clone()));
     genesis_client.submit(unregister_peer).unwrap();
     thread::sleep(pipeline_time * 2);
+    thread::sleep(std::time::Duration::from_secs(10)); // New P2P system takes a little bit of time to update connections.
```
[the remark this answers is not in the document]
appetrosyan: They aren't the only ones. I'm surprised this is an improvement over the old design. And I'm not saying this just because I wrote the old `p2p` crate.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+        let connected_to_peer_keys = {
+            let mut keys = Vec::new();
+            let mut connected_to_peers = self
+                .connected_to_peers
+                .lock();
+            for (public_key, _) in connected_to_peers.iter_mut() {
+                keys.push(public_key.clone());
+            }
+            keys
+        };
```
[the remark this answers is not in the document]
appetrosyan: I wonder if we can do better. I'm experimenting with decoupling from a vector, and just passing in an `iter()`. It should also elide an extra heap allocation.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+        }
+    }
+
+    pub(crate) fn post_to_network(&self, message: &NetworkMessage, recipients: &[PublicKey]) {
+        let mut to_disconnect_keys = Vec::new();
+
+        let mut connected_to_peers = self
+            .connected_to_peers
+            .lock();
+        for public_key in recipients.iter() {
```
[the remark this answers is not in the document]
appetrosyan: We can do one better if we use `if let`. The surprising thing about this refactor is that we have a ton of opportunities to exercise `match` to its full potential. I've not done so much pattern matching since Uni.

core/src/p2p.rs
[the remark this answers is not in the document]
appetrosyan: I advocate for extracting this into its own function and preserving the error. Moreover, the error itself is not enough. We need some context to see what's going on. Currently there's not nearly enough information to go around.

core/src/p2p.rs
```diff
@@ -0,0 +1,741 @@
...
+                        "Disconecting, cause: Could not post message to peer {}.",
+                        public_key
+                    );
+                    to_disconnect_keys.push(public_key.clone());
+                }
+            } else {
+                debug!("P2P post failed: Not connected.");
+            }
+        }
+        drop(connected_to_peers);
```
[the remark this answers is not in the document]
appetrosyan: I wonder if we'd be better off with an `AtomicCell` here. @SamHSmith had an argument for the locks not being too heavily contended, and this being OK to use Mutices held for this long.

core/src/p2p.rs
appetrosyan: It is. I'm going to work a little on `tracing` spans and `tracing_error` when the refactor is in an OK state.

core/src/p2p.rs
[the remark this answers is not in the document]
SamHSmith: In that case the thread to pursue is why are they contended? And the answer is most likely blocking I/O. The symptom is contented locks, but the real problem is that this is not a compute heavy task and so we should be able to not have contention.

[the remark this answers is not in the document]
appetrosyan: `async`

PR #2978, opened 2022-11-29
[feature]: #2231 Generate FFI wrapper API

### LINKED ISSUE #2231: FFI wrapper API

mversic:

### Description of the Change

* should be reviewed after #2994 
* the API is now such that it transfers ownership of vectors over FFI
* it's now possible to generate functions which delegate calls to FFI functions

It's not possible to return a vector from an FFI function and in instances where it is possible it's too inconvenient so I've updated the rule to allow ownership transfer for both opaque types and vectors. 

In the future, It's possible that this limitation will be further expanded to allow transferring of ownership for other types. Or we'll revert to not transferring ownership, in which case we'll disallow returning `Vec<T>` and only allow `Vec<&T>`. This warrants a separate issue to be discussed which I'll duly open

### Issue

One of the last major things to do is to support opaque types in `ffi_import`. I've already done some work around this so expect a PR soon

Closes #2231


primitives/src/conststr.rs
```diff
@@ -565,13 +565,21 @@ mod ffi {
         }
     }
 
+    // TODO: Shouldn it be cloned?
```
appetrosyan, line 568: It's safe to say that it should be. There's quite a bit of UB we could cause if we didn't do that.

ffi/src/repr_c.rs
```diff
@@ -1038,3 +1487,142 @@ union TransmuteHelper<R: Transmute> {
...
+    // Correct as long as `R::into_repr_c` doesn't return a reference to the store (`R: NonLocal`)
+    let store_borrow = &mut *addr_of_mut!(store);
+    CTypeConvert::try_from_repr_c(out_ptr, store_borrow)
+}
+
+#[cfg(test)]
+mod tests {
+    use alloc::alloc;
+
+    crate::def_ffi_fn! { dealloc }
```
mversic, line 1627: I need help here. I can't get the tests in `ffi_import` to dynamically link

mversic: if I enable it always, not just in `cfg(test)` then it works

ffi/tests/ffi_import.rs
```diff
@@ -0,0 +1,127 @@
+#![allow(unsafe_code, clippy::restriction, clippy::pedantic)]
+
+use std::{alloc, ops::Deref};
+
+use iroha_ffi::{ffi_import, FfiType, LocalRef, LocalSlice};
+
+iroha_ffi::def_ffi_fn! { dealloc }
```
mversic, line 7: I need help here. I can't get the tests in ffi_import to dynamically link against dealloc which is needed because I'm transferring ownership of vectors over ffi now

ffi/src/slice.rs
```diff
@@ -59,35 +71,42 @@ impl<T> SliceRef<T> {
...
+    pub const fn from_slice(source: Option<&[C]>) -> Self {
+        if let Some(slice) = source {
+            Self(slice.as_ptr(), slice.len())
+        } else {
+            Self(core::ptr::null(), 0)
+        }
     }
 
     /// Convert [`Self`] into a shared slice. Return `None` if data pointer is null
+    /// Unlike [`core::slice::from_raw_parts`], data pointer is allowed to be null
```
outoftardis, line 83:
```suggestion
    /// Convert [`Self`] into a shared slice. Return `None` if data pointer is null.
    /// Unlike [`core::slice::from_raw_parts`], data pointer is allowed to be null.
```

ffi/src/slice.rs
```diff
@@ -102,94 +121,79 @@ impl<T> SliceMut<T> {
...
-    pub fn from_slice(slice: &mut [T]) -> Self {
-        Self(slice.as_mut_ptr(), slice.len())
+    pub fn from_slice(source: Option<&mut [C]>) -> Self {
+        source.map_or_else(
+            || Self(core::ptr::null_mut(), 0),
+            |slice| Self(slice.as_mut_ptr(), slice.len()),
+        )
     }
 
     /// Convert [`Self`] into a mutable slice. Return `None` if data pointer is null
```
outoftardis, line 131:
```suggestion
    /// Convert [`Self`] into a mutable slice. Return `None` if data pointer is null.
```

ffi/src/slice.rs
```diff
@@ -1,50 +1,62 @@
...
+impl<C> Copy for OutBoxedSlice<C> {}
+impl<C> Clone for OutBoxedSlice<C> {
     fn clone(&self) -> Self {
         Self(self.0, self.1)
     }
 }
 
-impl<T> SliceRef<T> {
+impl<C> SliceRef<C> {
+    /// Set slice data ptr to null
```
outoftardis, line 47:
```suggestion
    /// Set the slice's data pointer to null
```

ffi/src/slice.rs
```diff
@@ -1,50 +1,62 @@
...
-use core::mem::MaybeUninit;
+use alloc::{boxed::Box, vec::Vec};
+use core::slice;
 
-use crate::{FfiReturn, OutPtrOf, ReprC, Result};
+use crate::ReprC;
 
-/// Immutable slice with a defined C ABI layout. Consists of data pointer and length
+/// Immutable slice `&[C]` with a defined C ABI layout. Consists of data pointer and length.
+/// `Option<&[C]>` can also be represented by setting the data pointer to a null pointer.
```
outoftardis, line 9:
```suggestion
/// Immutable slice `&[C]` with a defined C ABI layout. Consists of a data pointer and a length.
/// `Option<&[C]>` can also be represented by setting the data pointer to null.
```


or:

```suggestion
/// Immutable slice `&[C]` with a defined C ABI layout. Consists of a data pointer and a length.
/// If the data pointer is set to `null`, the struct represents `Option<&[C]>`.
```


similar changes to other structs below

ffi/src/repr_c.rs
```diff
@@ -1,31 +1,36 @@
-#![allow(trivial_casts, clippy::undocumented_unsafe_blocks)]
-
-//! Logic related to the conversion of IR types to equivalent robust C types. Most of the
-//! types that implement [`Ir`], i.e. provide conversion into IR, will get an automatic
-//! implementation of [`CType`] and consequently an implementation of [`FfiType`]
+#![allow(trivial_casts)]
 
+//! Logic related to the conversion of IR types to equivalent robust C types. Types that are mapped into
+//! one of the predefined [`Ir`] types, will be provided an automatic implementation of traits in this module.
```
outoftardis, line 4:
```suggestion
//! one of the predefined [`Ir`] types will be provided an automatic implementation of traits in this module.
```

ffi/src/repr_c.rs
```diff
@@ -1,31 +1,36 @@
...
+//!
+//! Traits in this module mainly exists to bridge the gap between IR and C type equivalents. User should
+//! only implement these traits if none of the predefined IR types provide an adequate mapping
```
outoftardis, line 7:
```suggestion
//! Traits in this module mainly exist to bridge the gap between IR and C type equivalents. User should
//! only implement these traits if none of the predefined IR types provide an adequate mapping.
```

ffi/src/repr_c.rs
```diff
@@ -1,31 +1,36 @@
...
+    FfiConvert, FfiOutPtr, FfiOutPtrRead, FfiOutPtrWrite, FfiOutput, FfiReturn, FfiType, LocalRef,
+    LocalSlice, ReprC, Result,
 };
 
-/// A type that can be converted into some C type. When implemented on [`Ir::Type`] the blanket
-/// implementations ensures that [`FfiType`] is also implemented on the type implementing [`Ir`].
-/// This trait mainly exists to move types of internal representation into C type equivalents.
+crate::decl_ffi_fn! { dealloc }
+
+/// A type that can be converted into some C type.
```
outoftardis, line 20:
```suggestion
/// A type that can be converted into a C type.
```

ffi/src/repr_c.rs
```diff
@@ -1,31 +1,36 @@
...
     /// C type current type can be converted into
     type ReprC: ReprC;
 }
 
-/// Conversion utility [`CType`]s. When implemented on [`Ir::Type`] the blanket implementation
-/// ensures that [`FfiType`] is also implemented on the type implementing [`Ir`]. This trait
-/// mainly exists to move types of internal representation into C type equivalents. When
-/// implementing make sure that the implementation doesn't conflict with [`NonLocal`]
-pub trait CTypeConvert<'itm, S, C>: Sized {
+/// Facilitates conversion of rust types to/from `ReprC` types.
```
outoftardis, line 29:
```suggestion
/// The trait facilitates conversion of rust types to/from `ReprC` types.
```

similar in other descriptions that start with "facilitates"

ffi/src/repr_c.rs
```diff
@@ -51,54 +56,138 @@ pub trait CTypeConvert<'itm, S, C>: Sized {
     unsafe fn try_from_repr_c(source: C, store: &'itm mut Self::FfiStore) -> Result<Self>;
 }
 
-/// Provides the capability to use the type as an out-pointer.
-/// This trait mainly exists to be implemented on IR types.
+/// Replaces output type of the wrapper function generated by `ffi_import`. Most notably, the
+/// necessity to replace the output type for a wrapper function arises when returning a reference
+/// that use store during conversion(i.e. that are cloned) or when returning an opaque type reference.
```
outoftardis, line 61:
```suggestion
/// The trait is used to replace the output type of the wrapper function generated by `ffi_import`.
/// 
/// Most notably, the
/// necessity to replace the output type for a wrapper function arises when the wrapper function returns:
/// - a reference that uses store during conversion (i.e. that are cloned)
/// - an opaque type reference
```

ffi/src/repr_c.rs
```diff
@@ -51,54 +56,138 @@ pub trait CTypeConvert<'itm, S, C>: Sized {
...
+    /// Check [`CConvert::try_from_repr_c`]
+    unsafe fn try_read_out(out_ptr: Self::OutPtr) -> Result<Self>;
 }
 
-/// Marker trait indicating that a type converted into corresponding [`CType`] doesn't
-/// reference conversion context. This is useful to determine which types can be returned
-/// from FFI function and how they are returned.
+/// Marker trait indicating that [`CTypeConvert::into_repr_c`] and [`CTypeConvert::try_from_repr_c`] don't
+/// return a reference to the store. This is useful to determine which(and how) types can be
+/// returned from an FFI function considering that, after return, local context is destroyed
```
outoftardis, line 111:
```suggestion
/// return a reference to the store. 
///
/// Since local context is destroyed after return, this trait is useful in determining which types can be
/// returned from an FFI function.
```

ffi/src/repr_c.rs
```diff
@@ -51,54 +56,138 @@ pub trait CTypeConvert<'itm, S, C>: Sized {
...
 /// # Example
 ///
-/// 1. `&mut [u8]`
-/// This type will be converted to [`SliceMut<u8>`] and during conversion will not make use
-/// of the store (in any direction). This means that the type's corresponding out-pointer
-/// will be `*mut SliceMut<u8>`, i.e. a pointer to a location where the slice's
-/// data pointer and length will be stored
+/// 1. `&[u8]` implements [`NonLocal`]
+/// This type will be converted to [`SliceRef<u8>`] and during conversion will not make use
+/// of the store(in any direction). The corresponding out-pointer will be `*mut SliceRef<u8>`
```
outoftardis, line 117:
```suggestion
/// of the store (in any direction). The corresponding out-pointer will be `*mut SliceRef<u8>`
```

ffi/src/repr_c.rs
```diff
@@ -1038,3 +1487,135 @@ union TransmuteHelper<R: Transmute> {
...
+) {
+    let mut store = Default::default();
+    // NOTE: Bypasses the erroneous lifetime check.
+    // Correct as long as `R::into_repr_c` doesn't return a reference to the store (`R: NonLocal`)
+    let store_borrow = &mut *addr_of_mut!(store);
+    out_ptr.write(CTypeConvert::into_repr_c(source, store_borrow));
+}
+
+/// Reads a rust value from an out-pointer of any type that doesn't return a reference
+/// reference to the store during serialization from an FFI-compatible type
```
outoftardis, line 1600:
```suggestion
/// Read a rust value from an out-pointer of any type that doesn't return a reference
/// to the store during serialization from an FFI-compatible type
```

ffi/src/repr_c.rs
```diff
@@ -1038,3 +1487,135 @@ union TransmuteHelper<R: Transmute> {
...
+    Ok(Vec::from_raw_parts(
+        vec.as_mut_ptr().cast(),
+        vec.len(),
+        vec.capacity(),
+    ))
+}
+
+/* ----------------------------------------------------------------------------------- */
+
+/// Writes a rust value into an out-pointer of any type that doesn't return a
```
outoftardis, line 1578:
```suggestion
/// Write a rust value into an out-pointer of any type that doesn't return a
```

ffi/src/lib.rs
```diff
@@ -246,6 +376,86 @@ pub struct Extern {
     __marker: core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
 }
 
+/// Defines the correct [`FfiOutput::ReturnType`] out of the given [`COutput::ReturnType`].
```
outoftardis, line 379:
```suggestion
/// Define the correct [`FfiOutput::ReturnType`] out of the given [`COutput::ReturnType`].
```

ffi/src/lib.rs
```diff
@@ -246,6 +376,86 @@ pub struct Extern {
...
+/// use iroha_ffi::FfiType;
+///
+/// #[derive(FfiType)]
+/// #[ffi_type(unsafe {robust})]
+/// #[repr(transparent)]
+/// struct Example(u32);
+///
+/// /*
+/// Due to the fact that implementations of traits for [`Transparent`] structures delegate to the inner
+/// type `FfiType` macro expansion will produce the following implementation of [`COutput`] for `Example`:
```
outoftardis, line 394:
```suggestion
/// type `FfiType`, macro expansion will produce the following implementation of [`COutput`] for `Example`:
```

ffi/src/lib.rs
```diff
@@ -76,28 +76,146 @@ pub trait FfiConvert<'itm, C: ReprC>: Sized {
...
+///     let output = FfiConvert::try_from_ffi(output, &mut out_store);
+///     LocalRef(out_store.0, core::marker::PhantomData)
+/// } */
+/// ```
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
+pub struct LocalRef<'data, R>(R, core::marker::PhantomData<&'data ()>);
+
+/// Shared slice that owns it's referent. Used to facilitate returning of types(e.g. `&[(u32, u32)]`),
+/// that use store during FFI serialization, from the wrapper functions generated by `ffi_import`
```
outoftardis, line 169:
```suggestion
/// Shared slice that owns its referent. 
/// This struct is used when wrapper functions generated by `ffi_import` return types that use store during FFI serialization (e.g. `&[(u32, u32)]`).
```

ffi/src/lib.rs
```diff
@@ -76,28 +76,146 @@ pub trait FfiConvert<'itm, C: ReprC>: Sized {
...
-    /// * All conversions from a pointer must ensure pointer validity beforehand
-    // TODO: It could return bool for successful vs not?
-    // It depends on whether or not `source` should be checked for validity?
-    unsafe fn write(self, source: C) -> Result<()>;
+    /// Check [`FfiConvert::try_from_ffi`]
+    unsafe fn try_read_out(out_ptr: Self::OutPtr) -> Result<Self>;
 }
 
+/// Reference that owns it's referent. Used to facilitate returning of types(e.g. `&(u32, u32)`),
+/// that use store during FFI serialization, from the wrapper functions generated by `ffi_import`
```
outoftardis, line 118: see comment for `LocalSlice` desc

ffi/src/lib.rs
```diff
@@ -76,28 +76,146 @@ pub trait FfiConvert<'itm, C: ReprC>: Sized {
     unsafe fn try_from_ffi(source: C, store: &'itm mut Self::FfiStore) -> Result<Self>;
 }
 
-/// Marker trait designating a type that can be returned from an FFI function as an out-pointer
+/// Replaces output type of the wrapper function generated by [`ffi_import`]. Most notably, the
+/// necessity to replace the output type for a wrapper function arises when returning a reference
+/// that use store during conversion(i.e. that are cloned) or when returning an opaque type reference.
```
outoftardis, line 81: see comment for similar function in another file

ffi/src/lib.rs
```diff
@@ -246,6 +378,86 @@ pub struct Extern {
...
+///
+/// impl COutput for Example {
+///     type ReturnType = u32;
+/// }
+///
+/// which is corrected via [`ReturnTypeOf`] so that functions generated by [`ffi_import`] return [`Example`]
+/// */
+/// ```
+///
+pub trait ReturnTypeOf<T> {
```
mversic, line 406: I'm quite unhappy that I had to introduce this trait, but on the upside I think all implementations that will ever be needed have been covered in this PR

ffi/derive/src/wrapper.rs
```diff
@@ -67,47 +72,100 @@ fn wrap_method(fn_descriptor: &FnDescriptor) -> TokenStream {
...
-
-    let mut stmts = quote! {};
     for arg in &fn_descriptor.input_args {
         stmts.extend(gen_arg_src_to_ffi(arg, false));
     }
+    if let Some(arg) = &fn_descriptor.output_arg {
+        let name = &arg.name();
+
+        stmts.extend(quote! {
+            let mut #name = core::mem::MaybeUninit::uninit();
```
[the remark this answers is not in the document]
mversic: out pointer can never be null, it needs to be a preallocated space for FFI function to fill with result.

let's say there is some ffi function with the following declaration:
```rs
    pub extern fn ffi_fn(out_ptr: *mut u32);
```

What I do here is roughly like this:
```rs
    let output = MaybeUninit::uninit();
    ffi_fn(out_ptr.as_mut_ptr());
    let output = output.assume_init();
```

leaving out unnecessary details

ffi/tests/ffi_import.rs
mversic: in the `repr_c.rs` module I'm calling `__dealloc` function which is only available at link time. You can see it's declared there with `decl_ffi_fn`. I would like to provide a definition for `__dealloc` function later - in this case in `ffi_import`

ffi/src/handle.rs
```diff
@@ -67,24 +67,18 @@ macro_rules! def_ffi_fn {
...
+            handle_id: <$crate::handle::Id as FfiType>::ReprC,
             handle_ptr: *const core::ffi::c_void,
-            output_ptr: *mut *mut core::ffi::c_void
+            out_ptr: *mut *mut core::ffi::c_void
         ) -> $crate::FfiReturn {
             $crate::def_ffi_fn!(@catch_unwind {
-                use core::borrow::Borrow;
-
                 // False positive - doesn't compile otherwise
                 #[allow(clippy::let_unit_value)]
```
Erigara, line 76: Seems to be unnecessary now.

ffi/src/primitives.rs
```diff
@@ -135,13 +210,36 @@ mod wasm {
...
+                    CTypeConvert::<Vec<NonWasmIntPrimitive>, _>::into_repr_c(self, &mut store);
+                    out_ptr.write(OutBoxedSlice::from_vec(Some(store)));
+                }
+            }
+            impl COutPtrRead<Vec<NonWasmIntPrimitive>> for Vec<$src> {
+                unsafe fn try_read_out(out_ptr: Self::OutPtr) -> Result<Self> {
+                    let slice = SliceMut::from_raw_parts_mut(out_ptr.as_mut_ptr(), out_ptr.len());
+                    let res = CTypeConvert::<Vec<NonWasmIntPrimitive>, _>::try_from_repr_c(slice, &mut ());
+                    let layout = Layout::new::<Vec<$src>>();
+                    __dealloc(out_ptr.as_mut_ptr().cast(), layout.size(), layout.align());
```
Erigara, line 238: Correct me if i miss something, but seems like we are trying to deallocate not memory vector points to, but vector itself, is this correct behavour?

mversic: Yes, this is intended behavior

mversic: yes, `Vec<T>` has to be deallocated, not `T` because(at the current moment?) `T` will never carry ownership. Specifically in this case it's about integers like `Vec<u8>`

mversic: very good observation, I believe I fixed it, have a 2nd look please

Erigara: Seems fine now.
One small concern calling `allocate` on ZST is `UB` and i'm not sure if it's fine to call `deallocate` if `slice.len() == 0`.

ffi/src/lib.rs
```diff
@@ -76,28 +78,148 @@ pub trait FfiConvert<'itm, C: ReprC>: Sized {
...
+///     let mut in_store = ();
+///     let a = a.into_ffi(a, &mut in_store);
+///
+///     let mut output = core::mem::MaybeUninit::uninit();
+///     __func_returns_non_local(a, output.as_mut_ptr());
+///
+///     let out_store = ();
+///     let output = output.assume_init();
+///
+///     // &[(u32, u32)] doesn't reference local scope of a function
```
Erigara, line 199:
```suggestion
///     // &[u32] doesn't reference local scope of a function
```

ffi/src/primitives.rs
mversic: looking at the implementation of `Vec`, you are correct again. Will fix

data_model/src/lib.rs
```diff
@@ -557,6 +557,7 @@ pub enum Value {
     FfiType,
     IntoSchema,
 )]
+#[ffi_type(opaque)]
```
mversic, line 560: troubles with `Option<&NumericValue>`. Opened a ticket to resolve at some point in the future

ffi/tests/ffi_import.rs
mversic: Roman saved the day, I was missing imports in other test files

ffi/src/repr_c.rs
```diff
@@ -117,6 +206,40 @@ impl<'itm, R: CTypeConvert<'itm, S, C> + Clone, S: Cloned, C: ReprC>
...
+        out_ptr.write(store.0.expect("Store must be initialized"));
+    }
+}
+impl<'itm, R: NonLocal<S> + CTypeConvert<'itm, S, R::ReprC> + Clone + 'itm, S: Cloned + 'itm>
+    COutPtrRead<&'itm S> for LocalRef<'itm, R>
+{
+    unsafe fn try_read_out(out_ptr: Self::OutPtr) -> Result<Self> {
+        let mut store = Default::default();
+        // NOTE: Bypasses the erroneous lifetime check.
+        // Correct as long as `R::into_repr_c` doesn't return a reference to the store (`R: NonLocal`)
```
Erigara, line 236: Maybe? (in other places too)

```suggestion
        // Correct as long as `R::try_from_repr_c` doesn't return a reference to the store (`R: NonLocal`)
```

ffi/src/slice.rs
```diff
@@ -1,50 +1,64 @@
...
+/// Owned slice `Box<[C]>` with a defined C ABI layout. Consists of a data pointer and a length.
+/// Used in place of a function out-pointer to transfer ownership of the slice to the caller.
+/// If the data pointer is set to `null`, the struct represents `Option<Box<[C]>>`.
 #[repr(C)]
-#[derive(Debug, Clone, Copy)]
-pub struct OutBoxedSlice<T>(pub *mut T, pub usize, pub *mut isize);
+#[derive(Debug)]
+pub struct OutBoxedSlice<C>(*mut C, usize);
+
+crate::decl_ffi_fn! { dealloc }
```
Erigara, line 27: Maybe move on top? after imports

Erigara (approved):

ffi/derive/tests/ui_fail/nested_owned_structure.rs
```diff
@@ -1,21 +1,18 @@
...
+use iroha_ffi::{ffi_export, FfiType};
 
 /// Ffi structure
 #[derive(Clone, FfiType)]
 pub struct FfiStruct;
 
 /// Return nested structure
 #[ffi_export]
-pub fn return_nested() -> Nested {
+pub fn return_nested() -> Vec<Vec<FfiStruct>> {
```
appetrosyan, line 11: Consider turning this into a doctest, and hard-coding the error code.

ffi/tests/ffi_import.rs
appetrosyan: Roman deserves a commendation.

appetrosyan (approved):

ffi/derive/tests/ui_fail/nested_owned_structure.rs
mversic: I'd rather leave it for if/when we'll be moving all ui tests to doc tests

PR #2986, opened 2022-12-01
[refactor] #2974: Prevent block spoofing without full revalidation

### LINKED ISSUE #2974: [suggestion]: Prevent block spoofing without full revalidation
Currently it's possible to change block transactions without changing resulting hash of the block.

This happens because block hash is computed based only on block header.

Erigara:

### Description of the Change

- Add `CommittedCandidateBlock` block state used for transferring between peers, requires hash revalidation to obtain `ComittedBlock`;
- Change `BlockComitted` message to send only hash and signatures;

### Issue

Closes #2974, #2955.

### Benefits

More optimized communication.

### Possible Drawbacks

Requires peer to preserve latest created block.

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -29,7 +29,7 @@ use serde::Serialize;
 
 use crate::{
     prelude::*,
-    sumeragi::network_topology::Topology,
+    sumeragi::network_topology::{Role, Topology},
     tx::{TransactionValidator, VersionedAcceptedTransaction},
 };
 
@@ -372,15 +372,15 @@ impl ValidBlock {
         HashOf::new(&self.header).transmute()
     }
 
-    /// Sign this block and get `ValidSignedBlock`.
+    /// Sign this block and get `SignedBlock`.
     ///
     /// # Errors
     /// Fails if signature generation fails
-    pub fn sign(self, key_pair: KeyPair) -> Result<ValidSignedBlock> {
+    pub fn sign(self, key_pair: KeyPair) -> Result<SignedBlock> {
         let signature = SignatureOf::from_hash(key_pair, &self.hash().transmute())
             .wrap_err(format!("Failed to sign block with hash {}", self.hash()))?;
         let signatures = SignaturesOf::from(signature);
-        Ok(ValidSignedBlock {
+        Ok(SignedBlock {
             header: self.header,
             rejected_transactions: self.rejected_transactions,
             transactions: self.transactions,
@@ -390,9 +390,9 @@ impl ValidBlock {
     }
 }
 
-/// After receiving first signature, `ValidBlock` can transform into `ValidSignedBlock`.
+/// After receiving first signature, `ValidBlock` can transform into `SignedBlock`.
 #[derive(Debug, Clone)]
-pub struct ValidSignedBlock {
+pub struct SignedBlock {
     /// Block header
     pub header: BlockHeader,
     /// Array of all rejected transactions in this block.
@@ -405,10 +405,12 @@ pub struct ValidSignedBlock {
     pub event_recommendations: Vec<Event>,
 }
 
-impl ValidSignedBlock {
+impl SignedBlock {
     /// Commit block to the store.
+    /// When calling this function, the user is responsible for the validity of the block signatures.
+    /// Preference should be given to [`Self::commit`], where signature verification is built in.
     #[inline]
-    pub fn commit(self) -> CommittedBlock {
+    pub fn commit_unchecked(self) -> CommittedBlock {
         let Self {
             header,
             rejected_transactions,
@@ -426,12 +428,38 @@ impl ValidSignedBlock {
         }
     }
 
+    /// Verify signatures and commit block to the store.
+    ///
+    /// # Errors
+    /// - If signatures verification fails
+    #[inline]
+    pub fn commit(self, topology: &Topology) -> Result<CommittedBlock, eyre::Report> {
+        topology
+            .filter_signatures_by_roles(
+                &[
+                    Role::ValidatingPeer,
+                    Role::Leader,
+                    Role::ProxyTail,
+                    Role::ObservingPeer,
+                ],
+                self.verified_signatures(),
+            )
+            .len()
+            .ge(&topology.min_votes_for_commit())
+            .then_some(())
+            .ok_or_else(|| {
+                eyre!("The block doesn't have enough valid signatures to be committed.")
+            })?;
+
+        Ok(self.commit_unchecked())
+    }
+
     /// Calculate the hash of the current block.
     pub fn hash(&self) -> HashOf<Self> {
         HashOf::new(&self.header).transmute()
     }
 
-    /// Add additional signatures for `ValidSignedBlock`.
+    /// Add additional signatures for `SignedBlock`.
     ///
     /// # Errors
     /// Fails if signature generation fails
@@ -476,8 +504,8 @@ impl ValidSignedBlock {
     }
 }
 
-impl From<&ValidSignedBlock> for Vec<Event> {
-    fn from(block: &ValidSignedBlock) -> Self {
+impl From<&SignedBlock> for Vec<Event> {
+    fn from(block: &SignedBlock) -> Self {
         block
             .transactions
             .iter()
@@ -565,7 +593,7 @@ impl VersionedCandidateBlock {
         wsv: &WorldStateView,
         latest_block: &HashOf<VersionedCommittedBlock>,
         block_height: u64,
-    ) -> Result<ValidSignedBlock, eyre::Report> {
+    ) -> Result<SignedBlock, eyre::Report> {
         self.into_v1()
             .revalidate(transaction_validator, wsv, latest_block, block_height)
     }
@@ -632,7 +660,7 @@ impl CandidateBlock {
         wsv: &WorldStateView,
         latest_block: &HashOf<VersionedCommittedBlock>,
         block_height: u64,
-    ) -> Result<ValidSignedBlock, eyre::Report> {
+    ) -> Result<SignedBlock, eyre::Report> {
         if self.is_empty() {
             bail!("Block is empty");
         }
@@ -643,7 +671,7 @@ impl CandidateBlock {
 
         if latest_block != &self.header.previous_block_hash {
             bail!(
-                "Latest block hash mismatch. Expected: {}, actual: {}",
+                "Mismatch between the actual and expected hashes of the latest block. Expected: {}, actual: {}",
                 latest_block,
                 &self.header.previous_block_hash
             );
@@ -651,7 +679,7 @@ impl CandidateBlock {
 
         if block_height + 1 != self.header.height {
             bail!(
-                "Block heights are in an inconsistent state. Expected: {}, actual: {}",
+                "Mismatch between the actual and expected heights of the block. Expected: {}, actual: {}",
                 block_height + 1,
                 self.header.height
             );
@@ -675,7 +703,7 @@ impl CandidateBlock {
             .eq(&header.transactions_hash)
             .then_some(())
             .ok_or_else(|| {
-                eyre!("Block header transactions hash does not match actual transactions hash")
+                eyre!("The transaction hash stored in the block header does not match the actual transaction hash.")
             })?;
 
         rejected_transactions
@@ -686,7 +714,7 @@ impl CandidateBlock {
             .unwrap_or(Hash::zeroed().typed())
             .eq(&header.rejected_transactions_hash)
             .then_some(())
-            .ok_or_else(|| eyre!("Block header rejected transactions hash does not match actual rejected transaction hash"))?;
+            .ok_or_else(|| eyre!("The hash of a rejected transaction stored in the block header does not match the actual hash or this transaction."))?;
 
         // Check that valid transactions are still valid
         let transactions = transactions
@@ -734,7 +762,7 @@ impl CandidateBlock {
             })
             .wrap_err("Error during transaction revalidation")?;
 
-        Ok(ValidSignedBlock {
+        Ok(SignedBlock {
             header,
             transactions,
             rejected_transactions,
@@ -744,9 +772,9 @@ impl CandidateBlock {
     }
 }
 
-impl From<ValidSignedBlock> for CandidateBlock {
-    fn from(valid_block: ValidSignedBlock) -> Self {
-        let ValidSignedBlock {
+impl From<SignedBlock> for CandidateBlock {
+    fn from(valid_block: SignedBlock) -> Self {
+        let SignedBlock {
             header,
             rejected_transactions,
             transactions,
@@ -769,15 +797,16 @@ impl From<ValidSignedBlock> for CandidateBlock {
     }
 }
 
-impl From<ValidSignedBlock> for VersionedCandidateBlock {
-    fn from(valid_block: ValidSignedBlock) -> Self {
+impl From<SignedBlock> for VersionedCandidateBlock {
+    fn from(valid_block: SignedBlock) -> Self {
         CandidateBlock::from(valid_block).into()
     }
 }
+
 declare_versioned_with_scale!(VersionedCommittedBlock 1..2, Debug, Clone, iroha_macro::FromVariant, IntoSchema, Serialize);
 
 impl VersionedCommittedBlock {
-    /// Converts from `&VersionedCommittedBlock` to V1 reference
+    /// Convert from `&VersionedCommittedBlock` to V1 reference
     #[inline]
     pub const fn as_v1(&self) -> &CommittedBlock {
         match self {
@@ -785,7 +814,7 @@ impl VersionedCommittedBlock {
         }
     }
 
-    /// Converts from `&mut VersionedCommittedBlock` to V1 mutable reference
+    /// Convert from `&mut VersionedCommittedBlock` to V1 mutable reference
     #[inline]
     pub fn as_mut_v1(&mut self) -> &mut CommittedBlock {
         match self {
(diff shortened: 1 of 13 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -699,113 +776,86 @@ pub fn run<F>(
...
+                        .sign(key_pair_clone)
+                        .expect("maybe we should handle this error");
+
+                    let voting_block = VotingBlock::new(signed_block.clone());
+                    voting_block_option = Some(voting_block);
                 }
-            } else {
-                should_sleep = true;
+                Some(msg) => {
+                    trace!(?msg, "Observing peer not handling message");
```
outoftardis, line 835:
```suggestion
                    trace!(?msg, "Sumeragi Observing Peer is not handling the message");
```

though it's not entirely clear to me what exactly this means and entails

core/src/sumeragi/main_loop.rs
```diff
@@ -699,113 +776,86 @@ pub fn run<F>(
...
+            match maybe_incoming_message.take() {
+                Some(Message::TransactionForwarded(TransactionForwarded { transaction })) => {
+                    match VersionedAcceptedTransaction::from_transaction(
+                        transaction.into_v1(),
+                        &sumeragi.transaction_limits,
+                    ) {
+                        Ok(transaction) => match sumeragi.queue.push(transaction, &state.wsv) {
+                            Err((_, crate::queue::Error::InBlockchain)) | Ok(_) => (),
+                            Err((_, err)) => {
+                                error!(%err, "Error while pushing transaction into queue?");
```
outoftardis, line 849: why `?` at the end?

core/src/sumeragi/main_loop.rs
```diff
@@ -699,113 +776,86 @@ pub fn run<F>(
...
-                                Ok(block) => block,
-                                Err(err) => {
-                                    warn!(%err);
-                                    continue;
-                                }
-                            };
-                            commit_block(sumeragi, block, &mut state);
+                        },
+                        Err(err) => {
+                            error!(%err, "Received transaction that did not pass transaction limits.");
```
outoftardis, line 853:
```suggestion
                            error!(%err, "Received a transaction that exceeds transaction limits.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
             }
         } else if state.current_topology.role(&sumeragi.peer_id) == Role::ValidatingPeer {
-            if let Some(incoming_message) = maybe_incoming_message.take() {
-                match incoming_message {
-                    Message::BlockCreated(block_created) => {
-                        let block = block_created.block;
+            match maybe_incoming_message.take() {
+                Some(Message::BlockCreated(BlockCreated { block })) => {
+                    if voting_block_option.is_some() {
+                        warn!("Already have block, ignoring.");
```
outoftardis, line 944:
```suggestion
                        warn!("Sumeragi Validating Peer already has a block. Ignoring.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
+                        continue;
+                    }
 
-                        if voting_block_option.is_some() {
-                            warn!("Already have block, ignoring.");
-                            continue;
-                        }
+                    trace!("I, a validating peer, have received a block.");
```
outoftardis, line 948:
```suggestion
                    trace!("Sumeragi Validating Peer received a block.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
-                            continue;
-                        }
+                    if state
+                        .current_topology
+                        .filter_signatures_by_roles(&[Role::Leader], block.verified_signatures())
+                        .is_empty()
+                    {
+                        error!(
+                            role = ?state.current_topology.role(&sumeragi.peer_id),
+                            "Rejecting Block as it is not signed by leader.",
```
outoftardis, line 957:
```suggestion
                            "The block is rejected as it is not signed by the leader.",
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
-                            sumeragi.events_sender.send(event).unwrap_or(0);
-                        }
+                    if state
+                        .current_topology
+                        .filter_signatures_by_roles(&[Role::Leader], block.verified_signatures())
+                        .is_empty()
+                    {
+                        error!(
+                            role = ?state.current_topology.role(&sumeragi.peer_id),
+                            "Rejecting Block as it is not signed by leader.",
```
outoftardis, line 1036:
```suggestion
                            "The block is rejected as it is not signed by the leader.",
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
-                        trace!("I, the proxy tail, have received a block.");
-
-                        if block.header().is_genesis() {
-                            warn!("Rejecting block because it is genesis.");
-                            continue;
-                        }
+            match maybe_incoming_message.take() {
+                Some(Message::BlockCreated(BlockCreated { block })) => {
+                    if voting_block_option.is_some() {
+                        warn!("Already have block, ignoring.");
```
outoftardis, line 1018:
```suggestion
                        warn!("Sumeragi Proxy Tail already has a block. Ignoring.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
-                            )
-                            .is_empty()
-                        {
-                            error!(
-                                role = ?state.current_topology.role(&sumeragi.peer_id),
-                                "Rejecting Block as it is not signed by leader.",
-                            );
-                            continue;
-                        }
+                    trace!("I, the proxy tail, have received a block.");
```
outoftardis, line 1022:
```suggestion
                    trace!("Sumeragi Proxy Tail received a block.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
-                            ) {
-                                Ok(block) => block,
-                                Err(err) => {
-                                    warn!(%err);
-                                    continue;
-                                }
-                            }
-                        };
+                    if block.header().is_genesis() {
+                        warn!("Rejecting block because it is genesis.");
```
outoftardis, line 1025:
```suggestion
                        warn!("The block is rejected because it is the genesis block.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -1294,50 +1285,63 @@ where
...
+                        .genesis_topology
+                        .clone()
+                        .take()
+                        .expect("We just checked that it is some");
+                    commit_block(sumeragi, block, state);
+                    info!("Genesis block received and committed.");
+                    return Err(EarlyReturn::GenesisBlockReceivedAndCommitted);
                 }
-            },
+                debug!("Received block that was not genesis.");
```
outoftardis, line 1343:
```suggestion
                debug!("Received a block that was not genesis.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -1294,50 +1285,63 @@ where
...
-                        commit_block(sumeragi, block, state);
-                        info!("Genesis block received and committed.");
-                        return Err(EarlyReturn::GenesisBlockReceivedAndCommitted);
                     }
-                    debug!("Received block that was not genesis.");
-                }
-                msg => {
-                    trace!(?msg, "Not handling message, waiting genesis.");
+                    msg => {
+                        trace!(?msg, "Not handling message, waiting genesis.");
```
outoftardis, line 1327:
```suggestion
                        trace!(?msg, "Not handling the message, waiting for genesis.");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +938,171 @@ pub fn run<F>(
...
+
+                    for sig in valid_signatures {
+                        block_signature_acc.push((hash, sig));
                     }
                 }
-            } else {
-                // if there is no message — sleep
-                should_sleep = true;
+                Some(msg) => {
+                    trace!(?msg, "Proxy tail not handling message");
```
outoftardis, line 1103:
```suggestion
                    trace!(?msg, "Sumeragi Proxy Tail is not handling the message");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -481,6 +478,78 @@ fn handle_role_agnostic_messages<F>(
...
+                        actual = %block.header().previous_block_hash,
+                        "Latest block hash mismatch."
+                    );
+                    return;
+                }
+                if state.latest_block_height + 1 != block.header().height {
+                    warn!(
+                        expected = state.latest_block_height + 1,
+                        actual = block.header().height,
+                        "Block heights are in an inconsistent state."
```
outoftardis, line 501: does this mean that their is a mismatch between actual and expected block heights? I think this message should be rewritten

core/src/block.rs
```diff
@@ -989,6 +983,141 @@ impl From<&CommittedBlock> for Vec<Event> {
...
+        self.as_v1()
+            .verified_signatures()
+            .map(SignatureOf::transmute_ref)
+    }
+
+    /// Revaidate transactions' hashes and produce [`VersionedCommittedBlock`]
+    ///
+    /// # Errors
+    /// - If transactions' hashes doesn't match hashes stored in block header
+    pub fn revalidate_hashes(self) -> Result<VersionedCommittedBlock, eyre::Report> {
```
mversic, line 1038: I believe that additional invariant that `CommittedBlock` offers over `ValidSignedBlock` is that it has enough valid signatures? In that case check in this function is not sufficient to produce `CommittedBlock`

mversic: if you're going to do the check after creating `CommittedBlock` then we don't need this state. I would consider accepting that as well but then you would have to remove the `CommittedBlock` state from the codebase. In this case you would do the appropriate validation of `ValidSignedBlock` depending on the context and where it's used. But I think I would prefer that we keep the `CommittedBlock` state

mversic (changes requested):

core/src/sumeragi/main_loop.rs
```diff
@@ -481,6 +478,78 @@ fn handle_role_agnostic_messages<F>(
...
+                if valid_signatures.len() >= state.current_topology.min_votes_for_commit() {
+                    commit_block(sumeragi, block, state);
+                }
+            }
+            Message::BlockCommitted(BlockCommitted { hash, signatures }) => {
+                let block = if voting_block_option.is_some()
+                    && hash
+                        == voting_block_option
+                            .as_ref()
+                            .expect("Voting block is `Some`")
```
mversic, line 523: I think you can write this without using expect

core/src/sumeragi/main_loop.rs
```diff
@@ -304,12 +303,8 @@ where
...
     for event in events_buffer.into_iter().chain(Vec::<Event>::from(&block)) {
         trace!(?event);
-        sumeragi
-            .events_sender
-            .send(event)
-            .map_err(|e| error!(%e, "Some events failed to be sent"))
-            .unwrap_or(0);
-        // Essentially log and ignore.
+        // Ignoring error is fine here because lack of event consumer is not an error
+        sumeragi.events_sender.send(event).unwrap_or(0);
```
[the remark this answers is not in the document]
Erigara: But for what? It's just way to ignore result of send/

core/src/block.rs
```diff
@@ -914,7 +914,7 @@ impl From<CommittedBlock> for ValidSignedBlock {
     }
 }
```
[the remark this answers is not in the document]
Erigara: Sure, this conversation was not used anyway.

core/src/sumeragi/main_loop.rs
Erigara: typo i believe

core/src/sumeragi/main_loop.rs
```diff
@@ -520,7 +589,10 @@ fn compare_view_change_index_and_block_height_to_old<F>(
         *has_sent_transactions = false;
 
         *old_view_change_index = current_view_change_index;
-        trace!("View change to attempt #{}", current_view_change_index);
+        error!(
+            %current_view_change_index,
+            "Consensus hiccup: New View change to attempt.",
```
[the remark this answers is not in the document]
Erigara: This message mean that peers failed to agreed on block during current view change, and we are updating view change to try again.

core/src/sumeragi/main_loop.rs
```diff
@@ -481,6 +478,78 @@ fn handle_role_agnostic_messages<F>(
...
+                            .hash()
+                {
+                    let mut block = voting_block_option
+                        .take()
+                        .expect("Voting block is `Some`")
+                        .block;
+                    block.signatures.extend(signatures);
+                    block
+                } else {
+                    error!("block committed is not relevant block");
```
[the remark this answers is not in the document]
Erigara: In this context `relevant` means that hash received from `BlockCommitted` message is the hash of block stored by peer.

core/src/sumeragi/main_loop.rs
Erigara: > though it's not entirely clear to me what exactly this means and entails

 This is needed for debugging purposes to detect if we forget to handle message we suppose to handle.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
outoftardis: then
```suggestion
            "No consensus: the peers failed to agreed on the block. Updating view change to try again.",
```

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
outoftardis: maybe this then
```suggestion
                    error!("The hash of the committed block does not match the hash of the block stored by the peer.");
```

core/src/sumeragi/message.rs
```diff
@@ -13,7 +13,10 @@ use iroha_version::prelude::*;
 use parity_scale_codec::{Decode, Encode};
 
 use super::view_change;
-use crate::{block::ValidSignedBlock, VersionedAcceptedTransaction, VersionedCandidateBlock};
+use crate::{
+    block::{ValidSignedBlock, VersionedCandidateCommittedBlock},
```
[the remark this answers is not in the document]
Erigara: Make sense, i will do this.

core/src/block.rs
Erigara: > I believe that additional invariant that `CommittedBlock` offers over `ValidSignedBlock` is that it has enough valid signatures?

Yes, `CommittedBlock` is block accepted by consensus (i.e. have enough transactions). 

> In that case check in this function is not sufficient to produce `CommittedBlock`.

Good point, this goes for `ValidSignedBlock::commit` method too.
Problem is that signatures verification part is logic strongly related to consensus, since we not only counting signatures but also do filtering by role.

What we probably can do is to provide function `verify_signatures<T>(hash: HashOf<T>) -> bool` inside `revalidate_hashes`. This way logic of signatures verification would still be on `sumeragi` side.

mversic: I think what would be best is to give all required arguments to the constructor method of `CommittedBlock`. Is it possible to maybe send reference to `sumeragi` or topology to the block constructor? If not, then I guess this sounds like the 2nd best approach, but you would have to write in the comment of the function all the invariants that need to be upheld when calling the constructor. I dislike the 2nd approach because it's not enforced by the type system but at the same time we cannot mark the function `unsafe`

from the top of my head, a 3rd possibility might be to create some intermediate struct in sumeragi that has some invariants and is then used in the block constructor. Not sure if this is viable

Erigara: Problem with 1st approach is that we hardcoded logic of verification.
And sometimes we need to handle logic a bit different: 
1) With genesis block we want to obtain `CommittedBlock` but we don't bothered by amount of signatures. 
2) We have quite a few places where we use `commit` method inside tests and add `Topology` everywhere would be quite troublesome.

core/src/block.rs
```diff
@@ -405,10 +405,10 @@ pub struct ValidSignedBlock {
     pub event_recommendations: Vec<Event>,
 }
 
-impl ValidSignedBlock {
+impl SignedBlock {
     /// Commit block to the store.
     #[inline]
-    pub fn commit(self) -> CommittedBlock {
+    pub fn commit_unchecked(self) -> CommittedBlock {
```
mversic, line 411: please add a description in the doc which invariants must hold true if this function is used

Erigara: @mversic there is no invariants really because in some cases we just want to obtain `ComittedBlock` bypassing signatures verification.

mversic: I would add a comment saying that when calling this function it is users responsibility to make sure there is enough of valid signatures and to prefer using `commit` instead when possible

Erigara: sure

mversic (approved):

core/src/block.rs
```diff
@@ -989,6 +990,177 @@ impl From<&CommittedBlock> for Vec<Event> {
...
+                eyre!("The block doesn't have enough valid signatures to be committed.")
+            })?;
+        self.revalidate_hashes()
+    }
+
+    /// Revalidate transaction hashes and produce [`CommittedBlock`]
+    ///
+    /// # Errors
+    /// - If transaction hashes don't match the hashes stored in the block header
+    pub fn revalidate_hashes(self) -> Result<CommittedBlock, eyre::Report> {
```
mversic, line 1123: I would also add a comment about favoring the use of `revalidate` and that the signatures are expected to be validated before calling this function

mversic (approved):

SamHSmith (approved):

Issue #2384, opened 2022-06-22
Deprecate `decode()` for  `decode_all()` 


Arjentix:
* Blocked by #2378 

As mentioned in [this comment] we temporary allow users to send extra bytes, but we print warning message. This was done to help SDK-developers.

After everything is stabilized we have to remove this and return an error every time something with extra bytes was sent

appetrosyan:
@arndey @0x009922 , this is an important change, included in RC6.

appetrosyan:
You should test this against the RC.6 peer, and verify. The interpretations can silently diverge.

PR #2990, opened 2022-12-04
[fix] #2384: Replace decode with decode_all.

SamHSmith:

### DIFF
--- a/version/derive/src/lib.rs
+++ b/version/derive/src/lib.rs
@@ -242,38 +242,14 @@ impl Parse for DeclareVersionedArgs {
 fn impl_decode_versioned(enum_name: &Ident) -> proc_macro2::TokenStream {
     quote! (
         impl iroha_version::scale::DecodeVersioned for #enum_name {
-            fn decode_versioned(input: &[u8]) -> iroha_version::error::Result<Self> {
-                use iroha_version::{error::Error, Version, UnsupportedVersion, RawVersioned};
-                use parity_scale_codec::Decode;
-
-                if let Some(version) = input.first() {
-                    if Self::supported_versions().contains(version) {
-                        let mut input = input.clone();
-                        Ok(Self::decode(&mut input)?)
-                    } else {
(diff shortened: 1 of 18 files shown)


ilchu (changes requested):
LGTM, will approve after missing CI fixes.

wasm/src/lib.rs
```diff
@@ -305,7 +305,7 @@ unsafe fn _decode_from_raw_in_range<T: Decode>(
     let bytes = Box::from_raw(core::slice::from_raw_parts_mut(ptr as *mut _, len));
```
mversic, line 305: there are a few places in this module that you should replace `Decode` bound with `DecodeAll` as you did in `smartcontracts/wasm.rs`

version/derive/src/lib.rs
```diff
@@ -242,25 +242,6 @@ impl Parse for DeclareVersionedArgs {
...
-                            *version,
-                            RawVersioned::ScaleBytes(input.to_vec())
-                        ))))
-                    }
-                } else {
-                    Err(Error::NotVersioned)
-                }
-            }
-
             fn decode_all_versioned(input: &[u8]) -> iroha_version::error::Result<Self> {
```
mversic, line 245: why does the implementation of `decode_all_versioned` use `::decode`?

mversic (changes requested):

wasm/src/lib.rs
mversic: also check if `DumpDecoded` in `tools/parity_scale_codec` should be modified

version/derive/src/lib.rs
SamHSmith: I thought I wouldn't change the implementation unnecessarily. But I've changed it to use decode_all now.

mversic (approved):

ilchu (approved):

mversic (approved):

ilchu (approved):

Issue #2993, opened 2022-12-07
Fix FFI serialization of references


mversic:
seems that serializing of references over FFI is incorrect, fix it

Issue #2669, opened 2022-08-29
Manually implement `FfiType` for `ConstString`


Erigara:
Since we have only one case of using union we could instead of making derive macroses even more complicated we could just derive this traits manually.

mversic:
I'm reviving this ticket for further consideration. Previous code is removed [here](https://github.com/hyperledger/iroha/pull/2994/files#r1044272689)

PR #2994, opened 2022-12-07
[feature] #2993: enable all references in FFI

### LINKED ISSUE #2854: serialize `Hash` as transparent in FFI
Looking at the `Hash` in `crypto/hash.rs` it can be seen that it's serialized as a an opaque pointer which is a shame considering that it's underlying type is FFI-safe and it can be transmuted into it. There is some compiler error popping up that should be investigated when `ffi_type(unsafe {robust})` is used

mversic:

### Description of the Change

* Opening this PR as a precursor to the FFI wrapper PR
* rename `NonTransmute` to `Cloned`
* annotate tests with `webassembly_test`
* make `Hash` transparent

### Issue

Closes #2854 #2993

### DIFF
--- a/crypto/src/hash.rs
+++ b/crypto/src/hash.rs
@@ -3,7 +3,6 @@ use alloc::{alloc::alloc, format, string::String, vec, vec::Vec};
 use core::{hash, marker::PhantomData, num::NonZeroU8};
 
 use derive_more::{DebugCustom, Deref, DerefMut, Display};
-use iroha_ffi::FfiType;
 use iroha_schema::prelude::*;
 use parity_scale_codec::{Decode, Encode};
 use serde::{Deserialize, Serialize};
@@ -29,10 +28,7 @@ ffi::ffi_item! {
         Ord,
         PartialOrd,
         IntoSchema,
-        FfiType,
     )]
-    // TODO: use #[ffi_type(unsafe {robust})] instead
-    #[ffi_type(opaque)]
     #[display(fmt = "{}", "hex::encode(self.as_ref())")]
     #[debug(fmt = "{}", "hex::encode(self.as_ref())")]
     #[repr(C)]
@@ -40,6 +36,14 @@ ffi::ffi_item! {
         more_significant_bits: [u8; Self::LENGTH - 1],
         least_significant_byte: NonZeroU8,
     }
+}
+
+// NOTE: Hash is FFI serialized as an array (a pointer in a function call, by value when part of a struct)
+iroha_ffi::ffi_type! {unsafe impl Transparent for Hash[[u8; Hash::LENGTH]] validated with {Hash::is_lsb_1} }
+
+impl iroha_ffi::option::Niche for Hash {
+    // NOTE: Any value that has lsb=0 is a niche value
+    const NICHE_VALUE: Self::ReprC = [0; Hash::LENGTH];
 }
 
 impl Hash {
--- a/ffi/src/repr_c.rs
+++ b/ffi/src/repr_c.rs
@@ -82,23 +82,23 @@ pub trait COutPtr<S>: CType<S> {
 pub unsafe trait NonLocal<S>: COutPtr<S> {}
 
 /// Type that cannot be transmuted into a [`ReprC`] type
-pub trait NonTransmute {}
+pub trait Cloned {}
 
-impl<R: Ir> NonTransmute for &R where R::Type: NonTransmute {}
-impl<R> NonTransmute for Box<R> {}
-impl<R> NonTransmute for &[R] {}
-impl<R> NonTransmute for Vec<R> {}
-impl<R: Ir, const N: usize> NonTransmute for [R; N] where R::Type: NonTransmute {}
+impl<R: Ir> Cloned for &R where R::Type: Cloned {}
+impl<R> Cloned for Box<R> {}
+impl<R> Cloned for &[R] {}
+impl<R> Cloned for Vec<R> {}
+impl<R: Ir, const N: usize> Cloned for [R; N] where R::Type: Cloned {}
 
 // NOTE: `CType` cannot be implemented for `&mut T`
-impl<R: Ir<Type = S> + CType<S>, S: NonTransmute> CType<&S> for &R {
-    type ReprC = *const <R as CType<R::Type>>::ReprC;
+impl<R: CType<S>, S: Cloned> CType<&S> for &R {
+    type ReprC = *const R::ReprC;
 }
-impl<'itm, R: Ir<Type = S> + CTypeConvert<'itm, S, C> + Clone, S: NonTransmute, C: ReprC>
-    CTypeConvert<'itm, &S, *const C> for &'itm R
+impl<'itm, R: CTypeConvert<'itm, S, C> + Clone, S: Cloned, C: ReprC>
+    CTypeConvert<'itm, &'itm S, *const C> for &'itm R
 {
-    type RustStore = (Option<C>, <R as CTypeConvert<'itm, R::Type, C>>::RustStore);
-    type FfiStore = (Option<R>, <R as CTypeConvert<'itm, R::Type, C>>::FfiStore);
+    type RustStore = (Option<C>, R::RustStore);
+    type FfiStore = (Option<R>, R::FfiStore);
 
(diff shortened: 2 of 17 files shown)


ffi/src/repr_c.rs
```diff
@@ -195,10 +195,10 @@ impl<'slice, R: Ir<Type = S> + CTypeConvert<'slice, S, C> + Clone, S: NonTransmu
     }
 }
 
-impl<R: Ir<Type = S> + CType<S>, S: NonTransmute> CType<Vec<S>> for Vec<R> {
```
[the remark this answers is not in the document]
mversic: Because the implementation of `FfiType` requires that both `Ir` and `Ctype` are implemented for some `T`. Because of this `Ir` and `Ctype` don't have to be mutually dependent. This looks simpler while achieving the same

So even if there is no explicit relationship between `Ir` and `CType` they both have to be implemented for a type to derive implementation of `FfiType`

mversic:
there are some difficulties, reverted to draft

primitives/src/conststr.rs
```diff
@@ -469,139 +470,6 @@ impl TryFrom<String> for InlinedString {
     }
 }
 
-mod ffi {
```
mversic, line 472: I have plans to restore this. I removed it because `Account::roles` returns `Vec<&RoleId`> which is translated into `Vec<&ConstString>`. This type cannot be returned from an FFI function at the moment because it would pass on ownership. In the #2978 I am introducing limited ownership transfer because it was too difficult (in some situations even impossible) to generate a wrapper API if there is no ownership transfer.

I maintain that this is not a particularly important piece of code and can be restored

mversic, line 472: when FFI function is generated for `Account::roles` it returns `Vec<&RoleId>` which ends up translated into `Vec<&ConstString>`. In the current implementation `Vec<T>` cannot be returned from an FFI function if T carries ownership

I'm reviving the #2669. It's too bothersome to deal with this now

crypto/src/hash.rs
```diff
@@ -36,11 +36,11 @@ ffi::ffi_item! {
...
     #[repr(transparent)]
     #[serde(transparent)]
-    // TODO: use #[ffi_type(unsafe {robust})] instead
-    #[ffi_type(opaque)]
+    // SAFETY: Hash has no trap representations in [u8; LENGTH]
+    #[ffi_type(unsafe {robust})]
     #[display(fmt = "{}", "hex::encode(_0)")]
     #[debug(fmt = "{{ Hash({}) }}", "hex::encode(_0)")]
-    pub struct Hash([u8; Self::LENGTH]);
+    pub struct Hash([u8; Hash::LENGTH]);
```
mversic, line 43: it would be nice if we could use zero hash for trap representation but we allow all zero hash as a signal

Erigara (approved):

crypto/src/hash.rs
mversic: yes, though it's not present yet, any method on `SignatureOf` that takes `HashOf`, etc.

Erigara: > any method on `SignatureOf` that takes `HashOf`, etc.

But this method takes `HashOf<T>` not `Option<HashOf<T>>`, or i miss something?

mversic: yes, but `HashOf` will be transmuted into `Hash` which will be transmuted into `[u8; u32]`. The function will end up ingesting `[u8; u32]`

Erigara: and what benefits it would give us if all zeroes would be considered niche value? 
i thought we use it only in combination with `Option`

mversic: you are correct, `Option` is required, this wasn't a good example

ffi/src/repr_c.rs
```diff
@@ -82,76 +82,76 @@ pub trait COutPtr<S>: CType<S> {
...
 /// Type that cannot be transmuted into a [`ReprC`] type
-pub trait NonTransmute {}
+pub trait Cloned {}
 
-impl<R: Ir> NonTransmute for &R where R::Type: NonTransmute {}
-impl<R> NonTransmute for Box<R> {}
-impl<R> NonTransmute for &[R] {}
-impl<R> NonTransmute for Vec<R> {}
-impl<R: Ir, const N: usize> NonTransmute for [R; N] where R::Type: NonTransmute {}
+impl<R: Ir> Cloned for &R where R::Type: Cloned {}
```
appetrosyan, line 87: Like the new naming convention.

ffi/src/repr_c.rs
```diff
@@ -82,76 +82,76 @@ pub trait COutPtr<S>: CType<S> {
...
+    CTypeConvert<'itm, &'itm S, *const C> for &'itm R
 {
-    type RustStore = (Option<C>, <R as CTypeConvert<'itm, R::Type, C>>::RustStore);
-    type FfiStore = (Option<R>, <R as CTypeConvert<'itm, R::Type, C>>::FfiStore);
+    type RustStore = (Option<C>, R::RustStore);
+    type FfiStore = (Option<R>, R::FfiStore);
 
-    fn into_repr_c(self, store: &'itm mut Self::RustStore) -> *const C {
-        store.0.insert(self.clone().into_repr_c(&mut store.1))
+    fn lower(self, store: &'itm mut Self::RustStore) -> *const C {
```
appetrosyan, line 103: This is a little less clear than `into_repr_c`

appetrosyan (dismissed):

ffi/src/repr_c.rs
mversic: I agree

primitives/src/conststr.rs
mversic: if you take `Domain::asset_definitions` this function when converted to an FFI function will obviously return a `Vec<T>`. On the other hand, making an entire `Vec<T>` opaque would incur an FFI call for every access. I grant that you could use some the predefined queries like `FindAllAssetDefinitions` or similar, which seems inconvenient.

ffi/src/std_impls.rs
```diff
@@ -20,7 +20,7 @@ ffi_type! {unsafe impl<'slice> Transparent for &'slice str[&'slice [u8]] validat
 #[cfg(feature = "non_robust_ref_mut")]
 ffi_type! {unsafe impl<'slice> Transparent for &'slice mut str[&'slice mut [u8]] validated with {|target| core::str::from_utf8(target).is_ok()} }
 ffi_type! {unsafe impl<T> Transparent for core::mem::ManuallyDrop<T>[T] validated with {|_| true} }
-ffi_type! {unsafe impl<T> Transparent for core::ptr::NonNull<T>[*mut T] validated with {|target: &*mut T| target.is_null()} }
+ffi_type! {unsafe impl<T> Transparent for core::ptr::NonNull<T>[*mut T] validated with {|target: &*mut T| !target.is_null()} }
```
appetrosyan, line 23: This, while subtle, should have been caught by tests. The fact that we don't have any, is a little worrying.

crypto/src/hash.rs
```diff
@@ -42,6 +38,8 @@ ffi::ffi_item! {
     }
 }
 
+iroha_ffi::ffi_type! {unsafe impl Transparent for Hash[[u8; Hash::LENGTH]] validated with {|target| Hash::is_lsb_1(target)} }
```
mversic, line 41: it will be FFI serialized as an array. Which means as a pointer in a function call, or by value when part of a struct

appetrosyan (dismissed):

primitives/src/conststr.rs
appetrosyan: We can always change it later. I wonder how much binary size we'd be trading, and if we can leverage this to reduce the size even further.

crypto/src/hash.rs
appetrosyan: I think this ought to be a comment in the text.

appetrosyan (approved):

ffi/src/std_impls.rs
mversic: I concur, but I don't have the bandwidth to test every type. I would be interested to see if OpenAI can help writing these boilerplate tests. I expect it would do an admirable job.

it's not so bad, as long as it's not UB  ...

Erigara (approved):

PR #2996, opened 2022-12-07
[feature] #2995: Soft fork detection

### LINKED ISSUE #2995: [suggestion]: Detect soft forks
### Feature request

I'd like iroha2 to be able to detect soft fork condition.

### Motivation

Without soft fork detection it's hard to debug inconsistent blockchain state.

### Who can help?

@Erigara

Erigara:
### Description of the Change

Add mechanism to detect soft forks during block sync.

### Issue

Closes #2995.

### Benefits

Easier to debug problems with inconsistent blockchain state.

### Possible Drawbacks

Added complexity to sumeragi.

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -81,9 +81,13 @@ impl Chain {
         ChainIterator::new(self)
     }
 
-    /// Latest block reference and its height.
-    pub fn latest_block(&self) -> Option<MapRef<u64, VersionedCommittedBlock>> {
-        self.blocks.get(&(self.blocks.len() as u64))
+    /// Get nth block from the end of the [`Chain`]
+    pub fn nth_back(&self, n: usize) -> Option<MapRef<u64, VersionedCommittedBlock>> {
+        self.blocks
+            .len()
+            .gt(&n)
+            .then_some(())
+            .and_then(|_| self.blocks.get(&((self.blocks.len() - n) as u64)))
     }
 
     /// Length of the blockchain.
@@ -211,6 +215,7 @@ impl PendingBlock {
         self,
         height: u64,
         previous_block_hash: HashOf<VersionedCommittedBlock>,
+        view_change_index: u64,
     ) -> ChainedBlock {
         ChainedBlock {
             transactions: self.transactions,
@@ -219,6 +224,7 @@ impl PendingBlock {
                 timestamp: self.timestamp,
                 consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
                 height: height + 1,
+                view_change_index,
                 previous_block_hash,
                 transactions_hash: Hash::zeroed().typed(),
                 rejected_transactions_hash: Hash::zeroed().typed(),
@@ -236,6 +242,7 @@ impl PendingBlock {
                 timestamp: self.timestamp,
                 consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
                 height: 1,
+                view_change_index: 0,
                 previous_block_hash: EmptyChainHash::default().into(),
                 transactions_hash: Hash::zeroed().typed(),
                 rejected_transactions_hash: Hash::zeroed().typed(),
@@ -253,6 +260,7 @@ impl PendingBlock {
                 timestamp: self.timestamp,
                 consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
                 height: 1,
+                view_change_index: 0,
                 previous_block_hash: EmptyChainHash::default().into(),
                 transactions_hash: Hash::zeroed().typed(),
                 rejected_transactions_hash: Hash::zeroed().typed(),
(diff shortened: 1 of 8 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -699,113 +799,86 @@ pub fn run<F>(
...
-                    }
+
+                    let block_clone = block.clone();
+                    let key_pair_clone = sumeragi.key_pair.clone();
+                    let signed_block = block_clone
+                        .sign(key_pair_clone)
+                        .expect("maybe we should handle this error");
+
+                    let voting_block = VotingBlock::new(signed_block.clone());
+                    voting_block_option = Some(voting_block);
```
SamHSmith, line 855: I presume this is for your bandwidth saving PR? Making sure observing peers have a copy of the block so that the proxy tail message can be changed to only include the hash of the block?

core/src/sumeragi/main_loop.rs
```diff
@@ -889,229 +964,171 @@ pub fn run<F>(
...
 
-                        for sig in valid_signatures {
-                            block_signature_acc.push((hash, sig));
-                        }
+                    instant_at_which_we_should_have_committed =
+                        Instant::now() + sumeragi.commit_time;
+                }
+                Some(Message::BlockSigned(BlockSigned { hash, signatures })) => {
+                    if voting_block_option.is_none()
```
SamHSmith, line 1107: Networks are weird. There is a case where we receive the vote before the block. Ideally we should therefore have single "pending vote" slot per foreign peer. The vote gets stored there until we receive the voting block and then we can check it as usual.

It adds a bit of complexity but it is worth it for good runtime behaviour, especially in a public blockchain.

core/src/sumeragi/main_loop.rs
```diff
@@ -1294,50 +1311,63 @@ where
...
+                    info!("Using network topology from genesis block");
+                    state.current_topology = block
+                        .header()
+                        .genesis_topology
+                        .clone()
+                        .take()
+                        .expect("We just checked that it is some");
+                    commit_block(sumeragi, block, state);
+                    info!("Genesis block received and committed.");
+                    return Err(EarlyReturn::GenesisBlockReceivedAndCommitted);
```
SamHSmith, line 1367: Should the success case really be an error?

SamHSmith (approved):
Well done, we can merge this as is, or you can include the vote counting special case I mentioned.

I recommend we merge this and then you can do the vote thing as a seperate ticket.

core/src/sumeragi/main_loop.rs
Erigara: Really good point! Have created #2999.

core/src/sumeragi/main_loop.rs
Erigara: yeap

core/src/sumeragi/main_loop.rs
Erigara: I think this was done initially to represent all ways we can exit from this function (since `()` in not reachable).
I agree that wrapping success case in `Err` looks a bit confusing.

core/src/block.rs
```diff
@@ -81,9 +81,10 @@ impl Chain {
         ChainIterator::new(self)
     }
 
-    /// Latest block reference and its height.
-    pub fn latest_block(&self) -> Option<MapRef<u64, VersionedCommittedBlock>> {
-        self.blocks.get(&(self.blocks.len() as u64))
+    /// Get block nth from the end of the [`Chain`]
+    pub fn latest_nth_block(&self, n: usize) -> Option<MapRef<u64, VersionedCommittedBlock>> {
+        self.blocks
+            .get(&(self.blocks.len().saturating_sub(n) as u64))
```
QuentinI, line 87: I think `saturating_sub` behaviour is non-obvious and should be either documented, or changed to return None when n > len

QuentinI: Maybe also rename `latest_nth_block` and friends to `nth_back` by analogue to `std::iter::DoubleEndedIterator`?

core/src/sumeragi/mod.rs
```diff
@@ -226,6 +227,12 @@ impl Sumeragi {
         *self.internal.latest_block_hash.lock()
     }
 
+    /// Get previous to latest block hash for use by the block synchronization subsystem.
+    #[allow(clippy::expect_used)]
```
QuentinI, line 231: Is it used though?

QuentinI (approved):
LGTM with some nitpicks

core/src/block.rs
Erigara: > `nth_back`
Thanks, sounds much better! 

> I think saturating_sub behaviour is non-obvious and should be either documented, or changed to return None when n > len

I will add docs

core/src/sumeragi/mod.rs
Erigara: not used indeed

SamHSmith (dismissed):

core/src/block.rs
QuentinI: > Will return `None`

Oh, right, didn't realize block height starts from 1.

QuentinI (dismissed):

QuentinI (approved):

SamHSmith (approved):

Issue #2675, opened 2022-08-30
Validate item names/ids without allocating `Vec`


Erigara:
We have a few places where we allocate a vector within the `FromStr` implementation for an element name/id when we don't really need to. 

Approach without allocation: data_model/src/trigger/mod.rs impl FromStr for Id.

appetrosyan:
https://docs.rs/zerocopy/latest/zerocopy/ Might be useful

Retssaze:
Where this data_model/src/trigger/mod.rs  is located?

appetrosyan:
Hi! This is an iroha version 2 issue, so you need to switch to the iroha2-dev branch.

Retssaze:
Let's cath up

suiwenfeng:
@appetrosyan `zerocopy` make sense on packat deserialization from network bytes, but how can it do validation on invalid input? for example, no `@` in account_id(which should have format like `name@domain_name`).

after look at `data_model/src/trigger/mod.rs`, just replace

``` rust
let vector: Vec<&str> = string.split('@').collect();
```

with 
``` rust
let mut split = s.split('$');
match (split.next(), split.next(), split.next()) {
```

That`s right?

suiwenfeng:
Ok, i will refactor such code. Method execution time did improved in my local benchmark.

<img width="945" alt="image" src="https://user-images.githubusercontent.com/106252855/206404485-fc0d2bcb-116e-45c1-849f-a17886e70b03.png">

Erigara:
Great to hear! Will be waiting for the PR.

suiwenfeng:
#3003

Retssaze:
> #3003

You took my task, what a shame!

6r1d:
While I am not sure it helps, there's a [list](https://github.com/hyperledger/iroha/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22) of several issues that may be interesting for the new contributors.

Erigara:
For example you can take a look at #2123.

Retssaze:
To [suiwenfeng](https://github.com/suiwenfeng): Sir, For sure you are very strong and quick, and a message above by [6r1d](https://github.com/6r1d) is definitely not only for you!

PR #3003, opened 2022-12-09
[refactor] Validate item names/ids without allocating Vec

suiwenfeng:
Signed-off-by: Wenfeng Sui <suiwenfeng@qq.com>

### DIFF
--- a/config/src/client.rs
+++ b/config/src/client.rs
@@ -144,9 +144,9 @@ impl ConfigurationProxy {
             let api_url = api_url.clone().to_string();
             let split = api_url.rsplit_once("://");
             match split {
-                Some((protocal, _)) => {
+                Some((protocol, _)) => {
                     // TODO: this is neither robust, nor useful. This should be enforced as a `FromStr` implementation.
-                    if protocal != "http" {
+                    if protocol != "http" {
                         eyre::bail!(ConfigError::ProxyBuildError(
                             "`TORII_API_URL` string: `{api_url}` only supports the `HTTP` protocol currently".to_owned()
                         ));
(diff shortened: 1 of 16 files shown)


config/src/client.rs
```diff
@@ -142,37 +142,45 @@ impl ConfigurationProxy {
...
-                ));
-            }
-            // TODO: this is neither robust, nor useful. This should be enforced as a `FromStr` implementation.
-            if split_api_url[0] != "http" {
-                eyre::bail!(ConfigError::ProxyBuildError(
-                    "`TORII_API_URL` string: `{api_url}` only supports the `HTTP` protocol currently".to_owned()
-                ));
+            let split = api_url.rsplit_once("://");
+            match split {
+                Some((protocal, _)) => {
```
6r1d, line 147: `protocal`, not a "protocol"? Same applies at 149, 167, 168

data_model/tests/data_model.rs
```diff
@@ -302,3 +302,18 @@ fn cannot_forbid_minting_on_asset_mintable_infinitely() -> Result<(), ParseError
     assert!(definition.forbid_minting().is_err());
     Ok(())
 }
+
+#[test]
+fn account_id_parsing() -> Result<(), ParseError> {
+    // `AccountId` should have format `name@domain_name`
+    let account_normal: AccountId = "test@hello".parse()?;
+    assert_eq!(account_normal.name.to_string(), String::from("test"));
```
Erigara, line 310: I would suggest to do the following: 
```suggestion
    assert_eq!(account_normal.name.as_ref(), "test");
```

data_model/tests/data_model.rs
```diff
@@ -302,3 +302,18 @@ fn cannot_forbid_minting_on_asset_mintable_infinitely() -> Result<(), ParseError
...
+    assert_eq!(account_normal.domain_id.name.to_string(), String::from("hello"));
+
+    let account_empty: Result<AccountId, _> = "@hello".parse();
+    assert!(account_empty.is_err());
+
+    let account_invalid: Result<AccountId, _> = "testhello".parse();
+    assert!(account_invalid.is_err());
```
Erigara, line 317: Wight be redundant but we can also test for string `"@"`.

Erigara (commented):
Good job!
Here few minor things you should do: 
1. We use standardize format for commit names, so you commit message should look like: `[refactor] #2675: Validate item names/ids without allocating Vec`.
2. Also we use `rustfmt` in our workflow , you can run it by following command: `cargo +nightly fmt --all`

data_model/tests/data_model.rs
suiwenfeng: i reopened a pr for some reason, you have to check this at #3004 now.

PR #3012, opened 2022-12-14
[refactor] #3008: Add niche value into `Hash`

### LINKED ISSUE #3008: [suggestion]: Add niche to `Hash`
### Feature request

Add niche for `Hash` memory representation.

### Motivation

This way we can get rid of using all zero hash as null value and handle such logic as `Option<Hash>`.

### Who can help?

@appetrosyan @Erigara

Erigara:

### Description of the Change

In order to achieve niche value optimization least significant bit of hash set to be always 1.  

### Issue

Closes #3008.

### Benefits

Don't use all zero hash as null value, instead use `Option<Hash>` without loses in size. 

### Possible Drawbacks

- Unsafe code
- Least significant bit of hash is always 1, which can cause compatibility issues with `blake2b` hashing algorithm. 
- Can't implement `Display` for `Option<Hash>`

### DIFF
--- a/core/src/wsv.rs
+++ b/core/src/wsv.rs
@@ -413,19 +413,20 @@ impl WorldStateView {
     /// Return a vector of blockchain blocks after the block with the given `hash`
     pub fn block_hashes_after_hash(
         &self,
-        hash: HashOf<VersionedCommittedBlock>,
+        hash: Option<HashOf<VersionedCommittedBlock>>,
     ) -> Vec<HashOf<VersionedCommittedBlock>> {
-        if hash == Hash::zeroed().typed() {
-            self.block_hashes.borrow().clone()
-        } else {
-            self.block_hashes
-                .borrow()
-                .iter()
-                .skip_while(|&x| *x != hash)
-                .skip(1)
-                .copied()
-                .collect()
-        }
+        hash.map_or_else(
+            || self.block_hashes.borrow().clone(),
+            |block_hash| {
+                self.block_hashes
+                    .borrow()
+                    .iter()
+                    .skip_while(|&x| *x != block_hash)
+                    .skip(1)
+                    .copied()
+                    .collect()
+            },
+        )
     }
 
     /// The same as [`Self::modify_world_multiple_events`] except closure `f` returns a single [`WorldEvent`].
@@ -638,29 +639,21 @@ impl WorldStateView {
         }
     }
 
-    /// Hash of latest block
-    pub fn latest_block_hash(&self) -> HashOf<VersionedCommittedBlock> {
-        self.block_hashes
-            .borrow()
-            .iter()
-            .last()
-            .map_or(Hash::zeroed().typed(), |hash| *hash)
+    /// Return the hash of the latest block
+    pub fn latest_block_hash(&self) -> Option<HashOf<VersionedCommittedBlock>> {
+        self.block_hashes.borrow().iter().nth_back(0).copied()
     }
 
-    /// View change index of nth back block
+    /// Return the view change index of the latest block
     pub fn latest_block_view_change_index(&self) -> u64 {
         self.kura
             .get_block_by_height(self.height())
             .map_or(0, |block| block.header().view_change_index)
     }
 
-    /// Hash of latest block
-    pub fn previous_block_hash(&self) -> HashOf<VersionedCommittedBlock> {
-        self.block_hashes
-            .borrow()
-            .iter()
-            .nth_back(1)
-            .map_or(Hash::zeroed().typed(), |hash| *hash)
+    /// Return the hash of the block one before the latest block
+    pub fn previous_block_hash(&self) -> Option<HashOf<VersionedCommittedBlock>> {
+        self.block_hashes.borrow().iter().nth_back(1).copied()
     }
 
     /// Get `Account` and pass it to closure.
@@ -1101,16 +1094,14 @@ mod tests {
         let mut block_hashes = vec![];
         for i in 1..=BLOCK_CNT {
             block.header.height = i as u64;
-            if let Some(block_hash) = block_hashes.last() {
-                block.header.previous_block_hash = *block_hash;
-            }
+            block.header.previous_block_hash = block_hashes.last().copied();
             let block: VersionedCommittedBlock = block.clone().into();
             block_hashes.push(block.hash());
             wsv.apply(&block).unwrap();
         }
 
         assert!(wsv
-            .block_hashes_after_hash(block_hashes[6])
+            .block_hashes_after_hash(Some(block_hashes[6]))
             .into_iter()
             .eq(block_hashes.into_iter().skip(7)));
     }
--- a/crypto/src/hash.rs
+++ b/crypto/src/hash.rs
@@ -1,6 +1,6 @@
 #[cfg(not(feature = "std"))]
 use alloc::{alloc::alloc, format, string::String, vec, vec::Vec};
-use core::{hash, marker::PhantomData};
+use core::{hash, marker::PhantomData, num::NonZeroU8};
 
 use derive_more::{DebugCustom, Deref, DerefMut, Display};
 use iroha_ffi::FfiType;
@@ -17,6 +17,7 @@ use crate::ffi;
 
 ffi::ffi_item! {
     /// Hash of Iroha entities. Currently supports only blake2b-32.
+    /// Least significant bit of hash is set to 1.
     #[derive(
         Clone,
         Copy,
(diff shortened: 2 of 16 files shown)


data_model/src/block_value.rs
```diff
@@ -24,19 +24,19 @@ declare_item! {
     #[derive(
         Debug, Clone, Display, PartialEq, Eq, Hash, Decode, Encode, Deserialize, Serialize, FfiType, IntoSchema,
     )]
-    #[display(fmt = "Block №{height} (hash: {transactions_hash});")]
+    #[display(fmt = "Block №{height} (hash: {transactions_hash:?});")]
```
appetrosyan, line 27: Given that this is the `Display` implementation, it could be worthwhile to use `unwrap_or("empty")` in this position. In most cases it's `Some`, so we're just making the implementation noisy.

crypto/src/hash.rs
```diff
@@ -27,36 +28,34 @@ ffi::ffi_item! {
...
     )]
-    #[repr(transparent)]
-    #[serde(transparent)]
     // TODO: use #[ffi_type(unsafe {robust})] instead
     #[ffi_type(opaque)]
-    #[display(fmt = "{}", "hex::encode(_0)")]
-    #[debug(fmt = "{{ Hash({}) }}", "hex::encode(_0)")]
-    pub struct Hash([u8; Self::LENGTH]);
+    #[display(fmt = "{}", "hex::encode(self.as_ref())")]
+    #[debug(fmt = "{{ Hash({}) }}", "hex::encode(self.as_ref())")]
```
appetrosyan, line 37: Make the two identical, or derive the default `fmt::Debug`.

crypto/src/hash.rs
```diff
@@ -27,36 +28,34 @@ ffi::ffi_item! {
...
+    #[repr(C)]
+    pub struct Hash {
+        bytes: [u8; Self::LENGTH - 1],
+        byte: NonZeroU8,
```
appetrosyan, line 41: These names are tolerable. But not the best. Given that the things are completely internal I'd propose calling them `more_significant_bits`, and `least_significant_byte`.

crypto/src/hash.rs
```diff
@@ -84,15 +83,53 @@ impl Hash {
...
+    where
+        S: serde::Serializer,
+    {
+        let hash: &[u8; Self::LENGTH] = self.as_ref();
+        hash.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for Hash {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
```
appetrosyan, line 117: This is incredibly dangerous. What if we set the LSB to 0 by accident, and the remaining hash is non-zero.

appetrosyan (changes requested):

crypto/src/hash.rs
Erigara: We would silently set LSB to 1 inside `prehashed` function, however i agree that we should explicitly return error if hash is not valid.

core/src/wsv.rs
```diff
@@ -654,13 +651,9 @@ impl WorldStateView {
...
     }
 
-    /// Hash of latest block
-    pub fn previous_block_hash(&self) -> HashOf<VersionedCommittedBlock> {
-        self.block_hashes
-            .borrow()
-            .iter()
-            .nth_back(1)
-            .map_or(Hash::zeroed().typed(), |hash| *hash)
+    /// Hash of previous to latest block
```
outoftardis, line 654:
```suggestion
    /// Return the hash of the block one before the latest block
```

data_model/src/block_value.rs
Erigara: Have to fallback to manual implementation of `Display`.

Erigara: Actually i think that we have buggy implementation of `Display`, we would like to print `Hash` of Block not hash of accepted transactions.

appetrosyan (dismissed):

crypto/src/hash.rs
```diff
@@ -27,36 +28,34 @@ ffi::ffi_item! {
...
     #[ffi_type(opaque)]
-    #[display(fmt = "{}", "hex::encode(_0)")]
-    #[debug(fmt = "{{ Hash({}) }}", "hex::encode(_0)")]
-    pub struct Hash([u8; Self::LENGTH]);
+    #[display(fmt = "{}", "hex::encode(self.as_ref())")]
+    #[debug(fmt = "{}", "hex::encode(self.as_ref())")]
+    #[repr(C)]
+    pub struct Hash {
+        more_significant_bits: [u8; Self::LENGTH - 1],
+        least_significant_byte: NonZeroU8,
```
mversic, line 41: are we guaranteed Rust will use niche value optimization in this case? I presume you verified, but is there some doc you can refer to?

crypto/src/hash.rs
```diff
@@ -80,19 +79,75 @@ impl Hash {
     pub const fn typed<T>(self) -> HashOf<T> {
         HashOf(self, PhantomData)
     }
+
+    /// Check if least significant bit of `[u8; Hash::LENGTH]` is 1
+    fn is_lsb_1(hash: &[u8; Self::LENGTH]) -> bool {
```
mversic, line 84: so hash will always be an odd number? how does this affect the hash distribution?

crypto/src/hash.rs
mversic: Distribution should remain uniform, but we're halving the space of all possible hashes. Does that have any impact?

mversic: Also, have you considered using `NonZeroU32` as inner repr instead?

crypto/src/hash.rs
appetrosyan: Most hash functions are designed to work with such pathologies, if we used it as a hash table entry (which we wouldn’t). The Merkle proof makes no assumptions about the parity, so it should work with odd (pun intended) hashes. 

For our purposes it’s more important to think about the hash space than hash collisions. All we sacrifice in this case, however is one single bit.

appetrosyan: If we really wanted to be pedantic, we would right shift the hash before use, but to me it seems superfluous.

mversic: > Most hash functions are designed to work with such pathologies, if we used it as a hash table entry (which we wouldn’t). The Merkle proof makes no assumptions about the parity, so it should work with odd (pun intended) hashes. 

I agree, but thought I better bring it up

> If we really wanted to be pedantic, we would right shift the hash before use, but to me it seems superfluous. 

I don't see how that would make any meaningful difference

crypto/src/hash.rs
Erigara: > I presume you verified, but is there some doc you can refer to?

I have tested this on the playground  ([link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=030c72bc484c66e843e5b4a04f27c59b)).

Beyond that number of cases is covered [by the compiler](https://github.com/rust-lang/rust/blob/master/src/test/ui/structs-enums/type-sizes.rs.)

crypto/src/hash.rs
Erigara: I think it would be uniform distributed across in odd numbers.

crypto/src/hash.rs
Erigara: > Also, have you considered using NonZeroU32 as inner repr instead?

Not really, what it would give us?

mversic: > > Also, have you considered using NonZeroU32 as inner repr instead?
> 
> Not really, what it would give us? 

Hash space wouldn't be cut in half, function calling convention would be that of an integer, sdk devs wouldn't have to make any modifications. These may not be important, just exploring the space of ideas

Erigara: > Hash space wouldn't be cut in half 

But there still would be possibility to face all zero u32, right? (very small one).

Erigara: Another thing to consider is that `(U32, NonZeroU32)` would have different alignment from `[u8;32]`.
However i thing it's not an issue, because `transmute` is unaffected by alignment: 
> Because transmute is a by-value operation, alignment of the transmuted values themselves is not a concern. As with any other function, the compiler already ensures both T and U are properly aligned. However, when transmuting values that point elsewhere (such as pointers, references, boxes…), the caller has to ensure proper alignment of the pointed-to values.

And when we do pointer casts we would cast value with higher alignment to value with lower alignment, so it should be fine.

SamHSmith (dismissed):
Looks good.

crypto/src/hash.rs
mversic: alignment is an interesting point to be made

> But there still would be possibility to face all zero u32, right? (very small one).

oh, you think that the hash function has spit out all-zero hash. I had a feeling this is guaranteed not to happen, I might have been wrong there

mversic (approved):

mversic (approved):

appetrosyan (approved):

PR #3028, opened 2022-12-20
[feature] #2813: Kagami generate default peer config

### LINKED ISSUE #2813: [ci]: Add a way to check peer configuration.json
Follow up task to #2778.

Erigara:

### Description of the Change

- Extend `kagami` to support peer configuration generation
- Add CI check for validity of peer configuration
 

### Issue

Closes #2813.

### Benefits

Automation and more checks.

### Possible Drawbacks

None.

```bash
cargo run --bin kagami -- config peer
```


configs/peer/config.json
```diff
@@ -37,7 +42,7 @@
     "ACTOR_CHANNEL_CAPACITY": 100
   },
   "QUEUE": {
-    "MAXIMUM_TRANSACTIONS_IN_BLOCK": 8192,
+    "MAXIMUM_TRANSACTIONS_IN_BLOCK": 512,
```
ilchu, line 45: Great that now the file will be kept in sync. IIRC this queue change was merged quite some time ago.

configs/peer/config.json
```diff
@@ -1,6 +1,9 @@
 {
-  "PUBLIC_KEY": null,
-  "PRIVATE_KEY": null,
+  "PUBLIC_KEY": "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0",
+  "PRIVATE_KEY": {
+    "digest_function": "ed25519",
+    "payload": "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
+  },
```
ilchu, line 6: Do you think it's better this way? My understanding was that it's obviously OK for test environments, but when trying to spin up a peer, we want to refrain from giving defaults, instead prompting the user to generate the key and provide it directly. More so for the keys, less so for `torii` and trusted peers, but still relevant.

tools/kagami/src/main.rs
```diff
@@ -363,43 +363,120 @@ mod genesis {
...
+                    )?),
+                    private_key: Some(PrivateKey::from_hex(
+                        Algorithm::Ed25519,
+                        "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
+                    )?),
+                    ..ConfigurationProxy::default()
+                }
+                .build()?;
+                writeln!(writer, "{}", serde_json::to_string_pretty(&config)?)
+                    .wrap_err("Failed to write.")
```
ilchu, line 428:
```suggestion
.wrap_err("Failed to write serialized config to the buffer.")
```
And maybe in a few other places throughout the file too.

ilchu (changes requested):

tools/kagami/src/main.rs
```diff
@@ -363,43 +363,120 @@ mod genesis {
...
+                            Algorithm::Ed25519,
+                            "038ae16b219da35aa036335ed0a43c28a2cc737150112c78a7b8034b9d99c9023f4e3e98571b55514edc5ccf7e53ca7509d89b2868e62921180a6f57c2f4e255"
+                        )?)),
+                        ..GenesisConfigurationProxy::default()
+                    }.into(),
+                    ..IrohaConfigurationProxy::default()
+                }
+                .build()?;
+                writeln!(writer, "{}", serde_json::to_string_pretty(&config)?)
+                    .wrap_err("Failed to write.")
```
Arjentix, line 478: Same as @ilchu has suggested in his [comment](https://github.com/hyperledger/iroha/pull/3028/files#r1055093571)

tools/kagami/src/main.rs
```diff
@@ -363,43 +363,120 @@ mod genesis {
...
+
+        impl<T: Write> RunArgs<T> for Args {
+            fn run(self, writer: &mut BufWriter<T>) -> Outcome {
+                let config = IrohaConfigurationProxy {
+                    public_key: Some(PublicKey::from_str(
+                        "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0",
+                    )?),
+                    private_key: Some(PrivateKey::from_hex(
+                        Algorithm::Ed25519,
+                        "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
```
Arjentix, line 456: Maybe we should remove this from default configuration so that real users won't use this as a real key?

outoftardis (commented):
kagami readme should be updated as well, right?

cli/src/lib.rs
```diff
@@ -256,7 +256,7 @@ impl Iroha {
         );
 
         let kura = Kura::from_configuration(&config.kura)?;
-        let wsv = WorldStateView::from_configuration(config.wsv, world);
+        let wsv = WorldStateView::from_configuration(config.wsv, world, Arc::clone(&kura));
```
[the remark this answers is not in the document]
Erigara: Yes, i suppose it's Sam's PR about blocks on kura.

tools/kagami/src/main.rs
Erigara: Good suggestion, thanks.

tools/kagami/src/main.rs
Erigara: Yeah, it probably be better to set `null` for then.

configs/peer/config.json
Erigara: Agree with you both, will try serialize proxy config.

Erigara:
> kagami readme should be updated as well, right?

@outoftardis i have updated kagami README could you please check it?

tools/kagami/README.md
```diff
@@ -142,3 +143,17 @@ The output should be identical to the [reference configuration](../../docs/sourc
...
+- Generate the default peer configuration:
+
+    ```bash
+    kagami config peer > peer-config.json
+    ```
+
+- Generate the default client configuration:
+
+    ```bash
+    kagami tokens > client-config.json
```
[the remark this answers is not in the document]
Erigara: yeah, thanks!

ilchu (approved):

configs/peer/config.json
```diff
@@ -25,9 +24,9 @@
     "GOSSIP_PERIOD_MS": 1000
   },
   "TORII": {
-    "P2P_ADDR": null,
-    "API_URL": null,
-    "TELEMETRY_URL": null,
+    "P2P_ADDR": "http://127.0.0.1:1337",
+    "API_URL": "http://127.0.0.1:8080",
+    "TELEMETRY_URL": "http://127.0.0.1:8180",
```
Arjentix, line 29: What about addressed? Do you think we need to keep them here? Or maybe better to keep them as nulls?

tools/kagami/src/main.rs
```diff
@@ -363,43 +363,103 @@ mod genesis {
...
+                    basic_auth: Some(Some(BasicAuth {
+                        web_login: WebLogin::new("mad_hatter")?,
+                        password: SmallStr::from_str("ilovetea"),
+                    })),
+                    public_key: Some(PublicKey::from_str(
+                        "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0",
+                    )?),
+                    private_key: Some(PrivateKey::from_hex(
+                        Algorithm::Ed25519,
+                        "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
```
Arjentix, line 422: If this will be `null` in the resulting config, maybe you can use `None` here?

Erigara: It won't be `null` because it's configuration for `client`, not `peer`.

configs/peer/config.json
Erigara: Yeah, lets keep them `null`, since this addresses would be replaced anyway to setup network

ilchu (approved):

Arjentix (approved):

ilchu (approved):

PR #3030, opened 2022-12-20
[feature] #3019: Add JSON5 support

Arjentix:
### Description of the Change

Added JSON5 support in user-provided configs, such as:

- config and genesis for `iroha`
- config, metadata, signature, instructions for `iroha_client_cli`

### Issue

- #3019 

### Benefits

Better human-reading format

### Possible Drawbacks

`json5` crate I use has lack of some functions and features unlike `serde_json`. But it should be enough to interact with user properly.

### Usage Examples or Tests

You can try running `iroha` or `iroha_client_cli` as always but also you can:
- use `json5` file extension (for default and your own configs)
- use all features of `json5` in your configs (which should not necessarily have `json5` extension), i.e. comments

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -9,9 +9,10 @@
     clippy::std_instead_of_core,
     clippy::std_instead_of_alloc
 )]
-use std::{panic, path::PathBuf, sync::Arc};
+use std::{borrow::Cow, convert::Infallible, panic, path::PathBuf, str::FromStr, sync::Arc};
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
+use eyre::ContextCompat;
 use iroha_actor::{broker::*, prelude::*};
 use iroha_config::{
     base::proxy::{LoadFromDisk, LoadFromEnv, Override},
@@ -50,22 +51,97 @@ pub struct Arguments {
     /// Set this flag on the peer that should submit genesis on the network initial start.
     pub submit_genesis: bool,
     /// Set custom genesis file path. `None` if `submit_genesis` set to `false`.
-    pub genesis_path: Option<PathBuf>,
+    pub genesis_path: Option<ConfigPath>,
     /// Set custom config file path.
-    pub config_path: PathBuf,
+    pub config_path: ConfigPath,
 }
 
-const CONFIGURATION_PATH: &str = "config.json";
-const GENESIS_PATH: &str = "genesis.json";
+const CONFIGURATION_PATH: &str = "config";
+const GENESIS_PATH: &str = "genesis";
+const POSSIBLE_CONFIG_EXTENSIONS: [&str; 2] = ["json", "json5"];
 const SUBMIT_GENESIS: bool = false;
 
 impl Default for Arguments {
     fn default() -> Self {
+        let extensions: Vec<_> = POSSIBLE_CONFIG_EXTENSIONS
+            .into_iter()
+            .map(ToOwned::to_owned)
+            .collect();
+
         Self {
             submit_genesis: SUBMIT_GENESIS,
-            genesis_path: Some(GENESIS_PATH.into()),
-            config_path: CONFIGURATION_PATH.into(),
+            genesis_path: Some(
+                ConfigPath::new(GENESIS_PATH).with_possible_extensions(extensions.clone()),
+            ),
+            config_path: ConfigPath::new(CONFIGURATION_PATH).with_possible_extensions(extensions),
         }
+    }
+}
+
+/// Wrapper around path to config file (i.e. config.json, genesis.json).
+///
+/// Provides abstraction above file extension.
+#[derive(Debug, Clone)]
+pub struct ConfigPath {
+    path: PathBuf,
+    possible_extensions: Vec<String>,
+}
+
+impl ConfigPath {
+    /// Construct new [`ConfigPath`] from exact `path`
+    #[inline]
+    pub fn new(path: impl Into<PathBuf>) -> Self {
+        Self {
+            path: path.into(),
+            possible_extensions: Vec::new(),
+        }
+    }
+
+    /// Add possible extensions to the config path.
+    ///
+    /// # Note
+    ///
+    /// After using this function simple `path` provided in [`ConfigPath::new()`] will not be
+    /// counted as a valid path.
+    /// Only `path` with `possible_extensions` will be used in other functions like
+    /// [`ConfigPath::exists()`].
+    #[must_use]
+    pub fn with_possible_extensions(self, possible_extensions: Vec<String>) -> Self {
+        Self {
+            possible_extensions,
+            ..self
+        }
+    }
+
+    /// Try to get first existing path applying possible extensions if there are some.
+    pub fn first_existing_path(&self) -> Option<Cow<PathBuf>> {
+        if self.possible_extensions.is_empty() {
+            self.path.exists().then_some(Cow::Borrowed(&self.path))
+        } else {
+            self.possible_extensions.iter().find_map(|extension| {
+                let path = self.path.with_extension(extension);
+                path.exists().then_some(Cow::Owned(path))
+            })
+        }
+    }
+
+    /// Check if config path exists applying possible extensions if there are some.
+    pub fn exists(&self) -> bool {
+        if self.possible_extensions.is_empty() {
+            self.path.exists()
+        } else {
+            self.possible_extensions
+                .iter()
+                .any(|extension| self.path.with_extension(extension).exists())
+        }
+    }
+}
+
+impl FromStr for ConfigPath {
+    type Err = Infallible;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(Self::new(s))
     }
 }
 
@@ -154,7 +230,13 @@ impl Iroha {
         query_judge: QueryJudgeBoxed,
     ) -> Result<Self> {
         let broker = Broker::new();
-        let file_proxy = ConfigurationProxy::from_path(&args.config_path);
+        let file_proxy = ConfigurationProxy::from_path(
+            &args
+                .config_path
+                .first_existing_path()
+                .wrap_err("Configuration file does not exist")?
+                .as_ref(),
+        );
         let env_proxy = ConfigurationProxy::from_env();
         let config = file_proxy.override_with(env_proxy).build()?;
 
@@ -170,7 +252,12 @@ impl Iroha {
         let genesis = if let Some(genesis_path) = &args.genesis_path {
             GenesisNetwork::from_configuration(
                 args.submit_genesis,
-                RawGenesisBlock::from_path(genesis_path)?,
+                RawGenesisBlock::from_path(
+                    genesis_path
+                        .first_existing_path()
+                        .wrap_err("Genesis configuration file does not exist")?
+                        .as_ref(),
+                )?,
                 Some(&config.genesis),
                 &config.sumeragi.transaction_limits,
             )
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -58,22 +58,15 @@ pub struct Arguments {
 
 const CONFIGURATION_PATH: &str = "config";
 const GENESIS_PATH: &str = "genesis";
-const POSSIBLE_CONFIG_EXTENSIONS: [&str; 2] = ["json", "json5"];
+const PRECONFIGURED_CONFIG_EXTENSIONS: [&str; 2] = ["json", "json5"];
 const SUBMIT_GENESIS: bool = false;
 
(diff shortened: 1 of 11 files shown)


Arjentix:
@6r1d I believe, after the merge of this PR it will be good to update docs somewhere to mention, that we now support JSON5.
Also you may find it usefull to leave comments inside configs in tutorials or somethins like that, because JSON5 supports it

client_cli/src/main.rs
```diff
@@ -868,8 +875,12 @@ mod json {
...
     impl RunArgs for Args {
         fn run(self, cfg: &ClientConfiguration) -> Result<()> {
-            let reader = BufReader::new(stdin());
-            let instructions: Vec<Instruction> = serde_json::from_reader(reader)?;
+            let mut reader = BufReader::new(stdin());
+            let mut raw_content = Vec::new();
+            reader.read_to_end(&mut raw_content)?;
+
+            let content = String::from_utf8(raw_content)?;
+            let instructions: Vec<Instruction> = json5::from_str(&content)?;
```
appetrosyan, line 883: Genesis doesn't exactly need it, but it doesn't hurt.

client_cli/src/main.rs
```diff
@@ -147,7 +147,16 @@ fn main() -> Result<()> {
...
-        Configuration::from_str("config.json")?
+        let config_path = iroha::ConfigPath::new("config").with_preconfigured_extensions();
+        #[allow(clippy::expect_used)]
+        Configuration::from_str(
+            config_path
+                .first_existing_path()
+                .wrap_err("Configuration file does not exist")?
+                .as_ref()
+                .to_str()
+                .expect("Default config path is a valid UTF-8 string, qed."),
```
appetrosyan, line 158:
```suggestion
                .expect("Default config path wasn't a  valid UTF-8 string."),
```

Remember this is an Error message. On that note, consider adding a `config_path` as an argument.

client_cli/src/main.rs
```diff
@@ -63,8 +63,8 @@ impl FromStr for Configuration {
     fn from_str(file: &str) -> Result<Self> {
         let deser_err_msg = format!("Failed to decode config file {} ", &file);
         let err_msg = format!("Failed to open config file {}", &file);
-        let file = File::open(file).wrap_err(err_msg)?;
-        let cfg = serde_json::from_reader(file).wrap_err(deser_err_msg)?;
+        let content = fs::read_to_string(file).wrap_err(err_msg)?;
```
appetrosyan, line 66: Buffering this way slightly increases RAM usage. But usually faster, @ilchu keep this in mind in the future.

cli/src/main.rs
```diff
@@ -85,12 +84,12 @@ fn print_help() {
     println!("pass `--version` or `-V` to print version information");
     println!();
     println!("Iroha 2 is configured via environment variables:");
-    println!("    IROHA2_CONFIG_PATH is the location of your `config.json`");
-    println!("    IROHA2_GENESIS_PATH is the location of `genesis.json`");
+    println!("    IROHA2_CONFIG_PATH is the location of your `config.json` or `config.json5`");
+    println!("    IROHA2_GENESIS_PATH is the location of `genesis.json` or `genesis.json5`");
```
appetrosyan, line 88: We don't actually check the extension, and given that the `json5` file is also `json`, it's kind of moot.

cli/src/lib.rs
```diff
@@ -170,7 +247,12 @@ impl Iroha {
         let genesis = if let Some(genesis_path) = &args.genesis_path {
             GenesisNetwork::from_configuration(
                 args.submit_genesis,
-                RawGenesisBlock::from_path(genesis_path)?,
+                RawGenesisBlock::from_path(
+                    genesis_path
+                        .first_existing_path()
+                        .wrap_err("Genesis configuration file does not exist")?
```
appetrosyan, line 253:
```suggestion
                        .wrap_err("Genesis  block file doesn't exist")?
```

cli/src/lib.rs
```diff
@@ -50,25 +51,95 @@ pub struct Arguments {
...
     /// Set custom config file path.
-    pub config_path: PathBuf,
+    pub config_path: ConfigPath,
 }
 
-const CONFIGURATION_PATH: &str = "config.json";
-const GENESIS_PATH: &str = "genesis.json";
+const CONFIGURATION_PATH: &str = "config";
+const GENESIS_PATH: &str = "genesis";
+const PRECONFIGURED_CONFIG_EXTENSIONS: [&str; 2] = ["json", "json5"];
```
QuentinI, line 61: A bit unclear why "preconfigured" instead of "supported" or "default". Not necessarily insisting on latter, just don't understand your what you mean by "preconfigured"

cli/src/lib.rs
```diff
@@ -50,25 +51,95 @@ pub struct Arguments {
...
     }
 }
 
+/// Wrapper around path to config file (i.e. config.json, genesis.json).
+///
+/// Provides abstraction above file extension.
+#[derive(Debug, Clone)]
+pub struct ConfigPath {
+    path: PathBuf,
+    with_preconfigured_extensions: bool,
```
QuentinI, line 80: Is there any reason we would ever want it set to false?

QuentinI: Also supporting this results in slightly confusing API that necessitates the note later

> ```rust
>    /// # Note
>    ///
>    /// After using this function simple `path` provided in [`ConfigPath::new()`] will not be
>    /// counted as a valid path.
>    /// Only `path` with `possible_extensions` will be used in other functions like
>    /// [`ConfigPath::exists()`].
> ```

client_cli/Cargo.toml
```diff
@@ -21,11 +21,12 @@ iroha_client = { version = "=2.0.0-pre-rc.11", path = "../client" }
 iroha_data_model = { version = "=2.0.0-pre-rc.11", path = "../data_model" }
 iroha_crypto = { version = "=2.0.0-pre-rc.11", path = "../crypto" }
 iroha_config = { version = "=2.0.0-pre-rc.11", path = "../config" }
+iroha = { version = "=2.0.0-pre-rc.11", path = "../cli" }
```
QuentinI, line 24: Can we move `ConfigPath` to `iroha_config` instead?

client_cli/src/main.rs
ilchu: That's how I've done it in the config's `LoadFromDisk` trait. As these are usually called once on startup, hope the effect is almost negligible.

cli/src/lib.rs
ilchu: Or maybe "allowed" would pass as well, though not insisting either.

client_cli/Cargo.toml
```diff
@@ -21,11 +21,12 @@ iroha_client = { version = "=2.0.0-pre-rc.11", path = "../client" }
...
 
 color-eyre = "0.6.2"
 clap = { version = "3.2.16", features = ["derive"] }
 dialoguer = { version = "0.10.2", default-features = false }
-serde_json = "1.0.83"
+json5 = "0.4.1"
```
ilchu, line 29: Is there any other crate where we technically don't need full `serde_json` functionality and could make a switch like this?

cli/src/lib.rs
ilchu: It also seems to me that for now it's more appropriate to have the extension list exhaustive, i.e. failing immediately when extension is neither of the current two. That would make for a more transparent error reporting as well, compared to if it happened further down the line.

client_cli/src/main.rs
Arjentix: I think it may be usefull for tutorials and debugging, where comments will be very helpful

client_cli/src/main.rs
Arjentix: I decided to use `to_string_lossy()` and remove this `expect()`. Also it will always valid, becuase it's the default `config.json` or `config.json5` path

client_cli/src/main.rs
Arjentix: I like how it was done before, but unfortunately `json5` crate has nothing like `serde_json::from_reader()`

cli/src/main.rs
Arjentix: Yeah, *json* is a valid *json5*, so I think it's okay to use `json5` crate to parse just *json*-files.

Or do you suggest to prohibit non *json*/*json5* extensions?

cli/src/lib.rs
Arjentix: Okay, I like *default*

cli/src/lib.rs
Arjentix: > Is there any reason we would ever want it set to false?

Yeas, when user provides their own path to config, then extension is clear. This abstraction is needed to find `config.json` or `config.json5` in *default* path

Arjentix: > It also seems to me that for now it's more appropriate to have the extension list exhaustive, i.e. failing immediately when extension is neither of the current two

Okay, I'll do it

cli/src/main.rs
Arjentix: @ilchu suggested it [here](https://github.com/hyperledger/iroha/pull/3030#discussion_r1054465803) so I'll resolve this

cli/src/lib.rs
Arjentix: > Also supporting this results in slightly confusing API that necessitates the note later

I was trying to provide an abstraction above path to a configuration. It's either some default expected path with two (or more in future) acceptable extensions or a path provided by user.

So we either add default extensions or not depending on user behaviour.

Don't you like this idea?

Arjentix: I'll change API a little bit to make it less confusing

cli/src/lib.rs
Arjentix: Or *allowed* if we disallow other extensions

client_cli/Cargo.toml
Arjentix: Good idea, but I can't find any more. It's related only to *cli* things like `cli` and `client_cli` crates. It's worked for `client_cli` here, but doesn't work for `cli`, because `torrii` inside it uses `serde_json`.

Also `json5` it-self depends on `serde_json` so we won't gain much of it

appetrosyan (dismissed):

cli/src/lib.rs
QuentinI: > Yes, when user provides their own path to config, then extension is clear. This abstraction is needed to find config.json or config.json5 in default path

Right, sorry, didn't catch that.

QuentinI (dismissed):
LGTM!

config/src/path.rs
```diff
@@ -0,0 +1,96 @@
...
+        match &self.0 {
+            Default(path) => ALLOWED_CONFIG_EXTENSIONS.iter().find_map(|extension| {
+                let path_ext = path.with_extension(extension);
+                path_ext.exists().then_some(Cow::Owned(path_ext))
+            }),
+            UserProvided(path) => path.exists().then_some(Cow::Borrowed(&path)),
+        }
+    }
+
+    /// Check if config path exists applying possible extensions if there are some.
```
outoftardis, line 87:
```suggestion
    /// Check if a config path exists by applying allowed extensions if there are any.
```

config/src/path.rs
```diff
@@ -0,0 +1,96 @@
...
+            .ok_or_else(|| Error::UserProvidedConfigFileHasNoExtension)?
+            .to_string_lossy();
+        if !ALLOWED_CONFIG_EXTENSIONS.contains(&extension.as_ref()) {
+            return Err(Error::InvalidExtensionError(extension.into_owned()));
+        }
+
+        Ok(Self(UserProvided(path)))
+    }
+
+    /// Try to get first existing path applying possible extensions if there are some.
```
outoftardis, line 76:
```suggestion
    /// Try to get first existing path by applying allowed extensions if there are any.
```

config/src/path.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[error("Provided by default config file has extension when it should not have one.")]
+    DefaultConfigFileHasExtension,
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Inner helper struct.
+///
+/// This could be [`Path`] itself, but it is not because enum can be constructed directly and
+/// we need checked way to construct it with written constructors.
```
outoftardis, line 32:
```suggestion
/// With this struct, we can construct `Path` with written constructors instead of doing it directly.
```

Arjentix:
Rebased two commits just to add sign-offs

QuentinI (approved):

ilchu (approved):

appetrosyan (approved):

PR #3032, opened 2022-12-21
[test] #2506: Add more tests on block validation

### LINKED ISSUE #2506: [PERFORMANCE BUG] When sending ~ 5000 assets in a row then Iroha 2 down without information in logs.

pesterev:

### Description of the Change

This test reproduces the root error of #2506. A detailed explanation of the bug is [here](https://github.com/hyperledger/iroha/issues/2506#issuecomment-1316913752).

### Issue

Closes #2506

### Benefits

Add more tests on the block validation.

### Possible Drawbacks

None

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -1136,7 +1136,12 @@ pub mod stream {
 mod tests {
     #![allow(clippy::restriction)]
 
+    use std::{str::FromStr, sync::Arc};
+
+    use iroha_data_model::prelude::*;
+
     use super::*;
+    use crate::kura::Kura;
 
     #[test]
     pub fn committed_and_valid_block_hashes_are_equal() {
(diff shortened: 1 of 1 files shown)


core/src/block.rs
```diff
@@ -1145,4 +1150,64 @@ mod tests {
...
+                alice_id.clone(),
+                [create_asset_definition.clone()].into(),
+                4000,
+            )
+            .sign(alice_keys.clone())
+            .expect("Valid");
+            crate::VersionedAcceptedTransaction::from_transaction(tx, &transaction_limits)
+                .expect("Valid")
+        };
+        let tx_b = {
```
QuentinI, line 1193: Couldn't we just clone here?

QuentinI (approved):

Erigara (approved):

QuentinI (approved):

Erigara (approved):

Issue #3027, opened 2022-12-20
Introduce lockfile to kura


QuentinI:
To be able to provide helpful information to the user when multiple instances of iroha are configured to use the same storage, such as in #2847, we need to detect such a situation in the first place.

PR #3033, opened 2022-12-21
[feature] #3027: implement lockfile for Kura

QuentinI:
### Description of the Change

Introduced a lockfile mechanism to Kura to prevent running multiple writing instances over one store.

### Issue

#3027 

### Benefits

As per #3027

### Possible Drawbacks

More complexity in Kura.

### Alternate Designs

- This PR uses typestate pattern to encode lock state in attempt to prevent accidental misuse when changing Kura code. One could argue for a simpler design that would allow slightly more room for mistakes.
- Storage is locked for full lifetime of Kura. There was an alternate design discussed where lock is acquired only when writing blocks, however it is worse for performance and still allows starting multiple Iroha instances backed by the same storage, postponing failure for indeterminate time in the future.

### DIFF
--- a/client/tests/integration/restart_peer.rs
+++ b/client/tests/integration/restart_peer.rs
@@ -55,6 +55,7 @@ fn restarted_peer_should_have_the_same_asset_amount() -> Result<()> {
             .find(|asset| asset.id().definition_id == asset_definition_id)
             .expect("Asset not found");
         assert_eq!(AssetValue::Quantity(quantity), *asset.value());
+        peer.stop();
     }
 
     {
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -2,20 +2,16 @@
 //! logic.  [`Kura`] is the main entity which should be used to store
 //! new [`Block`](`crate::block::VersionedCommittedBlock`)s on the
 //! blockchain.
-#![allow(
-    clippy::module_name_repetitions,
-    clippy::std_instead_of_core,
-    clippy::std_instead_of_alloc,
-    clippy::arithmetic_side_effects,
-    clippy::significant_drop_in_scrutinee
-)]
+#![allow(clippy::std_instead_of_alloc, clippy::arithmetic_side_effects)]
 use std::{
     fmt::Debug,
+    fs,
     io::{Read, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
     sync::Arc,
 };
 
+use derive_more::Deref;
 use iroha_config::kura::{Configuration, Mode};
 use iroha_crypto::HashOf;
 use iroha_logger::prelude::*;
@@ -23,6 +19,10 @@ use iroha_version::scale::{DecodeVersioned, EncodeVersioned};
 use parking_lot::Mutex;
 
 use crate::{block::VersionedCommittedBlock, handler::ThreadHandler};
+
+const INDEX_FILE_NAME: &str = "blocks.index";
+const DATA_FILE_NAME: &str = "blocks.data";
+const LOCK_FILE_NAME: &str = "kura.lock";
 
 /// The interface of Kura subsystem
 #[derive(Debug)]
@@ -32,7 +32,7 @@ pub struct Kura {
     /// The mode of initialisation of [`Kura`].
     mode: Mode,
     /// The block storage
-    block_store: Mutex<BlockStore>,
+    block_store: Mutex<BlockStore<Locked>>,
     /// The array of block hashes and a slot for an arc of the block. This is normally recovered from the index file.
     #[allow(clippy::type_complexity)]
     block_data: Mutex<
@@ -53,13 +53,12 @@ impl Kura {
     /// Fails if there are filesystem errors when trying
     /// to access the block store indicated by the provided
     /// path.
-    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
     pub fn new(
         mode: Mode,
         block_store_path: &Path,
         debug_output_new_blocks: bool,
     ) -> Result<Arc<Self>> {
-        let mut block_store = BlockStore::new(block_store_path);
+        let mut block_store = BlockStore::new(block_store_path).lock()?;
         block_store.create_files_if_they_do_not_exist()?;
 
         let block_plain_text_path = debug_output_new_blocks.then(|| {
@@ -81,7 +80,7 @@ impl Kura {
     /// Create a kura instance that doesn't write to disk. Instead it serves as a handler
     /// for in-memory blocks only.
     pub fn blank_kura_for_testing() -> Arc<Kura> {
-        let block_store = BlockStore::new(Path::new(""));
+        let block_store = BlockStore::fake_locked_for_tests(PathBuf::new());
         Arc::new(Self {
             mode: Mode::Strict,
             block_store: Mutex::new(block_store),
@@ -126,7 +125,7 @@ impl Kura {
     /// Fails if:
     /// - file storage is unavailable
     /// - data in file storage is invalid or corrupted
-    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
+    #[allow(clippy::expect_used)]
     pub fn init(&self) -> Result<Vec<VersionedCommittedBlock>> {
         let mut blocks = Vec::new();
 
@@ -242,7 +241,7 @@ impl Kura {
             let mut block_store_guard = kura.block_store.lock();
             if let Err(error) = block_store_guard.write_index_count(start_height as u64) {
                 error!("Failed to write index count, ERROR = {}", error);
-                panic!("Kura has encountered a fatal I/O error.");
+                panic!("Kura has encountered a fatal IO error.");
             }
 
             for block in blocks_to_be_written {
@@ -250,14 +249,14 @@ impl Kura {
 
                 if let Err(error) = block_store_guard.append_block_to_chain(&serialized_block) {
                     error!("Failed to store block, ERROR = {}", error);
-                    panic!("Kura has encountered a fatal I/O error.");
+                    panic!("Kura has encountered a fatal IO error.");
                 }
             }
         }
     }
 
     /// Get the hash of the block at the provided height.
-    #[allow(clippy::unwrap_in_result, clippy::expect_used)]
+    #[allow(clippy::expect_used)]
     pub fn get_block_hash(&self, block_height: u64) -> Option<HashOf<VersionedCommittedBlock>> {
         let hash_data_guard = self.block_data.lock();
         if block_height == 0 || block_height > hash_data_guard.len() as u64 {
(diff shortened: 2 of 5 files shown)


client/tests/integration/restart_peer.rs
```diff
@@ -55,6 +55,7 @@ fn restarted_peer_should_have_the_same_asset_amount() -> Result<()> {
             .find(|asset| asset.id().definition_id == asset_definition_id)
             .expect("Asset not found");
         assert_eq!(AssetValue::Quantity(quantity), *asset.value());
+        peer.stop();
```
pesterev, line 58: This design confuses me a bit.
After stopping, the peer goes into an undefined state, and all methods are still available to use. This may lead to misusing the peer or to even bugs imho.
It made me think about introducing StoppedPeer.
```rust
// drop the `peer` here
let stopped_peer: StoppedPeer = peer.stop();
// ....
// restart a new `peer` and drop the `stopped_peer`
let peer = stopped_peer.restart(...);
// or
let peer = PeerBuilder::new().restart(stopped_peer);
```

Erigara: Or maybe we can recreate peer in each scope? And rely on `Drop`.

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
+        // skip cloning the path.
+        self.try_lift_lock()?;
+        let mut path = PathBuf::new();
+        core::mem::swap(&mut path, &mut self.0);
 
-    // Above are the platform dependent functions.
-    // Below are the platform independent functions that use the above functions.
+        // We explicitly want to skip running the destructor
+        #[allow(clippy::mem_forget)]
+        core::mem::forget(self);
```
Erigara, line 418: Maybe? 
```suggestion
        // We explicitly want to skip running the destructor
        let mut me = core::mem::ManuallyDrop::new(self);
        let path = core::mem::take(&mut me.0);
```

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
+            path_to_blockchain: Unlocked::new(path.to_path_buf()),
         }
     }
+
+    /// Try to acquire a lockfile and convert self to
+    /// read-write [`BlockStore<Locked>`]
+    ///
+    /// # Errors
+    /// If failed to acquire a lockfile due to
+    /// I/O errors or it already existing
```
outoftardis, line 454:
```suggestion
    /// - I/O errors
    /// - lockfile already exists
```

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
 
-impl BlockStoreTrait for StdFileBlockStore {
+/// Operations available both on locked and unlocked store,
+/// i.e. only read operations.
+impl<L: Lock> BlockStore<L> {
+    /// Read a series of block indices from the block index file and
+    /// attempt to fill all of `dest_buffer`.
+    ///
+    /// # Errors
+    /// IO Error.
```
outoftardis, line 469: Here it's `IO` but you use `I/O` in another description, please be consistent

core/src/kura.rs
```diff
@@ -523,11 +509,91 @@ impl BlockStoreTrait for StdFileBlockStore {
...
+            .read(true)
+            .open(self.path_to_blockchain.join(DATA_FILE_NAME))?;
+        data_file.seek(SeekFrom::Start(start_location_in_data_file))?;
+        data_file.read_exact(dest_buffer)?;
+        Ok(())
+    }
+}
+
+/// Functions available only for lock-protected block store,
+/// i.e. ones writing to it.
```
outoftardis, line 562:
```suggestion
/// i.e. writing.
```

core/src/kura.rs
```diff
@@ -523,11 +509,91 @@ impl BlockStoreTrait for StdFileBlockStore {
...
+        Self {
+            path_to_blockchain: Locked(path),
+        }
+    }
+    /// Try to release the lockfile and convert self to
+    /// read-only [`BlockStore<Unlocked>`]
+    ///
+    /// # Errors
+    /// If failed to release a lockfile due to
+    /// I/O errors or it not existing
```
outoftardis, line 579:
```suggestion
    /// - I/O errors
    /// - lockfile doesn't exist
```

core/src/kura.rs
```diff
@@ -587,16 +648,49 @@ impl BlockStoreTrait for StdFileBlockStore {
         Ok(())
     }
 
-    fn create_files_if_they_do_not_exist(&mut self) -> Result<()> {
-        std::fs::create_dir_all(&self.path_to_blockchain)?;
+    /// Create the index and data files if they do not
+    /// already exist.
+    ///
+    /// # Errors
+    /// Fails if the any of the files don't exist
```
outoftardis, line 655:
```suggestion
    /// Fails if any of the files don't exist
```

core/src/kura.rs
```diff
@@ -587,16 +648,49 @@ impl BlockStoreTrait for StdFileBlockStore {
...
             .write(true)
             .create(true)
-            .open(self.path_to_blockchain.join(Self::DATA_FILE_NAME))?;
+            .open(self.path_to_blockchain.join(DATA_FILE_NAME))?;
+        Ok(())
+    }
+
+    /// Append `block_data` to this block store. First writing
+    /// the data to the data file and then creating a new index
+    /// for it in the index file.
```
outoftardis, line 672:
```suggestion
    /// Append `block_data` to this block store. First write
    /// the data to the data file and then create a new index
    /// for it in the index file.
```

core/src/kura.rs
```diff
@@ -587,16 +648,49 @@ impl BlockStoreTrait for StdFileBlockStore {
...
+    ///
+    /// # Errors
+    /// Fails if any of the required platform specific functions
```
outoftardis, line 675:
```suggestion
    /// Fails if any of the required platform-specific functions
```

core/test_network/src/lib.rs
```diff
@@ -519,6 +510,24 @@ impl Peer {
         self.temp_dir = Some(temp_dir);
     }
 
+    /// Stops the peer if running
```
outoftardis, line 513:
```suggestion
    /// Stop the peer if it's running
```

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
-    /// # Errors
-    /// IO Error.
-    fn read_block_indices(
-        &self,
-        start_block_height: u64,
-        dest_buffer: &mut [(u64, u64)],
-    ) -> Result<()>;
+// Marker trait for typestate. Not sealed, since while it
+// doesn't need to be implemented by a user, implementing it
+// shouldn't lead to problematic behaviour.
```
outoftardis, line 342:
```suggestion
/// Marker trait for typestate. Not sealed, because while it
/// doesn't need to be implemented by a user, implementing it
/// shouldn't lead to problematic behaviour.
```

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
-    /// IO Error.
-    ///
-    /// The most common reason this function fails is
-    /// that you did not call `create_files_if_they_do_not_exist`.
-    ///
-    /// Note that if there is an error, you can be quite sure all other
-    /// read and write operations will also fail.
-    fn write_index_count(&mut self, new_count: u64) -> Result<()>;
+/// Marker struct for typestate, signifies that store is
+/// not protected by a lockfile we own and thus can be used
```
outoftardis, line 347:
```suggestion
/// Marker struct for typestate. The marker signifies that the store is
/// not protected by a lockfile we own and thus can be used
```

who are `we` in this context?

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
-    /// # Errors
-    /// IO Error.
-    fn write_block_data(
-        &mut self,
-        start_location_in_data_file: u64,
-        block_data: &[u8],
-    ) -> Result<()>;
+/// Marker struct for typestate, signifies that store is
+/// protected by a lockfile owned by us and thus can be
+/// written into.
```
outoftardis, line 384: same as above

client/tests/integration/restart_peer.rs
QuentinI: > the peer goes into an undefined state

It's defined I think - basically Iroha instance is dropped, and all the context for it isn't. You shouldn't get any problems reusing it. I'll have a look again to double-check if there's any strange behaviour one can trigger that way.

>  It made me think about introducing StoppedPeer.

That sound good though, having type indicate state instead of struct of `Options` is almost always a cleaner solution IMO. That said, I think it's out of scope for this if `stop` indeed doesn't lead to any problems.

core/src/kura.rs
QuentinI: It's more of implementation detail, don't think we need to clutter the docs with it. Basically if you need to touch it, you are already changing this code directly.

core/src/kura.rs
QuentinI: > who are we in this context?

This `Kura` instance. I'll clarify it.

core/src/kura.rs
QuentinI: Could you elaborate please? Your suggestion feels overly succinct to me.

core/src/kura.rs
outoftardis: ok. it's just the similar comments just below this one are doc comments, I thought this might've been a typo

core/src/kura.rs
outoftardis: I'd make this similar to lines 438 and 461, maybe this is better:

```suggestion
/// i.e. write operations.
```

QuentinI:
@outoftardis thank you for proofreading!

core/src/kura.rs
```diff
@@ -734,8 +824,46 @@ mod tests {
...
     }
 
+    #[test]
+    fn lock_and_unlock() {
+        let dir = tempfile::tempdir().unwrap();
+        {
+            let _store = BlockStore::new(dir.path())
+                .lock()
+                .expect("Lock acquisition failed");
+            fs::File::open(dir.path().join(LOCK_FILE_NAME))
```
Erigara, line 834: You can use [`try_exist`](https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists).

Erigara (commented):
Great usage of typestates.

core/src/kura.rs
QuentinI: Nice, thanks

core/src/kura.rs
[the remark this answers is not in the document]
QuentinI: Oh, right, this is literally what `mem::take` is for. Thanks for the catch!

core/src/kura.rs
```diff
@@ -339,176 +337,164 @@ impl Kura {
...
+impl<L: Lock> BlockStore<L> {
+    /// Read a series of block indices from the block index file and
+    /// attempt to fill all of `dest_buffer`.
+    ///
+    /// # Errors
+    /// IO Error.
     #[allow(
         clippy::needless_range_loop,
         clippy::unwrap_used,
         clippy::unwrap_in_result
```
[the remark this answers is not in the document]
QuentinI: I've been waiting for working lints in CI to clean up `allow`s in Kura. Now that we know we are going to go with nightly clippy I'll just run Kura through it locally and fix/allow as needed.

client/tests/integration/restart_peer.rs
QuentinI: @Erigara @pesterev I looked through it once more and I'm convinced `stop` doesn't lead to any problematic behavior.

I agree it's not the best design, but I think we need to refactor `Peer` and `PeerBuilder` wholesale to do measurably better. What do you think?

pesterev: Got it. Let's go ahead with this.

Erigara: Agree with you let's return to this later.

Erigara (approved):

pesterev (approved):

Issue #3039, opened 2022-12-26
[refactor]: Introduce waiting buffer for the multisigs


Erigara:
Add waiting buffer for multi-signature transactions where they will be waiting for the more signatures.

PR #3035, opened 2022-12-22
[fix] #2963: Queue remove transactions correctly 

Erigara:
### Description of the Change

- Add queue size to metrics;
- Fix Queue to correctly remove transactions;

### Issue

Closes #2963.

### Benefits

- No infinite growth of the transaction queue;
- More diagnostic information in the `/status` response.  

### Possible Drawbacks

None.

### Usage Examples or Tests *[optional]*

Queue size field in the status response: 
```
curl http://127.0.0.1:8180/status
```

### Alternate Designs *[optional]*

Multi-signatures are still not handled correctly by the queue, i would suggest fix this in the separate PR.

PR #3043, opened 2022-12-27
[fix] #3038: Re-enable multisigs

### LINKED ISSUE #2595: Fix multisignature
After the merge of #2582 we will broke multisignature mechanism. It was working because of signature checking mistake (see #2580).

### LINKED ISSUE #3038: [fix]: Re-enable multisigs
Currently multi-signatures transactions is not working due to changes in the `Queue` and `Sumeragi`, this changes can be rollbacked in order to re-enable multisigs.

Erigara:
### Description of the Change

- Revert some changes made to `Queue` and `Sumeragi` to bring multisignatures back. 
- Turn on tests that works with multisigs.
- Add second queue for the multisignatures.

We need second queue for multisigns in order to prevent situation when multisig transactions prevent ordinary transactions from being processed. 

### Issue

Closes #2595, #3038, #3039.

### Benefits

Multisignatures are working again. 

### Possible Drawbacks

More complex queue.

### Usage Examples or Tests

Test for multisig functionality on one peer:

```bash
cargo test --package iroha_core --lib -- queue::tests::push_multisignature_tx --exact --nocapture
```

Test for multisig transaction gossiping between multiple peers: 

```bash
cargo test --package iroha_client --test mod -- integration::multisignature_transaction::multisignature_transactions_should_wait_for_all_signatures --exact --nocapture
```


core/src/sumeragi/main_loop.rs
```diff
@@ -231,12 +231,11 @@ fn gossip_transactions<F>(
     F: FaultInjection,
 {
     if last_sent_transaction_gossip_time.elapsed() > sumeragi.gossip_period {
-        let txs = state
-            .transaction_cache
-            .iter()
-            .take(sumeragi.gossip_batch_size as usize)
-            .cloned()
-            .collect::<Vec<_>>();
+        // Transactions are intentionally taken from the queue instead of the cache
```
QuentinI, line 234: This could impact performance, did you run any comparisons?

QuentinI (commented):
I assume this builds upon #3035, so looked only at the last commit.

core/src/sumeragi/main_loop.rs
Erigara: Not yet, i have problems running `tps` benchmark on `dev` branch.

Erigara: Result of `tps-dev` benchmark doesn't reveal any changes in the performance.

```
Before (fbc7d346):
time:   [381.34 tps 455.59 tps 529.53 tps]

After (81448627):
time:   [433.57 tps 503.22 tps 568.07 tps]
change: [+26.541% +9.4650% -11.143%] (p = 0.39 > 0.05)
No change in performance detected.
```

config/iroha_test_config.json
```diff
@@ -57,6 +57,7 @@
   "QUEUE": {
     "MAXIMUM_TRANSACTIONS_IN_BLOCK": 8192,
     "MAXIMUM_TRANSACTIONS_IN_QUEUE": 65536,
+    "MAXIMUM_TRANSACTIONS_IN_WAITING_BUFFER": 65536,
```
appetrosyan, line 60:
```suggestion
    "MAXIMUM_TRANSACTIONS_IN_SIGNATURE_BUFFER": 65536,
```

core/src/queue.rs
```diff
@@ -26,8 +27,10 @@ use crate::prelude::*;
 /// Multiple producers, single consumer
 #[derive(Debug)]
 pub struct Queue {
-    /// The queue proper
+    /// The queue for transactions that pass signature check
```
appetrosyan, line 30:
```suggestion
    /// The queue for transactions that passed signature check
```

core/src/queue.rs
```diff
@@ -147,49 +147,84 @@ impl Queue {
         tx: VersionedAcceptedTransaction,
         wsv: &WorldStateView,
     ) -> Result<(), (VersionedAcceptedTransaction, Error)> {
```
appetrosyan, line 149: Looks to me like the signature is a little off. Do you think it should be a tuple instead of `struct`?

core/src/queue.rs
```diff
@@ -147,49 +147,84 @@ impl Queue {
...
-                    old_tx
-                        .get_mut()
-                        .as_mut_v1()
-                        .signatures
-                        .extend(tx.as_v1().signatures.clone());
-                    return Ok(());
+        match self.check_tx(&tx, wsv) {
+            Err(e) => Err((tx, e)),
+            Ok(MustUse(signature_check)) => {
+                // Get `txs_len` before entry to avoid deadlock
```
appetrosyan, line 153: Or, get the lock explicitly and drop it when done. Picking up and almost immediately dropping a mutex guard is not the worst but it's inefficient.

core/src/queue.rs
```diff
@@ -147,49 +147,84 @@ impl Queue {
...
+                    }
+                    Entry::Vacant(entry) => entry,
+                };
+                if txs_len >= self.max_txs {
+                    return Err((tx, Error::Full));
                 }
-                Entry::Vacant(entry) => entry,
+
+                // Reason for such insertion order is to avoid situation
+                // when poped from the `queue` hash does not yet has corresponding (hash, tx) record in `txs`
```
appetrosyan, line 173:
```suggestion
                // Insert entry first so that the `tx` popped from `queue` will always have a `(hash, tx)` record in `txs`. 
```

core/src/queue.rs
```diff
@@ -147,49 +147,84 @@ impl Queue {
...
+                        .expect("Inserted just before match");
+                    (err_tx, Error::Full)
+                })
+            }
+        }
+    }
+
+    /// Pop single transaction from the waiting buffer.
+    ///
+    /// Records unsigned transaction in `seen`.
```
appetrosyan, line 193:
```suggestion
    /// Pop single transaction from the waiting buffer. Record unsigned transaction in `seen`.
```

core/src/queue.rs
```diff
@@ -147,49 +147,84 @@ impl Queue {
...
+        &self,
+        seen: &mut Vec<HashOf<VersionedSignedTransaction>>,
+        wsv: &WorldStateView,
+    ) -> Option<VersionedAcceptedTransaction> {
+        loop {
+            let hash = self.waiting_buffer.pop()?;
+            let entry = match self.txs.entry(hash) {
+                Entry::Occupied(entry) => entry,
+                // As practice shows this code is not `unreachable!()`.
+                // When transactions are submitted quickly it can be reached.
```
appetrosyan, line 204:
```suggestion
                // FIXME: Reachable under high load. Investigate, see if it's a problem. 
```

core/src/queue.rs
```diff
@@ -147,49 +147,84 @@ impl Queue {
...
-                let (_, err_tx) = self
-                    .txs
-                    .remove(&err_hash)
-                    .expect("Inserted just before match");
-                (err_tx, Error::Full)
-            })
+            match self.check_tx(entry.get(), wsv) {
+                Err(_) => {
+                    entry.remove_entry();
+                    continue;
```
appetrosyan, line 211: Technically this `continue` is redundant. Given that the other pattern is an early return, just reorder the arms and move this to the bottom.

core/src/queue.rs
```diff
@@ -249,21 +279,115 @@ impl Queue {
...
-        let out = std::iter::from_fn(|| self.pop(&mut seen, wsv))
+        let txs = txs_from_queue
+            .round_robin(txs_from_waiting_buffer)
             .filter(|tx| !transactions_hashes.contains(&tx.hash()))
             .take(self.txs_in_block - transactions.len());
-        transactions.extend(out);
+        transactions.extend(txs);
+
+        for (seen, queue) in [
+            (seen_queue, &self.queue),
```
appetrosyan, line 300: This function is an unholy mix of functional and imperative. Think we can rewrite parts of it to be one or the other?

core/src/queue.rs
```diff
@@ -249,21 +279,115 @@ impl Queue {
...
+            (seen_waiting_buffer, &self.waiting_buffer),
+        ] {
+            #[allow(clippy::expect_used)]
```
appetrosyan, line 303:
```suggestion

```
No longer necessary.

core/src/queue.rs
```diff
@@ -249,21 +279,115 @@ impl Queue {
...
+            seen.into_iter()
+                .try_for_each(|hash| queue.push(hash))
+                .expect("As we never exceed the number of transactions pending");
```
appetrosyan, line 306:
```suggestion
                .expect("Exceeded the number of transactions pending");
```

core/src/queue.rs
```diff
@@ -249,21 +279,115 @@ impl Queue {
...
+        }
+    }
+}
+
+/// Iterator which combine two iterators into the single one in round-robin fashion.
+///
+/// ```ignore
+/// [(a0,a1,a2,..),(b0,b1,b2,..)] -> (a0,b0,a1,b1,a2,b2,..)
+/// ```
+struct RoundRobinIterator<A, B> {
```
appetrosyan, line 316: Think this might be useful in `iroha_primitives`.

core/src/queue.rs
```diff
@@ -771,4 +1063,54 @@ mod tests {
         assert!(matches!(queue.push(tx, &wsv), Err((_, Error::InFuture))));
         assert_eq!(queue.txs.len(), 1);
     }
+
+    #[test]
```
appetrosyan, line 1067: Still more reasons to extract this into `iroha_primitives`.

core/src/queue.rs
Erigara: I think dropping lock is not option here because we need to insert value into the `entry` just after check for size.

core/src/queue.rs
Erigara: Sorry i'm not quite get your point on this one.

appetrosyan: Consider adding:
```rust 
struct Failure {
    tx: VersionedAcceptedTransaction, 
    err: Error
}
```

For clarity and ease of use.

core/src/queue.rs
Erigara: Hope it looks better now

core/src/queue.rs
```diff
@@ -137,59 +145,88 @@ impl Queue {
...
-    ///
-    /// Records unsigned transaction in `seen`.
-    #[allow(
-        clippy::expect_used,
-        clippy::unwrap_in_result,
-        clippy::cognitive_complexity
-    )]
-    fn pop(
+    /// Pop single transaction from the queue. Record unsigned transaction in `seen`.
+    fn pop_from_queue(
```
mversic, line 229: return the naming to just `pop`, obviously it's being popped from the queue

core/src/queue.rs
```diff
@@ -202,20 +239,15 @@ impl Queue {
...
                 Entry::Vacant(_) => continue,
             };
-            if self.check_tx(entry.get(), wsv).is_err() {
+
+            if !self.is_pending(entry.get(), wsv) {
                 entry.remove_entry();
                 continue;
             }
 
             // Transactions are not removed from the queue until expired or committed
```
mversic, line 248: and why not? Only sumeragi is ever popping transactions out of the queue. It might as well remove them from the queue

core/src/queue.rs
```diff
@@ -137,59 +145,88 @@ impl Queue {
...
-    #[allow(
-        clippy::expect_used,
-        clippy::unwrap_in_result,
-        clippy::cognitive_complexity
-    )]
-    fn pop(
+    /// Pop single transaction from the queue. Record unsigned transaction in `seen`.
+    fn pop_from_queue(
         &self,
         seen: &mut Vec<HashOf<VersionedSignedTransaction>>,
```
mversic, line 231: `seen` seems useless now? How can you have an unsigned transaction in the queue?

core/src/queue.rs
```diff
@@ -137,59 +145,88 @@ impl Queue {
...
-            self.queue.push(hash).map_err(|err_hash| {
-                let (_, err_tx) = self
-                    .txs
-                    .remove(&err_hash)
-                    .expect("Inserted just before match");
-                (err_tx, Error::Full)
-            })
+            match self.check_tx(entry.get(), wsv) {
+                Ok(MustUse(signature_check)) => {
+                    // Transactions are not removed from the queue until expired or committed
```
mversic, line 215: again, it's unclear why not exactly

mversic (changes requested):

client/tests/integration/multisignature_transaction.rs
```diff
@@ -12,7 +12,6 @@ use test_network::*;
 use super::Configuration;
 
 #[allow(clippy::too_many_lines)]
-#[ignore = "Multisignature is not working for now. See #2595"]
```
mversic, line 15: I think you should reference `#2595` in the PR

core/src/queue.rs
```diff
@@ -137,59 +145,88 @@ impl Queue {
...
-                    // MST case
-                    old_tx
-                        .get_mut()
-                        .as_mut_v1()
-                        .signatures
-                        .extend(tx.as_v1().signatures.clone());
-                    return Ok(());
+    ) -> Result<(), Failure> {
+        match self.check_tx(&tx, wsv) {
+            Err(err) => Err(Failure { tx, err }),
```
mversic, line 154: I think we should remove transaction from the queue here and maybe in `is_pending` as well

primitives/src/round_robin_iter.rs
```diff
@@ -0,0 +1,141 @@
+//! Contains round-robin iterator and related trait
+
+/// Iterator which combine two iterators into the single one in round-robin fashion.
+///
+/// ```ignore
+/// [(a0,a1,a2,..),(b0,b1,b2,..)] -> (a0,b0,a1,b1,a2,b2,..)
+/// ```
+pub struct RoundRobinIterator<A, B> {
```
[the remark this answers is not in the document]
Erigara: It is cyclic in sense that it pop item from the first iterator, then second, then first again and so on.

core/src/queue.rs
Erigara: Right, seen is indeed useless in case of `queue`.

cli/src/torii/routing.rs
```diff
@@ -116,7 +116,7 @@ pub(crate) async fn handle_instructions(
     #[allow(clippy::map_err_ignore)]
     queue
         .push(transaction, &sumeragi.wsv_mutex_access())
-        .map_err(|(tx, err)| {
+        .map_err(|queue::Failure { tx, err }| {
```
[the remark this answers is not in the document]
Erigara: @appetrosyan suggested to add `Failure` both options are fine to me.

config/src/queue.rs
```diff
@@ -18,6 +19,8 @@ pub struct Configuration {
     pub maximum_transactions_in_block: u32,
     /// The upper limit of the number of transactions waiting in the queue.
     pub maximum_transactions_in_queue: u32,
+    /// The upper limit of the number of transactions waiting for more signatures.
+    pub maximum_transactions_in_signature_buffer: u32,
```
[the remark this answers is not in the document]
Erigara: I think we can split `maximum_transactions_in_queue` in half for normal and multisig transactions, but 2 parameters give us more granular control.

core/src/queue.rs
```diff
@@ -108,24 +121,19 @@ impl Queue {
...
-        if tx.is_expired(self.tx_time_to_live) {
+    ) -> Result<MustUse<bool>, Error> {
+        if tx.is_in_future(self.future_threshold) {
+            Err(Error::InFuture)
+        } else if tx.is_expired(self.tx_time_to_live) {
             Err(Error::Expired)
         } else if tx.is_in_blockchain(wsv) {
             Err(Error::InBlockchain)
         } else {
             tx.check_signature_condition(wsv)
```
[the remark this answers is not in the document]
Erigara: Signatures are verified earlier in `AcceptedTransaction::from_transaction` method.

core/src/queue.rs
Erigara: I think there is 2 reasons: 

1. This way we prevent duplicates from being added into the queue. 
2. Gossiping take transactions from the queue so we want transactions to state here until transactions are in blockchain and we don't want to gossip them anymore.

core/src/queue.rs
Erigara: replied under another comment

core/src/queue.rs
Erigara: hmm, but it's `push` method, so transaction is not in the queue yet

core/src/queue.rs
Erigara: I will update documentation, because we record poped transactions in `seen` as well in order to put then back into the queue.

core/src/queue.rs
mversic: ok, then in the `is_pending` we should remove transactions?

core/src/queue.rs
mversic: > This way we prevent duplicates from being added into the queue.

how can we have duplicates in the queue when we use Set/Map

> Gossiping take transactions from the queue so we want transactions to state here until transactions are in blockchain and we don't want to gossip them anymore.

gossiping is different I agree, but any time we move transactions into sumeragi cache they should be removed from the queue immediately

core/src/queue.rs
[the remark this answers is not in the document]
mversic: `check_signature_condition` documentation is incorrect then

config/src/queue.rs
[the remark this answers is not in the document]
mversic: this is the reason why I don't like the approach with two queues. I presume that you've tried using only one queue but it didn't work out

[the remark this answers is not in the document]
Erigara: Actually approach with one queue works quite nice, with one exception multi-signatures could fill up whole queue preventing ordinary transactions from being added.
I'm not quite with current design too so feel free to suggest any idea.

core/src/queue.rs
Erigara: > how can we have duplicates in the queue when we use Set/Map

If we are talking about removing transactions from the queue then them should be removed from `Map` as well because otherwise we would end up in the situation where `Map` is never cleared. 

> gossiping is different I agree, but any time we move transactions into sumeragi cache they should be removed from the queue immediately

Then we would need to extend gossiping to partially take transactions from sumeragi cache partially from queue which sounds clumsily to me.

core/src/queue.rs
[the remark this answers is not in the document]
Erigara: i will update documentation

core/src/queue.rs
Erigara: `is_pending` is predicate function why we should remove transactions inside it?

mversic: ok

core/src/queue.rs
mversic: > Then we would need to extend gossiping to partially take transactions from sumeragi cache partially from queue which sounds clumsily to me.

valid point

core/src/queue.rs
```diff
@@ -137,59 +145,88 @@ impl Queue {
...
-                    .remove(&err_hash)
-                    .expect("Inserted just before match");
-                (err_tx, Error::Full)
-            })
+            match self.check_tx(entry.get(), wsv) {
+                Ok(MustUse(signature_check)) => {
+                    // Transactions are not removed from the queue until expired or committed
+                    seen.push(hash);
+                    if signature_check {
+                        return Some(entry.get().clone());
```
mversic, line 218: it might be worth to consider using an `Arc` instead of cloning since what we have is multiple owners (queue and sumeragi) of the same data. This is just a passing thought
 
 this is what I was after when talking about removing transactions from the queue

config/src/queue.rs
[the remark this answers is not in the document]
mversic: had a lengthy discussion with @Erigara on this matter. I'll make a write up. 2 queue solution remains

mversic (approved):

core/src/queue.rs
mversic: @Erigara says: `i’m not sure that this will work out because we still need cloning becasue block contatin all transactions`

this is a good point, but we have `transaction_cache` containing duplicates of transactions. It might still be worth to consider using `Arc` for that reason. On the other hand, I can hardly believe that `transaction_cache` in sumeragi is making any difference - especially after these changes are merged. We should remove it if possbile

Erigara: Created #3067

core/src/queue.rs
mversic: > is_pending is predicate function why we should remove transactions inside it?

is `is_pending` is also used in `all_transactions` but we don't remove from the queue

Erigara: This doesn't give much. Removing from map will not really free space for new transactions, because queues still will be occupied with hashes of removed transactions.

appetrosyan (approved):

mversic (approved):

PR #3047, opened 2022-12-30
[fix] #3046: Ensure Iroha can start gracefully on empty `./storage`

### LINKED ISSUE #3046: [BUG] Iroha fails if storage directory doesn't exist
### OS and Environment

Arch

### GIT commit hash

fbc7d346620a7dbaef5ae56442e794fce4540b2a

```bash 
IROHA2_CONFIG_PATH="<sample>/configs/peer/config.json" IROHA2_GENESIS_PATH="<sample>/configs/peer/genesis.json" cargo run -p iroha -- --submit-genesis
```

### Actual result

```
2022-12-30T17:33:21.589120Z  INFO iroha_p2p::network: Starting network actor listen_addr=127.0.0.1:1337
Error: 
   0: Failed reading/writing "./storage/kura.lock" from disk
   1: No such file or directory (os error 2)

```

### Expected result

Expect to succeed, instead. 

### Logs in JSON format

<details>
  
  ```json
  ```
</details>

@appetrosyan

### Notes

This is related to the fix of #3027. The implementation is slightly too eager. And needs fallback for the case if the directory doesn't exist.

appetrosyan:

### Description of the Change

Ensure that Iroha can start with a non-existent `./storage` directory. 

Various UI/UX improvements. 

### Issue

Closes #3046 

### Benefits

Graceful init

### Possible Drawbacks

None


cli/src/lib.rs
```diff
@@ -278,7 +275,13 @@ impl Iroha {
...
         );
 
-        let kura = Kura::from_configuration(&config.kura)?;
+        let kura = {
+            Kura::new(
+                config.kura.init_mode,
+                std::path::Path::new(&config.kura.block_store_path),
+                config.kura.debug_output_new_blocks,
+            )?
+        };
```
Erigara, line 284: Couldn't be just: 

```suggestion
        let kura = Kura::new(
            config.kura.init_mode,
            std::path::Path::new(&config.kura.block_store_path),
            config.kura.debug_output_new_blocks,
        )?;
```

core/src/kura.rs
```diff
@@ -362,15 +348,30 @@ impl Unlocked {
...
+                        .read(true)
+                        .write(true)
+                        .create_new(true)
+                        .open(path.clone())
+                    {
+                        Err(Error::IO(e, path))
+                    } else {
+                        Ok(Locked(self.0))
+                    }
+                }
```
Erigara, line 373: Do we need this error handling branch for block store directory creation?
Since then when `lock` function is called block store directory should already be created by `create_files_if_they_do_not_exist`  function.

core/src/kura.rs
```diff
@@ -362,15 +348,30 @@ impl Unlocked {
...
             .create_new(true)
-            .open(self.join(LOCK_FILE_NAME))
+            .open(path.clone())
         {
             match e.kind() {
                 std::io::ErrorKind::AlreadyExists => Err(Error::Locked(self.0)),
-                _ => Err(e.into()),
+                std::io::ErrorKind::NotFound => {
+                    std::fs::create_dir_all(self.0.clone())
+                        .map_err(|e| Error::IO(e, path.clone()))?;
```
Erigara, line 362: Probably should be `MkDir` error, no?

core/src/kura.rs
```diff
@@ -463,37 +465,48 @@ impl<L: Lock> BlockStore<L> {
     ///
     /// # Errors
     /// IO Error.
-    #[allow(clippy::needless_range_loop)]
+    #[warn(clippy::needless_range_loop)] // This could be a problem, because we might be truncating the blocks the wrong way around.
```
Erigara, line 468: What do you mean by this? 

Actually fix for this lint is rather simple, we replace:

```rust 
for index in 0..block_count {
    dest_buffer[index] = (
         // ...
    );
}
```

With: 

```rust
for current_buffer in dest_buffer.iter_mut() {
    *current_buffer = (
         // ...
    );
}
```

cli/src/lib.rs
```diff
@@ -306,7 +309,7 @@ impl Iroha {
         if Self::start_telemetry(telemetry, &config).await? {
             iroha_logger::info!("Telemetry started")
         } else {
-            iroha_logger::error!("Telemetry did not start")
+            iroha_logger::warn!("Telemetry not started")
```
Erigara, line 312: `info` maybe?

cli/src/lib.rs
appetrosyan: Yep

core/src/kura.rs
appetrosyan: That's actually why I needed to create this PR. It doesn't.

core/src/kura.rs
appetrosyan: I mean the logic of the operation is dangerous. And I'm afraid that block wouldn't do because of the coupling. There is a way to fix that lint, however, I wanted to focus on getting the code to work. Given that this wasn't merged, I'll draft and updraft once lints are fixed.

core/src/kura.rs
appetrosyan: No. Because we're checking for if the directory exists and then creating. 

In general I don't like the design of `std::io::error`, and would gladly change this to something more palatable

Erigara: Hmm, okay, i was concerned because in `create_files_if_they_do_not_exist` we return `MkDir` error for the same operation.

core/src/kura.rs
Erigara: Oh got it now, we create block folder store on already `locked` store (so `.create_files_if_they_do_not_exist()` in reality called after `.lock()`).
I'm still a bit concerned that we do the same thing twice.

cli/src/lib.rs
```diff
@@ -67,14 +67,12 @@ impl Default for Arguments {
...
-                    "Default genesis path `{}` has extension, but it should not have one. {e:?}",
-                    GENESIS_PATH
+                    "Default genesis path `{GENESIS_PATH}` has extension, but it should not have one. {e:?}",
                 )
             })),
             config_path: ConfigPath::default(CONFIGURATION_PATH).unwrap_or_else(|e| {
                 panic!(
-                    "Default config path `{}` has extension, but it should not have one. {e:?}",
-                    GENESIS_PATH
+                    "Default config path `{GENESIS_PATH}` has extension, but it should not have one. {e:?}",
```
mversic, line 75: @Arjentix I believe you've made a change recently in a case like this where you were able to make it a `lazy_static`?

cli/src/torii/utils.rs
```diff
@@ -49,10 +49,13 @@ pub mod body {
...
-        if let Some(error) = rejection.find::<VersionError>() {
-            return Ok(error.into_response());
-        }
-        Err(rejection)
+        rejection
+            .find::<VersionError>()
+            .map(|e| {
+                // generally don't implement a trait for &Error
+                e.into_response()
+            })
```
mversic, line 57:
```suggestion
            .map(SomeTrait::into_response)
```

maybe?

core/src/kura.rs
```diff
@@ -692,8 +753,11 @@ type Result<T, E = Error> = std::result::Result<T, E>;
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     /// Generic IO error
-    #[error("Failed reading/writing from disk")]
-    IO(#[from] std::io::Error),
+    #[error("Failed reading/writing {_1:?} from disk")]
+    IO(#[source] std::io::Error, PathBuf),
+    /// Failed to create the directory
+    #[error("Failed to create the directory")]
+    MkDir(std::io::Error, PathBuf),
```
mversic, line 760: isn't `MkDir` just a subset of `IO`? IMO it should be

core/src/kura.rs
appetrosyan: > I'm still a bit concerned that we do the same thing twice.

These things are accidentally the same. In reality there will be other files all of which need to be handled. I agree that maybe factoring it out into a function might be good, but given that it's just a sequence of operations, I don't necessarily think it'd help readability as much as obscure the intention.

core/src/kura.rs
appetrosyan: Good point. I didn't notice that.

cli/src/lib.rs
appetrosyan: The logic is as follows: 
- Use `trace` for showing information about normal operation, specifically which functions were called and returned from. 
- Use `debug` for information that you'd `println` if you were debugging a misbehaviour. Diagnostics, information status updates. E.g. if debugging deadlocks, calling `debug!("lock X locked")`, is a sane option. Don't overdo it, because `debug` statements add a conditional operation sometimes the branch predictor doesn't optimise them away. 
- Use `info` for regular feedback on normal processes. We have too much there currently. This information is supposed to tell node administrators what's going on, but not much else. It's only useful during startup and shutdown, because everything else falls either into debugging or error reporting. 
- Use `warn` for things that *could* be errors and node administrators should investigate, but which can also be considered part of the normal operation. Example: you could have not requested telemetry be initialised. Or you did request it, and it couldn't be initialised. The latter point should not stop Iroha from working, but is worth investigating. 
- Use `error` for things that require manual intervention from the node administrator. For example, disk is nearing full. Communication with an actor is supposed to occur, but it didn't. 

If we follow this, you can attach an alert to finding a `warn` or higher message. What's more important, to improve perofmrnace we set the logging to `error` in real deployments. So if you need the telemetry having the message in `warn` could tell you that it didn't start, because e.g. you forgot to enable it. Usually, you only need to disable telemetry when you need performance, like `unsafe` in Rust, so producing a warning makes sense.

cli/src/torii/utils.rs
appetrosyan: I'm not sure why this change shows up in my PR, maybe it was touched by LSP. But I agree. Maybe `ReponseExt::into_reponse`?

core/src/kura.rs
appetrosyan: I don't mind them being together, but I found that working with `io::Error` is a bit difficult. Pattern matching on this error is natural. Casting the trait object inside the `io::Error` is a lot more cumbersome and AFAIK not meant to be done at all.

appetrosyan: Also, this line shows up because of the conflicts. I'll need to rebase so it disappears.

cli/src/lib.rs
appetrosyan: Shows up due to conflict. Will disappear after rebase.

cli/src/lib.rs
Erigara: Yeah, i agree, thank you for clarification!

core/src/kura.rs
```diff
@@ -463,40 +481,50 @@ impl<L: Lock> BlockStore<L> {
...
+            .seek(SeekFrom::Start(start_location))
+            .map_err(|e| Error::IO(e, path.clone()))?;
         let mut buffer = [0; core::mem::size_of::<u64>()];
         // (start, length), (start,length) ...
-        for index in 0..block_count {
-            dest_buffer[index] = (
-                {
-                    index_file.read_exact(&mut buffer)?;
+        for current_buffer in dest_buffer.iter_mut() {
+            *current_buffer = BlockIndices {
```
Erigara, line 514: Nice, this place looks better with `BlockIndices`.

core/src/kura.rs
```diff
@@ -455,6 +464,15 @@ impl BlockStore<Unlocked> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+/// Lightweight wrapper for block indices in the block index file
+pub struct BlockIndices {
```
Erigara, line 469: Why not `BlockIndex`?

core/src/sumeragi/main_loop.rs
```diff
@@ -650,6 +650,7 @@ fn should_terminate(shutdown_receiver: &mut tokio::sync::oneshot::Receiver<()>)
     }
 }
 
+#[instrument(skip_all)]
 /// Execute the main loop of [`SumeragiWithFault`]
 #[instrument(skip(sumeragi, state, genesis_network, shutdown_receiver))]
```
Erigara, line 655: Shouldn't we remove this attribute then?

Erigara (approved):

core/src/kura.rs
```diff
@@ -362,15 +355,30 @@ impl Unlocked {
...
             .write(true)
             .create_new(true)
-            .open(self.join(LOCK_FILE_NAME))
+            .open(path.clone())
         {
             match e.kind() {
                 std::io::ErrorKind::AlreadyExists => Err(Error::Locked(self.0)),
-                _ => Err(e.into()),
+                std::io::ErrorKind::NotFound => {
+                    std::fs::create_dir_all(self.0.clone())
```
QuentinI, line 368: Can't we just call this pre-emptively?

core/src/kura.rs
```diff
@@ -720,6 +806,38 @@ mod tests {
...
+            start: 0,
+            length: 0,
+        }; N];
+        for idx in 0..value.len() {
+            ret[idx] = value[idx].into();
+        }
+        ret
+    }
+
+    impl PartialEq for BlockIndices {
```
QuentinI, line 820: Why not derive it?

core/src/kura.rs
```diff
@@ -720,6 +806,38 @@ mod tests {
...
+        fn eq(&self, other: &Self) -> bool {
+            self.start == other.start && self.length == other.length
+        }
+    }
+
+    impl PartialEq<(u64, u64)> for BlockIndices {
```
QuentinI, line 826: Is this used anywhere?

QuentinI (commented):
Nice for drive-by refactorings, Kura reads a bit too much like C for my liking.

core/src/kura.rs
appetrosyan: It doesn't give us much and falls in line with the "apologise if wrong, instead of asking" paradigm.

core/src/kura.rs
appetrosyan: Because it's under `cfg(test)` we don't actually need this outside of syntactic sugar.

core/src/kura.rs
appetrosyan: Yes. `cfg(test)` like before.

core/src/sumeragi/main_loop.rs
appetrosyan: Conflict artefact.

QuentinI (dismissed):

Erigara (approved):

cli/src/lib.rs
mversic: since you went through the trouble of explaining this in great detail, it would be a good idea to add this section to the README

mversic (approved):

core/src/lib.rs
```diff
@@ -1,4 +1,6 @@
 //! Iroha — A simple, enterprise-grade decentralized ledger.
+
+#![allow(incomplete_features)] // This feature is minor and breakage to be expected is going to be minimal.
 #![feature(adt_const_params)]
```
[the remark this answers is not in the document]
mversic: additional possible reason is that with `nightly` compiler I have a feeling that build cache is cleared much more often. I am not 100% sure of this, since it's still rather fast on M1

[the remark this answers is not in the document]
appetrosyan: M1 is faster than it has any right to be. We shouldn't expect people to have it though.

cli/src/lib.rs
Erigara: CONTRIBUTING probably

Issue #3053, opened 2023-01-10
Fix lints on Mac M1


Arjentix:
Our clippy lints check passes in CI, but fails on my Mac M1.

Need to fix it.

PR #3054, opened 2023-01-10
[refactor] #3053: Fix clippy lints

Arjentix:
### Description of the Change

Fix some clippy and `cargo check` lints

### Issue

- #3053 

### Benefits

No Mac M1 users have no lint errors hell

### Possible Drawbacks

None

### DIFF
--- a/config/src/path.rs
+++ b/config/src/path.rs
@@ -37,16 +37,16 @@ pub type Result<T> = std::result::Result<T, ExtensionError>;
 #[derive(Debug, Clone)]
 enum InnerPath {
     Default(PathBuf),
-    User(PathBuf),
+    UserProvided(PathBuf),
 }
 
 /// Wrapper around path to config file (i.e. config.json, genesis.json).
 ///
 /// Provides abstraction above user-provided config and default ones.
 #[derive(Debug, Clone)]
-pub struct ConfigPath(InnerPath);
+pub struct Path(InnerPath);
 
-impl ConfigPath {
+impl Path {
     /// Construct new [`Path`] from the default `path`.
     ///
     /// # Panics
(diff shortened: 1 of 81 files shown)


data_model/src/transaction.rs
```diff
@@ -414,6 +416,7 @@ impl Txn for Transaction {
     FfiType,
     IntoSchema,
 )]
+#[allow(clippy::use_debug)]
```
Erigara, line 419: We have module level `!#[allow(clippy::use_debug)]` do we need this allow for individual struct.

Erigara (dismissed):
LGTM

config/src/path.rs
```diff
@@ -1,87 +1,85 @@
...
-        Ok(Self(Default(path)))
+    /// # Panics
+    ///
+    /// Panics if `path` contains an extension.
+    #[allow(clippy::panic)]
+    pub fn default(path: &'static std::path::Path) -> Self {
+        assert!(
+            path.extension().is_none(),
+            "Default configuration path should have no extension"
+        );
```
mversic, line 60: While I understand that it reduces boilerplate downstream, I'm not really sure I support this.

mversic (approved):

config/src/path.rs
Arjentix: I did this because it's a programmer error and we want to know about this as far as we can

mversic: nvm

data_model/src/transaction.rs
Arjentix: Indeed. Fixed

Erigara (approved):

mversic (approved):

Issue #2123, opened 2022-04-19
Use enum for `PublicKey`/`PrivateKey` digest function


mversic:
Both `PublicKey` and `PrivateKey` store their respective digest functions as a string. They should store it as an `Algorithm` enum provided in the `iroha_crypto` crate.

Tezas-6174:
Can I please work on this issue?

Retssaze:
Thanks, very good one. It looks like I got what it is needed to do

appetrosyan:
Good. Let me know if you want to set up a meeting and discuss the work in more detail.

Retssaze:
Thank you for your kidness

Retssaze:
Hi Alex,

How are you doing? What about our Meeting?

K.S.

Retssaze:
It looks like I did something

PR #3055, opened 2023-01-10
[refactor] #2123: use `Algorithm` for storing digest function

Retssaze:
Signed-off-by: Retssaze <79268790+Retssaze@users.noreply.github.com>

### DIFF
--- a/.github/workflows/iroha2-dev.yml
+++ b/.github/workflows/iroha2-dev.yml
@@ -17,7 +17,7 @@ env:
 # `ubuntu-latest` and not `[self-hosted, Linux]`.
 jobs:
   dockerhub:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Linux, iroha2-dev-push]
     container:
       image: 7272721/i2-ci:nightly
     steps:
@@ -46,9 +46,8 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
-# FIXME: This workflow never worked properly. Should we remove it? 
   load-rs:
-    if: false
+    continue-on-error: true
     runs-on: ubuntu-latest
     container:
       image: 7272721/i2-ci:nightly


Erigara (changes requested):
Thank you for your contribution. 
I have few suggestions: 
1. Please use `rustfmt` you can run with command: `cargo +nightly-2022-12-22 fmt --all`
2. Fix lints, you can run linter by following command: `cargo +nightly-2022-12-22 lints clippy --workspace --tests --benches --examples --all-features`
3. Fix code in tests to compile and pass.
4. We use particular format for commit messages. In your case it would be something like this: ```[refactor] #2123: Use enum `Algorithm` for `PublicKey/PrivateKey` digest function```.
5. Instead of creating big pull request name you can fill template provided when you are open pr and pr name can match commit name. 
6. IMO we should add `Deserialize/Serialize` and `Decode/Encode` to `Algorithm` directly instead of doing conversion `ConstString -> Algorithm` inside of manual `PrivateKey/PublicKey` `De/Serialize` or `Encode/Decode` implementations.

Erigara:
@Retssaze ping me when PR will be ready for re-review.

Retssaze:
Dear @Erigara, please tell where to find this +nightly-2022-12-22?

ilchu:
> Dear @Erigara, please tell where to find this +nightly-2022-12-22?

This means that you install this particular version of the nightly toolchain (e.g. by calling `rustup toolchain add nightly-2022-12-22 --profile minimal` from the CLI) and then this flag should work as expected.

appetrosyan:
@Retssaze let us know when you're ready to re-review the PR by undrafting it.

Retssaze:
Dear @appetrosyan, please feel free to tell the feedback

crypto/src/lib.rs
```diff
@@ -69,7 +69,7 @@ impl std::error::Error for NoSuchAlgorithm {}
 
 ffi::ffi_item! {
     /// Algorithm for hashing
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, Display, FfiType)]
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Display, FfiType, IntoSchema, Serialize)]
```
Erigara, line 72: Current `Serialize/Deserialize` and `Encode/Decode` of Algorithm is not consistent.

You can check this by running this test: 

```rust

    #[test]
    fn key_pair_serialize_deserialize_consistent() {
        for algorithm in [
            Algorithm::Ed25519,
            Algorithm::Secp256k1,
            Algorithm::BlsNormal,
            Algorithm::BlsSmall,
        ] {
            let key_pair = KeyPair::generate_with_configuration(
                KeyGenConfiguration::default().with_algorithm(algorithm),
            )
            .expect("Failed to generate key pair");
            assert_eq!(
                key_pair,
                serde_json::to_string(&key_pair)
                    .and_then(|key_pair| serde_json::from_str(&key_pair))
                    .expect("Failed to de/serialize key")
            )
        }
    }
``` 

What i would suggest is to add direct implementations of `Deserialize/Serialize` and `Encode/Decode` to the `Algorithm` like that:

```rust
impl Decode for Algorithm {
    // ...
}

impl Encode for Algorithm {
    // ...
}

impl Serialize for Algorithm {
     // ...
}

impl Deserialize for Algorithm {
    // ...
}
```

This way we can simplify implementation of `Deserialize` and `Decode` for `PublicKey/PrivateKey`.

Also we need to add tests to ensure that `Serialize/Deserialize` and `Encode/Decode` of `Algorithm` is indeed consistent (you can use provided test as example).

Retssaze:
Should a list of methods for both keys be identical or differ?

Erigara:
@Retssaze 

> Should a list of methods for both keys be identical or differ?

I think differs because `PrivateKey` doesn't transferred between peers, so `Encode/Decode` shouldn't be implemented for them.

Retssaze:
Sure

crypto/src/lib.rs
appetrosyan: I'd recommend you add this test directly to the crate @Retssaze. Normally we don't need to check that what's serialised is also deserialised accordingly, but in this particular case, given the shift to manually implemented `Serialize`  this is a necessity.

Retssaze:
Dear Sirs, please assist me in solving issues with "workflows", since they are not related to code itself, and I cannot compile release with #![feature]` may not be used on the stable release channel  --> core/src/lib.rs:2:1

Erigara:
This mean that you should enable nightly toolchain:

```bash
rustup toolchain add nightly-2022-12-22 --profile minimal
rustup default nightly-2022-12-22
```

mversic:
@Retssaze rebase the branch and it should be clear to merge. Thank you for your work

crypto/src/lib.rs
```diff
@@ -516,14 +591,8 @@ impl<'de> Deserialize<'de> for PublicKey {
 
 impl Decode for PublicKey {
```
mversic, line 592: I believe that you can now derive `Decode`, `Deserialize` and maybe even `Serialize` for `PublicKey`/`PrivateKey`/`KeyPair`

appetrosyan:
One unit test is failing. You should add the `iroha_crypto::Algorithm` to the big list in `tools/parity_scale_decoder/src/generate_map.rs` into `fn generate_map()`. 

You also have an unnecessary truncating cast with `as u8`, instead of a fallible `try_into()`, which is the preferred method. 

Additionally the trivial cast needs to be addressed. 

I'd recommend you consider deriving the traits if you can, because the implementation is really not something you should do by hand unless it's absolutely necessary (and in this case it isn't).

appetrosyan:
One final thing that needs to be done before merging. You should've read the `Contributing.md` guide, which should explain how to create commits. As it is now, you have 44 commits for one small feature. We cannot accept the PR in that form and given that Signed your commits, we cannot rebase the commits for you. 

So what you should do is you should consider squashing the commits on your end. (**HINT**: `git rebase -I HEAD~44`) and give the commit the same name as the title of this Pull request. 

Finally, please make sure that all the tests pass.

PR #3058, opened 2023-01-11
[ci]: Adding five more self-hosted runners

BAStos525:

### Description of the Change
1. Add four more self-hosted runners for `I2::Dev::Tests` workflow.
2. Add a new self-hosted runner for `I2::Dev::Publish` workflow.
3. Make `load-rs` job inside `I2::Dev::Publish` workflow conditional.

### Issue
1. We can not run a two `I2::Dev::Tests` workflows in parallel on self-hosted runners (increase their amount from 4 to 8 for four tests jobs).
2. A `dockerhub` job  inside `I2::Dev::Publish` workflow takes a lot of time.
3. `I2::Dev::Publish` workflow fails if `load-rs` job was failed or `load-rs` job doesn't run.

### Benefits

1. Now a two`I2::Dev::Tests` workflows could be run in parallel.
2. Increase `dockerhub` job  inside `I2::Dev::Publish` speed.
3. Do not failI `2::Dev::Publish` workflow if `load-rs` job was fails.

### Possible Drawbacks

None.

### DIFF
--- a/.github/workflows/iroha2-dev.yml
+++ b/.github/workflows/iroha2-dev.yml
@@ -17,7 +17,7 @@ env:
 # `ubuntu-latest` and not `[self-hosted, Linux]`.
 jobs:
   dockerhub:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Linux, iroha2-dev-push]
     container:
       image: 7272721/i2-ci:nightly
     steps:
@@ -46,9 +46,8 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
-# FIXME: This workflow never worked properly. Should we remove it? 
   load-rs:
-    if: false
+    continue-on-error: true
     runs-on: ubuntu-latest
     container:
       image: 7272721/i2-ci:nightly


.github/workflows/iroha2-dev.yml
```diff
@@ -46,9 +46,8 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
-# FIXME: This workflow never worked properly. Should we remove it?
```
BAStos525, line 49: Actually there is not problem in `load-rs` gob here. Since it just trigger `dispatch` workflow from other repository, that `oad-rs-ci:dev` workflow in [soramitsu/iroha2-longevity-load-rs](https://github.com/soramitsu/iroha2-longevity-load-rs/actions/runs/3765611744/jobs/6401283080) should be fixed.

BAStos525 (commented):
```
  pull_request:
    branches: [iroha2-dev]
    paths:
      - '.github/workflows/**.yml'
```
Could you please clarify why this trigger was added? If I'm not mistaken, we can not share Actions secrets on `Pull Request` event.

appetrosyan:
@BAStos525 The idea is that the workflow must produce working `docker` images. While it would not be able to publish, the point is to confirm that it can produce the container and would be able to push if it needed to. 

This was meant for debugging and I left it in, because we might need to experiment with `cargo chef`

BAStos525:
> @BAStos525 The idea is that the workflow must produce working `docker` images. While it would not be able to publish, the point is to confirm that it can produce the container and would be able to push if it needed to.
> 
> This was meant for debugging and I left it in, because we might need to experiment with `cargo chef`

Okay, thank you for clarification. We still keep the branch with `cargo chef` experiment within the closed related PR.
By the way, this PR is about just to add a self-hosted runner for `iroha2-dev` image build and push.

appetrosyan (approved):

6r1d (approved):
Approving, but please note auth is failing [here](https://github.com/hyperledger/iroha/actions/runs/3892333160/jobs/6643624816)

PR #3060, opened 2023-01-12
[refactor]: simplify sumeragi

mversic:

### Description of the Change

I believe you will find it much more easier to understand now
I also left some comments in the code

The hanging issue still persists:
1. `Failed to push into queue err=InBlockchain` is observed:
```
2023-01-12T22:42:41.821378Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:54960}: iroha::torii::routing: Failed to push into queue tx_hash=d543e665272d74fd5f6733d1ca01869c78ac8633e7b5a49c36b563fb6baa51b9 err=InBlockchain
2023-01-12T22:42:41.821378Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:54955}: iroha::torii::routing: Failed to push into queue tx_hash=5b042f8b52a4d9cc1fca8f57d237cb8c5ff8f92aa58333f61314eb859dc5ae37 err=InBlockchain
2023-01-12T22:42:41.821378Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:54966}: iroha::torii::routing: Failed to push into queue tx_hash=5a272b4b5f92ad980bb23a7632672ebcf1ba076ca56cd7b967cbd88b182d90f7 err=InBlockchain
2023-01-12T22:42:41.821423Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:54960}: warp::filters::trace: unable to serve request (client error) status=400 error=None
2023-01-12T22:42:41.821423Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:54955}: warp::filters::trace: unable to serve request (client error) status=400 error=None
2023-01-12T22:42:41.821456Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:54966}: warp::filters::trace: unable to serve request (client error) status=400 error=None
2
```
2. And is followed by endless view changes where view change index increases:
```
2023-01-12T21:55:55.603155Z  WARN run: iroha_core::sumeragi::main_loop: Block not committed in due time, requesting view change... role=ObservingPeer
2023-01-12T21:55:55.607356Z  WARN run: iroha_core::sumeragi::main_loop: Block not committed in due time, requesting view change... role=ProxyTail
2023-01-12T21:55:55.610777Z  WARN run: iroha_core::sumeragi::main_loop: Block not committed in due time, requesting view change... role=Leader
2023-01-12T21:55:55.615403Z  WARN run: iroha_core::sumeragi::main_loop: Block not committed in due time, requesting view change... role=ValidatingPeer
2023-01-12T21:55:55.617158Z ERROR run: iroha_core::sumeragi::main_loop: No consensus, updating view change ... current_view_change_index=4
2023-01-12T21:55:55.618910Z ERROR run: iroha_core::sumeragi::main_loop: No consensus, updating view change ... current_view_change_index=4
2023-01-12T21:55:55.622082Z ERROR run: iroha_core::sumeragi::main_loop: No consensus, updating view change ... current_view_change_index=4
2023-01-12T21:55:55.625818Z ERROR run: iroha_core::sumeragi::main_loop: No consensus, updating view change ... current_view_change_index=4
2
```

This might also mean that there are no new incoming transactions

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -303,10 +303,13 @@ impl SignedBlock {
     /// Verify signatures and commit block to the store.
     ///
     /// # Errors
-    /// - If signatures verification fails
+    ///
+    /// Not enough signatures
     #[inline]
-    pub fn commit(self, topology: &Topology) -> Result<CommittedBlock, eyre::Report> {
-        topology
+    pub fn commit(mut self, topology: &Topology) -> Result<CommittedBlock, (Self, eyre::Report)> {
+        let verified_signatures = self.retain_verified_signatures();
+
+        if topology
             .filter_signatures_by_roles(
                 &[
                     Role::ValidatingPeer,
@@ -314,14 +317,16 @@ impl SignedBlock {
                     Role::ProxyTail,
                     Role::ObservingPeer,
                 ],
-                self.verified_signatures(),
+                verified_signatures,
             )
             .len()
-            .ge(&topology.min_votes_for_commit())
-            .then_some(())
-            .ok_or_else(|| {
-                eyre!("The block doesn't have enough valid signatures to be committed.")
-            })?;
+            .lt(&topology.min_votes_for_commit())
+        {
+            return Err((
+                self,
+                eyre!("The block doesn't have enough valid signatures to be committed."),
+            ));
+        }
 
         Ok(self.commit_unchecked())
     }
@@ -360,10 +365,11 @@ impl SignedBlock {
             ))
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
+    /// Return signatures that are verified with the `hash` of this block, removing all other
+    /// signatures.
     #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
-        self.signatures.verified_by_hash(self.hash())
+    pub fn retain_verified_signatures(&mut self) -> impl Iterator<Item = &SignatureOf<Self>> {
+        self.signatures.retain_verified_by_hash(self.hash())
     }
 
     /// Create dummy `ValidBlock`. Used in tests
@@ -463,11 +469,11 @@ impl VersionedCandidateBlock {
         self.as_v1().hash().transmute()
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
+    /// Return signatures that are verified with the `hash` of this block, removing all other signatures.
     #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
-        self.as_v1()
-            .verified_signatures()
+    pub fn retain_verified_signatures(&mut self) -> impl Iterator<Item = &SignatureOf<Self>> {
+        self.as_mut_v1()
+            .retain_verified_signatures()
             .map(SignatureOf::transmute_ref)
     }
 
@@ -480,7 +486,7 @@ impl VersionedCandidateBlock {
         self,
         transaction_validator: &TransactionValidator,
         wsv: &WorldStateView,
-        latest_block: &Option<HashOf<VersionedCommittedBlock>>,
+        latest_block: Option<HashOf<VersionedCommittedBlock>>,
         block_height: u64,
     ) -> Result<SignedBlock, eyre::Report> {
         self.into_v1()
@@ -511,10 +517,10 @@ impl CandidateBlock {
         HashOf::new(&self.header).transmute()
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
+    /// Return signatures that are verified with the `hash` of this block, removing all other signatures.
     #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
-        self.signatures.verified_by_hash(self.hash())
+    pub fn retain_verified_signatures(&mut self) -> impl Iterator<Item = &SignatureOf<Self>> {
+        self.signatures.retain_verified_by_hash(self.hash())
     }
 
     /// Check if there are no transactions in this block.
@@ -547,7 +553,7 @@ impl CandidateBlock {
         self,
         transaction_validator: &TransactionValidator,
         wsv: &WorldStateView,
-        latest_block: &Option<HashOf<VersionedCommittedBlock>>,
+        latest_block: Option<HashOf<VersionedCommittedBlock>>,
         block_height: u64,
     ) -> Result<SignedBlock, eyre::Report> {
         if self.is_empty() {
@@ -558,7 +564,7 @@ impl CandidateBlock {
             bail!("Block has committed transactions");
         }
 
-        if latest_block != &self.header.previous_block_hash {
+        if latest_block != self.header.previous_block_hash {
             bail!(
                 "Mismatch between the actual and expected hashes of the latest block. Expected: {:?}, actual: {:?}",
                 latest_block,
@@ -731,11 +737,12 @@ impl VersionedCommittedBlock {
         &self.as_v1().header
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
+    /// Return signatures that are verified with the `hash` of this block
     #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
+    pub fn signatures(&self) -> impl IntoIterator<Item = &SignatureOf<Self>> {
         self.as_v1()
-            .verified_signatures()
+            .signatures
+            .iter()
             .map(SignatureOf::transmute_ref)
     }
 
@@ -800,12 +807,6 @@ impl CommittedBlock {
     #[inline]
     pub fn hash(&self) -> HashOf<Self> {
         HashOf::new(&self.header).transmute()
-    }
-
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
-    #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
-        self.signatures.verified_by_hash(self.hash())
     }
 }
 
@@ -918,33 +919,30 @@ impl VersionedCandidateCommittedBlock {
         &self.as_v1().header
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
-    #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
-        self.as_v1()
-            .verified_signatures()
-            .map(SignatureOf::transmute_ref)
-    }
-
     /// Revalidate transaction hashes, verify signatures and produce [`VersionedCommittedBlock`]
     ///
     /// # Errors
     /// - If transaction hashes don't match the hashes stored in the block header
     /// - If signatures verification fails
-    pub fn revalidate(self, topology: &Topology) -> Result<VersionedCommittedBlock, eyre::Report> {
+    pub fn revalidate(
+        self,
+        topology: &Topology,
+    ) -> Result<VersionedCommittedBlock, (Self, eyre::Report)> {
         self.into_v1()
             .revalidate(topology)
             .map(VersionedCommittedBlock::from)
+            .map_err(|err| (err.0.into(), err.1))
     }
 
     /// Revalidate transaction hashes and produce [`VersionedCommittedBlock`]
     ///
     /// # Errors
     /// - If transaction hashes don't match the hashes stored in the block header
-    pub fn revalidate_hashes(self) -> Result<VersionedCommittedBlock, eyre::Report> {
+    pub fn revalidate_hashes(self) -> Result<VersionedCommittedBlock, (Self, eyre::Report)> {
         self.into_v1()
             .revalidate_hashes()
             .map(VersionedCommittedBlock::from)
+            .map_err(|err| (err.0.into(), err.1))
     }
 }
 
@@ -973,10 +971,10 @@ impl CandidateCommittedBlock {
         HashOf::new(&self.header).transmute()
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
+    /// Return signatures that are verified with the `hash` of this block, removing all other signatures.
     #[inline]
-    pub fn verified_signatures(&self) -> impl Iterator<Item = &SignatureOf<Self>> {
-        self.signatures.verified_by_hash(self.hash())
+    pub fn retain_verified_signatures(&mut self) -> impl Iterator<Item = &SignatureOf<Self>> {
+        self.signatures.retain_verified_by_hash(self.hash())
     }
 
     /// Revalidate transaction hashes, verify signatures and produce [`CommittedBlock`]
@@ -984,8 +982,13 @@ impl CandidateCommittedBlock {
     /// # Errors
     /// - If transaction hashes don't match the hashes stored in the block header
     /// - If signatures verification fails
-    pub fn revalidate(self, topology: &Topology) -> Result<CommittedBlock, eyre::Report> {
-        topology
+    pub fn revalidate(
+        mut self,
+        topology: &Topology,
+    ) -> Result<CommittedBlock, (Self, eyre::Report)> {
+        let verified_signatures = self.retain_verified_signatures();
+
+        if topology
             .filter_signatures_by_roles(
                 &[
                     Role::ValidatingPeer,
@@ -993,14 +996,16 @@ impl CandidateCommittedBlock {
                     Role::ProxyTail,
                     Role::ObservingPeer,
                 ],
-                self.verified_signatures(),
+                verified_signatures,
             )
             .len()
-            .ge(&topology.min_votes_for_commit())
-            .then_some(())
-            .ok_or_else(|| {
-                eyre!("The block doesn't have enough valid signatures to be committed.")
-            })?;
+            .lt(&topology.min_votes_for_commit())
+        {
+            return Err((
+                self,
+                eyre!("The block doesn't have enough valid signatures to be committed."),
+            ));
+        }
         self.revalidate_hashes()
     }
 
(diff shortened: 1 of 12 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -669,445 +669,285 @@ pub fn run<F>(
...
+    state: &State,
+    mut block: VersionedCandidateBlock,
+) -> Option<VotingBlock> {
+    let addr = &sumeragi.peer_id.address;
+    let role = state.current_topology.role(&sumeragi.peer_id);
+    trace!(%addr, %role, block_hash=%block.hash(), "Block received, voting...");
+
+    if state
+        .current_topology
+        .filter_signatures_by_roles(&[Role::Leader], block.retain_verified_signatures())
```
mversic, line 908: do you think there can be any problem if unverified signatures are removed from the block?

core/src/block_sync.rs
```diff
@@ -240,10 +240,10 @@ pub mod message {
                 }
                 Message::ShareBlocks(ShareBlocks { blocks, .. }) => {
                     use crate::sumeragi::message::{BlockSyncUpdate, Message, MessagePacket};
-                    for block in blocks {
+                    for block in blocks.clone() {
                         block_sync.sumeragi.incoming_message(MessagePacket::new(
                             Vec::new(),
-                            Message::BlockSyncUpdate(BlockSyncUpdate::from(block.clone())),
+                            Message::BlockSyncUpdate(BlockSyncUpdate { block }),
```
mversic, line 246: is this some leftover from the actor system? why do we send one by one block to sumeragi?

Erigara (commented):
Good refactoring

core/src/sumeragi/main_loop.rs
appetrosyan: This should be fine because we don't really do multisignature yet. When we do the mechanism is going to be different, so we can safely ignore it.

core/src/block_sync.rs
appetrosyan: I tried refactoring it under the assumption that it was a leftover. It ended up costing around 20% in performance. I suspect it's because of the prevalence of `Mutex` es in the `main_loop` which means that blocking for a significant amount of time can pessimism performance greatly. 

With that said, we might want to investigate other synchronisation methods. I'm willing to wager that Atomic pointers with an arena allocator would do very well in terms of performance.

core/src/sumeragi/main_loop.rs
```diff
@@ -463,123 +465,126 @@ fn handle_role_agnostic_messages<F>(
...
-                        .take()
-                        .expect("Voting block is `Some`")
-                        .block;
-                    block.signatures.extend(signatures);
-                    block
-                } else {
-                    error!("The hash of the committed block does not match the hash of the block stored by the peer.");
-                    return;
-                };
+                if let Some(mut voted_block) = voting_block.take() {
```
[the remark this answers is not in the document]
appetrosyan: `let else` to the rescue.

core/src/sumeragi/main_loop.rs
```diff
@@ -669,445 +667,287 @@ pub fn run<F>(
...
+
+fn add_signatures<const EXPECT_VALID: bool>(
+    block: &mut VotingBlock,
+    signatures: impl IntoIterator<Item = SignatureOf<SignedBlock>>,
+) {
+    for signature in signatures {
+        if let Err(err) = block.block.add_signature(signature) {
+            let err_msg = "Signature not valid";
+
+            if EXPECT_VALID == true {
```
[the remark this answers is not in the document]
appetrosyan: Defensive programming habit. Useful in C, next to useless in Rust. Still I don't mind it, given it's no less readable.

core/src/block.rs
```diff
@@ -303,25 +303,30 @@ impl SignedBlock {
     /// Verify signatures and commit block to the store.
     ///
     /// # Errors
-    /// - If signatures verification fails
+    ///
+    /// If there isn't enough valid signatures
```
appetrosyan, line 307:
```suggestion
    /// - Not enough signatures
    /// - Not all the [required](TODO, link to tutorial on multisig) signatures are valid
```

core/src/block.rs
```diff
@@ -360,10 +365,10 @@ impl SignedBlock {
             ))
     }
 
-    /// Return the signatures (as `payload`) that are verified with the `hash` of this block.
+    /// Return signatures that are verified with the `hash` of this block, remove all other.
```
appetrosyan, line 368:
```suggestion
    /// Return signatures that are verified with the `hash` of this block, removing all other signatures.
```

core/src/block.rs
```diff
@@ -918,33 +918,30 @@ impl VersionedCandidateCommittedBlock {
...
     /// Revalidate transaction hashes and produce [`VersionedCommittedBlock`]
     ///
     /// # Errors
     /// - If transaction hashes don't match the hashes stored in the block header
-    pub fn revalidate_hashes(self) -> Result<VersionedCommittedBlock, eyre::Report> {
+    pub fn revalidate_hashes(self) -> Result<VersionedCommittedBlock, (Self, eyre::Report)> {
         self.into_v1()
             .revalidate_hashes()
             .map(VersionedCommittedBlock::from)
+            .map_err(|err| (err.0.into(), err.1))
```
appetrosyan, line 944: Maybe we need to provide this information some other way. A new type perhaps? Tuple struct?

core/src/block.rs
```diff
@@ -973,34 +970,41 @@ impl CandidateCommittedBlock {
...
             )
             .len()
-            .ge(&topology.min_votes_for_commit())
-            .then_some(())
-            .ok_or_else(|| {
-                eyre!("The block doesn't have enough valid signatures to be committed.")
-            })?;
+            .lt(&topology.min_votes_for_commit())
+        {
+            return Err((
```
appetrosyan, line 1003: Consider `eyre::bail!((self, eyre!("The block doesn't have enough valid signatures to be committed")`. 

I also think that maybe a new error enum would do us a lot of good here. This message is repeated quite a lot, and `eyre::Report` discards any possibility of handling the error downstream.

core/src/tx.rs
```diff
@@ -509,7 +514,7 @@ mod tests {
         let mut chain = err.chain();
         assert_eq!(
             chain.next().unwrap().to_string(),
-            "Failed to accept transaction"
+            "Limits verification failed"
```
appetrosyan, line 517: This is a `ui` test. I feel we shouldn't mix them in with the regular tests. Perhaps this could be improved by using an `enum` instead of `&str`

crypto/src/signature.rs
```diff
@@ -476,21 +476,15 @@ impl<T> SignaturesOf<T> {
         self.signatures.insert(SignatureWrapperOf(signature));
     }
 
-    /// Returns signatures that have passed verification.
+    /// Returns signatures that have passed verification, remove all other.
```
appetrosyan, line 479:
```suggestion
    /// Return signatures that have passed verification, remove all other.
```

Missed a spot.

core/src/sumeragi/mod.rs
```diff
@@ -342,18 +342,23 @@ pub const TELEMETRY_INTERVAL: Duration = Duration::from_secs(5);
 #[non_exhaustive]
 pub struct VotingBlock {
     /// At what time has this peer voted for this block
-    pub voted_at: Duration,
+    pub voted_at: Instant,
```
appetrosyan, line 345: This problem should go away after we use the formally verified replacement for `chrono`.

core/src/sumeragi/main_loop.rs
```diff
@@ -48,9 +44,11 @@ where
     pub events_sender: EventsSender,
     /// The world state view instance that is used in public contexts
     pub wsv: Mutex<WorldStateView>,
-    /// TODO: good description
+    /// Time by which a newly created block should be committed. Prevents malicious nodes
+    /// to stall the network by not participating in consensus
```
appetrosyan, line 48:
```suggestion
    /// from stalling the network by not participating in consensus
```

core/src/sumeragi/main_loop.rs
```diff
@@ -463,123 +465,129 @@ fn handle_role_agnostic_messages<F>(
...
 
 #[allow(clippy::too_many_arguments)]
-fn compare_view_change_index_and_block_height_to_old<F>(
-    sumeragi: &SumeragiWithFault<F>,
+fn reset_state(
+    addr: &str,
     current_view_change_index: u64,
     old_view_change_index: &mut u64,
     current_latest_block_height: u64,
     old_latest_block_height: &mut u64,
```
appetrosyan, line 533: Should be a struct, no? 


There's something called negative cost abstractions, widely regarded in C++, when you can save lots of cycles by packing the data into a struct instead of passing each individually. It's an LLVM optimisation so it should work here too.

core/src/sumeragi/main_loop.rs
```diff
@@ -463,123 +465,129 @@ fn handle_role_agnostic_messages<F>(
...
 
+    if current_latest_block_height != *old_latest_block_height {
         *old_latest_block_height = current_latest_block_height;
+        was_commit_or_view_change = true;
     }
     if current_view_change_index != *old_view_change_index {
         current_topology.rebuild_with_new_view_change_count(current_view_change_index);
+        *old_view_change_index = current_view_change_index;
+        was_commit_or_view_change = true;
+        error!(%current_view_change_index, "No consensus, updating view change...");
```
appetrosyan:
```suggestion
        warn!(%current_view_change_index, "No consensus, updating view change...");
```

`error` is reserved for things that the user is expected to fix. Given that BFT is a selling point, asking the user to investigate is not particularly useful.

core/src/sumeragi/main_loop.rs
```diff
@@ -463,123 +465,129 @@ fn handle_role_agnostic_messages<F>(
...
+    use tokio::sync::oneshot::error::TryRecvError;
 
-        *old_view_change_index = current_view_change_index;
-        error!(
-            %current_view_change_index,
-            "No consensus: the peers failed to agreed on the block. Updating view change to try again",
-        );
+    match shutdown_receiver.try_recv() {
+        Ok(()) | Err(TryRecvError::Closed) => {
+            info!("Sumeragi Thread is being shutdown shut down.");
```
appetrosyan, line 570:
```suggestion
            info!("Sumeragi Thread is being shut down.");
```

core/src/sumeragi/mod.rs
mversic: there is no suggestion to edit here, right?

core/src/block.rs
mversic: I'd rather not

core/src/sumeragi/main_loop.rs
mversic: in this case your point is valid. However, due to the nature of distributed systems it's not possible to always say if error is in the system, user made or it's just a flake. Like when you get failed signature verification, this can be due to some misconfiguration, malicious peer, or just because some message came in late (which is normal behavior). These cases require manual inspection by a human and I lean toward marking them as `error`s in the log even if they may not be

core/src/block.rs
mversic: I don't want to make the block part of the error report. A new error variant could make more sense here, but I'd rather leave it as is atm

core/src/sumeragi/main_loop.rs
mversic: I understand, if I continue with a refactor I will make this change in another PR. Considering the state of sumeragi code that I found I think this is much better now.

core/src/tx.rs
mversic: yeah, the issue is that we rely on `eyre` error reports too often and then you have no choice but compare strings. Like in this case for `AcceptedTransaction::from_transaction`. Ideally, we should return error reports only when reporting errors outside rust (logging, cli, network). I will not change this now, this should be done systematically across codebase

core/src/block_sync.rs
mversic: good, as long it's been brought to our attention we can look into it in the future

SamHSmith: Originally sumeragi recieved BlockCommitted messages that originated both from other peers and block_sync. We could change it although I doubt it would help much since performance is dominated by WSV::apply.

core/src/sumeragi/main_loop.rs
```diff
@@ -635,479 +725,276 @@ pub fn run<F>(
...
-                        ),
-                        std::iter::once(state.current_topology.leader()),
-                    );
-                    has_sent_transactions = true;
-                    sent_transaction_time = Instant::now();
-                }
-            }
+        if round_start_time.elapsed() > sumeragi.pipeline_time() && !view_change_suggested {
+            // NOTE: Either there are malicious peers or there are no transactions.
+            // There is no meaningful performance hit when there are no transaction.
```
SamHSmith, line 764: The peer should not suggest a view change if it has no local transactions it wants to see get through. And the view change suggestions are not guaranteed to be delivered. In order to ensure a view change does eventually happen the suspecting peers must periodically call suggest_view_change() until the view change occurs.

SamHSmith (changes requested):
The refactor is good. Address comments and then you are in the clear. Most important thing is though that you become familiar with the code.

core/src/sumeragi/main_loop.rs
mversic: > The peer should not suggest a view change if it has no local transactions it wants to see get through

why not, what's the harm? Other peers will request it anyhow. I'm trying to keep it as simple as possible

> And the view change suggestions are not guaranteed to be delivered. In order to ensure a view change does eventually happen the suspecting peers must periodically call suggest_view_change() until the view change occurs.

got you, I believe it's fixed now

core/src/sumeragi/main_loop.rs
```diff
@@ -463,132 +559,134 @@ fn handle_role_agnostic_messages<F>(
...
+        *old_latest_block_height = current_latest_block_height;
+        *old_view_change_index = current_view_change_index;
 
-        *voting_block_option = None;
-        block_signature_acc.clear();
-        *has_sent_transactions = false;
+        *view_change_suggested = false;
+        *voting_block = None;
+        voting_signatures.clear();
+        *round_start_time = Instant::now();
```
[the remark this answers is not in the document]
mversic: I'm having a hard time understanding you here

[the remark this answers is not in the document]
SamHSmith: Actually after thinking about it this is less of an issue that I originally thought. So keeping it as it is is fine.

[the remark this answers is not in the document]
mversic: thank you

core/src/sumeragi/main_loop.rs
mversic: ok, this is a valid point. Let me see what I can do about it

mversic: I've addressed it with comments in the code. If you don't mind I prefer the view change is made at least every `pipeline_time` if there is no block committed(for whatever reason) or earlier if the block is committed

core/src/sumeragi/main_loop.rs
```diff
@@ -216,101 +223,206 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+        block_height=%state.latest_block_height,
+        %block_hash, "Committing block"
     );
-    sumeragi.kura.store_block_blocking(block);
-    state.current_topology.update_network_topology(&state.wsv);
 
-    // Transaction Cache
-    cache_transaction(state, sumeragi)
+    sumeragi.kura.store_block_blocking(committed_block);
+    current_topology.update_network_topology(&state.wsv);
```
mversic, line 423: I'm not sure if `update_network_topology` should be called after `refresh_at_new_block`? In all honesty, I think this should be joined into one function, because wsv can change peers only if block was committed. But the ordering might still be relevant

core/src/sumeragi/main_loop.rs
```diff
@@ -216,101 +224,206 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
+                _ => (),
+            }
+        }
+    }
 }
 
-#[allow(clippy::expect_used, clippy::ptr_arg)]
-fn gossip_transactions<F>(
+fn gossip_transactions<F: FaultInjection>(
     sumeragi: &SumeragiWithFault<F>,
```
Erigara, line 361: What do you think of making this method of `sumeragi`?

mversic: done, most of these functions can be made methods of `sumeragi`. This would require another round of refactor

core/src/sumeragi/main_loop.rs
SamHSmith: We can do it that way. We can always change back later if in production it is confusing.

core/src/sumeragi/main_loop.rs
mversic: @SamHSmith @Erigara

core/src/sumeragi/main_loop.rs
mversic: I made it so that it doesn't look confusing in the logs. Otherwise, I would have to special case and tangle up the code

core/src/sumeragi/main_loop.rs
Erigara: As i see it peers only can be added through committing new block (`Register<Peer>/Unregister<Peer>`) so seems like `refresh_at_new_block` should include `update_network_topology`.

Erigara (approved):

core/src/sumeragi/main_loop.rs
mversic: I agree with that point, I'm just not sure which should be done first updating the network or refreshing. I would say first update(this is only done if there was a peer added/removed through ISI) then refresh the view

Erigara: I think it can be done at the same time by using topology `Builder`.

core/src/sumeragi/mod.rs
```diff
@@ -342,18 +332,23 @@ pub const TELEMETRY_INTERVAL: Duration = Duration::from_secs(5);
...
+    /// Constructs new `VotingBlock` with current time.
     #[allow(clippy::expect_used)]
     pub fn new(block: SignedBlock) -> VotingBlock {
         VotingBlock {
-            voted_at: current_time(),
             block,
+            voted_at: Instant::now(),
         }
     }
+    /// Constructs new `VotingBlock` with the given time
```
outoftardis, line 349:
```suggestion
    /// Construct new `VotingBlock` with the given time
```

core/src/sumeragi/mod.rs
```diff
@@ -342,18 +332,23 @@ pub const TELEMETRY_INTERVAL: Duration = Duration::from_secs(5);
...
     /// At what time has this peer voted for this block
-    pub voted_at: Duration,
+    pub voted_at: Instant,
     /// Valid Block
     pub block: SignedBlock,
 }
 
 impl VotingBlock {
-    /// Constructs new `VotingBlock.`
+    /// Constructs new `VotingBlock` with current time.
```
outoftardis, line 341:
```suggestion
    /// Construct new `VotingBlock` with current time.
```

core/src/sumeragi/main_loop.rs
mversic: got you `at_block` actually doesn't do anything per se. Will try to fix this before merging

SamHSmith: Order does not matter. `update_network_topology` only does something if the peers set in the wsv has changed.

SamHSmith (dismissed):

core/src/sumeragi/main_loop.rs
mversic: I've made this final change

Erigara (approved):

appetrosyan (approved):

PR #3062, opened 2023-01-17
[feature] #2794: Reject Fieldless enums with explicit discriminants in FFI

### LINKED ISSUE #2794: Reject Fieldless enums with explicit discriminants in FFI
Currently, determining discriminant values of fieldless enums in FFI is done by incrementally counting the number of variants. This works for the common case, but is not entirely accurate. There can be holes in the sequentially increasing number for the discriminant, in the case that user defines a discriminant value by hand. Additionally, related to this, is finding a niche value for the fieldless enum if the enum doesn't take the whole range of the discriminant. This can be any value in the range. 

Example:

```rs
#[repr(u8)]
enum FieldlessEnum {
    A,       // discriminant = 0
    B = 43 - 1,
    C,        // discriminant = 43
    D = 255
}
```

This will most likely fail to compile or will result in an incorrect implementation. 
In this case niche value can be any value in the range `[1, 41]` or `[44, 254]`. I would recommend choosing a value that is most unlikely to be used by the enum in the future (assuming that additional variants will be added at some point). Largest available discriminant value might be a good candidate?

Erigara:

### Description of the Change

Derive macro `FfiType` now rejects fieldless enums with explicit discriminants.

### Issue

Closes #2794.

### Benefits

Less space to make errors.

### Possible Drawbacks

None.

### DIFF
--- a/ffi/derive/src/convert.rs
+++ b/ffi/derive/src/convert.rs
@@ -28,6 +28,12 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
             }
 
             if is_fieldless_enum(&item) {
+                if item.variants.iter().any(|v| v.discriminant.is_some()) {
+                    abort!(
+                        name,
+                        "fieldless enums with explicit discriminants are prohibited"
+                    );
+                }
                 derive_ffi_type_for_fieldless_enum(&input.ident, &item, &repr)
             } else {
                 let local = !is_non_local(&input.attrs);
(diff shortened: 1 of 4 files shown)


mversic (dismissed):

mversic:
I think you can also simiplify the `variant_discriminants` function

mversic (approved):

appetrosyan (approved):

PR #3064, opened 2023-01-17
[fix] #1170: Implement cloning-wsv-style soft-fork handling.

SamHSmith:


core/src/sumeragi/mod.rs
```diff
@@ -250,9 +251,38 @@ impl Sumeragi {
...
         sumeragi: Arc<Self>,
         genesis_network: Option<GenesisNetwork>,
+        block_hashes: Vec<HashOf<VersionedCommittedBlock>>,
     ) -> ThreadHandler {
         let wsv = sumeragi.wsv_mutex_access().clone();
 
+        for (i, block_hash) in block_hashes
+            .iter()
+            .take(block_hashes.len().saturating_sub(1))
+            .enumerate()
```
Erigara, line 261: I would suggest the following to avoid conversion: 

```suggestion
            .zip(1u64..)
```

core/src/sumeragi/mod.rs
```diff
@@ -250,9 +251,38 @@ impl Sumeragi {
...
+                "Kura init correctly reported the block hash."
+            );
+
+            wsv.apply(&block_ref)
+                .expect("Failed to apply block to wsv in init.");
+        }
+        let delayed_wsv = wsv.clone();
+
+        if let Some(block_hash) = block_hashes.last() {
+            let block_ref = sumeragi.internal.kura.get_block_by_hash(block_hash).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
```
Erigara, line 277: You can get block by height here as well, no?

core/src/sumeragi/mod.rs
```diff
@@ -283,6 +313,7 @@ impl Sumeragi {
             latest_block_view_change_index,
             current_topology,
             wsv,
+            delayed_wsv,
```
Erigara, line 316: What do you think about name `finalized_wsv` as it represent `wsv` that will never change?

core/src/sumeragi/main_loop.rs
```diff
@@ -109,6 +112,7 @@ pub struct State {
     /// as of now we keep them seperate for greater flexibility when
     /// optimizing.
     pub wsv: WorldStateView,
+    pub delayed_wsv: WorldStateView,
```
Erigara, line 115: Missing docs.

core/src/sumeragi/main_loop.rs
```diff
@@ -77,6 +77,9 @@ where
     pub message_sender: Mutex<mpsc::SyncSender<MessagePacket>>,
     /// Receiver channel.
     pub message_receiver: Mutex<mpsc::Receiver<MessagePacket>>,
+    /// Only used in testing. Causes the genesis peer to withhold blocks when it
+    /// is the proxy tail.
+    pub debug_force_soft_fork: bool,
```
Erigara, line 82: I wonder could we test soft-forks by other means because i'm not really fan of such approach. 

For example we can construct 2 `BlockSync` update message (with soft-fork in them) and feed them to sumeragi.

core/src/kura.rs
```diff
@@ -321,11 +336,13 @@ impl Kura {
         &self,
         block_hash: &HashOf<VersionedCommittedBlock>,
     ) -> Option<Arc<VersionedCommittedBlock>> {
-        self.block_data
+        let maybe_index = self
```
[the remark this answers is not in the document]
SamHSmith: sure thing

core/src/sumeragi/main_loop.rs
mversic: it's also possible to use compilation flags

SamHSmith: I will find a different solution.

core/src/sumeragi/mod.rs
SamHSmith: sure

core/src/sumeragi/mod.rs
SamHSmith: I have made the change.

core/src/sumeragi/mod.rs
SamHSmith: I was hesitant to do that originally but you are right it is better that way.

core/src/sumeragi/main_loop.rs
appetrosyan: While compile-time forcing is going to lead to quick testing, it doesn't seem to be the kind of thing that can be covered with unit tests. Integration tests, on the other hand need to work on the standard binary (because conditional compilation itself can produce spurious failures/successes). 


Can we come up with a realistic scenario that can come up in a deployment, force a soft-fork programmatically and externally (e.g. in the bash-script tests that we already have), and use that to test in the distributed environment?

core/src/kura.rs
```diff
@@ -165,30 +165,36 @@ impl Kura {
...
+        Ok(block_hashes)
     }
 
     #[allow(clippy::expect_used, clippy::cognitive_complexity, clippy::panic)]
     fn kura_receive_blocks_loop(
         kura: &Kura,
         mut shutdown_receiver: tokio::sync::oneshot::Receiver<()>,
     ) {
-        let mut written_block_count = kura.block_data.lock().len();
+        let (mut written_block_count, mut latest_block_hash) = {
```
appetrosyan, line 191: Clever.

core/src/kura.rs
```diff
@@ -335,6 +352,14 @@ impl Kura {
             .lock()
             .push((block.hash(), Some(Arc::new(block))));
     }
+
+    /// Replace the block in kura's in memory block store. Kura will write the block
+    /// to disk in due time.
```
appetrosyan, line 357:
```suggestion
    /// Replace the block in `Kura`'s in-memory block store. 
```

*In due time* isn't particularly informative, and relies on external code. Being terse is often good for maintainability.

core/src/kura.rs
```diff
@@ -335,6 +352,14 @@ impl Kura {
...
+    pub fn replace_top_block_blocking(&self, block: VersionedCommittedBlock) {
```
appetrosyan, line 358:
```suggestion
    pub fn replace_top_block(&self, block: VersionedCommittedBlock) {
```

The blocking property might change, and we don't really have a non-blocking alternative.

core/src/sumeragi/main_loop.rs
```diff
@@ -420,6 +431,50 @@ fn commit_block<F: FaultInjection>(
...
+    block: impl Into<VersionedCommittedBlock>,
+) {
+    let committed_block = block.into();
+    let block_hash = committed_block.hash();
+
+    state.wsv = state.finalized_wsv.clone();
+    state
+        .wsv
+        .apply(&committed_block)
+        .expect("Failed to apply block on WSV. This is absolutely not acceptable.");
```
appetrosyan, line 446:
```suggestion
        .expect("Failed to apply block to WSV. Bailing.");
```

AFAIK, this is common terminology for "It's too dangerous to continue running, so we're stopping the execution".

core/src/sumeragi/main_loop.rs
```diff
@@ -420,6 +431,50 @@ fn commit_block<F: FaultInjection>(
...
+    // state.previous_block_hash stays the same.
+    state.latest_block_height = committed_block.header().height;
+    state.latest_block_hash = Some(block_hash);
+    state.latest_block_view_change_index = committed_block.header().view_change_index;
+
+    // Update WSV copy that is public facing
+    {
+        let mut wsv = sumeragi.wsv.lock();
+        *wsv = state.wsv.clone();
+    }
```
appetrosyan, line 460:
```suggestion
        *sumeragi.wsv.lock() = state.wsv.clone();
```
Rust drops temporaries (values not bound to a name) immediately.

core/src/sumeragi/main_loop.rs
```diff
@@ -420,6 +431,50 @@ fn commit_block<F: FaultInjection>(
...
+    info!(
+        addr=%sumeragi.peer_id.address, %role,
+        block_height=%state.latest_block_height,
+        %block_hash, "Replacing top block"
+    );
+
+    sumeragi.kura.replace_top_block_blocking(committed_block);
+    current_topology.recreate(&state.wsv, block_hash);
+
+    // Transaction Cache
```
appetrosyan, line 474:
```suggestion
```

core/src/sumeragi/main_loop.rs
```diff
@@ -540,8 +595,9 @@ fn handle_role_agnostic_message<F: FaultInjection>(
                         peer_latest_block_view_change_index=?state.latest_block_view_change_index,
                         consensus_latest_block_hash=%block.hash(),
                         consensus_latest_block_view_change_index=%block.header().view_change_index,
-                        "Soft fork occurred: peer in inconsistent state. Currently only reload with wipe of the block storage will help."
+                        "Soft fork occurred: peer is inconsistent state. Rolling back and replacing top block."
```
appetrosyan, line 598:
```suggestion
                        "Soft fork: peer in inconsistent state. Rolling back and replacing top block."
```

Should read as "(A) Soft fork (has) (occurred): (and) (the) peer (is) in (an) inconsistent state". The *in* is not optional, as it is the active verb of the clause.

core/src/sumeragi/main_loop.rs
```diff
@@ -906,7 +966,11 @@ pub(crate) fn run<F: FaultInjection>(
                                 BlockCommitted::from(committed_block.clone()),
                             );
 
-                            sumeragi.broadcast_packet(msg, current_topology);
+                            if is_genesis_peer && sumeragi.debug_force_soft_fork {
+                                std::thread::sleep(sumeragi.pipeline_time() * 2);
```
appetrosyan, line 970: How reliable is this for generating a soft fork on weaker machines?

core/src/sumeragi/mod.rs
```diff
@@ -246,9 +248,38 @@ impl Sumeragi {
...
         let wsv = sumeragi.wsv_mutex_access().clone();
 
+        for (block_hash, i) in block_hashes
+            .iter()
+            .take(block_hashes.len().saturating_sub(1))
+            .zip(1u64..)
+        {
+            let block_height: u64 = i.try_into().unwrap();
+            let block_ref = sumeragi.internal.kura.get_block_by_height(block_height).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
+            assert_eq!(
```
appetrosyan, line 262: Good. More assertions please.

core/src/sumeragi/mod.rs
```diff
@@ -246,9 +248,38 @@ impl Sumeragi {
...
+
+        if !block_hashes.is_empty() {
+            let block_ref = sumeragi.internal.kura.get_block_by_height(block_hashes.len() as u64).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
+
+            wsv.apply(&block_ref)
+                .expect("Failed to apply block to wsv in init.");
+        }
+        *sumeragi.wsv_mutex_access() = wsv.clone();
+
+        info!("Sumeragi has loaded finished loading blocks and setting up the WSV");
```
appetrosyan, line 281:
```suggestion
        info!("Sumeragi has finished loading blocks and setting up the WSV");
```

core/src/sumeragi/main_loop.rs
SamHSmith: As far as I can tell 100 %. It produces the exact scenario mentioned in the whitepaper. Proxy tail commits the block locally but goes offline for some time while the rest of the network move on and produce a new block. The new block is guaranteed to happen within the pipeline time so when the peer rejoins it is in conflict with the rest of the network.

config/iroha_test_config.json
```diff
@@ -40,7 +40,8 @@
     },
     "ACTOR_CHANNEL_CAPACITY": 100,
     "GOSSIP_BATCH_SIZE": 500,
-    "GOSSIP_PERIOD_MS": 1000
+    "GOSSIP_PERIOD_MS": 1000,
+    "DEBUG_FORCE_SOFT_FORK": false
```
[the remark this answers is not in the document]
SamHSmith: Aleksandr and I have agreed that when we build in release mode it should not be included. So it never goes to users but we can use it to debug on a deployment.

core/src/sumeragi/main_loop.rs
SamHSmith: > Can we come up with a realistic scenario that can occur in a deployment, force a soft-fork programmatically and externally (e.g. in the bash-script tests that we already have), and use that to test in the distributed environment?

Conclusion was no. It must be triggered from the inside in order for it to happen consistently.

core/src/kura.rs
```diff
@@ -165,30 +165,36 @@ impl Kura {
             }
         }
 
-        info!("Loaded {} blocks at init.", blocks.len());
+        info!("Loaded {} blocks at init.", block_hashes.len());
```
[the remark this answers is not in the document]
SamHSmith: sure thing

client/Cargo.toml
```diff
@@ -17,7 +17,7 @@ is-it-maintained-open-issues = { repository = "https://github.com/hyperledger/ir
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-iroha_config = { version = "=2.0.0-pre-rc.11", path = "../config" }
+iroha_config = { version = "=2.0.0-pre-rc.11", path = "../config", features = [] }
```
mversic, line 20: is the same as `deafult-features == false`?

scripts/check.sh
```diff
@@ -3,27 +3,27 @@ set -e
 
 case $1 in
     "docs")
-        cargo run --bin kagami -- docs | diff - docs/source/references/config.md || {
+        cargo run --release --bin kagami -- docs | diff - docs/source/references/config.md || {
             echo 'Please re-generate docs using `cargo run --bin kagami -- docs > docs/source/references/config.md`'
```
mversic, line 7: then I guess you should update the `echo` message?

core/src/sumeragi/mod.rs
```diff
@@ -246,9 +246,38 @@ impl Sumeragi {
...
+            .iter()
+            .take(block_hashes.len().saturating_sub(1))
+            .zip(1u64..)
+        {
+            let block_height: u64 = i;
+            let block_ref = sumeragi.internal.kura.get_block_by_height(block_height).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
+            assert_eq!(
+                block_ref.hash(),
+                *block_hash,
+                "Kura init correctly reported the block hash."
```
mversic, line 263: isn't this reported if assertion fails? in which case the message should be negated, no?

core/src/sumeragi/mod.rs
```diff
@@ -246,9 +246,38 @@ impl Sumeragi {
...
+            );
+
+            wsv.apply(&block_ref)
+                .expect("Failed to apply block to wsv in init.");
+        }
+        let finalized_wsv = wsv.clone();
+
+        if !block_hashes.is_empty() {
+            let block_ref = sumeragi.internal.kura.get_block_by_height(block_hashes.len() as u64).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
```
mversic, line 272: shouldn't you be getting the last block from `kura`? If you're getting a block by it's height then it's not guaranteed you will get the last block

core/src/sumeragi/main_loop.rs
```diff
@@ -909,7 +966,11 @@ pub(crate) fn run<F: FaultInjection>(
                                 BlockCommitted::from(committed_block.clone()),
                             );
 
-                            sumeragi.broadcast_packet(msg, current_topology);
+                            if is_genesis_peer && sumeragi.debug_force_soft_fork {
```
mversic, line 969: this could be guarded by `debug_assertions`?

mversic (changes requested):

client/Cargo.toml
appetrosyan: Not exactly. But it doesn't matter for other reasons

https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

scripts/check.sh
appetrosyan: Or better yet, autogenerate it based on the command. 

In `bash` you can evaluate a string by enclosing it in `$()`.

core/src/sumeragi/mod.rs
appetrosyan: I would commend the inclusion of the message though.

core/src/sumeragi/main_loop.rs
appetrosyan: LLVM can see that this condition is always `false` so it won't matter all that much.

config/src/sumeragi.rs
```diff
@@ -16,6 +16,9 @@ const DEFAULT_ACTOR_CHANNEL_CAPACITY: u32 = 100;
 const DEFAULT_GOSSIP_PERIOD_MS: u64 = 1000;
 const DEFAULT_GOSSIP_BATCH_SIZE: u32 = 500;
 
+cfg_if::cfg_if! {
+    // not(feature = "hide_debug_config")
+    if #[cfg(debug_assertions)] {
```
appetrosyan, line 21: It seems overkill to pull in `cfg-if` for this. And frankly I don't think we need a complex compile-time configuration infrastructure.

config/src/sumeragi.rs
```diff
@@ -45,6 +48,44 @@ view! {
         pub gossip_batch_size: u32,
         /// Period in milliseconds for pending transaction gossiping between peers.
         pub gossip_period_ms: u64,
+
+        /// Only used in testing. Causes the genesis peer to withhold blocks when it
+        /// is the proxy tail.
+        pub debug_force_soft_fork: bool,
```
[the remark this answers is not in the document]
appetrosyan: Assuming the macro can cope with it.

config/src/sumeragi.rs
```diff
@@ -178,6 +221,7 @@ pub mod tests {
     use super::*;
 
     prop_compose! {
+        #[allow(unused_variables)]
```
appetrosyan, line 224: Not the worst way to approach this. It's fine because we have no guarantees about what the `prop_compose!` macro will do with the extra parameter.

core/src/sumeragi/main_loop.rs
```diff
@@ -660,15 +713,19 @@ pub(crate) fn run<F: FaultInjection>(
 ) {
     let addr = &sumeragi.peer_id.address;
 
-    if state.latest_block_height == 0 || state.latest_block_hash.is_none() {
+    let is_genesis_peer = if state.latest_block_height == 0 || state.latest_block_hash.is_none() {
```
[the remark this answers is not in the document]
appetrosyan: I don't mind this construction. The following steps are clear, and If you want to rewrite it, consider using a `match` with guard expressions. Something along the lines of 
```rust
match (state.latest_block_height, state.latest_block_hash, genesis_network) {
    (0, None, Some(genesis_network) => { init_commit ... ; true },
    (0, None, None) => {init_listen ...; false},
    _ => false
}
```

appetrosyan (approved):

config/src/sumeragi.rs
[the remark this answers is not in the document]
SamHSmith: I would have but it can't be done because of the view macro.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
SamHSmith: I prefer it as it is currently.

client/Cargo.toml
SamHSmith: Yea I forgot to revert this one.

config/src/sumeragi.rs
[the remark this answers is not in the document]
Erigara: Let's create ticket to came back later to this task.

scripts/check.sh
SamHSmith: thank you for catching.

core/src/sumeragi/mod.rs
SamHSmith: I think normally the assert message describes what is being asserted. Such that it then says, assertion failed: "Thing has correct value".

core/src/sumeragi/mod.rs
SamHSmith: The block hashes passed into this function was just produced by kura and contains all the blocks in the store.

core/src/sumeragi/main_loop.rs
SamHSmith: Sure? It's a small gain for release builds but I'll add it.

config/src/sumeragi.rs
SamHSmith: It was needed here in order to provide two versions of the view macro result.

appetrosyan (approved):

Erigara (approved):

mversic (approved):

PR #3074, opened 2023-01-23
[fix] #2456: Make genesis block unlimited

### LINKED ISSUE #2456: 404 when 1650 isi in genesis.
### GIT commit hash

fe7ea911

### Minimum working example

1. Run iroha with the 1650 isi in genesis
[genesis.txt](https://app.zenhub.com/files/181739240/70fe9b7b-a944-4e8e-96be-459ae4777748/download)

### Expected behaviour

I have to create genesis block with some assets. I generate instruction for domain and admin registration. Then I generate instruction for asset registration, 7 setKeyValue instructions and grant permission instruction few times. Total ~ 5870 isi. Then I put it in ONE genesis transaction and add it to genesis tr
ansactions that already had one small transaction. I expected it to work.

### Actual behaviour

```
iroha-iroha0-1  | 2023-01-17T07:49:17.396580Z ERROR iroha_core::genesis: Genesis transaction #0 failed error=Failed to accept transaction: Too many instructions in payload, max number is 4096, but got 6310
```
 Iroha throws 404 Not Found. Text: "  ,domain_name". If I reduce rows instructions amount to ~700 it works.

Also I have tried to split each asset with setkeyvalie and grant isi to separate tx, but result was the same, even with only one row (just for testing).

### Operating system

macOS

### Current environment

Docker Hub

### Who can help?
@astrokov7

Erigara:

### Description of the Change

- Add const generic to avoid limit check if transaction is part of genesis block.
- Genesis block early return error if it contains zero accepted transactions.

### Issue

Closes #2456.

### Benefits

Provide possibility for users to submit large genesis which should play well with block store merge tool. 

### Possible Drawbacks

User can try to submit too large genesis which would cause issues during network transfer.

### Usage Examples or Tests

```
cargo test --package iroha_core --lib -- tx::tests::genesis_transaction_ignore_limits --exact --nocapture
```

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -29,7 +29,7 @@ use serde::Serialize;
 use crate::{
     prelude::*,
     sumeragi::network_topology::{Role, Topology},
-    tx::{TransactionValidator, VersionedAcceptedTransaction},
+    tx::{TransactionOrigin, TransactionValidator, VersionedAcceptedTransaction},
 };
 
 /// Default estimation of consensus duration
@@ -482,15 +482,19 @@ impl VersionedCandidateBlock {
     /// # Errors
     /// Forward errors from [`CandidateBlock::revalidate`]
     #[inline]
-    pub fn revalidate(
+    pub fn revalidate<const TX_ORIGIN: TransactionOrigin>(
         self,
         transaction_validator: &TransactionValidator,
         wsv: &WorldStateView,
         latest_block: Option<HashOf<VersionedCommittedBlock>>,
         block_height: u64,
     ) -> Result<SignedBlock, eyre::Report> {
-        self.into_v1()
-            .revalidate(transaction_validator, wsv, latest_block, block_height)
+        self.into_v1().revalidate::<TX_ORIGIN>(
+            transaction_validator,
+            wsv,
+            latest_block,
+            block_height,
+        )
     }
 }
 
@@ -549,7 +553,8 @@ impl CandidateBlock {
     /// - There is a mismatch between candidate block previous block hash and actual latest block hash
     /// - Block header transaction hashes don't match with computed transaction hashes
     /// - Error during revalidation of individual transactions
-    pub fn revalidate(
+    #[allow(clippy::too_many_lines)]
+    pub fn revalidate<const TX_ORIGIN: TransactionOrigin>(
         self,
         transaction_validator: &TransactionValidator,
         wsv: &WorldStateView,
@@ -615,7 +620,10 @@ impl CandidateBlock {
             .into_iter()
             .map(VersionedSignedTransaction::into_v1)
             .map(|tx| {
-                AcceptedTransaction::from_transaction(tx, &transaction_validator.transaction_limits)
+                AcceptedTransaction::from_transaction::<TX_ORIGIN>(
+                    tx,
+                    &transaction_validator.transaction_limits,
+                )
             })
             .map(|accepted_tx| {
                 accepted_tx.and_then(|tx| {
@@ -638,7 +646,10 @@ impl CandidateBlock {
             .into_iter()
             .map(VersionedSignedTransaction::into_v1)
             .map(|tx| {
-                AcceptedTransaction::from_transaction(tx, &transaction_validator.transaction_limits)
+                AcceptedTransaction::from_transaction::<TX_ORIGIN>(
+                    tx,
+                    &transaction_validator.transaction_limits,
+                )
             })
             .map(|accepted_tx| {
                 accepted_tx.and_then(|tx| {
@@ -1181,8 +1192,10 @@ mod tests {
         let tx = Transaction::new(alice_id, [create_asset_definition].into(), 4000)
             .sign(alice_keys)
             .expect("Valid");
-        let tx = crate::VersionedAcceptedTransaction::from_transaction(tx, &transaction_limits)
-            .expect("Valid");
+        let tx = crate::VersionedAcceptedTransaction::from_transaction::<
+            { TransactionOrigin::ConsensusBlock },
+        >(tx, &transaction_limits)
+        .expect("Valid");
 
         // Creating a block of two identical transactions and validating it
         let transactions = vec![tx.clone(), tx];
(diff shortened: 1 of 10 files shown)


core/src/block.rs
```diff
@@ -482,15 +482,19 @@ impl VersionedCandidateBlock {
     /// # Errors
     /// Forward errors from [`CandidateBlock::revalidate`]
     #[inline]
-    pub fn revalidate(
+    pub fn revalidate<const IS_GENESIS: bool>(
```
mversic, line 485: was this my influence? :)

core/src/genesis.rs
```diff
@@ -147,43 +147,46 @@ impl GenesisNetworkTrait for GenesisNetwork {
...
+                        .expect("Should be `Some` when `submit_genesis` is true")
+                        .account_public_key
+                        .clone(),
+                    genesis_config
+                        .as_ref()
+                        .expect("Should be `Some` when `submit_genesis` is true")
+                        .account_private_key
+                        .clone()
+                        .ok_or_else(|| eyre!("Genesis account private key is empty."))?,
+                )?;
```
mversic, line 171:
```suggestion
                let genesis_config = genesis_config.as_ref().expect("Should be `Some` when `submit_genesis` is true");
                
                let genesis_key_pair = KeyPair::new(
                    genesis_config.account_public_key.clone(),
                    genesis_config.account_private_key.clone().ok_or_else(|| eyre!("Genesis account private key is empty."))?,
                )?;
```

core/src/genesis.rs
```diff
@@ -147,43 +147,46 @@ impl GenesisNetworkTrait for GenesisNetwork {
...
-                            .ok_or_else(|| eyre!("Genesis account private key is empty."))?,
-                    )?;
-
-                    raw_transaction.sign_and_accept(genesis_key_pair, tx_limits)
-                })
-                .enumerate()
-                .filter_map(|(i, res)| {
-                    res.map_err(|error| {
-                        let error_msg = format!("{error:#}");
-                        iroha_logger::error!(error = %error_msg, "Genesis transaction #{i} failed")
```
mversic, line 182: do we want to be forgiving in the genesis?

mversic: I mean why don't we insist that genesis is completely valid?

mversic (changes requested):

core/src/block.rs
Erigara: yes :) 
I think that we know exactly places in code where we expect genesis transaction so it would be convenient to use const generics.

core/src/genesis.rs
Erigara: We only remove limitation on number of isi and wasm size in genesis block.

Genesis block is kinda special in a sense that it could be larger than ordinary block:
1) When @QuentinI will finish with his merging tool all block store will be condensed into one genensis block (and it probably be larger than normal block)
2) There is no time limit on applying genesis block (so we can afford it to larger imo)

mversic: Let me rephrase. If a transaction in a genesis fails we seem to just skip it and log error. Why don't we fail blockchain initialization if there is even one invalid transaction?

mversic: This is orthogonal question to this PR. I understand that you've relaxed limits on genesis

Erigara: Oh, i see now, this is actually good point. IMO we shouldn't allow such behavior (what point in doing this after all).

mversic: Let's open a ticket to discuss further. There may be a good reason for this

Erigara: Created #3075.

mversic (dismissed):

core/src/genesis.rs
Erigara: Actually after relaxation on limits only one reason for invalid tx would be wrong signatures.

mversic (approved):

core/src/block.rs
appetrosyan: I applaud this approach. Although I personally think that `::<false>` doesn't really convey the idea.

core/src/genesis.rs
```diff
@@ -147,55 +147,44 @@ impl GenesisNetworkTrait for GenesisNetwork {
...
+            .transactions
+            .iter()
+            .map(|raw_transaction| {
+                raw_transaction.sign_and_accept(genesis_key_pair.clone(), tx_limits)
+            })
+            .enumerate()
+            .filter_map(|(i, res)| {
+                res.map_err(|error| {
+                    let error_msg = format!("{error:#}");
+                    iroha_logger::error!(error = %error_msg, "Genesis transaction #{i} failed")
```
appetrosyan, line 174:
```suggestion
                    iroha_logger::error!(error = %error_msg, transaction_num=i, "Genesis transaction failed")
```
Structured logging will let you extract the value instead of parsing the message.

core/src/genesis.rs
```diff
@@ -147,55 +147,44 @@ impl GenesisNetworkTrait for GenesisNetwork {
...
-            genesis_submission_delay_ms: genesis_config
-                .as_ref()
-                .expect("Should be `Some` when `submit_genesis` is true")
-                .genesis_submission_delay_ms,
+                .ok()
+            })
+            .collect::<Vec<_>>();
+        if transactions.is_empty() {
+            iroha_logger::error!("Genesis transaction set contains no valid transactions");
+            bail!("Genesis transaction set contains no valid transactions");
```
appetrosyan, line 181: This log and bail pattern is a problem.

core/src/block.rs
mversic: I agree, and they should be on stable soon. So we might consider using them right now

appetrosyan (dismissed):

core/src/genesis.rs
Erigara: I will just `bail` then.

appetrosyan (approved):

core/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 //! Iroha — A simple, enterprise-grade decentralized ledger.
+#![feature(adt_const_params)]
```
mversic, line 2: I support using nightly features that will reach stable soon
can we enable it project wide? probably not

mversic (approved):

core/src/lib.rs
Erigara: I don't find a way to enable feature project wide.

PR #3077, opened 2023-01-24
[feature] #3072: enable `ffi_import` in `iroha_crypto`

### LINKED ISSUE #3072: enable `ffi_import` for `iroha_crypto`
Enable the use of `iroha_crypto` in a dynamically linked library

mversic:

### Description of the Change

I don't want to make additional changes in this PR it's getting rather big:
- `ffi_import` still isn't enabled in `iroha_crypto`, because it conflicts with `data_model`

### Issue

Closes #3072


ffi/derive/src/wrapper.rs
```diff
@@ -29,36 +117,27 @@ fn gen_shared_fns(name: &Ident, attrs: &[syn::Attribute]) -> Vec<TokenStream> {
...
                         } else if path.is_ident("Ord") {
-                            derive_ord = true;
-                        } else if path.is_ident("PartialEq") || path.is_ident("PartialOrd") {
-                            // NOTE: These should be skipped
+                            shared_fn_impls.push(impl_ord_for_opaque(name));
+                        } else if path.is_ident("Setters")
+                            || path.is_ident("Getters")
+                            || path.is_ident("MutGetters")
+                        {
+                            // TODO: What to do about getters/setters?
```
Erigara, line 140: What's wrong with providing `getters/setters` for opaque type?

crypto/src/hash.rs
```diff
@@ -12,30 +13,15 @@ use ursa::blake2::{
...
-    #[display(fmt = "{}", "hex::encode(self.as_ref())")]
-    #[debug(fmt = "{}", "hex::encode(self.as_ref())")]
-    #[repr(C)]
-    pub struct Hash {
-        more_significant_bits: [u8; Self::LENGTH - 1],
-        least_significant_byte: NonZeroU8,
-    }
+/// Hash of Iroha entities. Currently supports only blake2b-32.
+/// The least significant bit of hash is set to 1.
+#[derive(Clone, Copy, Display, DebugCustom, Hash, Eq, PartialEq, Ord, PartialOrd, IntoSchema)]
```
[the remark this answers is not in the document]
mversic: It is implemented but with `ffi_type!` as you can see in the code. The reason is that we want to treat it as transparent with respect to `[u8; Hash::Size]` but can't derive it as such due to the fact that it's composed of two fields, not one.

ffi/derive/src/convert.rs
```diff
@@ -28,9 +28,12 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
                 if item.variants.iter().any(|v| v.discriminant.is_some()) {
                     abort!(
                         name,
-                        "fieldless enums with explicit discriminants are prohibited"
+                        "Fieldless enum with explicit discriminants are prohibited"
                     );
                 }
+                if without_repr(&input.attrs) {
+                    abort!(name, "Fieldless enum missing #[repr(int)]");
+                }
```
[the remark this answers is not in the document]
mversic: This, at least, used to be the case. Maybe rust is now offering this guarantee out-of-the-box?

crypto/src/lib.rs
```diff
@@ -59,32 +63,29 @@ pub const BLS_NORMAL: &str = "bls_normal";
...
-#[derive(Debug, Clone, Copy, Display, IntoSchema)]
+#[derive(Debug, Display, Clone, Copy, IntoSchema)]
 #[display(fmt = "Algorithm not supported")]
 pub struct NoSuchAlgorithm;
 
-#[allow(clippy::std_instead_of_alloc)]
 #[cfg(feature = "std")]
 impl std::error::Error for NoSuchAlgorithm {}
 
-ffi::ffi_item! {
```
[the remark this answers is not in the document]
mversic: It's not used in FFI. Current situation is that errors are serialized as strings and fetched separately. If this changes the code will not compile unless this error derives `FfiType`

Can you say specifically which other types you see problematic?

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
mversic: And I think it's no big deal to denote it explicitly in all cases, regardless of ffi

ffi/derive/src/wrapper.rs
```diff
@@ -29,36 +117,27 @@ fn gen_shared_fns(name: &Ident, attrs: &[syn::Attribute]) -> Vec<TokenStream> {
...
-
-                                        if clone_result != iroha_ffi::FfiReturn::Ok  {
-                                            panic!("Clone returned: {}", clone_result);
-                                        }
-
-                                        unsafe {iroha_ffi::FfiOutPtrRead::try_read_out(output.assume_init()).expect("Invalid output")}
-                                    }
-                                }
-                            });
+                        if path.is_ident("Copy") {
```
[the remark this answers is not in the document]
mversic: You cannot have both `Copy` and `Drop` implemented for the same type. I was thinking to add a comment but since this wouldn't compile if it is changed it would be obvious right away why it is so

ffi/derive/src/wrapper.rs
mversic: I'm still working on this. They will be provided

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
mversic: I don't know why fieldless enums don't have an ABI of the smallest possible int by default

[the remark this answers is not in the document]
Erigara: > I don't know why fieldless enums don't have an ABI of the smallest possible int by default

Seems like rustc chooses smallest posible int ([link](https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#layout-of-a-fieldless-enum)).

crypto/src/lib.rs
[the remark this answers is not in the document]
Erigara: On the second look i can't find any such place.

ffi/derive/src/convert.rs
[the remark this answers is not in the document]
mversic: > > I don't know why fieldless enums don't have an ABI of the smallest possible int by default
> 
> Seems like rustc chooses smallest posible int ([link](https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#layout-of-a-fieldless-enum)).

`In the absence of a #[repr] annotation, the number of bits used by the compiler are not defined and are subject to change.`

ffi/tests/getset.rs
```diff
@@ -3,15 +3,17 @@
 use std::mem::MaybeUninit;
 
 use getset::{Getters, MutGetters, Setters};
-use iroha_ffi::{ffi_export, FfiType};
 
-#[derive(Debug, Clone, PartialEq, Eq, FfiType)]
+#[derive(iroha_ffi::FfiType)]
```
mversic, line 7: the situation is such that `FfiType` needs to be declared as the outermost macro so it observes other derives. This is so that if getset derives are observed it will automatically produce FFI functions

data_model/src/role.rs
```diff
@@ -7,67 +7,38 @@ use std::collections::btree_set;
...
-        Getters,
-        Decode,
-        Encode,
-        Deserialize,
-        Serialize,
-        FfiType,
-        IntoSchema,
-    )]
-    #[cfg_attr(all(feature = "ffi_export", not(feature = "ffi_import")), iroha_ffi::ffi_export)]
-    #[cfg_attr(feature = "ffi_import", iroha_ffi::ffi_import)]
```
mversic, line 69: this is now automatically derived either with `FfiType` (added on `feature = "ffi_export`) or `iroha_ffi::ffi!` (added on `feature = "ffi_import"`)

data_model/src/role.rs
```diff
@@ -7,67 +7,38 @@ use std::collections::btree_set;
...
+    /// Identification of a role.
+    #[derive(Debug, Display, Clone, Constructor, FromStr, PartialEq, Eq, PartialOrd, Ord, Hash, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+    #[serde(transparent)]
+    #[repr(transparent)]
+    pub struct Id {
+        /// Role name, should be unique .
+        pub name: Name,
+    }
+
+    ffi_type(opaque)
```
mversic, line 34: should we treat all `Id`s as opaque in FFI context?

telemetry/Cargo.toml
```diff
@@ -35,7 +35,6 @@ prometheus = { version = "0.13.3", default-features = false }
 
 [build-dependencies]
 anyhow = "1.0.68"
-vergen = { version = "7.3.2", default-features = false, features = ["cargo", "git"] }
-# DO NOT BUMP THIS VERSION.
-git2 = { version = "=0.14.4", features = ["vendored-libgit2"] }
```
mversic, line 40: @appetrosyan was there a reason `libgit2` was statically linked?
seems that the version of `libgit2-sys` had to be fixated instead of `git2`

client_cli/src/main.rs
```diff
@@ -544,7 +544,7 @@ mod account {
     impl RunArgs for ListPermissions {
         fn run(self, cfg: &ClientConfiguration) -> Result<Box<dyn Serialize>> {
             let client = Client::new(cfg)?;
-            let find_all_permissions = FindPermissionTokensByAccountId { id: self.id.into() };
+            let find_all_permissions = FindPermissionTokensByAccountId::new(self.id);
```
mversic, line 547: it's a good question how this passes in the CI

ffi/derive/tests/ui_fail/derive_skip_field.rs
```diff
@@ -4,9 +4,9 @@ use getset::{Getters, Setters};
 use iroha_ffi::{ffi_export, FfiConvert, FfiType};
 
 /// FfiStruct
+#[ffi_export]
```
mversic, line 7: has to be outer most macro so that it's expanded before `derive` macro

ffi/src/handle.rs
```diff
@@ -72,6 +72,9 @@ macro_rules! def_ffi_fn {
             },
         }
     };
+    ( $prefix:literal $($fn_name:ident: {$($other:ty),+ $(,)?}),+ $(,)?) => {
+        $( $crate::def_ffi_fns! {$prefix $fn_name: $( $other ),+ } )+
+    };
```
mversic, line 77:
```suggestion
    ( $($prefix:literal)? $($fn_name:ident: {$($other:ty),+ $(,)?}),+ $(,)?) => {
        $( $crate::def_ffi_fns! {$prefix $fn_name: $( $other ),+ } )+
    };
```

I wanted to make `prefix` optional like this but I just don't know how to

Erigara: Most straight forward approach will be to just add one more rule without prefix.

ffi/tests/export_shared_fns.rs
```diff
@@ -2,9 +2,16 @@
...
 use std::{cmp::Ordering, mem::MaybeUninit};
 
-use iroha_ffi::{
-    def_ffi_fn, ffi_export, handles, FfiConvert, FfiOutPtrRead, FfiReturn, FfiType, Handle,
-};
+use iroha_ffi::{def_ffi_fns, ffi_export, FfiConvert, FfiOutPtrRead, FfiReturn, FfiType, Handle};
+
+iroha_ffi::handles! {FfiStruct1, FfiStruct2}
+
+def_ffi_fns! {""
```
Erigara, line 9: Maybe we can add something like `def_ffi_fns! { prefix: "...", ...}`? because at least for me it was not obvious what this do.

ffi/src/handle.rs
mversic: haha, that's it. Thank you

ffi/tests/export_shared_fns.rs
mversic: done

client_cli/src/main.rs
appetrosyan: Very likely a misunderstanding between `cargo` and build scripts. We always assume that you're working from the crate root (that's why CI passed). We also had `cargo-all-features` which entered all subdirs and ran tests from there, but it was done manually and I stopped doing that when we had too much work in October.

appetrosyan: Created #3311

ffi/src/handle.rs
appetrosyan: > Most straight forward approach will be to just add one more rule without prefix.

Actually that would also be faster, given that the rule captures are slower than rule matching.

core/src/tx.rs
```diff
@@ -54,11 +54,11 @@ impl TransactionValidator {
         is_genesis: bool,
         wsv: &WorldStateView,
     ) -> Result<VersionedValidTransaction, VersionedRejectedTransaction> {
-        if let Err(rejection_reason) = self.validate_internal(tx.clone(), is_genesis, wsv) {
+        if let Err(reason) = self.validate_internal(tx.clone(), is_genesis, wsv) {
             return Err(RejectedTransaction {
                 payload: tx.payload,
                 signatures: tx.signatures,
-                rejection_reason,
+                rejection_reason: reason,
```
appetrosyan, line 61: Please revert.

core/test_network/src/lib.rs
```diff
@@ -58,7 +58,7 @@ pub fn get_key_pair() -> KeyPair {
         .expect("Public key not in mulithash format"),
         PrivateKey::from_hex(
             Algorithm::Ed25519,
-            "9AC47ABF 59B356E0 BD7DCBBB B4DEC080 E302156A 48CA907E 47CB6AEA 1D32719E 7233BFC8 9DCBD68C 19FDE6CE 61582252 98EC1131 B6A130D1 AEB454C1 AB5183C0",
+            "9AC47ABF 59B356E0 BD7DCBBB B4DEC080 E302156A 48CA907E 47CB6AEA 1D32719E 7233BFC8 9DCBD68C 19FDE6CE 61582252 98EC1131 B6A130D1 AEB454C1 AB5183C0".as_ref(),
```
appetrosyan, line 61: Should we consider applying this formatting to other hard-coded keys?

crypto/src/lib.rs
```diff
@@ -61,18 +62,9 @@ pub const BLS_NORMAL: &str = "bls_normal";
...
-pub struct NoSuchAlgorithm;
-
-#[allow(clippy::std_instead_of_alloc)]
-#[cfg(feature = "std")]
-impl std::error::Error for NoSuchAlgorithm {}
-
 ffi::ffi_item! {
     /// Algorithm for hashing
-    #[derive(Debug, Display, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default, DeserializeFromStr, SerializeDisplay, Decode, Encode, FfiType, IntoSchema)]
+    #[derive(Debug, Display, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default, DeserializeFromStr, SerializeDisplay, Decode, Encode, IntoSchema)]
```
appetrosyan, line 68: This very likely would have been formatted with `rustfmt` if it weren't inside a function-like macro. Can we momentarily pull these out, format, and then put back in?

crypto/src/lib.rs
```diff
@@ -106,21 +98,23 @@ impl FromStr for Algorithm {
 }
 
 /// Options for key generation
+#[cfg(not(feature = "ffi_import"))]
 #[derive(Debug, Clone)]
-pub enum KeyGenOption {
+enum KeyGenOption {
```
appetrosyan, line 104: Why hide it?

crypto/src/lib.rs
```diff
@@ -163,24 +170,12 @@ impl KeyGenConfiguration {
 
     /// With algorithm
     #[must_use]
-    pub const fn with_algorithm(mut self, algorithm: Algorithm) -> Self {
+    pub fn with_algorithm(mut self, algorithm: Algorithm) -> Self {
```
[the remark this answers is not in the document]
appetrosyan: TBH `const` functions aren't particularly useful, given inline and macros, so I'd just drop the `const`.

crypto/src/lib.rs
```diff
@@ -489,35 +411,26 @@ impl From<PublicKey> for Multihash {
...
 ffi::ffi_item! {
     /// Private Key used in signatures.
-    #[derive(DebugCustom, Display, Clone, PartialEq, Eq, Deserialize, Serialize, FfiType)]
-    #[debug(fmt = "{{ digest: {digest_function}, payload: {payload:X?}}}")]
-    #[display(fmt = "{}", "hex::encode(payload)")]
-    #[allow(clippy::multiple_inherent_impl)]
+    #[derive(Clone, PartialEq, Eq, CopyGetters)]
+    #[cfg_attr(not(feature="ffi_import"), derive(derive_more::DebugCustom, Display, Deserialize, Serialize))]
+    #[cfg_attr(not(feature="ffi_import"), debug(fmt = "{{ digest: {digest_function}, payload: {payload:X?}}}"))]
+    #[cfg_attr(not(feature="ffi_import"), display(fmt = "{}", "hex::encode(payload)"))]
```
appetrosyan, line 418: This seems to be unique to the type, and IMO should be extended to a few more types.

crypto/src/lib.rs
```diff
@@ -560,23 +570,17 @@ impl PrivateKey {
...
 
-impl PrivateKey {
     /// Construct `PrivateKey` from hex encoded string
     ///
     /// # Errors
     ///
     /// If given payload is not hex encoded
-    pub fn from_hex(
-        digest_function: Algorithm,
-        payload: &(impl AsRef<[u8]> + ?Sized),
```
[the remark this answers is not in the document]
appetrosyan: TBH, all it saves is a single `.as_ref()` and removal probably reduced the binary size.

crypto/src/lib.rs
```diff
@@ -547,37 +456,152 @@ impl PrivateKey {
...
+        }
+    }
+
+    #[cfg(feature = "std")]
+    impl std::error::Error for MultihashConvertError {}
+
+    impl From<NoSuchAlgorithm> for MultihashConvertError {
+        fn from(_: NoSuchAlgorithm) -> Self {
+            Self {
+                reason: String::from("Digest function not supported"),
```
appetrosyan, line 491: Probably should delegate to `impl core::fmt::Display for NoSuchAlgorithm` instead.

appetrosyan:
```suggestion
                reason: "Digest function not supported".to_owned(),
```

crypto/src/lib.rs
```diff
@@ -547,37 +456,152 @@ impl PrivateKey {
...
+            }
+        }
+    }
+
+    /// Error which occurs when parsing [`PublicKey`]
+    #[derive(Debug, Display, Clone)]
+    pub enum KeyParseError {
+        /// Decoding hex failed
+        #[display(fmt = "Failed to decode. {_0}: {_1}")]
+        Decode(String, hex::FromHexError),
```
appetrosyan, line 501:
```suggestion
        Decode{
          candidate: String, 
          error: hex::FromHexError
        },
```

This would make it slightly more readable IMO.

crypto/src/lib.rs
```diff
@@ -547,37 +456,152 @@ impl PrivateKey {
...
+    }
+
+    #[cfg(feature = "std")]
+    impl std::error::Error for KeyParseError {}
+    /// Error when dealing with cryptographic functions
+    #[derive(Debug, Display)]
+    pub enum Error {
+        /// Returned when trying to create an algorithm which does not exist
+        #[display(fmt = "Algorithm doesn't exist")] // TODO: which algorithm
+        NoSuchAlgorithm,
```
appetrosyan, line 514: Any reason why it couldn't also accept a unit struct? Semantically speaking any place where we'd want one, we'd also want the other. DRY?

crypto/src/lib.rs
```diff
@@ -547,37 +456,152 @@ impl PrivateKey {
...
+        #[display(fmt = "General error. {_0}")] // This is going to cause a headache
+        Other(String),
+    }
+
+    #[cfg(feature = "std")]
+    #[cfg(not(feature = "ffi_import"))]
+    impl From<ursa::CryptoError> for Error {
+        fn from(source: ursa::CryptoError) -> Self {
+            match source {
+                ursa::CryptoError::NoSuchAlgorithm(_) => Self::NoSuchAlgorithm,
```
appetrosyan, line 540: I can make the type public if this helps.

crypto/src/lib.rs
```diff
@@ -547,37 +456,152 @@ impl PrivateKey {
...
+            }
+        }
+    }
+
+    #[cfg(feature = "std")]
+    impl From<NoSuchAlgorithm> for Error {
+        fn from(_: NoSuchAlgorithm) -> Self {
+            Self::NoSuchAlgorithm
+        }
+    }
```
appetrosyan, line 555: This should be implemented regardless of `std`. What should be made conditional is setting the `source` of the error (which doesn't take place here).

ffi/src/std_impls.rs
```diff
@@ -19,16 +15,14 @@ use crate::{
 // NOTE: This can be contested as it is nowhere documented that String is
 // actually transmutable into Vec<u8>, but implicitly it should be
 // SAFETY: String type should be transmutable into Vec<u8>
-ffi_type! {unsafe impl Transparent for String[Vec<u8>] validated with {|target| core::str::from_utf8(target).is_ok()} }
+ffi_type! {unsafe impl Transparent for String[Vec<u8>] {|target| core::str::from_utf8(target).is_ok()} }
 // NOTE: `core::str::as_bytes` uses transmute internally which means that
 // even though it's a string slice it can be transmuted into byte slice.
-ffi_type! {unsafe impl<'slice> Transparent for &'slice str[&'slice [u8]] validated with {|target| core::str::from_utf8(target).is_ok()} }
+ffi_type! {unsafe impl<'slice> Transparent for &'slice str[&'slice [u8]] {|target| core::str::from_utf8(target).is_ok()} }
```
appetrosyan, line 21: I like the old way better. It's self explanatory. This way it doesn't even really look like a lambda, so I don't immediately know what I'm supposed to do.

ffi/src/option.rs
```diff
@@ -90,5 +110,80 @@ where
...
+            1 => Ok(Some(R::try_from_ffi(source.1, store)?)),
+            _ => Err(FfiReturn::TrapRepresentation),
+        }
+    }
+}
+
+impl<R: FfiWrapperType> CWrapperType<WithoutNiche> for Option<R> {
+    type InputType = Option<R::InputType>;
+    type ReturnType = Option<R::ReturnType>;
+}
```
appetrosyan, line 145:
```suggestion
}

```

ffi/src/option.rs
```diff
@@ -90,5 +110,80 @@ where
...
+impl<R: FfiOutPtr> COutPtr<WithoutNiche> for Option<R> {
+    type OutPtr = FfiTuple2<<u8 as FfiOutPtr>::OutPtr, R::OutPtr>;
+}
```
appetrosyan, line 148:
```suggestion
}

```

ffi/src/option.rs
```diff
@@ -90,5 +110,80 @@ where
...
+
+                let mut value_out_ptr = MaybeUninit::uninit();
+                FfiOutPtrWrite::write_out(value, value_out_ptr.as_mut_ptr());
+                let value_out_ptr = unsafe { value_out_ptr.assume_init() };
+
+                out_ptr.write(FfiTuple2(discriminant_out_ptr, value_out_ptr));
+            }
+        }
+    }
+}
```
appetrosyan, line 175:
```suggestion
}

```

appetrosyan (changes requested):

ffi/src/std_impls.rs
mversic: I like the old way better as well but the macro rules didn't allow me to keep it like that after fixing the bug. I'll try to restore this

crypto/src/lib.rs
mversic: because it's just a container for values set on  `KeyGenConfiguration`. User can never interact with it

core/test_network/src/lib.rs
mversic: you mean uppercase space-separated?

ffi/src/std_impls.rs
mversic: fixed

crypto/src/lib.rs
mversic: can you elaborate what do you mean? yes, lots of types will have this guard

core/test_network/src/lib.rs
mversic: I'll do it in a separate PR

ffi/derive/tests/ui_fail/derive_skip_field.rs
mversic: I would normally prefer a function like macro when macro's expansion is position dependent but in this case this behavior seemed appropriate. I don't see how `no_std` and `no_main` relate to this. I've added a description

crypto/src/hash.rs
```diff
@@ -291,25 +274,47 @@ impl<T: IntoSchema> IntoSchema for HashOf<T> {
...
+#[cfg(any(feature = "ffi_export", feature = "ffi_import"))]
+mod ffi {
+    //! Manual implementations of FFI related functionality
+
+    use super::*;
+
+    // NOTE: Hash is FFI serialized as an array (a pointer in a function call, by value when part of a struct)
+    iroha_ffi::ffi_type! {
+        unsafe impl Transparent for Hash {
+            type Target = [u8; Hash::LENGTH];
```
mversic, line 286: do you think there could be any issue with treating Hash as transmutable into [u8; Hash::LENGTH] considering it's a repr(C) struct with two fields?

I'm slightly concerned about the effects on the calling convention, though probably that's not an issue

core/test_network/src/lib.rs
mversic: opened [#3316](https://github.com/hyperledger/iroha/pull/3316)

crypto/src/lib.rs
appetrosyan: This **should** apply to other types, but right now only applies to this.

crypto/src/hash.rs
appetrosyan: It would be a problem under some circumstances, but nothing comes to mind. IMO, we shouldn't try to solve a problem before it appears.

crypto/src/lib.rs
mversic: it's still not clear to me what you mean here. Can you point me to one such type that would require **this**?

crypto/src/lib.rs
mversic: I think I have addressed this?

crypto/src/lib.rs
mversic: I think I have addressed this?

crypto/src/lib.rs
mversic: this is no longer actual. I'm closing

ffi/tests/ffi_import_opaque.rs
```diff
@@ -128,7 +126,8 @@ fn take_and_return_opaque_ref() {
 #[webassembly_test::webassembly_test]
 fn fallible_output() {
     assert_eq!(Ok(42), OpaqueStruct::fallible_int_output(true));
-    assert!(OpaqueStruct::fallible_int_output(false).is_err());
+    // TODO:
+    //assert!(OpaqueStruct::fallible_int_output(false).is_err());
```
Erigara, line 130: Do you plan to do it in next PR?

Erigara (approved):

ffi/tests/ffi_import_opaque.rs
mversic: yes please, Error handling in FFI has some issues

PR #3083, opened 2023-01-25
[feature] #3082: Provide full data in `Created` event

### LINKED ISSUE #3082: Full delta in metadata events
### Feature request

Extend Iroha metadata events with the full delta. So each time any metadata is modified, the information about the precise kind of modification is retained, including the key and the value. 

### Motivation

Ease of use for orillion

### Who can help?

@mingela

Erigara:

### Description of the Change

`Created` events now contain builder object which provide all necessary data.

### Issue

Closes #3082 

### Benefits

Ease of use for orillion.

### Possible Drawbacks

Events became heavier.

### DIFF
--- a/client/tests/integration/events/data.rs
+++ b/client/tests/integration/events/data.rs
@@ -123,7 +123,8 @@ fn transaction_execution_should_produce_events(executable: Executable, port: u16
     // assertion
     for i in 0..4_usize {
         let domain_id = DomainId::new(i.to_string().parse().expect("Valid"));
-        let expected_event = DomainEvent::Created(domain_id).into();
+        let domain = Domain::new(domain_id);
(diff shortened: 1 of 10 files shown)


appetrosyan (commented):
I would consider adding metadata events just in case. So we can safely say that we have a full implementation of callbacks (i.e. they react to the state of the system when the event takes place).

Erigara:
> I would consider adding metadata events just in case. 

We already have events like `MetadataInserted/MetadataRemoved` which include inserted/deleted key value pair. Or you mean something else?

QuentinI (approved):
LGTM

QuentinI (approved):

appetrosyan (approved):

PR #3091, opened 2023-01-30
[fix] #3056: Fix `tps-dev` benchmark hanging

### LINKED ISSUE #3056: Benches hanging
After observing:
```
2023-01-11T07:42:46.989403Z  WARN request{method=POST path=/transaction version=HTTP/1.1 remote.addr=127.0.0.1:64177}: iroha::torii::routing: Failed to push into queue error=Transaction is already applied
```

the entire consensus reaches a deadlock

Erigara:

### Description of the Change

Problem was that transaction submitter accidentally produced duplicate transactions and it's thread paniced without blockchain reaching enough blocks.

- Fix verification for `If` instruction;
- Add nonce to tps benchmark transactions;
- Log and ignore benchmark transaction submitter;
- Transfer assets if have at least 2;
- Maximum TTL for transaction on client side;

### Issue

Closes #3056.

### Benefits

Stable benchmark.

### Possible Drawbacks

None.

### Usage Examples or Tests

```bash
cargo bench --bench tps-dev
```

### DIFF
--- a/client/benches/tps/utils.rs
+++ b/client/benches/tps/utils.rs
@@ -3,7 +3,16 @@
     clippy::std_instead_of_core,
     clippy::std_instead_of_alloc
 )]
-use std::{fmt, fs::File, io::BufReader, path::Path, str::FromStr as _, sync::mpsc, thread, time};
+use std::{
+    fmt,
+    fs::File,
+    io::BufReader,
+    path::Path,
+    str::FromStr as _,
+    sync::mpsc,
+    thread::{self, JoinHandle},
+    time,
+};
 
 use eyre::{Result, WrapErr};
 use iroha_client::client::Client;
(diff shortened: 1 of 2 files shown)


SamHSmith (approved):
lgtm

core/src/smartcontracts/isi/permissions/combinators.rs
```diff
@@ -47,7 +47,7 @@ impl<V: IsAllowed<Operation = Instruction>> IsAllowed for CheckNested<V> {
                     if_box
                         .otherwise
                         .as_ref()
-                        .map_or(ValidatorVerdict::Skip, |otherwise| {
+                        .map_or(ValidatorVerdict::Allow, |otherwise| {
```
mversic, line 50: can you explain this?

client/benches/tps/utils.rs
```diff
@@ -199,17 +221,39 @@ impl MeasurerUnit {
     }
 
     /// Spawn who periodically submits transactions
-    fn spawn_transaction_submitter(&self) {
+    fn spawn_transaction_submitter(&self, shutdown_signal: mpsc::Receiver<()>) -> JoinHandle<()> {
```
mversic, line 224: thank you, this was annoying

mversic (approved):

core/src/smartcontracts/isi/permissions/combinators.rs
Erigara: sure.
Suppose we have only `then` branch and it returns `Allow`, `else` branch will always return `Skip`. 
Then we select least permissive from `Allow` and `Skip` which is `Skip`, so there is no way to `If` instruction to be valid.

Issue #2373, opened 2022-06-19
Add an Iroha swarm analogue to Iroha 2


appetrosyan:
The Iroha swarm tool allows generating `docker-compose.yml` files for specific numbers of peers. 

Having such a tool would be useful for @Mingela and the DevOps team. 

Issues/PRs that **must** be solved in order to close this issue:

- [x] #3603 
- [x] #3604
- [x] #3605
  - **TODO:** this might be caused by outdated image in Dockerhub. If it works on `iroha2-dev` branch, it should **not** be a part of _this_ issue.
- [x] #3610 
- [ ] #3833

antonovmike:
Hi!

astrokov7:
Someone problem to fix that is we can't run the LTS branch through docker-compose from the box without additional edits to the image name lts instead dev. I've already been stuck here several times to a finding of a problem when pulled the fresh lts.
![telegram-cloud-photo-size-2-5229213929916448939-y.jpg](https://images.zenhubusercontent.com/62ff7ff6035024b7529178f9/d1897e2e-b852-4ea9-b170-6fc32554e0c7)

astrokov7:
Existing developments:
https://github.com/antonovmike/docker_swarm

appetrosyan:
Make it a subcommand of `kagami`.

0x009922:
So, what are the requirements for the tool, and what are the exact problems we need to solve?

My understanding so far about the problems:

- `docker-compose.yml` in the repository is not up-to-date. For example, it refers to `iroha2:dev` image while being in `iroha2-lts` branch and using it's configuration in `configs/peer`. This desynchronisation leads to compatibility issues.
- We have `docker-compose.yml`, `docker-compose-local.yml` and `docker-compose-single.yml`. They all have not many differences, so it's generally a boilerplate which we also need to synchronise (files with each other). Furthermore, they don't cover use cases when the consumer needs some custom configuration with N peers, not 1 or 4.

Could you desribe more with what @Mingela and DevOps struggle with?

My understanding so far about the solution:

1. The tool might be an interactive prompt CLI or a Web application. It depends on what exactly and how we need to build, I don't understand it yet. Also it depends on how the consumer will work with it, what will be more comfortable.
2. I agree with @appetrosyan that this functionality should be a part of `kagami`. It will keep all Iroha accompanying tooling in a single place, like `cargo` or `rustup` for Rust. Also, we have plans to give `kagami` GUI, so the first point might disappear on its own.
[...]

PR #3096, opened 2023-01-30
[feature] #3094: Generate network with `n` peers

### LINKED ISSUE #3094: [suggestion]: Generate iroha network with n peers
### Feature request

Generate iroha network with n peers

### Motivation

It will be useful for exploring fault tolerance behavior.

### Who can help?

_No response_

Erigara:

### Description of the Change

Extend `./scripts/test_env.sh` to work with arbitrary number of nodes.
Add option to kagami to generate peers in compact format.

### Issue

Closes #3094

### Benefits

Easier to test iroha with different number of peers.

### Possible Drawbacks

None.

```bash
# run 10 peers without docker
./scripts/test_env.sh setup bare 10
# run 10 peers with docker
./scripts/test_env.sh setup docker 10
```

### DIFF
--- a/scripts/test_env.sh
+++ b/scripts/test_env.sh
@@ -7,61 +7,63 @@ TEST=${TEST:-"./test"}
 HOST=${HOST:-"127.0.0.1"}
 IROHA2_CONFIG_PATH="$TEST/peers/config.json"
 IROHA2_GENESIS_PATH="$TEST/peers/genesis.json"
+IROHA2_PEER_COUNT=${IROHA2_PEER_COUNT:-"4"}
 
-# TODO: don't hard-code these, instead, generate them.
 declare -A public_keys
-public_keys[iroha0]='ed01201c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b'
-public_keys[iroha1]='ed0120cc25624d62896d3a0bfd8940f928dc2abf27cc57cefeb442aa96d9081aae58a1'
-public_keys[iroha2]='ed0120faca9e8aa83225cb4d16d67f27dd4f93fc30ffa11adc1f5c88fd5495ecc91020'
-public_keys[iroha3]='ed01208e351a70b6a603ed285d666b8d689b680865913ba03ce29fb7d13a166c4e7f1f'
 
 declare -A private_keys
-private_keys[iroha0]='282ed9f3cf92811c3818dbc4ae594ed59dc1a2f78e4241e31924e101d6b1fb831c61faf8fe94e253b93114240394f79a607b7fa55f9e5a41ebec74b88055768b'
-private_keys[iroha1]='3bac34cda9e3763fa069c1198312d1ec73b53023b8180c822ac355435edc4a24cc25624d62896d3a0bfd8940f928dc2abf27cc57cefeb442aa96d9081aae58a1'
-private_keys[iroha2]='1261a436d36779223d7d6cf20e8b644510e488e6a50bafd77a7485264d27197dfaca9e8aa83225cb4d16d67f27dd4f93fc30ffa11adc1f5c88fd5495ecc91020'
-private_keys[iroha3]='a70dab95c7482eb9f159111b65947e482108cfe67df877bd8d3b9441a781c7c98e351a70b6a603ed285d666b8d689b680865913ba03ce29fb7d13a166c4e7f1f'
 
 declare -A p2p_ports
-p2p_ports[iroha0]='1337'
-p2p_ports[iroha1]='1338'
-p2p_ports[iroha2]='1339'
-p2p_ports[iroha3]='1340'
+P2P_STARTING_PORT='1337'
 
 declare -A api_ports
-api_ports[iroha0]='8080'
-api_ports[iroha1]='8081'
-api_ports[iroha2]='8082'
-api_ports[iroha3]='8083'
+API_STARTING_PORT='8080'
 
 declare -A telemetry_ports
-telemetry_ports[iroha0]='8180'
-telemetry_ports[iroha1]='8181'
-telemetry_ports[iroha2]='8182'
-telemetry_ports[iroha3]='8183'
+TELEMETRY_STARTING_PORT='8180'
 
-IROHA_GENESIS_ACCOUNT_PUBLIC_KEY='ed01203f4e3e98571b55514edc5ccf7e53ca7509d89b2868e62921180a6f57c2f4e255'
-IROHA_GENESIS_ACCOUNT_PRIVATE_KEY="{ \"digest_function\": \"ed25519\", \"payload\": \"038ae16b219da35aa036335ed0a43c28a2cc737150112c78a7b8034b9d99c9023f4e3e98571b55514edc5ccf7e53ca7509d89b2868e62921180a6f57c2f4e255\" }"
+function generate_p2p_port {
+    P2P_PORT=$(($P2P_STARTING_PORT + $1))
+    p2p_ports[$1]=$P2P_PORT
+}
+
+function generate_api_port {
+    API_PORT=$(($API_STARTING_PORT + $1))
+    api_ports[$1]=$API_PORT
+}
+
+function generate_telemetry_port {
+    TELEMETRY_PORT=$(($TELEMETRY_STARTING_PORT + $1))
+    telemetry_ports[$1]=$TELEMETRY_PORT
+}
+
+function generate_peer_key_pair {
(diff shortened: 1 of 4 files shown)


scripts/test_env.sh
```diff
@@ -118,8 +183,22 @@ function clean_up_peers {
...
+            N_PEERS="$DEFAULT_N_PEERS"
+        else
+            N_PEERS="$3"
+            if [ "$N_PEERS" -le 0 ]
+            then
+                echo "Expected number of peers as non-zero positive number (> 0). Recieved: $3"
+                exit 1
+            fi
+        fi
+        echo "Strating iroha network with $N_PEERS peers"
```
mversic, line 200:
```suggestion
        echo "Starting iroha network with $N_PEERS peers"
```

mversic (approved):

mversic (approved):

scripts/test_env.sh
```diff
@@ -7,62 +7,66 @@ TEST=${TEST:-"./test"}
 HOST=${HOST:-"127.0.0.1"}
```
appetrosyan, line 7: This code is no longer what I'd call "reasonably small to be best done in `bash`". 

We should consider looking at other alternatives, e.g. `python`: 
https://unix.stackexchange.com/questions/24802/on-which-unix-distributions-is-python-installed-as-part-of-the-default-install

@astrokov7 had the idea of rewriting tests using `python`'s subprocess infrastructure, it might be worth it to do the same here.

scripts/test_env.sh
```diff
@@ -7,62 +7,66 @@ TEST=${TEST:-"./test"}
...
+
+function generate_telemetry_port {
+    TELEMETRY_PORT=$(($TELEMETRY_STARTING_PORT + $1))
+    telemetry_ports[$1]=$TELEMETRY_PORT
+}
+
+function generate_peer_key_pair {
+    mapfile -t -n 3 buffer < <($TEST/kagami crypto -c)
+    public_keys[$1]="${buffer[0]}"
+    private_keys[$1]=$(printf '{"digest_function": "%s", "payload": "%s"}' "${buffer[2]}" "${buffer[1]}")
```
appetrosyan, line 45: Seems like we might be doing something wrong in `kagami` if we need this. Should we adjust the syntax so we can just pipe the output? `kagami crypto -c --json`?

scripts/test_env.sh
```diff
@@ -7,62 +7,66 @@ TEST=${TEST:-"./test"}
...
+}
+
+function generate_genesis_key_pair {
+    mapfile -t -n 3 buffer < <($TEST/kagami crypto -c)
+    IROHA_GENESIS_ACCOUNT_PUBLIC_KEY="${buffer[0]}"
+    IROHA_GENESIS_ACCOUNT_PRIVATE_KEY=$(printf '{"digest_function": "%s", "payload": "%s"}' "${buffer[2]}" "${buffer[1]}")
+}
```
appetrosyan, line 52: I'd wager this is not what we want to do. If the genesis key pair is not specified, sure, generate a fresh pair. But it might not work in all contexts. If the environment variable is defined and set properly, we should just leave it alone and use it directly. 

That was actually the whole point of using environment variables for this purpose.

scripts/test_env.sh
```diff
@@ -7,62 +7,66 @@ TEST=${TEST:-"./test"}
...
 
 function generate_trusted_peers {
     echo -n "["
-    for iter in {0..2}
+    for iter in $(seq 0 $(($1-2)))
     do
-        trusted_peer_entry "iroha$iter"
+        trusted_peer_entry "$iter"
         echo -n ","
     done
```
appetrosyan, line 65: `printf`? It should make sense stylistically.

scripts/test_env.sh
```diff
@@ -90,24 +94,85 @@ function bulk_export {
...
+       generate_telemetry_port $peer
+       generate_peer_key_pair $peer
+       hosts[$peer]="$HOST"
+    done
+    SUMERAGI_TRUSTED_PEERS="$(generate_trusted_peers $1)"
+    for peer in $(seq 1 $(($1-1)))
+    do
+        run_peer $peer
+    done
+    run_peer 0 --submit-genesis
```
appetrosyan, line 125: This isn't quite what I'd like to happen. Overwriting the same environment variable can have unforeseen consequences. Again one more reason to consider writing this logic in Python.

scripts/test_env.sh
```diff
@@ -90,24 +94,85 @@ function bulk_export {
...
+cat <<eof
+  iroha$peer:
+    image: hyperledger/iroha2:dev
+    environment:
+      TORII_P2P_ADDR: '${hosts[$peer]}:${p2p_ports[$peer]}'
+      TORII_API_URL: '${hosts[$peer]}:${api_ports[$peer]}'
+      TORII_TELEMETRY_URL: '${hosts[$peer]}:${telemetry_ports[$peer]}'
+      IROHA_PUBLIC_KEY: '${public_keys[$peer]}'
+      IROHA_PRIVATE_KEY: '${private_keys[$peer]}'
+      SUMERAGI_TRUSTED_PEERS: '$SUMERAGI_TRUSTED_PEERS'
```
appetrosyan, line 163: Strip it out. The docker compose should be generated by a special program. We used to have `iroha_swarm` for iroha 1. 

More on this in https://github.com/hyperledger/iroha/issues/2373

scripts/test_env.sh
```diff
@@ -7,62 +7,66 @@ TEST=${TEST:-"./test"}
 HOST=${HOST:-"127.0.0.1"}
 IROHA2_CONFIG_PATH="$TEST/peers/config.json"
 IROHA2_GENESIS_PATH="$TEST/peers/genesis.json"
+DEFAULT_N_PEERS=4
```
appetrosyan, line 10:
```suggestion
IROHA2_PEER_COUNT=${IROHA2_PEER_COUNT:-"4"}
```

Follow common prefix. While `N` is universally known, the word `DEFAULT` doesn't really convey much. It won't set the number to that value **by default**, but always.

scripts/test_env.sh
```diff
@@ -118,8 +183,22 @@ function clean_up_peers {
 }
 
 case $1 in
-
     setup)
+        declare -i N_PEERS
+        if [ -z "$3" ]
+        then 
+            echo "Number of peers is not provided, using default value of $DEFAULT_N_PEERS"
+            N_PEERS="$DEFAULT_N_PEERS"
```
appetrosyan, line 191: See comment on line 10. Silent fallback might not be the most intuitive, but it is what we do for a few other variables.

scripts/test_env.sh
```diff
@@ -118,8 +183,22 @@ function clean_up_peers {
...
+        else
+            N_PEERS="$3"
+            if [ "$N_PEERS" -le 0 ]
+            then
+                echo "Expected number of peers as non-zero positive number (> 0). Recieved: $3"
+                exit 1
+            fi
```
appetrosyan, line 198: Similarly, this is not what we do for other variables.

scripts/test_env.sh
```diff
@@ -129,22 +208,37 @@ case $1 in
...
+            bare)
                 set_up_peers_common
                 bulk_export
-                run_4_peers
+                run_n_peers "$N_PEERS"
+                ;;
+            *)
+                echo 'Specify either `docker` or `bare`'
+                exit 1
         esac
```
appetrosyan, line 229: Consider stripping out.

scripts/test_env.sh
```diff
@@ -129,22 +208,37 @@ case $1 in
...
         ;;
-
     cleanup)
         case $2 in
             docker)
-                docker-compose rm -s -f;;
-            *)
+                docker-compose -f "$TEST/docker-compose.yml" rm -s -f
+                ;;
```
appetrosyan, line 235: Consider stripping out.

tools/kagami/src/main.rs
```diff
@@ -95,16 +96,25 @@ mod crypto {
         #[clap(long, short, group = "generate_from")]
         seed: Option<String>,
         /// Output the key-pair in JSON format
-        #[clap(long, short)]
+        #[clap(long, short, group = "format")]
         json: bool,
+        /// Output the key-pair without additional text
+        #[clap(long, short, group = "format")]
+        compact: bool,
```
appetrosyan, line 103: I'm all for this format, but IMO the fact that we need to do some `bash` scripting to use this information as part of JSON means that the format probably needs adjustment. Most people would assume that you can just assign the output of kagami to an environment variable and be done with it.

appetrosyan (approved):

pesterev (approved):

PR #3097, opened 2023-01-30
[fix] #3087: Collect votes from observing peers after view change.

SamHSmith:

### DIFF
--- a/client/tests/integration/unstable_network.rs
+++ b/client/tests/integration/unstable_network.rs
@@ -46,7 +46,6 @@ fn soft_fork() {
     );
 }
 
-#[ignore = "ignore, more in #2851"]
 #[test]
 fn unstable_network_7_peers_1_fault() {
     let n_peers = 7;
@@ -84,7 +83,7 @@ fn unstable_network(
     let (network, mut iroha_client) = rt.block_on(async {
         let mut configuration = Configuration::test();
         configuration.queue.maximum_transactions_in_block = MAXIMUM_TRANSACTIONS_IN_BLOCK;
-        configuration.logger.max_log_level = Level::ERROR.into();
+        configuration.logger.max_log_level = Level::INFO.into();
         #[cfg(debug_assertions)]
         {
             configuration.sumeragi.debug_force_soft_fork = force_soft_fork;
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -669,9 +669,16 @@ fn reset_state(
     voting_block: &mut Option<VotingBlock>,
     voting_signatures: &mut Vec<SignatureOf<SignedBlock>>,
     round_start_time: &mut Instant,
+    last_view_change_time: &mut Instant,
     view_change_time: &mut Duration,
 ) {
-    let mut was_commit_or_view_change = current_latest_block_height != *old_latest_block_height;
+    let mut was_commit_or_view_change = false;
+    if current_latest_block_height != *old_latest_block_height {
+        // Round is only restarted on a block commit, so that in the case of
+        // a view change a new block is immediately created by the leader
+        *round_start_time = Instant::now();
+        was_commit_or_view_change = true;
+    }
 
     if current_view_change_index != *old_view_change_index {
         current_topology.rebuild_with_new_view_change_count(current_view_change_index);
@@ -685,7 +692,7 @@ fn reset_state(
 
         *voting_block = None;
         voting_signatures.clear();
-        *round_start_time = Instant::now();
+        *last_view_change_time = Instant::now();
         *view_change_time = pipeline_time;
         info!(addr=%peer_id.address, role=%current_topology.role(peer_id), %current_view_change_index, "View change updated");
     }
@@ -749,6 +756,8 @@ pub(crate) fn run<F: FaultInjection>(
     let mut view_change_time = sumeragi.pipeline_time();
     // Instant when the current round started
     let mut round_start_time = Instant::now();
+    // Instant when the previous view change or round happened.
+    let mut last_view_change_time = Instant::now();
 
     while !should_terminate(&mut shutdown_receiver) {
         if should_sleep {
@@ -797,10 +806,11 @@ pub(crate) fn run<F: FaultInjection>(
             &mut voting_block,
             &mut voting_signatures,
             &mut round_start_time,
+            &mut last_view_change_time,
             &mut view_change_time,
         );
 
-        if round_start_time.elapsed() > view_change_time {
+        if last_view_change_time.elapsed() > view_change_time {
             let role = state.current_topology.role(&sumeragi.peer_id);
 
             if let Some(VotingBlock { block, .. }) = voting_block.as_ref() {
@@ -912,6 +922,17 @@ pub(crate) fn run<F: FaultInjection>(
             Role::ObservingPeer => match message {
                 Some(Message::BlockCreated(BlockCreated { block })) => {
                     if let Some(block) = vote_for_block(sumeragi, &state, block) {
+                        if current_view_change_index >= 1 {
+                            let block_hash = block.block.hash();
+
+                            let msg = MessagePacket::new(
+                                view_change_proof_chain.clone(),
+                                BlockSigned::from(block.block.clone()),
+                            );
+
+                            sumeragi.broadcast_packet_to(msg, [current_topology.proxy_tail()]);
+                            info!(%addr, %block_hash, "Block validated, signed and forwarded");
+                        }
                         voting_block = Some(block);
                     }
                 }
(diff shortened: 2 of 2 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -873,6 +888,7 @@ pub(crate) fn run<F: FaultInjection>(
                                 BlockCreated::from(new_block),
                             );
                             sumeragi.broadcast_packet(msg, current_topology);
+                            node_wants_block = true;
```
Erigara, line 891: I don't see `node_wants_block` set for `ObservingPeer` and `ProxyTail`.

Erigara (changes requested):

core/src/sumeragi/main_loop.rs
```diff
@@ -669,9 +672,15 @@ fn reset_state(
...
     voting_signatures: &mut Vec<SignatureOf<SignedBlock>>,
     round_start_time: &mut Instant,
+    last_view_change_time: &mut Instant,
     view_change_time: &mut Duration,
+    node_wants_block: &mut bool,
 ) {
-    let mut was_commit_or_view_change = current_latest_block_height != *old_latest_block_height;
+    let mut was_commit_or_view_change = false;
+    if current_latest_block_height != *old_latest_block_height {
+        *round_start_time = Instant::now();
```
mversic, line 681: Is this change relevant for this PR?

I suspect this change will bring about issues that would require the logic to become more convoluted. I will try to recall the problems if you think this change is relevant.

core/src/sumeragi/main_loop.rs
```diff
@@ -933,8 +963,15 @@ pub(crate) fn run<F: FaultInjection>(
...
-                            .filter_signatures_by_roles(&[Role::ValidatingPeer], &signatures);
+                        let valid_signatures = if current_view_change_index >= 1 {
+                            current_topology.filter_signatures_by_roles(
+                                &[Role::ValidatingPeer, Role::ObservingPeer],
+                                &signatures,
+                            )
+                        } else {
+                            current_topology
+                                .filter_signatures_by_roles(&[Role::ValidatingPeer], &signatures)
+                        };
```
mversic, line 974:
```suggestion
                        let roles = if current_view_change_index >= 1 {
                            &[Role::ValidatingPeer, Role::ObservingPeer]
                        } else {
                            &[Role::ValidatingPeer]
                        };
                        
                        let valid_signatures = current_topology.filter_signatures_by_roles(roles, &signatures);
```

consider me as a linter :), but this is what I would find more readable

core/src/sumeragi/main_loop.rs
```diff
@@ -911,7 +927,21 @@ pub(crate) fn run<F: FaultInjection>(
             },
             Role::ObservingPeer => match message {
                 Some(Message::BlockCreated(BlockCreated { block })) => {
-                    if let Some(block) = vote_for_block(sumeragi, &state, block) {
+                    if current_view_change_index >= 1 {
```
mversic, line 930: why don't we instead move the node role from observing to validating?

mversic (changes requested):

core/src/sumeragi/main_loop.rs
mversic: is this because you want to prevent view changes when there are no transactions?

core/src/sumeragi/main_loop.rs
Erigara: have the same thoughts

core/src/sumeragi/main_loop.rs
SamHSmith: That would be more complicated in my opinion. We are trying to move closer to the whitepaper. So some things make less sense because we don't do everything that the paper describes yet. But I think it's useful to adhere to it and then make a change like this once we are closer.

client/tests/integration/unstable_network.rs
```diff
@@ -46,7 +46,7 @@ fn soft_fork() {
     );
 }
 
-#[ignore = "ignore, more in #2851"]
+//#[ignore = "ignore, more in #2851"]
```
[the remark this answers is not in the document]
SamHSmith: This was on purpose but I will remove it for you.

core/src/sumeragi/main_loop.rs
SamHSmith: Yea it is in order to differentiate between when consensus is working and when it is not. You would be correct that it doesn't do that much now but for full adherence to the whitepaper this is an important distinction. For instance, now we shuffle all the peers for each view change or block commit. In paper you only shuffle within the A set when things are working, and only when consensus fails do you bring in observing peers. Basically when things are going well observing peers are never involved in block production.

core/src/sumeragi/main_loop.rs
SamHSmith: I will change it.

core/src/sumeragi/main_loop.rs
SamHSmith: https://github.com/hyperledger/iroha/blob/iroha2-dev/docs/source/iroha_2_whitepaper.md#28-consensus

Erigara: > For each round (e.g., block), the previous round's (block's) hash is used to determine an ordering over the set of nodes, such that there is a deterministic and canonical order for each block. In this ordering, the first nodes are grouped into a set called set . Under normal (non-faulty) conditions, consensus for a block is performed by set . The remaining nodes are grouped into a set called set . Under normal conditions, set acts as a passive set of validators to view and receive committed blocks, but otherwise they do not participate in consensus.

It's seems like we actually change set A on every new block.

core/src/sumeragi/main_loop.rs
```diff
@@ -911,7 +921,21 @@ pub(crate) fn run<F: FaultInjection>(
...
+                            );
+
+                            sumeragi.broadcast_packet_to(msg, [current_topology.proxy_tail()]);
+                            info!(%addr, %block_hash, "Block validated, signed and forwarded");
+
+                            voting_block = Some(block);
+                        }
+                    } else if let Some(block) = vote_for_block(sumeragi, &state, block) {
                         voting_block = Some(block);
                     }
```
Erigara, line 940: Consider small refactoring: 

```suggestion
                    if let Some(block) = vote_for_block(sumeragi, &state, block) {
                        if current_view_change_index >= 1 {
                            let block_hash = block.block.hash();

                            let msg = MessagePacket::new(
                                view_change_proof_chain.clone(),
                                BlockSigned::from(block.block.clone()),
                            );

                            sumeragi.broadcast_packet_to(msg, [current_topology.proxy_tail()]);
                            info!(%addr, %block_hash, "Block validated, signed and forwarded");
                        }
                        
                        voting_block = Some(block);
                    }
```

core/src/sumeragi/main_loop.rs
mversic: > For instance, now we shuffle all the peers for each view change or block commit. In paper you only shuffle within the A set when things are working, and only when consensus fails do you bring in observing peers. Basically when things are going well observing peers are never involved in block production

I would agree that we shouldn't involve observing peers when things go well. This way we're not wasting time on view changes to find A set that contains no faulty peers. I believe this is important if we want to maintain high throughput in an environment with a lot of faulty peers (`~f`).
However, in the case when there are no transactions, forgetting which nodes comprised a valid A set, and consequently wasting a few `sumeragi` cycles to find it again, is of little relevance.
Also, if we don't do a view change when there are no transactions we're going to have to bring back that mechanism for forwarding a transaction to a leader to make sure leader isn't faulty

I strive for simplicity. If this isn't convincing I will go about to find a loophole in the logic you implemented because as far as I remember there was one

core/src/sumeragi/main_loop.rs
```diff
@@ -797,10 +806,11 @@ pub(crate) fn run<F: FaultInjection>(
             &mut voting_block,
             &mut voting_signatures,
             &mut round_start_time,
+            &mut last_view_change_time,
             &mut view_change_time,
         );
 
-        if round_start_time.elapsed() > view_change_time {
+        if node_wants_block && last_view_change_time.elapsed() > view_change_time {
```
mversic, line 813: If you've seen my last comment, I had a small change of heart since then. It's good that you introduced `last_view_change_time` so that block is created immediately after a view change, i.e. the round is not restarted. However, I am still advocating that we should do view changes when network is idle as per [these](https://github.com/hyperledger/iroha/pull/3097/files#r1091892835) arguments. This is the only change I advocate in this PR:

```suggestion
        if last_view_change_time.elapsed() > view_change_time {
```

core/src/sumeragi/main_loop.rs
mversic: just add the following comment above (or you can find a more appropriate place):

```suggestion
        // Round is only restarted on a block commit, so that in the case of 
        // a view change a new block is immediately created by the leader
        *round_start_time = Instant::now();
```

core/src/sumeragi/main_loop.rs
mversic: sure, though I don't see this proposal conflicting with the whitepaper

mversic: if nobody has anything to say, feel free to close this conversation before merging

core/src/sumeragi/main_loop.rs
SamHSmith: will do.

core/src/sumeragi/main_loop.rs
SamHSmith: alright then.

mversic: you make the suggested change, I'll open an issue so we can discuss

core/src/sumeragi/main_loop.rs
SamHSmith: good suggestion

mversic (dismissed):

mversic (approved):

PR #3104, opened 2023-02-01
[refactor]: Remove adt_const_generics nightly feature

mversic:

### Description of the Change

Relevant [discussion](https://github.com/hyperledger/iroha/pull/3047#discussion_r1091485179)
If many of you disagree we'll reject this PR

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -29,7 +29,7 @@ use serde::Serialize;
 use crate::{
     prelude::*,
     sumeragi::network_topology::{Role, Topology},
-    tx::{TransactionOrigin, TransactionValidator, VersionedAcceptedTransaction},
+    tx::{TransactionValidator, VersionedAcceptedTransaction},
 };
 
 /// Default estimation of consensus duration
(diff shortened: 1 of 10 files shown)


core/src/block.rs
```diff
@@ -482,14 +482,14 @@ impl VersionedCandidateBlock {
     /// # Errors
     /// Forward errors from [`CandidateBlock::revalidate`]
     #[inline]
-    pub fn revalidate<const TX_ORIGIN: TransactionOrigin>(
+    pub fn revalidate<const IS_GENESIS: bool>(
```
mversic, line 485:
```suggestion
    pub fn revalidate<const IS_GENESIS: bool = false>(
```

I think this might be supported in the future which would make API much nicer for the default case

Erigara: this feature would be really nice to have

Erigara (approved):
Looks good for me. 

I find approach with enum to be too verbose + const generics enums  don't seems to be stabilized soon.

SamHSmith (approved):

appetrosyan (dismissed):
I'm not a huge fan of moving away, but given that the feature is marked as *incomplete* even though it should not break in any context the fact that this would let people use a stable compiler might be worth the trade-off in terms of readability.

appetrosyan (approved):

PR #3108, opened 2023-02-02
[feature] #3099: Sub-routing for status endpoint

### LINKED ISSUE #3099: [DOPS request] Add sub-routing to `status` endpoint
As per @BAStos525 's request, we should update the routing in Torii, to supplement the regular `/status` endpoint with sub-endpoints. Specifically, `/status/<param>` route should return the JSON object that corresponds to the field `<param>` in the struct `iroha_telemetry::metrics::Status`. 

## Example usage
```bash
curl http://127.0.0.1:8180/status/peers
``` 
returns `7` as JSON.

appetrosyan:
### Description

Improved sub-routing of the `status` endpoint. 

### Issue

Closes #3099 

### Benefits

One does not need to decode JSON in order to obtain a subset of the data. Specifically this can be used for makeshift healthchecking. 

### Possible Drawbacks

None

```sh
curl http://127.0.0.1:8180/status/blocks
```

Should return a single number, while 

```sh
curl curl http://127.0.0.1:8180/status
``` 

should return a set of JSON values.


cli/src/lib.rs
```diff
@@ -473,6 +485,84 @@ fn domains(configuration: &Configuration) -> [Domain; 1] {
...
+        pub fn or(&self, arg_group: &[&str; 2]) -> String {
+            format!(
+                "`{}` (short `{}`)",
+                arg_group[0].style(self.positive),
+                arg_group[1].style(self.minor)
+            )
+        }
+
+        /// Convenience method for ".json or .json5" pattern
+        pub fn json_ext(&self, name: &str) -> String {
```
ilchu, line 554:
```suggestion
        pub fn with_json_ext(&self, name: &str) -> String {
```
This could be more easily understable. Otherwise it reads almost inverted.

telemetry/src/metrics.rs
```diff
@@ -81,7 +81,7 @@ pub struct Metrics {
     pub view_changes: GenericGauge<AtomicU64>,
     /// Number of transactions in the queue
     pub queue_size: GenericGauge<AtomicU64>,
-    // Internal use only.
+    /// Internal use only. Needed for generating the response.
```
ilchu, line 84: Good catch. Appreciate the better docs wording throughout the PR overall.

cli/src/torii/routing.rs
```diff
@@ -423,16 +425,38 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -
...
+fn handle_status(sumeragi: &Sumeragi) -> Result<warp::reply::Json, Infallible> {
     if let Err(error) = sumeragi.update_metrics() {
         iroha_logger::error!(%error, "Error while calling `sumeragi::update_metrics`.");
     }
     let status = Status::from(&sumeragi.metrics_mutex_access());
     Ok(reply::json(&status))
 }
 
+#[cfg(feature = "telemetry")]
+#[allow(clippy::unused_async)]
```
ilchu, line 438: Do we still need this handler to have `async`, even though the other ones got rid of it?

cli/src/torii/routing.rs
```diff
@@ -204,8 +203,7 @@ enum Health {
     Healthy,
 }
 
-#[iroha_futures::telemetry_future]
```
ilchu, line 207: Just because it's not a particularly useful metric, I suppose?

scripts/test_env.sh
```diff
@@ -97,6 +97,8 @@ function run_peer () {
     STORAGE="$PEER/storage"
     mkdir -p "$STORAGE"
     KURA_BLOCK_STORE_PATH="$STORAGE"
+	LOG_FILE_PATH="$PEER/log.json"
+	MAX_LOG_LEVEL="TRACE"
```
ilchu, line 101: Speaking about the `test_env` script, it seems to fail for me when I run it with either the current `iroha2-dev` hash or likewise on this branch, complaining about failing to build the proxy without the `debug_force_soft_fork` being set -- so it was introduced by. Docker compose seems is still building fine. So it feels like despite what was mentioned in the soft-fork PR [discussion](https://github.com/hyperledger/iroha/pull/3064#discussion_r1073378147), it's still being exposed technically, as it's become part of the `ConfigurationProxy` via the derive macro. Should we create an issue for this?

cli/src/lib.rs
```diff
@@ -473,6 +485,84 @@ fn domains(configuration: &Configuration) -> [Domain; 1] {
...
+                positive: Style::new().green().bold(),
+                negative: Style::new().red().bold(),
+                highlight: Style::new().bold(),
+                minor: Style::new().green(),
+            }
+        }
+    }
+
+    /// Determine if message colourisation is to be enabled
+    pub fn should_enable_colour() -> bool {
```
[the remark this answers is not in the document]
appetrosyan: Refactoring artifact, it used to be the other way around, until I turned it into a series of `unwrap_or`.

cli/src/lib.rs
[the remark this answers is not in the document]
appetrosyan: Good point.

cli/src/main.rs
```diff
@@ -1,58 +1,88 @@
...
 use iroha_config::path::Path as ConfigPath;
 use iroha_core::prelude::AllowAll;
 use iroha_permissions_validators::public_blockchain::default_permissions;
+use owo_colors::OwoColorize as _;
+
+static HELP_ARG: [&str; 2] = ["--help", "-h"];
+static SUBMIT_ARG: [&str; 2] = ["--submit-genesis", "-s"];
+static VERSION_ARG: [&str; 2] = ["--version", "-V"];
+
+static REQUIRED_ENV_VARS: [(&str, &str); 7] = [
```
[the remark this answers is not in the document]
appetrosyan: Good point. They used to be `lazy_static!` but I forgot to change the type.

cli/src/torii/routing.rs
appetrosyan: > Just because it's not a particularly useful metric,

Yes. But also because the healthcheck is supposed to be quick and lightweight. Because you're going through the telemetry subsystem with its own `Future` the way it used to be pessimised the performance quite a bit, and not just for the health check. Just because the health check is running in its own thread doesn't mean it can't clog the `async` pipeline.

scripts/test_env.sh
appetrosyan: Aha! 

So here's my take on why this is. I'll try to fix it in this PR. 

Whe @SamHSmith was implementing the fix for #1170, we found that it wasn't possible to hide the configuration variable behind a feature flag, because of how `cargo` resolves optional features for multiple configurations. We ended up saying that the configuration parameter should only be present if `cfg(debug_assertions)` is also enabled. So a debug build (which we use for the testing) also retains that parameter as a requirement. 

Stop-gap solution is to add an env-var to silence the warning. 

Better solution is to make the variable optional and if it's not provided, set it to false. 

I'd also add a 
```rust 
warn!(%force_soft_forks=config.debug_force_soft_fork, "Running Iroha Debug build")
```
Just to be safe.

cli/src/torii/routing.rs
appetrosyan: Will fix in #3135.

ilchu (dismissed):
LGTM

BAStos525 (commented):
I conform that we got the parsed values of `peers`, `blocks`, `txs_accepted`, `txs_rejected`, `uptime`, `view_changes` and `queue_size` via sub-routing status endpoint.

BAStos525:
We shouldn't forget to update the [README](https://github.com/hyperledger/iroha/blob/iroha2-dev/docs/source/references/api_spec.md#endpoints-for-statusmetrics).

cli/src/lib.rs
```diff
@@ -200,7 +214,7 @@ impl Iroha {
             config,
             instruction_judge,
             query_judge,
-            broker,
+            Broker::new(),
```
mversic, line 217: broker can be implicit, i.e. removed from arguments?

cli/src/torii/routing.rs
```diff
@@ -278,14 +276,16 @@ async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::R
 
     let mut interval = tokio::time::interval(std::time::Duration::from_millis(10));
     loop {
+        // FIXME: cleanup.
```
mversic, line 279: what's this about?

telemetry/src/metrics.rs
```diff
@@ -198,7 +198,7 @@ impl Metrics {
     ///
     /// # Panics
     /// Never
-    #[allow(clippy::unused_self, clippy::expect_used)]
+    #[allow(clippy::unused_self)]
     pub fn current_time(&self) -> Duration {
```
mversic, line 202: why do we have `unused_self`?

scripts/test_env.sh
```diff
@@ -97,11 +98,14 @@ function run_peer () {
...
     TORII_P2P_ADDR="$HOST:${p2p_ports[$1]}"
     TORII_API_URL="$HOST:${api_ports[$1]}"
     TORII_TELEMETRY_URL="$HOST:${telemetry_ports[$1]}"
     IROHA_PUBLIC_KEY="${public_keys[$1]}"
     IROHA_PRIVATE_KEY="${private_keys[$1]}"
+	SUMERAGI_DEBUG_FORCE_SOFT_FORK="false"
```
mversic, line 108: formatting

mversic: for other above as well

mversic: you seem to have a problem with bash scripts

cli/src/torii/routing.rs
```diff
@@ -423,16 +425,38 @@ async fn handle_metrics(sumeragi: Arc<Sumeragi>, _network: Addr<IrohaNetwork>) -
...
+    // elide the creation of the value, and directly read the value
+    // behind the mutex.
+    let status = Status::from(&sumeragi.metrics());
+    match serde_json::to_value(status) {
+        Ok(value) => Ok(value
+            .get(segment)
+            .map_or_else(|| reply::json(&value), reply::json)),
+        Err(err) => {
+            iroha_logger::error!(%err, "Error while converting to JSON value");
+            Ok(reply::json(&None::<String>))
```
mversic, line 455: interesting, I didn't know you can add generic parameter to enum variant?

mversic (approved):

docs/source/references/api_spec.md
```diff
@@ -248,6 +248,8 @@ Also returns current status of peer in json string:
 ```
 __CAUTION__: Almost all fields are 64-bit integers and should be handled with care in JavaScript. Only the `nanos` field is 32-bit integer. See `iroha_telemetry::metrics::Status`.
 
+**Sub-routing**: To obtain the value of a specific field, one can append the name of the field to the path, e.g. `status/peers`. This returns the corresponding JSON value, inline, so strings are quoted, numbers are not and maps are presented as above.
```
mversic, line 251:
```suggestion
**Sub-routing**: To obtain the value of a specific field, one can append the name of the field to the path, e.g. `status/peers`. This returns the corresponding JSON value, inlined
```

This maybe sounds redundant?

cli/src/lib.rs
appetrosyan: I'd like to do a separate pass on those functions if you don't mind. We have an issue to refactor them into builders.

cli/src/torii/routing.rs
appetrosyan: Lots of refactoring that needs to be done in the loop.

telemetry/src/metrics.rs
appetrosyan: So that we can call this with the instance method syntax `thing.current_time`. In general this isn't a problem, it just signals that the unused parameter might be unintentional.

scripts/test_env.sh
appetrosyan: New work machine. Emacs by default uses tabs for shell scripts and C. I haven't had the time to fully migrate my old config to M1.

cli/src/torii/routing.rs
appetrosyan: Normally you don't need to.

docs/source/references/api_spec.md
appetrosyan: Can't be too careful. This *might* be redundant, but it saves the reader some thinking.

mversic (approved):

BAStos525 (approved):

PR #3118, opened 2023-02-07
[feature] #3116: Add sorting of any values (not only u128)

### LINKED ISSUE #3116: Sorting doesn't work if a value isn't `u128` type.
### Feature request

Currently, the Iroha 2 sorts a query result that only [has metadata with u128 as a value.](https://github.com/hyperledger/iroha/blob/iroha2-stable/cli/src/torii/routing.rs#L168-L174) That is a problem. It was probably implemented that way because there was no type cover generic number value then. So now we have `Value::Numeric` and can use it to compare metadata values. Or even can try to implement `Ord` for `Value` type.

### Motivation

Better sorting feature.

### Who can help?

@appetrosyan

pesterev:

### Description of the Change

Now we can sort by any values from metadata. Not only by u128. Also, this sorting implementation places all elements that cannot be sorted at the head of a result list. 

### Issue

Resolves #3116 

### Benefits

Better sorting feature.

### Possible Drawbacks

None

Take a look at tests in these changes.

Another design is to remove all elements that we can't sort.

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -6,7 +6,7 @@
 // FIXME: This can't be fixed, because one trait in `warp` is private.
 #![allow(opaque_hidden_inferred_bound)]
 
-use std::num::TryFromIntError;
+use std::{cmp::Ordering, num::TryFromIntError};
 
 use eyre::WrapErr;
 use futures::TryStreamExt;
@@ -170,32 +170,44 @@ pub(crate) async fn handle_queries(
 }
 
 fn apply_sorting_and_pagination(
-    mut vec_of_val: Vec<Value>,
+    vec_of_val: Vec<Value>,
     sorting: &Sorting,
     pagination: Pagination,
 ) -> Vec<Value> {
     if let Some(ref key) = sorting.sort_by_metadata_key {
-        let f = |value1: &Value| {
-            if let Value::Numeric(NumericValue::U128(num)) = value1 {
-                *num
-            } else {
-                0
-            }
-        };
-
-        vec_of_val.sort_by_key(|value0| match value0 {
-            Value::Identifiable(IdentifiableBox::Asset(asset)) => match asset.value() {
-                AssetValue::Store(store) => store.get(key).map_or(0, f),
-                _ => 0,
+        let mut pairs: Vec<(Option<Value>, Value)> = vec_of_val
+            .into_iter()
+            .map(|value| {
+                let key = match &value {
+                    Value::Identifiable(IdentifiableBox::Asset(asset)) => match asset.value() {
+                        AssetValue::Store(store) => store.get(key).cloned(),
+                        _ => None,
+                    },
+                    Value::Identifiable(v) => TryInto::<&dyn HasMetadata>::try_into(v)
+                        .ok()
+                        .and_then(|has_metadata| has_metadata.metadata().get(key))
+                        .cloned(),
+                    _ => None,
+                };
+                (key, value)
+            })
+            .collect();
+        pairs.sort_by(
+            |(left_key, _), (right_key, _)| match (left_key, right_key) {
+                (Some(l), Some(r)) => l.cmp(r),
+                (Some(_), None) => Ordering::Less,
+                (None, Some(_)) => Ordering::Greater,
+                (None, None) => Ordering::Equal,
             },
-            Value::Identifiable(v) => TryInto::<&dyn HasMetadata>::try_into(v)
-                .map(|has_metadata| has_metadata.metadata().get(key).map_or(0, f))
-                .unwrap_or(0),
-            _ => 0,
-        });
+        );
+        pairs
+            .into_iter()
+            .map(|(_, val)| val)
+            .paginate(pagination)
+            .collect()
+    } else {
+        vec_of_val.into_iter().paginate(pagination).collect()
     }
-
-    vec_of_val.into_iter().paginate(pagination).collect()
 }
 
 #[derive(serde::Serialize)]
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -777,6 +777,47 @@ impl Client {
         )
     }
 
+    /// Create a request with sorting and the given filter.
+    ///
+    /// # Errors
+    /// Fails if sending request fails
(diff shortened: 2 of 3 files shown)


client/src/client.rs
```diff
@@ -777,6 +777,47 @@ impl Client {
...
+        R: Query + Into<QueryBox> + Debug,
+        <R::Output as TryFrom<Value>>::Error: Into<eyre::Error>, // Seems redundant
+    {
+        self.request_with_pagination_and_filter_and_sorting(
+            request,
+            Pagination::default(),
+            sorting,
+            filter,
+        )
+    }
```
appetrosyan, line 800: Do we really need this function?

appetrosyan (dismissed):

client/src/client.rs
pesterev: Not really, but it's convenient function. Moreover we have `request_with_pagination_and_filter`.

pesterev:
Also, @arndey suggests adding one more query parameter to specify what the sorting should do with unsortable elements. For example,
sort_algo = "remove" - remove all unsortable elements
sort_algo = "head" - place unsortable elements at the head of a list
sort_algo = "tail" - place unsortable elements at the tail of a list

client/src/client.rs
Erigara: We have few places where we have functions like `do_X_and_Y`, `do_X_and_Z` which rely internally on `do_X_and_Y_and_Z`, maybe we can come up with better approach.

Erigara:
To me it's sound like unnecessary complications on the backend side.

Erigara (approved):

client/src/client.rs
appetrosyan: Maybe a proc macro? Optional arguments used to be a thing in Rust functional.

appetrosyan: Maybe a proc macro? Optional arguments used to be a thing in Rust functional.

pesterev: My suggestion is to make a builder (called `QueryRequestBuilder`) and to remove query `request_with_*` methods.

pesterev:
> To me it's sound like unnecessary complications on the backend side.

@Erigara we have to choose one of the behavior algorithms anyway if we aren't going to implement @arndey  suggestion. The current implementation in this PR is to place unsortable elements at the head of a result list. If it's okay, let's merge it.

Erigara:
> place unsortable elements at the head of a result list
fine to me

appetrosyan (dismissed):

client/src/client.rs
Erigara: Agree with builder approach. 
Other option we might consider: [link](https://www.thecodedmessage.com/posts/default-params/).

pesterev: I've opened [an issue](https://github.com/hyperledger/iroha/issues/3124) and mentioned your suggestion there.

pesterev:
Dear reviewers,

Andrey (@arndey) and I decided to change the algorithm of ordering unsortable elements, so now it places unsortable elements at the tail of a result list. And it makes sense. This variant of implementation is more convenient for manipulating a result list.

For example,
We have a list like this:
[ `A` has `1` in metadata, `B` has `0` in metadata, `C` has `nothing` in metadata, `D` also has `nothing` in metadata, `E` has `2` in metadata ]

The previous implementation returned => [`C`, `D`, `B`, `A`, `E`]
The current implementation will return => [`B`, `A`, `E`, `C`, `D`]

client/tests/integration/sorting.rs
```diff
@@ -52,7 +55,7 @@ fn correct_pagination_assets_after_creating_new_one() {
     let res = test_client
         .request_with_pagination_and_sorting(
             client::asset::by_account_id(account_id.clone()),
-            Pagination::new(Some(1), Some(5)),
+            Pagination::new(None, Some(5)),
```
pesterev, line 58: We skipped the `rose` here in the previous implementation. Now, the `rose` is placed at the tail.

cli/src/torii/routing.rs
```diff
@@ -170,32 +170,39 @@ pub(crate) async fn handle_queries(
 }
 
 fn apply_sorting_and_pagination(
-    mut vec_of_val: Vec<Value>,
+    vec_of_val: Vec<Value>,
     sorting: &Sorting,
     pagination: Pagination,
 ) -> Vec<Value> {
+    let mut iter: Box<dyn Iterator<Item = Value>> = Box::new(vec_of_val.into_iter());
```
appetrosyan, line 177: You need this binding to get rid of the `mut` in the signature?

appetrosyan (dismissed):

cli/src/torii/routing.rs
pesterev: I tried to use if-else to avoid `collect` calls and to get an iterator, but the iterators of both branches didn't match. That's why I used an object trait.

cli/src/torii/routing.rs
```diff
@@ -170,32 +170,44 @@ pub(crate) async fn handle_queries(
...
-
-        vec_of_val.sort_by_key(|value0| match value0 {
-            Value::Identifiable(IdentifiableBox::Asset(asset)) => match asset.value() {
-                AssetValue::Store(store) => store.get(key).map_or(0, f),
-                _ => 0,
+        let mut pairs: Vec<(Option<Value>, Value)> = vec_of_val
+            .into_iter()
+            .map(|value| {
+                let key = match &value {
+                    Value::Identifiable(IdentifiableBox::Asset(asset)) => match asset.value() {
```
appetrosyan, line 182: This `match` arm shadows numerical values which cannot be sorted by metadata. Have you tried a more specific pattern before a more general one? 

```suggestion
                    Value::Identifiable(IdentifiableBox::Asset(AssetValue(asset))) => asset.get(key).cloned()
```

appetrosyan (approved):

Erigara (approved):

cli/src/torii/routing.rs
pesterev: Opened [an issue](https://github.com/hyperledger/iroha/issues/3139) about sorting numeric values.

PR #3121, opened 2023-02-08
[feature] #3127: make data model structs opaque by default

### LINKED ISSUE #3127: Opaque/transparent API
Write a macro that will expose data model items as opaque when linked by clients but transparent when linked by iroha internal libraries. The behavior is controlled with `transparent_api` flag (renamed from `mutable_api`)

mversic:
### Description of the Change

The issue I've been facing is how to make data model structs transparent for `iroha_core` and opaque for the client:
* wrote a simple macro that converts a struct to an opaque struct when linked without `transparent_api` feature
* `transparent_api` should only be activated by the inner libraries of Iroha
* I opted for a function like macro because I want to incorporate FFI API expansion into this macro
* rename `warp` feature flag to `http`
* FFI feature is orthogonal to this and added with `ffi_type`

### Issue

Closes #3127

### Benefits

With this change we get the best of both worlds. We get opaque API for clients and transparent API for inner libraries of Iroha. I see no reason why Iroha libs wouldn't have access to inner workings of data model structs. It only forces us to clone or make weird workarounds 

### Possible Drawbacks

Not sure what can be done about opaque enums


data_model/src/account.rs
```diff
@@ -237,20 +235,18 @@ declare_item! {
     #[display(fmt = "({id})")] // TODO: Add more?
     pub struct Account {
         /// An Identification of the [`Account`].
-        id: Id,
+        pub id: Id,
```
appetrosyan, line 238: So if I understand it correctly, the public fields are always going to be available, so `Account` pointers have a transparent section, right?

data_model/src/account.rs
```diff
@@ -237,20 +235,18 @@ declare_item! {
...
         /// Assets in this [`Account`].
-        assets: AssetsMap,
+        pub assets: AssetsMap,
```
appetrosyan, line 240: So AssetsMap is itself opaque, but we can assign the assets field and don't need to go through two function calls to modify the map. Only one of those two calls would go through a VTable, so I wonder how much do we actually save.

data_model/src/account.rs
mversic: will be transparent only for `iroha_core`. Macro will make them public fields `pub(crate)` by default unless you activate `transparent_api`

data_model/derive/src/lib.rs
```diff
@@ -1,13 +1,42 @@
...
+        }
+
+        Ok(Self(items))
+    }
+}
+
+/// Macro for replacing
+#[proc_macro]
+#[proc_macro_error::proc_macro_error]
+pub fn model(input: TokenStream) -> TokenStream {
```
mversic, line 29: I will update the doc here

data_model/src/permission/validator.rs
```diff
@@ -45,12 +43,12 @@ declare_item! {
...
     pub struct Validator {
         id: Id,
-        #[getset(get = "pub")]
         /// Type of the validator
+        #[getset(get = "pub")]
         validator_type: Type,
-        // TODO: use another type like `WasmValidator`?
-        /// WASM code of the validator
         #[getset(get = "pub")]
+        /// WASM code of the validator
```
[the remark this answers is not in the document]
mversic: ok

mversic:
somebody made `iroha_core` a dependency of `iroha_client` so tests are compiled with `transmutable_api` when they shouldn't be. I'll try to take care of this in the future

Erigara (dismissed):
Clever design, fix doc test and will be good to go.

configs/peer/genesis.json
```diff
@@ -296,186 +296,184 @@
...
-          "Sequence": {
-            "instructions": [
-              {
-                "NewParameter": {
-                  "Parameter": "?BlockSyncGossipPeriod=10000",
-                  "source_id": {
-                    "AccountId": "alice@wonderland"
-                  }
+          "Sequence": [
+            {
```
mversic, line 300: API change

docs/source/references/schema.json
```diff
@@ -491,7 +491,7 @@
         {
           "name": "Parameter",
           "discriminant": 13,
-          "ty": "iroha_data_model::Parameter"
+          "ty": "iroha_data_model::parameter::Parameter"
```
mversic, line 494: API change

data_model/src/query.rs
```diff
@@ -155,39 +144,47 @@ impl Payload {
...
+#[repr(transparent)]
 pub struct QueryResult(pub Value);
 
 declare_versioned_with_scale!(VersionedPaginatedQueryResult 1..2, Debug, Clone, iroha_macro::FromVariant, IntoSchema);
 
 /// Paginated Query Result
-#[version_with_scale(n = 1, versioned = "VersionedPaginatedQueryResult")]
 #[derive(Debug, Clone, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+#[version_with_scale(n = 1, versioned = "VersionedPaginatedQueryResult")]
 pub struct PaginatedQueryResult {
```
mversic, line 188: not wrapped into `model` atm, because client should have access to the fields of query result. Could be contended

Erigara (dismissed):

data_model/derive/src/lib.rs
```diff
@@ -1,13 +1,82 @@
...
+/// use iroha_data_model_derive::model;
+///
+/// model! {
+///     struct DataModel {
+///        pub item1: u32,
+///        item2: u64
+///     }
+/// }
+///
+/// /* will produce:
```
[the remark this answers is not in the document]
mversic: I am aware, but i believe it doesn't matter

cli/src/torii/pagination.rs
```diff
@@ -0,0 +1,130 @@
+use iroha_data_model::prelude::*;
+
+/// Describes a collection to which pagination can be applied.
+/// Implemented for the [`Iterator`] implementors.
+pub trait Paginate: Iterator + Sized {
+    /// Returns a paginated [`Iterator`].
```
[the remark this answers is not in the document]
mversic: I was wondering where have you been :)

data_model/derive/src/api.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    if ffi_attrs.next().is_none() {
+        return quote! {
+            #(#attrs)*
+            #item
+        };
+    }
+
+    attrs.retain(|attr| *attr != syn::parse_quote! (#[ffi_type]));
+    let no_ffi_attrs: Vec<_> = attrs
```
QuentinI, line 16: I don't quite get what this does. What's the difference between `!attr.path.is_ident("ffi_type")` and `attr != syn::parse_quote! (#[ffi_type])`?

data_model/src/account.rs
```diff
@@ -47,230 +46,94 @@ pub type AccountsMap = btree_map::BTreeMap<<Account as Identifiable>::Id, Accoun
...
-        FfiType,
-        IntoSchema,
-    )]
+    #[derive(Debug, Display, Clone, IdEqOrdHash, Decode, Encode, Deserialize, Serialize, IntoSchema)]
     #[display(fmt = "[{id}]")]
+    #[ffi_type]
     pub struct NewAccount {
         /// Identification
-        pub id: <Account as Identifiable>::Id,
+        id: <Account as Identifiable>::Id,
```
QuentinI, line 103: Is there a reason to make it private? We lose the ability to easily disassemble it into constituent fields without cloning

docs/source/guides/develop-iroha-module.md
```diff
@@ -125,10 +125,10 @@ And see what it does to register a new Bridge:
 
 We will not discuss Bridge-related terminology here – the thing we want to look at is how we can compose these steps into one new Iroha Special Instruction.
 
-As you can see, we have `Instruction::If(...)` here. It's [the utility Iroha Special Instruction](references/glossary#utility-iroha-special-instruction).
+As you can see, we have `Instruction::from(...)` here. It's [the utility Iroha Special Instruction](references/glossary#utility-iroha-special-instruction).
```
QuentinI, line 128: This paragraph doesn't make sense with this change

version/derive/src/lib.rs
```diff
@@ -397,3 +401,19 @@ fn impl_declare_versioned(
         #json_impl
     )
 }
+
```
QuentinI, line 404: This is probably a leftover

QuentinI (commented):
As non-core consumer of data model, I quite like this change. I propose also generating something like `into_raw(self) -> (Field1, Field2, ...)` function for all the models, it would make my life even easier :)

data_model/src/account.rs
QuentinI: Same question goes for other `id`s

docs/source/guides/develop-iroha-module.md
mversic: a lot of things in this file are outdated

data_model/src/account.rs
mversic: I guess that's a valid point, I'll make them public. I left them hidden since `IdEqOrdHash` provides the `id` method.

data_model/derive/src/lib.rs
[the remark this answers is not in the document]
mversic: handling this would make the code more complicated, I think it's quite easy to see this is a no-op

mversic:
> As non-core consumer of data model, I quite like this change. I propose also generating something like `into_raw(self) -> (Field1, Field2, ...)` function for all the models, it would make my life even easier :)

I somewhat agree with you and this was the solution I considered. This macro could be "improved" with `impl From<Obj> for (Field1, Field2, ...)`. Can you tell me about a use-case for this? If the use-case is convincing I'll make an issue

docs/source/guides/develop-iroha-module.md
appetrosyan: Remove the file. It's better to have no docs than misleading ones.

data_model/derive/src/api.rs
QuentinI: Got it, thanks

QuentinI:
@mversic `into_raw` would help me immensely for upgrades in iroha squash, since most of what I do there is disassemble old items into the building blocks and reassemble them back into newer versions, making changes if necessary. For items with private fields it's a lot of cloning currently, sometimes I even have to resort to passing stuff through serde, which is prone to breakage.

appetrosyan:
Closed as not planned.

mversic:
> Closed as not planned.

I'll rebase it now, I thought it was a different PR. This one is ready to be merged

QuentinI (approved):

appetrosyan:
Tests?

data_model/src/error.rs
Erigara, line 1: Do we need this file at all after your changes?

data_model/src/isi.rs
```diff
@@ -903,14 +742,178 @@ impl NewParameterBox {
...
+    model! {
+        /// Instruction execution error type
+        #[derive(Debug, Display, FromVariant)]
+        #[cfg_attr(feature = "std", derive(thiserror::Error))]
+        // TODO: Only temporarily opaque because of InstructionExecutionFailure::Repetition
+        #[ffi_type(opaque)]
+        pub enum InstructionExecutionFailure {
+            /// Failed to find some entity
+            #[display(fmt = "Failed to find. {_0}")]
+
```
Erigara, line 773: I wonder why `rustfmt` won't remove this empty new line.

appetrosyan: `rustfmt` is fairly limited in what it changes. For example it doesn't convert tabs into spaces unless you tell it to do so (and for good reason). 

In most cases we assume that blank lines have a meaning, and help with readability. In this case it seems like a mistake.

Erigara (dismissed):

data_model/src/error.rs
mversic: no, I just noticed and removed it in #3072

core/src/lib.rs
```diff
@@ -102,6 +102,81 @@ pub mod handler {
...
+        /// impl core::fmt::Display for Keys {
+        ///     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        ///         format_comma_separated(self.0.iter().back_quoted(), ('{', '}'), f)
+        ///     }
+        /// }
+        ///
+        /// let keys = Keys(["key1".to_owned(), "key2".to_owned()]);
+        /// assert_eq!(keys.to_string(), "{`key1`, `key2`}");
+        /// ```
+        fn back_quoted(self) -> BackQuotedIterator<T, Self>;
```
appetrosyan, line 166: Consider extracting this into `iroha_primitives`.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+            }
+
+            let trigger_id = trigger.id;
+            map.insert(trigger_id.clone(), trigger.action);
+            self.ids.insert(trigger_id, event_type);
+            true
+        }
+
+        /// Get all contained trigger ids without a particular order
+        pub fn ids(&self) -> Vec<TriggerId> {
```
appetrosyan, line 110: `inline`.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+        ///
+        /// Returns `false` if a trigger with given id already exists
+        pub fn add_time_trigger(&self, trigger: Trigger<TimeEventFilter>) -> bool {
+            self.add_to(trigger, EventType::Time, &self.time_triggers)
+        }
+
+        /// Add trigger with [`ExecuteTriggerEventFilter`]
+        ///
+        /// Returns `false` if a trigger with given id already exists
+        pub fn add_by_call_trigger(&self, trigger: Trigger<ExecuteTriggerEventFilter>) -> bool {
```
appetrosyan, line 86: `inline`

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+        ///
+        /// Return `false` if a trigger with given id already exists
+        pub fn add_pipeline_trigger(&self, trigger: Trigger<PipelineEventFilter>) -> bool {
+            self.add_to(trigger, EventType::Pipeline, &self.pipeline_triggers)
+        }
+
+        /// Add trigger with [`TimeEventFilter`]
+        ///
+        /// Returns `false` if a trigger with given id already exists
+        pub fn add_time_trigger(&self, trigger: Trigger<TimeEventFilter>) -> bool {
```
appetrosyan, line 79: `inline`.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+        ///
+        /// Return `false` if a trigger with given id already exists
+        pub fn add_data_trigger(&self, trigger: Trigger<DataEventFilter>) -> bool {
+            self.add_to(trigger, EventType::Data, &self.data_triggers)
+        }
+
+        /// Add trigger with [`PipelineEventFilter`]
+        ///
+        /// Return `false` if a trigger with given id already exists
+        pub fn add_pipeline_trigger(&self, trigger: Trigger<PipelineEventFilter>) -> bool {
```
appetrosyan, line 72: `inline`

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+                        self.by_call_triggers.remove(id).map(|_| ()).expect(
+                            "`Set::by_call_triggers` doesn't contain required id. This is a bug",
+                        )
+                    }
+                })
+                .is_some()
+        }
+
+        /// Check if [`Set`] contains `id`.
+        pub fn contains(&self, id: &TriggerId) -> bool {
```
appetrosyan, line 226: `inline`

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
 //! instructions for triggers in Iroha.
 
-use iroha_data_model::{
-    error::{FindError, MathError},
-    prelude::*,
-};
+use iroha_data_model::{isi::error::MathError, prelude::*, query::error::FindError};
 use iroha_telemetry::metrics;
 
+pub mod set {
```
appetrosyan, line 7: If we restrict this module to `pub(crate)` we can reduce the binary size of `iroha_core`. Given that almost everything that it can depend on is in `core` anyway, restricting visibility will not take much effort.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+        pub fn handle_data_event(&self, event: DataEvent) {
+            for entry in self.data_triggers.iter() {
+                let id = entry.key();
+
+                if id.domain_id.is_some() && id.domain_id.as_ref() != event.domain_id() {
+                    continue;
+                }
+
+                self.match_and_insert_trigger(event.clone(), entry.pair());
+            }
```
appetrosyan, line 274: **Stylistic point (ignore at will)**

Consider rewriting this as a `filter`, where we just ignore the entries with 
```rust 
|entry | entry.key().domain_id.is_some() && entry.key().domain_id.as_ref() != event.domain_id() 
```
and then use a `for_each` to map `match_and_insert_trigger`.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+        /// Handle [`ExecuteTriggerEvent`].
+        ///
+        /// Find all actions that are triggered by `event` and store them.
+        /// These actions are inspected in the next [`Set::inspect_matched()`] call.
+        pub fn handle_execute_trigger_event(&self, event: ExecuteTriggerEvent) {
+            let Some(entry) = self.by_call_triggers.get(&event.trigger_id) else {
+                return
+            };
+
+            self.match_and_insert_trigger(event, entry.pair());
```
appetrosyan, line 286: **Stylistic point (ignore at will)**

This could be rewritten with a `map_or`. An `if let` also conveys the control flow better in this case.

core/src/smartcontracts/isi/triggers.rs
```diff
@@ -1,22 +1,518 @@
...
+        }
+
+        /// Handle [`PipelineEvent`].
+        ///
+        /// Find all actions that are triggered by `event` and store them.
+        /// These actions are inspected in the next [`Set::inspect_matched()`] call.
+        // Passing by value to follow other `handle_` methods interface
+        #[allow(clippy::needless_pass_by_value)]
+        pub fn handle_pipeline_event(&self, event: PipelineEvent) {
+            for entry in self.pipeline_triggers.iter() {
```
appetrosyan, line 296: **Stylistic point (ignore at will)**

Could be rewritten with a single `for_each` or `map`.

core/src/smartcontracts/wasm.rs
```diff
@@ -924,7 +924,8 @@ mod tests {
         let isi_hex = {
             let new_account_id = AccountId::from_str("mad_hatter@wonderland").expect("Valid");
             let register_isi = RegisterBox::new(Account::new(new_account_id, []));
-            encode_hex(Instruction::Register(register_isi))
+
```
appetrosyan, line 927:
```suggestion
```

core/src/smartcontracts/wasm.rs
```diff
@@ -1004,7 +1001,8 @@ mod tests {
...
```
appetrosyan, line 1004:
```suggestion
```

core/src/smartcontracts/wasm.rs
```diff
@@ -1051,7 +1049,8 @@ mod tests {
...
```
appetrosyan, line 1052:
```suggestion
```

tools/parity_scale_decoder/src/generate_map.rs
```diff
@@ -97,7 +97,7 @@ pub fn generate_map() -> DumpDecodedMap {
         BTreeSet<PublicKey>,
         BTreeSet<RoleId>,
         BTreeSet<SignatureOf<transaction::Payload>>,
-        BlockRejectionReason,
+        block::error::BlockRejectionReason,
```
appetrosyan, line 100: I'm not a huge fan of how this function works. I either think that we should use fully qualified names only, or that instead we should add to this list with various scoped `use` statements.

data_model/src/isi.rs
```diff
@@ -94,138 +83,175 @@ impl Instruction {
...
-}
-
-/// Composite instruction for a sequence of instructions.
-#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema, Hash)]
-pub struct SequenceBox {
-    /// Sequence of Iroha Special Instructions to execute.
-    pub instructions: Vec<Instruction>,
+macro_rules! isi {
+    ($($meta:meta)* $item:item) => {
+        crate::model! {
```
appetrosyan, line 88: `Derive(Display)` as well? I think the exception (`SequenceBox`) could just have the macro inlined.

data_model/src/transaction.rs
```diff
@@ -907,214 +803,241 @@ impl From<AcceptedTransaction> for SignedTransaction {
...
-    Hash,
-)]
-#[display(
-    fmt = "Transaction expired: consider increase transaction ttl (current {time_to_live_ms}ms)"
-)]
-pub struct TransactionExpired {
-    /// Transaction ttl.
-    pub time_to_live_ms: u64,
-}
+        impl serde::de::Visitor<'_> for Vis {
```
appetrosyan, line 829:
```suggestion
        impl serde::de::Visitor<'_> for Visitor {
```

data_model/src/transaction.rs
```diff
@@ -129,171 +114,112 @@ pub trait Txn {
...
-            }
-        }
-        deserializer.deserialize_str(Vis)
-    }
+model! {
+    /// Wrapper for byte representation of [`Executable::Wasm`].
+    ///
+    /// Uses **base64** (de-)serialization format.
+    #[derive(DebugCustom, Clone, PartialEq, Eq, Hash, Constructor, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+    #[debug(fmt = "<WASM is truncated>")]
```
appetrosyan, line 168:
```suggestion
impl core::fmt::Debug  for WasmSmartContract {
     fn format(...) -> core::fmt::Result {
         write!(f, "WASM binary, len={}", self.0.len())
     }
}
```

cli/src/torii/pagination.rs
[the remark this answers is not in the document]
mversic: filters as in event filters? they are needed on the client side so should remain in data_model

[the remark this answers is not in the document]
appetrosyan: Clarified in edit.

client/tests/integration/asset.rs
```diff
@@ -160,7 +160,7 @@ fn client_add_asset_with_decimal_should_increase_asset_amount() -> Result<()> {
         )),
     );
     let instructions: Vec<Instruction> = vec![create_asset.into(), mint.into()];
-    let tx = test_client.build_transaction(instructions.into(), metadata)?;
+    let tx = test_client.build_transaction(instructions, metadata)?;
```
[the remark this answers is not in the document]
mversic: this sort of generic API makes sense for client facing code. For internal code I would rather we didn't have generic apis because it might bloat the binary unnecessarily due to monomorphisation. We should resolver types before calling the function for iroha_core whenever possible

client_cli/Cargo.toml
```diff
@@ -35,4 +35,5 @@ erased-serde = "0.3.24"
 
 [build-dependencies]
 eyre = "0.6.8"
-vergen = { version = "7.3.2", default-features = false }
+vergen = { version = "7.4.2", default-features = false }
```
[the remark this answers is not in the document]
mversic: it's already this version in `Cargo.lock` so no change there

core/src/smartcontracts/isi/triggers.rs
mversic: how would changing visibility that reduce the size?

tools/parity_scale_decoder/src/generate_map.rs
mversic: if any is relevant type is missing they are reported because they are taken from schemas. We probably shouldn't even be listing them manually but rather generate from schemas. In any case this is not a problem except being tedious to list them all manually

data_model/src/isi.rs
mversic: I would have done that but `Display` wasn't playing nicely with my macro for some reason which I was not able to determine so I just left it out. I'll try one more time

docs/source/guides/develop-iroha-module.md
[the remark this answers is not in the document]
mversic: ofc

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: I am aware

data_model/src/isi.rs
mversic: this pops up: `cannot find value `_derive_more_display_formatter` in this scope`. I'm leaving it as is

cli/src/torii/pagination.rs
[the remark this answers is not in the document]
mversic: `PredicateBox` is used in `iroha_client`. But I've included it only with `http` feature now

core/src/smartcontracts/isi/triggers.rs
mversic: nvm, done it

appetrosyan (dismissed):

Erigara (approved):

Erigara (approved):

PR #3123, opened 2023-02-08
[refactor] #2943: Derive `HasOrigin`

### LINKED ISSUE #2943: Derive `HasOrigin` trait automatically
### Feature request

Currently `HasOrigin` trait for events relies on manual updating of its variants should any of them change. Instead we could have a derive macro doing this automatically for us as was discussed a long time ago.

### Motivation

Ergonomics, less leeway for error. Also see [comment](https://github.com/hyperledger/iroha/pull/2915#discussion_r1020196199).

### Who can help?

@Erigara @ilchu @s8sato

Erigara:
### Description of the Change

- Introduce `iroha_macro_utils` for macro building utilities reused through other crates;
- Add derive proc macro for `HasOrigin`.

### Issue

Closes #2943.

### Benefits

DRY.

### Possible Drawbacks

None.

### Usage Examples or Tests

Run macro doc tests.

```
cargo test --doc --package iroha_data_model_derive -- has_origin_derive --nocapture 
```

### DIFF
--- a/data_model/derive/src/lib.rs
+++ b/data_model/derive/src/lib.rs
@@ -5,6 +5,7 @@ use proc_macro::TokenStream;
 use syn::parse_macro_input;
 
 mod filter;
+mod has_origin;
 mod id;
 mod partially_tagged;
 
@@ -148,13 +149,11 @@ pub fn id_derive(input: TokenStream) -> TokenStream {
 /// }
 ///
 /// #[derive(Debug, Clone, IdOrdEqHash)]
-/// #[id(type = "LayerId")]
 /// pub struct Layer {
 ///     id: <Self as Identifiable>::Id,
 /// }
 ///
 /// #[derive(Debug, Clone, IdOrdEqHash)]
-/// #[id(type = "SubLayerId")]
 /// pub struct SubLayer {
 ///     id: <Self as Identifiable>::Id,
 /// }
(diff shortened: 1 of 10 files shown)


macro/utils/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+is-it-maintained-open-issues = { repository = "https://github.com/hyperledger/iroha" }
+maintenance = { status = "actively-developed" }
+
+[features]
+
+[dependencies]
+syn = "1.0.107"
+quote = "1.0.23"
+proc-macro2 = "1.0.49"
+proc-macro-error = "1.0.4"
```
ilchu, line 25: A newline missing here too

macro/utils/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+//! Module for various functions and structs to build macros in iroha.
```
ilchu, line 1: Might there be done other functions that we use across crates that could go in here?

macro/utils/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "2.0.0-pre-rc.13"
+authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
+edition = "2021"
+description = "This crate contains various functions and structs for creating macros in Iroha project."
+readme = "README.md"
+homepage = "https://github.com/hyperledger/iroha/tree/iroha2-dev"
+repository = "https://github.com/hyperledger/iroha/tree/iroha2-dev"
+license = "Apache-2.0"
+keywords = ["crypto", "blockchain", "ledger", "iroha", "macro"]
+categories = ["cryptography::cryptocurrencies", "encoding", "development-tools::profiling", "development-tools::debugging"]
```
ilchu, line 12: This looks like a good practice to me, but I believe no other crates have this? Maybe creating a small issue for this cosmetic change across all Iroha would be warranted.

data_model/derive/src/lib.rs
```diff
@@ -341,3 +342,82 @@ pub fn partially_tagged_serialize_derive(input: TokenStream) -> TokenStream {
...
+/// }
+///
+/// #[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
+/// pub struct SubLayerId {
+///     name: u32,
+///     parent_id: LayerId,
+/// }
+///
+/// #[derive(IdOrdEqHash, Debug, Clone)]
+/// #[id(type = "LayerId")]
```
ilchu, line 398: After the `IdOrdEqHash` refactor done by @mversic, these inner attributes are no longer needed, as it is sufficient for a field just to be named `id`. I was surprised the doc test still passed, and upon checking, it seems the macro defaults to looking at the field name currently and silently ignores the unneeded attr. That might be not what we want.

data_model/derive/src/lib.rs
```diff
@@ -341,3 +342,82 @@ pub fn partially_tagged_serialize_derive(input: TokenStream) -> TokenStream {
 pub fn partially_tagged_deserialize_derive(input: TokenStream) -> TokenStream {
     partially_tagged::impl_partially_tagged_deserialize(&parse_macro_input!(input))
 }
+
+/// Derive macro for `HasOrigin`.
+///
+/// Works only with single unnamed field enums.
```
ilchu, line 348:
```suggestion
/// Works only with enums containing single unnamed fields.
```

ilchu (changes requested):
Mostly cosmetic concerns, otherwise looks good

macro/utils/src/lib.rs
Erigara: Probably, my initial idea was that i would place here what i need to reuse. 
And in the future when working with other macros common functionality would emerged it could be extracted here.

macro/utils/Cargo.toml
Erigara: You mean `categories`? If so we already have then in number of crates.

data_model/derive/src/lib.rs
Erigara: Yeah, on container level macros are just ignored.
And this is not only place where we have such kind of behavior (ignoring unused attributes).

ilchu: I feel like that would make more sense, yeah. Because such attributes are attached when we want *certain* behavior, and silently ignoring it should become known to the end user. So be it a change in API or some mistaken usage, it's better to be explicit. If you create an issue, that would be much appreciated.

ilchu (approved):

appetrosyan (approved):

PR #3125, opened 2023-02-08
[fix] #3113: Revise unstable network tests.

SamHSmith:

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -9,6 +9,7 @@
     clippy::std_instead_of_core,
     clippy::std_instead_of_alloc
 )]
+use core::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
@@ -92,6 +93,10 @@ pub struct Iroha {
     thread_handlers: Vec<ThreadHandler>,
     /// Relay that redirects messages from the network subsystem to core subsystems.
     _sumeragi_relay: AlwaysAddr<FromNetworkBaseRelay>, // TODO: figure out if truly unused.
+    /// A boolean value indicating whether or not the peers will recieve data from the network. Used in
+    /// sumeragi testing.
+    #[cfg(debug_assertions)]
+    pub freeze_status: Arc<AtomicBool>,
 }
 
 impl Drop for Iroha {
@@ -104,6 +109,8 @@ impl Drop for Iroha {
 struct FromNetworkBaseRelay {
     sumeragi: Arc<Sumeragi>,
     broker: Broker,
+    #[cfg(debug_assertions)]
+    freeze_status: Arc<AtomicBool>,
 }
 
 #[async_trait::async_trait]
@@ -130,6 +137,11 @@ impl Handler<iroha_core::NetworkMessage> for FromNetworkBaseRelay {
 
     async fn handle(&mut self, msg: iroha_core::NetworkMessage) -> Self::Result {
         use iroha_core::NetworkMessage::*;
+
+        #[cfg(debug_assertions)]
+        if self.freeze_status.load(Ordering::SeqCst) {
+            return;
+        }
 
         match msg {
             SumeragiPacket(data) => {
@@ -357,9 +369,13 @@ impl Iroha {
             ),
         );
 
+        let freeze_status = Arc::new(AtomicBool::new(false));
+
         let sumeragi_relay = FromNetworkBaseRelay {
             sumeragi: Arc::clone(&sumeragi),
             broker: broker.clone(),
+            #[cfg(debug_assertions)]
+            freeze_status: freeze_status.clone(),
         }
         .start()
         .await
@@ -402,6 +418,7 @@ impl Iroha {
             torii,
             thread_handlers: vec![sumeragi_thread_handler, kura_thread_handler],
             _sumeragi_relay: sumeragi_relay,
+            freeze_status,
         })
     }
 
--- a/client/tests/integration/unstable_network.rs
+++ b/client/tests/integration/unstable_network.rs
@@ -1,10 +1,12 @@
 #![allow(clippy::restriction)]
 
-use std::{thread, time::Duration};
+use core::sync::atomic::Ordering;
+use std::thread;
 
 use iroha_client::client::{self, Client};
 use iroha_data_model::prelude::*;
 use iroha_logger::Level;
+use rand::seq::SliceRandom;
 use test_network::*;
 use tokio::runtime::Runtime;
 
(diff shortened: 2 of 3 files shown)


cli/src/lib.rs
```diff
@@ -93,6 +94,9 @@ pub struct Iroha {
     thread_handlers: Vec<ThreadHandler>,
     /// Relay that redirects messages from the network subsystem to core subsystems.
     _sumeragi_relay: AlwaysAddr<FromNetworkBaseRelay>, // TODO: figure out if truly unused.
+    /// A boolean value used to stop the peers ability to recieve data from the network. Used in
+    /// sumeragi testing.
+    pub freeze_status: Arc<AtomicU32>,
```
Erigara, line 99: I'm not fan of tendency to include testing specific variables into the primary modules. 
Maybe we can add some kind of wrapper that will drop messages?

mversic: @Erigara I agree. It's a sign of a bad design. Sometimes it can't be helped

cli/src/lib.rs
```diff
@@ -132,6 +137,10 @@ impl Handler<iroha_core::NetworkMessage> for FromNetworkBaseRelay {
     async fn handle(&mut self, msg: iroha_core::NetworkMessage) -> Self::Result {
         use iroha_core::NetworkMessage::*;
 
+        if self.freeze_status.load(Ordering::Relaxed) > 0 {
```
mversic, line 140: I would just use `SC`. If you think this is on a hot path then I will give it a deep thought. Ordering is hard to get right

client/tests/integration/unstable_network.rs
```diff
@@ -16,63 +18,33 @@ const MAXIMUM_TRANSACTIONS_IN_BLOCK: u32 = 5;
 fn unstable_network_4_peers_1_fault() {
     let n_peers = 4;
     let n_transactions = 20;
-    // Given that the topology will resolve view changes by shift for `n_peers` consequent times
-    // and having only 1 faulty peer guarantees us that at maximum in `n_peers` tries the block will be committed.
-    // So for the worst case scenario for `n_transactions` given 1 transaction per block
-    // the network will commit all of them in `n_peers * n_transactions` tries.
```
ilchu, line 22: Correct me if I'm wrong, but it feels that the condition described here hasn't changed, so I don't see a reason to remove the comment. Additional clarifications never hurt.

cli/src/lib.rs
ilchu: I agree that relaxed ordering could be a bit too much, but this is a classic use case for `Acquire`/`Release` IMO.

client/tests/integration/unstable_network.rs
```diff
@@ -126,22 +114,25 @@ fn unstable_network(
...
-                        && *asset.value() == AssetValue::Quantity(account_has_quantity)
-                })
-            },
-        )
-        .expect("Test case failure.");
+        // Return all peers to normal function.
+        for f in &freezers {
+            f.store(0, Ordering::Relaxed);
+        }
+    }
```
ilchu, line 137: So the solution to test flakiness is that instead of having preset offline peers, we now choose them randomly for each transaction and reset at the end of each? Also just to be on the safe side, maybe it's better to: 
```suggestion
        // Return all peers to normal function.
        for f in &freezers {
            f.compare_exchange(1, 0, Ordering::Acquire, Ordering::Relaxed);
        }
    }
```

ilchu (changes requested):

cli/src/lib.rs
mversic: I wouldn't be so reluctant to approve even `Acq/Rel`. `Relaxed` is just dangerous if there is no strong reason to use it

cli/src/lib.rs
mversic: not if it has to be cross-crate. Feature flag of compile flag would help

cli/src/lib.rs
SamHSmith: Relaxed is fine in this case since the atomic is not used to guard any seperate non-atomic data. (https://www.youtube.com/watch?v=ZQFzMfHIxng&t=3824s) Since we only have one reading thread and a seperate writing thread in C you wouldn't need atomics for this. I'm simply poking a value and want the other thread to eventually see the new value. Rust requires atomics for this because of it's safety rules.

Since this isn't performance critical in any way I will change it to SeqCst in order to ease the merging of the code.

cli/src/lib.rs
SamHSmith: I will put hide it under debug-assertions such that it does not get compiled in when we build for release. I do think however that we should clearly divide what is "public" and what is not. It is not appropriate to apply the same standards of public interfaces that are a part of our API that goes to customers to application internal interfaces. The internal interfaces should only be governed by principles that aid speed of development and producing the best product.

client/tests/integration/unstable_network.rs
```diff
@@ -126,22 +114,25 @@ fn unstable_network(
             .expect("Failed to create asset.");
         account_has_quantity += quantity;
         thread::sleep(pipeline_time);
-    }
 
-    thread::sleep(pipeline_time);
+        //Then
```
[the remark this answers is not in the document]
SamHSmith: Sure, I'll remove it.

core/test_network/src/lib.rs
```diff
@@ -171,6 +171,14 @@ impl Network {
             .collect()
     }
 
+    /// Collect the freeze handles from all the peers in the network.
+    pub fn get_freeze_status_handles(&self) -> Vec<Arc<AtomicU32>> {
```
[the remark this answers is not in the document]
SamHSmith: I will change it.

client/tests/integration/unstable_network.rs
SamHSmith: Not in this PR. But sumeragi changes have already been merged that make this comment out of date.

client/tests/integration/unstable_network.rs
SamHSmith: The acquire would be meaningless because there is not any data or memory access to synchronize. This PR does nothing for any test flakiness. What it does is prepare the tests for us moving closer the whitepaper. It makes them more proper tests of sumeragi that the future changes won't be able to dodge.

cli/src/lib.rs
```diff
@@ -92,6 +93,10 @@ pub struct Iroha {
     thread_handlers: Vec<ThreadHandler>,
     /// Relay that redirects messages from the network subsystem to core subsystems.
     _sumeragi_relay: AlwaysAddr<FromNetworkBaseRelay>, // TODO: figure out if truly unused.
+    /// A boolean value used to stop the peers ability to recieve data from the network. Used in
+    /// sumeragi testing.
```
outoftardis, line 97:
```suggestion
    /// A boolean value indicating whether or not peers can receive data from the network. Used in
    /// sumeragi testing.
```

ilchu (approved):

Erigara (dismissed):

cli/src/lib.rs
appetrosyan: Actually I doubt we can do better than `debug_assertions` recall that we will need to test the debug build using @astrokov7 's tests in Python.

cli/src/lib.rs
appetrosyan: My two cents: 
- @mversic and @ilchu , you're right, but a tad overzealous. IMO this can work with `Relaxed` just as well. We only need interior mutability. In C this would be a volatile variable. 
- @SamHSmith You're right, but a tad careless. If the ordering is not `SeqCst` and the reason isn't obvious from the same file, your justification should have been a comment in the code. 
- Not doing `SeqCst` is IMO a premature optimisation. Lock-free programming is a place where you *should* be careful.

Erigara (approved):

ilchu (dismissed):

Erigara (approved):

ilchu (approved):

PR #3106, opened 2023-02-01
[feature] #1618: Replace builtin validators with runtime validators

Arjentix:
### Description of the Change

Builtin validators replaced by runtime validators.

See commits for more details

### Issues

- Closes main issue: #1618
- Creates Associated issues :
  - #3169
  - #3170
  - #3171
  - #3172 
  - #3173 
  - #3174 
  - #3175
  - #3176 
  - #3177 
  - #3178 
  - #3179 
  - #3180
  - #3181 
  - #3182
  - #3184 
  - #3185 
  - #3186 
  - #3187 
  - #3188 
  - #3189 
  - #3190
  - #3191 
  - #3192 
  - #3193 
  - #3194 
  - #3196 
  - #3197 
  - #3198 
  - #3199 
  - #3200
  - #3201 
  - #3202 
  - #3203 
  - #3204 
  - #3205 
  - #3244 
  - #3206 
  - #3207 
  - #3209 
  - #3210
  - #3211
  - #3212 
  - #3213 
  - #3214 
  - #3215 
  - #3216 
  - #3217 
  - #3218 
  - #3219 
  - #3220
  - #3221 

# Followup tasks

See #3223 Epic on Zenhub

Issue #1618, opened 2021-11-17
Move permission validators into WASM Plugin


e-ivkov:
See [Triggers RFC](https://wiki.hyperledger.org/display/iroha/Triggers)

Also add corresponding instructions for their update at runtime.

appetrosyan:
@Arjentix please convert into epic and track changes

PR #3138, opened 2023-02-10
[refactor]: Remove `iroha_client` dependency on `iroha_core`

Erigara:


mversic (commented):
looks good. Is there something you find problematic?

Erigara:
> Is there something you find problematic?

Currently it's hart to remove `iroha_core` from `iroha_client` **dev-dependencies** because of `test_network`.

Another concern is that i have changed API of block streaming, because now streaming returns `BlockValue` (as in queries) instead of `VersionedCommittedBlock`.

mversic:
> Currently it's hart to remove iroha_core from iroha_client dev-dependencies because of test_network.

`iroha_core`as a `dev-dependency` is tolerable

> Another concern is that i have changed API of block streaming, because now streaming returns `BlockValue` (as in queries) instead of `VersionedCommittedBlock`.

This is what `BlockValue` was created for. It was a stop gap solution. A proper solution would be to move `VersionedCommittedBlock` to `data_model`. But that would be a big change

appetrosyan:
A few observations: 

The client tests are mostly for sanity checking. I don't really think that running them the way we do (via `test_network`) is a good idea. Here @SamHSmith 's decision to leave in the `debug_force_soft_fork` configuration parameter can be beneficial. Then we can bring up a docker compose config and re-use it. Ideally though, we shouldn't need these tests in particular and instead rely on the shell tests made by @astrokov7 . In any respect `client` tests are not something which should influence our architecture. 

The main reason why we chose creating an almost duplicate `BlockValue` with @QuentinI , is because of a quirk of Rust. Rust has no concept of workspace visibility (impossible without  restructuring). We don't want intermediate states of block processing to be private because they need to be visible to Sumeragi. We don't want them public  because that's a security risk. If we're fine with creating an extra crate this can be solved by creating an `iroha_block` crate, linking it directly to `iroha_core` and reexporting only the safe types from `iroha_data_model` can solve that issue, but it needs to be solved, if we move `*Block` outside `iroha_core`.

Erigara:
> The main reason why we chose creating an almost duplicate BlockValue with @QuentinI , is because of a quirk of Rust. Rust has no concept of workspace visibility (impossible without restructuring). We don't want intermediate states of block processing to be private because they need to be visible to Sumeragi. We don't want them public because that's a security risk. If we're fine with creating an extra crate this can be solved by creating an iroha_block crate, linking it directly to iroha_core and reexporting only the safe types from iroha_data_model can solve that issue, but it needs to be solved, if we move *Block outside iroha_core.

If i understand correctly work done by @mversic with `transaprent_api` we can include block structures into `iroha_data_model` because we can control visibility.

mversic:
> If i understand correctly work done by @mversic with transaprent_api we can include block structures into iroha_data_model because we can control visibility.

exactly, if you look at my PR #3127, for instance, `Register` instruction is public for `iroha_core` but will not be seen by `iroha_client`. `RegisterBox` is public for both crates

client/Cargo.toml
```diff
@@ -43,6 +42,7 @@ futures-util = "0.3.25"
 
 
 [dev-dependencies]
+iroha_core = { version = "=2.0.0-pre-rc.13", path = "../core", features = ["dev-telemetry", "telemetry"]}
```
mversic, line 45: why does it still remain dev dependency?

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+#![cfg(feature = "std")]
+
+use derive_more::Display;
+use iroha_crypto::HashOf;
+use iroha_primitives::fixed::FixedPointOperationError;
+use iroha_schema::IntoSchema;
+use parity_scale_codec::{Decode, Encode};
+use thiserror::Error;
+
+use crate::{block::VersionedCommittedBlock, metadata, permission, prelude::*, trigger};
```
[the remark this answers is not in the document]
Erigara: Sure, because client depends on this error structures. `QueryResponseHandler` expects to recieve `QueryError` ([link](https://github.com/hyperledger/iroha/blob/d4f3d85a189ccbe40081188d7f03f67a98282a18/client/src/client.rs#L85)).

client/Cargo.toml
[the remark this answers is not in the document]
Erigara: > why does it still remain dev dependency?

Because there is still number of things which are depends on core: 
1) `test_network` crate
2) `iroha_core::smartcontracts::wasm` module
3) permissions related stuff (which should be removed by #3106)
4) `iroha_core::genesis` module (we are building custom genesis in `million_accounts_genesis` example)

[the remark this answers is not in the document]
mversic: I think `2.` you can resolve in this PR and `3.` will be taken care of.
`1.` and `4.` are both about genesis if I'm not mistaken.

So it's only about genesis? We can fix this I believe, not necessarily in this PR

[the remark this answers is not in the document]
Erigara: Ok, will return to draft than and see what can be done.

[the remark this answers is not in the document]
mversic: arguably, we should have a predefined genesis file for those tests?

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[error("Repetition")]
+    Repetition(InstructionType, IdBox),
+    /// Failed to validate.
+    #[error("Failed to validate: {0}")]
+    Validate(#[from] ValidationError),
+}
+
+/// Enumeration of instructions which can have unsupported variants.
+#[derive(Debug, Display, Clone, Copy)]
+pub enum InstructionType {
```
mversic, line 84: for `Value` we use `strum_discriminants`. Can we apply the same here?

client/Cargo.toml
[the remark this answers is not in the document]
Erigara: I think we can and afaik @Arjentix done smt like this in #1618, made integration test use default `genesis.json`.

data_model/src/error.rs
Erigara: will try

client/Cargo.toml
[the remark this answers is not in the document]
mversic: either that or maybe extract that `GenesisNetwork` and stuff to a separate crate?

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[error("Query found wrong type of asset: {0}")]
+    Conversion(String),
+    /// Query without account.
+    #[error("Unauthorized query: account not provided")]
+    Unauthorized,
+}
+
+/// Instruction execution error type
+#[derive(Debug, Error)]
+pub enum IsiError {
```
mversic, line 46: this one we don't use in `client`?

mversic, line 46: this one we don't use in `client`?

client/Cargo.toml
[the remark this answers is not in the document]
Erigara: maybe this will be better option because store million account genesis is not really convenient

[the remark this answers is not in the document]
mversic: I see what you mean, you could produce it programatically and just paste it there, but it might be very large

data_model/src/error.rs
Erigara: Yes, but it uses `FindError` which is also used by `QueryError`,  so i thought it would make sense to group them.

client/Cargo.toml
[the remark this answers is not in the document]
mversic: it might not be bad that you handle this, but I think `transparent_api` will have to be used in integration tests. For instance, `Registrable::build` is only visible when `transparent_api` is active and integration tests rely heavily on being able to construct those. Another option is to make `Registrable` visible always, but that trait is useless for the client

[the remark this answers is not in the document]
appetrosyan: IMO this isn't exactly improving the architecture. We want to avoid bidirectional dependency if possible, such that `iroha_client` is slimmed down as far as we can make it go (and then some).

- Genesis is a special case. It's also mostly dead code. Extract into separate translation unit. IDK if there's a way to hint the linker that the code is single-shot, probably automatic. 
- There are circumstances where a separate process is a good choice. Example: `iroha_bootstrap` process that contains config parsing, genesis, bootstrapping and port mapping, which is a large executable with good UI, which hands over to a minimal `irohad` process, that only handles post-genesis  consensus, and responds to API requests. This is not going to save much memory, but it will allow control flow to follow the information flow. 
- Client tests must die. Real integration tests should work with separate processes built in `release` mode, wherever possible. 
- If `Registrable` isn't used, then it shouldn't be visible. However, if this entails adding another feature flag, I'd rather it were visible and unused, than bloat our feature matrix. AFAICT, there may be a way to make it work with `transparent_api`.

data_model/src/error.rs
appetrosyan: Actually, we also have a few other crates that do the same. 

https://docs.rs/enum-kinds/latest/enum_kinds/

Can we consolidate?

data_model/src/error.rs
appetrosyan:
```suggestion
pub enum InstructionExecutionFailure {
```

Similarly, `QueryError` -> `QueryExecutionFailure`. 

These names are much more informative and specific. We *know* that the `enum` is an `Error` because it has `#[derive(Error)` on top of it. I don't recommend hiding more things behind feature flags if the purpose is dead code elimination, given that LTO is a thing.

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+    Permission(permission::validator::DenialReason),
+    /// Query has wrong expression.
+    #[error("Query has a malformed expression: {0}")]
+    Evaluate(String),
+    /// Query found nothing.
+    #[error("Query found nothing: {0}")]
+    Find(#[from] Box<FindError>),
+    /// Query found wrong type of asset.
+    #[error("Query found wrong type of asset: {0}")]
+    Conversion(String),
```
appetrosyan, line 38: Same here.

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+    Revoke,
+    /// Set a chain-wide config parameter
+    SetParameter,
+    /// New chain-wide config parameter
+    NewParameter,
+}
+
+/// Type assertion error
+#[derive(Debug, Error, Decode, Encode, IntoSchema)]
+pub enum FindError {
```
appetrosyan, line 111: Given that this is no longer separated from the data by a crate boundary, I fully expect this struct is better generated by a macro.

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[error("Value passed to the parameter doesn't have the right type. {0}")]
+    ParameterValueType(#[from] Mismatch<Value>),
+    /// Asset Id mismatch
+    #[error("AssetDefinition Ids don't match. {0}")]
+    AssetDefinitionId(#[from] Box<Mismatch<<AssetDefinition as Identifiable>::Id>>),
+}
+
+/// Math error, which occurs during instruction execution
+#[derive(Debug, Clone, Error, Copy, PartialEq, Eq)]
+pub enum MathError {
```
appetrosyan, line 183:
```suggestion
pub enum AssetMathError {
```

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+    DomainViolation,
+    /// Unknown error. No actual function should ever return this if possible.
+    #[error("Unknown error")]
+    Unknown,
+    /// Encountered incompatible type of arguments
+    #[error("Binary operation does not support provided combination of arguments ({0}, {1})")]
+    BinaryOpIncompatibleNumericValueTypes(NumericValue, NumericValue),
+}
+
+impl From<FixedPointOperationError> for IsiError {
```
appetrosyan, line 207: When I get back, these should be simplified for the in-house implementation.

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+            FixedPointOperationError::Arithmetic => MathError::Unknown.into(),
+        }
+    }
+}
+
+impl From<FindError> for IsiError {
+    fn from(err: FindError) -> Self {
+        Self::Find(Box::new(err))
+    }
+}
```
appetrosyan, line 226: We could consider replacing `thiserror` with `derive_more`, given that this pattern would be covered with `derive_more::From`, and we can derive `Display` independently from `error`.

data_model/src/error.rs
mversic: > Can we consolidate?

@appetrosyan we've used to use `enum_kinds` in `iroha_data_model`. Idk why we switched to `strum`

data_model/src/transaction.rs
```diff
@@ -756,6 +756,110 @@ impl From<VersionedRejectedTransaction> for VersionedSignedTransaction {
...
+impl From<ValidTransaction> for AcceptedTransaction {
+    fn from(transaction: ValidTransaction) -> Self {
+        AcceptedTransaction {
+            payload: transaction.payload,
+            signatures: transaction.signatures,
+        }
+    }
+}
+
+impl From<VersionedRejectedTransaction> for VersionedAcceptedTransaction {
```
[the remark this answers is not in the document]
mversic: good point

config/src/sumeragi.rs
```diff
@@ -16,6 +16,11 @@ const DEFAULT_ACTOR_CHANNEL_CAPACITY: u32 = 100;
 const DEFAULT_GOSSIP_PERIOD_MS: u64 = 1000;
 const DEFAULT_GOSSIP_BATCH_SIZE: u32 = 500;
 
+/// Default estimation of consensus duration
+#[allow(clippy::integer_division)]
+pub const DEFAULT_CONSENSUS_ESTIMATION_MS: u64 =
+    DEFAULT_BLOCK_TIME_MS + (DEFAULT_COMMIT_TIME_LIMIT_MS / 2);
```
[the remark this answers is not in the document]
mversic: and shouldn't require `allow` lint

client/Cargo.toml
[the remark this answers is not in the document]
mversic: > There are circumstances where a separate process is a good choice. Example: iroha_bootstrap process that contains config parsing, genesis, bootstrapping and port mapping, which is a large executable with good UI, which hands over to a minimal irohad process, that only handles post-genesis consensus, and responds to API requests. This is not going to save much memory, but it will allow control flow to follow the information flow.

this is an interesting idea

> Client tests must die. Real integration tests should work with separate processes built in release mode, wherever possible.

absolutely

> If Registrable isn't used, then it shouldn't be visible. However, if this entails adding another feature flag, I'd rather it were visible and unused, than bloat our feature matrix. AFAICT, there may be a way to make it work with transparent_api.

no additional feature flags. `Registrable` is not visible unless `transparent_api` is activated

data_model/src/error.rs
appetrosyan: > Idk why we switched to strum.

`enum_kinds` is unmaintained. `strum` offers more functionality and is in many ways better suite to what we do.

cli/src/torii/routing.rs
```diff
@@ -312,8 +314,9 @@ async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::R
             _ = interval.tick() => {
                 if let Some(block) = kura.get_block_by_height(from_height) {
                     stream
+                        // TODO: to avoid clone `VersionedBlockMessage` could be split into sending and receiving parts
```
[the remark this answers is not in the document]
Erigara: I want to split `VersionedBlockMessage` not stream. 
I see it like this: 

```rust
struct BlockMessageSend(pub Arc<VersionedCommittedBlock>);

struct BlockMessageRecv(pub VersionedCommittedBlock);
```

This way there is no need in cloning block when we want to send it.

data_model/src/error.rs
Erigara: We can do not only asset related operations inside iroha. So i think `MathError` is more accurate in this case.

data_model/src/block.rs
```diff
@@ -0,0 +1,662 @@
...
+
+impl BlockHeader {
+    /// Checks if it's a header of a genesis block.
+    #[inline]
+    pub const fn is_genesis(&self) -> bool {
+        self.height == 1
+    }
+}
+
+impl PartialOrd for BlockHeader {
```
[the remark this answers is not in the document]
Erigara: I have done such reorganization, however i preferred to leave block names as is, because this way it is more consistent with change your proposed for `schema.json` about having unambiguous names.

docs/source/references/schema.json
```diff
@@ -71,18 +71,18 @@
   "Option<Duration>": {
     "Option": "Duration"
   },
-  "Option<iroha_core::sumeragi::network_topology::Topology>": {
-    "Option": "iroha_core::sumeragi::network_topology::Topology"
+  "Option<Vec<iroha_data_model::peer::Id>>": {
+    "Option": "Vec<iroha_data_model::peer::Id>"
```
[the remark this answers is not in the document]
Erigara: @appetrosyan you want this change in the current PR or it's ok to do this change in another PR?

data_model/src/error.rs
Erigara: Created #3266

data_model/src/error.rs
Erigara: Created #3267.

data_model/src/error.rs
Erigara: Not sure how to implement this as macro because `FindError` contain data from different sources.
Seems like more effort than benefits.

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: let's take this separately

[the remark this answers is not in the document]
Erigara: Created issue #3268 for discussion on unambiguous names.

data_model/src/error.rs
```diff
@@ -0,0 +1,243 @@
...
+
+/// Query errors.
+#[derive(Error, Debug, Decode, Encode, IntoSchema)]
+pub enum QueryError {
+    /// Query cannot be decoded.
+    #[error("Query cannot be decoded")]
+    Decode(#[from] Box<iroha_version::error::Error>),
+    /// Query has wrong signature.
+    #[error("Query has the wrong signature: {0}")]
+    Signature(String),
```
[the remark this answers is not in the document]
Erigara: Created #3269.

Erigara:
Further progresses is possible after merge of #3106 where permission related stuff is removed from the core and tests.

tools/parity_scale_decoder/src/generate_map.rs
```diff
@@ -469,9 +469,9 @@ mod tests {
     fn schemas_types_is_a_subset_of_map_types() {
         // These types **shouldn't** implement `Decode`. As such we need to make an exception.
         let exceptions = HashSet::from(type_names_arr![
-            Vec<iroha_core::genesis::GenesisTransaction>,
-            iroha_core::genesis::GenesisTransaction,
-            iroha_core::genesis::RawGenesisBlock,
+            Vec<iroha_genesis::GenesisTransaction>,
+            iroha_genesis::GenesisTransaction,
+            iroha_genesis::RawGenesisBlock,
```
mversic, line 474: is genesis required as a dependency of kagami?

mversic (changes requested):

client/Cargo.toml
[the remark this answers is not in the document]
mversic: ok. So this will be taken care of when we remove integration tests from client?

data_model/Cargo.toml
```diff
@@ -55,6 +55,7 @@ base64 = { version = "0.13.1", default-features = false, features = ["alloc"]}
 [dev-dependencies]
 iroha_core = { path = "../core", version = "=2.0.0-pre-rc.13" }
 iroha_client = { path = "../client", version = "=2.0.0-pre-rc.13" }
+iroha_genesis = { path = "../genesis", version = "=2.0.0-pre-rc.13" }
 iroha = { path = "../cli" }
 
 test_network = { path = "../core/test_network", version = "=2.0.0-pre-rc.13" }
```
mversic, line 61: can we remove this dependency? Maybe these tests should actually reside in the `client/tests/integration`?

mversic: I think only `find_rate_and_make_exchange_isi_should_succeed` should be moved to `iroha_client`

core/Cargo.toml
```diff
@@ -46,6 +46,7 @@ iroha_futures = { version = "=2.0.0-pre-rc.13", path = "../futures" }
 iroha_telemetry = { version = "=2.0.0-pre-rc.13", path = "../telemetry" }
 iroha_schema = { version = "=2.0.0-pre-rc.13", path = "../schema" }
```
mversic, line 47: I think `iroha_schema` dependency should be removed

client/Cargo.toml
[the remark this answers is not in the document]
Erigara: yes, we will depend on `core` until we stop starting iroha peers inside integration tests .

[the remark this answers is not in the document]
mversic: maybe you can write a `TODO` in the `client/Cargo.toml` and close this issue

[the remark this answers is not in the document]
mversic: excuse me, I meant you can resolve this conversation.

tools/parity_scale_decoder/src/generate_map.rs
Erigara: yes, we need it in tests

schema/gen/src/lib.rs
```diff
@@ -7,11 +7,8 @@
     clippy::std_instead_of_alloc
 )]
 
-use iroha_core::{
-    block::{stream::prelude::*, VersionedCandidateBlock},
-    genesis::RawGenesisBlock,
-    smartcontracts::isi::query::Error as QueryError,
-};
+use iroha_data_model::{block::stream::prelude::*, error::QueryExecutionFailure};
+use iroha_genesis::RawGenesisBlock;
```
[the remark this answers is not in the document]
Erigara: Ok, let's keep it in the schema for now, created #3284 for further investigation.

data_model/src/block.rs
```diff
@@ -0,0 +1,730 @@
...
+        pub view_change_index: u64,
+        /// Hash of a previous block in the chain.
+        /// Is an array of zeros for the first block.
+        pub previous_block_hash: Option<HashOf<VersionedCommittedBlock>>,
+        /// Hash of merkle tree root of the tree of valid transactions' hashes.
+        pub transactions_hash: Option<HashOf<MerkleTree<VersionedSignedTransaction>>>,
+        /// Hash of merkle tree root of the tree of rejected transactions' hashes.
+        pub rejected_transactions_hash: Option<HashOf<MerkleTree<VersionedSignedTransaction>>>,
+        /// Genesis topology
+        pub genesis_topology: Option<Vec<peer::Id>>,
```
mversic, line 70: oh, so you opted for not moving `Topology` to `data_model`

core/src/sumeragi/network_topology.rs
```diff
@@ -190,10 +191,10 @@ impl Builder {
 }
 
 /// The ordering of the peers which defines their roles in the current round of consensus.
-#[derive(Clone, Debug, Encode, Decode, IntoSchema, Serialize)]
+#[derive(Clone, Debug, Encode, Decode, Serialize)]
```
mversic, line 194: can you tell me if `Serialize` really required here?

Erigara: i will check

data_model/src/block.rs
mversic: I'm ok with how you did it

client/Cargo.toml
```diff
@@ -45,7 +45,9 @@ futures-util = "0.3.25"
 [dev-dependencies]
 iroha = { path = "../cli", features = ["dev-telemetry", "telemetry"] }
 iroha_wasm_builder = { version = "=2.0.0-pre-rc.13", path = "../wasm_builder" }
+iroha_genesis = { version = "=2.0.0-pre-rc.13", path = "../genesis" }
```
mversic, line 48: this should be pulled from `dependencies` section automatically?

tools/parity_scale_decoder/Cargo.toml
```diff
@@ -18,6 +18,7 @@ iroha_schema = { version = "=2.0.0-pre-rc.13", path = "../../schema"}
 iroha_core = { version = "=2.0.0-pre-rc.13", path = "../../core"}
```
mversic, line 18: is `core` required here?

data_model/src/transaction.rs
```diff
@@ -781,6 +781,132 @@ impl From<VersionedRejectedTransaction> for VersionedSignedTransaction {
...
+    #[inline]
+    fn payload(&self) -> &Payload {
+        &self.as_v1().payload
+    }
+}
+
+/// `AcceptedTransaction` — a transaction accepted by iroha peer.
+#[version_with_scale(n = 1, versioned = "VersionedAcceptedTransaction")]
+#[derive(Debug, Clone, Decode, Encode, Serialize)]
+pub struct AcceptedTransaction {
```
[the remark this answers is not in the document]
mversic: I would like to ask the same for `ValidTransaction`

data_model/src/block.rs
```diff
@@ -0,0 +1,730 @@
...
+pub use self::{
+    candidate::{CandidateBlock, VersionedCandidateBlock},
+    candidate_committed::{CandidateCommittedBlock, VersionedCandidateCommittedBlock},
+    chained::ChainedBlock,
+    committed::{CommittedBlock, VersionedCommittedBlock},
+    header::BlockHeader,
+    pending::{PendingBlock, VersionedPendingBlock},
+    signed::SignedBlock,
+    valid::ValidBlock,
+};
```
mversic, line 27: IMO the only block that we need in `data_model` is `CommittedBlock`

mversic: we don't want internal state transitions polluting our public API

client/Cargo.toml
Erigara: what do you mean?

mversic: it's duplicated, you can remove it

data_model/src/transaction.rs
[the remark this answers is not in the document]
Erigara: `AcceptedTransaction` is also used in `iroha_genesis`.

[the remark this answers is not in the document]
mversic: > `AcceptedTransaction` is also used by `iroha_genesis`.

I'll have to remember to exclude it with `transparent_api`

data_model/src/block.rs
Erigara: i agree

data_model/src/transaction.rs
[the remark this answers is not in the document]
Erigara: Maybe put `TODO`?

data_model/src/error.rs
mversic: @Erigara I believe this type could replace `InstructionExecutionFail::reason` type

cli/src/torii/routing.rs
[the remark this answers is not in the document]
appetrosyan: Separate issue, and explain clearly.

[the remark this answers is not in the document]
Erigara: Created #3297

data_model/src/error.rs
Erigara: Created #3298

appetrosyan (approved):

PR #3146, opened 2023-02-13
Fix in iroha-cpp library + added example of few transactions

baziorek:
Signed-off-by: Grzegorz Bazior <g.bazior@yodiss.pl>

### Description of the Change
I noticed that in examples queries were not working - after some time of investigation I noticed that `counter` must be greated than 0, but default  value is 0. So I changed into 1. Also I added another example file with error checking (it was my first time with GRPC in C++ so there can be errors)

### Benefits
Good example for somebody who wants to start developing Iroha 1 with iroha-cpp-lib + fix

### Possible Drawbacks
I don't know what `counter` is:D.

```
make irohalib_domainassetcreation_example
```

### DIFF
--- a/iroha-lib/examples/BatchExample.cpp
+++ b/iroha-lib/examples/BatchExample.cpp
@@ -5,7 +5,7 @@
 #include "model/TxBatch.hpp"
 
 
-iroha::protocol::Transaction generateTransaction(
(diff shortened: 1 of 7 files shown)


baziorek:
I was told that PR must be done from HL's reository, not from fork. So I copied the PR:
https://github.com/hyperledger/iroha/pull/3148
and it can be closed now

Issue #3155, opened 2023-02-15
Panic in one test leads to other tests failure


Arjentix:
While I was working on #3106 , I found a very strage tests behaviour. After a day of debugging it turns out to be our panic hook implementation!

So currently we register a new hook in `Iroha` start-up. Our hook first calls for previously set hook. The problem is that when we run tests in **parallel** they will register a lot of hooks. And hooks are not *test-local*, they are *process-local*.
So when one tests panics it calls for a general panic-hook (which will be set to the hook, registered by the last runned test), which not only sends shutdown message to a wrong `Iroha` instance of another test, but also call all previously registered hooks, so it will disable all other `Iroha` instances as well.

Also one thing made debugging much harder. I could not see logs comming from panic-hook, becuase I was inspecting one test output, but in fact it was another test which actually called for this panic-hook and so the log output belonged to that test. Mind blowing.

We need to fix this.

PR #3156, opened 2023-02-15
[fix] #3155: Fix panic-hook for tests

Arjentix:
### Description of the Change

See #3155 for bug details.

To fix it I used [thread-local-panic-hook](https://crates.io/crates/thread-local-panic-hook) crate. It's not popular at all, but it solves our problem.

So when `test-network` feature is enabled I set thread local panic hook instead of basic panic hook (process local).

Also I removed useless invocation of previously set hook.

### Issue

- #3155 

### Benefits

Panic in one test won't cause Iroha shutdowns in another tests.

### Drawbacks

When `test-network` feature is activated all sub-threads (e.g. Sumeragi thread) won't have specified panic-hook. Instead they will have a default one.

I and @appetrosyan decided that this this acceptable because of this:

- Sumeragi and etc. are designed to not panic because of client actions
- It works normally right now
- There will be no new integration tests in future. They will be rewritten in Python
- After rewriting tests this change will be rollbacked

### Usage Examples or Tests

Don't know how to prove that it fixes the problem. But it works for me in #3106 .

### Alternate Designs

There is an idea how to fix this problem with sub-threads but it doesn't worth implementation

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -9,7 +9,7 @@
     clippy::std_instead_of_core,
     clippy::std_instead_of_alloc
 )]
-use std::{panic, sync::Arc};
+use std::sync::Arc;
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
 use eyre::ContextCompat as _;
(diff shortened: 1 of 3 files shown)


cli/src/lib.rs
```diff
@@ -464,8 +484,12 @@ impl Iroha {
 
         let handle = task::spawn(async move {
             tokio::select! {
-                _ = sigint.recv() => {},
-                _ = sigterm.recv() => {},
+                _ = sigint.recv() => {
+                    iroha_logger::warn!("SIGINT received, shutting down...");
```
appetrosyan, line 488: info

cli/src/lib.rs
```diff
@@ -464,8 +484,12 @@ impl Iroha {
...
+                },
+                _ = sigterm.recv() => {
+                    iroha_logger::warn!("SIGTERM received, shutting down...");
```
appetrosyan, line 491: info

appetrosyan (approved):

Erigara (approved):

Issue #3153, opened 2023-02-14
[Testability `iroha_client_cli`] All data responses in `stdout` should be in `JSON`.


astrokov7:
### Feature request

To test `iroha_client_cli` conveniently (avoid parsing), we need all its responses to be formatted using `JSON`.
To see current responses run the following commands:
```bash
./iroha_client_cli domain list all
```
Or
```bash
./iroha_client_cli account list all
```
Or
```bash
./iroha_client_cli asset list all
```

### Actual `stdout`
<details>
  <summary>Printed Rust debug in stdout</summary>

```
User: alice@wonderland
[
    Domain {
        id: Id {
            name: "F",
        },
        accounts: {},
        asset_definitions: {},
        logo: None,
        metadata: Metadata {
            map: {},
        },
    },
    Domain {
        id: Id {
            name: "garden_of_live_flowers",
        },
        accounts: {
            Id {
                name: "carpenter",
                domain_id: Id {
                    name: "garden_of_live_flowers",
                },
            }: Account {
                id: Id {
                    name: "carpenter",
                    domain_id: Id {
                        name: "garden_of_live_flowers",
                    },
                },
                assets: {},
                signatories: {
                    { digest: ed25519, payload: 7233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0 },
                },
                signature_check_condition: SignatureCheckCondition(
                    EvaluatesTo {
                        expression: ContainsAny(
                            ContainsAny {
                                collection: EvaluatesTo {
                                    expression: ContextValue(
                                        ContextValue {
                                            value_name: "transaction_signatories",
                                        },
                                    ),
                                    _value_type: PhantomData,
                                },
                                elements: EvaluatesTo {
                                    expression: ContextValue(
                                        ContextValue {
                                            value_name: "account_signatories",
                                        },
                                    ),
                                    _value_type: PhantomData,
                                },
                            },
                        ),
                        _value_type: PhantomData,
                    },
                ),
                metadata: Metadata {
                    map: {},
                },
                roles: {},
            },
        },
        asset_definitions: {
            DefinitionId {
                name: "cabbage",
                domain_id: Id {
                    name: "garden_of_live_flowers",
                },
            }: AssetDefinitionEntry {
                definition: AssetDefinition {
                    id: DefinitionId {
                        name: "cabbage",
                        domain_id: Id {
                            name: "garden_of_live_flowers",
                        },
                    },
[...]

PR #3160, opened 2023-02-16
[feature] #3153: make serde_client_cli output machine-readable

QuentinI:
### Description of the Change

Adds `--json` flag to `client_cli` for usage in scripts.

### Issue

#3153 

### Benefits

Allows for automated testing

### Possible Drawbacks

None that I can think of

### Usage Examples or Tests

```shell
$ ./iroha_client_cli --json asset list all | jq '.[1].id'
"rose##alice@wonderland"
```


client_cli/src/main.rs
```diff
@@ -158,19 +168,34 @@ fn main() -> Result<()> {
...
+        eprintln!("This is a release build, debug information omitted from messages");
+    }
+
+    let subcommand_output = subcommand.run(&config)?;
+    if json {
+        println!("{}", serde_json::to_string_pretty(&subcommand_output)?)
+    } else {
+        println!(
+            "{}",
+            ron::Options::default()
```
QuentinI, line 194: The easiest way was to just rely on serde for both machine- and human-facing outputs.
I opted for [RON](https://docs.rs/ron/latest/ron/) for human-readable output, since it resembles what we had earlier (rust debug output) the most. In principle it could be any human-oriented format that's implemented for serde, like [gura](https://gura.netlify.app/)

appetrosyan: For completeness, here's our other options
- S-expressions would be fast and arguably readable. It's the one format I think could be used for both purposes, but neither as fast as JSON, nor as compact and readable as TOML. 
- JSON is a bit noisy, but it's a de-facto standard. It will look great once we've removed all the redundancies like `Identifiable` from the serialization `impl`s. 
- YAML is terrible format that needs to die. Far too much complexity traded for Python-like syntax. The only advantage those have over curly-braced JSON is that you have to highlight whitespace in order to determine if the nesting is correct. 
- XML is considered harmful. There are good reasons to use it, like for example wanting your data to be extensible, and mostly text. This application has a reputation for being [most unwise](http://harmful.cat-v.org/software/xml/). 
- RON is a 1/1 mapping onto Rust. It's very useful if we want to keep typing information and avoid redundancy. It was the **right call** to prefer this for machine readable interactions, even if this adds to the code size. 
- TOML is useful where data is concentrated on one level, e.g. flat or singly nested structures like our configuration file(s), or is extremely sparse (e.g. 
```toml
[Register.Identifiable.Domain]
id = "wonderland"
metadata = {}
```
Sadly, TOML is not the best option for our purposes, given that the syntax extensions it offers don't help for the kind of data that `iroha_client_cli` will return.

appetrosyan: I'd also encourage everyone to think about the effects of introducing yet another whitespace-sensitive language into Iroha. We have enough problems as it is with people not being able to decode our error messages. Adding whitespace errors into the mix might not be worth the extra load put on @6r1d

client_cli/src/main.rs
```diff
@@ -158,19 +168,34 @@ fn main() -> Result<()> {
...
-        &json5::to_string(&config).wrap_err("Failed to serialize configuration.")?
-    );
-    #[cfg(not(debug_assertions))]
-    eprintln!("This is a release build, debug information omitted from messages");
-    subcommand.run(&config)?;
+
+    if verbose {
+        eprintln!(
+            "User: {}@{}",
+            config.account_id.name, config.account_id.domain_id
```
appetrosyan, line 177: Comment on why this is not `account_id`'s `core::fmt::Display`

client_cli/src/main.rs
```diff
@@ -158,19 +168,34 @@ fn main() -> Result<()> {
...
+        );
+        #[cfg(debug_assertions)]
+        eprintln!(
+            "{}",
+            &json5::to_string(&config).wrap_err("Failed to serialize configuration.")?
+        );
```
appetrosyan, line 183: This is quaint. JSON5 is useful for accepting more input from the user (e.g. comments) without causing a deserialisation error. I think you could just use plain standard `json` here without losing anything major.

appetrosyan (changes requested):

client_cli/src/main.rs
QuentinI: No reason, can be changed. Probably leftover from when AccountId didn't implement Display or something

client_cli/src/main.rs
QuentinI: It was probably done to avoid introducing another explicit dependency. Since we're unlikely to somehow not transitively depend on serde_json anyway, this can be changed as well.

client_cli/src/main.rs
QuentinI: Re: whitespace - if we don't accept any input in hypothetical whitespace-sensitive language, we should be fine.
RON looks more natural to me anyways though.

Re: s-exps - as a Vim user I would argue they aren't readable :)

QuentinI:
Coverage is [failing](https://github.com/hyperledger/iroha/actions/runs/4203163925/jobs/7292221076) [again](https://github.com/hyperledger/iroha/actions/runs/4198281973/jobs/7281711180) (though intermittently)

client_cli/src/main.rs
Erigara: RON looks like fine choose, however i don't think that it is all that different from JSON in terms of readability.

Erigara (approved):

client_cli/src/main.rs
```diff
@@ -243,19 +262,14 @@ mod events {
         }
     }
 
-    pub fn listen(filter: FilterBox, cfg: &Configuration) -> Result<()> {
+    pub fn listen(filter: FilterBox, cfg: &Configuration) -> Result<Box<dyn Serialize>> {
         let iroha_client = Client::new(cfg)?;
-        println!("Listening to events with filter: {filter:?}");
-        for event in iroha_client
+        eprintln!("Listening to events with filter: {filter:?}");
```
mversic, line 267: why are you printing to `stderr`?

client_cli/src/main.rs
appetrosyan: IMO the fewer formats we use the better. The only exception is the DSL, because it's tailored to our use case. RON is fine because we were already using it. Shall we reconsider and move to JSON @Erigara ?

client_cli/src/main.rs
QuentinI: So that `iroha_client_cli` can be used in scripts, otherwise user would need to somehow separate debug output and the actual invocation results by hand.

appetrosyan:
RE: coverage, probably not your fault. Our tests are quite unstable since we moved to the new sumeragi.

client_cli/src/main.rs
```diff
@@ -76,6 +78,12 @@ pub struct Args {
     /// Sets a config file path
     #[structopt(short, long)]
     config: Option<Configuration>,
+    /// Output results in machine-readable JSON
+    #[structopt(short, long)]
+    json: bool,
+    /// More verbose ouput
```
pesterev, line 84:
```suggestion
    /// More verbose output
```

client_cli/src/main.rs
```diff
@@ -76,6 +78,12 @@ pub struct Args {
     /// Sets a config file path
     #[structopt(short, long)]
     config: Option<Configuration>,
+    /// Output results in machine-readable JSON
+    #[structopt(short, long)]
+    json: bool,
```
pesterev, line 83: I believe this flag should be enabled by default because JSON output is more useful in most cases:
+ pipe this CLI tool and parse it as JSON
+ easy to copy and paste into JSON explorers

pesterev (commented):
Overall looks good to me

client_cli/src/main.rs
QuentinI: If we agree JSON is more useful even for non-script usage we could then drop RON alltogether and make it JSON-only. @appetrosyan @Erigara what do you think?

Erigara: @QuentinI I'm for leaving only `json`. If we really need to we can always expand range of supported formats, but do other way around may cause some confusion for the users.

appetrosyan: For now, my opinion is that we should use JSON-only. If we decide to change it to something else later, there should be a good justification. 

It makes some sense to change the config to TOML, but only the config, the rest is better served with a simple format such as JSON.

client_cli/Cargo.toml
```diff
@@ -27,6 +27,10 @@ clap = { version = "3.2.23", features = ["derive"] }
 dialoguer = { version = "0.10.2", default-features = false }
 json5 = "0.4.1"
 once_cell = "1.16.0"
+lazy_static = "1.4.0"
+serde_json = "1.0.93"
+ron = "0.8.0"
```
Erigara, line 32: Should be removed

Erigara (approved):

appetrosyan (dismissed):

PR #3163, opened 2023-02-17
[documentation]: Actualise `api_spec.md` and fix doc comments

0x009922:
### Description of the Change

- Update `api_spec.md`: update struct names, WebSocket documentation details, fix broken links, format code.
- Update doc comments in `block.rs` and `events/mod.rs` so they don't contain outdated references as well. Also make some of them a little bit more concise.

### Issue

None.

### Benefits

Documentation will be up-to-date and clearer.

### Possible Drawbacks

None.


docs/source/references/api_spec.md
```diff
@@ -74,13 +85,10 @@ Whether each prerequisite object was found and `FindError`:
...
-**Expects**:
-
-First message after handshake from client: [`EventStreamSubscriptionRequest`](#iroha-structures)
-
-When server is ready to transmit events it sends: [`EventStreamSubscriptionAccepted`](#iroha-structures)
+**Communication**:
 
-The server sends `Event` and expects to receive [`EventReceived`](#iroha-structures) before sending the next event.
+After handshake, client should send [`VersionedEventSubscriptionRequest`](#iroha-structures). Then server
+sends [`VersionedEventMessage`](#iroha-structures).
```
outoftardis, line 91:
```suggestion
After handshake, the client sends [`VersionedEventSubscriptionRequest`](#iroha-structures). Then the server
sends [`VersionedEventMessage`](#iroha-structures).
```

outoftardis (approved):

outoftardis (approved):

outoftardis (approved):

docs/source/references/api_spec.md
```diff
@@ -323,37 +347,34 @@ Learn [how to use metrics](https://hyperledger.github.io/iroha-2-docs/guide/adva
...
 
-For more information on codec check [Substrate Dev Hub](https://substrate.dev/docs/en/knowledgebase/advanced/codec) and codec's [Github repository](https://github.com/paritytech/parity-scale-codec).
+For more information on codec check [Substrate Dev Hub](https://substrate.dev/docs/en/knowledgebase/advanced/codec) and
+codec's [GitHub repository](https://github.com/paritytech/parity-scale-codec).
 
 ## Reference Iroha Client Implementation
 
 [Iroha client in Rust.](../../../client)
 
 ## Iroha Structures
```
mversic, line 368: how is this section relevant?

docs/source/references/api_spec.md
```diff
@@ -30,39 +33,47 @@
...
 
 **Expects**:
+
 - Body: [`VersionedSignedQueryRequest`](#iroha-structures)
 - Query parameters:
-  + `start`: Optional parameter in queries where results can be indexed. Use to return results from specified point. Results are ordered where can be by id which uses rust's [PartialOrd](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable) and [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html) traits.
-  + `limit`: Optional parameter in queries where results can be indexed. Use to return specific number of results.
-  + `sort_by_metadata_key`: Optional parameter in queries. Use to sort results containing metadata with a given key.
+    - `start`: Optional parameter in queries where results can be indexed. Use to return results from specified point.
+      Results are ordered where can be by id which uses
```
mversic, line 40: something's not correct here?

docs/source/references/api_spec.md
```diff
@@ -30,39 +33,47 @@
...
-| Y | Y | Y | N | `FindError::Asset(AssetId)` |
+
+| Domain | Account | Asset Definition | Asset | `FindError`                                     |
+| ------ | ------- | ---------------- | ----- | ----------------------------------------------- |
+| N      | -       | -                | -     | `FindError::Domain(DomainId)`                   |
+| Y      | N       | -                | -     | `FindError::Account(AccountId)`                 |
+| Y      | -       | N                | -     | `FindError::AssetDefinition(AssetDefinitionId)` |
+| Y      | Y       | Y                | N     | `FindError::Asset(AssetId)`                     |
 
 #### Account Not Found 404
```
mversic, line 69: place it before asset

pesterev (approved):

docs/source/references/api_spec.md
```diff
@@ -225,30 +243,35 @@ Also returns current status of peer in json string:
...
+  txs_accepted: 3,
+  // Total number of rejected transactions
+  txs_rejected: 0,
+  // Uptime with nanosecond precision since creation of the genesis block
+  uptime: {
+    secs: 5,
+    nanos: 937000000,
+  },
+  // Number of view changes in the current round
+  view_changes: 0,
```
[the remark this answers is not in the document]
0x009922: I would better use JSDoc notation which is both readable & doc comments standard:

```json5
{
  /**
   * Number of connected peers, except for the reporting peer itself
   */
  peers: 3
}
```

What do you think?

docs/source/references/api_spec.md
0x009922: This section contains all the structures mentioned in the whole document. The purpose of it is to know the exact location of where structs are defined. 

At least it is my understanding of why this section was here. For me, as an SDK developer, it is useful, but I often found that the core team developers don't actualise this table.

docs/source/references/api_spec.md
0x009922: Nothing incorrect, that is the decision of Markdown formatted I used.

docs/source/references/api_spec.md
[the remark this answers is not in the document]
appetrosyan: If it's a standard, then by all means.

docs/source/references/api_spec.md
appetrosyan: > core team developers don't actualise this table. 

Actually if we could just delegate it to automatically generated docs, we won't need to.

docs/source/references/api_spec.md
appetrosyan: If you want us to use prettier, we should probably check doc formatting in CI. Should we add a ticket for that?

core/src/block.rs
```diff
@@ -711,23 +711,23 @@ impl From<SignedBlock> for VersionedCandidateBlock {
...
+    /// Convert to V1 reference
     #[inline]
     pub const fn as_v1(&self) -> &CommittedBlock {
         match self {
             Self::V1(v1) => v1,
         }
     }
 
-    /// Convert from `&mut VersionedCommittedBlock` to V1 mutable reference
+    /// Convert to V1 mutable reference
```
appetrosyan, line 722: These are self-explanatory and frankly not very well-documented. I'd be fine with `#[allow(missing_docs)]`

docs/source/references/api_spec.md
```diff
@@ -323,37 +359,34 @@ Learn [how to use metrics](https://hyperledger.github.io/iroha-2-docs/guide/adva
...
-- `EventStreamSubscriptionRequest`: `iroha_data_model::events::EventSubscriberMessage::SubscriptionRequest`
-- `EventStreamSubscriptionAccepted`: `iroha_data_model::events::EventPublisherMessage::SubscriptionAccepted`
-- `Event`: `iroha_data_model::events::EventPublisherMessage::Event`
-- `EventReceived`: `iroha_data_model::events::EventSubscriberMessage::EventReceived`
-
-- `BlockStreamSubscriptionAccepted`: `iroha_core::block::stream::BlockPublisherMessage::SubscriptionAccepted`
-- `BlockStreamSubscriptionRequest`: `iroha_core::block::stream::BlockSubscriberMessage::SubscriptionRequest`
-- `Block`: `iroha_core::block::stream::BlockPublisherMessage::Block`
-- `BlockReceived`: `iroha_core::block::stream::BlockSubscriberMessage::BlockReceived`
+| Short name                          | Fully qualified name                                           |
```
appetrosyan, line 382:
```suggestion
| Object name                          | Fully qualified object name                                           |
```

docs/source/references/api_spec.md
```diff
@@ -225,30 +243,47 @@ Also returns current status of peer in json string:
...
+   * Number of view changes in the current round
+   */
+  view_changes: 0,
 }
 ```
-__CAUTION__: Almost all fields are 64-bit integers and should be handled with care in JavaScript. Only the `nanos` field is 32-bit integer. See `iroha_telemetry::metrics::Status`.
 
-**Sub-routing**: To obtain the value of a specific field, one can append the name of the field to the path, e.g. `status/peers`. This returns the corresponding JSON value, inline, so strings are quoted, numbers are not and maps are presented as above. 
+**CAUTION**: Almost all fields are 64-bit integers and should be handled with care in JavaScript. Only the `nanos` field
+is 32-bit integer. See `iroha_telemetry::metrics::Status`.
```
appetrosyan, line 282:
```suggestion
is a 32-bit integer. See `iroha_telemetry::metrics::Status`.
```

docs/source/references/api_spec.md
```diff
@@ -225,30 +243,47 @@ Also returns current status of peer in json string:
...
-  + Number of committed blocks (block height)
-  + Total number of accepted transactions
-  + Total number of rejected transactions
-  + Uptime with nanosecond precision since creation of the genesis block
-  + Number of view changes in the current round
 
-```json
+200 OK reports status:
+
+```json5
```
appetrosyan, line 249: This is **not** what you would receive as a response, because we don't provide JSON5 formatted replies. This either needs to be clarified or removed.

docs/source/references/api_spec.md
```diff
@@ -204,13 +220,15 @@ Acceptable values are `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, corresponding t
...
 - 200 OK: The peer is up.
-Also returns current status of peer in json string:
+  Also returns current status of peer in json string:
+
 ```
 "Healthy"
 ```
 
-## Endpoints for [status/metrics](./config.md#toriitelemetry_url)
+## Endpoints for Status/Metrics
```
appetrosyan, line 231: Why remove the link?

docs/source/references/api_spec.md
```diff
@@ -146,28 +152,34 @@ continues to stream blocks as they are added to the blockchain.
...
 curl -X GET -H 'content-type: application/json' http://127.0.0.1:8080/configuration -d '{"Docs" : ["torii"]} ' -i
 ```
 
 **Responses**:
+
 - 200 OK: Field was found and either doc or value is returned in json body.
 - 404 Not Found: Field wasn't found
 
-### Configuration
+### Update Configuration
```
appetrosyan, line 182: Please revert. 

This is the configuration endpoint. 

Also, you don't update the configuration of something, you *configure* something. Configuration means also the process of configuring something.

docs/source/references/api_spec.md
appetrosyan: On that note, the wrap column should be set to 60, rather than 80 for markup.

docs/source/references/api_spec.md
0x009922: Because it was a dead link.

docs/source/references/api_spec.md
0x009922: I'd rather clarify than remove comments

core/src/block.rs
0x009922: Should I remove all edited comments of this kind and replace them with `#[allow(missing_docs)]`?

docs/source/references/api_spec.md
0x009922: I think it would be great if the whole `api_spec` is replaced with, for example, top-level documentation of the crate that is responsible for API. It will allow at least to use direct rustdoc links.

docs/source/references/api_spec.md
0x009922: > If you want us to use prettier

No, I don't. I just used it once to fix ugly formatting. In general I don't think it is worth to setup a CI to check docs formatting, and manual formatting is just find (until the document eventually becomes ugly again)

docs/source/references/api_spec.md
0x009922: Clarified.

docs/source/references/api_spec.md
appetrosyan: So should we add the doc comments to `iroha_cli`? If this reduces the maintenance, and we only need to maintain the doc comments in one place, it'd save us a lot of headache.

appetrosyan: See #3296

docs/source/references/api_spec.md
appetrosyan: > WDYM? I understand the sentence, but am confused by "for markup".

Subtle point. Typographers work with 80 column-wide screens but the actual fill-column threshold for optimal viewing is 60 characters for text. We can get away with 80 in programming mostly because most well-indented programs are 90% blank space. 

> No, I don't. I just used it once to fix ugly formatting. In general I don't think it is worth to setup a CI to check docs formatting, and manual formatting is just fine (until the document eventually becomes ugly again)

OK.

core/src/block.rs
appetrosyan: Yep

0x009922: Done.

appetrosyan (approved):

core/src/block.rs
```diff
@@ -441,23 +440,23 @@ mod candidate {
     }
 
     impl VersionedCandidateBlock {
-        /// Convert from `&VersionedCandidateBlock` to V1 reference
+        #[allow(missing_docs)]
```
mversic, line 443: can we just discuss what is the policy with documentation here?

6r1d (approved):
LGTM

Issue #2133, opened 2022-04-21
Revise the view change method


s8sato:
Besides shift & shuffle? See [discussion](https://github.com/hyperledger/iroha/pull/2124#discussion_r854453439)

s8sato:
I'd suggest:
- Shuffle on the nth view change, where n % 3 == 0
- Shift by `f`, otherwise

The idea is to maximize the difference of the set of peers.

This design
- affects how `BlockSynchronizer` reproduces topologies
- might be just a makeshift until the emergence of #1211

s8sato:
I’ve noticed that an unintended change has broken network liveness, so I’m reopening this issue.

Since PR #3250, view changes have been implemented as a simple rotation. This means that even if there are enough healthy peers overall, if faulty peers are scattered throughout the topology, no amount of rotation can guarantee 2f+1. In other words, the PR has reverted [this issue](https://lf-hyperledger.atlassian.net/wiki/spaces/iroha/pages/21012889/Change+View+Drawback+in+Sumeragi
) to an unresolved state.

To reproduce the problem, for example, start five peers in a swarm, randomly stop two of them, and observe that transactions fail to commit.

Unless we adopt Hijiri, we should reintroduce topology shuffling.

Issue #3168, opened 2023-02-17
Find a way to check genesis.json in CI overcoming cargo builds reproducibility problem


Arjentix:
Originally posted by me [here](https://github.com/hyperledger/iroha/pull/3106#issuecomment-1433730905).

After the merge of #3106 there will be a problem, that local `kagami` will generate different `genesis.json` rather than `kagami` in CI. This is caused by [cargo builds are not reproducibile](https://github.com/rust-lang/cargo/issues/5505) problem, which results in generating different WASM code for validator, which is inserted in our `genesis.json`.

We need to find a way to properly check our genesis in CI.

Arjentix:
One of disscussed solutions.

Store a path to a wasm blob and some kind of build-hash inside genesis. Build-hash should take macros, changes in dependencies and etc into account. In CI don't check the exact wasm blob file

PR #3250, opened 2023-02-22
[fix] #2133: Rewrite topology to be closer the whitepaper.

SamHSmith:


crypto/src/signature.rs
```diff
@@ -51,7 +51,7 @@ pub type Payload = Vec<u8>;
 pub struct Signature {
     /// Public key that is used for verification. Payload is verified by algorithm
     /// that corresponds with the public key's digest function.
-    public_key: PublicKey,
+    pub public_key: PublicKey,
```
mversic, line 54: please restore. This is part of the public facing API and we don't want to be exposing internals of our structures to the public. If you think this forces you to do much cloning in the `iroha_core` we should introduce `transparent_api` as we're going to do in #3127

core/src/sumeragi/main_loop.rs
```diff
@@ -836,17 +869,21 @@ fn reset_state(
...
 
-    if current_view_change_index != *old_view_change_index {
-        current_topology.rebuild_with_new_view_change_count(current_view_change_index);
+    while *old_view_change_index < current_view_change_index {
+        *old_view_change_index += 1;
+        error!(addr=%peer_id.address, "Rotating the entire topology.");
+        current_topology.rotate_all();
         was_commit_or_view_change = true;
+
+        println!("{:?}", current_topology);
```
mversic, line 881: print?

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,174 @@
...
+                .collect::<Vec<PeerId>>(),
+        );
+        self.sorted_peers = new_list;
+    }
+    /// Called after each failed attempt to create a block.
+    pub fn rotate_all(&mut self) {
+        let top = self.sorted_peers.remove(0);
+        self.sorted_peers.push(top);
+    }
+    /// Called after each successful block commit.
```
outoftardis, line 156:
```suggestion
    /// Re-arrange the set of peers after each successful block commit.
```

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,174 @@
...
+                        .iter()
+                        .position(|p| p.public_key == pid.public_key)
+                        .is_none()
+                })
+                .cloned()
+                .collect::<Vec<PeerId>>(),
+        );
+        self.sorted_peers = new_list;
+    }
+    /// Called after each failed attempt to create a block.
```
outoftardis, line 151:
```suggestion
    /// Rotate peers after each failed attempt to create a block.
```

crypto/src/signature.rs
SamHSmith: I didn't realize that the derived Getter trait added a getter. Will revert this change.

core/src/sumeragi/main_loop.rs
SamHSmith: forgot to remove this.

Erigara:
@SamHSmith can you please explain why we need to carry topology in every block since every peer anyway uses it's own topology to verify that received topology is correct?

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,182 @@
...
-        &self.sorted_peers
+    /// What role does this peer have in the topology. If it is not in the toplogy
+    /// the function will return `Role::ObservingPeer`.
+    // This lint is a bad suggestion.
+    #[allow(clippy::option_if_let_else)]
+    pub fn role(&self, peer_id: &PeerId) -> Role {
+        if let Some(index) = self
+            .sorted_peers
+            .iter()
+            .position(|p| p.public_key == peer_id.public_key)
```
mversic, line 110:
```suggestion
            .position(|p| p == peer_id)
```

peers are compared by public keys

mversic (changes requested):

SamHSmith:
> @SamHSmith can you please explain why we need to carry topology in every block since every peer anyway uses it's own topology to verify that received topology is correct?

The old topology was based on a block has and a view change index. This is bad for security since the leader can influence the block hash and therefore influence the future topology.

Now that we have a constantly evolving topology whos operations are destructive, rolling back becomes harder. Instead of keeping old topologies around locally we store it in the block header. This way when there is a softfork or odd consensus a peer can see that enough peers signed this block and adopt the topology. Topology is checked while verifying but not when receiving block through syncing.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,182 @@
...
-    pub fn validating_peers(&self) -> &[PeerId] {
-        let a_set = self.peers_set_a();
-        if a_set.len() > 1 {
-            &a_set[1..(a_set.len() - 1)]
-        } else {
-            &[]
-        }
+    /// How many faulty peers can this topology tolerate.
+    pub fn max_faults(&self) -> usize {
+        (self.sorted_peers.len().saturating_sub(1)) / 3
```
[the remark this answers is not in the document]
SamHSmith: Yes, although many of our tests create empty topologies.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,182 @@
...
+        Topology {
+            sorted_peers: peers.into_iter().collect(),
+        }
     }
-
-    /// Answers if the consensus stage is required with the current number of peers.
+    /// Is consensus required, aka are there more than 1 peer.
     pub fn is_consensus_required(&self) -> bool {
-        self.min_votes_for_commit() > 1
+        self.sorted_peers.len() > 1
```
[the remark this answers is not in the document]
SamHSmith:
```
min_votes_for_commit = max_faults * 2 + 1
max_faults = (peer_count - 1) / 3

min_votes_for_commit = ((peer_count - 1) / 3) * 2 + 1
min_votes_for_commit = (2 * peer_count - 2) / 3 + 1

| peer_count | min_votes_for_commit |
|        1          |          1                          |
|        2          |          1                          |
|        3          |          2                          |
|        4          |          3                          |
```
Interestingly the correct answer seems to be `> 2`, in all the time I've worked with sumeragi I've thought of the is_consensus_required as "are we running a solo node". Probably an artifact of how the code behaved before the rewrite. This is a good catch, I will change it back.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,182 @@
...
-            let peers = sort_peers_by_hash_and_counter(peers, &at_block, last_shuffled_at);
-            shift_peers_by_n(peers, since_last_shuffle)
-        };
-        Ok(Topology { sorted_peers })
-    }
-}
-
 /// The ordering of the peers which defines their roles in the current round of consensus.
-#[derive(Clone, Debug, Encode, Decode, IntoSchema, Serialize)]
+#[derive(PartialEq, Eq, Clone, Debug, Encode, Decode, IntoSchema, Serialize)]
```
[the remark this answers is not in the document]
SamHSmith: I will change it.

core/src/sumeragi/network_topology.rs
SamHSmith: Good to know thanks.

SamHSmith:
The genesis check is failing for some reason. It should not. And it succeeds on my machine.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,182 @@
...
             .collect()
     }
-
-    /// Sorted peers that this topology has.
-    pub fn sorted_peers(&self) -> &[PeerId] {
-        &self.sorted_peers
+    /// What role does this peer have in the topology. If it is not in the toplogy
+    /// the function will return `Role::ObservingPeer`.
+    // This lint is a bad suggestion.
+    #[allow(clippy::option_if_let_else)]
```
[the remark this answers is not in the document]
appetrosyan: Leave them in, if you like disable them in this crate, but not repo-wide

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,183 @@
...
     ) -> Vec<SignatureOf<T>> {
-        let roles: HashSet<Role> = roles.iter().copied().collect();
-        let public_keys: HashSet<_> = roles
-            .iter()
-            .flat_map(|role| role.peers(self))
-            .map(|peer| peer.public_key)
-            .collect();
+        let mut public_keys = Vec::new();
+        for role in roles {
+            let role_public_keys = if self.max_faults() == 0 {
```
appetrosyan, line 63: You can match on a tuple. e.g. 

```rust 
match (role, self.max_faults()) { 
   (Role::Leader, 0) => vec![self.sorted_peers[0].public_key.clone()],
   // which you can only see because of matching twice. 
   (Role::Leader, _) => vec![self.sorted_peers[0].public_key.clone()]
   // And similarly for other patterns. 
}

```

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,183 @@
...
     }
-
-    /// Verifies that this `message` was signed by the `signature` of a peer with specified `role`.
-    ///
-    /// # Errors
-    /// Fails if there are no such peer with this key and if signature verification fails
-    // TODO: Not used?
-    pub fn verify_signature_with_role(
+    /// Filter signatures by roles in the topology.
+    #[allow(clippy::comparison_chain)]
```
appetrosyan, line 55:
```suggestion

```

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,183 @@
...
-    }
-
-    /// Apply a `view change`, i.e. change the topology in case there were faults in the consensus round.
-    #[allow(clippy::expect_used)]
-    pub fn rebuild_with_new_view_change_count(&mut self, view_change_count: u64) {
-        *self = Builder::new().with_peers(self.sorted_peers.iter().cloned()).build(view_change_count).expect(
-            "Invalid topology. At this stage the error is unrecoverable. Please restart the peer.",
-        )
+    /// Create a new topology.
+    pub fn new<T: IntoIterator<Item = PeerId>>(peers: T) -> Self {
```
appetrosyan, line 37:
```suggestion
    pub fn new(peers: impl IntoIterator<Item = PeerId>) -> Self {
```

Minor point, but this is the preferred approach

core/src/sumeragi/main_loop.rs
```diff
@@ -976,11 +1011,11 @@ pub(crate) fn run<F: FaultInjection>(
 
             if let Some(VotingBlock { block, .. }) = voting_block.as_ref() {
                 // NOTE: Suspecting the tail node because it hasn't yet committed a block produced by leader
-                warn!(%role, block=%block.hash(), "Block not committed in due time, requesting view change...");
+                warn!(peer_public_key=?sumeragi.peer_id.public_key, %role, block=%block.hash(), "Block not committed in due time, requesting view change...");
```
appetrosyan, line 1014:
```suggestion
                warn!(peer_public_key=%sumeragi.peer_id.public_key, %role, block=%block.hash(), "Block not committed in due time, requesting view change...");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -976,11 +1011,11 @@ pub(crate) fn run<F: FaultInjection>(
...
             } else {
                 // NOTE: Suspecting the leader node because it hasn't produced a block
                 // If the current node has a transaction, the leader should have as well
-                warn!(%role, "No block produced in due time, requesting view change...");
+                warn!(peer_public_key=?sumeragi.peer_id.public_key, %role, "No block produced in due time, requesting view change...");
```
appetrosyan, line 1018:
```suggestion
                warn!(peer_public_key=%sumeragi.peer_id.public_key, %role, "No block produced in due time, requesting view change...");
```

core/src/sumeragi/main_loop.rs
```diff
@@ -1065,6 +1100,15 @@ fn vote_for_block<F: FaultInjection>(
         return None;
     }
 
+    if block.header().committed_with_topology != state.current_topology {
+        error!(
+            %addr, %role, block_topology=?block.header().committed_with_topology, my_topology=?state.current_topology, hash=%block.hash(),
```
appetrosyan, line 1105:
```suggestion
            %addr, %role, block_topology=%block.header().committed_with_topology, my_topology=%state.current_topology, hash=%block.hash(),
```

Consider implementing `core::fmt::Display` for `block_header` and `Topology`.

Erigara:
> The genesis check is failing for some reason. It should not. And it succeeds on my machine.

@SamHSmith i think this is due to not reproducibility of wasm builds, current workaround is to take genesis from CI. 

This problem was recently introduced by runtime validators PR and will be solved later #3168.

Erigara:
> The old topology was based on a block has and a view change index. This is bad for security since the leader can influence the block hash and therefore influence the future topology.

So now topology is kinda predetermined, right?

core/src/sumeragi/network_topology.rs
appetrosyan: Although, sometimes, it's better to do `impl FromIterator for ...`

core/src/sumeragi/network_topology.rs
```diff
@@ -355,188 +197,3 @@ pub enum Role {
...
-        match self {
-            Role::Leader => vec![network_topology.leader().clone()],
-            Role::ValidatingPeer => network_topology.validating_peers().to_vec(),
-            Role::ObservingPeer => network_topology.peers_set_b().to_vec(),
-            Role::ProxyTail => vec![network_topology.proxy_tail().clone()],
-        }
-    }
-}
-
-#[cfg(test)]
```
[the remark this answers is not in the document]
SamHSmith: They don't test things that are present in the new code. The new code is so simple that have dedicated tests is a very marginal benefit. You would just end up implementing it twice so that one copy of the code could test the other.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,183 @@
...
+        );
+        self.sorted_peers = new_list;
+    }
+    /// Rotate peers after each failed attempt to create a block.
+    pub fn rotate_all(&mut self) {
+        let top = self.sorted_peers.remove(0);
+        self.sorted_peers.push(top);
+    }
+    /// Re-arrange the set of peers after each successful block commit.
+    pub fn rotate_a_set(&mut self) {
```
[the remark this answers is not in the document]
SamHSmith: will change

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,183 @@
...
-
-    /// Sorted peers that this topology has.
-    pub fn sorted_peers(&self) -> &[PeerId] {
-        &self.sorted_peers
+    /// What role does this peer have in the topology. If it is not in the toplogy
+    /// the function will return `Role::ObservingPeer`.
+    // This lint is a bad suggestion.
+    #[allow(clippy::option_if_let_else)]
+    pub fn role(&self, peer_id: &PeerId) -> Role {
+        if let Some(index) = self.sorted_peers.iter().position(|p| p == peer_id) {
```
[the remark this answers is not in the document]
SamHSmith: sure

core/src/sumeragi/network_topology.rs
SamHSmith: sure

core/src/sumeragi/network_topology.rs
SamHSmith: I'm keeping this one because the code is not better if I follow the lint suggestion.

core/src/sumeragi/main_loop.rs
SamHSmith: sure

core/src/sumeragi/main_loop.rs
SamHSmith: sure

core/src/sumeragi/main_loop.rs
SamHSmith: I don't think it would be useful.

appetrosyan (approved):

appetrosyan (dismissed):

core/src/genesis.rs
```diff
@@ -88,31 +84,14 @@ async fn try_get_online_topology(
     network_topology: &Topology,
     network: Addr<IrohaNetwork>,
 ) -> Result<Topology> {
-    let (online_peers, offline_peers) =
+    let (online_peers, _offline_peers) =
         check_peers_status(this_peer_id, network_topology, network).await;
```
mversic, line 88: change `check_peers_status` so that it returns only online peers

core/src/sumeragi/main_loop.rs
```diff
@@ -628,6 +657,9 @@ fn handle_message<F: FaultInjection>(
                 let voting_block_hash = voted_block.block.hash();
 
                 if hash == voting_block_hash.transmute() {
+                    // The manipulation of the topology relies upon all peers seeing the same signature set.
+                    // Therefore we must clear the signatures and accept what the proxy tail giveth.
+                    voted_block.block.signatures.clear();
                     add_signatures::<true>(&mut voted_block, signatures.transmute());
```
mversic, line 663: I don't like the `clear` being introduced

```suggestion
                    voted_block.block.signatures = signatures;
```

what would happen in this case?

core/src/sumeragi/main_loop.rs
```diff
@@ -842,17 +875,19 @@ fn reset_state(
         // a view change a new block is immediately created by the leader
         *round_start_time = Instant::now();
         was_commit_or_view_change = true;
+        *old_view_change_index = 0;
```
mversic, line 878: are you sure you want to reset this on every commit?

core/src/sumeragi/main_loop.rs
```diff
@@ -842,17 +875,19 @@ fn reset_state(
...
     }
 
-    if current_view_change_index != *old_view_change_index {
-        current_topology.rebuild_with_new_view_change_count(current_view_change_index);
+    while *old_view_change_index < current_view_change_index {
+        *old_view_change_index += 1;
+        error!(addr=%peer_id.address, "Rotating the entire topology.");
+        current_topology.rotate_all();
```
mversic, line 884: what is the advantage of rotating entire topology `current_view_change_index - old_view_change_index` times vs only once?

mversic: is it so that all peers reach the same topology? and if you didn't do it many times they could reach different topologies?

core/src/sumeragi/main_loop.rs
```diff
@@ -421,8 +416,25 @@ fn commit_block<F: FaultInjection>(
...
+    );
+    current_topology.rotate_set_a();
+    current_topology.update_peer_list(
+        &state
+            .wsv
+            .peers_ids()
+            .iter()
+            .map(|id| id.clone())
+            .collect::<Vec<PeerId>>(),
+    );
```
mversic, line 435: this code is duplicated

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,174 @@
...
+        let mut i = 0;
+        while i < self.sorted_peers.len() {
+            if to_lift_up.contains(&self.sorted_peers[i].public_key) {
+                i += 1;
+            } else {
+                observing.insert(0, self.sorted_peers.remove(i)); // This has to be insert(0) and not push in order to preserve order.
+            }
+        }
+        self.sorted_peers.extend(observing);
     }
```
mversic, line 176: should I mention that worst case complexity is quite high for this simple operation here?

core/src/sumeragi/main_loop.rs
Erigara: We have relevant task for this #3254.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,174 @@
...
-            .at_block(block)
-            .build(0)
-            .expect("Invalid topology");
+    /// Get proxy tail's peer id.
+    pub fn proxy_tail(&self) -> &PeerId {
+        &self.sorted_peers[self.min_votes_for_commit()]
+    }
+    /// Add or remove peers from the topology.
+    pub fn update_peer_list(&mut self, new_peer_list: &[PeerId]) {
+        // The complexity here is to not two peer_ids that are different but have the same key.
```
mversic, line 135: something's wrong with the comment

mversic: also, isn't this impossible? Shouldn't public key be a unique identifier for a peer? @appetrosyan @Erigara

mversic: yes, @SamHSmith please check comment on `PartialEq` implementation for `PeerId`

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,174 @@
...
+    #[allow(clippy::option_if_let_else)]
+    pub fn role(&self, peer_id: &PeerId) -> Role {
+        match self.sorted_peers.iter().position(|p| p == peer_id) {
+            Some(index) if index == 0 => Role::Leader,
+            Some(index) if index < self.min_votes_for_commit() => Role::ValidatingPeer,
+            Some(index) if index == self.min_votes_for_commit() => Role::ProxyTail,
+            Some(_) => Role::ObservingPeer,
+            None => {
+                trace!(%peer_id, "Peer is not in topology.");
+                Role::ObservingPeer
```
mversic, line 121: maybe we should panic here instead

core/src/genesis.rs
SamHSmith: okay

core/src/sumeragi/main_loop.rs
SamHSmith: That would create a security issue. `add_signatures` contains validation code that must run.

core/src/sumeragi/main_loop.rs
SamHSmith: Yes. View change index is a round local value.

core/src/sumeragi/main_loop.rs
SamHSmith: It is so that all peers reach the same topology. And if we don't do it the same amount of times on each peer then they reach different topologies and cannot reach consensus. You are correct.

core/src/sumeragi/network_topology.rs
SamHSmith: No because N is small and it's all in cache so it's not a problem.

core/src/sumeragi/network_topology.rs
```diff
@@ -5,341 +5,174 @@
...
+                .filter(|pid| !self.sorted_peers.iter().any(|p| &p == pid))
+                .cloned()
+                .collect::<Vec<PeerId>>(),
+        );
+        self.sorted_peers = new_list;
+    }
+    /// Rotate peers after each failed attempt to create a block.
+    pub fn rotate_all(&mut self) {
+        let top = self.sorted_peers.remove(0);
+        self.sorted_peers.push(top);
```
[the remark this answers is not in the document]
SamHSmith: sure

core/src/sumeragi/main_loop.rs
mversic: but it didn't used to be?

core/src/sumeragi/main_loop.rs
mversic: This is not good because you're breaking `SignaturesOf` invariants. Check docs on `SignaturesOf`

core/src/sumeragi/network_topology.rs
SamHSmith: Yea I have now rewritten it.

core/src/sumeragi/network_topology.rs
mversic: would you elaborate, how can tests rely on this?

core/src/sumeragi/main_loop.rs
appetrosyan: I'm with @mversic here. We need to be extra careful with the consensus. Whatever is the response to this becomes a comment.

core/src/genesis.rs
```diff
@@ -130,13 +107,13 @@ async fn check_peers_status(
         .peers;
     iroha_logger::info!(peer_count = peers.len(), "Peers status");
 
-    let (online, offline): (SmallVec<_>, SmallVec<_>) = network_topology
-        .sorted_peers()
+    let (online, _offline): (SmallVec<_>, SmallVec<_>) = network_topology
+        .sorted_peers
         .iter()
         .cloned()
         .partition(|id| peers.contains(&id.public_key) || this_peer_id.public_key == id.public_key);
```
mversic, line 114:
```suggestion
        .filter(|id| peers.contains(&id.public_key) || this_peer_id.public_key == id.public_key);
```

core/src/sumeragi/main_loop.rs
appetrosyan: `SignaturesOf` is worth encapsulating. That said, we don't even know off the top of our heads what invariants need to be upheld, and that's a problem

core/src/sumeragi/network_topology.rs
appetrosyan: @mversic Why do you think we should panic here? We ought to be able to recover in this situation.

core/src/sumeragi/main_loop.rs
SamHSmith: Good point. I've now updated the docs to remove that invariant.

core/src/sumeragi/main_loop.rs
SamHSmith: I wrote the old code. I wrote this code. I assure you the behaviour is the same. When I say round, I mean what happens after committing block N and before committing N+1. Within a round 0-INF view changes can occure. It is important to keep track of this value so that we can shift the topology the correct number of times. When a block has been committed the view change index is now zero, then we increment it with each topology rotation.

core/src/sumeragi/network_topology.rs
SamHSmith: They don't use the consensus code but will create blocks with empty topologies. If we panic here then the tests with empty topologies also panic.

mversic: > @mversic Why do you think we should panic here? We ought to be able to recover in this situation.

ok, I don't see how it's possible for this situation to occur. I thought it's kinda equivalent to `assert` statements we have around

core/src/sumeragi/main_loop.rs
appetrosyan: Appealing to authority isn't addressing the comment. 

If this isn't obvious to @mversic, a senior engineer, it's not obvious to whomever is going to read the code afterwards.

core/src/genesis.rs
```diff
@@ -88,31 +83,13 @@ async fn try_get_online_topology(
     network_topology: &Topology,
     network: Addr<IrohaNetwork>,
 ) -> Result<Topology> {
-    let (online_peers, offline_peers) =
-        check_peers_status(this_peer_id, network_topology, network).await;
+    let online_peers = check_peers_status(this_peer_id, network_topology, network).await;
     let set_a_len = network_topology.min_votes_for_commit();
     if online_peers.len() < set_a_len {
         return Err(eyre!("Not enough online peers for consensus."));
```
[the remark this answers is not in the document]
SamHSmith: sure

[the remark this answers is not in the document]
SamHSmith: sure

core/src/sumeragi/network_topology.rs
mversic: so for empty topologies there is only one peer, the one executing the test? shouldn't `self.sorted_peers` then contain that one peer?

you are free to disregard this comment

core/src/sumeragi/main_loop.rs
```diff
@@ -421,8 +416,25 @@ fn commit_block<F: FaultInjection>(
...
     );
 
+    *current_topology = committed_block.header().committed_with_topology.clone();
+    current_topology.lift_up_peers(
+        &committed_block
+            .signatures()
+            .into_iter()
+            .map(|s| s.public_key().clone())
+            .collect::<Vec<PublicKey>>(),
+    );
```
Erigara, line 426: Isn't this approach has potential for abuse too? `ProxyTail` can manipulate whose signatures will go into the block and can control their orded.

core/src/sumeragi/network_topology.rs
SamHSmith: It's the tests that don't even set up a peer. All they do is create a blockchain. And since the topology exists on the blockheader it get's used. But since there are no peers it is empty.

core/src/sumeragi/main_loop.rs
SamHSmith: Yes, but it is less. Order is preserved. So the proxy tail is able to promote b set peers into the a set or demote a set peers into the b set. But is not able to reorder arbitrarely.

mversic (approved):

appetrosyan (approved):

PR #3258, opened 2023-02-27
[feature] #3010: Add transaction expire pipeline event

### LINKED ISSUE #3010: too-small value for `TRANSACTION_TIME_TO_LIVE_MS`,  under a high load, notify failure
### When a client provides a too-small value for TRANSACTION_TIME_TO_LIVE_MS, which is not enough to process transactions under a high load, Then the logging indicates this reason for failure and suggests a solution.

Erigara:
### Description of the Change

- Add transaction expired rejection reason for pipeline event. 
- Improve error message on expiration.

### Issue

Resolves #3010.

### Benefits

User now can observe if transaction is expired and dropped.

### Possible Drawbacks

Not every peer produces the same expire events, because peers can have different transactions in the queue.


core/src/queue.rs
```diff
@@ -205,53 +213,57 @@ impl Queue {
         &self,
         seen: &mut Vec<HashOf<VersionedSignedTransaction>>,
         wsv: &WorldStateView,
+        expired_transactions: &mut Vec<VersionedAcceptedTransaction>,
     ) -> Option<VersionedAcceptedTransaction> {
-        loop {
-            let hash = self.signature_buffer.pop()?;
-            let entry = match self.txs.entry(hash) {
-                Entry::Occupied(entry) => entry,
-                // FIXME: Reachable under high load. Investigate, see if it's a problem.
```
QuentinI, line 213: We lost this FIXME

core/src/queue.rs
```diff
@@ -205,53 +213,57 @@ impl Queue {
...
+    }
+
+    /// Pop single transaction either from the queue or waiting buffer
+    #[inline]
+    fn pop_from<const SKIP_SIGNATURE_CHECK: bool>(
+        &self,
+        queue: &ArrayQueue<HashOf<VersionedSignedTransaction>>,
+        seen: &mut Vec<HashOf<VersionedSignedTransaction>>,
+        wsv: &WorldStateView,
+        expired_transactions: &mut Vec<VersionedAcceptedTransaction>,
```
QuentinI, line 240: Not sure if it makes sense or I'm being naive here, but since all we end up using is hash and ttl, maybe it makes sense not to retain the full struct here?

Erigara: Good point, i considered such approach, but decided to leave queue a bit more general.

data_model/src/transaction.rs
```diff
@@ -897,6 +948,9 @@ pub enum TransactionRejectionReason {
     /// Genesis account can sign only transactions in the genesis block.
     #[display(fmt = "The genesis account can only sign transactions in the genesis block.")]
     UnexpectedGenesisAccountSignature,
+    /// Transaction gets expired.
+    #[display(fmt = "Transaction rejected due to being expired: {}", self.0)]
```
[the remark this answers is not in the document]
Erigara: From my experience only `TransactionExpired` error message is shown to the user so there is no redundancy in that sense.

core/src/queue.rs
QuentinI: I'm mostly worried this might impact performance. Let's see if anyone performance-savvy has input on this, otherwise I'll try and look into it myself

core/src/queue.rs
```diff
@@ -205,53 +213,58 @@ impl Queue {
...
         wsv: &WorldStateView,
+        expired_transactions: &mut Vec<VersionedAcceptedTransaction>,
+    ) -> Option<VersionedAcceptedTransaction> {
+        // NOTE: `SKIP_SIGNATURE_CHECK=true` because `queue` contains only transactions for which signature check is `true`.
+        self.pop_from::<true>(&self.queue, seen, wsv, expired_transactions)
+    }
+
+    /// Pop single transaction either from the queue or waiting buffer
+    #[inline]
+    fn pop_from<const SKIP_SIGNATURE_CHECK: bool>(
```
Arjentix, line 235: Interesting move. This should well optimized by compiler, I believe

core/src/sumeragi/main_loop.rs
```diff
@@ -935,9 +935,18 @@ pub(crate) fn run<F: FaultInjection>(
...
+            &state.wsv,
+            &mut state.transaction_cache,
+            &mut expired_transactions,
+        );
+        sumeragi.send_events(
+            expired_transactions
+                .iter()
+                .map(Txn::expired_event)
+                .collect::<Vec<_>>(),
+        );
```
Arjentix, line 949: I don't understand well the code in this file, but are you sure this is a good place to send events to the client? Shouldn't we record expired events into block as well as we do for failed events? And only then send this info?

data_model/src/transaction.rs
```diff
@@ -122,6 +129,24 @@ pub trait Txn {
         let tx_timestamp = Duration::from_millis(self.payload().creation_time);
         tx_timestamp.saturating_sub(crate::current_time()) > threshold
     }
+
+    /// Create expired pipeline event for the given transaction.
+    #[cfg(feature = "std")]
+    fn expired_event(&self) -> Event
```
Arjentix, line 135: While I understand that this was the easiest way to somehow incapsulate the logic of creating *expired event* related to transaction, I wonder, if it may be usefull to introduce strong-typing for expired transaction?

It may look like this:

```rust
impl Transaction {
    pub fn mark_as_expired(self) -> ExpiredTransaction {
        // ...
    }
}

pub struct ExpiredTransaction(Transaction);

/// I'm not a huge fan of this `From for Event` approach, but we already use this a lot
impl From<ExpiredTransaction> for Event {
    // ...
}
```

Or may be even better in future:

```rust
pub struct Transaction<State> {
    // ...
}

enum State {
    Raw,
    Validated,
    Committed,
    Rejected,
    Expired,
}
```

Just a suggestion. Not sure it will be useful

data_model/src/transaction.rs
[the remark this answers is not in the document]
Arjentix: Why do we even have this kind of duplication? Maybe we can fix it in another PR?

core/src/queue.rs
Erigara: I think so, at least i have tested this such approach and it worked ([link](https://godbolt.org/z/zc58d7GW5)).

data_model/src/transaction.rs
Erigara: I would like to try later approach you proposed one day, but `Expired` state feel like overkill to me, because only use we have for `Expired` transactions is to convert then into `Event`.
However i would like to hear what other reviewers think.

data_model/src/transaction.rs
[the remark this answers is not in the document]
Erigara: Feel free to create issue.

core/src/sumeragi/main_loop.rs
Erigara: After some discussion in the DM we concluded that inclusion of `expired_transactions` is not necessary.

data_model/src/transaction.rs
Arjentix: I agree. After discussion we have we conclude that we don't need `Expired` transactions in `Block`. So it's okay for me to leave as is

data_model/src/transaction.rs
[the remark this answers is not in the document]
Arjentix: Created #3281

Arjentix (approved):

QuentinI (approved):

core/src/queue.rs
appetrosyan: Why are you asking, for someone? Benchmark.

Erigara: @QuentinI @appetrosyan  benchmarking results don't show any performance degradation. 

**NOTE**: Since we are measuring tps and not time results are inverted ("+" is "-", regressed is improved).  

```
Transaction TTL=u64::MAX

Before (aec7aaa5):
time:   [1479.7 tps 1588.9 tps 1693.2 tps]

After:
time:   [1354.4 tps 1595.6 tps 1812.5 tps]
change: [-15.814% +0.4261% +16.030%] (p = 0.96 > 0.05)
No change in performance detected.

Transaction TTL=1_000

Before (aec7aaa5):
# Most of the transactions just expired
time:   [159.19 tps 191.36 tps 223.89 tps] 

After:
time:   [227.29 tps 250.73 tps 273.30 tps]
change: [+9.1402% +31.029% +60.827%] (p = 0.01 < 0.05)
Performance has regressed.
```

Arjentix (dismissed):

ilchu (approved):

appetrosyan (approved):

Issue #3263, opened 2023-02-28
Rename `_iroha_wasm_main()` for triggers and validators


Arjentix:
Originally suggested by @mversic [here](https://github.com/hyperledger/iroha/pull/3106#discussion_r1119633398).

The idea is to rename `_iroha_wasm_main()` to:

1. `_iroha_trigger_main()` -- for triggers;
2. `_iroha_validator_main()` -- for validators.

This will require changes on both Iroha side and procedural macros side.

mversic:
no, my suggestion was to have only one name `_iroha_wasm_main` instead of multiple different

PR #3264, opened 2023-03-01
[feature] #3144: Wasm logging

### LINKED ISSUE #3144: [suggestion] Plug wasm smartcontracts into the Iroha logging system
### Feature request

Route all printing done in smartcontracts to an appropriate log level, with appropriate span, and print using the logging libraries already present in Iroha. 

### Benefits: 

- Logging level affects the verbosity of smartcontracts
- Logging statements are scoped to the correct smartcontract. 
- Smartcontract execution feedback can be routed to searchable log formats (e.g. JSON, `journalctl` etc)

### Motivation

@Mingela mentioned that during the original validation of the blockstore after restarting, many messages from smartcontracts are printed to `stdout`. 

### Who can help?

@appetrosyan

mversic:
### Description of the Change

* replace `lazy_static` with `once_cell`
* replace `anyhow` with `eyre`
* use `SeqCst` vs other memory orderings
* remove `ParseError` from `wasm::Error`
* fix wasm tests
* add logging to smart contracts

### Issue

Closes #3144 #3265 #3263 

### Benefits

Log now looks like this:
image


logger/src/layer.rs
```diff
@@ -144,7 +144,7 @@ impl<S: Subscriber> EventInspectorTrait for LevelFilter<S> {
 
     fn event(&self, event: &Event<'_>) {
         let level = Self::level_as_u8(*event.metadata().level());
-        if level >= CURRENT_LEVEL.load(Ordering::Relaxed) {
+        if level >= CURRENT_LEVEL.load(Ordering::SeqCst) {
```
mversic, line 147: this one is worth leaving `Relaxed`

wasm/src/logging.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Error
+    ERROR,
+}
+
+impl MaxLogLevel {
+    fn get(&self) -> LogLevel {
+        use core::sync::atomic::Ordering;
+
+        let inner = &self.0;
+        if inner.load(Ordering::SeqCst) == u8::MAX {
```
mversic, line 34: I might consider using `Relaxed`

wasm/src/logging.rs
```diff
@@ -0,0 +1,168 @@
...
+    WARN,
+    /// Error
+    ERROR,
+}
+
+impl MaxLogLevel {
+    fn get(&self) -> LogLevel {
+        let mut log_level = self.0.load(Ordering::SeqCst);
+
+        if log_level == u8::MAX {
```
mversic, line 34: an alternative to lazy loading would be to query host log level at the beginning of every smart contract

wasm/src/logging.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[cfg(not(test))]
+    use host::query_max_log_level as host_query_max_log_level;
+    #[cfg(test)]
+    use tests::_query_max_log_level_mock as host_query_max_log_level;
+
+    // SAFETY: `host_query_max_log_level` must return a valid log level
+    #[allow(clippy::cast_possible_truncation)]
+    unsafe {
+        core::mem::transmute(host_query_max_log_level() as u8)
+    }
```
mversic, line 71: I'm just assuming here that host is returning a valid value

wasm/src/logging.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[cfg(not(test))]
+mod host {
+    #[link(wasm_import_module = "iroha")]
+    extern "C" {
+        /// Get the max log level set on the host
+        pub(super) fn query_max_log_level() -> u32;
+
+        /// Log string withe the host logging system
```
outoftardis, line 51:
```suggestion
        /// Log string with the host logging system
```

wasm/src/lib.rs
```diff
@@ -74,8 +77,12 @@ unsafe extern "C" fn _iroha_wasm_dealloc(offset: *mut u8, len: usize) {
     let _box = Box::from_raw(core::slice::from_raw_parts_mut(offset, len));
 }
 
+/// Implementing types can be executed on the host
 pub trait Execute {
+    /// Return value of the execution
```
outoftardis, line 82:
```suggestion
    /// Return the result of the execution
```

core/src/smartcontracts/wasm.rs
```diff
@@ -439,8 +445,8 @@ impl<'wrld> Runtime<'wrld> {
         Ok(value_offset)
     }
 
-    /// Host defined function which prints given string. When calling
-    /// this function, module serializes ISI to linear memory and
+    /// Host defined function which prints the given string. When calling
+    /// this function, module serializes string to linear memory and
```
outoftardis, line 449:
```suggestion
    /// Host-defined function which prints the given string. When this function
    /// is called, the module serializes the string to linear memory and
```

core/src/smartcontracts/wasm.rs
```diff
@@ -459,6 +465,56 @@ impl<'wrld> Runtime<'wrld> {
         Ok(())
     }
 
+    /// Host defined function which logs the given string. When calling
+    /// this function, module serializes string to linear memory and
+    /// provides log level, offset and length as parameters
```
outoftardis, line 470:
```suggestion
    /// Host-defined function which logs the given string. When this function
    /// is called, the module serializes the string to linear memory and
    /// provides log level, offset and length as parameters
```

wasm/src/lib.rs
mversic: this is not correct IMO

outoftardis: `the resulting value` maybe?

mversic: ok

core/src/smartcontracts/wasm.rs
```diff
@@ -459,6 +465,56 @@ impl<'wrld> Runtime<'wrld> {
...
+    /// If log level or string decoding fails
+    fn log(
+        mut caller: Caller<State>,
+        log_level: u32,
+        offset: WasmUsize,
+        len: WasmUsize,
+    ) -> Result<(), Trap> {
+        let error_msg = Trap::new(format!("{}: not a valid log level", log_level));
+
+        if let Ok(log_level) = log_level.try_into() {
```
Erigara, line 488: Consider `let else`.

cli/Cargo.toml
```diff
@@ -62,7 +62,7 @@ thiserror = "1.0.38"
 tokio = { version = "1.23.0", features = ["sync", "time", "rt", "io-util", "rt-multi-thread", "macros", "fs", "signal"] }
 warp = "0.3.3"
 serial_test = "0.8.0"
-lazy_static = "1.4.0"
+once_cell = "1.16.0"
```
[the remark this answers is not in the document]
mversic: It's considered to be integrated into std

core/src/smartcontracts/wasm.rs
```diff
@@ -459,6 +465,56 @@ impl<'wrld> Runtime<'wrld> {
...
+    /// # Errors
+    ///
+    /// If log level or string decoding fails
+    fn log(
+        mut caller: Caller<State>,
+        log_level: u32,
+        offset: WasmUsize,
+        len: WasmUsize,
+    ) -> Result<(), Trap> {
+        let error_msg = Trap::new(format!("{}: not a valid log level", log_level));
```
Arjentix, line 486: This is super picky, but ideally you could write this:

```suggestion
        let construct_error_msg = || Trap::new(format!("{}: not a valid log level", log_level));
```

This will reduce extra allocation on happy path which is the most common

wasm/src/logging.rs
```diff
@@ -0,0 +1,168 @@
...
+use super::*;
+
+// NOTE: `u8::MAX` is a sentinel value for an undefined log level
+static MAX_LOG_LEVEL: MaxLogLevel = MaxLogLevel(AtomicU8::new(u8::MAX));
+
+/// Struct which holds a valid [`LogLevel`] stored as an integer
+struct MaxLogLevel(AtomicU8);
+
+/// Log level supported by the host
+// NOTE: This struct must be exact duplicate of `config::logger::Level`
```
Arjentix, line 12: But we can't just import it, right?

wasm/src/logging.rs
Arjentix: So right now we query this at the first attempting to log?

Arjentix:
> Also I'm interested in motivation of moving from `lazy_static` to `once_cell`, I can't find it in #3265.

Okay, I see [here](https://github.com/hyperledger/iroha/pull/3264#discussion_r1121846326) now

wasm/src/logging.rs
mversic: yes

actor/src/actor_id.rs
```diff
@@ -18,7 +18,7 @@ impl ActorId {
     pub fn new(name: Option<&'static str>) -> Self {
         Self {
             name,
-            id: ACTOR_ID_COUNTER.fetch_add(1, Ordering::SeqCst),
+            id: ACTOR_ID_COUNTER.fetch_add(1, Ordering::Relaxed),
```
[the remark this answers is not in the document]
mversic: although I'm against using orderings other than `SeqCst`, in this case it's obviously true that ordering doesn't matter as long as every actor gets a different ID. The atomic is used only as a counter

client/tests/integration/unstable_network.rs
```diff
@@ -131,7 +131,7 @@ fn unstable_network(
 
         // Return all peers to normal function.
         for f in &freezers {
-            f.store(false, Ordering::Relaxed);
+            f.store(false, Ordering::SeqCst);
```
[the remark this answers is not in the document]
mversic: maybe and depending on the architecture. But this is in tests so not relevant and also unlikely that this speed difference can be measured

wasm/src/logging.rs
mversic: I don't want to be dealing with `iroha_logger` crate as yet another dependency in `wasm` and making that crate `no_std` and FFI compatible just for a single enum. I tried moving this enum into `data_model` so I can share it but it just gave me headaches

config/src/logger.rs
```diff
@@ -9,45 +9,42 @@ use iroha_config_base::{
...
     TRACE,
-}
-
-// TODO: derive when bump version to 1.62
-impl Default for Level {
-    fn default() -> Self {
-        DEFAULT_MAX_LOG_LEVEL
-    }
+    /// Debug
+    DEBUG,
```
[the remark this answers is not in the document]
mversic: I don't really understand what you're saying. This enum has uppercase variants so that it would look like `tracing::Level`

core/src/smartcontracts/wasm.rs
```diff
@@ -14,7 +14,12 @@ use iroha_config::{
     base::proxy::Builder,
     wasm::{Configuration, ConfigurationProxy},
 };
-use iroha_data_model::{permission, prelude::*, ParseError};
+use iroha_data_model::{
+    permission::{self, validator},
+    prelude::*,
+};
+// NOTE: Using error_span so that span info is logged on every event
+use iroha_logger::{error_span as wasm_log_span, prelude::tracing::Span, Level as LogLevel};
```
mversic, line 22: are we ok with this?

Cargo.lock
```diff
@@ -2168,6 +2169,7 @@ dependencies = [
  "color-eyre",
  "derive_more",
  "iroha_config",
+ "iroha_data_model",
```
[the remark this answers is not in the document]
mversic: this was a leftover from when I tried to move `Level` into `data_model`, removed it now. IMO we don't want `iroha_logger` as `iroha_data_model` dependency

core/src/smartcontracts/wasm.rs
```diff
@@ -867,7 +927,7 @@ mod tests {
 
     #[test]
     fn execute_query_exported() -> Result<(), Error> {
-        let account_id = AccountId::from_str("alice@wonderland")?;
+        let account_id = AccountId::from_str("alice@wonderland").expect("Valid");
```
Erigara, line 946: Relevant comment from the other PR: [link](https://github.com/hyperledger/iroha/pull/3106#discussion_r1121635117).

core/src/smartcontracts/wasm.rs
Erigara: i think so

wasm/src/lib.rs
```diff
@@ -23,11 +23,13 @@ pub use iroha_data_model as data_model;
 pub use iroha_wasm_derive::entrypoint;
 pub use parity_scale_codec::{DecodeAll, Encode};
 
-pub mod debug;
+mod debug;
```
Erigara, line 26: why `debug` became private?

wasm/src/logging.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// Log `obj` with desired log level
+pub fn log<T: Encode + ?Sized>(log_level: LogLevel, obj: &T) {
+    #[cfg(not(test))]
+    use host::log as host_log;
+    #[cfg(test)]
+    use tests::_log_mock as host_log;
+
+    if log_level >= MAX_LOG_LEVEL.get() {
```
Erigara, line 81: Currently smart-contract won't change it'sloglevel if host change it's loglevel, however i don't think it's significant problem.

wasm/src/logging.rs
Erigara: imo such lazy evaluation is fine, why load anything if it won't be used

core/src/smartcontracts/wasm.rs
```diff
@@ -439,8 +447,61 @@ impl<'wrld> Runtime<'wrld> {
         Ok(value_offset)
     }
 
-    /// Host defined function which prints given string. When calling
-    /// this function, module serializes ISI to linear memory and
+    fn query_max_log_level(mut caller: Caller<State>) -> Result<WasmUsize, Trap> {
+        unimplemented!()
```
Erigara, line 451: why it's unimplemented?

mversic: sorry, I found this was missing and am working on it

wasm/src/lib.rs
mversic: I didn't notice. I can either make `debug` module private or remove `pub use debug::*` statement. The latter seems like a better option. What do you say?

Erigara: let's remove `pub use debug::*`

Erigara (dismissed):

config/src/logger.rs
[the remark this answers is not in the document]
Arjentix: Okay, got it

core/src/smartcontracts/wasm.rs
Arjentix: Yeah

wasm/src/logging.rs
appetrosyan: As a Haskellite I vote for lazy. Logging isn't critical. If you don't have logging you care about performance and would appreciate us not loading the infrastructure if you don't use it.

wasm/src/logging.rs
appetrosyan: If it's invalid, the worst that could happen is a random `u8`. 

Just saturate the log to the lowest level: `trace`.

wasm/src/log.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/// Query the max log level set on the host
+fn query_max_log_level() -> Level {
+    #[cfg(not(test))]
+    use host::query_max_log_level as host_query_max_log_level;
+    #[cfg(test)]
+    use tests::_query_max_log_level_mock as host_query_max_log_level;
+
+    // SAFETY: `host_query_max_log_level` must return a valid log level
```
appetrosyan, line 69: I would recommend saturating the log level rather than triggering UB.

wasm/src/log.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let bytes = obj.encode();
+        let ptr = bytes.as_ptr();
+        let len = bytes.len();
+
+        unsafe { host_log(log_level_id, ptr, len) }
+    }
+}
+
+/// Construct a new event
```
appetrosyan, line 94: I assume we're trying to copy what's done in `iroha_logger`. Can't we combine them?

appetrosyan (approved):
This PR

wasm/src/log.rs
mversic: that's true, but it can be implemented differently. And I don't think it's worth intersecting these crates

Erigara (approved):

appetrosyan (approved):

PR #3280, opened 2023-03-03
[fix] #2123: Return to `PublicKey` de/serialization from multihash

### LINKED ISSUE #2123: Use enum for `PublicKey`/`PrivateKey` digest function

Erigara:
### Description of the Change

Return `PublicKey` to de/serialization from multihash string.

### Issue

Closes #2123.

### Benefits

Reverse breaking change.

### Possible Drawbacks

None.

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -27,6 +27,7 @@ iroha_schema = { path = "../schema" }
 derive_more = { version = "0.99.17", default-features = false, features = ["deref", "deref_mut", "display"] }
 parity-scale-codec = { version = "3.2.1", default-features = false, features = ["derive", "full"] }
 serde = { version = "1.0.151", default-features = false, features = ["derive"] }
+serde_with = { version = "2.1.0", default-features = false, features = ["macros"] }
 hex = { version = "0.4.3", default-features = false, features = ["alloc", "serde"] }
 openssl-sys = { version = "0.9.80", features = ["vendored"], optional = true }
 ursa = { version = "0.3.7", optional = true }
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -30,8 +30,9 @@ use iroha_ffi::FfiType;
 use iroha_schema::IntoSchema;
 pub use merkle::MerkleTree;
 use multihash::{DigestFunction as MultihashDigestFunction, Multihash};
-use parity_scale_codec::{Decode, Encode, Error as ScaleError};
+use parity_scale_codec::{Decode, Encode};
 use serde::{Deserialize, Serialize};
+use serde_with::{DeserializeFromStr, SerializeDisplay};
 pub use signature::*;
 #[cfg(feature = "std")]
 pub use ursa;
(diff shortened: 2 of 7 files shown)


crypto/src/lib.rs
```diff
@@ -403,6 +328,25 @@ impl KeyPair {
     }
 }
 
+#[cfg(feature = "std")]
```
appetrosyan, line 331: This is an interesting approach. The most readable `impl<'de> Deserialize<'de>` which wasn't redundant. I wish we used this approach more.

appetrosyan (dismissed):

pesterev (approved):
LGTM

crypto/Cargo.toml
```diff
@@ -27,6 +27,7 @@ iroha_schema = { path = "../schema" }
 derive_more = { version = "0.99.17", default-features = false, features = ["deref", "deref_mut", "display"] }
 parity-scale-codec = { version = "3.2.1", default-features = false, features = ["derive", "full"] }
 serde = { version = "1.0.151", default-features = false, features = ["derive"] }
+serde_with = {version = "2.1.0", default-features = false, features = ["macros"]}
```
Arjentix, line 30:
```suggestion
serde_with = { version = "2.1.0", default-features = false, features = ["macros"] }
```

crypto/src/lib.rs
```diff
@@ -267,7 +192,7 @@ pub enum Error {
     #[display(fmt = "Signing failed. {_0}")]
     Signing(String),
     /// Returned when an error occurs during key generation
-    #[display(fmt = "Key generation failed. {_0} sss")]
```
Arjentix, line 270: Lol

crypto/src/lib.rs
```diff
@@ -355,7 +280,7 @@ impl KeyPair {
 
                 SignatureOf::new(key_pair.clone(), &dummy_payload)?
                     .verify(&dummy_payload)
-                    .map_err(|_err| Error::KeyGen(String::from("Key pair mismatch sss")))?;
```
Arjentix, line 358: Lol x2

crypto/src/lib.rs
```diff
@@ -403,6 +328,25 @@ impl KeyPair {
...
+        #[derive(Deserialize)]
+        struct KeyPair {
+            public_key: PublicKey,
+            private_key: PrivateKey,
+        }
+
+        let key_pair = KeyPair::deserialize(deserializer)?;
+        Self::new(key_pair.public_key, key_pair.private_key).map_err(D::Error::custom)
+    }
+}
```
Arjentix, line 348: I don't get it. Original `KeyPair` has the exact same fields. What does this custom impl does that can't be done with `Deserialize` derive?

Arjentix: To future readers. Roman answered in DM that it's done to use the `new()` method to check keys consistency.

In that case I suggest to note this in the code and/or comments. I.e. we can remove internal `KeyPair` to `KeyPairUnchecked`.

appetrosyan: I thought it was obvious from the signature. IMO KeyPairCandidate is a bit better, and a comment just above ::new to emphasise the consistency check would help.

crypto/Cargo.toml
appetrosyan: On that note, I feel we should consider adding a TOML auto-formatter to the CI as well as cargo fmt. In most cases the overhead would be minimal.

appetrosyan: #3282

crypto/src/lib.rs
```diff
@@ -71,101 +72,25 @@ impl std::error::Error for NoSuchAlgorithm {}
 
 ffi::ffi_item! {
     /// Algorithm for hashing
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, FfiType, IntoSchema)]
+    #[derive(Debug, Display, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, FfiType, IntoSchema, SerializeDisplay, DeserializeFromStr, Encode, Decode, Default)]
```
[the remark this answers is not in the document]
Erigara: Sure, i will change this

Arjentix (approved):

pesterev (approved):

Issue #3286, opened 2023-03-07
[refactor] Create `iroha_derive_primitives`


Arjentix:
Need to extract `params` and `parse_keywords!` from `iroha_wasm_derive` to a separate crate. That will be useful for future `iroha_wasm_validator_derive` crate

Issue #3226, opened 2023-02-21
Extract validators wasm API from `iroha_wasm`


Arjentix:
`iroha_wasm::validator` and `iroha_wasm_derive` validators derives become pretty huge. I suggest to move all validators-related stuff into another crate (i.e. `iroha_wasm_validator`) and import it from validators

PR #3287, opened 2023-03-07
Extract `iroha_wasm_validator` crate from `iroha_wasm`

Arjentix:
### Description of the Change

- Updated version of `iroha_wasm`
- Created `iroha_derive_primitives`
- Created `iroha_wasm_validator`

### Issue

Closes:

- #3226 
- #3285 
- #3286

### Benefits

Code which is useful only for *Validators* will not be compiled for other things like *Triggers* and *Smart contracts*

### Possible Drawbacks

Can't see any


derive_primitives/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "iroha_derive_primitives"
+version = "2.0.0-pre-rc.13"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+syn = {version = "1", default-features = false, features = ["full", "extra-traits", "derive"] }
+quote = "1.0"
+proc-macro2 = "1.0.50"
```
mversic, line 11: please make sure you use uniform versions

wasm/derive/Cargo.toml
```diff
@@ -9,7 +9,8 @@ edition = "2021"
 proc-macro = true
 
 [dependencies]
-syn = {version = "1", default-features = false, features = ["full", "extra-traits", "derive"] }
+iroha_derive_primitives = { version = "2.0.0-pre-rc.13", path = "../../derive_primitives" }
+syn = { version = "1", default-features = false, features = ["full", "extra-traits", "derive"] }
 quote = "1.0"
 proc-macro2 = "1.0.50"
```
mversic, line 15: here as well

mversic:
what I would prefer is to make hierarchy such that all `wasm` related crates are part of the same workspace. I don't like the workspace split, but it's necessary (i don't remember but probably because nightly is required). You can split code into as many crates as you see fit

wasm/src/lib.rs
```diff
@@ -24,26 +23,47 @@ pub use iroha_wasm_derive::entrypoint;
...
-    // raise a `trap` in the `WebAssembly` if execution of said WASM panics.
-    unreachable!("Program should have aborted")
-}
-
-#[no_mangle]
-#[cfg(not(test))]
-#[alloc_error_handler]
-fn oom(layout: ::core::alloc::Layout) -> ! {
-    panic!("Allocation({} bytes) failed", layout.size())
+/// Initialization macro for programs using `iroha_wasm`.
```
QuentinI, line 30: We should probably mention in crate-level docs that it's advisable to call this.

QuentinI (commented):
Since at this point `wasm` in validators is just an implementation detail, as we don't have validators of any other sort, I'd say we can drop `wasm` prefix for brevity

wasm_validator/derive/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Validators are only checking if an operation is **invalid**, not if it is valid.
+/// A validator can either deny the operation or pass it to the next validator if there is one.
+///
+/// # Attributes
+///
+/// This macro must have an attribute describing entrypoint parameters.
+///
+/// The syntax is:
+/// `#[iroha_wasm::validator_entrypoint(params = "[<type>,*]")]`, where `<type>` is one of:
+/// - `authority` is a signer account id who submits an operation
```
0x009922, line 25:
```suggestion
/// The syntax is:
/// 
/// ```ignore
/// #[iroha_wasm::validator_entrypoint(params = "[<type>,*]")]
/// ```
///
/// where `<type>` is one of:
///
/// - `authority` is a signer account id who submits an operation
```

wasm_validator/derive/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+/// - `expression` is an expression that is being validated
+///
+/// Exactly one parameter of *operation to validate* kind must be specified.
+/// `authority` is optional.
+/// Parameters will be passed to the entrypoint function in the order they are specified.
+///
+/// ## Authority
+///
+/// A real function parameter type corresponding to the `authority` should have
+/// `iroha_wasm::data_model::prelude::AccountId` type.
```
0x009922, line 38:
```suggestion
/// [`iroha_wasm::data_model::prelude::AccountId`] type.
```

I think here and further it makes sense to leverage [Rustdoc linking feature](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html), so it will be able to catch outdated/wrong links if they occur.

Arjentix: I'd like to, but I can't. `iroha_wasm_validator_derive` does not depend on `iroha_wasm` or `iroha_wasm_validator`

Arjentix:
> Since at this point wasm in validators is just an implementation detail, as we don't have validators of any other sort, I'd say we can drop wasm prefix for brevity

@QuentinI , I understand your idea, but I think, I can move `wasm_valdiator` into `wasm/validator` as @mversic suggested [here](https://github.com/hyperledger/iroha/pull/3287#issuecomment-1459859963). With that I remove redundant `wasm` in folder name, but I think it make sense to leave `iroha_wasm_validator` as a crate name because of this folder structure.

appetrosyan:
>  but I think it makes sense to leave iroha_wasm_validator as a crate name because of this folder structure.

Personally I think that the folder structure and namespacing should exist separately. I'm with @QuentinI  on this one.

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -40,7 +47,43 @@ jobs:
         if: always()
         run: cargo doc --no-deps --quiet
 
-  tests:
+  wasm-validator-static-analysis:
```
mversic, line 50: do we have to have two steps for wasm static analysis?

wasm/validator/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+# Enables debugging tools such as `dbg()` and `DebugUnwrapExt`
+debug = ["iroha_wasm/debug"]
```
mversic, line 13: seems unnecessary?

wasm/validator/derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "iroha_validator_derive"
+
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+
+license.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
mversic, line 10: please remove this line in a few places

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -12,6 +12,10 @@ on:
       - 'wasm/**.json'
       - 'wasm/**.toml'
       - 'wasm/**.yml'
+      - 'wasm/validator/**.rs'
+      - 'wasm/validator/**.json'
+      - 'wasm/validator/**.toml'
+      - 'wasm/validator/**.yml'
```
mversic, line 18: don't previous `**` capture all this already?

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -23,8 +27,11 @@ env:
   RUSTUP_TOOLCHAIN: nightly-2022-12-22
 
 jobs:
-  static-analysis:
+  wasm-static-analysis:
```
mversic, line 30:
```suggestion
 static-analysis:
```

it's obvious from the workflow name

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -23,8 +27,11 @@ env:
...
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: wasm
```
mversic, line 34: this is already set at the top

client/tests/integration/smartcontracts/Cargo.toml
```diff
@@ -25,3 +25,4 @@ codegen-units = 1   # Further reduces binary size but increases compilation time
 
 [workspace.dependencies]
 iroha_wasm = { path = "../../../../wasm", features = ["debug"]}
+iroha_validator = { path = "../../../../wasm/validator", features = ["debug"]}
```
mversic, line 28:
```suggestion
iroha_validator = { path = "../../../../wasm/validator"}
```

it's already set by `iroha_wasm`

wasm/validator/Cargo.toml
mversic: it doesn't do anything except active feature of another crate. User should just activate that feature directly on `iroha_wasm` crate

wasm/validator/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+iroha_wasm = { version = "2.0.0-pre-rc.13", path = ".." }
+iroha_validator_derive = { path = "derive" }
```
mversic, line 17: missing version

wasm/validator/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
```
mversic, line 1: there should be only one `.cargo` at workspace level

wasm/validator/derive/src/conversion/asset_definition.rs
```diff
@@ -11,7 +11,7 @@ pub fn impl_derive_ref_into_asset_definition_owner(input: TokenStream) -> TokenS
     impl_from(
         &input.ident,
         &input.generics,
-        &syn::parse_quote!(::iroha_wasm::validator::pass_conditions::asset_definition::Owner),
+        &syn::parse_quote!(::iroha_validator::pass_conditions::asset_definition::Owner),
         &syn::parse_quote!(asset_definition_id),
     )
     .into()
```
mversic, line 17: IMO `mod.rs`, `account.rs`, `asset.rs` and `asset_definition.rs` are too small to be in separate modules. I would just place all this code in `conversion.rs` and delete those

.github/workflows/iroha2-dev-pr-wasm.yaml
Arjentix: Indeed

.github/workflows/iroha2-dev-pr-wasm.yaml
Arjentix: I belive, that no, thanks to your suggestion about workspace

wasm/validator/Cargo.toml
Arjentix: > it doesn't do anything except active feature of another crate. User should just activate that feature directly on iroha_wasm crate

How if user will only link to `iroha_validator`?

Arjentix: I see, that it worked for `client/tests/integration/smartcontracts`. But that just becasue it has `iroha_wasm` as a direct depenency as well

mversic:
while you're at it maybe you would also move `wasm_builder` -> `wasm/builder`?

wasm/validator/Cargo.toml
mversic: > I see, that it worked for `client/tests/integration/smartcontracts`. But that just becasue it has `iroha_wasm` as a direct depenency as well

my suggestion was that whenever user needs `debug` feature they would bring the `iroha_wasm` as explicit dependency considering that it's always there as implicit dependency. I'm not sure if this is preferable or not, maybe someone else would care to comment

mversic (approved):

wasm/validator/Cargo.toml
mversic: had these two crates `iroha_wasm` and `wasm_validator` been truly independent I wouldn't mention this

Arjentix:
> while you're at it maybe you would also move wasm_builder -> wasm/builder?

This won't work, I'm afraid. `iroha_wasm_builder` should be used in build-scripts and should have no contraints to be `no-std` compatible or being able to compile for `wasm32-unknown-unknown` as `wasm` workspace does

wasm/validator/Cargo.toml
Arjentix: I'm not sure it's a good thing, but yes, we can wait for someone else

mversic (dismissed):

mversic (approved):

permission_validators/Cargo.toml
```diff
@@ -47,10 +47,10 @@ members = [
 ]
 
 [workspace.dependencies]
-iroha_wasm = { path = "../wasm", features = ["debug"]}
+iroha_validator = { path = "../wasm/validator", features = ["debug"]}
```
mversic, line 50: missing version

wasm/src/lib.rs
```diff
@@ -23,16 +26,44 @@ use parity_scale_codec::{DecodeAll, Encode};
...
-    loop {}
+/// Initialization macro for programs using `iroha_wasm`.
+/// Requires `alloc_error_handler` feature to be enabled.
+/// Should be called once at global scope.
+///
+/// # Example
+///
+/// ```no_run
+/// #![no_std]
+/// #![feature(alloc_error_handler)]
```
mversic, line 41: remove this

wasm/src/lib.rs
```diff
@@ -23,16 +26,44 @@ use parity_scale_codec::{DecodeAll, Encode};
...
+            loop {}
+        }
+
+        #[no_mangle]
+        #[cfg(not(test))]
+        #[alloc_error_handler]
+        fn oom(layout: ::core::alloc::Layout) -> ! {
+            panic!("Allocation({} bytes) failed", layout.size())
+        }
+    };
```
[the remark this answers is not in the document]
mversic: considering `init` is now just a panic handler, I would consider, in no particular order: 
1. removing it and implementing it manually in every crate that needs it
2. renaming it to panic_handler or something
4. removing it and using [this](https://github.com/korken89/panic-halt/blob/master/src/lib.rs) instead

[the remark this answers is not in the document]
Arjentix: Yes, 3'rd option seems good for me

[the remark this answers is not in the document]
Arjentix: But it requires to use it like:

```rust
#[cfg(not(test))]
extern crate panic_halt;
```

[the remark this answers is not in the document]
mversic: ofc, however you add panic_handler it will have to be guarded from being included into tests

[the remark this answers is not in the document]
Arjentix: Ok, I'm fixing it

mversic (approved):

wasm/validator/Cargo.toml
mversic: @appetrosyan @Erigara @QuentinI thoughts?

QuentinI: I'm in favor of leaving it be. Doesn't cost us anything and gives more intuitive UX.

Issue #3232, opened 2023-02-21
Workspace package inheritance


mversic:
Since [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds) it's possible to use `workspace.package` for a crate to inherit package info from the workspace.

Issue #3289, opened 2023-03-08
Workspace dependencies inheritance


mversic:
Since [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds) it's possible to use `workspace.dependencies` for a crate to inherit dependencies from the workspace.

Related to #3232

PR #3290, opened 2023-03-08
[reafctor] #3232: Share workspace metadata

### LINKED ISSUE #3285: [bug] Update `iroha_wasm` version to `2.0.0-pre-rc.13`

mversic:
### Issue

Closes #3285 #3232


client/tests/integration/smartcontracts/mint_rose/Cargo.toml
```diff
@@ -1,10 +1,9 @@
 [package]
 name = "mint_rose"
+
+edition = "2021"
 version = "2.0.0-pre-rc.13"
```
pesterev, line 5: This is left here because we can't share workspace metadata in integration test crates, right?

wasm/Cargo.toml
```diff
@@ -1,10 +1,11 @@
...
 name = "iroha_wasm"
-version = "0.1.0"
-authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
+
 edition = "2021"
+version = "2.0.0-pre-rc.13"
+authors = ["Iroha 2 team <https://github.com/orgs/soramitsu/teams/iroha2>"]
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+license = "Apache-2.0"
```
mversic, line 8: I'm not sure whether to use `license-file = ../LICENSE` here instead because this crate doesn't include the license and is not part of the workspace

client/tests/integration/smartcontracts/mint_rose/Cargo.toml
mversic: As you can see below it's not part of the workspace. But I will try to make a virtual manifest for these `smartcontract` tests and share from there

mversic: I applied this

wasm/Cargo.toml
appetrosyan: I prefer licence files. This is standard such as README.

ffi/src/handle.rs
```diff
@@ -66,6 +66,7 @@ macro_rules! def_ffi_fn {
         /// All of the given pointers must be valid and the given handle id must match the expected
         /// pointer type
         #[no_mangle]
+        #[export_name = concat!(module_path!(), "__clone")]
```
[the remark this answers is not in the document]
mversic: pardon me, I'm still fixing this. You're observation is correct

permission_validators/Cargo.toml
```diff
@@ -1,5 +1,37 @@
...
+name = "iroha_main_validator"
+
+edition.workspace = true
+version.workspace = true
+authors.workspace = true
+
+license.workspace = true
+
+[lib]
+# A smartcontract should be linked dynamically so that it may link to functions exported
```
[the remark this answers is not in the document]
mversic: Then I'll just remove the comment all together. It's obvious that it's linked dynamically

core/Cargo.toml
```diff
@@ -60,20 +63,15 @@ crossbeam-queue = "0.3.8"
...
 parking_lot = { version = "0.12.1", features = ["deadlock_detection"] }
-
-# transitive dependencies
-anyhow = "1.0.68"
-derive_more = "0.99.17"
-getset = "0.1.2"
 itertools = "0.10.5"
-
-tempfile = "3.3.0"
+derive_more = "0.99.17"
```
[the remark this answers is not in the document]
mversic: why this? I will update other files as well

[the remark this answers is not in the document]
mversic: I'd leave this suggestion for #3289

cli/Cargo.toml
```diff
@@ -1,15 +1,18 @@
...
+authors.workspace = true
+
+description.workspace = true
+repository.workspace = true
+homepage.workspace = true
+documentation.workspace = true
+
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
```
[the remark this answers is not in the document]
mversic: so I've added these for `iroha_core`, `iroha_cli` and `iroha_client`. For other crates I was a bit conservative and left some things out as this info didn't quite seem to be a fit for those crates

wasm/Cargo.toml
mversic: are you saying to keep this `license = "Apache-2.0"` and duplicate `LICENSE` file in the `wasm` crate. I think that's the best

client/tests/integration/smartcontracts/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+    "create_nft_for_every_user_trigger",
+    "deny_new_validators_validator",
+    "mint_rose",
+]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+strip = "debuginfo" # Remove debugging info from the binary
```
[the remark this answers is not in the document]
mversic: #3293

wasm/Cargo.toml
mversic: the choice here is:
1. use `license-file = ../LICENSE` in every workspace except root
2. use `license = "Apache-2.0"` and duplicate `LICENSE` file in every workspace we have (root, wasm, permission_validators)

pesterev (commented):
looks fine to me, but there are empty lines in some toml files
like
```toml
+ authors.workspace = true
+
+ license.workspace = true
```

ffi/src/handle.rs
```diff
@@ -58,15 +58,16 @@ macro_rules! def_ffi_fn {
...
         /// FFI function equivalent of [`Clone::clone`]
         ///
         /// # Safety
         ///
         /// All of the given pointers must be valid and the given handle id must match the expected
         /// pointer type
         #[no_mangle]
-        $vis unsafe extern "C" fn __clone(
+        $(#[export_name = concat!($prefix, "__clone")])?
+        unsafe extern "C" fn __clone(
```
mversic, line 70: current behavior is such that one `__clone` is generated per crate because of orphan rule and to avoid the risk of collision of `Handle::ID`s across crates. Since there are multiple versions of `__clone` they are disambiguated by adding a prefix. Currently handle id is unique per crate

It would be possible to generate only one `__clone` across all crates but that approach has the risk of two structs from different crates sharing the same `Handle::ID` by accident resulting in UB because of invalid void pointer cast. In that case we would have to make sure that id is globally unique across all crates

data_model/src/lib.rs
```diff
@@ -1591,6 +1591,17 @@ pub mod ffi {
...
+    #[cfg(any(crate_type = "dylib", crate_type = "cdylib"))]
+    #[cfg(all(feature = "ffi_export", not(feature = "ffi_import")))]
+    mod dylib {
+        #[cfg(not(feature = "std"))]
+        use alloc::alloc;
+        #[cfg(feature = "std")]
+        use std::alloc;
+
+        iroha_ffi::def_ffi_fn! {dealloc}
+    }
```
mversic, line 1603: unlike other extern functions, there should be only one `__dealloc` function per generated dynamic library

appetrosyan: And there's no way of enforcing this invariant at compile time. I doubt the macro trickery is worth it, I'd just add a comment in the text of the program to that effect.

ffi/src/handle.rs
appetrosyan: 1. I'd like this in the text of the program, preferably as a doc-string. 
2. I'd like a couple of unit tests that show us how to break this invariant. 
3. Hard-coding the export name is **much** better because it decouples the module path from the actual implementation. With that said, I'd rather we had a more stable system.

appetrosyan (dismissed):
LGTM, two minor comments

ffi/src/handle.rs
mversic: > 1. I'd like this in the text of the program, preferably as a doc-string. 

I agree

> 2. I'd like a couple of unit tests that show us how to break this invariant.

Maybe I can make a test for this. This would be testing that linker refuses to link two crates that have exported a function with the same name.

> 3. Hard-coding the export name is **much** better because it decouples the module path from the actual implementation. With that said, I'd rather we had a more stable system. 

What would a "more stable" system be?

data_model/src/lib.rs
mversic: I believe this can only be enforced at link time. I will add a comment

This function could be defined once per crate, but there is no need as it would be the same for every crate

ffi/src/handle.rs
appetrosyan: Sorry, I was confused. The doc should specifically contrast the brittle approach (one global `__clone`) with the current implementation per-crate `__clone`

appetrosyan (approved):

appetrosyan (approved):

Issue #3234, opened 2023-02-21
Use `proc_macro_error` in `iroha_wasm_derive`


Arjentix:

mversic:
@DCNick3 please handle this while transitioning to syn 2.0

mversic:
this is not relevant anymore

PR #3299, opened 2023-03-10
[feature] #3260: Wasm type inference

### LINKED ISSUE #3260: Type inference for triggers and validators
Originally suggested by @mversic [here](https://github.com/hyperledger/iroha/pull/3106#discussion_r1115262135).

We can infere the type of input parameters by looking at function parameter type names. It won't work in 100% cases because user can name his type the same way or do something else. So we need to keep the old system as well, but make it optional.

## Example

Old:

```rust
#[entrypoint(params = "[query]")]
pub fn validate(_: QueryBox) -> Verdict {
    Verdict::Deny("No queries are allowed".to_owned())
}
```

New:

```rust
#[entrypoint]
pub fn validate(_: QueryBox) -> Verdict {
    Verdict::Deny("No queries are allowed".to_owned())
}
```

mversic:
### Issue

Closes #3260 #3234


wasm/derive/src/entrypoint.rs
```diff
@@ -74,6 +57,29 @@ impl super::params::ConstructArg for ParamType {
...
+        };
+
+        args.push(if is_authority(ty) {
+            ParamType::Authority
+        } else if is_triggering_event(ty) {
+            ParamType::TriggeringEvent
+        } else {
+            abort!(
+                input,
+                "Unable to infer function parameter type. Use #[entrypoint(params = \"[...]\")]"
```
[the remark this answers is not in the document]
mversic: No, considering your suggestion with trait approach

wasm/derive/src/validator/entrypoint.rs
```diff
@@ -84,6 +78,35 @@ impl crate::params::ConstructArg for ParamType {
...
+            ParamType::Expression
+        } else {
+            abort!(
+                input,
+                "Unable to infer function parameter type. Use #[entrypoint(params = \"[...]\")]"
+            )
+        });
+    }
+
+    args.into_iter().map(|arg| arg.construct_arg()).collect()
```
pesterev, line 107: It seems we can avoid a vector allocation here.
```suggestion
    sig.inputs
        .iter()
        .map(|input| {
            let syn::FnArg::Typed(syn::PatType {ty, ..}) = input else {
                panic!("Receiver types are not allowed")
            };

            let param_type = if is_authority(ty) {
                ParamType::Authority
            } else if is_transaction(ty) {
                ParamType::Transaction
            } else if is_instruction(ty) {
                ParamType::Instruction
            } else if is_query(ty) {
                ParamType::Query
            } else if is_expression(ty) {
                ParamType::Expression
            } else {
                abort!(
                input,
                "Unable to infer function parameter type. Use #[entrypoint(params = \"[...]\")]"
            )
            };
            param_type.construct_arg()
        })
        .collect()
```

wasm/derive/src/entrypoint.rs
```diff
@@ -74,6 +57,29 @@ impl super::params::ConstructArg for ParamType {
     }
 }
 
+fn infer_args(sign: &syn::Signature) -> Punctuated<syn::Expr, syn::token::Comma> {
+    let mut args = vec![];
```
pesterev, line 61: the same point as [here](https://github.com/hyperledger/iroha/pull/3299#discussion_r1132646776)

wasm/derive/Cargo.toml
```diff
@@ -12,6 +12,7 @@ proc-macro = true
 
 [dependencies]
 syn = {version = "1", default-features = false, features = ["full", "extra-traits", "derive"] }
-quote = "1.0.23"
+proc-macro-error = "1.0.4"
```
Arjentix, line 15: Hm, this probably closes #3234

wasm/derive/src/entrypoint.rs
```diff
@@ -3,96 +3,31 @@
...
         vis,
         sig,
         mut block,
     } = parse_macro_input!(item);
 
-    assert!(
-        syn::ReturnType::Default == sig.output,
-        "Exported function must not have a return type"
-    );
+    if syn::ReturnType::Default != sig.output {
```
Arjentix, line 19: Yoda notation?

wasm/derive/src/entrypoint.rs
```diff
@@ -3,96 +3,31 @@
...
-        Attr::Params(params_attr) => params_attr.construct_args(),
-        Attr::Empty => Punctuated::new(),
-    };
     let fn_name = &sig.ident;
+    let args = sig.inputs.iter().map(|arg| {
+        let syn::FnArg::Typed(syn::PatType {ty, ..}) = arg else {
+            abort!(arg, "Receiver types are not supported");
+        };
+
+        quote! { <#ty as iroha_wasm::QueryHost>::query() }
```
Arjentix, line 29: Interesting solution, I like it

wasm/src/lib.rs
```diff
@@ -64,6 +62,74 @@ unsafe extern "C" fn _iroha_wasm_dealloc(offset: *mut u8, len: usize) {
     let _box = Box::from_raw(core::slice::from_raw_parts_mut(offset, len));
 }
 
+/// Marks a type whose values can be queried from the host environment and passed
+/// as arguments to `entrypoint` of a smart contract
+pub trait QueryHost {
```
Arjentix, line 67:
```suggestion
pub trait QueryHost: sealed::Sealed {
```

I think we should prevent users to write impls for this trait for their types

Arjentix (commented):
More deletions than insertions and clear entrypoint macros usage, cool!

permission_validators/src/lib.rs
```diff
@@ -8,9 +8,7 @@ extern crate alloc;
 use iroha_wasm::data_model::prelude::*;
 
 /// Validate any [`Instruction`](iroha_wasm::data_model::isi::Instruction).
-//
-// TODO: Not exhaustive list. Add more validators here.
```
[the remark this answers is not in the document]
appetrosyan: As long as there's a tracking ticket, removing the `TODO` is fine.

wasm/derive/src/entrypoint.rs
mversic: let me speak like yoda

wasm/derive/src/entrypoint.rs
mversic: credits go to @Erigara

wasm/src/lib.rs
```diff
@@ -53,8 +53,111 @@ unsafe extern "C" fn _iroha_wasm_dealloc(offset: *mut u8, len: usize) {
...
+        #[cfg(test)]
+        use tests::_iroha_wasm_query_authority_mock as host_query_authority;
+
+        // Safety: ownership of the returned result is transferred into `_decode_from_raw`
+        unsafe { decode_with_length_prefix_from_raw(host_query_authority()) }
+    }
+}
+
+impl sealed::Sealed for Event {}
+impl QueryHost for Event {
```
mversic, line 81: @Arjentix I think this makes sense?

wasm/src/lib.rs
```diff
@@ -53,8 +53,111 @@ unsafe extern "C" fn _iroha_wasm_dealloc(offset: *mut u8, len: usize) {
...
+        #[cfg(test)]
+        use tests::_iroha_wasm_query_triggering_event_mock as host_query_triggering_event;
+
+        // Safety: ownership of the result is transferred into `_decode_from_raw`
+        unsafe { decode_with_length_prefix_from_raw(host_query_triggering_event()) }
+    }
+}
+
+impl sealed::Sealed for NeedsPermissionBox {}
+impl QueryHost for NeedsPermissionBox {
```
mversic, line 94: @Arjentix Does this make sense?

permission_validators/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > As long as there's a tracking ticket, removing the TODO is fine.

Ok, no problem then

wasm/src/lib.rs
```diff
@@ -53,8 +53,111 @@ unsafe extern "C" fn _iroha_wasm_dealloc(offset: *mut u8, len: usize) {
...
+}
+
+/// Marks a type whose values can be queried from the host environment and passed
+/// as arguments to `entrypoint` of a smart contract
+pub trait QueryHost: sealed::Sealed {
+    /// Return requested [`Self`] object from the host
+    fn query() -> Self;
+}
+
+impl sealed::Sealed for AccountId {}
```
Arjentix, line 67: I like keeping such impls inside `sealed` mod, but that's just my preference

wasm/src/lib.rs
Arjentix: Yes, usefull for triggers. And will panic if we try to get this from validator, I believe

wasm/src/lib.rs
Arjentix: Interesting moment... Well, let's keep it for now

Arjentix: Shouldn't we implement this for every sub-type of `Event` as well?

Arjentix (approved):

pesterev (dismissed):

wasm/src/lib.rs
appetrosyan: Probably

wasm/derive/src/entrypoint.rs
appetrosyan: Jokes aside, this is useful in C-like languages.

appetrosyan (dismissed):

mversic:
This was not a good approach. Also things around validators are changing rapidly. Closing at the moment

Issue #3132, opened 2023-02-09
[bug]: `NewParamerter` isi is ignored in validators


Erigara:
We are currently ignoring `NewParameter` instruction inside `ChangeParametersOnlyForSignerAccount` and `GrantedChangeConfigParameters` validators. 
But since parameters currently doesn't affect anything + we are moving towards runtime validators, this is not something we want to fix right now imo.
I will create ticket to keep track of this issue.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/3131#discussion_r1101196373_

My suggestion is to fix this when we transit to runtime permission validators.

PR #3300, opened 2023-03-12
[fix] #3132: Add `NewParameter` validator

Arjentix:
### Description of the Change

- Added new permission validator for `NewParameter` isi
- This validator added to the main permission validator
- Increased max gas limit

### Issue

Closes

- #3132 

### Benefits

Validator for a new instruction

### Possible Drawbacks

None

### Usage Examples or Tests

We don't write new tests until we have new test system with Python


Erigara (approved):

permission_validators/parameter/new/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+extern crate alloc;
+
+use iroha_wasm::{
+    data_model::prelude::*,
+    validator::{pass_conditions, prelude::*},
+};
+
+/// Strongly-typed representation of `can_grant_permission_to_create_parameters` permission token.
+#[derive(Token, Validate, Clone, Copy)]
+#[validate(pass_conditions::AlwaysDeny)] // Can be granted only in genesis
```
appetrosyan, line 16:
```suggestion
#[validate(pass_conditions::OnlyGenesis)]
```

tools/kagami/src/main.rs
```diff
@@ -692,7 +692,7 @@ mod tokens {
     #[derive(StructOpt, Debug, Clone, Copy)]
     pub struct Args;
 
-    pub fn permission_token_definitions() -> Result<[PermissionTokenDefinition; 17]> {
+    pub fn permission_token_definitions() -> Result<[PermissionTokenDefinition; 20]> {
```
appetrosyan, line 695: I wonder if Rust has means of inferring the size. `const` generic maybe?

appetrosyan (dismissed):

tools/kagami/src/main.rs
Erigara: I'm not sure that const generic will work here because the size it's not something we pass inside function.

permission_validators/parameter/new/src/lib.rs
Arjentix: Will fix in my another openned PR: #3226

tools/kagami/src/main.rs
Arjentix: Looks like there is a nightly feature: [rust-lang/rust/#85077](https://github.com/rust-lang/rust/issues/85077)

Erigara: IMO we can just switch to `Vec` since this place isn't performance critical at all.

Arjentix: And I can't make it const or static, because of `token::Definition::with_params()`

Arjentix: Fixed using `Vec`

PR #3301, opened 2023-03-13
[refactor] #3254: Refactor `commit_block()` and `replace_top_block()`

### LINKED ISSUE #3254: Refactor `commit_block()` and `replace_top_block()`
This two functions repeats pretty the same code. And it also contained a bug.

So we need to extract their common functuonallity into another function. Or maybe remove one of this functions at all.

Originally suggested by @mversic [here](https://github.com/hyperledger/iroha/pull/3106#discussion_r1114184943).

Erigara:
### Description of the Change

Add `update_topology` and `update_state` function to encapsulate common functionality. 

### Issue

Resolves #3254.

### Benefits

DRY.

### Possible Drawbacks

None.

### Alternate Designs

I have tried const generic approach but it doesn't look very nice.

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -385,55 +385,20 @@ fn commit_block<F: FaultInjection>(
     block: impl Into<VersionedCommittedBlock>,
 ) {
     let committed_block = block.into();
-    let block_hash = committed_block.hash();
 
     state.finalized_wsv = state.wsv.clone();
-    state
-        .wsv
-        .apply(&committed_block)
-        .expect("Failed to apply block on WSV. Bailing.");
-    sumeragi.send_events(state.wsv.events_buffer.replace(Vec::new()));
-
-    // Update WSV copy that is public facing
-    *sumeragi.wsv.lock() = state.wsv.clone();
-
-    // This sends "Block committed" event, so it should be done
-    // AFTER public facing WSV update
-    sumeragi.send_events(&committed_block);
-
+    update_state(state, sumeragi, &committed_block);
     state.previous_block_hash = state.latest_block_hash;
-    state.latest_block_height = committed_block.header().height;
-    state.latest_block_hash = Some(committed_block.hash());
-    state.latest_block_view_change_index = committed_block.header().view_change_index;
-
-    let current_topology = &mut state.current_topology;
-    let role = current_topology.role(&sumeragi.peer_id);
 
     info!(
-        addr=%sumeragi.peer_id.address, %role,
+        addr=%sumeragi.peer_id.address,
+        role=%state.current_topology.role(&sumeragi.peer_id),
         block_height=%state.latest_block_height,
-        %block_hash, "Committing block"
+        block_hash=%committed_block.hash(),
+        "Committing block"
     );
 
-    *current_topology = Topology {
-        sorted_peers: committed_block.header().committed_with_topology.clone(),
-    };
-    current_topology.lift_up_peers(
-        &committed_block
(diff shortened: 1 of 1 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -448,60 +413,71 @@ fn replace_top_block<F: FaultInjection>(
...
             .peers_ids()
             .iter()
             .map(|id| id.clone())
             .collect::<Vec<PeerId>>(),
     );
+    state.current_topology = topology;
+}
 
-    sumeragi.kura.replace_top_block(committed_block);
+fn update_state<F: FaultInjection>(
```
appetrosyan, line 459: This refactor doesn't seem to do much other than extract the logic into a single function. The function isn't particularly reusable either. I'm not against merging it, but I'd think that a dedicated refactor PR should do a little more, what do you think?

Erigara: @appetrosyan what direction should be taken in your opinion?
Intention was to extract reused functionality to separate functions to avoid duplication and divergence in behavior.

I tried to use `const generic bool` value but than code is polluted with `if` statements which doesn't look very nice to me.

appetrosyan: > I tried to use const generic bool value but than code is polluted with if statements which doesn't look very nice to me.

Those aren't particularly nice looking in rust. In C++ it's a different situation. 

> Intention was to extract reused functionality (state update and topology update) to separate functions to avoid duplication and divergence in behavior.

That's a good endeavour, and we should merge it, my question is whether we could expand the scope of this refactoring to more aspects of this same file? Compress it somewhat?

Erigara: Yeah, i absolutely agree that some tests for sumeragi is would be beneficial. Do you mind separate ticket for this?

appetrosyan: #3302

appetrosyan (dismissed):

appetrosyan (approved):

core/src/sumeragi/main_loop.rs
```diff
@@ -448,60 +413,71 @@ fn replace_top_block<F: FaultInjection>(
...
 
-    *current_topology = Topology {
+    update_topology(state, &committed_block);
+
+    sumeragi.kura.replace_top_block(committed_block);
+
+    cache_transaction(state, sumeragi)
+}
+
+fn update_topology(state: &mut State, committed_block: &VersionedCommittedBlock) {
```
pesterev, line 436: Seems it should be a method of `State`, but I suppose you have a strong reason to keep it as a separate function, right?

pesterev (commented):
LGTM

core/src/sumeragi/main_loop.rs
appetrosyan: I agree that this is a mutator method and it **should** be associated to state (i.e. written in an `impl` block). 

> other functions that also looks like "method" functions, but some reason we keep them as separate functions.

This is mostly stylistic, but I can give you a rational reason why that's a good idea. Functions associated to types are in a privileged position with respect to access modifiers. That's why you need to be careful about the visibility of these functions. Everything is a method is an OOP concept, and not necessarily a useful one. It can help readability, and it definitely would in this particular case, but sometimes it's not very helpful. 

IMO, given that we're refactoring the file, we should try and look at other potentially associated `fn`s and try to add them to the respective `impl` blocks.

appetrosyan (approved):

pesterev (approved):

Issue #3249, opened 2023-02-22
Block hash should include signatures.


SamHSmith:

SamHSmith:
Right now the signatures are not included in the block hash because the consensus needs to be able to refer to the block during the signing process. This is a security problem since the validity of the blockchain is rooted in the signatures and as we move closer to the whitepaper they also affect how the topology evolves. Therefore the hash should be split into a consensus used hash that refers to the contents of the block without signatures and the default hash should include signatures.

PR #3306, opened 2023-03-14
[fix] #3249: Split block hashes into partial and complete versions.

SamHSmith:
### Description of the Change

Right now the signatures are not included in the block hash because the consensus needs to be able to refer to the block during the signing process. This is a security problem since the validity of the blockchain is rooted in the signatures and as we move closer to the whitepaper they also affect how the topology evolves. Therefore the hash should be split into a consensus used hash that refers to the contents of the block without signatures and the default hash should include signatures.

### DIFF
--- a/crypto/src/signature.rs
+++ b/crypto/src/signature.rs
@@ -103,6 +103,16 @@ impl Signature {
 
         Ok(())
     }
+
+    /// Get the payload of the public key in this signature.
+    pub fn key_payload(&self) -> &[u8] {
+        self.public_key.payload()
+    }
+
+    /// Get the encrypted payload of this signature.
+    pub fn signature_payload(&self) -> &[u8] {
+        &self.payload
+    }
 }
 
 // TODO: Enable in ffi_import
--- a/data_model/src/block.rs
+++ b/data_model/src/block.rs
@@ -12,14 +12,14 @@ use derive_more::Display;
 use getset::Getters;
 #[cfg(feature = "std")]
 use iroha_crypto::SignatureOf;
-use iroha_crypto::{HashOf, MerkleTree, SignaturesOf};
+use iroha_crypto::{Hash, HashOf, MerkleTree, SignaturesOf};
 use iroha_schema::IntoSchema;
 use iroha_version::{declare_versioned_with_scale, version_with_scale};
 use parity_scale_codec::{Decode, Encode};
 use serde::{Deserialize, Serialize};
 
 pub use self::{
-    committed::{CommittedBlock, VersionedCommittedBlock},
+    committed::{CommittedBlock, PartialBlockHash, VersionedCommittedBlock},
     header::BlockHeader,
 };
 use crate::{events::prelude::*, model, peer, transaction::prelude::*};
@@ -83,6 +83,27 @@ mod header {
         pub const fn is_genesis(&self) -> bool {
             self.height == 1
         }
+        /// Serialize the header's data for hashing purposes.
+        pub fn payload(&self) -> Vec<u8> {
+            let mut data = Vec::new();
+            data.extend(&self.timestamp.to_le_bytes());
+            data.extend(&self.consensus_estimation.to_le_bytes());
+            data.extend(&self.height.to_le_bytes());
+            data.extend(&self.view_change_index.to_le_bytes());
+            if let Some(hash) = self.previous_block_hash.as_ref() {
+                data.extend(hash.as_ref());
+            }
+            if let Some(hash) = self.transactions_hash.as_ref() {
+                data.extend(hash.as_ref());
+            }
+            if let Some(hash) = self.rejected_transactions_hash.as_ref() {
+                data.extend(hash.as_ref());
+            }
+            for id in &self.committed_with_topology {
+                data.extend(id.payload());
+            }
+            data
+        }
     }
 
     impl PartialOrd for BlockHeader {
@@ -112,6 +133,33 @@ mod committed {
     #[model]
     pub mod model {
         use super::*;
+
+        /// The hash of a [`VersionedCommittedBlock`] used for signing in consensus.
+        /// The normal [`Hashof<VersionedCommittedBlock>`] will change based on who
+        /// has signed the block. If you want to compare the contents of a block only
+        /// use this hash instead.
+        #[derive(
+            Debug,
+            Display,
+            Clone,
+            Copy,
+            PartialEq,
+            Eq,
+            Getters,
+            Decode,
+            Encode,
+            Deserialize,
+            Serialize,
+            IntoSchema,
+        )]
+        #[display(fmt = "({internal})")]
+        #[repr(transparent)]
+        #[serde(transparent)]
+        #[ffi_type(unsafe {robust})]
+        pub struct PartialBlockHash {
+            /// The hash value.
+            pub internal: Hash,
+        }
 
         /// The `CommittedBlock` struct represents a block accepted by consensus
         #[version_with_scale(n = 1, versioned = "VersionedCommittedBlock")]
@@ -178,12 +226,21 @@ mod committed {
             }
         }
 
-        /// Calculate the hash of the current block.
-        /// `VersionedCommitedBlock` should have the same hash as `VersionedCommitedBlock`.
+        /// Calculate the [`PartialBlockHash`] of this block.
         #[cfg(feature = "std")]
         #[inline]
+        pub fn partial_hash(&self) -> PartialBlockHash {
+            match self {
+                Self::V1(v1) => v1.partial_hash(),
+            }
+        }
+
+        /// Calculate the [`HashOf<VersionedCommittedBlock>`] for this block.
+        #[cfg(feature = "std")]
         pub fn hash(&self) -> HashOf<Self> {
-            self.as_v1().hash().transmute()
+            match self {
+                Self::V1(v1) => v1.hash().transmute(),
+            }
         }
 
         /// Return signatures that are verified with the `hash` of this block
@@ -213,12 +270,26 @@ mod committed {
     }
 
     impl CommittedBlock {
-        /// Calculate the hash of the current block.
-        /// `CommitedBlock` should have the same hash as `ValidBlock`.
+        /// Calculate the partial hash of the current block.
+        /// [`CommitedBlock`] should have the same partial hash as [`PendingBlock`].
+        #[cfg(feature = "std")]
+        #[inline]
+        pub fn partial_hash(&self) -> PartialBlockHash {
+            PartialBlockHash {
+                internal: Hash::new(self.header.payload()),
+            }
+        }
+        /// Calculate the complete hash of the block that includes signatures.
         #[cfg(feature = "std")]
         #[inline]
         pub fn hash(&self) -> HashOf<Self> {
-            HashOf::new(&self.header).transmute()
+            let mut data = Vec::new();
+            data.extend(self.header.payload());
+            for s in self.signatures.iter() {
+                data.extend(s.key_payload());
+                data.extend(s.signature_payload());
+            }
+            Hash::new(&data).typed()
         }
 
         /// Return signatures that are verified with the `hash` of this block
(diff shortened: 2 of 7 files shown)


outoftardis:
> PerfectBlockHash (as in finished)

That doesn't seem like a good naming choice, to be honest. "Perfected" might, in certain contexts, be considered a synonym for "finished", but I don't see how `PerfectBlockHash` would be interpreted correctly.

SamHSmith:
> > PerfectBlockHash (as in finished)
> 
> That doesn't seem like a good naming choice, to be honest. "Perfected" might, in certain contexts, be considered a synonym for "finished", but I don't see how `PerfectBlockHash` would be interpreted correctly.

I will change it to FinishedBlockHash.

data_model/src/block.rs
```diff
@@ -145,11 +183,27 @@ mod committed {
...
         #[cfg(feature = "std")]
         #[inline]
         pub fn hash(&self) -> HashOf<Self> {
-            self.as_v1().hash().transmute()
+            match self {
+                Self::V1(v1) => v1.hash().transmute(),
+            }
+        }
+        #[cfg(feature = "std")]
+        pub fn perfect_hash(&self) -> PerfectBlockHash {
```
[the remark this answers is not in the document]
Arjentix: Maybe we should use `HashOf<Self>` as final hash and something like `raw_hash()` for hash without signatures?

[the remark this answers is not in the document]
Arjentix: Seems more logical to me. And also it doesn't require to change a lot of code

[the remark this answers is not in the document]
SamHSmith: I think it requires *a lot* of code change. All of our signing and cryptography code assumes that a SignatureOf<Thing> uses the HashOf<Thing> as the basis for the signature. As for which one you should use, always use PerfectBlockHash/FinishedBlockHash unless you have reason not to do so.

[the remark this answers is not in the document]
Arjentix: So in most cases we need `FinishedBlockHash`. I think, it's better to make this behaviour the default one with `hash()` and `HashOf<Self>`. Isn't cryptography part just a one of our modules? And right now you have changed a lot of modules just to replace `HashOf<...>` with `FinishedBlockHash`

core/benches/validation.rs
```diff
@@ -224,7 +198,6 @@ criterion_group!(
 );
 criterion_group!(
     blocks,
-    chain_blocks,
```
Arjentix, line 227: Why removed?

core/src/smartcontracts/isi/block.rs
```diff
@@ -41,7 +44,11 @@ impl ValidQuery for FindBlockHeaderByHash {
         let block = wsv
             .all_blocks_by_value()
             .find(|block| block.hash() == hash)
-            .ok_or_else(|| QueryExecutionFailure::Find(Box::new(FindError::Block(hash))))?;
+            .ok_or_else(|| {
+                QueryExecutionFailure::Find(Box::new(FindError::Block(FinishedBlockHash {
+                    internal: *hash,
+                })))
+            })?;
```
Arjentix, line 51: Looks like we searching by just `hash()` but error says about `FinishedBlockHash`

data_model/src/block.rs
```diff
@@ -128,11 +171,30 @@ mod committed {
...
+        /// `VersionedCommitedBlock` should have the same hash as `VersionedCandidateCommittedBlock`.
         #[cfg(feature = "std")]
         #[inline]
         pub fn hash(&self) -> HashOf<Self> {
-            self.as_v1().hash().transmute()
+            match self {
+                Self::V1(v1) => v1.hash().transmute(),
+            }
+        }
+        /// Calculate the `FinishedBlockHash` for this block.
```
Arjentix, line 182:
```suggestion
        /// Calculate the  [`FinishedBlockHash`] for this block.
```

data_model/src/block.rs
```diff
@@ -128,11 +171,30 @@ mod committed {
         }
 
         /// Calculate the hash of the current block.
-        /// `VersionedCommitedBlock` should have the same hash as `VersionedCommitedBlock`.
+        /// `VersionedCommitedBlock` should have the same hash as `VersionedCandidateCommittedBlock`.
```
Arjentix, line 174:
```suggestion
        /// [`VersionedCommitedBlock`] should have the same hash as [`VersionedCandidateCommittedBlock`].
```

data_model/src/block.rs
```diff
@@ -128,11 +171,30 @@ mod committed {
...
+                    data.extend(v1.header.payload());
+                    for s in v1.signatures.iter() {
+                        data.extend(s.key_payload());
+                        data.extend(s.signature_payload());
+                    }
+                    FinishedBlockHash {
+                        internal: Hash::new(&data),
+                    }
+                }
+            }
```
Arjentix, line 197: Is it okay, that `finished_hash()` does not explicitly include `hash()`?

docs/source/references/schema.json
```diff
@@ -1070,6 +1063,16 @@
...
+  "iroha_data_model::block::committed::FinishedBlockHash": {
+    "Struct": {
+      "declarations": [
+        {
+          "name": "internal",
+          "ty": "iroha_crypto::hash::Hash"
+        }
+      ]
+    }
+  },
```
Arjentix, line 1075: Shouldn't this will be default hash for queries from SDK libraries? If yes, then this will complicate the usage from their side. Probably better to remove `internal` from public interface. I.e. using `PerfectHash(Hash)` or manually implementing `IntoSchema` and (de-)serialization

data_model/src/error.rs
Arjentix, line 1: @mversic , I think, you have deleted this file? Was there any reason?

mversic: I deleted it, errors are now in different modules

data_model/src/block.rs
```diff
@@ -79,6 +100,28 @@ mod committed {
...
+        Deserialize,
+        PartialEq,
+        Eq,
+        Display,
+        Hash,
+    )]
+    pub struct FinishedBlockHash {
+        /// The hash value.
+        pub internal: Hash,
+    }
```
mversic, line 123:
```suggestion
    model! {
        /// The hash of a `VersionedCommittedBlock` that has gone through consensus.
        /// What makes this hash special is that it also verifies the signatures that
        /// were applied to the block during consensus.
        #[derive(Debug, Display, Clone, Copy, PartialEq, Eq, Hash, Decode, Encode, Deserialize, Serialize, IntoSchema)]
        #[serde(transparent)]
        #[repr(transparent)]
        // SAFETY: FinishedBlockHash has no trap representations in Hash
        #[ffi_type(unsafe {robust})]
        pub struct FinishedBlockHash {
            /// The hash value.
            pub internal: Hash,
        }
    }
```

data_model/src/block.rs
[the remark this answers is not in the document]
SamHSmith: You were right. I am changing this to be inline with your suggestion.

SamHSmith:
This PR is waiting for this other PR to be merged first. https://github.com/hyperledger/iroha/pull/3323

core/benches/validation.rs
SamHSmith: Already removed on dev, conversation does not apply anymore.

core/src/smartcontracts/isi/block.rs
SamHSmith: This is reverted to the original now.

docs/source/references/schema.json
SamHSmith: You are correct and that is why I went with the partial hash scheme.

data_model/src/block.rs
SamHSmith: not applicable anymore

crypto/src/signature.rs
```diff
@@ -110,6 +110,15 @@ impl Signature {
 
         Ok(())
     }
+
+    /// Get the payload of the public key in this signature.
+    pub fn key_payload(&self) -> &[u8] {
+        self.public_key.payload()
+    }
+    /// Get the encrypted payload of this signature.
```
0x009922, line 118:
```suggestion
    }

    /// Get the encrypted payload of this signature.
```

data_model/src/block.rs
```diff
@@ -83,6 +104,21 @@ mod committed {
 
     use super::*;
 
+    model! {
+        /// The hash of a `VersionedCommittedBlock` that has gone through consensus.
```
0x009922, line 108:
```suggestion
        /// The hash of a [`VersionedCommittedBlock`] that has gone through consensus.
```

data_model/src/block.rs
```diff
@@ -132,11 +168,20 @@ mod committed {
...
-        /// `VersionedCommitedBlock` should have the same hash as `VersionedCommitedBlock`.
+        /// `VersionedCommitedBlock` should have the same hash as `VersionedCandidateCommittedBlock`.
         #[cfg(feature = "std")]
         #[inline]
+        pub fn partial_hash(&self) -> PartialBlockHash {
+            match self {
+                Self::V1(v1) => v1.partial_hash(),
+            }
+        }
+        /// Calculate the `HashOf<VersionedCommittedBlock>` for this block.
```
0x009922, line 179:
```suggestion
        /// Calculate the [`HashOf<VersionedCommittedBlock>`] for this block.
```

0x009922 (commented):
I guess these problems might be already resolved "in parallel" before squashing, right?

0x009922:
Having at least this as the PR description would be nice.

> Right now the signatures are not included in the block hash because the consensus needs to be able to refer to the block during the signing process. This is a security problem since the validity of the blockchain is rooted in the signatures and as we move closer to the whitepaper they also affect how the topology evolves. Therefore the hash should be split into a consensus used hash that refers to the contents of the block without signatures and the default hash should include signatures.

\- https://github.com/hyperledger/iroha/issues/3249#issuecomment-1440943982

data_model/src/block.rs
```diff
@@ -175,12 +220,26 @@ mod committed {
     }
 
     impl CommittedBlock {
-        /// Calculate the hash of the current block.
-        /// `CommitedBlock` should have the same hash as `ValidBlock`.
+        /// Calculate the partial hash of the current block.
+        /// [`CommitedBlock`] should have the same partial hash as [`PendingBlock`].
+        #[cfg(feature = "std")]
+        #[inline]
+        pub fn partial_hash(&self) -> PartialBlockHash {
```
mversic, line 227: this is only used by `iroha_core`. If it will never be used by `iroha_client` or `iroha_wasm` then you should move this into `iroha_core`. I know this is going to be a bother now, but we're trying to have clean public API. This PR should make no changes to `data_model`

You can bypass orphan rule if you define `PartialHash` trait so you can implement for structs in `iroha_data_model` like `CommittedBlock`

data_model/src/peer.rs
```diff
@@ -54,6 +54,13 @@ impl PeerId {
             public_key: public_key.clone(),
         }
     }
+    /// Serialize the data contained in this Id for use in hashing.
+    pub fn payload(&self) -> Vec<u8> {
+        let mut data = Vec::new();
+        data.extend(self.address.bytes());
+        data.extend(self.public_key.payload());
+        data
+    }
```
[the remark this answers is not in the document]
SamHSmith: Yes, but, if we allow the address to be manipulated a hacker could render the network broken. Where to connect to is an important peice of data that must also be secured by the hashing.

data_model/src/block.rs
SamHSmith: Users will want to have access to this value because it's very important for debugging. And all code that depended on the behavior of the old hash is to use this one instead because it is a preservation. Partial hash = Old Hash. Therefore it should stay.

data_model/src/peer.rs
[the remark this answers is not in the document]
mversic: but hacker can manipulate the address to affect the hash value?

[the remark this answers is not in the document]
SamHSmith: We should probably verify in sumeragi that addresses have not been fiddled with. But that is a seperate issue.

mversic (dismissed):

ilchu (dismissed):

ilchu (approved):

mversic (approved):

PR #3310, opened 2023-03-17
[feature] #2097: Store asset-registering transactions

### LINKED ISSUE #2097: Store asset-definition-registering transactions
When an asset (definition) gets registered the hash of the transaction that registered it should be stored somewhere so that it can be queried. 

The preferred solution is to store the data inside the block store only, so that the data can be restored across invocations of Iroha.

ilchu:
### Description of the Change
* Augmented `Account` with a dashmap storing asset registering transactions.
* New query to get those transactions.

### Issue
Closes #2097.

### Benefits
Ease of use and querying for SDK developers.

### Possible Drawbacks
* Fatter `Account` structs.
* Concerns about event implications. Now (un)registering an asset emits two events instead of one, and e.g. using `FindError::Asset` to report failing to retrieve a value from the map might be not ideal semantically. 

### Alternate Designs
Does the `AssetDefinition` need the same functionality? I.e. do we want to augment `Domain` with the same kind of map?

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1489,6 +1489,13 @@ pub mod asset {
     pub fn by_id(asset_id: impl Into<EvaluatesTo<<Asset as Identifiable>::Id>>) -> FindAssetById {
         FindAssetById::new(asset_id)
     }
+
+    /// Construct a query to get an asset definition registering transaction by its id
+    pub fn register_tx_by_definition_id(
+        asset_definition_id: impl Into<EvaluatesTo<<AssetDefinition as Identifiable>::Id>>,
+    ) -> FindAssetRegisteringTransaction {
+        FindAssetRegisteringTransaction::new(asset_definition_id)
+    }
 }
 
 pub mod block {
--- a/client/tests/integration/queries/asset_register_tx.rs
+++ b/client/tests/integration/queries/asset_register_tx.rs
@@ -1,0 +1,47 @@
+#![allow(clippy::restriction)]
+
+use std::thread;
+
+use eyre::Result;
+use iroha_client::client::asset;
+use iroha_crypto::KeyPair;
+use iroha_data_model::prelude::*;
+use test_network::*;
+
+use crate::integration::Configuration;
+
+#[test]
+fn find_asset_register_tx_returns_the_right_hash() -> Result<()> {
+    let (_rt, _peer, test_client) = <PeerBuilder>::new().with_port(10_690).start_with_runtime();
+    wait_for_genesis_committed(&vec![test_client.clone()], 0);
+    let pipeline_time = Configuration::pipeline_time();
+    thread::sleep(pipeline_time * 2);
+
+    let domain_id: DomainId = "domain".parse()?;
+    let create_domain = RegisterBox::new(Domain::new(domain_id.clone()));
+    test_client.submit_blocking(create_domain)?;
+
+    let account_id: AccountId = "account@domain".parse()?;
+    let (public_key, _) = KeyPair::generate()?.into();
+    let account = Account::new(account_id, [public_key]);
+    let create_account = RegisterBox::new(account);
+    test_client.submit_blocking(create_account)?;
+
+    let asset_definition_id_1 = AssetDefinitionId::new("xor".parse()?, domain_id.clone());
+    let create_asset = RegisterBox::new(AssetDefinition::quantity(asset_definition_id_1.clone()));
+    let asset_reg_tx_hash_1 = test_client.submit_blocking(create_asset)?;
+
+    let asset_definition_id_2 = AssetDefinitionId::new("val".parse()?, domain_id);
+    let create_asset = RegisterBox::new(AssetDefinition::quantity(asset_definition_id_2.clone()));
+    let asset_reg_tx_hash_2 = test_client.submit_blocking(create_asset)?;
+
+    let query_result_1 =
+        test_client.request(asset::register_tx_by_definition_id(asset_definition_id_1))?;
+    let query_result_2 =
+        test_client.request(asset::register_tx_by_definition_id(asset_definition_id_2))?;
+
+    assert_eq!(query_result_1, asset_reg_tx_hash_1.into());
+    assert_eq!(query_result_2, asset_reg_tx_hash_2.into());
+
+    Ok(())
+}
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -14,6 +14,7 @@ pub mod wsv;
 use core::time::Duration;
 
 use dashmap::{DashMap, DashSet};
+use iroha_crypto::HashOf;
 use iroha_data_model::prelude::*;
 use parity_scale_codec::{Decode, Encode};
 use tokio::sync::broadcast;
@@ -47,6 +48,11 @@ pub type PermissionTokensMap = DashMap<<Account as Identifiable>::Id, Permission
 /// API to work with a collections of [`PermissionTokenDefinitionId`] : [`PermissionTokenDefinition`] mappings.
 pub type PermissionTokenDefinitionsMap =
     DashMap<<PermissionTokenDefinition as Identifiable>::Id, PermissionTokenDefinition>;
+
+/// API to work with collection of key([`AssetDefinitionId`])-value([`HashOf<VersionedSignedTransaction>`])
+/// pairs.
+pub type AssetRegisterTxsMap =
+    DashMap<<AssetDefinition as Identifiable>::Id, HashOf<VersionedSignedTransaction>>;
 
 /// Type of `Sender<Event>` which should be used for channels of `Event` messages.
 pub type EventsSender = broadcast::Sender<Event>;
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -14,12 +14,15 @@ use dashmap::{mapref::entry::Entry, DashMap};
 use eyre::{Report, Result};
 use iroha_config::queue::Configuration;
 use iroha_crypto::HashOf;
-use iroha_data_model::transaction::prelude::*;
+use iroha_data_model::{
+    expression::Context, isi::Instruction, prelude::AssetDefinition, transaction::prelude::*,
+    Identifiable, RegistrableBox,
+};
 use iroha_primitives::{must_use::MustUse, riffle_iter::RiffleIter};
 use rand::seq::IteratorRandom;
 use thiserror::Error;
 
-use crate::{prelude::*, tx::CheckSignatureCondition as _};
+use crate::{prelude::*, smartcontracts::Evaluate, tx::CheckSignatureCondition as _};
 
 /// Lockfree queue for transactions
 ///
@@ -113,6 +116,39 @@ impl Queue {
             .filter(|e| self.is_pending(e.value(), wsv))
             .map(|e| e.value().clone())
             .collect()
+    }
+
+    /// Returns latest asset registering transaction hash.
+    pub fn latest_asset_reg_tx_hash(
+        &self,
+        wsv: &WorldStateView,
+        asset_definition_id: &<AssetDefinition as Identifiable>::Id,
+    ) -> Option<HashOf<VersionedSignedTransaction>> {
+        self.txs
+            .iter()
+            .filter(|e| -> bool {
+                self.is_pending(e.value(), wsv)
+                    && match e.value().payload().instructions {
+                        Executable::Instructions(ref instructions) => {
+                            instructions.iter().any(|i| match i {
+                                Instruction::Register(reg_box) => {
+                                    let context = Context::new();
+                                    reg_box.object.evaluate(wsv, &context).map_or(
+                                        false,
+                                        |reg_box| {
+                                            matches!(reg_box, RegistrableBox::AssetDefinition(inner) if inner.id() == asset_definition_id)
+                                        },
+                                    )
+                                }
+                                _ => false,
+                            })
+                        }
+                        // TODO: actually track asset registering via WASM smartcontracts as well
+                        Executable::Wasm(_) => false,
+                    }
+            })
+            .max_by_key(|e| e.value().payload().creation_time)
+            .map(|e| *e.key())
     }
 
     /// Returns `n` randomly selected transaction from the queue.
@@ -341,6 +377,16 @@ impl Queue {
             expired_transactions.extend(expired_txs);
         })
     }
+
+    /// Create a queue to be used in test WSV only.
+    pub fn default_queue_for_testing() -> std::sync::Arc<Self> {
+        use iroha_config::{base::proxy::Builder, queue::ConfigurationProxy};
+        std::sync::Arc::new(Queue::from_configuration(
+            &ConfigurationProxy::default()
+                .build()
+                .expect("Default queue config should always build"),
+        ))
+    }
 }
 
 #[cfg(test)]
@@ -400,19 +446,20 @@ mod tests {
     fn push_tx() {
         let key_pair = KeyPair::generate().unwrap();
         let kura = Kura::blank_kura_for_testing();
-        let wsv = Arc::new(WorldStateView::new(
-            world_with_test_domains([key_pair.public_key().clone()]),
-            kura.clone(),
-        ));
-
-        let queue = Queue::from_configuration(&Configuration {
+        let queue = Arc::new(Queue::from_configuration(&Configuration {
             maximum_transactions_in_block: 2,
             transaction_time_to_live_ms: 100_000,
             maximum_transactions_in_queue: 100,
             ..ConfigurationProxy::default()
                 .build()
                 .expect("Default queue config should always build")
-        });
+        }));
+
+        let wsv = Arc::new(WorldStateView::new(
+            world_with_test_domains([key_pair.public_key().clone()]),
+            &Arc::clone(&queue),
+            kura.clone(),
+        ));
 
         queue
             .push(accepted_tx("alice@wonderland", 100_000, key_pair), &wsv)
@@ -425,19 +472,21 @@ mod tests {
 
         let key_pair = KeyPair::generate().unwrap();
         let kura = Kura::blank_kura_for_testing();
-        let wsv = Arc::new(WorldStateView::new(
-            world_with_test_domains([key_pair.public_key().clone()]),
-            kura.clone(),
-        ));
 
-        let queue = Queue::from_configuration(&Configuration {
+        let queue = Arc::new(Queue::from_configuration(&Configuration {
             maximum_transactions_in_block: 2,
             transaction_time_to_live_ms: 100_000,
             maximum_transactions_in_queue: max_txs_in_queue,
             ..ConfigurationProxy::default()
                 .build()
                 .expect("Default queue config should always build")
-        });
+        }));
+
+        let wsv = Arc::new(WorldStateView::new(
+            world_with_test_domains([key_pair.public_key().clone()]),
+            &Arc::clone(&queue),
+            kura.clone(),
+        ));
 
         for _ in 0..max_txs_in_queue {
             queue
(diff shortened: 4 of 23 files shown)


QuentinI:
I don't really see why we would need this. Did we perhaps want this for `AssetDefinition` s, not `Asset`s?

core/src/smartcontracts/isi/account.rs
```diff
@@ -29,13 +29,19 @@ pub mod isi {
             wsv: &WorldStateView,
         ) -> Result<(), Self::Error> {
             let asset_id = self.object.id();
+            let account_id = asset_id.account_id.clone();
```
Arjentix, line 32: I can't leave a comment for a proper line, but don't you think @appetrosyan meant `Register<AssetDefinition>` in #2097 rather than `Register<Asset>`? Btw I don't like that we even have `Register<Asset>`

data_model/src/query.rs
```diff
@@ -1135,6 +1135,48 @@ pub mod asset {
...
+        Ord,
+        Hash,
+    )]
+    #[display(fmt = "Find asset registering transaction of `{id}` asset")]
+    pub struct FindAssetRegisteringTransaction {
+        /// `Id` of an [`Asset`] .
+        pub id: EvaluatesTo<AssetId>,
+        /// `Id` of an [`Account`] owning the asset .
+        pub account_id: EvaluatesTo<AccountId>,
+    }
```
Arjentix, line 1161: Why you need both `asset_id` and `account_id` if you can infer `account_id` from `asset_id`?

core/src/smartcontracts/isi/account.rs
```diff
@@ -29,13 +29,19 @@ pub mod isi {
...
 
             match wsv.asset(asset_id) {
                 Err(err) => match err {
                     QueryExecutionFailure::Find(find_err)
                         if matches!(*find_err, FindError::Asset(_)) =>
                     {
                         assert_can_register(&asset_id.definition_id, wsv, self.object.value())?;
+                        let latest_tx_hash = wsv.latest_transaction_hash();
```
Arjentix, line 40: Hm... So you use not the hash of current block, but the hash of the previous one. However docs for `Account::add_asset_register_tx()` and the corresponding field don't document this

core/src/smartcontracts/isi/asset.rs
```diff
@@ -695,4 +695,25 @@ pub mod query {
...
+            let account_id = self
+                .account_id
+                .evaluate(wsv, &Context::default())
+                .wrap_err("Failed to get account id")
+                .map_err(|e| Error::Evaluate(e.to_string()))?;
+            iroha_logger::trace!(%asset_id, %account_id);
+            let tx_hash = wsv
+                .map_account(&account_id, |account| account.asset_register_tx(&asset_id))?
+                .ok_or_else(|| FindError::Asset(asset_id))?;
+            Ok(tx_hash)
```
Arjentix, line 716: I think query should respond with the actual hash of transaction, which registered asset. But, I believe, current implementation will return hash just before the one we need

core/src/wsv.rs
```diff
@@ -943,6 +943,18 @@ impl WorldStateView {
...
+    #[allow(clippy::expect_used)]
+    pub fn latest_transaction_hash(&self) -> HashOf<VersionedSignedTransaction> {
+        self.latest_block_ref()
+            .expect("WSV should have at least genesis block")
+            .as_v1()
+            .transactions
+            .last()
+            .expect("Block should have at least one transaction")
+            .hash()
+    }
```
Arjentix, line 956: I think it should either return `Option<...>` or document that it will panic if there are no transactions yet

data_model/src/events/data/events.rs
```diff
@@ -351,6 +351,8 @@ mod account {
         Deleted(AccountId),
         AuthenticationAdded(AccountId),
         AuthenticationRemoved(AccountId),
+        AssetRegisterTxAdded(AccountId),
+        AssetRegisterTxRemoved(AccountId),
```
Arjentix, line 355: I strongly believe, that it should happen on every `Register<AssetDefinition>`, so I think it's obvious and don't deserve any separate events

data_model/src/asset.rs
```diff
@@ -38,6 +40,11 @@ pub type AssetDefinitionsMap =
 pub type AssetTotalQuantityMap =
     btree_map::BTreeMap<<AssetDefinition as Identifiable>::Id, NumericValue>;
 
+/// [`AssetRegisterTxMap`] provides an API to work with collection of key([`AssetId`])-value([`HashOf<VersionedSignedTransaction>`])
+/// pairs.
+pub type AssetRegisterTxsMap =
+    btree_map::BTreeMap<<Asset as Identifiable>::Id, HashOf<VersionedSignedTransaction>>;
```
Arjentix, line 46: In PR description you write about `DashMap`

Arjentix (changes requested):

data_model/src/query.rs
appetrosyan: Doubt we could infer for definitions.

core/src/smartcontracts/isi/asset.rs
appetrosyan: Why *believe* when you can test. @ilchu Consider adding a query test to `iroha_client`. Given that @astrokov7 is ignoring my requests for updates, we should make an exception.

core/src/wsv.rs
appetrosyan: 1. `clippy::expect_used` is no longer necessary. 
2. Instead each case of a panic has to be justified. Keep in mind that an `Option<HashOf...` is as efficient memory and instruction wise as panicking, due to niche-value optimisation. 
3. This function should be inline, because a simple case of pointer arithmetic now suddenly comes with two checks and a function call overhead. 
4. I don't quite see why we need this particular function at all. Inlining it into the text of the program might actually make it easier to refactor when `as_v1` isn't the only version.

data_model/src/asset.rs
appetrosyan: Just remove the reference to the concrete type used in the implementation. Anyone interested enough in this can look it up in the text of the program.

core/src/smartcontracts/isi/account.rs
QuentinI: Glossary:

| Thing | What is it? | US dollar analogy |
|--------|--------|--------|
`Asset` | Particular asset value for particular account | US dollars in my bank deposit
`AssetDefinition` |  Declaration that an asset exists with such and such value type | US dollar as a currency
`Register<AssetDefinition>` | Create a new asset definition | Declare that there now exists a currency called US dollars (Coinage Act of 1792)
`Register<Asset>` | Initialize an asset value for particular account | Open a new deposit for US dollar in my bank account (maybe with some non-zero initial balance)
`Mint<Asset>` | Add some quantity to a particular `Asset` | US Treasury mints some dollars and puts them into my bank deposit (which sure would be nice) 

Before #2227 we didn't have such a thing as `Register<Asset>`. It was a representable instruction, but it wasn't implemented and errored out if one tried to execute it, which is why  #2227 was created in the first place.

After #2227 it does what's described in the table above: creates an `Asset` with some initial value. If `Asset` with the same `AssetDefinition` id is present for this account, it errors out. Crucially,  __it's not the only way to initialize an `Asset`__. If you make a transfer or mint to an account for which the asset isn't initialized, it will automatically be created.

ilchu: Thanks a lot for such a detailed explanation! It definitely wasn't clear to me which one would we want to modify in this PR despite going through the docs and code, so I went for a wrongful guess. Thus I agree that such a rename would be beneficial, and could incorporate it into this PR as it doesn't seem to warrant a separate issue, as long as @appetrosyan agrees.

Arjentix: To me it seems like renaming `Asset` to `AssetBalance` and `AssetDefinition` to `Asset` will be pretty painful.  And not only for us, but for SDK devs too.

And I still don't understand why we can't just use `Mint<Asset>` and remove `Register<Asset>`

QuentinI: To be clear, I'm not saying we should rename stuff, just saying how one can think about them to make it less confusing. I agree actually renaming this would be quite painful.

The problem with removing `Register<Asset>` is that it's easy to remove the implementation, but hard to make the instruction actually unrepresentable.

appetrosyan: > To me it seems like renaming Asset to AssetBalance and AssetDefinition to Asset will be pretty painful. And not only for us, but for SDK devs too.

So let's lump it in with the account restructuring. 

> The problem with removing Register<Asset> is that it's easy to remove the implementation, but hard to make the instruction actually unrepresentable.

We need a matrix of representable and implemented instructions.

ilchu: Tracking it with #3331.

core/src/wsv.rs
```diff
@@ -943,6 +943,18 @@ impl WorldStateView {
         txs
     }
 
+    /// Get hash of the last transaction in the latest block.
+    #[allow(clippy::expect_used)]
```
mversic, line 947: this lint is no longer required

data_model/src/query.rs
```diff
@@ -732,6 +734,21 @@ pub mod asset {
         type Output = bool;
     }
 
+    query! {
+        /// [`FindAssetRegisteringTransaction`] Iroha Query gets [`AssetDefinitionId`] as input and finds
+        /// [`Hash`] of the transaction that registered this asset.
+        #[derive(Display)]
+        #[display(fmt = "Find asset registering transaction of `{id}` asset")]
+        pub struct FindAssetRegisteringTransaction {
```
mversic, line 742:
```suggestion
        // SAFETY: `FindAssetRegisteringTransaction` has no trap representation in `EvaluatesTo<AssetDefinitionId>`
        #[ffi_type(unsafe {robust})]
        #[repr(transparent)]
        pub struct FindAssetRegisteringTransaction {
```

client/src/client.rs
```diff
@@ -1489,6 +1489,13 @@ pub mod asset {
     pub fn by_id(asset_id: impl Into<EvaluatesTo<<Asset as Identifiable>::Id>>) -> FindAssetById {
         FindAssetById::new(asset_id)
     }
+
+    /// Construct a query to get an asset definition registering transaction by its id
+    pub fn register_tx_by_definition_id(
```
Arjentix, line 1494:
```suggestion
    pub fn registering_tx_by_definition_id(
```

client/tests/integration/queries/asset_register_tx.rs
```diff
@@ -0,0 +1,43 @@
...
+    let account_id: AccountId = "account@domain".parse()?;
+    let (public_key, _) = KeyPair::generate()?.into();
+    let account = Account::new(account_id.clone(), [public_key]);
+    let create_account = RegisterBox::new(account);
+    test_client.submit_blocking(create_account)?;
+
+    let asset_definition_id = AssetDefinitionId::new("xor".parse()?, domain_id);
+    let create_asset = RegisterBox::new(AssetDefinition::quantity(asset_definition_id.clone()));
+    let asset_reg_tx_hash = test_client.submit_blocking(create_asset)?;
+    dbg!(&asset_reg_tx_hash);
```
Arjentix, line 33: I think you can remove it as long as this test passes

client/tests/integration/queries/asset_register_tx.rs
```diff
@@ -0,0 +1,43 @@
...
+    thread::sleep(pipeline_time * 4);
+
+    let query_result =
+        test_client.request(asset::register_tx_by_definition_id(asset_definition_id))?;
+    dbg!(&query_result);
```
Arjentix, line 38: Same

client/tests/integration/queries/asset_register_tx.rs
```diff
@@ -0,0 +1,43 @@
...
+use eyre::Result;
+use iroha_client::client::asset;
+use iroha_crypto::KeyPair;
+use iroha_data_model::prelude::*;
+use test_network::*;
+
+use crate::integration::Configuration;
+
+#[test]
+fn find_asset_register_tx_returns_the_right_hash() -> Result<()> {
```
Arjentix, line 14: Nice to have a test

core/src/lib.rs
```diff
@@ -48,6 +49,11 @@ pub type PermissionTokensMap = DashMap<<Account as Identifiable>::Id, Permission
 pub type PermissionTokenDefinitionsMap =
     DashMap<<PermissionTokenDefinition as Identifiable>::Id, PermissionTokenDefinition>;
 
+/// API to work with collection of key([`DefinitionId`])-value([`HashOf<VersionedSignedTransaction>`])
+/// pairs.
```
Arjentix, line 53: Can you run `cargo doc --no-deps` and check if this doc comment passes without warnings? Because I'm not sure about `HashOf<VersionedSignedTransaction>`

core/src/smartcontracts/isi/domain.rs
```diff
@@ -85,6 +83,15 @@ pub mod isi {
                 .map_err(Error::Validate)?;
 
             let asset_definition_id = asset_definition.id().clone();
+            let latest_tx_hash = wsv
+                .latest_transaction_hash()
```
Arjentix, line 87: So you still use latest hash, but not the current hash. Your test shouldn't pass then

core/src/smartcontracts/isi/domain.rs
```diff
@@ -158,6 +165,10 @@ pub mod isi {
                 })?;
             }
 
+            if !wsv.remove_asset_register_tx(&asset_definition_id) {
```
Arjentix, line 168: Maybe we should store this info in `AssetDefinition` directly? Not it wsv?

data_model/src/query.rs
```diff
@@ -732,6 +734,21 @@ pub mod asset {
...
+            /// `Id` of an [`AssetDefinition`] .
+            pub id: EvaluatesTo<AssetDefinitionId>,
+        }
+    }
+
+    impl Query for FindAssetRegisteringTransaction {
+        type Output = iroha_crypto::Hash;
```
Arjentix, line 749:
```suggestion
        type Output = HashOf<VersionedCommittedBlock>;
```

Why not like so?

mversic: I think you mean:
```suggestion
        type Output = HashOf<VersionedSignedTransaction>;
```
 Are we finding transaction or block?

core/src/smartcontracts/isi/domain.rs
mversic: that would make more sense

core/src/smartcontracts/isi/domain.rs
```diff
@@ -85,6 +85,14 @@ pub mod isi {
                 .map_err(Error::Validate)?;
 
             let asset_definition_id = asset_definition.id().clone();
+            if let Some(latest_tx_hash) = wsv.queue_latest_asset_reg_tx_hash() {
```
mversic, line 88: the query should return latest transactions on-chain not from the queue

core/src/wsv.rs
```diff
@@ -98,7 +101,8 @@ pub struct WorldStateView {
     pub metric_tx_amounts: std::cell::Cell<f64>,
     /// Count of how many mints, transfers and burns have happened.
     pub metric_tx_amounts_counter: std::cell::Cell<u64>,
-
+    /// Reference to the pending tx queue from sumeragi.
+    queue: Arc<Queue>,
```
mversic, line 105: if queue is now part of wsv then you should remove queue from `sumeragi` and have it access queue through wsv

core/src/queue.rs
```diff
@@ -115,6 +117,37 @@ impl Queue {
...
 
+    /// Returns latest asset registering transaction hash.
+    pub fn latest_asset_reg_tx_hash(
+        &self,
+        wsv: &WorldStateView,
+    ) -> Option<HashOf<VersionedSignedTransaction>> {
+        self.txs
+            .iter()
+            .filter(|e| -> bool {
+                self.is_pending(e.value(), wsv)
```
mversic, line 128: Are you sure this transaction is guaranteed to be the one that is being part of the block in the current sumeragi round?

core/src/queue.rs
```diff
@@ -115,6 +117,37 @@ impl Queue {
...
+                                        false,
+                                        |reg_box| {
+                                            matches!(reg_box, RegistrableBox::AssetDefinition(_))
+                                        },
+                                    )
+                                }
+                                _ => false,
+                            })
+                        }
+                        Executable::Wasm(_) => false,
```
mversic, line 144: you know wasm can also register an asset?

mversic: this is going to get way more complicated

core/src/lib.rs
ilchu: This one resolved no problem, but there were some other warnings to be fixed, so I'll deal with that.

data_model/src/query.rs
ilchu: It is indeed a transaction, yes.

core/src/wsv.rs
ilchu: Should I do the same for kura then, as it's also present in both?

core/src/queue.rs
ilchu: Created #3357 and left a TODO there in case it might be big.

core/src/queue.rs
ilchu: To make the match more precise, added another argument to match against the `AssetDefinitionId`.

core/src/smartcontracts/isi/domain.rs
ilchu: This was a [suggestion](https://github.com/hyperledger/iroha/pull/3310#discussion_r1142115067) by @appetrosyan to save space. Though if instead of a map we're storing just a single hash with every `AssetDefinition`, that might make the logic even more lightweight.

core/src/queue.rs
mversic: for me this is a big no no. What is the point of introducing a feature that you know doesn't work? You could say that you made half the work done, but properly handling this might require a code that is very different than the current code in which case the current code is and would be useless. Just in case that someone would think I'm nitpicking I would like to get opinions from other colleagues

core/src/queue.rs
mversic: that still might not be correct, you can have two transactions registering the same asset in the queue and you don't know which one is the one that is being voted for in the current consensus round. I'd like to hear other opinions

core/src/queue.rs
mversic: if you remove the query than I'd be ok with merging the code as is because then it cannot be misused

core/src/smartcontracts/isi/domain.rs
mversic: yeah, we don't want to store redundant info on chain so storing it in `wsv` is the way to go

mversic: IMO, you should consider storing a hash of the block that registered the transaction, not hash of a transaction. Then when queried you should search the block for the actual transaction

core/src/smartcontracts/isi/tx.rs
```diff
@@ -3,15 +3,16 @@
 use eyre::{Result, WrapErr};
 use iroha_data_model::{
     prelude::*,
-    query::error::{FindError, QueryExecutionFailure},
+    query::error::{FindError, QueryExecutionFailure as Error},
```
0x009922, line 6: Why not using `QueryExecutionFailure` as-is? Would make more sense than just `Error`, taking into account that this error is not exclusive here (there is also `FindError`)

core/src/wsv.rs
```diff
@@ -1024,6 +1045,43 @@ impl WorldStateView {
...
+        asset_definition_id: &AssetDefinitionId,
+    ) -> Option<iroha_crypto::Hash> {
+        self.world
+            .asset_register_txs
+            .get(asset_definition_id)
+            .map(|tx| (*tx).into())
+    }
+
+    /// Add a [`AssetDefinitionId`]-[`HashOf<VersionedSignedTransaction>`] pair to [`World`]'s corresponding [`map`](crate::AssetRegisterTxsMap) if the account does not have it yet.
+    /// Return a Boolean value indicating whether or not the [`AssetRegisterTxsMap`](crate::AssetRegisterTxsMap) already had this key.
```
0x009922, line 1061:
```suggestion
    /// Add a [`AssetDefinitionId`]-[`HashOf<VersionedSignedTransaction>`] pair to [`World`]'s corresponding [`AssetRegisterTxsMap`](crate::AssetRegisterTxsMap) if the account does not have it yet.
    /// Return a Boolean value indicating whether or not the [`AssetRegisterTxsMap`](crate::AssetRegisterTxsMap) already had this key.
```

data_model/src/query.rs
```diff
@@ -843,13 +863,23 @@ pub mod asset {
         }
     }
 
+    impl FindAssetRegisteringTransaction {
+        /// Construct [`FindAssetRegisteringTransaction`].
```
0x009922, line 867:
```suggestion
        #[allow(missing_docs)]
```

appetrosyan:
I don't think this is salvageable without a major restructuring.

PR #3317, opened 2023-03-21
[refactor]: #3268 Globally unique item identifiers

### LINKED ISSUE #3268: [suggestion]: Unambiguous schema names
## Motivation

This PR is largely an under-the-hood improvement, yet the SDKs must do extra work to account for the change of the locations of `Block` structures. No semantic changes took place, yet work must be done manually. 

If all schema types have unambiguous names, we can drop the qualified path, and never have to worry about it.

_Originally posted by @appetrosyan in https://github.com/hyperledger/iroha/pull/3138#discussion_r1113743083_

mversic:
### Description of the Change

Rename items so that their names are globally unique within `Iroha` workspace

Closes #3268 

### Issue

Attaching `#[no_mangle]` to exported function whose name is derived based on the name of the struct was causing collisions. This was the preferred solution. 

### Possible Drawbacks

we might want to consider implementing some tests to make sure `iroha_schema::TypeId` is unique for every type in the codebase

In addition to this we can also consider implementing a hook for custom naming exported functions


data_model/derive/src/api.rs
```diff
@@ -10,13 +10,13 @@ pub fn process_item(mut input: syn::DeriveInput) -> TokenStream {
     }
 
     let non_transparent_item = quote! {
-        #[cfg(not(feature = "transparent_api"))]
+        #[cfg(all(not(test), not(feature = "transparent_api")))]
```
mversic, line 13: IMO `transparent_api` affecting integration tests is not desireable, ergo this change

client/benches/tps/utils.rs
```diff
@@ -222,15 +222,15 @@ impl MeasurerUnit {
         let submitter = self.client.clone();
         let interval_us_per_tx = self.config.interval_us_per_tx;
         let instructions = self.instructions();
-        let alice_id = <Account as Identifiable>::Id::from_str("alice@wonderland")
-            .expect("Failed to parse account id");
+        let account_id = account_id(self.name);
```
appetrosyan, line 225: Postfix notation? 
```rust
self.name.account_id()
```

client_cli/src/main.rs
```diff
@@ -441,9 +441,9 @@ mod account {
                 format!("Failed to deserialize signature condition from file {}", &s);
             let content = fs::read_to_string(s).wrap_err(err_msg)?;
             let condition: Box<Expression> = json5::from_str(&content).wrap_err(deser_err_msg)?;
-            Ok(Self(SignatureCheckCondition(EvaluatesTo::new_unchecked(
-                condition,
-            ))))
+            Ok(Self(SignatureCheckCondition::new(
+                EvaluatesTo::new_unchecked(condition),
+            )))
```
appetrosyan, line 446: Eliding the check here isn't a good idea. Let's use the checked variant just to be safe.

appetrosyan (commented):
Overall, this PR has a lot of changes, and only one commit. This is precisely why we have a `[refactor]:` tag with no issue attached, so we can strip out changes that don't contribute to the functionality in the ticket.

data_model/src/query.rs
```diff
@@ -1280,52 +1292,28 @@ pub mod http {
...
 
-    model! {
         /// Sized container for all possible Query results.
         #[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
         #[version_with_scale(n = 1, versioned = "VersionedQueryResult")]
         #[serde(transparent)]
         #[repr(transparent)]
+        // TODO: This should be a separate type, not just wrap Value because it infects Value
+        // with variants that can only ever be returned, i.e. can't be used in instructions
+        // enum QueryResult { ... }
```
appetrosyan, line 1313: Create a ticket for it. It could be problematic, because then the SDK would need to double all of their `Value`-based codegen. With that said, if there's no codegen, there's no problem.

data_model/src/query.rs
```diff
@@ -1343,46 +1331,55 @@ pub mod http {
...
                     account_id,
-                    filter,
+                    filter: PredicateBox::default(),
                 },
             }
         }
 
-        /// Consumes self and returns a signed [`QueryRequest`].
+        /// Construct a new request with the `query`.
+        #[must_use]
```
appetrosyan, line 1350: `#[inline]`

data_model/src/query.rs
```diff
@@ -1343,46 +1331,55 @@ pub mod http {
...
+        pub fn with_filter(mut self, filter: PredicateBox) -> Self {
+            self.payload.filter = filter;
+            self
+        }
+
+        /// Consumes self and returns a signed [`QueryBuilder`].
         ///
         /// # Errors
         /// Fails if signature creation fails.
         pub fn sign(
```
appetrosyan, line 1360: `#[inline]`

appetrosyan (commented):
We need to notify all dependent projects of the changes in this PR.

data_model/src/name.rs
```diff
@@ -17,9 +17,9 @@ model! {
     /// [`Domain`](`crate::domain::Domain`) name or
     /// [`Account`](`crate::account::Account`) name.
     #[derive(DebugCustom, Display, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Serialize, IntoSchema)]
-    #[ffi_type(opaque)]
-    #[repr(transparent)]
     #[serde(transparent)]
+    #[repr(transparent)]
+    #[ffi_type(opaque)]
```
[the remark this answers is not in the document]
mversic: I guess I have some personal convention. No, we shouldn't establish this convention, rather I should not have rearranged it

client/benches/tps/utils.rs
mversic: this is actually calling a function in this module, not a method on `name`

version/Cargo.toml
```diff
@@ -24,7 +24,6 @@ http = ["std", "warp"]
 [dependencies]
 iroha_version_derive = { version = "=2.0.0-pre-rc.13", path = "derive", default-features = false, optional = true }
 iroha_macro = { version = "=2.0.0-pre-rc.13", path = "../macro", default-features = false }
-iroha_schema = { version = "=2.0.0-pre-rc.13", path = "../schema", default-features = false }
```
appetrosyan, line 27: Nice decoupling.

schema/gen/src/lib.rs
```diff
@@ -1,100 +1,63 @@
...
         VersionedPaginatedQueryResult,
-        VersionedSignedQueryRequest,
-        VersionedSignedTransaction,
-        QueryExecutionFailure,
-
-        RegistrableBox,
-
-        // Even though these schemas are not exchanged between server and client,
-        // they can be useful to the client to generate and validate their hashes
-        MerkleTree<VersionedSignedTransaction>,
```
appetrosyan, line 46: I'd keep this type.

mversic: I'm the author of that comment. I removed it because:
1. `MerkleTree` is never exchanged between core and client, only `HashOf<MerkleTree>` is
2. if anyone has to construct `MerkleTree` schema would be of little use for them because it's only a `Vec<Node>`

data_model/src/query.rs
mversic: opened #3344

schema/src/lib.rs
```diff
@@ -1,52 +1,105 @@
 //! Module for schematizing rust types in other languages for translation.
 
 #![no_std]
+#![allow(unsafe_code)]
```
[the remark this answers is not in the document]
mversic: a leftover, because initially I made `TypeId` an unsafe trait.

schema/src/lib.rs
```diff
@@ -1,52 +1,105 @@
...
+}
 
 /// `IntoSchema` trait
-pub trait IntoSchema {
-    /// Returns unique type name.
-    // TODO: Should return &str if possible or be immutable string
+pub trait IntoSchema: TypeId {
+    /// Name under which a type is represented in the schema
+    // TODO: Should return &str or ConstString.
     fn type_name() -> String;
```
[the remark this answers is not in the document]
Erigara: hmm, ok, shouldn't than `IntoSchema::type_name` had default implementation which uses `TypeId::id` to avoid duplication in few places?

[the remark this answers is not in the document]
mversic: I thought so as well, but no because of generic parameters. If you look at `Box<SignatureWrapperOf<T>>`, by your suggestion `IntoSchema::type_name` would end up being `Box<SignatureWrapperOf<T>` which is incorrect because it should to be `Box<SignatureOf<T>>`

To illustrate (if I used `TypeId::id` as the default):
```rs
trait TypeId {
    fn type_name() -> String {
        // NOTE: this would return `format!("Box<{}>", T::id())` 
        // but it should return `format!("Box<{}>", T::type_name())
        T::id()
    }
}
impl<T: TypeId> TypeId for Box<T> {
    fn id() -> String {
        format!("Box<{}>", T::id())
    }
}
impl<T: IntoSchema> IntoSchema for Box<T> {
    // ...
}
```

data_model/src/transaction.rs
```diff
@@ -1014,9 +950,73 @@ pub mod error {
...
+    #[cfg(feature = "std")]
+    use std::vec;
+
+    use iroha_version::declare_versioned_with_scale;
     use warp::{reply::Response, Reply};
 
-    impl Reply for super::VersionedPendingTransactions {
+    use super::*;
+
+    declare_versioned_with_scale!(VersionedPendingTransactions 1..2, Debug, Clone, FromVariant, IntoSchema);
```
Erigara, line 963: Should `PendingTransactions` be versioned at all? we use this type only in queries

mversic: idk, better safe than sorry?

Erigara: yeah, let's leave it as is for now

Erigara (approved):

schema/src/serialize.rs
```diff
@@ -0,0 +1,317 @@
...
+            map.serialize_entry("name", &self.data.tag)?;
+            map.serialize_entry("type", self.type_name(type_id))?;
+            map.serialize_entry("discriminant", &self.data.discriminant)?;
+            map.end()
+        } else {
+            let mut map = serializer.serialize_map(Some(2))?;
+            map.serialize_entry("name", &self.data.tag)?;
+            map.serialize_entry("discriminant", &self.data.discriminant)?;
+            map.end()
+        }
```
0x009922, line 147:
```suggestion
        if let Some(type_id) = self.data.ty {
            let mut map = serializer.serialize_map(Some(3))?;
            map.serialize_entry("name", &self.data.tag)?;
            map.serialize_entry("type", self.type_name(type_id))?;
            map.serialize_entry("dis", &self.data.discriminant)?;
            map.end()
        } else {
            let mut map = serializer.serialize_map(Some(2))?;
            map.serialize_entry("name", &self.data.tag)?;
            map.serialize_entry("dis", &self.data.discriminant)?;
            map.end()
        }
```

docs/source/references/schema.json
0x009922, line 1: I like this change very much, in general. It is like a fresh breath from the perspective of SDK.

A few points:

- If we make major changes in schema, I have a proposal to use [internally or adjacently tagged enums](https://serde.rs/enum-representations.html) for each schema entry. It is a bit easier to work with, at least in TypeScript/JavaScript. E.g., not `"AssetDefinitionEventFilter": { "Enum": [...] }`, but `{ "kind": "Enum", ... }`
- Shorten `discriminant` field to `dis` or smth. It is pretty long and it is used very often. We already use `len` instead of `length`.
- I see that structs are now defined simply with a record, e.g.:

  **Before**

  ```json
  {
    "iroha_crypto::PublicKey": {
      "Struct": {
        "declarations": [
          {
            "name": "digest_function",
            "ty": "iroha_crypto::Algorithm"
          },
          {
            "name": "payload",
            "ty": "Vec<u8>"
          }
        ]
      }
    }
  }
  ```

  **After**

  ```json
  {
    "PublicKey": {
      "Struct": {
        "digest_function": "Algorithm",
        "payload": "Vec<u8>"
      }
    }
  }
  ```

  I like that it is shorter, but it might cause problems. In general, JSON objects are deserialised as hash-maps, in which the order of fields is not necessarily preserved, so codegen tools might generate incorrect codecs, because for SCALE the order does matter.

0x009922 (changes requested):
I think the bug with `Array<X, N>` is critical.

docs/source/references/api_spec.md
```diff
@@ -388,9 +387,9 @@ codec's [GitHub repository](https://github.com/paritytech/parity-scale-codec).
...
 |-------------------------------------|----------------------------------------------------------------|
 | `VersionedSignedTransaction`        | `iroha_data_model::transaction::VersionedSignedTransaction`    |
-| `VersionedSignedQueryRequest`       | `iroha_data_model::query::VersionedSignedQueryRequest`         |
+| `VersionedSignedQuery`              | `iroha_data_model::query::VersionedSignedQuery`                |
 | `VersionedPaginatedQueryResult`     | `iroha_data_model::query::VersionedPaginatedQueryResult`       |
-| `QueryError`                        | `iroha_core::smartcontracts::isi::query::Error`                |
+| `QueryExecutionFailure`             | `iroha_core::smartcontracts::isi::query::Error`                |
 | `FindError`                         | `iroha_core::smartcontracts::isi::error::FindError`            |
 | `VersionedBlockSubscriptionRequest` | `iroha_core::block::stream::VersionedBlockSubscriptionRequest` |
 | `VersionedBlockMessage`             | `iroha_core::block::stream::VersionedBlockMessage`             |
```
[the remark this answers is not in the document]
mversic: oh cool

docs/source/references/schema.json
```diff
@@ -1,5404 +1,3435 @@
...
+      "type": "u32",
+      "len": 4
+    }
+  },
+  "Array<u16, 8>": {
+    "Array": {
+      "type": "u32",
+      "len": 8
+    }
+  },
```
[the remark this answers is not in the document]
mversic: interesting, I'll look it up

schema/src/serialize.rs
mversic: I disagree here, we don't use abbreviations much. Is there some benefit?

docs/source/references/schema.json
mversic: > I like that it is shorter, but it might cause problems. In general, JSON objects are deserialised as hash-maps, in which the order of fields is not necessarily preserved, so codegen tools might generate incorrect codecs, because for SCALE the order does matter.

damn that's a good point

mversic: > Shorten discriminant field to dis or smth. It is pretty long and it is used very often. We already use len instead of length.

we might not even need discriminants if we disallow explicit discriminant enums. What you think?

mversic: > If we make major changes in schema, I have a proposal to use [internally or adjacently tagged enums](https://serde.rs/enum-representations.html) for each schema entry. It is a bit easier to work with, at least in TypeScript/JavaScript. E.g., not "AssetDefinitionEventFilter": { "Enum": [...] }, but { "kind": "Enum", ... }

let me see what I can do

docs/source/references/schema.json
```diff
@@ -1,5404 +1,3435 @@
...
+  "FindAllBlockHeaders": null,
+  "FindAllBlocks": null,
+  "FindAllDomains": null,
+  "FindAllParameters": null,
+  "FindAllPeers": null,
+  "FindAllPermissionTokenDefinitions": null,
+  "FindAllRoleIds": null,
+  "FindAllRoles": null,
+  "FindAllTransactions": null,
+  "FindAssetById": {
```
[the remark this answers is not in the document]
0x009922: But there is only one reference (to each):

```rust
enum QueryBox {
  FindAllAccounts(FindAllAccounts),
  // ...
}
```

Isn't it possible to remove an associated unit struct at all?

```rust
enum QueryBox {
  FindAllAccounts,
  // ...
}
```

docs/source/references/schema.json
0x009922: > > Shorten discriminant field to dis or smth. It is pretty long and it is used very often. We already use len instead of length.
> 
> we might not even need discriminants if we disallow explicit discriminant enums. What you think?

You mean, if versioned enums will not have custom discriminants?
If it is possible to avoid it, it would be cool. At least we can omit discriminants at all in case when they are not custom.

schema/src/serialize.rs
mversic: removed `discriminant` as it's not used

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: fixed

version/derive/src/lib.rs
```diff
@@ -374,7 +364,7 @@ fn impl_declare_versioned(
         pub enum #enum_name {
             #(
                 /// This variant represents a particular version.
-                #scale_variant_attributes #json_variant_attributes
```
mversic, line 377: @0x009922 was this a bug or a feature?
Look [here](https://github.com/paritytech/parity-scale-codec) for what `codec` does.
E.g.:
```rs
enum VersionedEnum {
    #[codec(index = 1)]
    V1(u32),
    #[codec(index = 2)]
    V2(u64)
```

by default when serialized discriminants would be 0, 1, but with `codec` it is 1, 2

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: This is how it looks in schema:
```json
  "QueryBox": {
    "Enum": [
      {
        "name": "FindAllAccounts",
        "type": "FindAllAccounts"
      }
      ...
    ]
}
```

Look at `"type"`. There `QueryBox` references a type `FindAllAccounts`. We don't know what it is so we go and look up it's definition in the schema and we find this:
```json
  "FindAllAccounts": null,
```

This is the usual schema resolution process. I suppose that we can flatten it like this or something similar:
```rs
```json
  "QueryBox": {
    "Enum": [
      {
        "name": "FindAllAccounts",
        "type": null
      }
      ...
    ]
}
```

but then we can do that for any type, it's not that `()` is special

docs/source/references/schema.json
mversic: have a look now, I've made changes except internally tagging enums. Would that really make a difference?

mversic: I believe I can do something like this: 
```json
  "DomainEventFilter": {
    "type": "Enum",
    "variants": [
      {
        "name": "ByCreated"
      },
      {
        "name": "ByDeleted"
      },
      {
        "name": "ByMetadataInserted"
      },
      {
        "name": "ByMetadataRemoved"
      },
      {
        "name": "ByAccount",
        "type": "FilterOpt<AccountFilter>"
      },
      {
        "name": "ByAssetDefinition",
        "type": "FilterOpt<AssetDefinitionFilter>"
      }
    ]
  },
  "DomainFilter": {
    "type": "Struct",
    "fields": [
      {
        "name": "origin_filter",
        "type": "FilterOpt<OriginFilter<DomainEvent>>"
      },
      {
        "name": "event_filter",
        "type": "FilterOpt<DomainEventFilter>"
      }
    ]
  }
```

if you think this is better?

version/derive/src/lib.rs
mversic: Iroha tests are failing now because of this. I can restore it, but tell me what you think

docs/source/references/schema.json
mversic: these are not enums we're talking about. I can do something like this:
```json
  "DomainEventFilter": {
    "type": "Enum",
    "variants": [
      {
        "name": "ByCreated"
      },
      {
        "name": "ByDeleted"
      },
      {
        "name": "ByMetadataInserted"
      },
      {
        "name": "ByMetadataRemoved"
      },
      {
        "name": "ByAccount",
        "type": "FilterOpt<AccountFilter>"
      },
      {
        "name": "ByAssetDefinition",
        "type": "FilterOpt<AssetDefinitionFilter>"
      }
    ]
  },
  "DomainFilter": {
    "type": "Struct",
    "fields": [
      {
        "name": "origin_filter",
        "type": "FilterOpt<OriginFilter<DomainEvent>>"
      },
      {
        "name": "event_filter",
        "type": "FilterOpt<DomainEventFilter>"
      }
    ]
  }
```

if you think this is better? But it might be harder to parse because you first have to read `type` and then based on that field deserialize other fields

version/derive/src/lib.rs
mversic: I reverted it. Don't want to deal with this now. It only affects how enums are serialized with SCALE

0x009922: I cannot find the reference, but I remember that `Versioned` enums were made to have custom discriminants intentionally. 1 for v1, 2 for v2 etc. So, this is a feature?

docs/source/references/schema.json
0x009922: It is not as efficient for `serde`, but it is more ergonomic for JS )
It is easier to assert and narrow the type in TypeScript.

version/derive/src/lib.rs
mversic: > I cannot find the reference, but I remember that Versioned enums were made to have custom discriminants intentionally. 1 for v1, 2 for v2 etc. So, this is a feature?

exactly. I've reverted now

docs/source/references/schema.json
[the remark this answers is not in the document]
0x009922: Isn't it possible to remove these unit structures completely from the code base? Do they make any sense in Rust code?

docs/source/references/schema.json
mversic: > It is not as efficient for `serde`, but it is more ergonomic for JS ) It is easier to assert and narrow the type in TypeScript.

efficiency is not an issue. Deserializing will never happen on the `Rust`/`serde` side. Only on the SDK side.

> But it might be harder to parse because you first have to read type and then based on that field deserialize other fields

I can't say about Typescript/JS, but I remember this conditional deserialization was a pain in the ass for me when I was a Java dev. You have to deserialize one field and then based on that field decide what structure are you deserializing. IMO it's better to know upfront what structure are you deserializing

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: > Isn't it possible to remove these unit structures completely from the code base? Do they make any sense in Rust code?

Yes, they make sense in the rust code, and no they cannot be removed from the codebase. They are not just part of the `QueryBox` they can be used on their own. `QueryBox` is only a convenience serialization format. It just so happens that they don't have any fields inside.

Maybe you would like to serialize them like this:
```json
  "FindAllAccounts": {},
```

but this might not be conveying the appropriate semantics. I've used `serde::Serializer::serialize_unit_struct` precisely because this is a unit struct

docs/source/references/schema.json
0x009922: Okay, it is a minor issue, I am okay if it remains as it is now.

docs/source/references/schema.json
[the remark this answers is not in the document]
0x009922: > Yes, they make sense in the rust code, and no they cannot be removed from the codebase.

Okay then. Is it possible to remain them in Rust, but make them invisible from the perspective in schema? I think it makes sense because there is no point to have these structs in the schema.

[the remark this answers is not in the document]
mversic: The point of the schema is to show you how it's done in the reference implementation (i.e. Rust) so that a dev can reflect this API in a specific sdk. We should strive to have a uniform developing experience of Iroha across different programming languages. If in schema you see:
```json
"QueryBox": {
  "Enum": [
  {
    "name": "FindAllAccounts",
    "type": "FindAllAccounts"
  }
  ...
  ]
}
"FindAllAccounts": null,
```

It means that in rust it looks and is used like:
```rs
pub struct FindAllAccounts;
pub enum QueryBox {
    FindAllAccounts(FindAllAccounts),
}

let query = FindAllAccounts;
let query = FindAllAccounts {}; // both are valid
let boxed_query = QueryBox::from(query); // if you want to box the query
```

whereas in Java it might look close to this:
```java
public interface Query {}

public class FindAllAccounts implements Query {
    // Schema says that this type should exist and be empty
}

public class QueryBox {
    Query query;

    public<Q extends Query> QueryBox from(Q query) {
        this.query = query;
    }
}

FindAllAccounts query = FindAllAccounts();
QueryBox boxed_query = QueryBox::from(query);
```

I don't know whether or not it will be possible for an sdk API to reflect Rust API or will there be some limitations but at least they will know by looking at schema how it is supposed to look like.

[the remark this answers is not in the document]
0x009922: > The point of the schema is to show you how it's done in the reference implementation (i.e. Rust)

I am completely disagree here. I think that the point of the schema is to be completely agnostic to any implementation and to reflect pure Iroha data model namespace and how it is represented in SCALE. Ideally, even Rust side should depend on the schema, not the opposite. It shouldn't have Rust implementation details, it should only have SCALE spec details. The schema should be the source of truth which is established first, and which is then implemented on each platform. 

That's why I think these structs are redundand - if the schema was the source of truth, it wouldn't have this extra noise.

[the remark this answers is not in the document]
0x009922: Keeping the schema as the source of truth is also good in the terms of documentations and availability of API to the users, IMO.

version/derive/src/lib.rs
0x009922: Actually, I've remembered that the intention was to make the first byte of transferred data reflect the version of it.

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: > I am completely disagree here. I think that the point of the schema is to be completely agnostic to any implementation and to reflect pure Iroha data model namespace and how it is represented in SCALE. Ideally, even Rust side should depend on the schema, not the opposite. It shouldn't have Rust implementation details, it should only have SCALE spec details. The schema should be the source of truth which is established first, and which is then implemented on each platform.

ok, I see where you're coming from and I mostly agree with that view point. The question then is: should we allow for empty type definitions in the schema. Can we say like this: if in the schema there is a type definition that is empty like in `"FindAllAccounts": null`, an sdk developer is free to not implement the empty type, but if he chooses to implement the empty type, type should be named `FindAllAccounts`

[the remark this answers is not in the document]
mversic: I would also encourage the dev to implement the empty type, but I'm not sdk dev so I don't have enough broad perspective to do so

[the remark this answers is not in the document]
mversic: I'm ok with changing the representation of an empty type as long as we agree to keep them. Some of possible variations are:
```json
  "FindAllAccounts": {
    "Tuple": []
    }
```
```json
  "FindAllAccounts": {
    "Struct": {}
    }
```

but serde decides for me that the best format is `FindAllAccounts: null` because this format doesn't assume anything about the structure of an empty type (it can be struct, enum, tuple, array, ...)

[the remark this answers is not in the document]
0x009922: > Can we say like this: if in the schema there is a type definition that is empty like in "FindAllAccounts": null, an sdk developer is free to not implement the empty type, but if he chooses to implement the empty type, type should be named FindAllAccounts

Now it works this way, at least in JS SDK.

> I would also encourage the dev to implement the empty type, but I'm not sdk dev so I don't have enough broad enough perspective to give such advices

There is a type to handle unit structs, at least in JS SDK.

---

My point is to ask the question from the perspective of schema, without knowing about Rust impl: does it make sense to have `FindAllAccounts`, or in general, `QueryBox::FindAllAccounts(())`, instead of having just `QueryBox::FindAllAccounts`? I would answer "No, it doesn't, it's just a noise and the second variant is clearer."

version/derive/src/lib.rs
0x009922: @mversic, please consider my question before marking as resolved )

docs/source/references/schema.json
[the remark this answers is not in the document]
mversic: opened #3359

docs/source/references/api_spec.md
```diff
@@ -34,7 +34,7 @@
 
 **Expects**:
 
-- Body: [`VersionedSignedQueryRequest`](#iroha-structures)
+- Body: [`VersionedSignedQuery`](#iroha-structures)
```
0x009922, line 37:
```suggestion
- Body: `VersionedSignedQuery`
```

There are other places that should be fixed too.

docs/source/references/api_spec.md
```diff
@@ -382,17 +381,3 @@ codec's [GitHub repository](https://github.com/paritytech/parity-scale-codec).
 ## Reference Iroha Client Implementation
 
 [Iroha client in Rust.](../../../client)
-
-## Iroha Structures
```
0x009922, line 386: You've removed the section, but there are still links to it, i.e. `#iroha-structures`

0x009922 (approved):

Erigara (approved):

Issue #2212, opened 2022-05-15
Hook installation fail in `color-eyre`


ilchu:
When instantiating a peer through `iroha` cli, `color_eyre` fails to install its hooks in the `with_genesis` [function](https://github.com/hyperledger/iroha/blob/iroha2-dev/cli/src/lib.rs#L149). This is weird as this should be the first initialization. As mentioned by @appetrosyan, this could be caused by some tokio task trying to do the same thing separately, and needs further investigation, as for now we're just logging and ignoring it.

QuentinI:
Couldn't reproduce neither on current HEAD nor on commits from around issue creation date.

PR #3321, opened 2023-03-22
[fix] #3031: Fix the UI/UX of missing configuration parameters

### LINKED ISSUE #3031: Enhance `ConfigurationProxy` failure reporting
> With that said, we should print *all* of the missing fields, not just the first. Given that these errors are generated in a macro, I think there can be some reusable logic

I.e. refactor the config proxy build and/or generation routine so that it lists all the fields which did not have acceptable values that would allow it to build.

Second, make the build message more informative instead of just `Proxy failed at build stage`.

Finally, `eprintln!` the received proxy to make debugging less painful.

appetrosyan:
### Description of the Change

Improve UI/UX of incorrect configuration specification. 

### Issue

Closes #3031 

### Benefits

Better UI

### Possible Drawbacks

Slightly worse readability of the `iroha_cli` crate, particularly the `main.rs` file.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -13,7 +13,6 @@ use core::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
-use eyre::ContextCompat as _;
 use iroha_config::{
     base::proxy::{LoadFromDisk, LoadFromEnv, Override},
     iroha::{Configuration, ConfigurationProxy},
@@ -32,6 +31,8 @@ use iroha_core::{
 };
 use iroha_data_model::prelude::*;
 use iroha_genesis::{GenesisNetwork, GenesisNetworkTrait, RawGenesisBlock};
+use iroha_logger::prelude::span;
+use iroha_p2p::OnlinePeers;
 use tokio::{
     signal,
     sync::{broadcast, mpsc, Notify},
@@ -152,72 +153,6 @@ impl NetworkRelay {
 }
 
 impl Iroha {
-    /// To make `Iroha` peer work all actors should be started first.
-    /// After that moment it you can start it with listening to torii events.
-    ///
-    /// # Side effect
-    /// - Prints welcome message in the log
-    ///
-    /// # Errors
-    /// - Reading genesis from disk
-    /// - Reading telemetry configs
-    /// - telemetry setup
-    /// - Initialization of [`Sumeragi`]
-    #[allow(clippy::non_ascii_literal)]
-    pub async fn new(args: &Arguments) -> Result<Self> {
-        let mut config = args
-            .config_path
-            .first_existing_path()
-            .map_or_else(
-                || {
-                    eprintln!(
-                        "Configuration file not found. Using environment variables as fallback."
-                    );
-                    ConfigurationProxy::default()
-                },
-                |path| ConfigurationProxy::from_path(&path.as_path()),
-            )
-            .override_with(ConfigurationProxy::from_env())
-            .build()?;
-
-        if style::should_disable_color() {
-            config.disable_panic_terminal_colors = true;
-            // Remove terminal colors to comply with XDG
-            // specifications, Rust's conventions as well as remove
-            // escape codes from logs redirected from STDOUT. If you
-            // need syntax highlighting, use JSON logging instead.
-            config.logger.terminal_colors = false;
-        }
-
-        let telemetry = iroha_logger::init(&config.logger)?;
-        iroha_logger::info!(
-            git_commit_sha = env!("VERGEN_GIT_SHA"),
-            "Hyperledgerいろは2にようこそ！(translation) Welcome to Hyperledger Iroha {}!",
-            env!("CARGO_PKG_VERSION")
-        );
-
-        let genesis = if let Some(genesis_path) = &args.genesis_path {
-            GenesisNetwork::from_configuration(
-                args.submit_genesis,
-                RawGenesisBlock::from_path(
-                    genesis_path
-                        .first_existing_path()
-                        .wrap_err_with(|| {
-                            format!("Genesis block file {genesis_path:?} doesn't exist")
-                        })?
-                        .as_ref(),
-                )?,
-                Some(&config.genesis),
-                &config.sumeragi.transaction_limits,
-            )
-            .wrap_err("Failed to initialize genesis.")?
-        } else {
-            None
-        };
-
-        Self::with_genesis(genesis, config, telemetry).await
-    }
-
     fn prepare_panic_hook(notify_shutdown: Arc<Notify>) {
         #[cfg(not(feature = "test-network"))]
         use std::panic::set_hook;
@@ -271,20 +206,14 @@ impl Iroha {
     /// - Reading telemetry configs
     /// - telemetry setup
     /// - Initialization of [`Sumeragi`]
-    #[allow(clippy::too_many_lines)] // This is actually easier to understand as a linear sequence of init statements.
+    #[allow(clippy::too_many_lines)]
+    #[iroha_logger::log(name="init", skip_all)]// This is actually easier to understand as a linear sequence of init statements.
     pub async fn with_genesis(
         genesis: Option<GenesisNetwork>,
         config: Configuration,
         telemetry: Option<iroha_logger::Telemetries>,
     ) -> Result<Self> {
-        if !config.disable_panic_terminal_colors {
-            if let Err(e) = color_eyre::install() {
-                let error_message = format!("{e:#}");
-                iroha_logger::error!(error = %error_message, "Tried to `color_eyre::install()` twice",);
-            }
-        }
         let listen_addr = config.torii.p2p_addr.clone();
-        iroha_logger::info!(%listen_addr, "Starting peer");
         let network = IrohaNetwork::start(listen_addr, config.public_key.clone())
             .await
             .wrap_err("Unable to start P2P-network")?;
@@ -306,11 +235,13 @@ impl Iroha {
 
         // Validate every transaction in genesis block
         if let Some(ref genesis) = genesis {
+            let span = span!(tracing::Level::TRACE, "genesis").entered();
             let wsv_clone = wsv.clone();
 
             transaction_validator
                 .validate_every(genesis.iter().cloned(), &wsv_clone)
                 .wrap_err("Transaction validation failed in genesis block")?;
+            span.exit();
         }
 
         let block_hashes = kura.init()?;
@@ -498,6 +429,25 @@ fn genesis_domain(configuration: &Configuration) -> Domain {
 
     domain
 }
+
+pub fn combine_configs(args: &Arguments) -> color_eyre::eyre::Result<Configuration> {
+    args
+        .config_path
+        .first_existing_path()
+        .map_or_else(
+            || {
+                eprintln!(
+                    "Configuration file not found. Using environment variables as fallback."
+                );
+                ConfigurationProxy::default()
+            },
+            |path| ConfigurationProxy::from_path(&path.as_path()),
+        )
+        .override_with(ConfigurationProxy::from_env())
+        .build()
+        .map_err(Into::into)
+}
+
 
 pub mod style {
     //! Style and colouration of Iroha CLI outputs.
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -30,9 +30,8 @@ use iroha_core::{
     IrohaNetwork,
 };
 use iroha_data_model::prelude::*;
-use iroha_genesis::{GenesisNetwork, GenesisNetworkTrait, RawGenesisBlock};
+use iroha_genesis::GenesisNetwork;
 use iroha_logger::prelude::span;
-use iroha_p2p::OnlinePeers;
 use tokio::{
     signal,
     sync::{broadcast, mpsc, Notify},
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -43,6 +43,7 @@ use torii::Torii;
 mod event;
 pub mod samples;
 mod stream;
+pub mod style;
 pub mod torii;
 
 /// Arguments for Iroha2.  Configuration for arguments is parsed from
@@ -441,6 +442,10 @@ fn genesis_domain(configuration: &Configuration) -> Domain {
     domain
 }
 
+/// Combine configuration proxies from several locations, preferring `ENV` vars over config file
+///
+/// # Errors
+/// - if config fails to build
 pub fn combine_configs(args: &Arguments) -> color_eyre::eyre::Result<Configuration> {
     args.config_path
         .first_existing_path()
@@ -454,84 +459,6 @@ pub fn combine_configs(args: &Arguments) -> color_eyre::eyre::Result<Configurati
         .override_with(ConfigurationProxy::from_env())
         .build()
         .map_err(Into::into)
-}
-
-pub mod style {
-    //! Style and colouration of Iroha CLI outputs.
-    use owo_colors::{OwoColorize, Style};
-
-    /// Styling information set at run-time for pretty-printing with colour
-    #[derive(Clone, Copy, Debug)]
-    pub struct Styling {
-        /// Positive highlight
-        pub positive: Style,
-        /// Negative highlight. Usually error message.
-        pub negative: Style,
-        /// Neutral highlight
-        pub highlight: Style,
-        /// Minor message
-        pub minor: Style,
-    }
-
-    impl Default for Styling {
-        fn default() -> Self {
-            Self {
-                positive: Style::new().green().bold(),
-                negative: Style::new().red().bold(),
-                highlight: Style::new().bold(),
-                minor: Style::new().green(),
-            }
-        }
-    }
-
-    /// Determine if message colourisation is to be enabled
-    pub fn should_disable_color() -> bool {
-        supports_color::on(supports_color::Stream::Stdout).is_none()
-            || std::env::var("TERMINAL_COLORS")
-                .map(|s| !s.as_str().parse().unwrap_or(true))
-                .unwrap_or(false)
-    }
-
-    impl Styling {
-        #[must_use]
-        /// Constructor
-        pub fn new() -> Self {
-            if should_disable_color() {
-                Self::no_color()
-            } else {
-                Self::default()
-            }
-        }
-
-        fn no_color() -> Self {
-            Self {
-                positive: Style::new(),
-                negative: Style::new(),
-                highlight: Style::new(),
-                minor: Style::new(),
-            }
-        }
-
-        /// Produce documentation for argument group
-        pub fn or(&self, arg_group: &[&str; 2]) -> String {
-            format!(
-                "`{}` (short `{}`)",
-                arg_group[0].style(self.positive),
-                arg_group[1].style(self.minor)
-            )
-        }
-
-        /// Convenience method for ".json or .json5" pattern
-        pub fn with_json_file_ext(&self, name: &str) -> String {
-            let json = format!("{name}.json");
-            let json5 = format!("{name}.json5");
-            format!(
-                "`{}` or `{}`",
-                json.style(self.highlight),
-                json5.style(self.highlight)
-            )
-        }
-    }
 }
 
 #[cfg(not(feature = "test-network"))]
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -76,8 +76,16 @@ impl Default for Arguments {
     }
 }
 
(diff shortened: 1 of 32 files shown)


config/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 //! Aggregate configuration for different Iroha modules.
-
+#![feature(absolute_path)]
```
mversic, line 2: do we have to use unstable feature? It's only for `Display` implementation

config/src/iroha.rs
```diff
@@ -88,48 +87,56 @@ impl ConfigurationProxy {
...
+                    });
                 }
             } else {
-                eyre::bail!(ConfigError::ProxyBuildError(
-                    "Torii config should have at least `p2p_addr` provided for sumeragi finalisation"
-                        .to_owned()
-                ))
+                return Err(ConfigError::MissingField{
+                    field: "p2p_addr",
+					message: "Torii config should have at least `p2p_addr` provided for sumeragi finalisation",
```
mversic, line 138: you have tabs in a few places

cli/src/main.rs
```diff
@@ -52,10 +70,19 @@ async fn main() -> Result<(), color_eyre::Report> {
...
             print_help(&styling)?;
+			// NOTE: because of how `color_eyre` works, we need to
+			// install the hook before creating any instance of
+			// `eyre::Report`, otherwise the `eyre` default reporting
+			// hook is going to be installed automatically.
+
+			// This results in a nasty repetition of the
+			// `color_eyre::install().unwrap()` pattern, which is the
+			// lesser of two evils
+			color_eyre::install().expect("CRITICAL");
```
mversic, line 85: I assume this is related to #2212

mversic: Also, I see this hook install around the codebase in different crates. If it's expected to fail, you think we could make it into a function?

cli/src/main.rs
```diff
@@ -84,7 +116,42 @@ async fn main() -> Result<(), color_eyre::Report> {
         }
     }
 
-    <iroha::Iroha>::new(&args).await?.start().await?;
+    // <iroha::Iroha>::new(&args).await?.start().await?;
```
mversic, line 119: seems accidental?

mversic (changes requested):

cli/src/main.rs
appetrosyan: It's not expected to fail, the point of this particular implementation is that we know for sure that we haven't installed the hook. And we have no reason to expect that the user would want to disable it (vic config params, because they're unreadable). Afterwards, it's a different story.

config/src/lib.rs
appetrosyan: In this case, there's a good reason: `std::fs::absolute` is `unstable`, but it does exactly what we need. 

Given that we're using a `nightly` only toolchain, and this function will very likely be added without changes, I took the liberty of adding this feature. If we do move back to `stable` we might want to reconsider it, but for now it does what we need.

config/src/iroha.rs
appetrosyan: I'm guessing `rustfmt` isn't replacing them with spaces.

config/base/src/lib.rs
```diff
@@ -300,54 +299,82 @@ pub mod derive {
...
+        },
 
-        /// To be used for [`Self::UnknownField`] variant construction.
-        #[inline]
-        pub fn concat_error_string(field: &[String]) -> String {
-            field.join(".")
-        }
+        // /// Used in [`Builder`] trait for build errors
+        // #[error("Proxy failed at build stage due to: {0}")]
+        // ProxyBuildError(String),
```
ilchu, line 368: Commented code. Also good to know this source of confusion is finally gone.

config/base/src/lib.rs
```diff
@@ -300,54 +299,82 @@ pub mod derive {
...
+        #[error("The value {} of {} is wrong. \nPlease change the value.", .value, .field)]
+        #[serde(skip)]
+        InsaneValue {
+            /// The value of the field that's incorrect
+            value: String,
+            /// Field name that contains invalid value
+            field: &'static str,
+            /// Additional message to be added as `color_eyre::suggestion`
+            message: String,
+            // docstring: &'static str,  // TODO: Inline the docstring for easy access
```
ilchu, line 363: Or maybe generating a link such as `iroha/docs/source/references/config.md#queuetransaction_time_to_live_ms` would be easier?

config/src/iroha.rs
```diff
@@ -88,48 +87,56 @@ impl ConfigurationProxy {
...
+    pub fn finish(&mut self) -> Result<(), ConfigError> {
         if let Some(sumeragi_proxy) = &mut self.sumeragi {
             // First, iroha public/private key and sumeragi keypair are interchangeable, but
             // the user is allowed to provide only the former, and keypair is generated automatically,
             // bailing out if key_pair provided in sumeragi no matter its value
             if sumeragi_proxy.key_pair.is_some() {
-                eyre::bail!(ConfigError::ProxyBuildError(
-                    "Sumeragi should not be provided with `key_pair` directly as it is instantiated via Iroha config. Please set the `KEY_PAIR` to `null` or omit them entirely."
-                        .to_owned()))
+                return Err(ConfigError::ProvidedInferredField{
```
ilchu, line 96:
```suggestion
                return Err(ConfigError::ProvidedInferredField {
```

config/base/src/lib.rs
```diff
@@ -356,24 +383,27 @@ pub mod runtime_upgrades;
...
 
     /// Marker traits for types for which views are implemented
     pub trait HasView {}
 
     /// Wrapper structure used to check if type implements `[HasView]`
-    /// If `T` doesn't implement [`HasView`] then `NoView::IS_HAS_VIEW` (`false`) will be used
-    /// Otherwise `IsHasView::IS_HAS_VIEW` (`true`) from `impl` block will shadow `NoView::IS_HAS_VIEW`
-    pub struct IsHasView<T>(core::marker::PhantomData<T>);
+    /// If `T` doesn't implement [`HasView`] then
+    /// [`NoView::IS_INSTANCE_HAS_VIEW`] (`false`) will be used Otherwise
```
ilchu, line 399:
```suggestion
    /// [`NoView::IS_INSTANCE_HAS_VIEW`] (`false`) will be used, otherwise
```

cli/src/lib.rs
```diff
@@ -154,93 +157,25 @@ impl Handler<iroha_core::NetworkMessage> for FromNetworkBaseRelay {
...
         //
         // A downside of this approach is that this panic hook will not work for
         // threads created by Iroha itself (e.g. Sumeragi thread).
         //
-        // # TODO
-        //
-        // Remove this when all Rust integrations tests will be converted to a
-        // separate Python tests.
+        // # TODO: Remove this when all Rust integrations tests will
+        // be converted to a separate Python tests.
```
ilchu, line 178:
```suggestion
        // be converted to separate Python tests.
```

logger/src/lib.rs
```diff
@@ -283,6 +283,7 @@ pub fn install_panic_hook() -> Result<(), Report> {
         .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
         .is_ok()
     {
+        println!("Installing eyre hook");
```
ilchu, line 286: Why not use `self::prelude::trace!` here? And also maybe a bit more specific:
```suggestion
        println!("Installing eyre hook for test network");
```

cli/src/lib.rs
```diff
@@ -497,6 +432,71 @@ impl Iroha {
...
+                    // escape codes from logs redirected from STDOUT. If you
+                    // need syntax highlighting, use JSON logging instead.
+                    config.logger.terminal_colors = false;
+                }
+                config
+            }
+            Err(e) => {
+                use iroha_config::base::derive::Error::*;
+
+                color_eyre::install()?;
```
ilchu, line 475: Do we really need to call it here as long as the hook is already installed in `cli/src/main.rs` and that's its only call site?

cli/src/main.rs
```diff
@@ -56,6 +74,16 @@ async fn main() -> Result<(), color_eyre::Report> {
...
+            // WORKAROUND for #2212: because of how `color_eyre`
+            // works, we need to install the hook before creating any
+            // instance of `eyre::Report`, otherwise the `eyre`
+            // default reporting hook is going to be installed
+            // automatically.
+
+            // This results in a nasty repetition of the
+            // `color_eyre::install().unwrap()` pattern, which is the
+            // lesser of two evils
+            color_eyre::install().expect("CRITICAL");
```
ilchu, line 86: This is a bit confusing to me as we call the hook inside this function and a few times in `.map_err(..)` invocations. Is it really needed or would moving the call to the top of `main()` help?

config/src/lib.rs
appetrosyan: > Tracking issue https://github.com/rust-lang/rust/issues/92750#issuecomment-1476829107 mentions that we can use Path::canonicalize() instead.

No because a lot of the time, we don't care if the file exists.

config/src/path.rs
```diff
@@ -15,13 +15,13 @@ pub const ALLOWED_CONFIG_EXTENSIONS: [&str; 2] = ["json", "json5"];
 pub enum ExtensionError {
     /// User provided config file without extension.
     #[error(
-        "Provided config file has no extension, allowed extensions are: {:?}.",
+        "No valid file extension found. Allowed file extensions are: {:?}.",
```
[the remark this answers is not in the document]
appetrosyan: Also used with genesis. So config is confusing.

config/base/src/lib.rs
appetrosyan: Created #3326

config/src/lib.rs
mversic: I'm up for any solution which doesn't rely on introducing an unstable feature. Otherwise, we should take this up for a vote with the team as we discussed on the retro

appetrosyan: Let's keep it to `stable` then.

logger/src/lib.rs
appetrosyan: Actually I meant to remove it.

cli/src/lib.rs
appetrosyan: We do. Courtesy of `eyre` installing a default hook when a `Report` is constructed, the hook installation being irreversible and us having the brilliant foresight of introducing a config parameter for colors, rather than using XDG standards, we actually have to initialise the correct hook **each** time we return an error.

config/src/iroha.rs
appetrosyan: Wonder why `rustfmt` doesn't fix this.

cli/src/main.rs
```diff
@@ -27,6 +29,17 @@ const REQUIRED_ENV_VARS: [(&str, &str); 7] = [
...
+/// To make `Iroha` peer work all actors should be started first.
+/// After that moment it you can start it with listening to torii events.
+///
+/// # Side effect
+/// - Prints welcome message in the log
+///
+/// # Errors
+/// - Reading genesis from disk
+/// - Reading telemetry configs
+/// - telemetry setup
```
0x009922, line 41:
```suggestion
/// - Telemetry setup
```

config/base/src/lib.rs
```diff
@@ -356,24 +380,27 @@ pub mod runtime_upgrades;
 pub mod view {
     //! Module for view related traits and structs
 
-    /// Marker trait to set default value `IS_HAS_VIEW` to `false`
+    /// Marker trait to set default value [`IsInstanceHasView::IS_INSTANCE_HAS_VIEW`] to `false`
     pub trait NoView {
         /// [`Self`] doesn't implement [`HasView`]
         const IS_HAS_VIEW: bool = false;
```
0x009922, line 386: Maybe it would be more idiomatic to use "strong type" instead of a `bool`:

```rust
enum ViewExistence {
  Exists,
  Absent
}
```

config/src/client.rs
```diff
@@ -124,20 +124,30 @@ impl ConfigurationProxy {
         if let Some(tx_ttl) = self.transaction_time_to_live_ms {
             // Really small TTL would be detrimental to performance
             if tx_ttl < TTL_TOO_SMALL_THRESHOLD {
-                eyre::bail!(
-                    ConfigError::ProxyBuildError("`TRANSACTION_TIME_TO_LIVE_MS`, network throughput may be compromised for values less than {TTL_TOO_SMALL_THRESHOLD}".to_owned())
-                );
+                eyre::bail!(ConfigError::InsaneValue {
+                    field: "TRANSACTION_TIME_TO_LIVE_MS",
```
0x009922, line 128: I believe that hard-coding SCREAMING field name should be replaced with _at least_ a `const`:

```rust
struct ConfigFieldsScreaming;

impl ConfigFieldsScreaming {
  pub const TxTimeToLiveMs: &'static str = "TRANSACTION_TIME_TO_LIVE_MS";
}

ConfigError::InsaneValue { field: ConfigFieldsScreaming::TxTimeToLiveMs }
```

config/base/src/lib.rs
appetrosyan: The biggest benefit of this approach is that you need to use a `match` or an `if let` instead of `if`. Users of the library don't expect a yes/no question to be answered with an enum, unless there's a good reason. using 
```rust 
enum HasView {
  Has(View),
  Not
}
```
would be better, but I don't see how to do it yet.

config/src/client.rs
appetrosyan: In theory, yes. In practice, 1. Rustc would complain that your constant isn't upper-snake-case (screaming), and 2. The Rust-y name for the constant is going to be precisely the same as its string representation. 

Best practices from other languages would suggest that naming a constant is a good idea, citing refactoring ability. These don't apply here for the following reasons: 
1. The symbol name is straightforward to find/replace, and it would also affect docs (which is good). 
2. And if you did change the string, you probably also need to change the variable name, because it is misleading. 
3. We don't want the config parameters to change names often if at all. Most headaches that @BAStos525 @ilchu and @Mingela are facing are because we change this too frequently.

data_model/src/asset.rs
```diff
@@ -387,6 +388,12 @@ impl fmt::Display for AssetId {
     }
 }
 
+impl fmt::Debug for AssetId {
```
appetrosyan, line 391: I know this is kind-of inconsistent, but I didn't really want to abuse proc-macros too much.

data_model/src/isi.rs
```diff
@@ -18,41 +18,57 @@ use crate::{model, Registered};
 
 model! {
     /// Sized structure for all possible Instructions.
-    #[derive(Debug, Display, Clone, PartialEq, Eq, Hash, FromVariant, EnumDiscriminants, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+    #[derive(DebugCustom, Display, Clone, PartialEq, Eq, Hash, FromVariant, EnumDiscriminants, Decode, Encode, Deserialize, Serialize, IntoSchema)]
     #[strum_discriminants(name(InstructionType), derive(Display), cfg_attr(any(feature = "ffi_import", feature = "ffi_export"), derive(iroha_ffi::FfiType)), allow(missing_docs), repr(u8))]
     #[ffi_type(opaque)]
     pub enum InstructionBox {
         /// `Register` variant.
+        #[debug(fmt = "{_0:?}")]
```
appetrosyan, line 26: Unfortunately, putting this once at the header didn't work.

crypto/src/lib.rs
```diff
@@ -177,6 +177,46 @@ ffi::ffi_item! {
...
+        match source {
+            ursa::CryptoError::NoSuchAlgorithm(_) => Self::NoSuchAlgorithm,
+            ursa::CryptoError::ParseError(source) => Self::Parse(source),
+            ursa::CryptoError::SigningError(source) => Self::Signing(source),
+            ursa::CryptoError::KeyGenError(source) => Self::KeyGen(source),
+            ursa::CryptoError::DigestGenError(source) => Self::DigestGen(source),
+            ursa::CryptoError::GeneralError(source) => Self::Other(source),
+        }
+    }
+}
```
mversic, line 218: Usually I wouldn't mind reordering but this is going to create unnecessary conflict with #3076 PR

config/base/derive/src/documented.rs
```diff
@@ -193,7 +193,13 @@ fn impl_get_recursive(ast: &StructWithFields) -> proc_macro2::TokenStream {
...
                 [stringify!(#ident)] => {
                     serde_json::to_value(&#l_value)
-                        .map_err(|e| ::iroha_config_base::derive::Error::field_error(stringify!(#ident), e))?
+                        .map_err(
+							|error|
+							::iroha_config_base::derive::Error::FieldDeserialization {
+								field: stringify!(#ident),
+								error
+							}
+						)?
```
mversic, line 202: tabs

cli/src/main.rs
```diff
@@ -43,7 +61,11 @@ async fn main() -> Result<(), color_eyre::Report> {
     if env::args().any(|a| SUBMIT_ARG.contains(&a.as_str())) {
         args.submit_genesis = true;
         if let Ok(genesis_path) = env::var("IROHA2_GENESIS_PATH") {
-            args.genesis_path = Some(ConfigPath::user_provided(&genesis_path)?);
+            args.genesis_path = Some(
+                ConfigPath::user_provided(&genesis_path)
+                    .wrap_err_with(|| "Required, because `--submit-genesis` was specified.")
+                    .wrap_err_with(|| format!("Could not read `{genesis_path}`"))?,
```
mversic, line 67: I like it how you use `wrap_err` multiple times to add more messages

mversic (dismissed):

crypto/src/lib.rs
appetrosyan: Good point

mversic (approved):

config/base/derive/src/documented.rs
mversic: still tabs

cli/src/lib.rs
```diff
@@ -271,20 +217,14 @@ impl Iroha {
     /// - Reading telemetry configs
     /// - telemetry setup
     /// - Initialization of [`Sumeragi`]
-    #[allow(clippy::too_many_lines)] // This is actually easier to understand as a linear sequence of init statements.
+    #[allow(clippy::too_many_lines)]
+    #[iroha_logger::log(name = "init", skip_all)] // This is actually easier to understand as a linear sequence of init statements.
```
QuentinI, line 221: Explainer comment slid down a line

cli/src/style.rs
```diff
@@ -0,0 +1,75 @@
...
+            highlight: Style::new().bold(),
+            minor: Style::new().green(),
+        }
+    }
+}
+
+/// Determine if message colourisation is to be enabled
+pub fn should_disable_color() -> bool {
+    supports_color::on(supports_color::Stream::Stdout).is_none()
+        || std::env::var("TERMINAL_COLORS")
```
QuentinI, line 31: Since `supports_color` takes [NO_COLOR](http://no-color.org/) into account, do we really need to read one more env variable?

config/src/iroha.rs
```diff
@@ -88,48 +87,56 @@ impl ConfigurationProxy {
...
                 sumeragi_proxy.key_pair =
                     Some(KeyPair::new(public_key.clone(), private_key.clone())?);
             } else {
-                eyre::bail!(ConfigError::ProxyBuildError(
-                    "Iroha public and private key not supplied, instantiating `sumeragi` keypair is impossible. Please provide `PRIVATE_KEY` and `PUBLIC_KEY` variables."
-                        .to_owned()
-                ))
+                return Err(ConfigError::MissingField {
+                    field: "PUBLIC_KEY and PRIVATE_KEY",
+                    message: "The sumeragi keypair is not provided in the example configuration. It's done this way to ensure you don't re-use the example keys in production, and know how to generate new keys. Please have a look at \n\nhttps://hyperledger.github.io/iroha-2-docs/guide/configure/keys.html\n\nto learn more.\n\n-----",
```
QuentinI, line 107: Maybe paraphrasing it to be impersonal would be more polite? Reading that something is fool-proofed is nicer when there's room to assume it's not _you_ who may be the fool in question.

crypto/src/lib.rs
```diff
@@ -178,6 +178,46 @@ ffi::ffi_item! {
...
 }
 
+/// Error when dealing with cryptographic functions
+#[derive(Debug, Display, serde::Deserialize)]
+pub enum Error {
+    /// Returned when trying to create an algorithm which does not exist
+    #[display(fmt = "Algorithm doesn't exist")] // TODO: which algorithm
+    NoSuchAlgorithm,
+    /// Occurs during deserialization of a private or public key
+    #[display(fmt = "Key could not be parsed. {_0}")]
```
QuentinI, line 188:
```suggestion
    #[display(fmt = "`Failed to parse key. {_0}")]
```
Similar to the rest

QuentinI (approved):

PR #3322, opened 2023-03-22
[CI]: New PR template

appetrosyan:
### Hear ye, Hear ye

We've had an old issue template for a while now. About time we modernise it a little.


6r1d (approved):
Thanks, this looks great. Let's add it! :-)

PR #3323, opened 2023-03-23
[fix] #3247: Removed fault injection from sumeragi.

SamHSmith:

### DIFF
--- a/core/benches/kura.rs
+++ b/core/benches/kura.rs
@@ -1,6 +1,6 @@
 #![allow(missing_docs, clippy::restriction)]
 
-use std::{str::FromStr as _, sync::Arc};
+use std::str::FromStr as _;
 
 use byte_unit::Byte;
 use criterion::{criterion_group, criterion_main, Criterion};
@@ -42,15 +42,19 @@ async fn measure_block_size_for_n_validators(n_validators: u32) {
         iroha_core::kura::Kura::new(iroha_config::kura::Mode::Strict, dir.path(), false).unwrap();
     let _thread_handle = iroha_core::kura::Kura::start(kura.clone());
 
-    let block = PendingBlock::new(vec![tx], Vec::new())
-        .chain_first()
-        .validate(
-            &TransactionValidator::new(transaction_limits),
-            &Arc::new(WorldStateView::new(World::new(), kura)),
-        );
-    let mut block = block
-        .sign(KeyPair::generate().expect("Failed to generate KeyPair"))
-        .unwrap();
+    let mut block = BlockBuilder {
+        transactions: vec![tx],
+        event_recommendations: Vec::new(),
+        height: 1,
+        previous_block_hash: None,
+        view_change_index: 0,
+        committed_with_topology: iroha_core::sumeragi::network_topology::Topology::new(Vec::new()),
+        key_pair: KeyPair::generate().expect("Failed to generate KeyPair"),
+        transaction_validator: &TransactionValidator::new(transaction_limits),
+        wsv: WorldStateView::new(World::new(), kura),
+    }
+    .build();
+
     for _ in 1..n_validators {
         block = block
             .sign(KeyPair::generate().expect("Failed to generate KeyPair."))
--- a/core/benches/validation.rs
+++ b/core/benches/validation.rs
@@ -132,99 +132,46 @@ fn validate_transaction(criterion: &mut Criterion) {
     println!("Success count: {success_count}, Failure count: {failure_count}");
 }
 
-fn chain_blocks(criterion: &mut Criterion) {
+fn sign_blocks(criterion: &mut Criterion) {
     let keys = KeyPair::generate().expect("Failed to generate keys");
     let transaction =
         AcceptedTransaction::accept::<false>(build_test_transaction(keys), &TRANSACTION_LIMITS)
             .expect("Failed to accept transaction.");
-    let block = PendingBlock::new(vec![transaction.into()], Vec::new());
-    let mut previous_block_hash = block.clone().chain_first().hash();
-    let mut success_count = 0;
-    let _ = criterion.bench_function("chain_block", |b| {
-        b.iter(|| {
-            success_count += 1;
-            let new_block = block.clone().chain(
-                success_count,
-                Some(previous_block_hash.transmute()),
-                0,
-                Topology::new(Vec::new()),
-            );
-            previous_block_hash = new_block.hash();
-        });
-    });
-    println!("Total count: {success_count}");
-}
-
-fn sign_blocks(criterion: &mut Criterion) {
-    let keys = KeyPair::generate().expect("Failed to generate keys");
-    let transaction = AcceptedTransaction::accept::<false>(
-        build_test_transaction(keys.clone()),
-        &TRANSACTION_LIMITS,
-    )
-    .expect("Failed to accept transaction.");
     let transaction_validator = TransactionValidator::new(TRANSACTION_LIMITS);
-    let block = PendingBlock::new(vec![transaction.into()], Vec::new())
-        .chain_first()
-        .validate(
-            &transaction_validator,
-            &Arc::new(build_test_and_transient_wsv(keys)),
-        );
     let key_pair = KeyPair::generate().expect("Failed to generate KeyPair.");
+    let kura = iroha_core::kura::Kura::blank_kura_for_testing();
+
     let mut success_count = 0;
     let mut failures_count = 0;
     let _ = criterion.bench_function("sign_block", |b| {
-        b.iter(|| match block.clone().sign(key_pair.clone()) {
-            Ok(_) => success_count += 1,
-            Err(_) => failures_count += 1,
+        b.iter(|| {
+            let block = BlockBuilder {
+                transactions: vec![transaction.clone().into()],
+                event_recommendations: Vec::new(),
+                height: 1,
+                previous_block_hash: None,
+                view_change_index: 0,
+                committed_with_topology: Topology::new(Vec::new()),
+                key_pair: key_pair.clone(),
+                transaction_validator: &transaction_validator,
+                wsv: WorldStateView::new(World::new(), kura.clone()),
+            }
+            .build();
+
+            match block.sign(key_pair.clone()) {
+                Ok(_) => success_count += 1,
+                Err(_) => failures_count += 1,
+            }
         });
     });
     println!("Success count: {success_count}, Failures count: {failures_count}");
-}
-
-fn validate_blocks(criterion: &mut Criterion) {
-    // Prepare WSV
-    let key_pair = KeyPair::generate().expect("Failed to generate KeyPair.");
-    let domain_name = "global";
-    let account_id = AccountId::new(
-        "root".parse().expect("Valid"),
-        domain_name.parse().expect("Valid"),
-    );
-    let (public_key, _) = key_pair.into();
-    let account = Account::new(account_id, [public_key]).build();
-    let domain_id = DomainId::from_str(domain_name).expect("is valid");
-    let mut domain = Domain::new(domain_id).build();
-    assert!(domain.add_account(account).is_none());
-    // Pepare test transaction
-    let keys = KeyPair::generate().expect("Failed to generate keys");
-    let transaction =
-        AcceptedTransaction::accept::<false>(build_test_transaction(keys), &TRANSACTION_LIMITS)
-            .expect("Failed to accept transaction.");
-    let block = PendingBlock::new(vec![transaction.into()], Vec::new()).chain_first();
-    let transaction_validator = TransactionValidator::new(TRANSACTION_LIMITS);
-    let kura = iroha_core::kura::Kura::blank_kura_for_testing();
-    let _ = criterion.bench_function("validate_block", |b| {
-        b.iter(|| {
-            block.clone().validate(
-                &transaction_validator,
-                &Arc::new(WorldStateView::new(
-                    World::with([domain.clone()], BTreeSet::new()),
-                    kura.clone(),
-                )),
-            )
-        });
-    });
 }
 
 criterion_group!(
     transactions,
     accept_transaction,
     sign_transaction,
-    validate_transaction,
-    validate_blocks
+    validate_transaction
 );
-criterion_group!(
-    blocks,
-    chain_blocks,
-    sign_blocks, // validate_blocks
-);
+criterion_group!(blocks, sign_blocks);
 criterion_main!(transactions, blocks);
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -16,897 +16,590 @@ use eyre::{bail, eyre, Context, Result};
 use iroha_config::sumeragi::DEFAULT_CONSENSUS_ESTIMATION_MS;
 use iroha_crypto::{HashOf, KeyPair, MerkleTree, SignatureOf, SignaturesOf};
 use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
-use iroha_version::{declare_versioned_with_scale, version_with_scale};
 use parity_scale_codec::{Decode, Encode};
 
-pub use self::{
-    candidate::{CandidateBlock, VersionedCandidateBlock},
-    candidate_committed::{CandidateCommittedBlock, VersionedCandidateCommittedBlock},
-    chained::ChainedBlock,
-    pending::{PendingBlock, VersionedPendingBlock},
-    signed::SignedBlock,
-    valid::ValidBlock,
-};
 use crate::{
     prelude::*,
     sumeragi::network_topology::{Role, Topology},
     tx::TransactionValidator,
 };
 
-mod pending {
-    use super::*;
-
-    declare_versioned_with_scale!(VersionedPendingBlock 1..2, Debug, Clone, iroha_macro::FromVariant);
-
-    /// Transaction data is permanently recorded in files called
-    /// blocks.  This is the first stage of a [`Block`]s life-cycle.
-    #[version_with_scale(n = 1, versioned = "VersionedPendingBlock")]
-    #[derive(Debug, Clone, Decode, Encode)]
-    pub struct PendingBlock {
-        /// Unix time (in milliseconds) of block forming by a peer.
-        pub timestamp: u128,
-        /// array of transactions, which successfully passed validation and consensus step.
-        pub transactions: Vec<VersionedAcceptedTransaction>,
-        /// Event recommendations.
-        pub event_recommendations: Vec<Event>,
-    }
-
-    impl PendingBlock {
-        /// Create a new [`PendingBlock`] from transactions.
-        #[inline]
-        pub fn new(
-            transactions: Vec<VersionedAcceptedTransaction>,
-            event_recommendations: Vec<Event>,
-        ) -> Self {
-            #[allow(clippy::expect_used)]
-            let timestamp = crate::current_time().as_millis();
-            // TODO: Need to check if the `transactions` vector is empty. It shouldn't be allowed.
-            Self {
-                timestamp,
-                transactions,
-                event_recommendations,
-            }
-        }
-
-        /// Chain block with the existing blockchain.
-        pub fn chain(
-            self,
-            height: u64,
-            previous_block_hash: Option<HashOf<VersionedCommittedBlock>>,
-            view_change_index: u64,
-            committed_with_topology: Topology,
-        ) -> ChainedBlock {
-            ChainedBlock {
-                transactions: self.transactions,
-                event_recommendations: self.event_recommendations,
-                header: BlockHeader {
-                    timestamp: self.timestamp,
-                    consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
-                    height: height + 1,
-                    view_change_index,
-                    previous_block_hash,
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: committed_with_topology.sorted_peers,
-                },
-            }
-        }
-
-        /// Create a new blockchain with current block as a first block.
-        pub fn chain_first_with_topology(self, genesis_topology: Topology) -> ChainedBlock {
-            ChainedBlock {
-                transactions: self.transactions,
-                event_recommendations: self.event_recommendations,
-                header: BlockHeader {
-                    timestamp: self.timestamp,
-                    consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
-                    height: 1,
-                    view_change_index: 0,
-                    previous_block_hash: None,
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: genesis_topology.sorted_peers,
-                },
-            }
-        }
-
-        /// Create a new blockchain with current block as a first block.
-        pub fn chain_first(self) -> ChainedBlock {
-            ChainedBlock {
-                transactions: self.transactions,
-                event_recommendations: self.event_recommendations,
-                header: BlockHeader {
-                    timestamp: self.timestamp,
-                    consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
-                    height: 1,
-                    view_change_index: 0,
-                    previous_block_hash: None,
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: Vec::new(),
-                },
-            }
-        }
-    }
+/// Transaction data is permanently recorded in chunks called
+/// blocks.
+#[derive(Debug, Clone, Decode, Encode)]
+pub struct PendingBlock {
+    /// Block header
+    pub header: BlockHeader,
+    /// Array of rejected transactions.
+    pub rejected_transactions: Vec<VersionedRejectedTransaction>,
+    /// Array of transactions, which successfully passed validation and consensus step.
+    pub transactions: Vec<VersionedValidTransaction>,
+    /// Event recommendations.
+    pub event_recommendations: Vec<Event>,
+    /// Signatures of peers which approved this block
+    pub signatures: SignaturesOf<Self>,
 }
 
-mod chained {
-    use super::*;
-
-    /// When [`PendingBlock`] chained with a blockchain it becomes [`ChainedBlock`]
-    #[derive(Debug, Clone, Decode, Encode)]
-    pub struct ChainedBlock {
-        /// Block header
-        pub header: BlockHeader,
-        /// Array of transactions, which successfully passed validation and consensus step.
-        pub transactions: Vec<VersionedAcceptedTransaction>,
-        /// Event recommendations.
-        pub event_recommendations: Vec<Event>,
-    }
-
-    impl ChainedBlock {
-        /// Calculate the hash of the current block.
-
-        #[inline]
-        pub fn hash(&self) -> HashOf<Self> {
-            HashOf::new(&self.header).transmute()
-        }
-
-        /// Validate block transactions against the current state of the world.
-        pub fn validate(
-            self,
-            transaction_validator: &TransactionValidator,
-            wsv: &WorldStateView,
-        ) -> ValidBlock {
-            let mut txs = Vec::new();
-            let mut rejected = Vec::new();
-
-            let wsv = wsv.clone();
-            for tx in self.transactions {
-                match transaction_validator.validate(tx.into_v1(), self.header.is_genesis(), &wsv) {
-                    Ok(tx) => txs.push(tx),
-                    Err(tx) => {
-                        iroha_logger::warn!(
-                            reason = %tx.as_v1().rejection_reason,
-                            caused_by = ?tx.as_v1().rejection_reason.source(),
-                            "Transaction validation failed",
-                        );
-                        rejected.push(tx)
-                    }
-                }
-            }
-            let mut header = self.header;
-            header.transactions_hash = txs
-                .iter()
-                .map(VersionedValidTransaction::hash)
-                .collect::<MerkleTree<_>>()
-                .hash();
-            header.rejected_transactions_hash = rejected
-                .iter()
-                .map(VersionedRejectedTransaction::hash)
-                .collect::<MerkleTree<_>>()
-                .hash();
-            let event_recommendations = self.event_recommendations;
-            // TODO: Validate Event recommendations somehow?
-            ValidBlock {
-                header,
-                rejected_transactions: rejected,
-                transactions: txs,
-                event_recommendations,
-            }
-        }
-    }
+/// Builder for `PendingBlock`
+pub struct BlockBuilder<'a> {
+    /// Block's transactions.
+    pub transactions: Vec<VersionedAcceptedTransaction>,
+    /// Block's event recommendations.
+    pub event_recommendations: Vec<Event>,
+    /// The height of the block.
+    pub height: u64,
+    /// The hash of the previous block if there is one.
+    pub previous_block_hash: Option<HashOf<VersionedCommittedBlock>>,
+    /// The view change index this block was committed with. Produced by consensus.
+    pub view_change_index: u64,
+    /// The topology thihs block was committed with. Produced by consensus.
+    pub committed_with_topology: Topology,
+    /// The keypair used to sign this block.
+    pub key_pair: KeyPair,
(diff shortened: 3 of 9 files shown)


core/benches/kura.rs
```diff
@@ -42,15 +42,18 @@ async fn measure_block_size_for_n_validators(n_validators: u32) {
...
+    let mut block = PendingBlock::new(
+        vec![tx],
+        Vec::new(),
+        1,
+        None,
+        0,
+        iroha_core::sumeragi::network_topology::Topology::new(Vec::new()),
+        KeyPair::generate().expect("Failed to generate KeyPair"),
+        &TransactionValidator::new(transaction_limits),
+        &Arc::new(WorldStateView::new(World::new(), kura)),
```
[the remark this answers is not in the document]
appetrosyan: In python we'd use named parameters (`**kwargs`). Here, the next best thing is the struct literal syntax. 

```rust
PendingBlockStub {
   transactions: Vec<VersionedAcceptedTransaction>,
   event_recommendations: Vec<Event>,
   height: u64,
   previous_block_hash: Option<HashOf<VersionedCommittedBlock>>,
   view_change_index: u64,
   committed_with_topology: Topology,
   key_pair: KeyPair,
   transaction_validator: &TransactionValidator,
   wsv: &WorldStateView,
}
```
Where you get the benefit of not worrying about filling in all the details thanks to the ability to only define the struct fields that matter and using `..PendingBlockStub::default()` later. This allows you not to have to specify things like  height (mostly it's `1` unless otherwise specified) and `view_change_index`. 

Finally, you just `impl From<PendingBlockStub> for PendingBlock` and postfix the call to the struct literal with `.into()`.

appetrosyan (changes requested):

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-                    previous_block_hash,
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: committed_with_topology.sorted_peers,
-                },
-            }
-        }
+impl PendingBlock {
+    /// Create a new [`PendingBlock`] from transactions.
+    #[inline]
```
appetrosyan, line 49: This function for sure is not a good `inline` candidate, given the register pressure.

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-                    view_change_index: 0,
-                    previous_block_hash: None,
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: genesis_topology.sorted_peers,
-                },
+        let mut txs = Vec::new();
+        let mut rejected = Vec::new();
+
+        let wsv = wsv.clone();
```
appetrosyan, line 79: Consider making this function accept a `WorldStateView` and cloning one level above.

core/benches/kura.rs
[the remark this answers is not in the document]
appetrosyan: After looking at the `new` method, it might be a good idea to just add 
```rust
impl PendingBlockStub { 
  fn build() -> PendingBlock {
  ```

File: core/src/block.rs [hunk 4]
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: Vec::new(),
-                },
-            }
+        header.transactions_hash = txs
+            .iter()
+            .map(VersionedValidTransaction::hash)
+            .collect::<MerkleTree<_>>()
+            .hash();
```
appetrosyan, line 97: Good

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-    impl ValidBlock {
-        /// Calculate hash of the current block.
+    /// Create dummy [`ValidBlock`]. Used in tests
+    ///
+    /// # Panics
+    /// If generating keys or block signing fails.
+    #[allow(clippy::restriction)]
+    #[cfg(test)]
+    pub fn new_dummy() -> Self {
+        #[allow(clippy::expect_used)]
```
appetrosyan, line 219:
```suggestion

```
It's allowed project-wide.

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-        /// Array of all rejected transactions in this block.
-        pub rejected_transactions: Vec<VersionedRejectedTransaction>,
-        /// Array of all valid transactions in this block.
-        pub transactions: Vec<VersionedValidTransaction>,
-        /// Signatures of peers which approved this block.
-        pub signatures: SignaturesOf<Self>,
-        /// Event recommendations.
-        pub event_recommendations: Vec<Event>,
-    }
+pub trait Auditable: Sized {
```
appetrosyan, line 248: Nice to see traits.

appetrosyan:
```suggestion
pub trait Revalidate: Sized {
```
Rust has a different naming convention for interfaces.

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-
-            CommittedBlock {
-                event_recommendations,
-                header,
-                rejected_transactions,
-                transactions,
-                signatures: signatures.transmute(),
-            }
-        }
+        let wsv = wsv.clone();
```
appetrosyan, line 302: Another implicit `wsv` clone.

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-        /// Array of all valid transactions in this block.
-        pub transactions: Vec<VersionedValidTransaction>,
-        /// Signatures of peers which approved this block.
-        pub signatures: SignaturesOf<Self>,
-        /// Event recommendations.
-        pub event_recommendations: Vec<Event>,
-    }
+pub trait Auditable: Sized {
+    fn revalidate<const IS_GENESIS: bool>(
+        self,
```
appetrosyan, line 250: Also, there's a question of whether or not we want for it to be Sized. Revalidate implies that you're using the object Read-only. So why take ownership?

Is it for the chain syntax? 

If there's a good reason for it, there ought to be a comment in the text of the function declaration. But I'd also consider rewriting this with a `&self`.

core/src/block.rs
```diff
@@ -19,894 +19,614 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
+                    entity_kind: PipelineEntityKind::Transaction,
+                    status: PipelineStatus::Validating,
+                    hash: transaction.hash().into(),
+                }
+                .into()
+            }))
+            .chain([PipelineEvent {
+                entity_kind: PipelineEntityKind::Block,
+                status: PipelineStatus::Validating,
+                hash: block.hash().into(),
```
appetrosyan, line 626: Good use of functional style

core/src/sumeragi/main_loop.rs
```diff
@@ -81,7 +66,7 @@ pub struct SumeragiWithFault<F: FaultInjection> {
     pub debug_force_soft_fork: bool,
 }
 
-impl<F: FaultInjection> Debug for SumeragiWithFault<F> {
+impl Debug for Sumeragi {
```
appetrosyan, line 69: Good. I don't mind a manual impl in this case.

core/src/block.rs
SamHSmith: I agree. This was a copy/paste/refactor mistake.

core/src/sumeragi/mod.rs
```diff
@@ -26,6 +25,7 @@ use network_topology::{Role, Topology};
 
 use crate::handler::ThreadHandler;
 
+/// The main event loop that powers sumeragi.
```
0x009922, line 28: I see that in other cases module-level doc tests are *usually* located inside of a module, not outside of it.

core/src/block.rs
```diff
@@ -19,894 +19,635 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
                 .transactions
                 .iter()
                 .map(VersionedValidTransaction::hash)
                 .collect::<MerkleTree<_>>()
                 .hash()
-                .ne(&self.header.transactions_hash)
+                .ne(&block.header.transactions_hash)
             {
-                return Err((self, eyre!("The transaction hash stored in the block header does not match the actual transaction hash.")));
+                return Err((block.into(), eyre!("The transaction hash stored in the block header does not match the actual transaction hash.")));
```
ilchu, line 604: Do you think having a more structured error (e.g. its own error type) in `block` would be beneficial? Getting rid of `eyre` in this module would also allow to get rid of `Sized` bound on `Revalidate`, though I agree it's minor.

core/src/block.rs
```diff
@@ -19,894 +19,635 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-            block_height: u64,
-        ) -> Result<SignedBlock, eyre::Report> {
-            if self.is_empty() {
-                bail!("Block is empty");
-            }
-
-            if self.has_committed_transactions(wsv) {
-                bail!("Block has committed transactions");
-            }
+impl Revalidate for VersionedCommittedBlock {
```
ilchu, line 429: I'm a bit on the fence about implementing this for `VersionedCommittedBlock` instead of just `CommittedBlock`. If we moved all the helper methods from VCB into CB, both methods in this trait could have a default impl, as the destructuring from `V1` seems to be the only difference. And also semantically it seems to make more sense, as I believe versioning is more of a data model concern, and we shouldn't be dealing with it in core.

core/src/sumeragi/main_loop.rs
```diff
@@ -13,17 +10,7 @@ use tracing::{span, Level};
 use super::*;
 use crate::{block::*, sumeragi::tracing::instrument};
 
-/// Fault injection for consensus tests
-pub trait FaultInjection: Send + Sync + Sized + 'static {}
```
ilchu, line 17: Great to know that it could be deleted that easily. Non-generic signatures are much easier to read.

ilchu (changes requested):

core/src/block.rs
mversic: `eyre` should be removed from `iroha_core` and errors should be made strongly typed. `eyre` is not an appropriate error handling solution for libraries. We're slowly removing it so this might be a good time to remove it from block validation api

core/src/block.rs
SamHSmith: The cloning was introduce by someone other than me in order to solve a bug. This is a immutable reference but it infact has interior mutability. If we don't clone then the wsv is modified and that is incorrect behaviour. We need a local copy we can mutate because we need to validate a series of transactions and how they affect each other.

core/src/block.rs
SamHSmith: Issue created. https://github.com/hyperledger/iroha/issues/3353

core/src/sumeragi/main_loop.rs
```diff
@@ -608,10 +584,16 @@ fn handle_message<F: FaultInjection>(
         (Message::BlockSyncUpdate(BlockSyncUpdate { block }), _) => {
             info!(%addr, %role, hash=%block.hash(), "Block sync update received");
 
-            let block = match block.revalidate(current_topology) {
+            let block_hash = block.hash();
+            let block = match block.revalidate::<true>(
```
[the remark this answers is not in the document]
SamHSmith: I'm not that convinced. But I wouldn't be against it if you think it's valuable. In that case though I think it's better if you implemented the way you want in a seperate PR. I'm not going to do it justice. Here is the issue I created.

core/src/block.rs
```diff
@@ -19,894 +19,635 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-                    timestamp: self.timestamp,
-                    consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
-                    height: 1,
-                    view_change_index: 0,
-                    previous_block_hash: None,
-                    transactions_hash: None,
-                    rejected_transactions_hash: None,
-                    committed_with_topology: genesis_topology.sorted_peers,
-                },
+        let wsv = self.wsv.clone();
```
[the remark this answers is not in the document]
SamHSmith: Pasting my comment from other conversation:
The cloning was introduce by someone other than me in order to solve a bug. This is a immutable reference but it infact has interior mutability. If we don't clone then the wsv is modified and that is incorrect behaviour. We need a local copy we can mutate because we need to validate a series of transactions and how they affect each other.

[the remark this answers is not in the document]
SamHSmith: The plan is to replace this with change sets. But that's will wait on general WSV refactoring.

core/src/block.rs
```diff
@@ -19,894 +19,635 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
+            .iter()
+            .map(VersionedValidTransaction::hash)
+            .collect::<MerkleTree<_>>()
+            .hash();
+        header.rejected_transactions_hash = rejected
+            .iter()
+            .map(VersionedRejectedTransaction::hash)
+            .collect::<MerkleTree<_>>()
+            .hash();
+        // TODO: Validate Event recommendations somehow?
```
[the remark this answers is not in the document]
SamHSmith: I don't know. This comment is not mine, I'm simply forwarding it.

core/src/block.rs
SamHSmith: I would like to keep this issue focused on the collapsing of the block types. Therefore I have created this issue concerning eyre, https://github.com/hyperledger/iroha/issues/3353

core/src/block.rs
SamHSmith: Normally I'm not for future proofing but in this case I think it is important. By having all the methods on VCB we cover for the introduction of other block versions that do things differently. Currently we assume that a v1 block will always be convertible to a v2 block. This is not the case. Imagine just the simple scenario that after a v2 block has been committed we do not want to allow any more v1 blocks. As I've written it we simply switch on the block type and can implement this logic. The previous model were we convert everything into the latest version and use that would not support this case.

ilchu: I see, that makes quite some sense. Just had a natural urge to completely get rid of code reduplication, but it's not always achievable in real life.

core/src/block.rs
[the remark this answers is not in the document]
ilchu: Gotcha, just was a bit confusing without some clarifying comment there. I do vaguely remember such conversation happening, though.

ilchu (approved):

core/src/block.rs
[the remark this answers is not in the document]
Erigara: Imo we can avoid such confusion in future by accepting `WorldStateView` as parameter, not `&WorldStateView`.

core/src/block.rs
```diff
@@ -19,894 +19,630 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
+    ///
+    /// # Errors
+    /// - Block is empty
+    /// - Block has committed transactions
+    /// - There is a mismatch between candidate block height and actual blockchain height
+    /// - There is a mismatch between candidate block previous block hash and actual latest block hash
+    /// - Block header transaction hashes don't match with computed transaction hashes
+    /// - Error during revalidation of individual transactions
+    #[allow(clippy::too_many_lines)]
+    fn revalidate<const IS_GENESIS: bool>(
```
Erigara, line 435: In case of `CommittedBlock` we only want to revalidate number of signatures and check transactions hashes.

Erigara (changes requested):
I see that this PR overlapping with #3356 in terms of block refactoring so we need to decide how we gonna merge them.
As i wrote earlier i'm for leveraging type system to prevent accidental bugs and it's what Marin did with introduction of `Candidate<T>`.
So what should we do? @SamHSmith @mversic

core/benches/validation.rs
```diff
@@ -181,50 +165,11 @@ fn sign_blocks(criterion: &mut Criterion) {
     println!("Success count: {success_count}, Failures count: {failures_count}");
 }
 
-fn validate_blocks(criterion: &mut Criterion) {
```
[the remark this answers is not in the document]
SamHSmith: The refactor broke it and it was not a good benchmark. I could have rewritten it to test how fast we could revalidate the block. But then again the block has only one transaction. It is completely pointless, all it measures is the overhead.

core/src/block.rs
mversic: > There is a bug in validation

in general, I don't we should be structuring our code around an assumption that it's buggy. Maybe this case is different, I won't go into it deeply

SamHSmith: > in general, I don't think we should be structuring our code around an assumption that it's buggy. Maybe this case is different, I won't go into it deeply

I agree. But I believe this is a case of "don't trust, verify" that we should be upholding as a blockchain project.

core/src/sumeragi/message.rs
```diff
@@ -89,14 +83,12 @@ pub enum Message {
 #[non_exhaustive]
 pub struct BlockCreated {
     /// The corresponding block.
-    pub block: VersionedCandidateBlock,
+    pub block: PendingBlock,
```
Erigara, line 86: Why not `VersionedPendingBlock`?

core/src/sumeragi/message.rs
```diff
@@ -143,14 +135,12 @@ impl From<CommittedBlock> for BlockCommitted {
 #[non_exhaustive]
 pub struct BlockSyncUpdate {
     /// The corresponding block.
-    pub block: VersionedCandidateCommittedBlock,
+    pub block: VersionedCommittedBlock,
```
Erigara, line 138: If we anyway do full re-validation on block sync there is no point in transferring  `VersionnedCommittedBlock` we can just use `VersionedPendingBlock`.

core/src/block.rs
SamHSmith: @appetrosyan Can we have your opinion on this matter?

core/src/block.rs
```diff
@@ -19,894 +19,630 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
-    pub struct PendingBlock {
-        /// Unix time (in milliseconds) of block forming by a peer.
-        pub timestamp: u128,
-        /// array of transactions, which successfully passed validation and consensus step.
-        pub transactions: Vec<VersionedAcceptedTransaction>,
-        /// Event recommendations.
-        pub event_recommendations: Vec<Event>,
-    }
+/// Builder for `PendingBlock`
+pub struct PendingBlockStub<'a> {
```
Erigara, line 48: Maybe `BlockBuilder`?

core/src/block.rs
```diff
@@ -19,894 +19,630 @@ use iroha_data_model::{block::*, events::prelude::*, transaction::prelude::*};
...
 
-mod pending {
-    use super::*;
+declare_versioned_with_scale!(VersionedPendingBlock 1..2, Debug, Clone, iroha_macro::FromVariant);
+
+/// Transaction data is permanently recorded in chunks called
+/// blocks.
+#[version_with_scale(n = 1, versioned = "VersionedPendingBlock")]
+#[derive(Debug, Clone, Decode, Encode)]
+pub struct PendingBlock {
```
Erigara, line 34: What about just `Block`?

core/src/block.rs
appetrosyan: I agree with both to an extent. 

Right now, we should do the full re-verification, until we conduct a full audit of `sumeragi` our main assumption should be that the code is exceedingly buggy, unless proven otherwise. 

In the future, when we can formally verify that the business logic was executed correctly; which at present doesn't seem possible, but given several audits, and thorough test coverage and fuzzing, should be doable; we should short-circuit double checking.

core/src/block.rs
appetrosyan: `PendingBlock` is much more informative.

core/src/sumeragi/message.rs
appetrosyan: Bad idea. The revalidation could be a precaution, but semantically  the return type is `VersionedCommittedBlock`. Keeping it that way helps with formal verification.

core/src/block.rs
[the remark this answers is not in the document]
appetrosyan: Rust in general prefers allocations be pushed as far up the call stack as possible. Thus accepting a `wsv` by reference to then clone is considered a bad practice (with caveats).

core/src/sumeragi/message.rs
SamHSmith: Blocks by consensus are always of the "most recent" version. And instead of promoting all versions to the latest version we should just use the "latest version" in the code. We can make an informed decision about how to handle live upgrades to a network in the future. The current approach gave the impression that this scheme was thought out when I don't think it was. Having a versioned committed block is important though.

I forgot to remove VersionedPendingBlock and have now done so.

core/src/block.rs
SamHSmith: I will rename it.

appetrosyan (approved):

core/src/sumeragi/message.rs
Erigara: Yes, untyped, unverified and untrusted and thats why we want to indicate that by type and enforce revalidation When receiving such type through the network.

core/src/block.rs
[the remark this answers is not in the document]
Erigara: Why this was resolved? there is neither changes nor ticket or reasons why this won't work?
Imo this change is pretty small to be separate ticket.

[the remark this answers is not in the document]
SamHSmith: It felt like bikeshedding to me since it will be returned to a pass by reference anyway after we have implemented the new-wsv.

core/src/sumeragi/message.rs
SamHSmith: In my opinion this would cause again the large amount of types. I think we should avoid turning our programming problems into typesystem problems. But if this was introduced again by another PR I would not be against it.

mversic: please do not resolve discussions without getting a green light from the OP. 

Also, in this case, it is @Erigara 's opinion as well as mine that you should return the candidate blocks in this PR. Nobody is going to be reverting your changes after you. I will not be going into discussion why it's better to enforce invariants in the type system, it's a common sense. All you should know atm is that majority vote is against you on this topic and we govern ourselves democratically as a team

core/src/block.rs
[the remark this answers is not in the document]
mversic: whether or not it's bikeshedding, you should not be resolving discussions as you see fit. Discussion can only be closed if given permission by the OP or other. Mutual agreement has to be reached

core/src/sumeragi/message.rs
appetrosyan: We wouldn't have made it a hard requirement to resolve all conversations before merging if it wasn't necessary. 

Also, the **whole point of code review** is to identify areas where a separate pair of eyes has a different opinion. @Erigara and @mversic have made it clear that they feel strongly about one way of doing things. You don't get to ignore that opinion, you **muat address it**.

core/src/block.rs
```diff
@@ -16,897 +16,628 @@ use eyre::{bail, eyre, Context, Result};
...
+    /// - Block has committed transactions
+    /// - There is a mismatch between candidate block height and actual blockchain height
+    /// - There is a mismatch between candidate block previous block hash and actual latest block hash
+    /// - Block header transaction hashes don't match with computed transaction hashes
+    /// - Error during revalidation of individual transactions
+    #[allow(clippy::too_many_lines)]
+    fn revalidate<const IS_GENESIS: bool>(
+        &self,
+        transaction_validator: &TransactionValidator,
+        wsv: &WorldStateView,
```
Erigara, line 288: The same comment about implicit `wsv` clone.

core/src/sumeragi/main_loop.rs
```diff
@@ -605,16 +593,34 @@ fn handle_message<F: FaultInjection>(
...
+            let block = block_sync_update.extract_block_unchecked();
             info!(%addr, %role, hash=%block.hash(), "Block sync update received");
 
-            let block = match block.revalidate(current_topology) {
-                Ok(block) => block,
-                Err((block, error)) => {
-                    error!(%addr, %role, hash=%block.hash(), ?error, "Block not valid");
+            let block_hash = block.hash();
+            if block
+                .revalidate::<true>(
```
Erigara, line 602: This is not genesis transaction so must be

```suggestion
                .revalidate::<false>(
```

core/src/sumeragi/main_loop.rs
```diff
@@ -605,16 +593,34 @@ fn handle_message<F: FaultInjection>(
...
+                    &state.wsv,
+                    state.latest_block_hash,
+                    state.latest_block_height,
+                )
+                .is_err()
+            {
+                /* If the block fails validation we must check again using the finaziled wsv.
+                When a soft-fork occurs the consensus-block may be valid on the previous
+                wsv but not the current one. */
+                if let Err(error) = block.revalidate::<true>(
```
Erigara, line 613: Here too:

```suggestion
                if let Err(error) = block.revalidate::<false>(
```

core/src/sumeragi/main_loop.rs
```diff
@@ -605,16 +593,34 @@ fn handle_message<F: FaultInjection>(
...
+                    &sumeragi.transaction_validator,
+                    &state.finalized_wsv,
+                    state.latest_block_hash,
+                    state.latest_block_height,
```
Erigara, line 617: If we are revalidating against `finalized_wsv` shouldn't we use appropriate `block_height` and `block_hash`?

core/src/sumeragi/message.rs
```diff
@@ -143,15 +139,49 @@ impl From<CommittedBlock> for BlockCommitted {
...
+                transaction_validator,
+                wsv,
+                latest_block,
+                block_height,
+            )?;
+            Ok(self.block)
         }
     }
+    /// Bypass typesystem enforced validation.
+    pub fn extract_block_unchecked(self) -> VersionedCommittedBlock {
```
Erigara, line 182: This method is redundant, it's used in only in single place where we can just call `validate_and_extract_block` two times (we can chain them using [`or_else`](https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else)).

core/src/sumeragi/message.rs
```diff
@@ -143,15 +139,49 @@ impl From<CommittedBlock> for BlockCommitted {
...
+        Self { block }
+    }
+}
+
+impl BlockSyncUpdate {
+    /// Extract block from block sync update message.
+    ///
+    /// # Errors
+    /// - When the block is invalid.
+    pub fn validate_and_extract_block(
```
Erigara, line 156: In my opinion it will be better to enforce type restrictions as early as possible (preferable at system boundary ) and not on the spot.
In our case it will be blocks received as part of `ShareBlocks` message.

core/src/sumeragi/message.rs
```diff
@@ -89,14 +87,12 @@ pub enum Message {
 #[non_exhaustive]
 pub struct BlockCreated {
```
Erigara, line 88: We were talking about revalidation of `CommittedBlock`, but the same applies for the `PendingBlock`.
We can't trust that we receive validated block and must revalidate it.

Erigara (changes requested):
In general i think it's possible to minimize umber of block states to 4: `PendingBlock`, `Candidate<PendingBlock>`, `CommittedBlock` and `Candidate<ComittedBlock>`.

And with smaller amount we start to lose information about block invariants.

core/src/sumeragi/message.rs
```diff
@@ -143,15 +139,49 @@ impl From<CommittedBlock> for BlockCommitted {
...
+    /// # Errors
+    /// - When the block is invalid.
+    pub fn validate_and_extract_block(
+        self,
+        transaction_validator: &TransactionValidator,
+        wsv: &WorldStateView,
+        latest_block: Option<HashOf<VersionedCommittedBlock>>,
+        block_height: u64,
+    ) -> Result<VersionedCommittedBlock, eyre::Report> {
+        if block_height == 0 {
```
Erigara, line 163: Initial idea behind `const generic` was that we know exact places in code where we expect block with certain height (0 or more than zero) and know we fall back to runtime check.

core/src/sumeragi/main_loop.rs
SamHSmith: true!

core/src/sumeragi/message.rs
SamHSmith: The chaining causes borrow checker problems but I will make your change nonetheless.

core/src/sumeragi/message.rs
SamHSmith: I don't think that works because block sync does not know if the block is valid or not. Only sumeragi knows because validity is relative to the current state.

core/src/sumeragi/message.rs
SamHSmith: Is that not what sumeragi is for? ValidatingPeers + ProxyTail all validate the block and the leader does not because it created it? Maybe there's a case for the ObservingPeers but they are supposed to receive all their blocks through block sync. (I know this is currently not the case but I have a ticket for this to do.)

core/src/sumeragi/message.rs
SamHSmith: I have created this issue, https://github.com/hyperledger/iroha/issues/3372.

core/src/sumeragi/message.rs
Erigara: Yeah, but validating peers and proxy tail receive `BlockCreated` message and we want to make sure that peer don't accidentally forget to validate it.

core/src/sumeragi/message.rs
Erigara: I mean that we still can use `const generics`, not that we should remove them. Because we only expect genesis block inside wait for genesis loop, no?
I don't think that soft fork on genesis block is possible.

core/src/sumeragi/main_loop.rs
```diff
@@ -352,11 +337,22 @@ impl<F: FaultInjection> SumeragiWithFault<F> {
...
-                            match block.revalidate_hashes() {
-                                Ok(block) => block,
-                                Err((_, error)) => {
+                            match block_sync_update.validate_and_extract_block(
+                                &self.transaction_validator,
+                                state.wsv.clone(),
+                                state.latest_block_hash,
+                                state.latest_block_height,
+                            ) {
+                                Ok(block) => match revalidate_hashes(block) {
```
Erigara, line 348: Hashes are already checked inside `validate_and_extract` block, so we can remove this call.

core/src/sumeragi/message.rs
Erigara: > Only sumeragi knows because validity is relative to the current state.

Yeah and thus it won't be possible to extract `CommittedBlock` inside block sync so we prevent wrong usage of it.

core/src/block.rs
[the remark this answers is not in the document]
SamHSmith: This change has been made now.

core/src/sumeragi/message.rs
SamHSmith: So this conversation can be resolved?

core/src/sumeragi/message.rs
SamHSmith: The genesis wait loop calls this function because it recieves block sync messages.

SamHSmith: Should I create a seperate version of the extract function for use by the genesis code only?

core/src/sumeragi/main_loop.rs
SamHSmith: True! And the function can be removed as well.

core/src/sumeragi/message.rs
SamHSmith: I have now added this.

Erigara (approved):

appetrosyan (approved):

PR #3334, opened 2023-03-28
[refactor] #3238: Optimize WASM triggers with preloading

### LINKED ISSUE #3238: Optimize Triggers execution by storing pre-loaded wasm
It turns out that we can load wasm once and then execute it multiple times much faster, because loading takes a lot of time for `wasmtime`. After loading we get `wasmtime::Module` and that we can store istead of just raw wasm blob.

Blocked by #3106 because this PR introduces `iroha_core::wasm` refactoring which will allow for simple pre-loading and execution.

Arjentix:
## Description

All WASM triggers now are being preloaded on `Register` step. Previously it was done every time trigger was being executed.

Also updated results of queries which return `Trigger`. Now that won't return WASM executable inside trigger, because Iroha doesn't store it anymore.

To be fair, I think we need to benchmark our triggers, because I can't provide any proof right now.

### Linked issue

* Closes #3238 

### Benefits

Triggers work faster

### Checklist


data_model/src/trigger.rs
```diff
@@ -34,100 +34,50 @@ model! {
     #[display(fmt = "@@{id}")]
     #[getset(get = "pub")]
     #[ffi_type]
-    pub struct Trigger<F: Filter> {
+    pub struct Trigger<F: Filter, E = Executable> {
```
mversic, line 37: yet nowhere do you use the implicit default

mversic (changes requested):

data_model/src/trigger.rs
Arjentix: Yeah, will remove it

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -275,6 +279,17 @@ fn unregister_trigger() -> Result<()> {
...
+    let found_action = found_trigger.action;
+    let found_trigger = Trigger::new(
+        found_trigger.id,
+        Action::new(
+            found_action.executable.expect("Already checked"),
+            found_action.repeats,
+            found_action.technical_account,
+            found_action.filter,
+        ),
+    );
```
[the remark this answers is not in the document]
Arjentix: Yes, so the trigger returned by `FindTriggerById` has another type than original *trigger*, because of executable. We now don't return WASM, only instructions if there are some. So at the beginning `found_trigger` has `Option<Vec<Instruction>>` as `executable`. We need to unwrap it to be able to compare triggers

data_model/src/query.rs
```diff
@@ -1005,24 +1005,24 @@ pub mod trigger {
...
         pub struct FindTriggerById {
             /// The Identification of the trigger to be found.
-            pub id: EvaluatesTo<<Trigger<FilterBox> as Identifiable>::Id>,
+            pub id: EvaluatesTo<<Trigger<FilterBox, Executable> as Identifiable>::Id>,
         }
     }
 
     impl Query for FindTriggerById {
-        type Output = Trigger<FilterBox>;
+        type Output = Trigger<FilterBox, Option<Vec<Instruction>>>;
```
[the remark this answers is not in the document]
Arjentix: > It's not as if this information is lost, it's on the blockchain.

I also thought about that and realized, that we respond only with information about current world state. We won't respond with deleted account even if it existed before.

Same for WASM inside triggers. It exists in block history, so technically it's possible to retrieve this info somehow. But our queries works only with current state. And raw WASM is never presented in the state.

I think, we should handle this problem in another way. With some kind of history queries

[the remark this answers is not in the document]
Arjentix: Also in 99% there is no meaning to query WASM, because you probably can't do anythin useful with it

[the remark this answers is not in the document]
Arjentix: And storing useless raw WASM will just take a lot of space. And that only for such strange queires

data_model/src/trigger.rs
```diff
@@ -264,8 +187,61 @@ pub mod action {
...
+        /// Get action metadata
+        fn metadata(&self) -> &Metadata;
+
+        /// Check if action is mintable.
+        fn mintable(&self) -> bool;
+
+        /// Convert action to a boxed representation
+        fn into_boxed(self) -> Action<FilterBox, Self::Executable>;
+
+        /// Same as `into_boxed()` but clones `self`
```
0x009922, line 237:
```suggestion
        /// Same as [`into_boxed`] but clones `self`
```

data_model/src/query.rs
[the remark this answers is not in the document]
0x009922: > Also in 99% there is no meaning to query WASM, because you probably can't do anything useful with it

There is, for Iroha explorer. Although it's not an easy task to display it in a readable way, it is possible to transform it to text and show to an explorer user.

data_model/src/isi.rs
```diff
@@ -869,6 +872,16 @@ pub mod error {
...
+
+        /// Invalid instruction parameter error
+        #[derive(Debug, Display, Clone, PartialEq, Eq)]
+        #[ffi_type(opaque)]
+        #[repr(u8)]
+        pub enum InvalidParameterError {
+            /// Invalid WASM binary
+            #[display(fmt = "Invalid WASM binary: {_0}")]
+            Wasm(String),
+        }
```
0x009922, line 884: Why is this enum-error uses `derive_more::Display` and not `thiserror::Error`? I though the latter is designed specifically to handle enum-errors.

data_model/src/isi.rs
```diff
@@ -342,11 +342,11 @@ isi! {
     #[display(fmt = "EXECUTE `{trigger_id}`")]
     #[serde(transparent)]
     #[repr(transparent)]
-    // SAFETY: `ExecuteTriggerBox` has no trap representation in `Trigger<FilterBox> as Identifiable>::Id`
+    // SAFETY: `ExecuteTriggerBox` has no trap representation in `Trigger<FilterBox, Executable> as Identifiable>::Id`
```
0x009922, line 345: I thought it is a good practice to put **"Safety"** section to doccomments ([reference](https://cheats.rs/#idiomatic-rust)). So, it will be also possible to leverage doc links.

mversic: there is a difference between declaring `unsafe` function/trait and implementing one or writing an `unsafe` block of code.

data_model/src/isi.rs
mversic: because in `no_std` environments you cannot derive `thiserror::Error`

data_model/src/query.rs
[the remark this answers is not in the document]
mversic: > There is, for Iroha explorer. Although it's not an easy task to display it in a readable way, it is possible to transform it to text and show to an explorer user.

is it really useful? because you cannot get rust source code, you can only get wasm binary or `wat` format at best which is hardly usable IMO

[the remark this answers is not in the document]
Erigara: > We won't respond with deleted account even if it existed before.

IMO situation with account is a bit different because account deletion will be on chain.

And in case of trigger difference is that we store trigger differently between block store and world. 

Current return type of query is kinda confusing to me, but searching through blocks or store original wasm doesn't seems to be efficient.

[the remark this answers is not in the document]
Erigara: So should we gather opinions from the SDK devs?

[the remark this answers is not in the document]
Arjentix: After a discussion on todays status we decided to return optimized WASM

data_model/src/isi.rs
Arjentix: @0x009922 , can we close this?

data_model/src/isi.rs
Arjentix: @0x009922 , can we close this?

client/tests/integration/triggers/by_call_trigger.rs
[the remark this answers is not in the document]
Arjentix: @mversic , can we close this?

data_model/src/lib.rs
```diff
@@ -882,6 +883,64 @@ impl TryFrom<Value> for RegistrableBox {
     }
 }
 
+impl TryFrom<Value> for trigger::Trigger<FilterBox, Option<Vec<InstructionBox>>> {
```
Erigara, line 886: It seems like this `impl` is no longer needed and can be removed? (and one below)

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -0,0 +1,644 @@
...
+        }
+    }
+}
+
+impl From<LoadedExecutable> for Executable {
+    fn from(executable: LoadedExecutable) -> Self {
+        match executable {
+            LoadedExecutable::Wasm(module) => Executable::Wasm(WasmSmartContract::new(
+                module
+                    .serialize()
```
Erigara, line 613: We should inform users that obtained bytes are not wasm smartcontract, but wasmtime inner representation of module.

Arjentix: I don't know the best way to do it. Should I introduce a new enum just for that?

Erigara: >  Should I introduce a new enum just for that?

Not sure either. What do you think about returning `Trigger<FilterBox, LoadedExecutable>` we can define `Encode` and `Decode` for `LoadedExecutable`?

Arjentix: I'm afraid it won't work because we have to specify query return type in `iroha_data_model`, but I think we don't want `wasmtime` to be a dependency of `iroha_data_model`

Erigara: Oh, i see, so we don't have that much options. So i think you should add another enum.

Arjentix: Check out new version, please

Erigara (dismissed):

Erigara (approved):

appetrosyan (approved):
Looks good!

PR #3336, opened 2023-03-29
[feature] #3015: Support for niche optimization in FFI

### LINKED ISSUE #3015: Option and niche optimization in FFI
There is no support for converting `Option<T>` for types which have no niche optimization. 

There is some basic support for niche optimizations in FFI, often times inadequate. A good example of this is `NumericValue` which had to be marked as `opaque` because of a very specific issue where `NumericValue` and `&NumericValue` have a niche value but out pointer of `Option<&NumericValue>` doesn't. This is and issue specific for for `NonLocal`s which are `Cloned` (such as `NumericValue`)

mversic:
### Linked issue

Closes #3015 

### Checklist


ffi/src/lib.rs
```diff
@@ -268,98 +266,100 @@ pub enum FfiReturn {
...
+/// }
 ///
-/// // CAUTION: Struct is not robust albeit it's `#[repr(C)]`
-/// // ffi_type! {unsafe impl Robust for NonRobustStruct}
+/// // Validation function is `|_| true` implicitly indicating
+/// // this type is robust with respect to the wrapped type
+/// ffi_type! {
+///     unsafe impl<T> Transparent for Wrapper<T> {
+///         type Target = T;
+///     }
```
mversic, line 290: somehow this test doesn't work

mversic: if someone knows why let me know, if not I'll make a different test

ffi/src/std_impls.rs
```diff
@@ -1,64 +1,65 @@
...
-ffi_type! {impl<K, V> Opaque for BTreeMap<K, V> }
-ffi_type! {impl<K> Opaque for BTreeSet<K> }
+ffi_type! {
+    unsafe impl<'slice> Transparent for &'slice mut str {
+        type Target = &'slice mut [u8];
+
+        validation_fn=unsafe {|target| core::str::from_utf8(target).is_ok()},
+        niche_value=SliceMut::null_mut()
+    }
+}
```
mversic, line 42: this is the best API I could come up with for non-robust transparent items

ffi/src/std_impls.rs
```diff
@@ -1,64 +1,65 @@
...
+ffi_type! {
+    unsafe impl<T> Transparent for core::mem::ManuallyDrop<T> {
+        type Target = T;
+    }
+}
```
mversic, line 47: this is the best API I could come up with for non-robust transparent items

ffi/src/std_impls.rs
```diff
@@ -1,64 +1,65 @@
...
+ffi_type! {
+    unsafe impl<T> Transparent for core::ptr::NonNull<T> {
+        type Target = *mut T;
+
+        validation_fn=unsafe {|target: &*mut T| !target.is_null()},
+        niche_value=core::ptr::null_mut()
+    }
+}
```
mversic, line 55: this is the best API I could come up with for non-robust transparent items. Any ideas?

mversic: The things I dislike:
* code in `validation_fn` is just plugged into `unsafe fn validate` method. This is bad because one can write any unsafe code and it will compile. I've added `validate_fn=unsafe{..}` marker to at least visually mark the unsafety 
* for robust transparent types the `ffi_type`doesn't require `validation_fn` because it's inferred to always be true. This is bad because user can forget to write a validation function if one is required

I've added appropriate description for the macro but I'm still concerned about potential misuses

ffi/tests/transparent.rs
```diff
@@ -31,6 +31,33 @@ pub struct TransparentStruct {
     _zst3: PhantomData<String>,
 }
```
mversic, line 32: Deriving `FfiType` for a transparent item is problematic because either:
1. the user has to write a validation function
2. the user must mark the type as robust 

This API is bad for the following reasons:
* devs often don't understand what `#[ffi_type(unsafe {robust})]` does
* `#[ffi_type(unsafe {robust})]` implements unsafe trait in a macro so the user might be unaware of the safety implications. This is visually mitigated with the use of `unsafe {robust}` in the attribute

When validation function is required the user has to write something like the following impl (taken from this module):
```rs
iroha_ffi::ffi_type! {
    unsafe impl Transparent for NonRobustTransparent {
        type Target = NonRobustTransparentInner;

        validation_fn=unsafe {|target| target != &[0; 4]},
        niche_value=[0; 4]
    }
}
```

I could also require the use to write an impl like this in the robust case instead of using `ffi_type(unsafe {robust})` like so:
```rs
iroha_ffi::ffi_type! {
    unsafe impl Transparent for TransparentStruct {
        type Target = GenericTransparentStruct<()>;
    }
}
```

crypto/src/hash.rs
```diff
@@ -39,11 +39,13 @@ ffi::ffi_item! {
...
+    unsafe impl Transparent for Hash {
+        type Target = [u8; Hash::LENGTH];
 
-impl iroha_ffi::option::Niche for Hash {
-    // NOTE: Any value that has lsb=0 is a niche value
-    const NICHE_VALUE: Self::ReprC = [0; Hash::LENGTH];
+        validation_fn=unsafe {Hash::is_lsb_1},
+        niche_value = [0; Hash::LENGTH]
+    }
 }
```
mversic, line 49: do you think there could be any issue with treating Hash as transmutable into [u8; Hash::LENGTH] considering it's a repr(C) struct with two fields?

I'm slightly concerned about the effects on the calling convention, though probably that's not an issue

data_model/src/block.rs
```diff
@@ -79,7 +81,10 @@ mod committed {
 
     use super::*;
 
-    declare_versioned_with_scale!(VersionedCommittedBlock 1..2, Debug, Clone, FromVariant, IntoSchema, Serialize, Deserialize, PartialEq, Eq, Hash);
+    #[cfg(any(feature = "ffi_import", feature = "ffi_export"))]
```
Arjentix, line 84: As I can see, we have no feature called `ffi_import`

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,182 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
-            if item.variants.is_empty() {
-                abort!(name, "uninhabited enum's cannot be instantiated");
-            }
-
             if is_fieldless_enum(&item) {
+                if item.variants.len() == 1 {
+                    // NOTE: one-variant fieldless enums have representation of ()
+                    return derive_ffi_type_for_opaque_item(name, &input.generics);
+                }
+                if without_repr(&input.attrs) {
```
Arjentix, line 35: You call `find_attr(..., "repr")` before. Could you use info from that function? With that you won't need `without_repr()`

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,182 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
+                    abort!(name, "Fieldless enum missing #[repr(int)]");
+                }
                 if item.variants.iter().any(|v| v.discriminant.is_some()) {
                     abort!(
                         name,
-                        "fieldless enums with explicit discriminants are prohibited"
+                        "Fieldless enums with explicit discriminants are prohibited"
```
Arjentix, line 41: I can't leave this diff comment for a proper line, but here it is:

```rust
if let Some(variant) = item.variants.iter().find(|v| v.discriminant.is_some()) {
    abort!(
        variant,
        "Fieldless enums with explicit discriminants are prohibited",
    )
}
```

I think with that it will show an error for a concrete variant

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,182 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
             }
         }
-        Data::Struct(_) => {
-            if is_owning(&input.attrs, &input.data) {
-                abort!(
-                    name,
-                    "Structure contains raw pointers. If the pointers don't own the data, attach `#[ffi_type(unsafe {non_owning})`"
-                )
-            }
+        Data::Struct(ref item) => {
```
Arjentix, line 51: But you don't use `ref` in `Data::Enum(item)`...

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,182 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
+                        .named
+                        .iter()
+                        .map(|field| &field.ty)
+                        .for_each(add_bound),
+                    syn::Fields::Unnamed(fields) => fields
+                        .unnamed
+                        .iter()
+                        .map(|field| &field.ty)
+                        .for_each(add_bound),
+                    syn::Fields::Unit => {}
```
Arjentix, line 69:
```suggestion
                    syn::Fields::Unit => unreachable!("There can't be any unit-fields in struct")
```

I think it shows better your idea, if I understood it. Without it a reader (me) starts thinking, why do we ingore it

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,182 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
+                }
+
+                derive_unsafe_repr_c(&input.ident, &input.generics)
```
Arjentix, line 72: Looks scary. Does user know about his responsibilities? Or is it totaly safe for the end-user?

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,182 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
-                abort!(
-                    name,
-                    "Structure contains raw pointers. If the pointers don't own the data, attach `#[ffi_type(unsafe {non_owning})`"
-                )
-            }
+        Data::Struct(ref item) => {
+            let ffi_type_impl = derive_ffi_type_for_repr_c(&input);
+
+            let repr_c_impl = {
+                let predicates = &mut input.generics.make_where_clause().predicates;
```
Arjentix, line 55: Weird function to me, but that's on `syn` side, I understand it.
Because `make_` sounds like it returns an object by value. And by reading it looks like you take a mutable reference to a temporary value, which is forbidden in Rust

Arjentix: It's even more weird that `syn` allows you to insert new clauses like that

Arjentix (commented):
Reviewed a couple of files. Will continue later

data_model/src/block.rs
mversic: yes we do, it's just not finished yet so is inert

ffi/derive/src/convert.rs
mversic: is that a problem?

ffi/derive/src/convert.rs
mversic: it is totally safe in this case. At this point struct has `#[repr(C)]` and all it's fields implement `ReprC`

ffi/derive/src/convert.rs
mversic: it's `syn` API, I've used it a lot of times

ffi/src/std_impls.rs
Arjentix: As far as I don't understand a half of code in `ffi` directory, I can't suggest anything meaningful

ffi/tests/transparent.rs
Arjentix: As we disscussed in DM some days ago, we can consider approach with unsafe marker trait:

```rust
#[derive(SomeUnsafeFfiStuff)] // Won't compile without the following impl
pub struct MyStruct { ... }

// SAFETY: I've read the docs, I know what I'm doing
unsafe impl MeaningfulFfiMarker for MyStruct {}
```

Arjentix (commented):
Oof, very hardcore topic

ffi/derive/tests/ui_fail/non_robust_repr_c.stderr
```diff
@@ -0,0 +1,19 @@
+error[E0277]: the trait bound `bool: ReprC` is not satisfied
+ --> tests/ui_fail/non_robust_repr_c.rs:4:38
+  |
+4 | #[derive(Clone, Copy, PartialEq, Eq, FfiType)]
+  |                                      ^^^^^^^ the trait `ReprC` is not implemented for `bool`
```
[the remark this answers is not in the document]
mversic: If you look at docs for `ReprC` you will see it requires for the type to be **robust**, not only `repr(C)`

ffi/derive/tests/ui_fail/non_robust_repr_c.stderr
```diff
@@ -0,0 +1,19 @@
...
+  = help: the following other types implement trait `ReprC`:
+            *const R
+            *mut R
+            FfiTuple10<A, B, C, D, E, F, G, H, I, J>
+            FfiTuple11<A, B, C, D, E, F, G, H, I, J, K>
+            FfiTuple12<A, B, C, D, E, F, G, H, I, J, K, L>
+            FfiTuple1<A>
+            FfiTuple2<A, B>
+            FfiTuple3<A, B, C>
+          and $N others
```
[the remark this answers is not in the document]
mversic: It's also there in std I believe

ffi/derive/src/convert.rs
```diff
@@ -20,140 +27,178 @@ pub fn derive_ffi_type(mut input: DeriveInput) -> TokenStream {
...
-                const NICHE_VALUE: <#inner as iroha_ffi::FfiType>::ReprC = <#inner as iroha_ffi::option::Niche>::NICHE_VALUE;
+        Data::Struct(item) => {
+            if let Some(first_field) = item.fields.iter().next() {
+                &first_field.ty
+            } else {
+                // NOTE: Fieldless structs have representation of ()
+                return derive_ffi_type_for_opaque_item(name, &input.generics);
             }
+        }
+        Data::Union(_) => unreachable!("https://github.com/rust-lang/rust/issues/6040"),
```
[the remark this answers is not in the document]
mversic: it's missing `5`, it should have been `60405`

ffi/derive/src/convert.rs
mversic: ok, made the change. But mind you it wasn't just about style. One arm was capturing by value others by reference

ffi/derive/src/lib.rs
```diff
@@ -276,21 +280,23 @@ pub fn ffi_import(_attr: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 fn is_opaque(input: &syn::DeriveInput) -> bool {
+    let repr = find_attr(&input.attrs, "repr");
```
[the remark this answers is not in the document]
mversic: don't you think optimizing compiler would do that?

ffi/derive/tests/ui_fail/non_robust_repr_c.stderr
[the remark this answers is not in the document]
mversic: what is interesting there?

ffi/derive/tests/ui_fail/non_robust_repr_c.stderr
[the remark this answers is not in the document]
mversic: This library doesn't allow types that have trap representations to cross ffi boundary without checking that bit-value is indeed valid for the given type. If we allowed this it could cause UB. In case of the `bool` type, it only has two valid bit-values in Rust, but all bit-values are valid in C

ffi/derive/src/convert.rs
mversic: You can't write a diverging thing like that without the `ref` capture in the match arm

ffi/derive/tests/ui_fail/non_robust_repr_c.stderr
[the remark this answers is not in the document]
Arjentix: Why it's not a real number, why its `$N`?

ffi/derive/src/lib.rs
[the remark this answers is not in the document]
Arjentix: I don't know actually. If it will be smart enough to understand, that `find_attr()` doesn't produce any side effects, I think. But to me it looks like moving this will be more logical too

ffi/tests/transparent.rs
appetrosyan: IMO having a few example impls outside the macro context is a good idea. While the blanket impl could also serve that purpose and the marker trait(s) could indicate different varieties of robust types. 

I do agree that the documentation on FFI needs to be expanded. With luck, we should be getting another tech writer, and should have a thoroughly documented FFI.

crypto/src/hash.rs
appetrosyan: Worth investigating, but shouldn't hold this PR back. FFI is not deployed and we have the luxury of not using it in production

ffi/src/std_impls.rs
appetrosyan: My 2 cents: 

> ... is bad because one can write any unsafe code and it will compile.

Which is true of most `unsafe`. That's why people are discouraged from using `unsafe` unless they know what they're doing. It's just as bad as `eval`, but `eval` is only a problem when exposed at a prompt. 

We can't prevent all misuse, all we can do is provide safe alternatives, for typical cases. A few pseudo-marker traits (I believe you'd need to specify types) with verified sensible `validate` functions will do the trick. Doubt we can do much beyond that. 

> for robust transparent types the ffi_typedoesn't require validation_fn because it's inferred to always be true. This is bad because user can forget to write a validation function if one is required

As long as we have a tutorial that explains when a type is considered robust, this isn't our responsibility. In fact, I'll ask you this question: _who is the target audience for the FFI library_? People providing dynamic runtimes don't necessarily need to know the ins and outs of the system in order to build up an API. They don't need to know how to build a robust type; using primitive types _only_ is enough, and we have good coverage. We have more than sufficiently many types  to build a robust runtime, with some inefficiencies only apparent after years of usage and deployment in real projects. 

Supporting all possible types is a premature optimisation. But more to the point, we don't necessarily need to provide implementations directly. We have good examples with `ConstString` and a few other primitives, where we demonstrate the worst case scenarios and how to handle them. The people who are to be qualified to interact with the FFI are more than qualified to get what they need by looking at the code and at this point, writing a short tutorial and stabilising the FFI API would be all we need. 

The API you've come up with is reasonably easy to understand, a huge improvement over substrate, and sufficient for our purposes. What more do we need?

appetrosyan (approved):

ffi/derive/tests/ui_fail/non_robust_repr_c.stderr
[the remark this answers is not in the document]
mversic: @Arjentix can this be resolved?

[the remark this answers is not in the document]
mversic: @Arjentix ping

Arjentix (dismissed):

appetrosyan (dismissed):

appetrosyan (approved):

Arjentix (approved):

PR #3340, opened 2023-03-30
[fix] #2581: reduce noise in logs

### LINKED ISSUE #2581: Reduce noise in logs
In a real deployment it is almost impossible to combine health-checks and actual log investigation, because `warp` produces logs regardless of whether or not they're useful. 

In this example, we have two lines taken up by what is effectively noise. 

At a minimum we should be able to reduce the log level of `warp` independently. 

The default should be set to `DEBUG` or better `TRACE` as the information is effectively a filter trace, which isn't particularly useful at our stage of development, but might be useful for debugging purposes.

0x009922:
## Description

I've researched for possible ways on how to configure logging in `warp`. I found that it's not easy.

Provided warp's tracing filter has hard-coded log levels. To implement something similar, but with patched levels, we need to implement `Filter`. This is not possible because [library authors decided](https://github.com/seanmonstar/warp/issues/555#issuecomment-622568143) that this should be kept internally and consumers should build their filters in a "convenient" way with `warp::any()`.

So, we decided to exclude `/health` route from logging at all, as this is the primary issue.

Also, I found that `WasmSmartContract` debug repr was not that is expected, and fixed it.

### Linked issue

Closes #2581  

### Benefits

- less noise in logs
- less noise in TRACE logs (fixed WASM debug repr)

### Checklist

### DIFF
--- a/data_model/src/transaction.rs
+++ b/data_model/src/transaction.rs
@@ -165,7 +165,7 @@ model! {
     ///
     /// Uses **base64** (de-)serialization format.
     #[derive(DebugCustom, Clone, PartialEq, Eq, Hash, Constructor, Decode, Encode, Deserialize, Serialize, IntoSchema)]
-    #[debug(fmt = "WASM binary(len = {})", self.0.len())]
+    #[debug(fmt = "WASM binary(len = {})", "self.0.len()")]
     #[serde(transparent)]
     #[repr(transparent)]
     // SAFETY: `WasmSmartContract` has no trap representation in `Vec<u8>`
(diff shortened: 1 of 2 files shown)


data_model/src/transaction.rs
```diff
@@ -165,7 +165,7 @@ model! {
     ///
     /// Uses **base64** (de-)serialization format.
     #[derive(DebugCustom, Clone, PartialEq, Eq, Hash, Constructor, Decode, Encode, Deserialize, Serialize, IntoSchema)]
-    #[debug(fmt = "WASM binary(len = {})", self.0.len())]
+    #[debug(fmt = "WASM binary(len = {})", "self.0.len()")]
```
mversic, line 168: @Arjentix I believe this is how it should be done

0x009922: Without it it was `[0, 1, 2, 3]` etc with all the bytes inside.

appetrosyan (approved):

PR #3351, opened 2023-04-03
[refactor] #3349: Remove `EvaluateOnHost` from smart contracts

### LINKED ISSUE #3349: Host evaluation is largely redundant for instructions
We should strive to minimize interactions of wasm smart contract and host environment. The reason why a `wasm` smart contract has to call a host defined function to evaluate an expression is because it needs access to the `wsv`. However, of all the implementations of trait `Evaluate` in `core/smartcontracts`, `wsv` is only required for executing an `Expression::Query`.

`EvaluateOnHost` should leverage this by not doing a host call unless absolutely required

mversic:
## Description

Evaluation of expressions on the host is unnecessary so it's been moved into the smart contract itself. `wsv` is only used in one expression evaluation and that is `Expression::Query`. Had we not had this expression, every expression could be evaluated deterministically(whatever the current state of the world may be)

This should not bloat the binary once dynamic linking is in place because evaluation will be taking place in the dynamically linked wasm not the smart contract itself

### Linked issue

Closes #3349

### Benefits

* the less interaction with the host the better the performance

### Checklist


data_model/src/evaluate.rs
```diff
@@ -1,105 +1,107 @@
...
-use super::Evaluate;
-use crate::{prelude::ValidQuery, wsv::WorldStateView};
+/// Context of expression evaluation, holding (name, value) pairs for resolving identifiers
+pub trait Context: Clone {
+    /// Execute query against the current state of `Iroha`
+    ///
+    /// # Errors
+    ///
+    /// If query evaluation/execution fails
+    fn query(&self, query: &QueryBox) -> Result<Value, QueryExecutionFailure>;
```
Arjentix, line 22: Okay, I see, you did it so that we can execute query from WASM without direct wsv access... But it still doesn't make sense to me from logic side of view

wasm/src/lib.rs
```diff
@@ -85,40 +87,33 @@ impl ExecuteOnHost for data_model::query::QueryBox {
...
 
-    fn evaluate(&self) -> Result<Self::Value, Self::Error> {
-        #[cfg(not(test))]
-        use host::evaluate_on_host as host_evaluate_on_host;
-        #[cfg(test)]
-        use tests::_iroha_wasm_evaluate_on_host_mock as host_evaluate_on_host;
+impl iroha_data_model::evaluate::Context for Context {
+    fn query(&self, query: &QueryBox) -> Result<Value, QueryExecutionFailure> {
+        Ok(query.execute())
+    }
```
Arjentix, line 109: I can't see a reason to not to allow user to just call `query.execute()` in its own. So why do we need this?

data_model/src/evaluate.rs
```diff
@@ -1,105 +1,107 @@
...
-            Contains(contains) => contains.evaluate(wsv, context)?,
-            ContainsAll(contains_all) => contains_all.evaluate(wsv, context)?,
-            ContainsAny(contains_any) => contains_any.evaluate(wsv, context)?,
-            Where(where_expression) => where_expression.evaluate(wsv, context)?,
-            ContextValue(context_value) => context_value.evaluate(wsv, context)?,
-            Multiply(multiply) => multiply.evaluate(wsv, context)?,
-            Divide(divide) => divide.evaluate(wsv, context)?,
-            Mod(modulus) => modulus.evaluate(wsv, context)?,
-            RaiseTo(raise_to) => raise_to.evaluate(wsv, context)?,
+            Query(query) => context.query(query)?,
```
mversic, line 72: because you cannot execute query from `data_model`

wasm/src/lib.rs
mversic: look [here](https://github.com/hyperledger/iroha/pull/3351/files#r1157745952)

core/src/smartcontracts/mod.rs
```diff
@@ -48,16 +35,44 @@ pub trait ValidQuery: Query {
...
+            values: BTreeMap::new(),
+            wsv,
+        }
+    }
+}
+
+impl iroha_data_model::evaluate::Context for Context<'_> {
+    fn query(&self, query: &QueryBox) -> Result<Value, QueryExecutionFailure> {
+        query.execute(self.wsv)
+    }
```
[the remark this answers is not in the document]
mversic: I don't think it can be any other way. If this is not ok for you would you like to:

1. rename `Context` to something else?
2 entirely remove `Expression::Query`?

data_model/src/evaluate.rs
```diff
@@ -1,105 +1,107 @@
...
+    ///
+    /// # Errors
+    ///
+    /// If query evaluation/execution fails
+    fn query(&self, query: &QueryBox) -> Result<Value, QueryExecutionFailure>;
+
+    /// Return a reference to the [`Value`] corresponding to the [`Name`].
+    fn get(&self, name: &Name) -> Option<&Value>;
+
+    /// Gets an iterator over the values in this context, in order by [`Name`].
```
QuentinI, line 27: This seems incorrect

QuentinI (changes requested):

core/src/smartcontracts/mod.rs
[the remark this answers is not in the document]
QuentinI: > Context can execute Query? This doesn't make sense to me

I would rephrase it as "context can be queried", which sounds reasonable

data_model/src/evaluate.rs
```diff
@@ -1,105 +1,107 @@
...
     expression::{prelude::*, Expression},
     isi::error::{InstructionExecutionFailure as Error, MathError},
     prelude::*,
-    query::error::FindError,
+    query::error::{FindError, QueryExecutionFailure},
 };
 
-use super::Evaluate;
-use crate::{prelude::ValidQuery, wsv::WorldStateView};
+/// Context of expression evaluation, holding (name, value) pairs for resolving identifiers
```
QuentinI, line 15: Could you maybe expand on this? It's not obvious neither what are the identifiers we're talking about nor why we need them

core/src/smartcontracts/mod.rs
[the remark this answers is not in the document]
mversic: The difference is that Context now includes state of Iroha at the time of executing a query, not only identifiers found in the where clause of the expression.

data_model/src/evaluate.rs
mversic: you are correct, it can be a pain to figure out what `Context` does. I've updated the docs now with a neat example

core/src/smartcontracts/mod.rs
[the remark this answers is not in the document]
mversic: We could evaluate all expressions on the client side had it not been for the `Expression:Query`. Atm, it is the only expressions that needs `wsv`

wasm/src/lib.rs
Arjentix: Okay, makes sense

core/src/smartcontracts/isi/mod.rs
```diff
@@ -267,7 +266,7 @@ impl Execute for SetKeyValueBox {
             IdBox::DomainId(id) => {
                 SetKeyValue::<Domain, Name, Value>::new(id, key, value).execute(authority, wsv)
             }
-            _ => Err(Error::Unsupported(InstructionType::SetKeyValue)),
+            _ => Err(Error::Evaluate(InstructionType::SetKeyValue.into())),
```
Arjentix, line 269: Why so? I think `Unsupported` error clearly shows that this instructions with this parameter types is not supported. I don't quite understand what `Evaluate` error means here. Maybe it least something like `Evaluate(Unsupported(...))`?

data_model/src/isi.rs
```diff
@@ -757,50 +757,63 @@ pub mod error {
...
-            /// Failed to validate.
-            #[display(fmt = "Failed to validate: {_0}")]
-            Validate(#[cfg_attr(feature = "std", source)] ValidationError),
+        }
+
+        /// Type error
+        #[derive(Debug, Display, Clone, PartialEq, Eq, FromVariant)]
+        // TODO: Only temporarily opaque because of problems with FFI
+        #[ffi_type(opaque)]
+        pub enum EvaluationError {
```
Arjentix, line 807: Okay, I see a comment about type here. Maybe rename it somehow? I.e. `TypeEvaluationError`?

data_model/src/evaluate.rs
```diff
@@ -1,299 +1,277 @@
...
+    fn get(&self, name: &Name) -> Option<&Value>;
+
+    /// Update this context with given values.
+    fn update(&mut self, other: impl IntoIterator<Item = (Name, Value)>);
+}
+
+/// Calculate the result of the expression without mutating the state.
+pub trait Evaluate {
+    /// The resulting type of the expression.
+    type Value;
```
Arjentix, line 51: Isn't it always `iroha_data_model::Value`?

permission_validators/account/remove_key_value/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub fn validate(authority: <Account as Identifiable>::Id, instruction: Instructi
     };
 
     let IdBox::AccountId(account_id) = remove_key_value.object_id()
-        .evaluate()
+        .evaluate(&Context::new())
```
Arjentix, line 47: What about:

```rust
pub trait Evaluate {
    // ...
    
    fn evaluate_default<C: Context + Default>(&self) -> Result<Self::Value, Error> {
        Evaluate::evaluate(self, Context::default())
    }
}
```

Arjentix (dismissed):

data_model/src/evaluate.rs
mversic: no, `EvaluatesTo<V>` evaluates to a concrete type `V`

core/src/smartcontracts/isi/mod.rs
mversic: IMO `Unsupported` is a type of evaluation error because your expression failed to produce a valid type. `Evaluate` is about whether the instruction is valid in the sense that there is at least one state of Iroha that it can be executed against. If evaluation fails this means that instruction is not a valid construct according to the specification of Iroha DSL. 

`Validate` on the other hand is a separate step that verifies that a valid instruction can be executed at this time on the current state of Iroha. This is about permissions or invalid operations like divide amount in metadata by 0, etc.

Interestingly, `Expression::Query` is the only expression that requires `wsv`. If we wanted to pre-evaluate instructions we could do so by reducing every expression either to a value or to a `f(query)`. This would bring us a benefit of smaller chain size and less repeated calculation. In this case we can even consider client side evaluation

data_model/src/isi.rs
mversic: I'm closing this comment in favor of the one before

permission_validators/account/remove_key_value/src/lib.rs
mversic: that is a good idea

mversic: it's not possible to implement `Context::default` because it needs to get a reference to `wsv`

QuentinI (approved):

permission_validators/account/remove_key_value/src/lib.rs
Arjentix: Yes, but possible for wasm:: Context, I think. Because it doesn't need wsv. And that would be enough to make use out of this function

config/src/wasm.rs
```diff
@@ -3,7 +3,7 @@
 use iroha_config_base::derive::{Documented, Proxy};
 use serde::{Deserialize, Serialize};
 
-const DEFAULT_FUEL_LIMIT: u64 = 20_000_000;
+const DEFAULT_FUEL_LIMIT: u64 = 25_000_000;
```
Arjentix, line 6: So we do more computations on the WASM side without extra communication with the host. That's why it increased?

permission_validators/account/remove_key_value/src/lib.rs
mversic: mind you, it's part of the trait. I can't implement `Default` for `core::Context`

config/src/wasm.rs
mversic: yes, it should reduce with my next PR where I would reduce computation on wasm side

permission_validators/account/remove_key_value/src/lib.rs
Arjentix: You don't have to make `Default` to be a part of `Context` trait. You can just add this function with additional `Default` bound to the `Evaluate` trait

Arjentix: It will be impossible to call this function for `core::Context` but will be possible for `wasm:: Context`

mversic: let's leave it for my next PR, it'll make a difference there

PR #3356, opened 2023-04-05
[refactor]: Block and sumeragi refactor

### LINKED ISSUE #3397: [bug]: CommittedBlock revalidation miss vote counting

mversic:
## Description

I was poking around the codebase and refactored a few things:
* with `ProofBuilder` we're assured that proof always has at least one signatures
* with `BlockBuilder` only block that is chained and signed is called a block. Intermediate stages are encapsulated with the builder
* the lifecycle of blocks is now `SignedBlock` -> `ValidBlock` -> `CommittedBlock`
* only `VersionedSignedBlock` is allowed across network
* optimized `CommittedBlock` impls of `Eq` and `Ord`
* move `AcceptedTransaction` into `iroha_genesis`
* refined our `Value` enum

### Linked issue

Closes #3397 

### Checklist


core/src/block.rs
```diff
@@ -243,21 +226,52 @@ mod valid {
...
+            match self {
+                Self::V1(v1) => v1,
+            }
+        }
+    }
+
+    /// Signed block
+    #[version_with_scale(n = 1, versioned = "VersionedSignedBlock")]
+    #[derive(Debug, Clone, Decode, Encode)]
     pub struct SignedBlock {
```
Arjentix, line 264: Shouldn't it be renamed as `Signed` like all other `Block`-related stuff in this module?

core/src/block.rs
```diff
@@ -421,109 +435,81 @@ mod signed {
...
+            self.0.as_mut_v1().retain_verified_signatures()
         }
 
         /// Revalidate a block against the current state of the world.
         ///
         /// # Errors
-        /// Forward errors from [`CandidateBlock::revalidate`]
-        #[inline]
+        /// - Block is empty
+        /// - Block has committed transactions
```
Arjentix, line 468: What's wrong with that? I don't get it

core/src/block.rs
```diff
@@ -16,116 +16,172 @@ use eyre::{bail, eyre, Context, Result};
...
+        }
+
+        /// Create [`BlockBuilder`]
         #[inline]
+        // NOTE: Allowed because of assert!
+        #[allow(clippy::missing_panics_doc)]
         pub fn new(
             transactions: Vec<VersionedAcceptedTransaction>,
+            commit_topology: Topology,
             event_recommendations: Vec<Event>,
```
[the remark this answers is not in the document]
mversic: Should it be optional or is it that we just haven't implemented it?

core/src/block.rs
mversic: it goes like this: `BlockBuilder<Pending>` - >`BlockBuilder<Chained>` -> `SignedBlock`

core/src/block.rs
Erigara: Yes, exactly. We want to avoid double-spending and stuff like that.

core/src/tx.rs
```diff
@@ -85,19 +84,18 @@ impl TransactionValidator {
         wsv: &WorldStateView,
     ) -> Result<(), TransactionRejectionReason> {
         for tx in txs {
-            self.validate_internal(tx.into_v1(), true, wsv)?;
+            self.validate_internal::<true>(tx.into_v1(), wsv)?;
```
pesterev, line 87: I know the following isn't related to this PR, but, according to "true" here is hard-coded, it seems the current implementation implies that we should use this function only on genesis. I don't think this is a reliable assumption because nothing prevents to use it on non-genesis. I would add IS_GENESIS here and use the flag hard-coding above on the call-stack where we make sure the peer is genesis.

mversic: that is a good observation

core/src/block.rs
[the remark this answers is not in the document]
mversic: until we implement it, I would leave it as required in the constructor so we're reminded of it. When event recommendations are implemented we can change the behavior

cli/src/lib.rs
```diff
@@ -319,7 +319,7 @@ impl Iroha {
             let wsv_clone = wsv.clone();
 
             transaction_validator
-                .validate_every(genesis.iter().cloned(), &wsv_clone)
+                .validate_every::<true>(genesis.iter().cloned(), &wsv_clone)
```
appetrosyan, line 322: I'm not at all a fan of this API. I'd rather we had a similarly named method in two different traits. The user is left guessing as to what `<true>` means in this context. It was slightly better when we had an enum, but now I'm convinced we really need two trait extension methods and only one in scope.

core/benches/kura.rs
```diff
@@ -42,12 +45,10 @@ async fn measure_block_size_for_n_validators(n_validators: u32) {
         iroha_core::kura::Kura::new(iroha_config::kura::Mode::Strict, dir.path(), false).unwrap();
     let _thread_handle = iroha_core::kura::Kura::start(kura.clone());
 
-    let block = PendingBlock::new(vec![tx], Vec::new())
-        .chain_first()
-        .validate(
-            &TransactionValidator::new(transaction_limits),
-            &Arc::new(WorldStateView::new(World::new(), kura)),
-        );
+    let block = BlockBuilder::new(vec![tx], Topology::new(Vec::new()), Vec::new()).chain_first(
```
appetrosyan, line 48: How does this improve the API?

appetrosyan: I now know less about the meanings of the other objects. Are you sure that positional arguments here make sense?

docs/source/references/schema.json
```diff
@@ -602,14 +602,14 @@
         "name": "header",
         "type": "BlockHeader"
       },
-      {
```
appetrosyan, line 605: Wasn't this supposed to be stable?

data_model/src/transaction.rs
```diff
@@ -330,6 +330,20 @@ impl Transaction for VersionedSignedTransaction {
     }
 }
 
+impl From<ValidTransaction> for SignedTransaction {
+    fn from(source: ValidTransaction) -> Self {
+        let ValidTransaction {
+            payload,
+            signatures,
+        } = source;
```
appetrosyan, line 338: You can (should) do this in the function signature. 

Also, this function should be marked as inline (it's always available thanks to the prelude), and IMO it would be much shorter if you just used `Self {payload: source.payload, signatures: source.signatures.into()`.

data_model/src/block.rs
```diff
@@ -95,10 +95,10 @@ mod committed {
         pub struct CommittedBlock {
             /// Block header
             pub header: BlockHeader,
-            /// Array of rejected transactions.
-            pub rejected_transactions: Vec<VersionedRejectedTransaction>,
             /// array of transactions, which successfully passed validation and consensus step.
             pub transactions: Vec<VersionedValidTransaction>,
+            /// Array of rejected transactions.
```
appetrosyan, line 100: Is this change necessary? IMO it changes the schema for no apparent reason. 

Should we just agree that fields should be sorted alphabetically, to prevent unnecessary changes in the schema?

crypto/src/merkle.rs
```diff
@@ -261,7 +261,7 @@ mod tests {
 
     fn test_hashes(n_hashes: u8) -> Vec<HashOf<()>> {
         (1..=n_hashes)
-            .map(|i| Hash::prehashed([i; Hash::LENGTH]).typed())
+            .map(|i| Hash::prehashed([i; Hash::LENGTH]).typed_unchecked())
```
appetrosyan, line 264: What's your opinion on marking functions like this as `unsafe`? I heard some people complain that it disables optimisations, but I found the opposite to be true, most of the time, Rust takes into account if you have unsafe blocks and it can eliminate more code if it can know that you can't create an invalid object. 

This comes from age-old C++ optimisations where LLVM tracks some invariants if you have a limited set of constructors.

core/src/wsv.rs
```diff
@@ -386,6 +386,7 @@ impl WorldStateView {
     ///
     /// # Errors
     /// - There is no account with such name.
+    // NOTE: Allowed because of assert!
```
appetrosyan, line 389:
```suggestion
    /// # Panics
    /// - if [`Account::add_asset`] fails.   
```

If you're documenting allowing an undocumented panic, you should just document the panic.

core/src/sumeragi/view_change.rs
```diff
@@ -25,64 +25,84 @@ pub enum Error {
...
-        let mut buf = [0_u8; Hash::LENGTH + std::mem::size_of::<u64>()];
-        if let Some(hash) = self.latest_block_hash {
-            buf[..Hash::LENGTH].copy_from_slice(hash.as_ref());
-        }
-        buf[Hash::LENGTH..].copy_from_slice(&self.view_change_index.to_le_bytes());
-        // Now we hash the buffer to produce a payload that is completely
-        // different between view change proofs in the same sumeragi round.
-        Hash::new(buf)
+impl ProofBuilder {
+    /// Create new [`Self`]
```
appetrosyan, line 47:
```suggestion
    /// Constructor from index. 
```

core/src/sumeragi/view_change.rs
```diff
@@ -25,64 +25,84 @@ pub enum Error {
...
+
+    /// Add latest block hash to the proof. This function can be skipped
+    /// only if the genesis block has not yet been committed
+    #[must_use]
+    pub fn with_latest_block_hash(
+        mut self,
+        latest_block_hash: HashOf<VersionedCommittedBlock>,
+    ) -> Self {
+        self.0.latest_block_hash = Some(latest_block_hash);
+        self
```
appetrosyan, line 66: Just a thought, maybe doing it by reference is a good idea?

core/src/sumeragi/view_change.rs
```diff
@@ -25,64 +25,84 @@ pub enum Error {
...
     /// This way peers vote for changing the view (changing the roles of peers).
     ///
     /// # Errors
     /// Can fail during creation of signature
-    pub fn sign(&mut self, key_pair: KeyPair) -> Result<()> {
-        let signature = Signature::new(key_pair, self.signature_payload().as_ref())?;
-        self.signatures.push(signature);
-        Ok(())
+    pub fn sign(mut self, key_pair: KeyPair) -> Result<SignedProof> {
+        let proof_payload = (self.0.latest_block_hash, self.0.view_change_index);
```
appetrosyan, line 75: This should be a fielded (non-tuple) struct at the very least.

core/src/sumeragi/view_change.rs
```diff
@@ -129,7 +149,7 @@ impl ProofChain {
         self.truncate(valid_count);
     }
 
-    /// Attempt to insert a view chain proof into this `ProofChain`.
+    /// Attempt to insert a view change proof into this `ProofChain`.
```
appetrosyan, line 152:
```suggestion
    /// Attempt to insert a view change proof into this [`ProofChain`].
```

core/benches/validation.rs
```diff
@@ -181,46 +194,11 @@ fn sign_blocks(criterion: &mut Criterion) {
     println!("Success count: {success_count}, Failures count: {failures_count}");
 }
 
-fn validate_blocks(criterion: &mut Criterion) {
```
appetrosyan, line 184: I'd create a ticket for a block validation benchmark. We need this one badly. Removing it should not be done without promising to replace it with something.

core/src/block.rs
```diff
@@ -16,155 +16,101 @@ use eyre::{bail, eyre, Context, Result};
...
+#[derive(Debug, Clone, Decode, Encode)]
+pub struct BlockBuilder<B>(B);
+
 mod pending {
-    use super::*;
 
-    declare_versioned_with_scale!(VersionedPendingBlock 1..2, Debug, Clone, iroha_macro::FromVariant);
+    use super::*;
 
     /// Transaction data is permanently recorded in files called
```
appetrosyan, line 41:
```suggestion
    /// First stage in the life-cycle of a [`Block`]. In the beginning the block is assumed to be verified and to contain only accepted transactions. Additionally the block must retain events emitted during the execution of on-chain logic during the previous round, which might then be processed by the trigger system. 
```

core/src/block.rs
```diff
@@ -16,155 +16,101 @@ use eyre::{bail, eyre, Context, Result};
...
-        /// Unix time (in milliseconds) of block forming by a peer.
-        pub timestamp: u128,
-        /// array of transactions, which successfully passed validation and consensus step.
-        pub transactions: Vec<VersionedAcceptedTransaction>,
-        /// Event recommendations.
-        pub event_recommendations: Vec<Event>,
+    /// blocks. This is the first stage of a [`Block`]s life-cycle.
+    #[derive(Debug, Clone)]
+    pub struct Pending {
+        timestamp: u128,
```
appetrosyan, line 45:
```suggestion
        /// Unix timestamp (i.e. count of milliseconds passed since the Unix null date). 
        timestamp: u128,
```

core/src/block.rs
```diff
@@ -16,155 +16,101 @@ use eyre::{bail, eyre, Context, Result};
...
+        transactions: Vec<VersionedAcceptedTransaction>,
```
appetrosyan, line 46:
```suggestion
        /// Collection of transactions which passed consensus and have been accepted. 
        transactions: Vec<VersionedAcceptedTransaction>,
```

core/src/block.rs
```diff
@@ -16,155 +16,101 @@ use eyre::{bail, eyre, Context, Result};
...
+        commit_topology: Topology,
```
appetrosyan, line 47:
```suggestion
        /// The topology at the time of block commit. 
        commit_topology: Topology,
```

core/src/block.rs
```diff
@@ -16,155 +16,101 @@ use eyre::{bail, eyre, Context, Result};
...
+        event_recommendations: Vec<Event>,
```
appetrosyan, line 48:
```suggestion
        /// Event recommendations for use in triggers and off-chain work
        event_recommendations: Vec<Event>,
```

core/src/block.rs
```diff
@@ -16,155 +16,101 @@ use eyre::{bail, eyre, Context, Result};
...
     }
 
-    impl PendingBlock {
-        /// Create a new [`PendingBlock`] from transactions.
+    impl BlockBuilder<Pending> {
+        /// Create [`BlockBuilder`]
         #[inline]
+        // NOTE: Allowed because of assert!
+        #[allow(clippy::missing_panics_doc)]
```
appetrosyan, line 55: Document the panic instead.

appetrosyan (changes requested):

core/src/sumeragi/main_loop.rs
```diff
@@ -780,40 +782,49 @@ fn process_message_independent<F: FaultInjection>(
...
+
+                                        sumeragi.broadcast_packet(msg, current_topology);
+                                        commit_block(sumeragi, state, committed_block);
+                                    }
+                                    Err(err) => error!(%addr, role=%Role::Leader, ?err),
+                                }
                             }
-                            Err(err) => error!(%addr, role=%Role::Leader, ?err),
                         }
+                        Err(error) => error!(%error, "Failed to sign block"),
```
appetrosyan, line 827:
```suggestion
                        Err(error) => error!(?error, "Failed to sign block"),
```

core/src/sumeragi/main_loop.rs
```diff
@@ -1176,36 +1210,23 @@ fn sumeragi_init_commit_genesis<F: FaultInjection>(
...
+        Ok(block) => {
+            info!(%addr, hash=%block.hash(), "Genesis block created");
+
+            sumeragi.send_events(&block);
+            let msg = MessagePacket::new(ProofChain::default(), BlockCreated::from(block.clone()));
+            sumeragi.broadcast_packet(msg, &state.current_topology);
+            // Omit signature verification during genesis round
+            commit_block(sumeragi, state, block.commit_unchecked());
+        }
+        Err(error) => error!(%error, "Failed to sign block"),
```
appetrosyan, line 1229:
```suggestion
        Err(error) => error!(?error, "Failed to sign block"),
```

core/benches/kura.rs
mversic: > I now know less about the meanings of the other objects.

What information that was obvious to you in the previous version of the code are you now missing? I don't see much difference with respect to the readability

I suppose I had a different perspective in mind. With builders I'm differentiating between required and optional arguments. To create a block you always have to provide list of transactions, a topology and a list of event recommendations, then you chain the block (either as genesis or with the previous block) and then you sign it. At this point I say you have constructed a block.

>  Are you sure that positional arguments here make sense?

can I get a recommendation of what you would do here?

docs/source/references/schema.json
mversic: it's the order of fields in the struct you are referring to? Should I revert it?

data_model/src/block.rs
mversic: ok, reverting this. It's not relevant

mversic: > Should we just agree that fields should be sorted alphabetically, to prevent unnecessary changes in the schema?

I don't support this idea

mversic: > Is this change necessary? IMO it changes the schema for no apparent reason.

reason was to make it more in line with order of fields in the `BlockHeader`

core/src/sumeragi/main_loop.rs
```diff
@@ -517,32 +517,35 @@ fn suggest_view_change<F: FaultInjection>(
...
+    match view_change_proof {
+        Ok(proof) => {
+            view_change_proof_chain
+                .insert_proof(
+                    &state.current_topology.sorted_peers,
+                    state.current_topology.max_faults(),
+                    state.latest_block_hash,
+                    proof,
+                )
+                .unwrap_or_else(|err| error!("{err}"));
```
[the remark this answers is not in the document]
mversic: ofc, it was a lapse

core/benches/validation.rs
mversic: this is now part of `chain_blocks` benchmark. Block is validated when it is being chained. This step cannot be tested separately

data_model/src/block.rs
mversic: I can revert it, but considering recent schema changes I thought it might not be problematic

crypto/src/merkle.rs
Arjentix: I agree with @mversic , `unsafe` should stay for memory and UB issues while business logic should be marked on another level. `unchecked` is fine for me

core/src/block.rs
```diff
@@ -551,67 +548,67 @@ mod candidate {
...
             rejected_transactions
                 .iter()
-                .map(VersionedSignedTransaction::hash)
+                .map(VersionedRejectedTransaction::hash)
                 .collect::<MerkleTree<_>>()
                 .hash()
                 .eq(&header.rejected_transactions_hash)
                 .then_some(())
-                .ok_or_else(|| eyre!("The hash of a rejected transaction stored in the block header does not match the actual hash or this transaction."))?;
+                .ok_or_else(|| eyre!("Tash of a rejected transaction stored in the block header does not match the actual hash or this transaction."))?;
```
Arjentix, line 595: Funny

core/benches/validation.rs
appetrosyan: This could be a problem.

core/src/block.rs
```diff
@@ -664,149 +661,105 @@ mod candidate {
...
     }
 
-    impl CandidateCommittedBlock {
+    impl Candidate<CommittedBlock> {
         /// Calculate the hash of the current block.
         /// [`CommittedBlock`] should have the same hash as [`ValidBlock`].
         #[inline]
         pub fn hash(&self) -> HashOf<Self> {
-            HashOf::new(&self.header).transmute()
+            HashOf::new(&self.0.header).transmute()
```
Arjentix, line 752: Same with `typed_unchecked()`

Arjentix (dismissed):

data_model/src/block.rs
appetrosyan: We should not be changing the order of fields in the schema willy nilly. So either we impose an order and generate the schema exactly as seen in the struct, or we decouple the order of fields in the schema from the order of field definitions in the struct. 

Refactors should never break the API.

core/src/block.rs
```diff
@@ -175,66 +131,104 @@ mod chained {
...
+    impl BlockBuilder<Chained> {
         /// Sign this block and get [`SignedBlock`].
         ///
         /// # Errors
+        ///
         /// Fails if signature generation fails
         pub fn sign(self, key_pair: KeyPair) -> Result<SignedBlock> {
-            let signature = SignatureOf::from_hash(key_pair, &self.hash().transmute())
-                .wrap_err(format!("Failed to sign block with hash {}", self.hash()))?;
+            let hash = HashOf::new(&self.0.header).transmute();
```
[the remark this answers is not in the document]
mversic: no because then it would have to be first encoded like this:
```suggestion
            let hash = Hash::new(&self.0.header.encode()).typed_unchecked();
```

core/benches/kura.rs
appetrosyan: > I don't see much difference with respect to the readability

Why change something if it doesn't improve readability. Especially if something close to it could. 

> What information that was obvious to you in the previous version of the code are you now missing?

`PendingBlock` is constructed. 

In the new version, `Topology` is explicitly constructed from a sensible default, which raises the question, why wasn't it implicit (as before). 

> With builders ...

Which would look more like 

```suggestion
    PendingBlockBuilder::new(vec![tx])
      .topology(Vec::new().into())
      .events(Vec::new())
      .build()?
      .chain_first(
```

> ... I'm differentiating between required and optional arguments (that have some sensible default)

By that definition only one argument is required: the topology in this example is constructed from an empty vector (which can be considered its `Default`) and event recommendations are empty, which is also a natural and sensible default. Given you already have builders and a run-time check to convert to a block, the following would be a viable API: 

```rust
    PendingBlockBuilder{
      txn: vec![tx], 
      ..Default::default()
      // topology: Topology::new(Vec::new()), // implied
      // events: Vec::new(),  // implied
    }
    .check()? 
    .chain_first(
```
Which could make things a bit more readable. As is, I don't see why it couldn't have also been 

```suggestion
    pending_block(vec![tx], vec![].into(), Vec::new()).chain_first(
```

or better yet: 

```suggestion
    PendingBlock::from(tx).chain_first(
```

## Disclaimer

This is a very minor point. 

I'd rather we merged the code with this minor annoyance to be fixed later (or never) than spend another day with 17+ unmerged PRs.

appetrosyan (approved):

core/src/block.rs
```diff
@@ -12,594 +11,826 @@
...
+                            signatures: block.signatures.transmute(),
+                        }),
+                        error,
+                    )
+                })?;
+
+            if IS_GENESIS {
+                Ok(block.commit_unchecked())
+            } else {
+                block.commit(current_topology).map_err(|(block, error)| {
```
mversic, line 800: I think that #3247 has accidentally removed verification that the committed block has required number of signatures

core/src/sumeragi/main_loop.rs
```diff
@@ -1039,21 +1041,31 @@ pub(crate) fn run(
...
+        } else {
+            process_message_independent::<false>(
+                sumeragi,
+                &mut state,
+                &mut voting_block,
+                current_view_change_index,
+                &mut view_change_proof_chain,
+                &round_start_time,
+            );
+        }
```
Erigara, line 1062: I think `is_genesis_peer` variable confused you here. It's used only for tests in soft fork detection. We shouldn't apply genesis round rules for such peer.

Erigara: It should be named something like `peer_with_soft_fork`.

mversic: is it better now?

core/src/sumeragi/main_loop.rs
```diff
@@ -535,14 +532,14 @@ fn enqueue_transaction(sumeragi: &Sumeragi, wsv: &WorldStateView, tx: VersionedS
 }
 
 #[allow(clippy::too_many_lines)]
-fn handle_message(
+fn handle_message<const IS_GENESIS: bool>(
```
Erigara, line 535: I think this function shouldn't have const generic as it's called only in non-genesis rounds.

core/src/sumeragi/main_loop.rs
Erigara: Yes, feel free to resolve.

core/src/sumeragi/main_loop.rs
mversic: what about [this](https://github.com/hyperledger/iroha/pull/3356#discussion_r1160039422)

Erigara: Sounds reasonable to me.

core/src/block.rs
Erigara: I created issue for this yesterday, assign yourself if you will fix this in current PR. #3397

appetrosyan (approved):

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -357,7 +357,7 @@ fn trigger_in_genesis_using_base64() -> Result<()> {
 
     // Registering trigger in genesis
     let mut genesis = GenesisNetwork::test(true).expect("Expected genesis");
-    match &mut genesis.transactions[0].as_mut_v1().payload.instructions {
+    match &mut genesis.transactions[0].as_mut_v1().0.payload.instructions {
```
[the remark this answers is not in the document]
mversic: > I see many places where we use this as_mut_v1().0.payload boilerplate and also know that we have an immutable method to retrieve payload

we shouldn't be having mutable getters but I didn't want to go into handling this since we're moving to python based integration tests. Getters are there only to opacify our data model structs and those structs are immutable from the perspective of the end user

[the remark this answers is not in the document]
pesterev: Got you

appetrosyan (approved):

data_model/src/block.rs
```diff
@@ -136,24 +131,21 @@ mod committed {
         /// `VersionedCommitedBlock` should have the same hash as `VersionedCommitedBlock`.
         #[cfg(feature = "std")]
         #[inline]
-        pub fn hash(&self) -> HashOf<Self> {
-            self.as_v1().hash().transmute()
+        pub fn hash(&self) -> HashOf<BlockPayload> {
```
mversic, line 134: pay attention to this change, the same is done for transactions. This makes it possible to interchangeably use hashes for different blocks/transactions without using `transmute`

data_model/src/lib.rs
```diff
@@ -611,19 +611,16 @@ model! {
...
         #[serde_partially_tagged(untagged)]
         Id(IdBox),
         #[serde_partially_tagged(untagged)]
         Identifiable(IdentifiableBox),
         PublicKey(PublicKey),
         SignatureCheckCondition(SignatureCheckCondition),
-        TransactionValue(TransactionValue),
         TransactionQueryResult(TransactionQueryResult),
         PermissionToken(permission::PermissionToken),
-        Hash(Hash),
```
mversic, line 626: hash is now typed in `Value` enum

data_model/src/lib.rs
```diff
@@ -611,19 +611,16 @@ model! {
...
-        TransactionLimits(transaction::TransactionLimits),
-        LengthLimits(LengthLimits),
+        Limits(LimitsValue),
         #[serde_partially_tagged(untagged)]
         Id(IdBox),
         #[serde_partially_tagged(untagged)]
         Identifiable(IdentifiableBox),
         PublicKey(PublicKey),
         SignatureCheckCondition(SignatureCheckCondition),
-        TransactionValue(TransactionValue),
```
mversic, line 623: removed as redundant, using `TransactionQueryResult` instead

data_model/src/lib.rs
```diff
@@ -611,19 +611,16 @@ model! {
         Name(Name),
         Vec(#[skip_from] #[skip_try_from] Vec<Value>),
         LimitedMetadata(metadata::Metadata),
-        MetadataLimits(metadata::Limits),
-        TransactionLimits(transaction::TransactionLimits),
-        LengthLimits(LengthLimits),
```
mversic, line 616: grouped under one entry

crypto/src/signature.rs
```diff
@@ -179,6 +177,12 @@ impl<T> Ord for SignatureOf<T> {
     }
 }
 
+impl<T> core::hash::Hash for SignatureOf<T> {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+    }
+}
```
Arjentix, line 184: Did `PhantomData<T>` affect hash somehow?

data_model/src/query.rs
```diff
@@ -1156,16 +1156,16 @@ pub mod transaction {
         #[derive(Display)]
         #[display(fmt = "Find transaction with `{hash}` hash")]
         #[repr(transparent)]
-        // SAFETY: `FindTransactionByHash` has no trap representation in `EvaluatesTo<Hash>`
+        // SAFETY: `FindTransactionByHash` has no trap representation in `EvaluatesTo<HashOf<TransactionPayload>>`
```
Arjentix, line 1159:
```suggestion
        // SAFETY: [`FindTransactionByHash`] has no trap representation in `EvaluatesTo<HashOf<TransactionPayload>>`
```

data_model/src/lib.rs
```diff
@@ -651,6 +648,28 @@ model! {
...
+        /// Block hash
+        Block(HashOf<BlockPayload>),
+        /// Transaction hash
+        Transaction(HashOf<TransactionPayload>),
+    }
+
+    /// Enum for all supported hash types
+    #[derive(DebugCustom, Display, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, FromVariant, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+    #[ffi_type]
+    pub enum LimitsValue {
```
Arjentix, line 665: Copy-pasted docs

crypto/src/signature.rs
mversic: no, it was to remove the lint `derive_hash_xor_eq`

data_model/src/query.rs
```diff
@@ -1238,11 +1238,11 @@ pub mod block {
         #[derive(Display)]
         #[display(fmt = "Find block header with `{hash}` hash")]
         #[repr(transparent)]
-        // SAFETY: `FindBlockHeaderByHash` has no trap representation in `EvaluatesTo<Hash>`
+        // SAFETY: `FindBlockHeaderByHash` has no trap representation in `EvaluatesTo<HashOf<BlockPayload>>`
```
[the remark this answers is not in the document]
mversic: this is // comment so it doesn't make sense to me

core/src/block.rs
```diff
@@ -12,594 +11,798 @@
...
-            event_recommendations: self.event_recommendations,
-            signatures,
+        /// Chain the block with existing blockchain.
+        pub fn chain(
+            self,
+            height: u64,
+            previous_block_hash: Option<HashOf<BlockPayload>>,
+            view_change_index: u64,
+            transaction_validator: &TransactionValidator,
+            wsv: WorldStateView,
```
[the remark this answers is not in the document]
mversic: what problem?

core/src/block.rs
```diff
@@ -12,594 +11,798 @@
...
-                }
+        fn make_header(
+            timestamp_ms: u128,
+            height: u64,
+            previous_block_hash: Option<HashOf<BlockPayload>>,
+            view_change_index: u64,
+            transactions: &[VersionedValidTransaction],
+            rejected_transactions: &[VersionedRejectedTransaction],
+            commit_topology: Topology,
+        ) -> BlockHeader {
```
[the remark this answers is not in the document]
mversic: `BlockHeaderBuilder` would indicate that there are some optional fields but, except for the genesis block, there are none

core/src/block.rs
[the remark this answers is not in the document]
Arjentix: That's unclear from usage side what two integers are doing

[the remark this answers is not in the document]
mversic: why do you think it's unclear? can you suggest something here?

data_model/src/block.rs
appetrosyan: Clever

data_model/src/lib.rs
appetrosyan: Relevant for #3418

core/src/block.rs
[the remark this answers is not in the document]
mversic: I tried, but I still think this is useless. I create `RawHeader` only to have it destructured into `BlockHeader` in the next step. Because `make_header` takes arguments and assembles them into `BlockHeader`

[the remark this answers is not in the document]
Arjentix: Ok, I don't insist. That won't be used in many places so the current version is ok for me

core/src/block.rs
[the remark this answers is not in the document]
Arjentix: I was about to suggest the same idea with `RawHeder` but I think it's okay like this too

Arjentix (dismissed):

Arjentix (dismissed):

appetrosyan (dismissed):

Arjentix:
I'll approve when all tests pass

core/src/sumeragi/main_loop.rs
```diff
@@ -517,90 +504,77 @@ fn handle_message(
...
+                    {
+                        debug!(%addr, %role, %block_hash, ?error, "Block appears to be invalid. Revalidating...");
+
+                        /* If the block fails validation we must check again using the finaziled wsv.
+                        When a soft-fork occurs the consensus-block may be valid on the previous
+                        wsv but not the current one. */
+                        match block.validate(
+                            state.latest_block_height.saturating_sub(1),
+                            state.previous_block_hash,
+                            &state.current_topology,
```
Erigara, line 536: I think in case of soft-fork revalidation we should pass not `current_topology`, but `previous_topology`. Suppose that wrong block added some new peers.
I can try to address this in my #3405.

core/src/block.rs
```diff
@@ -12,594 +11,871 @@
...
+                return Err((self, msg));
+            }
+
+            if !self.header().is_genesis()
+                && topology
+                    .filter_signatures_by_roles(&[Role::Leader], self.signatures())
+                    .is_empty()
+            {
+                return Err((self, eyre!("Block is not signed by the leader")));
+            }
```
Erigara, line 391: Shouldn't this check be part of `commit` method?

core/src/lib.rs
```diff
@@ -51,6 +51,12 @@ pub type PermissionTokenDefinitionsMap =
 /// Type of `Sender<Event>` which should be used for channels of `Event` messages.
 pub type EventsSender = broadcast::Sender<Event>;
 
+/// Implemented for entities that produce events
+pub trait EventProducer {
+    /// Produce events.
+    fn produce_events(&self) -> Vec<Event>;
```
Erigara, line 57: Cool, `Into<Vec<Event>>` was a bit weird.

core/src/queue.rs
```diff
@@ -171,11 +171,12 @@ impl Queue {
...
                     Entry::Occupied(mut old_tx) => {
                         // MST case
+                        // TODO: Unnecessary validation of signatures happens here
                         old_tx
                             .get_mut()
-                            .as_mut_v1()
-                            .signatures
-                            .extend(tx.as_v1().signatures.clone());
+                            .add_signatures(tx.into_v1().0.signatures)
+                            .expect("Valid");
```
Erigara, line 178: To be honest `expect("Valid")` doesn't tell us more then just `unwrap()`.

core/src/sumeragi/message.rs
```diff
@@ -165,40 +145,18 @@ impl From<VersionedCommittedBlock> for BlockCommitted {
...
-    }
-    /// Get hash of block.
-    pub fn hash(&self) -> HashOf<VersionedCommittedBlock> {
-        self.block.hash()
-    }
-}
+    pub block: VersionedSignedBlock,
+}
+
+//impl From<VersionedCommittedBlock> for BlockSyncUpdate {
```
Erigara, line 151: Commented code.

core/src/sumeragi/main_loop.rs
```diff
@@ -247,70 +245,48 @@ impl Sumeragi {
...
+                                continue;
+                            }
+                        },
+                        Err((block, err)) => {
+                            warn!(%addr, hash=%block.hash(), ?err, "Failed to commit genesis block");
+                            continue;
+                        }
+                    };
+
                     if block.header().is_genesis() {
```
Erigara, line 290: Imo we should check this before start validation.

core/src/sumeragi/main_loop.rs
```diff
@@ -340,18 +314,33 @@ fn commit_block(sumeragi: &Sumeragi, state: &mut State, block: impl Into<Version
...
+    );
+
+    // TODO: It should not be ok to commit without proxy tail signature here?
+    let committed_block = match block.commit_without_proxy_tail_signature(&state.current_topology) {
+        Ok(block) => block,
+        Err((_, err)) => {
+            error!(?err, %block_hash, "Failed to commit replacement block. Unable to resolve soft fork.");
+            return;
+        }
+    };
```
Erigara, line 343: I don't think this is correct. 

First of all why would we check `Leader` signature is present, but skip `ProxyTail` signature. 
Since keep track of all topology changes, i think we should just check that block has enough valid signatures.
Otherwise we can try to recreate proper topology from initial topology for this round + amount of view changes (it should be possible if randomness is not involved) and then do proper validation.

Second is that we are using **current** topology when we commit **previous** block.

core/src/sumeragi/main_loop.rs
mversic: I wanted to ask you to specifically review soft fork resolution code

core/src/block.rs
mversic: no, every block should have leader signature. It will be rejected before commit if it doesn't have it. On `commit` it's guaranteed that the block has it, otherwise it's impossible to create `ValidBlock`

core/src/lib.rs
mversic: and the code is moved from `data_model` into `iroha_core`

core/src/sumeragi/main_loop.rs
```diff
@@ -517,90 +505,82 @@ fn handle_message(
...
+                state.wsv.clone(),
+            ) {
                 Ok(block) => block,
-                Err(error) => {
-                    error!(%addr, %role, %block_hash, ?error, "Block not valid.");
-                    return;
+                Err((block, error)) => {
+                    let header = block.header();
+
+                    // TODO: What is the correct condition for soft-fork?
```
Erigara, line 522: I think correct condition is when:
1) block pass validation against `finalized_wsv` (this will include height check, hash check)
2) block `view_change_index` is greater then currently committed one.

core/src/sumeragi/main_loop.rs
```diff
@@ -517,90 +505,82 @@ fn handle_message(
...
+                    #[allow(clippy::redundant_else)]
+                    if state.latest_block_height == header.height
+                        && state.previous_block_hash == header.previous_block_hash
+                        && state.latest_block_hash != Some(block_hash)
```
Erigara, line 526: This check seems redundant because we check that `state.latest_block_view_change_index < header.view_change_index` so in such case hashes can't be equal.

core/src/block.rs
Erigara: I just think about of block sync.
Consider situation when peer is behind for some reason (it's current `view_change_index` is 1 for example) and rest of the network committed block (at `view_change_index` 10).
In such case it might be possible that `Leader at view_change_index 2 != Leader at view_change_index 10` and slow peer will reject such block.

core/src/sumeragi/main_loop.rs
mversic: > I don't think this is correct.

I know, but the tests are failing if I change it to `.commit`, that is why I put the `TODO`. I think every block should have both `Leader` and `ProxyTail` signatures.

core/src/sumeragi/main_loop.rs
mversic: is there is a slight chance they will be equal for different output?

appetrosyan: > I hope that blake2 protects us from collisions. 

It's the best that we have at the moment. Definitely better than SHA-3 in terms of collision resistance (according to [Blake 2 spec](https://www.blake2.net/blake2.pdf)). 

More important question is what can we do about collisions. At this point switching over to any other cryptographic hash would serve to pessimise performance. We **could** come up with a perfect hash but we'd need to verify that it is perfect, and that would involve at least one cryptographer and would be an endless source of problems. Re-using signatures is well... an attack vector we can't really do anything about. 

At the same time, in order for this to be a likely attack vector, you'd need multiple blocks with sensible data to come up with the same hash. I.e. you shouldn't worry about the situation in which the attacker can find one block with the right hash and useful state transitions, but any arbitrary number of blocks and an algorithm to generate them. At this point you have bigger problems than just consensus not working. That's why cryptographic libraries are almost always designed to be as easy to hotpatch as possible, these cryptographic hashes develop a correlation and sometimes these correlations can be exploited. 

Best we can do then, is patch it. Not much use worrying about it in today's code.

data_model/src/block.rs
```diff
@@ -10,274 +10,281 @@ use core::{cmp::Ordering, fmt::Display};
...
-        use super::*;
+    #[derive(
+        Debug, Display, Clone, Eq, Getters, Decode, Encode, Deserialize, Serialize, IntoSchema,
+    )]
+    #[display(fmt = "({header})")]
+    #[getset(get = "pub")]
+    #[allow(missing_docs)]
+    #[ffi_type]
+    pub struct BlockPayload {
+        /// Block header
```
SamHSmith, line 103: What is the motivation for this + the block trait? Seems more complicated in my opinion.

core/src/sumeragi/main_loop.rs
SamHSmith: I can confirm this. When two blocks fit the same slot in the chain, that is a soft-fork and we revert and use the one with the highest view change index.

core/src/sumeragi/main_loop.rs
```diff
@@ -516,90 +513,83 @@ fn handle_message(
...
+                        && state.latest_block_view_change_index < header.view_change_index
+                    {
+                        debug!(%addr, %role, %block_hash, ?error, "Block appears to be invalid. Revalidating...");
+
+                        /* If the block fails validation we must check again using the finaziled wsv.
+                        When a soft-fork occurs the consensus-block may be valid on the previous
+                        wsv but not the current one. */
+                        match ValidBlock::validate_without_validating_topology(
+                            block,
```
SamHSmith, line 544: Why are we not validating the topology?

core/src/sumeragi/main_loop.rs
```diff
@@ -516,90 +513,83 @@ fn handle_message(
...
-                    actual = block.header().height,
-                    expected = state.latest_block_height + 1,
-                    "Mismatch between the actual and expected height of the block."
-                );
-                return;
-            }
-
-            commit_block(sumeragi, state, block);
+            match block.commit_without_proxy_tail_signature(current_topology) {
+                Ok(committed_block) => commit_block(sumeragi, state, committed_block),
```
SamHSmith, line 567: Without proxy tail signature??

SamHSmith (changes requested):
Maybe some of the refactors in this PR could be split up? We could have a more productive conversation then about what to change / keep.

core/src/block.rs
```diff
@@ -1,605 +1,701 @@
...
+    }
 
-                Ok(())
-            }
+    impl CommittedBlock {
+        #[deprecated(
+            since = "2.0.0-pre-rc.13",
+            note = "This method exists only because some tests are failing, but it shouldn't"
+        )]
+        pub(crate) fn commit_without_validation(
```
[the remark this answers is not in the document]
mversic: It is only used for genesis after reading blocks from `kura`. I just left it the way it was previously because I don't understand the logic there

core/src/sumeragi/main_loop.rs
Arjentix: Shouldn't we add a test for that case? Maybe in another PR.

Or there was one test failing because of that?

Erigara: @Arjentix i had some tests for soft fork detection in #3405, but i think we still need to cover case with topology.

core/src/block.rs
```diff
@@ -1,605 +1,704 @@
...
+                            let tx_result = if block.header().is_genesis() {
+                                AcceptedTransaction::accept::<true>(tx, limits)
+                            } else {
+                                AcceptedTransaction::accept::<false>(tx, limits)
+                            };
+
+                            match tx_result {
+                                Ok(tx) => Ok(tx),
+                                Err((_tx, err)) => Err(err).wrap_err("Failed to accept transaction")
+                            }
```
Arjentix, line 377: Maybe replace with `map_err()`?

core/src/kura.rs
```diff
@@ -100,7 +95,12 @@ impl Kura {
         });
 
         let shutdown = move || {
-            let _result = shutdown_sender.send(());
+            if let Err(error) = shutdown_sender.send(()) {
+                iroha_logger::error!(
+                    ?error,
+                    "Failed to send shut down signal to kura. Thead might already be shut down."
```
Arjentix, line 101:
```suggestion
                    "Failed to send shut down signal to kura. Thread might already be shut down."
```

core/src/kura.rs
```diff
@@ -123,15 +123,15 @@ impl Kura {
...
         for block in block_indices {
             // This is re-allocated every iteration. This could cause a problem.
             let mut block_data_buffer = vec![0_u8; block.length.try_into()?];
 
             match block_store.read_block_data(block.start, &mut block_data_buffer) {
-                Ok(_) => match VersionedCommittedBlock::decode_all_versioned(&block_data_buffer) {
+                Ok(_) => match VersionedSignedBlock::decode_all_versioned(&block_data_buffer) {
                     Ok(decoded_block) => {
-                        block_hashes.push(decoded_block.hash());
+                        block_hashes.push(Block::hash(&decoded_block));
```
Arjentix, line 134: Is it a style preference or does this affect something?

data_model/src/query.rs
```diff
@@ -1150,7 +1150,7 @@ pub mod transaction {
 
     impl FindTransactionByHash {
         ///Construct [`FindTransactionByHash`].
```
Arjentix, line 1152:
```suggestion
        /// Construct [`FindTransactionByHash`].
```

core/src/validator.rs
```diff
@@ -71,9 +71,9 @@ impl Validator {
         &self,
         wsv: &WorldStateView,
         authority: &<Account as Identifiable>::Id,
-        operation: impl Into<data_model_validator::NeedsValidationBox>,
+        operation: data_model_validator::NeedsValidationBox,
     ) -> Result<()> {
-        let operation = operation.into();
+        let operation = operation;
```
Arjentix, line 76: This line is useless now. Also what's the problem of `impl Into` signature?

data_model/src/account.rs
```diff
@@ -163,22 +163,6 @@ pub mod model {
...
 
-impl AccountId {
-    #[cfg(feature = "transparent_api")]
-    const GENESIS_ACCOUNT_NAME: &str = "genesis";
-
-    /// Construct [`Id`] of the genesis account.
-    #[inline]
-    #[must_use]
-    #[cfg(feature = "transparent_api")]
-    pub fn genesis() -> Self {
```
Arjentix, line 174: How can we construct genesis account now?

client/tests/integration/unstable_network.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::restriction)]
+#![cfg(debug_assertions)]
```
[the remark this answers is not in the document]
mversic: because there is some config parameter on Iroha that has `cfg(debug_assertions)`. And when I ran tests with `--release` it didn't compile. No, I don't think we ever run tests with `--release`

data_model/src/account.rs
mversic: it's now a constant `GENESIS_ACCOUNT` and is stored in `iroha_genesis`

mversic:
I will close this because I'm splitting it out in multiple PRs

Issue #3130, opened 2023-02-09
[suggestion] Fully asynchronous client API


appetrosyan:
### Feature request

I propose including a version of the API calls which is fully asynchronous and consider further extending the scope of the issue to include generic HTTP transport. 

Specifically we want to ensure most implementations don't need to provide an `async_client` implementation (c.f. https://github.com/soramitsu/iroha2-longevity-load-rs/tree/iroha2-dev/src/async_client), and can instead specify their preferred dependencies in feature flags and rely on the provided library. 

Alternatively, a feasible solution would be to fully generify the API and provide specific implementations as separate crates, e.g. `iroha_client_awc` and/or `iroha_client_reqwest` as plugins. 

### Motivation

#3117 shows a method of obtaining cancel-able thread-based transaction submission. However, the API chosen for the initial implementation results in a counter-intuitive API, one which most of our users are likely to ignore. 

More importantly, field testing showed that the use cases in which threading would offer advantages are few and far between. Our original assumption about the workflow being mostly thread based turned out to be completely wrong as evidenced by the inclusion of an `async_client` in the blockchain explorer, longevity load script as well as several internal projects. 

It should also be considered a performance optimisation, given that threading is likely to not fully utilise the threads given the usage pattern. 

### Who can help?

@pesterev @appetrosyan @0x009922 

## Meeting minutes for Feb. 10. 

Transport agnosticism not as important as assumed, hard-coding `awc` has advantages. Code size impact is minimal. 
- [ ] Make `iroha_client` a lightweight API crate, 
- [ ] Use a separate `iroha_client_awc` which implements the transport logic to be used in e.g. `iroha_client_cli`. 
- [ ] Update `iroha_client_cli` to use new crate, 
- [ ] Update `longevity-load-rs` to use new crate, 
- [ ] (Optional) update `blockchain_explorer` to use new crate, 
- [ ] (Optional) consider writing using run-time agnostic libraries. 
- [ ] (Optional) consider enabling static dispatch `async` in trait methods. 
- [ ] In so doing, prefer being `async` friendly, because most workflows need `async`.

appetrosyan:
Comment due to @0x009922:
<quote>
I think it is not very hard to make the client's API fully Future-based, although it doesn't seem trivial to achieve the most important point (imo): iroha_client should be able to use a generic HTTP-transport instead of having a hard-code dependency such as awc, reqwest etc.

[...]

PR #3373, opened 2023-04-13
[feature] #1786: Add iroha-native types for socket addresses

### LINKED ISSUE #1786: Strong typing for socket addresses
Currently all socket addresses are stored as strings. This is not ideal, because we don't know if the configuration files should contain the protocol specifier or not. For example, there are many instances in the code, where a string check for presence of `http://` is done, and a format pre-pend. 

We should probably either use one of the `std::net` types, or (better), a custom type that can be converted into the types that we need.

QuentinI:
### Linked issue

Closes #1786

### Benefits

Less stringly-typed code

### Checklist


tools/kagami/src/main.rs
```diff
@@ -504,8 +504,8 @@ mod config {
         impl<T: Write> RunArgs<T> for Args {
             fn run(self, writer: &mut BufWriter<T>) -> Outcome {
                 let config = ConfigurationProxy {
-                    torii_api_url: Some(SmallStr::from_str(DEFAULT_API_URL)),
-                    torii_telemetry_url: Some(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL)),
+                    torii_api_url: Some(format!("http://{DEFAULT_API_ADDR}").parse().unwrap()),
```
appetrosyan, line 507: Consider using either `expect` or `?` preferably.

appetrosyan: You must communicate to the user that you have asserted the correctness and infallibility of the `parse()` call.

scripts/tests/panic_on_invalid_genesis.sh
```diff
@@ -2,8 +2,8 @@
 set -ex
 # Setup env
 export TORII_P2P_ADDR='127.0.0.1:1341'
-export TORII_API_URL='127.0.0.1:8084'
-export TORII_TELEMETRY_URL='127.0.0.1:8184'
+export TORII_API_ADDR='127.0.0.1:8084'
```
appetrosyan, line 5: Renaming the field is going to cause issues when upgrading. Please revert.

primitives/src/addr.rs
```diff
@@ -6,34 +6,92 @@
...
+        Hash,
+        DeserializeFromStr,
+        SerializeDisplay,
+        Encode,
+        Decode,
+        IntoSchema,
+    )]
     #[repr(transparent)]
-    pub struct Ipv4Addr([u8; 4]);
+    pub struct Ipv4Addr(pub [u8; 4]);
```
appetrosyan, line 89: Please keep this field private if possible.

appetrosyan (approved):

tools/kagami/src/main.rs
appetrosyan: Consider `?` or `expect`

scripts/tests/panic_on_invalid_genesis.sh
appetrosyan, line 5: Please rename back to `URL`. I know that there's a subtle difference, but it's not worth the headache during upgrade.

appetrosyan (changes requested):

primitives/src/addr.rs
```diff
@@ -100,12 +190,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
     }
 }
 
+impl core::str::FromStr for Ipv6Addr {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
appetrosyan, line 196: While this is acceptable, The trait is implemented for `std::net::SocketAddrv6` (and v4). You also know that this is only used with `serde` with the `std` feature enabled. As such you can avoid having to do all this work, by deferring to those primitives under a feature flag.

primitives/src/addr.rs
```diff
@@ -100,12 +190,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
...
+        }
+    }
+}
+
 impl core::fmt::Display for Ipv6Addr {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         // TODO: Implement omission of zeroes.
         for i in 0_usize..7_usize {
-            write!(f, "{i:x}:")?; // Need hexadecimal
+            write!(f, "{:x}:", self[i])?; // Need hexadecimal
```
appetrosyan, line 244: Good catch!

p2p/src/network.rs
```diff
@@ -167,13 +166,12 @@ impl_handle_methods! {
...
     /// Listening address for incoming connections. Must parse into [`std::net::SocketAddr`]
-    listen_addr: String,
+    listen_addr: SocketAddr,
     /// Current [`Peer`]s in [`Peer::Ready`] state.
     peers: HashMap<PublicKey, RefPeer<T>>,
     /// Map from [`std::net::IpAddr`] of the untrusted remote [`Peer`]:
     /// inserted by [`DisconnectPeer`] and removed by [`ConnectPeer`] from Sumeragi.
-    /// In case the [`String`] represents an unresolved hostname, the first reconnection is not refused
-    untrusted_peers: HashSet<String>,
+    untrusted_peers: HashSet<IpAddr>,
```
appetrosyan, line 174: Looks like we need to think about equality among `IpAddr` instances. For example should the peer `127.0.0.1` be equal to `localhost` equal to `::1/128`?

primitives/src/addr.rs
QuentinI: Good catch, will do

p2p/src/network.rs
appetrosyan: Similarly, the loopback is different to `localhost` but only in very specific circumstances. We need to be able to resolve the addresses somehow.

p2p/src/handshake.rs
```diff
@@ -0,0 +1,264 @@
+//! This is a WIP module containing the refactored state-machine based
```
appetrosyan, line 1: I think that this module should be removed. @Erigara fixed `p2p` to look much better, and this is no longer necessary.

core/test_network/src/lib.rs
```diff
@@ -832,29 +836,25 @@ impl TestConfiguration for Configuration {
...
-            small::SmallStr::from_str(api_url)
-        } else {
-            small::SmallStr::from_str(&("http://".to_owned() + api_url))
-        };
-        configuration.torii_telemetry_url = if telemetry_url.starts_with("http") {
-            small::SmallStr::from_str(telemetry_url)
-        } else {
-            small::SmallStr::from_str(&("http://".to_owned() + telemetry_url))
-        };
+        configuration.torii_api_url = format!("http://{api_url}")
```
appetrosyan, line 841: We're testing the parsing? Shouldn't this be handled by a unit test inside `primitives` and all other instances, handled by a purpose built-macro that does some checks?

core/test_network/src/lib.rs
```diff
@@ -662,11 +665,12 @@ impl PeerBuilder {
...
 type PeerWithRuntimeAndClient = (Runtime, Peer, Client);
 
-fn local_unique_port() -> Result<String> {
-    Ok(format!(
-        "127.0.0.1:{}",
-        unique_port::get_unique_free_port().map_err(Error::msg)?
-    ))
+fn local_unique_port() -> Result<SocketAddr> {
+    Ok(SocketAddrV4 {
+        ip: Ipv4Addr([127, 0, 0, 1]),
```
appetrosyan, line 670: If this is why you made the bytes public, I'd recommend using a constructor through a macro.

core/test_network/src/lib.rs
```diff
@@ -598,9 +601,9 @@ impl PeerBuilder {
     pub fn build(&mut self) -> Result<Peer> {
         let mut peer = Peer::new()?;
         if let Some(port) = self.port.take() {
-            peer.p2p_address = format!("127.0.0.1:{port}");
-            peer.api_address = format!("127.0.0.1:{}", port + 1);
-            peer.telemetry_address = format!("127.0.0.1:{}", port + 2);
+            peer.p2p_address = socket_addr!(127,0,0,1; port);
```
appetrosyan, line 604: Any particular reason why the syntax is not `127.0.0.1:port`? Should we have a ticket to rewrite as proc-macro that uses the standard syntax?

configs/client_cli/config.json
```diff
@@ -9,8 +9,8 @@
     "web_login": "mad_hatter",
     "password": "ilovetea"
   },
-  "TORII_API_URL": "http://127.0.0.1:8080",
-  "TORII_TELEMETRY_URL": "http://127.0.0.1:8180",
+  "TORII_API_URL": "http://127.0.0.1:8080/",
```
appetrosyan, line 12: If this slash is important: 
1. Consider coping with its absence. 
2. Document the necessary slash.

cli/src/samples.rs
```diff
@@ -65,9 +65,9 @@ pub fn get_config_proxy(peers: HashSet<PeerId>, key_pair: Option<KeyPair>) -> Co
             ..iroha_config::sumeragi::ConfigurationProxy::default()
         }),
         torii: Some(iroha_config::torii::ConfigurationProxy {
-            p2p_addr: Some(DEFAULT_TORII_P2P_ADDR.to_owned()),
-            api_url: Some(DEFAULT_API_URL.to_owned()),
-            telemetry_url: Some(DEFAULT_TORII_TELEMETRY_URL.to_owned()),
+            p2p_addr: Some(DEFAULT_TORII_P2P_ADDR.clone()),
+            api_addr: Some(DEFAULT_API_ADDR.clone()),
```
appetrosyan, line 69: I'd consider making the address `const`-constructible.

client/src/client.rs
```diff
@@ -1649,8 +1652,12 @@ mod tests {
...
                     .expect("This account ID should be valid"),
             ),
-            torii_api_url: Some(SmallStr::from_str(DEFAULT_API_URL)),
-            torii_telemetry_url: Some(SmallStr::from_str(DEFAULT_TORII_TELEMETRY_URL)),
+            torii_api_url: Some(format!("http://{DEFAULT_API_ADDR}").parse().unwrap()),
+            torii_telemetry_url: Some(
+                format!("http://{DEFAULT_TORII_TELEMETRY_ADDR}")
+                    .parse()
+                    .unwrap(),
+            ),
```
appetrosyan, line 1660: Not a huge fan of calling `format!(//..).parse()`.

client/src/http.rs
```diff
@@ -115,7 +118,7 @@ pub mod ws {
     ///
     ///     fn init(self) -> InitData<R, Self::Next> {
     ///         InitData::new(
-    ///             R::new(Method::GET, "http://localhost:3000"),
+    ///             R::new(Method::GET, "http://localhost:3000".parse().unwrap()),
```
appetrosyan, line 121:
```suggestion
    ///             R::new(Method::GET, "http://localhost:3000".parse().expect("`localhost` is a valid URL, port `3000` is sensible, `http` is supported")),
```

client/src/http.rs
```diff
@@ -288,16 +292,14 @@ pub mod ws {
...
-            "ws://".to_owned() + http_uri
-        } else {
-            return Err(eyre!("No schema in web socket uri provided. {}", uri));
-        };
+    /// Fails if passed URL doesn't have a valid protocol
+    pub fn transform_ws_url(mut url: Url) -> Result<Url> {
+        match url.scheme() {
+            "https" => url.set_scheme("wss").expect("Valid substitution"),
+            "http" => url.set_scheme("ws").expect("Valid substitution"),
+            _ => return Err(eyre!("Provided URL scheme is not http or https: {}", url)),
```
appetrosyan, line 300:
```suggestion
            _ => return Err(eyre!("Provided URL scheme is neither `http` nor `https`: {}", url)),
```

core/test_network/src/lib.rs
QuentinI: In client config `api_url` is an actual URL, not a socket address, and, bizzarely, there's no reasonable way to construct `url::Url` other that parsing a string.

core/test_network/src/lib.rs
QuentinI: Because 127.0 is a float, so standard syntax can't be expressed as a `macro_rules` macro.
I didn't think this was important enough to warrant a proc-macro, but I could rewrite it as one if you think it's a good idea

core/test_network/src/lib.rs
appetrosyan: Interesting. I'd imagine that a macro like `url!(http://{api_url}:8080/post_tx) would be tough to write, but if it expands to something along the lines of 
```rust
Url { 
   protocol: url::Http(), 
   host: api_url.clone(), 
   port: Some(8080), 
   segments: vec!["post_tx"], 
}
```
Could do the trick.

client/src/client.rs
QuentinI: Me neither, just no other way to construct `url::Url`. I can look up a replacement URL implementation if you feel this is way to unseemly

core/test_network/src/lib.rs
appetrosyan: I'd relegate it to a separate issue.

client/src/client.rs
appetrosyan: Come to think of it. Leave it for a separate ticket.

primitives/src/addr.rs
QuentinI: Actually, deserialization creeps into `iroha_data_model::Peer`, which would also then need to implement Deserialize conditionally on `std`. That seems like an inconsistent API to me.

client/src/client.rs
QuentinI: Opened https://github.com/hyperledger/iroha/issues/3385

core/test_network/src/lib.rs
QuentinI: Resolving to be implemented as part of https://github.com/hyperledger/iroha/issues/3385

p2p/src/network.rs
QuentinI: We're not dealing with hostnames here, so that one is not a problem to consider.

Comparing supposedly equal IPs sounds quite complicated. One could run two peers listening on `127.0.0.1` and `::1` respectively for example. I'd say we should just leave it as is.

QuentinI:
`kagami genesis` seems to have become is non-reproducible with runtime validators.

cli/src/samples.rs
appetrosyan: Doubt that can be done in general.

primitives/src/addr.rs
appetrosyan: IMO, the `Peer` as a whole is an oversight if allowed on the smartcontract side: reason being that the `Peer` operations are registering and de-registering, which **should** be done rarely and with human oversight. 

Moreover, any form of `serde` deserialization has to do with running code in the peer, (not with the smartcontract) which usually depends on `std`. I don't think it's inconsistent, it's just not exactly what we're doing at the moment. 

If you'd like this to be a separate issue, I agree, but not that it shouldn't be done in general.

QuentinI: > IMO, the Peer as a whole is an oversight if allowed on the smartcontract side: reason being that the Peer operations are registering and de-registering, which should be done rarely and with human oversight.

Runtime validators will want to use it though, and they share API surface with smartcontracts AIUI.

> Moreover, any form of serde deserialization has to do with running code in the peer, (not with the smartcontract) which usually depends on std. I don't think it's inconsistent, it's just not exactly what we're doing at the moment.

That I agree with, probably losing serde support for `Peer` in `no_std` environment is not a big deal. I'll update the PR.

QuentinI: > If you'd like this to be a separate issue, I agree, but not that it shouldn't be done in general.

On further examination, I'd prefer to make this a separate issue. It chains into a lot of changes in `data_model`. Opened https://github.com/hyperledger/iroha/issues/3407

appetrosyan (dismissed):

primitives/src/addr.rs
```diff
@@ -6,30 +6,96 @@
...
+    /// Ipv6 address contains more than one '::' abbreviation
+    UnexpectedAbbreviation,
+}
 
 ffi::ffi_item! {
-    /// An Iroha-native version of `std::net::Ipv4Addr`, duplicated here to remain `no_std` compatible.
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, From, AsRef, IntoIterator, Decode, Encode, Deserialize, Serialize, IntoSchema)]
-    #[serde(transparent)]
+    /// An Iroha-native version of `std::net::Ipv4Addr`, duplicated here
+    /// to remain `no_std` compatible.
```
Erigara, line 69: But `SocketAddr` is available in [`core`}(https://doc.rust-lang.org/stable/core/net/enum.SocketAddr.html) and since `1.69` construction will be available in cons context.

client/src/http.rs
```diff
@@ -13,7 +14,7 @@ pub use http::{Method, Response, StatusCode};
 pub trait RequestBuilder {
     /// Create a new builder with specified method and URL. Entrypoint for most client operations.
     #[must_use]
-    fn new(method: Method, url: impl AsRef<str>) -> Self;
+    fn new(method: Method, url: Url) -> Self;
```
0x009922, line 17: It is kind of a breaking change. When builder is used for different HTTP-client crates, they might have their own abstractions over URL, and this might lead to extra allocations.

Though I don't think it is critical, since the whole client API is going to be drastically changed:

- #3130

primitives/src/addr.rs
```diff
@@ -6,30 +6,96 @@
...
 use iroha_schema::IntoSchema;
 use parity_scale_codec::{Decode, Encode};
 use serde::{Deserialize, Serialize};
+use serde_with::{DeserializeFromStr, SerializeDisplay};
 
-use crate::ffi;
+use crate::{conststr::ConstString, ffi};
+
+#[macro_export]
+/// Convenience macro to concisely construct a `[SocketAddr]`
```
0x009922, line 19:
```suggestion
/// Convenience macro to concisely construct a [`SocketAddr`]
```

primitives/src/addr.rs
```diff
@@ -6,30 +6,96 @@
...
+    NoPort,
+    /// Ipv6 address contains more than one '::' abbreviation
+    UnexpectedAbbreviation,
+}
 
 ffi::ffi_item! {
-    /// An Iroha-native version of `std::net::Ipv4Addr`, duplicated here to remain `no_std` compatible.
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, From, AsRef, IntoIterator, Decode, Encode, Deserialize, Serialize, IntoSchema)]
-    #[serde(transparent)]
+    /// An Iroha-native version of `std::net::Ipv4Addr`, duplicated here
```
0x009922, line 68:
```suggestion
    /// An Iroha-native version of [`std::net::Ipv4Addr`], duplicated here
```

primitives/src/addr.rs
```diff
@@ -60,24 +150,33 @@ impl Ipv4Addr {
 }
 
 ffi::ffi_item! {
-    /// An Iroha-native version of `std::net::Ipv6Addr`, duplicated here to remain `no_std` compatible.
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, From, AsRef, IntoIterator, Decode, Encode, Deserialize, Serialize, IntoSchema)]
-    #[serde(transparent)]
+    /// An Iroha-native version of `std::net::Ipv6Addr`, duplicated here
```
0x009922, line 153:
```suggestion
    /// An Iroha-native version of [`std::net::Ipv6Addr`], duplicated here
```

primitives/src/addr.rs
```diff
@@ -98,12 +202,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
...
         for i in 0_usize..7_usize {
-            write!(f, "{i:x}:")?; // Need hexadecimal
+            write!(f, "{:x}:", self[i])?; // Need hexadecimal
         }
         write!(f, "{:x}", self[7])
     }
 }
+
+ffi::ffi_item! {
+    /// An Iroha-native version of `std::net::IpAddr`, duplicated here
```
0x009922, line 263:
```suggestion
    /// An Iroha-native version of [`std::net::IpAddr`], duplicated here
```

primitives/src/addr.rs
```diff
@@ -98,12 +202,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
...
+    pub enum IpAddr {
+        /// Ipv4 variant
+        V4(Ipv4Addr),
+        /// Ipv6 variant
+        V6(Ipv6Addr),
+    }
+}
+
+ffi::ffi_item! {
+    /// This struct provides an Iroha-native version of `std::net::SocketAddrV4`. It is duplicated here
```
0x009922, line 291:
```suggestion
    /// This struct provides an Iroha-native version of [`std::net::SocketAddrV4`]. It is duplicated here
```

primitives/src/addr.rs
```diff
@@ -98,12 +202,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
...
+        let (ip, port) = value.split_once(':').ok_or(ParseError::NoPort)?;
+        Ok(Self {
+            ip: ip.parse()?,
+            port: port.parse().map_err(|_| ParseError::InvalidPort)?,
+        })
+    }
+}
+
+ffi::ffi_item! {
+    /// This struct provides an Iroha-native version of `std::net::SocketAddrV6`. It is duplicated here
```
0x009922, line 344:
```suggestion
    /// This struct provides an Iroha-native version of [`std::net::SocketAddrV6`]. It is duplicated here
```

primitives/src/addr.rs
```diff
@@ -98,12 +202,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
...
+        let port = port.parse().map_err(|_| ParseError::InvalidPort)?;
+        Ok(Self {
+            host: host.into(),
+            port,
+        })
+    }
+}
+
+ffi::ffi_item! {
+    /// This enum provides an Iroha-native version of `std::net::SocketAddr`. It is duplicated here
```
0x009922, line 441:
```suggestion
    /// This enum provides an Iroha-native version of [`std::net::SocketAddr`]. It is duplicated here
```

primitives/src/addr.rs
```diff
@@ -98,12 +202,586 @@ impl core::ops::Index<usize> for Ipv6Addr {
...
+        /// An Ipv6 socket address.
+        Ipv6(SocketAddrV6),
+        /// A socket address identified by hostname
+        Host(SocketAddrHost),
+    }
+}
+
+impl SocketAddr {
+    /// Returns `Some(IpAddr)` for `Ipv4` and `Ipv6` variants,
+    /// `None` otherwise.
```
0x009922, line 470:
```suggestion
    /// Extracts [`IpAddr`] from [`Self::Ipv4`] and [`Self::Ipv6`] variants
```

primitives/src/addr.rs
0x009922, line 1: Isn't it a too big wheel to reinvent? Why not [`no-std-net`](https://docs.rs/no-std-net/latest/no_std_net/)?

QuentinI: Existing wheels aren't SCALE-compatible unfortunately.

primitives/src/addr.rs
QuentinI: As far as I understand it's still not SCALE-compatible.

Erigara: By SCALE-compatible you mean that it doesn't implement `Encode/Decode`?

primitives/src/addr.rs
mversic: @0x009922 as for the `feature(ip_in_core)` we're trying to keep `Iroha` to build on stable

config/src/client.rs
```diff
@@ -151,57 +152,28 @@ impl ConfigurationProxy {
...
-                        value: api_url.to_string(),
-                        message: ", because it's missing the connection protocol (e.g. `http://`)"
-                            .to_owned(),
-                    });
-                }
+            if api_url.scheme() != "http" {
+                eyre::bail!(ConfigError::InsaneValue {
+                    field: "TORII_API_URL",
+                    value: api_url.to_string(),
+                    message: ", because we only support HTTP".to_owned(),
```
QuentinI, line 159: Do we though? As far as I understand client is fully capable of HTTPS?

Erigara (dismissed):

appetrosyan (approved):

Erigara (approved):

PR #3395, opened 2023-04-18
[tests]: Remove false negatives from coverage

appetrosyan:
## Description

Remove coverage instrumentation from trivial and functions which are only checkable in a real environment

### Linked issue
**TODO**

### Benefits

Fewer false-negatives in coverage

### Checklist

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(coverage_nightly, feature(no_coverage))]
 //! Common primitives for a CLI instance of Iroha. If you're
 //! customising it for deployment, use this crate as a reference to
 //! add more complex behaviour, TUI, GUI, monitoring, etc.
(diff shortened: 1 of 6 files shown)


cli/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg_attr(coverage_nightly, feature(no_coverage))]
```
mversic, line 1: is this a mistake? in other places you use `#[cfg_attr(coverage_nightly,no_coverage)]`

mversic: nvm, I see the docs on `cargo-llvm-cov`

mversic: Seems like that in `cargo-llvm-cov` but I'm confused by this

mversic (approved):

cli/src/lib.rs
appetrosyan: So the point is we need to enable this nightly feature conditional on running coverage via nightly. 

The remaining places just mark functions as `no_coverage`.

appetrosyan:
It's ugly and it doesn't nearly help coverage as much as I'd hoped. We could achieve the same results by just excluding the interface crates from coverage calculations, which would also be easier to deal with in terms of toolchain upgrades. 

I'm afraid for the time being, the only way to improve coverage is by writing more tests. Even trivial ones.

PR #3394, opened 2023-04-18
[fix] #3393: Break communication deadlock loop in actors

Erigara:
## Description

- Add tokio console support.
- Fixed bug with communication deadlock.
- Refactor sumeragi/network in terms of responsibility for connecting/disconnecting peers. 

### Linked issue

Closes #3393, #3392 

### Checklist

- [x] I've read `CONTRIBUTING.md`
- [x] I've used the standard signed-off commit format (or will squash just before merging)
- [x] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

PR #3405, opened 2023-04-20
[fix] #3396: Tps benchmark fails block missing

### LINKED ISSUE #3396: [bug]: Tps benchmark fails block missing
I see two reasons why it can happen currently: 

1. In case of soft fork block committed event will be emitted twice with only advancing height once.
2. Block committed event is emitted before block is actually stored by kura, so there is possibility that client will ask for block not yet stored by kura.

Erigara:
## Description

- Refactor inside topology + tests
- Refactored block sync handling + tests
- Fixed order when block is committed + tests

### Linked issue

Closes #3396 

### Benefits

Iroha less flaky.

### Notes

Want to merge #3356 and #3394 first. 

### Checklist


client/tests/integration/events/pipeline.rs
```diff
@@ -86,3 +86,32 @@ impl Checker {
         })
     }
 }
+
+#[test]
+fn committed_block_must_be_available_in_kura() {
```
Erigara, line 91: This test was flaky before i made changes.

Erigara:
I will split this PR in multiple PRs.

PR #3419, opened 2023-04-24
[refactor] #3388: Make`model!` a module-level attribute macro

### LINKED ISSUE #3388: Make `model!` be a module attribute macro
After a discussion with @mversic we decided to re-implement `model!` macro as a module attribute macro.
New usage will look like so:

```rust
#[model]
pub mod model {
    // Struct goes here...
}
```

Pros:

1. Better IDEs support. They can't always analyze what's going on inside `model!` macro, but can work with module like that
2. Less noise in the codebase
3. You don't skip some struct accidentally

Arjentix:
## Description

* `model` was rewritten as an attribute macro which should be applied on `pub mod model`.
* Refactored all usages
* Created `model_single` for edge cases which works the same as old `model!`

### Linked issue

* Closes #3388  

### Benefits

See #3388 for benefits

### Checklist


data_model/derive/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license.workspace = true
 proc-macro = true
 
 [dependencies]
-syn = "1.0.107"
+syn = { version = "1.0.107", features = ["full", "extra-traits"] }
```
appetrosyan, line 14: If this is used once, it should be used everywhere. We don't win anything by reducing the feature set, not even compilation time.

data_model/src/account.rs
```diff
@@ -60,7 +63,23 @@ model! {
     ///
     /// let id = "user@company".parse::<AccountId>().expect("Valid");
     /// ```
-    #[derive(Debug, Display, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Constructor, Getters, Decode, Encode, DeserializeFromStr, SerializeDisplay, IntoSchema)]
+    #[derive(
```
QuentinI, line 66: Seems like we have an editor war on `derive`s.
@mversic's  PRs have them in one line, if I'm not mistaken. Maybe we should sign a peace treaty to have less noise in diffs?

version/derive/src/lib.rs
```diff
@@ -173,7 +173,7 @@ fn impl_version(args: Vec<NestedMeta>, item: TokenStream) -> TokenStream2 {
     let alias_type_name = format_ident!("_{}V{}", versioned_struct_name, version_number);
     quote!(
         /// Autogenerated alias type to link versioned item to its container.
-        type #alias_type_name = #struct_name;
+        pub type #alias_type_name = #struct_name;
```
QuentinI, line 176: Why do we need this?

data_model/src/account.rs
appetrosyan: IMO, this is on `rustfmt`. That program is next to useless for imposing a consistent style, because it fails to remove tabs in my PRs. And here the threshold for when to break up derives is set either too low or too high. 

The **solution** to the problem is to perform a self-review on the draft and ensure that any change that you commit is necessary. I can tell @mversic does that because he often leaves comments on his own PRs and I think it's a good practice.

data_model/derive/Cargo.toml
mversic: this is tracked by #3289

version/derive/src/lib.rs
mversic: good catch. This should remain private

data_model/src/account.rs
Arjentix: I'm on the side of `rustfmt`. We can configure it as we wish, but let's keep the formating it provides

version/derive/src/lib.rs
Arjentix: Yes, it probably should be `pub(super)`

mversic: `pub(crate)` is also fine

Arjentix:
> Also, now we have two macros model and model_single because of issues with macro expansion order.

Yes, but `model_single!` used only in three modules. And that's only because we wanted to reduce the amount of code by using other macros.

> The only possible benefit here is that the code now looks more like substrate code?

No, there are other benefit to me:

- We still have better IDE support, because `rust-analyzer` can read modules (and format it too)
- To me module named `model` shows better what is going on

But I also can see some downsides:

- Indeed having `model_single` isn't very good
- Code size has been increased (probably only because of formated derives)

Arjentix:
But I'm also not sure if it worth it to merge this PR. I just tried to implement it and now we can see all cons and pros

Arjentix:
`substrate` way probably works for them because they have limited number of things that could be declared inside `pallet` macro. But we have unlimited number of sometimes repetative structs and enums

Arjentix:
Interesting that even `syn` uses macros like our old `model!` or current `model_single!`: [source](https://docs.rs/syn/latest/src/syn/item.rs.html#219-231)

Arjentix:
Waiting for other opinions on this topic.

I think I won't update this PR until we decide if we need this PR, because it may be just a waste of time if we close it

data_model/src/account.rs
appetrosyan: If formatting is not enforced by `rustfmt` it's not enforced. With that said, the whole point of moving to an attribute macro was to allow for tooling to inspect the internals of the module, and hence format. 

With that said, https://lkml.org/lkml/2020/5/29/1038, the 80 char limit isn't as uniformly accepted as it used to be.

appetrosyan: That's exactly what I said, no? 

> With that said, the whole point of moving to an attribute macro was to allow for tooling to inspect the internals of the module, and hence format.

version/derive/src/lib.rs
appetrosyan: If in doubt, prefer `pub(crate)`. It makes re-arranging modules less annoying.

appetrosyan:
My two cents. 

TL;DR **We need this, let's merge**. 

> The only possible benefit here is that the code now looks more like substrate code?~

I would hardly classify this as a benefit. And yes while this is a far cry from using types to implement constants, and traits to reflect structures (the OOP virus is quite contagious), it's not a comparison we should invite. 

Using a module + attr macro has the following benefits, though. 

## Automatic namespacing

This has a lot more benefits than one might think at first glance. We know that the path segment to a type will always have `<thing>::model::<struct>`, which would allow us to tidy the structures and coupled with `use` make the code either less verbose, or more explicit. 

Furthermore, because of the way Rust handles modules, every object used in a `model` has to be imported. If `model` forbids the use of `use super::*`, we now have an explicit list of types, that can be parsed out from the `use` statements. This can be used to generate warnings for misbehaving types later down the line, when the developer UX becomes a priority. 

## Tooling support

As we've seen here, the contents of function-like macros are ignored in formatting by default. Reason is fairly obvious; because you can, and are encouraged to, define arbitrary syntax that might potentially be sensitive to things that in normal Rust are irrelevant (for example, "{}" is not the same in `print` as in a plain `&str`). This already led to `rustfmt` applying to more code than with `model!`. For good or ill, we have a style guide enforced by `rustfmt`. This change improves the readability. 

Right now, `rust-analyzer` can still get confused inside an annotated module, but less so. Eventually, perhaps it will work as advertised, or another better language server will emerge. Then the difference would be vital. 

## Minimum surprise

`#[cfg(test)]` is fairly common. 

A construction like the one in the old `model` is not. 

## #[cfg_attr]

The code generated by `model` is used for the `ffi`. We don't use it yet. Expanding the macro when you need to is what we should be doing in order to reduce compilation time and ensure good developer UX. Being able to only expand these macros conditionally can help a lot. 

## Other attr-like proc macros

Right now these things are only added to the `model` macro because that's the only place they're used. What if we have another API that we need to support for all `data_model` entities. E.g. we want substrate-compatibility? 

Compare the readability of 

```rust
#[model]
#[substrate_compat]
#[js_bind]
#[jvm_bind]
pub mod model {

}
``` 
to 
```rust
model! {
	substrate_compat! {
		js_bind! {
			jvm_bind{
				//... 
			}
		}
	}
}

**Comment by mversic** [comment]
following the discussion, I see the benefits of attribute like macros vs function like ones. I'll approve once updated

File: data_model/derive/src/lib.rs [hunk 4]
```diff
@@ -51,38 +44,50 @@ impl syn::parse::Parse for Items {
...
+pub fn model_single(input: TokenStream) -> TokenStream {
+    model::process_item(parse_macro_input!(input)).into()
+}
 
-    quote::quote! { #(#items)* }.into()
+/// Stub for usage inside [`model`] macro.
+/// Will make [`model`] macro to produce the same item but with
+/// [`iroha_ffi_derive::FfiType`] when `ffi_export` feature is enabled and `ffi_import` is disabled.
+#[proc_macro_attribute]
+pub fn ffi_type(_attr: TokenStream, _input: TokenStream) -> TokenStream {
```
mversic, line 89: why do we have this?

data_model/src/transaction.rs
```diff
@@ -229,6 +247,143 @@ model! {
...
+        /// The reason for rejecting this transaction during the validation pipeline.
+        #[getset(get = "pub")]
+        pub rejection_reason: error::TransactionRejectionReason,
+    }
+
+    /// `AcceptedTransaction` — a transaction accepted by iroha peer.
+    #[version_with_scale(n = 1, versioned = "VersionedAcceptedTransaction")]
+    #[derive(Debug, Clone, Decode, Encode, Serialize)]
+    #[ffi_type]
+    pub struct AcceptedTransaction {
```
mversic, line 357: previously it was `pub(crate)`. This was a feature where a struct would be turned private if built without `transparent_api`. In this case this transaction is useless to `iroha_client` or `iroha_wasm` so it was removed from public API.

data_model/src/transaction.rs
```diff
@@ -229,6 +247,143 @@ model! {
...
+        PartialEq,
+        Eq,
+        Hash,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    #[display(fmt = "{self:?}")] // TODO ?
```
mversic, line 270:
```suggestion
    #[display(fmt = "{}", "self.hash()")]
```

would you fix the display?

data_model/src/transaction.rs
mversic: the same was with `QueryPayload` and `AcceptTransactionFailure`

data_model/src/query.rs
```diff
@@ -7,27 +7,36 @@ use alloc::{boxed::Box, format, string::String, vec::Vec};
...
+
+            $(
+                #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+                #[derive(parity_scale_codec::Decode, parity_scale_codec::Encode)]
+                #[derive(serde::Deserialize, serde::Serialize)]
+                #[derive(iroha_schema::IntoSchema)]
+                $($meta)*
+                $item
+            )+
         }
```
mversic, line 40:
```suggestion
        pub mod model{
            use super::*; $(
                #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
                #[derive(parity_scale_codec::Decode, parity_scale_codec::Encode)]
                #[derive(serde::Deserialize, serde::Serialize)]
                #[derive(iroha_schema::IntoSchema)]
                $($meta)*
                $item )+
        }
```

totally irrelevant, but this is how I prefer for macro to look like because it's the closest to how the actual source code would be laid out so it makes it easier for me to read it

mversic: I format the code as if it is source code and then put macro tokens around it

data_model/derive/src/filter.rs
```diff
@@ -200,14 +200,13 @@ pub fn impl_filter(event: &EventEnum) -> TokenStream {
     let filter_doc = format!(" Filter for {event_ident} entity");
 
     quote! {
-        iroha_data_model_derive::model! {
```
mversic, line 203: what is the reason to remove this?

data_model/derive/src/lib.rs
```diff
@@ -33,12 +19,19 @@ impl syn::parse::Parse for Items {
 /// Additionally, this macro will rewrite private items as public when `transparent_api` is active.
 /// If an item should remain private regardless of consumer library, just don't wrap it in this macro.
 ///
+/// Items private fields will be marked as `pub(super)`, so that it will be possible to write
+/// proper `impl`s for such items.
```
mversic, line 23: I don't see a reason why you would make private fields be anything else. IMO a private field should remain private

data_model/derive/src/model.rs
```diff
@@ -1,8 +1,49 @@
...
+        );
+    }
+
+    let items_code = content.as_ref().map_or_else(Vec::new, |(_, items)| {
+        items.iter().cloned().map(process_item).collect()
+    });
+
+    quote! {
+        #(#attrs)*
+        #[doc = "Module with exported structs and enums forming *model* of the parent module"]
```
mversic, line 33: I would use `allow(missing_docs)` here. We use it already in similar situations

mversic, line 33:
```suggestion
        #[allow(missing_docs)]
```

I would use `allow(missing_docs)` here. We use it already in similar situations

data_model/derive/src/lib.rs
Arjentix: Because without this it will be enable to compile code inside `model` mod which uses `#[ffi_type]`. I just replicated the same behaviour as before. This annotation will be processed inside `#[model]` macro

data_model/derive/src/lib.rs
Arjentix: Making them `pub(super)` enables to write `impl`s in parent modules. So that we try to keep only struct and enum declarations inside `model` macro. Do you think we should move impl into `model` as well?

mversic: why doesn't the developer just make it `pub(super)` then? If I make a field private I would expect it to stay private. Either don't allow for a field to be private or leave it private. It's ok to hide away struct internals, but but I would never reveal additional details about the struct.

data_model/derive/src/lib.rs
mversic: > This annotation will be processed inside #[model] macro

doesn't this processing remove this attribute? I don't think it has to be declared as a proc macro

data_model/derive/src/lib.rs
Arjentix: Okay, agree, it will be more explicit

data_model/derive/src/lib.rs
Arjentix: No, looks like `rustc` wants `#[ffi_type]` to be declared before `#[model]` expansion

mversic: does this mean that macros inside the module are expanded first? and module level macros are expanded afterwards

data_model/src/transaction.rs
Arjentix: I can't because there is no `hash()` function without `std` feature enabled. Btw why do we use cryptographic hash for it? Shouldn't it be just a basic non-cryptographic hash just to identify this transaction?

mversic: but `hash` is in `core::hash`?

data_model/derive/src/lib.rs
Arjentix: I don't think so. I think it's like derive attributes. `rustc` wont allow to use undefined attributes despite that any of derive macros can use it

mversic: interesting. And we can't move the `ffi_type` code into this function because then we would have to make sure it's top-level macro? I would panic inside this function then because it can only be used when wrapped in `model!` macro

Arjentix: Hm, good idea, I'll try this

Arjentix: It worked!

data_model/derive/src/filter.rs
Arjentix: You're right, there should be `model_single!`

data_model/src/transaction.rs
Arjentix: So they shouldn't be `ffi_type` then?

mversic: no

data_model/derive/src/lib.rs
Arjentix: I was wrong. I was meslead by another error which was fixed by that. Now everything works without it

data_model/derive/src/filter.rs
```diff
@@ -245,14 +246,12 @@ fn impl_event_filter(event: &EventEnum) -> proc_macro2::TokenStream {
     let event_filter_doc = format!(" Event filter for {event_ident} entity");
 
     quote! {
-        iroha_data_model_derive::model! {
```
mversic, line 248: also missing `model_single`

data_model/derive/src/lib.rs
```diff
@@ -51,40 +41,57 @@ impl syn::parse::Parse for Items {
...
+// /// [`iroha_ffi_derive::FfiType`] when `ffi_export` feature is enabled and `ffi_import` is disabled.
+// #[proc_macro_attribute]
+// #[proc_macro_error::proc_macro_error]
+// pub fn ffi_type(_attr: TokenStream, input: TokenStream) -> TokenStream {
+//     let input = proc_macro2::TokenStream::from(input);
+//     proc_macro_error::abort!(
+//         input,
+//         "`ffi_type` attribute macro should be used only inside `model` macro",
+//     );
+// }
```
mversic, line 93:
```suggestion
```

data_model/src/isi.rs
```diff
@@ -8,19 +8,88 @@ use core::fmt::Debug;
...
+            #[derive(iroha_schema::IntoSchema)]
+            #[getset(get = "pub")]
+            $($meta)*
+            $item
+        }
+    };
+}
+
+#[model]
+pub mod model {
```
mversic, line 64: IMO you should move all `isi!` instructions into this module

mversic: I see you did the same for events

data_model/src/query.rs
```diff
@@ -7,27 +7,34 @@ use alloc::{boxed::Box, format, string::String, vec::Vec};
...
     trigger::*,
 };
-use crate::{account::Account, model, Identifiable, Value};
+use crate::{account::Account, Identifiable, Value};
+
+macro_rules! queries {
+    ($($($meta:meta)* $item:item)+) => {
+        pub use self::model::*;
+
+        #[iroha_data_model_derive::model]
```
mversic, line 28: If I'm not mistaken now you have `model` module inside `model` module? I would rather use `model_single`

data_model/src/isi.rs
mversic: mah, nvm

QuentinI (approved):

PR #3425, opened 2023-04-25
[fix] #3411: Properly handle peers simultaneous connection

### LINKED ISSUE #3411: [bug]: Iroha peers simultaneous connection
At the moment when two peers are trying to connect each other at the same time it's possible to face situation when them don't get connected.

Erigara:
## Description

- Prevent simultaneous connections by selecting who will be passive and active when connecting.
- Prevent connection duplication by keeping track of outgoing connections.
- If simultaneous connection happen add policy to resolve which connection to preserve. 

### Linked issue

Closes #3411.

### Benefits

Test flakes at much smaller rate, i run it like 10000 times without any error + works a lot faster.

There still can be cases where we can lose message (and thus `multiple_networks` test can fail).

### Checklist

### DIFF
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -5,7 +5,13 @@
 use std::{io, net::AddrParseError};
 
 use iroha_crypto::ursa::{
-    encryption::symm::prelude::ChaCha20Poly1305, kex::x25519::X25519Sha256, CryptoError,
+    blake2::{
+        digest::{Update, VariableOutput},
+        VarBlake2b,
+    },
+    encryption::symm::prelude::ChaCha20Poly1305,
+    kex::x25519::X25519Sha256,
+    CryptoError,
 };
 pub use network::message::*;
 use parity_scale_codec::{Decode, Encode};
@@ -157,4 +163,16 @@ pub(crate) mod unbounded_with_len {
             Ok(())
         }
     }
+}
+
+/// Create Blake2b hash as u64 value
+pub fn blake2b_hash(slice: impl AsRef<[u8]>) -> u64 {
+    const U64_SIZE: usize = core::mem::size_of::<u64>();
+    let hash = VarBlake2b::new(U64_SIZE)
+        .expect("Failed to create hash with given length")
+        .chain(&slice)
+        .finalize_boxed();
+    let mut bytes = [0; U64_SIZE];
+    bytes.copy_from_slice(&hash);
+    u64::from_be_bytes(bytes)
 }
--- a/p2p/src/network.rs
+++ b/p2p/src/network.rs
@@ -14,10 +14,11 @@ use iroha_logger::prelude::*;
 use message::*;
 use tokio::{
     net::{TcpListener, TcpStream},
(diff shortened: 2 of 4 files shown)


p2p/src/lib.rs
```diff
@@ -150,3 +156,15 @@ pub(crate) mod unbounded_with_len {
         }
     }
 }
+
+/// Create Blake2b hash as u64 value
+pub fn blake2b_hash(slice: impl AsRef<[u8]>) -> u64 {
```
Erigara, line 161: I'm not sure should i place this in `iroha_crypto` because it's used only in `iroha_p2p` crate.

mversic: the only difference is hash length? The proper thing to do would be to make `iroha_crypto::Hash` parametrized with hash length. Which could be a very large change through the codebase, ergo I think you should leave this here

Erigara: The difference is that returned value is ’u64’. I initially want to do arithmetic operations on hashes, but currently i'm using only comparison. So i think i can switch to ’Hash’ directly. But with ’u64’ we still save quite some place. 
What do you think?

SamHSmith (dismissed):
Very good solution to the problem of which connection to preserve. I did not think of this when I was trying to rewrite p2p. Do you think the sumeragi view change indices inspired this design?

p2p/src/lib.rs
appetrosyan: Let's leave as is.

p2p/src/network.rs
```diff
@@ -191,40 +192,40 @@ struct NetworkBase<T: Pload, K: Kex, E: Enc> {
 
 impl<T: Pload, K: Kex, E: Enc> NetworkBase<T, K, E> {
     /// [`Self`] task.
+    #[log(skip(self), fields(listen_addr=self.listen_addr, public_key=%self.public_key))]
     async fn run(mut self) {
         // TODO: probably should be configuration parameter
-        let mut update_topology_interval = tokio::time::interval(Duration::from_millis(1000));
+        let mut update_topology_interval = tokio::time::interval(Duration::from_millis(100));
```
appetrosyan, line 198: Ideally we shouldn't have these long-term. However, I'm conscious that sometimes arbitrary numbers have to be pulled out of a hat to make things work.

Erigara: I agree, in theory it should be possible to remove periodic topology update at all.
We will just try to reconnect if previous connection was terminated.
But for know i think periodic approach is ok.

Erigara:
> Do you think the sumeragi view change indices inspired this design?

Im not sure what exactly affected my design.
Initially i was thinking about adding some kind of value that one peer will send to another and this value will be used as resolve connections, but i didn't want to introduce another stage of handshake.
And then i got insight that we already have such value.

mversic (approved):

mversic (approved):

mversic (approved):

SamHSmith (approved):

PR #3432, opened 2023-04-27
[refactor] #3412: Move transaction gossiping in separate actor

### LINKED ISSUE #3412: [refactor]: Exclude transaction gossiping out of sumeragi
At the moment transaction gossiping is orthogonal to want main purpose of sumergai.
So this functionality can be excluded into separate actor to simplify sumeargi.

Erigara:
## Description

Add new actor for transaction gossiping.

### Linked issue

Closes #3412 

### Benefits

Sumeragi is more streamlined.

### Checklist


core/src/gossiper.rs
```diff
@@ -0,0 +1,167 @@
+//! Gossiper is actor which is responsible for transaction gossiping
```
mversic, line 1: why is it not also responsible for block gossiping?

Erigara: we already have block sync, do you think we should combine them?

Erigara: i want to keep scope of actor narrow

mversic: let's rename it to `TransactionGossiper` then, because we have multiple forms of gossiping

core/src/gossiper.rs
```diff
@@ -0,0 +1,167 @@
...
+    sumeragi: Arc<Sumeragi>,
+    /// Limits that all transactions need to obey, in terms of size
+    /// of WASM blob and number of instructions.
+    transaction_limits: TransactionLimits,
+}
+
+impl Gossiper {
+    /// Start [`Self`] actor.
+    pub fn start(self) -> GossiperHandle {
+        let (message_sender, message_receiver) = mpsc::channel(1);
```
mversic, line 56: why does it have capacity of 1?

core/src/gossiper.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Send [`TransactionGossip`] to actor
+    pub async fn gossip(&self, gossip: TransactionGossip) {
+        self.message_sender
+            .send(gossip)
+            .await
+            .expect("Gossiper must handle messages until there is at least one handle to it")
+    }
+}
+
+/// Actor to gossip transactions and receive transaction gossips
```
mversic, line 34: isn't this struct only receiving, `GossiperHandler` is responsible for sending?

core/src/gossiper.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+}
+
+fn enqueue_transaction(
+    tx: VersionedSignedTransaction,
+    queue: &Queue,
+    wsv: &WorldStateView,
+    transaction_limits: &TransactionLimits,
+) {
```
mversic, line 131:
```suggestion
fn enqueue_transaction(&self, tx) {
```

I would make it a method

core/src/gossiper.rs
Erigara: I think we can insist on difference but still use `TransactionGossiper` to highlight what we actually gossiping.

core/src/gossiper.rs
Erigara: No, we generate gossips every `gossip_period` and broadcast them. 

And `GossiperHandle` is used to provide recieved gossips from `Network` to `Gossiper`.

mversic (approved):

SamHSmith (dismissed):
Looks good.

mversic (approved):

PR #3434, opened 2023-04-28
[feature] #3168: Provide genesis validator in separate file

### LINKED ISSUE #3168: Find a way to check genesis.json in CI overcoming cargo builds reproducibility problem

Erigara:
## Description

Add ability to provide validator in separate file not inlined in genesis.

### Linked issue

Closes #3168 

### Benefits

Genesis isn't overloaded with inlined validator. 

### Checklist


tools/kagami/README.md
```diff
@@ -136,7 +137,13 @@ You can also choose output format:
...
 
     ```bash
-    kagami genesis --synthetic --domains n --accounts-per-domain m --assets-per-domain p
+    kagami genesis synthetic --domains n --accounts-per-domain m --assets-per-domain p
+    ```
+
+ - Generate a default genesis block in JSON format and provide path to the validator file (it could be absolute or relative to genesis file)
+
+    ```bash
+    kagami genesis --validator-path ./validator.wasm
```
mversic, line 146: is this argument always required? I would make it positional if so

Erigara: No, you can omit it and genesis will be generated with validator inlined (previous behavior).

mversic: why would we want that?

Erigara: i thought that it may be more convenient for some setups to provide everything in single file and want to remain backwards compatible

data_model/src/transaction.rs
```diff
@@ -395,6 +395,13 @@ impl AsRef<[u8]> for WasmSmartContract {
     }
 }
 
+impl WasmSmartContract {
+    /// Create [`Self`] from raw wasm bytes
+    pub fn from_raw(blob: Vec<u8>) -> Self {
+        Self(blob)
+    }
+}
```
mversic, line 403: there is `WasmSmartContract::new`. Remove either one

core/test_network/src/lib.rs
```diff
@@ -75,10 +75,9 @@ impl TestGenesis for GenesisNetwork {
 
         // TODO: Fix this somehow. Probably we need to make `kagami` a library (#3253).
         let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
-        let content = std::fs::read_to_string(manifest_dir.join("../../configs/peer/genesis.json"))
-            .expect("Failed to read data from configs/peer/genesis.json");
-        let mut genesis: RawGenesisBlock =
-            json5::from_str(&content).expect("Failed to deserialize genesis block from config");
```
mversic, line 81: I'm going to guess you can remove `json5` dependency

data_model/src/validator.rs
```diff
@@ -94,6 +94,16 @@ pub mod model {
     }
 }
 
+impl Validator {
+    /// Create empty mock validator to use as stub in tests
+    // TODO: consider feature like `mock` to provide convenience methods for tests
```
mversic, line 99: You think this is needed for unit tests? For integration tests it's not needed definitely

tools/kagami/README.md
mversic: I'm not fan of this. We still have a privilege to not be backwards compatible in this case. It's added complexity for no tangible reason. Having said that, I don't mind supporting both `kagami genesis` and `kagami genesis --validator-path ./validator.wasm`. But both commands should produce two files IMO

data_model/src/validator.rs
Erigara: Yes, i was thinking about unit tests in genesis crate. Unfortunately ‘#[cfg(test)]‘ doesn't work across crates.

tools/kagami/README.md
Erigara: At the moment we don't produce 2 files, we only producing genesis file with provided path. And validator itself should be generated in separate command.

data_model/src/validator.rs
mversic: it should be possible to write small inline wasm validator that always passes for unit tests. No need for feature in that case

tools/kagami/README.md
Erigara: > I'm not fan of this. We still have a privilege to not be backwards compatible in this case. It's added complexity for no tangible reason. Having said that, I don't mind supporting both `kagami genesis` and `kagami genesis --validator-path ./validator.wasm`. But both commands should produce two files IMO

In any case i don't mind to provide only new behaviour. I just don't want to cause trouble to sdk devs and devops. Because @appetrosyan was against removal of configuration params, so here i thought in the same vein.

mversic: I'm aware of your motivation. As you wish. It's good

Erigara: Let's then wait for other reviewer opinion on this matter.

mversic (dismissed):

data_model/src/transaction.rs
appetrosyan: This needs to be `#[inline]` for move elision to work properly. Also neither of the two methods are very informative. `from_compiled` sounds a lot better than `from_raw` because raw here could mean 
1. The source code of the Verifier, 
2. The results of compiling with the WASM target
3. The results of `wasmtime` JIT.

appetrosyan: In general given that we'd be using WASM for a lot more than just on-chain scripting, we might want to consider making the object hierarchy more pronounced: example, `wasm::SmartContract` and `wasm::Verifier` being two newtype wrappers around blobs, which however, `impl trait wasm::Executable`. Extract common logic into extension traits.

data_model/src/validator.rs
appetrosyan: 1. We decided to remove `mock` from as many crates as possible. This is going backwards. 
2. it's better to have different samples as files than it is to have them be generated for mocking purposes only. 
3. As @mversic has said on several previous occasions, if a state is unrepresentable, tests that work with unrepresentable state are meaningless.

genesis/Cargo.toml
```diff
@@ -15,5 +15,6 @@ iroha_schema = { version = "=2.0.0-pre-rc.13", path = "../schema" }
 derive_more = { version = "0.99.17", default-features = false, features = ["deref"]}
 serde = { version = "1.0.151", features = ["derive"] }
 serde_json = "1.0.91"
+json5 = "0.4.1"
```
appetrosyan, line 18: The whole point behind introducing JSON5 was to make use of comments and other syntactic sugar. I'd suggest either hard forking out genesis block and expanding it with explanatory comments (more work for `kagami` less work for us, when we're trying to explain stuff to users), or making `genesis.json` be strict JSON.

tools/kagami/README.md
appetrosyan: > Because @appetrosyan was against removal of configuration params, so here i thought in the same vein.

That is correct. However, thankfully, here nobody's ever worked with the inline validators outside the core team, so we have flexibility. 

Some software has insane requirements. Sometimes they'll limit sizes of files, sometimes subvolumes that can be mounted shall be restricted. Given that we support both `inline` and included validators, I strongly suggest keeping the functionality in. 

From the perspective of UX, the default should be what we prefer for our CI. I'd add `kagami genesis --inline` as an alternative to `kagami genesis` which warns you that it created a file by default and tells you about the final form: `kagami genesis --compiled-validator-path <path>` for the override.

appetrosyan (dismissed):

tools/kagami/README.md
mversic: then I would make it part of the serialization step. IMO `RawGenesisBlock` should always work with _Inline_ validator, but in `kagami` or during serialization you can convert the file path to inline validator. @Erigara do you think that is possible? That way you would also be truly backwards compatible because with the current approach you have modified schema

mversic: then I would make it part of the serialization step. IMO RawGenesisBlock should always work with Inline validator, but in kagami, during deserialization or in constructor you can convert the file path to inline validator. @Erigara do you think that is possible? That way you would also be truly backwards compatible because with the current approach you have modified schema

two files are then just a convenience

mversic: we can consider deserialization approach for configs and other files to split them up into smaller constituents for reasons such as portability or file size

Erigara: >  do you think that is possible?

@mversic, it was exactly approach i took initially, unfortunately as you said we lose ability to generate proper genesis in kagami and that's why i decide to took another approach. Imo manually move validator into separate file would be pain in the ass.

data_model/src/validator.rs
Erigara: I will remove this method since i observed that we don't really need it.

genesis/Cargo.toml
Erigara: Ok, let's for now make `genesis.json` strict JSON, we can later expand it to JSON5.

tools/kagami/README.md
mversic: fine by me

data_model/src/transaction.rs
Erigara: I've extracted your comment into separate ticket to further discussion (#3443).

mversic (approved):

tools/kagami/README.md
```diff
@@ -126,17 +127,23 @@ You can also choose output format:
 - Generate a genesis block in JSON format:
 
     ```bash
-    kagami genesis
+    kagami genesis --inlined-validator
```
mversic, line 130: what happens if I provide both `--inlined-validator` and `--compiled-validator-path`?

mversic (approved):

scripts/check.sh
```diff
@@ -8,7 +8,7 @@ case $1 in
             exit 1
         };;
     "genesis")
-        cargo run --release --bin kagami -- genesis | diff - configs/peer/genesis.json || {
+        cargo run --release --bin kagami -- genesis --compiled-validator-path ./validator.wasm | diff - configs/peer/genesis.json || {
```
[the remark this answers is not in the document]
appetrosyan: Actually doing the compiling from rust source makes sense, given that there's a specific set of steps with a containerised environment. I'd keep it as is.

appetrosyan (approved):

tools/kagami/README.md
Erigara: Fail with error. Clap support arg groups (accept only one argument).

Issue #3437, opened 2023-05-01
[suggestion] Standardize env variable format


mversic:
### Feature request

Should we prefix all environment variables with `IROHA_`?  ATM sometimes we're using `IROHA_`(like `IROHA_GENESIS_ACCOUNT_PUBLIC_KEY`), sometimes `IROHA2_`(like `IROHA2_CONFIG_PATH`) and sometimes no prefix(like `MAX_LOG_LEVEL`).

IMO we should either:
1. decide on one convention - like `IROHA_` prefix for all env variables
2. allow multiple conventions - like `IROHA_` prefix and no prefix for all env variables

### Motivation

Simple is better than complex. Using different prefixes creates pain points and confusion for the users because they cannot figure out the name of the environment variable just by looking at the config file

### Who can help?

@astrokov7 @outofardis @6r1d @appetrosyan

appetrosyan:
I think no, will provide explanation and close.

PR #3439, opened 2023-05-01
[refactor]: add prefix IROHA_LOG to logger related environment variables

mversic:
### Linked issue

related to #3437 but doesn't close

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -231,7 +231,7 @@ Follow these commit guidelines:
 
 <details> <summary> Expand to learn how to change the log level or write logs to a JSON.</summary>
 
-If one of your tests is failing, you may want to decrease the maximum logging level. By default, Iroha only logs `INFO` level messages, but retains the ability to produce both `DEBUG` and `TRACE` level logs. This setting can be changed either using the `MAX_LOG_LEVEL` environment variable for code-based tests, or using the `/configuration` endpoint on one of the peers in a deployed network.
+If one of your tests is failing, you may want to decrease the maximum logging level. By default, Iroha only logs `INFO` level messages, but retains the ability to produce both `DEBUG` and `TRACE` level logs. This setting can be changed either using the `LOG_MAX_LEVEL` environment variable for code-based tests, or using the `/configuration` endpoint on one of the peers in a deployed network.
 
 While logs printed in the `stdout` are sufficient, you may find it more convenient to produce `json`-formatted logs into a separate file and parse them using either [node-bunyan](https://www.npmjs.com/package/bunyan) or [rust-bunyan](https://crates.io/crates/bunyan).
 
@@ -252,7 +252,7 @@ RUSTFLAGS="--cfg tokio_unstable" cargo build --features tokio-console
 ```
 
 Port for tokio console can by configured through `TOKIO_CONSOLE_ADDR` configuration parameter (or environment variable).
-Using tokio console require log level to be `TRACE`, can be enabled through configuration parameter or environment variable `MAX_LOG_LEVEL`.
+Using tokio console require log level to be `TRACE`, can be enabled through configuration parameter or environment variable `LOG_MAX_LEVEL`.
 
 Example of running iroha with tokio console support using `scripts/test_env.sh`:
 
@@ -260,7 +260,7 @@ Example of running iroha with tokio console support using `scripts/test_env.sh`:
 # 1. Compile iroha
 RUSTFLAGS="--cfg tokio_unstable" cargo build --features tokio-console
 # 2. Run iroha with TRACE log level
-MAX_LOG_LEVEL=TRACE ./scripts/test_env.sh setup
+LOG_MAX_LEVEL=TRACE ./scripts/test_env.sh setup
 # 3. Access iroha. Peers will be available on ports 5555, 5556, ...
 tokio-console http://127.0.0.1:5555
 ```
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ You can find the detailed list of all available endpoints in the [API specificat
 
 By default, Iroha provides logs in a human-readable format and prints them out to `stdout`.
 
-The logging level can be changed either via a [configuration option](./docs/source/references/config.md#loggermax_log_level) or at run-time using the `configuration` endpoint.
+The logging level can be changed either via a [configuration option](./docs/source/references/config.md#loggermax_level) or at run-time using the `configuration` endpoint.
 
 <details><summary>Example: changing log level</summary>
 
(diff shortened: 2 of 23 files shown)


CONTRIBUTING.md
```diff
@@ -231,11 +231,11 @@ Follow these commit guidelines:
 
 <details> <summary> Expand to learn how to change the log level or write logs to a JSON.</summary>
 
-If one of your tests is failing, you may want to decrease the maximum logging level. By default, Iroha only logs `INFO` level messages, but retains the ability to produce both `DEBUG` and `TRACE` level logs. This setting can be changed either using the `MAX_LOG_LEVEL` environment variable for code-based tests, or using the `/configuration` endpoint on one of the peers in a deployed network.
+If one of your tests is failing, you may want to decrease the maximum logging level. By default, Iroha only logs `INFO` level messages, but retains the ability to produce both `DEBUG` and `TRACE` level logs. This setting can be changed either using the `IROHA_LOGGER_MAX_LEVEL` environment variable for code-based tests, or using the `/configuration` endpoint on one of the peers in a deployed network.
```
appetrosyan, line 234: This is better than the way things used to be how exactly? 

## Namespacing

Environment variables are not namespaced. Every systems programmer and systems administrator is well aware of this fact. This means that unlike most programming languages, the OS works under the assumption of overlap. 

Case in point. Environment variables are used to express common preferences: `export NO_COLOR=true` disables all colour for all applications, which is what we added to Iroha recently. As such, the more standard APIs and environment variables we use, the better. `RUST_LOG=INFO` is standard, whether we like it or not. If we want to make the process easier for users, it is far better to support some standard, than it is to come up with your own internal logic, which falls in line with the Unix Philosophy. 

As for the specifics. `LOG=INFO` seems the most appropriate in a vacuum. `RUST_LOG=INFO` seems is the standard I think we should support. `MAX_LOG_LEVEL=INFO` is somewhat verbose, but accurate. 

As for the remainder: 
- `IROHA_LOGGER_MAX_LEVEL` — Is unnecessarily verbose. There is no need for the environment variable to be namespaced because there realistically is no circumstance under which the environment variable is used for two applications unintentionally. 
- `IROHA_LOGGER_FILE_PATH` — is unnecessarily verbose _and_ confusing _and_ inaccurate. This isn't the _Logger_'s file path, this is the file path of the file which will become the log or in other words it's the log-file path. I can't see why it can't equally be `LOG_OUTPUT='/tmp/iroha.log'`. 

So generally, I'd invite a discussion before we continue along those lines. The whole point of deserializing from the environment is to allow individualised configuration either in `docker-compose` or on the command line. In this situation, we shouldn't have a bunch of `IROHA_LOGGER_*` environment variables, instead we should have one which contains the equivalent JSON configuration. There should be one for each: `Sumeragi`, `Torii`, and `Kura`. This is a way of keeping the configuration in one place, grouped and structured. Defining extra environment variables individually as overrides is acceptable, but with one caveat. 

Logging configuration is both the most likely to change during deployment, during testing and the one that should be made the most convenient. Namespacing in log configuration is counter productive. 

## Further discussion

@Erigara @QuentinI

mversic: > Case in point. Environment variables are used to express common preferences:

I agree and understand your point. But to me that is not the only concern and the two are in conflict. The intention was (as per #3437) to have easy and consistent mapping between configuration and environment variables, i.e. a naming convention. The idea was that it's better to be verbose than having to figure out how to name environment variable for a parameter that you wish to override in config. This for me is a pain point. Previously, I would just add prefix `IROHA_` and follow the trail in json dictionary. Another argument in favor of naming convention is that it is self documenting.

> IROHA_LOGGER_FILE_PATH — is unnecessarily verbose and confusing and inaccurate

I made the correction to a more sensible name taking `RUST_LOG` as a role model

> IROHA_LOGGER_MAX_LEVEL - Is unnecessarily verbose

verbosity is the unfortunate consequence of having a mapping between config file and environment variable.

This PR doesn't have to be accepted, it is here as an example. For me as a user of `Iroha` the unclear naming convention of environment variables has been only an inconvenience. We can also come up with a different naming convention?

Erigara: Imo we should have standard naming convention as @mversic suggested.
To me it's confusing too to determine which variables are prefixed and which are not. 
As for overlap i think most of the time app is ran in some containerized environment (docket, k8s) where only single app exists, so reusing env vars is not really the case.

Erigara:
What do you think of renaming `IROHA2_CONFIG_PATH` and `IROHA2_GENESIS_PATH` to `IROHA_...`?

Erigara (dismissed):

mversic:
> What do you think of renaming `IROHA2_CONFIG_PATH` and `IROHA2_GENESIS_PATH` to `IROHA_...`?

yes, I believe I agree. This should be addressed in #3437

CONTRIBUTING.md
QuentinI: I think we should just unprefix everything. If colliding env variables really is a problem for someone somewhere we could also automatically support `IROHA_` prefix for everything.

SamHSmith (dismissed):
I see no issue with this change. When I started working on iroha I was expecting env vars to be prefixed.

CONTRIBUTING.md
appetrosyan: Frankly, env vars are designed for you to re-use the names. It's just that simple. Running things with un-prefixed env vars is what env vars were designed to do. If we use them to pass configuration parameters, then we should have `MAX_LOG_LEVEL` and `LOG_OUTPUT` and not `JP.CO.SORAMITSU.HYPERLEDGER.IROHA.LOGGER.MAX_LOG_LEVEL`.

README.md
```diff
@@ -170,7 +170,7 @@ You can find the detailed list of all available endpoints in the [API specificat
 
 By default, Iroha provides logs in a human-readable format and prints them out to `stdout`.
 
-The logging level can be changed either via a [configuration option](./docs/source/references/config.md#loggermax_log_level) or at run-time using the `configuration` endpoint.
+The logging level can be changed either via a [configuration option](./docs/source/references/config.md#loggermax_level) or at run-time using the `configuration` endpoint.
```
mversic, line 173: @outfardis is this a bug?

CONTRIBUTING.md
mversic: @appetrosyan this can be closed?

README.md
outoftardis: no, why? what do you mean? from what I can see the link is updated to match changes in config.md

mversic: I mean is that it's lumped together into `loggermax_level`

outoftardis: yep, it works

appetrosyan:
Given a recent discussion on the config RFC with @0x009922 I'm closing this PR until further notice. Logging behaviour should be controlled by standard env variables, and definitely not prefixed with `IROHA_`.

PR #3455, opened 2023-05-05
[feature] #3454: make LTS the default

### LINKED ISSUE #3454: [feature] make LTS the default
I noticed this issue while resolving a problem with a user

mversic:
## Description

I noticed this while resolving an issue for a user from the community chat
We should target LTS by default

### Linked issue

Closes #3454

### Checklist


configs/peer/legacy_lts/config.json
mversic, line 1: I don't see a point in having these files. If you wish to work with legacy lts, you check out a branch and configs are there. I'd be happy to restore if someone disagrees

Erigara (changes requested):
You delete `configs/peer/legacy_lts`, but `docker-compose.yml` still try to mount it, no?

mversic:
> You delete `configs/peer/legacy_lts`, but `docker-compose.yml` still try to mount it, no?

yeah, stupid of me. Fixed now

configs/peer/update_configs.sh
```diff
@@ -1,8 +1,6 @@
 #!/bin/sh
-# This script is intended for release updates,
-# when LTS and Stable branch configurations may change.
-# It downloads the previous configurations from raw.githubusercontent.com.
```
mversic, line 4: this was wrong because docker image ID would  would be pointing to `hyperledger/iroha2:lts` but configs would be from previous lts. This is why when updating docker images we have to make sure that they are updated in this order exactly:
1. lts
2. stable
3. dev

.github/workflows/iroha2-release.yml
```diff
@@ -52,22 +52,39 @@ jobs:
...
     runs-on: ubuntu-latest
     container:
       image: hyperledger/iroha2-ci:nightly-2022-12-22
+    steps:
+      - uses: actions/checkout@v3
+        ref: iroha2-dev
+      - name: Setup git config
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
```
mversic, line 65: is this ok for DCO rules? IMO it should be. We could replace with a dummy address like [here](https://github.com/orgs/community/discussions/41101)

mversic, line 65: is this ok for DCO rules? IMO it should be. We could replace with a dummy address like [here](https://github.com/orgs/community/discussions/41101)

configs/peer/update_legacy_configs.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/sh
-# This script is intended for release updates,
-# when LTS and Stable branch configurations may change.
-# It downloads the previous configurations from raw.githubusercontent.com.
```
mversic, line 4: this was wrong because docker image ID would would be pointing to `hyperledger/iroha2:lts` but configs could be from previous lts or current lts. It would depend on whether the current release is becoming lts.

.github/workflows/iroha2-release.yml
```diff
@@ -52,10 +52,28 @@ jobs:
...
+      - name: Update configs
+        run: |
+          ./scripts/update_configs.sh lts
+          ./scripts/update_configs.sh stable
+      - name: Commit config changes
+        run: |
+          git commit -am "[documentation]: Update lts/stable configs following a release" --signoff
+          git push origin iroha2-dev
+
+  load-rs:
```
appetrosyan, line 75: I'd skip this (or allow failures) until @BAStos525 fixes the workflow on the side of the load program. False-positives are a problem.

appetrosyan (dismissed):

docker-compose.dev.yml
0x009922, line 1: All 4 peers will share the same storage located in `/app/soramitsu/iroha`, wouldn't they?

mversic: Yes, it's so that they can access source code

appetrosyan (approved):

0x009922 (approved):

PR #3474, opened 2023-05-12
[fix] #3444: Remove custom discriminants from scale encoding

### LINKED ISSUE #3444: [BUG] Versioned structures in the schema miss discriminant info
### GIT commit hash

b1855454

### Description

After the merge of #3317, the schema doesn't contain info about enum discriminants anymore. It was discussed in a topic (https://github.com/hyperledger/iroha/pull/3317#discussion_r1159516469). I thought that if discriminants aren't explicitly specified anymore, then the default sequential discriminants computation should be used. It turned out not to be the case: `VersionedSignedTransaction` and `VersionedSignedQuery` still have custom discriminants (`1` for `V1`) while it is not specified in the schema:

https://github.com/hyperledger/iroha/blob/b1855454ac03603fb62a3a6491b5492ab26f9777/docs/source/references/schema.json#L3978-L3993

### What is expected

It is expected that the schema should not have such implitic details at all. If there are custom discriminants for enum variants, they should be specified.

### Desired solution

I think that it is not necessary to specify discriminants for all enums, it would be too verbose. However, optional `discriminant` field should exist for such a case. The approach of discriminants computation should be the same as it works in Rust ([Enumerations > Discriminants - The Rust Reference](https://doc.rust-lang.org/reference/items/enumerations.html#discriminants)). For example:

```rust
enum Foo {
    Bar,            // 0
    Baz = 123,      // 123
    Quux,           // 124
}
```

```json5
{
  Foo: {
    Enum: [
      { name: 'Bar' },                     // 0
      { name: 'Baz', discriminant: 123 },  // 123
      { name: 'Quuz' }                     // 124
    ]
  }
}
```

mversic:
## Description

This is a very contentious issue for sdk devs. Let's see what they think as well
This PR might be rejected

### Linked issue

Closes #3444

### Checklist


Erigara (dismissed):

0x009922 (commented):
After a short discussion with @appetrosyan, we decided that the solution would be not to remove custom discriminants, but to specify them explicitly for a whole enum in which they are custom.

appetrosyan (approved):
It looks like this PR should have two commits instead of one, changing the behaviour of `#[codec(skip)]` isn't exactly in the scope of #3444

appetrosyan (dismissed):
It looks like this PR should have two commits instead of one, changing the behaviour of `#[codec(skip)]` isn't exactly in the scope of #3444

appetrosyan:
@0x009922 This PR can be merged, it just doesn't close #3444. For that we need some mechanism to determine if the SCALE-assigned discriminant is trivially computable. This can be done with property-based testing. 

**Or** we could add a flag to `kagami schema` that adds discriminant information to the schema.

appetrosyan (approved):

Arjentix (approved):

Arjentix (approved):

appetrosyan:
Given that we have resolved the issues on the SDK side, and this does not fully address #3444 I believe it is useful to just close this PR.

Issue #1615, opened 2021-11-16
testing techniques: failure points


appetrosyan:
As per #1592, [fail](https://docs.rs/fail/0.5.0/fail/) is a good candidate to test several modes of failure that bring down our test coverage, particularly in case of error correction code that isn't being tested, because the error conditions never arise.

PR #3477, opened 2023-05-12
[fix] #3233: Refactor error types

### LINKED ISSUE #3233: [BUG] Iroha 2 error messages are duplicated
### OS and Environment

MacOS, Docker Hub

### GIT commit hash

b1855454

1. Try to register an existing domain using the client cli `./iroha_client_cli domain register --id=job`
2. Observe the duplicated and unclear error messages in the console output.

### Actual result

The error messages are duplicated and unclear, making it difficult to understand the issue.
```
Error: 
   0: Failed to create domain
   1: Failed to submit transaction SignedTransaction { payload: TransactionPayload { account_id: alice@wonderland, instructions: [RegisterBox { object: NewDomain(NewDomain { id: DomainId { name: "job" }, logo: None, metadata: Metadata { map: {} } }) }], creation_time: 1682518246134, time_to_live_ms: 100000, nonce: None, metadata: {} }, signatures: {iroha_crypto::signature::SignatureOf<iroha_data_model::transaction::TransactionPayload>({ pub_key: ed01207233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0, payload: A1B60609A79A23035C0EEEFAD60549BA7C7806B2C81DF32E2A6D919605098862ECA4DB28667A4A2DADD85A599DF7CE01FC8896973BDA794799BF3A0AA7C0A703 })} }
   2: Transaction rejected
   3: Transaction was rejected: Transaction rejected due to insufficient authorisation: Action not permitted: WASM error: Exported function call failed: Repetition
      wasm backtrace:
          0: 0x2416c - <unknown>!iroha_wasm::encode_and_execute::h7cdcd596fd856adc
          1: 0x2413a - <unknown>!<iroha_data_model::isi::InstructionBox as iroha_wasm::ExecuteOnHost>::execute::h4959493f58ba6043
          2: 0x2045e - <unknown>!iroha_default_validator::validate_and_execute_instruction::h0bbb69e9d4fc5d5d
          3: 0x203e2 - <unknown>!iroha_default_validator::validate_and_execute_transaction::h04877105a9176dd0
          4: 0x20969 - <unknown>!iroha_default_validator::validate::h4140f3a88639d737
          5: 0x208e1 - <unknown>!_iroha_wasm_main
   3: 
   4: Transaction rejected due to insufficient authorisation: Action not permitted: WASM error: Exported function call failed: Repetition
      wasm backtrace:
          0: 0x2416c - <unknown>!iroha_wasm::encode_and_execute::h7cdcd596fd856adc
          1: 0x2413a - <unknown>!<iroha_data_model::isi::InstructionBox as iroha_wasm::ExecuteOnHost>::execute::h4959493f58ba6043
          2: 0x2045e - <unknown>!iroha_default_validator::validate_and_execute_instruction::h0bbb69e9d4fc5d5d
          3: 0x203e2 - <unknown>!iroha_default_validator::validate_and_execute_transaction::h04877105a9176dd0
          4: 0x20969 - <unknown>!iroha_default_validator::validate::h4140f3a88639d737
          5: 0x208e1 - <unknown>!_iroha_wasm_main
   4: 
   5: Action not permitted: WASM error: Exported function call failed: Repetition
      wasm backtrace:
          0: 0x2416c - <unknown>!iroha_wasm::encode_and_execute::h7cdcd596fd856adc
          1: 0x2413a - <unknown>!<iroha_data_model::isi::InstructionBox as iroha_wasm::ExecuteOnHost>::execute::h4959493f58ba6043
          2: 0x2045e - <unknown>!iroha_default_validator::validate_and_execute_instruction::h0bbb69e9d4fc5d5d
          3: 0x203e2 - <unknown>!iroha_default_validator::validate_and_execute_transaction::h04877105a9176dd0
          4: 0x20969 - <unknown>!iroha_default_validator::validate::h4140f3a88639d737
          5: 0x208e1 - <unknown>!_iroha_wasm_main
   5: 

Location:
   client/src/client.rs:525

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets. 
```

### Expected result

The error message should be concise, clear, and without duplication, providing a clear understanding of the issue. For example:
```
Error: Failed to register domain 'job'
Reason: Domain already exists
Location: client/src/client.rs:525
```

@astrokov7

### Notes
For some reason an error message from validator returned to user (`iroha_client`) contains error **3** times instead of one. Need to fix this.

Arjentix:
### Linked issue

- Closes #3233 

### Benefits

Old error message:

```
Error: Transaction rejected

Caused by:
   0: Transaction was rejected: Transaction rejected due to insufficient authorisation: Action not permitted: Validator denied the operation `SignedTransaction { payload: TransactionPayload { account_id: admin@admin, instructions: [TransferBox { source_id: AssetId(rose##alice@wonderland), object: 1_u32, destination_id: AccountId(admin@admin) }], creation_time: 1683932346090, time_to_live_ms: 100000, nonce: None, metadata: {} }, signatures: {iroha_crypto::signature::SignatureOf<iroha_data_model::transaction::model::TransactionPayload>({ pub_key: ed0120419AAFB36674C689D2B612EC6903364DDDB0CE8FF0F4031D8E3F8ACE88F42B5F, payload: 9ACCE845B6CD44ED13A14915528B3F09BECFF36E14765D426A0276F9B986340513B7B8CB7D2CB3DC9C080387EEA81472DCA5EA37E3E7C7BD2465D9A0B275B50D })} }`: `Can't transfer assets of another account`
   1: Transaction rejected due to insufficient authorisation: Action not permitted: Validator denied the operation `SignedTransaction { payload: TransactionPayload { account_id: admin@admin, instructions: [TransferBox { source_id: AssetId(rose##alice@wonderland), object: 1_u32, destination_id: AccountId(admin@admin) }], creation_time: 1683932346090, time_to_live_ms: 100000, nonce: None, metadata: {} }, signatures: {iroha_crypto::signature::SignatureOf<iroha_data_model::transaction::model::TransactionPayload>({ pub_key: ed0120419AAFB36674C689D2B612EC6903364DDDB0CE8FF0F4031D8E3F8ACE88F42B5F, payload: 9ACCE845B6CD44ED13A14915528B3F09BECFF36E14765D426A0276F9B986340513B7B8CB7D2CB3DC9C080387EEA81472DCA5EA37E3E7C7BD2465D9A0B275B50D })} }`: `Can't transfer assets of another account`
   2: Action not permitted: Validator denied the operation `SignedTransaction { payload: TransactionPayload { account_id: admin@admin, instructions: [TransferBox { source_id: AssetId(rose##alice@wonderland), object: 1_u32, destination_id: AccountId(admin@admin) }], creation_time: 1683932346090, time_to_live_ms: 100000, nonce: None, metadata: {} }, signatures: {iroha_crypto::signature::SignatureOf<iroha_data_model::transaction::model::TransactionPayload>({ pub_key: ed0120419AAFB36674C689D2B612EC6903364DDDB0CE8FF0F4031D8E3F8ACE88F42B5F, payload: 9ACCE845B6CD44ED13A14915528B3F09BECFF36E14765D426A0276F9B986340513B7B8CB7D2CB3DC9C080387EEA81472DCA5EA37E3E7C7BD2465D9A0B275B50D })} }`: `Can't transfer assets of another account`
```

New error message:

```
Error: Transaction was rejected

Caused by:
   0: Validation failed
   1: Not permitted
   2: Validator denied the operation [TRANSFER `1` FROM `rose##alice@wonderland` TO `admin@admin`] by `admin@admin`
   3: Can't transfer assets of another account
```

### Further work

We should return a more detailed error from validator. I.e. it should explain on which instruction validation of transaction failed.
Also we should try to avoid using `Trap` because hard to work with when trying to analyze errors.

### Checklist


core/src/smartcontracts/wasm.rs
```diff
@@ -69,25 +69,80 @@ pub mod import {
...
-    #[error("Module instantiation failure: {0}")]
+    #[error("Module instantiation failure")]
     Instantiation(#[source] eyre::Report),
     /// Expected named export not found in module
-    #[error("Named export not found: {0}")]
+    #[error("Named export not found")]
     ExportNotFound(#[source] eyre::Report),
+    /// Error exporting memory from wasm to iroha
+    #[error("Failed to export memory")]
+    MemoryExport(#[from] MemoryExportError),
```
mversic, line 82: this should be under `Initialization` or `Instantiation`? Also, this error is hardly observable because it would mean someone has tapped into our framework

core/src/smartcontracts/wasm.rs
```diff
@@ -69,25 +69,80 @@ pub mod import {
...
+    HostExecution(#[source] eyre::Report),
+    /// Stack overflow, heap overflow or other limits exceeded
+    #[error("Execution limits exceeded")]
+    ExecutionLimitsExceeded(#[source] eyre::Report),
+    /// Some integer error like division by zero or overflow
+    #[error("Numerical error")]
+    Integer(#[source] eyre::Report),
+    /// Other kind of trap
+    #[error("Other")]
+    Other(#[source] Trap),
```
mversic, line 116:
```suggestion
    Other(#[source] eyre::Report),
```

please don't expose `Trap`

core/src/smartcontracts/wasm.rs
```diff
@@ -69,25 +69,80 @@ pub mod import {
...
+    }
+}
+
+// TODO: We probably don't need this and should return [`Error`]
+// from imported functions like `execute_instruction`
+impl From<Error> for Trap {
+    fn from(error: Error) -> Self {
+        Trap::new(error.to_string())
+    }
+}
```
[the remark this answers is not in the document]
mversic: > Yes, we should return `Result` from `execute_instruction` and `execute_query`

Would you consider making this change in this PR? I don't think it would be too involving

data_model/src/transaction.rs
```diff
@@ -1046,6 +1104,31 @@ pub mod error {
...
+            Hash,
+            Decode,
+            Encode,
+            Deserialize,
+            Serialize,
+            IntoSchema,
+        )]
+        #[ffi_type(opaque)]
+        #[cfg_attr(feature = "std", derive(thiserror::Error))]
+        pub enum AccountDoesNotExistError {
```
mversic, line 1123: we already have `FindError`. Can we use it please?

mversic (changes requested):

data_model/src/transaction.rs
```diff
@@ -985,14 +1012,45 @@ pub mod error {
...
-        #[serde(transparent)]
-        #[repr(transparent)]
-        // SAFETY: `NotPermittedFail` has no trap representation in `String`
-        #[ffi_type(unsafe {robust})]
-        pub struct NotPermittedFail {
-            /// The cause of failure.
-            pub reason: String,
+        #[ffi_type(opaque)]
+        #[cfg_attr(feature = "std", derive(thiserror::Error))]
+        pub enum ValidationFailure {
```
mversic, line 1017: I see some are named with suffix `Fail` and some with `Failure`

core/src/smartcontracts/wasm.rs
Arjentix: Currently we're trying to get this export i.e. on every call to `execute_query` or `execute_instruction`, so it's not an `Initialization` or `Instantiation` error.

But I'll try to refactor this

core/src/smartcontracts/wasm.rs
Arjentix: It was exposed before, but ok, I'll change it

core/src/smartcontracts/wasm.rs
```diff
@@ -69,25 +69,80 @@ pub mod import {
...
+pub enum ExportFnCallError {
+    /// Failed to execute something on the host side
+    #[error("Failed to execute operation")]
+    HostExecution(#[source] eyre::Report),
+    /// Stack overflow, heap overflow or other limits exceeded
+    #[error("Execution limits exceeded")]
+    ExecutionLimitsExceeded(#[source] eyre::Report),
+    /// Some integer error like division by zero or overflow
+    #[error("Numerical error")]
+    Integer(#[source] eyre::Report),
```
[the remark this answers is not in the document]
Arjentix: Hm, I did it probably because `TrapCode` have 3 integer related error codes. But in fact this isn't used at our side.

Should I move these error to `Other`? Or it doesn't hurt?

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
Arjentix: > Would you consider making this change in this PR? I don't think it would be too involving

Ok, will do

core/src/smartcontracts/wasm.rs
mversic: ok, yeah. But can we get this error after initialization?

data_model/src/transaction.rs
Arjentix: I used it at the beginning but then decided to create another type, because `FindError` contains too much variants which are meaningless in this place

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
mversic: I'm not sure. If you leave it I would like a better description of what this error is

data_model/src/transaction.rs
Arjentix: Will replace with `ValidationFail`

core/src/smartcontracts/wasm.rs
Arjentix: I will try to call `get_memory()` once and store it in state, so that we can't get this error at real runtime

data_model/src/transaction.rs
mversic: I would use `FindError`. It's not unusual to have a big error variant that encompasses all possible errors but which cannot occur for some system. Just take a look at `std::io::ErrorKind`, it contains errors for both network and filesystem and these don't overlap entirely

mversic: and it's easier for the user to be dealing with a smaller number of items

Arjentix: > It's not unusual to have a big error variant that encompasses all possible errors but which cannot occur for some system.

Yeah, I know, but I'm just not a fan of that approach. I'd better do something like [this](https://sabrinajewson.org/blog/errors) (although it's too verbose)

core/src/smartcontracts/wasm.rs
Arjentix: Okay, actually it's not so trivial, because to get memory we should already have `State` instantiated, so it's not possible to store `wasmtime::Memory` without using `Option` which I don't like here.

And still we could get this error after initialization (it's wasm engine creation time) and after instantiation (it's concrete wasm blobl instantiation time). So I'd better keep it like it is.

> Also, this error is hardly observable because it would mean someone has tapped into our framework

Yes, but we still should handle this type of error, because user can submit any wierd wasm code. That's why I call it `InternalError` but I do not use panic, because it's not *Iroha* bug, it's admin mistake.

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
Arjentix: Ok, let's remove it because it's useless

core/src/smartcontracts/wasm.rs
Arjentix: Why are you sure that on wasm intialization step (which one, actually?) wasmtime also checks for `memory` export?

Arjentix: I think if it would be like so then there would be a way to just call `memory()` on something

mversic: I think you are wrong. The access to memory is name based, it's not engraved into the type system. The compiler cannot know if any particular access is correct. But we have established this invariant when the module was instantiated so we don't need to worry about it anymore. 

Also, if this error is possible at runtime, it is very appropriate to just `Trap`. This error cannot be handled in the program in any reasonable way

mversic: I think you are wrong. The access to memory is name based, it's not engraved into the type system. The compiler cannot know if any particular access is correct. But we have established this invariant when the module was instantiated so we don't need to worry about it anymore. 

Also, if this error is possible at runtime, it is very appropriate to just `Trap`. This error cannot be handled in the program in any reasonable way

Arjentix: >  But we have established this invariant when the module was instantiated so we don't need to worry about it anymore.

I don't see that in the code. We don't check that `memory` is a really valid memory at instantiation time.
Do you suggest to implement it and then just panic if we can't get memory later?

data_model/src/transaction.rs
```diff
@@ -1065,30 +1148,33 @@ pub mod error {
         // TODO: Temporarily opaque
         #[ffi_type(opaque)]
         pub enum TransactionRejectionReason {
+            /// Account does not exist
+            AccountDoesNotExist(#[cfg_attr(feature = "std", source)] AccountDoesNotExistError),
```
Erigara, line 1152: I don't see `#[display]` for this variant, is it done on purpose?

data_model/src/transaction.rs
Erigara: I would vote for `FindError`, because otherwise we bring additional types into the schema. 
And as i see it we use `TransactionRejectionReason` for error reporting not error handling, so redundant variants doesn't seem to be huge deal.

core/src/smartcontracts/wasm.rs
```diff
@@ -69,25 +69,80 @@ pub mod import {
...
-    #[error("Runtime initialization failure: {0}")]
+    #[error("Runtime initialization failure")]
     Initialization(#[source] eyre::Report),
     /// Module could not be compiled or instantiated
-    #[error("Module instantiation failure: {0}")]
+    #[error("Module instantiation failure")]
     Instantiation(#[source] eyre::Report),
     /// Expected named export not found in module
-    #[error("Named export not found: {0}")]
+    #[error("Named export not found")]
```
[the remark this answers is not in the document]
appetrosyan: I prefer this style in the code-base. This is much simpler and gives more freedom in some situations.

core/src/smartcontracts/wasm.rs
appetrosyan: I would recommend we revisit this after #1615 is implemented. If the code is dead, we should see it in full-coverage failure point test. 

I agree with @mversic about the general sentiment, I just don't think that reasoning about code with regards to memory safety is sufficient (If it were, Rust wouldn't exist). 

So I would recommend splitting this off into a separate ticket and closing the conversation.

data_model/src/transaction.rs
appetrosyan: `FindError` is my preference as well. There might be need for special handling of accounts after the new schema is introduced, which I'll do later.

cli/src/torii/routing.rs
```diff
@@ -76,7 +77,17 @@ impl VerifiedQuery {
...
+                let error = ValidationFail::from(error);
+                if let ValidationFail::InternalError(mes) = &error {
+                    error!(
+                        error = mes,
+                        "Internal error occurred during query validation, \
+                         is Runtime Validator correct?"
+                    )
+                }
+                error
+            })?;
```
0x009922, line 90: Producing side effects in a mapping function isn't very nice, although it looks clear

Arjentix: Yes, but function that should be ok with side effects is called `inspect` and it's not stabilized yet. So for now I think we have no other choice

core/src/smartcontracts/wasm/codec/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+///
+/// Usually, to retrieve the encoded object both pointer and the length of the allocation
+/// are provided. However, due to the lack of support for multivalue return values in stable
+/// `WebAssembly` it's not possible to return two values from a wasm function without some
+/// shenanignas. In those cases, only one value is sent which is pointer to the allocation
+/// with the first element being the length of the encoded object following it.
+pub fn encode_with_length_prefix<T: Encode>(obj: &T) -> Vec<u8> {
+    // Compile-time size check
+    #[allow(clippy::let_unit_value)]
+    let _ = <SizeChecker<T> as DoesSizeFitInWasmUsize>::RESULT;
```
mversic, line 111: this is cool

core/src/smartcontracts/wasm/codec/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// Helper struct to check type size at compile time.
+struct SizeChecker<T>(std::marker::PhantomData<T>);
+
+impl<T> DoesSizeFitInWasmUsize for SizeChecker<T> {
+    const RESULT: () = assert!(
+        core::mem::size_of::<T>() < (u32::MAX as usize - core::mem::size_of::<WasmUsize>()),
+        "Size doesn't fit in 2^32"
+    );
+}
```
mversic, line 142:
```suggestion
/// Helper struct to check type size at compile time.
struct SizeChecker<T>(std::marker::PhantomData<T>);

impl<T> SizeChecker<T> {
    const RESULT: () = assert!(
        core::mem::size_of::<T>() < (u32::MAX as usize - core::mem::size_of::<WasmUsize>()),
        "Size doesn't fit in 2^32"
    );
}
```

make it associated const

core/src/smartcontracts/wasm/codec/src/lib.rs
mversic: is this const expression guaranteed to be evaluated. Does it matter that the result is bound to a placeholder?

core/src/smartcontracts/wasm/mod.rs
```diff
@@ -223,90 +323,54 @@ pub struct Runtime<'wrld> {
...
-        let alloc_fn = Self::get_alloc_fn(&mut caller)?;
-        let memory = Self::get_memory(&mut caller)?;
-
-        let query: QueryBox = Self::decode_from_memory(&memory, &caller, offset, len)?;
-
-        let wsv = caller.data().wsv;
+    /// Execute `query` on host
+    #[allow(clippy::needless_pass_by_value)]
+    #[codec::wrap]
+    fn execute_query(query: QueryBox, state: &State) -> Result<Value, QueryExecutionFailure> {
```
mversic, line 349:
```suggestion
    fn execute_query(state: &State, query: QueryBox) -> Result<Value, QueryExecutionFailure> {
```

I would rather reverse it and say that args[0] is always `State`

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+///
+/// Describing all possible input and output signatures would be a very big table,
+/// so see detailed signature by expanding generated code (i.e. with `cargo expand`).
+///
+/// # Key notes
+///
+/// 1. If there is something to (de-)code (input or output) generated signature will always
+/// return `Result<..., Trap>`.
+/// 2. If your function returns `T` on success, then generated function will
+/// return `Result<WasmUsize, Trap>`, where `WasmUsize` is the offset of encoded `T`
```
mversic, line 23: offset prefixed with length

mversic (changes requested):

core/src/smartcontracts/wasm/mod.rs
mversic: irrelevant

wasm/src/log.rs
```diff
@@ -55,7 +55,7 @@ mod host {
         /// # Warning
         ///
         /// This function doesn't take ownership of the provided allocation
-        pub(super) fn log(log_level: u8, ptr: *const u8, len: usize);
+        pub(super) fn log(ptr: *const u8, len: usize);
```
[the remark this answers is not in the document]
mversic: not very relevant

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -16,7 +16,8 @@ use iroha_wasm::{data_model::prelude::*, prelude::*};
 fn main() {
     iroha_wasm::info!("Executing trigger");
 
-    let accounts = FindAllAccounts.execute();
+    let accounts = FindAllAccounts.execute().dbg_unwrap();
```
mversic, line 19: we don't allow unwrap in the code base. What do others think?

data_model/src/asset.rs
```diff
@@ -191,6 +191,7 @@ pub mod model {
         Eq,
         PartialOrd,
         Ord,
+        Hash,
```
mversic, line 194: can you point out where is this necessary?

data_model/src/isi.rs
```diff
@@ -910,22 +910,32 @@ pub mod error {
...
     pub mod model {
+        use core::ops::RangeInclusive;
+
         use super::*;
+        use crate::ValidationFail;
+
+        /// Error type used only when WASM smart contract fails to execute instruction
+        #[derive(Debug, Display, PartialEq, Eq, FromVariant, Encode, Decode)]
+        #[cfg_attr(feature = "std", derive(thiserror::Error))]
+        pub enum WasmInstructionRejectionReason {
```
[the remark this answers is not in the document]
mversic: IMO `InstructionExecutionFailure` should be just a variant of `ValidationFail`

wasm/src/lib.rs
```diff
@@ -51,18 +52,18 @@ unsafe extern "C" fn _iroha_wasm_dealloc(offset: *mut u8, len: usize) {
...
     /// Execute instruction on the host
-    fn execute(&self);
+    fn execute(&self) -> Result<(), WasmInstructionRejectionReason>;
 }
 
 /// Implementing queries can be executed on the host
 pub trait QueryHost: Query {
     /// Execute query on the host
-    fn execute(&self) -> Self::Output;
+    fn execute(&self) -> Result<Self::Output, QueryExecutionFailure>;
```
mversic, line 61: query can fail validation as well

data_model/src/lib.rs
```diff
@@ -839,8 +840,8 @@ pub mod model {
         Identifiable(IdentifiableBox),
         PublicKey(PublicKey),
         SignatureCheckCondition(SignatureCheckCondition),
-        TransactionValue(TransactionValue),
-        TransactionQueryResult(TransactionQueryResult),
+        TransactionValue(Box<TransactionValue>),
+        TransactionQueryResult(Box<TransactionQueryResult>),
```
mversic, line 844: what happened here?

data_model/src/isi.rs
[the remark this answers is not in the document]
mversic: this is a code smell to me. We should know are we validating or executing an instruction beforehand.
Let's leave as is but keep in mind. I think this distinction will become more clear in the future

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
Arjentix: It's an example of a smartcontract, and here it has no way to recover from this error. So why not?

wasm/src/lib.rs
mversic: let's disregard this atm

core/src/smartcontracts/wasm.rs
Arjentix: This error type was refactored, so check out new version and leave a comment for it. Closing this

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
Arjentix: So I taked 2 weeks but I did it. Check out new code. Closing this

data_model/src/transaction.rs
Arjentix: Done

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
Arjentix: This error type was refactored, check out new code. Closing this

core/src/smartcontracts/wasm.rs
```diff
@@ -270,10 +325,17 @@ impl<'wrld> Runtime<'wrld> {
         let wsv = caller.data().wsv;
         wsv.validator_view()
             .validate(wsv, &caller.data().account_id, query.clone())
-            .map_err(|error| NotPermittedFail {
-                reason: error.to_string(),
+            .map_err(|error| {
+                let error = ValidationFailure::from(error);
+                if let ValidationFailure::InternalError(mes) = &error {
```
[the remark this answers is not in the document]
Arjentix: Agree about`msg`. I always forget this name. Refactored all `mes` to `msg` in our codebase

data_model/src/transaction.rs
Arjentix: It was a mistake, thanks. Btw I wonder why it compiles

data_model/src/isi.rs
```diff
@@ -957,35 +967,34 @@ pub mod error {
...
         pub enum EvaluationError {
             /// Asset type assertion error
-            #[display(fmt = "Failed to evaluate expression: {_0}")]
-            Expression(evaluate::Error),
+            #[display(fmt = "Failed to evaluate expression")]
+            Expression(#[cfg_attr(feature = "std", source)] evaluate::Error),
             /// Parameter type assertion error
-            #[display(fmt = "Instruction not supported: {_0}")]
+            #[display(fmt = "Instruction of type `{_0}` is not supported")]
             Unsupported(InstructionType),
```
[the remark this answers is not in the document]
Arjentix: Indeed, removed!

core/src/smartcontracts/isi/account.rs
```diff
@@ -314,7 +317,11 @@ pub mod isi {
                     error!(%permission, "Revoking non-existent token");
```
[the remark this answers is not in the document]
Arjentix: Removed logging and added another `return Err`. Yes, I think It's two different error cases

data_model/src/query.rs
```diff
@@ -1420,7 +1420,45 @@ pub mod error {
...
+            PartialEq,
+            Eq,
+            Hash,
+            Serialize,
+            Deserialize,
+            Decode,
+            Encode,
+            IntoSchema,
+        )]
+        // TODO: Only temporary
```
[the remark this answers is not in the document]
Arjentix: I think it was a copy-paste mistake. This comment was related to `#[ffi_type(opaque)]` and it looks like there should be no such todo and it should be just `#[ffi_type]`

wasm/src/lib.rs
```diff
@@ -146,7 +146,7 @@ pub fn query_triggering_event() -> Event {
 /// # Traps
 ///
 /// Host side will generate a trap if this function was not called from a validator.
-pub fn query_operation_to_validate() -> NeedsValidationBox {
+pub fn query_operation_to_validate() -> Option<NeedsValidationBox> {
```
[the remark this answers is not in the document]
Arjentix: Because previously we used to just return `Trap` if `State` does not contains such operation.
And I decided to make this handling on user-side. Like recoverable error

data_model/src/lib.rs
```diff
@@ -904,6 +905,85 @@ pub mod model {
...
+        #[cfg(feature = "transparent_api")]
+        InternalError(
+            /// Contained error message if its used internally.
+            /// Never serialized to not to expose internal errors to the end user.
+            #[codec(skip)]
+            #[serde(skip)]
+            String,
+        ),
+        #[cfg(not(feature = "transparent_api"))]
+        InternalError,
```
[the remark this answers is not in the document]
Arjentix: Ok, then I will just never (de-)serialize `String` inside of it. So that it will be filled inside Iroha but will be empty for clients

data_model/src/isi.rs
```diff
@@ -1067,22 +1076,23 @@ pub mod error {
...
             /// Invalid WASM binary
             #[display(fmt = "Invalid WASM binary: {_0}")]
             Wasm(String),
+            /// Invalid name length (i.e. too long [`AccountId`])
+            #[display(
+                fmt = "Name must be between {} and {} (inclusive) characters in length",
+                "_0.start()",
+                "_0.end()"
+            )]
+            NameLength(RangeInclusive<u32>),
```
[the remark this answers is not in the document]
Arjentix: I can't find where we have it

core/src/smartcontracts/wasm/codec/src/lib.rs
Arjentix: Nice, thanks

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// Macro to wrap function with normal parameters and return value to another one which will
+/// feet `wasmtime` specifications.
+///
+/// Describing all possible input and output signatures would be a very big table,
+/// so see detailed signature by expanding generated code (i.e. with `cargo expand`).
+///
+/// # Key notes
+///
+/// 1. If there is something to (de-)code (input or output) generated signature will always
```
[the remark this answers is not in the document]
Arjentix: Changed to `encode or decode`. Ok?

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+/// return `Result<..., Trap>`.
+/// 2. If your function returns `T` on success, then generated function will
+/// return `Result<WasmUsize, Trap>`, where `WasmUsize` is the offset of encoded `T`
+/// 3. If your function returns [`Result`] with `Trap` on [`Err`], generated function will pop it up
```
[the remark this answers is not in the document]
Arjentix: I also thought so at the beginning. But then I realized that users may also want to return *unrecoverable errors from smart contract point of view*. 
his is done for `log` function where user can submit inexistent level number.

data_model/src/lib.rs
Arjentix: Yeah, so now without `Box`'s compiler will fail to calculate `Value` size.

Explanation for `TransactionValue`
```
`<-`  means "contains"

Value <- TransactionValue <- VersionedRejectedTransaction <- RejectedTransaction <- TransactionRejectionReason <- ValidationFail <- InstructionExecutionFailure <- InstructionEvaluationError <- (new) TypeError <- Mismatch<Value>
```

QuentinI: Assuming error is an exceptional case rather than the norm, maybe it would be more efficient to bring Box closer to the end of the chain?

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+use std::ops::Deref;
+
+use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
+use proc_macro_error::{abort, diagnostic, proc_macro_error, Diagnostic, Level, OptionExt as _};
+use quote::quote;
+use syn::{parse_quote, punctuated::Punctuated};
+
+/// Macro to wrap function with normal parameters and return value to another one which will
+/// feet `wasmtime` specifications.
```
QuentinI, line 13:
```suggestion
/// meet `wasmtime` specifications.
```
🦶🦶

QuentinI (changes requested):

wasm/src/lib.rs
```diff
@@ -203,7 +214,7 @@ mod host {
         ///
         /// This function doesn't take ownership of the provided allocation
         /// but it does transfer ownership of the result to the caller
-        pub(super) fn execute_instruction(ptr: *const u8, len: usize);
+        pub(super) fn execute_instruction(ptr: *const u8, len: usize) -> *const u8;
```
Erigara, line 217: What is returned value here?

Erigara (commented):
Looks good to me, rebase and apply @QuentinI comments and ping me for reapproval.

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
mversic: but user should never return `Trap`. Codec should modify the function so that it returns trap?

core/src/smartcontracts/wasm/mod.rs
```diff
@@ -1,22 +1,22 @@
 //! This module contains logic related to executing smartcontracts via
 //! `WebAssembly` VM Smartcontracts can be written in Rust, compiled
 //! to wasm format and submitted in a transaction
-#![allow(clippy::expect_used, clippy::doc_link_with_quotes)]
 #![allow(
-    clippy::arithmetic_side_effects,
-    clippy::std_instead_of_core,
-    clippy::std_instead_of_alloc
+    clippy::expect_used,
```
mversic, line 5: not needed

wasm/validator/src/lib.rs
```diff
@@ -55,6 +54,19 @@ macro_rules! deny {
     };
 }
 
+/// Like `?` but uses [`deny`] macro instead of `Err`.
+#[macro_export]
+macro_rules! propagate {
```
mversic, line 59: I would inline this to keep it simple. It just confuses users and makes us have to explain one more thing in an API that is complicated as it is

wasm/src/lib.rs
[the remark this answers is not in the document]
mversic: I'm not sure I agree with this. This indicates that smart contract is incorrectly implemented or that we have a bug in Iroha. This is not recoverable IMO

The only way I can think of how someone could use this is to have the same smart contract for both trigger and validation functionality. I don't find this very useful

data_model/src/metadata.rs
```diff
@@ -76,24 +76,17 @@ pub mod model {
...
-        /// The actual *entry* size in bytes
-        actual: usize,
-    },
+#[derive(
+    Debug, Display, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Encode, Decode, IntoSchema,
+)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+pub enum MetadataError {
+    /// Metadata entry is too big
+    #[display(fmt = "Metadata entry is roo big")]
```
mversic, line 85:
```suggestion
    #[display(fmt = "Metadata entry is too big")]
```

core/src/smartcontracts/wasm/mod.rs
```diff
@@ -643,149 +652,46 @@ impl<'wrld> Runtime<'wrld> {
...
+
+        // Check mandatory exports
+        let _ = Self::get_memory(&mut (&instance, &mut store))?;
+        let _ =
+            Self::get_typed_func::<WasmUsize, WasmUsize>(&instance, store, export::WASM_ALLOC_FN)?;
+        let _ = Self::get_typed_func::<(WasmUsize, WasmUsize), ()>(
+            &instance,
+            store,
+            export::WASM_DEALLOC_FN,
+        )?;
```
mversic, line 692: isn't this done automatically by `wasmtime` when linking? I'm quite sure it will not allow you to instantiate a module with undefined imports. so it's redundant

mversic: I meant to say this check is useless because this will never fail

core/src/smartcontracts/wasm/mod.rs
```diff
@@ -284,153 +356,79 @@ impl<'wrld> Runtime<'wrld> {
...
     ///
     /// If log level or string decoding fails
-    fn log(
-        mut caller: Caller<State>,
-        log_level: u32,
-        offset: WasmUsize,
-        len: WasmUsize,
-    ) -> Result<(), Trap> {
+    #[codec::wrap]
+    fn log((log_level, msg): (u8, String), state: &State) -> Result<(), Trap> {
```
mversic, line 425: do you think we can just use `panic!` or `assert!` and not return a `Trap`. I would prefer that because it indicates an unrecoverable error. This was a mistake on my part that I've implemented this to return `Result`

mversic (changes requested):

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
mversic: check my other comment. `log` function should panic instead IMO

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
mversic: no, I meant that only `decode` can fail. Why you mention `encode`?

data_model/src/isi.rs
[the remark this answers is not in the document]
mversic: yes, even though it looks like it's possible, actually it's impossible to reach this point. This will fail to validate first

wasm/src/lib.rs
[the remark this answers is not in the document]
Arjentix: Ok, let's make it a `Trap`

data_model/src/lib.rs
Arjentix: Makes sense

core/src/smartcontracts/wasm/mod.rs
```diff
@@ -284,153 +356,79 @@ impl<'wrld> Runtime<'wrld> {
...
-        let state = caller.data();
-        let event = state
-            .triggering_event
-            .as_ref()
-            .ok_or_else(|| Trap::new("There is no triggering event".to_owned()))?;
-
-        let bytes = Self::encode_with_length_prefix(event)?;
-        let event_offset = Self::encode_bytes_into_memory(&bytes, &memory, &alloc_fn, &mut caller)?;
-        Ok(event_offset)
+    #[codec::wrap]
```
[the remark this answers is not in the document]
Arjentix: Although the overall code size has increased. But yes, I think it's better now.

data_model/src/name.rs
```diff
@@ -47,26 +47,26 @@ impl Name {
...
+        Ok(())
     }
 
     /// Check if `candidate` string would be valid [`Name`].
     ///
     /// # Errors
     /// Fails if not valid [`Name`].
     fn validate_str(candidate: &str) -> Result<(), ParseError> {
-        const FORBIDDEN_CHARS: [char; 4] = ['@', '#', '$', '%'];
+        const FORBIDDEN_CHARS: [char; 3] = ['@', '#', '$'];
```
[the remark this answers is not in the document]
Arjentix: Yeah, `%` should be removed long ago

wasm/src/lib.rs
Arjentix: `Result<(), ValidationFail>`

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
mversic: ok, but both of these examples should panic, not return error

data_model/src/isi.rs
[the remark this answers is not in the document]
mversic: can we make validator return this error as well?

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
Arjentix: I thought that if we panic -- then it will be a unrecoverable error on Iroha side. And if we return `Trap` -- it will be an unrecoverable error on wasm side. No?

data_model/src/isi.rs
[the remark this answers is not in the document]
Arjentix: I mean that user-provided validator can skip this check, so we should do it on our side too

wasm/src/lib.rs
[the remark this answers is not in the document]
Arjentix: Or panic

core/src/smartcontracts/wasm/codec/src/lib.rs
Arjentix: The author of the article I mentioned doesn't do that but ok

Arjentix: I get this error:

```
error[E0401]: can't use generic parameters from outer function
   --> core/src/smartcontracts/wasm/codec/src/lib.rs:123:38
    |
120 | pub fn encode_with_length_prefix<T: Encode>(obj: &T) -> Vec<u8> {
    |                                  - type parameter from outer function
...
123 |     const RESULT: () = SizeChecker::<T>::RESULT;
    |                                      ^ use of generic parameter from outer function
```

Very strange. Let's keep it like it's already done

Arjentix: The best description I've found:

> The error message is not the best, but items like fns and consts inside a function definition behave as if they were actually declared outside the function. The only difference is that their visibility is restricted to inside the containing function.

Looks like it possible only with a nightly feature.

[source](https://users.rust-lang.org/t/error-e0401-cant-use-generic-parameters-from-outer-function/84512)

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
Arjentix: They return `Trap`. Isn't it the same? Can you please explain the difference between panicking and returning Trap?

cli/src/torii/mod.rs
```diff
@@ -68,15 +68,19 @@ pub enum Error {
...
-        Find(_) => StatusCode::NOT_FOUND,
+    match validation_error {
+        NotPermitted(_) => StatusCode::FORBIDDEN,
+        QueryFailed(query_error) => match query_error {
+            Evaluate(_) | Conversion(_) => StatusCode::BAD_REQUEST,
+            Signature(_) | Unauthorized => StatusCode::UNAUTHORIZED,
+            Find(_) => StatusCode::NOT_FOUND,
+        },
+        TooComplex => StatusCode::UNPROCESSABLE_ENTITY,
+        InstructionFailed(_) | InternalError(_) => StatusCode::INTERNAL_SERVER_ERROR,
```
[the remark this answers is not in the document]
Arjentix: Hm... In fact I think it's unreachable... I'll make it so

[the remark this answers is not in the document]
Arjentix: I fixed that place, check it out

client/tests/integration/query_errors.rs
```diff
@@ -22,7 +22,9 @@ fn non_existent_account_is_specific_error() {
         .expect_err("Should error");
 
     match err {
-        ClientQueryError::QueryError(QueryExecutionFailure::Find(err)) => match *err {
+        ClientQueryError::Validation(ValidationFail::QueryFailed(QueryExecutionFailure::Find(
+            err,
+        ))) => match *err {
```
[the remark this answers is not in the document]
Arjentix: Can't say anything here, this is not my code, I just fixed error type here. What do you suggest?

Arjentix:
Addressed part of the comments. Will address others later

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
mversic: yes, but not quite. If you `panic!` inside the rust program it's as if you have returned `Trap` explicitly. Therefore, any function that returns `Result<_, Trap>` can be transformed into an equivalent function that panics. 

Our functions should never return `Trap` explicitly, this should be handled by your `codec::wrap`

core/src/smartcontracts/wasm/mod.rs
```diff
@@ -285,154 +353,83 @@ impl<'wrld> Runtime<'wrld> {
...
-
-        let memory = Self::get_memory(&mut caller)?;
-        let msg: String = Self::decode_from_memory(&memory, &caller, offset, len)?;
-
-        let _span = caller.data().log_span.enter();
-        match LogLevel::from_repr(log_level).ok_or_else(error_msg)? {
+        let _span = state.log_span.enter();
+        match LogLevel::from_repr(log_level)
+            .ok_or_else(|| Trap::new(format!("{log_level}: not a valid log level")))?
+        {
```
mversic, line 432:
```suggestion
        match LogLevel::from_repr(log_level).expect(format!("{log_level}: not a valid log level")) {
```

core/src/smartcontracts/wasm/mod.rs
mversic:
```suggestion
    fn log((log_level, msg): (u8, String), state: &State) {
```

function signature should look like this

client/tests/integration/query_errors.rs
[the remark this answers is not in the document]
QuentinI: Nevermind, I misread the diff, sorry

core/src/smartcontracts/wasm/codec/derive/src/lib.rs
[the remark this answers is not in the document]
Arjentix: We decided to implement it in another task: #3555

core/src/smartcontracts/wasm/mod.rs
Arjentix: Created #3555

core/src/smartcontracts/wasm/mod.rs
Arjentix: Created #3555

wasm/src/lib.rs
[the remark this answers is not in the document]
Arjentix: Fixed

data_model/src/lib.rs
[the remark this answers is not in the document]
Arjentix: Fixed

data_model/src/lib.rs
Arjentix: Fixed

core/src/smartcontracts/wasm/mod.rs
Arjentix: > I'm quite sure it will not allow you to instantiate a module with undefined imports. so it's redundant

How does wasmtime know which imports we require? Why isn't it possible to load a smartcontract without allocation function?

QuentinI (dismissed):

data_model/src/isi.rs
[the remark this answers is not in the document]
mversic: ok, but shouldn't our default validator return this error?

cli/src/torii/routing.rs
```diff
@@ -66,19 +67,23 @@ impl VerifiedQuery {
...
                 "Signature public key doesn't correspond to the account.",
-            )));
+            ))
+            .into());
         }
-        wsv.validator_view()
-            .clone() // Cloning validator is a cheap operation
-            .validate(wsv, &self.payload.account_id, self.payload.query.clone())
-            .map_err(|err| QueryExecutionFailure::Permission(err.to_string()))?;
+        wsv.validator_view().clone().validate(
```
mversic, line 82:
```suggestion
        wsv.validate(
```

isn't this a lot more sensible? I'm not asking you to make this change in this PR. I'm just wondering what you think

data_model/src/isi.rs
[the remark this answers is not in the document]
mversic: ok, let's leave this aside. Do you think `DenialReason` should be an enum with one of the variants being `UnsupportedInstruction`? Should we be offering a set of predefined error types that a validator can return?

This is not in the scope of this PR

core/src/smartcontracts/wasm/mod.rs
mversic: got it

wasm/validator/src/lib.rs
mversic: > This macro is used in 13 places

This metric should be taken with a grain of salt. If I wrote a macro like this and used in in 100 places:
```rs
macro_rules! ret {
    ( $val:ident ) => {
        return $val
    }
}

fn return_val() -> usize {
    let a = 12;
    ret!(a);
}
```

would you say this is a useful macro? IMO this is what `propagate` is doing. It replaces a perfectly readable Rust syntax. It barely reduces the amount of repetition in the code. Not every code repetition is bad, especially if the amount of code is small. A macro should have some meat in it

Maybe someone else has an opinion?

data_model/src/isi.rs
```diff
@@ -995,32 +1034,63 @@ pub mod error {
...
-            ParameterValueType(Mismatch<Value>),
+            #[display(fmt = "Value passed to the parameter doesn't have the right type, {_0}")]
+            ParameterValueType(#[cfg_attr(feature = "std", source)] Box<Mismatch<Value>>),
             /// Asset Id mismatch
-            #[display(fmt = "AssetDefinition Ids don't match. {_0}")]
-            AssetDefinitionId(Mismatch<<AssetDefinition as Identifiable>::Id>),
+            #[display(fmt = "AssetDefinition Ids don't match, {_0}")]
+            AssetDefinitionId(
+                #[cfg_attr(feature = "std", source)]
+                Box<Mismatch<<AssetDefinition as Identifiable>::Id>>,
```
mversic, line 1066:
```suggestion
                Mismatch<<AssetDefinition as Identifiable>::Id>,
```

not sure, but I don't think it's needed? unless the linter is complaining

Arjentix: Yeah, it's complaining about variant sized because I introduced `Box` for another variant as we discussed somwhere above

wasm/validator/src/lib.rs
Arjentix: Ok, most of the time it's

```rust
if propagate!(...) {
    pass!()
}
```

Will remove this macro

appetrosyan: Some Lisp best practices about macros: 
1. They are meant to reduce cognitive load. Sometimes longer, but simpler code is easier to understand. Sam is going overboard, but I tend to lean in that direction. This is in the middle. It's like the old `try!` macro. But it's not very ergonomic, if we're honest. 
2. Macros need to give you the context of what they do. `propagate` is about the most neutral name a thing can have. It tells me nothing about the context. 
3. It needs to focus the attention of the reader on the key aspects. To me it looks like the error and the validator are the most important. Contrast this with `?`, where it focuses you on the fallible operation. 
4. Macros are more brittle than functions. If you can avoid writing one, use a generic instead (in Lisp, it'd be a regular function). 

For the record, when the `FromResidual` trait is stable, we'd be able to override the behaviour of `?`, so I think we can live with extra code until then.

Arjentix: Removed, check out new code

data_model/src/isi.rs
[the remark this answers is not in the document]
Arjentix: In my PR it returns this `NameLength` error already

cli/src/torii/routing.rs
Arjentix: I'm not sure. I don't like WSV being god-object. And by implementing `validate()` on it we'll make this types relations even harder.

And current `validator_view()` just returns a reference to the validor. So that `wsv` doesn't take the reponsibillity for functions provided by validator

mversic: > And current `validator_view()` just returns a reference to the validor. So that wsv doesn't take the reponsibillity for functions provided by validator

`validator_view()` is always followed by `.clone()`. At least we can move the `clone()` into the `validator_view` function itself

@Erigara your thoughts?

mversic (approved):

cli/src/torii/routing.rs
Erigara: > > And current `validator_view()` just returns a reference to the validor. So that wsv doesn't take the reponsibillity for functions provided by validator
> 
> `validator_view()` is always followed by `.clone()`. At least we can move the `clone()` into the `validator_view` function itself
> 
> @Erigara your thoughts?

 Don't have strong opinion about this. But moving clone into the `validator_view` will be less repetative.
However i think we need to rename method in such case because there is no immutable view just clone.

Erigara (approved):

cli/src/torii/routing.rs
Arjentix: Created #3560

PR #3485, opened 2023-05-17
[refactor]: Standardize transaction API

mversic:
### Linked issue

Closes #{issue_number} 

### Checklist


data_model/src/trigger.rs
```diff
@@ -226,7 +212,7 @@ pub mod action {
             /// Technical account linked to this trigger. The technical
             /// account must already exist in order for `Register<Trigger>` to
             /// work.
-            pub technical_account: crate::account::AccountId,
+            pub technical_account: AccountId,
```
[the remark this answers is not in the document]
Arjentix: Yes

docs/source/references/schema.json
```diff
@@ -3333,19 +3333,9 @@
...
-        "type": "SignaturesOf<TransactionPayload>"
-      },
-      {
-        "name": "rejection_reason",
-        "type": "TransactionRejectionReason"
-      }
+    "Tuple": [
+      "VersionedSignedTransaction",
+      "TransactionRejectionReason"
     ]
```
0x009922, line 3339: As for JS SDK, working with structs is easier than with tuples in JS/TS. So I would prefer to have this instead:

```rust
struct RejectedTransaction {
  transaction: VersionedSignedTransaction,
  reason: TransactionRejectionReason
}
```

0x009922 (commented):
PR has too many places I am not competent at.

cli/src/lib.rs
```diff
@@ -245,22 +244,8 @@ impl Iroha {
...
-        // Validate every transaction in genesis block
-        if let Some(ref genesis) = genesis {
-            let span = span!(tracing::Level::TRACE, "genesis").entered();
-            let wsv_clone = wsv.clone();
-
-            transaction_validator
-                .validate_every(genesis.iter().cloned(), &wsv_clone)
-                .wrap_err("Transaction validation failed in genesis block")?;
-            span.exit();
-        }
```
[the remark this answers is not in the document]
Erigara: Idea when i made this change was to catch errors in genesis as early as possible.
Do you think we should validate genesis in sumeragi?

config/src/wsv.rs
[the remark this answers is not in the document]
Erigara: I would be careful with removing `wsv` configuration. 
Because some of configuration `wsv` used not only by `sumeragi` but also other actors like `gossiper` or `torii`.
So we will need a way to propagate this changes.

My another concern is that for soft-fork resolution we might need previous version of some parameters. 
Currently this is achieved by packing relevant configuration in `wsv`.

cli/src/lib.rs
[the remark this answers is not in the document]
mversic: Wherever we validate we should transition from SignedBlock to ValidatedBlock. I have done this in the other refactor PR and will revive it after this one is merged

Please allow me the benefit of the doubt atm

config/src/wsv.rs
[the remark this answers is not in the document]
mversic: Yeah sure. It just seems weird to have `wsv` in the configuration.

cli/src/lib.rs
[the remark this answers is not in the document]
Erigara: Sure, feel free to resolve

config/kita
Erigara, line 1: I think this was added accidentally

core/src/smartcontracts/isi/world.rs
```diff
@@ -423,11 +423,11 @@ pub mod query {
     impl ValidQuery for DoesAccountHavePermissionToken {
         #[metrics("does_account_have_permission")]
         fn execute(&self, wsv: &WorldStateView) -> Result<Self::Output, Error> {
-            let account_id = wsv
+            let authority = wsv
                 .evaluate(&self.account_id)
                 .map_err(|e| Error::Evaluate(e.to_string()))?;
 
-            wsv.map_account(&account_id, |account| {
+            wsv.map_account(&authority, |account| {
```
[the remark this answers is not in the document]
Erigara: What is SSO?

core/src/tx.rs
```diff
@@ -14,19 +14,129 @@
...
+    }
+
+    /// Payload of the transaction
+    pub fn payload(&self) -> &TransactionPayload {
+        self.0.payload()
+    }
+
+    pub(crate) fn signatures(&self) -> &SignaturesOf<TransactionPayload> {
+        self.0.signatures()
+    }
```
Erigara, line 121: Maybe implement `Deref`?

core/src/smartcontracts/isi/world.rs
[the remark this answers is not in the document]
mversic: small string optimization

schema/gen/src/lib.rs
```diff
@@ -42,20 +42,21 @@ pub fn build_schemas() -> MetaMap {
...
         VersionedPaginatedQueryResult,
         VersionedSignedQuery,
-        VersionedPendingTransactions,
+
+        // Never referenced, but present in type signature. Like `PhantomData<X>`
         UpgradableBox,
         RegistrableBox,
         MerkleTree<VersionedSignedTransaction>,
+
+        // TODO: Should genesis belong to schema? #3284
```
Erigara, line 58: I think this TODO could be removed since issue is closed.

data_model/src/transaction.rs
```diff
@@ -306,125 +142,56 @@ pub mod model {
...
-        Deserialize,
-        Serialize,
-        IntoSchema,
+        Debug, Display, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Serialize, IntoSchema,
     )]
-    #[display(fmt = "{self:?}")] // TODO ?
+    #[cfg_attr(not(feature = "std"), display(fmt = "Signed transaction"))]
+    #[cfg_attr(feature = "std", display(fmt = "{}", "self.hash()"))]
     #[ffi_type]
+    // TODO: All fields in this struct should be private
```
Erigara, line 158: Because we don't want them to be modified independently?

data_model/src/transaction.rs
```diff
@@ -667,232 +403,69 @@ impl Ord for TransactionQueryResult {
...
-}
 
-impl From<VersionedRejectedTransaction> for VersionedSignedTransaction {
-    fn from(transaction: VersionedRejectedTransaction) -> Self {
-        match transaction {
-            VersionedRejectedTransaction::V1(transaction) => SignedTransaction {
-                payload: transaction.payload,
-                signatures: transaction.signatures,
+    impl SignedTransactionCandidate {
+        fn validate(mut self) -> Result<SignedTransaction, &'static str> {
```
Erigara, line 423: Why `&'static str` as error?

core/src/tx.rs
mversic: I could. But semantically it would be more appropriate to implement `AsRef`

data_model/src/transaction.rs
mversic: Because we want the codebase to be version independent. Nobody should be using V1 structs outside data_model. All operations should be done on VersionedXXX structs

data_model/src/transaction.rs
mversic: Because it's only for deserialize/decode. And they work with str

client/tests/integration/permissions.rs
```diff
@@ -253,19 +250,15 @@ fn permissions_differ_not_only_by_names() {
...
+    );
 
-    let grant_shoes_access_tx = TransactionBuilder::new(
-        mouse_id,
-        vec![allow_alice_to_set_key_value_in_shoes],
-        100_000,
-    )
-    .sign(mouse_keypair)
-    .expect("Failed to sign mouse transaction");
+    let grant_shoes_access_tx = TransactionBuilder::new(mouse_id, Duration::from_millis(100_000))
```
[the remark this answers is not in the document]
mversic: yes, that is a good point

core/src/tx.rs
Erigara: I mean that we `Deref` we could just call `tx.payload()` without implementing this methods for `VersionedSignedTransaction`.

docs/source/references/schema.json
appetrosyan: It's also easier in Rust. You can destructure directly and with no ambiguity.

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
mversic: actually, I found that the field `SignedTransaction::time_to_live_ms` was never used so I removed it entirely

docs/source/references/schema.json
mversic: > As for JS SDK, working with structs is easier than with tuples in JS/TS.

schema can be defined independently of Rust types



> It's also easier in Rust. You can destructure directly and with no ambiguity.

I applied your suggestions. Though for posterity check the comment [here](https://github.com/hyperledger/iroha/pull/3485#discussion_r1198785396).

core/src/tx.rs
Erigara: After reading attached post i agree with your point.

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
Erigara: On latest `dev` i can see the following:

```rust
    fn is_expired(&self, transaction_time_to_live: Duration) -> bool {
        let tx_timestamp = Duration::from_millis(self.payload().creation_time);
        crate::current_time().saturating_sub(tx_timestamp)
            > core::cmp::min(
                transaction_time_to_live,
                Duration::from_millis(self.payload().time_to_live_ms),
            )
    }
```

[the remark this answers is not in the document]
mversic: ok, you are correct. Do we find this useful though?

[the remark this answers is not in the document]
mversic: makes sense, I'll revert this

client/tests/integration/asset.rs
```diff
@@ -195,6 +196,7 @@ fn client_add_asset_with_decimal_should_increase_asset_amount() -> Result<()> {
 #[test]
 fn client_add_asset_with_name_length_more_than_limit_should_not_commit_transaction() -> Result<()> {
     let (_rt, _peer, test_client) = <PeerBuilder>::new().with_port(10_520).start_with_runtime();
+    wait_for_genesis_committed(&vec![test_client.clone()], 0);
```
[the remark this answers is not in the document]
mversic: 1.5x - 2x

config/src/client.rs
```diff
@@ -79,10 +81,12 @@ pub struct Configuration {
     /// Status URL.
     pub torii_telemetry_url: Url,
     /// Proposed transaction TTL in milliseconds.
-    pub transaction_time_to_live_ms: u64,
+    pub transaction_time_to_live_ms: Option<NonZeroU64>,
     /// Transaction status wait timeout in milliseconds.
     pub transaction_status_timeout_ms: u64,
     /// The limits to which transactions must adhere to
+    // NOTE: If you want this functionality, implement it in the app manually
+    #[deprecated(note = "This parameter is not used and takes no effect")]
```
0x009922, line 89:
```suggestion
    #[deprecated(note = "This parameter is not used and takes no effect. If you want this functionality, implement it in the app manually.")]
```

Saying to user what should they do with a deprecation helps.

core/src/queue.rs
```diff
@@ -14,27 +14,79 @@ use dashmap::{mapref::entry::Entry, DashMap};
...
 use iroha_logger::{debug, info, trace, warn};
 use iroha_primitives::{must_use::MustUse, riffle_iter::RiffleIter};
 use rand::seq::IteratorRandom;
 use thiserror::Error;
 
-use crate::{prelude::*, tx::CheckSignatureCondition as _};
+use crate::prelude::*;
+
+impl AcceptedTransaction {
+    //// TODO: We should have another type of transaction like `CheckedTransaction` in the type system?
```
0x009922, line 31:
```suggestion
    // TODO: We should have another type of transaction like `CheckedTransaction` in the type system?
```

0x009922 (commented):
LGTM, although I don't understand most of the changes.

data_model/src/block.rs
```diff
@@ -186,11 +179,11 @@ mod committed {
             /// Array of rejected transactions.
             // TODO: Derive with getset once FFI impl is fixed
             #[getset(skip)]
-            pub rejected_transactions: Vec<VersionedRejectedTransaction>,
+            pub rejected_transactions: Vec<RejectedTransaction>,
```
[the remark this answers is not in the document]
Erigara: I'm fine with `RejectedTransaction`. 
And if i remember correctly SDK prefer not to work with tuples.

schema/gen/src/lib.rs
Erigara: TODO is still there.

Erigara (approved):
LGTM, couple of small comments

cli/src/main.rs
```diff
@@ -119,24 +119,27 @@ async fn main() -> Result<(), color_eyre::Report> {
...
-            )?,
-            Some(&config.genesis),
-            &config.wsv.transaction_limits,
-        )
-        .wrap_err("Failed to initialize genesis.")?
-    } else {
-        None
-    };
+    let genesis =
+        if let Some(genesis_path) = args.submit_genesis.then_some(()).and(args.genesis_path) {
```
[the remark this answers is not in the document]
appetrosyan: Marin prefers the imperative style.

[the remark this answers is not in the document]
mversic: > Marin prefers the imperative style. 

Actually, i don't care. I prefer that we have a uniform style. But sometimes I dont find jt worth my time to enforce it religiously. Ww need to have a faster turnover of PRs. I think every one of us should ask themselves is this comment really worth delaying the merge? I will always apply every comment if I'm neutral or in favor of it. In this case there was a `?` inside the closure which I didn't want to figure out how. My mistake was that I closed the discussion instead of asking for help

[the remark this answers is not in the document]
mversic: I change it to functional style. It was very easy, I was lazy to do it. And it's indeed looking a lot more readable

Erigara (dismissed):

wasm/validator/src/permission.rs
```diff
@@ -165,7 +165,7 @@ pub mod trigger {
     ) -> Result<bool> {
         FindTriggerById::new(trigger_id)
             .execute()
-            .map(|trigger| trigger.action().technical_account() == authority)
+            .map(|trigger| trigger.action().authority() == authority)
```
appetrosyan, line 168: We should back-propagate this change, the somewhat obscure name `technical_account` was chosen precisely because it was in the triggers RFC. Every deviation from that RFC was documented so should this one.

schema/src/lib.rs
```diff
@@ -262,6 +265,52 @@ macro_rules! impl_schema_int {
 
 impl_schema_int!(u128, u64, u32, u16, u8, i128, i64, i32, i16, i8);
 
+impl TypeId for NonZeroU32 {
```
appetrosyan, line 268: Let's hope the SDK devs know how to create one.

client/tests/integration/queries/asset.rs
```diff
@@ -99,20 +99,20 @@ fn find_asset_total_quantity() -> Result<()> {
...
-            .map(|asset| RegisterBox::new(asset).into())
+            .map(RegisterBox::new)
             .collect::<Vec<_>>();
         test_client.submit_all_blocking(register_asset)?;
 
         let mint_asset = asset_ids
             .iter()
             .cloned()
-            .map(|asset_id| MintBox::new(to_mint.clone(), asset_id).into());
+            .map(|asset_id| MintBox::new(to_mint.clone(), asset_id));
```
appetrosyan, line 109: Rust desperately needs partial application/currying.

client/tests/integration/transfer_asset.rs
```diff
@@ -69,13 +69,14 @@ fn simulate_transfer<
         IdBox::AssetId(AssetId::new(asset_definition_id.clone(), alice_id.clone())),
     );
 
+    let instructions: [InstructionBox; 3] = [
+        // create_alice.into(), We don't need to register Alice, because she is created in genesis
```
appetrosyan, line 73: Good comment.

client/tests/integration/tx_history.rs
```diff
@@ -57,12 +57,12 @@ fn client_has_rejected_and_acepted_txs_should_return_tx_history() -> Result<()>
...
 
-    let mut prev_creation_time = 0;
+    let mut prev_create_time = core::time::Duration::from_millis(0);
     for tx in &transactions {
-        assert_eq!(&tx.payload().account_id, &account_id);
+        assert_eq!(tx.payload().authority(), &account_id);
         //check sorted
-        assert!(tx.payload().creation_time >= prev_creation_time);
-        prev_creation_time = tx.payload().creation_time;
+        assert!(tx.payload().create_time() >= prev_create_time);
```
appetrosyan, line 64: `creation_time` is more grammatically correct. But `create_time` is acceptable.

config/src/client.rs
```diff
@@ -97,31 +104,29 @@ impl Default for ConfigurationProxy {
             basic_auth: Some(None),
             torii_api_url: None,
             torii_telemetry_url: None,
-            transaction_time_to_live_ms: Some(DEFAULT_TRANSACTION_TIME_TO_LIVE_MS),
+            transaction_time_to_live_ms: Some(None),
```
appetrosyan, line 107: I would still argue that we need a default for this.

config/src/queue.rs
```diff
@@ -5,8 +5,8 @@ use serde::{Deserialize, Serialize};
 
 const DEFAULT_MAX_TRANSACTIONS_IN_QUEUE: u32 = 2_u32.pow(16);
 const DEFAULT_MAX_TRANSACTIONS_IN_SIGNATURE_BUFFER: u32 = 2_u32.pow(16);
-// 24 hours
-const DEFAULT_TRANSACTION_TIME_TO_LIVE_MS: u64 = 24 * 60 * 60 * 1000;
+/// Default length of time transaction will be kept if not committed
+pub const DEFAULT_TRANSACTION_TIME_TO_LIVE_MS: u64 = 24 * 60 * 60 * 1000; // 24 hours
```
appetrosyan, line 9: Might be a good idea to consider a stronger type that can be deserialized from JSON. E.g. 
```json
{
  "TRANSACTION_TIME_TO_LIVE_MS": {
    "hours": 24
  }
}
```

configs/client/config.json
```diff
@@ -11,7 +11,7 @@
   },
   "TORII_API_URL": "http://127.0.0.1:8080/",
   "TORII_TELEMETRY_URL": "http://127.0.0.1:8180/",
-  "TRANSACTION_TIME_TO_LIVE_MS": 100000,
+  "TRANSACTION_TIME_TO_LIVE_MS": null,
```
appetrosyan, line 14: Probably should be returned.

cli/src/torii/routing.rs
```diff
@@ -29,8 +29,7 @@ use iroha_data_model::{
     predicate::PredicateBox,
     prelude::*,
     query,
-    query::error::QueryExecutionFailure,
-    transaction::InBlock,
+    query::error::QueryExecutionFail,
```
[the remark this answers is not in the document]
mversic: it's out of consistency. All other errors are named with suffix `Fail`. It was either rename all other types to suffix `Failure` or change this to suffix `Fail`. I chose this to limit the change in the schema as much as I can. When we'll be having a discussion on error types (which I think we should, just not quite yet) we can choose more appropriate naming

client/tests/integration/asset.rs
```diff
@@ -97,9 +97,9 @@ fn client_add_asset_quantity_to_existing_asset_should_increase_asset_amount() ->
             account_id.clone(),
         )),
     );
-    let instructions: Vec<InstructionBox> = vec![create_asset.into(), mint.into()];
+    let instructions: [InstructionBox; 2] = [create_asset.into(), mint.into()];
```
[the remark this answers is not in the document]
mversic: I like this suggestion. I'm going to open a ticket for it

cli/src/torii/routing.rs
```diff
@@ -112,17 +111,15 @@ pub(crate) async fn handle_instructions(
...
-        .map_err(Error::AcceptTransaction)?
-        .into();
+    let transaction = AcceptedTransaction::accept(transaction, &transaction_limits)
+        .map_err(Error::AcceptTransaction)?;
     queue
         .push(transaction, &wsv)
         .map_err(|queue::Failure { tx, err }| {
             iroha_logger::warn!(
-                tx_hash=%tx.hash(), ?err,
+                tx_hash=%tx.payload().hash(), ?err,
```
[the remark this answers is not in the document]
mversic: > This seems contentious

agree, but actually, this hasn't changed hash in this place. We've always been using payload hash as transaction hash. I was inspired with @SamHSmith work in #3306 . This is the exact same change but for transactions, not blocks. Once the transactions has left the queue it's signature set is stable and we can add it to the hash

wasm/validator/src/permission.rs
mversic: so I should go to RFC and make a note of the change?

schema/src/lib.rs
mversic: the beauty of it is that they don't have to. They can continue using `Option<u32>` because both types have the same serialized representation. Niche optimization is not visible in the serialized format

client/tests/integration/transfer_asset.rs
mversic: not mine though

config/src/client.rs
mversic: check the comment [here](https://github.com/hyperledger/iroha/pull/3485/files#diff-38804a7b8ccf61b62b61b4f22618d2c5a9ec747a42041d54ac4cfd44239fd79eR92)

configs/client/config.json
mversic: I'm closing this in favor of discussion [here](https://github.com/hyperledger/iroha/pull/3485#discussion_r1216936773)

config/src/queue.rs
mversic: maybe, @0x009922 can we include this suggestion in configuration discussions?

config/src/client.rs
mversic: can you give an argument for it?

client/tests/integration/tx_history.rs
mversic: silly change, I'll revert this

mversic: we might consider `creation_timestamp` vs `creation_time`

data_model/src/isi.rs
```diff
@@ -1190,12 +1206,8 @@ pub mod error {
             #[display(fmt = "Invalid WASM binary: {_0}")]
             Wasm(String),
             /// Invalid name length (i.e. too long [`AccountId`])
-            #[display(
-                fmt = "Name must be between {} and {} (inclusive) characters in length",
-                "_0.start()",
-                "_0.end()"
-            )]
-            NameLength(RangeInclusive<u32>),
+            #[display(fmt = "Name length violation")]
```
Erigara, line 1209: Why have you decided to simplify this error message? Wouldn't it will be useful for user to know exact limits on error.

Erigara (approved):

data_model/src/isi.rs
mversic: Because `Range` doesn't implement `PartialEq`

config/src/queue.rs
0x009922: For user convenience, we might use an approach from an incredibly popular [`ms` node package](https://github.com/vercel/ms):

```js
ms('2 days')  // 172800000
ms('1d')      // 86400000
ms('10h')     // 36000000
ms('2.5 hrs') // 9000000
ms('2h')      // 7200000
ms('1m')      // 60000
ms('5s')      // 5000
ms('1y')      // 31557600000
ms('100')     // 100
ms('-3 days') // -259200000
ms('-1h')     // -3600000
ms('-200')    // -200
```

schema/src/lib.rs
0x009922: While it is effectively still `u32`, it is not redundant to let SDK devs know that some particular `u32` should not be zero. Such information is useful and might be handled depending on platform.

wasm/validator/src/permission.rs
appetrosyan: No. You're good. I've opened a ticket and transferred to docs.

schema/src/lib.rs
appetrosyan: Java and Kotlin have some support for contract-based programming annotations. @NotNull and @NonZero.

config/src/client.rs
appetrosyan: We have this option in general because blockchain operators want to fight DoS attacks and manage bandwidth. Given how much is stored in RAM, this value must be set as low as possible, and easy to change. 

IMO a low reasonable value does the job just as well as asking the user to set it explicitly. Remember this is the client configuration, so we're probably dealing with the front-end where the Swift or Java program can figure out the value on its own, which is likely to just pick a default. This way, this is a default that's controlled upstream.

data_model/src/isi.rs
appetrosyan: We could implement `PartialOrd` for the entire struct manually, and still have the extra information. I still think that this should be a separate type.

cli/src/torii/routing.rs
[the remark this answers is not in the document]
appetrosyan: I'm a bit out of the loop on that one.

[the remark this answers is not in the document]
mversic: So there is both `HashOf<TransactionPayload>` and `HashOf<SignedTransaction>`. Previously they were synonymous, with this change they will be distinct hashes. The 1st one hashing only payload, and the 2nd one hashing both payload and signature set. If we didn't have MST, we could be using `HashOf<SignedTransaction>` in all places

config/src/queue.rs
mversic: I'll close this then

data_model/src/isi.rs
mversic: opened an [issue](https://github.com/hyperledger/iroha/issues/3590)

client/tests/integration/asset.rs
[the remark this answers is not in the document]
mversic: [issue](https://github.com/hyperledger/iroha/issues/3591)

core/src/smartcontracts/isi/world.rs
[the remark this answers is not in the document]
appetrosyan: We could have separate pipelines for these things

Erigara (approved):

PR #3486, opened 2023-05-18
[refactor] #3363: Genesis API

### LINKED ISSUE #3363: Genesis API
I'm not at all a fan of this API. I'd rather we had a similarly named method in two different traits. The user is left guessing as to what `<true>` means in this context. It was slightly better when we had an enum, but now I'm convinced we really need two trait extension methods and only one in scope.

_Originally posted by @appetrosyan in https://github.com/hyperledger/iroha/pull/3356#discussion_r1161496792_

ilchu:
## Description

* Introduce sealed traits for block revalidation and accepting transactions
* Make wrappers around those two functions for public use

### Linked issue

Closes #3363 

### Benefits
More idiomatic and self-descriptive usage.

### Checklist

### DIFF
--- a/core/src/sumeragi/message.rs
+++ b/core/src/sumeragi/message.rs
@@ -13,11 +13,7 @@ use iroha_version::prelude::*;
 use parity_scale_codec::{Decode, Encode};
 
 use super::view_change;
-use crate::{
-    block::{PendingBlock, Revalidate},
-    tx::TransactionValidator,
-    WorldStateView,
-};
+use crate::block::PendingBlock;
 
 declare_versioned_with_scale!(VersionedPacket 1..2, Debug, Clone, iroha_macro::FromVariant);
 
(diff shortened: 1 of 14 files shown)


data_model/src/transaction.rs
```diff
@@ -141,6 +142,60 @@ pub trait Sign {
     ) -> Result<SignedTransaction, iroha_crypto::error::Error>;
 }
 
+/// Sealed trait for accepting transactions. Downstream users
+/// should only use its extension traits [`InGenesis`] and
+/// [`InBlock`] according to the use case.
+#[sealed]
+#[cfg(feature = "std")]
+pub trait Accept: Sized {
```
mversic, line 150:
```suggestion
#[cfg(feature = "transparent-api")]
pub trait Accept: Sized {
```

mversic (approved):

appetrosyan (approved):
Reads much better. Glad you chose to go with a sealed trait.

core/src/sumeragi/message.rs
```diff
@@ -169,7 +146,7 @@ impl From<VersionedCommittedBlock> for BlockCommitted {
 #[non_exhaustive]
 pub struct BlockSyncUpdate {
     /// The corresponding block.
-    block: VersionedCommittedBlock,
+    pub block: VersionedCommittedBlock,
```
Erigara, line 149: This field was private on purpose, to always be sure that block will be revalidated before usage. 
However i think we can return to this in separate PR.

Erigara (dismissed):
I think merge commit accidentally happened.

core/src/sumeragi/message.rs
ilchu: Created #3492.

mversic (approved):

mversic (dismissed):

mversic (dismissed):

Erigara (approved):

mversic (approved):

PR #3491, opened 2023-05-19
[fix] #3444: return discriminant into schema

### LINKED ISSUE #3444: [BUG] Versioned structures in the schema miss discriminant info

mversic:
### Linked issue

Closes #3444 

### Checklist


schema/tests/enum_with_various_discriminants.rs
```diff
@@ -28,18 +30,22 @@ fn discriminant() {
                 variants: vec![
                     EnumVariant {
                         tag: "A".to_owned(),
+                        discriminant: 1,
                         ty: None,
                     },
                     EnumVariant {
                         tag: "B".to_owned(),
+                        discriminant: 1,
```
mversic, line 38: it's funny that with SCALE we can have two structs with the same discriminant

mversic: one more reason why I dislike allowing discriminants

docs/source/references/schema.json
0x009922, line 1: Looks verbose now, but at least it is correct.

As for me, I would be ok if discriminants are specified only for versioned structs, but complete explicitness is fine too.

0x009922 (approved):

schema/tests/enum_with_various_discriminants.rs
0x009922: Why wouldn't you add a guard into the proc macro to prevent such collision?

appetrosyan: Because there's still ways of implementing `Encode` and `Decode` with a collision, e.g. without using our macro, I think we also need a marker trait.  This will ensure that the user is shown an error message if they do something potentially buggy.

appetrosyan (approved):

PR #3475, opened 2023-05-12
[feature] #2373: `kagami swarm`

0x009922:
## Description

This PR introduces `kagami swarm` command and some related refactoring and fixes.

This command allows to generate a directory with everything needed to run `docker compose up` in it. It allows to specify variable number of peers and source of the images.

**Examples:**

- Generate docker-compose configuration in `./compose-iroha` directory with 2 peers using `GitHub` source:
  ```bash
  kagami swarm --build-from-github --peers 2 --outdir ./compose-iroha
  ```

  **Generated files:**

  ```
  [1.1M]  ./compose-iroha
  ├── [761K]  config
  │   ├── [2.1K]  config.json
  │   ├── [ 13K]  genesis.json
  │   └── [746K]  validator.wasm
  ├── [2.1K]  docker-compose.yml
  └── [323K]  iroha-cloned
      └── ...
  ```

  <details> <summary><code>docker-compose.yml</code></summary>
  
  ```yaml
  version: '3.8'
  services:
    iroha0:
      build: ./iroha-cloned
      environment:
        IROHA_PUBLIC_KEY: ed0120F958FCF0FCD15BD3A3C50F5A3DA9DF97D1B49D9BD6E482A109AE8082BCC54198
        IROHA_PRIVATE_KEY: '{"digest_function":"ed25519","payload":"a9ba80952c6f1a87e1039931c5355c026f0e6d2eff3ab5ba38c4411933c5226df958fcf0fcd15bd3a3c50f5a3da9df97d1b49d9bd6e482a109ae8082bcc54198"}'
        TORII_P2P_ADDR: iroha0:1337
        TORII_API_URL: iroha0:8080
        TORII_TELEMETRY_URL: iroha0:8180
        IROHA_GENESIS_ACCOUNT_PUBLIC_KEY: ed01200D80056A4CB7E32F6E0593C8708C03340D3D1E7F1050E635DBE0A821C236E8E4
        IROHA_GENESIS_ACCOUNT_PRIVATE_KEY: '{"digest_function":"ed25519","payload":"3e51c0c06622a95a27d4936d706318fc7430c652136d5348d55db0d47835783c0d80056a4cb7e32f6e0593c8708c03340d3d1e7f1050e635dbe0a821c236e8e4"}'
[...]

Issue #3507, opened 2023-05-21
[suggestion] Enhance configuration reference format


0x009922:
### Feature request

This might be split into 2 separate tasks:

1. Emit configuration reference as a machine-readable file, e.g. JSON
2. Format that structured machine-readable reference as a structured Markdown

The infrastructure for both is almost done. This complain is only about the format.

Related issue:

- #1639 

#### 1. Machine-readable

Sample format:

```json5
{
  schema: [
    // leaf root-level field
    // this field doesn't have `value.default`, i.e. it is REQUIRED
    {
      kind: "leaf",
      name: "public_key",
      documentation: "Public key of this peer",
      env: "IROHA_PUBLIC_KEY",
      value: {
        typedoc: "`PublicKey`",
        related_types: ["PublicKey"],
      },
    },
    // group of nested fields
    {
      kind: "nested",
      name: "logger",
      documentation: "Logger configuration",
      items: [
        // this field has `value.default`, i.e. it is OPTIONAL
        {
          kind: "leaf",
          name: "max_log_level",
          documentation: "Maximum log level",
          env: "MAX_LOG_LEVEL",
          value: {
            typedoc: `"TRACE" | "DEBUG" | "INFO" | "WARN" | "ERROR" | "FATAL"`,
            // this
            default: `"INFO"`,
          },
        },
      ],
    },
    // Contains references to a complex structures
    {
      kind: "nested",
      name: "sumeragi",
      documentation: "Sumeragi configuration",
      items: [
        {
          kind: "leaf",
          name: "trusted_peers",
          documentation: "Optional list of predefined trusted peers.",
          env: "SUMERAGI_TRUSTED_PEERS",
          value: {
            typedoc: "`PeerId`[]",
            default: `\`[]\``,
            example: `[{"address":"iroha0:1337", "public_key": "ed01201C61FAF8FE94E253B93114240394F79A607B7FA55F9E5A41EBEC74B88055768B"}]`,
            related_types: ["PeerId"],
          },
        },
      ],
    },
  ],
  types: {
    PeerId: {
      kind: "struct",
      documentation: "Peer id?",
      fields: [
        {
          name: "address",
          documentation: "Peer address",
          value: {
            typedoc: "string",
            example: `"iroha0:1337"`,
          },
        },
        {
          name: "public_key",
          documentation: "Peer public key",
          value: {
            typedoc: "`PublicKey`",
            related_types: ["PublicKey"],
          },
        },
      ],
    },
[...]

PR #3510, opened 2023-05-22
[documentation] #3509: Generate Kagami CLI help as Markdown

### LINKED ISSUE #3509: [documentation] Auto-generate Kagami docs from clap
### Description

The documentation for Kagami (in its README) is outdated. It will be even more outdated after the merge of #3475.

### Your suggestions

Use [`clap-markdown`](https://crates.io/crates/clap-markdown) or similar tool.

0x009922:
## Description

_Self-explanatory PR title._

### Linked issue

Closes #3509

### Benefits

Same documentation (from doccomments) for both CLI and GitHub.

### Checklist

- [x] Self-review
- [ ] Merge #3475 
- [ ] Doc review
- [x] Initial code review
- [x] Fix comments
- [ ] Re-review
- [ ] Decide whether we should proceed with `clap-markdown` or with a self-made tool
- [ ] Create a follow-up issue: generate kagami help in CI


tools/kagami/CLI.md
0x009922, line 1: I don't like the format that `clap-markdown` produces for a couple of reasons:

- Mix of `##` and `#####`
- Using bold in headers: `##### **Options:**`
- Inconsistent spacing
- `↴` character

Not critical though

0x009922: Also, it ignores additional text in arguments doccomments and only includes the first line.

0x009922:
It is not much code to write a better docgen implementation. It might be used for other CLIs as well.

tools/kagami/CLI.md
ilchu: What do you mean by this? From what I see, the doccomments are included in full (e.g. for genesis or crypto). Or did you mean that it doesn't preserve formatting without the `verbatim` attribute?

tools/kagami/CLI.md
```diff
@@ -0,0 +1,315 @@
...
+## `kagami config`
+
+Generate the default client/peer configuration
+
+**Usage:** `kagami config <COMMAND>`
+
+###### **Subcommands:**
+
+* `client` — 
+* `peer` —
```
ilchu, line 221: Would be good to have short descriptions here as well. Could be done by carrying over the old examples directly into the `Args::Config` docstring, just like you did with `Args::Genesis`, I suppose.

tools/kagami/CLI.md
ilchu, line 1: Also would prefer some different title, as either the old `README.md` or maybe `kagami.md` sound more descriptive to me. And maybe have this markdown generation a part of our CI, just like e.g. schema is (could be a separate small PR though).

tools/kagami/CLI.md
```diff
@@ -0,0 +1,315 @@
...
+
+**Usage:** `kagami docs`
+
+
+
+## `kagami tokens`
+
+Generate a list of predefined permission tokens and their parameters
+
+**Usage:** `kagami tokens`
```
ilchu, line 249: The comment about `Args::Config` also applies here, despite this part being insubstantial. Even though it's just two small and obvious examples there, I don't think we lose much by omitting them, but being more explicit never hurts.

tools/kagami/src/main.rs
```diff
@@ -66,18 +160,26 @@ pub enum Args {
...
+    /// issue](https://github.com/hyperledger/iroha/issues/3473)). If you are going to run it outside
     /// of the repo, make sure to pass `--no-default-configuration` flag.
     ///
     /// Be careful with specifying a Dockerhub image as a source: Kagami Swarm only guarantees that
     /// the docker-compose configuration it generates is compatible with the same Git revision it
     /// is built from itself. Therefore, if specified image is not compatible with the version of Swarm
     /// you are running, the generated configuration might not work.
+    #[clap(verbatim_doc_comment)]
     Swarm(swarm::Args),
+    /// Render CLI help message as Markdown
```
ilchu, line 181:
```suggestion
    /// Render `kagami` documentation and examples as Markdown
```
As CLI sounds too generic.

ilchu (changes requested):

ilchu:
> It is not much code to write a better docgen implementation. It might be used for other CLIs as well.

Been thinking about this too, you mean ditching the whole `Documented` trait and doing something similar for config endpoint docs?

0x009922:
> > It is not much code to write a better docgen implementation. It might be used for other CLIs as well.
> 
> Been thinking about this too, you mean ditching the whole `Documented` trait and doing something similar for config endpoint docs?

@ilchu,

Hmm, as far as I understand you are speaking about Iroha configuration documentation, which is generated by `kagami docs`. I am speaking not about it. (Although you are welcome to put your thoughts here: #3507)

Here I mean that `clap-markdown` implementation is not complex, and we can write our own clap-based markdown generator (e.g. `iroha_clap_markdown` crate), with a format that we consider a better option.

tools/kagami/CLI.md
```diff
@@ -0,0 +1,315 @@
...
+###### **Options:**
+
+* `--image <IMAGE>` — Use specified docker image
+* `--build <PATH>` — Use local path location of the Iroha source code to build images from
+* `--build-from-github` — Clone `hyperledger/iroha` repo from the revision Kagami is built itself, and use the cloned source code to build images from
+* `-p`, `--peers <PEERS>` — How many peers to generate within the docker-compose
+* `--outdir <OUTDIR>` — Target directory where to place generated files
+* `--outdir-force` — Re-create the target directory if it already exists
+* `--no-default-configuration` — Do not create default configuration in the `<outdir>/config` directory
+* `-s`, `--seed <SEED>` — Might be useful for deterministic key generation
```
0x009922, line 297: Here is the example of that `clap-markdown` doesn't render additional documentation for arguments. Even if I put `#[clap(verbatim_doc_comment)]`.

For example, `--no-default-configuration`'s full documentation is as follows:

> Do not create default configuration in the `<outdir>/config` directory.
> 
> Default `config.json`, `genesis.json` and `validator.wasm` are generated and put into
> the `<outdir>/config` directory. That directory is specified in the Docker Compose
> `volumes` field.
> 
> If you don't need the defaults, you could set this flag. The `config` directory will be
> created anyway, but you should put the necessary configuration there by yourself.
>

tools/kagami/CLI.md
0x009922: It definitely should be a part of CI.

0x009922: @ilchu, see: https://github.com/hyperledger/iroha/pull/3510#discussion_r1210888850

tools/kagami/CLI.md
ilchu: Ah, now I see it, thanks. So it doesn't recurse into the submodules args fully. That should be trivially fixable with a home-brew implementation (and sounds really close to what `Documented` does already anyways).

ilchu:
> @ilchu,
> 
> Hmm, as far as I understand you are speaking about Iroha configuration documentation, which is generated by `kagami docs`. I am speaking not about it. (Although you are welcome to put your thoughts here: #3507)

That's indeed what I was talking about. Then I'll see if I actually can come up with anything actionable up there!

QuentinI:
I'm a bit bothered that running subcommands with `--help` (e.g. `kagami crypto --help`) will output a wall of raw Markdown. I'm not sure the longer docs is even useful there as opposed to keeping it in documentation only; and of course unprocessed Markdown looks quite jarring in a help message

appetrosyan (approved):

6r1d:
To summarize our recent conversation, I like the format for the generated docs. Thank you for the detailed demo with Asciinema.

0x009922:
See: https://github.com/hyperledger/iroha/issues/3509#issuecomment-1581766055

PR #3514, opened 2023-05-22
[fix] #3498: Fix soft-fork detection

### LINKED ISSUE #3498: [bug]: Soft-fork detection is broken
At the moment soft-fork detection doesn't work as intended.

Erigara:
## Description

- Remove redundant fields from sumeragi;
- Fix soft-fork detection and signatrue validation;
- Add tests.

### Linked issue

Closes #3498 

### Benefits

Topology work correctly + new tests.

### Checklist

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -26,7 +26,6 @@ use sealed::sealed;
 use crate::{
     prelude::*,
     sumeragi::network_topology::{Role, Topology},
-    tx::TransactionValidator,
 };
 
 /// Transaction data is permanently recorded in chunks called
@@ -46,39 +45,36 @@ pub struct PendingBlock {
 }
 
 /// Builder for `PendingBlock`
-pub struct BlockBuilder<'a> {
+pub struct BlockBuilder {
     /// Block's transactions.
     pub transactions: Vec<VersionedAcceptedTransaction>,
     /// Block's event recommendations.
     pub event_recommendations: Vec<Event>,
-    /// The height of the block.
-    pub height: u64,
-    /// The hash of the previous block if there is one.
-    pub previous_block_hash: Option<HashOf<VersionedCommittedBlock>>,
     /// The view change index this block was committed with. Produced by consensus.
     pub view_change_index: u64,
     /// The topology thihs block was committed with. Produced by consensus.
     pub committed_with_topology: Topology,
     /// The keypair used to sign this block.
     pub key_pair: KeyPair,
-    /// The transaction validator to be used when validating the block.
-    pub transaction_validator: &'a TransactionValidator,
     /// The world state to be used when validating the block.
     pub wsv: WorldStateView,
 }
 
-impl BlockBuilder<'_> {
+impl BlockBuilder {
     /// Create a new [`PendingBlock`] from transactions.
     pub fn build(self) -> PendingBlock {
         let timestamp = crate::current_time().as_millis();
+        let height = self.wsv.height() + 1;
+        let previous_block_hash = self.wsv.latest_block_hash();
+        let transaction_validator = self.wsv.transaction_validator();
         // TODO: Need to check if the `transactions` vector is empty. It shouldn't be allowed.
 
         let mut header = BlockHeader {
             timestamp,
             consensus_estimation: DEFAULT_CONSENSUS_ESTIMATION_MS,
-            height: self.height,
+            height,
             view_change_index: self.view_change_index,
-            previous_block_hash: self.previous_block_hash,
+            previous_block_hash,
             transactions_hash: None,
             rejected_transactions_hash: None,
             committed_with_topology: self.committed_with_topology.sorted_peers,
@@ -88,10 +84,7 @@ impl BlockBuilder<'_> {
         let mut rejected = Vec::new();
 
         for tx in self.transactions {
-            match self
-                .transaction_validator
-                .validate(tx.into_v1(), header.is_genesis(), &self.wsv)
-            {
+            match transaction_validator.validate(tx.into_v1(), header.is_genesis(), &self.wsv) {
                 Ok(tx) => txs.push(tx),
                 Err(tx) => {
                     iroha_logger::warn!(
@@ -270,13 +263,7 @@ impl PendingBlock {
 pub trait Revalidate: Sized {
     /// # Errors
     /// - When the block is deemed invalid.
-    fn revalidate<const IS_GENESIS: bool>(
-        &self,
-        transaction_validator: &TransactionValidator,
-        wsv: WorldStateView,
-        latest_block: Option<HashOf<VersionedCommittedBlock>>,
-        block_height: u64,
-    ) -> Result<(), eyre::Report>;
+    fn revalidate<const IS_GENESIS: bool>(&self, wsv: WorldStateView) -> Result<(), eyre::Report>;
 
     /// Return whether or not the block contains transactions already committed.
     fn has_committed_transactions(&self, wsv: &WorldStateView) -> bool;
@@ -287,20 +274,8 @@ pub trait Revalidate: Sized {
 pub trait InGenesis: Sized + Revalidate {
     /// # Errors
     /// - When the block is deemed invalid.
-    fn revalidate(
-        &self,
-        transaction_validator: &TransactionValidator,
-        wsv: WorldStateView,
-        latest_block: Option<HashOf<VersionedCommittedBlock>>,
-        block_height: u64,
-    ) -> Result<(), eyre::Report> {
-        <Self as Revalidate>::revalidate::<true>(
-            self,
-            transaction_validator,
-            wsv,
-            latest_block,
-            block_height,
-        )
+    fn revalidate(&self, wsv: WorldStateView) -> Result<(), eyre::Report> {
+        <Self as Revalidate>::revalidate::<true>(self, wsv)
     }
 }
 
@@ -309,20 +284,8 @@ pub trait InGenesis: Sized + Revalidate {
 pub trait InBlock: Sized + Revalidate {
     /// # Errors
     /// - When the block is deemed invalid.
-    fn revalidate(
-        &self,
-        transaction_validator: &TransactionValidator,
-        wsv: WorldStateView,
-        latest_block: Option<HashOf<VersionedCommittedBlock>>,
-        block_height: u64,
-    ) -> Result<(), eyre::Report> {
-        <Self as Revalidate>::revalidate::<false>(
-            self,
-            transaction_validator,
-            wsv,
-            latest_block,
-            block_height,
-        )
+    fn revalidate(&self, wsv: WorldStateView) -> Result<(), eyre::Report> {
+        <Self as Revalidate>::revalidate::<false>(self, wsv)
     }
 }
 
@@ -342,13 +305,11 @@ impl Revalidate for PendingBlock {
     /// - Block header transaction hashes don't match with computed transaction hashes
     /// - Error during revalidation of individual transactions
     #[allow(clippy::too_many_lines)]
-    fn revalidate<const IS_GENESIS: bool>(
-        &self,
-        transaction_validator: &TransactionValidator,
-        wsv: WorldStateView,
-        latest_block: Option<HashOf<VersionedCommittedBlock>>,
-        block_height: u64,
-    ) -> Result<(), eyre::Report> {
+    fn revalidate<const IS_GENESIS: bool>(&self, wsv: WorldStateView) -> Result<(), eyre::Report> {
+        let latest_block_hash = wsv.latest_block_hash();
+        let block_height = wsv.height();
+        let transaction_validator = wsv.transaction_validator();
+
         if self.transactions.is_empty() && self.rejected_transactions.is_empty() {
             bail!("Block is empty");
         }
@@ -357,10 +318,10 @@ impl Revalidate for PendingBlock {
             bail!("Block has committed transactions");
         }
 
-        if latest_block != self.header.previous_block_hash {
+        if latest_block_hash != self.header.previous_block_hash {
             bail!(
                     "Mismatch between the actual and expected hashes of the latest block. Expected: {:?}, actual: {:?}",
-                    latest_block,
+                    latest_block_hash,
                     &self.header.previous_block_hash
                 );
         }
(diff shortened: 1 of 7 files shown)


client/benches/tps/utils.rs
```diff
@@ -294,7 +294,7 @@ impl MeasurerUnit {
         let transfer_rose = TransferBox::new(
             asset_id(self.name),
             1_u32.to_value(),
-            asset_id(self.next_name),
+            account_id(self.next_name),
```
appetrosyan, line 297: Good catch. It was broken but we didn't spot it, because we don't routinely run benchmarks.

core/src/sumeragi/network_topology.rs
```diff
@@ -42,14 +42,17 @@ impl Topology {
     pub fn is_consensus_required(&self) -> bool {
         self.min_votes_for_commit() > 1
     }
+
```
appetrosyan, line 45: In general, I would avoid whitespace changes if possible. GitHub could have implemented filtering those ages ago, but they likely never will, so this is just making life harder for the reviewer.

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
         topology.update_peer_list(new_peers);
         assert_eq!(extract_ports(&topology), vec![0, 2, 5, 7])
     }
+
+    #[test]
+    fn filter_by_role() {
```
appetrosyan, line 361: Very good unit test. I'd recommend that you make it parametric, so that it accepts the number of key pairs (the macro would have to be inlined), so we could convert this to a `proptest`.

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+            topology.filter_signatures_by_roles(&[Role::ProxyTail], signatures.iter());
+        assert_eq!(proxy_tail_signatures.len(), 1);
+        assert_eq!(proxy_tail_signatures[0].public_key(), peers[4].public_key());
+
+        let validating_peers_signatures =
+            topology.filter_signatures_by_roles(&[Role::ValidatingPeer], signatures.iter());
+        assert_eq!(validating_peers_signatures.len(), 3);
+        assert!(validating_peers_signatures
+            .iter()
+            .all(|s| peers[1..4].iter().any(|p| p.public_key() == s.public_key())));
```
appetrosyan, line 392: This isn't exactly checking for deterministic topologies. It might also miss some cases where the peers all have the same public key (which is a problem).

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+
+        let observing_peers_signatures =
+            topology.filter_signatures_by_roles(&[Role::ObservingPeer], signatures.iter());
+        assert_eq!(observing_peers_signatures.len(), 2);
+        assert!(observing_peers_signatures
+            .iter()
+            .all(|s| peers[5..].iter().any(|p| p.public_key() == s.public_key())));
```
appetrosyan, line 399: Same here. I would suggest defining a function which actually checks  that these match the peers you would expect iterated in the correct order.

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+    }
+
+    #[test]
+    fn roles() {
```
appetrosyan, line 403: Consider converting into proptest.

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn proxy_tail() {
+        let peers = peers![0, 1, 2, 3, 4, 5, 6];
+        let topology = Topology::new(peers.clone());
+
+        assert_eq!(topology.proxy_tail(), &peers[4]);
+    }
```
appetrosyan, line 431: Consider making a `proptest`.

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+
+    #[test]
+    fn leader() {
```
appetrosyan, line 434: Not a good `proptest` candidate.

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+        let mut signatures = key_pairs
+            .iter()
+            .enumerate()
+            .filter(|(i, _)| *i == 2) // Retain only last signature
+            .map(|(_, key_pair)| SignatureOf::new(key_pair.clone(), &dummy).expect("Failed to sign"))
+            .collect::<Result<SignaturesOf<_>, _>>()
+            .expect("Failed to create `SignaturesOf`");
+
+        let result = topology.verify_signatures(&mut signatures, HashOf::new(&dummy));
+        assert_eq!(result, Ok(()))
```
appetrosyan, line 484:
```suggestion
        assert!(topology.verify_signatures(&mut signatures, HashOf::new(&dummy)).is_ok())
```

core/src/sumeragi/network_topology.rs
```diff
@@ -234,4 +356,202 @@ mod tests {
...
+        let mut signatures = key_pairs
+            .iter()
+            .map(|key_pair| SignatureOf::new(key_pair.clone(), &dummy).expect("Failed to sign"))
+            .collect::<Result<SignaturesOf<_>, _>>()
+            .expect("Failed to create `SignaturesOf`");
+
+        assert_eq!(
+            topology.verify_signatures(&mut signatures, HashOf::new(&dummy)),
+            Ok(())
+        );
```
appetrosyan, line 461:
```suggestion
        assert!(
            topology.verify_signatures(&mut signatures, HashOf::new(&dummy)).is_ok()
        );
```

core/src/sumeragi/network_topology.rs
```diff
@@ -163,16 +262,39 @@ pub enum Role {
...
+    /// Error when block doesn't have proxy tail signature
+    #[error("The block doesn't have proxy tail signature")]
+    ProxyTailMissing,
+    /// Error when block doesn't have leader signature
+    #[error("The block doesn't have leader signature")]
+    LeaderMissing,
+}
+
 #[cfg(test)]
 mod tests {
```
appetrosyan, line 285: Good unit test coverage would have caught the error early. This is why it's important.

appetrosyan: For each test, I'd refer to the relevant section of the whitepaper so we can verify which parts are to-spec and which are not.

core/src/sumeragi/network_topology.rs
```diff
@@ -163,16 +262,39 @@ pub enum Role {
     ProxyTail,
 }
 
+/// Error during signature verification
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SignatureVerificationError {
+    /// Error when block doesn't have enough valid signatures
+    #[error("The block doesn't have enough valid signatures to be committed ({votes_count} out of {min_votes_for_commit}).")]
```
appetrosyan, line 269: I was thinking of adding some primitive error types. We have common patterns like not enough of X, which has verbatim the same error message, that is inconsistently replicated all over the codebase. 

```rust
struct Insufficient<T: Ord> {
    expected: T, 
    actual: T, 
}
```
with some basic convenience functions.

core/src/sumeragi/network_topology.rs
```diff
@@ -125,29 +128,125 @@ impl Topology {
...
+
+        topology.update_topology(block, new_peers);
+
+        // Rotate all once for every view_change
+        topology.rotate_all_n(view_change_index);
+
+        topology
+    }
+
+    /// Check if block's signatures meet requirements for given topology
```
appetrosyan, line 198: Please refer to requirements in the whitepaper, or (better) inline the requirements to the docstring. 

Consensus is important, and good documentation here pays off in hours saved debugging.

core/src/sumeragi/network_topology.rs
```diff
@@ -125,29 +128,125 @@ impl Topology {
...
+            .filter_signatures_by_roles(&[Role::Leader], signatures.iter())
+            .is_empty()
+        {
+            return Err(SignatureVerificationError::LeaderMissing);
+        }
+
+        if self
+            .filter_signatures_by_roles(&[Role::ProxyTail], signatures.iter())
+            .is_empty()
+        {
```
appetrosyan, line 244: How do we tell that we have enough `ValidatinPeer` signatures? It's not obvious to me from the code.

core/src/sumeragi/network_topology.rs
Erigara: I would like to keep is this way because this way we can see `rhs` and what exact error we received.

appetrosyan (changes requested):

core/src/sumeragi/network_topology.rs
```diff
@@ -125,29 +128,125 @@ impl Topology {
...
+        self.rotate_all_n(1);
     }
+
+    /// Rotate peers n times where n is a number of failed attempt to create a block.
+    pub fn rotate_all_n(&mut self, n: usize) {
+        let len = self.sorted_peers.len();
+        if let Some(rem) = n.checked_rem(len) {
+            self.sorted_peers.rotate_left(rem);
+        }
+    }
```
appetrosyan, line 153: There should be a way to do this in $O(1)$

core/src/sumeragi/network_topology.rs
```diff
@@ -125,29 +128,125 @@ impl Topology {
     pub fn leader(&self) -> &PeerId {
         &self.sorted_peers[0]
     }
+
     /// Get proxy tail's peer id.
     pub fn proxy_tail(&self) -> &PeerId {
-        &self.sorted_peers[self.min_votes_for_commit()]
+        &self.sorted_peers[self.min_votes_for_commit() - 1]
```
appetrosyan, line 134: Comment on why $-1$

core/src/sumeragi/network_topology.rs
```diff
@@ -105,15 +108,15 @@ impl Topology {
...
     /// the function will return `Role::ObservingPeer`.
     // This lint is a bad suggestion.
-    #[allow(clippy::option_if_let_else)]
     pub fn role(&self, peer_id: &PeerId) -> Role {
         match self.sorted_peers.iter().position(|p| p == peer_id) {
             Some(index) if index == 0 => Role::Leader,
-            Some(index) if index < self.min_votes_for_commit() => Role::ValidatingPeer,
-            Some(index) if index == self.min_votes_for_commit() => Role::ProxyTail,
+            Some(index) if index < self.min_votes_for_commit() - 1 => Role::ValidatingPeer,
+            Some(index) if index == self.min_votes_for_commit() - 1 => Role::ProxyTail,
```
appetrosyan, line 119: Comment on why we need to do `-1`. This can save searching far and wide to find the reason.

core/src/sumeragi/network_topology.rs
Erigara: We check for total amount of signatures that we have at least 2f+1 signatures and then individually for proxy tail and leader signatures.

core/src/sumeragi/network_topology.rs
Erigara: Newline between methods is common pattern across our codebase, so should i revert this change?

core/src/sumeragi/main_loop.rs
```diff
@@ -1082,3 +1008,263 @@ fn early_return(
...
+
+        (wsv, kura, block)
+    }
+
+    #[test]
+    fn block_sync_invalid_block() {
+        let (finalized_wsv, _, mut block) = create_data_for_test();
+        let wsv = finalized_wsv.clone();
+
+        // Mailfrom block to make it invalid
```
0x009922, line 1159:
```suggestion
        // Malform block to make it invalid
```

?

core/src/sumeragi/main_loop.rs
```diff
@@ -1082,3 +1008,263 @@ fn early_return(
...
+
+    #[test]
+    fn block_sync_invalid_soft_fork_block() {
+        let (finalized_wsv, kura, mut block) = create_data_for_test();
+        let wsv = finalized_wsv.clone();
+
+        kura.store_block(block.clone());
+        wsv.apply(&block).expect("Failed to apply block");
+
+        // Mailfrom block to make it invalid
```
0x009922, line 1174:
```suggestion
        // Malform block to make it invalid
```

Same

core/src/sumeragi/network_topology.rs
0x009922: As for me, I don't have any cognitive pressure from these newlines. Instead, it looks satisfying.

core/src/sumeragi/network_topology.rs
0x009922: Or add self-explanatory local (or non-local) constant:

```rust
const SIZE_AND_INDEX_DIFFERENCE = 1;
```

It will be compiled away anyway (haven't checked though).

core/src/sumeragi/network_topology.rs
```diff
@@ -105,15 +108,15 @@ impl Topology {
             .cloned()
             .collect()
     }
+
     /// What role does this peer have in the topology. If it is not in the toplogy
     /// the function will return `Role::ObservingPeer`.
     // This lint is a bad suggestion.
```
0x009922, line 114:
```suggestion
```

core/src/sumeragi/network_topology.rs
Erigara: I moved this logic into `proxy_tail_index` private function with required clarifications.

core/src/sumeragi/network_topology.rs
Erigara: So how this can look like imo:

1. Use circular linked list and have a pointer to current head of the list.
2. In such case `rotate_set_a` operation will be to just moving leader element to the proxy tail position which is `O(1)`.
3. `rotate_all_n` will be just moving head pointer `n` elements forward

To optimize allocation fragmentation we can use [typed_arena](https://docs.rs/typed-arena/latest/typed_arena) for node storage.

However `update_peer_list` and `lift_up_peers` might be not so optimal in such case.

Erigara: Hmm, probably situation can be even simpler than that. 
If we use observation that we construct topology once per round and then only use rotate all.
We can just store index of current leader, then `rotate_all` will be just increment of this index.

Erigara: Created #3519 for further discussion.

core/src/sumeragi/network_topology.rs
appetrosyan: Add as comments and I'll approve

core/src/sumeragi/network_topology.rs
Erigara: Created #3524.

core/src/sumeragi/network_topology.rs
appetrosyan: > Newline between methods is common pattern across our codebase, so should i revert this change?

Not this instance. Just don't add newlines to existing code if you have a PR with important contents. 

This is less of a problem, because you use conventional commits.

appetrosyan (approved):

core/src/block.rs
```diff
@@ -46,39 +42,36 @@ pub struct PendingBlock {
...
 }
 
-impl BlockBuilder<'_> {
+impl BlockBuilder {
     /// Create a new [`PendingBlock`] from transactions.
     pub fn build(self) -> PendingBlock {
         let timestamp = crate::current_time().as_millis();
+        let height = self.wsv.height() + 1;
+        let previous_block_hash = self.wsv.latest_block_hash();
+        let transaction_validator = self.wsv.transaction_validator();
```
SamHSmith, line 66: Smart.

SamHSmith (approved):
I don't see any issues with this.

PR #3515, opened 2023-05-22
[fix] #3396: Store block in kura before emitting block comitted event

### LINKED ISSUE #3396: [bug]: Tps benchmark fails block missing

Erigara:
### Linked issue

Closes #3396  

### Benefits

No situation when block is committed but not yet in kura.

### Checklist


core/src/sumeragi/main_loop.rs
```diff
@@ -318,16 +318,14 @@ impl Sumeragi {
...
         self.cache_transaction();
     }
 
     fn replace_top_block(&mut self, block: impl Into<VersionedCommittedBlock>) {
-        let committed_block = block.into();
+        let committed_block = Arc::new(block.into());
 
         self.wsv = self.finalized_wsv.clone();
-        self.update_state(&committed_block);
+        self.update_state(&Arc::clone(&committed_block), Kura::replace_top_block);
```
appetrosyan, line 328: Nice.

core/src/sumeragi/main_loop.rs
```diff
@@ -367,22 +363,30 @@ impl Sumeragi {
         self.connect_peers(&self.current_topology);
     }
 
-    fn update_state(&mut self, committed_block: &VersionedCommittedBlock) {
+    fn update_state(
+        &mut self,
+        committed_block: &Arc<VersionedCommittedBlock>,
```
appetrosyan, line 368: We don't want to increment the refcount, but at the same time we don't want to generify with an `AsRef`, to not accept any other smart pointer?

appetrosyan (dismissed):
LGTM. Minor comments only.

core/src/kura.rs
```diff
@@ -319,17 +319,17 @@ impl Kura {
...
-            .lock()
-            .push((block.hash(), Some(Arc::new(block))));
+    pub fn store_block(&self, block: impl Into<Arc<VersionedCommittedBlock>>) {
+        let block = block.into();
+        self.block_data.lock().push((block.hash(), Some(block)));
     }
 
     /// Replace the block in `Kura`'s in memory block store.
-    pub fn replace_top_block(&self, block: VersionedCommittedBlock) {
+    pub fn replace_top_block(&self, block: impl Into<Arc<VersionedCommittedBlock>>) {
```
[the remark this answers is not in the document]
Erigara: I want to make changes in such a way that don't break previous code, and such signature work just fin in this case.

core/src/sumeragi/main_loop.rs
Erigara: It's just clippy complaining that this value could be passed by reference instead of value.
I decide to obey instead of adding ignore because see no harm in such signature.

core/src/sumeragi/main_loop.rs
```diff
@@ -367,22 +363,30 @@ impl Sumeragi {
...
             .expect("Failed to apply block on WSV. Bailing.");
 
         self.send_events(self.wsv.events_buffer.replace(Vec::new()));
 
         // Parameters are updated before updating public copy of sumeragi
         self.update_params();
 
+        // Block must be stored by kura after updating WSV because apply block might fail and we will store corrupted block
+        // but before updating public copy of wsv and emitting BlockCommitted event to avoid situation when block is not yet stored in kura
+        kura_store_block(&self.kura, Arc::clone(committed_block));
```
0x009922, line 382:
```suggestion
        // https://github.com/hyperledger/iroha/issues/3396:
        // Iroha should emit `BlockCommitted` event only after Kura stored the block.
        // However, WSV update should happen first, in order to avoid storaging a corrupted block
        kura_store_block(&self.kura, Arc::clone(committed_block));
```

SamHSmith (approved):

client/benches/tps/utils.rs
```diff
@@ -294,7 +294,7 @@ impl MeasurerUnit {
         let transfer_rose = TransferBox::new(
             asset_id(self.name),
             1_u32.to_value(),
-            asset_id(self.next_name),
+            account_id(self.next_name),
```
ilchu, line 297: So the benchmark too was doing a somewhat off thing? I wonder if it would help if we could have the benchmarks integrated in our CI.

ilchu (approved):

client/benches/tps/utils.rs
Erigara: Yes, basically API was changed, but benchmark is not.
I think it will help a lot to at least run tps oneshot in the CI. 
Because i fixed such errors on multiple occasions already.

core/src/sumeragi/main_loop.rs
```diff
@@ -367,22 +363,31 @@ impl Sumeragi {
...
             .expect("Failed to apply block on WSV. Bailing.");
 
         self.send_events(self.wsv.events_buffer.replace(Vec::new()));
 
         // Parameters are updated before updating public copy of sumeragi
         self.update_params();
 
+        // https://github.com/hyperledger/iroha/issues/3396
+        // Iroha should emit `BlockCommitted` event only after Kura stored the block.
+        // However, WSV update should happen first, in order to avoid storing a corrupted block
```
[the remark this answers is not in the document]
Erigara: I like wording on this i will put this in the other PR.

Issue #3516, opened 2023-05-22
[suggestion] Allow querying for original WASM blobs/sources 


QuentinI:
### Feature request

Provide an API to match WASM blobs as stored on blockchain with `LoadedModules` as they are in the WSV.
As a way more ambitious goal - figure out building WASM contracts reproducibly and do that on the node, allowing to trace WASM back to the source code.

### Motivation

Trivial motivation for me: needed by iroha-squash to report which contract is which when squashing blockstore.
Overall motivation -  SCs would be easier to inspect in public blockchain use cases.

mversic:
as I mentioned [here](https://github.com/hyperledger/iroha/pull/3517#issuecomment-1558838514) returning compiled wasm was an oversight. We should change the API of `FindTriggerXXX` queries to return whatever user has submitted, not our internal representation

mversic:
If we go with this suggestion (and I strongly think we should), I don't see how we can have queries that return both optimized or original wasm blob. We should make a choice what these `FindTriggerXXX` queries should return.

@Mingela do you mind us changing the behavior of these queries to return original wasm smart contract as was submitted (and as is recorded on the blockchain). Was there a reason why you would need a compiled one? How would you use it?

Arjentix:
We still need optimized WASM to be able to locally debug wasm if something goes wrong.

As I mentioned before, our queries report with the current world view. And world never contains original wasm, it always stores the optimized version.

I think it should be some different query of even different *kind* of query. And I don't think it was an oversight.

mversic:
> As I mentioned before, our queries report with the current world view. And world never contains original wasm, it always stores the optimized version.

world state view is not a concept that end user should ever hear about. Justifying API decisions based on internal representation (and wsv is an IR) of data is wrong IMO and opens a whole new can of worms.

> And I don't think it was an oversight

ok, let's not call it that

Let me make a summary, trigger queries should return original wasm because:
1. this is what user expects to be returned (needing an optimized wasm, is for the power user, not common user)
2. returning optimized wasm exposes our internal representation (it tells everyone that we're using `wasmtime`)
3. optimized wasm representation is not stable. This ties our public facing API to a concrete implementation of wasm VM (`wasmtime` ATM). Not only that, it also ties public API to a specific version of `wasmtime` since even that is not stable
4. different Iroha nodes could be running different versions of `wasmtime`. This itself makes returning optimized wasm almost entirely useless 
5. you will still be able to debug, even more so with original wasm

[...]

PR #3517, opened 2023-05-22
[feature] #3516: retain original blob hash in LoadedExecutable

QuentinI:
## Description

Stop-gap solution to make recovering original WASM blobs easier for WorldStateView API consumers such as iroha-squash.

### Linked issue

Partially addresses #3516

### Benefits

As per description, squashing blockchain is nigh impossible otherwise, since we need to report which smartcontracts to put where.

### Checklist


core/src/smartcontracts/isi/triggers/mod.rs
```diff
@@ -6,7 +6,7 @@ use iroha_data_model::{
 };
 use iroha_telemetry::metrics;
 
-pub(crate) mod set;
```
QuentinI, line 9: Could be made more granular, but almost the whole module leaks through with `WorldStateView::triggers` anyways, so I decided against it unless someone objects.

appetrosyan: That was the original intention. `WorldStateView::triggers` is inspired by SBCL hooks.

appetrosyan (approved):

0x009922 (commented):
Nothing to say, just looked through.

mversic:
I think this whole problem is because our `FindTriggerXXX` queries don't return the trigger as stored on the blockchain, rather they return our internal representation (preloaded wasm module) of the smart contract. This is inconvenient, but also ties us to a specific wasm runtime because now we're leaking our internals through the API

appetrosyan (approved):

Arjentix (approved):
I like that solution of our problem

PR #3490, opened 2023-05-18
[feature] #3376: Add Iroha Client CLI pytest automation framework

astrokov7:
## Description

This PR introduces automated tests for the Iroha Client CLI, using Python's Pytest framework. The new test suite includes a collection of tests that verify the functionality of the client CLI in various scenarios.

PR #3547, opened 2023-05-30
[fix] #3350: Core testing

### LINKED ISSUE #3350: Investigate core module unit testing.

SamHSmith:
## Description

This PR adds five new tests for sumeragi inside the core trait. I still need to refactor the PR to eliminate the duplicated code between the tests.

### Linked issue

Closes #3350

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -73,6 +73,8 @@ hex = "0.4.3"
 byte-unit = "4.0.18"
 once_cell = "1.16.0"
 tempfile = "3.3.0"
+iroha = { version = "=2.0.0-pre-rc.16", path = "../cli" }
+
 
 [[bench]]
 name = "validation"
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -3,11 +3,10 @@
 use std::sync::mpsc;
 
 use iroha_data_model::{block::*, transaction::error::TransactionExpired};
-use iroha_p2p::UpdateTopology;
 use tracing::{span, Level};
 
 use super::*;
-use crate::{block::*, sumeragi::tracing::instrument};
+use crate::block::*;
 
 /// `Sumeragi` is the implementation of the consensus.
 pub struct Sumeragi {
@@ -31,8 +30,12 @@ pub struct Sumeragi {
     pub max_txs_in_block: usize,
     /// Kura instance used for IO
     pub kura: Arc<Kura>,
-    /// [`iroha_p2p::Network`] actor address
-    pub network: IrohaNetwork,
+    /// A function pointer used when posting sumeragi packets to the network.
+    pub post_procedure: SumeragiMessagePostProcedure,
+    /// A function pointer used when sending the new network topology to the network subsystem.
+    pub update_topology_procedure: SumeragiUpdateNetworkTopologyProcedure,
+    /// A function pointer used to query the currently connected peers.
+    pub get_connected_peers_procedure: SumeragiGetCurrentlyConnectedProcedure,
     /// Receiver channel.
     pub message_receiver: mpsc::Receiver<MessagePacket>,
     /// Only used in testing. Causes the genesis peer to withhold blocks when it
@@ -69,19 +72,6 @@ impl Debug for Sumeragi {
 }
 
 impl Sumeragi {
-    /// Send a sumeragi packet over the network to the specified `peer`.
-    /// # Errors
-    /// Fails if network sending fails
-    #[instrument(skip(self, packet))]
-    #[allow(clippy::needless_pass_by_value)] // TODO: Fix.
-    fn post_packet_to(&self, packet: MessagePacket, peer: &PeerId) {
-        let post = iroha_p2p::Post {
-            data: NetworkMessage::SumeragiPacket(Box::new(packet.into())),
-            peer_id: peer.clone(),
-        };
-        self.network.post(post);
-    }
-
     #[allow(clippy::needless_pass_by_value, single_use_lifetimes)] // TODO: uncomment when anonymous lifetimes are stable
     fn broadcast_packet_to<'peer_id>(
         &self,
@@ -89,22 +79,26 @@ impl Sumeragi {
         ids: impl IntoIterator<Item = &'peer_id PeerId> + Send,
     ) {
         for peer_id in ids {
-            self.post_packet_to(msg.clone(), peer_id);
+            (self.post_procedure)(msg.clone(), peer_id);
         }
     }
 
     #[allow(clippy::needless_pass_by_value)]
     fn broadcast_packet(&self, msg: MessagePacket) {
-        let broadcast = iroha_p2p::Broadcast {
-            data: NetworkMessage::SumeragiPacket(Box::new(msg.into())),
-        };
-        self.network.broadcast(broadcast);
+        for peer_id in self
+            .current_topology
+            .sorted_peers
+            .iter()
+            .filter(|p| **p != self.peer_id)
+        {
+            (self.post_procedure)(msg.clone(), peer_id);
+        }
     }
 
     /// Connect or disconnect peers according to the current network topology.
     fn connect_peers(&self, topology: &Topology) {
         let peers = topology.sorted_peers.clone().into_iter().collect();
-        self.network.update_topology(UpdateTopology(peers));
+        (self.update_topology_procedure)(peers);
     }
 
     /// The maximum time a sumeragi round can take to produce a block when
@@ -230,8 +224,6 @@ impl Sumeragi {
     }
 
     fn sumeragi_init_commit_genesis(&mut self, genesis_network: GenesisNetwork) {
-        std::thread::sleep(Duration::from_millis(250));
-
         info!("Initializing iroha using the genesis block.");
 
         assert_eq!(self.wsv.height(), 0);
--- a/core/src/sumeragi/mod.rs
+++ b/core/src/sumeragi/mod.rs
@@ -7,6 +7,7 @@
     clippy::std_instead_of_alloc
 )]
 use std::{
+    collections::HashSet,
     fmt::{self, Debug, Formatter},
     sync::{mpsc, Arc},
     time::{Duration, Instant},
@@ -22,7 +23,7 @@ use iroha_telemetry::metrics::Metrics;
 use network_topology::{Role, Topology};
 use tokio::sync::watch;
 
-use crate::handler::ThreadHandler;
+use crate::{handler::ThreadHandler, IrohaNetwork};
 
 pub mod main_loop;
 pub mod message;
@@ -35,9 +36,11 @@ use self::{
     message::{Message, *},
     view_change::{Proof, ProofChain},
 };
-use crate::{
-    block::*, kura::Kura, prelude::*, queue::Queue, EventsSender, IrohaNetwork, NetworkMessage,
-};
+use crate::{block::*, kura::Kura, prelude::*, queue::Queue, EventsSender, NetworkMessage};
+
+type SumeragiMessagePostProcedure = Arc<(dyn Fn(MessagePacket, &PeerId) + Send + Sync)>;
+type SumeragiUpdateNetworkTopologyProcedure = Arc<(dyn Fn(HashSet<PeerId>) + Send + Sync)>;
+type SumeragiGetCurrentlyConnectedProcedure = Arc<(dyn Fn() -> HashSet<PeerId> + Send + Sync)>;
 
 /*
 The values in the following struct are not atomics because the code that
@@ -55,13 +58,13 @@ struct LastUpdateMetricsData {
 #[derive(Clone)]
 pub struct SumeragiHandle {
     public_wsv_receiver: watch::Receiver<WorldStateView>,
-    message_sender: mpsc::SyncSender<MessagePacket>,
     metrics: Metrics,
     last_update_metrics_mutex: Arc<Mutex<LastUpdateMetricsData>>,
-    network: IrohaNetwork,
+    get_connected_peers_procedure: SumeragiGetCurrentlyConnectedProcedure,
     kura: Arc<Kura>,
     queue: Arc<Queue>,
     _thread_handle: Arc<ThreadHandler>,
+    message_sender: mpsc::SyncSender<MessagePacket>, // This must be kept alive until the thread exits, ergo it's after _thread_handle.
 }
 
 impl SumeragiHandle {
@@ -94,9 +97,8 @@ impl SumeragiHandle {
     /// - If either mutex is poisoned
     #[allow(clippy::cast_precision_loss)]
     pub fn update_metrics(&self) -> Result<()> {
-        let online_peers_count: u64 = self
-            .network
-            .online_peers(std::collections::HashSet::len)
+        let online_peers_count: u64 = (self.get_connected_peers_procedure)()
+            .len()
             .try_into()
             .expect("casting usize to u64");
 
@@ -192,13 +194,58 @@ impl SumeragiHandle {
         SumeragiStartArgs {
             configuration,
             events_sender,
+            wsv,
+            queue,
+            kura,
+            genesis_network,
+            block_hashes,
+            network,
+        }: SumeragiStartArgs,
+    ) -> SumeragiHandle {
+        let network2 = network.clone();
+        let network3 = network.clone();
+        SumeragiHandle::start_closure(SumeragiStartClosureArgs {
+            configuration,
+            events_sender,
+            wsv,
+            queue,
+            kura,
+            genesis_network,
+            block_hashes,
+            post_procedure: Arc::new(move |packet, peer_id| {
+                let post = iroha_p2p::Post {
+                    data: NetworkMessage::SumeragiPacket(Box::new(packet.into())),
+                    peer_id: peer_id.clone(),
+                };
+                network.post(post);
+            }),
+            update_topology_procedure: Arc::new(move |network_topology| {
+                network2.update_topology(iroha_p2p::UpdateTopology(network_topology));
+            }),
+            get_connected_peers_procedure: Arc::new(move || {
+                network3.online_peers(std::collections::HashSet::clone)
+            }),
+        })
+    }
+
+    /// Start [`Sumeragi`] actor and return handle to it. Takes network closures
+    /// as parameters. If you want to pass `IrohaNetwork` handle use `start`.
+    ///
+    /// # Panics
+    /// May panic if something is of during initialization which is bug.
+    pub fn start_closure(
+        SumeragiStartClosureArgs {
+            configuration,
+            events_sender,
             mut wsv,
             queue,
             kura,
-            network,
             genesis_network,
             block_hashes,
-        }: SumeragiStartArgs,
+            post_procedure,
+            update_topology_procedure,
+            get_connected_peers_procedure,
+        }: SumeragiStartClosureArgs,
     ) -> SumeragiHandle {
         let (message_sender, message_receiver) = mpsc::sync_channel(100);
 
@@ -261,7 +308,9 @@ impl SumeragiHandle {
             block_time: Duration::from_millis(configuration.block_time_ms),
             max_txs_in_block: configuration.max_transactions_in_block as usize,
             kura: Arc::clone(&kura),
-            network: network.clone(),
+            post_procedure,
+            update_topology_procedure,
+            get_connected_peers_procedure: Arc::clone(&get_connected_peers_procedure),
             message_receiver,
             debug_force_soft_fork,
             current_topology,
@@ -288,7 +337,6 @@ impl SumeragiHandle {
 
         let thread_handle = ThreadHandler::new(Box::new(shutdown), thread_handle);
         SumeragiHandle {
-            network,
             queue,
             kura,
             message_sender,
@@ -299,6 +347,7 @@ impl SumeragiHandle {
                 metric_tx_amounts: 0.0_f64,
                 metric_tx_amounts_counter: 0,
             })),
+            get_connected_peers_procedure,
             _thread_handle: Arc::new(thread_handle),
         }
     }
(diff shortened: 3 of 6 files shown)


Erigara (changes requested):
I'm not sure about taken direction for this tests. 

This tests is not granular enough to consider then unit tests imo.

It resemble my our ordinary integration tests but with some functionality ripped from other actors and without client part. 

If semi intergation tests is direction we want to take then i think we need to run all our actors but without torii and client part + change p2p `Network` to communicate through rust channels (call it smt like `LocalNetwork`) and make it use the same `IrohaNetwork` actor handle.

core/tests/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+    }
+    panic!("Sumeragi did not submit genesis.");
+}
+
+#[test]
+fn sumeragi_all_peers_commit_genesis() {
+    let mut configuration_for_logger = {
+        let mut sample_proxy =
+            iroha::samples::get_config_proxy(HashSet::new(), Some(KeyPair::generate().unwrap()));
+        let env_proxy = ConfigurationProxy::from_env();
```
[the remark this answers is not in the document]
mversic: should a unit test depend on configuration? I don't think so. IMO we should hardcode some value for `TransactionLimits` and logger if we need telemetry

core/src/sumeragi/main_loop.rs
```diff
@@ -32,8 +31,12 @@ pub struct Sumeragi {
     pub max_txs_in_block: usize,
     /// Kura instance used for IO
     pub kura: Arc<Kura>,
-    /// [`iroha_p2p::Network`] actor address
```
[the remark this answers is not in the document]
Arjentix: I agree with others.
Although making code testable requires to change the working code, it can be done with a more convinient way.
Maybe `mockall` crate might help here.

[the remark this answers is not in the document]
Arjentix: Or maybe you can create something like:

```rust
#[cfg(test)]
pub fn set_post_procedure(procedure: SumeragiMessagePostProcedure) {}
```

And then use it in tests

core/tests/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+use iroha_core::sumeragi::main_loop::Sumeragi;
+
+use iroha_data_model::transaction::Accept;
+
+
+/// Get a standardised key-pair from the hard-coded literals.
+///
+/// # Panics
+/// Programmer error. Given keys must be in proper format.
+pub fn get_key_pair() -> KeyPair {
```
Arjentix, line 28: This comment makes me think that this function accepts some kind of keys but it's not. I'd replace it like this:

```suggestion
/// Programmer error. Hardcoded keys must be in proper format.
pub fn get_key_pair() -> KeyPair {
```

core/tests/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+        )
+        .expect("Public key not in mulithash format"),
+        PrivateKey::from_hex(
+            Algorithm::Ed25519,
+            "9AC47ABF59B356E0BD7DCBBBB4DEC080E302156A48CA907E47CB6AEA1D32719E7233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0".as_ref(),
+        ).expect("Private key not hex encoded")
+    ).expect("Key pair mismatch")
+}
+
+static GENESIS_KEY : Mutex<Option<KeyPair>> = Mutex::new(None);
```
Arjentix, line 41: Can you use `once_cell::sync::Lazy` here as well?

core/tests/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+static mut GENESIS_CACHE : Mutex<Option<RawGenesisBlock>> = Mutex::new(None);
+
+impl TestGenesis for GenesisNetwork {
+    fn test(cfg: Configuration, submit_genesis: bool) -> Option<Self> {
+        // TODO: Fix this somehow. Probably we need to make `kagami` a library (#3253).
+        let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+        let content = std::fs::read_to_string(manifest_dir.join("../configs/peer/genesis.json"))
+            .expect("Failed to read data from configs/peer/genesis.json");
+        
+        let mut genesis: RawGenesisBlock = unsafe {
```
Arjentix, line 91: I think code like this needs `SAFETY` comment. Probably you should guarantee that this function will never be called in parallel or somtehing like that

core/tests/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+    let peer_ids = vec![PeerId::new(&"localhost:0".parse().unwrap(), &public_key), PeerId::new(&"localhost:1".parse().unwrap(), other_key.public_key())];
+
+    let mut configuration = {
+        let mut sample_proxy =
+            iroha::samples::get_config_proxy(peer_ids.iter().cloned().collect(), Some(keypair.clone()));
+        let env_proxy = ConfigurationProxy::from_env();
+        
+        sample_proxy.override_with(env_proxy)
+                    .build()
+                    .expect("Test Iroha config failed to build. This is either a programmer error or a compiler bug.")
```
Arjentix, line 171: Not sure it worth mentioning that it could be a compiler bug. At the end we don't write it in other panic messages. Have we ever encountered such a bug?

core/tests/mod.rs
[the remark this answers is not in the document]
Arjentix: Same in other places

core/tests/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+    let keypairs : Vec<KeyPair> = (0..7).into_iter().map(|n| KeyPair::generate().unwrap()).collect();
+    
+    let peer_ids : Vec<PeerId> = keypairs.iter().enumerate().map(|(i, kp)| PeerId::new(&format!("localhost:{}", i).parse().unwrap(), kp.public_key())).collect();
+
+    let mut sumeragis : HashMap::<PeerId, SumeragiHandle> = HashMap::new();
+    let mut configurations : HashMap::<PeerId, Configuration> = HashMap::new();
+    
+    for (index, (keypair, peer_id)) in keypairs.iter().zip(peer_ids.iter()).enumerate() {
+
+    let mut configuration = iroha::samples::get_config_proxy(peer_ids.iter().cloned().collect(), Some(keypair.clone())).build().unwrap();
```
Arjentix, line 261: It's hard to read without spaces in loops

core/tests/mod.rs
Arjentix, line 1: I understand it's WIP, but I'd consider splitting this file into functions to improve readability

core/tests/mod.rs
appetrosyan: Formatting. @SamHSmith if you're using my latest emacs, `apheleia` mode handles the formatting on save.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
SamHSmith: IrohaNetwork is a defined interface yes. But it only allows async api. Sync api is required for the softfork test. I will make sure closures don't leak into other crates.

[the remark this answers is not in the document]
Erigara: > But it only allows async api 

What do you mean? Most of the `NetworkBaseHandle` are sync expect `strart` and `wait_for_online_peers`.

core/tests/mod.rs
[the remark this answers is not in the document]
SamHSmith: I copied this from the existing tests. I think aleksandr thought this useful for changing the loglevel in order to debug the tests. But I actually agree with you both and will remove these.

core/tests/mod.rs
SamHSmith: Not any more, switched to https://github.com/focus-editor/focus

Will format before I undraft.

core/src/sumeragi/mod.rs
```diff
@@ -54,13 +59,13 @@ struct LastUpdateMetricsData {
...
     public_wsv: Arc<Mutex<WorldStateView>>,
-    message_sender: mpsc::SyncSender<MessagePacket>,
     metrics: Metrics,
     last_update_metrics_mutex: Arc<Mutex<LastUpdateMetricsData>>,
-    network: IrohaNetwork,
+    get_connected_peers_procedure: SumeragiGetCurrentlyConnectedProcedure,
     kura: Arc<Kura>,
     queue: Arc<Queue>,
     _thread_handle: Arc<ThreadHandler>,
+    message_sender: mpsc::SyncSender<MessagePacket>, // This must be kept alive until the thread exits, ergo it's after _thread_handle.
```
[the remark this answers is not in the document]
appetrosyan: Actually, the better way to express this is by `impl Drop`. Just because `Arc::drop` was called doesn't mean the thread is stopped, just that the refcount got decremented.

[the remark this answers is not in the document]
appetrosyan: In fact instead of silently shutting off the `Sender`, it's a good idea to block until the refcount reaches zero and panic if it doesn't in a reasonable amount of time.

core/tests/mod.rs
appetrosyan: Well compiler bugs are not ucnommon, but they are hardly easily reproducible, so outright saying something is a compiler bug is just disrespectful of the `rustc` team.

core/tests/mod.rs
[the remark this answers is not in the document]
appetrosyan: For the record, there's a few things that this allows us to do. It's not just about changing the default log level, that can be controlled via a build param in the config proxy at compile time. 

I used this to fuzz configuration parameters. It's redundant now given #3490.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
SamHSmith: > And can you please elaborate in which way sync api is required for soft fork test?

I need to block the sumeragi thread in order to trigger the softfork at the right moment. When it calls the closure I can simply call thread sleep inside the closure.

core/src/sumeragi/mod.rs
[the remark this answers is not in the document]
SamHSmith: I disagree. There are no copies of the ThreadHandle arc outside of the SumeragiHandle. So there is a one to one relationship between the number of SumeragiHandle's, Arc<ThreadHandler> and SyncSenders.

core/tests/mod.rs
SamHSmith: I switched it to be lazy initialized.

core/tests/mod.rs
SamHSmith: Can this conversation be closed?

core/Cargo.toml
```diff
@@ -72,6 +72,8 @@ hex = "0.4.3"
 byte-unit = "4.0.18"
 once_cell = "1.16.0"
 tempfile = "3.3.0"
+iroha = { version = "=2.0.0-pre-rc.13", path = "../cli" }
```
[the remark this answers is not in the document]
SamHSmith: This is the only function used `iroha::samples::get_config_proxy`.

core/src/sumeragi/mod.rs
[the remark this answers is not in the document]
appetrosyan: In which case it's not supposed to be an Arc, which is an owning reference.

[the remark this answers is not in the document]
SamHSmith: I kind of agree with you. But it's not for this PR. I have opened an issue. https://github.com/hyperledger/iroha/issues/3584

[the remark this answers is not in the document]
appetrosyan: Splitting off into another issue is always good. 

For future reference, it's typically better to use "Reference in new issue", that way you don't go through the template, and refer to the conversation in-context, without having to add things manually.

appetrosyan (approved):

core/src/sumeragi/main_loop.rs
```diff
@@ -258,6 +250,11 @@ impl Sumeragi {
             "Genesis transaction set contains invalid transactions"
         );
 
+        assert!(
+            block.rejected_transactions.is_empty(),
```
Erigara, line 254: This code is duplicate from the above.

SamHSmith: Good catch. Git be gitting. I'll remove it.

core/Cargo.toml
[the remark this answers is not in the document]
appetrosyan: I'd consider extracting it into `core` and removing the dependency.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
Arjentix: So have you considered `mockall` crate? I'm just not a big fan of using dyn closures here. I think there should be a better way

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+use once_cell::sync::Lazy;
+use rand::{seq::IteratorRandom, SeedableRng};
+
+#[test]
+fn sumeragi_commits_and_broadcasts_genesis() {
+    init_telemetry();
+
+    let pb_hash = Arc::new(Mutex::new(None));
+    let pb_hash2 = pb_hash.clone();
+    /* Can't have 1 peer, then block is not sent. Can't have < 4 peers because non genesis peers get upset. */
```
Arjentix, line 29:
```suggestion
    // Can't have 1 peer, then block is not sent. Can't have < 4 peers because non genesis peers get upset
```

This form for me looks more modern and convinient with the rest of our code

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+            if let message::Message::BlockCreated(block) = packet.message {
+                let vb: VersionedCommittedBlock = block.block.commit_unchecked().into();
+                *pb_hash2.lock().unwrap() = Some(vb.hash());
+            }
+        }),
+    );
+    let sumeragis = sumeragis.lock().unwrap();
+    let sumeragi = sumeragis.get(&peer_ids[0]).unwrap();
+
+    for _ in 0..300 {
```
Arjentix, line 42: Please, make `300` a constant with meaningful name. Right now it's a magic number

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+        {
+            assert_eq!(
+                genesis_block.hash(),
+                sumeragi
+                    .apply_wsv(WorldStateView::clone)
+                    .block_hashes_after_hash(None)[0]
+            );
+            return;
+        }
+        sleep_ms(100);
```
Arjentix, line 128: Why do we need to wait actually? Can we refactor this to some event listenning? Maybe in another issue

core/tests/mod.rs
Arjentix: Same for other places where it's hard to understand what's going on. I.e. when `_` is used as a variable for iteration

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+
+    let (_keypairs, peer_ids, _configs, queues, sumeragis) = set_up_network(
+        7,
+        Arc::new(move |packet, peer_id, sumeragis| {
+            let sumeragis = sumeragis.lock().unwrap();
+            sumeragis.get(peer_id).unwrap().incoming_message(packet);
+        }),
+    );
+
+    for block_height in 1..10 {
```
Arjentix, line 145: Can you add some comments describing what's going on inside such a big blocks of code? I'd prefer splitting into functions, but I understand that it might be hard sometimes

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+                        if sumeragi.apply_wsv(|wsv| {
+                            wsv.block_hashes_after_hash(None).len() != block_height
+                        }) {
+                            success = false;
+                        }
+                    }
+                    if success {
+                        break;
+                    }
+                }
```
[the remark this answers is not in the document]
Arjentix: Also can this small loop be refactored with iterators? Like:

```rust
let success = sumeragis
    .values()
    .all(|sumeragi| {
         sumeragi.apply_wsv(|wsv| {
             wsv.block_hashes_after_hash(None).len() != block_height
         })
    });
```

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+            queue.push(transaction.clone().into(), &wsv).unwrap();
+        }
+    }
+}
+
+#[test]
+fn sumeragi_with_faults_commit_blocks() {
+    init_telemetry();
+
+    let peer_count = 7;
```
Arjentix, line 222: Better to make it `const`

core/tests/mod.rs
appetrosyan: As you can tell, following a different convention with `//` is not a terrible idea.

core/tests/mod.rs
appetrosyan: Personally don't mind it. Magic numbers especially in tests actually help with readability. 

**However**, What I would like instead, is for this function to be parametric (non-test), and that you could change the number of iterations at the call site inside of a test with a proper name. 

This would later allow us to convert the regular test into a `proptest` and handle unforeseen edgecases.

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+                    sumeragi
+                        .apply_wsv(WorldStateView::clone)
+                        .block_hashes_after_hash(None)[0]
+                );
+                return;
+            }
+        }
+        sleep_ms(100);
+    }
+    panic!("Sumeragi did not submit genesis.");
```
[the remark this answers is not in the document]
appetrosyan: UK vs. US approaches. The US convention is to put the full stop outside the quote like so: 

```
"This is reported speech". 
```

while the UK convention is 

```
"This is reported speech."
```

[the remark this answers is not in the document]
appetrosyan: With that said, @Arjentix is right, because Rust panic messages cater to the US convention. So no full stop in either panic or error messages.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
mversic: we should definitely explore better ways as part of this PR. I'm not fan of `mockall` because it allows data races if you're not careful of how you're using it's API. One of the reasons we use Rust is precisely to not have data races. We can, however, replicate the compile time injection trick it uses

There is [this](https://github.com/ryo33/mry) interesting testing framework, but I haven't used it so I can't tell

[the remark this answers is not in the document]
appetrosyan: I'm afraid Mry is worse than `mockall` for reliability.

core/tests/mod.rs
appetrosyan: It's more descriptive, but not better IMO. 

The compiler already knows that this is a value binding. Everything is immutable, and no enclosing scope is present, so it's effectively a compile-time constant.

core/tests/mod.rs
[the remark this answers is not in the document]
appetrosyan: +1 for iterators in general.

core/tests/mod.rs
appetrosyan: On that note, most tests need extra comments. 

You have to describe a strict set of preconditions with an example, a strict set of operations, and a strict set of postconditions. With something as complex as this, you might need to also elaborate on specific pre-conditions.

appetrosyan: What's concerning here is that Rust cannot differentiate `test` function code from the code that it's covering. You're pushing the coverage up, but you're not covering the code that's in use. Given such a complex test, it looks to me that you're effectively mocking what you should be testing instead. 

Consider if we can remove this in favour of #3490 ?

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+fn sleep_ms(t: u64) {
+    std::thread::sleep(std::time::Duration::from_millis(t));
+}
+
+/// Trait used to differentiate a test instance of `genesis`.
+pub trait TestGenesis: Sized {
+    /// Construct Iroha genesis network and optionally submit genesis
+    /// from the given peer.
+    fn test(cfg: Configuration, submit_genesis: bool) -> Option<Self>;
+}
```
appetrosyan, line 582: This reminds me of code that we removed last year. I still think that we ought to avoid mocking and instead try to integration test what we can. 

@Erigara is right in that the scope of the current tests is bigger than unit tests, but smaller than integration. This lies in the uncanny valley of placebo testing: it's not utilising the full code paths and requires a lot of effort to maintain, but it pushes coverage up significantly. 

If we can't test core without resorting to such tricks, then maybe we ought to refine our work on proper integration and functional tests.

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+
+#[allow(clippy::needless_pass_by_value, clippy::type_complexity)]
+fn set_up_network_with_genesis(
+    peer_count: usize,
+    commit_genesis: bool,
+    post_proc: Arc<
+        (dyn Fn(MessagePacket, &PeerId, &Arc<Mutex<HashMap<PeerId, SumeragiHandle>>>)
+             + Send
+             + Sync),
```
appetrosyan, line 452: No. This is not how you do it. 

1. You should not have an atomically reference counted function pointer. EVER. The functional approach is to take this by reference. 
2. I don't think a function pointer is a good choice here. See https://stackoverflow.com/questions/56420597/should-i-pass-function-objects-by-value-or-by-reference. You are actively making the compiler do more work. The end result is very likely the same thanks to LLVM's devirtualiser. 
3. And this more contentious; the type that holds this function value, should itself be generic. It can make things a bit uglier, and before GATs far uglier than they would now. Still that's more clear with the intent. 
4. This should really be a `struct -> struct` function or better yet `impl From<ArgStruct> for ReturnStruct`. Rust has robust destructuring, the only way a tuple is different from a struct is in that it's ambiguous about names, less verbose and far less error-prone.

core/tests/mod.rs
```diff
@@ -0,0 +1,646 @@
...
+                    sleep_ms(20000);
+                }
+            }
+
+            let sumeragis = sumeragis.lock().unwrap();
+            sumeragis.get(peer_id).unwrap().incoming_message(packet);
+        }),
+    );
+
+    for block_height in 1..5 {
```
appetrosyan, line 351: I don't mind this not being a separate function, though you could extract a few things here as parameters. 

There's a lot of mutable state here though, which hurts readability. Loads of back-references. Can we try and do something about that?

Arjentix: +1 for removing mutable variables

SamHSmith:
This is not the direction we have decided to go with regards to testing.

PR #3563, opened 2023-06-05
[feature] #3088: Introduce queue throttling

### LINKED ISSUE #3088: `SignatureCheckCondition` can block transaction queue
At the moment, `SignatureCheckCondition` can evaluate to `true` or `false`.

#### Ideally, `SignatureCheckCondition` should evaluate to one of three states:
* passed
* failed (and waiting for additional requirements to be met on a transaction)
* impossible (no matter how long the peer waits this transaction can never pass)

Considering that we don't have the `impossible` state in the current implementation of condition check, it is possible for a malicious user to write such a condition check that can never be fulfilled and flood the Iroha queue with transactions consequently blocking the queue. 

@Erigara in #3043 was aware of this so he implemented a two queue solution. There is one queue for transaction that have passed signature check (are valid and can be added to a block). The other queue is for transactions that have not passed signature check. 1st queue cannot be affected by this problem, but the 2nd can. 

**Even if `SignatureCheckCondition` was such that it had 3 states, the problem is that the malicious user can still write an incorrect condition and we're back to square one**

#### Some of possible solutions would be:
1. formally verify the user given signature check is correct (I don't believe this is possible?)
2. simulate transaction execution to verify that it signature check condition can actually be fulfilled
3. don't give user the possibility to write any sort of signature check, rather offer a set of predefined signature checks that they can choose from and mint for their account

I believe 3rd is the most easy that we can offer in the short term and I don't believe that we need the full flexibility of custom defined signature check conditions. A set of predefined might be sufficient? How do we use the signature check except for supporting MSTs?

Also, if this problem is fixed, we can go back to using just one queue for transactions in Iroha. Two queues were the temporary preferred solution because of this problem

Erigara:
## Description

Instead of introducing separate queue for MST transaction and double required space for queue. 
It was decided to approach problem from different angle and restrict amount of transactions user can have pending in the queue.

### Linked issue

Closes #3088 

### Benefits

Single user can't flood transaction queue.

### Checklist


config/src/queue.rs
```diff
@@ -4,7 +4,7 @@ use iroha_config_base::derive::{Documented, Proxy};
 use serde::{Deserialize, Serialize};
 
 const DEFAULT_MAX_TRANSACTIONS_IN_QUEUE: u32 = 2_u32.pow(16);
-const DEFAULT_MAX_TRANSACTIONS_IN_SIGNATURE_BUFFER: u32 = 2_u32.pow(16);
+const DEFAULT_MAX_TRANSACTIONS_IN_QUEUE_PER_USER: u32 = 2_u32.pow(16);
```
Erigara, line 7: By setting this option to be equal to `DEFAULT_MAX_TRANSACTIONS_IN_QUEUE` we preserve old behavior. 
Not to break tests and benchmarks which can rely on a lot of transactions submitted by single user.

core/src/queue.rs
```diff
@@ -479,166 +469,6 @@ mod tests {
         ));
     }
 
-    #[test]
-    fn push_tx_when_signature_buffer_is_full() {
```
Erigara, line 483: This test became meaningless and thus removed.

core/src/queue.rs
```diff
@@ -27,16 +27,16 @@ use crate::{prelude::*, tx::CheckSignatureCondition as _};
...
-    signature_buffer: ArrayQueue<HashOf<VersionedSignedTransaction>>,
-    /// [`VersionedAcceptedTransaction`]s addressed by `Hash`.
+    /// [`VersionedAcceptedTransaction`]s addressed by `Hash`
     txs: DashMap<HashOf<VersionedSignedTransaction>, VersionedAcceptedTransaction>,
+    /// Amount of transactions per user in the queue
+    txs_per_user: DashMap<AccountId, usize>,
     /// The maximum number of transactions in the queue
     max_txs: usize,
+    /// The maximum number of transactions in the queue per user. Used to apply throttling
+    max_txs_per_user: usize,
```
Erigara, line 39: Alternatively we can use `Option<NonZeroUsize>`.
And skip user transaction number check if `max_txs_per_user` is `None`

mversic: I don't think this should ever be skipped. It's more appropriate the way you've done it

appetrosyan: I'd prefer that.

core/src/queue.rs
```diff
@@ -322,46 +294,64 @@ impl Queue {
...
+                *occupied.get_mut() += 1;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn decrease_per_user_tx_count(&self, account_id: &AccountId) {
+        let Entry::Occupied(mut occupied) = self.txs_per_user
+            .entry(account_id.clone()) else { panic!("Call to decrease always should be paired with increase count. This is a bug.") };
```
mversic, line 347:
```suggestion
        let mut occupied = self.txs_per_user[account_id.clone()];
```

should be as good + removes runtime check

appetrosyan (approved):

core/src/queue.rs
Erigara: Unfortunately `DashMap` doesn't implement `IndexMut` trait.

mversic (approved):

appetrosyan (approved):

PR #3571, opened 2023-06-06
[fix] #3546: Fix time trigger tests get stuck

### LINKED ISSUE #3546: [bug]: the test `pre_commit_trigger_should_be_executed` get stuck sometimes

Erigara:
## Description

Change order in which we submit first transaction and start listening for events.

### Linked issue

Closes #3546 

### Benefits

Test doesn't get stuck.

### Checklist

### DIFF
--- a/client/tests/integration/triggers/time_trigger.rs
+++ b/client/tests/integration/triggers/time_trigger.rs
@@ -21,7 +21,6 @@ macro_rules! const_assert {
 /// so it's impossible to create fully reproducible test scenario.
 ///
 /// But in general it works well and this test demonstrates it
-#[ignore = "ignore, more in #2851"]
 #[test]
 #[allow(clippy::cast_precision_loss)]
 fn time_trigger_execution_count_error_should_be_less_than_15_percent() -> Result<()> {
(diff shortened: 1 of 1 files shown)


pesterev (approved):

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -34,6 +34,9 @@ fn time_trigger_execution_count_error_should_be_less_than_15_percent() -> Result
     wait_for_genesis_committed(&vec![test_client.clone()], 0);
     let start_time = current_time();
 
+    // Start listening BEFORE submitting any transaction not to miss any block committed event
+    let event_listener = get_block_committed_event_listener(&test_client)?;
```
Arjentix, line 38: So this test is ignored, but you fixed it too? Interesting. Mabye we should unignore it if it works fine now?

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -178,6 +181,9 @@ fn mint_nft_for_every_user_every_1_sec() -> Result<()> {
     let (_rt, _peer, mut test_client) = <PeerBuilder>::new().with_port(10_780).start_with_runtime();
     wait_for_genesis_committed(&vec![test_client.clone()], 0);
 
+    // Start listening BEFORE submitting any transaction not to miss any block committed event
+    let event_listener = get_block_committed_event_listener(&test_client)?;
```
Arjentix, line 185: Shouldn't we create this just before the last `submit()` before submitting sample instructions?

I mean shouldn't we put this before line 240 (in current diff)

client/tests/integration/triggers/time_trigger.rs
appetrosyan: The whole point is to unignore these tests. We only remove the tests that are redundant after the Pytest 
sdk was merged

client/tests/integration/triggers/time_trigger.rs
Erigara: For what? there is no transactions submitted between this two points and iroha don't produce empty blocks so effect should be the same

Erigara: Yes, indeed

pesterev (approved):

Issue #3535, opened 2023-05-26
[suggestion] Establish the list of magic types in the schema


0x009922:
### Feature request

The data-model schema has a set of types that aren't specified in the schema and are kind-of a part of standard preset. This preset is not established or mentioned anywhere. Currently, it includes such primitives like `Bool`, `String`, `u8`, `u16` etc. 

I ask to define the set of standard types explicitly. I propose to use SCALE primitives as a reference:

- `i8`
- `u8`
- `i16`
- `u16`
- `i32`
- `u32`
- `i64`
- `u64`
- `i128`
- `u128`
- `compact`
- `string`
- `bool`
- `()` (or `unit`, zero-size type)

These types might be mentioned somewhere in the schema or its documentation, and should be considered by codegenerators as magic types that should not be looked up in the schema. For better explicitness, these types might be represented as schema with a prefix, like `std::u8`,  or `scale::bool`, or `core::string`.

### Motivation

Remove yet another seamless convention from the schema and make its usage clearer for SDK developers and related tooling.

### Who can help?

@mversic, @Erigara

appetrosyan:
@arndey I think it'd be important if you pitch into this discussion

arndey:
> ### Feature request
> The data-model schema has a set of types that aren't specified in the schema and are kind-of a part of standard preset. This preset is not established or mentioned anywhere. Currently, it includes such primitives like `Bool`, `String`, `u8`, `u16` etc.
> 
> I ask to define the set of standard types explicitly. I propose to use [SCALE primitives](https://docs.substrate.io/reference/scale-codec/) as a reference:
> 
> * `i8`
> * `u8`
> * `i16`
> * `u16`
> * `i32`
> * `u32`
> * `i64`
> * `u64`
> * `i128`
> * `u128`
> * `compact`
> * `string`
> * `bool`
> * `()` (or `unit`, zero-size type)
> 
> These types might be mentioned somewhere in the schema or its documentation, and should be considered by codegenerators as magic types that should not be looked up in the schema. For better explicitness, these types might be represented as schema with a prefix, like `std::u8`, or `scale::bool`, or `core::string`.
> 
> ### Motivation
> Remove yet another seamless convention from the schema and make its usage clearer for SDK developers and related tooling.
> 
> ### Who can help?
> @mversic, @Erigara

We (Java/Kotlin SDK team) can be flexible in this regard

Mingela:
[...]

PR #3572, opened 2023-06-06
[fix] #3162: Forbid 0 height in block streaming request

### LINKED ISSUE #3162: [BUG] Block Stream doesn't emit blocks when requested height is 0
### OS and Environment

macOS 13.1 (22C65)

### GIT commit hash

fe826be8

- Save the JS file into `bundle.js` - https://pastebin.com/NZ0a8wFV
  It is a dependency-free JavaScript bundle with rolled up `iroha-javascript` code responsible for handling Block Stream API and a simple workflow to reproduce the bug at the very end of the bundle.
- Run Iroha in any way you like it.
- Run `bundle.js` with Node.js 16/18:
  ```shell
  DEBUG=@iroha2* node bundle.js \
    // optionally pass Torii API URL
    http://localhost:8080
  ```
- See "Block stream is ready" message in console
- Commit some block into Iroha and see what is printed from the script

### Actual result

Nothing is printed - Iroha doesn't send any messages via WebSocket.

### Expected result

I expected Iroha to send committed blocks.

### Logs in JSON format

I can't provide you with any useful logs.

_No response_

### Notes

_No response_

Erigara:
## Description

Use `NonZeroU64` for block streaming to prevent user from sending 0 as block height.

### Linked issue

Closes #3162 

### Benefits

Clear error.

### Checklist


Arjentix:
> Use `NonZeroU62`

Did you mean `NonZeroU64`?

schema/gen/src/lib.rs
```diff
@@ -421,10 +421,12 @@ types!(
     u32,
     u64,
     u8,
+    NonZeroU64,
```
Arjentix, line 424: Probably schema should have been updated after this change?

schema/src/lib.rs
```diff
@@ -262,6 +265,52 @@ macro_rules! impl_schema_int {
 
 impl_schema_int!(u128, u64, u32, u16, u8, i128, i64, i32, i16, i8);
 
+impl TypeId for NonZeroU32 {
```
Arjentix, line 268: So these impls for `NonZeroU32` are useless for now, right? Because there are no `NonZeroU32` in schema

mversic: Not useless. The implementation is such that NonZeroU32/NonZeroU64 are mapped into u32/u64 respectively. That is why there is no change in the schema.json

data_model/src/block.rs
```diff
@@ -399,13 +399,15 @@ pub mod stream {
...
+        use core::num::NonZeroU64;
+
         use super::*;
 
         /// Request sent to subscribe to blocks stream starting from the given height.
         #[version_with_scale(n = 1, versioned = "VersionedBlockSubscriptionRequest")]
         #[derive(Debug, Clone, Copy, Constructor, Decode, Encode, IntoSchema)]
         #[repr(transparent)]
-        pub struct BlockSubscriptionRequest(pub u64);
+        pub struct BlockSubscriptionRequest(pub NonZeroU64);
```
0x009922, line 410: Such a change should be reflected in the generated schema, so that SDKs have a way to handle it specifically.

Also, such a type is a part of the "magic types" list: #3535

0x009922 (changes requested):

data_model/src/block.rs
mversic: > Such a change should be reflected in the generated schema, so that SDKs have a way to handle it specifically.

I would change the phrasing to `could` instead of `should`.

I'm not sure I entirely agree here. There is a point to be made about simplicity. Do we really want to overload schema with handling niche values for types? If sdk devs want to they still can do so even if it is not specified in the schema. Some implicit knowledge might be fine

On the other hand do we want to consider introducing some sort of type metadata in the schema? This info could then be embedded in the schema's metadata

mversic: So rather than having a distinct type for every non-robust version of the container type we could rather decorate the type with metadata

mversic: I take back what I've said. Look [here](https://github.com/hyperledger/iroha/issues/3535#issuecomment-1579905464). All information has to be in the type name, we can't use type metadata. So if we want this we need a special type

cli/src/torii/routing.rs
```diff
@@ -317,14 +317,14 @@ async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::R
...
-                if let Some(block) = kura.get_block_by_height(from_height) {
+                if let Some(block) = kura.get_block_by_height(from_height.get()) {
                     stream
                         // TODO: to avoid clone `VersionedBlockMessage` could be split into sending and receiving parts
                         .send(VersionedBlockMessage::from(
                             BlockMessage(VersionedCommittedBlock::clone(&block)),
                         ))
                         .await?;
-                    from_height += 1;
+                    from_height = from_height.saturating_add(1);
```
appetrosyan, line 327: Are you sure saturation is the right choice here? I'd rather this were a panic.

tools/parity_scale_decoder/src/main.rs
```diff
@@ -6,6 +6,7 @@
 )]
 #![allow(clippy::print_stdout, clippy::use_debug, clippy::unnecessary_wraps)]
 
+use core::num::NonZeroU64;
```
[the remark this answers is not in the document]
Erigara: This is required due to expansion of `generate_map!` which is defined inside `types!` macro.

schema/gen/src/lib.rs
```diff
@@ -421,10 +421,12 @@ types!(
...
 );
 
 #[cfg(test)]
 mod tests {
+    use core::num::NonZeroU64;
```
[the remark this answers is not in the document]
Erigara: This is required due to expansion of `generate_map!` which is defined inside `types!` macro.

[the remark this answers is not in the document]
Erigara: Check response for your similar comment.

schema/src/lib.rs
Erigara: I decided to add implementation for `NonZeroU32` in one go with `NonZeroU64` but it's not used in the schema at the moment.

data_model/src/block.rs
Erigara: @0x009922 would you mind if we currently leave `NonZeroU64` represenation as is and left a comment about this magic schema types ticket for further discussion? 

From my point of view adding such types into the schema wouldn't add much value. Since iroha side would need to check that received value is indeed non-zero. 

So usage of `NonZeroU64` is more like safe-guard for iroha rather than distinct type.

What i'm trying to say that sending height of zero is logical error, not type error.

appetrosyan (approved):

Arjentix (approved):

0x009922 (approved):

Issue #3230, opened 2023-02-21
Permission Tokens analysis


Arjentix:
Runtime validator contains all permission tokens it checks. We need a way to get all this tokens from it and use it in our genesis (`kagami genesis` subcommand). Right now all tokens are hardcoded, which is not good.

PR #3588, opened 2023-06-08
[feature] #3587: Allow different states in wasm::Runtime & link-time …

### LINKED ISSUE #3587: [feature] Allow different states in `wasm::Runtime` & link-time checking
Initially it's a sub-task for #3230 .

# Problems

The problem is that in #3230 I need to introduce another entypoint which should be executed with data different from our current `State`. For example there will be no `account_id`, but current desing forces me to use it.

Also with current design we skip link-time checking. We always link all functions. But for example `query_triggering_event()` function shouldn't be executed from *Smart Contract* or *Validator*, it should be executed only from *Trigger*.

# Solution

Make `Runtime` generic over state it uses. This will allow split current big `State` into different types and use different `Linker`'s for `Runtime` to implement link-time checking.

# How to test from outside world

When this will be implemented there will be no way to execute inappropriate functions from WASM.

Things that should fail:

1. Executing `query_triggering_event()` from *Smart Contract* or *Validator*
2. Execution `query_operation_to_validate()` from *Smart Contract* or *Trigger*

Arjentix:
## Description

- Added support for different state types into `Runtime`
- Added support for `state` attribute into `iroha_wasm_codec_derive`, so that there will be a way to specify state type when it's unable to infer it from parameters
- Added different `wasmtime::Linker` configurations for different states

### Linked issue

* Closes #3587  

### Benefits

- Ability to expand state types which I'm blocked with in #3230 
- Link-time checking for inappropriate function calls from WASM

See issue description for more details.

### Checklist


core/src/tx.rs
```diff
@@ -250,7 +250,7 @@ impl TransactionValidator {
     ) -> Result<(), TransactionRejectionReason> {
         debug!("Validating wasm");
 
-        wasm::RuntimeBuilder::new()
+        wasm::RuntimeBuilder::<wasm::state::SmartContract>::new()
```
appetrosyan, line 253: What happened here?

Arjentix: Now `RuntimeBuilder` is generic over state which will be used in runtime. Here I specify that we need a `Runtime` for *Smart Contract* executio

Arjentix: I can make another API for `RuntimeBuilder`. For example something like this:

```rust
struct RuntimeBuilder;

struct RuntimeBuilderWithState<S> {
    // ...
}

impl RuntimeBuilder {
    pub fn smart_contract() -> RuntimeBuilderWithState<state::SmartContract> {
        // ...
    }
}
```

wasm_codec/derive/src/lib.rs
```diff
@@ -99,13 +143,32 @@ fn gen_output(
     }
 }
 
+/// [`TokenStream2`] wrapper which will be lazily evaluated
+///
+/// Implements [`quote::ToTokens`] trait
+struct LazyTokenStream<F>(once_cell::unsync::Lazy<TokenStream2, F>);
```
Erigara, line 149: Can you explain why do we need this?

Erigara (approved):

wasm_codec/derive/src/lib.rs
Arjentix: It's a workaround to keep the code in `gen_body()` more clear. I wrap `memory` and `alloc` retrieving into this structure. But this retrieving requires me to know the *state* type. And I don't need `memory` and `alloc` in all cases.

So in some cases I don't need any of them and this lazy code will never be evaluated. This gives us:
1. A little bit faster compliation (probably a very little decrease)
2. Checking for *state* type presence only when we need it (this is mainly why I did it)

core/src/smartcontracts/wasm.rs
```diff
@@ -746,12 +811,104 @@ impl RuntimeBuilder {
...
+            let mut linker = Linker::new(engine);
+
+            create_imports!(linker,
+                import::EXECUTE_ISI_FN_NAME => execute_instruction_as_validator,
+                import::EXECUTE_QUERY_FN_NAME => execute_query_as_validator,
+                import::QUERY_AUTHORITY_FN_NAME => query_authority,
+                import::QUERY_OPERATION_TO_VALIDATE_FN_NAME => query_operation_to_validate,
+                import::QUERY_MAX_LOG_LEVEL => query_max_log_level,
+                import::LOG_FN_NAME => log,
+                import::DBG_FN_NAME => dbg,
```
mversic, line 900:
```suggestion
                import::EXECUTE_ISI_FN_NAME => execute_instruction_as_validator,
                import::EXECUTE_QUERY_FN_NAME => execute_query_as_validator,
                import::QUERY_OPERATION_TO_VALIDATE_FN_NAME => query_operation_to_validate,
                import::QUERY_MAX_LOG_LEVEL => query_max_log_level,
```

these should always be defined, so you can just hardcode them in the macro

core/src/smartcontracts/wasm.rs
```diff
@@ -629,111 +726,79 @@ impl<'wrld> Runtime<'wrld> {
...
-    fn execute_smart_contract_with_state(
-        &mut self,
-        bytes: impl AsRef<[u8]>,
-        state: State,
-    ) -> Result<()> {
-        let mut store = self.create_store(state);
-        let smart_contract = self.create_smart_contract(&mut store, bytes)?;
+    #[allow(clippy::needless_pass_by_value)]
+    #[codec::wrap]
+    fn execute_query_as_validator(
```
mversic, line 731:
```suggestion
    fn execute_query(
```

why not just this? it's implicit in the type

core/src/smartcontracts/wasm.rs
```diff
@@ -629,111 +726,79 @@ impl<'wrld> Runtime<'wrld> {
...
+        query.execute(state.common_mut().wsv).map_err(Into::into)
     }
 
-    fn execute_main_with_store(
-        instance: &wasmtime::Instance,
-        store: &mut wasmtime::Store<State>,
-    ) -> Result<()> {
-        let main_fn = Self::get_typed_func(instance, store, export::WASM_MAIN_FN_NAME)?;
+    #[codec::wrap]
+    fn execute_instruction_as_validator(
```
mversic, line 741:
```suggestion
    fn execute_instruction(
```

it's implicit in the type

core/src/smartcontracts/wasm.rs
```diff
@@ -246,115 +263,246 @@ impl Validator {
...
-        let wsv: &mut WorldStateView = state.wsv;
-        let called_from_validator = state.operation_to_validate.is_some();
-        if called_from_validator {
-            // NOTE: Validator has already validated the query
-        } else {
-            // NOTE: Smart contract (not validator) is trying to execute the query, validate it first
-            // TODO: Validation should be skipped when executing smart contract.
-            // There should be two steps validation and execution. First smart contract
-            // is validated and then it's executed. Here it's validating in both steps.
-            // Add a flag indicating whether smart contract is being validated or executed
```
mversic, line 353: isn't this comment still valid? please restore it

core/src/smartcontracts/wasm.rs
mversic: I wasn't referring to those functions

core/src/smartcontracts/wasm.rs
Arjentix: I've tried this, but it doesn't work. The reason is that original `execute_query` is defined for generic `Runtime` so there will be a conflict.

In other words technically for `Runtime<state::Validator>` exist both `execute_query` and `execute_query_as_validator`.

core/src/smartcontracts/wasm.rs
Arjentix: See the response for `execute_query` above

core/src/smartcontracts/wasm.rs
mversic: > In other words technically for Runtime<state::Validator> exist both execute_query and execute_query_as_validator.

please don't do that. Maybe you can define a trait with default implementation. If not, you can have 3 `execute_query` methods defined on `Runtime<state::Trigger>`, `Runtime<state::Validator>`, `Runtime<state::SmartContract>`. If you're worried about code duplication you can define a freestanding method that you delegate to from `Runtime<state::Trigger>`, `Runtime<state::SmartContract>`

core/src/tx.rs
appetrosyan: I'd add it to another PR.

core/src/smartcontracts/wasm.rs
```diff
@@ -246,115 +263,246 @@ impl Validator {
...
+        fn common_mut(&mut self) -> &mut Common<'wrld>;
+
+        /// Get mutable reference to store limits
+        ///
+        /// # Note
+        ///
+        /// This is a workaround for lifetime issues when using [`StoreLimits`]
+        /// inside of [`wasmtime::Store::limiter()`] where we can't use
+        /// `&mut s.common_mut().store_limits`
+        fn limits_mut(&mut self) -> &mut StoreLimits;
```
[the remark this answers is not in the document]
appetrosyan: Honestly, it's not too ugly.

data_model/src/transaction.rs
```diff
@@ -428,6 +428,7 @@ mod candidate {
     }
 
     impl SignedTransactionCandidate {
+        #[cfg_attr(not(feature = "std"), allow(unused_mut))]
```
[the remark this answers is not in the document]
mversic: also the deserialization is straight out incorrect in `no_std`

crypto/src/merkle.rs
```diff
@@ -13,7 +13,9 @@ use std::collections::VecDeque;
 use iroha_schema::{IntoSchema, TypeId};
 use parity_scale_codec::{Decode, Encode};
 
-use crate::{Hash, HashOf};
+#[cfg(feature = "std")]
+use crate::Hash;
```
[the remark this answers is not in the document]
Arjentix: Did `cargo check-all-features` and `cargo test-all-features`. Fixed errors

core/src/smartcontracts/wasm.rs
Arjentix: Done

data_model/src/transaction.rs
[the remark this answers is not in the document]
mversic: > What do you mean here?

it's possible to create `SignedTransaction` that has invalid signature. Which should be disallowed at the type system. In `no_std` `SignedTransaction` should not implement `Deserialize` at all

mversic (approved):

data_model/src/transaction.rs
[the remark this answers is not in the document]
Arjentix: Got it, let's move it to another task, cause it may be not very trivial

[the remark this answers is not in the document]
Arjentix: Created #3607

Erigara (approved):

PR #3600, opened 2023-06-12
[fix] #3246: Don't involve observing peer without fault.

SamHSmith:
## Description

Don't involve observing peers in consensus when view_change_index == 0.

### Linked issue

#3246

I have verified that this PR has the intended effect by observing the output of the unstable network tests.

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -75,6 +75,9 @@ impl Sumeragi {
     #[instrument(skip(self, packet))]
     #[allow(clippy::needless_pass_by_value)] // TODO: Fix.
     fn post_packet_to(&self, packet: MessagePacket, peer: &PeerId) {
+        if peer == &self.peer_id {
+            return;
+        }
         let post = iroha_p2p::Post {
             data: NetworkMessage::SumeragiPacket(Box::new(packet.into())),
             peer_id: peer.clone(),
@@ -536,8 +539,10 @@ fn handle_message(
 
                     sumeragi.broadcast_packet_to(msg, [current_topology.proxy_tail()]);
                     info!(%addr, %block_hash, "Block validated, signed and forwarded");
+                    *voting_block = Some(v_block);
+                } else {
+                    error!(%addr, %role, "Received BlockCreated message, but shouldn't");
                 }
-                *voting_block = Some(v_block);
             }
         }
         (Message::BlockCreated(block_created), Role::ProxyTail) => {
@@ -580,6 +585,7 @@ fn handle_message(
     }
 }
 
+#[allow(clippy::too_many_lines)]
 fn process_message_independent(
     sumeragi: &mut Sumeragi,
     voting_block: &mut Option<VotingBlock>,
@@ -624,7 +630,17 @@ fn process_message_independent(
                             view_change_proof_chain.clone(),
                             BlockCreated::from(new_block),
                         );
-                        sumeragi.broadcast_packet(msg);
+                        if current_view_change_index >= 1 {
+                            sumeragi.broadcast_packet(msg);
+                        } else {
+                            sumeragi.broadcast_packet_to(
+                                msg,
+                                current_topology
+                                    .sorted_peers
+                                    .iter()
+                                    .take(current_topology.min_votes_for_commit()),
+                            );
+                        }
                     } else {
                         match new_block.commit(current_topology) {
                             Ok(committed_block) => {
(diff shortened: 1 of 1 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -662,13 +676,33 @@ fn process_message_independent(
...
                             sumeragi.broadcast_packet(msg);
+                        } else {
+                            sumeragi.broadcast_packet_to(
+                                msg,
+                                current_topology
+                                    .sorted_peers
+                                    .iter()
+                                    .zip(0..current_topology.min_votes_for_commit())
+                                    .map(|(p, _)| p),
+                            );
```
Erigara, line 689: This code will try to send message to the peer itself. We should skip sender.

Erigara:
Are we okay we repercussions this solution has?

As i see it this will increase latency because observing peers will have to receive latest blocks through block sync.

And observing peers currently aren't involved in the consensus per say, proxy tail doesn't wait for them so we just send block for them and move on.

I think we should have at least some arguments why we want or don't want to have this cnange.

core/src/sumeragi/main_loop.rs
SamHSmith: I have added a check in the post_packet_to function.

SamHSmith:
Makoto has said that we should do it this way. And I can cite the whitepaper: 
> The leader then sends the block to each node in set a. Each peer in set  then validates and signs the block, and sends it to the proxy tail.

Later on regarding failure cases of the A set.

> * a peer may not sign off on a block
>     * if the lack of a signature causes a block to not commit, a new node will be brought in from set b

Erigara (approved):

mversic:
> Makoto has said that we should do it this way. And I can cite the whitepaper:
> 
> > The leader then sends the block to each node in set a. Each peer in set  then validates and signs the block, and sends it to the proxy tail.

This states that leader should not send block to observing peers. Arguably, proxy tail should still send block to all peers, not just set a. Do you think that waiting for block sync would noticeably degrade performance?

SamHSmith:
> This states that leader should not send block to observing peers. Arguably, proxy tail should still send block to all peers, not just set a. Do you think that waiting for block sync would noticeably degrade performance?

That's the thing though. The proxy tail does not send the block, it sends the partial block hash + signatures. It notifies the a set that they should commit the block.

Not involving the B set should improve performance. Block sync needs to be able to operate faster than consensus anyway, because otherwise a peer who has been offline could never catch up. The B set catching up through block sync distributes the load better, since all of the A set can be sources for block sync.

There will probably be a latency increase if one of the A set peers is faulty, because B set might not be exactly up to date. But, that's why we then move that peer to the B set so that things can continue smoothly.

core/src/sumeragi/main_loop.rs
```diff
@@ -75,6 +75,9 @@ impl Sumeragi {
     #[instrument(skip(self, packet))]
     #[allow(clippy::needless_pass_by_value)] // TODO: Fix.
     fn post_packet_to(&self, packet: MessagePacket, peer: &PeerId) {
+        if peer == &self.peer_id {
+            return;
+        }
```
appetrosyan, line 80: Took us long enough. Good catch.

core/src/sumeragi/main_loop.rs
```diff
@@ -536,8 +539,10 @@ fn handle_message(
 
                     sumeragi.broadcast_packet_to(msg, [current_topology.proxy_tail()]);
                     info!(%addr, %block_hash, "Block validated, signed and forwarded");
+                    *voting_block = Some(v_block);
+                } else {
+                    error!(%addr, %role, "Received BlockCreated message, but shouldn't");
```
appetrosyan, line 544:
```suggestion
                    error!(%peer=addr, %role, "Received `BlockCreated` message. See the topology on other peers to see if it's topology mismatch.");
```

appetrosyan (approved):
Looks alright. The latency increase might not scale, because this reduces the communication overhead and network load. We should not assume that things have changed, but measure. A 20% latency increase is fine, a $20\times$ is not.

Erigara:
> Looks alright. The latency increase might not scale, because this reduces the communication overhead and network load. We should not assume that things have changed, but measure. A 20% latency increase is fine, a 20× is not.

It was like decrease from 1300 to 1100 tps when i measured it (about 10% difference).
But i real world with larger amount of peers situation might be better. 
Also we can fine tune block sync frequency to decrease latency.

PR #3608, opened 2023-06-13
[fix] #3596: `iroha_client_cli` show blocks/events

### LINKED ISSUE #3596: [BUG]
### OS and Environment

MacOS ventura 13.3, DockerHub

### GIT commit hash

c82f2c09

the problem started to be observed from commit a8d62b51d79cb6e9f273b746bcc3530983730ae7

deploy iroha2-dev
deploy client_cli
./iroha_client_cli blocks 1

### Actual result

./iroha_client_cli blocks 1 -> does not display the current state of the blocks, the output in the terminal is empty

### Expected result

./iroha_client_cli blocks 1 -> displays the current state of the blocks

### Logs in JSON format

nothing todo

@timofeevmd

### Notes

_No response_

Erigara:
## Description

Pass closure inside subcomands to do printing. 

### Linked issue

Closes #3596 

### Benefits

Works correctly.

### Checklist

### DIFF
--- a/client_cli/src/main.rs
+++ b/client_cli/src/main.rs
@@ -14,7 +14,7 @@
 use std::{
     fmt,
     fs::{self, read as read_file},
-    io::stdin,
+    io::{stdin, stdout},
     str::FromStr,
     time::Duration,
 };
@@ -111,27 +111,55 @@ pub enum Subcommand {
     Json(json::Args),
 }
 
+/// Context inside which command is executed
+pub trait RunContext {
+    /// Get access to configuration
+    fn configuration(&self) -> &ClientConfiguration;
+    /// Serialize and print data
+    ///
+    /// # Errors
+    /// - if serialization fails
+    /// - if printing fails
+    fn print_data(&mut self, data: &dyn Serialize) -> Result<()>;
+}
+
+struct PrintJsonContext<W> {
+    write: W,
+    config: ClientConfiguration,
+}
+
+impl<W: std::io::Write> RunContext for PrintJsonContext<W> {
+    fn configuration(&self) -> &ClientConfiguration {
+        &self.config
+    }
+
+    fn print_data(&mut self, data: &dyn Serialize) -> Result<()> {
+        writeln!(&mut self.write, "{}", serde_json::to_string_pretty(data)?)?;
+        Ok(())
+    }
+}
+
 /// Runs subcommand
 pub trait RunArgs {
     /// Runs command
     ///
     /// # Errors
     /// if inner command errors
-    fn run(self, cfg: &ClientConfiguration) -> Result<Box<dyn Serialize>>;
+    fn run(self, context: &mut dyn RunContext) -> Result<()>;
 }
 
 macro_rules! match_all {
-    (($self:ident, $cfg:ident), { $($variants:path),* $(,)?}) => {
+    (($self:ident, $context:ident), { $($variants:path),* $(,)?}) => {
         match $self {
-            $($variants(variant) => RunArgs::run(variant, $cfg),)*
+            $($variants(variant) => RunArgs::run(variant, $context),)*
         }
     };
 }
 
 impl RunArgs for Subcommand {
-    fn run(self, cfg: &ClientConfiguration) -> Result<Box<dyn Serialize>> {
+    fn run(self, context: &mut dyn RunContext) -> Result<()> {
         use Subcommand::*;
-        match_all!((self, cfg), { Domain, Account, Asset, Peer, Events, Wasm, Blocks, Json })
+        match_all!((self, context), { Domain, Account, Asset, Peer, Events, Wasm, Blocks, Json })
     }
 }
 
@@ -174,9 +202,12 @@ fn main() -> Result<()> {
         );
     }
 
-    let subcommand_output = subcommand.run(&config)?;
-    println!("{}", serde_json::to_string_pretty(&subcommand_output)?);
-    Ok(())
+    let mut context = PrintJsonContext {
+        write: stdout(),
+        config,
+    };
+
+    subcommand.run(&mut context)
 }
 
 /// Submit instruction with metadata to network.
@@ -186,10 +217,10 @@ fn main() -> Result<()> {
 #[allow(clippy::shadow_unrelated)]
 pub fn submit(
     instructions: impl Into<Executable>,
-    cfg: &ClientConfiguration,
     metadata: UnlimitedMetadata,
-) -> Result<Box<dyn Serialize>> {
-    let iroha_client = Client::new(cfg)?;
+    context: &mut dyn RunContext,
+) -> Result<()> {
+    let iroha_client = Client::new(context.configuration())?;
     let instructions = instructions.into();
     #[cfg(debug_assertions)]
     let err_msg = format!("Failed to build transaction from instruction {instructions:?}");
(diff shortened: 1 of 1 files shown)


Arjentix (approved):
Not sure I'd did it in the same way, but it's fine for me

client_cli/src/main.rs
```diff
@@ -239,23 +250,31 @@ mod events {
...
+            self,
+            cfg: &ClientConfiguration,
+            callback: F,
+        ) -> Result<()> {
             let filter = match self {
                 Args::Pipeline => FilterBox::Pipeline(PipelineEventFilter::new()),
                 Args::Data => FilterBox::Data(DataEventFilter::AcceptAll),
             };
-            listen(filter, cfg)
+            listen(filter, cfg, callback)
```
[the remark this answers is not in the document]
Erigara: Yes, when you need to serialize and print value you call `callback` and pass your value inside.

client_cli/src/main.rs
```diff
@@ -117,21 +117,29 @@ pub trait RunArgs {
...
         match $self {
-            $($variants(variant) => RunArgs::run(variant, $cfg),)*
+            $($variants(variant) => RunArgs::run(variant, $cfg, $callback),)*
         }
     };
 }
 
 impl RunArgs for Subcommand {
-    fn run(self, cfg: &ClientConfiguration) -> Result<Box<dyn Serialize>> {
+    fn run<F: FnMut(&dyn Serialize) -> Result<()>>(
```
0x009922, line 136: `FnMut(&dyn Serialize) -> Result<()>` is repeated over and over and over again. Could you use an alias for that?

client_cli/src/main.rs
```diff
@@ -117,21 +117,29 @@ pub trait RunArgs {
     ///
     /// # Errors
     /// if inner command errors
-    fn run(self, cfg: &ClientConfiguration) -> Result<Box<dyn Serialize>>;
+    fn run<F: FnMut(&dyn Serialize) -> Result<()>>(
+        self,
+        cfg: &ClientConfiguration,
+        callback: F,
+    ) -> Result<()>;
```
0x009922, line 124: I don't like that each `RunArgs` implementor has to accept a magic `callback` and pass it explicitly to the final destination. Also, each time you have to specify the type of callback, which is verbose.

I think it would be better to introduce a trait like `RunContext`:

```rust
trait RunContext {
  fn configuration(&self) -> &ClientConfiguration;

  fn print_data(&self, value: impl Serialize) -> Result<()>;
}
```

That is something you can pass everywhere instead of `cfg` + `callback`. This approach is even less cost in terms of runtime (you don't need to pass 2 args, and you can simply `impl RunContext` with callback instead of passing a closure).

client_cli/src/main.rs
0x009922: I also don't understand why it should be `FnMut` and not `Fn`?

0x009922 (changes requested):

client_cli/src/main.rs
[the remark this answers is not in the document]
mversic: instead of passing in callback, could you make it a method on `RunArgs` trait?

mversic (dismissed):

client_cli/src/main.rs
0x009922: Runtime cost is definitely not an issue, of course. But I wonder if code organisation will be better with an umbrella trait.

Erigara: I will try to implement your approach. I think it will be look nicer that current one.

client_cli/src/main.rs
```diff
@@ -111,27 +111,55 @@ pub enum Subcommand {
...
+    write: W,
+    config: ClientConfiguration,
+}
+
+impl<W: std::io::Write> RunContext for PrintJsonContext<W> {
+    fn configuration(&self) -> &ClientConfiguration {
+        &self.config
+    }
+
+    fn print_data(&mut self, data: &dyn Serialize) -> Result<()> {
```
0x009922, line 136: Why it should be `dyn Serialize` and not `impl Serialize`?

client_cli/src/main.rs
```diff
@@ -111,27 +111,55 @@ pub enum Subcommand {
...
+}
+
 /// Runs subcommand
 pub trait RunArgs {
     /// Runs command
     ///
     /// # Errors
     /// if inner command errors
-    fn run(self, cfg: &ClientConfiguration) -> Result<Box<dyn Serialize>>;
+    fn run(self, context: &mut dyn RunContext) -> Result<()>;
```
0x009922, line 148: Same as with `dyn Serialize`: why not use just generics?

```suggestion
    fn run<C: RunContext>(self, context: &mut C) -> Result<()>;
```

client_cli/src/main.rs
```diff
@@ -186,10 +217,10 @@ fn main() -> Result<()> {
 #[allow(clippy::shadow_unrelated)]
 pub fn submit(
```
0x009922, line 218: Btw, it seems that `context` is used only within `submit` and `listen` functions. So, maybe you can define a "miniclient" struct that has `submit` and `listen` functions, and does the printing internally?

```rust
trait MiniClient {
  fn listen(&mut self, height: NonZeroU64) -> Result<()>;

  fn submit(&mut self, isi: impl Into<Executable>, metadata: UnlimitedMetadata) -> Result<()>;
}
```

`RunArgs` might work with `MiniClient` trait. The struct that implements `MiniClient` might contain both configuration and debugging logic, which `RunArgs` does not care about.

0x009922 (approved):
LGTM, although I got a new code organisation idea.

client_cli/src/main.rs
Erigara: It can be `impl Serialize` but in this case code size would be bigger because compiled code would have monomorphized version of function.

If in the future we would introduce another `context` like `PrintRONContext`. 
Size of code will be proportional to `n_contexts * m_serialized_objects`.

Atm difference is not that big through `9.2Mb` for generics vs `9.1Mb` for dyn object (`9.2Mb` vs `9.8Mb` if we would add `RON` serialization support).

client_cli/src/main.rs
Erigara: Answered in the comment above.

client_cli/src/main.rs
Erigara: I should notice that `context` is also used in queries + `listen` for blocks/events functions are not the same.

Erigara: Created #3611.

client_cli/src/main.rs
0x009922: I think that for the client the compiled binary size is more important than a tiny performance benefit we would get with `impl`. So yeah, I am fine with `dyn`.

client_cli/src/main.rs
```diff
@@ -111,27 +111,55 @@ pub enum Subcommand {
     Json(json::Args),
 }
 
+/// Context inside which command is executed
+pub trait RunContext {
+    /// Get access to configuration
+    fn configuration(&self) -> &ClientConfiguration;
+    /// Serialize and print data
```
Arjentix, line 118:
```suggestion
    fn configuration(&self) -> &ClientConfiguration;
    
    /// Serialize and print data
```

Arjentix (approved):

client_cli/src/main.rs
Erigara: Discussed in DM, let's do it as part of other task

PR #3610, opened 2023-06-15
[fix] #3473: Make `kagami validator` runnable from outside `iroha` dir

### LINKED ISSUE #3473: [BUG] Kagami could not generate validator when run outside of the repo dir
### OS and Environment

macOS 13.2.1 (22D68)

### GIT commit hash

d7e054d

1. Create a directory:
   ```bash
   mkdir test && cd test
   ```
2. Install `kagami` binary there:
   
   ```bash
   cargo install --git https://github.com/hyperledger/iroha.git \
      --rev d7e054d3aca37f1a93383d2318e7b46fa76e0010 \ # or simply `--branch iroha2-dev`
      --root ./cargo-install-root \
      kagami
   ```
3. Run compiled kagami:
   ```bash
   ./cargo-install-root/bin/kagami validator
   ```

### Actual result

```
Error:
   0: Failed to construct absolute path for: ../../default_validator
   1: Expected `CARGO_MANIFEST_DIR` environment variable
   2: environment variable not found
```

### Expected result

Expected to get the compiled wasm blob.

### Logs in JSON format

N/A

_No response_

### Notes

Kagami relies to be always executed in the root of the repo:

https://github.com/hyperledger/iroha/blob/d7e054d3aca37f1a93383d2318e7b46fa76e0010/tools/kagami/src/main.rs#L815-L826

I don't think this is correct - Kagami should be a standalone, self-contained binary with "all batteries included". So, the wasm binary should be inlined into the binary itself.

This functionality is essential for #2373, which I am working on now. Thus, I assign this issue to myself.

ilchu:
## Description

Instead of failing with a `CARGO_MANIFEST_DIR` env var not found error, `kagami` now copies and builds the necessary packages for the default validator.

### Linked issue

Closes #3473. 

### Benefits
`kagami` gets closer to being a location-independent executable

### Checklist

- [x] Make sure it works with swarm, do the `swarm.rs` adjustments
- [x] <s>Write some tests? Needs to be integration presumably, as we're testing the executable independence</s> <s>EDIT: added a new check in `scripts/check.sh`, but it will pass only after merge, as the generated version SHA should be from the `iroha2-dev` branch to work</s> CI-related things to be addressed separately in #3700 


tools/kagami/src/main.rs
```diff
@@ -849,13 +853,69 @@ mod validator {
...
+            "derive_primitives",
+            "ffi",
+            "macro",
+            "primitives",
+            "schema",
+            "version",
+            "wasm",
+        )
+        .dir(out_dir.as_ref())
+        .run()?;
```
ilchu, line 891: Not really happy here with how many packages from workspace it required to pull up, which is mostly due to `wasm` crate dependencies. Maybe it would make sense to introduce a new feature in that crate so that it would allow compiling only `wasm/validator`, which would make the pull slimmer. For now I've chosen the route of minimal change.

tools/kagami/src/main.rs
```diff
@@ -849,13 +853,69 @@ mod validator {
...
+        cmd!(
+            "git",
+            "clone",
+            "-n",
+            "--depth=1",
+            "--filter=tree:0",
+            "--branch=iroha2-dev",
+            GIT_ORIGIN,
+            out_dir.as_ref()
+        )
```
0x009922, line 869: It is an issue for `kagami swarm --build-from-github`, because it clones the repo already, and it would be redundant to make it twice for validator building.

tools/kagami/src/main.rs
```diff
@@ -849,13 +853,69 @@ mod validator {
         }
     }
 
+    pub fn clone_default_validator(out_dir: impl AsRef<Path>) -> Outcome {
+        const GIT_ORIGIN: &str = "https://github.com/hyperledger/iroha.git";
```
0x009922, line 857: Same const is duplicated in `kagami::swarm`. We should probably move all meta consts in a single place.

tools/kagami/src/main.rs
0x009922: Do we need to optimise it this way?

This partial clone is really hard to cover with tests, thus it is very error prone: the workspace configuration might change and it might be far from obvious that this part will break.

I think it will be simpler and less error-prone if we work with a whole repo.

0x009922 (changes requested):

tools/kagami/src/main.rs
```diff
@@ -849,13 +853,69 @@ mod validator {
...
+
+        // Sparse directory copy as suggested in https://stackoverflow.com/a/52269934/10669935
+        cmd!(
+            "git",
+            "clone",
+            "-n",
+            "--depth=1",
+            "--filter=tree:0",
+            "--branch=iroha2-dev",
```
0x009922, line 866: It might not necessarily be `iroha2-dev`.

tools/kagami/src/main.rs
appetrosyan: I can see this later on, when we have stabilised the libraries. But right now, I'd rather you did what @0x009922 asks you to do.

wasm_builder/Cargo.toml
```diff
@@ -10,4 +10,5 @@ categories = ["development-tools::build-utils"]
 
 [dependencies]
 eyre = "0.6.8"
+path-absolutize = "3.1.0"
```
appetrosyan, line 13: Let's hope we won't need it for long. It's already an unstable feature.

tools/kagami/src/main.rs
ilchu: Remade it to employ swarm's functions.

tools/kagami/src/main.rs
ilchu: Got rid of this func entirely, but didn't really refactor out the consts and common functionality yet.

appetrosyan:
I suggest a script that tests that this works independently of being run via `cargo`.

tools/kagami/src/main.rs
```diff
@@ -705,6 +705,10 @@ mod docs {
 }
 
 mod validator {
```
appetrosyan, line 707: I'd suggest breaking up `main` into smaller modules, given that we have them conveniently defined, just use `rust-analyzer` to inline them.

tools/kagami/src/main.rs
```diff
@@ -713,18 +717,44 @@ mod validator {
     impl<T: Write> RunArgs<T> for Args {
         fn run(self, writer: &mut BufWriter<T>) -> Outcome {
             writer
-                .write_all(&construct_validator()?)
+                .write_all(&construct_validator(None)?)
```
appetrosyan, line 720: Passing options as arguments is usually an anti-pattern. In this case, I'd use `dependency injection`, to move the logic up the call stack.

tools/kagami/src/main.rs
```diff
@@ -713,18 +717,44 @@ mod validator {
...
+                    Ok(source_dir.path().to_owned())
+                }
+            },
+            |manifest_dir| {
+                Path::new(&manifest_dir)
+                    .join("../../default_validator")
+                    .canonicalize()
+                    .wrap_err("Failed to canonicalize path")
+            },
+        )?;
```
appetrosyan, line 755: You should pass `validator_path` as an argument and instead compute it from another function.

tools/kagami/src/swarm.rs
```diff
@@ -20,13 +20,13 @@ use ui::UserInterface;
 
 use super::{ClapArgs, Outcome};
 
-const GIT_REVISION: &str = env!("VERGEN_GIT_SHA");
-const GIT_ORIGIN: &str = "https://github.com/hyperledger/iroha.git";
+pub const GIT_REVISION: &str = env!("VERGEN_GIT_SHA");
+pub const GIT_ORIGIN: &str = "https://github.com/hyperledger/iroha.git";
```
appetrosyan, line 24: Non-public constants are usually a mistake.

tools/kagami/src/swarm.rs
```diff
@@ -232,7 +232,12 @@ enum PrepareConfigurationStrategy {
 }
 
 impl PrepareConfigurationStrategy {
```
appetrosyan, line 234: The word `Strategy` is a pattern of naming where people name their structures after gang-of-four design patterns. To see why this is a bad convention see the image below.
 
9c00c55e6d36b73d11a5858fd547a250-73305366

We should attempt to give this a better name.

0x009922: Hmm, the current "strategy" naming in combination with the enum variant names seems ok for me. I tried to use a different naming, and decided this one to be the best.

The image hasn't helped me to understand why "strategy" is bad in this case. Just a thought that it is avoiding the word "strategy" because of its "bad" history.

Do you have some options for how the naming could be different?

tools/kagami/src/swarm.rs
```diff
@@ -256,8 +261,14 @@ impl PrepareConfigurationStrategy {
...
-                let validator = super::validator::construct_validator()
-                    .wrap_err("Failed to construct the validator")?;
+                let validator = match source {
+                    ResolvedImageSource::Build { ref path } => {
+                        super::validator::construct_validator(Some(path))
+                            .wrap_err("Failed to construct the validator")?
+                    }
+                    _ => super::validator::construct_validator(None)
+                        .wrap_err("Failed to construct the validator")?,
+                };
```
0x009922, line 271: Hmmph, it doesn't work this way.

Whether `kagami swarm` should build the validator or not depends not on the source of image (`build`, `build-from-github` or `image`), but on `--no-default-configuration` option. The validator is a part of the configuration in `<dir>/config` directory, which is volumed in `docker-compose.yml` and does not depend on image source.

So, generating the validator or not directly depends on `self`: `Self::GenerateDefault` means to generate the configuration, including the validator, and `Self::GenerateOnlyDirectory` means... what it says.

tools/kagami/src/main.rs
```diff
@@ -713,18 +717,44 @@ mod validator {
...
-    pub fn construct_validator() -> color_eyre::Result<Vec<u8>> {
+    pub fn construct_validator(
+        swarm_build_dir: Option<&AbsolutePath>,
+    ) -> color_eyre::Result<Vec<u8>> {
         let build_dir = tempfile::tempdir()
             .wrap_err("Failed to create temp dir for runtime validator output")?;
+        let source_dir = tempfile::tempdir()
+            .wrap_err("Failed to create temp dir for default runtime validator sources")?;
+
+        // If `CARGO_MANIFEST_DIR` is set, it means the `kagami` binary is run from inside Iroha repo,
```
0x009922, line 733:
```suggestion
        // If `CARGO_MANIFEST_DIR` is set, it means the `kagami` binary is run with `cargo run`,
```

I am not sure, but when we tested this with @timofeevmd, it turned out that even running a compiled `kagami` binary from the root of the repo doesn't work, because `CARGO_MANIFEST_DIR` env variable is not set.

0x009922 (changes requested):
The change in `kagami swarm` logic is incorrect.

tools/kagami/src/swarm.rs
0x009922: Maybe it should be put into a docstring, though

ilchu: We're doing it inside of a `Self::GenerateDefault` match arm if you double check above the diff. The distinction here is whether we need to clone `iroha` behind the scenes (the arm with `None` passed to the `construct_validator` func) or if we have the shallow clone already.

tools/kagami/src/swarm.rs
ilchu: I do agree that this naming a bit cumbersome, but still well within readable territory, so I wouldn't insist on changing it. As for alternatives, the only thing coming to mind is maybe `GenerateConfig`? And the same could apply to `PrepareDirectoryStrategy`.

tools/kagami/src/swarm.rs
ilchu: I could put the consts into `kagami`'s `mod.rs` after applying this [suggestion](https://github.com/hyperledger/iroha/pull/3610/files#r1242012154) which would be neat.

tools/kagami/src/swarm.rs
0x009922: Indeed, I am blind, my apologies.

tools/kagami/src/swarm.rs
```diff
@@ -464,6 +475,12 @@ impl AsRef<OsStr> for AbsolutePath {
     }
 }
 
+impl From<PathBuf> for AbsolutePath {
+    fn from(path: PathBuf) -> Self {
+        Self { path }
+    }
+}
```
0x009922, line 482: I used `AbsolutePath` newtype following "parse-not-validate" pattern. In other words, my intention was that `AbsolutePath` should be constructed only with `PathBuf` which is absolute. Here `AbsolutePath` is constructed without any validation.

I would suggest changing it to `TryFrom` and put resolution logic inside. It is essentially what `AbsolutePath::absolutize` is.

tools/kagami/src/main.rs
```diff
@@ -713,18 +717,44 @@ mod validator {
...
+        // called from `swarm`
+        let validator_path = std::env::var("CARGO_MANIFEST_DIR").map_or_else(
+            |_| {
+                if let Some(swarm_dir) = swarm_build_dir {
+                    Ok(swarm_dir.join("../../default_validator"))
+                } else {
+                    swarm::shallow_git_clone(
+                        swarm::GIT_ORIGIN,
+                        swarm::GIT_REVISION,
+                        &AbsolutePath::from(source_dir.path().join(swarm::DIR_CLONE)),
```
0x009922, line 744:
```suggestion
                        &AbsolutePath::absolutize(source_dir.path().join(swarm::DIR_CLONE))?,
```

As I sad in another comment, `AbsolutePath` should be constructed through validation.

tools/kagami/src/swarm.rs
0x009922: I am fine with `GenerateConfig`. Or `PrepareConfig`.

0x009922 (changes requested):
I am primarily confused by `impl From<PathBuf> for AbsolutePath`

tools/kagami/src/swarm.rs
appetrosyan: >  It's not necessary to expose everything from a module "just in case".

True. This is not true of constants. 

> I personally think that exposing only essential parts at first place is a better practice. Especially when it is fixable by adding a single pub keyword.

It is true of complex behaviours. The whole point of a constant is that you change it in one place, and the change is propagated.

tools/kagami/src/swarm.rs
appetrosyan: If it weren't for the Rust bureaucracy, I'd recommend using the standard library methods.

tools/kagami/src/main.rs
ilchu: Changed the wording, but the new version also works with compiled standalone, added a little CI check for that too.

appetrosyan (approved):

0x009922 (approved):

PR #3617, opened 2023-06-16
[feature] #3598: Permission token analysis (WASM side)

### LINKED ISSUE #3598: Permission Token Analysis (WASM side)
This is a sub-issue of #3230

Need to implement permission token analysis on WASM side:
Steps:

- [x] Add attribute macro for new entrypoint
- [x] Implement new entrypoint for default validator
- [x] Implement new entrypoint for test validator
- [x] Uncomment & fix the code on Iroha side connect these two parts and make it working

Arjentix:
## Description

- Added new trait `AssociatedConstant<T>` into `iroha_data_model`. Useful for `Token` derive
- Updated `Token` trait and derive
- Updated `iroha_validator::entrypoint` attribute macro
- Implemented new entrypoints in all validators (default and tests)
- Remove tokens from Kagami
- Uncomment code on Iroha side
- Added test

### Linked issue

- Closes #3598 

### Benefits

No more duplication of `PermissionTokenDefinition`s which currently should be the same as the ones defined in the validator

### Checklist

- [x] (optional) I've written unit (not unit but integration) tests for the code changes


data_model/src/lib.rs
```diff
@@ -258,6 +258,12 @@ impl<EXPECTED, GOT> EnumTryAsError<EXPECTED, GOT> {
 #[cfg(feature = "std")]
 impl<EXPECTED: Debug, GOT: Debug> std::error::Error for EnumTryAsError<EXPECTED, GOT> {}
 
+/// Trait to define associated constant of type `T`
+pub trait AssociatedConstant<T> {
```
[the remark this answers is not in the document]
mversic: to begin with we should move this into `wasm/validator` crate

mversic (changes requested):

data_model/src/permission.rs
```diff
@@ -127,15 +127,21 @@ impl PermissionTokenDefinition {
         }
     }
 
-    /// Add parameters to the [`Definition`] replacing any parameters previously defined
+    /// Get *id* of the [`PermissionTokenDefinition`]
+    #[inline]
+    pub const fn id(&self) -> &PermissionTokenId {
```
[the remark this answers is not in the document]
mversic: if it's because of the `const`ness I don't like that reason

core/src/smartcontracts/isi/world.rs
```diff
@@ -374,26 +371,26 @@ pub mod isi {
...
+                .permission_token_definitions()
+                .values()
+                .cloned()
+                .collect::<HashSet<_>>();
+            let new_permission_token_definitions =
+                HashSet::from_iter(new_permission_token_definitions);
+
+            old_permission_token_definitions
+                .difference(&new_permission_token_definitions)
+                .map(|definition| Unregister::<PermissionTokenDefinition> {
```
mversic, line 384: maybe these should not be ISI anymore, and should be directly registered/unregistered because then you allow for our executor to also execute these operations by itself?

core/src/smartcontracts/isi/world.rs
```diff
@@ -358,13 +355,13 @@ pub mod isi {
             let raw_validator = self.object;
             let engine = wsv.engine.clone(); // Cloning engine is cheap
 
-            let (new_validator /*, new_permission_token_definitions*/) =
+            let (new_validator, new_permission_token_definitions) =
                 || -> Result<_, crate::smartcontracts::wasm::error::Error> {
```
mversic, line 359: why is this a closure?

client/tests/integration/upgrade.rs
```diff
@@ -54,14 +109,5 @@ fn validator_upgrade_should_work() -> Result<()> {
...
 
-    // Check that admin can transfer alice's rose now
-    // Creating new transaction instead of cloning, because we need to update it's creation time
-    let transfer_rose_tx = TransactionBuilder::new(admin_id)
-        .with_instructions([transfer_alice_rose])
-        .sign(admin_keypair)?;
-    client
-        .submit_transaction_blocking(&transfer_rose_tx)
-        .expect("Should succeed");
-
```
mversic, line 65: why remove

core/src/smartcontracts/isi/world.rs
mversic: also, in `client/tests/integration/permissions.rs` there is a test that tries to register a permission token. This should be disallowed, no?

data_model/src/lib.rs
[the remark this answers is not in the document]
mversic: I opened a PR [here](https://github.com/Arjentix/iroha/pull/1)

configs/peer/genesis.json
```diff
@@ -102,347 +102,6 @@
           }
         }
       },
-      {
-        "Register": {
```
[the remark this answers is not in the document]
Arjentix: Interesting, but I don't think it makes sence.

1. It's harder to implement
2. It's harder to maintain. Users have to remember to update both `json` and `rs` files
3. Another abstraction which makes the whole system even more complicated
4. No room for runtime migrations (which I plan to implement in future). For example if new `Validator` changes the name of some token it may want to reassing it to the users who owned it's previous version.
5. Tokens and Validator are very tied, I can't see a reason to decouple them

core/src/smartcontracts/wasm.rs
```diff
@@ -848,7 +850,8 @@ impl<'wrld> Runtime<state::Validator<'wrld>> {
         let mut store = self.create_store(state);
         let instance = self.instantiate_module(module, &mut store)?;
 
-        let validate_fn = Self::get_typed_func(&instance, &mut store, export::WASM_MAIN_FN_NAME)?;
+        let validate_fn =
+            Self::get_typed_func(&instance, &mut store, export::VALIDATOR_VALIDATE_FN_NAME)?;
```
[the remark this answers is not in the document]
Arjentix: Because it looks like there is no sence in having it the same as `Smart Conract` and `Trigger` in the same time while we have another entrypoint. To me the new name is more logical

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+        &self,
+        expression: &E,
+    ) -> core::result::Result<E::Value, EvaluationError> {
+        self.0.evaluate(expression)
+    }
+}
+
+/// Entrypoint to return permission token definitions defined in this validator.
+#[entrypoint]
+pub fn permission_tokens() -> Vec<PermissionTokenDefinition> {
```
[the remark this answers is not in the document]
Arjentix: > 2. This entrypoint function should be automatically generated with #[entrypoint] on validate function

I think it would be very unclear from user point of view. Also it would require to know the type of validator, so I'll need to add another attribute thing in this macro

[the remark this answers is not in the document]
Arjentix: > 1. I think this should be a method on the Validate trait

Before we didn't have something like `Validator::validate()` as an entrypoint. It's a custom functions which permforms matching over `NeedsValidationBox`. Maybe we should create this method inside `Validate` trait as well?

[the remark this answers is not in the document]
Arjentix: I mean, I did it in the same way it was before. And we should either do it like that or replace with trait methods, after which entrypoint functions will probably be just wrappers around methods

data_model/src/permission.rs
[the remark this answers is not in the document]
Arjentix: > isn't this method already given by deriving `IdEqOrdHash`?

Doesn't look like so. This derive only implements comparision and hash traits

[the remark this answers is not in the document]
Arjentix: Oh, you're actually right

data_model/src/lib.rs
```diff
@@ -55,7 +55,7 @@ use serde_with::{DeserializeFromStr, SerializeDisplay};
 use strum::EnumDiscriminants;
 
 pub use self::model::*;
-use crate::{account::SignatureCheckCondition, name::Name, transaction::TransactionValue};
+use crate::{account::SignatureCheckCondition, name::Name, transaction::TransactionValue}; // Hack for macros
```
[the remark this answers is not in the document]
Arjentix: Indeed, I think, initially it was for something else

data_model/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > to begin with we should move this into wasm/validator crate as it's only used there

Currently yes. But I don't understand why it's not okay to keep this in `data_model`. Also how would I derive it on `Value` if I won't have it inside `data_model`?

I'll take a look at your PR

data_model/derive/src/variant_discriminant.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+pub fn impl_variant_discriminant(enum_: &VariantDiscriminantEnum) -> TokenStream {
+    let discriminant_type = &enum_.discriminant_type;
+    let impls = enum_.variants.iter().map(|variant| {
+        let Variant {
+            ty,
+            discriminant_name,
+        } = variant;
+        // In order to make doc-tests work, we need to not to use full path to `AssociatedConstant`
```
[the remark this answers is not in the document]
Arjentix: No, the problem is in doc-test for `variant_discriminant_derive()` function

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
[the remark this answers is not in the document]
mversic: > I think it would be very unclear from user point of view. Also it would require to know the type of validator, so I'll need to add another attribute thing in this macro

What do you mean by type of validator? What types of validators are there?
My point here was that user should not be bothered with writing this export. It should be autogenerated

core/src/smartcontracts/isi/world.rs
Arjentix: > there is a test that tries to register a permission token. This should be disallowed, no?

Yes, thank you

Arjentix: > maybe these should not be ISI anymore

I'm not very sure. We probably should discuss this. But I will ban it's usage by our defalut validator

core/src/smartcontracts/isi/world.rs
Arjentix: Because there are two places with errors inside. And I should convert this errors with the same way. So this closure is more like `try-catch`

core/src/smartcontracts/isi/world.rs
mversic: > I'm not very sure

Can you comment on why this operation should be allowed to be executed by the user? Because anyone will be able to execute this operation from the validator

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > I don't want to see more bloat in the validator_with_admin

I think the problem is that it was already bloated and we should try to make it more clear in another way. This new entrypoint doesn't add much lines but keep things explicit for user (what I thought we consider to be good).

> What do you mean by type of validator?

I mean that macro doesn't know on which type to call `<_ as Validator>::permission_tokens()`. We should pass this information to the macro

core/src/smartcontracts/isi/world.rs
Arjentix: > > there is a test that tries to register a permission token. This should be disallowed, no?
> 
> Yes, thank you

Turns out these are ignored tests. Ignored for a long time and I don't why. We should revisit #2581. I'm not sure if I should dig into these tests in this PR. I think we can find more outdated stuff in there

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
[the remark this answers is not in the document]
mversic: > I think the problem is that it was already bloated and we should try to make it more clear in another way. This new entrypoint doesn't add much lines but keep things explicit for user (what I thought we consider to be good).

there is no reason for this to be explicit. If explicit was always preferred we wouldn't have `entrypoint` macro. The problem is that it's hard to hide it

> I mean that macro doesn't know on which type to call <_ as Validator>::permission_tokens(). We should pass this information to the macro

I see, this is a pervasive problem. `ExpressionEvaluator` has the same problem. Perhaps we'll come up with a way to hide these away. Let me close this and make another suggestion

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+    // We need to come up with a better way.
+    if let Some(pos) = tokens.iter().position(|definition| {
+        definition == &iroha_validator::default::domain::tokens::CanUnregisterDomain::definition()
+    }) {
+        tokens.remove(pos);
+    }
+
+    tokens.push(token::CanControlDomainsLives::definition());
+    tokens
+}
```
mversic, line 161:
```suggestion
pub fn permission_tokens() -> Vec<PermissionTokenDefinition> {
    CustomValidator::permission_tokens()
}

trait Validate {
    fn permission_tokens() -> Vec<PermissionTokenDefinition>;
}
```

Erigara:
@Arjentix why do we need `entrypoint` function to access permission tokens? Can't we access `const` array defined inside wasm blob or something like this?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+pub fn permission_tokens() -> Vec<PermissionTokenDefinition> {
+    let mut tokens = iroha_validator::default::permission_tokens();
+
+    // TODO: Not very convenient usage.
+    // We need to come up with a better way.
+    if let Some(pos) = tokens.iter().position(|definition| {
+        definition == &iroha_validator::default::domain::tokens::CanUnregisterDomain::definition()
+    }) {
+        tokens.remove(pos);
+    }
```
Erigara, line 157: Why no `retain`?

Arjentix:
> why do we need `entrypoint` function to access permission tokens? Can't we access `const` array defined inside wasm blob or something like this

@Erigara, cool idea, but you've seen `validator_with_custom_token`. Currently I have no idea how to modify default array and keep it constant in the same time

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
Arjentix: Because `retain()` will iterate over all collection, while `position()` will stop on the first found result

Erigara: Do you expect size of permission tokens be that big that it will make any difference?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
Arjentix: It shoul probably be:

```rust
impl Validate for CustomValidator {
    fn permission_tokens() -> Vec<PermissionTokenDefinition> {
        let mut tokens = DefaultValidator::permission_tokens();
        // ...
    }
}
```

data_model/src/lib.rs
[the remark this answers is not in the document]
mversic: I'll extract this discussion into a separate issue with a separate PR

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
mversic: Not a big deal, but I liked the uniformity of naming all `entrypoint` functions `_iroha_main`. @Erigara @appetrosyan your vote

genesis/src/lib.rs
```diff
@@ -65,12 +65,14 @@ impl GenesisNetwork {
                 .clone()
                 .ok_or_else(|| eyre!("Genesis account private key is empty."))?,
         )?;
-        let transactions_iter = raw_block.transactions.into_iter();
         #[cfg(not(test))]
-        let transactions_iter =
-            transactions_iter.chain(std::iter::once(GenesisTransactionBuilder {
-                isi: vec![UpgradeBox::new(Validator::try_from(raw_block.validator)?).into()],
-            }));
+        let transactions_iter = std::iter::once(GenesisTransactionBuilder {
```
Erigara, line 69: Please add comments with clarification why do we want such order.

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
Erigara: ok, i just thought that it will be more compact
feel free to resolve

core/src/smartcontracts/isi/world.rs
Erigara: I think if all permission tokens are controlled by `validator` there is no point in preserving `Register/Unregister<PermissionToken>`. It will just add additional surface for vulnerabilities.

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
Arjentix: > ok, i just thought that it will be more compact
feel free to resolve

Appreciate your opinion, but, although as Marin said I win only some comparisions, I'd like to keep it this way. Just a feeling 😄

Erigara: `swap_remove` will do the job if we don't care about order of permission tokens

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
Erigara: I vote for preserving old name.

core/src/smartcontracts/isi/world.rs
```diff
@@ -374,26 +371,26 @@ pub mod isi {
...
-            //     .cloned()
-            //     .map(|definition| Register::<PermissionTokenDefinition> { object: definition })
-            //     .try_for_each(|new| new.execute(authority, wsv))?;
+            let old_permission_token_definitions = wsv
+                .permission_token_definitions()
+                .values()
+                .cloned()
+                .collect::<HashSet<_>>();
+            let new_permission_token_definitions =
+                HashSet::from_iter(new_permission_token_definitions);
```
mversic, line 380: I think it's possible that `new_permission_token_definitions` will contain duplicates which will silently be discarded. It would be better if you used `HashSet::insert` and return error if a duplicate is found. There is no guarantee that there will not be two permission token structs of the same name

core/src/smartcontracts/wasm.rs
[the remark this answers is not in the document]
appetrosyan: I vote for the new naming convention.

core/src/smartcontracts/isi/world.rs
appetrosyan: I think that this should be removed. But I don't think that it necessarily needs to be removed in this PR.

genesis/src/lib.rs
```diff
@@ -65,12 +65,16 @@ impl GenesisNetwork {
...
                 .ok_or_else(|| eyre!("Genesis account private key is empty."))?,
         )?;
-        let transactions_iter = raw_block.transactions.into_iter();
         #[cfg(not(test))]
-        let transactions_iter =
-            transactions_iter.chain(std::iter::once(GenesisTransactionBuilder {
-                isi: vec![UpgradeBox::new(Validator::try_from(raw_block.validator)?).into()],
-            }));
+        // First instruction should be Validator upgrade.
+        // This makes possible to grant permissions to users in genesis.
```
appetrosyan, line 70: Excellent. This needs to be documented in the official docs.

data_model/derive/src/lib.rs
```diff
@@ -484,3 +485,45 @@ pub fn partially_tagged_deserialize_derive(input: TokenStream) -> TokenStream {
...
+/// enum MyEnumKind {
+///    Unsigned,
+///    String,
+///    Boolean,
+/// }
+///
+/// assert_eq!(<u32 as AssociatedConstant<MyEnumKind>>::VALUE, MyEnumKind::Unsigned);
+/// assert_eq!(<String as AssociatedConstant<MyEnumKind>>::VALUE, MyEnumKind::String);
+/// assert_eq!(<bool as AssociatedConstant<MyEnumKind>>::VALUE, MyEnumKind::Boolean);
+/// ```
```
appetrosyan, line 523:
```suggestion
/// Derive macro implementing the [`AssociatedConstant<T>`](iroha_data_model::AssociatedConstrant) trait. 
/// # Notes
/// You should import `AssociatedConstant`. 
///
/// # Attributes
///
/// `#[variant_discriminant(name(DiscriminantType))]` attribute where `DiscriminantType` is the name of the `enum` that should serve as the discriminant enumeration. 
///
/// # Examples
///
/// ```
/// use iroha_data_model_derive::VariantDiscriminant;
/// use iroha_data_model::AssociatedConstant;
///
/// #[derive(VariantDiscriminant)]
/// #[variant_discriminant(name(MyEnumKind))]
/// enum MyEnum {
///     Unsigned(u32),
///     String(String),
///     Boolean(bool),
/// }
///
/// #[derive(Debug, PartialEq, Eq)]
/// enum MyEnumKind {
///    Unsigned,
///    String,
///    Boolean,
/// }
///
/// assert_eq!(<u32 as AssociatedConstant<MyEnumKind>>::VALUE, MyEnumKind::Unsigned);
/// assert_eq!(<String as AssociatedConstant<MyEnumKind>>::VALUE, MyEnumKind::String);
/// assert_eq!(<bool as AssociatedConstant<MyEnumKind>>::VALUE, MyEnumKind::Boolean);
/// ```
```

appetrosyan (commented):
This needs to have a large `dev_update`, probably with a detailed explanation.

core/src/smartcontracts/isi/world.rs
mversic: since there is a concern about modifying `genesis.json` @Arjentix you mark these two methods as deprecated and have them `panic`/return error if somebody calls them accidentally

genesis/src/lib.rs
Arjentix: This does not affect user api, because validator is specified in a special field and we automatically move it to be the first instruction

core/src/smartcontracts/isi/world.rs
Arjentix: Removed these instructions

mversic (approved):

mversic (dismissed):

client/tests/integration/upgrade.rs
```diff
@@ -33,17 +36,66 @@ fn validator_upgrade_should_work() -> Result<()> {
...
+    let definitions = client.request(FindAllPermissionTokenDefinitions)?;
+    assert!(!definitions
+        .iter()
+        .any(|definition| definition.id() == &can_unregister_domain_token_id));
+
+    // Check that `can_control_domain_lives` exists
+    let can_control_domain_lives: PermissionTokenId = "can_control_domain_lives".parse()?;
+    assert!(!definitions
+        .iter()
+        .any(|definition| definition.id() == &can_control_domain_lives));
```
mversic, line 83:
```suggestion
    // Check that `can_control_domain_lives` exists
    let can_control_domain_lives: PermissionTokenId = "can_control_domain_lives".parse()?;
    assert!(definitions
        .iter()
        .any(|definition| definition.id() == &can_control_domain_lives));
```

shouldn't this be like this? I noticed it fails in my PR

client/tests/integration/permissions.rs
```diff
@@ -381,7 +348,7 @@ mod token_parameters {
         let (_rt, _peer, client) = <PeerBuilder>::new().with_port(10_750).start_with_runtime();
         wait_for_genesis_committed(&[client.clone()], 0);
 
-        register_test_token_definition(&client);
+        // register_test_token_definition(&client);
```
Erigara, line 351: Why commented?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/// Entrypoint to return permission token definitions defined in this validator.
+#[entrypoint]
+pub fn permission_tokens() -> Vec<PermissionTokenDefinition> {
+    CustomValidator::permission_tokens()
+}
+
+/// Allow operation if authority is `admin@admin` and if not,
+/// fallback to [`DefaultValidator::validate()`].
```
Erigara, line 169: I believe this comment is not correct.

client/tests/integration/permissions.rs
mversic: we don't support registering token definitions through ISI anymore, only through validator migration

Erigara: i understand that, but why to preserve commented code?

client/tests/integration/upgrade.rs
Arjentix: Right. Copy-paste mistake. I wonder why it doesn't fail for me

client/tests/integration/permissions.rs
Arjentix: Yes, I left it commented so that someone who will work in #2581 have some glue on how it worked before

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
Arjentix: Damn copy-paste

client/tests/integration/upgrade.rs
Arjentix: It's because of typo

mversic (approved):

Erigara (approved):

PR #3618, opened 2023-06-17
[refactor]: strong type hash in Value

mversic:
## Description

* use strongly typed hashes in `Value` enum
* remove unnecessary boxing of structs
* move `TransactionValue`/`TransactionQueryResult` from `transaction.rs` into `query.rs`

### Checklist


0x009922:
@mversic, could you add some description to the PR?

data_model/src/query.rs
```diff
@@ -153,12 +162,89 @@ pub mod model {
...
+    #[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+    #[ffi_type]
+    pub enum TransactionValue {
+        /// Committed transaction
+        Transaction(VersionedSignedTransaction),
+        /// Rejected transaction with reason of rejection
+        RejectedTransaction(RejectedTransaction),
+    }
+
+    /// `TransactionQueryResult` is used in `FindAllTransactions` query
```
Arjentix, line 176:
```suggestion
    /// [`TransactionQueryResult`] is used in [`FindAllTransactions`] query
```

data_model/src/query.rs
```diff
@@ -1147,7 +1233,7 @@ pub mod transaction {
 
     impl FindTransactionByHash {
         ///Construct [`FindTransactionByHash`].
```
Arjentix, line 1235:
```suggestion
        /// Construct [`FindTransactionByHash`].
```

Arjentix (approved):
I remember we had `HashOf` usage before, but then it was changed to `Hash`. I think it was done by @SamHSmith, do you remember why?

schema/gen/src/lib.rs
```diff
@@ -103,24 +103,11 @@ types!(
     BlockMessage,
     BlockRejectionReason,
     BlockSubscriptionRequest,
-    Box<Account>,
```
appetrosyan, line 106: Interesting that so many of the types are redundant thanks to one small change.

data_model/src/lib.rs
```diff
@@ -1153,11 +1178,9 @@ where
     }
 }
 
-// TODO: This macro looks very similar to `from_and_try_from_value_identifiable`
-// and `from_and_try_from_value_identifiablebox` macros. It should be possible to
-// generalize them under one macro
+// TODO: The following macros looks very similar. Try to generalize them under one macro
```
appetrosyan, line 1181: Honestly, don't think that that's a good idea. Generating stuff from macros itself impairs readability. Adding another layer makes it very difficult to reason about `macro_rules`. I would suggest instead turning these into `derive` macros, if possible.

data_model/src/lib.rs
```diff
@@ -650,27 +654,27 @@ pub mod model {
     pub enum RegistrableBox {
         /// [`Peer`](`peer::Peer`) variant.
         #[display(fmt = "Peer {_0}")]
-        Peer(Box<<peer::Peer as Registered>::With>),
+        Peer(<peer::Peer as Registered>::With),
```
appetrosyan, line 657: This isn't exactly equivalent to `Box`-ing on the outside. Consider that the enum before had a fixed size of $2 \times \text{usize}$ and now has the size of the biggest `Registered::With`. 

As such `Box`-ing `val.into()` defeats the whole point of boxing to reduce wasted RAM space. Consider removing `Box` es entirely.

appetrosyan (approved):

data_model/src/lib.rs
mversic: > Honestly, don't think that that's a good idea. Generating stuff from macros itself impairs readability

having worked with macros for a while, I generally agree with this sentiment. Not sure in this case

Issue #3318, opened 2023-03-22
[bug] Transaction revalidation order


Arjentix:
By looking at `core/srd/block.rs` I think there is an issue with our revalidation mechanism. Need to check if I'm right or not.

## Steps to test

1. Submit some failing transaction. I.e. spend asset that user don't have or something like that
2. Submit transaction that could make tx *1* be successful if thix tx *2* preceeded it. I.e. mint some asset amount for user

## What I expect to happen

I believe that peer other than the one you have submitted these transactions will complain about tx *1* because it should fail but it won't.

So if you submit transaction to **peer 1**, then it will have tx *1* failed and tx *2* succeed. And all other peers should have some failure in logs, because they will have both transactions successful.

## What should happen in valid Iroha 2

Nothing wrong, just every peer agrees about tx *1* being failed

SamHSmith:
You are right there is a subtle problem here. Assume the sequence Mint, Register, Mint. The first mint fails and gets put in the rejected transactions. But in validation those get validated after, so it actually passes this time. We need to revalidate transactions, in the same order as during creation.

Arjentix:
> But in validation those get validated after, so it actually passes this time. We need to revalidate transactions, in the same order as during creation.

Yeah, that's what I've seen in the code and decided to create this issue. Have you managed to reproduce this bug?

SamHSmith:
Yes I reproduced it in a test and managed to fix it.

PR #3619, opened 2023-06-17
#3318: Interleave rejected and accepted transactions in block

SamHSmith:
## Description

Adds an index on rejected transactions that is then used to revalidate in the same order as when the block was created.

### Linked issue #3318

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -18,7 +18,7 @@ use iroha_data_model::{
     block::*,
     events::prelude::*,
     peer::PeerId,
-    transaction::{error::TransactionRejectionReason, prelude::*, RejectedTransaction},
+    transaction::{error::TransactionRejectionReason, prelude::*},
 };
 use iroha_genesis::GenesisTransaction;
 use parity_scale_codec::{Decode, Encode};
@@ -124,10 +124,8 @@ pub enum TransactionRevalidationError {
 pub struct PendingBlock {
     /// Block header
     pub header: BlockHeader,
-    /// Array of rejected transactions.
-    pub rejected_transactions: Vec<RejectedTransaction>,
-    /// Array of transactions, which successfully passed validation and consensus step.
-    pub transactions: Vec<VersionedSignedTransaction>,
+    /// Array of transactions.
+    pub transactions: Vec<TransactionValue>,
     /// Event recommendations.
     pub event_recommendations: Vec<Event>,
     /// Signatures of peers which approved this block
@@ -171,34 +169,36 @@ impl BlockBuilder {
         };
 
         let mut txs = Vec::new();
-        let mut rejected = Vec::new();
 
         for tx in self.transactions {
             match transaction_validator.validate(tx, height == 1, &mut self.wsv) {
-                Ok(transaction) => txs.push(transaction),
+                Ok(transaction) => txs.push(TransactionValue {
+                    tx: transaction,
+                    error: None,
+                }),
                 Err((transaction, error)) => {
                     iroha_logger::warn!(
                         reason = %error,
                         caused_by = ?error.source(),
                         "Transaction validation failed",
                     );
-                    rejected.push(RejectedTransaction { transaction, error })
+                    txs.push(TransactionValue {
+                        tx: transaction,
+                        error: Some(error),
+                    });
                 }
             }
         }
         header.transactions_hash = txs
             .iter()
-            .map(VersionedSignedTransaction::hash)
+            .filter(|tx| tx.error.is_none())
+            .map(|tx| tx.tx.hash())
             .collect::<MerkleTree<_>>()
             .hash();
-        header.rejected_transactions_hash = rejected
+        header.rejected_transactions_hash = txs
             .iter()
-            .map(
-                |RejectedTransaction {
-                     transaction,
-                     error: _,
-                 }| transaction.hash(),
-            )
+            .filter(|tx| tx.error.is_some())
+            .map(|tx| tx.tx.hash())
             .collect::<MerkleTree<_>>()
             .hash();
         // TODO: Validate Event recommendations somehow?
@@ -212,7 +212,6 @@ impl BlockBuilder {
 
         PendingBlock {
             header,
-            rejected_transactions: rejected,
             transactions: txs,
             event_recommendations: self.event_recommendations,
             signatures,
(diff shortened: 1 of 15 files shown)


data_model/src/transaction.rs
```diff
@@ -172,6 +172,7 @@ pub mod model {
     #[ffi_type]
     pub struct RejectedTransaction {
         pub transaction: VersionedSignedTransaction,
+        pub block_order_index: u64,
```
mversic, line 175: position of a transaction within a block should not be a property of a transaction, but a property of the block. You should explore way to move `block_order_index` into the block

mversic: on 2nd thought, this might be fine. I want to see other opinions

mversic: Here is my suggestion. Currently we have block defined like this:
```rs
        pub struct CommittedBlock {
            pub header: BlockHeader,
            pub rejected_transactions: Vec<RejectedTransaction>,
            pub transactions: Vec<VersionedSignedTransaction>,
        }
```

we could define it like this:
```rs
        pub struct CommittedBlock {
            pub header: BlockHeader,
            /// This Vec contains all transactions rejected and accepted
            pub transactions: Vec<VersionedSignedTransaction>,
            /// This Vec contains location in the previous Vec + rejection reason
            pub rejections: Vec<(u64, TransactionRejectionReason)>
        }
```

SamHSmith: I will take your suggestion.

SamHSmith: Actually doing this change would require a lot more refactoring. Since I am not do do such thing I think we should merge this PR as is. Roman and I have already discussed how this could be unified into a single transaction queue with a boolean to indicate transaction failure. I have created the issue here. #3623

SamHSmith:
Tests fail when run at the same time. Seems like a port alloction problem.

data_model/src/transaction.rs
Erigara: > on 2nd thought, this might be fine. I want to see other opinions

I also think that order information should be part of the block.
It will code cleaner i think. 
Additionally i don't think that we should give this information to user on transactions.

Erigara (changes requested):

client/tests/integration/asset_propagation.rs
```diff
@@ -59,3 +59,59 @@ fn client_add_asset_quantity_to_existing_asset_should_increase_asset_amount_on_a
     )?;
     Ok(())
 }
+
+#[test]
+fn tx_order_same_in_validation_and_revalidation() -> Result<()> {
+    // Given
+    let (_rt, network, client) = <Network>::start_test_with_runtime(4, Some(10_450));
```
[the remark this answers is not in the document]
SamHSmith: Ah thank you for that insight.

I will add an additional test in block.rs.

data_model/src/transaction.rs
SamHSmith: Ok I will pursue the single tx array.

client/tests/integration/asset_propagation.rs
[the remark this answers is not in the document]
SamHSmith: On second thought I agree with you and will only have a test in block.rs.

data_model/src/transaction.rs
Arjentix: Why can't we just store an array of marked transactions? Why to store rejected transactions in a separate field. I can't see where we use this

core/src/queue.rs
```diff
@@ -223,7 +223,7 @@ impl Queue {
     pub fn push(&self, tx: AcceptedTransaction, wsv: &WorldStateView) -> Result<(), Failure> {
         trace!(?tx, "Pushing to the queue");
         if let Err(err) = self.check_tx(&tx, wsv) {
-            warn!("Failed to evaluate signature check");
+            warn!("Failed to evaluate signature check. Error = {}", err);
```
Arjentix, line 226:
```suggestion
            warn!(Error=%err, "Failed to evaluate signature check);
```

I think it's more convinient to use `tracing` builtin things

core/src/block.rs
```diff
@@ -620,68 +631,61 @@ fn revalidate_transactions(
...
     // Check that valid transactions are still valid
-    let _transactions = transactions
-        .iter()
-        .cloned()
-        .map(|tx| {
-            if is_genesis {
-                Ok(AcceptedTransaction::accept_genesis(GenesisTransaction(tx)))
+    let mut tx_iter = transactions.iter().cloned().peekable();
+    let mut rejected_tx_iter = rejected_transactions.iter().cloned().peekable();
+    for tx_index in 0..u64::MAX {
```
Arjentix, line 636: Oh no. I'm sure there is a way to not to iterate over such a big amount of numbers. There should be a way to place exit condition in the loop condition here

data_model/src/transaction.rs
Erigara: Only place i can think of is merkle tree root hash, but we can compute this hash anyway.

data_model/src/transaction.rs
```diff
@@ -165,12 +166,27 @@ pub mod model {
...
+        pub error: Option<error::TransactionRejectionReason>,
+    }
+
+    /// `TransactionQueryResult` is used in `FindAllTransactions` query
+    #[derive(
+        Debug, Clone, PartialEq, Eq, Getters, Decode, Encode, Deserialize, Serialize, IntoSchema,
+    )]
+    #[getset(get = "pub")]
+    #[ffi_type]
+    pub struct TransactionQueryResult {
```
mversic, line 185: please return this struct in the `query.rs` module. It only exists because of querying

data_model/src/transaction.rs
```diff
@@ -165,12 +166,27 @@ pub mod model {
...
     #[ffi_type]
-    pub struct RejectedTransaction {
-        pub transaction: VersionedSignedTransaction,
-        pub error: error::TransactionRejectionReason,
+    pub struct TransactionValue {
+        /// Committed transaction
+        pub tx: VersionedSignedTransaction,
+        /// Reason of rejection
+        pub error: Option<error::TransactionRejectionReason>,
+    }
```
mversic, line 177: I like the new layout

Erigara (approved):

PR #3621, opened 2023-06-19
[feature]  #3468: Server-side cursor

### LINKED ISSUE #3468: [suggestion] Re-entrant lazily evaluated pagination
### Feature request

To fix #3379 we should explore defaulting to reasonable pagination and optimising the behaviour for paginated query results to not cause a stoppage in Iroha. 

Specifically pagination should be: 
- [ ] Re-entrant. There should be a mechanism to request a query result, paginate it, and request a different part of the query result at a later time. 
- [ ] Lazily-evaluated. We only want to compute the parts of the query which are relevant to generating the current page. If the query asks for all domains, but only the first 100 are present in the page, we should "pause" the evaluation until a different page is requested. 
- [ ] Optimal. If the 3-rd page is requested, we only resume evaluation to find the 3rd page of results and nothing more.  
- [ ] Idempotent. Requesting the same page twice is not an error. The result is identical. 
- [ ] Functional reactive. If the data changed during the evaluation of the query, the result consistent with the old data is returned. 
- [ ] Sessions expire. If a paginated result was requested and not resumed for a period of time, the re-entrant session expires and new data is returned. 
- [ ] Resettable. The session can be reset under certain conditions to be discussed with @gv-timur and @arndey. 

### Motivation

Needed for internal projects

### Who can help?

@mversic

mversic:
## Description

* disallow deserialization of queries with incorrect signature
* return iterator over query execution result instead of Vec (iterator is lazy evaluated)
* add server-side cursor to return query result in batches

### Linked issue

Closes #3468 

### Checklist


core/src/wsv.rs
```diff
@@ -737,24 +751,13 @@ impl WorldStateView {
...
-    pub fn peers(&self) -> Vec<Peer> {
-        let mut vec = self
-            .world
-            .trusted_peers_ids
-            .iter()
-            .map(|peer| Peer::new((*peer).clone()))
-            .collect::<Vec<Peer>>();
-        vec.sort();
-        vec
+    pub fn peers(&self) -> impl ExactSizeIterator<Item = &PeerId> {
```
appetrosyan, line 754: Much more functional. I like it.

core/src/wsv.rs
```diff
@@ -552,7 +563,7 @@ impl WorldStateView {
     ///
     /// # Errors
     /// Fails if there is no domain
-    pub fn domain(&self, id: &<Domain as Identifiable>::Id) -> Result<&Domain, FindError> {
+    pub fn domain<'wsv>(&'wsv self, id: &DomainId) -> Result<&'wsv Domain, FindError> {
```
appetrosyan, line 566: Glad we got rid of the associated typing. It didn't make much sense anyway.

core/src/wsv.rs
```diff
@@ -159,39 +155,58 @@ impl WorldStateView {
...
-        tokens
+
+        Ok(tokens.into_iter())
     }
 
     /// Return a set of permission tokens granted to this account not as part of any role.
+    ///
+    /// # Errors
+    ///
+    /// - if there is no account
```
appetrosyan, line 193:
```suggestion
    /// - `account_id` is not found in `self.world`. 
```

core/src/wsv.rs
```diff
@@ -159,39 +155,58 @@ impl WorldStateView {
...
     }
 
     /// Return a set of all permission tokens granted to this account.
-    pub fn account_permission_tokens(&self, account: &Account) -> BTreeSet<PermissionToken> {
-        let mut tokens: BTreeSet<PermissionToken> =
-            self.account_inherent_permission_tokens(account).collect();
+    ///
+    /// # Errors
+    ///
+    /// - if there is no account
```
appetrosyan, line 169:
```suggestion
    /// - if `account_id` is not found in `self`
```

core/src/wsv.rs
```diff
@@ -159,39 +155,58 @@ impl WorldStateView {
...
     /// # Errors
     /// Fails if there is no domain or account
-    pub fn account_assets(&self, id: &AccountId) -> Result<Vec<Asset>, QueryExecutionFail> {
-        self.map_account(id, |account| account.assets.values().cloned().collect())
+    pub fn account_assets(
+        &self,
+        id: &AccountId,
+    ) -> Result<impl ExactSizeIterator<Item = Asset> + '_, QueryExecutionFail> {
+        self.map_account(id, |account| account.assets.values().cloned())
     }
```
appetrosyan, line 163: This is a little out of line with the other similar functions. Why doesn't this return `ExactSizeIterator<Item = &Asset>` and allow the user to then apply `Iter::cloned` to them?

core/src/sumeragi/main_loop.rs
```diff
@@ -1049,7 +1056,7 @@ trait ApplyBlockStrategy {
     fn before_update_hook(sumeragi: &mut Sumeragi);
 
     /// Operation to invoke in kura to store block.
-    fn kura_store_block(kura: &Kura, block: Arc<VersionedCommittedBlock>);
+    fn kura_store_block(kura: &Kura, block: VersionedCommittedBlock);
```
appetrosyan, line 1059: Excellent, now we have tighter control over the lifetime of the blocks.

cli/src/torii/routing.rs
```diff
@@ -135,49 +73,34 @@ pub(crate) async fn handle_queries(
...
-    let request: VerifiedQuery = request.try_into().map_err(ValidationFail::from)?;
-
-    let (result, filter) = {
-        let mut wsv = sumeragi.wsv_clone();
-        let (valid_request, filter) = request.validate(&mut wsv)?;
-        let original_result = valid_request.execute(&wsv).map_err(ValidationFail::from)?;
-        (filter.filter(original_result), filter)
-    };
+) -> Result<Scale<VersionedQueryResult>> {
+    let mut wsv = sumeragi.wsv_clone();
```
appetrosyan, line 77: Need to find a way to elide this clone. 

From a common-sense point of view, queries should **not** mutate the `wsv` and as such we not only don't need to clone it, we only need an immutable borrow. If we could externalise the mutation, segregate it and make it reversible, the `&wsv` should be enough.

cli/src/torii/routing.rs
```diff
@@ -207,7 +130,7 @@ fn apply_sorting_and_pagination(
             .paginate(pagination)
             .collect()
     } else {
-        vec_of_val.into_iter().paginate(pagination).collect()
+        iter.paginate(pagination).collect()
```
appetrosyan, line 133: So the magic must be happening in the `paginate` method.

client/src/client.rs
```diff
@@ -105,10 +105,10 @@ where
         // Separate-compilation friendly response handling
         fn _handle_query_response_base(
             resp: &Response<Vec<u8>>,
-        ) -> QueryHandlerResult<VersionedPaginatedQueryResult> {
+        ) -> QueryHandlerResult<VersionedQueryResult> {
```
appetrosyan, line 108: I wonder if we can make `Versioned` a generic wrapper?

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -118,10 +118,12 @@ fn change_asset_metadata_after_1_sec() -> Result<()> {
...
-    let value = test_client.request(FindAssetDefinitionKeyValueByIdAndKey {
-        id: asset_definition_id.into(),
-        key: key.into(),
-    })?;
+    let value = test_client
+        .request(FindAssetDefinitionKeyValueByIdAndKey {
+            id: asset_definition_id.into(),
+            key: key.into(),
+        })
+        .map(Into::into)?;
```
appetrosyan, line 126:
```suggestion
        ?.into();
```

core/quang
```diff
@@ -0,0 +1,10 @@
+1. repositories:
+    - cbindgen(git@github.com:mozilla/cbindgen.git),
+    - iroha2(git@github.com:hyperledger/iroha.git) iroha2-stable branch
+    - iroha2-java(git@github.com:hyperledger/iroha-java.git)
+
+2. run iroha nodes and connect to it via iroha_client_cli and http app
+3. run cbindgen on a small example:
+ - export a function from Rust
+ - generate C bindings with cbindgen
+ - link this C lib from C and Java
```
appetrosyan, line 10: Nice to see some progress on the intern.

core/src/smartcontracts/wasm.rs
```diff
@@ -697,7 +698,14 @@ impl<'wrld, S: state::GetCommon<'wrld>, R: DefaultExecute> ExecuteOperations<'wr
             .clone() // Cloning validator is a cheap operation
             .validate(wsv, &common_state.authority, query.clone())?;
 
-        query.execute(wsv).map_err(Into::into)
+        query
+            .execute(wsv)
+            .map_err(Into::into)
+            .map(|lazy_value| match lazy_value {
+                LazyValue::Value(value) => value,
+                // NOTE: Returning references to the host system is a security risk
```
appetrosyan, line 706: More info on this please.

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -251,7 +255,8 @@ impl Set {
 
                 Some(result)
             })
-            .collect()
+            .collect::<Vec<_>>()
```
appetrosyan, line 258: This looks odd.

data_model/src/query.rs
```diff
@@ -1357,43 +1374,96 @@ pub mod http {
...
+                    .map_err(D::Error::custom)
+            }
+        }
+    }
+
+    #[cfg(feature = "transparent_api")]
+    impl VersionedSignedQuery {
+        /// Return query signature
+        pub fn signature(&self) -> &SignatureOf<QueryPayload> {
+            let VersionedSignedQuery::V1(query) = self;
```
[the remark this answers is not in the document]
mversic: I agree, I'm fixing it here and there when I can. Except in the `data_model` no other crate should be version aware

data_model/src/query.rs
```diff
@@ -1357,43 +1374,96 @@ pub mod http {
...
     // TODO: This is the only structure whose inner fields are exposed. Wrap it in model macro?
     #[derive(Debug, Clone, Decode, Encode, Deserialize, Serialize, IntoSchema)]
-    #[version_with_scale(n = 1, versioned = "VersionedPaginatedQueryResult")]
-    pub struct PaginatedQueryResult {
+    #[version_with_scale(n = 1, versioned = "VersionedQueryResult")]
+    pub struct QueryResult {
         /// The result of the query execution.
-        pub result: QueryResult,
-        /// The filter that was applied to the Query result. Returned as a sanity check, but also to ease debugging on the front-end.
-        pub filter: PredicateBox,
```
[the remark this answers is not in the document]
mversic: if possible I'll return it. It gave me some itch

data_model/src/query.rs
```diff
@@ -1357,43 +1374,96 @@ pub mod http {
...
-            Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema,
-        )]
-        #[version_with_scale(n = 1, versioned = "VersionedQueryResult")]
-        #[serde(transparent)]
-        #[repr(transparent)]
-        // TODO: This should be a separate type, not just wrap Value because it infects Value
-        // with variants that can only ever be returned, i.e. can't be used in instructions
-        // enum QueryResult { ... }
-        pub struct QueryResult(pub Value);
+    mod candidate {
```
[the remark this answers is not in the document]
mversic: > If not, I don't necessarily want for you to change that, just to propose that related functions (e.g. constructors, setters, getters etc.) be grouped like this, while other business functions are separated with a single blank line.

sometimes it's intentional sometimes it's not. If I see them logically grouped I omit newline

data_model/src/query.rs
```diff
@@ -183,7 +183,22 @@ pub mod model {
...
+    fn from(source: MetadataValue) -> Self {
+        source.0
+    }
+}
+
+impl From<Value> for MetadataValue {
+    fn from(source: Value) -> Self {
+        Self(source)
     }
 }
```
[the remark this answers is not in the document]
mversic: yeah, I think there is no pain in making it public. It's that we by default mostly wrap everything as opaque

core/src/wsv.rs
```diff
@@ -694,6 +699,15 @@ impl WorldStateView {
...
 
+    fn account(&self, id: &AccountId) -> Result<&Account, FindError> {
+        self.domain(&id.domain_id).and_then(|domain| {
+            domain
+                .accounts
+                .get(id)
+                .ok_or_else(|| FindError::Account(id.clone()))
+        })
+    }
```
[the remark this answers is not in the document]
mversic: > What is the utility of these private getters?

none in particular, but we have it in other places. If you want this change it would require a systematic audit of wsv. I don't think we have the bandwidth for that atm

core/src/wsv.rs
mversic: yeah, that was my fault. I introduced that long time ago. Recently, I saw this in the documentation examples and figured it's unintelligible to a newcomer, just a complexity for no benefit.

core/src/wsv.rs
mversic: I agree, there must have been some pesky lifetime issue, I'll try to return a reference

core/src/sumeragi/main_loop.rs
mversic: yes, if we could also return a `&Block` instead of `Arc<Block>` from `kura` it would be a lot better. I had to go through some escapades with to return iterator over blocks because of this with the help of @Erigara

cli/src/torii/routing.rs
mversic: yes and no. Cloning is not here because of mutation but to relieve contention on re-entrant pagination. We don't want to clone on queries that return a non-iterable value, but for paginated queries we have to store the snapshot of the wsv

appetrosyan: You're right. I need to reformulate the comment. 

What threw me off is that this logic is not supposed to be in the `routing` module of Torii. 

## Cloning WSV

It's fine if the clones are cheap and that's what we intend to have eventually. In that case this logic should be intrinsic to how all queries are handled. In other words, the fact that we're cloning isn't an interface detail anymore. We should describe whether a query needs to clone the `wsv` in its `ValidQuery` impl,  if we don't clone in all circumstances. 

Unfortunately, we also can't make the persistent data structure for a cheap WSV with off-the-shelf components. We need to go custom and this is why for all intents and purposes until `wsv` clones are cheap we should treat them as expensive. 

## COW WSV

Given expensive `wsv` clones, and/or expensive `clone()` + `execute()` combined operations, it might make sense to hook into the event system and ensure that the `wsv` is only updated if a relevant event was emitted somewhere else. This gives you a more granular `Cow`. 

Another point is that evaluating the query to completion might be as fast as cloning the `wsv`. In that case we should be even more granular about what we clone and what we borrow.

appetrosyan: ## Querying by block height

In either case, it might also be worth considering persistent `wsv` snapshots for each block height. Each query optionally includes the block height for which it must execute. If an appropriate `wsv` is stored in the cache, it goes as before. The `wsv` cloning can be done asynchronously so we don't block normal operations and lazily so we don't need to clone each time we receive a query. 

Once the snapshot is stored, the data there is immutable, so multiple queries can operate on the same data simultaneously. This would insert a delay until the current block is committed, but until that happens the information isn't really *in* the blockchain. 

The logic can then be reused for a lightweight archival node.

Erigara: > Given expensive wsv clones, and/or expensive clone() + execute() combined operations, it might make sense to hook into the event system and ensure that the wsv is only updated if a relevant event was emitted somewhere else. This gives you a more granular Cow.

We already do that, we can watch for updates of `wsv` through sumeragi `wsv_updated` method. Problem is that it isn't easily integrated into `torii`.

mversic: > We should describe whether a query needs to clone the wsv in its ValidQuery impl, if we don't clone in all circumstances.

I think this insight alone will solve a lot of headache. I can do this if `ValidQuery::execute(&wsv)` will never return references to the `wsv` and if it returns iterator then `Iterator::Item` will never contain a reference to the `wsv`. Considering how `ValidQuery::execute` is used I think this approach should work.

data_model/src/query.rs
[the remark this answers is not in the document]
appetrosyan: The people who requested this information are under your direct supervision. You can return it, **or** you can convince them that it's unnecessary. I personally think that sending extra information in a debug context is always good.

cli/src/torii/routing.rs
appetrosyan: > We already do that, we can watch for updates of wsv through sumeragi wsv_updated method. Problem is that it isn't easily integrated into torii.

Hence the very first thing I mentioned was that this code **does not belong in `torii`**.

core/src/kura.rs
```diff
@@ -319,17 +319,17 @@ impl Kura {
     }
 
     /// Put a block in kura's in memory block store.
-    pub fn store_block(&self, block: impl Into<Arc<VersionedCommittedBlock>>) {
-        let block = block.into();
-        self.block_data.lock().push((block.hash(), Some(block)));
+    pub fn store_block(&self, block: VersionedCommittedBlock) {
```
Erigara, line 322: It was `impl Into<Arc<VersionedCommittedBlock>>` on purpose to avoid redundant cloning.

core/src/sumeragi/main_loop.rs
```diff
@@ -313,7 +307,7 @@ impl Sumeragi {
         &mut self,
         block: impl Into<VersionedCommittedBlock>,
     ) {
-        let committed_block = Arc::new(block.into());
+        let committed_block = block.into();
```
Erigara, line 310: Why you've decided not to use `Arc` here, it was used to avoid redundant cloning?

core/src/sumeragi/main_loop.rs
```diff
@@ -335,20 +329,33 @@ impl Sumeragi {
...
+            .map(|s| s.public_key())
+            .cloned()
+            .collect::<Vec<PublicKey>>();
+
         // https://github.com/hyperledger/iroha/issues/3396
         // Kura should store the block only upon successful application to the internal WSV to avoid storing a corrupted block.
         // Public-facing WSV update should happen after that and be followed by `BlockCommited` event to prevent client access to uncommitted data.
-        Strategy::kura_store_block(&self.kura, Arc::clone(&committed_block));
+        // TODO: Redundant clone
+        Strategy::kura_store_block(&self.kura, committed_block);
```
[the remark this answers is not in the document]
Erigara: It was introduced by me, to distinguish between committing new block and replacing old block, didn't come up with better name back then.

core/src/kura.rs
mversic: I avoided redundant clone with other means

core/src/sumeragi/main_loop.rs
mversic: I avoided redundant clone with other means

mversic: > Why do you think it will be better?

There is no need for `Arc` in this case. Also, I'd like if we could work toward giving out `&T` instead of `Arc<T>` from `Kura`. It's a bit of a pain point but can be worked around

> Also afaik it wasn't possible to avoid all clones (L347)

that comment is incorrect. I left it before I removed the unnecessary cloning. I believe I have successfully removed all clones

Erigara: > that comment is incorrect 

Indeed, good job then. Cool that it possible to avoid `Arc`ing. Feel free to resolve.

cli/src/torii/routing.rs
mversic: I think I resolved most concerns related to this

client/src/http_default.rs
```diff
@@ -25,8 +25,9 @@ fn header_name_from_str(str: &str) -> Result<HeaderName> {
 }
 
 /// Default request builder implemented on top of `attohttpc` crate.
+#[derive(Debug)]
 pub struct DefaultRequestBuilder {
-    inner: Result<AttoHttpRequestBuilder>,
+    pub inner: Result<AttoHttpRequestBuilder>,
```
appetrosyan, line 30: Suggest refactoring this later.

data_model/src/query/cursor.rs
```diff
@@ -0,0 +1,98 @@
...
+
+            if let Some(query_id) = candidate.query_id {
+                Ok(ForwardCursor {
+                    query_id: Some(query_id),
+                    cursor: candidate.cursor,
+                })
+            } else if candidate.cursor.is_some() {
+                Err(D::Error::custom("Cursor missing query id"))
+            } else {
+                Ok(ForwardCursor::default())
```
Arjentix, line 58: What's the meaning of default forward cusor? Shoulnd't it always be attached to some query?

data_model/src/query/cursor.rs
```diff
@@ -0,0 +1,98 @@
...
+        fn decode<I: Input>(input: &mut I) -> Result<Self, parity_scale_codec::Error> {
+            let candidate = ForwardCursorCandidate::decode(input)?;
+
+            if let Some(query_id) = candidate.query_id {
+                Ok(ForwardCursor {
+                    query_id: Some(query_id),
+                    cursor: candidate.cursor,
+                })
+            } else if candidate.cursor.is_some() {
+                println!("KITA: {:?}", candidate.cursor);
```
Arjentix, line 73: KITA?

data_model/src/query/pagination.rs
```diff
@@ -0,0 +1,60 @@
...
+)]
+pub struct Pagination {
+    /// limit of indexing
+    pub limit: Option<NonZeroU32>,
+    /// start of indexing
+    // TODO: Rename to offset
+    pub start: Option<NonZeroU64>,
+}
+
+impl From<Pagination> for Vec<(&'static str, NonZeroU64)> {
```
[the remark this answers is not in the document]
Arjentix: I see, it's not yours code initially, but I still wonder

cli/src/torii/cursor.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+impl<I: Iterator> Iterator for Batched<I> {
+    type Item = I::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next()
+    }
+}
```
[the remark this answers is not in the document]
mversic: yes, batching is something that should be opaque to the user

data_model/src/query/pagination.rs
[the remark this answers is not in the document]
mversic: for attaching query strings to URL in client

core/src/smartcontracts/isi/triggers/mod.rs
```diff
@@ -237,7 +247,7 @@ pub mod query {
 
     impl ValidQuery for FindTriggerKeyValueByIdAndKey {
```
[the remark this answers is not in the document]
mversic: I think so

client/tests/integration/asset.rs
```diff
@@ -218,19 +232,22 @@ fn client_add_asset_with_name_length_more_than_limit_should_not_commit_transacti
     iroha_logger::info!("Creating another asset");
     thread::sleep(pipeline_time * 4);
 
-    let asset_definition_ids = test_client
+    let mut asset_definition_ids = test_client
```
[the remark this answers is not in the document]
mversic: because it's iterator

data_model/src/query/cursor.rs
mversic: it's an empty/unused cursor. It has no query_id defined

cli/src/torii/mod.rs
```diff
@@ -27,10 +29,44 @@ use warp::{
...
+    }
+
+    // TODO: Add notifier channel to enable graceful shutdown
+    fn expired_query_cleanup(self: Arc<Self>, idle_time: Duration) -> tokio::task::JoinHandle<()> {
+        tokio::task::spawn(async move {
+            loop {
+                sleep(idle_time).await;
+
+                self.queries
+                    .retain(|_, (_, last_access_time)| last_access_time.elapsed() <= idle_time);
```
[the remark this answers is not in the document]
mversic: I don't understand you

appetrosyan (approved):

cli/src/torii/mod.rs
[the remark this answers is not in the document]
Arjentix: No objections. I think I've left this comment for previous version of code, which you've updated while I still was reviewing.

client/tests/integration/non_mintable.rs
```diff
@@ -34,7 +34,8 @@ fn non_mintable_asset_can_be_minted_once_but_not_twice() -> Result<()> {
     // We can register and mint the non-mintable token
     test_client.submit_transaction(&tx)?;
     test_client.poll_request(client::asset::by_account_id(account_id.clone()), |result| {
-        result.iter().any(|asset| {
+        let assets = result.collect::<QueryResult<Vec<_>>>().expect("Valid");
```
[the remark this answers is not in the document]
Arjentix: *Pain in the brain* time

data_model/src/query/pagination.rs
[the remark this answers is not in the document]
Arjentix: What I don't like about tuples is that in most of the times they're not readable. I'd make it a method.
But it's not a scope of this PR.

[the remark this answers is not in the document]
Arjentix: - Created https://github.com/hyperledger/iroha/issues/3740

data_model/src/query/cursor.rs
Arjentix: And why do we need this kind of unitialized cursor? Maybe the code that uses it should use `Option` instead?

cli/src/torii/routing.rs
```diff
@@ -232,15 +202,18 @@ async fn handle_pending_transactions(
...
-            .collect::<Vec<_>>(),
-    ))
+
+    let query_response = queue
+        .all_transactions(&wsv)
+        .map(Into::into)
+        .paginate(pagination)
+        .collect::<Vec<_>>();
+    // TODO:
+    //.batched(fetch_size)
```
Arjentix, line 214: I don't understand when it should be done. Consider adding description or create an issue

cli/src/torii/routing.rs
```diff
@@ -232,15 +202,18 @@ async fn handle_pending_transactions(
     sumeragi: SumeragiHandle,
     pagination: Pagination,
 ) -> Result<Scale<Vec<VersionedSignedTransaction>>> {
+    // TODO: Don't clone wsv here
```
Arjentix, line 205: Is it applicable here? It's transactions, so I don't think so

cli/src/torii/routing.rs
mversic: #3742

cli/src/torii/routing.rs
```diff
@@ -129,54 +83,70 @@ pub(crate) async fn handle_instructions(
...
-        let (valid_request, filter) = request.validate(&mut wsv)?;
-        let original_result = valid_request.execute(&wsv).map_err(ValidationFail::from)?;
-        (filter.filter(original_result), filter)
-    };
+    sorting: Sorting,
+    pagination: Pagination,
+
+    cursor: ForwardCursor,
+) -> Result<Scale<VersionedBatchedResponse<Value>>> {
+    // TODO: Remove wsv clone
```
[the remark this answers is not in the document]
mversic: ok, then there will be a conflict and we'll resolve the TODO, no?

cli/src/torii/routing.rs
mversic: it's pending transactions. Those are in the queue. Look at how `wsv` is used after this close, only by reference

appetrosyan (dismissed):

data_model/src/query/cursor.rs
mversic: well honestly, dealing with query strings in warp is a pain. It was rejecting my other attempts, this was a thing that I found works

Arjentix: Understandable

cli/src/torii/routing.rs
Arjentix: I see

Arjentix (approved):

appetrosyan (approved):

PR #3634, opened 2023-06-21
[refactor] #3615: Preserve wsv after validation

### LINKED ISSUE #3615: [refactor]: Implement Runtime Executor RFC
Link to the [RFC](https://wiki.hyperledger.org/display/iroha/Promote+Iroha+Runtime+Validator+into+Iroha+Runtime+Executor).

Erigara:
## Description

Reuse wsv created during validation.

### Linked issue

Partially closes #3615 

### Benefits

Improve performance by executing instructions only once.

```
validate_blocks/validate_blocks                                                                          
      time:   [705.06 ms 709.27 ms 713.55 ms]
      change: [-46.177% -45.775% -45.415%] (p = 0.00 < 0.10)
      Performance has improved.
```

### Checklist

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -133,7 +133,7 @@ pub struct PendingBlock {
 }
 
 /// Builder for `PendingBlock`
-pub struct BlockBuilder {
+pub struct BlockBuilder<'a> {
     /// Block's transactions.
     pub transactions: Vec<AcceptedTransaction>,
     /// Block's event recommendations.
@@ -145,12 +145,12 @@ pub struct BlockBuilder {
     /// The keypair used to sign this block.
     pub key_pair: KeyPair,
     /// The world state to be used when validating the block.
-    pub wsv: WorldStateView,
+    pub wsv: &'a mut WorldStateView,
 }
 
-impl BlockBuilder {
+impl BlockBuilder<'_> {
     /// Create a new [`PendingBlock`] from transactions.
-    pub fn build(mut self) -> PendingBlock {
+    pub fn build(self) -> PendingBlock {
         let timestamp = crate::current_time().as_millis();
         let height = self.wsv.height() + 1;
         let previous_block_hash = self.wsv.latest_block_hash();
@@ -171,7 +171,7 @@ impl BlockBuilder {
         let mut txs = Vec::new();
 
         for tx in self.transactions {
-            match transaction_validator.validate(tx, height == 1, &mut self.wsv) {
+            match transaction_validator.validate(tx, height == 1, self.wsv) {
                 Ok(transaction) => txs.push(TransactionValue {
                     tx: transaction,
                     error: None,
@@ -344,7 +344,7 @@ pub trait Revalidate: Sized {
     /// - When the block is deemed invalid.
     fn revalidate<const IS_GENESIS: bool>(
         &self,
-        wsv: WorldStateView,
+        wsv: &mut WorldStateView,
     ) -> Result<(), BlockRevalidationError>;
 
     /// Return whether or not the block contains transactions already committed.
@@ -356,7 +356,7 @@ pub trait Revalidate: Sized {
 pub trait InGenesis: Sized + Revalidate {
     /// # Errors
     /// - When the block is deemed invalid.
-    fn revalidate(&self, wsv: WorldStateView) -> Result<(), BlockRevalidationError> {
+    fn revalidate(&self, wsv: &mut WorldStateView) -> Result<(), BlockRevalidationError> {
         <Self as Revalidate>::revalidate::<true>(self, wsv)
     }
 }
@@ -366,7 +366,7 @@ pub trait InGenesis: Sized + Revalidate {
 pub trait InBlock: Sized + Revalidate {
     /// # Errors
     /// - When the block is deemed invalid.
-    fn revalidate(&self, wsv: WorldStateView) -> Result<(), BlockRevalidationError> {
+    fn revalidate(&self, wsv: &mut WorldStateView) -> Result<(), BlockRevalidationError> {
         <Self as Revalidate>::revalidate::<false>(self, wsv)
     }
 }
@@ -389,7 +389,7 @@ impl Revalidate for PendingBlock {
     #[allow(clippy::too_many_lines)]
     fn revalidate<const IS_GENESIS: bool>(
         &self,
-        mut wsv: WorldStateView,
+        wsv: &mut WorldStateView,
     ) -> Result<(), BlockRevalidationError> {
         let latest_block_hash = wsv.latest_block_hash();
         let block_height = wsv.height();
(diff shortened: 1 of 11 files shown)


SamHSmith (dismissed):

core/src/block.rs
```diff
@@ -133,7 +133,7 @@ pub struct PendingBlock {
 }
 
 /// Builder for `PendingBlock`
-pub struct BlockBuilder {
+pub struct BlockBuilder<'a> {
```
Arjentix, line 136: Doesn't clippy complain about one letter lifetime name? I'd better name it `world` as it's done in `wasm` module

core/src/kura.rs
```diff
@@ -115,7 +115,7 @@ impl Kura {
     /// - file storage is unavailable
     /// - data in file storage is invalid or corrupted
     #[iroha_logger::log(skip_all, name = "kura_init")]
-    pub fn init(&self) -> Result<Vec<HashOf<VersionedCommittedBlock>>> {
+    pub fn init(&self) -> Result<usize> {
```
Arjentix, line 118: Can you please add a doc comment to explain what does this return type mean?

core/src/wsv.rs
```diff
@@ -335,15 +335,34 @@ impl WorldStateView {
     /// - If timestamp conversion to `u64` fails
     #[iroha_logger::log(skip_all, fields(block_height))]
     pub fn apply(&mut self, block: &VersionedCommittedBlock) -> Result<()> {
```
Arjentix, line 337: Is it used anywhere after your changes?

Arjentix (dismissed):

core/src/kura.rs
0x009922: Or express it as a newtype, like `BlockCount(usize)`, so that it will be literally a self-documenting code.

core/src/sumeragi/mod.rs
```diff
@@ -205,33 +205,32 @@ impl SumeragiHandle {
...
-            let block_ref = kura.get_block_by_height(block_height).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
-            assert_eq!(
-                block_ref.hash(),
-                *block_hash,
-                "Kura init correctly reported the block hash."
-            );
-
-            wsv.apply(&block_ref)
+        let mut blocks_iter = (1..=block_count).map(|block_height| {
+            kura.get_block_by_height(block_height as u64).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.")
```
0x009922, line 214: `.expect()` error messages should express not the error that happened, but the condition or assumption that was **expected** to be valid.

I would rewrite it to

> Sumeragi should be able to load the block that was reported as presented. If not, the block storage was probably disconnected.

Not sure about using hints like "please check" is appropriate for panic messages. If advanced error handling is needed, it should be a `Result`, not `panic!` (not sure).

core/src/sumeragi/mod.rs
```diff
@@ -205,33 +205,32 @@ impl SumeragiHandle {
...
-        if !block_hashes.is_empty() {
-            let block_ref = kura.get_block_by_height(block_hashes.len() as u64).expect("Sumeragi could not load block that was reported as present. Please check that the block storage was not disconnected.");
+        // finalized_wsv is one block behind
+        let finalized_wsv = wsv.clone();
 
-            wsv.apply(&block_ref)
+        if let Some(latest_block) = blocks_iter.next() {
+            latest_block
+                .revalidate(&mut wsv)
+                .expect("Failed to revalidate block in init.");
```
0x009922, line 232:
```suggestion
                .expect("The block should be valid in init");
```

Or.. what assumption is expected? Why "init" assumes that revalidation should never fail?

0x009922 (approved):
LGTM, but there are possible minor improvements.

core/src/sumeragi/mod.rs
Erigara: > Why "init" assumes that revalidation should never fail?
 
Because we store only valid blocks in the block store.

mversic: Can we trust that block store cannot be tampered with?

Erigara: We can't but what best course of action would be in such case?

We panic so user have to manually inspect what happen in the block store (e.g. clear it).

core/src/wsv.rs
Erigara: We use it in some test where we don't validate block.

core/src/sumeragi/mod.rs
Erigara: I agree, implemented your changes.

core/src/sumeragi/mod.rs
appetrosyan: > Can we trust that block store cannot be tampered with?

Actually this is an involved question. 

1. It may be subject to bit-rot. We don't have a mechanism for counteracting this, but it's trivial to fix and detect: check if any of the blocks in the block-store has the wrong hash in an `async` task, and if so, refresh from other peers. 

2. The block could be tampered with maliciously. This could be either because a party gained access to the peer via a CVE, or because the peer administrator was corrupt. At this point all we need to do is convey to the peer admin that if they continue using an invalid block store that they'd lose money. If you can tamper with the block store, you can easily tamper with a build of a ledger to which you have the source code. 

### Corollary

The original message of "Failed to revalidate block at initialisation" is adequate.

I _would_, if possible, tell the user how to fix the block store, but given that it's now a single file with multiple blocks in binary, we'd need extra tooling or very good docs.

core/src/sumeragi/mod.rs
```diff
@@ -205,33 +205,33 @@ impl SumeragiHandle {
...
+        // finalized_wsv is one block behind
+        let finalized_wsv = wsv.clone();
 
-            wsv.apply(&block_ref)
+        if let Some(latest_block) = blocks_iter.next() {
+            latest_block
+                .revalidate(&mut wsv)
+                .expect("The block should be valid in init");
+            wsv.apply_without_execution(latest_block.as_ref())
                 .expect("Failed to apply block to wsv in init.");
```
0x009922, line 235:
```suggestion
                .expect("Block application in init should not fail");
```

0x009922 (dismissed):
LGTM, but more detailed explanation about `init` assumptions might help.

core/src/wsv.rs
```diff
@@ -335,15 +335,34 @@ impl WorldStateView {
...
+        self.execute_transactions(block.as_v1())?;
+        debug!("All block transactions successfully executed");
+
+        self.apply_without_execution(block)?;
+
+        Ok(())
+    }
+
+    /// Apply transactions without actually executing them.
+    /// It's assumed that block's transaction was already executed (as part of validation for example).
```
0x009922, line 347: I wonder if it is possible to express this assumption via type system. Just a thought.

0x009922 (approved):

SamHSmith (approved):

core/src/wsv.rs
Erigara: Maybe we can use newtype pattern or something.

PR #3645, opened 2023-06-24
[fix] #3516: make FindTrigger queries return original WASM

### LINKED ISSUE #3516: [suggestion] Allow querying for original WASM blobs/sources

QuentinI:
## Description

Retain original WASM blobs in `TriggerSet`
Substitute internal representation for original when querying for Triggers

### Linked issue

Closes #3516 

### Benefits

No more leaking internal representation
`FindTrigger` queries return actually useful WASM

### Checklist

### DIFF
--- a/core/src/smartcontracts/isi/triggers/set.rs
+++ b/core/src/smartcontracts/isi/triggers/set.rs
@@ -11,7 +11,6 @@
 use core::cmp::min;
 use std::{collections::HashMap, fmt};
 
-use iroha_crypto::HashOf;
 use iroha_data_model::{
     events::Filter as EventFilter,
     isi::error::{InstructionExecutionError, MathError},
@@ -50,6 +49,8 @@ pub struct Set {
     by_call_triggers: HashMap<TriggerId, LoadedAction<ExecuteTriggerEventFilter>>,
     /// Trigger ids with type of events they process
     ids: HashMap<TriggerId, EventType>,
+    /// Original [`WasmSmartContract`]s by [`TriggerId`] for querying purposes.
+    original_contracts: HashMap<TriggerId, WasmSmartContract>,
     /// List of actions that should be triggered by events provided by `handle_*` methods.
     /// Vector is used to save the exact triggers order.
     matched_ids: Vec<(Event, TriggerId)>,
@@ -77,6 +78,7 @@ impl Clone for Set {
             time_triggers: self.time_triggers.clone(),
             by_call_triggers: self.by_call_triggers.clone(),
             ids: self.ids.clone(),
+            original_contracts: self.original_contracts.clone(),
             matched_ids: Vec::default(),
         }
     }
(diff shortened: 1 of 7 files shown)


docs/source/references/schema.json
```diff
@@ -4845,10 +4844,6 @@
       {
         "name": "serialized",
         "type": "Vec<u8>"
-      },
-      {
-        "name": "blob_hash",
-        "type": "HashOf<WasmSmartContract>"
```
mversic, line 4851: this is good. I think that schema should also remove all references to `OptimizedExecutable`

data_model/src/query.rs
```diff
@@ -1073,15 +1073,15 @@ pub mod trigger {
     }
 
     impl Query for FindTriggerById {
-        type Output = Trigger<FilterBox, OptimizedExecutable>;
+        type Output = Trigger<FilterBox, Executable>;
```
mversic, line 1076:
```suggestion
        type Output = Trigger<FilterBox>;
```

I would prefer this if possible? I'm not sure though on how it's used exactly

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -326,6 +339,7 @@ impl Set {
     ///
     /// Return `false` if [`Set`] doesn't contain the trigger with the given `id`.
     pub fn remove(&mut self, id: &TriggerId) -> bool {
+        self.original_contracts.remove(id);
```
mversic, line 342: this is an interesting change? Was this a bug?

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -50,6 +49,8 @@ pub struct Set {
     by_call_triggers: HashMap<TriggerId, LoadedAction<ExecuteTriggerEventFilter>>,
     /// Trigger ids with type of events they process
     ids: HashMap<TriggerId, EventType>,
+    /// Original [`WasmSmartContract`]s by [`TriggerId`] for querying purposes.
+    original_contracts: HashMap<TriggerId, WasmSmartContract>,
```
mversic, line 53: while this may be fine atm, this will waste a lot of memory. A proper thing to do would be to store a reference to the block (block height actually) where this trigger is stored like this:

```suggestion
    original_contracts: HashMap<TriggerId, usize>,
```

mversic: also this would best be stored in `wsv` just like `WorldStateView::transactions`

docs/source/references/schema.json
QuentinI: I'm not actually sure how it still gets there

data_model/src/query.rs
QuentinI: AFAIU that would require refactoring how we handle optimized executables altogether. I'm not against it, but I feel like it should be a separate issue

docs/source/references/schema.json
mversic: you should remove `OptimizedExecutable` from `data_model` and move it into `iroha_core`

mversic: `OptimizedExecutable` should not be part of the public API. Everything in `data_model` is public API

PR #3652, opened 2023-06-26
[fix] #3352: Split up control flow and data messages

### LINKED ISSUE #3352: Split control flow from data processing in sumeragi
At the moment we have only one **control flow** message in `sumeragi`, namely `ViewChangeSuggested`, and it is processed alongside other **data** messages. This is not optimal for many reasons. One of the problems is that they share the same queue which introduces a lag in the processing of control flow messages due to which under high contention(lot of gossiping, etc.) nodes can stay out of sync for prolong periods of time

Additionally, control flow messages should also have priority in processing because some of the data messages waiting in the queue (that have arrived before the control message) could already be outdated.

This separation can be achieved with a separate queue for control flow messages or by changing the existing message queue into `PriorityQueue` where control flow messages have a higher priority than data messages

SamHSmith:
## Description

Create a second incomming packet queue inside sumeragi.

Closes #3352


core/src/sumeragi/main_loop.rs
```diff
@@ -130,28 +132,57 @@ impl Sumeragi {
...
+                packet.view_change_proofs,
+                &self.current_topology.sorted_peers,
+                self.current_topology.max_faults(),
+                self.wsv.latest_block_hash(),
+            ) {
+                trace!(%error, "Failed to add proofs into view change proof chain")
+            }
+
+            packet.message
+        })
```
appetrosyan, line 185: I see you've embraced the functional style. 

https://isocpp.org/blog/2014/11/functional-carmack

core/src/sumeragi/main_loop.rs
```diff
@@ -130,28 +132,57 @@ impl Sumeragi {
...
+                Ok(packet) => Some(packet),
+                Err(recv_error) => {
+                    match recv_error {
+                        mpsc::TryRecvError::Empty => None,
+                        mpsc::TryRecvError::Disconnected => {
+                            panic!("Sumeragi message pump disconnected. This is not a recoverable error.")
+                            // TODO: Use early return.
+                        }
+                    }
                 }
```
appetrosyan, line 170: This `match` pattern is equivalent to `map_err` followed by `ok()`.

core/src/sumeragi/main_loop.rs
```diff
@@ -130,28 +132,57 @@ impl Sumeragi {
...
+        let mut packet = if *control_message_in_a_row_counter < MAX_CONTROL_MSG_IN_A_ROW {
+            *control_message_in_a_row_counter += 1;
+            match self.control_message_receiver.try_recv() {
+                Ok(packet) => Some(packet),
+                Err(recv_error) => match recv_error {
+                    mpsc::TryRecvError::Empty => None,
+                    mpsc::TryRecvError::Disconnected => {
+                        panic!("Sumeragi control message pump disconnected. This is not a recoverable error.")
+                        // TODO: Use early return.
+                    }
```
appetrosyan, line 151: I don't think that this is accidentally the same as the exact same as the `packet` assignment later, all the way down to the identical `TODO`. You can extract the identical parts into a function.

core/src/sumeragi/main_loop.rs
```diff
@@ -33,6 +33,8 @@ pub struct Sumeragi {
     pub kura: Arc<Kura>,
     /// [`iroha_p2p::Network`] actor address
     pub network: IrohaNetwork,
+    /// Receiver channel, for control flow messages.
+    pub control_message_receiver: mpsc::Receiver<MessagePacket>,
```
mversic, line 37:
```suggestion
    pub control_message_receiver: mpsc::Receiver<ViewChangeSuggested>,
```

we want it to be a separate type not a variant of `Message`

core/src/sumeragi/main_loop.rs
```diff
@@ -130,28 +132,57 @@ impl Sumeragi {
...
-                if let Err(error) = view_change_proof_chain.merge(
-                    packet.view_change_proofs,
-                    &current_topology.sorted_peers,
-                    current_topology.max_faults(),
-                    self.wsv.latest_block_hash(),
-                ) {
-                    trace!(%error, "Failed to add proofs into view change proof chain")
-                }
-                Some(packet.message)
+    fn receive_network_packet(
```
mversic, line 135: while I think this change is a good start, we still have control messages processed in the same thread as data messages. Your thoughts?

Erigara:
Since we now separate queue for service messages could we use remove view change proof from every message?

core/src/sumeragi/main_loop.rs
```diff
@@ -130,28 +132,57 @@ impl Sumeragi {
...
+        &self,
+        view_change_proof_chain: &mut ProofChain,
+        control_message_in_a_row_counter: &mut usize,
+    ) -> Option<Message> {
+        const MAX_CONTROL_MSG_IN_A_ROW: usize = 25;
+
+        let mut packet = if *control_message_in_a_row_counter < MAX_CONTROL_MSG_IN_A_ROW {
```
Erigara, line 142: We need this counter to prevent choke of data message queue, right?

core/src/sumeragi/main_loop.rs
SamHSmith: Doing it in a seperate thread would be counter productive. The sumeragi thread runs as an event loop. These control flow messages are very easy to process and so they should be consumed basically instantly compared to the data messages.

core/src/sumeragi/main_loop.rs
SamHSmith: You are exactly correct.

SamHSmith:
> Since we now separate queue for service messages could we use remove view change proof from every message?

Not a bad idea actually, but I would like to do that in a seperate PR in that case.

core/src/sumeragi/main_loop.rs
SamHSmith: I'm trying to write code that will get merged faster.

core/src/sumeragi/main_loop.rs
SamHSmith: Not worth it. It's too small.

core/src/sumeragi/main_loop.rs
mversic: My concern here was that a data message could take long to process and thus prevent control flow messages from being processed ASAP. Is this a valid concern?

SamHSmith: > My concern here was that a data message could take long to process and thus prevent control flow messages from being processed ASAP. Is this a valid concern?

I don't think so because what would you otherwise do with a control flow message? If it is possible to send a data message that takes a long time to process that is a bug. Because it is a vector for a dos attack. Therefore we for instance check tx count against max allowed first.

So yes in a sense your concern is valid. But the real concern is not stopping control messages from being processed. But that it must be impossible to send any message that takes too long to process. If someone sends a message that makes the sumeragi thread grind to a halt, that is an incorrect message and it should have been rejected early.

mversic: > But that it must be impossible to send any message that takes too long to process. If someone sends a message that makes the sumeragi thread grind to a halt, that is an incorrect message and it should have been rejected early.

good point

appetrosyan (approved):
LGTM.

appetrosyan (dismissed):

core/src/sumeragi/message.rs
```diff
@@ -76,6 +76,22 @@ pub enum Message {
     ViewChangeSuggested,
 }
```
mversic, line 77:
```suggestion
}
```

we should not need this variant in data messages anymore

core/src/sumeragi/main_loop.rs
```diff
@@ -130,28 +132,55 @@ impl Sumeragi {
...
+                packet.view_change_proofs,
+                &self.current_topology.sorted_peers,
+                self.current_topology.max_faults(),
+                self.wsv.latest_block_hash(),
+            ) {
+                trace!(%error, "Failed to add proofs into view change proof chain")
+            }
+
+            packet.message
+        })
```
mversic, line 183:
```suggestion
        packet.map_or_else(|| {
            *control_message_in_a_row_counter = 0;
            packet = self
                .message_receiver
                .try_recv()
                .map_err(|recv_error| {
                    assert!(
                        recv_error != mpsc::TryRecvError::Disconnected,
                        "Sumeragi message pump disconnected. This is not a recoverable error."
                    )
                })
                .ok();
        }, |packet| {
            if let Err(error) = view_change_proof_chain.merge(
                packet.view_change_proofs,
                &self.current_topology.sorted_peers,
                self.current_topology.max_faults(),
                self.wsv.latest_block_hash(),
            ) {
                trace!(%error, "Failed to add proofs into view change proof chain")
            }

            packet.message
        })
```

core/src/sumeragi/mod.rs
```diff
@@ -186,7 +187,14 @@ impl SumeragiHandle {
...
     /// Deposit a sumeragi network message.
     pub fn incoming_message(&self, msg: MessagePacket) {
-        if let Err(error) = self.message_sender.try_send(msg) {
+        if matches!(&msg.message, Message::ViewChangeSuggested) {
+            if let Err(error) = self.control_message_sender.try_send(ControlFlowMessage {
+                view_change_proofs: msg.view_change_proofs,
+            }) {
+                self.metrics.dropped_messages.inc();
+                error!(?error, "This peer is faulty. Incoming control messages have to be dropped due to low processing speed.");
+            }
```
mversic, line 196: what happens at this point if a control message is dropped?

core/src/sumeragi/message.rs
mversic, line 77:
```suggestion
}
```

IMO this shouldn't be mixed with the data message anymore

core/src/sumeragi/mod.rs
SamHSmith: Consensus tolerates it unless this is happens on more that f peers in the topology 3f + 1

mversic: github will not allow me comment there but shouldn't [this](https://github.com/hyperledger/iroha/pull/3652/files#diff-900707b74d552f7c23161afeabbf4501209b25c4cb77c61b3aa178baf27a7c00R783) line then be:

```rust
*old_view_change_index = current_view_change_index;
```

SamHSmith: I think I broke your link by pushing. Can you find it again?

core/src/sumeragi/main_loop.rs
SamHSmith: I have taken this suggestion and gone a little further.

core/src/sumeragi/mod.rs
mversic: [link](https://github.com/hyperledger/iroha/pull/3652/files#diff-900707b74d552f7c23161afeabbf4501209b25c4cb77c61b3aa178baf27a7c00R795)

SamHSmith: No. Because then rotate_all would be called too few times on the topology.

PR #3659, opened 2023-06-28
[fix] #3543: Increment quantity change, not total

### LINKED ISSUE #3543: [bug]: `metric_tx_amounts` add total asset value and not increment
If we look at `tx_amounts` metrics definition: "average amount involved in a transaction on this peer"

In implementation of `Asset`-related instructions (`Mint`, `Burn`, `Transfer`) we always add `total` quantity value for `metrics_tx_amounts`. When it seems like should add only actual value being transferred, minted or burnt. 

For example:
- if user A had 10 apples and then mint 1 apple we should add 1 to `metrics_tx_amounts` (and currently we add 11);
- if user A had 10 apples and then burn 1 apple we should add 1 to `metrics_tx_amounts` (and currently we add 9);
- if user A had 10 apples and then transferred 1 apple to user B who already had 6 apples then we should add 1 to `metrics_tx_amounts` (an currently we add 7).

SamHSmith:
### Linked issue

Closes #3543

### DIFF
--- a/core/src/smartcontracts/isi/asset.rs
+++ b/core/src/smartcontracts/isi/asset.rs
@@ -213,7 +213,7 @@ pub mod isi {
                 &asset_id,
                 <Self as AssetInstructionInfo>::DEFAULT_ASSET_VALUE,
             )?;
-            let quantity = {
+            let new_quantity = {
                 let asset = wsv.asset_mut(&asset_id)?;
                 let quantity: &mut Self = asset
                     .try_as_mut()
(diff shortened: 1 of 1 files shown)


core/src/smartcontracts/isi/asset.rs
```diff
@@ -222,13 +222,12 @@ pub mod isi {
...
                     .ok_or(MathError::Overflow)?;
-                *quantity
+                mint.object
             };
 
-            // TODO: add only value actually being minted (https://github.com/hyperledger/iroha/issues/3543)
             #[allow(clippy::float_arithmetic)]
             {
-                wsv.metric_tx_amounts += quantity.into_metric();
+                wsv.metric_tx_amounts += new_quantity.into_metric();
```
Erigara, line 230: maybe we can then use just `mint.object` directly?

Erigara (approved):

core/src/smartcontracts/isi/asset.rs
SamHSmith: Maybe, but I think it's slightly easier to understand like this.

appetrosyan (approved):

Erigara:
I think we should also include how to test section for our QA team.

PR #3661, opened 2023-06-29
[feature] #3236: Enhance iroha_wasm_builder with cache, better errors and minor API changes

### LINKED ISSUE #3236: Cache `iroha_wasm_builder` output

Arjentix:
## Description

Now you can configure `iroha_wasm_builder` output dir in a three ways (in order of priority):

1. Explicitly use `out_dir()` method
2. Set `IROHA_WASM_BUILDER_OUT_DIR` env var
3. Do nothing if you are in a build-script. Directory will be inherited from `OUT_DIR` env var which is set by `cargo`

By default `target` directory of running crate will be used.

Other than that optimization *caching* is also implemented.
This is done by comparing `sha256` hashes of non-omptimized wasm-files. If hash before and after build are equal, then non-omptimized file wasn't changed, which means that optimized wasm file should also stay unchanged and we don't have to spent time on running `wasm-opt`.

Also `build()` will fail now if it's being called on workspace. I think, it was an error that we didn't have that before. It's an error because we run `build()` only to get `Output` from it, where `Output` is a binary encoded WASM file, which we can't choose if we build workspace.

### Linked issue

- Closes #3236 

### Benefits

Cached wasm binaries allows us not to rebuild them everytime from scratch. I.e. cached validator builds now take much less time. Same for tests which sumbit wasm.

### Checklist


tools/kagami/src/main.rs
```diff
@@ -703,13 +703,10 @@ mod validator {
...
 
     pub fn construct_validator() -> color_eyre::Result<Vec<u8>> {
-        let build_dir = tempfile::tempdir()
-            .wrap_err("Failed to create temp dir for runtime validator output")?;
-
         // FIXME: will it work when Kagami is run outside of the iroha dir?
         //        https://github.com/hyperledger/iroha/issues/3473
         let wasm_blob = iroha_wasm_builder::Builder::new("../../default_validator")
-            .out_dir(build_dir.path())
+            // .out_dir(build_dir.path())
```
Erigara, line 709: Commented code

configs/peer/validatator.wasm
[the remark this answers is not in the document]
Arjentix: Lol, indeed. Happened while testing

data_model/src/block.rs
```diff
@@ -315,24 +315,26 @@ mod committed {
...
+                        }
+                        .into()
+                    },
+                    |error| {
+                        PipelineEvent {
+                            entity_kind: PipelineEntityKind::Transaction,
+                            status: PipelineStatus::Rejected(error.clone().into()),
+                            hash: tx.payload().hash().into(),
+                        }
+                        .into()
```
Erigara, line 334: Nitpicking :) we can use additional `map(Into::into)` to transform pipeline event and do not repeat `.into()` for both branches

Erigara (approved):

data_model/src/block.rs
```diff
@@ -315,24 +315,19 @@ mod committed {
...
-                    }
-                    .into(),
-                    Some(error) => PipelineEvent {
-                        entity_kind: PipelineEntityKind::Transaction,
-                        status: PipelineStatus::Rejected(error.clone().into()),
-                        hash: tx.payload().hash().into(),
-                    }
-                    .into(),
-                });
+            let tx = block.transactions.iter().cloned().map(|tx| {
```
appetrosyan, line 318: DRY. Good

wasm_builder/src/lib.rs
```diff
@@ -63,7 +64,9 @@ impl<'path, 'out_dir> Builder<'path, 'out_dir> {
 
     /// Set smartcontract build output directory.
     ///
-    /// Default is `OUT_DIR` environment variable.
+    /// By default out directory will be taken from `IROHA_WASM_BUILDER_OUT_DIR` or
+    /// `OUT_DIR` environment variables (following this order).
+    /// If none of them is set, then `target` directory will be used.
```
appetrosyan, line 69:
```suggestion
    /// By default the output directory will be assigned either from `IROHA_WASM_BUILDER_OUT_DIR` or
    /// `OUT_DIR` environment variables, in this order.
    /// If neither is set, then the `target` directory will be used.
```

wasm_builder/src/lib.rs
```diff
@@ -96,16 +99,17 @@ impl<'path, 'out_dir> Builder<'path, 'out_dir> {
...
         self.into_internal()?.build()
     }
 
     fn into_internal(self) -> Result<internal::Builder<'out_dir>> {
+        let abs_path = Self::absolute_path(self.path)?;
         Ok(internal::Builder {
-            absolute_path: self.absolute_path()?,
+            absolute_path: abs_path
+                .canonicalize()
+                .wrap_err_with(|| format!("Failed to canonicalize path: {}", abs_path.display()))?,
```
appetrosyan, line 112: More details = better.

wasm_builder/src/lib.rs
```diff
@@ -115,24 +119,33 @@ impl<'path, 'out_dir> Builder<'path, 'out_dir> {
...
+            .or_else(|| env::var_os("OUT_DIR"))
+            .map(PathBuf::from)
+            .map_or_else(
+                || {
+                    const PATH: &str = "target";
+                    let abs_path = Self::absolute_path(PATH)?;
+                    std::fs::DirBuilder::new()
+                        .recursive(true)
+                        .create(&abs_path)
+                        .wrap_err("Failed to create directory")?;
```
appetrosyan, line 132:
```suggestion
                        .wrap_err_with(|| format!"Failed to create directory {abs_path}")?;
```

wasm_builder/src/lib.rs
```diff
@@ -280,16 +320,46 @@ impl Output {
...
+            file = self
+                .wasm_file
+                .file_stem()
+                .map_or_else(|| "output".into(), OsStr::to_string_lossy),
+            ext = self
+                .wasm_file
+                .extension()
+                .map_or_else(|| "wasm".into(), OsStr::to_string_lossy),
+        ));
+
```
appetrosyan, line 339: Th8s is perhaps the best usage of `format!` I've seen. Good job!

appetrosyan (approved):
Very clean code. Minor comments about comments

wasm_builder/src/lib.rs
```diff
@@ -280,16 +320,46 @@ impl Output {
...
-        let optimized_file = tempfile::NamedTempFile::new_in(self.out_dir.path())?
-            .path()
-            .to_owned();
+        let optimizer = wasm_opt::OptimizationOptions::new_optimize_for_size();
         optimizer.run(self.wasm_file, optimized_file.as_path())?;
 
         Ok(Self {
             wasm_file: optimized_file,
-            out_dir: self.out_dir,
+            previous_hash: self.previous_hash,
```
pesterev, line 362: I don't understand why we didn't return a new hash as `previous_hash` here. According to the condition above, the `previous_hash` equals `None` or an outdated hash, it isn't?
```rust
Ok(Self {
    wasm_file: optimized_file,
    previous_hash: new_hash, // or reuse `current_hash`
})
```

pesterev, line 362: I don't understand why we didn't return a new hash as previous_hash here. According to the condition above, the previous_hash equals None or an outdated hash, it isn't?
```rust
Ok(Self {
    wasm_file: optimized_file,
    previous_hash: new_hash, // or reuse `current_hash`
})
```

0x009922 (approved):
LGTM. Although I am confused by the "Cache `iroha_wasm_builder` output" title, because:

- it seems that only its optimisation output is cached
- PR contains changes in the builder API

I would rename to "Enhance `iroha_wasm_builder` with cache, better errors and minor API changes"

wasm_builder/src/lib.rs
Arjentix: Hm, you're right, that way we can also protect us from double optimizing in stupid cases like optimize()?.optimize()?.
I'll fix that

Arjentix:
> it seems that only its optimisation output is cached

Actually no. Build step caching is done automatically by `cargo`. What I did is removing temporary directory for build artifacts

wasm_builder/src/lib.rs
```diff
@@ -115,24 +119,33 @@ impl<'path, 'out_dir> Builder<'path, 'out_dir> {
...
-            .ok_or_else(|| eyre!("Expected `OUT_DIR` environment variable"))
-            .map(Into::into)
+        env::var_os("IROHA_WASM_BUILDER_OUT_DIR")
+            .or_else(|| env::var_os("OUT_DIR"))
+            .map(PathBuf::from)
+            .map_or_else(
+                || {
+                    const PATH: &str = "target";
+                    let abs_path = Self::absolute_path(PATH)?;
+                    std::fs::DirBuilder::new()
```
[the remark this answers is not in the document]
Arjentix: Yeah, cool thing from `std`. I also didn't know about it before I start searching on how to create directory recursively

appetrosyan (approved):

Erigara (approved):

PR #3675, opened 2023-07-04
[feature] #3383: Implement a macro that parses a socket address at compile time

### LINKED ISSUE #3383: [suggestion] Function-like proc-macro for parsing sockets
### Feature request

We want something like 
```rust
socket!(127.0.0.1:8080);
```

### Motivation

Can't do with `macro_rules!` because `127.0` parses to a float. 

### Who can help?

@ilchu

DCNick3:
### Linked issue

Closes #3383 

### Limitations

This can't replace all uses of the existing `socket_addr!` macro, as it parses only literal addresses. Some of the `socket_addr!` usages use an expression as port number: https://github.com/hyperledger/iroha/blob/19984d91c2add850a197050e6cbee52af1bc6fa2/core/test_network/src/lib.rs#L598. Supporting this is possible, but would require a more complicated parser.

As 5/11 usages of the original macro use it, it probably doesn't make sense to merge it without support for variable ports

It also doesn't support hostname addresses, but they are not used anywhere except in the doc-test for the `socket_addr!` macro. Probably fine going without it, but should be relatively easy to implement if needed.

### Checklist

- [x] Parse IPv4 and IPv6 addresses
- [x] Allow usage of arbitrary expressions in port position
- [x] Replace old usages of `socket_addr!` macro


primitives/derive/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+mod socket;
+
+#[proc_macro]
+pub fn socket(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
Arjentix, line 4: Please, add a doc comment here. Also some doc-tests will be good under `Examples` section

primitives/derive/src/socket.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn smoke() {
```
Arjentix, line 77: Interesting name, but I think that all `tests` modules contain *smoke* tests. I don't think there is much sence in this naming. We prefer to keep test names in the style of what should happen.

So I suggest to rename it to `parse_ipv4`. Same for `v6`

primitives/derive/src/socket.rs
```diff
@@ -0,0 +1,118 @@
...
+                        port: #port
+                    }
+                )
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Arjentix, line 73: Btw nice practice to have tests for proc macros

primitives/derive/src/socket.rs
```diff
@@ -0,0 +1,118 @@
...
+    result
+}
+
+pub fn socket_impl(input: TokenStream) -> TokenStream {
+    let input = stringify_tokens(input);
+
+    let addr = match input.parse::<SocketAddr>() {
+        Ok(addr) => addr,
+        Err(e) => {
+            let message = format!("Failed to parse {:?}: {}", input, e);
```
Arjentix, line 37:
```suggestion
            let message = format!("Failed to parse `{:?}`: {}", input, e);
```

I prefer to isolate user input.

Arjentix (commented):
Congratulations with your first PR! Nice!

In PR description you say that custom port is not supported. But by looking at the code it looks like it is supported, no?

Cargo.toml
```diff
@@ -30,6 +30,7 @@ members = [
   "derive_primitives",
   "genesis",
   "primitives",
+  "primitives/derive",
```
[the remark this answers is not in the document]
DCNick3: AFAIU:

`derive_primitives` is primitives for derive
`primitives_derive` is derives for primitives

It's... A bit confusing, but I think it makes sense?

It was suggested by Aleksandr (https://github.com/hyperledger/iroha/issues/3383#issuecomment-1619733411)...

DCNick3:
> In PR description you say that custom port is not supported. But by looking at the code it looks like it is supported, no?

I forgot the link to an example..

https://github.com/hyperledger/iroha/blob/19984d91c2add850a197050e6cbee52af1bc6fa2/core/test_network/src/lib.rs#L598

Basically, it uses an arbirary expression instead of an integer literal for port. This won't work with the "stringify and parse it" approach

primitives/derive/src/socket.rs
DCNick3: It's already using the `Debug` formatting, but it will have double quotes not back ticks. 

Using `Display` with backticks should be fine

primitives/derive/src/socket.rs
```diff
@@ -0,0 +1,118 @@
...
+        match token_tree {
+            TokenTree::Group(g) => {
+                let inner = stringify_tokens(g.stream());
+
+                let bracketed = match g.delimiter() {
+                    Delimiter::Parenthesis => format!("({})", inner),
+                    Delimiter::Brace => format!("{{{}}}", inner),
+                    Delimiter::Bracket => format!("[{}]", inner),
+                    Delimiter::None => inner,
+                };
```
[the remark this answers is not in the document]
DCNick3: Not really.. It's "without inserting any spaces in between" that is important

By default it produces a bunch of spaces between tokens, looking something like `127.0 . 0.1 : 8080`.

primitives/derive/Cargo.toml
Arjentix, line 1: Add an empty line at the end, please

primitives/derive/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+mod socket_addr;
+
+/// Convenience macro to concisely construct a [`SocketAddr`]
```
Arjentix, line 3: I'm afraid this won't resolve. You can try by running `cargo doc --no-deps`. Although we have a lot of unresolved links we shouldn't make more. And we don't want to add `iroha_primitives` dependency just for that. So I suggest this:

```suggestion
/// Convenience macro to concisely construct a `SocketAddr`
```

Arjentix:
So what blocks you to undraft this PR?

DCNick3:
> So what blocks you to undraft this PR?

Using expressions as port numbers, which is done a lot throughout the codebase. Will probably tackle this tomorrow

primitives/derive/src/lib.rs
DCNick3: Yeah, fair enough

DCNick3: Hm... There is actually a way to make it resolve:

When re-exporting the macro in `iroha_primitives`, one can force the re-export to be inline and write the docs there. Then the macro will be documented in `iroha_primitives` crate with a resolved `SocketAddr` link. It doesn't seem to work for intellij-rust though, only shows up in rustdoc. Not sure if it's worth it.

```rust
/// Convenience macro to concisely construct a [`SocketAddr`]
///
/// # Examples
/// ```
/// # use iroha_primitives::addr::socket_addr2;
/// let localhost = socket_addr2!(127.0.0.1:8080);
/// let remote = socket_addr2!([2001:db8::1]:8080);
/// ```
#[doc(inline)]
pub use iroha_primitives_derive::socket_addr as socket_addr2;
```

0x009922 (commented):
LGTM. Some comments:

- I wonder, which syntax would you use for expressions interpolation? I have a couple of ideas:

  ```rust
  // format-style?
  socket_addr!("127.0.0.1:{}", port);
  
  // something custom?
  socket_addr!(127.0.0.1:~port);
  ```

- Will you replace all usages of the old macro?

0x009922:
Also, please make meaningful checklist in the PR's description (the current one is template for external contributors, AFAIC). It might make sense to move some TODOs from your description/code there.

DCNick3:
> I wonder, which syntax would you use for expressions interpolation? I have a couple of ideas:

I _think_ it's possible to parse the natural syntax: `socket_addr!(127.0.0.1:port)`. Everything before the colon token is converted to string and parsed as an IP address, while everything after the colon would be parsed as an expression.

> Will you replace all usages of the old macro?

Yes, this is what I am planning to do

DCNick3:
`rustfmt` wants `socket_addr!(127.0.0.1: port);` to be formatted as `socket_addr!(127.0.0 .1: port);`...

There is an option to exclude a certain macro invocations from being formatted, but 

1) the config option parsing seems broken https://github.com/rust-lang/rustfmt/issues/5816
2) the toolchain used in CI doesn't support it

I guess we would have to live with this ugly formatting for some time?

Arjentix (approved):

0x009922 (approved):

PR #3685, opened 2023-07-06
[refactor]: move instruction len from data_model into core

mversic:
## Description

the less functionality we expose in public API the better

### Linked issue

Closes #{issue_number} 

### Checklist

### DIFF
--- a/client/tests/integration/events/pipeline.rs
+++ b/client/tests/integration/events/pipeline.rs
@@ -116,7 +116,14 @@ fn committed_block_must_be_available_in_kura() {
         .expect("Failed to submit transaction");
 
     let event = event_iter.next().expect("Block must be committed");
-    let Ok(Event::Pipeline(PipelineEvent { entity_kind: PipelineEntityKind::Block, status: PipelineStatus::Committed, hash })) = event else { panic!("Received unexpected event") };
+    let Ok(Event::Pipeline(PipelineEvent {
+        entity_kind: PipelineEntityKind::Block,
+        status: PipelineStatus::Committed,
+        hash,
+    })) = event
+    else {
+        panic!("Received unexpected event")
+    };
     let hash = HashOf::from_untyped_unchecked(hash);
 
     peer.iroha
--- a/core/src/smartcontracts/isi/mod.rs
+++ b/core/src/smartcontracts/isi/mod.rs
@@ -443,7 +443,7 @@ pub mod prelude {
 mod tests {
     #![allow(clippy::restriction)]
 
-    use core::str::FromStr;
+    use core::str::FromStr as _;
     use std::sync::Arc;
 
     use iroha_crypto::KeyPair;
--- a/core/src/smartcontracts/isi/query.rs
+++ b/core/src/smartcontracts/isi/query.rs
@@ -100,7 +100,9 @@ mod tests {
     use std::str::FromStr as _;
 
     use iroha_crypto::{Hash, HashOf, KeyPair};
-    use iroha_data_model::{block::VersionedCommittedBlock, transaction::TransactionLimits};
+    use iroha_data_model::{
+        block::VersionedCommittedBlock, query::error::FindError, transaction::TransactionLimits,
+    };
     use once_cell::sync::Lazy;
 
     use super::*;
@@ -384,7 +386,10 @@ mod tests {
             .sign(ALICE_KEYS.clone())?;
         let wrong_hash = unapplied_tx.hash();
         let not_found = FindTransactionByHash::new(wrong_hash).execute(&wsv);
-        assert!(not_found.is_err());
+        assert!(matches!(
+            not_found,
+            Err(Error::Find(FindError::Transaction(_)))
+        ));
 
         let found_accepted = FindTransactionByHash::new(va_tx.hash()).execute(&wsv)?;
         if found_accepted.transaction.error.is_none() {
--- a/core/src/tx.rs
+++ b/core/src/tx.rs
@@ -45,6 +45,34 @@ pub enum AcceptTransactionFail {
     SignatureVerification(#[source] SignatureVerificationFail<TransactionPayload>),
 }
 
+fn instruction_size(isi: &InstructionBox) -> usize {
+    use InstructionBox::*;
+
+    match isi {
+        Register(isi) => isi.object.len() + 1,
+        Unregister(isi) => isi.object_id.len() + 1,
+        Mint(isi) => isi.destination_id.len() + isi.object.len() + 1,
+        Burn(isi) => isi.destination_id.len() + isi.object.len() + 1,
+        Transfer(isi) => isi.destination_id.len() + isi.object.len() + isi.source_id.len() + 1,
+        If(isi) => {
+            let otherwise = isi.otherwise.as_ref().map_or(0, instruction_size);
+            isi.condition.len() + instruction_size(&isi.then) + otherwise + 1
+        }
+        Pair(isi) => {
+            instruction_size(&isi.left_instruction) + instruction_size(&isi.right_instruction) + 1
+        }
+        Sequence(isi) => isi.instructions.iter().map(instruction_size).sum::<usize>() + 1,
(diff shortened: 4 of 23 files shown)


wasm_codec/derive/src/lib.rs
mversic, line 1: this file contains only formatting changes because I used newer compiler version

client/tests/integration/events/pipeline.rs
```diff
@@ -116,7 +116,14 @@ fn committed_block_must_be_available_in_kura() {
...
     let event = event_iter.next().expect("Block must be committed");
-    let Ok(Event::Pipeline(PipelineEvent { entity_kind: PipelineEntityKind::Block, status: PipelineStatus::Committed, hash })) = event else { panic!("Received unexpected event") };
+    let Ok(Event::Pipeline(PipelineEvent {
+        entity_kind: PipelineEntityKind::Block,
+        status: PipelineStatus::Committed,
+        hash,
+    })) = event
+    else {
+        panic!("Received unexpected event")
+    };
```
appetrosyan, line 126: This looks like a manual `expect` implementation.

core/src/smartcontracts/isi/mod.rs
```diff
@@ -40,6 +40,125 @@ pub trait Registrable {
...
+            Fail(fail_box) => fail_box.len(),
+            SetKeyValue(set_key_value) => set_key_value.len(),
+            RemoveKeyValue(remove_key_value) => remove_key_value.len(),
+            Grant(grant_box) => grant_box.len(),
+            Revoke(revoke_box) => revoke_box.len(),
+            ExecuteTrigger(execute_trigger) => execute_trigger.len(),
+            SetParameter(set_parameter) => set_parameter.len(),
+            NewParameter(new_parameter) => new_parameter.len(),
+            Upgrade(upgrade_box) => upgrade_box.len(),
+        }
```
appetrosyan, line 65: https://docs.rs/syntactic-for/latest/syntactic_for/index.html

Might be useful here.

core/src/smartcontracts/isi/mod.rs
```diff
@@ -40,6 +40,125 @@ pub trait Registrable {
...
+    }
+}
+
+impl ValidInstruction for SetKeyValueBox {
```
appetrosyan, line 69: Probably better done in a macro.

appetrosyan: Also, do we really need this trait? Can't we have a single `fn` with a top-level match statement? 

Because we're not really using dynamic dispatch (even though if we did, LLVM devirtualiser would still emplace the function calls), and as mentioned above, we're not really checking that the correct `len` is implemented.

core/src/smartcontracts/isi/mod.rs
appetrosyan: Also we ought to have some a way to differentiate `len` implementations. Right now, there's no telling if this is `ExactSizeIterator::len`, `ExecuteTrigger::len` or `<ExecuteTrigger as ValidInstruction>::len`. 

Maybe calling the function something else, or even using `dyn ValidInstruction`. Or even better inline the implementation as a `match` branch.

core/src/smartcontracts/isi/query.rs
```diff
@@ -384,7 +384,7 @@ mod tests {
             .sign(ALICE_KEYS.clone())?;
         let wrong_hash = unapplied_tx.hash();
         let not_found = FindTransactionByHash::new(wrong_hash).execute(&wsv);
-        assert!(matches!(not_found, Err(_)));
+        assert!(not_found.is_err());
```
appetrosyan, line 387: Another option is to set the exact `Err` type. Given that these errors are user facing, I'd hard-code the error type into the `matches!` invocation.

core/src/smartcontracts/mod.rs
```diff
@@ -26,6 +28,13 @@ pub trait Execute {
     fn execute(self, authority: &AccountId, wsv: &mut WorldStateView) -> Result<(), Error>;
 }
 
+/// This trait should be implemented for all Iroha Instructions.
+#[allow(clippy::len_without_is_empty)]
```
appetrosyan, line 32:
```suggestion

```

And implement `is_empty()`. 

OR, better yet, call it something else, like `instruction_size` (semantically, this is a potentially heterogeneous structure, so it makes more sense to call it size).

core/src/sumeragi/main_loop.rs
```diff
@@ -1167,6 +1167,7 @@ fn handle_block_sync(
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::redundant_clone)]
```
appetrosyan, line 1170: What's triggering the presumed false-positive?

data_model/src/isi.rs
```diff
@@ -135,124 +132,27 @@ pub mod model {
...
+    impl Instruction for MintBox {}
+    impl Instruction for BurnBox {}
+    impl Instruction for TransferBox {}
+    impl Instruction for GrantBox {}
+    impl Instruction for RevokeBox {}
+    impl Instruction for SetParameterBox {}
+    impl Instruction for NewParameterBox {}
+    impl Instruction for UpgradeBox {}
+    impl Instruction for ExecuteTriggerBox {}
+    impl Instruction for FailBox {}
```
appetrosyan, line 150: Probably useful to use `syntactic_for` here.

wasm_codec/derive/src/lib.rs
appetrosyan: Which version? We should want to update CI.

mversic: the one I saw in `Dockerfile`

client/tests/integration/events/pipeline.rs
mversic: 1. it's also refuting all event patterns that are not pipeline
2. it's in tests

data_model/src/isi.rs
mversic: interesting crate, after my update I don't think it's useful in this case

appetrosyan (approved):

appetrosyan:
`TRYBUILD=overwrite` and it should work.

appetrosyan (dismissed):

appetrosyan (approved):

core/src/tx.rs
```diff
@@ -45,6 +45,34 @@ pub enum AcceptTransactionFail {
...
+        Sequence(isi) => isi.instructions.iter().map(instruction_size).sum::<usize>() + 1,
+        SetKeyValue(isi) => isi.object_id.len() + isi.key.len() + isi.value.len() + 1,
+        RemoveKeyValue(isi) => isi.object_id.len() + isi.key.len() + 1,
+        Grant(isi) => isi.object.len() + isi.destination_id.len() + 1,
+        Revoke(isi) => isi.object.len() + isi.destination_id.len() + 1,
+        SetParameter(isi) => isi.parameter.len() + 1,
+        NewParameter(isi) => isi.parameter.len() + 1,
+        Upgrade(isi) => isi.object.len() + 1,
+        Fail(_) | ExecuteTrigger(_) => 1,
+    }
```
0x009922, line 73: 1. What does magic $+ 1$ means? Small comment or self-descriptive const here might help to understand it.
2. I see that $+ 1$ is applied for each instruction. Probably, it would be cleaner:
   ```rs
   let base_size = match isi { /* ... */ };
   base_size + 1
   ```

core/src/tx.rs
```diff
@@ -45,6 +45,34 @@ pub enum AcceptTransactionFail {
     SignatureVerification(#[source] SignatureVerificationFail<TransactionPayload>),
 }
 
+fn instruction_size(isi: &InstructionBox) -> usize {
```
0x009922, line 48: On one hand I feel that it would be a more composable way to re-write it as a trait (`impl InstructionSize for InstructionBox`, `impl InstructionSize for RegisterBox` etc). On the other hand, this function looks quite minimal and just works, why not.

0x009922 (approved):

core/src/tx.rs
mversic: opened an issue #3726

PR #3686, opened 2023-07-07
[fix] #3607: Do not allow decoding/deserialization of `SignedTransaction` in `no_std`

### LINKED ISSUE #3607: [bug] `SignedTransaction` and `SignedTransactionCandidate` should not implement `Decode` / `Deserialize` in `no_std`
Originally posted by @mversic [here](https://github.com/hyperledger/iroha/pull/3588#discussion_r1228115374)

# How to test

Honestly I don't think it's worth spending time on tests like this.

## Before changes

Currently it's possible to create a Rust package which will link to `iroha_data_model` with `default-features = false` enabled. That will disable `std` feature. After that you should be able to decode / deserialize `SignedTransaction` with wrong signatures, because they won't be checked at `no_std`

## After changes

You should get a compile error if you try to decode / deserialize `SignedTransaction` from `no_std` environment.

DCNick3:
No way to validate signatures (ursa is not available) - no way to enforce the "signatures are valid" invariant

Better to not support the decoding at all

IMO it still makes sense to decode `SignedTransactionCandidate`, as it does not carry invariants (at least I think so?)

Closes #3607


DCNick3:
> Transaction validation failed reason=Validation failed caused_by=Some(InternalError("Other"))

So, ugh... How do you debug a validator?

appetrosyan:
Rebuild Validator and see if it fixes the problem.

DCNick3:
It got worse. It failed only one test before probably because that test was actually re-building the validator

pesterev (approved):

PR #3698, opened 2023-07-11
Displaydoc

### LINKED ISSUE #3374: Reduce repetition in `derive(Error)` docstring and error message
If the docstring is just text and the format of the error is too, just use the doc-string in both places. If the error `Display` impl has a format string,  the doc-string could be concatenated to the doc-string. E.g. 
```rust
/// Too many inputs 
#[error("{inputs}, maximum: {maximum}")]
TooManyInputs
```
which displays as  `Too many inputs. 3 maximum: 2`

DCNick3:
## Description

Use `displaydoc` instead of `derive_more`'s `Display` derive or `thiserror`'s `error` attribute.

This unifies error's doc-comments and `Display` implementation, as they are often repeated/convey the same information.

Sometimes `#[ignore_extra_doc_attributes]` is used. In this case only the first line of the doc-comment is used for the `Display`, leaving the others only for documentation

### Linked issue

Closes #3374

### Checklist

- [x] `iroha_promitives`
- [x] `iroha_data_model`
- [ ] `iroha_p2p`
- [ ] `iroha_config`
- [ ] `iroha_core`
- [ ] `iroha`
- [ ] make sure all tests pass


data_model/src/isi.rs
```diff
@@ -936,50 +936,42 @@ pub mod error {
             Encode,
             IntoSchema,
         )]
+        #[ignore_extra_doc_attributes]
```
0x009922, line 939: Why is it needed here? It seems that the doc of `InstructionExecutionError` is a single line already.

If it is needed for `::InvariantViolation` variant, isn't it possible to set the attribute for the variant only?

data_model/src/isi.rs
```diff
@@ -1096,34 +1080,29 @@ pub mod error {
             IntoSchema,
         )]
         // TODO: Only temporarily opaque because of InstructionExecutionError::BinaryOpIncompatibleNumericValueTypes
+        #[ignore_extra_doc_attributes]
```
0x009922, line 1083: Same as my above comment.

data_model/src/isi.rs
```diff
@@ -1198,15 +1171,16 @@ pub mod error {
             Encode,
             IntoSchema,
         )]
+        #[ignore_extra_doc_attributes]
```
0x009922, line 1174: Same as my above comment.

data_model/src/query.rs
```diff
@@ -1520,48 +1515,33 @@ pub mod error {
...
+            /// Failed to find permission token
             PermissionToken(PermissionTokenFindError),
-            /// Failed to find specified [`Parameter`] variant.
-            #[display(fmt = "Failed to find specified parameter variant: `{_0}`")]
+            /// Parameter with id `{0}` not found
             Parameter(ParameterId),
-            /// Failed to find [`PublicKey`] at some account.
-            #[display(fmt = "Failed to find public key: `{_0}`")]
+            /// Failed to find [`PublicKey`]: `{0}`
             PublicKey(PublicKey),
```
0x009922, line 1545: **Inconsistency**: for `::PermissionToken` you deleted doc links (<code>[\`PermissionToken\`]</code>, <code>[\`AccountId\`]</code>), but you preserved it for `::PublicKey` (<code>[\`PublicKey\`]</code>).

So, there are two options (as I see):

1. Remove doc links completely. It would be more consistent with the rest of the code (in other errors you have removed it). 
2. Return doc links for `::PermissionToken`, and possibly in other places as well. Since doc links do not make sense for `displaydoc` (AFAIU), they should be placed below the first doc line.

I would prefer the 1st option. IMO, we should either write docs for `cargo doc` or for `displaydoc`, not for both. Otherwise, it will lead to dirt. With this PR errors' doc strings will be used explicitly for `Display`, not for API documentation, because it uses formatting that doesn't make sense for `cargo doc`.

primitives/src/fixed.rs
```diff
@@ -140,26 +140,20 @@ impl Fixed {
...
     Overflow,
     /// Division by zero
-    #[display(fmt = "Division by zero")]
     DivideByZero,
     /// Domain violation. E.g. computing `sqrt(-1)`
-    #[display(fmt = "Domain violation")]
     DomainViolation,
-    /// Arithmetic
-    #[display(fmt = "Unknown Arithmetic error")]
+    /// Unknown Arithmetic error
```
0x009922, line 156: Why capitalised*?

```suggestion
    /// Unknown arithmetic error
```

<sup>*I am not trying to offend matematicians</sup>

0x009922 (commented):
LGTM in general.

Would be great to get reviews from our technical writers, imo. They might help to make error messages more comprehensible (or at least grammatically correct), thus, make life easier for both developers and users.

data_model/src/isi.rs
DCNick3: `displaydoc` doesn't support this =(


`#[ignore_extra_doc_attributes]` only works when placed on the whole type

data_model/src/query.rs
DCNick3: Yeah, the doc links definitely do not make sense for the `displaydoc`

With this in mind, I also lean towards the first option. Losing the links is kinda sad, but if we put them at the second line we would be back to square one with duplication between docs and error messages

client/Cargo.toml
```diff
@@ -37,6 +37,7 @@ serde = { version = "1.0.151", features = ["derive"] }
 serde_json = "1.0.91"
 base64 = "0.13.1"
 thiserror = "1.0.38"
+displaydoc = "0.2.4"
```
Arjentix, line 40: What do you think about making it a workspace depenency? We already have an issue for that, so I think we can start adding new depenencies as workspace's.

That way you won't have to specify all options like version everytime

config/base/src/lib.rs
```diff
@@ -395,6 +408,17 @@ pub mod derive {
             }
         }
     }
+
+    #[test]
+    fn unknown_field_fmt() {
```
Arjentix, line 413: Nice

config/base/src/lib.rs
```diff
@@ -395,6 +408,17 @@ pub mod derive {
...
+        assert_eq!(
+            format!(
+                "{}",
+                Error::UnknownField(Field(vec!["a".into(), "b".into()]))
+            ),
```
Arjentix, line 418:
```suggestion
                Error::UnknownField(Field(vec!["a".into(), "b".into()])).to_string()
```

core/src/kura.rs
```diff
@@ -771,37 +771,28 @@ impl BlockStore<Locked> {
...
+#[derive(thiserror::Error, Debug, displaydoc::Display)]
 pub enum Error {
-    /// Generic IO error
-    #[error("Failed reading/writing {_1:?} from disk")]
+    /// Failed reading/writing {1:?} from disk
     IO(#[source] std::io::Error, PathBuf),
-    /// Failed to create the directory
-    #[error("Failed to create the directory {_1:?}")]
+    /// Failed to create the directory {1:?}
     MkDir(std::io::Error, PathBuf),
```
Arjentix, line 779: This will probably omit underlying error in display impl. Can you, please, refactor that? Something like: 

```rust
MkDr {
    source: std::io::Error,
    path: PathBuf,
}
```

Arjentix: `thiserror` should be smart enough to handle `source` field, but better to check the docs

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -572,19 +571,16 @@ impl From<LoadedExecutable> for OptimizedExecutable {
 }
 
 /// [`Set::mod_repeats()`] error
-#[derive(Debug, Clone, thiserror::Error)]
+#[derive(Debug, Clone, thiserror::Error, displaydoc::Display)]
 pub enum ModRepeatsError {
-    /// Trigger not found error
-    #[error("Trigger with id = {0} not found")]
+    /// Trigger with id = {0} not found
```
Arjentix, line 576:
```suggestion
    /// Trigger with id = `{0}` not found
```

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -572,19 +571,16 @@ impl From<LoadedExecutable> for OptimizedExecutable {
...
     /// Trigger repeats count overflow error
-    #[error("{0}")]
     RepeatsOverflow(#[from] RepeatsOverflowError),
 }
 
-/// Trigger repeats count overflow error
-#[derive(Debug, Copy, Clone, thiserror::Error)]
-#[error("Trigger repeats count overflow")]
+#[derive(Debug, Copy, Clone, thiserror::Error, displaydoc::Display)]
+/// Trigger repeats count overflow
```
Arjentix, line 583:
```suggestion
/// Trigger repeats count overflow
#[derive(Debug, Copy, Clone, thiserror::Error, displaydoc::Display)]
```

core/src/smartcontracts/wasm.rs
```diff
@@ -74,42 +74,36 @@ pub mod error {
...
         /// Export which should always be present is missing
-        #[error("Mandatory export error")]
         MandatoryExport(#[from] ExportError),
     }
 
-    /// Export error
-    #[derive(Debug, Copy, Clone, thiserror::Error)]
-    #[error("Failed to export `{export_name}`")]
+    #[derive(Debug, Copy, Clone, thiserror::Error, displaydoc::Display)]
+    /// Failed to export `{export_name}`
```
Arjentix, line 106:
```suggestion
    /// Failed to export `{export_name}`
    #[derive(Debug, Copy, Clone, thiserror::Error, displaydoc::Display)]
```

core/src/smartcontracts/wasm.rs
```diff
@@ -173,16 +163,13 @@ pub mod error {
...
 
     /// Exported function call error
-    #[derive(Debug, thiserror::Error)]
+    #[derive(Debug, thiserror::Error, displaydoc::Display)]
     pub enum ExportFnCallError {
         /// Failed to execute something on the host side
-        #[error("Failed to execute operation on host")]
         HostExecution(#[source] eyre::Report),
         /// Stack overflow, heap overflow or other limits exceeded
-        #[error("Execution limits exceeded")]
```
Arjentix, line 182: I'd better keep `Execution limits exceeded`

core/src/smartcontracts/wasm.rs
```diff
@@ -173,16 +163,13 @@ pub mod error {
     }
 
     /// Exported function call error
-    #[derive(Debug, thiserror::Error)]
+    #[derive(Debug, thiserror::Error, displaydoc::Display)]
     pub enum ExportFnCallError {
         /// Failed to execute something on the host side
-        #[error("Failed to execute operation on host")]
```
Arjentix, line 179: Same. I suggest to keep previous error message

primitives/src/fixed.rs
```diff
@@ -140,26 +140,20 @@ impl Fixed {
...
+    /// Failed to produce fixed point number
     Conversion(#[cfg_attr(feature = "std", source)] fixnum::ConvertError),
     /// Overflow
-    #[display(fmt = "Overflow")]
     Overflow,
     /// Division by zero
-    #[display(fmt = "Division by zero")]
     DivideByZero,
     /// Domain violation. E.g. computing `sqrt(-1)`
-    #[display(fmt = "Domain violation")]
```
Arjentix, line 159: Keep previous

primitives/src/addr.rs
```diff
@@ -17,7 +17,7 @@ use serde_with::{DeserializeFromStr, SerializeDisplay};
 use crate::{conststr::ConstString, ffi};
 
 #[cfg_attr(feature = "std", derive(thiserror::Error))]
-#[derive(Debug, Clone, Display, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, displaydoc::Display)]
 /// Error when parsing IP or socket address
```
Arjentix, line 21:
```suggestion
/// Error when parsing address
#[derive(Debug, Clone, Copy, PartialEq, Eq, displaydoc::Display)]
```

I don't like `or`'s in user-facing messages. Like we a trying to guess

data_model/src/evaluate.rs
```diff
@@ -617,20 +616,16 @@ pub mod model {
...
-        #[display(fmt = "{_0}: Value not found in context")]
+        /// `{0}`: Value not found in context
         Find(
             #[skip_from]
             #[skip_try_from]
             String,
         ),
-        /// Conversion EvaluationError
-        #[display(fmt = "Conversion EvaluationError: {_0}")]
+        /// Conversion EvaluationError: `{0}`
```
Arjentix, line 628:
```suggestion
        /// Conversion evaluation error: `{0}`
```

data_model/src/evaluate.rs
```diff
@@ -617,20 +616,16 @@ pub mod model {
...
         /// Failed due to math exception
-        #[display(fmt = "Math error")]
         Math(#[cfg_attr(feature = "std", source)] MathError),
-        /// Validation error
-        #[display(fmt = "Validation failed")]
+        /// Validation failed
         Validation(#[cfg_attr(feature = "std", source)] Box<ValidationFail>),
-        /// Value not found in context.
-        #[display(fmt = "{_0}: Value not found in context")]
+        /// `{0}`: Value not found in context
```
Arjentix, line 622:
```suggestion
        /// `{0}`: Value not found in the context
```

Sounds better, but I'm not sure

p2p/src/lib.rs
```diff
@@ -45,44 +45,34 @@ pub mod boilerplate {
 }
 
 /// Errors used in [`crate`].
-#[derive(Debug, Error)]
+#[derive(Debug, Error, displaydoc::Display)]
 pub enum Error {
-    /// Failed to read or write
-    #[error("Failed IO operation: {0}.")]
+    /// Failed IO operation
     Io(#[source] std::sync::Arc<io::Error>),
```
Arjentix, line 51: I wonder why it's `Arc`, but ok

data_model/src/query.rs
```diff
@@ -1520,48 +1515,33 @@ pub mod error {
...
             Role(RoleId),
-            /// Failed to find [`PermissionTokenDefinition`] by id.
-            #[display(fmt = "Failed to find permission token definition by id: `{_0}`")]
+            /// Permission token definition with id `{0}` not found
             PermissionTokenDefinition(PermissionTokenId),
-            /// Failed to find [`PermissionToken`] for [`AccountId`].
-            #[cfg_attr(not(feature = "std"), display(fmt = "Failed to find permission token"))]
-            #[cfg_attr(feature = "std", error(transparent))]
+            /// Failed to find permission token
             PermissionToken(PermissionTokenFindError),
```
Arjentix, line 1541: I suggest to derive `thiserror::Error` and use `#[from]` here

data_model/src/metadata.rs
```diff
@@ -91,19 +91,14 @@ pub mod model {
...
     /// Metadata exceeds overall length limit
-    #[display(fmt = "Metadata exceeds overall length limit")]
     OverallSize(#[cfg_attr(feature = "std", source)] SizeError),
-    /// Empty path
-    #[display(fmt = "Path specification empty")]
+    /// Path specification empty
     EmptyPath,
-    /// Middle path segment is missing. I.e. nothing was found at that key
-    #[display(fmt = "{_0}: path segment not found")]
+    /// {0}: path segment not found, i.e. nothing was found at that key
```
Arjentix, line 99:
```suggestion
    /// `{0}`: path segment not found, i.e. nothing was found at that key
```

data_model/src/metadata.rs
```diff
@@ -91,19 +91,14 @@ pub mod model {
...
     MissingSegment(Name),
-    /// Middle path segment is not nested metadata. I.e. something was found, but isn't an instance of [`Metadata`]
-    #[display(fmt = "{_0}: path segment not an instance of metadata")]
+    /// {0}: path segment not an instance of metadata. I.e. something was found, but isn't an instance of Metadata
```
Arjentix, line 101:
```suggestion
    /// `{0}`: path segment not an instance of metadata

Arjentix (commented):
Nice idea to use this crate

client/Cargo.toml
DCNick3: I filed #3708 for that yesterday =)
Probably makes sense to make it here too (there will be merge conflicts anyway..)

p2p/src/lib.rs
appetrosyan: To be cloneable. If the structure wasn't private, I'd actually look into the underlying issue, because an `IO` error could be raised in multiple situations. Unfortunately, the standard library doesn't want that use case and an `IO` error can only be parsed as text.

core/src/smartcontracts/isi/triggers/set.rs
DCNick3: Today I learned: derive macros actually do get doc-comments before the `#[derive(...)]` in their input

appetrosyan (approved):

appetrosyan (approved):

Issue #3651, opened 2023-06-26
[suggestion] WSV snapshot


mversic:
### Feature request

It is in the scope of this ticket to add a way for Iroha to serialize and deserialize `wsv`.

We can do this:
1. periodically with a configurable dumping period
2. we can implement another endpoint in `Torii` to force dumping of `wsv` to the disk.

Once restarted, the node will read the `wsv` from the disk and reapply any excess transactions from the block store on top of it. The node will try reading a serialized `wsv` only if it is started with a specific flag, otherwise not

We can also make the network exchange the `wsv` when started similarly to what we do with broadcasting genesis block. IMO this would require more work and development time for too small gains

### Motivation

We need a way to restart Iroha without it having to reapply all transactions from the block store to get to the current WSV. This is a highly requested feature because it takes 3hrs to restart Iroha for Orillion team

### Who can help?

@appetrosyan @mversic

QuentinI:
I vaguely remember this discussed as a possible Kura init mode, where we periodically store snapshots of WSV in blockstore and restart by applying only txns that happened after the snapshot.
I'd say that makes the most sense to me.

mversic:
you are referring to [this](https://github.com/hyperledger/iroha/issues/1915) discussion

mversic:
I didn't think taking a `wsv` snapshot should be mixed with `Kura` init modes

appetrosyan:
I would recommend adding a SIGUSR1 handler too, since this is what most programs would do anyway.

timofeevmd:
Environment: mac m1
iroha2 commit: ee60fc82
setup: on system

step for reproduce:

- `./scripts/test_env.sh setup`
-`kill -2 ${pid}`
- ./scripts/test_env.sh setup`

console output
`
+ BLOCK_COUNT=
+ echo -1
-1
+ '[' '' = 1 ']'
+ echo 'Sleeping 1 sec...'
Sleeping 1 sec...
+ sleep 1
+ for i in $(seq 1 $1)
+ echo 'Waiting for genesis block to be created, attempt 16/20'
Waiting for genesis block to be created, attempt 16/20
++ curl --fail -X GET http://127.0.0.1:8180/status/blocks
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[...]

PR #3701, opened 2023-07-12
Kura init mode, wsv serialization basics

### LINKED ISSUE #1915: Implement missing `kura` init modes.
Kura currently has a `fast` and `strict` init mode, but both do the exact same thing. Given that the block storage had been optimised by @SamHSmith we should consider further improving the boot-up times and add a quick mode that skips some checks.

SamHSmith:
## Description

This PR is not done. But I want to get feedback on the strategy taken for serializing the WSV. If you think I should do something differently or that I've missed something please tell me.

### Linked issue

Closes #1915 

### Benefits

Fast init and boot for peers after power outage. 

### Checklist

- [x] Draft PR released
- [x] Strategy analysed
- [x] Weakpoints discovered
- [ ] Weakpoints addressed
- [ ] Regular PR
- [ ] Tests (via Pytest)

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -17,13 +17,15 @@ use iroha_crypto::HashOf;
 use iroha_data_model::block::VersionedCommittedBlock;
 use iroha_logger::prelude::*;
 use iroha_version::scale::{DecodeVersioned, EncodeVersioned};
+use parity_scale_codec::{DecodeAll, Encode};
 use parking_lot::Mutex;
 
-use crate::handler::ThreadHandler;
+use crate::{handler::ThreadHandler, wsv::WSVData, WorldStateView};
 
 const INDEX_FILE_NAME: &str = "blocks.index";
 const DATA_FILE_NAME: &str = "blocks.data";
 const LOCK_FILE_NAME: &str = "kura.lock";
+const WSV_FILE_NAME: &str = "wsv.snapshot";
 
 /// The interface of Kura subsystem
 #[derive(Debug)]
@@ -44,6 +46,8 @@ pub struct Kura {
     >,
     /// Path to file for plain text blocks.
     block_plain_text_path: Option<PathBuf>,
+    /// Path to blockchain directory.
+    block_store_path: PathBuf,
 }
 
 impl Kura {
@@ -73,6 +77,7 @@ impl Kura {
             block_store: Mutex::new(block_store),
             block_data: Mutex::new(Vec::new()),
             block_plain_text_path,
+            block_store_path: block_store_path.to_owned(),
         });
 
         Ok(kura)
@@ -87,6 +92,20 @@ impl Kura {
             block_store: Mutex::new(block_store),
             block_data: Mutex::new(Vec::new()),
             block_plain_text_path: None,
+            block_store_path: PathBuf::new(),
+        })
+    }
+
+    /// Create a kura instance that doesn't write to disk. Instead it serves as a handler
+    /// for in-memory blocks only. With tempdir path.
+    pub fn blank_kura_for_testing_with_dir(temp_dir: &Path) -> Arc<Kura> {
+        let block_store = BlockStore::fake_locked_for_tests(temp_dir.to_owned());
+        Arc::new(Self {
+            mode: Mode::Strict,
+            block_store: Mutex::new(block_store),
+            block_data: Mutex::new(Vec::new()),
+            block_plain_text_path: None,
+            block_store_path: temp_dir.to_owned(),
         })
     }
 
@@ -327,6 +346,35 @@ impl Kura {
         let mut data = self.block_data.lock();
         data.pop();
         data.push((block.hash(), Some(block)));
+    }
+
+    /// Store WSV to disk.
+    pub fn store_wsv(&self, wsv: &WorldStateView) {
+        let data: WSVData = wsv.into();
+        let path = self.block_store_path.join(WSV_FILE_NAME);
+
+        std::thread::spawn(move || {
+            let bin_data: Vec<u8> = data.encode();
+
+            let mut wsv_file = std::fs::OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .create(true)
+                .open(path.clone())
+                .expect("Failed to open wsv file");
+
+            wsv_file
+                .write_all(&bin_data)
+                .expect("Failed to write wsv file");
+        });
+    }
+
+    /// Load WSVData from disk.
+    pub fn try_load_wsv(&self) -> Option<WSVData> {
+        let path = self.block_store_path.join(WSV_FILE_NAME);
+        let binding = std::fs::read(path).ok()?;
+        let mut slice = binding.as_slice();
+        WSVData::decode_all(&mut slice).ok()
     }
 }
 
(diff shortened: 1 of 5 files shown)


core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -539,6 +540,42 @@ pub enum LoadedExecutable {
...
+            repeats: r.repeats.clone(),
+            authority: r.authority.clone(),
+            filter: r.filter.clone(),
+            metadata: r.metadata.clone(),
+        }
+    }
+}
+
+#[derive(Encode, Decode)]
+pub struct SerialAction<F> {
```
Arjentix, line 564: This struct looks exactly the same as `data_model::trigger::Action`, no?

Arjentix: `data_model::trigger::Action<F, Executable>` to be precise

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -539,6 +540,42 @@ pub enum LoadedExecutable {
     Instructions(Vec<InstructionBox>),
 }
 
+impl<F: Clone> From<&Action<F, LoadedExecutable>> for SerialAction<F> {
+    fn from(r: &Action<F, LoadedExecutable>) -> Self {
+        let exec = match &r.executable {
+            LoadedExecutable::Wasm(lw) => Executable::Wasm(WasmSmartContract(
+                lw.module
+                    .serialize()
+                    .expect("This failing means iroha was compiled incorrectly"),
```
Arjentix, line 549: Not Iroha. Trigger -- is being writte by external users. Take a look at another place in this file where `serialize()` is used

core/src/wsv.rs
```diff
@@ -139,6 +142,110 @@ impl Clone for WorldStateView {
     }
 }
 
+/// WSV data to that can be serialized and deserialized.
+#[derive(Encode, Decode)]
+pub struct WSVData {
```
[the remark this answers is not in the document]
Arjentix: And what's the problem of just (en-)coding in the special way? Why do we need a separate proxy type?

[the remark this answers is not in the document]
Arjentix: I'm just afraid this is hard to keep in sync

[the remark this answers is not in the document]
Arjentix: > Like get serialized world state from another peer and good to go?

Oh, I see, it's just to boot kura quickly. But this thing might be good as well. If we trust to peer of course

core/src/kura.rs
```diff
@@ -328,6 +347,35 @@ impl Kura {
...
+            let mut wsv_file = std::fs::OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .create(true)
+                .open(path.clone())
+                .expect("Failed to open wsv file");
+
+            wsv_file
+                .write_all(&bin_data)
+                .expect("Failed to write wsv file");
```
Arjentix, line 368: Please, specify a path to the file in panics

core/src/kura.rs
```diff
@@ -328,6 +347,35 @@ impl Kura {
...
+        });
+    }
+
+    /// Load WSVData from disk.
+    pub fn try_load_wsv(&self) -> Option<WSVData> {
```
Arjentix, line 373: I'd use explicit errors. If it seems like to much unrelated works consider creating a new task please

core/src/kura.rs
```diff
@@ -991,4 +1039,26 @@ mod tests {
...
+    #[test]
+    #[should_panic]
+    fn write_then_read_wsv_from_disk_should_fail() {
+        let dir = tempfile::tempdir().unwrap();
+        let kura = Kura::blank_kura_for_testing_with_dir(&dir.path());
+        let mut wsv = WorldStateView::new(crate::World::default(), kura.clone());
+
+        std::thread::sleep(std::time::Duration::from_secs(1));
+        let loaded_data = kura.try_load_wsv().unwrap();
+    }
```
Arjentix, line 1052:
```suggestion
    #[test]
    fn read_wsv_from_disk_without_write_should_fail() {
        let dir = tempfile::tempdir().unwrap();
        let kura = Kura::blank_kura_for_testing_with_dir(&dir.path());
        let mut wsv = WorldStateView::new(crate::World::default(), kura.clone());

        assert!(kura.try_load_wsv().is_none())
    }
```

core/src/kura.rs
```diff
@@ -991,4 +1039,26 @@ mod tests {
...
+
+    #[test]
+    fn write_then_read_wsv_from_disk() {
+        let dir = tempfile::tempdir().unwrap();
+        let kura = Kura::blank_kura_for_testing_with_dir(&dir.path());
+        let mut wsv = WorldStateView::new(crate::World::default(), kura.clone());
+
+        kura.store_wsv(&wsv);
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
Arjentix, line 1061: Oof, kind of unstable. Consider refactor so that user can decide if they want to detach or join thread

core/src/kura.rs
Arjentix, line 373: To be honest what Kura looks now may lead to *Zombie* state. This is the problem known to lazy initialization and etc.

So user can create Kura with `new()` and may forget to load `wsv`. I understand that with this approach you're trying to solve different loading modes problem. What I suggest is to have one creation method which will accept loading mode.

appetrosyan:
I'm fine with the general approach, but I think it has a few places where more work is known to be required.

core/src/kura.rs
appetrosyan: `Option` is fine, if the error that you return has context only when forwarded up the stack.

core/src/kura.rs
appetrosyan: Or don't panic. This looks like an error that could be forwarded up and handled later.

core/src/smartcontracts/isi/triggers/set.rs
appetrosyan: Well, technically both could cause this. However, you shouldn't lead peoeple down a single path if there are multiple causes. It doesn't look to me that this is an error with a single unambiguous cause.

core/src/smartcontracts/isi/triggers/set.rs
appetrosyan: They are not semantically the same. So I'm fine with some duplication

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -41,13 +42,13 @@ pub type LoadedAction<F> = Action<F, LoadedExecutable>;
...
+    pub data_triggers: HashMap<TriggerId, LoadedAction<DataEventFilter>>,
     /// Triggers using [`PipelineEventFilter`]
-    pipeline_triggers: HashMap<TriggerId, LoadedAction<PipelineEventFilter>>,
+    pub pipeline_triggers: HashMap<TriggerId, LoadedAction<PipelineEventFilter>>,
     /// Triggers using [`TimeEventFilter`]
-    time_triggers: HashMap<TriggerId, LoadedAction<TimeEventFilter>>,
+    pub time_triggers: HashMap<TriggerId, LoadedAction<TimeEventFilter>>,
     /// Triggers using [`ExecuteTriggerEventFilter`]
-    by_call_triggers: HashMap<TriggerId, LoadedAction<ExecuteTriggerEventFilter>>,
+    pub by_call_triggers: HashMap<TriggerId, LoadedAction<ExecuteTriggerEventFilter>>,
```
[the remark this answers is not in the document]
appetrosyan: I don't mind it being `pub(crate)` but if that doesn't work, I think we should go down the path of what @Arjentix has suggested.

core/src/validator.rs
```diff
@@ -62,6 +62,17 @@ impl Validator {
         })
     }
 
+    pub fn into_data_model(&self) -> data_model_validator::Validator {
+        data_model_validator::Validator {
+            wasm: crate::WasmSmartContract(
+                self.loaded_validator
+                    .module
+                    .serialize()
+                    .expect("This failing means iroha was compiled incorrectly"),
```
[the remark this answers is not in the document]
appetrosyan:
```suggestion
                    .expect("The loaded validator failed to serialize, which is not a recoverable error. Please consult our tutorial\n https://hyperledger.github.io/iroha-2-docs/guide/troubleshooting/overview.html\n to find the potential causes and fixes to this issue. "),
```

core/src/wsv.rs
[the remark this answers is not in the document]
mversic: if we need a proxy type can't we use a newtype pattern? like this:

```rs
pub struct WsvData(WorldStateView)
```

are there differences for the proxy struct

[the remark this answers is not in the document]
mversic: in any case, I'm sure this can be refactored

mversic:
will this affect @QuentinI 's work on #3651 ?

core/src/kura.rs
appetrosyan: No preference here. 

The `#[should_panic]` approach conveys that this is a negative test case more clearly, but this is also shown in the name of the test.

core/src/smartcontracts/isi/triggers/set.rs
appetrosyan: On that note, I'd like this struct to be defined before it's being used. Could you move it above `impl From`?

SamHSmith:
Closed in favor of #3724

PR #3703, opened 2023-07-12
[fix] #3613: Do not tolerate invalid signatures on transactions

### LINKED ISSUE #3613: [suggestion] Don't tolerate invalid signatures in a transaction
ATM we tolerate invalid transaction with invalid signatures, we just remove the invalid signature from signature set. What happens then is that transaction is accepted and lingers in the queue until it expires or until it's added more signatures (like a MST). @arndey encountered this issue where his transaction submit would timeout instead of being straight rejected. Check [here](https://github.com/hyperledger/iroha/blob/iroha2-dev/data_model/src/transaction.rs#L433)

DCNick3:
## Description

Fail deserializing/decoding transactions with invalid signatures 

Closes #3613 

## Unsolved questions

1. There are other places where the "retain valid signatures" logic is used: https://github.com/hyperledger/iroha/blob/c0a59c002b822a9dbca9bba376651d0e7457547a/core/src/sumeragi/network_topology.rs#L233 and https://github.com/hyperledger/iroha/blob/c0a59c002b822a9dbca9bba376651d0e7457547a/core/src/sumeragi/main_loop.rs#L1042 Maybe the logic in these places should be changed too? Or would it break some assumptions?

2. Error handling here is a bit funky. It returns an ad-hoc `&'static str` as an error type, potentially losing some context. It _could_ make sense to defined another error type just for this.

### DIFF
--- a/data_model/src/transaction.rs
+++ b/data_model/src/transaction.rs
@@ -377,12 +377,8 @@ mod candidate {
 
     impl SignedTransactionCandidate {
         #[cfg(feature = "std")]
-        fn validate(mut self) -> Result<SignedTransaction, &'static str> {
-            // TODO: Should we tolerate invalid signatures?
-            if self.retain_verified_signatures().is_empty() {
-                return Err("Transaction contains no valid signatures");
-            }
-
+        fn validate(self) -> Result<SignedTransaction, &'static str> {
+            self.validate_signatures()?;
             self.validate_instructions()
         }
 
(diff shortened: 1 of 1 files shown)


data_model/src/transaction.rs
```diff
@@ -405,12 +401,10 @@ mod candidate {
         }
 
         #[cfg(feature = "std")]
-        fn retain_verified_signatures(
-            &mut self,
-        ) -> Vec<&iroha_crypto::SignatureOf<TransactionPayload>> {
+        fn validate_signatures(&self) -> Result<(), &'static str> {
             self.signatures
-                .retain_verified_by_hash(self.payload.hash())
```
pesterev, line 412: The same pattern (removing invalid signatures) uses on the Sumeragi level. Shouldn't it be replaced there too?

mversic: That's for blocks, not transactions. There it makes sense to tolerate invalid signatures

Arjentix (approved):

SamHSmith (approved):

PR #3707, opened 2023-07-13
[feature] #3692: Validator migration

### LINKED ISSUE #3692: Executor migration
Currently we have `permission_tokens()` entrypoint inside `Executor` (which is still called `Validator`).

Need to make this entrypoint more flexible. Because, for example, validator may want to reassign renamed permission tokens and there is no way to do it currently. We need a better *hook* for that.

That's what *migration* does. It's a terminology taken from `Substrate`.

There should be `migration()` entrypoint which will be allowed to do anything with `wsv`, but only once -- when we upgrade *Executor*

Arjentix:
## Description

- Removed `permission_tokens()` entrypoint
- Created `migrate()` entrypoint
- Implemented migration on validators side
- Updated `entrypoint` macro
- Added new Runtime for migration including imports to (un-)register permission tokens
- Permission tokens update is now the responsibillity of validators
- Added test to check that failed migration makes no side effects

### Linked issue

- Closes #3692 

### Benefits

- Possibility to write any custom logic to perform migration from one validator to another

### Checklist


core/src/smartcontracts/isi/account.rs
```diff
@@ -611,4 +614,33 @@ pub mod query {
             .map_err(Into::into)
         }
     }
+
+    impl ValidQuery for FindAccountsWithPermissionTokenDefinition {
```
mversic, line 618: we don't need yet another query. You can achieve the same using `FindAllAccounts` + filter

wasm/validator/src/default.rs
```diff
@@ -112,18 +112,30 @@ pub(crate) use map_all_crate_tokens;
...
+            }
 
-        map_all_crate_tokens!(add_to_vec);
+            map_all_crate_tokens!(add_to_set);
+            set
+        };
 
-        v
+        Self::unregister_old_permission_tokens(&old_definitions, &new_definitions)?;
+        Self::register_permission_tokens(&old_definitions, &new_definitions)
```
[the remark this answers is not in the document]
Arjentix: > we have to make sure that it's impossible to call any of migration functions (line unregister_old_permission_tokens) at any time except during migration

Actually this is already done by our `wasm::Runtime` design.

I think that *Validator/Executor* should be able to run custom hook during upgrade, which migration is. If we will make migration in two sides: something on Iroha side and something on Wasm side that will be too implicit and will require more work from both sides

core/src/smartcontracts/isi/account.rs
Arjentix: As I understand we don't have a way to attach filter to a query in WASM. So I need to do it on WASM side. Where is the edge of what we should put into Iroha and what into client?
Although validator is not a basic client and this query has a big chance to be popular in this scenarios

mversic: > Where is the edge of what we should put into Iroha and what into client?

That's an interesting question. IMO there are 2 categories of queries:

1. queries executed by a client (can only be done through ISI)
2. queries executed by a smart contract (can be done through ISI or custom `extern` function)

Now for the 2nd category. What I'm objecting to is that we've started expanding ISI queries just for the purposes of **internal execution** of validator. No client will likely ever execute these queries. IMO, that's not ISI should be structured, rather the set of queries should be expanded according to the needs of the client. Internally used queries should be `extern` functions exposed by the host just like any other entrypoint

We should get more opinions here

Arjentix: I can't see much difference between creating `extern` function and creating new query. But with query this becomes usefull for basic clients as well, so double kill, because this shorthand query might be valuable for them as well

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -2,6 +2,11 @@
 //! [`token::CanControlDomainLives`] permission token.
 //!
 //! It also doesn't have [`iroha_validator::default::domain::tokens::CanUnregisterDomain`].
+//!
+//! In migration it replaces [`iroha_validator::default::domain::tokens::CanUnregisterDomain`]
+//! with [`token::CanControlDomainLives`] for all accounts.
+//! So it doesn't matter which domain user was able to unregister before migration, they will
+//! get access to control all domains. Remember that this is just a test example.
```
appetrosyan, line 9: This feels like it needs a more thorough explanation in the tutorial repo. We've accumulated a lot of material on the runtime validator/executor, so we should update the docs. I'd recommend doing it sooner rather than later.

wasm/validator/src/default.rs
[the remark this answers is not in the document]
appetrosyan: We had a relatively long discussion on this topic, so I think it's worth reiterating the key points. 

1. Our attack vector should exclude the Executor binary compromise. Not because it can't happen, but because it's _game over_ and nothing can be done to mitigate the consequences. 
2. The only constraint that the `migration` entrypoint must satisfy is that it is reproducible, and reasonable. What constitutes _reasonable_ very much depends on the blockchain, hence why the executor is the responsible party. Violating either constraint can be fixed by removing the migration from block history and starting over. 
3. If Iroha handles migrations, it must be made aware of (and be taught how to enforce the) executor invariants. This means that code is (as @Arjentix mentioned) shared between Iroha and the executor. 

Under these circumstances, it might be worth exploring both ideas, but it seems that a lot of the advantages of host-side migrations are compromised by points 1, 2 and 3. IMO we'll have a smaller API surface area if we allow the executor, an already important piece of the blockchain security, to also be responsible for its own migrations.

wasm/validator/src/default.rs
```diff
@@ -155,6 +167,44 @@ impl DefaultValidator {
             host: iroha_wasm::Host,
         }
     }
+
+    fn unregister_old_permission_tokens(
+        old_definitions: &BTreeSet<PermissionTokenDefinition>,
```
appetrosyan, line 172: Consider adding 
```rust
type Permissions = BTreeSet<PermissionTokenDefinition>
```

appetrosyan (approved):

ffi/derive/tests/ui_fail/array_as_ptr.stderr
```diff
@@ -2,7 +2,7 @@ error[E0308]: mismatched types
  --> tests/ui_fail/array_as_ptr.rs:9:17
   |
 9 |     __array_arg(arg);
-  |     ----------- ^^^ expected `*mut [u32; 2]`, found array `[u32; 2]`
+  |     ----------- ^^^ expected `*mut [u32; 2]`, found `[u32; 2]`
```
appetrosyan, line 5: TRYBUILD=overwrite

core/src/wsv.rs
```diff
@@ -1044,6 +1047,140 @@ impl WorldStateView {
...
+                return Err(FindError::Role(role_id.clone()).into());
+            }
+        }
+
+        self.emit_events(events);
+
+        Ok(())
+    }
+
+    /// Remove all tokens with specified definition id from all accounts in all domains
```
appetrosyan, line 1145: I'm a bit concerned with leaving this in long-term. 

Hopefully we'll have some opportunity to discuss the new Iroha ISA, and these functions will be generic.

Arjentix: What does *ISA* stand for?

core/src/smartcontracts/isi/account.rs
Arjentix: - Created #3720

wasm/validator/src/default.rs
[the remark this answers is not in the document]
Arjentix: After a discussion with @appetrosyan and @mversic we decided to keep following current approach AFAIU.

@mversic , if you're agree, close the discussion, please

wasm/validator/src/lib.rs
```diff
@@ -112,8 +112,15 @@ macro_rules! declare_tokens {
...
-    fn permission_tokens() -> Vec<PermissionTokenDefinition>;
+    /// Migrate previous validator to the current version.
+    ///
+    /// This function should be called by `migrate` entrypoint,
+    /// which will be called by Iroha only once just before upgrading validator.
+    ///
+    /// # Errors
+    ///
+    /// Concrete errors are specific to the implementation.
+    fn migrate() -> MigrationResult;
```
mversic, line 123: I get the feeling `migrate` doesn't belong into `Validate` trait. Let's have a short discussion on our naming. We should adjust our naming a little bit since we've moved from validator to executor. Can you give a thought about the naming conventions

client/tests/integration/smartcontracts/validator_with_migration_fail/src/lib.rs
mversic, line 1: it's not your fault, but we have a lot of boilerplate here. We should consider how to reduce it

mversic (changes requested):

wasm/validator/src/lib.rs
Arjentix: I agree that `migrate` doesn't look nice in `Validate` trait. My hope is that we eventually rename it to `Execute`

wasm/validator/src/default.rs
```diff
@@ -155,6 +169,44 @@ impl DefaultValidator {
...
+            .difference(old_definitions)
+            .try_for_each(|definition| {
+                crate::iroha_wasm::register_permission_token_definition(definition)
+            })
+            .map_err(|error| {
+                Self::format_err(error, "Failed to register new permission token definitions")
+            })
+    }
+
+    pub fn format_err<C>(error: ValidationFail, context: C) -> MigrationError
```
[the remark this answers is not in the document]
Arjentix: Ok, will remove

wasm/validator/Cargo.toml
```diff
@@ -11,11 +11,12 @@ default = ["default-validator"]
 # Enables debugging tools such as `dbg()` and `DebugUnwrapExt`
 debug = ["iroha_wasm/debug"]
 # Enables predefined default validator
-default-validator = []
+default-validator = ["dep:anyhow"]
 
 [dependencies]
 iroha_wasm = { version = "2.0.0-pre-rc.16", path = ".." }
 iroha_validator_derive = { version = "2.0.0-pre-rc.16", path = "derive" }
+anyhow = {version = "1.0.71", default-features = false, optional = true}
```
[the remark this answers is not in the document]
Arjentix: Indeed. Will fix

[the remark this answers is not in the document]
Arjentix: Looks like these dependencies are used only in this crate and won't be used anywhere else in this workspace

wasm/validator/Cargo.toml
```diff
@@ -11,11 +11,12 @@ default = ["default-validator"]
 # Enables debugging tools such as `dbg()` and `DebugUnwrapExt`
 debug = ["iroha_wasm/debug"]
 # Enables predefined default validator
-default-validator = []
+default-validator = ["dep:anyhow"]
```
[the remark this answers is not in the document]
Arjentix: It works, but `dep` syntax is better. It shows that `anyhow` is not a feature name, but a dependency name. Take a look [here](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies)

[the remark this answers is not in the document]
Arjentix: I'd like if we used `dep` syntax everywhere

core/src/smartcontracts/isi/world.rs
```diff
@@ -352,54 +227,26 @@ pub mod isi {
...
+                })?
+                .map_err(|migration_error| {
+                    InvalidParameterError::Wasm(format!(
+                        "{:?}",
+                        eyre::eyre!(migration_error).wrap_err("Migration failed"),
+                    ))
                 })?;
 
             let world = wsv.world_mut();
             let _ = world.upgraded_validator.insert(new_validator);
```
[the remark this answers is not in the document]
Arjentix: Yes. Also I've added a test to check that validator won't be upgraded after failed migration

client/tests/integration/smartcontracts/validator_with_migration_fail/src/lib.rs
Arjentix: What do you mean exactly? Boilerplate between `default_validator`, `validator_with_admin` and this one? Or just in this file?

If you mean just this file then I don't understand you.

wasm/validator/src/lib.rs
Arjentix: As well as all *validator* related terminology will be renamed to *executor* related.

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
appetrosyan: That is unfortunately true and some work will need to be duplicated. But we expect that either RC18, RC19 or RC20 are going to be LTS candidates. This functionality is going to be utilised heavily.

wasm/validator/Cargo.toml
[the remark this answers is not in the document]
appetrosyan: Created #3754

appetrosyan (dismissed):

client/tests/integration/events/pipeline.rs
```diff
@@ -100,7 +100,7 @@ impl Checker {
 
 #[test]
 fn committed_block_must_be_available_in_kura() {
-    let (_rt, peer, client) = <PeerBuilder>::new().with_port(10_985).start_with_runtime();
+    let (_rt, peer, client) = <PeerBuilder>::new().with_port(11_040).start_with_runtime();
```
0x009922, line 103:
```suggestion
    let (_rt, peer, client) = <PeerBuilder>::new().with_any_port().start_with_runtime();
```

Would be nice to have a `#[cfg(test)]` method `::with_any_port()` which find any available port on a machine.

Related:

- https://github.com/hyperledger/iroha/issues/2137

data_model/src/validator.rs
```diff
@@ -79,6 +79,12 @@ pub mod model {
 /// Result type that every validator should return.
 pub type Result<T = (), E = crate::ValidationFail> = core::result::Result<T, E>;
 
+/// Migration error type.
+pub type MigrationError = String;
+
+/// Result type for `migration()` entrypoint.
```
0x009922, line 85: `migration()`? Or `migrate()`?

Since there is no way to make a doc link to the exact function name, maybe omit the exact name here at all?

```suggestion
/// Result type for a validator's `migrate()` entrypoint.
```

wasm/src/lib.rs
```diff
@@ -188,7 +188,8 @@ pub fn query_triggering_event() -> Event {
 ///
 /// # Traps
 ///
-/// Host side will generate a trap if this function was not called from a validator.
+/// Host side will generate a trap if this function was not called from a
+/// validator `validate()` entrypoint.
```
0x009922, line 192:
```suggestion
/// Host side will generate a trap if this function was not called from a
/// validator's `validate()` entrypoint.
```

wasm/src/lib.rs
```diff
@@ -199,6 +200,22 @@ pub fn query_operation_to_validate() -> NeedsValidationBox {
...
+/// Set new [`PermissionTokenSchema`].
+///
+/// # Errors
+///
+/// - If execution on Iroha side failed
+///
+/// # Traps
+///
+/// Host side will generate a trap if this function was not called from a
+/// validator `migrate()` entrypoint.
```
0x009922, line 212:
```suggestion
/// Host side will generate a trap if this function was not called from a
/// validator's `migrate()` entrypoint.
```

wasm/validator/derive/src/entrypoint.rs
```diff
@@ -66,12 +66,8 @@ pub fn impl_entrypoint(attr: TokenStream, item: TokenStream) -> TokenStream {
...
     match &fn_item.sig.ident {
         fn_name if fn_name == "validate" => impl_validate_entrypoint(attr, fn_item),
-        fn_name if fn_name == "permission_token_schema" => {
-            impl_permission_token_schema_entrypoint(&attr, fn_item)
-        }
-        _ => panic!(
-            "Validator entrypoint name should be either `validate` or `permission_token_schema`"
-        ),
+        fn_name if fn_name == "migrate" => impl_migrate_entrypoint(&attr, fn_item),
+        _ => panic!("Validator entrypoint name should be either `validate` or `migrate`"),
```
0x009922, line 70: Actually, it is possible to define entrypoints names (`"validate"`, `"migrate"`, `"permission_tokens"`) as consts.

Thus:

- It is possible to re-use them in the panic message here;
- If moved to a separate crate, it is possible to mention then in doc comments and avoid pasting everywhere `migrate()` and `validate()`.

The outcome is a more coherent and refactoring-prone codebase and documentation.

client/tests/integration/smartcontracts/validator_with_migration_fail/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "validator_with_migration_fail"
+
+edition.workspace = true
```
0x009922, line 4: Wow, I haven't known such a syntax is allowed in TOML.

0x009922 (dismissed):
LGTM.

Although I think about an improvement: define entrypoints names as consts. Then it will be possible to re-use it across the codebase (except the source code of validators themselves) and mention them in doccomments. It doesn't seem difficult to do. More details in one of my comments.

client/tests/integration/events/pipeline.rs
Erigara: @0x009922 initially we used such approach with [unique_port](https://github.com/soramitsu/iroha2-unique_port) crate, unfortunately it didn't work very well because there were no ipc and accidentally 2 different tests were trying to take the same port. 
When we run tests they are run in different processes and we need some form of ipc to distribute available ports without large overhead.

wasm/validator/derive/src/entrypoint.rs
appetrosyan: refactoring-prone usually means that you want to refactor it often. You probably mean easy-to-refactor.

client/tests/integration/smartcontracts/validator_with_migration_fail/Cargo.toml
appetrosyan: One of the main reasons why even Python is starting to use TOML in their package configurations.

appetrosyan (approved):

wasm/validator/derive/src/entrypoint.rs
Arjentix: Created #3757

appetrosyan (dismissed):

data_model/src/lib.rs
```diff
@@ -555,6 +555,7 @@ pub mod parameter {
 }
 
 #[model]
+#[allow(irrefutable_let_patterns)] // Triggered from derives macros
```
appetrosyan, line 558: Fix `irrefutable_let_patterns` lint generated in the `#[model]` macro.

appetrosyan:
Merging in the interest of time, given that the changes were previously approved by @0x009922

PR #3708, opened 2023-07-13
[refactor] #3289: Use workspace inheritance

### LINKED ISSUE #3232: Workspace package inheritance

### LINKED ISSUE #3289: Workspace dependencies inheritance

DCNick3:
## Description

This PR extracts all common dependencies (2 usages or more) into a workspace level, so that it is easier to update them in lockstep.

Also, all iroha crates besides except inner `_derive` crates are also specified at workspace level.

### Issues

There are two places where I couldn't unify the dependencies:

1. `tokio-tungstenite` of versions 0.16 and 0.17 is used
https://github.com/hyperledger/iroha/blob/39504a5a8029080e0a7fe93b0f06a74710dba347/client/Cargo.toml#L44-L43

2. `clap` v3 and v4 is used:
https://github.com/hyperledger/iroha/blob/39504a5a8029080e0a7fe93b0f06a74710dba347/client_cli/Cargo.toml#L31-L30

I haven't tried too hard, but it seems that it would require some code changes to update these places. 

### Linked issue

Closes #3289

### DIFF
--- a/logger/Cargo.toml
+++ b/logger/Cargo.toml
@@ -8,23 +8,23 @@ authors.workspace = true
 license.workspace = true
 
 [dependencies]
-iroha_config = { version = "=2.0.0-pre-rc.16", path = "../config" }
+iroha_config = { workspace = true }
 
-color-eyre = "0.6.2"
-serde_json = "1.0.91"
-tracing = "0.1.37"
-tracing-core = "0.1.30"
-tracing-futures = { version = "0.2.5", default-features = false, features = ["std-future", "std"] }
-tracing-subscriber = { version = "0.3.16", default-features = false, features = ["fmt", "ansi"] }
-tracing-bunyan-formatter = { version = "0.3.4", default-features = false }
-tokio = { version = "1.23.0", features = ["sync"] }
+color-eyre = { workspace = true }
+serde_json = { workspace = true }
+tracing = { workspace = true }
+tracing-core = { workspace = true }
+tracing-futures = { workspace = true, features = ["std-future", "std"] }
+tracing-subscriber = { workspace = true, features = ["fmt", "ansi"] }
+tracing-bunyan-formatter = { workspace = true }
+tokio = { workspace = true, features = ["sync"] }
 console-subscriber =  { version = "0.1.8", optional = true }
-once_cell = "1.16.0"
-derive_more = "0.99.17"
+once_cell = { workspace = true }
+derive_more = { workspace = true }
 tracing-error = "0.2.0"
 
 [dev-dependencies]
-tokio = { version = "1.23.0", features = ["macros", "time", "rt"] }
+tokio = { workspace = true, features = ["macros", "time", "rt"] }
 
 
 [features]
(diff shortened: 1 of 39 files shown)


tools/parity_scale_decoder/Cargo.toml
```diff
@@ -13,20 +13,22 @@ license.workspace = true
...
 colored = "2.0.0"
 
 [build-dependencies]
-iroha_data_model = { version = "=2.0.0-pre-rc.16", path = "../../data_model"}
-parity-scale-codec = { version = "3.2.1", default-features = false }
+iroha_data_model = { workspace = true }
+
+parity-scale-codec = { workspace = true }
 serde_json = "1.0.91"
 serde = "1.0.151"
```
pesterev, line 34: seems to be missing here

DCNick3: I _think_ it's fine?

I just added an empty line between `iroha_*` deps and other deps, like is done in most other `Cargo.toml`s

logger/Cargo.toml
```diff
@@ -8,23 +8,23 @@ authors.workspace = true
...
+iroha_config = { workspace = true }
 
-color-eyre = "0.6.2"
-serde_json = "1.0.91"
-tracing = "0.1.37"
+color-eyre = { workspace = true }
+serde_json = { workspace = true }
+tracing = { workspace = true }
 tracing-core = "0.1.30"
 tracing-futures = { version = "0.2.5", default-features = false, features = ["std-future", "std"] }
```
0x009922, line 17: Shouldn't it be moved as well?

AFAIU if `tracing` is specified there, these related crates might be moved there as well.

tools/parity_scale_decoder/Cargo.toml
0x009922, line 34: Why in all other places the workspace version of `serde` is used, but here is not?

0x009922 (approved):
Aahh, so pleasant set of changes.

As for `clap`'s v3 vs v4: it was quite easy to migrate in Kagami. However, it is better to do in a separate PR.

logger/Cargo.toml
DCNick3: Yeah, probably it makes sense to move co-dependent crates

tools/parity_scale_decoder/Cargo.toml
DCNick3: It's an oversight =)

appetrosyan (approved):

PR #3710, opened 2023-07-17
[feature] #3688: Implement basic arithmetic as proc macro

### LINKED ISSUE #3688: [dsl]: Implement simple arithmetic parser
### Feature request

- [ ] Addition
- [ ] Subtraction
- [ ] Multiplication
- [ ] Division
- [ ] Type checking 

### Motivation

Part of the DSL internship

### Who can help?

@appetrosyan

RealNeGate:
## Description

Implement skeleton DSL crate. Arithmetic and boolean logic handled as a procedural macro. Uses `data_model::expression`. 

### Linked issue

Closes #3688 

### Benefits

DSL for expressions.

### Checklist

### DIFF
--- a/dsl/Cargo.toml
+++ b/dsl/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "iroha_dsl"
+version = "2.0.0-pre-rc.18"
+edition = "2021"
+authors = ["Yasser"]
+
+[lib]
+proc-macro = true
+
+[dependencies]
+quote = "1.0"
+proc-macro2 = "1.0"
+litrs = "0.4.0"
+
+[dev-dependencies]
+iroha_data_model = { workspace = true }
+iroha_crypto = { workspace = true }
+iroha_config = { workspace = true }
+iroha_client = { workspace = true }
+serde_json = "1.0"
+
+
--- a/dsl/src/lib.rs
+++ b/dsl/src/lib.rs
@@ -1,0 +1,205 @@
+use std::{convert::TryFrom, iter::Peekable, str::FromStr};
+
+use litrs::Literal;
+use proc_macro2::{
+    token_stream::IntoIter,
+    TokenStream, TokenTree,
+    TokenTree::{Ident, Punct},
+};
+use quote::quote;
+
+#[derive(PartialEq)]
+enum ExprType {
+    Nil,
+    Int,
+    Bool,
+}
+
+// within the compiler, these generate tokens for
+// Value
+struct ParsedExpr {
+    t: ExprType,
+    tokens: TokenStream,
+}
+
+struct Operator(i32, ExprType, &'static str);
+
+fn parse_literal(it: &mut Peekable<IntoIter>) -> ParsedExpr {
+    let token = it
+        .next()
+        .expect("failed to parse literal, hit end of string");
+
+    match token {
+        Punct(ref x) => {
+            if x.as_char() == '(' {
+                let v = parse_expr(it);
+
+                it.next()
+                    .expect("expected closing paren, hit end of string");
+                return v;
+            }
+        }
+
+        // boolean literals
+        Ident(ref x) => match x.to_string().as_str() {
+            "true" => {
+                return ParsedExpr {
+                    t: ExprType::Bool,
+                    tokens: quote! { true },
+                }
+            }
+            "false" => {
+                return ParsedExpr {
+                    t: ExprType::Bool,
+                    tokens: quote! { false },
+                }
+            }
+
+            // unary not
+            "not" => {
+                let v = parse_literal(it).tokens;
+                return ParsedExpr {
+                    t: ExprType::Bool,
+                    tokens: quote! { Not::new(#v) },
+                };
+            }
+            _ => panic!("error: unknown identifier"),
+        },
+
+        // kinda ugly but we fallthrough basically
+        _ => {}
+    }
+
+    // integer literals
+    if let Ok(Literal::Integer(i)) = Literal::try_from(token) {
+        let v = i.value::<u64>().expect("i don't think this integer fits?");
+        return ParsedExpr {
+            t: ExprType::Int,
+            tokens: quote! { #v },
+        };
+    }
+
+    ParsedExpr {
+        t: ExprType::Nil,
+        tokens: TokenStream::new(),
+    }
+}
+
+fn precedence(it: &mut Peekable<IntoIter>) -> Option<Operator> {
+    match it.peek() {
+        Some(Punct(x)) => match x.as_char() {
+            // arithmatic
+            '+' => Some(Operator(4, ExprType::Int, "Add")),
+            '-' => Some(Operator(4, ExprType::Int, "Subtract")),
+            '*' => Some(Operator(3, ExprType::Int, "Multiply")),
+
+            // compares
+            '=' => Some(Operator(2, ExprType::Int, "Equal")),
+            '>' => Some(Operator(2, ExprType::Int, "Greater")),
+            '<' => Some(Operator(2, ExprType::Int, "Less")),
+            _ => None,
+        },
+        Some(Ident(ref x)) => match x.to_string().as_str() {
+            "and" => Some(Operator(1, ExprType::Bool, "And")),
+            "or" => Some(Operator(1, ExprType::Bool, "Or")),
+            _ => None,
+        },
+        _ => None,
+    }
+}
+
+/// precedence walking
+fn parse_binop(it: &mut Peekable<IntoIter>, min_prec: i32) -> ParsedExpr {
+    let mut lhs = parse_literal(it);
+    while let Some(prec) = precedence(it) {
+        it.next();
+        if prec.0 < min_prec {
+            break;
+        }
+
+        let op = proc_macro2::TokenStream::from_str(prec.2).unwrap();
+        let rhs = parse_literal(it);
+
+        assert!(
+            lhs.t == prec.1 && rhs.t == prec.1,
+            "cannot perform binary operator on these!"
+        );
+
+        let lhs_tokens = lhs.tokens;
+        let rhs_tokens = rhs.tokens;
+        lhs = ParsedExpr {
+            t: prec.1,
+            tokens: quote! { #op::new(#lhs_tokens, #rhs_tokens) },
+        };
+    }
+
+    lhs
+}
+
+fn is_ident(a: &TokenTree, b: &'static str) -> bool {
+    if let Ident(ref x) = a {
+        return x.to_string().as_str() == b;
+    }
+
+    false
+}
+
+fn parse_expr(it: &mut Peekable<IntoIter>) -> ParsedExpr {
+    if is_ident(it.peek().expect("hit end of string"), "if") {
+        it.next();
+
+        let cond_tokens = parse_binop(it, 0).tokens;
+        assert!(
+            is_ident(&it.next().expect("hit end of string"), "then"),
+            "expected 'then'"
+        );
+
+        let true_case = parse_binop(it, 0);
+        assert!(
+            is_ident(&it.next().expect("hit end of string"), "else"),
+            "expected 'else'"
+        );
+
+        let false_case = parse_binop(it, 0);
+        assert!(
+            true_case.t == false_case.t,
+            "both types in a conditional must match"
+        );
+
+        let true_tokens = true_case.tokens;
+        let false_tokens = false_case.tokens;
+        return ParsedExpr {
+            t: ExprType::Int,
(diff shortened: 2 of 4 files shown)


DCNick3:
It seems that you are doing the parsing completely by-hand.. I think it might be beneficial to use [syn's parsing framework](https://docs.rs/syn/latest/syn/parse/index.html). Using it also makes error handling easier and would actually report spans for where the error occurs (as opposed to just `panic`ing)

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        Punct(ref x) => if x.as_char() == '(' {
+            let v = parse_expr(it);
+
+            it.next().expect("expected closing paren, hit end of string");
+            return v
+        },
+
+        // boolean literals
+        Ident(ref x) => match x.to_string().as_str() {
+            "true"  => return ParsedExpr{ t: ExprType::Bool, tokens: quote!{ true }},
```
DCNick3, line 43: You should probably re-use the tokens you parsed instead of making a new `TokenStream` with `quote!`. This way you would be able to keep span information

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            "false" => return ParsedExpr{ t: ExprType::Bool, tokens: quote!{ false }},
+
+            // unary not
+            "not"   => {
```
DCNick3, line 47: A bit of bikeshed: I think that a DSL used in Rust is better off using a syntax that resembles Rust. Thus, suggest using `!` to denote the "not" operation

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+            // compares
+            '=' => Some((2, ExprType::Int, "Equal")),
+            '>' => Some((2, ExprType::Int, "Greater")),
+            '<' => Some((2, ExprType::Int, "Less")),
+            _ => None,
+        },
+
+        Some(Ident(ref x)) => match x.to_string().as_str() {
+            "and" => Some((1, ExprType::Bool, "And")),
```
DCNick3, line 83: Same as before, suggest using `&&` and `||`

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    let mut lhs = parse_literal(it);
+    while let Some(prec) = precedence(it) {
+        it.next();
+        if prec.0 < min_prec { break }
+
+        let op = proc_macro2::TokenStream::from_str(prec.2).unwrap();
+        let rhs = parse_literal(it);
+
+        if lhs.t != prec.1 || rhs.t != prec.1 {
+            panic!("error: cannot perform binary operator on these!");
```
DCNick3, line 103: Just a note: AFAIK, you would be extending this crate to cover more expressions. For larger DSLs (and for better error reporting) it often makes sense to split the parsing and type-checking into separate passes. Not sure it makes sense in your case though, use your judgement here =)

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        return false;
+    }
+}
+
+fn parse_expr(it: &mut Peekable<IntoIter>) -> ParsedExpr {
+    if is_ident(it.peek().expect("hit end of string"), "if") {
+        it.next();
+
+        let cond_tokens = parse_binop(it, 0).tokens;
+        if !is_ident(&it.next().expect("hit end of string"), "then") {
```
DCNick3, line 127: Ditto about non-rusty syntax

dsl/src/lib.rs
appetrosyan: Why not both?

dsl/src/lib.rs
appetrosyan: I'd suggest keeping `and` and `or`, and maybe adding support for `&&` and `||` for those who are more used to the masochistic languages.

dsl/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[dependencies]
+quote = "1.0"
+proc-macro2 = "1.0"
+litrs = "0.4.0"
+
+[dev-dependencies]
+iroha_data_model = { version = "=2.0.0-pre-rc.16", path = "../data_model" }
+iroha_crypto = { version = "=2.0.0-pre-rc.16", path = "../crypto" }
+iroha_config = { version = "=2.0.0-pre-rc.16", path = "../config" }
+iroha_client = { version = "=2.0.0-pre-rc.16", path = "../client" }
```
appetrosyan, line 21: We should remember to update these after #3232 is merged. If it gets merged before this, we should update these packages to use workspace inheritance just as well.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
+extern crate quote;
+extern crate litrs;
+extern crate proc_macro2;
```
appetrosyan, line 3: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#an-exception

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use quote::quote;
+
+use proc_macro2::TokenTree;
+use proc_macro2::TokenStream;
+use proc_macro2::TokenTree::Punct;
+use proc_macro2::TokenTree::Ident;
+use proc_macro2::token_stream::IntoIter;
+use std::iter::Peekable;
```
appetrosyan, line 12:
```suggestion
use proc_macro2::{TokenTree, TokenStream, TokenTree::{Punct, Ident}, token_stream::IntoIter};

use std::iter::Peekable;
```

appetrosyan: This is one of the things that would be suggested to you by `cargo fmt --all`.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use std::convert::TryFrom;
+use std::str::FromStr;
+use litrs::Literal;
+
+#[derive(PartialEq)]
+enum ExprType {
+    Nil, Int, Bool
```
appetrosyan, line 20: Familiar name. 

I suggest you have a look at [non_exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html) and see if it fits the role of `Nil`.

dsl/src/lib.rs
appetrosyan: Another more abstract reason is that you won't be repeating `true` and `true` and making the compiler do extra work.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+fn parse_literal(it: &mut Peekable<IntoIter>) -> ParsedExpr {
+    let token = it.next().expect("failed to parse literal, hit end of string");
+
+    match token {
+        Punct(ref x) => if x.as_char() == '(' {
+            let v = parse_expr(it);
+
+            it.next().expect("expected closing paren, hit end of string");
```
appetrosyan, line 37: https://docs.rs/proc-macro-error/latest/proc_macro_error/

We use it all over the code base.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            // unary not
+            "not"   => {
+                let v = parse_literal(it).tokens;
+                return ParsedExpr{ t: ExprType::Bool, tokens: quote!{ Not::new(#v) }}
+            }
+            _ => panic!("error: unknown identifier"),
+        },
+
+        // kinda ugly but we fallthrough basically
+        _ => {},
```
appetrosyan, line 55: Don't worry, there will be some sophisticated parsing here soon.

dsl/src/lib.rs
appetrosyan: We'll get to it, don't worry :)

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        let true_case = parse_binop(it, 0);
+        if !is_ident(&it.next().expect("hit end of string"), "else") {
+            panic!("expected 'else'")
+        }
+
+        let false_case = parse_binop(it, 0);
+        if true_case.t != false_case.t {
+            panic!("both types in a conditional must match")
```
appetrosyan, line 138: Very good for now. Later I'd consider adding more diagnostic info to such errors. For example, which expression has which type, and why. What you're doing here is lightweight type inference that might become a bit hard to do later.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+/// use iroha_data_model::{prelude::*, ParseError};
+///
+/// fn main() {
+///     assert_eq!(make_answer!(54654*5 + 1), Add::new(Multiply::new(54654_u64, 5_u64), 1_u64));
+///     println!("{}", make_answer!(not true and false));
+///     println!("{}", make_answer!(if 4 = 4 then 64 else 32));
+/// }
+/// ```
+#[proc_macro]
+pub fn make_answer(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
appetrosyan, line 169:
```suggestion
pub fn expr(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```

dsl/src/lib.rs
RealNeGate: I think it would work just fine, i'll make the relevant changes

dsl/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+autobins = false
+
+[lib]
+proc-macro = true
+path = "src/lib.rs"
+
+[dependencies]
+quote = "1.0"
+proc-macro2 = "1.0"
+litrs = "0.4.0"
```
ilchu, line 15: A more conventional choice would be to just use `syn::Lit` -- and as @DCNick3 mentioned, you'd want some of its other capabilities sooner or later. E.g. I feel `syn::visit` module could especially come in handy when the tree walks get more elaborate.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    // integer literals
+    if let Ok(Literal::Integer(i)) = Literal::try_from(token) {
+        let v = i.value::<u64>().expect("i don't think this integer fits?");
+        return ParsedExpr{ t: ExprType::Int, tokens: quote!{ #v }}
+    }
+
+    ParsedExpr{ t: ExprType::Nil, tokens: TokenStream::new() }
+}
+
+fn precedence(it: &mut Peekable<IntoIter>) -> Option<(i32, ExprType, &'static str)> {
```
ilchu, line 67: A minor gripe, but I feel such return signatures call for a custom enum instead to make it more readable.

dsl/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        let lhs_tokens = lhs.tokens;
+        let rhs_tokens = rhs.tokens;
+        lhs = ParsedExpr{ t: prec.1, tokens: quote!{ #op::new(#lhs_tokens, #rhs_tokens) }};
+    }
+
+    lhs
+}
+
+fn is_ident(a: &TokenTree, b: &'static str) -> bool {
```
ilchu, line 114: Returning to the `syn` comment already made: this functionality is provided by `syn::Path::is_ident`.

ilchu:
As another side note: you might have a somewhat different workflow, but adding this new crate to the whole `iroha` workplace will make it discoverable from the project root. Besides that, working on a spinoff branch instead of on top of your `iroha2-dev` branch is more comfortable for quickly checking out the changes you've made when reviewing.

dsl/Cargo.toml
appetrosyan: Created #3756

dsl/src/lib.rs
RealNeGate: went with `expr` as the name

dsl/src/lib.rs
RealNeGate: changed it to an `Option<Operator>` to avoid the uglier tuple, I'm curious if that's fine

dsl/src/lib.rs
RealNeGate: Removed these

dsl/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        .next()
+        .expect("failed to parse literal, hit end of string");
+
+    match token {
+        Punct(ref x) => {
+            if x.as_char() == '(' {
+                let v = parse_expr(it);
+
+                it.next()
+                    .expect("expected closing paren, hit end of string");
```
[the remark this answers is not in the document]
appetrosyan: Created #3786

dsl/src/lib.rs
appetrosyan: Created #3787

appetrosyan (approved):

PR #3721, opened 2023-07-19
[ci] #3700: CI check for `kagami validator`

### LINKED ISSUE #3700: [ci] Check that `kagami validator` is runnable as standalone
### Feature request

Add a CI check to ensure that now `kagami validator` can be run as e.g. `target/debug/kagami <cmd>`, i.e. with no `CARGO_*` env variables set.

### Motivation

`kagami` being truly portable makes it an independent tool just as it should be.

### Who can help?

@ilchu

ilchu:
## Description

Added a dumb CI check to verify that we can run `kagami validator` as a standalone binary with no errors.

### Linked issue

Closes #3700 

### Benefits
Peace of mind. Even though this is not the most intended usage, our QA already came across this problem. 

### Checklist

### DIFF
--- a/tools/kagami/build.rs
+++ b/tools/kagami/build.rs
@@ -1,8 +1,16 @@
 //! Provides macros used to get the version
 
-use std::error::Error;
+use std::{env::set_var, error::Error};
 
 fn main() -> Result<(), Box<dyn Error>> {
+    // We default to upstream's latest hash (usually `iroha2-dev`)
+    // to enable shallow cloning in swarm
+    let repo = git2::Repository::open("../..")?;
+    let sha = match repo.revparse_single("@{u}") {
+        Ok(rev) => rev.id(),
+        Err(_) => repo.revparse_single("HEAD")?.id(),
+    };
+    set_var("VERGEN_GIT_SHA", sha.to_string());
     vergen::EmitBuilder::builder().git_sha(false).emit()?;
     Ok(())
 }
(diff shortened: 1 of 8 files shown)


tools/kagami/build.rs
```diff
@@ -1,8 +1,12 @@
 //! Provides macros used to get the version
 
-use std::error::Error;
+use std::{env::set_var, error::Error};
 
 fn main() -> Result<(), Box<dyn Error>> {
+    // We default to iroha2-dev's latest hash to enable shallow cloning in swarm
+    let repo = gix::open("../..")?;
+    let sha = repo.rev_parse_single("iroha2-dev")?;
+    set_var("VERGEN_GIT_SHA", sha.to_string());
```
0x009922, line 9: Not sure if Swarm should bind to a particular branch (e.g. `iroha2-dev`, `iroha2-stable` etc). IMO, it makes more sense to bind to the same revision Kagami is built itself, whatever branch it is. Also, I think I wrote this in Kagami CLI help.

DCNick3: I _think_ using `HEAD` as revision specification here would do the necessary thing

ilchu: The shallow git clone for the validator fails in case of using just `HEAD` as it tries to grab the wrong SHA - presumably on only this PR, but still. This whole contraption is there to circumvent it. I guess I'll have to go for `HEAD~1`. This is likely to break CI for PRs with many commits though..

DCNick3: Yeah, that makes sense.. There's also `git merge-base` (don't know gix-speak for it). Though, I am not sure if using older version than the PR contains even makes sense for a CI check..

appetrosyan:
Closed as the issue lost relevance.

PR #3484, opened 2023-05-17
[refactor] #2817: Remove interior mutability from `WorldStateView`

Erigara:
## Description

Changes:

- Add wsv related benchmarks;
- Remove interior-mutability from wsv;
- Optimize cloning of wsv for different actors;
- ~~Add persistent structures for wsv~~;

### Linked issue

Closes #3472 

### Bemchmarks

```text
base:
    time:   [1.6779 s 1.6927 s 1.7093 s]
no interior mutability: 
    time:   [1.2139 s 1.2439 s 1.2818 s]
persitent structures:
    time:   [12.324 s 12.386 s 12.446 s]
```

### Notes 

Persistent structures were removed from this PR, because they decrease performance quite significantly (at least in `rpds` crate). 

### Benefits

- New benchmarks.
- No interior mutability in the WSV, no potential for deadlock.
- Performance for block application is a bit better.
- ~~Clone is very cheep fir wsv and constant in time~~.

### Downsides

Cloning is not as cheep as with persistent structures, but amount of clones could be minimized: 
What's more is that cloning could be minimized:
1. Partially i did this in this PR for block_sync and gossiper actors, they clone wsv once per block.
2. If @mversic idea with wasm executor will be pulled of, we wouldn't need to clone wsv for verification.
3. For `torii` access to wsv could be done through intermediate actor and communicate through channels.

### Checklist

- [x] I've read `CONTRIBUTING.md`
- [x] I've used the standard signed-off commit format (or will squash just before merging)
- [x] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

PR #3723, opened 2023-07-19
[feature] #3371: Split validator entrypoint

### LINKED ISSUE #3371: Separate WASM functions for transaction, instruction and query validation
Currently our WASM Runtime Permission Validator (Verifier in future) has only one entrypoint which accepts `NeedsPermissionBox`.

As @mversic [mentioned](https://github.com/hyperledger/iroha/pull/3329#discussion_r1152922013) it will be good to split this functions and we decided to make it three different functions. That way we can link them to user smartcontracts pretty much directly.

Arjentix:
# Review only the latest commit!!!

This PR is based on top of #3707 which is not currently merged. So ignore the commit from #3707.

## Description

`validate()` entrypoint is splitted into:
1. `validate_transaction()`
2. `validate_instruction()`
3. `validate_query()`

`validate_query()` now is allowed to take only `&wsv`, not `&mut wsv`.

Also removed `WorldStateView::upgraded_validator` field, as it was redundant after #3484.

### Linked issue

- Closes #3371  

### Benefits

- Query validation now can't mutate `wsv`. That removes redundant clonning inside torii and also removes possibility of peers desynchronization (because client queries aren't on chain, but bevore this PR valiator can mutate `wsv` when checking them)
- More clear validators code

### Checklist


wasm/validator/derive/src/entrypoint.rs
QuentinI, line 1: Please update the docs at `wasm/validator/derive/src/lib.rs`

wasm/src/lib.rs
```diff
@@ -179,20 +179,40 @@ pub fn query_triggering_event() -> Event {
     unsafe { decode_with_length_prefix_from_raw(host_query_triggering_event()) }
 }
 
-/// Query an operation which is to be validated
+/// Query [`VersionedSignedTransaction`] to validate.
```
QuentinI, line 182: Why is `query` used for these functions instead of something like `get`? It feels as if it's in some way related to Iroha `Query`

core/src/wsv.rs
```diff
@@ -73,9 +74,22 @@ pub struct World {
...
+    fn default() -> Self {
+        Self {
+            parameters: Parameters::default(),
+            trusted_peers_ids: PeersIds::default(),
+            domains: DomainsMap::default(),
+            roles: crate::RolesMap::default(),
+            account_permission_tokens: crate::PermissionTokensMap::default(),
+            permission_token_definitions: crate::PermissionTokenDefinitionsMap::default(),
+            triggers: TriggerSet::default(),
+            validator: Some(Validator),
```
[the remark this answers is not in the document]
Arjentix: Oh, I keep receiving this question everytime.

The problem is that we have unit-tests in `core`. At the same time `Validator` should be supplied by user at runtime (in genesis or in `Upgrade` instruction). Without `validator` set to `Some` unit-tests become kind of useless (and failing), because every transaction in tests requires validation. So we have a `MockValidator` which is aliased to `Validator` when `#[cfg(test)]`. It preforms no validation, but just executes provided instructions and queries (remember that *Validator* is in fact *Executor* and we will rename it in future).

I'll add a comment in this code.

wasm/src/lib.rs
Arjentix: I'm not particularly on `query_` side, but it's named so for historical reasons, I believe. Need other opinions here.
@mversic , @Erigara  ?

Erigara (commented):
LGTM, however as i understand #3707 should be merged first.

core/src/wsv.rs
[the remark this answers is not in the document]
QuentinI: @Arjentix makes sense, I'm okay with that if there's a comment.
I like @Erigara's idea, but that's out of scope for this PR of course.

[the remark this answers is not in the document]
appetrosyan: Created #3752

wasm/src/lib.rs
```diff
@@ -184,20 +183,40 @@ pub fn query_triggering_event() -> Event {
...
 ///
-/// Host side will generate a trap if this function was not called from a
-/// validator `validate()` entrypoint.
-pub fn query_operation_to_validate() -> NeedsValidationBox {
-    #[cfg(not(test))]
-    use host::query_operation_to_validate as host_query_operation_to_validate;
-    #[cfg(test)]
-    use tests::_iroha_wasm_query_operation_to_validate_mock as host_query_operation_to_validate;
+/// Host side will generate a trap if this function was called not from a
+/// validator `validate_transaction()` entrypoint.
```
0x009922, line 191: Same as here: https://github.com/hyperledger/iroha/pull/3707#discussion_r1277071830

core/src/smartcontracts/wasm.rs
```diff
@@ -67,8 +74,12 @@ mod import {
...
     /// Name of the imported function to query event that triggered the smart contract execution
     pub const QUERY_TRIGGERING_EVENT_FN_NAME: &str = "query_triggering_event";
-    /// Name of the imported function to query operation that is to be verified
-    pub const QUERY_OPERATION_TO_VALIDATE_FN_NAME: &str = "query_operation_to_validate";
+    /// Name of the imported function to query transaction that is to be verified
+    pub const QUERY_TRANSACTION_TO_VALIDATE_FN_NAME: &str = "query_transaction_to_validate";
+    /// Name of the imported function to query instruction that is to be verified
+    pub const QUERY_INSTRUCTION_TO_VALIDATE_FN_NAME: &str = "query_instruction_to_validate";
+    /// Name of the imported function to request query that is to be verified
+    pub const QUERY_QUERY_TO_VALIDATE_FN_NAME: &str = "query_query_to_validate";
```
0x009922, line 82: Ah, here they are. Would be great to restructure these consts so that it is possible to reuse them even in the validator macros.

core/src/smartcontracts/wasm.rs
```diff
@@ -379,43 +388,52 @@ pub mod state {
...
 
-    /// Trait to retrieve common data from concrete state
-    pub trait GetCommon<'wrld>: LogSpan + LimitsMut {
-        /// Get common data
-        fn common(&self) -> &Common<'wrld>;
+    /// Trait to immutable reference to [`WorldStateView`]
+    pub trait Wsv {
+        /// Get wsv
+        fn wsv(&self) -> &WorldStateView;
+    }
```
0x009922, line 395: > Trait to _get an_ immutable reference to [`WorldStateView`]

> Get immutable [`WorldStateView`]

_Cannot leave an edit suggestion while reviewing a commit_

core/src/smartcontracts/wasm.rs
```diff
@@ -841,113 +929,323 @@ impl<'wrld> Runtime<state::Trigger<'wrld>> {
...
+        debug!(%instruction, "Executing as Trigger");
+
+        // TODO: Validation should be skipped when executing smart contract.
+        // There should be two steps validation and execution. First smart contract
+        // is validated and then it's executed. Here it's validating in both steps.
+        // Add a flag indicating whether smart contract is being validated or executed
+        let wsv: &mut WorldStateView = state.common.wsv;
+        wsv.validator()
+                .clone() // Cloning validator is a cheap operation
+                .validate_instruction(wsv, &state.common.authority, instruction)
```
0x009922, line 973: This code fragment literally replicates the fragment on lines +888 to +895. Since the expanded comment is also replicated, suggest moving it into a separate function.

core/src/smartcontracts/wasm.rs
```diff
@@ -992,56 +1290,38 @@ impl<'wrld> Runtime<state::validator::Migrate<'wrld>> {
...
+/// Panics with error message if called, because it should never be called from
+/// `migrate()` entrypoint.
+impl<'wrld> import::QueryOperationsToValidate<state::validator::Migrate<'wrld>>
+    for Runtime<state::validator::Migrate<'wrld>>
+{
+    #[codec::wrap]
+    fn query_transaction_to_validate(
+        _state: &state::validator::Migrate<'wrld>,
+    ) -> VersionedSignedTransaction {
+        panic!("Validator `migrate()` entrypoint should not query transaction to validate")
```
0x009922, line 1314: Another place where "query" looks confusing (for me). "get transaction" feels more natural.

0x009922 (commented):
I think there is too much hard code (mostly in error messages and doc comments) with exact validator entrypoints. Although it might be solved in a separate PR.

core/src/smartcontracts/wasm.rs
Arjentix: Created #3757

core/src/smartcontracts/wasm.rs
```diff
@@ -465,58 +473,107 @@ pub mod state {
...
 
         use super::*;
 
-        /// State for executing `validate()` entrypoint of validator
-        pub struct Validate<'wrld> {
+        /// Struct to encapsulate common state for `validate_transaction()` and
+        /// `validate_instruction()` entrypoints.
+        ///
+        /// *Mutably* means that [`WorldStateView`] will be mutated.
+        pub struct ValidateMutably<'wrld, T> {
```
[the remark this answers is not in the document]
QuentinI: @Arjentix

wasm/validator/derive/src/entrypoint.rs
QuentinI: @Arjentix

core/src/wsv.rs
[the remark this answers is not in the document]
QuentinI: @Arjentix docs

QuentinI (approved):

wasm/validator/derive/src/entrypoint.rs
```diff
@@ -140,11 +128,7 @@ fn impl_migrate_entrypoint(attr: &TokenStream, fn_item: syn::ItemFn) -> TokenStr
 
     assert!(
         matches!(sig.output, syn::ReturnType::Type(_, _)),
-        "Validator `permission_tokens()` entrypoint must have `MigrationResult` return type"
-    );
-    assert!(
-        attr.is_empty(),
-        "`#[entrypoint]` macro for Validator `migrate` entrypoint accepts no attributes"
+        "Validator `migrate()` entrypoint must have `MigrationResult` return type"
```
appetrosyan, line 131: One trick that you could do here is to insert the actual type with the fully qualified path. What this will do is also type check the output of the function.

default_validator/src/lib.rs
```diff
@@ -1,38 +1,44 @@
 //! Iroha default validator.
+
```
appetrosyan, line 2: This file is a little confusing. A while ago I fully expected to find the entire validator here as a reference implementation. 

Why did you decide to make this a separate crate, but also not to provide the actual logic inside of it? Wouldn't it make more sense to inline at least the top level functions instead of just having one line per function?

core/src/smartcontracts/wasm.rs
```diff
@@ -64,11 +71,15 @@ mod import {
     /// Name of the imported function to execute queries
     pub const EXECUTE_QUERY_FN_NAME: &str = "execute_query";
     /// Name of the imported function to query trigger authority
-    pub const QUERY_AUTHORITY_FN_NAME: &str = "query_authority";
+    pub const GET_AUTHORITY_FN_NAME: &str = "get_authority";
     /// Name of the imported function to query event that triggered the smart contract execution
-    pub const QUERY_TRIGGERING_EVENT_FN_NAME: &str = "query_triggering_event";
-    /// Name of the imported function to query operation that is to be verified
-    pub const QUERY_OPERATION_TO_VALIDATE_FN_NAME: &str = "query_operation_to_validate";
+    pub const GET_TRIGGERING_EVENT_FN_NAME: &str = "get_triggering_event";
```
appetrosyan, line 76: I feel like this form of manual namespacing is part of the reason why Rust so heavily encourages modules. I propose:
```rust 
pub mod fn_names {
    QUERY_TRIGGERING_EVENT: &str = "query_triggering_event",
}
```

Another point here is that the convention should perhaps be to take the common name and using the fact that the `mod` is a namespace just downcase the `stringify!` or even better, just `#[allow]` constants to not be uppercase and have equality between the variable and its string value. Maybe even to the point of just using the variable name and construcing the `&str` in place when necessary?

appetrosyan (approved):
I'd love to have a more thorough review, but unfortunately, I believe that it is better to have a stable API by the time we release RC18 and to adjust RC19 accordingly.

wasm/validator/derive/src/entrypoint.rs
appetrosyan: Closed in favour of #3764

default_validator/src/lib.rs
appetrosyan: Closed in favour of #3765

wasm/validator/derive/src/entrypoint.rs
Arjentix: @appetrosyan , this is actually done in line 143

PR #3727, opened 2023-07-20
[refactor] introduce syn 2.0 + manyhow showcase

DCNick3:
## Description

This PR adds `syn` 2.0 to the workspace (under the name of `syn2`, to allow co-existence of different versions).

It then updates `iroha_version_derive` to use syn2, as well as moving from unmaintained `proc-macro-error` to `manyhow`.

There are some rough edges in the API, but they are pretty minor and, seemingly, will be resolved soon (see https://github.com/ModProg/manyhow/issues/2#issuecomment-1643604170, https://github.com/ModProg/manyhow/issues/6).

### Benefits

Unlike `proc-macro-error`, `manyhow` is actively maintained and has cleaner API that does not rely on global state.

### Checklist

- [ ] Replace uses of `Accumulator` with upstream `Emitter`


version/derive/Cargo.toml
```diff
@@ -11,11 +11,14 @@ license.workspace = true
...
-syn = { workspace = true, features = ["full"] }
+syn2 = { workspace = true, features = ["default", "full"] }
 quote = { workspace = true }
 proc-macro2 = { workspace = true }
-proc-macro-error = { workspace = true }
-rustc-hash = { workspace = true }
+
+# using fork for `darling` support
+# Tracked at https://github.com/ModProg/manyhow/pull/7
+manyhow = { git = "https://github.com/DCNick3/manyhow.git", rev = "3f48c16fbf6fa8f45dd9428ff191722863867be1", features = ["darling"] }
```
appetrosyan, line 20: Ditto. This feels like we'll get into shape really soon.

version/derive/src/lib.rs
```diff
@@ -91,100 +82,64 @@ pub fn version_with_json(attr: TokenStream, item: TokenStream) -> TokenStream {
...
-            } else {
-                None
-            }
-        })
-        .collect();
+#[derive(FromMeta)]
+struct VersionArgs {
+    n: u32,
+    versioned: syn2::Ident,
+}
```
appetrosyan, line 112: Just give them the right names. `n` is **very** confusing, especially once it becomes ambiguous.

version/derive/src/lib.rs
```diff
@@ -91,100 +82,64 @@ pub fn version_with_json(attr: TokenStream, item: TokenStream) -> TokenStream {
...
 
-    for name in args_map.keys() {
-        if ![VERSION_NUMBER_ARG_NAME, VERSIONED_STRUCT_ARG_NAME].contains(&name.as_str()) {
-            abort!(name.span(), "Unknown field");
+fn impl_version(args: TokenStream, item: TokenStream) -> Result<TokenStream> {
+    let args = NestedMeta::parse_meta_list(args)?;
+    let VersionArgs {
+        n: version_number,
+        versioned: versioned_struct_name,
```
appetrosyan, line 118:
```suggestion
       version_number,
       versioned_struct_name,
```

Like this.

version/derive/tests/ui_fail/unknown_field_in_version_attribute.stderr
```diff
@@ -1,10 +1,8 @@
-error: Unknown field
- --> tests/ui_fail/unknown_field_in_version_attribute.rs:7:1
+error: Unknown field: `derive`
+ --> tests/ui_fail/unknown_field_in_version_attribute.rs:7:50
   |
 7 | #[version(n = 1, versioned = "VersionedMessage", derive = "Clone")]
-  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-  |
-  = note: this error originates in the attribute macro `version` (in Nightly builds, run with -Z macro-backtrace for more info)
+  |                                                  ^^^^^^^^^^^^^^^^
```
appetrosyan, line 5: How far can we push `manyhow`. It looks like we would be able to provide more context and more details in the error message. 

E.g. `unknown field: ``derive`` `, the macro `version` can only accept `n`, `versioned` (see previous comment about naming convention.

version/derive/src/lib.rs
appetrosyan: Or better: 
```suggestion
semver,
versioned_alias,
```

version/derive/src/lib.rs
DCNick3: Should I just change the macro interface?

version/derive/tests/ui_fail/unknown_field_in_version_attribute.stderr
DCNick3: Hmm.. I think it's possible to provide a manual hint with what the `#[version]` macro should look like.

DCNick3: tbh, I don't think it's a good idea. If we just add a blanket reminder what the attribute should look like for any error, it risks giving irrelevant messages.

version/derive/src/lib.rs
appetrosyan: Yes please

version/derive/tests/ui_fail/unknown_field_in_version_attribute.stderr
appetrosyan: This is not a hard requirement, I'll approve regardless of if it's implemented. 

However, AFAICT, a lot of our attributes are simple key-value pairs and for keys which are known ahead of time. For this use case, all you need to do is add a generic method to show a message for the recognised attributes. If all this does is print `help` and each such proc macro has a `help` message it does the job.

appetrosyan: Created #3736

appetrosyan (dismissed):

appetrosyan (dismissed):

appetrosyan (dismissed):

macro/derive/src/lib.rs
```diff
@@ -12,19 +15,117 @@ const SKIP_TRY_FROM_ATTR: &str = "skip_try_from";
...
+    skip_from: bool,
+    skip_try_from: bool,
+    skip_container: bool,
+}
+
+// implementing manually, because darling can't parse attributes that are not under some unified attr
+// It expects us to have a common attribute that will contain all the fields, like:
+// #[hello(skip_from, skip_container)]
+// The already defined macro API uses `skip_from` and `skip_container` attributes without any qualification
+// Arguably, this is also more convenient for the user (?)
```
Erigara, line 106: Imo `#[from_variant(skip_from, ...)]` is more readable since we can clearly see relation between derive macro and attribute.

Erigara (approved):
Good job!

macro/derive/src/lib.rs
DCNick3: That's true. It is a bit more wordy, but the `skip_*` attrs are not used that much..

macro/derive/src/lib.rs
```diff
@@ -189,94 +292,47 @@ fn try_into_variant(
...
-            const VARIANT_COUNT: usize = #variant_count;
-        }
-    }
-    .into()
+            let span = variant.span();
+            let field = variant
+                .fields
+                .iter()
+                .next()
+                .expect("BUG: this should've been checked when parsing FromVariantInput");
```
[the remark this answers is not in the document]
DCNick3: I don't think so, not with today's darling

appetrosyan (approved):

Erigara (approved):

PR #3738, opened 2023-07-24
(more) Syn 2.0 updates

DCNick3:
## Description

NOTE: This PR is based on #3727, so only last 3 commits are _actually_ belong to this PR

This PR refactors `iroha_primitives_derive` crate and `VariantCount` macro in `iroha_derive`. The gist of the changes: use `manyhow` to report errors and `darling` to parse attributes. Try to improve UX where possible. 

I've kept API the same for all the macros, but it may be useful to change some of them: 
- confusingly, attributes for `VariantCount` are placed on fields instead of enum variants. I suggest we move them to variants, as it's really a property of the variant
- the naming of `#[skip_try_from]` is confusing in the context of `#[skip_from]`: the different in names is in the `try`, but they correspond to different directions of the generated conversions: `from` is for `Variant` ->  `Enum` and `try_from` is for `Enum` -> `Variant`. I suggest changing the name to `#[skip_into]` or `#[skip_try_into]` to properly represent the change of conversion direction, even if the actual generated trait is `From`
- Maybe change `#[skip_container]` to `#[skip_from_container]`?

I would like some input on this from other team members though ^^

### Checklist

- [ ] #3727 merged
- [ ] All macros documented properly
- [ ] API changes figured out


DCNick3:
The changes are now part of #3727

PR #3759, opened 2023-07-28
[refactor]: update `iroha_schema_derive` to use syn 2.0

DCNick3:
Also parse attributes with `darling` and handle errors with `manyhow`

## Description

This is continuation of work started in #3727, touching on `iroha_schema_derive`.

There was an unused API in `TypeId` derive: `#[type_id(bound = "...")]`, I removed support for it for now.

### Benefits

Same as before: less code, better error messages.

### Checklist

- [x] update the `iroha_schema_derive` crate
- [x] add more ui tests covering all the API
- [ ] Address comments

### DIFF
--- a/schema/derive/tests/ui_pass/derive_into_schema.rs
+++ b/schema/derive/tests/ui_pass/derive_into_schema.rs
@@ -37,7 +37,6 @@ pub struct FakeString {}
 #[derive(IntoSchema)]
 pub enum Enum {
     Zero,
-    // how does it work? We don't allow discriminants, but still try to parse them?
     One,
     #[codec(index = 42)]
     FortyTwo,
--- a/schema/tests/schema_json.rs
+++ b/schema/tests/schema_json.rs
@@ -1,0 +1,235 @@
+//! This test checks how the json-serialized schema looks like.
+
+#![allow(dead_code)]
+#![allow(unused_tuple_struct_fields)]
+
+use iroha_schema::IntoSchema;
+use serde_json::json;
+
+/// It expects to have three parameters: a type definition item, an expected schema type name and a JSON schema.
+///
+/// The json is passed to the `serde_json::json!` macro, so it can be a string, an array or an object.
+///
+/// Only the schema of the type itself is checked, not the schema of its fields.
+///
+/// NOTE: this macro doesn't support generics.
+macro_rules! check_schema {
+    ($(#[$($meta:tt)*])* struct $ty:ident, $type_name:ident, $json:tt) => {{
+        #[derive(IntoSchema)]
+        $(#[$($meta)*])*
+        struct $ty;
+        check_schema!(@impl $ty, $type_name, $json);
+    }};
+    ($(#[$($meta:tt)*])* struct $ty:ident ($($body:tt)*), $type_name:ident, $json:tt) => {{
+        #[derive(IntoSchema)]
+        $(#[$($meta)*])*
+        struct $ty($($body)*);
+        check_schema!(@impl $ty, $type_name, $json);
+    }};
+    ($(#[$($meta:tt)*])* struct $ty:ident {$($body:tt)*}, $type_name:ident, $json:tt) => {{
+        #[derive(IntoSchema)]
+        $(#[$($meta)*])*
+        struct $ty {$($body)*}
+        check_schema!(@impl $ty, $type_name, $json);
+    }};
+    ($(#[$($meta:tt)*])* enum $ty:ident {$($body:tt)*}, $type_name:ident, $json:tt) => {{
+        #[derive(IntoSchema)]
+        $(#[$($meta)*])*
+        enum $ty {$($body)*}
+        check_schema!(@impl $ty, $type_name, $json);
+    }};
+    (@impl $ty:ident, $type_name:ident, $json:tt) => {{
+        assert_eq!(
+            $ty::type_name(),
+            stringify!($type_name),
+            "Type name of {} is not equal to the expected one",
+            stringify!($ty)
+        );
+        let __schema = serde_json::value::to_value(&$ty::schema())
+            .expect("Failed to serialize schema to JSON");
+        assert_eq!(
+            __schema.get(&$ty::type_name()).unwrap().clone(),
+            json!($json),
+            "Schema of {} is not equal to the expected one",
+            stringify!($ty)
+        );
+    }};
+}
+
+#[test]
+fn test_struct() {
+    check_schema!(
+        struct EmptyNamedStruct {},
+        EmptyNamedStruct,
+        {"Struct": []}
+    );
+
+    // this behaviour is weird...
+    check_schema!(
+        struct EmptyTupleStruct(),
+        EmptyTupleStruct,
+        null
+    );
+    check_schema!(
+        struct UnitStruct,
+        UnitStruct,
+        null
+    );
+
+    check_schema!(
+        struct NormalStruct {
+            normal_field_1: u32,
+            normal_field_2: u32,
+        },
+        NormalStruct,
+        {"Struct": [
+            {"name": "normal_field_1", "type": "u32"},
+            {"name": "normal_field_2", "type": "u32"}
+        ]}
+    );
+    check_schema!(
+        struct NewtypeStruct(u32),
+        NewtypeStruct,
+        "u32"
+    );
+    check_schema!(
+        struct TupleStruct(u32, u32),
+        TupleStruct,
+        {"Tuple": [
+            "u32", "u32"
+        ]}
+    );
+}
+
+#[test]
+fn test_struct_codec_attr() {
+    check_schema!(
+        struct SkipField {
+            #[codec(skip)]
+            skipped_field: u32,
+            normal_field: u32,
+        },
+        SkipField,
+        {"Struct": [
+            {"name": "normal_field", "type": "u32"}
+        ]}
+    );
+    check_schema!(
+        struct CompactField {
+            #[codec(compact)]
+            compact_field: u32,
+        },
+        CompactField,
+        {"Struct": [
+            {"name": "compact_field", "type": "Compact<u32>"}
+        ]}
+    );
+}
+
+#[test]
+fn test_transparent() {
+    check_schema!(
+        #[schema(transparent)]
+        struct TransparentStruct(u32),
+        u32,
+        {"Int": "FixedWidth"}
+    );
+    check_schema!(
+        #[schema(transparent = "u32")]
+        struct TransparentStructExplicitInt {
+            a: u32,
+            b: i32,
+        },
+        u32,
+        {"Int": "FixedWidth"}
+    );
+    check_schema!(
+        #[schema(transparent = "String")]
+        struct TransparentStructExplicitString {
+            a: u32,
+            b: i32,
+        },
+        String,
+        "String"
+    );
+    check_schema!(
+        #[schema(transparent = "String")]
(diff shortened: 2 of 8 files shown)


version/derive/tests/ui_fail/unknown_field_in_version_attribute.stderr
```diff
@@ -1,8 +1,16 @@
+<<<<<<< HEAD
```
appetrosyan, line 1: This is a bug. Please fix.

appetrosyan (changes requested):

schema/tests/transparent_types.rs
```diff
@@ -0,0 +1,73 @@
...
+use iroha_schema::prelude::*;
+use parity_scale_codec::{Decode, Encode};
+
+#[derive(Decode, Encode, IntoSchema)]
+#[schema(transparent)]
+struct TransparentStruct(u32);
+
+#[derive(Decode, Encode, IntoSchema)]
+#[schema(transparent = "u32")]
+struct TransparentStructExplicitInt {
```
appetrosyan, line 14: These test cases need either comments or doc-comments.

schema/tests/schema_json.rs
```diff
@@ -0,0 +1,239 @@
...
+    }};
+}
+
+// NOTE: this doesn't check generic types
+
+/// It expects to have three parameters: a type definition item, an expected schema type name and a JSON schema.
+///
+/// The json is passed to the `serde_json::json!` macro, so it can be a string, an array or an object.
+///
+/// Only the schema of the type itself is checked, not the schema of its fields.
```
appetrosyan, line 35:
```suggestion


/// It expects to have three parameters: a type definition item, an expected schema type name and a JSON schema.
///
/// The json is passed to the `serde_json::json!` macro, so it can be a string, an array or an object.
///
/// Only the schema of the type itself is checked, not the schema of its fields.
///
/// # NOTE
/// This macro doesn't check generic types. 
```

schema/derive/tests/ui_pass/derive_into_schema.rs
```diff
@@ -13,17 +13,34 @@ pub struct Foo2<T, K> {
 
 #[derive(IntoSchema)]
 pub struct Foo3<T: Clone> {
-   _value1: T,
+    _value1: T,
```
appetrosyan, line 16: One baffling decision in `rustfmt` is the fact that it doesn't handle whitespace. A lot of the time the whole reason one wants to have an autoformatter is to ensure that the pesky spaces are inserted.

schema/derive/tests/ui_pass/derive_into_schema.rs
```diff
@@ -13,17 +13,34 @@ pub struct Foo2<T, K> {
...
+pub struct AutoTransparent(u32);
+
+#[derive(IntoSchema)]
+#[schema(transparent = "String")]
+pub struct FakeString {}
+
+#[derive(IntoSchema)]
+pub enum Enum {
+    Zero,
+    // how does it work? We don't allow discriminants, but still try to parse them?
```
appetrosyan, line 40: We used to have those. We decided not to add complexity by allowing custom discriminants but preserve the behaviour for the future. So that technically it all works.

schema/derive/src/lib.rs
```diff
@@ -1,54 +1,36 @@
...
-use quote::quote;
-use syn::{
-    parse::{Parse, ParseStream, Result},
-    parse_macro_input, parse_quote,
-    spanned::Spanned,
-    Attribute, Data, DataEnum, DataStruct, DeriveInput, Expr, Field, Fields, FieldsNamed,
-    FieldsUnnamed, GenericParam, Generics, LitStr, Meta, NestedMeta, Type,
-};
+// darling-generated code triggers this lint
+#![allow(clippy::option_if_let_else)]
```
appetrosyan, line 5: @mversic and I believe that we should stop enforcing the functional style project-wide. It should be reserved for crates that users are likely to touch, and crates which are already written in a functional style for consistency. We should remember to remove those.

schema/derive/src/lib.rs
```diff
@@ -1,54 +1,36 @@
...
+// it's fine to have some functions return Result without returning Err
+// this allows some functions to have more uniform signatures
+#![allow(clippy::unnecessary_wraps)]
```
appetrosyan, line 8: No. 

1) You are lying to the user of your lib. 
2) While sometimes the return of the result is required because of a function signature, you don't have it in a macro crate. You are also extremely unlikely to have it in a macro crate all things considered. 
3) If it does happen you return `Result<Thing, Infallible>` customarily, which at least helps the user see that it is not a fallible operation. 
4) If you **did** in fact have the need to return a `Result` because of trait bounds (see point 2), you could do so by taking an infallible function and returning a closure that wraps it in an `Ok()` after evaluation. It's much neater in Haskell than in Rust, true, but you still get the benefit of removing these instances if and when you need them, without actually having to refactor the function signature.

appetrosyan (changes requested):

schema/derive/src/lib.rs
DCNick3: I like when most of my functions return `Result` (with a universal error type), mostly for "symmetry" and for ease of adding error cases in the future (no need to rewrite the signature and call sites), but I am not going to die on this hill.

appetrosyan: Rust did not go with the symmetrical solution of exceptions (panics) for error handling, expressly for this purpose. 

If the function suddenly becomes fallible, the change in the signature requires some downstream modification. If you don't change the signature but make a previously infallible function falllible, you no longer have a nice compilation error to tell you what needs to be done. A function which didn't fail previously suddenly failing now is an API breaking change.

Arjentix: The only reason I understand is future changes. Although for that library developers basically use complex wrappers around real errors, use `#[non_exhaustive]` and etc.

**But** you're not writing extremely portable library for hundreds of users. You're writing our internal crate, it's even not in the crates.io. So I agree with Aleksandr.

primitives/src/conststr.rs
```diff
@@ -271,6 +270,33 @@ impl WrapperTypeDecode for ConstString {
...
 
+// darling doesn't support unions, so this can't be derived
+impl TypeId for ConstString {
+    fn id() -> Ident {
+        #[cfg(not(feature = "std"))]
+        use alloc::string::ToString;
+        "ConstString".to_string()
+    }
+}
+impl IntoSchema for ConstString {
```
Arjentix, line 281:
```suggestion
}

impl IntoSchema for ConstString {
```

primitives/src/conststr.rs
```diff
@@ -271,6 +270,33 @@ impl WrapperTypeDecode for ConstString {
     type Wrapped = String;
 }
 
+// darling doesn't support unions, so this can't be derived
+impl TypeId for ConstString {
+    fn id() -> Ident {
+        #[cfg(not(feature = "std"))]
+        use alloc::string::ToString;
```
Arjentix, line 277:
```suggestion
        #[cfg(not(feature = "std"))]
        use alloc::string::ToString as _;
```

primitives/src/conststr.rs
```diff
@@ -271,6 +270,33 @@ impl WrapperTypeDecode for ConstString {
...
+        "ConstString".to_string()
+    }
+}
+impl IntoSchema for ConstString {
+    fn type_name() -> Ident {
+        #[cfg(not(feature = "std"))]
+        use alloc::string::ToString;
```
Arjentix, line 284: I suggest to import in once at the module level

primitives/src/conststr.rs
```diff
@@ -271,6 +270,33 @@ impl WrapperTypeDecode for ConstString {
...
+        "String".to_string()
```
Arjentix, line 285: Shouldn't it be `ConstString`?

Arjentix: Or better just reuse implementation from `TypeId`

schema/derive/src/lib.rs
```diff
@@ -62,34 +41,123 @@ fn impl_type_id(input: &mut DeriveInput) -> TokenStream2 {
...
+
+    fn from_word() -> darling::Result<Self> {
+        Ok(Self::Transparent(None))
+    }
+
+    fn from_string(value: &str) -> darling::Result<Self> {
+        let ty = syn2::parse_str(value)?;
+        Ok(Self::Transparent(Some(ty)))
+    }
+}
```
Arjentix, line 63: Can't this be derived using darling derives?

schema/tests/schema_json.rs
```diff
@@ -0,0 +1,239 @@
...
+    ($ty:ident, $type_name:ident, $json:tt) => {{
+        assert_eq!(
+            $ty::type_name(),
+            stringify!($type_name),
+            "Type name of {} is not equal to the expected one",
+            stringify!($ty)
+        );
+        let __schema = serde_json::value::to_value(&$ty::schema())
+            .expect("Failed to serialize schema to JSON");
+        // println!("{}: {}", stringify!($ty), __schema);
```
Arjentix, line 19: Commented debug code

schema/tests/schema_json.rs
```diff
@@ -0,0 +1,239 @@
...
+///
+/// Only the schema of the type itself is checked, not the schema of its fields.
+///
+/// NOTE: this macro doesn't support generics.
+macro_rules! check_schema {
+    ($(#[$($meta:tt)*])* struct $ty:ident, $type_name:ident, $json:tt) => {{
+        #[derive(IntoSchema)]
+        $(#[$($meta)*])*
+        struct $ty;
+        check_schema_impl!($ty, $type_name, $json);
```
Arjentix, line 41: Better use [this technique](https://stackoverflow.com/questions/54406388/what-does-an-symbol-mean-in-a-rust-declarative-macro)

primitives/src/conststr.rs
DCNick3: This is how the derive works: it's a transparent type so it reports itself as a `String`. For `TypeId`, however, transparent types do not matter and it reports the actual type name

schema/derive/src/lib.rs
DCNick3: Not with this syntax I believe.. The only things that can be constructed from a word (smth like `#[schema(transparent)]`) are `bool` and `()`, but then there wouldn't be a way to provide an optional explicit type

schema/tests/schema_json.rs
DCNick3: So, instead of defining an `_impl` macro, I should create a matcher with `@impl` containing the same stuff? Sounds neat

schema/derive/src/lib.rs
Arjentix: Ok, then how does this code reflects that only `transparent` word can construct `Transparent` struct?

DCNick3: It doesn't. In `darling` data model the name of the meta matters only in the `FromAttributes` (and alike). So, to change the kind of word used to create the `Transparent` struct, you change the name if the field in the `SchemaAttributes` struct

```rust
#[derive(Debug, Clone, FromAttributes)]
#[darling(attributes(schema))]
struct SchemaAttributes {
    // change the name here
    transparent: Transparent,
}
```

appetrosyan (approved):

appetrosyan (approved):

Arjentix (approved):

PR #3770, opened 2023-07-31
[refactor] #3752: Replace `MockValidator` with `Initial`

### LINKED ISSUE #3752: [refactor]: Replace `MockValidator` with `Default`
Maybe we can design so that when empty wsv is created it's already have minimal validator that only execute instructions (and `Upgrade` instruction in genesis will replace it)?

By doing that we can keep design more uniform.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/3723#discussion_r1272997794_

Arjentix:
## Description

- Removed `MockValidator` and its usage with `#[cfg(test)]`
- `Validator` is now an enum with `Initial` variant set by default
- This `Initial` variant will perform no permission checking, just operation execution

### Linked issue

- Closes #3752 

### Benefits

No divergence between real and test code

### Question

Should we allow genesis without Validator then?

### Checklist


core/src/wsv.rs
```diff
@@ -976,9 +952,7 @@ impl WorldStateView {
     /// Panics if validator is not initialized.
     /// Possible only before applying genesis.
```
Erigara, line 953: I belive panic docs in no longer true, right?

core/src/smartcontracts/isi/world.rs
```diff
@@ -218,32 +218,21 @@ pub mod isi {
...
-                        .migrate(wsv, authority)
-                        .map(|migration_result| migration_result.map(|_: ()| new_validator))
-                })
-                .map_err(|error| {
-                    InvalidParameterError::Wasm(format!("{:?}", eyre::Report::from(error)))
-                })?
+
+            // Cloning validator to avoid multiple mutable borrows.
+            // Also it's a cheap operation.
+            let mut cloned_validator = wsv.validator().clone();
```
Erigara, line 225: Maybe `upgraded_validator`?

Erigara (approved):
Great this change allowed to throw away mocking code.

mversic (dismissed):

core/src/validator.rs
```diff
@@ -77,18 +77,33 @@ impl Validator {
...
-            transaction,
-        )?
+
+        match self {
+            Self::Initial => {
+                let (_authority, Executable::Instructions(instructions)) = transaction.into() else {
+                    return Ok(());
+                };
+                for isi in instructions {
+                    self.validate_instruction(wsv, authority, isi)?;
```
mversic, line 88:
```suggestion
                    isi.execute(authority, wsv).map_err(Into::into)?;
```

I would rather have it like this

mversic (approved):

PR #3772, opened 2023-08-01
[feature] #3276: Add `Log` isi 

### LINKED ISSUE #3276: [suggestion] Instruction to print message in the log
### Feature request

It will be useful to have instruction to print info in the `Iroha 2` logs.

Originally requested by @arndey.

### Motivation

With this it will be easier to debug triggers written only with instructions

### Who can help?

@arjentix

Erigara:
## Description

`Log` is was edded to make debugging of isi only triggers easier.

### Linked issue

Closes #3276  

### Benefits

Debugging should be easier.

### Drawbacks

We add new isi which will need to be maintained.

### How to test 

1. Submit `Log` isi to iroha through `cat log_isi.json | ./iroha_client_cli json`

```json
[
  {"Log": {"LogLevel":"ERROR","msg":{"String":"Cool error"}}}
]
```

2. Grep iroha message for this log message
3. Check that it have appropriate log level

### Migrate WASM to use `Log` isi instead of `log` function

After discussion with @mversic and @Arjentix it was decided not to make this migration because `Log` isi approach face multiple issues in `WASM` context: 

1. We lose `Span`s.
4. Not every WASM entrypoint is allowed to execute instructions at all.
5. Since every instruction is executed by validator we can't distinguish whatever this isi came from smart-contract or validator itself.  

Unless this problems will be solved migration is not desirable.

### Checklist

- [x] Move log `Level` into data model
- [x] Implement `Log` isi


appetrosyan:
I need to create some issues according to the PR description.

Arjentix (approved):

appetrosyan (dismissed):

Arjentix (approved):

appetrosyan (approved):

PR #3775, opened 2023-08-01
[feature] #3641: Human-readable permission token payload

### LINKED ISSUE #3641: Make permission token payload human readable in `genesis.json`
So the bad part is that we loose readability in genesis. Just mentioning

_Originally posted by @Arjentix in https://github.com/hyperledger/iroha/pull/3625#discussion_r1239551590_

Arjentix:
## Description

**Scale** codec for Permission Token Payloads was replaced with **JSON**.

That's the most congruent way to fix the problem of readability. Permission tokens aren't used frequently and their main purpose is to be transparent for users. So their payload should be always readable.

Also did refactor a bit.

### Linked issue

- Closes #3641  

### Checklist


data_model/src/permission.rs
```diff
@@ -91,10 +97,10 @@ impl PermissionTokenSchema {
 
 impl PermissionToken {
     /// Construct [`Self`]
-    pub fn new<T: Encode>(definition_id: PermissionTokenId, payload: &T) -> Self {
+    pub fn new(definition_id: PermissionTokenId, payload: serde_json::Value) -> Self {
         Self {
             definition_id,
-            payload: payload.encode(),
+            payload: StringWithJson(payload),
```
Erigara, line 103: Why not to convert`Value` to the `String` inside constructor? Because common pattern for this struct is `.to_string()` before actual use which might be costly.

Erigara (dismissed):

wasm/validator/src/permission.rs
```diff
@@ -48,6 +51,8 @@ pub enum PermissionTokenConversionError {
     // TODO: Improve this error
     /// Unexpected parameter value.
     Value(alloc::string::String),
```
mversic, line 53: Do we need `Value` and `Param` variants anymore?

wasm/validator/Cargo.toml
```diff
@@ -20,6 +20,7 @@ iroha_validator_derive = { version = "2.0.0-pre-rc.18", path = "derive" }
 anyhow = {version = "1.0.71", default-features = false, optional = true}
 
 parity-scale-codec = { version = "3.1.5", default-features = false }
```
mversic, line 22: do we still need SCALE?

mversic (changes requested):

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -113,7 +116,10 @@ impl CustomValidator {
             .iter()
             .try_for_each(|(account, domain_id)| {
                 RevokeBox::new(
-                    PermissionToken::new(can_unregister_domain_definition_id.clone(), domain_id),
+                    PermissionToken::new(
+                        can_unregister_domain_definition_id.clone(),
+                        json!(domain_id),
```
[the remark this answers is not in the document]
mversic: no big deal. Especially if you apply [this](https://github.com/hyperledger/iroha/pull/3775#discussion_r1281449756) comment. Then it's hidden away

data_model/src/permission.rs
Arjentix: The main point is `Serialize` and `Deserialize` derives. I'd like it to be consistent with other json in `genesis.json`. It will be weird if it will be string in json format inside another json

Arjentix: I'll change the name

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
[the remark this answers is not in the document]
Arjentix: > it feels like we're tied to serde_json

Yes, as well, as we tied to `parity_scale_codec`. But I don't see much problems here.
The problem of keeping `String` is that it will be Serialized/Deserialized as string, not as JSON. See my response for Roman's comment

tools/kagami/src/genesis.rs
```diff
@@ -163,15 +164,18 @@ pub fn generate_default(validator_path: Option<PathBuf>) -> color_eyre::Result<R
...
+        .expect("At least one transaction is expected");
+    for isi in [
+        mint.into(),
+        mint_cabbage.into(),
+        grant_permission_to_set_parameters.into(),
+        parameter_defaults.into(),
+        register_user_metadata_access,
+    ] {
+        first_tx.push_instruction(isi);
+    }
```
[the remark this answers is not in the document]
Arjentix: When how do we say to which transaction we append instruction? Genesis contains transactions, not instructions directly

wasm/validator/Cargo.toml
Arjentix: Indeed, we don't

wasm/validator/src/permission.rs
Arjentix: You're right, we don't

data_model/src/permission.rs
Erigara: > The main point is `Serialize` and `Deserialize` derives. I'd like it to be consistent with other json in `genesis.json`. It will be weird if it will be string in json format inside another json

We can provide our own `Serialize` and `Deserialize` which will do conversion `Serialized` -> `serde_json::Value` -> `String` (and wise versa). What do you think?

tools/kagami/src/genesis.rs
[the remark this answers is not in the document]
mversic: as far as I am aware it always contains only single transaction

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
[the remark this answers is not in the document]
mversic: let's move discussion to that comment

tools/kagami/src/genesis.rs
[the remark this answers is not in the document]
mversic: and in case there are multiple why exactly would we change specifically first transaction of all transactions?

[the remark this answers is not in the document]
Arjentix: > as far as I am aware it always contains only single transaction

Yes, we just always use corner case.

> and in case there are multiple why exactly would we change specifically first transaction of all transactions?

I think there is no real reason.

But having `append_instruction()` in `genesis` does not tell me that it will append instruction to the latest transaction.

data_model/src/permission.rs
Arjentix: And what's the point of doing that if it would be the same, we just move conversion from `Encode/Decode` to `Serialize/Deserialize`?

While having `serde_json::Value` protects us from currupting json with out actions, which `String` cannot provide

data_model/src/permission.rs
```diff
@@ -117,6 +125,57 @@ impl core::fmt::Display for PermissionToken {
...
 }
 
+impl Encode for EncodableJson {
+    fn encode_to<W: parity_scale_codec::Output + ?Sized>(&self, dest: &mut W) {
+        self.0.to_string().encode_to(dest)
+    }
+
+    fn size_hint(&self) -> usize {
+        self.0.to_string().size_hint()
+    }
```
0x009922, line 135: Encoding JSON twice might be expensive. Although it is not critical now.

0x009922 (approved):

data_model/src/permission.rs
Erigara: > And what's the point of doing that if it would be the same, we just move conversion from Encode/Decode to Serialize/Deserialize?

We can validate that string which we received is valid json without actually parsing it:

```rust
let _: serde::de::IgnoredAny = serde_json::from_str(data)?;
```

> While having serde_json::Value protects us from currupting json with our actions, which String cannot provide

Which actions may corrupt permission token payload?

And we anyway have to deserialize/decode permission token inside validator, so we can verify that we received valid json.

tools/kagami/src/genesis.rs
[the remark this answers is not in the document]
mversic: ok, it's not related to this PR anyhow

data_model/src/permission.rs
Arjentix: Will address this under [this](https://github.com/hyperledger/iroha/pull/3775#discussion_r1287191647) discussion

data_model/src/permission.rs
Arjentix: Check out new code

Erigara (approved):

0x009922 (dismissed):

0x009922 (approved):

Arjentix:
@mversic , waiting for your approval

mversic (approved):

PR #3784, opened 2023-08-07
[feature] #3781: Validate genesis transactions

### LINKED ISSUE #3781: Validate genesis transactions
Currently we do not validate transactions in genesis for historical reasons.

There are a couple of problems with that:

1. We cannot validate permission tokens usage in genesis
2. Validator may want to have all transactions to be passed through it so that it can store some information inside

The solution is to validate genesis transaction but with some additional info, so that Validator could know that it's from genesis.
@mversic proposed to pass block number to the validator. Seems a good solution to me, because with that users can also implement some interesting logic like "Let users do X after the block N" (like a permission delayed in time).

Arjentix:
## Description

- Added new parameter `block_height` for all Validator entrypoints
- `block_height` can also be achieved from Smartcontracts and Triggers but it's not mandatory in entrypoint signature
- Implemented real genesis validation on Iroha side
- Added test

### Linked issue

- Closes #3781 

### Benefits

- Genesis Permission Tokens validation
- Validator will get **all** transactions, which it may want to be able to configure its state somehow

### Checklist


client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
```diff
@@ -116,34 +122,39 @@ impl ExpressionEvaluator for CustomValidator {
 }
 
 #[entrypoint]
-pub fn migrate() -> MigrationResult {
-    CustomValidator::migrate()
+pub fn migrate(block_height: u64) -> MigrationResult {
+    CustomValidator::migrate(block_height)
```
appetrosyan, line 126: This seems like a costly oversight. We should have a model for the upgrade mechanism

wasm/validator/src/lib.rs
```diff
@@ -168,11 +168,14 @@ pub trait Validate: Visit {
     ///
     /// If `migrate()` entrypoint fails then the whole `Upgrade` instruction
     /// will be denied and previous validator will stay unchanged.
-    fn migrate() -> MigrationResult;
+    fn migrate(block_height: u64) -> MigrationResult;
 
-    /// Validator verdict.
+    /// Get validator verdict.
```
appetrosyan, line 173: Revert

wasm/validator/src/lib.rs
```diff
@@ -168,11 +168,14 @@ pub trait Validate: Visit {
...
     fn verdict(&self) -> &Result;
 
+    /// Get current block height.
```
appetrosyan, line 176:
```suggestion
    /// Current block height.
```

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
```diff
@@ -116,34 +122,39 @@ impl ExpressionEvaluator for CustomValidator {
...
 
     validator.visit_instruction(&authority, &instruction);
 
     validator.0.verdict
 }
 
 #[entrypoint]
-pub fn validate_query(authority: AccountId, query: QueryBox) -> Result {
-    let mut validator = CustomValidator(DefaultValidator::new());
+pub fn validate_query(authority: AccountId, query: QueryBox, block_height: u64) -> Result {
```
0x009922, line 156: Seems like `authority`, `query` and `block_height` (and whatever will appear in future) should be packed into a single `ValidateQueryEntrypointPayload` struct. Same applies to other entrypoints (i.e. `migrate()`, `validate_instruction()` etc).

Erigara (commented):
Do i understand correctly that we actually `run` validator inside during genesis round because `Upgrade<Validator>` is the first instruction in genesis and after that every isi will be validated using runtime validator?

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
Erigara: Agree with @0x009922. We can pack this fields into the struct inside `#[entrypoint]` macro for example.

Arjentix:
> Do i understand correctly that we actually run validator inside during genesis round because Upgrade<Validator> is the first instruction in genesis and after that every isi will be validated using runtime validator?

Yes.

We could remove mandatory validator in genesis. Then all instructions will be validated by `Initial` validator, which just executes instructinos

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
Arjentix: - Created #3791

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
Arjentix: I don't get it. Can you explain more, please?

wasm/src/lib.rs
```diff
@@ -292,6 +299,13 @@ mod host {
         /// This function does transfer ownership of the result to the caller
         pub(super) fn get_query_to_validate() -> *const u8;
 
+        /// Get current block height of the chain.
+        ///
+        /// # Warning
+        ///
+        /// This function does transfer ownership of the result to the caller
+        pub(super) fn get_block_height() -> *const u8;
```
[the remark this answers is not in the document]
Arjentix: We use our own codec to pass data between Iroha and Wasm. It was introduced by @mversic. Basically we just encode every time using *parity scale codec*.
I'm not sure that bypassing this codec makes much sence for this edge case

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
Arjentix: Yes, previously we could do that using inderect proofs. Right now we have `block_height` and `DefaultValidator` will fail to migrate if it was uploaded not as the first block

Erigara (approved):

appetrosyan (approved):

PR #3788, opened 2023-08-08
[fix] #2245: Nix build iroha node binary as AppImage

### LINKED ISSUE #2245: Add packaging for Linux and MacOS
After almost 12 hours of incessant work on one of the DevOps machines (s6.stg1), it was found that the system requirements to compile any of the supporting binaries for Linux requires a significant amount of computational resources. 

As a stop-gap solution the executable can be taken directly from the pipeline (provided it's compiled with the `musl` libc support, but in the long run, we must consider adding a portable executable format to the pipeline. 

There are several major candidates
1. [AppImage](https://appimage.org/)
2. [snap](https://snapcraft.io/)
3. [Nix](https://nixos.org/)

Each have their own unique advantages, and disadvantages. 

My personal preference is with Appimage, for the simple reason that it requires no host infrastructure aside from a working libc, POSIX shell and an ELF-compatible executable format. Distributing updates through this channel is a little problematic but doable. It also doesn't bind us to any packaging standard, any packaging distribution format and almost everything can be done in-house if we need to. 

Snap (and to a lesser extent Flatpak) are not designed with distribution of server-grade applications in mind. As such they mostly offer desktop-oriented (but still portable) features. Packaging for Flatpak is relatively easier, while Snap is more flexible in configuration. As we already have docker, offering either of the two packages is a little redundant, but offers much lower overhead in general. 

Nix is a much more refined approach, and would be the ideal long-term solution, whilst also requiring the biggest short-term change to our workflow. Writing a `nix` package requires knowledge of the `nix` programming language. While, personally, I'd be happy to do that (I have more contributions to NixOS than any other GNU+Linux distribution), it'll require every single person involved with the CI/CD to also know the language. 

Nix is used to great effect in the functional programming world and is the preferred DevOps platform for Scrive. It alone can cover the ground of both Docker, a proper system-wide package manager (no overhead), and cover the major unix-like platforms Linux and MacOS support running Nix packages natively. Even if we decide not to go with Nix in this particular instance, it might be a good idea to revisit at some point after the release of the LTS.

SamHSmith:
## Description

I have found an AppImage bundler for nix and converted it for use in our nix flake. You can now run `nix build .#appimage` to build the iroha node binary as an AppImage. Ideally we would build all the projects binaries but for now that was outside of my skills with nix.

### Linked issue

Partially closes #2245  

### Checklist


QuentinI:
Why did you choose to add a copy of `nix-appimage` to the iroha repo?

SamHSmith:
> Why did you choose to add a copy of `nix-appimage` to the iroha repo?

Because it was only available to use from the commandline. It was not built to be used from within a derivation. So I modified it to support that usecase.

appetrosyan (approved):

Issue #3240, opened 2023-02-21
Guard against secrets leakage


mversic:
Make use of the [secrecy](https://docs.rs/secrecy/latest/secrecy/) crate to prevent the secrets like `PrivateKey`(are there any other?) from being leaked accidentally by being printed in logs or serialized. `Secret` wrapper also makes sure that upon dropping the field is [zeroize](https://docs.rs/zeroize/latest/zeroize/index.html)d in memory

DCNick3:
I don't think it's feasible at least before we remove the config documentation from the API (as planned by hyperledger/iroha-rfcs#8). Removing `Serialize` impl from the key seems to have troubles with the `iroha_config_base::proxy::Documented` trait. It'll be easier to add secrecy after the removal than to try to work around that now IMO.

0x009922:
`Documented` trait will be removed by:

- https://github.com/hyperledger/iroha/pull/4127

nxsaken:
@dima74 should this be closed?

Issue #2824, opened 2022-10-04
Move keys out of configuration


ilchu:
As discussed on a call with @appetrosyan and @Erigara, we want to move towards configuration which would not keep track of any configuration parameters that are common to all the network, such as consensus parameters and so on, but instead would bootstrap those via some stand-in process and maybe IPC.

A similar logic should apply to keys: they are not configuration parameters per se, but important credentials that should rather be initialized once during the lifetime of a peer, thus also obsoleting the need for the keep-in-sync routine for keys done in `Configuration`'s `finish()`.

mversic:
related to #3240

0x009922:
@mversic, why was is closed?

Issue #2135, opened 2022-04-21
Representation of key pair in configuration files


mversic:
It has been [mentioned](https://github.com/hyperledger/iroha/pull/2130#discussion_r855210627) recently that in `client::Configuration`/`cli::Configuration`/`genesis::Configuration` we store `public_key` and `private_key` as separate variables as in:

```rs
pub struct Configuration {
    pub public_key: PublicKey,
    pub private_key: PrivateKey,
}
```
when they could be stored as a key pair as in:
```rs
pub struct Configuration {
    pub key_pair: KeyPair,
}
```

This is probably related to the deserialization either from config files or env variables. It should be investigated whether this switch can be made and will there be any tolerable consequences for the API of config files or env variables.

s8sato:
One possibility is to stop deserializing to keys from config files or environment variables altogether
and configure it only when sign-up (keys generation) or login

dmitrivenger:
#3240

mversic:
@0x009922 is this still actual?

0x009922:
On the user layer (reading from a file/env), it is still represented by separate `public_key` and `private_key` fields. However, when the user layer is parsed further, a we construct a `KeyPair` and then put it into the type-strong view of the config.

As for what format is good for the config file, I am not sure if nesting these two fields into a `key_pair` will make much difference:

```toml
public_key = "..."
private_key = { algorithm = "...", payload = "..." }
```

```toml
[key_pair]
public_key = "..."
private_key = { algorithm = "...", payload = "..." }
```

I'd vote for making things flat if possible (i.e. don't introduce `key_pair` nesting).

PR #3795, opened 2023-08-11
[RFC]: Configuration Overhaul

0x009922:
I think this place is the most convenient to gather the team's feedback.

### Checklist

- [x] Internal review
- [x] Library authors & users review
- [ ] Tech writers review
- [x] DevOps review

### DIFF
--- a/docs/source/rfc-configuration-overhaul.md
+++ b/docs/source/rfc-configuration-overhaul.md
@@ -511,11 +511,27 @@ struct Telemetry {
 }
 ```
 
-The telemetry functionality in Iroha relies heavily on the `name` field. Without it:
+While this flat structure is convenient to represent the user configuration of telemetry (i.e. the content of `telemetry` section in the configuration file), there are some possible concerns:
 
-- If other fields are set but `name` is omitted, the system could issue a warning indicating the ignored fields.
+- If `name` is omitted, but some of the other fields are set, the system could warn that telemetry will not start unless `name` is provided.
 - If the `name` is given, but neither `url` nor `file` are specified, the system should either issue a warning or terminate, signaling that the telemetry configuration is incomplete.
 - If both the `url` and `file` fields are filled (indicating a potential conflict), a warning about the conflict should be raised, followed by a system termination.
+
+These points might be taken into account during the *resolution* stage. A _resolved type_ for telemetry configuration might avoid these conflicts by design:
+
+```rust
+struct TelemetryResolved {
+    name: String,
+    target: TelemetryResolvedTarget,
+    min_retry_period: u64,
+    max_retry_delay_exponent: u8,
+}
+
+enum TelemetryResolvedTarget {
+    Url(Url),
+    File(std::path::PathBuf),
+}
+```
 
 #### Summary
 
--- a/docs/source/rfc-configuration-overhaul.md
+++ b/docs/source/rfc-configuration-overhaul.md
@@ -321,9 +321,9 @@ This proposal aims to standardise the naming conventions within Iroha's configur
 
 #### See Also
 
-- [todo](#proposal-1---use-toml) [[80ad14 Proposal 1 - Use TOML]]
-- [todo](#proposal-4---better-aliases-for-env) [[80ad8 Proposal 4 - Better Aliases for ENV]]
-- [todo](#proposal-5---define-deprecation-and-migration-policy) [[80ad13 Proposal 5 - Define Deprecation and Migration Policy]]
+- [Proposal 1 - Use TOML](#proposal-1---use-toml)
+- [Proposal 4 - Better Aliases for ENV](#proposal-4---better-aliases-for-env)
+- [Proposal 5 - Define Deprecation and Migration Policy](#proposal-5---define-deprecation-and-migration-policy)
 - [Representation of key pair in configuration files · Issue #2135 · hyperledger/iroha](https://github.com/hyperledger/iroha/issues/2135)
 
 ### Proposal 4 - Better Aliases for ENV
@@ -353,9 +353,9 @@ By transitioning to more user-friendly ENV variable names and reducing reliance 
 
 #### See Also
 
-- [[80ad10 Proposal 2 - Reference Before Implementation]](#proposal-2---reference-before-implementation)
-- [[80ad9 Proposal 3 - Consistent Naming of Configuration Parameters]](#proposal-3---consistent-naming-of-configuration-parameters)
-- [[80ad12 Proposal 7 - Trace Configuration Resolution]](#proposal-7---trace-configuration-resolution)
+- [Proposal 2 - Reference Before Implementation](#proposal-2---reference-before-implementation)
+- [Proposal 3 - Consistent Naming of Configuration Parameters](#proposal-3---consistent-naming-of-configuration-parameters)
+- [Proposal 7 - Trace Configuration Resolution](#proposal-7---trace-configuration-resolution)
 
 ### Proposal 5 - Define Deprecation and Migration Policy
 
@@ -465,7 +465,7 @@ While the trace provides detailed information, caution should be taken to avoid 
 
 #### Further Integration
 
-Incorporating these traces into configuration-related error messages can further streamline troubleshooting, enhancing user experience. (Refer to: [[80ad11 Proposal 6 - Exhaustive Error Messages]](#proposal-6---exhaustive-error-messages))
+Incorporating these traces into configuration-related error messages can further streamline troubleshooting, enhancing user experience. (Refer to: [Proposal 6 - Exhaustive Error Messages](#proposal-6---exhaustive-error-messages))
 
 #### Summary
 
--- a/docs/source/rfc-configuration-overhaul.md
+++ b/docs/source/rfc-configuration-overhaul.md
@@ -48,17 +48,14 @@ The inconsistency between the `IROHA2` and `IROHA` prefixes can be confusing. Of
 
 #### Inconsistent Naming of Configuration Parameters
 
-Current configuration demonstrates some inconsistencies in parameter naming, leading to potential user confusion. Specifically:
+Current configuration demonstrates some inconsistencies in parameter naming, leading to potential user confusion. To address this, some refinements are proposed:
 
-- In the `torii` section:
-  - We have terms like `p2p_addr`, `api_url`, and `telemetry_url`. While "addr" and "url" may convey similar intents, the interchangeability of these terms is non-intuitive.
-  - Using prefixes instead of suffixes, like transitioning from `p2p_addr` to `addr_p2p`, progresses from a general term to a specific one, simplifying interpretation.
-- In the `logger` section:
-  - The term `max_log_level` could be simplified to `log_level` or just `level` to make it more concise without losing meaning.
-- In the `wsv` section:
-  - The field `wasm_runtime_config` has a redundant `_config` suffix. Given the nature of these parameters as configurations, such suffixes are extraneous and can be dropped for clarity.
+- The `torii` section uses terms like `p2p_addr`, `api_url`, and `telemetry_url`. These are actually addresses in the form of `localhost:8080` and not full URLs, which would include components like protocol and path. To standardize, these should all be changed to use `addr` (e.g., `addr_p2p`, `addr_api`, `addr_telemetry`).
+- Consider moving addresses to their respective domain scopes. For example, instead of `torii.telemetry_url`, we could have `telemetry.addr`.
+- In the `logger` section, the term `max_log_level` could be simplified to `log_level` or just `level` to make it more concise without losing meaning. Although users are more likely to set this parameter using an environment variable.
+- In the `wsv` section, the field `wasm_runtime_config` has a redundant `_config` suffix. Given these are all configuration parameters, such suffixes can be eliminated for clarity.
 
-Implementing a consistent naming convention will further streamline the configuration process for users.
+Implementing a consistent naming convention like this will simplify the user's configuration process and reduce ambiguity.
 
 #### Redundant SCREAMING_CASE in JSON Configuration
 
--- a/docs/source/rfc-configuration-overhaul.md
+++ b/docs/source/rfc-configuration-overhaul.md
@@ -79,7 +79,7 @@ address this, some refinements are proposed:
 - In the `logger` section, the term `max_log_level` could be simplified to `log_level` or just `level` to make it more
   concise without losing meaning. Although users are more likely to set this parameter using an environment variable.
 - In the `wsv` section, the field `wasm_runtime_config` has a redundant `_config` suffix. Given these are all
-  configuration parameters, such suffixes can be eliminated for clarity.
+  _configuration_ parameters, such suffixes can be eliminated for clarity.
 
 Implementing a consistent naming convention like this will simplify the user's configuration process and reduce
 ambiguity.
--- a/docs/source/rfc-configuration-overhaul.md
+++ b/docs/source/rfc-configuration-overhaul.md
@@ -112,7 +112,7 @@ Iroha and might not be familiar with Rust. Here are some examples:
 
 The use of rustisms like that overcomplicates the configuration reference and affects the user experience. Further, it
 doesn't actually provide any extra information to a user familiar with Rust, as environment variables and JSON values
-cannot be `None` due to being strictly typed and null-able regardless of whether the type specifies `Option`.
+cannot be `None` due to being stringly typed and null-able regardless of whether the type specifies `Option`.
 
 ### Unhelpful Error Messages
 
--- a/docs/source/rfc-configuration-overhaul.md
+++ b/docs/source/rfc-configuration-overhaul.md
@@ -1,0 +1,712 @@
+# Iroha 2 Configuration Overhaul
+
+## Introduction
+
+Configuration is often the first point of interaction for users with Hyperledger Iroha 2. It's crucial to make this experience as smooth as possible. However, the current system, marred by issues like redundant fields, ambiguous naming, and unclear error messages, necessitates a thorough review. This RFC sets out to propose vital improvements for a more intuitive and efficient configuration experience.
+
+## Background
+
+### Redundant Fields
+
+There are configuration fields that shouldn't be initialized by the user, meaning that including them in the configuration reference is redundant. For example, here's the [excerpt](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#sumeragi-default-null-values) about configuration options for Sumeragi:
+
+> **Sumeragi: default `null` values**
+>
+> A special note about sumeragi fields with `null` as default: only the `trusted_peers` field out of the three can be initialized via a provided file or an environment variable.
+>
+> The other two fields, namely `key_pair` and `peer_id`, go through a process of finalization where their values are derived from the corresponding ones in the uppermost Iroha config (using its `public_key` and `private_key` fields) or the Torii config (via its `p2p_addr`). This ensures that these linked fields stay in sync, and prevents the programmer error when different values are provided to these field pairs. Providing either `sumeragi.key_pair` or `sumeragi.peer_id` by hand will result in an error, as it should never be done directly.
+
+As we can see, if the user tries to initialize `key_pair` and `peer_id` via the config file or env variables, they'll get an error. We are creating a terrible user experience by presenting the user with an option to configure something they should not be configuring.
+
+We should avoid exposing fields like this to the user.
+
+### Naming Issues
+
+The configuration parameters naming presents a few challenges.
+
+#### Overcomplicated ENV Variables
+
+ENV variables for all configuration parameters are named according to the internal module names. These technical designations might resonate with developers but can be confusing or alienating for users not deeply immersed in the project's inner mechanisms. Many FOSS projects gravitate towards generic, descriptive names to ensure broader comprehension and enhanced usability.
+
+#### Inconsistent Naming of ENV Variables
+
+The output from the `iroha --help` command displays a mix of naming conventions:
+
+```
+Iroha 2 is configured via environment variables:
+    IROHA2_CONFIG_PATH is the location of your `config.json` or `config.json5`
+    IROHA2_GENESIS_PATH is the location of your `genesis.json` or `genesis.json5`
+...
+    IROHA_TORII: Torii (gateway) endpoint configuration
+    IROHA_SUMERAGI: Sumeragi (emperor) consensus configuration
+    IROHA_KURA: Kura (storage). Configuration of block storage
+    IROHA_BLOCK_SYNC: Block synchronisation configuration
+...
+```
+
+The inconsistency between the `IROHA2` and `IROHA` prefixes can be confusing.
+
+#### Inconsistent Naming of Configuration Parameters
+
+Current configuration demonstrates some inconsistencies in parameter naming, leading to potential user confusion. Specifically:
+
+- In the `torii` section:
+  - We have terms like `p2p_addr`, `api_url`, and `telemetry_url`. While "addr" and "url" may convey similar intents, the interchangeability of these terms is non-intuitive.
+  - Using prefixes instead of suffixes, like transitioning from `p2p_addr` to `addr_p2p`, progresses from a general term to a specific one, simplifying interpretation.
+- In the `logger` section:
+  - The term `max_log_level` could be simplified to `log_level` or just `level` to make it more concise without losing meaning.
+- In the `wsv` section:
+  - The field `wasm_runtime_config` has a redundant `_config` suffix. Given the nature of these parameters as configurations, such suffixes are extraneous and can be dropped for clarity.
+
+Implementing a consistent naming convention will further streamline the configuration process for users.
+
+#### Redundant SCREAMING_CASE in JSON Configuration
+
+The adoption of SCREAMING_CASE for parameters within the JSON configuration appears out of place. Such uppercase field names might be customary for environment variables, but in a JSON setting, they don't offer any distinctive benefit. Here's a snippet from the current configuration showcasing this:
+
+```json
+{
+  "PUBLIC_KEY": null,
+  "PRIVATE_KEY": null,
+  "DISABLE_PANIC_TERMINAL_COLORS": false
+}
+```
+
+---
+
+In sum, these issues underscore the need for clearer, more intuitive naming conventions in Iroha 2's configuration.
+
+### Rustisms in the Configuration Reference
+
+Configuration reference contains Rust-specific expressions that wouldn't make sense to the end-user who is configuring Iroha and might not be familiar with Rust. Here are some examples:
+
+- [Using `Option<..>`](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#option)
+- [Using `std::path::PathBuf`](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#loggerlog_file_path)
+
+The use of rustisms like that overcomplicates the configuration reference and affects the user experience.
+
+### Unhelpful Error Messages
+
+Some of the error messages related to Iroha configuration are inconsistent and misleading. Here's a quick rundown of various scenarios, associated errors and possible confusions.
+
+#### 1. No configuration file
+
+Running Iroha without a configuration file results in the following log and error:
+
+```
+IROHA_TORII: environment variable not found
+IROHA_SUMERAGI: environment variable not found
+IROHA_KURA: environment variable not found
+IROHA_BLOCK_SYNC: environment variable not found
+IROHA_PUBLIC_KEY: environment variable not found
+IROHA_PRIVATE_KEY: environment variable not found
+IROHA_GENESIS: environment variable not found
+Configuration file not found. Using environment variables as fallback.
+Error:
+   0: Please add `PUBLIC_KEY and PRIVATE_KEY` to the configuration.
+```
+
+Several things are wrong with this error message:
+
+- The information about using environment variables as fallback comes after the messages about them not being found.
+- Hinting to add `PUBLIC_KEY` and `PRIVATE_KEY` to the configuration does not explain what happened (the absence of the config file and no environment variables to fallback to). This hint is also not helpful as there is no information on whether public and private keys should be added to the config file or to ENV variables, and no information about the ENV variables these fields are mapped to.
+
+#### 2. Path to non-existent config file
+
+Providing a path to a configuration file that doesn't exist results in the exact same error as if there was no configuration file specified at all.
+
+While it makes sense to silently fallback to ENV variables if the user doesn't provide a path to a config file, in case when the user **does** specify the path to a config file, it would be better for the program to fail with an appropriate error message.
+
+#### 3. Empty config file
+
+Running Iroha with an empty `config.json` results in the error that does not provide any information about the fallback to the environment variables even though it does happen. This behaviour is inconsistent.
+
+```
+Error:
+   0: Please add `PUBLIC_KEY and PRIVATE_KEY` to the configuration.
+```
+
+#### 4. Config file with only `PUBLIC_KEY` and `PRIVATE_KEY` specificed
+
+Running Iroha with a config file that only contains `PUBLIC_KEY` and `PRIVATE_KEY` results in the following error:
+
+```
+Error:
+   0: Please add `p2p_addr` to the configuration.
+```
+
+This error message is misleading and inconsistent:
+
+- If the `p2p_addr` field is added to the config file, the error stays the same.
+- The `p2p_addr` field is not a root-level field, it's actually a part of the `TORII` configuration and should be `TORII.P2P_ADDR`.
+- Error messages asking to add public and private keys use uppercase for configuration fields, while in this error the field name is written in lowercase.
+
+#### 5. Config file with only `PUBLIC_KEY`, `PRIVATE_KEY`, and `TORII.P2P_ADDR` specified
+
+Running Iroha with a config file that only contains `PUBLIC_KEY`, `PRIVATE_KEY`, and `TORII.P2P_ADDR` results in the following error:
+
+```
+Error:
+   0: Please add `api_url` to the configuration.
+```
+
+Comparing this scenario to the previous one, we can notice that the `TORII.API_URL` was missing before as well but the previous error message didn't mention it.
+
+There are also two inconsistencies:
+
+- Similar to `p2p_addr` in the previous scenario, this field name is lowercase, while for public and private keys it was uppercase.
+- While `TORII.P2P_ADDR` and `TORII.API_URL` share the same root-level (`TORII`), their names are different: "addr" derived from "address" and "URL". Logically these are both addresses or URLs, and it would make sense for the names to be aligned.
+
+#### 6. Config file with invalid `PUBLIC_KEY`
+
+Providing an invalid `PUBLIC_KEY` in the `config.json` results in the same error as when there is no config file at all, or an empty config file:
+
+```
+Error:
+   0: Please add `PUBLIC_KEY and PRIVATE_KEY` to the configuration.
+```
+
+This is not a helpful error message as it does not mention what exactly is wrong, why and where the parsing failed.
+
+#### 7. Invalid `IROHA2_PUBLIC_KEY` environment variable
+
+Providing an invalid `IROHA2_PUBLIC_KEY` environment variable results in the following error:
+
+```
+Error:
+   0: Failed to build configuration from env
+   1: Failed to deserialize the field `IROHA_PUBLIC_KEY`: JSON5: Key could not be parsed. Key could not be parsed. Invalid character 's' at position 4
+   2: JSON5: Key could not be parsed. Key could not be parsed. Invalid character 's' at position 4
+```
+
+While this message is more helful than the ones we discussed above, there are still issues:
+
+- The message contains repetition.
+- Without an input snippet, the `Invalid character 's' at position 4` part of the message is not as helpful as it could be.
+
+#### 8. Config file with extra fields
+
+Providing a configuration file with extra fields that are not supposed to be there does not produce any errors at all.
+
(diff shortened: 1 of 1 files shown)


docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+>
+> Network configuration
+>
+> Has type `Option<network::ConfigurationProxy>`. Can be configured via environment variable `IROHA_NETWORK`
+>
+> ```json
+> {
+>   "ACTOR_CHANNEL_CAPACITY": 100
+> }
+> ```
```
0x009922, line 227: Looks like a bug of GitHub's diff view.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+As we can see, `network` is not actually a configuration field in itself. Instead, there are various `network.*` fields that can be configured. This bring a question of what will happen when we set the `IROHA_NETWORK` environment variable as mentioned in the excerpt above? Will it override the nested `IROHA_NETWORK_CAPACITY` fields? Is there a use case for providing an ability to set `IROHA_NETWORK` all-in-one through an environment variable?
+
+### Chaotic Code Organisation
+
+Internally, not all of the configuration-related logic is contained within a single `iroha_config` crate. Instead, some of the configuration resolution logic is located in other crates, such as `iroha_cli`. This makes the configuration-related code error-prone and harder to maintain.
```
6r1d, line 233: Wow, thanks for pointing that out.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+- [Proposal 3 - Consistent Naming of Configuration Parameters](#proposal-3---consistent-naming-of-configuration-parameters)
+- [Proposal 4 - Better Aliases for ENV](#proposal-4---better-aliases-for-env)
+- [Proposal 5 - Define Deprecation and Migration Policy](#proposal-5---define-deprecation-and-migration-policy)
+
+### Proposal 2 - Reference Before Implementation
+
+#### Objective
+
+Establish a clear and comprehensive configuration reference prior to actual code implementation. This approach ensures that any new features or changes are well-documented, understandable, and in alignment with the project's goals.
```
6r1d, line 263: This is going to be great for supporting Iroha! Thank you!

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+- A detailed configuration reference document outlining the structure, default values, data types, and descriptions for each configuration parameter.
+
+**Potential Risks**:
+
+- Misalignment with user needs if not adequately consulted.
+- Overcomplication or oversimplification of configuration design.
+
+**Feedback Loop**:
+
+- Regularly consult with both developers and potential users to ensure design alignment with their expectations and use cases.
```
6r1d, line 574: Let's start with that on some of our bi-weekly meetings, and from that point on, we can mention the updates in our community.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+The inconsistency between the `IROHA2` and `IROHA` prefixes can be confusing.
+
+#### Inconsistent Naming of Configuration Parameters
+
+Current configuration demonstrates some inconsistencies in parameter naming, leading to potential user confusion. Specifically:
+
+- In the `torii` section:
+  - We have terms like `p2p_addr`, `api_url`, and `telemetry_url`. While "addr" and "url" may convey similar intents, the interchangeability of these terms is non-intuitive.
+  - Using prefixes instead of suffixes, like transitioning from `p2p_addr` to `addr_p2p`, progresses from a general term to a specific one, simplifying interpretation.
```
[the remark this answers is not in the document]
QuentinI: Furthermore, I'd argue that `p2p` is the more general term here if interpreted in context of iroha. I feel like parameters are better off being grouped by subsystem rather than by name of knob they tweak.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+### Proposal 1 - Use TOML
+
+#### Objective
+
+Transition to TOML as Iroha's standard configuration format to provide a cleaner, more intuitive setup and to align with Rust community standards.
+
+#### Rationale
+
+- **Human-Friendly Syntax:** TOML's format is easier to read and understand compared to JSON, especially given its support for comments.
```
QuentinI, line 245: If we want to adopt TOML we should probably cut down on nesting. In my opinion TOML doesn't handle it well

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+- **User Accessibility:** Generifying names improves understanding and avoids confusion. Using internal module names can be restrictive and lead to misconfiguration.
+- **Consistency:** A uniform naming convention for environment variables enhances predictability, aiding users in setting up configurations without constantly referring to documentation.
+- **Transition Support:** New aliases will coexist with existing ones, ensuring no disruptions and aiding users during the transitional phase.
+
+#### Proposed Changes
+
+1. **Rename Based on Functionality:** Shift from code-based naming to function-based naming. For example:
+   - `TORII_API_URL` becomes `API_ENDPOINT_URL`.
+   - Convert `KURA_INIT_MODE` to `BLOCK_REVALIDATION_MODE`.
```
QuentinI, line 345: I feel like we should be conservative with this sort of change, lest it results in developers having to keep in mind two parallel naming schemes.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+#### Objective
+
+Implement a mechanism that displays the sequence in which configuration is determined. This allows users to grasp the sources from which configuration values are derived, whether they're being overridden, and which resort to default values.
+
+#### Rationale
+
+Transparency in configuration resolution is vital for troubleshooting and ensuring the correct setup. By providing clear feedback on how parameters are set, users can identify potential misconfigurations or conflicts quickly.
+
+#### Proposed Tracing Output
```
QuentinI, line 430: Should probably include the resultant config values after all resolution stages.

Erigara (commented):
Don't have any objections.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+
+#### Inconsistent Naming of Configuration Parameters
+
+Current configuration demonstrates some inconsistencies in parameter naming, leading to potential user confusion. Specifically:
+
+- In the `torii` section:
+  - We have terms like `p2p_addr`, `api_url`, and `telemetry_url`. While "addr" and "url" may convey similar intents, the interchangeability of these terms is non-intuitive.
+  - Using prefixes instead of suffixes, like transitioning from `p2p_addr` to `addr_p2p`, progresses from a general term to a specific one, simplifying interpretation.
+- In the `logger` section:
+  - The term `max_log_level` could be simplified to `log_level` or just `level` to make it more concise without losing meaning.
```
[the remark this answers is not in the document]
0x009922: It is under `logger` section, thus its full path is `logger.level`. `logger.log` looks awkward. However, in terms of ENV vars `LOG` is fine.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+- In the `wsv` section:
+  - The field `wasm_runtime_config` has a redundant `_config` suffix. Given the nature of these parameters as configurations, such suffixes are extraneous and can be dropped for clarity.
+
+Implementing a consistent naming convention will further streamline the configuration process for users.
+
+#### Redundant SCREAMING_CASE in JSON Configuration
+
+The adoption of SCREAMING_CASE for parameters within the JSON configuration appears out of place. Such uppercase field names might be customary for environment variables, but in a JSON setting, they don't offer any distinctive benefit. Here's a snippet from the current configuration showcasing this:
```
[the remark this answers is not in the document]
0x009922: I want to clarify that this section is about SCREAMING_CASE _exclusively_ in the JSON configuration.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+Several things are wrong with this error message:
+
+- The information about using environment variables as fallback comes after the messages about them not being found.
+- Hinting to add `PUBLIC_KEY` and `PRIVATE_KEY` to the configuration does not explain what happened (the absence of the config file and no environment variables to fallback to). This hint is also not helpful as there is no information on whether public and private keys should be added to the config file or to ENV variables, and no information about the ENV variables these fields are mapped to.
+
+#### 2. Path to non-existent config file
+
+Providing a path to a configuration file that doesn't exist results in the exact same error as if there was no configuration file specified at all.
+
+While it makes sense to silently fallback to ENV variables if the user doesn't provide a path to a config file, in case when the user **does** specify the path to a config file, it would be better for the program to fail with an appropriate error message.
```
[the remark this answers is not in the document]
0x009922: I checked this behaviour after May. Maybe I will re-check it again.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+```
+
+While this message is more helful than the ones we discussed above, there are still issues:
+
+- The message contains repetition.
+- Without an input snippet, the `Invalid character 's' at position 4` part of the message is not as helpful as it could be.
+
+#### 8. Config file with extra fields
+
+Providing a configuration file with extra fields that are not supposed to be there does not produce any errors at all.
```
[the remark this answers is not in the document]
0x009922: Sure. This section is focused on the background, the current state of the configuration. The actual proposals come later.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+> }
+> ```
+
+As we can see, `network` is not actually a configuration field in itself. Instead, there are various `network.*` fields that can be configured. This bring a question of what will happen when we set the `IROHA_NETWORK` environment variable as mentioned in the excerpt above? Will it override the nested `IROHA_NETWORK_CAPACITY` fields? Is there a use case for providing an ability to set `IROHA_NETWORK` all-in-one through an environment variable?
+
+### Chaotic Code Organisation
+
+Internally, not all of the configuration-related logic is contained within a single `iroha_config` crate. Instead, some of the configuration resolution logic is located in other crates, such as `iroha_cli`. This makes the configuration-related code error-prone and harder to maintain.
+
+## Proposals
```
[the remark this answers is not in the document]
0x009922: You mean to emphasise that the proposals shouldn't be necessarily implemented all-or-nothing, but instead each point might be considered in isolation?

docs/source/rfc-configuration-overhaul.md
0x009922: Currently `wsv` has the deepest nesting:

```toml
[wsv]
assets_metadata_limits = {
  max_len = 1000,
  max_entry_byte_size = 32
}

# or

[wsv]
wasm_runtime_config.max_memory = 1_000_000

# or

[wsv.ident_length_limits]
min = 10
max = 32
```

I personally don't see any problem with that. The problems that come to my mind are:

- Readability: if nesting goes much deeper
- Familiarity: in comparison to JSON/YAML, TOML's syntax for nesting might be less intuitive

Could you expand your opinion?

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
0x009922: No objections here. I think what you have said should be already clear in the RFC.

docs/source/rfc-configuration-overhaul.md
0x009922: Addressed, changed the sample output.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+```
+
+The inconsistency between the `IROHA2` and `IROHA` prefixes can be confusing.
+
+#### Inconsistent Naming of Configuration Parameters
+
+Current configuration demonstrates some inconsistencies in parameter naming, leading to potential user confusion. Specifically:
+
+- In the `torii` section:
+  - We have terms like `p2p_addr`, `api_url`, and `telemetry_url`. While "addr" and "url" may convey similar intents, the interchangeability of these terms is non-intuitive.
```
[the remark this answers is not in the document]
0x009922: I am confused with your explanation.

AFAIC, in reality, they are all addresses, i.e. `<host>:<port>`. URL usually contains more details such as schema, path, query etc, which is not relevant here.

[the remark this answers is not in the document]
appetrosyan: In principle.  In practice, the user doesn't care to know or be educated, but does care if they think that they're missing something.

docs/source/rfc-configuration-overhaul.md
appetrosyan: There are no formats that handle nesting well, with the possible exception of Nix. IMO the nesting we have is adequate for the format not to be confusing for users in TOML (or YAML).

docs/source/rfc-configuration-overhaul.md
appetrosyan: Renaming would only have the developers remember one old scheme and a newer one.  With deprecation warnings it should be relatively easy.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
appetrosyan: I think we can get rid of the internal codename as well, so something along the lines of 
```toml
telemetry.addr = "localhost:8081"
```

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
appetrosyan: Good point.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
appetrosyan: After a bit of thinking, it's better to leave the env vars alone for the time being.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
appetrosyan: Yes.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
0x009922: Addressed in the document.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
0x009922: Addressed.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,721 @@
...
+In sum, these issues underscore the need for clearer, more intuitive naming conventions in Iroha 2's configuration.
+
+### Rustisms in the Configuration Reference
+
+Configuration reference contains Rust-specific expressions that wouldn't make sense to the end-user who is configuring Iroha and might not be familiar with Rust. Here are some examples:
+
+- [Using `Option<..>`](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#option)
+- [Using `std::path::PathBuf`](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#loggerlog_file_path)
+
+The use of rustisms like that overcomplicates the configuration reference and affects the user experience.
```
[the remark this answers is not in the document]
0x009922: Addressed.

0x009922:
Addressed all of the comments as of now. The document looks finalised.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+
+Configuration reference contains Rust-specific expressions that wouldn't make sense to the end-user who is configuring
+Iroha and might not be familiar with Rust. Here are some examples:
+
+- [Using `Option<..>`](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#option)
+- [Using `std::path::PathBuf`](https://github.com/hyperledger/iroha/blob/35ba182e1d1b6b594712cf63bf448a2edefcf2cd/docs/source/references/config.md#loggerlog_file_path)
+
+The use of rustisms like that overcomplicates the configuration reference and affects the user experience. Further, it
+doesn't actually provide any extra information to a user familiar with Rust, as environment variables and JSON values
+cannot be `None` due to being strictly typed and null-able regardless of whether the type specifies `Option`.
```
ilchu, line 115:
```suggestion
cannot be `None` due to being stringly typed and null-able regardless of whether the type specifies `Option`.
```
Because this line got me confused TBH, and I believe that's what @appetrosyan said himself.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+IROHA_PRIVATE_KEY: environment variable not found
+IROHA_GENESIS: environment variable not found
+Configuration file not found. Using environment variables as fallback.
+Error:
+   0: Please add `PUBLIC_KEY and PRIVATE_KEY` to the configuration.
+```
+
+Several things are wrong with this error message:
+
+- The information about using environment variables as fallback comes after the messages about them not being found.
```
Mingela, line 141: moreover, there shouldn't be any explicit mention of an environment variable in an error, a config file is recommended approach, an env variable is just an alternative, hence generic 'value not set' should be mentioned

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+  config file and no environment variables to fallback to). This hint is also not helpful as there is no information on
+  whether public and private keys should be added to the config file or to ENV variables, and no information about the
+  ENV variables these fields are mapped to.
+
+#### 2. Path to non-existent config file
+
+Providing a path to a configuration file that doesn't exist results in the exact same error as if there was no
+configuration file specified at all.
+
+While it makes sense to silently fallback to ENV variables if the user doesn't provide a path to a config file, in case
```
Mingela, line 152: I guess ENV has the most priority for a parameter so it's rather not a fallback, though I get the sense anyway

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+or an empty config file:
+
+```
+Error:
+   0: Please add `PUBLIC_KEY and PRIVATE_KEY` to the configuration.
+```
+
+This is not a helpful error message as it does not mention what exactly is wrong, why and where the parsing failed.
+
+#### 7. Invalid `IROHA2_PUBLIC_KEY` environment variable
```
Mingela, line 216: good research of the all potential cases 👍 wondering if there is other behavior for other combinations of the parameters etc 😱

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+> ```json
+> {
+>   "ACTOR_CHANNEL_CAPACITY": 100
+> }
+> ```
+
+As we can see, `network` is not actually a configuration field in itself. Instead, there are various `network.*` fields
+that can be configured. This bring a question of what will happen when we set the `IROHA_NETWORK` environment variable
+as mentioned in the excerpt above? Will it override the nested `IROHA_NETWORK_CAPACITY` fields? Is there a use case for
+providing an ability to set `IROHA_NETWORK` all-in-one through an environment variable?
```
Mingela, line 283: `IROHA_NETWORK` is just a prefix/level, not the parameter, so should be treated as an excessive value

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+Internally, not all of the configuration-related logic is contained within a single `iroha_config` crate. Instead, some
+of the configuration resolution logic is located in other crates, such as `iroha_cli`. This makes the
+configuration-related code error-prone and harder to maintain.
+
+## Proposals
+
+The proposals outlined in this section are designed to collectively enhance the configuration system within Iroha.
+However, they are not mutually required. Each proposal can be considered and implemented independently.
+
+### Proposal 1 - Use TOML
```
Mingela, line 296: how about `yaml` alternative proposal?
common for deployments

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+
+##### Redundant Fields
+
+In the background about [redundant fields](#redundant-fields), it's mentioned that these fields were made available to
+end users primarily because the user config wasn't separated from the resolved config. However, in a two-stage
+configuration resolution, these redundant fields can be managed more efficiently.
+
+##### Root-Level `public_key` and `private_key`
+
+When users provide their configuration input, it should be possible for them to specify only the private key. From this,
```
Mingela, line 623: if we implement this logic we might just drop providing `public_key` from the configuration whatsoever

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+```
+
+While this flat structure is convenient to represent the user configuration of telemetry (i.e. the content of
+`telemetry` section in the configuration file), there are some possible concerns:
+
+- If `name` is omitted, but some of the other fields are set, the system could warn that telemetry will not start unless
+  `name` is provided.
+- If the `name` is given, but neither `url` nor `file` are specified, the system should either issue a warning or
+  terminate, signaling that the telemetry configuration is incomplete.
+- If both the `url` and `file` fields are filled (indicating a potential conflict), a warning about the conflict should
```
Mingela, line 653: what kind of conflict might take place?

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+  be raised, followed by a system termination.
+
+These points might be taken into account during the _resolution_ stage. A _resolved type_ for telemetry configuration
+might avoid these conflicts by design:
```
Mingela, line 657: agree, such things rather should be prevented by design

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+
+```rust
+struct TelemetryResolved {
+    name: String,
+    target: TelemetryResolvedTarget,
```
Mingela, line 662: though why can't we have multiple in parallel?

docs/source/rfc-configuration-overhaul.md
0x009922: _Currently_, due to implementation details, one can pass `IROHA_NETWORK` env var and it will work:

```bash
# set a particular field
IROHA_NETWORK_CAPACITY=100 ./iroha

# same
IROHA_NETWORK="{\"ACTOR_CHANNEL_CAPACITY\":100}" ./iroha
```

> **Note:** it is relevant not only to `network.*`, but to all other nestings, e.g. `torii.*`, `sumeragi.*` etc.

The problem now is that `network.*` treated not as a namespace of parameters, but also a parameter itself. I don't think we need such a feature. Moreover, although it is partially highlighted in the configuration reference, it is not well documented and produces questions I have highlighted in the RFC.

Will rephrase this part.

docs/source/rfc-configuration-overhaul.md
0x009922: Indeed. Well, let's say, in our case our Telemetry supports only single target. I should indicate it in this section.

docs/source/rfc-configuration-overhaul.md
0x009922: I imply that `url` and `file` are mutually exclusive. I.e., Telemetry might have only one emit target.

docs/source/rfc-configuration-overhaul.md
0x009922: Keys configuration has a few issues:

- #2135
- #2824
- #3240

I am not sure what of these should be included into the RFC. Thus, I chose to stick with the status quo.

Let me put a note somewhere in the RFC about it and mention these issues.

docs/source/rfc-configuration-overhaul.md
0x009922: Well, choice between TOML and YAML is more about a personal preference, I believe. However, we prefer TOML primarily because:

- Simple and explicit syntax, more readable
- Strong types, less unexpected results
- Indentation doesn't matter
- Better error messages
- It is a standard for Rust ecosystem

As for making YAML/JSON/whatever an additional configuration format, it might badly affect the overall quality of the configuration system due to it being more generalised, thus e.g. less able to provide some details in error messages.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,893 @@
...
+
+- The message contains repetition.
+- Without an input snippet, the `Invalid character 's' at position 4` part of the message is not as helpful as it could
+  be.
+
+#### 8. Config file with extra fields
+
+Providing a configuration file with extra fields that are not supposed to be there does not produce any errors at all.
+
+This might lead to a bad user experience when user expects some options to apply, but doesn't have any idea that those
```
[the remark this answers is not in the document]
0x009922: I am confused, because the purpose of default values is exactly to be applied when nothing is specified at all.

As for accidentally having an outdated ENV var, this RFC includes defining a deprecation policy. According to it, when an ENV var renaming takes place, the user will have a window between versions when they will get a warning/error for the deprecated variable before it is deleted entirely.

docs/source/rfc-configuration-overhaul.md
0x009922: I see, let me try to change the phrasing.

docs/source/rfc-configuration-overhaul.md
mversic: IMO `TOML` is the simplest of the three and is best suited for Iroha purposes

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
Mingela: In that case we should at least specify what fields have default values, I guess such a set should be as little as possible (i.e. log level etc only)

docs/source/rfc-configuration-overhaul.md
Mingela: That's kinda exotic, we'd rather exclude this kind of possibility imo

docs/source/rfc-configuration-overhaul.md
Mingela: My main motivation is that configuration is rather related to deployment, whereas `TOML` is common for a project (development) configuration in Rust, though I don't deny there deployment tools relying on `TOML` may exist. DevOps input may be really valuable here.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
0x009922: Hmm, from internal discussions I remember the opposite request: we should have as many defaults as possible.

Anyway, on the configuration reference writing stage we will decide in details which fields we should have, what default values should be etc. At that stage we will also collect feedback from external users.

docs/source/rfc-configuration-overhaul.md
0x009922: Rephrased.

docs/source/rfc-configuration-overhaul.md
0x009922: Rephrased this bit, hope it is clearer now.

docs/source/rfc-configuration-overhaul.md
0x009922: Included this note as a footnote.

docs/source/rfc-configuration-overhaul.md
0x009922: Added a note that here `url` and `file` are implied to be mutually exclusive for the purposes of this example.

Mingela (approved):
lgtm

docs/source/rfc-configuration-overhaul.md
6r1d: While my current position is unrelated to DevOps, I'm confident I have enough experience to discuss DevOps issues.

For me, `TOML` is more readable, and the advantage of `JSON`/`YAML`/`TOML` is that we learn them quickly as human beings.

If we assume that people are configuring Iroha, they either know `TOML` or will know it as soon as they need to use it.
If we assume we're using GPT to configure it (for some reason), it knows the format and will most likely get stuck on the variable names. Real people may make the same mistake unless they have the config reference doc for the latest Iroha version with a lot of examples.

> whereas TOML is common for a project (development) configuration in Rust, though I don't deny there deployment tools relying on TOML may exist

TOML is intended to be a configuration file format, according to Tom Preston-Werner at https://toml.io/en/: 

> A config file format for humans.

> TOML aims to be a minimal configuration file format that's easy to read due to obvious semantics. TOML is designed to map unambiguously to a hash table. TOML should be easy to parse into data structures in a wide variety of languages.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
6r1d: In my opinion, both of you have a point.
I have a question regarding the deployment workflow (which may be covered by a part of the RFC and I'm missing something).
Let's assume a person who never used Iroha 2 performs the deployment.
How do they see the configuration parameters that were applied as defaults?
Maybe a verbose-running mode would help debug the deployment configurations at a later point.

docs/source/rfc-configuration-overhaul.md
6r1d: I agree with @0x009922 here. If we don't see which field has an issue, how do we debug an Iroha deployment?

docs/source/rfc-configuration-overhaul.md
6r1d: I think that a regular deployment with multiple SDKs doing a request will be helpful.
If (for example) Tuesday's stand-up starts with the automatic news about the `dev`, `stable` or `lts` issues, it'll help both in responding to the community (we know the actual state) and quickly resolving the potential problems.
(I am not sure triggering it only on each release is valid, because Docker may change and influence the behavior of Iroha deployment)

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
Mingela: That's why we should just provide an example ('default') exhaustive configuration explicitly, but don't allow parameters not being provided with values

Cre-eD (approved):
LGTM

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
mversic: > That's why we should just provide an example ('default') exhaustive configuration explicitly, but don't allow parameters not being provided with values

I respectfully strongly disagree. Because this makes any change in the configuration a breaking change for every customer instead of only for those that actually use the parameter. We want to minimize the possibility of breaking changes downstream and we want to commit to as little of API as possible

Let me give an example of how a configuration file should look like (I take inspiration from `mysql.conf`):
```toml
[sumeragi]
private_key = "pq239874294u42139384u243"
# block_time_ms: 2000
```

1. All possible parameters should be present in the configuration file
2. Parameters that have a default value **should be commented out** (but still present in the file)
3. If a parameter has a default value the default value **must be visible** in the config file

This way a user is aware of all possible parameter and what their default values are, even if they will only use a small subset of parameters

>  How do they see the configuration parameters that were applied as defaults?

@6r1d  configuration should be printed in the log or we should at least provide an option to do so. We can print additional debug information in the log such as the information was the parameter given a default value or was set by the user, was it set through a file or an environment variable, etc.

docs/source/rfc-configuration-overhaul.md
mversic: IMO `yaml` is unnecessarily complicated and it's a mess to deal with spaces. On the other hand, `JSON` would be fine except that it doesn't support comments, and we really want to be able to comment out parameters in the config file. `TOML` fits our needs perfectly

if you look at `mysql.conf` it's configuration format is quite similar to `TOML`.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
Mingela: @mversic the suggestion is definitely valid, though we still have to determine what parameters should not be allowed to have the default value, e.g. things like metadata size limit could be risky if a user starts a network without proper attention to the consequences of not configuring that specifically to their use case or constraints.
On the other hand, if someone migrates to a newer version of Iroha, updating the corresponding configuration to match that is not a decision making factor overhead (esp. compared to a blockstore migration).

To conclude, I'd not insist on my view, either choice is fine to me. If we allow the defaults we just should be much more careful about this for every parameter now and in the future.

docs/source/rfc-configuration-overhaul.md
AlexStroke, line 1: I'm concerned about placing RFCs in iroha/docs. While RFCs are vital for discussions, they aren't finalized documentation. Housing them with main docs might mislead users. Consider a separate place for RFCs to avoid confusion.

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,912 @@
...
+- In the `logger` section, the term `max_log_level` could be simplified to `log_level` or just `level` to make it more
+  concise without losing meaning. Although users are more likely to set this parameter using an environment variable.
+- In the `wsv` section, the field `wasm_runtime_config` has a redundant `_config` suffix. Given these are all
+  _configuration_ parameters, such suffixes can be eliminated for clarity.
+
+Implementing a consistent naming convention like this will simplify the user's configuration process and reduce
+ambiguity.
+
+#### Redundant SCREAMING_CASE in JSON Configuration
+
```
adaagava, line 88: Maybe we need rule of using the SCREAMING_CASE ?
For example, use it just for parameters that doesn't have the preset value and must be filled in by user?

AlexStroke (changes requested):

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
0x009922: @Mingela, do I get you correctly: your point is that there are more parameters which should be provided by the user than we have now?

In other words, configuration like metadata size limits should be defined by the user anyway.

docs/source/rfc-configuration-overhaul.md
0x009922: Totally agree. This is WIP, we will find a better place for it.

docs/source/rfc-configuration-overhaul.md
0x009922: This is a creative suggestion, but I am not convinced that it would simplify the configuration process more than it would complicate it. The ideal configuration format should be both consistent and simple, and stick to a single, straightforward naming scheme. Instead of relying on syntax to highlight important fields, it's better to clearly identify those in the documentation. After all, this approach is somewhat unconventional and diverges from the norms you might see in many other software projects.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
mversic: we can always decide post factum on the set of parameters that will be required vs the ones that will have a default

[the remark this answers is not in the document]
Mingela: > do I get you correctly: your point is that there are more parameters which should be provided by the user than we have now?
> 
> In other words, configuration like metadata size limits should be defined by the user anyway.

@0x009922, I'm afraid I'm not entirely aware of all the parameters which have defaults, but in my opinion things like metadata limits should be described in the docs properly and left up to a user. 

I've just come to another idea to consider which is to introduce a config mode, i.e. `strict` vs `relaxed` (or like advanced/basic) based on which a user may omit specifying many parameters (prohibiting or enforcing defaults) keeping that decided by them still. On the other hand, we probably should not complicate things that much..

docs/source/rfc-configuration-overhaul.md
Mingela: It will be a different repo or wiki or smth like that

docs/source/rfc-configuration-overhaul.md
```diff
@@ -0,0 +1,912 @@
...
+- **User Transition**: Given the main goal of this RFC — simplifying the configuration for end users — we anticipate a
+  minimal migration pain for current users. The proposed changes, informed by earlier sections of this RFC, are intended
+  to streamline and simplify the process for them.
+- **Deployment & Migration**: The revamped configuration could bring new deployment methods or procedures. Clear
+  documentation will be essential to guide system administrators and users through these changes.
+
+### User Impact
+
+- **Enhanced User Experience**: The main drive behind these changes is to offer users a more intuitive configuration
```
adaagava, line 865: Do we need some sort of tool for easy converting from old config to the new one?

0x009922:
The RFC is moved here:

- https://github.com/hyperledger/iroha-rfcs/pull/3

docs/source/rfc-configuration-overhaul.md
0x009922: Probably. Either a tool or a migration guide. I believe it is not a major point, since we don't have many users as of now, but it is worth it note this in the RFC.

docs/source/rfc-configuration-overhaul.md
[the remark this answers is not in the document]
0x009922: > things like metadata limits should be described in the docs

Everything should no doubt be described in the docs!

> `strict` vs `relaxed`

It might be convenient. Let's leave this discussion to the stage of the actual documentation reference creation. It will be the best time to make decisions based on what might be the best for users.

Also, I had an idea of having `extends` mechanism. With it, users might _extend_ existing presets. For example:

```toml
extends = "./base.toml"
# or even some built-in preset?
extends = "built-in:strict"
```

`extends` mechanism plays well with the "Trace Configuration Resolution" section.

docs/source/rfc-configuration-overhaul.md
0x009922: Addressed: https://github.com/hyperledger/iroha-rfcs/pull/3/commits/f89aff51901362918e395c081d3c88132bc3a39d

PR #3809, opened 2023-08-13
[documentation]: Release roadmap

appetrosyan:
The proposed release roadmap submitted for review as a recommendation.


docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
+#+TITLE: Recommended Road
```
WRRicht3r, line 1:
```suggestion
#+TITLE: Recommended Release Roadmap for IROHA v2
```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+
+* State of the project
+
+Iroha v2 is a capable ledger with notable implemented subroutines that
+allow it to be deployed successfully as a private blockchain ledger:
+1. It has relatively good performance characteristics regarding
+   throughput (post RC9)
+2. It has good performance characteristics regarding transaction
+   finality and latency (post RC19 also query latency).
+3. It has support for Turing complete permission validation, able to
```
WRRicht3r, line 13:
```suggestion
3. It has support for Turing complete permission validation and is able to
```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+   sandbox, with a full array of events which can trigger these
+   behaviours.
+6. The ledger has three SDKs within the JVM, JavaScript and Python
+   ecosystems, allowing the users to easily adapt the Iroha node to
+   their requirements. 
+
+
+The ledger has been demonstrated to work during the ADB CSD/RTGS
+linkages project, and is slated to be used in two projects internally
+by Soramitsu which only utilise the private blockchain ledger
```
WRRicht3r, line 27:
```suggestion
by SORAMITSU which only utilise the private blockchain ledger
```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+upgrade-able) permission validator modules, allowing the blockchain
+invariants to evolve and change with the operation of the blockchain.
+It is capable of resolving simple soft-forks, and has tooling to
+quickly restart a blockchain by squashing all of the information and
+history into a single genesis block. However, it does not retain
+enough flexibility to be used as a single ledger for both private and
+public blockchains. And the following issues have been identified.
+
+1. The more efficient ISI-based language is not usable. The
+   expressions are not sufficiently expressive.
```
WRRicht3r, line 48: 🤯

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+2. WASM is a sub-optimal medium for storage of executable logic. It
+   alone is responsible for 80% of the complexity present in the code
+   base and around 95% of technical debt is related to some mitigation
+   of a shortcoming of WASM in the blockchain.
+
+   The advantages and disadvantages for various media storing Turing
+   complete logic on chain must also be discussed and shall be
+   published as an RFC by the end of August 2023.
+
+3. Iroha lacks any facilities for correlating the consensus process
```
WRRicht3r, line 65:
```suggestion
3. Iroha lacks any facilities to correlate the consensus process
```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+   with the blockchain data. This prevents Proof-of-work,
+   Proof-of-stake, slashing, governance and democracy from being
```
WRRicht3r, line 67:
```suggestion
   Proof-of-stake, slashing, governance, and democracy from being
```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+operating-system-guaranteed interface. One could, in principle write
+smartcontracts in regular =x86= assembly to offer the best size
+optimisations. This can be combined with a scripting language also
+being native supported on-chain, because languages like =lua= and
+Python offer support for native linkage as part of the standard
+interpreter. The JVM would offer a wider range of compatible programs,
+and though the full runtime library might be in excess of a few
+hundred megabytes, this cost must be paid once up-front and provide
+compatibility with a wide range of languages
+[[https://en.wikipedia.org/wiki/List_of_JVM_languages]].
```
WRRicht3r, line 255:
```suggestion

```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+dynamic linkage offers the widest compatibility, as it is a ubiquitous
+operating-system-guaranteed interface. One could, in principle write
+smartcontracts in regular =x86= assembly to offer the best size
+optimisations. This can be combined with a scripting language also
+being native supported on-chain, because languages like =lua= and
+Python offer support for native linkage as part of the standard
+interpreter. The JVM would offer a wider range of compatible programs,
+and though the full runtime library might be in excess of a few
+hundred megabytes, this cost must be paid once up-front and provide
+compatibility with a wide range of languages
```
WRRicht3r, line 254:
```suggestion
compatibility with a [wide range of languages](https://en.wikipedia.org/wiki/List_of_JVM_languages).
```

WRRicht3r (changes requested):

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+
+The ledger has been demonstrated to work during the ADB CSD/RTGS
+linkages project, and is slated to be used in two projects internally
+by Soramitsu which only utilise the private blockchain ledger
+functionality.
+
+The ledger can support un-sharded monolithic blockchains, with limited
+export capability (unlike Iroha v1, v2 uses an internal representation
+and doesn't use a database). Its consensus was designed with Byzantine
+fault tolerance in mind, however, the algorithm -- =sumeragi= has some
```
6r1d, line 33: Should the name be capitalized?

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+intrinsic flaws outlined in [[https://arxiv.org/pdf/1707.01873.pdf][an IBM Research article]], which is notable
+because IBM research originally designed the B-Chain algorithm which
+serves as a prototype for =sumeragi=.
```
6r1d, line 36: same as above

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+1. The more efficient ISI-based language is not usable. The
+   expressions are not sufficiently expressive.
+
+   While this can be somewhat mitigated by the introduction of the
+   DSL, the main expressiveness-boosting facilities of ISI are not
+   usable, and thus only add overhead. A thorough exploration of a
+   restructuring of the Iroha Instruction Set Architecture is needed
+   and shall be published as an RFC in August 2023.
+
+2. WASM is a sub-optimal medium for storage of executable logic. It
```
6r1d, line 56: Thanks for pointing it out :-)

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+
+In this case the following is true:
+- The ISI and Expression language is kept as is,
+- The WASM infrastructure is continued to be built as per the original plan
+- Public blockchain features are added and until the previously agreed
+  upon set of features is implemented (TODO: link Zenhub), no v2.0.0
+  release is carried out.
+- =sumeragi= is kept as the only supported consensus protocol and is
+  supplemented with features necessary to support uses in a public
+  blockchain, but no concerns outlined in the IBM research article is
```
6r1d, line 147: "are addressed"?

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+for hosting multiple versions of multiple machines for validation.
+This greatly simplifies deployment of new nodes, attracting more users
+in the public blockchain space, as well as reducing the operational
+costs to private blockchains.
+
+Removal of WASM, will definitely eliminate the need to implement
+shared-memory dynamic linkage, and instead use the standard-native
+protocol. C-ABI dynamic linkage is implemented by the operating
+system, resulting in the best outcome. For Java some bridging code may
+be needed. For =lua= and Python, the dynamic linkage is an
```
6r1d, line 235: I wonder if I should point out the existence of [LuaJIT](https://luajit.org/), which is heavily optimized for `x86` and `x86-64` platforms? If we're talking about Lua, we must consider maximizing its performance.

6r1d (changes requested):

6r1d:
Thank you. This is an excellent roadmap, please address my questions, I believe some clarification may help others.

docs/source/release_roadmap.org
appetrosyan: @WRRicht3r This is an org document, the link syntax is slightly different: 
```org
[[link text][link destination]]
```

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
+#+TITLE: Recommended Road
+#+AUTHOR: Aleksandr Petrosyan
+#+DATE: 2023-08-13
+
+* State of the project
+
+Iroha v2 is a capable ledger with notable implemented subroutines that
+allow it to be deployed successfully as a private blockchain ledger:
```
Erigara, line 8: I have a feeling that some points sounds a bit subjective. It will be better if we can provide some quantitative measurements here.
Like "good performance characteristics" compared to what?
Or "is a capable ledger" does it have feature parity with it's competitors in the private blockchain context?

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+
+   While this can be somewhat mitigated by the introduction of the
+   DSL, the main expressiveness-boosting facilities of ISI are not
+   usable, and thus only add overhead. A thorough exploration of a
+   restructuring of the Iroha Instruction Set Architecture is needed
+   and shall be published as an RFC in August 2023.
+
+2. WASM is a sub-optimal medium for storage of executable logic. It
+   alone is responsible for 80% of the complexity present in the code
+   base and around 95% of technical debt is related to some mitigation
```
Erigara, line 58: Can you please elaborate on this numbers?

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+   changes upstream, and thus leading to incompatibility between the
+   Iroha APIs for a public and private blockchain.
+
+8. Iroha has no facilities to extend the expression language. This
+   dovetails into point 1, and leads to even less logic being encoded
+   as an ISI or =Expression= and instead relying heavily on WASM.
+
+9. WASM is only well-supported by Rust. While WASM is a
+   general-purpose binary format, only a few languages support it
+   natively: =C++=, =Go=, =nim=, =zig= and =Rust=. =Java=, =Kotlin=
```
Erigara, line 100: This languages cover pretty large share of used programming languages according to this [statistics](https://www.statista.com/statistics/793628/worldwide-developer-survey-most-used-languages/). Also i would noticed that this list is not [full](https://github.com/appcypher/awesome-wasm-langs).

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+   and a few other languages have experimental support with little in
+   terms of either interest or effort.
+
+   Note that all of the above languages are ahead-of-time compiled
+   systems programming languages which require low-level understanding
```
Erigara, line 105: Nitpicking but i wouldn't call `Java` or `Kotlin` system programming language.
This might indicate conformation bias regarding usage of `WASM`.

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+guaranteed backwards compatibility, such as Java, native C-ABI
+libraries, and an embedded scripting language all eliminate the need
+for hosting multiple versions of multiple machines for validation.
+This greatly simplifies deployment of new nodes, attracting more users
+in the public blockchain space, as well as reducing the operational
+costs to private blockchains.
+
+Removal of WASM, will definitely eliminate the need to implement
+shared-memory dynamic linkage, and instead use the standard-native
+protocol. C-ABI dynamic linkage is implemented by the operating
```
Erigara, line 233: I think providing `C-ABI` compatible layer for iroha already proved to be non-trivial task and still ongoing effort.

Erigara (commented):
In general i agree with need to refactor of ISI, however i have some doubts about blaming `WASM` in most of our problems, isn't it picking scapegoat? would like to see RFC with further explanations

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+3. It has support for Turing complete permission validation, able to
+   enforce arbitrarily complex logic on-chain
+4. It has support for a small interpreted VM called the Iroha Special
+   Instructions which allow small-footprint transactions
+5. It has support for Turing complete smartcontracts living in a
+   sandbox, with a full array of events which can trigger these
+   behaviours.
+6. The ledger has three SDKs within the JVM, JavaScript and Python
+   ecosystems, allowing the users to easily adapt the Iroha node to
+   their requirements.
```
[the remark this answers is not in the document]
appetrosyan: Yes. I've added it as point 7.

docs/source/release_roadmap.org
appetrosyan: Fixed.

docs/source/release_roadmap.org
appetrosyan: English plurals are weird.

docs/source/release_roadmap.org
appetrosyan: Remind me of that when I create the WASM-away RFC.

docs/source/release_roadmap.org
appetrosyan: Good point.

docs/source/release_roadmap.org
appetrosyan: `no_std`. Much of the dancing around in the `ffi` library has more to do with what is possible on WASM. The fact that the builds are not reproducible can technically be lumped in here as well, given that things like [reproducible central](https://reproducible-builds.org/docs/jvm/) exist for JVM languages.

docs/source/release_roadmap.org
appetrosyan: Actually, providing a C-ABI compatible layer proved to be a trivial task. What was a problem was making it suitable for smart contract use. @mversic can say more.

docs/source/release_roadmap.org
appetrosyan: For one, this list already contains a half-truth: Rust only supports wasm32-unknown-unknown as a [tier 2 target](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2) and wasm64 is a tier 3. 

Second, Java does not officially support it as a target, instead providing a "use-at-your-own-risk" workaround (you can see that in your second link), corroborated by [official WASM docs](https://webassembly.org/getting-started/developers-guide/). 

Thirdly, I don't think that that statistic supports your claim. In fact, the first item that technically qualifies (see second point) is Java, which ranks 7-th. So at best, we cover only 30% of developers worldwide, unless we start shipping v8 and allowing people to write financial code in JavaScript (not saying it couldn't be done, but it's like hammring a nail with a screwdriver). 

Fourth, awesome-* listicles are not reliable sources of information. One such listicle listed emacs as a terminal editor, gave Vim and Neovim one entry (despite the two projects being extremely different) and listed VSCode and VSCodium as two separate entries, despite the former being a proprietary version of the latter.

appetrosyan: I will admit that the phrasing makes it look like this list is meant to be exhaustive. I'll fix that.

docs/source/release_roadmap.org
appetrosyan: Will fix. I included Java and Kotlin _post factum_ and forgot to propagate the changes.

appetrosyan:
@Erigara 

> In general I agree with need to refactor of ISI

Thank you. 

> however i have some doubts about blaming WASM in most of our problems. Would like to see RFC with further considerations.

I would like to clarify this a bit. We can still do what we want to do with WASM. We can do it with any binary format. Most binary formats are "good enough". However, I don't think that the trade-offs have been considered well and that we can do far better than WASM. I have not done a detailed comparison which would make my point clear, this will come later.

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+   While this can be somewhat mitigated by the introduction of the
+   DSL, the main expressiveness-boosting facilities of ISI are not
+   usable, and thus only add overhead. A thorough exploration of a
+   restructuring of the Iroha Instruction Set Architecture is needed
+   and shall be published as an RFC in August 2023.
+
+2. WASM is a sub-optimal medium for storage of executable logic. It
+   alone is responsible for 80% of the complexity present in the code
+   base and around 95% of technical debt is related to some mitigation
+   of a shortcoming of WASM in the blockchain.
```
Mingela, line 59: tl;dr it also limits SDK usability

referring to SDKs, a developer willing to use smart contracts is basically forced to write ones in Rust since compiling almost everything else to WASM is experimental/prototypical/highly challenging
moving that to the side of SDKs is an overkill at this point

docs/source/release_roadmap.org
Mingela: Unfortunately, most of stuff mentioned as 'supporting WASM' provide hardly usable, very limited, experimental prototypes. So it rather indicates whether some attempts were made in that direction at all, not more. Agree with Aleksandr on that.

docs/source/release_roadmap.org
Mingela: ah, I see it basically mentioned later, nvm

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+
+Furthermore, any change in the API will be accompanied by an extra
+copy of the =iroha_data_model= crate embedded into the Iroha runtime
+in addition to the copies that will occupy space in the Executor. 
+
+As such a monthly release cadence is the preferred solution as it
+allowed for a stable communication channel with the SDKs and the
+community, further boosting the profile of the project.
+
+* Scenario 2: ISA refactor only
```
Mingela, line 193: did `Scenario 1` get missed on purpose

Mingela:
> I would like to clarify this a bit. We can still do what we want to do with WASM. We can do it with any binary format. Most binary formats are "good enough". However, I don't think that the trade-offs have been considered well and that we can do far better than WASM. I have not done a detailed comparison which would make my point clear, this will come later.

@appetrosyan as you may know there was similar comparison in the past, please don't hesitate addressing some points/comments from there too https://wiki.hyperledger.org/display/iroha/Scripting+Languages+and+Runtimes+for+Iroha2+Smart+Contracts

docs/source/release_roadmap.org
Erigara: @Mingela hmm, i don’t see how picking other runtime will stop limiting SDK devs. If we will pick lua then sevs have to write contracts in lua. If jvm them write contract in jvm based language and so on…
Can you explain it a bit further?

appetrosyan: Well, this will be expanded upon in the WASM-specific document. The main issue with WASM specifically is that it is tailored to a web-based use, which has some countervailing tendencies to what we see in general. There's no pressure to stabilise modules, hence no dynamic linkage. It is meant to transport low-level compiled instructions thus no or non-existent run-time support is present. WASM takes up more RAM. WASM must be compiled. WASM is not backwards compatible and WASM must be AOT compiled and then JIT-ted for the particular run-time.  

With a different standard we get a bit more freedom, because the different standard is designed for specific uses. The JVM is perhaps the closest point of comparison. It too has a limited number of supported languages that can target it, but a much wider coverage. Because most JVM languages assume a host-side garbage collector, one has automatic memory management without either the borrow checker, or having to embed GC. Java is smaller for the same source code, because more of the data is embedded in the runtime. 

The runtime itself can be large, but because of late binding the actual memory usage is smaller. Plus we don't have to come up with a dynamic linkage mechanism, we have one. We also have functional multithreading, and 64-bit support. Another important aspect for why JVM would be better is the guaranteed backwards compatibility. Java does not make breaking changes, and you don't have to keep an older version of a runtime in order to preserve compatibility. This means that instead of a complex system of host-side runtime selection, we can just upgrade to the newer version. 

Thus with Java we get a slightly less complex language for smartcontracts that can do slightly more, with the added benefit of being a huge ecosystem tailored to running application logic (and not DOM manipulation), that is significantly more mature. WASM is not the worst possible standard, but we can alleviate a lot of the problems. 

AND THEN, we get to interpreted languages, which have lightweight syntax, don't allow you to do what's not intended, can live concurrently and allow for a lot of transparency. Once I give you the WASM blob, you have a whole song and dance to verify that it belongs to the source code that I gave you. Most people wouldn't bother and that discourages audits. At the same time, saying that the smartcontract is an actual source snippet that is interpreted in the blockchain, tells you a lot more about the logic of the smartcontract and gives you a lot more trust. 

Because for an interpreted language we must expose the FFI bindings, we could then just say "you know what, we don't care, as long as it conforms to some very simple protocol, use Python if that rocks your boat". Then the language becomes an implementation detail for each blockchain, and we're not confined to _one_ choice.

docs/source/release_roadmap.org
appetrosyan: Yes. I forgot to update the numbers.

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+linkages project, and is slated to be used in two projects internally
+by SORAMITSU which only utilise the private blockchain ledger
+functionality.
+
+The ledger can support un-sharded monolithic blockchains, with limited
+export capability (unlike Iroha v1, v2 uses an internal representation
+and doesn't use a database). Its consensus was designed with Byzantine
+fault tolerance in mind, however, the algorithm -- Sumeragi has some
+intrinsic flaws outlined in [[https://arxiv.org/pdf/1707.01873.pdf][an IBM Research article]], which is notable
+because IBM research originally designed the B-Chain algorithm which
```
takemiyamakoto, line 37: not sure what this means. Sisi Duan did this for her PhD thesis at UC Davis

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+The ledger has been demonstrated to work during the ADB CSD/RTGS
+linkages project, and is slated to be used in two projects internally
+by SORAMITSU which only utilise the private blockchain ledger
+functionality.
+
+The ledger can support un-sharded monolithic blockchains, with limited
+export capability (unlike Iroha v1, v2 uses an internal representation
+and doesn't use a database). Its consensus was designed with Byzantine
+fault tolerance in mind, however, the algorithm -- Sumeragi has some
+intrinsic flaws outlined in [[https://arxiv.org/pdf/1707.01873.pdf][an IBM Research article]], which is notable
```
takemiyamakoto, line 36: the paper doesn't mention any intrinsic flaws in B-Chain anywhere in the paper

docs/source/release_roadmap.org
```diff
@@ -0,0 +1,363 @@
...
+
+Iroha v2 is a capable ledger with notable subroutines that allow it to
+be deployed successfully as a private blockchain ledger:
+1. It has relatively good performance characteristics regarding
+   throughput (post RC9).
+2. It has good performance characteristics regarding transaction
+   finality and latency (post RC19 also good query latency).
+3. It has support for Turing complete permission validation, which
+   allows Iroha to enforce arbitrarily complex invariants on-chain. 
+4. It has support for a small interpreted VM called the Iroha Special
```
takemiyamakoto, line 15: Is it correct to call it an "interpreted VM"

docs/source/release_roadmap.org
appetrosyan: not of BChain. I've added quotes to make it clearer what I mean.

docs/source/release_roadmap.org
appetrosyan: Clarified.

6r1d:
Moved to https://github.com/hyperledger/iroha-rfcs/pull/6

PR #3813, opened 2023-08-15
[refactor] #3794: Differentiate between smart contract and trigger entrypoints

### LINKED ISSUE #3794: Differentiate between smartcontract and trigger entrypoints
We need to differentiate between this two in the same way we do for *Validator* -- create another entrypoint macro and import in triggers.

With that we can introduce a proper trigger payload + remove `iroha_derive_primitives` crate (it'll become useless and can be replaced with `darling` if needed)

Arjentix:
## Description

- Added `iroha_trigger` crate under `wasm/trigger` directory
- Removed `get_authority()` wasm imports
- Added payloads for smart contract and trigger and corresponging imports
- Removed entrypoint attribute parsing, arguments are mandatory now
- Removed `derive_primitives` crate because it becomed useless and can be replaced with `darling` crate if needed

### Linked issue

Closes #3794 

### Benefits

- Smart contracts and triggers now also follow the idea of payloads, which makes argument expanding easier
- No useless `derive_primitives` crate
- Less thinking when writing triggers and smart contracts, because you don't need to worry about macro attributes

### Checklist


client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -10,11 +8,11 @@ extern crate panic_halt;
...
 use alloc::{format, string::ToString};
 
-use iroha_wasm::{data_model::prelude::*, prelude::*};
+use iroha_trigger::prelude::*;
 
-#[iroha_wasm::main]
-fn main() {
-    iroha_wasm::info!("Executing trigger");
+#[iroha_trigger::main]
+fn main(_owner: AccountId, _event: Event) {
```
Erigara, line 14: Why arguments are unpacked from payload?

Erigara: Ok, it's current approach to unpack payloads, which might be changed in the future.

Erigara (approved):

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
Arjentix: After a discussion with Roman in DM decided to move it into #3819

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -10,11 +8,11 @@ extern crate panic_halt;
...
+    iroha_trigger::info!("Executing trigger");
```
appetrosyan, line 15: Minor point about language. A smart-contract is a simple "on trigger execute Executable" concept. A trigger is the mechanism that decides that something needs to run, not the thing whose `main` you're calling. 

The old naming convention was somewhat generic, I admit, but it wasn't conceptually dissonant. 

```suggestion
    iroha_hook::info!("Executing trigger");
```

or maybe

```suggestion
    iroha_smartcontract::info!("Executing trigger");
```

Personally prefer the `hook` name, because it's shorter and makes it clear that it acts like lisp hooks do, minus the dynamic typing.

core/src/smartcontracts/wasm.rs
```diff
@@ -850,7 +824,24 @@ impl<'wrld> Runtime<state::SmartContract<'wrld>> {
         let mut store = self.create_store(state);
         let smart_contract = self.create_smart_contract(&mut store, bytes)?;
 
-        Self::execute_main_with_store(&smart_contract, &mut store)
+        let main_fn = Self::get_typed_func(
+            &smart_contract,
```
appetrosyan, line 828: What triggers me is that this is called a smart-contract and _not_ a trigger, while the entrypoint to what's being called by the trigger is called the _trigger main_.

core/src/smartcontracts/wasm.rs
```diff
@@ -850,7 +824,24 @@ impl<'wrld> Runtime<state::SmartContract<'wrld>> {
...
+            &mut store,
+            export::fn_names::SMART_CONTRACT_MAIN,
+        )?;
+
+        // NOTE: This function takes ownership of the pointer
```
appetrosyan, line 833:
```suggestion
        // NOTE: This function takes ownership `store`. 
```

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
appetrosyan: There is a different approach that one can take based on a comment below.

data_model/src/wasm.rs
```diff
@@ -13,8 +13,10 @@ pub mod export {
         pub const WASM_ALLOC: &str = "_iroha_wasm_alloc";
         /// Exported function to deallocate memory
         pub const WASM_DEALLOC: &str = "_iroha_wasm_dealloc";
-        /// Name of the exported entry for smart contract or trigger execution
-        pub const WASM_MAIN: &str = "_iroha_wasm_main";
+        /// Name of the exported entry for smart contract execution
+        pub const SMART_CONTRACT_MAIN: &str = "_iroha_smart_contract_main";
+        /// Name of the exported entry for trigger execution
+        pub const TRIGGER_MAIN: &str = "_iroha_trigger_main";
```
appetrosyan, line 19: Maybe here we need to make a distinction along a different axis. This is related to the above discussion of language.  

We have three things that fit the definition of [smart contracts](https://en.wikipedia.org/wiki/Smart_contract): the single shot executable, or an on-chain program that is sent instead of a transaction, a hook which maps triggers to callbacks and an executor. 

So we could call the three as follows: transaction main, hook main and executor main, propagating the names throughout this pull request.

appetrosyan (commented):
In addition to the nomenclature discussions, I'd like to raise another issue. Iroha need not support only WASM. So far we've been naming the modules after the implementation detail, a binary standard for on-chain executable storage. We could, instead, name it after the role that it fulfils: `smartctontract`, consolidating the interaction with `iroha_core`.

SamHSmith (approved):

Issue #3800, opened 2023-08-11
[inconsistency]: `WsvSeed` is named after what it deserializes (WSV), while `WasmSeed` is named after what it contains (wasmtime runtime).


appetrosyan:
> Naming inconsistency: `WsvSeed` is named after what it deserializes (WSV), while `WasmSeed` is named after what it contains (wasmtime runtime).

Maybe name it `KuraSeed` for consistency?

_Originally posted by @DCNick3 in https://github.com/hyperledger/iroha/pull/3724#discussion_r1273625144_

Issue #3798, opened 2023-08-11
[feature] handle soft fork in `wsv` snapshot


appetrosyan:
> We probably need to store `finalized_wsv` (one block behind `wsv`) because otherwise in case where peer died when it has wrong latest block (soft-fork) we will need to somehow restore finalized wsv in order to resolve soft-fork.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/3724#discussion_r1272294686_

PR #3815, opened 2023-08-16
[refactor] #3796: Add `SnapshotMaker` actor

### LINKED ISSUE #3276: [suggestion] Instruction to print message in the log

### LINKED ISSUE #3796: [refactor] loading `wsv` shouldn't be the part of `kura`.
> Imo loading wsv shouldn't be the part of `kura`.

> We can try to recover `wsv` from backup file and then load necessary blocks (starting from certain height) from kura.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/3724#discussion_r1272262851_

Erigara:
## Description

Separate wsv snapshots into separate actor.

### Linked issue

Closes #3796, #3798, #3800  

### Benefits

More configuration options.

### Checklist

- [x] Create actor for snapshot creation
- [x] Add periodic snapshots
- [x] Write snapshot to tmp file
- [x] Check consistency between `kura` and snapshot
- [x] Save `finalized_wsv` instead of `wsv`

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -17,22 +17,18 @@ use iroha_data_model::block::VersionedCommittedBlock;
 use iroha_logger::prelude::*;
 use iroha_version::scale::{DecodeVersioned, EncodeVersioned};
 use parking_lot::Mutex;
-use serde::{de::DeserializeSeed, Serialize};
 
-use crate::{
-    handler::ThreadHandler,
-    wsv::{WorldStateView, WsvSeed},
-};
+use crate::handler::ThreadHandler;
 
 const INDEX_FILE_NAME: &str = "blocks.index";
 const DATA_FILE_NAME: &str = "blocks.data";
-const SNAPSHOT_FILE_NAME: &str = "snapshot.data";
 const LOCK_FILE_NAME: &str = "kura.lock";
 
 /// The interface of Kura subsystem
 #[derive(Debug)]
 pub struct Kura {
     /// The mode of initialisation of [`Kura`].
+    #[allow(dead_code)]
     mode: Mode,
     /// The block storage
     block_store: Mutex<BlockStore>,
@@ -116,43 +112,8 @@ impl Kura {
     /// - file storage is unavailable
     /// - data in file storage is invalid or corrupted
     #[iroha_logger::log(skip_all, name = "kura_init")]
-    pub fn init(self: &Arc<Self>) -> Result<(Option<WorldStateView>, BlockCount)> {
+    pub fn init(self: &Arc<Self>) -> Result<BlockCount> {
         let block_store = self.block_store.lock();
-
-        let mut wsv = None;
-        let mut skip_block_count = 0usize;
-
-        if matches!(self.mode, Mode::Fast) {
-            let mut data = Vec::new();
-            match block_store.read_snapshot_data(&mut data) {
-                Ok(_) => {
-                    let seed = WsvSeed { kura: self.clone() };
-                    let mut deserializer = serde_json::Deserializer::from_slice(&data);
-                    match seed.deserialize(&mut deserializer) {
-                        Ok(read_wsv) => {
-                            skip_block_count = read_wsv
-                                .height()
-                                .try_into()
-                                .expect("We don't have 4 billion blocks.");
-                            iroha_logger::trace!(?skip_block_count, "Read WSV from snapshot");
-                            wsv = Some(read_wsv);
-                        }
-                        Err(e) => {
-                            iroha_logger::warn!(
-                                ?e,
-                                "Kura initialized in Fast mode but failed to deserialize snapshot file"
-                            )
-                        }
-                    }
-                }
-                Err(e) => {
-                    iroha_logger::warn!(
-                        ?e,
-                        "Kura initialized in Fast mode but failed to read snapshot file"
-                    )
-                }
-            }
-        }
 
         let block_index_count: usize = block_store
             .read_index_count()?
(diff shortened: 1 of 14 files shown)


core/src/snapshot.rs
```diff
@@ -0,0 +1,181 @@
+//! This module contains [`WorldStateView`] snapshot actor service.
+#![allow(
+    clippy::std_instead_of_core,
+    clippy::std_instead_of_alloc,
+    clippy::arithmetic_side_effects
+)]
+use std::{io::Read, path::Path, sync::Arc, time::Duration};
```
Arjentix, line 7:
```suggestion
)]

use std::{io::Read, path::Path, sync::Arc, time::Duration};
```

More lines to the god of lines!

core/src/snapshot.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Name of the temporary [`WorldStateView`] snapshot file.
+const SNAPSHOT_TMP_FILE_NAME: &str = "snapshot.tmp";
+
+/// Errors produced by [`Snapshotter`] actor.
+pub type Result<T, E = ErrReport> = core::result::Result<T, E>;
+
+/// [`Snapshotter`] actor handle.
+#[derive(Clone)]
+pub struct SnapshotterHandle {
+    /// Don't used for actually sending messages but to signal that there is no more handles to [`Snapshotter`]
```
Arjentix, line 32:
```suggestion
    /// Not used to actually send messages but to signal that there is no more handles to [`Snapshotter`]
```

core/src/snapshot.rs
```diff
@@ -0,0 +1,181 @@
...
+        let sumeragi = self.sumeragi.clone();
+        let path_to_snapshot = self.snapshot_path.clone();
+        let handle = tokio::task::spawn_blocking(move || -> Result<u64> {
+            sumeragi.apply_wsv(|wsv| {
+                Self::try_write_snapshot(wsv, &path_to_snapshot)?;
+                Ok(wsv.height())
+            })
+        });
+
+        match handle.await {
```
Arjentix, line 101: I'm not sure this optimization actually does something.
Inside the task there is no async operations, so there is nothing `tokio` can stop at. I believe that `handle.await` will always be just `Ready` and in fact it's a still blocking operation.

core/src/snapshot.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    /// Try deserialize [`WorldStateView`] from snapshot file
+    ///
+    /// # Errors
+    /// - IO errors
+    /// - Deserialization errors
+    pub fn try_read_snapshot(
+        path_to_snapshot: impl AsRef<Path>,
```
Arjentix, line 121:
```suggestion
        snapshot_dir: impl AsRef<Path>,
```

core/src/snapshot.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// Serialize and write snapshot to file,
+    /// overwriting any previously stored data.
+    ///
+    /// # Errors
+    /// - IO errors
+    /// - Serialization errors
+    pub fn try_write_snapshot(
+        wsv: &WorldStateView,
+        path_to_snapshot: impl AsRef<Path>,
```
Arjentix, line 156:
```suggestion
        snapshot_dir: impl AsRef<Path>,
```

core/src/snapshot.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(wsv)
+    }
+
+    /// Serialize and write snapshot to file,
+    /// overwriting any previously stored data.
+    ///
+    /// # Errors
+    /// - IO errors
+    /// - Serialization errors
+    pub fn try_write_snapshot(
```
Arjentix, line 154: Shouldn't it be private?

core/src/snapshot.rs
Erigara: This way we allow `tokio` executor to switch to another task, [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) was deliberately created for this purpose (and [`Future`](https://docs.rs/tokio/latest/src/tokio/runtime/task/join.rs.html#312-344) impl `JoinHandle`).

Erigara:
I'm not very happy about solution for snapshotting finalized wsv, because we add yet another clone of wsv. I think we should rethink and refactor how we work with wsv in general.

SamHSmith (approved):

core/src/kura.rs
```diff
@@ -18,22 +18,18 @@ use iroha_data_model::block::VersionedCommittedBlock;
...
 const DATA_FILE_NAME: &str = "blocks.data";
-const SNAPSHOT_FILE_NAME: &str = "snapshot.data";
 const LOCK_FILE_NAME: &str = "kura.lock";
 
 /// The interface of Kura subsystem
 #[derive(Debug)]
 pub struct Kura {
     /// The mode of initialisation of [`Kura`].
+    #[allow(dead_code)]
     mode: Mode,
```
pesterev, line 33: Why not remove it?

```suggestion
```

Erigara: We still have plans to bring it back. I’m going to work on it after vacations.

SamHSmith (approved):

core/src/snapshot.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Error (de)serializing [`WorldStateView`] snapshot
+    Serialization(#[from] serde_json::Error),
+    /// Snapshot is inconsistent state. Snapshot has greater height ({snapshot_height}) than kura block store ({kura_height})
+    MismatchedHeight {
+        /// Amount of block hashes stored by snapshot
+        snapshot_height: usize,
+        /// Amount of blocks stored by [`Kura`]
+        kura_height: usize,
+    },
+    /// Snapshot is inconsistent state. Hash of the block at height {height} is different between snapshot ({snapshot_block_hash}) and kura ({kura_block_hash})
```
Arjentix, line 217:
```suggestion
    /// Snapshot is in a non-consistent state. Hash of the block at height {height} is different between snapshot ({snapshot_block_hash}) and kura ({kura_block_hash})
```

core/src/wsv.rs
```diff
@@ -292,20 +292,20 @@ pub struct WorldStateView {
 }
 
 /// Context necessary for deserializing [`WorldStateView`]
```
Arjentix, line 294:
```suggestion
/// Context necessary for deserializing [`Kura`]
```

Arjentix: Maybe move to the `kura` module then?

Arjentix (approved):

core/src/wsv.rs
Erigara: @Arjentix we need this for deserializing `WorldStateView`, i renamed it for the sake of consistency. Check out #3800.

pesterev (approved):

PR #3817, opened 2023-08-17
[RFC]: ISA restructuring

appetrosyan:


docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,515 @@
...
+particular instruction in a transaction fails to apply, all
+instructions effects must be reversed. But in the given model, such a
+concept might not be possible to define, and as such, the only way to
+enforce atomicity, given that any transaction can affect potentially
+an unconstrained number of objects, the result is that the largest
+object in the blockchain must be cloned before each application. There
+are relatively efficient ways of mitigating this with persistent data
+structures but all of them pale in comparison with the efficiency of
+having a simpler instruction set in which a tiny addition to a number
+is reverted by changing that tiny number back. This is not impossible
```
Erigara, line 187: I think this claim is not that obvious.
With current default configuration we can have up to 4096 instructions per transaction if we will adopt RISC isi we can expect this number to increase. In such case reverting N instructions might be more costly then using something like MVCC which is commonly used for database management systems, which also can provide us WSV view at particular height, which might be very useful for heavy quires to provide snapshot isolation and providing finalized wsv to resolve soft-forks.

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,732 @@
...
+module itself are met (e.g. governance vote).  This in turn exposed
+several inefficiencies and some tension between what is considered a
+blockchain invariant and what comprises a blockchain-agnostic,
+universal invariant that must be upheld on all Iroha-based
+blockchains.  To address those inefficiencies it was decided that
+instead of a dynamically loadable permission validator, there would be
+a dynamically loadable executor, which is responsible not only for
+checking if the instruction application violates any of the
+invariants, and checks for permissions, but is responsible wholly for
+upholding the blockchain invariants.
```
Arjentix, line 52: I'm not very sure about that, becuase it requires users to understand even more things and edge cases given that they already have to know too much to just be able to write validator/executor.

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,732 @@
...
+instructions is submitted, in principle, the executor can allow or
+deny the applications of certain instruction combinations.  At which
+point the permission tokens became ill-defined, as their meaning is
+undefined until a specific executor is loaded.  Within these
+constraints, the distinction between blockchain and host invariants
+becomes clearly defined: everything that isn't enforced by a failure
+in applying an instruction is a blockchain invariant, including
+permissions. The executor is responsible for enforcing the blockchain
+invariants, and enforcement of host instructions is propagated from
+the host to the executor as well.
```
Arjentix, line 66: > and enforcement of host instructions is propagated from the host to the executor as well

What does it mean, I don't get it

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,732 @@
...
+arbitrarily complex. However, by virtue of being interpreted, the
+expressions are slow. Due to having the user compose expressions and
+dealing with a threadbare model that does not have a front-end in the
+form of a DSL, the model isn't particularly expressive either.
+Further, to add insult to injury, while this model could in principle
+be extensible, given that expressions within /certain tags only/ are
+implemented, it would be possible to add custom, blockchain-specific
+expressions. However, at present this is not possible. As an example
+of where this functionality would have been useful, is CSD/RTGS
+linkages Proof of concept. In which an entire smart-contract had been
```
Arjentix, line 97: Can we write this info in a public document?

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,732 @@
...
+a /very/ specific financial connotation. Furthermore, the =Store=
+asset becomes more similar to a piece of =Metadata= than an =Asset=
+making it difficult to differentiate between the two. In other words,
+despite the words being specifically chosen to avoid confusion,
+because of the lack of a general purpose
+=ThisInstructionOperatesOnAssets= and
+=ThisInstructionOperatesOnMetadata=, the API discoverability is hurt
+and the generality is hurt also. Insufficient generality also
+manifests as having too many instructions: the current exhaustive list
+o finstructions is the following enumeration:
```
Arjentix, line 128:
```suggestion
of instructions is the following enumeration:
```

docs/source/rfc/isa_restructuring.org
Arjentix: Isn't something what @SamHSmith wanted to implement using versions?

Arjentix: Also as I understand the idea is to have complex instructions implemented as a DSL inside Executor. That way we could also implement reverting instructions inside Executor.

I would expect that we could decrease the amount of complex instructions used by user in transaction but increase the amount of native instructions used by Executor to express complex instructions.

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,732 @@
...
+different elements of the ISA. We have both a =SequenceBox=, and
+multiple instructions comprising a single transaction. There's both an
+=If= as an =instruction= and an =If= as an =Expression=. Further
+obfuscating the types, each instruction accepts an expression which
+either wraps a =Value= or a collection of operations that are
+ill-defined for some of the potential variants of =Value=.
+
+Finally, compounding the issue, the Iroha special instructions were
+meant to be accompanied by a domain-specific language that is only now
+being written.
```
Arjentix, line 215: Were they really meant to be used in DSL? I haven't seen such an issue/task, while I've seen a lot of overs old ones

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,732 @@
...
+- The instruction set can be completely altered and tailored to
+  specific blockchain needs. Expressions can be eliminated or extended
+  to arbitrarily complex operations.
+- All logic is defined in one place and thus easy to audit.
+- All logic can be patched at runtime.
+
+This approach necessarily requires the Executor to be implemented in a
+compiled language. Further, while it can cope with the current choice
+of WASM, it is affected by some of the limitations of said binary
+format. For instance: because WASM is 32-bit-only, only 4GiB of world
```
Arjentix, line 244: What about [wasm64-unknown-unknown](https://doc.rust-lang.org/rustc/platform-support/wasm64-unknown-unknown.html)?

docs/source/rfc/isa_restructuring.org
appetrosyan: > Isn't something what @SamHSmith wanted to implement using versions?

I was about to say that. 

MVCC doesn't care if the instruction set is RISC (and only one of the proposals entails using a RISC ISA). 

It is true that we can have up to 4096 instructions per transaction. It is also true that under this proposal, frequently used combinations are to be encoded as expressions and not as bare instructions, so the newly proposed blockchain is likely to be smaller, not larger. 

In variant 2, I propose having RISC, but of a very particular kind. It comprises essentially the fragment of the final state, so the collection of instructions can be viewed as a `WorldStateView`. What this means is that @SamHSmith will get to implement something very similar to working sets, but with a generalised application. 

> I would expect that we could decrease the amount of complex instructions used by user in transaction but increase the amount of native instructions used by Executor to express complex instructions.

Indeed that's the intention. The idea is that RISC is a very general flexible language that is to be generated by smartcontracts, and which can be sent from any compliant client to any compliant host (in which case the executor simply validates the transaction). 

Any other complex operation which is likely to be re-used, is an expression with implementation-defined representation. So if Iroha is running a scientific lab, or a set of smart meters where none of our domain-specific financial instructions are useful, most of the blockchain contains `ExpressionBox` variants, which make sense for this application.

docs/source/rfc/isa_restructuring.org
appetrosyan: That is a problem that is hidden in the old approach also. We just make it explicit.

docs/source/rfc/isa_restructuring.org
appetrosyan: Roughly, that the host no longer checks if an instruction violates any invariants. There's no distinction between host invariants and blockchain invariants anymore and so the logic that says "Instruction X bad, because it caused overflow" is moved up into the executor as a static library, not as a virtual call of a host function.

docs/source/rfc/isa_restructuring.org
appetrosyan: At this point the entire report is public, so yes. Before the end of my notice period I'll upload the report to the Iroha 2 docs repo, so we can actually refer to the precise locations where I point out what we would have found useful.

docs/source/rfc/isa_restructuring.org
appetrosyan: As per [rust-lang](https://doc.rust-lang.org/rustc/platform-support/wasm64-unknown-unknown.html):

> This target is not a stable target. The [memory64](https://github.com/webassembly/memory64) WebAssembly proposal is still in-progress and not standardized. This means that there are not many engines which implement the memory64 feature and if they do they're likely behind a flag, for example: ...

Which I should include in the documentation.

docs/source/rfc/isa_restructuring.org
mversic: I don't see a problem with this paragraph. Even if blockchain doesn't use executor for upholding blockchain invariants it certainly can do so

docs/source/rfc/isa_restructuring.org
mversic: > At which point the permission tokens became ill-defined, as their meaning is undefined until a specific executor is loaded

I don't agree with this point. Permission tokens are not ill-defined until executor is loaded, rather they don't exist at all (i.e. permission tokens set is empty and you can verify this with `FindPermissionTokens` query)

docs/source/rfc/isa_restructuring.org
appetrosyan: There was a `dsl` crate that I removed in November 2021. The intentions were not always committed to GitHub, because we used to use Jira for the issue tracking until August 2021.

docs/source/rfc/isa_restructuring.org
appetrosyan: Addressed it in text.

6r1d:
Moved to https://github.com/hyperledger/iroha-rfcs/pull/7

docs/source/rfc/isa_restructuring.org
```diff
@@ -0,0 +1,787 @@
...
+of host instructions is propagated from the host to the executor as
+well.
+
+* Shortcomings of the current ISA
+
+** The Expression problem
+
+Firstly, while the Iroha blockchain ledger has an expression language
+that is dangerously close to being Turing complete (more precisely by
+being a lambda calculus), the expression language is largely unused.
```
0x009922, line 91: > the expression language is largely unused

So true

PR #3824, opened 2023-08-18
[refactor] #3761: Rewrite `scripts/test-env.sh`

### LINKED ISSUE #3761: [suggestion] Rewrite `test_env.sh` to POSIX shell
### Feature request

We should provide a rewrite of the `test_env.sh` that can be run on regular POSIX shell; namely [`dash`](https://wiki.archlinux.org/title/Dash). 

### Motivation

Bash is not a universal shell, and not all distributions provide the bash version that has associative arrays. 

### Who can help?

_No response_

ilchu:
## Description

Rewrite the `test_env.sh` script as python. This should help with running it on systems that don't have bash as default due to python being more ubiquitous.

### Linked issue

Closes #3761. 

### Benefits
Slightly more descriptive docs and messages produced by the script, presumably increased readability.

### Checklist

- [x] I've run the script with multiple configurations and with the oldest officially supported version of python to make sure it's sufficiently backwards compatible


scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+                        it kills all processes named `iroha`, so proper caution is advised")
+
+    parser.add_argument("--out_dir", "-o", default="./test", type=pathlib.Path,
+                        help="Directory to store config and log files. \
+                        Defaults to `./test`. If setup was run with a custom directory, \
+                        the same directory must be provided to cleanup, otherwise only \
+                        peer processes will be destroyed")
+    parser.add_argument("--host_addr", "-a", default="127.0.0.1", type=ipaddress.IPv4Address,
+                        help="IP address of the host machine. Used in trusted peer \
+                        generation. Defaults to localhost without port specified")
```
DCNick3, line 185: > without port specified

I don't quite get what it means..  AFAIK `ipaddress.IPv4Address` doesn't allow the port to be specified

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+def setup(args):
+    logging.info("Starting iroha network with %d peers", args.n_peers)
+    setup_dir = pathlib.Path(args.out_dir)
+    try:
+        os.mkdir(setup_dir)
+    except FileExistsError:
+        logging.error("Test directory `%s` already exists", setup_dir)
+        sys.exit(1)
+    try:
+        shutil.copy2("./target/debug/iroha_client_cli", setup_dir)
```
DCNick3, line 141: In case the script is ran not from the root of the repo, this error will be triggered. Maybe check that it is the case or even locate it automagically and make it the workdir?

Or, for better code clarity, use it explicitly, like the `setup_dir`

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+
+class Network:
+    """
+    A network of bootstrapped peers to run on bare metal.
+    """
+    def __init__(self, setup_dir: pathlib.Path, host_addr: ipaddress.IPv4Address, n_peers: int = 4):
+        logging.info("Setting up test environment...")
+        self.setup_dir = setup_dir
+        self.n_peers = n_peers
+        peers_dir = setup_dir.joinpath("peers")
```
DCNick3, line 26: `pathlib` supports a funny syntax for path joins: `setup_dir / "peers"`. Sometimes it might be easier to read

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+        os.environ["IROHA2_GENESIS_PATH"] = str(setup_dir.joinpath("peers/genesis.json"))
+        os.environ["IROHA_GENESIS_ACCOUNT_PUBLIC_KEY"] = self.peers[0].public_key
+        os.environ["IROHA_GENESIS_ACCOUNT_PRIVATE_KEY"] = self.peers[0].private_key
+
+        self.generate_trusted_peers()
+    
+    def generate_trusted_peers(self):
+        self.trusted_peers = []
+        for peer in self.peers:
+            self.trusted_peers.append(f'{{"address": "{peer.host_addr}:{peer.p2p_port}", "public_key": "{peer.public_key}"}}')
```
DCNick3, line 54: Why not use the `json` module to generate the JSON here? (you would need to first construct the object using dicts and lists and serialize it with `json.dumps`)

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+        json_keypair = json.loads(str_keypair)
+        # public key is a string, private key is a json object
+        self.public_key = json_keypair['public_key']
+        self.private_key = json.dumps(json_keypair['private_key'])
+
+    def run(self, is_genesis: bool = False):
+        logging.info("Running peer %s...", self.name)
+        peer_dir = self.setup_dir.joinpath(f"peers/{self.name}")
+        os.mkdir(peer_dir)
+        os.mkdir(peer_dir.joinpath("storage"))
```
DCNick3, line 94: ditto about path joins in this function

scripts/test_env.py
appetrosyan: Given that Python is quack typed, I'd err on the side of safety.

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+    def __init__(self, setup_dir: pathlib.Path, host_addr: ipaddress.IPv4Address, n_peers: int = 4):
+        logging.info("Setting up test environment...")
+        self.setup_dir = setup_dir
+        self.n_peers = n_peers
+        peers_dir = setup_dir.joinpath("peers")
+        os.mkdir(peers_dir)
+        try:
+            shutil.copy2("./configs/peer/config.json", peers_dir)
+            shutil.copy2("./configs/peer/genesis.json", peers_dir)
+            shutil.copy2("./configs/peer/validator.wasm", peers_dir)
```
appetrosyan, line 31: I'd recommend making these configurable later down the line. If we run this script from a different location it should be possible to retrieve the project root.

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+        except FileNotFoundError:
+            logging.error("Some of the config files are missing. \
+                          Please provide them in the `./configs/peer` directory")
```
appetrosyan, line 34: Avoid hard-coding paths if possible.

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+            sys.exit(1)
+        try:
+            shutil.copy2("./target/debug/iroha", peers_dir)
+        except FileNotFoundError:
+            logging.error("Please build the `iroha` by running `cargo build --bin iroha`")
```
appetrosyan, line 39: Or you could prompt the user if you want to do that on their behalf. No harm in asking a `y/n` question.

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+        self.peers = [Peer(setup_dir, host_addr, i) for i in range(n_peers)]
+
+        os.environ["IROHA2_CONFIG_PATH"] = str(setup_dir.joinpath("peers/config.json"))
+        os.environ["IROHA2_GENESIS_PATH"] = str(setup_dir.joinpath("peers/genesis.json"))
+        os.environ["IROHA_GENESIS_ACCOUNT_PUBLIC_KEY"] = self.peers[0].public_key
+        os.environ["IROHA_GENESIS_ACCOUNT_PRIVATE_KEY"] = self.peers[0].private_key
+
+        self.generate_trusted_peers()
+    
+    def generate_trusted_peers(self):
```
appetrosyan, line 51: I wouldn't mind if this function were inlined. A linear sequence of steps is often easier to follow than a set of functions which are used and called once.

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def run(self):
+        self.peers[0].run(is_genesis=True)
+        for peer in self.peers[1:]:
+            peer.run()
+
+class Peer:
+    """
+    A single bootstrapped peer. Could be a genesis node or a regular peer.
+    Should not be run directly, but rather as a part of a Network.
```
appetrosyan, line 65: >  Should not be run directly,

Which is more than likely going to be ignored by Pythonistas that "know better". At least indicate it with the `_`.

scripts/test_env.py
appetrosyan: Ditto about Python being weird and often problematic.

scripts/test_env.py
appetrosyan, line 1: One overarching problem is that this python script assumes UNIX-like paths and has implicit assumptions about where it's being run from. Maybe we should consider defining `IROHA_PROJECT_ROOT`?

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+        os.environ["TORII_API_URL"] = f"{self.host_addr}:{self.api_port}"
+        os.environ["TORII_TELEMETRY_URL"] = f"{self.host_addr}:{self.telemetry_port}"
+        os.environ["TOKIO_CONSOLE_ADDR"] = f"{self.host_addr}:{self.tokio_console_port}"
+
+        genesis_arg = "--submit-genesis" if is_genesis else ""
+        # FD never gets closed
+        log_file = open(peer_dir.joinpath("out.log"), "w")
+        # These processes are created detached from the parent process already
+        subprocess.Popen([iroha_bin, f"{genesis_arg}"],
+                    stdout=log_file, stderr=subprocess.STDOUT)
```
appetrosyan, line 116: Notice that in the `exec` in my script, I rename the processes to be `iroha0` and so on. Consider adding this here.

scripts/test_env.py
appetrosyan: Another thing I noticed is the inconsistent use of `f"{}"` strings. I'd commit to the newer style, given that it's now basically a standard.

scripts/test_env.py
```diff
@@ -0,0 +1,191 @@
...
+    else:
+        raise argparse.ArgumentTypeError(f"Argument {arg} must be a positive integer")
+
+def main(args):
+    logging.basicConfig(level=logging.INFO if args.verbose else logging.WARNING,
+        format="[%(asctime)s] [%(levelname)s] %(message)s")
+    if args.command == "setup":
+        setup(args)
+    elif args.command == "cleanup":
+        cleanup(args)
```
appetrosyan, line 130: Can we use the same convention as iroha?

scripts/test_env.py
DCNick3: `click` library is great for that btw

scripts/test_env.py
ilchu: Or just pass it as an optional argument? Defaulting to the current directory or such. And with string formatting, I was following the old-fashioned logging docs' examples, never really checking if the newer style works. Will fix if it does.

scripts/test_env.py
ilchu: Was following the style of the original script here, but that makes sense. I'm trying to avoid using external dependencies though to make it runnable on bare systems.

scripts/test_env.py
ilchu: But that's what actually is being done here. I tried to do child renaming with `exec`, but `subprocess` was being weird about it and many hours were lost. So I opted for the dumber approach where I rename and copy the binaries into each peer's directory. The only difference is that the process name includes full relative path to it when viewed with `ps` or `top`.

ilchu: But that's what actually is being done here. I tried to do child renaming with `exec`, but `subprocess` was being weird about it and many hours were lost. So I opted for the dumber approach where I rename and copy the binaries into each peer's directory. The only difference is that the process name includes full relative path to it when viewed with `ps` or `top`.

scripts/test_env.py
ilchu: What I meant here is just that the port should not supplied in the host string after the address itself. I'll rephrase it to make it more clear.

scripts/test_env.py
ilchu: Re: POSIX path abuse, found this in `pathlib` docs:
> class pathlib.Path(*pathsegments)[¶](https://docs.python.org/3/library/pathlib.html#pathlib.Path)
A subclass of [PurePath](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath), this class represents concrete paths of the system’s path flavour (instantiating it creates either a [PosixPath](https://docs.python.org/3/library/pathlib.html#pathlib.PosixPath) or a [WindowsPath](https://docs.python.org/3/library/pathlib.html#pathlib.WindowsPath)) ...

scripts/test_env.py
```diff
@@ -0,0 +1,240 @@
...
+
+def setup(args):
+    logging.info(f"Starting iroha network with {args.n_peers} peers...")
+    try:
+        os.mkdir(args.out_dir)
+    except FileExistsError:
+        logging.error(f"Test directory `{args.out_dir}` already exists")
+        sys.exit(5)
+    try:
```
DCNick3, line 184: Maybe it would be cleaner if all the "try to copy, prompt to build" logic was handled in one function? Otherwise there's a fair amount of code duplication..

scripts/test_env.py
```diff
@@ -0,0 +1,240 @@
...
+    if int(arg) > 0:
+        return int(arg)
+    else:
+        raise argparse.ArgumentTypeError(f"Argument {arg} must be a positive integer")
+
+def prompt_build_bin(bin_name: str, root_dir: pathlib.Path, target_dir: pathlib.Path):
+    logging.error(f"The binary `{bin_name}` wasn't found in `{root_dir}` directory")
+    while True:
+        prompt = input(f"Build it by running `cargo build --bin {bin_name}`? (y/n)\n")
+        if prompt.lower() in ["y", "yes"]:
```
DCNick3, line 146: Maybe make `y` the default? So, if the user presses enter it just builds it

(this is often denoted by making the `Y` capitalized in the prompt)

SamHSmith (dismissed):

scripts/test_env.py
ilchu: I did put that logic into a `prompt_build_bin` func though. And other exception handling, while a bit cumbersome, still feels at least as terse as the previous bash script, and definitely more readable.

scripts/test_env.py
```diff
@@ -0,0 +1,244 @@
...
+                        Defaults to 4. If setup was run with a custom number of peers, \
+                        the same number doesn't need to be provided to cleanup as \
+                        it kills all processes named `iroha`, so proper caution is advised")
+
+    parser.add_argument("--out_dir", "-o", default="./test", type=pathlib.Path,
+                        help="Directory to store config and log files. \
+                        Defaults to `./test`. If setup was run with a custom directory, \
+                        the same directory must be provided to cleanup, otherwise only \
+                        peer processes will be destroyed")
+    parser.add_argument("--host_addr", "-a", default="127.0.0.1", type=ipaddress.IPv4Address,
```
0x009922, line 231:
```suggestion
    parser.add_argument("--host-ip", "-a", default="127.0.0.1", type=ipaddress.IPv4Address,
```

scripts/test_env.py
```diff
@@ -0,0 +1,244 @@
...
+                        directory of choice. This is meant to be run from `iroha` root dir. \
+                        The `cleanup` command will kill all peer processes \
+                        that were started by the `setup` command and remove the test directory")
+    parser.add_argument("n_peers", nargs="?", default=4, type=pos_int,
+                        help="Number of peers to bootstrap. \
+                        Defaults to 4. If setup was run with a custom number of peers, \
+                        the same number doesn't need to be provided to cleanup as \
+                        it kills all processes named `iroha`, so proper caution is advised")
+
+    parser.add_argument("--out_dir", "-o", default="./test", type=pathlib.Path,
```
0x009922, line 226:
```suggestion
    parser.add_argument("--out-dir", "-o", default="./test", type=pathlib.Path,
```

`kebab-case` CLI args naming is more consistent with other CLIs in the repo.

scripts/test_env.py
```diff
@@ -0,0 +1,244 @@
...
+                        help="Directory to store config and log files. \
+                        Defaults to `./test`. If setup was run with a custom directory, \
+                        the same directory must be provided to cleanup, otherwise only \
+                        peer processes will be destroyed")
+    parser.add_argument("--host_addr", "-a", default="127.0.0.1", type=ipaddress.IPv4Address,
+                        help="IP address of the host machine. Used in trusted peer \
+                        generation. Defaults to localhost. Note that the port/s shouldn't \
+                        be provided as for each peer's endpoints they're assigned automatically")
+    parser.add_argument("--root_dir", "-r", default=".", type=pathlib.Path,
```
0x009922, line 235:
```suggestion
    parser.add_argument("--root-dir", "-r", default=".", type=pathlib.Path,
```

scripts/test_env.py
```diff
@@ -0,0 +1,244 @@
...
+        os.environ["TORII_P2P_ADDR"] = f"{self.host_addr}:{self.p2p_port}"
+        os.environ["TORII_API_URL"] = f"{self.host_addr}:{self.api_port}"
+        os.environ["TORII_TELEMETRY_URL"] = f"{self.host_addr}:{self.telemetry_port}"
+        os.environ["TOKIO_CONSOLE_ADDR"] = f"{self.host_addr}:{self.tokio_console_port}"
+
+        genesis_arg = "--submit-genesis" if is_genesis else ""
+        # FD never gets closed
+        log_file = open(peer_dir.joinpath(".log"), "w")
+        # These processes are created detached from the parent process already
+        subprocess.Popen([iroha_bin, f"{genesis_arg}"],
```
0x009922, line 137:
```suggestion
        subprocess.Popen([iroha_bin, genesis_arg],
```

scripts/test_env.py
```diff
@@ -0,0 +1,244 @@
...
+    else:
+        raise argparse.ArgumentTypeError(f"Argument {arg} must be a positive integer")
+
+def prompt_build_bin(bin_name: str, root_dir: pathlib.Path, target_dir: pathlib.Path):
+    logging.error(f"The binary `{bin_name}` wasn't found in `{root_dir}` directory")
+    while True:
+        prompt = input(f"Build it by running `cargo build --bin {bin_name}`? (Y/n)\n")
+        if prompt.lower() in ["y", "yes", ""]:
+            subprocess.run(
+                ["cargo", "build", "--bin", f"{bin_name}"],
```
0x009922, line 152:
```suggestion
                ["cargo", "build", "--bin", bin_name],
```

Maybe I miss something...

0x009922 (approved):

scripts/test_env.py
DCNick3: It's just that the

```python
try:
    shutil.copy2(f"{args.root_dir}/target/debug/somebin", args.out_dir)
except FileNotFoundError:
    prompt_build_bin("somebin", args.root_dir, args.out_dir)
```

snippet is repeated a bunch of times, while it can itself be part of the `prompt_build_bin` function (probably would need another name). The repetition is not _that_ bad tbh, but I don't see where it would hurt to reduce it

0x009922 (approved):

scripts/test_env.py
ilchu: Good point, sorry for misunderstanding.

0x009922 (approved):

DCNick3 (approved):

PR #3831, opened 2023-08-21
[fix] #3059: Simplify kura lock file code

### LINKED ISSUE #3059: Kura graceful shutdown
sometimes when executing benchmarks and sending a shutdown signal the following message appears:
```
2023-01-11T15:00:14.233857Z  WARN iroha_core::kura: Couldn't remove lockfile at Locked("/var/folders/hv/g1qbnmwj2ylc6nx9mdsbb2dr0000gn/T/.tmp6XYfkx"). Either the lockfile was deleted externally while Iroha instance was running, or there was an IO error. Check the block store with kura_inspector and then delete the lockfile manually. err=Tried to unlock block store by deleting lockfile at /var/folders/hv/g1qbnmwj2ylc6nx9mdsbb2dr0000gn/T/.tmp6XYfkx, but it couldn't be found
```

SamHSmith:
### Linked issue

Closes #3059  

### Checklist

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -11,7 +11,6 @@ use std::{
     sync::Arc,
 };
 
-use derive_more::Deref;
 use iroha_config::kura::Mode;
 use iroha_crypto::HashOf;
 use iroha_data_model::block::VersionedCommittedBlock;
@@ -36,7 +35,7 @@ pub struct Kura {
     /// The mode of initialisation of [`Kura`].
     mode: Mode,
     /// The block storage
-    block_store: Mutex<BlockStore<Locked>>,
+    block_store: Mutex<BlockStore>,
     /// The array of block hashes and a slot for an arc of the block. This is normally recovered from the index file.
     #[allow(clippy::type_complexity)]
     block_data: Mutex<
@@ -62,7 +61,7 @@ impl Kura {
         block_store_path: &Path,
         debug_output_new_blocks: bool,
     ) -> Result<Arc<Self>> {
-        let mut block_store = BlockStore::new(block_store_path).lock()?;
+        let mut block_store = BlockStore::new(block_store_path, LockStatus::Unlocked);
         block_store.create_files_if_they_do_not_exist()?;
 
         let block_plain_text_path = debug_output_new_blocks.then(|| {
@@ -84,10 +83,9 @@ impl Kura {
     /// Create a kura instance that doesn't write to disk. Instead it serves as a handler
     /// for in-memory blocks only.
     pub fn blank_kura_for_testing() -> Arc<Kura> {
-        let block_store = BlockStore::fake_locked_for_tests(PathBuf::new());
         Arc::new(Self {
             mode: Mode::Strict,
-            block_store: Mutex::new(block_store),
+            block_store: Mutex::new(BlockStore::new(&PathBuf::new(), LockStatus::Locked)),
             block_data: Mutex::new(Vec::new()),
             block_plain_text_path: None,
         })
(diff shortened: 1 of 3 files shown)


core/src/kura.rs
```diff
@@ -62,7 +62,7 @@ impl Kura {
         block_store_path: &Path,
         debug_output_new_blocks: bool,
     ) -> Result<Arc<Self>> {
-        let mut block_store = BlockStore::new(block_store_path).lock()?;
+        let mut block_store = BlockStore::new(block_store_path, false);
```
appetrosyan, line 65: Consider 
```suggestion
        let mut block_store = BlockStore::new(block_store_path, BlockStoreType::Unlocked);
```

core/src/kura.rs
```diff
@@ -84,10 +84,9 @@ impl Kura {
     /// Create a kura instance that doesn't write to disk. Instead it serves as a handler
     /// for in-memory blocks only.
     pub fn blank_kura_for_testing() -> Arc<Kura> {
-        let block_store = BlockStore::fake_locked_for_tests(PathBuf::new());
         Arc::new(Self {
             mode: Mode::Strict,
-            block_store: Mutex::new(block_store),
+            block_store: Mutex::new(BlockStore::new(&PathBuf::new(), true)),
```
appetrosyan, line 89:
```suggestion
            block_store: Mutex::new(BlockStore::new(&PathBuf::new(), BlockStoreType::Locked)),
```

core/src/kura.rs
```diff
@@ -536,9 +415,48 @@ pub struct BlockIndex {
...
+                        }
+                    }
+                    _ => Err(Error::IO(e, path)),
+                }
+                .expect("Kura must be able to lock the blockstore");
+            }
+        }
+        BlockStore {
+            path_to_blockchain: path.to_path_buf(),
+        }
```
appetrosyan, line 457:
```suggestion
     else {
        BlockStore {
            path_to_blockchain: path.to_path_buf(),
        }
     }
```

The control flow is a bit confusing with this level of nesting.

core/src/kura.rs
```diff
@@ -536,9 +415,48 @@ pub struct BlockIndex {
...
+            let path = path.join(LOCK_FILE_NAME);
+            if let Err(e) = fs::File::options()
+                .read(true)
+                .write(true)
+                .create_new(true)
+                .open(path.clone())
+            {
+                match e.kind() {
+                    std::io::ErrorKind::AlreadyExists => Err(Error::Locked(path)),
+                    std::io::ErrorKind::NotFound => {
```
appetrosyan, line 431: One of the worst designed features of the standard library.

appetrosyan (commented):
Like it. Less code. In principle, could be done with a `const generic` and I'd honestly prefer an `enum` to make things a bit easier to understand, but no major objections.

core/src/kura.rs
SamHSmith: I reordered to make it slightly easier on the eye.

core/src/kura.rs
```diff
@@ -449,81 +391,17 @@ pub struct BlockCount {
...
-    /// - IO errors
-    /// - lockfile already exists
-    pub fn lock(self) -> Result<BlockStore<Locked>> {
-        Ok(BlockStore {
-            path_to_blockchain: self.path_to_blockchain.lock()?,
-        })
+impl Drop for BlockStore {
+    fn drop(&mut self) {
+        let path = self.path_to_blockchain.join(LOCK_FILE_NAME);
+        let _ = fs::remove_file(path); // we don't care if this succeeds or not
```
0x009922, line 404: Shouldn't it at least print a warning/debug message in case of failure?

core/src/kura.rs
```diff
@@ -536,9 +414,53 @@ pub struct BlockIndex {
...
+        if matches!(already_locked, LockStatus::Unlocked) {
+            let path = path.join(LOCK_FILE_NAME);
+            if let Err(e) = fs::File::options()
+                .read(true)
+                .write(true)
+                .create_new(true)
+                .open(path.clone())
+            {
+                match e.kind() {
+                    _ => Err(Error::IO(e, path)),
```
0x009922, line 434: I am probable blind, but match arms evaluation depends on order ([example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=10a0a792c96012f37cde12e27528ee01)). I wonder why rustc doesn't print a warning about unreachable arms here.

core/src/kura.rs
```diff
@@ -536,9 +414,53 @@ pub struct BlockIndex {
...
-/// Operations available both on locked and unlocked store,
-/// i.e. only read operations.
-impl<L: Lock> BlockStore<L> {
+pub enum LockStatus {
+    Locked,
+    Unlocked,
+}
+
+impl BlockStore {
+    /// Create a new read-only block store in `path`.
```
0x009922, line 423: Should this docstring contain information about possible panic inside?

core/src/kura.rs
pesterev: > I wonder why rustc doesn't print a warning about unreachable arms here.

it does i've checked

core/src/kura.rs
```diff
@@ -536,9 +414,53 @@ pub struct BlockIndex {
...
+                                    .open(path.clone())
+                                {
+                                    Err(Error::IO(e, path))
+                                } else {
+                                    Ok(())
+                                }
+                            }
+                        }
+                    }
+                }
```
pesterev, line 455:
```suggestion
                match e.kind() {
                    std::io::ErrorKind::AlreadyExists => Err(Error::Locked(path)),
                    std::io::ErrorKind::NotFound => std::fs::create_dir_all(&path)
                        .map_err(|e| Error::MkDir(e, path.clone()))
                        .and_then(|_| {
                            fs::File::options()
                                .read(true)
                                .write(true)
                                .create_new(true)
                                .open(&path)
                                .map_err(|e| Error::IO(e, path))
                        }),
                    _ => Err(Error::IO(e, path)),
                }
```
it seems matching have to be fixed, also you can simplify it using combinators and remove redundant clones.

core/src/kura.rs
SamHSmith: No. The purpose of this PR is to not do that.

0x009922: I don't really understand why it shouldn't be done, but I am fine with that.

0x009922 (approved):

pesterev (approved):

PR #3832, opened 2023-08-22
[feature] #3812 adding registerbox to DSL

RealNeGate:
## Description

Getting started on a syntax for `RegisterBox` to replace writing `RegisterBox::new(Domain::new("neverland".parse()?))` with `expr!(register "neverland")`, currently there's no way to describe objects which aren't domains. It also makes sense to separate the name detection logic from the DSL.

### Benefits

There's a consistent grammar across different kinds of resources like domains, assets, accounts which can be inferred when creating objects within the DSL. For instance, all assets will be `asset#domain`.

https://hyperledger.github.io/iroha-2-docs/guide/blockchain/naming.html

### Checklist

- [x] syntax for `register`
- [x] Domains
- [ ] Assets
- [x] AssetDefinitions
- [ ] Accounts
- [ ] Moving detection code into separate crate
- [ ] CI checks

### DIFF
--- a/dsl/src/lib.rs
+++ b/dsl/src/lib.rs
@@ -1,5 +1,6 @@
 use std::{convert::TryFrom, iter::Peekable, str::FromStr};
 
+use enum_dispatch::enum_dispatch;
 use litrs::Literal;
 use proc_macro2::{
     token_stream::IntoIter,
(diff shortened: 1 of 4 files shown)


dsl/src/lib.rs
```diff
@@ -71,17 +92,29 @@ fn parse_literal(it: &mut Peekable<IntoIter>) -> ParsedExpr {
...
+            return ParsedExpr {
+                t: ExprType::Domain,
+                tokens: quote!{ Domain::new(#v.parse()?) },
+            };
+        },
 
-    ParsedExpr {
-        t: ExprType::Nil,
-        tokens: TokenStream::new(),
+        _ => {
```
appetrosyan, line 112: Some diagnostics here. You can bind it to 

```suggestion
        val => {
            println!("{val:?}");
```

DCNick3:
Just a design note: wouldn't it be beneficial to have separate macros for expressions and instructions? It feels weird getting different kinds of entities from the same macro depending on the content..

RealNeGate:
> Just a design note: wouldn't it be beneficial to have separate macros for expressions and instructions? It feels weird getting different kinds of entities from the same macro depending on the content..

Sorry for the late reply, yes that's going to be the next step

PR #3835, opened 2023-08-22
[RFC]: on-chain scripting

appetrosyan:


docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,571 @@
...
+
+Sand boxing does not exist by default. In order to sandbox the smart
+contract execution environment, one must execute the smart contract
+from a confined process, which is easily achievable using
+firejail. This is a sub-optimal solution, compared to having a
+pessimistic confinement as there is an extra package that must be
+fetched which constitutes another dynamic dependency for Iroha which
+would not be the case if we used =wasmtime=. 
+
+Reproducility in Rust is problematic as in the case of WASM.
```
0x009922, line 480:
```suggestion
Reproducibility in Rust is problematic as in the case of WASM.
```

docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,571 @@
...
+well-established at this point and its use is largely situational.
+
+Native support for SDKs varies but is usually as good or better than
+with WASM.  JavaScript can not be interpreted without a native
+runtime, with that said, in lieu of embedding a full v8 interpreter,
+one can use a project like [[https://github.com/vercel/pkg][pkg]] to produce self-contained quasi-native
+applications.  This could be done with WASM, but with WASM the
+infrastructure to make use of this binary would not be present. The
+JVM compatibility is fixed by adding a Java runtime environment, which
+can be done once, and used directly.  As an addded benefit there are
```
0x009922, line 507:
```suggestion
can be done once, and used directly.  As an added benefit there are
```

docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,571 @@
...
+functions in the standard c library, there is no maintenance to speak
+of.
+
+To summarise, a native platform library is by almost all criteria
+either a direct upgrade, or a minor improvement. The one instance in
+which there is significantly more work to be done is the question of
+confinement.  What can generously be called a minor inconvenience is
+the fact that running native libraries on heterogeneous networks,
+utilising multiple architectures has to come with a separate emulation
+layer.  Experience with running =x86= containers on Apple silicone
```
DCNick3, line 529: Using fat binaries (or some other way to have the same code compiled for different architectures) may be a solution, though it makes the size and auditability even worse

docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,954 @@
...
+concept, it is not used to limit execution, but instead the even
+parity substrate decides weights via a benchmarking process.
+
+The cons are equally underdeveloped.
+
+1. Requires nightly* for the smart contracts themselves
+2. Difficult to limit memory further than 3GB in =wasmtime= (needs more
+   research)
+3. No Interface Types Support - so it has to be simulated through
+   passing serialized bytes or mapping memory
```
DCNick3, line 134: Btw, I think there _is_ support for interface types (now called wasm [Component Model](https://github.com/WebAssembly/component-model)). One project I know of using it is [Ambient](https://github.com/AmbientRun/Ambient/tree/main/crates/wasm)

docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,954 @@
...
+benefit of the dynamic linkage already.  However most objects in the
+Iroha data model are defined as Rust structures, so interactions with
+them must be done via either intermediate representations, or dynamic
+linkage. As such dynamic linkage is a must.  Fortunately, it is easier
+to do that with Java than it is with Rust, even though this is at the
+cost of performance, most dispatch in Java is dynamic by design, thus
+making the question of handling object lifetimes inside libraries much
+easier.  Dynamic linkage of some form, thus becomes a necessity.
+
+Java is opaque: it was designed for the open-closed principle, where
```
DCNick3, line 751: > Java is opaque

I would argue that JVM code (if not obfuscated) is much easier to inspect than x86/wasm, as the compilation preserves more structure of the program (along with class/function names!). A lot of times, a good decompiler can easily recover the equivalent java source code

docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,954 @@
...
+
+Haskell can produce some of the smallest binaries owing to a
+sophisticated runtime that must be present, and provided it uses
+dynamic linkage.  Default static linkage even for small programs
+produces binaries comparable in size to WASM.  It can also be stored
+as source code, and compiled in place.  Owing to ease of composition
+and a terse syntax Haskell is used in code golfing.  Haskell does not
+have as many complex invariants to uphold as does Rust, thus it can do
+dynamic linkage much more simply than Rust.  However, linking
+pro-actively, which is to say =dlopen= comes [[Haskell. in
```
DCNick3, line 859: Broken link?

docs/source/rfc/on-chain-scripting.org
```diff
@@ -0,0 +1,954 @@
...
+statement and in all situations the result of the evaluation will be
+completely unchanged.  Identifying bugs in a functional codebase is
+much easier.
+
+Sand boxing is one of the key areas where Haskell shines.  All Haskell
+programs are purely functional (unlike e.g. OCaml that is functional
+by default, but retains an imperative subset).  Any deviations must be
+carried through the type system, and unlike =unsafe= in Rust, which
+can occur silently inside of functions, functions which can cause
+input and output to happen, must carry the information through the
```
DCNick3, line 888: I don't think that's true, you can do crimes in Haskell too. For example, [`unsafePerformIO`](https://hackage.haskell.org/package/base-4.18.0.0/docs/System-IO-Unsafe.html#v:unsafePerformIO) takes an `IO a` (which represents some operation that produces `a` with side-effects) and gives you `a`, with all the side-effects

I do agree however, that it is much easier to ensure that some code has no side-effects/doesn't depend on environment/uses no random

docs/source/rfc/on-chain-scripting.org
appetrosyan: Fixed.

docs/source/rfc/on-chain-scripting.org
appetrosyan: fn::7

6r1d (approved):
Security-wise, Haskell would be the best choice: it'll be easiest to make it behave predictably.

While it will require adding some infrastructural parts, the valuable advantage is controlling access to randomness and side effects. Being able to both interpret and compile the source provides an ability to upload the source, having a warranty of blockchain events being equal to the ones we'd expect from the smart contracts, which may be necessary for legal applications.

In addition to that, we've discussed SPIR-v quite a bit. It'll allow us to get an excellent performance, but integrating it and getting the safety guarantees Haskell provides would be more complex.

Lua is another interesting candidate since the performance means more in executor than a smart contract. Using a simple interpreted language like Lua for the task would also contribute to the transparency of the executor, contributing to the smart contract security.

6r1d:
Moved to https://github.com/hyperledger/iroha-rfcs/pull/5

PR #3854, opened 2023-08-31
[refactor] #2573: Use a more efficient representation for immutable bytes values

### LINKED ISSUE #2573: Immutable payload in `iroha_crypto`
Check [this](https://github.com/hyperledger/iroha/pull/2557#discussion_r934203281) comment. Payload (currently a `Vec<u8>`) in structures like `PublicKey`, `PrivateKey` and `Signature` is immutable for the lifetime of the wrapping type. This means that capacity field of the `Vec` is redundant and could be removed to reduce the size. A similar thing was done in #1985

DCNick3:
## Description

Introduce an `ImmutableBytes` type that is a newtype around `Box<[u8]>` and use it in `PublicKey`, `PrivateKey`, `Signature` and `Multihash` objects.

While `PublicKey` and `PrivateKey` can be further specialized, as their size can be known without storing separately, this is not true for `Signature`, as it can take any user-provided slice as its payload. `Multihash` can also be specialized, but it's used as a temporary representation, so it is not really helpful in reducing resident memory consumption.

### Linked issue

Closes #2573

### Benefits

- This uses 8 bytes less memory for every `PublicKey`, `PrivateKey`, `Signature`and `Multihash` objects


crypto/src/lib.rs
```diff
@@ -918,15 +925,19 @@ mod tests {
...
+                        )
+                        .expect("Failed to decode public key.")
                     )
-                    .expect("Failed to decode public key.")
                 },
                 private_key: PrivateKey {
                     digest_function: Algorithm::BlsSmall,
-                    payload: hex_decode("0000000000000000000000000000000060F3C1AC9ADDBBED8DB83BC1B2EF22139FB049EECB723A557A41CA1A4B1FED63")
-                    .expect("Failed to decode private key"),
+                    payload: ImmutableBytes::new(
```
0x009922, line 937: This pattern is repeated in tests a lot:

```rs
ImmutableBytes::new(hex_decode("deadbeef").expect("failed"));
```

Maybe, define a test-only method for it?

```rs
impl ImmutableBytes {
    #[cfg(test)]
    fn from_hex(hex: AsRef<str>) -> Self {
        ...
    }
}
```

0x009922 (dismissed):

crypto/src/lib.rs
DCNick3: Made it into a regular fallible method, as it had a use outside of tests

crypto/src/immutable_bytes.rs
```diff
@@ -0,0 +1,90 @@
...
+#[cfg(feature = "std")]
+use std::{boxed::Box, vec::Vec};
+
+use iroha_schema::IntoSchema;
+use parity_scale_codec::{Compact, Decode, Encode};
+use serde::{Deserialize, Serialize};
+
+/// Stores bytes that are not supposed to change during the runtime of the program
+///
+/// This is a more efficient than `Vec<u8>` because it does not have to store the capacity field
```
pesterev, line 13: This sounds like a compact pointer around `[u8]`, not an immutable. This pointer is good for keeping something in the storage compactly, but if we consider `Arc<[u8]>` as an immutable data structure, that may be more efficient because it has cheap copies.
I'd rename it to `CompactBytes` or something like that IMHO.

DCNick3: Done

crypto/src/immutable_bytes.rs
mversic, line 1: I would name it `ConstVec` and move it to `primitives` crate like we did for `ConstString`

DCNick3: Hm, definitely makes sense to have in primitives. The naming consistency is nice too

DCNick3: Although, it's not really a vec currently, as it can only hold bytes

0x009922 (dismissed):
LGTM. Although, I like the idea of moving it into the `primitives` crate first before merging.

crypto/src/immutable_bytes.rs
```diff
@@ -0,0 +1,90 @@
...
+// The encoded representation is the same as &[u8] or Vec<u8>
+impl Encode for CompactBytes {
+    fn size_hint(&self) -> usize {
+        mem::size_of::<u32>() + self.len()
+    }
+
+    fn encode_to<T: parity_scale_codec::Output + ?Sized>(&self, dest: &mut T) {
+        let len = u32::try_from(self.len())
+            .expect("Attempted to serialize an ImmutableBytes that is too large.");
+        let len = Compact(len);
```
mversic, line 54: I think that if you don't specify in the schema that you're using compact encoding it will break JDKs. Actually, if you look at `ConstString`, I think you can delegate encoding to `Vec<u8>` like this:

```rs
impl WrapperTypeEncode for CompactBytes {}

impl WrapperTypeDecode for CompactBytes {
    type Wrapped = Vec<u8>;
}
```

crypto/src/immutable_bytes.rs
```diff
@@ -0,0 +1,90 @@
...
+impl Decode for CompactBytes {
+    fn decode<I: parity_scale_codec::Input>(
+        input: &mut I,
+    ) -> Result<Self, parity_scale_codec::Error> {
+        let Compact(length) = <Compact<u32>>::decode(input)?;
+        let mut out = vec![0u8; length as usize];
+        input.read(&mut out)?;
+        Ok(Self::new(out))
+    }
+}
```
mversic, line 70: I think that if you don't specify in the schema that you're using compact encoding it will break JDKs. Actually, if you look at `ConstString`, I think you can delegate encoding to `Vec<u8>` like this:

```rs
impl WrapperTypeEncode for CompactBytes {}

impl WrapperTypeDecode for CompactBytes {
    type Wrapped = Vec<u8>;
}
```

crypto/src/immutable_bytes.rs
mversic: > Although, it's not really a vec currently, as it can only hold bytes

maybe `ConstBytesVec`?

crypto/src/immutable_bytes.rs
DCNick3: Curiously, the same can be said about the `ConstString`: it doesn't have cheap copies, but its name can suggest this..

crypto/src/immutable_bytes.rs
DCNick3: Didn't notice that such a mechanism existed. Using `WrapperTypeEncode`/`WrapperTypeDecode` now

crypto/src/immutable_bytes.rs
DCNick3: Decided on `ConstBytes`, which is a bit shorter

0x009922 (dismissed):

crypto/src/immutable_bytes.rs
pesterev: I think these are not really suitable names for these data structures, and a bit confusing when reading them. But it's just my opinion, and I guess good documentation eliminates this problem at all.

pesterev (dismissed):

mversic (changes requested):

primitives/src/const_vec.rs
```diff
@@ -0,0 +1,99 @@
...
+use alloc::{boxed::Box, format, string::String, vec::Vec};
+use core::ops::Deref;
+#[cfg(feature = "std")]
+use std::{boxed::Box, vec::Vec};
+
+use iroha_schema::{IntoSchema, MetaMap, Metadata, TypeId, VecMeta};
+use parity_scale_codec::{WrapperTypeDecode, WrapperTypeEncode};
+use serde::{Deserialize, Serialize};
+
+/// Stores bytes that are not supposed to change during the runtime of the program in a compact way
```
mversic, line 12: well technically, it could possible to mutate the `ConstVec`, just not to resize it. But we don't need that

mversic (approved):

0x009922 (approved):

Issue #2851, opened 2022-10-10
A number of tests had to be ignored


6r1d:
Since #2848, a number of slow tests were ignored explicitly, because those tests were too slow, breaking the CI.
At some point, the functionality of such tests has to be restored.

appetrosyan:
Need to restore tests, and find a way to trigger `pytest` via `cargo`

PR #3855, opened 2023-08-31
[fix] #3853: Compare permission token payload as JSON

### LINKED ISSUE #3853: [BUG] )Permission token payloads comparison fails
Since we store payload as a JSON encoded string we should also compare it as JSON but it is compared as a string. Otherwise a single whitespace can break the comparison

mversic:
## Description

1. `StringWithJson` should implement `PartialEq` comparison as JSON
this was causing bugs when `iroha-java` would serialize token payload vs when `iroha` `serde` would do so 
2. remove `DefaultValidator` struct from `iroha_validator`
it's been used incorrectly, because the delegated method must always call back into original validator whereas calling into `DefaultValidator` would prevent this. Interestingly, removing `DefaultValidator` didn't increase code duplication which was the argument for introducing it in the first place
3. removed `DoesAccountHavePermissionToken`
there is little reason to think it measurably optimizes the operation of finding a token for an account especially since it's only executed from the validator
4. discovered #3857

### Linked issue

Closes #3853 

### Checklist


client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -39,22 +41,29 @@ mod token {
     use super::*;
 
     /// Token to identify if user can (un-)register domains.
-    #[derive(Token, ValidateGrantRevoke, Decode, Encode, IntoSchema, Serialize, Deserialize)]
+    #[derive(
+        PartialEq, Token, ValidateGrantRevoke, Decode, Encode, IntoSchema, Serialize, Deserialize,
```
appetrosyan: Why not eq?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -152,8 +161,16 @@ impl CustomValidator {
     }
 }
 
-impl Visit for CustomValidator {
-    fn visit_register_domain(&mut self, authority: &AccountId, _register_domain: Register<Domain>) {
+macro_rules! defaults {
```
appetrosyan, line 172: This macro seems to be use more than once. Maybe it can later be extracted into a top level module?

data_model/src/permission.rs
```diff
@@ -127,6 +142,12 @@ impl StringWithJson {
         Self(payload.to_string())
     }
 }
+impl PartialEq for StringWithJson {
+    fn eq(&self, other: &Self) -> bool {
+        serde_json::from_str::<serde_json::Value>(&self.0).unwrap()
```
appetrosyan, line 147: Can this become an unwrap unchecked?

default_validator/src/lib.rs
```diff
@@ -3,14 +3,164 @@
...
+        Self {
+            verdict: Ok(()),
+            block_height,
+            host: iroha_wasm::Host,
+        }
+    }
+
+    fn ensure_genesis(block_height: u64) -> MigrationResult {
+        if block_height != 0 {
+            return Err("Default Validator is intended to be used only in genesis. \
```
appetrosyan, line 41: As a slight recommendation, o'd wager this doesn't need to be represented as a string, but rather there ought to be an enum for this

appetrosyan (dismissed):

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
mversic: added

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
mversic: there is #3856

default_validator/src/lib.rs
mversic: I'm not entirely convinced because this is just an example validator. And there is no way of enforcing that every 1st validator will implement this check

appetrosyan (approved):

default_validator/src/lib.rs
appetrosyan: It doesn't need to.just like not every http response is not emitted by every server

client/tests/integration/permissions.rs
```diff
@@ -293,3 +285,54 @@ fn permissions_differ_not_only_by_names() {
...
+            // NOTE: Introduced additional whitespaces in the serialized form
+            "{ \"asset_id\" : \"xor#wonderland#mouse@wonderland\" }",
+        ),
+        alice_id,
+    );
+
+    let transaction = TransactionBuilder::new(mouse_id)
+        .with_instructions([allow_alice_to_set_key_value_in_mouse_asset])
+        .sign(mouse_keypair)
+        .expect("Failed to sign mouse transaction");
```
0x009922, line 325: Not a suitable message here.

```suggestion
        .expect("Should sign mouse transaction fine");
```

```suggestion
        .unwrap();
```

client/tests/integration/permissions.rs
```diff
@@ -293,3 +285,54 @@ fn permissions_differ_not_only_by_names() {
...
+    iroha_client
+        .submit_transaction_blocking(&transaction)
+        .expect("Failed to grant permission to alice.");
```
0x009922, line 328:
```suggestion
        .unwrap();
```

client/tests/integration/permissions.rs
```diff
@@ -293,3 +285,54 @@ fn permissions_differ_not_only_by_names() {
...
+
+    // Check that alice can indeed mint mouse asset
+    let set_key_value =
+        SetKeyValueBox::new(mouse_asset, Name::from_str("color")?, "red".to_owned());
+    iroha_client
+        .submit_blocking(set_key_value)
+        .expect("Failed to mint asset for mouse.");
```
0x009922, line 335:
```suggestion
        .unwrap();
```

0x009922 (approved):
I am not familiar with the changes well. On the shallow look, LGTM. Since this PR is blocking us from `rc19` release, I am approving this with fingers crossed.

client/tests/integration/permissions.rs
```diff
@@ -61,127 +59,124 @@ fn genesis_transactions_are_validated() {
...
     iroha_client
         .request(client::asset::by_account_id(id.clone()))
         .expect("Failed to execute request.")
         .collect::<QueryResult<Vec<_>>>()
         .expect("Failed to execute request.")
 }
 
-#[ignore = "ignore, more in #2851"]
 #[test]
+#[ignore = "ignore, more in #2851"]
```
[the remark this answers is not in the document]
mversic: I tried, but it's broken. It was broken before this PR so I think it should be resolved in the scope of #2851

default_validator/src/lib.rs
mversic: created #3860

PR #3869, opened 2023-09-06
[refactor] #3526: Allow only a fixed set of signature verification conditions

### LINKED ISSUE #3526: [suggestion] Offer only predefined set of signature check conditions
Related to #3088

The issue is described in the previous ticket. We have to make sure that our signature check conditions are valid and don't suffer from halting problem

DCNick3:
## Description

Instead of allowing arbitrary iroha expressions in `SignatureVerificationCondition`, allow only a fixed set of verification conditions

### Linked issue

Closes #3526

### Checklist

- [x] Decide on the set of signature check conditions we want to allow
- [x] Add unit tests
- [x] Implement without using expressions


data_model/src/account.rs
```diff
@@ -148,24 +147,28 @@ pub mod model {
...
-    pub struct SignatureCheckCondition(pub EvaluatesTo<bool>);
+    #[ffi_type(opaque)]
+    #[allow(clippy::enum_variant_names)]
+    pub enum SignatureCheckCondition {
+        #[default]
+        AnyAccountSignatures,
+        AllAccountSignatures,
+        #[display(fmt = "[{_0:?}]")]
+        AllCustomSignatures(Vec<PublicKey>),
+    }
```
mversic, line 171: we should have a discussion which variants the `SignatureConditionCheck` enum should contain. @Erigara @Arjentix your thoughs?

data_model/src/account.rs
```diff
@@ -148,24 +147,28 @@ pub mod model {
...
-    #[ffi_type(unsafe {robust})]
-    pub struct SignatureCheckCondition(pub EvaluatesTo<bool>);
+    #[ffi_type(opaque)]
+    #[allow(clippy::enum_variant_names)]
+    pub enum SignatureCheckCondition {
+        #[default]
+        AnyAccountSignatures,
+        AllAccountSignatures,
+        #[display(fmt = "[{_0:?}]")]
+        AllCustomSignatures(Vec<PublicKey>),
```
mversic, line 170: I believe this `Vec` is immutable. Could we expand #2573 to support other immutable `Vec` types not only bytes?

data_model/src/account.rs
Erigara: IMO we should have generalized form `N` out of `M` signatures (`AnyAccountSignatures` is `1` out of `M`, `AllAccountSignatures` is `M` out of `M`).

Also idk how useful `AllCustomSignatures` would be since set of pubic keys could be adjusted (through `Mint` and `Burn` and then `M` out of `M` is used).

Other than that i not sure which conditions we will need. 

Should be have `WASM` script for signature check condition (with access to transaction)?

data_model/src/account.rs
DCNick3: Yeah, I think so

DCNick3: I've pushed these changes to #3854 that generalize this. Would either need to wait for it to merge or use some git-fu..

data_model/src/account.rs
DCNick3: What I don't like in the "N out of M` model is that for account signatures the `M` can be changed, potentially making the previously set `SignatureCheckCondition` not what the user intended

mversic: > What I don't like in the "N out of Mmodel is that for account signatures theMcan be changed, potentially making the previously setSignatureCheckCondition` not what the user intended

this makes a good point. What happens to the `SignatureCheckCondition` when a key is removed from `Account::signatories`. Should the change be reflected?

DCNick3: Sounds quite reasonable

DCNick3: Hm, currently the default is to allow the tx to pass if _any_ of the keys from account signatories are used, not _all_. Removing the _any_ option breaks some tests

mversic: I would consider that a bug. Which tests are that?

DCNick3: The `transaction_signed_by_new_signatory_of_account_should_pass` explicitly adds a new key to the signatory and tries to sign a transaction with it. 

There are other failed tests, but the failure may be transient, haven't looked at them deeper

core/src/queue.rs
```diff
@@ -53,25 +52,21 @@ impl AcceptedTransaction {
...
+        SignatureCheckCondition::AnyAccountSignatures => account_signatories
+            .iter()
+            .any(|transaction_signatory| transaction_signatories.contains(transaction_signatory)),
+        SignatureCheckCondition::AllAccountSignatures => account_signatories
+            .iter()
+            .all(|transaction_signatory| transaction_signatories.contains(transaction_signatory)),
+        SignatureCheckCondition::AllCustomSignatures(required_signatures) => required_signatures
+            .iter()
+            .all(|required_signature| transaction_signatories.contains(required_signature)),
+    }
```
Erigara, line 69: Maybe this logic should be method of `SignatureCheckCondition` itself?

DCNick3: Done

data_model/src/account.rs
mversic: except that I would add custom signatures on top of `AnyAccountSignatures` and `AllAccountSignatures` not as an alternative

DCNick3: Currently, I've implemented the following conditions:

1. require any of the `Account::signatories`
2. require all of the `Account::signatories`
3. require all of the `Account::signatories` + additional keys

DCNick3: > Because IMHO we'll always want some of account signatories to be included

That's fair. I also think we can simplify by having only the version with the additional keys. With the empty additional keys list the logic would be equivalent. Also, the empty boxed slice [doesn't allocate](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=add8ce577f7c50575b0e1e3763249c79), so the perf should be the same

Erigara: > we don't need N out of M

I don't insist on N out of M. I agree that change in M is valid concern. I think main usecase for `N out of M` is some kind of majority voting (majority, supermajority) which we can add later.

data_model/src/account.rs
```diff
@@ -154,18 +142,20 @@ pub mod model {
...
-    #[ffi_type(unsafe {robust})]
-    pub struct SignatureCheckCondition(pub EvaluatesTo<bool>);
+    #[ffi_type(opaque)]
+    #[allow(clippy::enum_variant_names)]
+    pub enum SignatureCheckCondition {
+        #[display(fmt = "AnyAccountSignatureOr({_0:?})")]
+        AnyAccountSignatureOr(ConstVec<PublicKey>),
+        #[display(fmt = "AllAccountSignaturesAnd({_0:?})")]
+        AllAccountSignaturesAnd(ConstVec<PublicKey>),
+    }
```
mversic, line 158: neat solution

Erigara (dismissed):
LGTM

data_model/src/account.rs
```diff
@@ -345,7 +314,148 @@ impl FromStr for AccountId {
     }
 }
 
+impl Default for SignatureCheckCondition {
+    fn default() -> Self {
+        Self::AnyAccountSignatureOr(ConstVec::new_empty())
+    }
+}
```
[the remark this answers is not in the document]
DCNick3: No, `Default` for enums works [only on unit variants](https://github.com/rust-lang/rust/blob/c38b8a8c621e0c946af1b74f57bb8cc028e0060d/src/test/ui/macros/macros-nonfatal-errors.rs#L86C64-L86C64) marked with `#[default]`.

data_model/src/account.rs
```diff
@@ -345,7 +314,148 @@ impl FromStr for AccountId {
...
+
+impl SignatureCheckCondition {
+    /// Shorthand to create a [`SignatureCheckCondition::AnyAccountSignatureOr`] variant without additional allowed signatures.
+    #[inline]
+    pub fn any_account_signature() -> Self {
+        Self::AnyAccountSignatureOr(ConstVec::new_empty())
+    }
+
+    /// Shorthand to create a [`SignatureCheckCondition::AllAccountSignaturesAnd`] variant without additional allowed signatures.
```
0x009922, line 330:
```suggestion
    /// Shorthand to create a [`SignatureCheckCondition::AllAccountSignaturesAnd`] variant without additional required signatures.
```

0x009922 (approved):

data_model/src/account.rs
```diff
@@ -345,7 +314,148 @@ impl FromStr for AccountId {
...
+
+        check_signature_check_condition(&condition, &[], &[], false);
+        check_signature_check_condition(&condition, &[&key1], &[], false);
+        check_signature_check_condition(&condition, &[], &[&key1], false);
+        check_signature_check_condition(&condition, &[&key1], &[&key1], true);
+        check_signature_check_condition(&condition, &[&key1], &[&key2], false);
+        check_signature_check_condition(&condition, &[&key1, &key2, &key3], &[&key1], true);
+        check_signature_check_condition(&condition, &[&key1, &key2, &key3], &[&key2], true);
+        check_signature_check_condition(&condition, &[&key1, &key2, &key3], &[&key3], true);
+    }
```
0x009922, line 406: Hmmm, I am not proficient in that topic, but couldn't it be covered with `proptest`s?

0x009922 (approved):

Erigara (approved):

data_model/src/account.rs
DCNick3: I think it __might__ be useful.. Though, tbh, for such a simple implementation they might be overkill.

Also, the (possibly) unexpected behaviour of `All` mode with the empty set of account keys would be hard to discover (even though it's probably not possible to make an account w/o keys :sweat_smile:)

PR #3872, opened 2023-09-06
[feature] #3868: Add iroha_swarm disclaimer header

### LINKED ISSUE #3868: [suggestion] Add `this file is autogenerated` into `iroha_swarm` output
Before #3864 I had no idea that our `docker-compose.dev.*` files are generated with `iroha_swarm`.

I suggest to add a comment like:

```yml
# This file is autogenerated with `iroha_swarm <used options>`
```

@0x009922 , what do you think?

Arjentix:
## Description

Added disclaimer about files being generated by `iroha_swarm`

### Linked issue

- Closes #3868 

### Benefits

Now it's obvious that files are generated by a program and not written by hand

### Checklist


DCNick3:
Why shouldn't it be the default? Is the user supposed to edit it in some cases or smth?

docker-compose.dev.local.yml
```diff
@@ -1,3 +1,6 @@
+# This file is generated by `iroha_swarm -p 4 -s Iroha --disclaimer-header --force --config-dir ./configs/peer --build . --outfile docker-compose.dev.local.yml`.
```
0x009922, line 1: It seems that this banner tries to put all the parameters so it is self-contained and reproducible only using them. However, the output of `iroha_swarm` also depends on CWD.

I think it is not necessary to put ALL of the parameters here. Most of them you can infer on your own. The only parameter that is impossible (or hard?) to infer is `--seed`.

tools/swarm/src/cli.rs
```diff
@@ -20,6 +20,9 @@ pub struct Cli {
     /// overwrite the file anyway, pass `--force` flag.
     #[arg(long, short)]
     pub outfile: PathBuf,
+    /// Print a disclaimer header at the top of the generated file.
+    #[arg(long)]
+    pub disclaimer_header: bool,
```
0x009922, line 25:
```suggestion
    /// Put a banner in the file saying that the file is generated. It includes the cryptography seed in order to help with reproducibility.
    #[arg(long)]
    pub banner: bool,
```

Or `--no-banner`?

0x009922 (changes requested):

docker-compose.dev.local.yml
Arjentix: I think it's obvious that this command was executed from the iroha main directory. I'd like to keep all parameters listed because it's hard to infer them. I either have to:

1. Read carefully the whole file, count peers, check if they use image or local build and etc. This is not comfortable and I have to remember all existing parameters, which might be even harder in future if we'll have more of them.
2. Read message from the CI, which I'll now very lately

tools/swarm/src/cli.rs
```diff
@@ -20,6 +20,11 @@ pub struct Cli {
     /// overwrite the file anyway, pass `--force` flag.
     #[arg(long, short)]
     pub outfile: PathBuf,
+    /// Put a banner in the file saying that the file is generated.
```
DCNick3, line 23: The doc-comment should probably be updated to reflect the inverted logic of the flag

docker-compose.dev.local.yml
0x009922: > I think it's obvious that this command was executed from the iroha main directory

It is the case for _our_ docker-compose files, but it is not a general purpose of `iroha_swarm`. This tool is standalone and might be executed from anywhere.

Indeed, the parameters are not easy to infer. Still, I am personally lean against putting all of the parameters into the banner, because CWD **must** be included for completeness, and it might look _dirty_. IMO, the minimal purpose of the banner is only to notice that the file is generated and not hand-written, therefore the reader should probably do some research on their own before editing it by hand.

I wouldn't say I insist on removing the parameters at all, but if we put parameters, CWD should be considered too.

DCNick3: Including CWD will lead to reproducibility troubles, especially with how we check in CI that the `yaml` file is up-to-date

Arjentix: > Including CWD will lead to reproducibility troubles, especially with how we check in CI that the yaml file is up-to-date

Actually not since I've already modified how we check them in CI -- it skips first 2 lines now.

But looks like I have to rollback it, fine

0x009922 (approved):

DCNick3 (approved):

PR #3873, opened 2023-09-07
[feature] #1915: Implement Fast kura init mode

### LINKED ISSUE #1915: Implement missing `kura` init modes.

Erigara:
## Description

Implement `fast` kura init mode.
Which can skip loading blocks from disk and thus bootstrap startup time. 

### Linked issue

Closes #1915 

### Benefits

Improve startup time.

### Downsides

Take additional space for storing hashes on disk (32 bytes per hash) time/memory tradeoff.

`fast` init mode should be used with caution because it can detect some mutations in hashes file. 

### How to test

1. Run iroha with `strict` kura init mode
2. Create couple of blocks
3. Check that `storage/blocks.hashes` is created
4. Shutdown iroha peer
5. Switch to `fast` kura init mode in the configuration
6. Start peer again
7. Check that peer is running correctly
8. Shutdown iroha peer ones more
9. Make backup of `storage/blocks.hashes` (`storage/blocks.hahses.old`)
10. Remove some data from data from `storage/blocks.hashes` or whole file entirely
11. Start peer
12. Check that kura failed to load in `fast` mode and failed back to STRICT mode (grep for “Hashes file is broken”)
13. Check that peer is running correctly
14. Check that `storage/blocks.hashes` is restored (`storage/blocks.hashes` == `storage/blocks.hashes.old`)
15. Check that peer is running correctly

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -6,13 +6,13 @@
 use std::{
     fmt::Debug,
     fs,
-    io::{Read, Seek, SeekFrom, Write},
+    io::{BufWriter, Read, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
     sync::Arc,
 };
 
 use iroha_config::kura::Mode;
-use iroha_crypto::HashOf;
+use iroha_crypto::{Hash, HashOf};
 use iroha_data_model::block::VersionedCommittedBlock;
 use iroha_logger::prelude::*;
 use iroha_version::scale::{DecodeVersioned, EncodeVersioned};
@@ -22,7 +22,10 @@ use crate::handler::ThreadHandler;
 
 const INDEX_FILE_NAME: &str = "blocks.index";
 const DATA_FILE_NAME: &str = "blocks.data";
+const HASHES_FILE_NAME: &str = "blocks.hashes";
 const LOCK_FILE_NAME: &str = "kura.lock";
+
+const SIZE_OF_BLOCK_HASH: u64 = std::mem::size_of::<HashOf<VersionedCommittedBlock>>() as u64;
 
 /// The interface of Kura subsystem
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


core/src/kura.rs
```diff
@@ -651,7 +829,8 @@ impl BlockStore {
     /// # Errors
     /// Fails if any of the required platform-specific functions
     /// fail.
-    pub fn append_block_to_chain(&mut self, block_data: &[u8]) -> Result<()> {
+    pub fn append_block_to_chain(&mut self, block: &VersionedCommittedBlock) -> Result<()> {
```
DCNick3, line 832: Stronger typing is nice!

DCNick3 (dismissed):

core/src/kura.rs
```diff
@@ -113,38 +116,81 @@ impl Kura {
...
                     }
-                },
-                Err(error) => {
-                    error!(?error, "Malformed block index or corrupted block data file. Not reading any blocks beyond this height.");
-                    break;
                 }
             }
+
+            block_store.overwrite_block_hashes(&block_hashes)?;
         }
```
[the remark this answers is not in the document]
Erigara: Good suggestion, indeed much more readable + allow to write fast init mode in imperative style.

core/src/kura.rs
```diff
@@ -498,6 +542,73 @@ impl BlockStore {
         // Each entry is 16 bytes.
     }
 
+    /// Read a series of block hashes from the block hashes file and
+    /// attempt to fill all of `dest_buffer`.
```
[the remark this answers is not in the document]
Erigara: Leftover.

core/src/kura.rs
```diff
@@ -498,6 +542,73 @@ impl BlockStore {
...
+    ///
+    /// # Errors
+    /// IO Error.
+    pub fn read_block_hashes(
```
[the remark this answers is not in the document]
Erigara: No, we call this function from `Kura`.

core/src/kura.rs
```diff
@@ -498,6 +542,73 @@ impl BlockStore {
...
+        let mut hashes = Vec::with_capacity(block_count);
+        let mut buffer = [0; Hash::LENGTH];
+        for _ in 0..block_count {
+            hashes_file
+                .read_exact(&mut buffer)
+                .map_err(|e| Error::IO(e, path.clone()))?;
+            let hash = HashOf::from_untyped_unchecked(Hash::prehashed(buffer));
+            hashes.push(hash);
+        }
+        Ok(hashes)
```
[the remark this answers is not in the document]
Erigara: Your suggestion reminded me that we can use `collect` to create `Result<Vec<T>, Error>`, so thanks.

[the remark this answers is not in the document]
Arjentix: Wow, I forgot about that too, nice feature

core/src/kura.rs
```diff
@@ -620,6 +728,67 @@ impl BlockStore {
...
+        let mut hashes_file = std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .open(path.clone())
+            .map_err(|e| Error::IO(e, path.clone()))?;
+        let start_location = block_height * SIZE_OF_BLOCK_HASH;
+        if start_location + SIZE_OF_BLOCK_HASH
+            > hashes_file
+                .metadata()
+                .map_err(|e| Error::IO(e, path.clone()))?
```
Arjentix, line 752: I'm not sure my idea makes sense in this case, but I can see this pattern repeated a lot.
So just an idea, maybe for future things:

```rust
trait AddIoContextExt<T> {
    fn add_io_context(self, path: Path) -> Result<T, Error>;
}

impl<T> AddIoContext<T> for Result<T, std::io::Error> {
    fn add_io_context(self, path: Path) -> Result<T, Error> {
        self.map_err(|e| Error::IO(e, path))
    }
}


let len = hashes_file
    .metadata()
    .add_io_context(path.clone())?;

```

It does not reduce the lines of code but to me looks simplier

Arjentix (approved):

core/src/kura.rs
Arjentix: Or maybe even better:

```rust
trait AddErrContextExt<T> {
    type Context;

    fn add_err_context(self, context: Context) -> Result<T, Error>;
}

impl<T> AddErrContext<T> for Result<T, std::io::Error> {
    type Context = Path;

    fn add_err_context(self, path: Self::Context) -> Result<T, Error> {
        self.map_err(|e| Error::IO(e, path))
    }
}
```

So that it becomed usefull not only for `Error::IO`

Arjentix: I probaly just took the idea from [anyhow's `with_context()`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.with_context)

core/src/kura.rs
```diff
@@ -694,6 +844,22 @@ pub enum Error {
     Locked(PathBuf),
     /// Conversion of wide integer into narrow integer failed. This error cannot be caught at compile time at present
     IntConversion(#[from] std::num::TryFromIntError),
+    /// Blocks count differs hashes file and index file
+    HashesFileHeightMismatch,
+}
+
+trait AddErrContextExt<T> {
```
DCNick3, line 851: This feels like re-implementing [`snafu`](https://docs.rs/snafu/latest/snafu/).. Probably fine for one file, be we should consider using `snafu`..

DCNick3 (approved):

core/src/kura.rs
Erigara: Good point, let's consider this refactoring if such pattern will appear in other places too

Arjentix: What I don't understand about snafu is that it adds `Snafu` suffix to types. Why? It looks ugly to me

Arjentix (approved):

core/src/kura.rs
DCNick3: `Snafu` suffix is for context helpers - all values for your error variant, excluding the error itself. Instead of writing `res.map_err(|e| Error::IoError { error: e, path })` you would write something akin to `res.context(|| IoSnafu { path })`.

It also obviously can derive `From` if your error variant does not need any context. Or you can use the `Snafu` struct to disambiguate between two variants

Issue #3878, opened 2023-09-08
📖 [documentation] Add the documentation-related Iroha configurations


6r1d:
We're actively using the "[Code snippets](https://hyperledger.github.io/iroha-2-docs/documenting/snippets.html)" feature, and I'm trying to have more.
While doing https://github.com/hyperledger/iroha-2-docs/pull/394, I want to reach some granularity.

I can either have `Cargo.toml` for the "[Client setup](https://hyperledger.github.io/iroha-2-docs/guide/rust.html#_1-iroha-2-client-setup)" section of the [Rust tutorial](https://hyperledger.github.io/iroha-2-docs/guide/rust.html) here or add it to the snippets directory in Iroha 2 documentation. It has less chance of remaining unnoticed for a long time if it belongs to the main repo, at least in my opinion.

Collecting a list of doc-related configurations that depend on the release in the main repo seems OK to me, but someone may disagree.

Please comment here or in https://github.com/hyperledger/iroha/pull/3879.

mversic:
should be tracked in iroha-documentation repo

PR #3870, opened 2023-09-06
[docs]: #391 move the Rust examples

6r1d:
## Description

This is a draft PR that moves code from the blocks in `iroha-2-docs` repo.

### Linked issue

Closes #391

### Benefits

Docs will be more clean and the doctests will be more detailed.

### Checklist

- [x] I've read `CONTRIBUTING.md`
- [x] I've used the standard signed-off commit format (or will squash just before merging)
- [x] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

PR #3879, opened 2023-09-08
[documentation] #3878: Add the documentation-related Iroha configurat…

6r1d:
Original issue: https://github.com/hyperledger/iroha/issues/3878

We're actively using the "[Code snippets](https://hyperledger.github.io/iroha-2-docs/documenting/snippets.html)" feature, and I'm trying to have more.
While doing https://github.com/hyperledger/iroha-2-docs/pull/394, I want to reach some granularity.

I can either have `Cargo.toml` for the "[Client setup](https://hyperledger.github.io/iroha-2-docs/guide/rust.html#_1-iroha-2-client-setup)" section of the [Rust tutorial](https://hyperledger.github.io/iroha-2-docs/guide/rust.html) here or add it to the snippets directory in Iroha 2 documentation. It has less chance of remaining unnoticed for a long time if it belongs to the main repo, at least in my opinion.

Collecting a list of doc-related configurations that depend on the release in the main repo seems OK to me, but someone may disagree.

Please comment here.

### DIFF
--- a/configs/documentation/Cargo.rc19.toml
+++ b/configs/documentation/Cargo.rc19.toml
@@ -1,0 +1,5 @@
+[dependencies]
+iroha_client = { version = "=2.0.0-pre-rc.19", path = "~/Git/iroha/client" }
+iroha_data_model = { version = "=2.0.0-pre-rc.19", path = "~/Git/iroha/data_model" }
+iroha_crypto = { version = "=2.0.0-pre-rc.19", path = "~/Git/iroha/crypto" }
+iroha_config = { version = "=2.0.0-pre-rc.19", path = "~/Git/iroha/config" }
--- a/configs/documentation/README.md
+++ b/configs/documentation/README.md
@@ -1,0 +1,8 @@
+# Documentation-related configurations
+
+This directory contains the configuration files used by for Iroha documentation.
+These files are maintained and updated depending on the internal changes in Iroha.
+
+## Configuration list
+
+* Cargo.*.toml - Cargo configuration for the Rust tutorial.


0x009922:
Let me try to understand better what is the problem we need to solve. 

To clarify the context, the documentation piece is the following:

> Once the initial `v2.0.0` release is complete, we plan to create a package on [crates.io](https://crates.io/) with all the documentation. In the meantime, you could use the local copy that you've just created in the [previous step](https://hyperledger.github.io/iroha-2-docs/guide/build.html) as a local installation in your client's `Cargo.toml`:
> 
> ```toml
> [dependencies]
> iroha_client = { version = "=2.0.0-pre-rc.13", path = "~/Git/iroha/client" }
> iroha_data_model = { version = "=2.0.0-pre-rc.13", path = "~/Git/iroha/data_model" }
> iroha_crypto = { version = "=2.0.0-pre-rc.13", path = "~/Git/iroha/crypto" }
> iroha_config = { version = "=2.0.0-pre-rc.13", path = "~/Git/iroha/config" }
> ```

You say:

> It has less chance of remaining unnoticed for a long time if it belongs to the main repo, at least in my opinion.

So the problem is that this snippet (and other similar ones) might get outdated, right?

This doesn't make much sense for me. First of all, I have a strong confusion about versioning, releases, `v2.0.0`, `rc.13`, and `path = ~/Git/iroha/crypto`. It is **so** overcomplicated for the tutorial guide to mention all of this. IMO, the approach of the tutorial must be as straightforward as possible. It is hardly achievable now, because we have `stable`, `dev` and `lts` channels all with different APIs and guarantees about backwards and forwards compatibility.

**What should the documentation rely on???**

IMO, it should rely on something single.

Coming back to the original issue. IMO, there is no need to integrate documentation-related `Cargo.toml`s into the main Iroha repository. A one that is integrated into the repo will look like this:

```toml
[dependencies]
iroha_client.workspace = true
# or
iroha_client = { path = "../../client" }
```

And this doesn't make any sense for the documentation. If we want to achieve a "real-world"-like Iroha-based project configuration, we might create a real-world Iroha-based project and store it as a separate repo, e.g. `hyperledger/iroha-2-docs-rust-tutorial-source`. In that project, you can define a working `Cargo.toml`, as close as possible to the project users might create. (Actually, users might clone that repo and run everything step-by-step, make changes in code, etc).

```toml
[dependencies]
iroha_client = { "git" = "https://github.com/hyperledger/iroha.git", branch = "iroha2-stable" }
```

This is how you can use Iroha **at the moment, in reality**. Shall we create such a project? Might be a good idea. We will have a truly working independent sample. Now we have mostly ephemeral snippets of code, integrated into the main repo with crutches.

Also, I want to make another point. IMHO, things like describing a `Cargo.toml` for a working project should *not* be possible to outdate that easy. If it happens, it means Iroha API is not yet stable *enough* to be documented on the level of "beginner user tutorial".

In conclusion, I think we need to very clearly identify the versioning of Iroha 2, the guarantees and boundaries. The documentation needs a single, clear, and stable target to document.

0x009922:
@6r1d, same as with #3870 - closing this in favor of the examples repo?

Issue #3892, opened 2023-09-15
Implement `Retrieve` isi output recovery on client side


Arjentix:
This is a subtask of #3718.

After `Retrieve` instruction being implemented internaly in Iroha we need to add a way to send the instruction (transaction in fact) output to the client who submitted it.
Also need to update client SDK and add a test

- [ ] Iroha responds with transaction output
- [ ] Client SDK supports this response
- [ ] `Retrieve` test

Arjentix:
Deprecated since we decided to not to proceed with this approach

PR #3893, opened 2023-09-15
[feature] #3891: Implement `Retrieve` isi internally

### LINKED ISSUE #3891: Implement `Retrieve` isi internally
This is a subtask of #3718. 
This division is done to reduce the amount of code to review in PRs.

Need to impement `Retrieve` instruction on Iroha side. So that clients can send this instruction, but can not request the result, because that part will be implemented in another issue.
This requires a lot of changes inside core, data_model and validator as well.

Arjentix:
## Description

- Added `Retrieve` isi
- Added `Output` associated type for `Instruction` trait, because now instructions can return something other than just `()`
- Refactored validators to support these new output types
- Fixed error types in `iroha_wasm_codec`
- `ValidationFail::InternalError` is now more descriptive, because it's filled with debug formated error now

### Linked issue

- Closes #3891 

### Follow-up task

- #3892

### Benefits

- Users can send `Retrieve` instruction
- Validator infrastructure is now more flexible with different verdict types

### Downsides

- Validator infrastructure becomed a bit more complicated

### Checklist


Erigara (commented):
Before extending `iroha` api surface we should consider whatever what we are trying to do is possible by means we already. 

For example if user what to execute query and make some changes based on it's result it's already possible. 
If user what to additionally get query result back, user can write result into some special store asset. And retrieve it either through normal query or through event subscription. 

Any thoughts @mversic?

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
```diff
@@ -14,7 +14,10 @@ use iroha_validator::{
 extern crate panic_halt;
 
 struct Validator {
-    verdict: Result,
+    transaction_verdict: MaybeUninitialized<TransactionValidationResult>,
```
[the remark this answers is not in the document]
Arjentix: Yes, it's my custom type, because otherwise for transaction it would be `Option<Result<Option<Option<InstructionBox::Output>>>>`. So I decided to add semantic value for a type at least for the first `Option`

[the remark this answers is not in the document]
Arjentix: Which name could you suggest? I agree that it sounds similar, but I have no other ideas

Arjentix:
> If user what to additionally get query result back, user can write result into some special store asset. And retrieve it either through normal query or through event subscription.

I thought about that but to me it sounds more like a crutch. Also from whos behalf this write should be performed? And what if such changes are restricted by user permissions? I think we should solve it on the protocol level, not on the business-logic level.

Event sending sounds interesting for that case. However that part with sending actual result will be impelemented in #3892

Erigara:
>  Also from whos behalf this write should be performed?

On behalf of a user who submitted transaction?

> And what if such changes are restricted by user permissions?

Maybe i misunderstood you but if such actions are restricted why would we allow them through different isi?

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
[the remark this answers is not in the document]
Erigara: > Which name could you suggest?

`VerdictOption`, `MaybeVerdict`

Arjentix:
>  Maybe i misunderstood you but if such actions are restricted why would we allow them through different isi?

I mean that using validator users can forbid to modify metadata or store assets at all. Our internal thing should not rely on that.

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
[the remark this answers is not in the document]
Arjentix: I like `MaybeVerdict`, will fix

Erigara:
> Won't it be too implicit?

But we already make any action on behalf of user who submitted transaction, no?

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
```diff
@@ -14,7 +14,10 @@ use iroha_validator::{
 extern crate panic_halt;
 
 struct Validator {
-    verdict: Result,
+    transaction_verdict: MaybeVerdict<TransactionValidationOutput>,
+    instruction_verdict: MaybeVerdict<InstructionValidationOutput>,
+    query_verdict: MaybeVerdict<QueryValidationOutput>,
```
mversic, line 19: why do we need multiple verdicts?

default_validator/src/lib.rs
```diff
@@ -171,26 +197,30 @@ pub fn validate_transaction(
     authority: AccountId,
     transaction: VersionedSignedTransaction,
     block_height: u64,
-) -> Result {
+) -> Result<TransactionValidationOutput> {
```
mversic, line 200: IMO validation functions of the executor should never return output value. Output value should be cached on the host side when the host computes it (which happens on `.execute`). The return value should not go through executor, this is just a waste of resoureces

core/src/smartcontracts/isi/query.rs
```diff
@@ -25,6 +25,18 @@ pub enum LazyValue<'a> {
     Iter(Box<dyn Iterator<Item = Value> + 'a>),
 }
 
+impl LazyValue<'_> {
+    /// Compute the lazy value.
+    ///
+    /// Returns the [`Value`] if it's already computed, or collects the iterator into [`Value::Vec`].
+    pub fn collect(self) -> Value {
```
mversic, line 32: I really dig this change. Could we do something similar for the client as well?

default_validator/src/lib.rs
Arjentix: Our Executor already validates transaction by executing all instructions. And after that host system will execute transaction for a second time.

I think it makes sence for Executor to actually return the result of execution, if it already have to execute transaction.

Erigara: > And after that host system will execute transaction for a second time.

No, we removed second execution

mversic: and return it to who, it's just pipelined from host to Executor and back to host

Arjentix:
After a discussion with @mversic and @Erigara we decided to not to proceed with this and freeze this task until we have a more suitable architecture

Issue #3882, opened 2023-09-11
Rewrite iroha_data_model_derive to use syn 2.0


DCNick3:
Part of #3682 

Do the same work that was done in https://github.com/hyperledger/iroha/pull/3727 and https://github.com/hyperledger/iroha/pull/3759 and #3814 , but for `iroha_data_model_derive`

Can also probably do a drive-by fixes/improvements described in #3737 and #2437

PR #3896, opened 2023-09-18
[refactor] #3856: Proc macro for default validator boilerplate

### LINKED ISSUE #3856: Write a macro to remove bloat from custom validators
If one compares different validators in tests, one can see there is a lot of bloat that one has to write to make a custom validator. Let's remove this difficult and confusing manual effort from the user by providing a macro

ilchu:
## Description

Now the different flavors of validators can be generated with less repetition.

### Linked issue

Closes #3856. 

### Benefits
Less non-obvious boilerplate when defining custom validators.

### Checklist

### DIFF
--- a/client/tests/integration/smartcontracts/executor_with_admin/src/lib.rs
+++ b/client/tests/integration/smartcontracts/executor_with_admin/src/lib.rs
@@ -2,175 +2,33 @@
 //! If authority is not `admin@admin` then default validation is used as a backup.
 
 #![no_std]
-#![allow(missing_docs, clippy::missing_errors_doc)]
 
 #[cfg(not(test))]
 extern crate panic_halt;
 
-use iroha_executor::{
-    data_model::evaluate::{EvaluationError, ExpressionEvaluator},
-    parse,
-    prelude::*,
-    smart_contract,
-};
+use iroha_executor::{parse, prelude::*, smart_contract};
 use lol_alloc::{FreeListAllocator, LockedAllocator};
 
 #[global_allocator]
 static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
 
+#[derive(Constructor, ValidateEntrypoints, ExpressionEvaluator, Validate, Visit)]
+#[visit(custom(visit_instruction))]
 struct Executor {
     verdict: Result,
     block_height: u64,
     host: smart_contract::Host,
 }
 
-impl Executor {
-    /// Construct [`Self`]
-    pub fn new(block_height: u64) -> Self {
-        Self {
-            verdict: Ok(()),
-            block_height,
-            host: smart_contract::Host,
-        }
-    }
-}
-
-macro_rules! defaults {
-    ( $($executor:ident $(<$param:ident $(: $bound:path)?>)?($operation:ty)),+ $(,)? ) => { $(
-        fn $executor $(<$param $(: $bound)?>)?(&mut self, authority: &AccountId, operation: $operation) {
-            iroha_executor::default::$executor(self, authority, operation)
-        } )+
-    };
-}
-
-impl Visit for Executor {
-    fn visit_instruction(&mut self, authority: &AccountId, isi: &InstructionExpr) {
-        if parse!("admin@admin" as AccountId) == *authority {
-            pass!(self);
-        }
-
-        iroha_executor::default::visit_instruction(self, authority, isi);
+fn visit_instruction(executor: &mut Executor, authority: &AccountId, isi: &InstructionExpr) {
+    if parse!("admin@admin" as AccountId) == *authority {
+        pass!(executor);
     }
 
-    defaults! {
-        visit_unsupported<T: core::fmt::Debug>(T),
-
-        visit_transaction(&SignedTransaction),
-        visit_expression<V>(&EvaluatesTo<V>),
-        visit_sequence(&SequenceExpr),
-        visit_if(&ConditionalExpr),
-        visit_pair(&PairExpr),
-
-        // Peer validation
-        visit_unregister_peer(Unregister<Peer>),
-
-        // Domain validation
-        visit_unregister_domain(Unregister<Domain>),
-        visit_transfer_domain(Transfer<Account, DomainId, Account>),
-        visit_set_domain_key_value(SetKeyValue<Domain>),
-        visit_remove_domain_key_value(RemoveKeyValue<Domain>),
-
-        // Account validation
-        visit_unregister_account(Unregister<Account>),
-        visit_mint_account_public_key(Mint<PublicKey, Account>),
-        visit_burn_account_public_key(Burn<PublicKey, Account>),
-        visit_mint_account_signature_check_condition(Mint<SignatureCheckCondition, Account>),
-        visit_set_account_key_value(SetKeyValue<Account>),
-        visit_remove_account_key_value(RemoveKeyValue<Account>),
-
-        // Asset validation
-        visit_register_asset(Register<Asset>),
-        visit_unregister_asset(Unregister<Asset>),
-        visit_mint_asset(Mint<NumericValue, Asset>),
-        visit_burn_asset(Burn<NumericValue, Asset>),
-        visit_transfer_asset(Transfer<Asset, NumericValue, Account>),
-        visit_set_asset_key_value(SetKeyValue<Asset>),
-        visit_remove_asset_key_value(RemoveKeyValue<Asset>),
-
-        // AssetDefinition validation
-        visit_unregister_asset_definition(Unregister<AssetDefinition>),
-        visit_transfer_asset_definition(Transfer<Account, AssetDefinitionId, Account>),
-        visit_set_asset_definition_key_value(SetKeyValue<AssetDefinition>),
-        visit_remove_asset_definition_key_value(RemoveKeyValue<AssetDefinition>),
-
-        // Permission validation
-        visit_grant_account_permission(Grant<PermissionToken>),
-        visit_revoke_account_permission(Revoke<PermissionToken>),
-
-        // Role validation
-        visit_register_role(Register<Role>),
-        visit_unregister_role(Unregister<Role>),
-        visit_grant_account_role(Grant<RoleId>),
-        visit_revoke_account_role(Revoke<RoleId>),
-
-        // Trigger validation
-        visit_unregister_trigger(Unregister<Trigger<TriggeringFilterBox>>),
-        visit_mint_trigger_repetitions(Mint<u32, Trigger<TriggeringFilterBox>>),
-        visit_burn_trigger_repetitions(Burn<u32, Trigger<TriggeringFilterBox>>),
-        visit_execute_trigger(ExecuteTrigger),
-
-        // Parameter validation
-        visit_set_parameter(SetParameter),
-        visit_new_parameter(NewParameter),
-
-        // Upgrade validation
-        visit_upgrade_executor(Upgrade<iroha_executor::data_model::executor::Executor>),
-    }
-}
-
-impl Validate for Executor {
-    fn verdict(&self) -> &Result {
-        &self.verdict
-    }
-
-    fn block_height(&self) -> u64 {
-        self.block_height
-    }
-
-    fn deny(&mut self, reason: ValidationFail) {
-        self.verdict = Err(reason);
-    }
-}
-
-impl ExpressionEvaluator for Executor {
-    fn evaluate<E: Evaluate>(&self, expression: &E) -> Result<E::Value, EvaluationError> {
-        self.host.evaluate(expression)
-    }
+    iroha_executor::default::visit_instruction(executor, authority, isi);
 }
 
 #[entrypoint]
 pub fn migrate(_block_height: u64) -> MigrationResult {
     Ok(())
-}
-
-#[entrypoint]
-pub fn validate_transaction(
-    authority: AccountId,
-    transaction: SignedTransaction,
-    block_height: u64,
-) -> Result {
-    let mut executor = Executor::new(block_height);
-    executor.visit_transaction(&authority, &transaction);
-    core::mem::forget(transaction);
-    executor.verdict
-}
-
-#[entrypoint]
-pub fn validate_instruction(
-    authority: AccountId,
-    instruction: InstructionExpr,
-    block_height: u64,
-) -> Result {
-    let mut executor = Executor::new(block_height);
-    executor.visit_instruction(&authority, &instruction);
-    core::mem::forget(instruction);
-    executor.verdict
-}
-
-#[entrypoint]
-pub fn validate_query(authority: AccountId, query: QueryBox, block_height: u64) -> Result {
-    let mut executor = Executor::new(block_height);
-    executor.visit_query(&authority, &query);
-    core::mem::forget(query);
-    executor.verdict
 }
--- a/smart_contract/executor/derive/src/lib.rs
+++ b/smart_contract/executor/derive/src/lib.rs
@@ -1,10 +1,14 @@
 //! Crate with executor-related derive macros.
 
+use iroha_macro_utils::Emitter;
+use manyhow::manyhow;
 use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use syn::{parse_macro_input, parse_quote, DeriveInput};
 
 mod conversion;
+mod default;
 mod entrypoint;
 mod token;
 mod validate;
@@ -144,8 +148,8 @@ pub fn derive_token(input: TokenStream) -> TokenStream {
     ValidateGrantRevoke,
     attributes(validate, validate_grant, validate_revoke)
 )]
-pub fn derive_validate(input: TokenStream) -> TokenStream {
-    validate::impl_derive_validate(input)
+pub fn derive_validate_grant_revoke(input: TokenStream) -> TokenStream {
+    validate::impl_derive_validate_grant_revoke(input)
 }
 
 /// Should be used together with [`ValidateGrantRevoke`] derive macro to derive a conversion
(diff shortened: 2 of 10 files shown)


client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
```diff
@@ -4,163 +4,24 @@
 #![no_std]
 #![allow(missing_docs, clippy::missing_errors_doc)]
```
mversic, line 5: can you remove these lints now?

client/tests/integration/smartcontracts/validator_with_migration_fail/src/lib.rs
```diff
@@ -10,131 +10,17 @@ extern crate panic_halt;
...
+use iroha_validator::{data_model::ValidationFail, parse, prelude::*};
+
+default_validator!(
+    impl Visit for Validator {
+        fn visit_query(&mut self, _authority: &AccountId, _query: &QueryBox) {
+            self.deny(ValidationFail::NotPermitted(
+                "All queries are forbidden".to_owned(),
+            ));
         }
     }
```
mversic, line 22:
```suggestion
default_validator!(
    pub struct Validator {
        data1: u32,    // Just an example of additional fields in the validator
    }

    impl Visit for Validator {
        fn visit_query(&mut self, _authority: &AccountId, _query: &QueryBox) {
            self.deny(ValidationFail::NotPermitted(
                "All queries are forbidden".to_owned(),
            ));
        }
    }
```
You should not implicitly define a custom validator especially since you're referring to it even outside of the macro. Except for being able to control the name of you validator struct, you should also be able to define additional fields in the struct (such as `data1` in my example)

mversic (changes requested):

wasm/validator/derive/src/default.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    let idents_to_exclude = custom_items.0.as_ref().map_or_else(
+        || vec![],
+        |impl_| {
+            let items = &impl_.items;
+            items
+                .iter()
+                .map(|item| match item {
+                    syn::ImplItem::Method(method) => &method.sig.ident,
+                    _ => panic!("Only methods are expected in `Visit` impl blocks"),
```
mversic, line 197: don't use panics. @DCNick3 what are we using instead of `proc-macro-error` now?

DCNick3: `manyhow`, and, where applicable, `iroha_macro_utils::Emitter` (though it's not yet merged, see #3882)

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -299,32 +191,3 @@ pub fn migrate(_block_height: u64) -> MigrationResult {
...
-    validator.visit_instruction(&authority, &instruction);
-    validator.verdict
-}
-
-#[entrypoint]
-pub fn validate_query(authority: AccountId, query: QueryBox, block_height: u64) -> Result {
-    let mut validator = Validator::new(block_height);
-    validator.visit_query(&authority, &query);
-    validator.verdict
-}
```
Arjentix, line 330: To be honest, I'd prefer to leave all `entrypoint`s as is. So that it will be clear which functions can be executed by the host

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
```diff
@@ -56,22 +52,33 @@ mod token {
     pub struct CanControlDomainLives;
 }
 
-struct Validator {
-    verdict: Result,
-    block_height: u64,
-    host: iroha_wasm::Host,
-}
+default_validator!(
```
Arjentix, line 55: Can this be a derive macro instead? Or probably two. Something like:

```rust
#[derive(Validate, Visit)]
// Tells `Visit` macro to fallback to `visit_register_domain()` function from this module
// instead of `iroha_wasm_validator::default::visit_register_domain()`
#[visit(custom(visit_register_domain))]
pub struct Validator {
    // Here goes probably the same fields for every validator,
    // But I don't think it's a big problem and prefer transparency
}

pub fn visit_register_domain(&mut Validator, authority: &AccountId, isi: Register<Domain) {
    // ...
}
```

Arjentix: @mversic , what do you think?

Arjentix (changes requested):

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
mversic: > @mversic , what do you think?

IMO your suggestion looks more complicated. Why would you enforce derive macro?

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
ilchu: I don't think so as the generated code is still exactly same. And it's mostly firing up because of the many `visit_...` functions whose names are self-descriptive.

wasm/validator/derive/src/default.rs
ilchu: Will do that. I suppose it's better not to touch refactoring the remainder of the `validator_derive` crate to keep things granular?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
ilchu: I was thinking about going that route as well, but the biggest problem I've seen is that the visit functions should be inside an `impl Visit` block, and we can only have one that's generated by the macro. The idea of an attribute specifying a fallback has somehow slipped me. 😢 I'd say that with such a scheme the code looks a bit more cleaner and compact, and re: complicatedness I think it's only marginally worse, and only in terms of implementation, whereas user-facing experience is indeed more straightforward.

client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs
mversic: ok, but can you attach it to specific functions? this should not be a burden on the user

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
mversic: @ilchu you will find that you can't alter struct's body with derive macro. Which means that you cannot hide those repeating fields like `verdict`. Also, how do you measure complicatedness? I find it quite incomprehensible to read the code written by @Arjentix whereas I had no problem in reading your original code

I'd like to hear @Erigara and @DCNick3 opinions here. Do you prefer function like macro or derive macro in this case?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
mversic: it doesn't sound entirely logical to me that this is part of the `default_validator` macro

we can:
1. leave it as is (user will copy this bloat for every validator)
2. make it part of the `default_validator` as was done in this PR. (But we need a way to be able to provide custom impls)
3. make another macro that is somehow related to `default_validator` macro (it needs to know the name of the validator struct generated by it)

DCNick3:
I may be a bit out of the loop, but why do we need to include all the non-modified methods of the `Visit` trait in the impl? Doesn't the default impl already call those without any explicit delegation, as all those methods have default bodies in the trait definition?

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
DCNick3: I like the `derive(Validate)`. To "hide" extra fields you can put them all into a struct `ValidatorBase` (name up for bike shed) defined in `iroha_validator`.

As for the `derive(Visit)`, I am not even sure why this is required, see https://github.com/hyperledger/iroha/pull/3896#issuecomment-1724995062.

Arjentix: > Item-level function-like macros should be avoided IMO, they are often too much magic

I agree. `derive`-macros has explicit abilities and clearly stays to which trait they're attached to.

> Which means that you cannot hide those repeating fields like verdict

Yes, but `derive`-macro can panic if some field is missing

Arjentix:
@DCNick3 , the problem we have is that `Visit` is declared inside `iroha_data_model` and by all meanings its default implementations should not do anything. That's the idea taken from `syn` by @mversic.
And then we have default validator which reimplements pretty much all methods of `Visit` trait. It also requires additional methods like `verdict()` and etc, which are implemented under `Validate` trait.

DCNick3:
Yeah, I see the inspiration from syn visitors, I think it's a good design. But I don't understand why all the generated `fn visit_something(&mut self, s: Something) { default::visit_something(self, s); }` methods are needed.

After a little bit of digging, I think they were necessary when `DefaultValidator` was a thing (see https://github.com/hyperledger/iroha/blob/d57ab25a8a57f10a3954471ed0fd260dfd0c7596/client/tests/integration/smartcontracts/validator_with_admin/src/lib.rs#L18). But now that it's gone I don't think the generated code should include all those delegating methods and instead use the default impls (the same way it's done in serde).

```rust
impl Visit for Validator {
    fn visit_instruction(&mut self, authority: &AccountId, isi: &InstructionBox) {
        if parse!("admin@admin" as AccountId) == *authority {
            pass!(self);
        }

        iroha_validator::default::visit_instruction(self, authority, isi);
    }
}
```

I don't think this^ needs any proc macro help, this is very minimal

Arjentix:
What do you mean by default impls, if `iroha_data_model::visit::Visit` cannot provide default impls related to validator? That's a more general thing, it can be used for any Iroha expression traversion, i.e. DSL

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
DCNick3: I think keeping the entrypoints separate is a good idea. Bikeshed syntax:

```rust
default_entrypoints!{
  [
    validate_transaction,
    validate_instruction,
    validate_query,
  ] => MyValidator
}
```

DCNick3:
> That's a more general thing, it can be used for any Iroha expression traversion

I see, I did notice that the trait is not specific to the validator.

Then again, I would propose to make a duplicating trait specific to validator. It _would_ have `iroha_validator::default::*` methods as its default trait implementations. With this, there would no longer be a need for the user to use a proc macro for the delegating boilerplate, a standard rust feature will handle it for them.

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
ilchu: I feel it's more of a question about explicitness and what we would want to provide to the end user. The current func-like macro solution hides more things, but allows arbitrary magic (and thus has more assumptions baked in too), making it more frustrating when it actually breaks. The derive approach is more of an opposite. And after this discussion I'd rather lean on the side of being explicit.

mversic:
> Then again, I would propose to make a duplicating trait specific to validator. It would have iroha_validator::default::* methods as its default trait implementations. With this, there would no longer be a need for the user to use a proc macro for the "delegating visitor" boilerplate, a standard rust feature will handle it for them.

no, because we don't want surprising default implementations for `Validate` trait. Every method should remain empty and just visit the node offering you the possibility to add your own behavior. There is a difference between "empty" validator which is a sensible default and "DefaultValidator" which

mversic:
> > Then again, I would propose to make a duplicating trait specific to validator. It would have iroha_validator::default::* methods as its default trait implementations. With this, there would no longer be a need for the user to use a proc macro for the "delegating visitor" boilerplate, a standard rust feature will handle it for them.
> 
> no, because we don't want surprising default implementations for `Validate` trait. Every method should remain empty and just visit the node offering you the possibility to add your own behavior. There is a difference between "empty" validator which is what you get when you implement `Visit` and default validation functions which you can use to override those empty impls.
> 
> We want to be able to override empty methods with some validation, not the other way around

let me put it this way. The true default validator is empty, i.e. everything goes. However, we offer you some building blocks in the `default.rs` to create your own validator and if you want to use all of them we provide you with a macro `default_validator`

Arjentix:
Also, as @mversic mentioned somewhere, users should be quite careful with using things from `default.rs`. For example, these functions expect some permission tokens to be defined, and if user do not provide them in `migrate()` entrypoint, things can go wrong.

That's another reason to not to make things from `default.rs` really be the default trait implementations. They rely too much on `migrate()`.

client/tests/integration/smartcontracts/validator_with_custom_token/src/lib.rs
ilchu: Went for a `#[omit(entrypoints(..)` syntax here, should work just as well.

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+
+/// Implements the `iroha_validator::Validate` trait for the given `Executor` struct. As
+/// this trait has a `iroha_validator::prelude::Visit`, and the latter has an
+/// `iroha_validator::iroha_data_model::evaluate::ExpressionEvaluator`
+/// bound, at least these two should be implemented as well.
+///
+/// Emits a compile error if the struct didn't have all the expected fields with corresponding
+/// types, i.e. `verdict`: `iroha_validator::prelude::Result`, `block_height`: `u64` and
+/// `host`: `iroha_validator::iroha_wasm::Host`, though technically only `verdict` and
+/// `block_height` are needed. The types can be unqualified, but not aliased.
```
Arjentix, line 211: `iroha_validator` path is outdated, now it's `iroha_executor`

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+///    host: iroha_wasm::Host,
+/// }
+///
+/// // Custom visit function should supply a `&mut Validator` as first argument
+/// fn visit_query(validator: &mut Validator, _authority: &AccountId, _query: &QueryBox) {
+///     validator.deny(ValidationFail::NotPermitted(
+///         "All queries are forbidden".to_owned(),
+///     ));
+/// }
+/// ```
```
Arjentix, line 256: Same

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+    let result = default::impl_derive_validate(&mut emitter, &input);
+
+    emitter.finish_token_stream_with(result)
+}
+
+/// Implemets the `iroha_validator::prelude::Visit` trait on a given `Executor` struct.
+/// Users can supply custom overrides for any of the visit functions as freestanding functions
+/// in the same module or omit some of the visit functions completely via the `#[omit(visit(...))]`
+/// attribute by supplying corresponding visit function names inside of it, otherwise a default
+/// implementation from `iroha_validator::default` module is used.
```
Arjentix, line 230: You write ` ... or omit ...` but in fact looks like users *have* to use `omit` to provide their own functions, no?

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+///
+/// #[derive(Constructor, Entrypoints, ExpressionEvaluator, Validate, Visit)]
+/// #[omit(entrypoints(validate_query))]
+/// pub struct Validator {
+///    verdict: Result,
+///    block_height: u64,
+///    host: iroha_wasm::Host,
+/// }
+///
+/// ```
```
Arjentix, line 297: Also the problem with outdated `iroha_validator` path

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+    };
+
+    let result = default::impl_derive_visit(&mut emitter, &input);
+
+    emitter.finish_token_stream_with(result)
+}
+
+/// Implements three default entrypoints on a given `Executor` struct: `validate_transaction`,
+/// `validate_query` and `validate_instruction`. The `migrate` entrypoint is implied to be
+/// implemented manually by the user if needed.
```
Arjentix, line 273: `migrate()` entrypoint is mandatory as well as others, so `if needed` doesn't look right to me

Arjentix: I see, you don't want to generate `migrate()` because it will be custom in most of the cases. Than how about changing `Entrypoints` to `ValidateEntrypoints`

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+    emitter.finish_token_stream_with(result)
+}
+
+/// Implements `iroha_validator::iroha_data_model::evaluate::ExpressionEvaluator` trait
+/// for the given `Executor` struct.
+///
+/// Emits a compile error if the struct didn't have all the expected fields with corresponding
+/// types, i.e. `verdict`: `iroha_validator::prelude::Result`, `block_height`: `u64` and
+/// `host`: `iroha_validator::iroha_wasm::Host`, though technically only `host` is needed.
+/// The types can be unqualified, but not aliased.
```
Arjentix, line 318: Same with `iroha_validator`

smart_contract/executor/derive/src/lib.rs
```diff
@@ -195,3 +199,155 @@ pub fn derive_ref_into_account_owner(input: TokenStream) -> TokenStream {
...
+}
+
+/// Implements a constructor for the given `Executor` struct. If the `Executor` has any custom fields
+/// (i.e. different from the expected fields listed below), they will be included into the constructor
+/// automatically and will need to be passed into `new()` function explicitly. In the default case,
+/// only the `block_height` needs to be supplied manually.
+///
+/// Emits a compile error if the struct didn't have all the expected fields with corresponding
+/// types, i.e. `verdict`: `iroha_validator::prelude::Result`, `block_height`: `u64` and
+/// `host`: `iroha_validator::iroha_wasm::Host`. The types can be unqualified, but not aliased.
```
Arjentix, line 340: Same with `iroha_validator`

client/tests/integration/smartcontracts/executor_with_admin/src/lib.rs
```diff
@@ -2,172 +2,33 @@
...
-    smart_contract,
-};
+use iroha_executor::{parse, prelude::*, smart_contract};
 use lol_alloc::{FreeListAllocator, LockedAllocator};
 
 #[global_allocator]
 static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
 
+#[derive(Constructor, Entrypoints, ExpressionEvaluator, Validate, Visit)]
+#[omit(visit(visit_instruction))]
```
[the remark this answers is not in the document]
ilchu: It was a bit easier to implement this way via darling through a shared field on a receiver, but I'll try refactoring it otherwise.

smart_contract/executor/derive/src/lib.rs
ilchu: Indeed, good catch in wording.

client/tests/integration/smartcontracts/executor_with_admin/src/lib.rs
```diff
@@ -2,172 +2,34 @@
...
 
-impl Executor {
-    /// Construct [`Self`]
-    pub fn new(block_height: u64) -> Self {
-        Self {
-            verdict: Ok(()),
-            block_height,
-            host: smart_contract::Host,
-        }
+#[allow(dead_code)] // Used in `Visit` derive.
```
Arjentix, line 23: If so then why does it trigger the warning...?

Arjentix: Having this attribute seems to be dangerous to me. It may mean that this function is really not used and macro was expanded to something else

Arjentix: Same for all other validator files

Arjentix: And looks like that is exactly why integration test with validator upgrade is failing

smart_contract/executor/derive/src/lib.rs
Arjentix: I don't understand how they may not need them if this functions are the part of the `Visit` trait and they will be implemented anyway

ilchu: But as @mversic mentioned above, the default `visit_..` fns just visit the nodes without additional behavior, and the `Visit` impl just brings them into immediate scope. It is true that we're not actually omitting things listed in the attribute. So thinking about it again, maybe it makes more sense to call this `replace` instead of `omit` as that's what's actually being done.

Arjentix: Yeah, I'd like `replace` more. Or maybe something like `custom`

Arjentix (commented):
No more objections, just https://github.com/hyperledger/iroha/pull/3896#discussion_r1373020651

Arjentix (approved):

client/tests/integration/smartcontracts/executor_with_admin/src/lib.rs
```diff
@@ -2,172 +2,33 @@
...
 
 #[global_allocator]
 static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
 
+#[derive(Constructor, Entrypoints, ExpressionEvaluator, Validate, Visit)]
+#[omit(visit(visit_instruction))]
 struct Executor {
     verdict: Result,
     block_height: u64,
     host: smart_contract::Host,
```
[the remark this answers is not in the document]
ilchu: I thought about hiding these fields in an `ExecutorBase` auto-generated struct as @DCNick3 suggested in a [comment](https://github.com/hyperledger/iroha/pull/3896#discussion_r1329718902), but felt that some external code might rely on the current `Executor` layout and it would be hard to enforce this invariant. I could still create an issue if you feel this feature is relevant.

mversic (dismissed):
lgtm, I'm just questioning should `ValidateEntrypoints` be a derive macro?

mversic:
after you merge to dev, please create a backport to rc20

mversic (approved):

Arjentix (approved):

PR #3903, opened 2023-09-20
[fix] #3899: Fix topology mismatch bug

### LINKED ISSUE #3899: [bug] Topology mismatch when starting network
When I start docker compose or `test_env.py` I can see `TopologyMismatch` error. Because of that genesis is not being committed.

Arjentix:
## Description

- Fixed trusted peer deserialization. `HashSet` was replaced by `Vec`
- Some minor refactoring

However I don't quite understand this change fixes the main problem. I was testing with `test_env.py` which runs the same binary multiple times, so hashing should stay the same.

### Linked issue

- Closes #3899

### Benefits

- No more `TopologyMismatch` error on network startup

### Checklist


mversic (approved):

Erigara:
I think that there is another reason for this bug.

Before block API reactor we don't check topology for genesis block and now we do that (so we trust genesis peer to give us topology).

Erigara (dismissed):

config/src/sumeragi.rs
```diff
@@ -122,52 +123,8 @@ impl Configuration {
...
-                result.insert(value);
-            }
-
-            Ok(result)
-        }
-    }
-
-    let visitor = UniqueVisitor(core::marker::PhantomData);
-    deserializer.deserialize_seq(visitor)
+    #[serde(deserialize_with = "UniqueVec::deserialize_display")]
```
mversic, line 126: why do you need to refer to deserialization function explicitly? Should it be delegated to the correct function by default?

core/src/block.rs
```diff
@@ -434,39 +435,9 @@ mod valid {
...
-                    return Err((
-                        self,
-                        SignatureVerificationError::NotEnoughSignatures {
-                            votes_count,
-                            min_votes_for_commit: topology.min_votes_for_commit(),
-                        }
-                        .into(),
-                    ));
+            if !self.payload().header.is_genesis() {
+                if let Err(err) = topology.verify_signatures(self.signatures()) {
```
mversic, line 439: this is ok, although as you can see it does more checks then is strictly required.
If you could I would rather you moved this function into `block.rs` and remove redundant checks

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub const fn new() -> Self {
+        Self(Vec::new())
+    }
+
+    /// Try construct [`UniqueVec`] from iterator.
+    ///
+    /// # Errors
+    ///
+    /// - If there are duplicates in the sequence.
+    pub fn try_from_iter<I>(iter: I) -> Result<Self, DuplicationError>
```
mversic, line 58: you should rather implement `FromIterator` that returns `Result<Self, DuplicationError>`

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub fn clear(&mut self) {
+        self.0.clear()
+    }
+
+    /// Returns reference to an internal mutable slice.
+    ///
+    /// # Warning
+    ///
+    /// Caller is responsible for ensuring that elements in slice will remain unique.
+    pub fn as_mut_unchecked(&mut self) -> &mut [T] {
```
mversic, line 116: having this method can break the invariant of uniqueness and be the source of potential bugs. I would remove it

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,334 @@
...
+        }
+        Ok(Self(inner))
+    }
+}
+
+impl<T: PartialEq> From<UniqueVec<T>> for Vec<T> {
+    fn from(value: UniqueVec<T>) -> Self {
+        value.0
+    }
+}
```
mversic, line 140: IMO implementing `FromIterator` is sufficient. I would remove these

config/src/sumeragi.rs
Arjentix: I've found no way to have `deserialize()` using `Display` when it's implemented on `T`, and not using it in other cases.
We need `Display` to have a more precise user-facing error as it was before.

So now default `deserialize()` does not restrict `T` anyhow and provides a message just about a problematic place, `deserialize_debug()` prints also debug repr of duplicated value and `deserialize_display()` prints duplicated value using `Disoplay`.

primitives/src/unique_vec.rs
Arjentix: I cannot return `Err` from `FromIterator`

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,334 @@
...
+        Ok(v)
+    }
+
+    /// Try extend [`UniqueVec`] from iterator.
+    /// If an error occurs, all previously added elements will remain in the [`UniqueVec`].
+    ///
+    /// # Errors
+    ///
+    /// - If there are duplicates in the sequence.
+    pub fn try_extend<I>(&mut self, iter: I) -> Result<(), DuplicationError>
```
mversic, line 73: I would prefer if you just implemented `Extend` trait instead of providing `try_extend` method. It should be fine to remove duplicates in the process of extending. Maps or sets do a similar thing where they reject some of the `other` collection members silently

Arjentix: Hm, haven't thought about that. Makes sense, will do

primitives/src/unique_vec.rs
mversic: yes, you can. You have to make it like this:

```rs
impl<T: PartialEq> FromIterator<UniqueVec<T>> for Result<Vec<T>, DuplicationError>
```

Arjentix: Em... The thing you wrote works like `flat()`, I think you meant:

```rust
impl<T: PartialEq> FromIterator<T> for Result<UniqueVec<T>, DuplicationError>
```

Arjentix: I think you meant something different, because the thing you wrote works like `flatten()`.

Maybe you meant:

```rust
impl<T: PartialEq> FromIterator<T> for Result<UniqueVec<T>, DuplicationError>
```

which won't work, see [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fadcc2c9b71498fe0ab7355c1336cc4d)

primitives/src/unique_vec.rs
Arjentix: See https://github.com/hyperledger/iroha/pull/3903#discussion_r1335687887

Arjentix: `FromIterator` is [implemented](https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA,+E%3E%3E-for-Result%3CV,+E%3E) for `Result` in std, but it's not sufficient in our case

primitives/src/unique_vec.rs
Arjentix: Wait. We want it to fail to deserialize if there are any duplicates. Why don't we want the same explicit error in other cases?

Arjentix: Just checked how serde deserialization works for `HashSet`. It ignores all duplicates. Here is a workaround:
https://github.com/serde-rs/serde/issues/1607

Arjentix: Should I provide the same kind of API?

mversic: so, our choice here is to either:
1. have all custom fallible methods
2. always silently remove duplicates
 - in this case we can deserialize list which contains duplicates, and implement `FromIterator` and `Extend`
 
 > Should I provide the same kind of API?

I would prefer it that way. To me 2. would make more sense for `UniqueVec` if we look at it separately from it's use as a vec of peers. For a particular case where throwing error is required it can be done via a custom deserializing function
 
Did you consider using [itertools::all_unique](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.all_unique)

primitives/src/unique_vec.rs
mversic: I think that slapping `_unchecked` is only for constructors. I don't think you can apply it to getters. Can you point me to some example in `std`

mversic: it's not that the method is `unchecked` and if you maintain invariants at that place all will be good. The problem is that consumers of the slice you return `&mut [T]` have no way of knowing they can break the code by mutating it

mversic: you should implement mutating methods directly on the `UniqueVec` without exposing mutable slice directly

mversic: or you can drain `UniqueVec` into a `Vec` do the operation and collect it back into `UniqueVec`

Arjentix: > or you can drain UniqueVec into a Vec do the operation and collect it back into UniqueVec. This sounds even better to me

Nice, that should fix the original problem!

primitives/src/unique_vec.rs
Arjentix: > Did you consider using

I did not, will take a look, thanks

> To me 2 would make more sense for `UniqueVec`

Ok, that's ok for me. I just wanted to be sure that these two options are clear for us and we should just choose one of them. Will do.

core/src/sumeragi/network_topology.rs
```diff
@@ -158,38 +158,31 @@ impl Topology {
...
+            self.ordered_peers.as_mut_unchecked().rotate_left(rem);
         }
     }
 
     /// Re-arrange the set of peers after each successful block commit.
     pub fn rotate_set_a(&mut self) {
         let rotate_at = self.min_votes_for_commit();
         if rotate_at > 0 {
-            self.ordered_peers[..rotate_at].rotate_left(1);
+            self.ordered_peers.as_mut_unchecked()[..rotate_at].rotate_left(1);
```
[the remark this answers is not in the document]
Arjentix: > Though std unchecked is used to indicate unsafe methods.

Yes, I've also checked for that... I don't know a better name for *Business Logic Errors*.
But in another comment we decided to remove this method, so no problems.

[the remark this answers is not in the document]
mversic: > Though `std` `unchecked` is used to indicate `unsafe` methods.

indeed, it seems that `_unchecked` suffix implies `unsafe` in `std` because there is no `_unchecked` method that is not also `unsafe`. But it is my understanding that `_unchecked` warns you of potential logical errors, whereas `unsafe` warns you of potential UB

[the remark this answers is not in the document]
Arjentix: For logic errors usually `panic!` and friends work fine, but in the case we could only do that after. For example every next operation after `_unchecked()` would check for uniqueness and panic if assertion failed.

primitives/src/unique_vec.rs
Arjentix: Works now after we decided to ignore duplciates

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,363 @@
...
+impl<T: PartialEq> Default for UniqueVec<T> {
+    fn default() -> Self {
+        Self(Vec::default())
+    }
+}
+
+impl<T: PartialEq> UniqueVec<T> {
+    /// Create new [`UniqueVec`].
+    #[must_use]
+    pub const fn new() -> Self {
```
mversic, line 46: if I'm not mistaken you can use `derive_more::{Constructor, Deref, AsRef}`

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,363 @@
...
+        $(v.push($x);)+
+        v
+    }};
+}
+
+/// Wrapper type for [`Vec`] which ensures that all elements are unique.
+#[derive(
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize, Encode, Decode, IntoSchema,
+)]
+pub struct UniqueVec<T: PartialEq>(Vec<T>);
```
mversic, line 35:
```suggestion
pub struct UniqueVec<T>(Vec<T>);
```

is this bound required on the type? if you remove it you'll be able to derive more things, like `Default`

core/src/sumeragi/network_topology.rs
```diff
@@ -320,47 +268,28 @@ pub enum Role {
...
 #[cfg(test)]
 mod tests {
     use iroha_crypto::KeyPair;
+    use iroha_primitives::unique_vec;
 
     use super::*;
-    use crate::block::ValidBlock;
 
-    macro_rules! peers {
+    #[macro_export]
```
mversic, line 278: exported for what reason?

mversic: there are other ways to make it accessible from `block.rs`

primitives/src/unique_vec.rs
```diff
@@ -0,0 +1,363 @@
...
+    ///     numbers: UniqueVec<u32>,
+    /// }
+    ///
+    /// let err = serde_json::from_str::<Config>(r#"{"numbers": [1, 2, 3, 2, 4, 5]}"#).unwrap_err();
+    /// assert_eq!(
+    ///     err.to_string(),
+    ///     "Duplicated value at line 1 column 25",
+    /// );
+    /// ```
+    pub fn deserialize_failing_on_duplicates<D>(deserializer: D) -> Result<Self, D::Error>
```
mversic, line 165: IMHO these specialized deserialization functions should be defined where they are used. In this case it's the config

primitives/src/unique_vec.rs
Erigara: Imo type should have as little bounds as possible and bounds should be moved to places where actually required. 

For example the same pattern is used across std ([`BTreeMap`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html)).

primitives/src/unique_vec.rs
Arjentix: `Deref` ans `AsRef` work fine, but `Contructor` generates `new(vec: Vec<T>) -> Self`, which is not what I need

primitives/src/unique_vec.rs
Arjentix: > Imo type should have as little bounds as possible and bounds should be moved to places where actually required.

Indeed I think the same in most of the cases, but IMO sometimes it makes sence to restrict generic in the type. However I don't this this is the case, will fix

primitives/src/unique_vec.rs
Arjentix: I don't see the reason to make it less flexible if we already have more generic implementation...

@Erigara ?

core/src/sumeragi/network_topology.rs
Arjentix: Oof, yes. Working at late nights have its own consequences

primitives/src/unique_vec.rs
Arjentix: > Would orphan rule allow this? (we anyway can workaround this by using plain function)

Yes, as you said, we can use plain function, and that's how `#[serde(deserialize_with = "foo")]` works.

> Do we expect that this funtion can be reused in other places except config, if yes i would leave it here.

I don't know, maybe someday... But in that day someone who will write this code may not know that such function is already implemented in config and will re-implement it

Erigara: I don't mind leave it here.

mversic (approved):

Issue #3622, opened 2023-06-19
[suggestion] Deprecate `cargo lints`


appetrosyan:
### Feature request

Inline all of the general `deny` lints into the top levels of each affected crate. 

### Motivation

`cargo lints` is a package that is effecitvely a thin wrapper around `clippy`. The issue with building our CI around it is as follows
1. It applies to the entire workspace. Some lints are enforced as a matter of policy, e.g. groups like `pedantic` and `restriction`. However, most are specific to the crates in which they exist. For example, `arithmetic_side_effects` is only sensible in crates where we do arithmetic. This rules out `logger` and `telemetry`. 
2. It applies to tests and benchmarks, even though the standards for those crates are different. Sometimes expressive tests take shortcuts for the sake of readability and coverage. 
3. It requires more tools to be installed on the potential contributor's system. 
4. `cargo lints` hasn't been updated since May 2021. Likely won't be. 
5. It doesn't properly use `clippy-driver` which can potentially break Rust in the future. 

### Who can help?

@appetrosyan

DCNick3:
With the merge of #3904, `cargo-lints` is no longer being used and points 3, 4 and 5 no longer apply. We might still want to have the lints be more granular

mversic:
closed in favor of #3927

PR #3904, opened 2023-09-20
[ci] #3622: Use cargo manifest lints instead of unmaintained cargo-lints

DCNick3:
Since the implementation of rust-lang/rfcs#3389, it is now possible to specify workspace-level lints for rustc and clippy. This PR updates the cargo configuration and CI to use this new feature instead of cargo-lints.

Note that it was only stabilized in `nightly-2023-09-10`. Using it with out current toolchain requires either a -Zlints flag or a modification to `.cargo/config.toml`:

```toml
[unstable]
lints = true
```

Also note that unlike the original suggestion in #3622, this doesn't make the lints crate-level, but merely replaces a clunky unmaintained tool with a standard solution for configuring lints.

In particular, this PR:
- Removes old lints.toml configuration files for cargo-lints
- Adds [lint] tables to Cargo.toml of the root and wasm workspaces. The lints are duplicated between the two
- Replaces `cargo lints clippy` invocations with `cargo clippy -Zlints` in CI scripts
- Silences/fixes some new lints that popped up due to differences in how between `cargo lints` and workspaces
- Does a drive-by fix to iroha_genesis: it now too shares cargo metadata as do other crates

### Checklist

- [ ] Make sure it passes CI


mversic (approved):

Erigara (approved):
Nice work

PR #3908, opened 2023-09-21
[fix]: `mkdir -r` with store path, not lock path

0x009922:
## Description

TBH, I am lazy to prepare a proper reproduction of the bug. In my case, Iroha repeatedly failed to start, complaining about existence of the lock file all the time. After all, it turned out that `./storage/kura.lock`, created by Iroha, **is a directory, not a file**.

This PR makes variables naming clearer, and passes actual block-store path into `mkdir -r`, not the path of the lockfile.

### DIFF
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -413,20 +413,20 @@ impl BlockStore {
     ///
     /// # Panics
     /// * if you pass in `LockStatus::Unlocked` and it is unable to lock the block store.
-    pub fn new(path: &Path, already_locked: LockStatus) -> Self {
+    pub fn new(store_path: &Path, already_locked: LockStatus) -> Self {
         if matches!(already_locked, LockStatus::Unlocked) {
-            let path = path.join(LOCK_FILE_NAME);
+            let lock_path = store_path.join(LOCK_FILE_NAME);
             if let Err(e) = fs::File::options()
                 .read(true)
                 .write(true)
                 .create_new(true)
-                .open(path.clone())
+                .open(lock_path.clone())
             {
                 match e.kind() {
-                    std::io::ErrorKind::AlreadyExists => Err(Error::Locked(path)),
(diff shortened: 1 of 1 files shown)


core/src/kura.rs
```diff
@@ -413,43 +413,43 @@ impl BlockStore {
...
+                .open(lock_path.clone())
             {
                 match e.kind() {
-                    std::io::ErrorKind::AlreadyExists => Err(Error::Locked(path)),
+                    std::io::ErrorKind::AlreadyExists => Err(Error::Locked(lock_path)),
                     std::io::ErrorKind::NotFound => {
-                        match std::fs::create_dir_all(&path)
-                            .map_err(|e| Error::MkDir(e, path.clone()))
+                        match std::fs::create_dir_all(&store_path)
+                            .map_err(|e| Error::MkDir(e, store_path.into()))
```
0x009922, line 429: The actual problem was here: `create_dir_all(&path)`, where `path` is the lock file's path.

DCNick3 (approved):

core/src/kura.rs
```diff
@@ -413,43 +413,43 @@ impl BlockStore {
...
+    pub fn new(store_path: &Path, already_locked: LockStatus) -> Self {
         if matches!(already_locked, LockStatus::Unlocked) {
-            let path = path.join(LOCK_FILE_NAME);
+            let lock_path = store_path.join(LOCK_FILE_NAME);
             if let Err(e) = fs::File::options()
                 .read(true)
                 .write(true)
                 .create_new(true)
-                .open(path.clone())
+                .open(lock_path.clone())
```
pesterev, line 423: You can take a reference instead of making a clone.
```suggestion
                .open(&lock_path)
```

pesterev (approved):

DCNick3 (dismissed):

DCNick3 (approved):

pesterev (approved):

Issue #3906, opened 2023-09-21
[BUG]: Default validator memory failure


Erigara:
When large transaction is submitted (register 100 domains, 100_000 accounts and 100_000 asset definitions) validator failed with following error: 

```
error while executing at wasm backtrace:\n    0: 0x82d62 - <unknown>!<wee_alloc::LargeAllocPolicy as wee_alloc::AllocPolicy>::should_merge_adjacent_free_cells::ha0fc908c83f63ea9\n    1: 0x82d94 - <unknown>!wee_alloc::WeeAlloc::dealloc_impl::{{closure}}::h162cafb800fa43f5\n    2: 0x77c5 - <unknown>!<wee_alloc::WeeAlloc as core::alloc::global::GlobalAlloc>::dealloc::hd41631e66f421fb4\n    3: 0x9fb0 - <unknown>!core::ptr::drop_in_place<alloc::boxed::Box<iroha_data_model::expression::model::Expression>>::h26bdf7c39b483337\n    4: 0x256a0 - <unknown>!core::ptr::drop_in_place<iroha_data_model::isi::model::InstructionBox>::hc1a70663e73bc5fe\n    5: 0x242d5 - <unknown>!<I as iroha_wasm::ExecuteOnHost>::execute::h2b1c20a6b977bc88\n    6: 0x31f7f - <unknown>!<iroha_default_validator::Validator as iroha_data_model::visit::Visit>::visit_instruction::ha3c91381150e154a\n    7: 0x3dfe2 - <unknown>!_iroha_validator_validate_transaction
```

PR #3912, opened 2023-09-22
[fix] #3905: Fix apply_blocks and validate_blocks benchmark

### LINKED ISSUE #3905: [BUG]: `apply_blocks` and `validate_blocks` fails
Currently `apply_blocks` and `validate_blocks` fails silently.

Erigara:
## Description

Benchmarks now measure what they suppose to measure.
Fix memory issue inside wasm validator.

### Linked issue

Closes #3905, #3906 

### Benefits

Fixed benchmarks.


wasm/src/lib.rs
```diff
@@ -24,13 +24,14 @@ use data_model::{
...
 pub use iroha_wasm_derive::main;
+use lol_alloc::{FreeListAllocator, LockedAllocator};
 use parity_scale_codec::{DecodeAll, Encode};
 
 pub mod debug;
 pub mod log;
 
 #[global_allocator]
-static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
```
DCNick3, line 34: Looking at the docs, `LockedAllocator` is only required if multi-threading is used in wasm. With [AssumeSingleThreaded](https://docs.rs/lol_alloc/0.4.0/lol_alloc/struct.AssumeSingleThreaded.html), a few bytes may be shaved off.

core/src/validator.rs
```diff
@@ -27,10 +27,12 @@ impl From<wasm::error::Error> for ValidationFail {
...
                 match call_error {
                     ExecutionLimitsExceeded(_) => Self::TooComplex,
-                    HostExecution(error) | Other(error) => Self::InternalError(error.to_string()),
+                    HostExecution(error) | Other(error) => {
+                        Self::InternalError(format!("{error:#}"))
+                    }
                 }
             }
-            _ => Self::InternalError(err.to_string()),
+            _ => Self::InternalError(format!("{err:#}")),
```
DCNick3, line 35: How does this change the error message? Would it have proper newlines, or smth?

core/benches/blocks/common.rs
```diff
@@ -136,10 +172,12 @@ fn restore_every_nth(
     Ok(instructions)
 }
 
-fn build_wsv(account_id: &AccountId, key_pair: &KeyPair) -> WorldStateView {
+pub fn build_wsv(account_id: &AccountId, key_pair: &KeyPair) -> WorldStateView {
     let kura = iroha_core::kura::Kura::blank_kura_for_testing();
     let mut wsv = WorldStateView::new(World::with([], BTreeSet::new()), kura);
     wsv.config.transaction_limits = TransactionLimits::new(u64::MAX, u64::MAX);
+    wsv.config.wasm_runtime_config.fuel_limit = u64::MAX;
+    wsv.config.wasm_runtime_config.max_memory = u32::MAX;
```
DCNick3, line 180: Is this necessary even after fixing the memory leak? Or is there just no sense to limit these in a benchmark situation? :thinking:

DCNick3 (commented):
It's a bit hard to see what actually changed with all the renamings and stuff..

I understand what's the problem with the `wee_alloc` memory leak, but what the problem with `apply_blocks` and `validate_blocks`?

What was the problem with

core/src/validator.rs
Erigara: It will print causes of error ([link](https://docs.wasmtime.dev/api/wasmtime/struct.Error.html#display-representations))

core/benches/blocks/common.rs
Erigara: > Or is there just no sense to limit these in a benchmark situation? 🤔

Exactly, in the benchmark we are submitting large block that won't normally pass limitations in the validator.

Erigara:
> but what the problem with apply_blocks and validate_blocks?

@DCNick3 so problem was that:
1) Not enough fuel
2) Transactions don't have enough rights to be executed with default_validator

core/src/validator.rs
DCNick3: Ah, that's just an anyhow error. Makes sense.

DCNick3 (approved):

SamHSmith (approved):

PR #3913, opened 2023-09-22
[refactor] #3886: Remove 'Versioned' prefix from versioned containers

### LINKED ISSUE #3886: Rename versioned structs
Our naming should hide that structures are versioned. For example this means to rename `VersionedSignedBlock` -> `SignedBlock` and `SignedBlock` -> `V1SignedBlock` (or something similar). The idea is that the user should operate on `SignedBlock` and shouldn't ever be bothered by the fact that it may or may not be versioned. Versioning should be opaque 

In a sense this means we should revert our naming convention

mversic:
## Description

I think this makes a better API.
Would you agree?

### Linked issue

Closes #3886

### Checklist

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -26,9 +26,9 @@ use iroha_core::{
 use iroha_data_model::{
     block::{
         stream::{BlockMessage, BlockSubscriptionRequest},
-        VersionedSignedBlock,
+        SignedBlock,
     },
-    http::{BatchedResponse, VersionedBatchedResponse},
+    http::{BatchedResponse, BatchedResponseV1},
     prelude::*,
     query::{ForwardCursor, Pagination, Sorting},
 };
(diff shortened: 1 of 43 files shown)


Arjentix (approved):
Btw, I think the next step of implementing opaque versioning should look like this:

```rust
pub enum SomeData {
    V1(V1SomeData),
    V2(V2SomeData),
}

struct V1SomeData {
    i: i32,
    s: String,
}

struct V2SomeData {
    s: String,
    u: u32
}

impl SomeData {
    #[deprecated(`i` is deprecated, use `u` instead)] // Not sure about this tag
    pub fn i(&self) -> Option<i32> {
        match self {
            V1(v1) => Some(v1.i),
            V2(_) => None,
        }
    }
    
    pub fn s(&self) -> Option<&String> {
        match self {
            V1(v1) => Some(&v1.s),
            V2(v2) => Some(&v2.s),
        }
    }
    
    pub fn u(&self) -> Option<u32> {
        match self {
            V1(_) => None,
            V2(v2) => Some(v2.u),
        }
    }
}
```

Otherwise I don't understand how you would like to implemet opaqueness

Erigara (dismissed):
LGTM

Erigara (dismissed):

Arjentix (approved):

Arjentix (approved):

Erigara (approved):

Issue #1232, opened 2021-07-15
Add Kura benchmarks


takemiyamakoto:
Write new benchmark tests for Kura and Storage.

PR #3914, opened 2023-09-22
[fix] #1232: Introduce startup time benchmark

SamHSmith:
## Description

You can now run `scripts/test_env.py setup -b`

### Linked issue

Closes #{issue_number} 

### Checklist


core/src/block.rs
```diff
@@ -446,12 +446,18 @@ impl Revalidate for VersionedCommittedBlock {
                             .view_change_index
                             .try_into()
                             .map_err(|_| BlockRevalidationError::ViewChangeIndexTooLarge)?;
+                            
+println!("PLEBIAN view change index = {}", view_change_index);
+println!("PLEBIAN previous block hash = {}", last_committed_block.hash());
```
mversic, line 451: remove

core/src/sumeragi/main_loop.rs
```diff
@@ -805,13 +795,18 @@ fn reset_state(
     while *old_view_change_index < current_view_change_index {
         *old_view_change_index += 1;
         error!(addr=%peer_id.address, "Rotating the entire topology.");
-        current_topology.rotate_all();
```
mversic, line 808: you can now remove this method from `network_topology.rs`

core/src/sumeragi/main_loop.rs
```diff
@@ -785,6 +774,7 @@ fn reset_state(
     old_view_change_index: &mut u64,
     current_latest_block_height: u64,
     old_latest_block_height: &mut u64,
+    lastest_block: &VersionedCommittedBlock,
```
mversic, line 777: if you send both latest block and voting block, aren't some of the arguments given to this fn redundant?

core/src/block.rs
```diff
@@ -446,12 +446,18 @@ impl Revalidate for VersionedCommittedBlock {
...
+                            
                         Topology::recreate_topology(
                             &last_committed_block,
                             view_change_index,
                             new_peers,
                         )
                     };
+println!("PLEBIAN topology = {:?}", topology);
+println!("PLEBIAN new block previous block hash = {:?}", block.header.previous_block_hash);
```
mversic, line 460: also remove please

core/src/block.rs
mversic, line 451: remove

Arjentix:
> If you want times for release mode you must copy the release binary to target/debug/iroha

IMO there should be a flag in `test_env.py` to run binary from `target/release`.
Although it can be done in another PR.

configs/peer/config.json
```diff
@@ -14,10 +15,10 @@
     "PEER_ID": null,
     "BLOCK_TIME_MS": 2000,
     "TRUSTED_PEERS": null,
-    "COMMIT_TIME_LIMIT_MS": 4000,
+    "COMMIT_TIME_LIMIT_MS": 40000,
```
Arjentix, line 18: Why 10x increase...? O_o

Arjentix: Maybe we should have a separate config for tests/benches? Like we already have `iroha_test_config.json`.

I think `configs/peer/config.json` should be a reference for real world blockchains, not a range for tests/benches.

core/src/sumeragi/main_loop.rs
```diff
@@ -805,14 +798,19 @@ fn reset_state(
...
         was_commit_or_view_change = true;
     }
 
     // Reset state for the next round.
     if was_commit_or_view_change {
         *old_latest_block_height = current_latest_block_height;
 
+        *current_topology = Topology::recreate_topology(
+            &lastest_block,
+            current_view_change_index as usize,
```
Arjentix, line 810: Please, use failing conversion and `expect()`. We do that in other places and I find it more explicit

core/src/sumeragi/main_loop.rs
```diff
@@ -672,7 +657,14 @@ fn process_message_independent(
                         transactions,
                         event_recommendations,
                         view_change_index: current_view_change_index,
-                        committed_with_topology: sumeragi.current_topology.clone(),
+                        committed_with_topology: Topology::recreate_topology(
+                            &sumeragi
+                                .wsv
+                                .latest_block_ref()
+                                .expect("WSV must have blocks"),
+                            current_view_change_index as usize,
```
Arjentix, line 665: Same

scripts/test_env.py
```diff
@@ -36,20 +37,32 @@ def __init__(self, args: argparse.Namespace):
                           Please provide them in the `{args.root_dir}/configs/peer` directory")
             sys.exit(1)
         copy_or_prompt_build_bin("iroha", args.root_dir, peers_dir)
-
+
```
Arjentix, line 40: Lol, your editor adds this and my removes. But it's fine

scripts/test_env.py
```diff
@@ -18,6 +18,7 @@
 import urllib.error
 import urllib.request
 
+is_startup_benchmark = False
```
Arjentix, line 21: Can you not use global variables? It worked fine without it before so I'm sure there is a way to pass this value as a function parameter

scripts/test_env.py
```diff
@@ -36,20 +37,32 @@ def __init__(self, args: argparse.Namespace):
...
         os.environ["IROHA2_GENESIS_PATH"] = str(peers_dir.joinpath("genesis.json"))
         os.environ["IROHA_GENESIS_ACCOUNT_PUBLIC_KEY"] = self.peers[0].public_key
         os.environ["IROHA_GENESIS_ACCOUNT_PRIVATE_KEY"] = self.peers[0].private_key
+        
+        saved = self.peers[0].public_key
+        if is_startup_benchmark:
+            self.peers[0].public_key = "ed012072AED89EDB3CAA5EA203C6CED29A23CB387C9F5C48DB025E524DC508E18A8E22"
+            self.peers[1].public_key = "ed01200B06FCBBD0724DC11D25FC1A88B110B756712462DD6C0D762411E1A98B295AC8"
+            self.peers[2].public_key = "ed01203D922F3E7834F7AF20DAE2DE1AB2F48B4CF8FE95C62D372897692A64143F82C7"
+            self.peers[3].public_key = "ed01208BC55A04A1FC36A15DC63C59E8979FB7080E0937DDE11F2C3F783E9E167ABF4E"
```
Arjentix, line 53: As I can see we generate private keys randomly (without seed) using `kagami`.
So why do you reassign them here and with hardcoded values? Especially when `private_key` is not changed

scripts/test_env.py
```diff
@@ -36,20 +37,32 @@ def __init__(self, args: argparse.Namespace):
...
 
         logging.info("Generating trusted peers...")
         self.trusted_peers = []
         for peer in self.peers:
             peer_entry = {"address": f"{peer.host_ip}:{peer.p2p_port}", "public_key": peer.public_key}
             self.trusted_peers.append(json.dumps(peer_entry))
         os.environ["SUMERAGI_TRUSTED_PEERS"] = f"[{','.join(self.trusted_peers)}]"
+        
+        self.peers[0].public_key = saved
```
Arjentix, line 62: Why to save it, then change, then reload saved value?
Do some code between this assignments relies on that? If so then it's too implicit, it should be function parameters or something

scripts/test_env.py
Arjentix, line 1: Btw why do we use python script to benchmark start-up if previously we used Rust benchmarks?

Arjentix (changes requested):

config/src/iroha.rs
```diff
@@ -24,6 +24,8 @@ view! {
         pub private_key: PrivateKey,
         /// Disable coloring of the backtrace and error report on panic
         pub disable_panic_terminal_colors: bool,
+        /// Exit after initialization for startup time testing.
+        pub exit_after_init: bool,
```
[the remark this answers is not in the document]
mversic: > For example sending requests to Iroha and when it responses we send shutdown signal?
That souns like a real world observable benchmark.

could we watch the `/status` endpoint until `Status:blocks` is incremented to 1 or more

> Or we can analyze Iroha logs somehow.

I wouldn't go down this route

> Or we can introduce new off-chain event type, something like PeerStatus.

I would rather go with `/status` endpoint

scripts/test_env.py
SamHSmith: It was to get the node to accept the saved benchmark blockchain. It's not necessary now so I have removed it in the latest version of this PR.

config/src/iroha.rs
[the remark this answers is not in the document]
Arjentix: Yeah, we already have this check for 1 block in the script. I just thought that's not what we want to measure

core/src/sumeragi/main_loop.rs
```diff
@@ -241,11 +241,32 @@ impl Sumeragi {
...
 
-        let transactions: Vec<_> = genesis_network
-            .transactions
-            .into_iter()
-            .map(AcceptedTransaction::accept_genesis)
-            .collect();
+        let transactions: Vec<_> = {
+            let mut transactions: Vec<_> = genesis_network.transactions.into_iter().collect();
+
+            // Copy sumeragi config params into chain.
```
Arjentix, line 247: Interesting, but doesn't it contradict `SetParameter` instruction?

Also why this change is needed in the benchmark context?

ffi/derive/src/attr_parse/repr.rs
```diff
@@ -136,7 +136,7 @@ impl FromAttributes for Repr {
                     ),
                     Meta::List(list) => {
                         let Some(tokens) = accumulator.handle(
-                            syn2::parse2::<ReprTokens>(list.tokens.clone()).map_err(Into::into)
+                            syn2::parse2::<ReprTokens>(list.tokens.clone()).map_err(Into::into),
```
Arjentix, line 139: Looks like you used `cargo fmt` of a newer version or something, which changed a lot of file. Just wondering, no objections

scripts/test_env.py
```diff
@@ -50,7 +51,10 @@ def __init__(self, args: argparse.Namespace):
             peer_entry = {"address": f"{peer.host_ip}:{peer.p2p_port}", "public_key": peer.public_key}
             self.trusted_peers.append(json.dumps(peer_entry))
         os.environ["SUMERAGI_TRUSTED_PEERS"] = f"[{','.join(self.trusted_peers)}]"
-
+        
+        if is_startup_benchmark:
+            del self.peers[1:]
```
Arjentix, line 56: Can't you just use `1` instead of `args.n_peers` in line 41?

config/src/iroha.rs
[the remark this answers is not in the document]
SamHSmith: The reason I think this way is the best is that it allows us to easily use existing profiling tools. They are all built to measure total runtime of the program and monitor it while it's running.

cli/src/lib.rs
```diff
@@ -270,6 +270,11 @@ impl Iroha {
             block_count,
         });
 
+        if config.exit_after_init {
+            iroha_logger::error!("Exiting after init due to configuration.");
+            return Err(eyre!("Exiting after init due to configuration."));
+        }
```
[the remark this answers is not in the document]
SamHSmith: They don't run on the main thread. The thing we are interested in profiling is wsv code. The benchmark is measuring init time but the init that is important is the reading in and processing of the entire blockchain.

configs/peer/config.json
SamHSmith: This was me trying to get things to work when there were bugs. It has been removed now.

scripts/test_env.py
SamHSmith: Because this is easily extended to make use of existing profiling tools that profile running binaries. For example, when Spall( https://gravitymoth.itch.io/spall ) adds rust support we will be able to get a nanosecond accurate trace of iroha while it is running this benchmark. The same cannot be said for the current benchmarks inside rust.

core/src/sumeragi/main_loop.rs
SamHSmith: I thought that it was a bug that changing config parameters did not affect iroha, but this change has been removed.

scripts/test_env.py
SamHSmith: I would like to, but iroha refuses to start unless there are enough other peers in the network_topology. So I let it setup the full network topology and then this stops all but one node from running.

SamHSmith:
I have split off the profiler work to a seperate issue, #4083, it was taking too long due to weird issues that needed solving.

Erigara:
If i understand scope of this PR correctly (measure raw performance of application of blocks to iroha storage) we already have `apply_blocks` and `validate_blocks` benchmarks and their's oneshot equivalents which could be profiled using binary profiler.

We can extend then to actually load blocks from store or add new one, in such case our profiling data wouldn't be polluted by some irrelevant stuff + we don't need to add extra flags inside iroha itself.

config/src/iroha.rs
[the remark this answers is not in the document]
Erigara: I also would perfer using `/status` to stop iroha after initialization.

> The reason I think this way is the best is that it allows us to easily use existing profiling tools. 

I don't see how using `/status` to kill iroha prevent us from measuring it's performance using binary profiler.

SamHSmith:
> If i understand scope of this PR correctly (measure raw performance of application of blocks to iroha storage) we already have `apply_blocks` and `validate_blocks` benchmarks and their's oneshot equivalents which could be profiled using binary profiler.

The important difference are that those are synthetic benchmarks. The benchmark blockchain I have used here is obviously not representative. But we can easily swap in any customers or real world deployement blockchain and then reproducibly benchmark using that.

I'm curious how you are attatching a profiler to apply_blocks or validate_blocks? Maybe I just did not know how, but I found it impossibly difficult when I tried it. This benchmark plugs into the most universal api for profilers, "run this binary and see how long it runs for".

Erigara:
> The important difference are that those are synthetic benchmarks.

I agree, that's why i suggested to improve them by using block strore (or add new benchmark). 

> I'm curious how you are attatching a profiler to apply_blocks or validate_blocks? 

Like this: 

```bash

# build oneshot equivalent of bechmark in release mode but with debug info
RUSTFLAGS=-g cargo build --release --example apply_blocks 
# run this binary with profiler ([samply](https://github.com/mstange/samply) in my case)
samply record ./target/release/examples/apply_blocks
```

Erigara:
Actually what did you use to produce block store?

Erigara:
>  You mean that we could create a benchmark like the existing ones that would load an existing benchmark blockstore from disk?

Yes, exactly

SamHSmith:
In that case I guess we'll close this in favor of that strategy.

PR #3917, opened 2023-09-25
[refactor]: Move roles in `WSV`

### LINKED ISSUE #3681: [suggestion] Phase out `Account::roles`
### Feature request

`Account::roles` should be removed and replaced with an asset of type `AssetType::Store` linked to every account. The accompanying asset table should be created by the executor on every `Register<Account>`. Preferably the logic around permission tokens and roles would be completely removed from `iroha_core` and would reside only within executor/validator

### Motivation

* Roles are not a relevant concept for permissionless blockchains
* more modular design (less things hard coded)

### Alternative design

Instead of storing roles in `Account` we can store them in `wsv` just like we do for permission tokens. In this case we don't need another asset linked to every account. This seems like a simpler approach and might be preferred

### Who can help?

_No response_

Erigara:
## Description

Move accounts' roles into `wsv`.

This PR takes alternate design from #3681 since it might be preferable to operate on roles as first class citizens (e.g. have specific queries, isi, ...). 

### Linked issue

Closes #3681 

### Benefits

Move effective range quires over all assets, roles.

Cloning account is less expensive.


data_model/src/account.rs
```diff
@@ -104,17 +96,13 @@ pub mod model {
     pub struct Account {
         /// An Identification of the [`Account`].
         pub id: AccountId,
-        /// Assets in this [`Account`].
-        pub assets: AssetsMap,
```
mversic, line 108: why would you move assets out of the account?

data_model/src/account.rs
```diff
@@ -314,6 +245,20 @@ impl FromStr for AccountId {
     }
 }
 
+impl Ord for AccountId {
```
mversic, line 248: why custom impl?

data_model/src/asset.rs
```diff
@@ -511,6 +509,20 @@ impl FromStr for AssetId {
     }
 }
 
+impl Ord for AssetId {
```
mversic, line 512: same question, why custom impl?

data_model/src/account.rs
Erigara: To store every asset next to each other, we have different queries that iterate over assets inside particular domain.

Also in case of asset definition removal we iterate through every domain, every account inside it. It much effective to store them in one place.

mversic: ok, I get it, you want to make `Account` more shallow. I do not agree with this necessarily.
The reason for removing `Account::roles` was because it can be expressed through a "special" asset given to the account. This way there would be no difference in API between public vs private blockchains because it would be expressed through the common concept. This is what I would have preferred

data_model/src/account.rs
Erigara: I did that to enforce particular ordering without changing fields orded which will cause changes in scale coded.

Erigara: So i want account to be firstly sort by their domain and next by name. This way we assure that in `BTreeMap` accounts with same domain are next to each other.

data_model/src/asset.rs
Erigara: answered above

data_model/src/account.rs
Erigara: > The reason for removing Account::roles was because it can be expressed through a "special" asset given to the account. This way there would be no difference in API between public vs private blockchains because it would be expressed through the common concept. This is what I would have preferred

But we still have `PermissionToken` in the `data_model` (through it's content could be arbitrary), do you think it should be removed too?

How usable it will to the user? Because users will have to query assets, write to assets instead or roles.

data_model/src/account.rs
mversic: you can reorder them, we're breaking API for RC20 lts anyhow

0x009922: > to enforce particular ordering without changing fields order

The only way for SDKs now to sort data within sorted maps is to rely on how default `Ord` impl works. If there are any overrides of its behaviour, it might be impossible to implement sorting on the side of SDK (unless additional information is provided in the schema).

Erigara: > you can reorder them, we're breaking API for RC20 lts anyhow

i see, i this case reordering will be preferred solution

data_model/src/account.rs
mversic: > How usable it will to the user? Because users will have to query assets, write to assets instead or roles.

I don't see why necessarily. We can still keep the specialized API for dealing with roles (with `Grant` and `Revoke`). 
I still think `Account::roles` should be an asset of an account

> But we still have `PermissionToken` in the `data_model` (through it's content could be arbitrary), do you think it should be removed too?

I don't understand the question. If you're asking me whether I would prefer this to be an asset of account, then yes. Why? Because when a user queries `FindAccountById` it gets all the info about an account. 

Is cloning accounts something that we do extensively? I don't think so, and I don't think making accounts more shallow will bring performance benefits

mversic: the way I thought this would work is that `Roles` asset definition would be created by the Executor. If the blockchain is private, i.e. will deal with the concept of roles, otherwise not. And `Grant`/`Revoke` would delegate to mint/burn `Roles` asset

mversic: the way I thought this would work is that `Roles` asset definition would be created by the Executor. If the blockchain is private, i.e. will deal with the concept of roles, otherwise not. And `Grant`/`Revoke` would delegate to mint/burn `Roles` asset

mversic: I wouldn't mind if we used `Mint`/`Burn` terminology instead `Grant`/`Revoke`. IMO they are just overloads of the same operation. In fact, I think this is what we should do. We want Iroha to be a simple blockchain with a simple API

if we can make our ISA extendable (by providing a mechanism to define custom instructions in the Executor). We would be able to add `Grant`/`Revoke` as custom instructions

Erigara: > Is cloning accounts something that we do extensively? I don't think so, and I don't think making accounts more shallow will bring performance benefits

Currently we do quite a few wsv clones which result in cloning of all accounts, also quires like `FindAllAccounts` result in account cloning.

> and I don't think making accounts more shallow will bring performance benefits

It bring us benefits in particular context such as when we need to iterate over assets (not necessary for single account). Because if's more effective to store entities next to each other without pointer indirection. 

> I don't understand the question. If you're asking me whether I would prefer this to be an asset of account, then yes. Why? Because when a user queries FindAccountById it gets all the info about an account.

Ok, so our plans is to remove `PermissionToken` and `Role` from the codebase, so it will be concepts defined on the level of executor, right? 

> We can still keep the specialized API for dealing with roles (with Grant and Revoke).

In such case having specialized API seems contradictory to me, because we might have some special API for possibly non-existent thing (if executor writer don't bother to use permissions and roles). Also we don't have a control how exactly this concepts will be implemented (maybe executor writer would decide to store permissions/roles) in account metadata not account assets.

Erigara: > We want Iroha to be a simple yet extendable blockchain with a simple API
> if we can make our ISA extendable (by providing a mechanism to define custom instructions in the Executor). We would be able to add Grant/Revoke as custom instructions

We probably should define what `Executor` is allowed to do. Extreme would case would be that iroha is responsible for storing data + consensus. And actual interpretation of data, execution of transactions will be executor responsibility.

mversic: > Currently we do quite a few wsv clones which result in cloning of all accounts, also quires like FindAllAccounts result in account cloning.

1. if you store roles in the `wsv` cloning `wsv` will still clone all roles
2. yes, this is why we have a query `FindRolesByAccountId`. We shouldn't though, we should use filters

> Also we don't have a control how exactly this concepts will be implemented (maybe executor writer would decide to store permissions/roles) in account metadata not account assets.

IMO we should remove `Metadata` from an `Account`. It should also be just an asset of an account

mversic: > We probably should define what Executor is allowed to do. Extreme would case would be that iroha is responsible for storing data + consensus. And actual interpretation of data, execution of transactions will be executor responsibility.

It could be seen as an interpreter that transforms higher level API into RISC instructions that operate directly on the blockchain

Erigara: @mversic, let's summarize this a bit:
1. let's try to remove `Role`s
2. do you think that `Asset`s should be returned to the `Account`? (we can gather other opinion's as well)

Erigara: To elaborate on the shallow account i see it as tables in `SQL` where we have foreign key to the entity in our case `Assets` already have `AccountId` which points to the `Account`.

mversic: > @mversic, let's summarize this a bit:
> 1. let's try to remove `Role`s
> 2. do you think that `Asset`s should be returned to the `Account`? (we can gather other opinion's as well)

1. Yes, though I would prefer you make it into an asset right now, it's fine even if you make it part of wsv atm. Switch from wsv to asset will not be an API breaking change and we can do that later of it's too much work now
2. I would open an issue to discuss swallowing of accounts

Erigara: Created #3921, however scope is more broad.

Erigara:
After discussion with @mversic we decided to move `roles` from account to wsv, but otherwise leave everything as is.

core/src/lib.rs
```diff
@@ -48,6 +48,9 @@ pub type RolesMap = HashMap<RoleId, Role>;
 /// API to work with a collections of [`AccountId`] [`Permissions`] mappings.
 pub type PermissionTokensMap = HashMap<AccountId, Permissions>;
 
+/// API to work with a collections of [`AccountId`][`RoleId`] mappings.
+pub type AccountRolesSet = BTreeSet<role::RoleIdWithOwner>;
```
pesterev, line 52: Do you consider to use `BTreeMap<AccountId, BTreeSet<RoleId>>`? Could you please explain why using pairs (`RoleIdWithOwner`s) is better? I assume it has the same time complexity and doesn't require implementing `MaxMinExt`.

core/src/wsv.rs
```diff
@@ -1245,14 +1268,72 @@ impl WorldStateView {
...
+
+    use super::*;
+    use crate::role::RoleIdWithOwner;
+
+    /// Key for range quires over account for roles
+    #[derive(PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
+    pub struct RoleIdByAccount<'asset> {
+        account_id: &'asset AccountId,
+        role_id: MinMaxExt<&'asset RoleId>,
+    }
```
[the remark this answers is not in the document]
Erigara: Leftover, will fix

client/benches/torii.rs
```diff
@@ -50,7 +50,7 @@ fn query_requests(criterion: &mut Criterion) {
     let mut group = criterion.benchmark_group("query-requests");
     let domain_id: DomainId = "domain".parse().expect("Valid");
     let create_domain = RegisterBox::new(Domain::new(domain_id.clone()));
-    let account_id = AccountId::new("account".parse().expect("Valid"), domain_id.clone());
+    let account_id = AccountId::new(domain_id.clone(), "account".parse().expect("Valid"));
```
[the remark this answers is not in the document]
Erigara: I reodered fields, derived `Constrictor` switched them too.

core/src/lib.rs
```diff
@@ -101,6 +104,37 @@ pub mod handler {
...
+    use super::*;
+
+    /// [`RoleId`] with owner [`AccountId`] attached to it.
+    #[derive(
+        Debug,
+        Clone,
+        Constructor,
+        PartialEq,
+        Eq,
+        PartialOrd,
```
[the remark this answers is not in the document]
mversic: comment no, tests maybe

client/benches/torii.rs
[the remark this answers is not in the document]
mversic: I don't think there was need to reorder constructor args, but I'm fine with it both ways

core/src/lib.rs
Erigara: Yes, i've considered this option.
My primary motivation was to reduce amount of pointer indirection and thus optimize for cases where we scan through all accounts and all roles in it (e.g. when role is removed we have to remove it from every account).

primitives/src/cmpext.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+impl<T: Ord> Ord for MinMaxExt<T> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match (self, other) {
+            (Self::Value(lhs), Self::Value(rhs)) => lhs.cmp(rhs),
+            (lhs, rhs) if lhs == rhs => Ordering::Equal,
+            (Self::Min, _) | (_, Self::Max) => Ordering::Less,
+            (Self::Max, _) | (_, Self::Min) => Ordering::Greater,
```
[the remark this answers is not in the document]
Erigara: You can check clauses above where we compare equal values, so `Min.cmp(Min) == Equal`

core/src/lib.rs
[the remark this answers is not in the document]
Erigara: > I mean tests which ensures that placing different RoleIdWithOwner in BTreeSet will group them by account_id

I will add tests for comparasion of `AccountId`s, but i don't think we should test `BTreeSet`.

core/src/wsv.rs
```diff
@@ -1245,14 +1268,72 @@ impl WorldStateView {
...
+        /// Convert bound to dyn object
+        pub fn to_dyn_range_bounds(
+            (left_bound, right_bound): &'asset (Self, Self),
+        ) -> impl RangeBounds<dyn AsRoleIdByAccount + 'asset> {
+            let bounds: (
+                Bound<&'asset dyn AsRoleIdByAccount>,
+                Bound<&'asset dyn AsRoleIdByAccount>,
+            ) = (Bound::Excluded(left_bound), Bound::Excluded(right_bound));
+            bounds
+        }
```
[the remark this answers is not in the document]
Erigara: Indeed! Thanks for suggestion, code is more clear now i believe.

core/src/lib.rs
Erigara: Also i believe `MaxMinExt` might be useful in different context as well (if we will decide to implement #3921).

data_model/src/account.rs
```diff
@@ -79,10 +79,10 @@ pub mod model {
     #[getset(get = "pub")]
     #[ffi_type]
     pub struct AccountId {
-        /// [`Account`]'s name.
-        pub name: Name,
         /// [`Account`]'s [`Domain`](`crate::domain::Domain`) id.
         pub domain_id: DomainId,
+        /// [`Account`]'s name.
+        pub name: Name,
```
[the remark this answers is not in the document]
Erigara: Added test

client/benches/torii.rs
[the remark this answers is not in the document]
Erigara: Ok, let's preserve original order in constructor.

core/src/smartcontracts/isi/account.rs
```diff
@@ -380,12 +381,14 @@ pub mod isi {
                 .into_iter()
                 .map(|token| token.definition_id);
 
-            wsv.account_mut(&account_id).and_then(|account| {
-                if !account.remove_role(&role_id) {
-                    return Err(FindError::Role(role_id.clone()));
-                }
-                Ok(())
-            })?;
+            // TODO: clone could be avoided through `borrow`
```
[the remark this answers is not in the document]
Erigara: Implemented.

core/src/lib.rs
Arjentix: But the downside of this approach is memory footprint. I think we should start considering string interning (or string pooling)

primitives/src/cmpext.rs
```diff
@@ -0,0 +1,144 @@
...
+        pub trait $trait {
+            /// Extract key
+            fn as_key(&self) -> $key;
+        }
+
+        impl $trait for $key {
+            fn as_key(&self) -> $key {
+                *self
+            }
+        }
```
Arjentix, line 83: I still don't understand how does this work... How can you create a movable value from a reference... It should be possible only if `$key` implements `Copy`

Arjentix: Oh, I see, your `RoleIdByAccount` actualy derives `Copy`

Arjentix: Was it intenional that this macro works properly only with `$key` implementing `Copy`?

core/src/lib.rs
```diff
@@ -114,3 +173,47 @@ pub mod prelude {
         wsv::{World, WorldStateView},
     };
 }
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
Erigara: You mean create `role` module in `iroha_core`? Because if you are taking about `iroha_data_model` it's opposite of what i did before.

 https://github.com/hyperledger/iroha/pull/3917#discussion_r1335508283

core/src/lib.rs
Erigara: > But the downside of this approach is the memory footprint. I think we should start considering string interning (or string pooling)

Yeah, about string pooling, if `domain` and `account` are small enough (less 16 bytes each) they will be inlined and no allocation will happen.

primitives/src/cmpext.rs
Erigara: > Was it intenional that this macro works properly only with $key implementing Copy?

Yeah, key expected to be light weighted value consisting of references or values which can be copied.

core/src/lib.rs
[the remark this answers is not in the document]
Arjentix: Yes, I mean `role` module in the same file, so it's `iroha_core`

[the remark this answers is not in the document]
Erigara: Do we nest `tests` modules? I haven't seen this before through codebase.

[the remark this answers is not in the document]
Arjentix: Hm... Well I just think, that keeping tests closer to the code they test, is a good idea

Arjentix (approved):

mversic (approved):

Issue #3612, opened 2023-06-15
[suggestion] Make illegal instructions unrepresentable


appetrosyan:
### Feature request

Some combinations of instructions are not legal and never will be, for example `Mint<String>`. As such we may be able to constrain the expression language to prevent illegal values to be representable and as such stop them from being decoded. 

### Motivation

This was not previously possible because of lack of GATs in stable Rust. Now that the support had been stabilised, we can express the correct constraints in the trait bounds. And perhaps refactor the `Expression` and `Instruction` systems to be more strict. 

One suggestion is to rethink the expressions as not just something that generates data for an instruction, but an entire transaction. So instead of 
```rust
pub enum Executable {
    Instruction(Vec<InstructionBox>), 
    WASM(WasmBlob), 
}

// Where instructions have a nested expressionBox for every value
```
we can have 

```rust
pub enum Transaction {
    Register(RegisterBox),
    Mint(MintBox), 
    Burn(BurnBox), 
    Transfer(TransferBox),
    // ... 
    Expression(ExpressionBox), 
    Wasm(WasmBlob),
}
```

This has the advantage of allowing us to use an optimised code path for common instructions. Removing the redundant top-level `Expression::Sequence`. Remove the need for `Expression::Raw`. Allow injecting the expected type information into the expression and fail eagerly. 

More importantly, it allows us to clean out arithmetic from string-related code, string-related code from numerical asset logic, booleans from almost all places, and potentially inline queries whose type is known ahead of time. This is important, because we will then not have to deal with `FindAllAccounts` as an argument to `Transfer`. Finally, this trivialises #2409. 

This has one significant drawback and I propose only considering it, if we are certain that the benefits outweigh the costs. 

### Who can help?

@

mversic:
> Remove the need for `Expression::Raw`. Allow injecting the expected type information into the expression and fail eagerly.

If I understand you correctly, you're saying that we can have our expression system inherently strongly typed. Since we kinda already have this with `EvaluatesTo<T>` not much should change from the API standpoint but we would get more optimized underlying representation of expression, faster deserialization, etc. This would entail removing `EvaluatesTo<T>` and using `Expression<T>` instead of `Expression`. We can do this because **we know type expression will evaluate to when constructing the expression**. Some parsers don't do it this way, namely Rust `syn` library. I'd like to hear why some parsers choose to be strongly typed and some don't 

> potentially inline queries whose type is known ahead of time

but every query output type is known ahead of time?

[...]

PR #3930, opened 2023-09-27
[refactor]: Rename ISI from *Box to *Expr

mversic:
## Description

1. This was a complete misnomer because `Box` suffix if for enums used to group objects for serialization purpose
2. This is an interlude into another PR where I will introduce proper `Box`ed instructions which will not contain expressions inside them, i.e. they will be evaluated instructions
3. I'm only doing this rename in separate PR first so that changes are not overwhelming to review

### Linked issue

Closes #{issue_number} 

### Checklist


Arjentix:
> This is an interlude into another PR where I will introduce proper Boxed instructions which will not contain expressions inside them, i.e. they will be evaluated instructions

So do you mean that non existing instructions will be unrepresentable?

mversic:
> So do you mean that non existing instructions will be unrepresentable?

what do you mean by a non existing instruction?

Arjentix (approved):
I think clicking `Viewed` was harder than searh-n-replace done in this PR 😄

Arjentix:
> what do you mean by a non existing instruction?

Like `Register<u32>` or something like that. I think we have an issue for that: https://github.com/hyperledger/iroha/issues/3612

mversic:
> Like `Register<u32>` or something like that. I think we have an issue for that: #3612

oh, I get what you mean. The answer is kinda yes. Atm, we evaluate expressions twice redundantly, once in executor and then again on the host side. The idea is to move Expression evaluation into executor (and only Executor). What will be representable in the Expression model implemented in Executor is the responsibility of a particular expression implementation. Iroha will offer a RISC API that expression can evaluate into. In that sense expression system will be extandable by writing a custom executor

Erigara (approved):

Arjentix (approved):

PR #3933, opened 2023-09-28
[fix] #3928: Make failing the wasm tests actually fail the CI

### LINKED ISSUE #3928: [BUG] failed wasm tests do not fail the CI
For some reason (maybe because of `webassembly_test`), failed tests do not make the CI fail, making it useless ([example of a job that should've failed](https://github.com/hyperledger/iroha/actions/runs/6308478359/job/17126693246))

DCNick3:
## Description

Add a new `iroha_wasm_test_runner` tool. It mostly duplicates the `webassembly-test-runner`, but actually returns a non-zero exit code if there are any test failures. As an added bonus 

This tool is expected to be installed on the developer's system with `cargo install`. It's not going to change often, so I don't think there's a risk of it going out-of-sync with the state of the repo.

Also fix the `evaluate_expression` test, which was never passing before due to a typo.

### Linked issue

Closes #3928 

### Benefits

- Actually makes the CI check that the wasm test

### Alternatives

An alternative would be to continue to use the upstream test runner when/if https://github.com/matklad/webassembly-test/pull/1 gets merged, but I wouldn't bet on it.

### Checklist

- [x] make sure CI passes

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -23,7 +23,6 @@ RUN rustup component add llvm-tools-preview clippy
 RUN rustup component add rust-src
 RUN rustup component add rustfmt
 RUN rustup target add wasm32-unknown-unknown
-RUN cargo install webassembly-test-runner
 RUN cargo install cargo-llvm-cov
 
 # TODO: Figure out a way to pull in libgit2, which doesn't crash if this useless variable is gone.
--- a/tools/wasm_test_runner/src/main.rs
+++ b/tools/wasm_test_runner/src/main.rs
@@ -1,0 +1,82 @@
+//! A tool to run `WebAssembly` tests
+//!
+//! This copies functionality of `webassembly-test-runner`, but with an ability to indicate failure with an exit code.
+
+use std::process::ExitCode;
+
+use anyhow::{bail, Result};
+use wasmtime::{Engine, Instance, Module, Store};
+
+struct TestMeta<'a> {
+    name: &'a str,
+    ignore: bool,
+}
+
+fn main() -> Result<ExitCode> {
+    let argv0 = std::env::args().next().unwrap();
+
+    let file = match std::env::args().nth(1) {
+        Some(it) => it,
+        None => {
+            bail!("usage: {} tests.wasm", argv0);
+        }
+    };
+    // Modules can be compiled through either the text or binary format
+    let engine = Engine::default();
+    let module = Module::from_file(&engine, &file)?;
+    let mut tests = Vec::new();
+    for export in module.exports() {
+        if let Some(name) = export.name().strip_prefix("$webassembly-test$") {
+            let mut ignore = true;
+            let name = name.strip_prefix("ignore$").unwrap_or_else(|| {
+                ignore = false;
+                name
+            });
+            tests.push((export, TestMeta { name, ignore }));
+        }
+    }
+    let total = tests.len();
+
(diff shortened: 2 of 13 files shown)


mversic (dismissed):
I see you also opened a PR to `webassembly-test-runner` repo. Do you think we should use it when your fix is merged?

DCNick3:
Yeah, I think it's fine to use it when/if it'll get merged. But, AFAIK, matklad is not doing any rust stuff recently, so it might be a while.

DCNick3:
The `iroha_wasm::tests::execute_instruction` test fails when `CARGO_INCREMENTAL` is `0`, but passes when it's `1`. I am not sure how to even approach debugging this...

mversic:
@Arjentix can you help Nikita debug this?

Arjentix:
Will try

DCNick3:
I have some progress, but not much: it's probably a result of some unsoundness, or a compiler bug, as modifying the code (like inserting a call to a log function) can make the failure disappear. It also seems to occur in release regardless of whether the incremental compilation is used

DCNick3:
It's a double free caused by the confusion abt whether the function should take ownership.

While debugging this, I've added ad-hoc logging capabilities to the test runner. Not sure if this logger should stay. (It would allow us to move back to the upstream `webassembly-test-runner`, as my PR was merged)

DCNick3:
Made another PR to the upstream (for printing the backtrace for test failures). If it gets merged I'll get rid of the `iroha_wasm_test_runner`.

mversic:
> Removed the ad-hoc logging system & made another PR to the upstream (for printing the backtrace of test failures). If it gets merged I'll get rid of the `iroha_wasm_test_runner`.

ok, since this PR is unlikely to conflict with any of the current work we can wait a few days before merging

6r1d (dismissed):
LGTM, thanks for a detailed explanation, hopefully we can rebuild the image soon :-)

Dockerfile.build
```diff
@@ -12,7 +12,7 @@ RUN rustup component add llvm-tools-preview clippy
 RUN rustup component add rust-src
 RUN rustup component add rustfmt
 RUN rustup target add wasm32-unknown-unknown
-RUN cargo install webassembly-test-runner
+RUN cargo install --git https://github.com/matklad/webassembly-test.git webassembly-test-runner
```
mversic, line 15: could we request a new version of `webassembly-test-runner` to be released?

DCNick3:
Reverted to hosting the wasm runner tool in monorepo, as the upstream would still be effectively only maintained by us

mversic (approved):

tools/wasm_test_runner/src/main.rs
```diff
@@ -0,0 +1,82 @@
+//! A tool to run `WebAssembly` tests
+//!
+//! This copies functionality of `webassembly-test-runner`, but with an ability to indicate failure with an exit code.
+
+use std::process::ExitCode;
+
+use anyhow::{bail, Result};
```
pesterev, line 7: Have you considered to use `wasmtime::Result` and not to depend on `anyhow` in this case?

pesterev (approved):

tools/wasm_test_runner/src/main.rs
DCNick3: `wasmtime::Result` is pricesely the `anyhow::Result`, so I don't think it's beneficial

PR #3935, opened 2023-09-28
[refactor] #3934: Migrate `iroha_wasm_derive` and `iroha_validator_derive` to syn 2.0

### LINKED ISSUE #3934: Migrate `iroha_smart_contract_derive`, `iroha_executor_derive` and `iroha_trigger_derive` to syn 2.0
Part of https://github.com/hyperledger/iroha/issues/3682

Do the same work that was done in https://github.com/hyperledger/iroha/pull/3727 and https://github.com/hyperledger/iroha/pull/3759 and https://github.com/hyperledger/iroha/issues/3814 , but for `iroha_wasm_derive` and `iroha_validator_derive`

DCNick3:
## Description

Use syn 2.0, manyhow and darling to enhance the error messages

### Linked issue

Closes #3934 

### Checklist

- [ ] make sure the CI passes
- [ ] add tests?


Arjentix:
@DCNick3 , any plans to update this? Sorry for pressure, just having a lot of openned PRs makes me feel frustrated.

Also seems like something is wrong, because there are too many lines changed...

Arjentix:
@DCNick3 , any plans to update this? Sorry for pressure, just having a lot of openned PRs makes me feel frustrated.

Also seems like something is wrong, because there are too many lines changed...

DCNick3:
Yes, I am planning to get back to this after I finish the ursa replacement.

The "lots of line changed" probably comes from the history rewrite that happened not so long ago. Ig I'll hate to do some git-fu to make it point to the new commits

mversic:
> Also seems like something is wrong, because there are too many lines changed...

it's because we rewrote history to fix DCO

smart_contract/trigger/derive/Cargo.toml
```diff
@@ -15,6 +15,11 @@ workspace = true
 proc-macro = true
 
 [dependencies]
-syn.workspace = true
-quote.workspace = true
-proc-macro2.workspace = true
+iroha_macro_utils = { workspace = true }
+
+syn = { workspace = true }
```
mversic, line 20: do we still need `syn`?

DCNick3: No I don't, fixed now

smart_contract/derive/Cargo.toml
```diff
@@ -14,6 +14,9 @@ workspace = true
 proc-macro = true
 
 [dependencies]
-syn.workspace = true
-quote.workspace = true
-proc-macro2.workspace = true
+iroha_macro_utils = { workspace = true }
+
+syn2 = { workspace = true }
```
pesterev, line 19: Should we name it `syn` instead of `syn2`, if we already don't support `syn` version 1 in this crate? Or we going to rename it after migrating in all crates?

DCNick3: Is there a way to rename a workspace dependency? If there is, I think that it would make sense to use it here

0x009922: Using [`package`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml) won't work?

smart_contract/executor/derive/src/lib.rs
```diff
@@ -46,9 +43,25 @@ mod validate;
...
+#[manyhow]
 #[proc_macro_attribute]
 pub fn entrypoint(attr: TokenStream, item: TokenStream) -> TokenStream {
-    entrypoint::impl_entrypoint(attr, item)
+    let mut emitter = Emitter::new();
+
+    if !attr.is_empty() {
+        emit!(
+            emitter,
+            "`#[entrypoint]` macro for Validator entrypoints accepts no attributes"
```
0x009922, line 54:
```suggestion
            "`#[entrypoint]` macro for Executor entrypoints accepts no attributes"
```

smart_contract/executor/derive/src/validate.rs
```diff
@@ -37,119 +36,136 @@ enum ValidateAttribute {
     },
 }
 
-impl ValidateAttribute {
-    fn from_attributes<'attr, A>(attributes: A) -> Self
-    where
-        A: IntoIterator<Item = &'attr Attribute>,
-    {
+impl FromAttributes for ValidateAttribute {
+    #[allow(clippy::or_fun_call)] // nope, this is what I actually meant
```
0x009922, line 40: Not sure why does this lint need to be disabled?

smart_contract/derive/Cargo.toml
DCNick3: Nope, renaming deps inherited from the workspace doesn't seem to be implemented as of now https://github.com/rust-lang/cargo/issues/12546

smart_contract/executor/derive/src/validate.rs
DCNick3: Updated with a comment, duplicating it here:

```rust
    // we use `Option::or` to select the first specified condition in case of duplicates
    // but we still _want_ to validate that each attribute parses successfully
    // this is to ensure that we provide the user with as much validation as possible, instead of bailing out early
    // `Option::or_else` would NOT work here, as it would not validate conditions after the first valid one
```

6r1d:
Read The Docs check should be fixed by https://github.com/hyperledger/iroha/pull/4056

0x009922 (dismissed):

0x009922 (approved):

VAmuzing (approved):

0x009922 (approved):

VAmuzing (approved):

PR #3940, opened 2023-10-02
[refactor] #3640: place permission tokens in a separate module

### LINKED ISSUE #3640: Define permission tokens in a separate crate
Just an idea. We can make *Validator* to be a library too. That way we don't have to use strings like that.

From the other side this is not the way we wan't users to interact with our system and it seems like a thing only for us. But maybe there are some benefits for users too?

_Originally posted by @Arjentix in https://github.com/hyperledger/iroha/pull/3625#discussion_r1237658928_

mversic:
## Description

* moved all tokens to a separate module 
I'll still explore the way of how to make them in a separate crate
* optimized permission execution

### Linked issue

Closes #3640 

### Checklist


wasm/validator/src/tokens.rs
Erigara, line 1: Why tokens defined here and not in `default_validator`?

Erigara (dismissed):

wasm/validator/src/tokens.rs
mversic: because atm it wasn't possible to do so. I would hope that we would be able to separate them into a crate that can be imported by both Executor and `iroha_client`

Erigara (dismissed):

smart_contract/executor/src/tokens.rs
Arjentix, line 1: These tokens are still the default ones. So I think it should be `default/tokens.rs`

smart_contract/executor/src/tokens.rs
```diff
@@ -0,0 +1,478 @@
...
+///     use iroha_schema::IntoSchema;
+///     use iroha_executor_derive::{Token, ValidateGrantRevoke};
+///     use serde::{Deserialize, Serialize};
+///
+///     #[derive(Clone, PartialEq, Deserialize, Serialize, IntoSchema, Token, ValidateGrantRevoke)]
+///     #[validate(iroha_executor::permission::OnlyGenesis)]
+///     pub struct MyToken;
+/// }
+/// ```
+macro_rules! declare_tokens {
```
Arjentix, line 29: I think this might be usefull inside `lib.rs` so that users might also use it without affiliation with default tokens

smart_contract/executor/src/tokens.rs
```diff
@@ -0,0 +1,478 @@
...
+    "CanRegisterAssetsWithDefinition" => crate::tokens::asset::CanRegisterAssetsWithDefinition,
+    "CanUnregisterAssetsWithDefinition" => crate::tokens::asset::CanUnregisterAssetsWithDefinition,
+    "CanUnregisterUserAsset" => crate::tokens::asset::CanUnregisterUserAsset,
+    "CanBurnAssetsWithDefinition" => crate::tokens::asset::CanBurnAssetsWithDefinition,
+    "CanBurnUserAsset" => crate::tokens::asset::CanBurnUserAsset,
+    "CanMintAssetsWithDefinition" => crate::tokens::asset::CanMintAssetsWithDefinition,
+    "CanTransferAssetsWithDefinition" => crate::tokens::asset::CanTransferAssetsWithDefinition,
+    "CanTransferUserAsset" => crate::tokens::asset::CanTransferUserAsset,
+    "CanSetKeyValueInUserAsset" => crate::tokens::asset::CanSetKeyValueInUserAsset,
+    "CanRemoveKeyValueInUserAsset" => crate::tokens::asset::CanRemoveKeyValueInUserAsset,
```
Arjentix, line 97: Looks like this repetition could be easily handled inside macro

mversic: I don't know how

Arjentix: The usage could be something like:
```rust
declare_tokens! {
    crate::tokens::asset::{
        CanRegisterAssetsWithDefinition,
        CanUnregisterAssetsWithDefinition,
        // ...
    };
    crate::tokens::account::{
        // ...
    };
    // ...
}
```

And for macro implementation you can use `stringify!`

mversic: check out now, this is the best I was able to come up with

Arjentix (approved):

smart_contract/executor/src/tokens.rs
Arjentix: I see, it's hard because you need to use interpolation inside inner macro

Erigara (approved):

PR #3944, opened 2023-10-02
[refactor]: Include smart contract code into the workspace

mversic:
## Description

* renamed `wasm` crate to `smart_contract` (`wasm` is just one of potentially many VMs)
* included `smart_contract` crate in the workspace

### Linked issue

Closes #{issue_number} 

### Benefits

we no longer have a split in the workspace.

### Downsides

writer of the smart contract now must explicitly include allocator

### Checklist


DCNick3 (commented):
Looks reasonable! I see that you also tried to move all wasm-specific changes from `data_model` into `core::smartcontracts::wasm`, which, ig, is in preparation of introducing other on-chain scripting technologies?

Is the requirement of explicit allocator a technical limitation, or just decision to give move freedom to smart contract authors?

mversic:
> Looks reasonable! I see that you also tried to move all wasm-specific changes from `data_model` into `core::smartcontracts::wasm`, which, ig, is in preparation of introducing other on-chain scripting technologies?

I moved them because including `iroha_data_model` as dependency of smart contract proc macro crates was failing to link due to some outdated dependency of `ursa`. We're unfortunately still dependent of `ursa` which has reached EOL

> Is the requirement of explicit allocator a technical limitation, or just decision to give move freedom to smart contract authors?

I don't quite understand what you're asking here. If a user is in `no_std` environment they have to define allocator. Now they can choose which one they want to use

data_model/src/name.rs
```diff
@@ -127,15 +127,15 @@ impl<'de> Deserialize<'de> for Name {
 
         let name = ConstString::deserialize(deserializer)?;
         Self::validate_str(&name)
-            .map(|_| Self(name))
```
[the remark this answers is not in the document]
mversic: I'm not sure what you wanted. I hope I made the right change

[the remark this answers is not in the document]
DCNick3: This is not what I meant, but I like what you did even more =)

DCNick3 (dismissed):

DCNick3 (approved):

smart_contract/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+        unsafe {
+            decode_with_length_prefix_from_raw(encode_and_execute(
+                &isi_box,
+                host_execute_instruction,
+            ))
+        }
+    }
+}
+
+// TODO: Remove the Clone bound. It can be done by custom serialization/deserialization to QueryBox
```
ilchu, line 84: I understand that this and other todos around what used to be the `wasm` crate hail from the expression validator PR, but I didn't see any issues created for it AFAICT. Aren't they somewhat forgotten?

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -9,13 +8,15 @@ extern crate panic_halt;
 use alloc::{format, string::ToString};
 
 use iroha_trigger::prelude::*;
+use lol_alloc::{FreeListAllocator, LockedAllocator};
+
+#[global_allocator]
+static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
```
ilchu, line 14: While I appreciate the ~5 times less memory overhead, I am a bit uncertain if it's as well tested as the default allocator, and that's what the author warns about as well. No big objection about introducing it, though.

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -23,23 +22,6 @@ env:
...
-
-      - name: Format
-        run: cargo fmt --all -- --check
-      - name: Lints
-        if: always()
-        run: cargo clippy -Zlints --workspace --benches --tests --examples --quiet
-      - name: Documentation
-        if: always()
-        run: cargo doc --no-deps --quiet
-
```
ilchu, line 42: Why do we choose to stop checking lints for the wasm related code here?

ilchu (approved):
LGTM.

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
DCNick3: Seems to work better than now-unmaintained and leaky `wee_alloc` at least..

.github/workflows/iroha2-dev-pr-wasm.yaml
mversic: we don't, it's part of the workspace now, it should be linted in `iroha2-dev-pr-static.yml`

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
mversic: > While I appreciate the ~5 times less memory overhead, I am a bit uncertain if it's as well tested as the default allocator, and that's what the author warns about as well. No big objection about introducing it, though.

I didn't introduce it in this PR, it was introduced by @Erigara to mitigate memory leak in `wee_alloc`. Having said that, I agree with you, and after this PR is merged users will be given freedom to choose allocator, i.e. they will no longer be pinned down to allocator chosen by us

smart_contract/src/lib.rs
mversic: opened #3972

PR #3949, opened 2023-10-03
[refactor]: move expression len out of public API into core

mversic:
## Description

let's not pollute public API with internal stuff

### Linked issue

Closes #{issue_number} 

### Checklist

### DIFF
--- a/core/src/tx.rs
+++ b/core/src/tx.rs
@@ -44,7 +44,151 @@ pub enum AcceptTransactionFail {
     UnexpectedGenesisAccountSignature,
 }
 
+mod len {
+    use iroha_data_model::{expression::*, query::QueryBox, Value};
+
+    pub trait ExprLen {
+        fn len(&self) -> usize;
+    }
+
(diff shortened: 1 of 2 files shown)


Erigara:
@mversic we use `len` to check limit in `max_instruction_number` `TransactionLimits`. How user can evaluate it's transaction limits if `len` is moved into core?

mversic:
> @mversic we use `len` to check limit in `max_instruction_number` `TransactionLimits`. How user can evaluate it's transaction limits if `len` is moved into core?

they already cannot do it because the code for `instruction_size` is in the core since #3949, I just moved the remainder. Until I see a request for such a feature in the client, I wouldn't like to keep it in the public API. Public API can always be expanded, but not reduced. Also, in the new ISA expressions will be evaluated in Executor

That being said, `max_instruction_number` sounds like a misnomer since it tracks expressions

Erigara (approved):

DCNick3 (approved):

PR #3956, opened 2023-10-04
[refactor] #3950: Merge API and Telemetry endpoints into a single server

### LINKED ISSUE #3950: Merge API and Telemetry endpoints into a single server
## Description

Iroha has two set of endpoints, with two separate servers listening on two separate addresses. One set, "API", is responsible for main endpoints like transaction, queries, health check etc. Another one, "Telemetry", is responsible for status, metrics, and api version. The API server runs in any case, while the Telemetry one is started conditionally. They are configured by the required `torii.api_url` field, and by the optional `torii.telemetry_url`[^1] field.

IMO, this design choice is questionable for a numerous reasons:

- Separation doesn't seem to have any runtime/administration benefit, maybe even the opposite
- There is no paths overlap between routes
- It overcomplicates code of Torii, clients implementation, documentation and configuration

In general, I don't see any benefits of the separation. Therefore, I propose to use a single gateway for all endpoints, configurable with `torii.address` parameter.

[^1]: Both are actually *addresses* in a form of `localhost:8080`

DCNick3:
## Description

This PR merges the Telemetry endpoint into the main API endpoint

### Linked issue

Closes #3950

### Benefits

Less complexity, no more threading multiple port numbers to configuration

### Checklist

- [x] check that the endpoints actually work (I don't think they are covered by any tests?)
- [x] make ci pass


mversic (approved):

0x009922 (approved):
I would also like to revise the paths themselves:

- `/status` → `/telemetry/status`
- `/metrics` → `/telemetry/metrics`

mversic:
> I would also like to revise the paths themselves:
> 
>     * `/status` → `/telemetry/status`
> 
>     * `/metrics` → `/telemetry/metrics`

opened  #3959

PR #3958, opened 2023-10-04
[feature] #3929: Lazy queries inside wasm

### LINKED ISSUE #3929: Lazy queries inside WASM
We support lazy (kind-of) queries for basic clients, but it's not supported for WASM. 

This creates some problems. For example because of that we cannot move our queries execution into Executor causing stragne divergence: transctions and instructions are executed, but queries are not.

Need to implement lazy evaluation for WASM as well

Arjentix:
## Description

- Live query store moved from Torii to Core
- Now wasm can use lazy queries
- Basic clients should send either Query with pagination and sorting or cursor. Not everything together.
- Reduced `iroha::Configuration` stack memory footprint 

### Linked issue

- Closes #3929 

### Checklist


cli/src/torii/routing.rs
```diff
@@ -81,122 +79,31 @@ async fn handle_instructions(
...
+                    fetch_size,
+                    sorting,
+                    pagination,
+                    cursor,
+                },
+                wsv,
+            )
+        })
+        .map(Scale)
+        .map_err(Into::into)
```
mversic, line 106: Please rename `handle_query` to `handle_query_result`

I was just talking to @Erigara the other day how this function only uses `wsv` to execute the query. Could we make it like this:
```suggestion
    sumeragi
        .apply_wsv(|wsv| valid_request.execute(wsv).map_err(ValidationFail::from))
        .map(|query_result| {
            live_query_store.handle_query_result(
                WithQueryParameters {
                    query,
                    fetch_size,
                    sorting,
                    pagination,
                    cursor,
                },
                wsv,
            )
        })
        .map(Scale)
        .map_err(Into::into)
```

core/src/smartcontracts/wasm.rs
```diff
@@ -543,11 +584,18 @@ pub mod state {
                 self.0.wsv
             }
         }
+
+        impl QueryHandle for Migrate<'_> {
```
mversic, line 588: is it necessary to support lazy queries in Migrate?

core/src/query/store.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    /// Execute query performing sorting and pagination without any validation.
+    ///
+    /// Usable for queries executed by validator.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::ConnectionClosed`] if [`QueryService`] is dropped,
+    /// - Otherwise throws up query handling errors.
+    pub fn handle_query_unchecked(
```
mversic, line 189: what is `unchecked`? You say that it doesn't validate query? Do you mean it doesn't go thgrough executor validation?

mversic (changes requested):
this is the first time we're effectively breaking sandboxing by allowing user submitted smart contracts to hold references to host resources. Let's think about security aspect of that

core/src/query/store.rs
Arjentix: Yes, and other checks like signature check as well, because it's just a `QueryBox` not `SignedQuery`

core/src/smartcontracts/wasm.rs
Arjentix: Why not? Users may want to perform queries from `migrate()` the same way they might want to do it in basic smart contracts

Arjentix:
> this is the first time we're effectively breaking sandboxing by allowing user submitted smart contracts to hold references to host resources. Let's think about security aspect of that

What security risks do you see?

I can't see any new problems for now. All users can do with wasm they already can do with a simple client.
Also keep in mind that our wasm smart contracts and etc. are one-shot and stateless

cli/src/torii/routing.rs
mversic: I propose to construct API in such a way that `apply_wsv` takes as little time as possible to execute. In this case I would just like to extract `valid_request.execute(wsv)` out of `handle_query`

Arjentix: Ok, I see, makes sense. I'll do that

core/src/smartcontracts/wasm.rs
```diff
@@ -1348,17 +1438,18 @@ impl<'wrld> import_traits::SetPermissionTokenSchema<state::validator::Migrate<'w
...
+    query_handle: LiveQueryStoreHandle,
     engine: Option<Engine>,
     config: Option<Configuration>,
     linker: Option<Linker<S>>,
 }
 
 impl<S> RuntimeBuilder<S> {
     /// Creates a new [`RuntimeBuilder`]
-    pub fn new() -> Self {
+    pub fn new(query_handle: LiveQueryStoreHandle) -> Self {
```
mversic, line 1450: why is query store given from the outside? Does this mean that query store is shared between different smart contracts?

core/src/query/store.rs
mversic: it just seems to me like you are abusing the suffix `_unchecked` here

Arjentix: Haha, maybe. Ok, anyway, after your suggestion [here](https://github.com/hyperledger/iroha/pull/3958#discussion_r1347000475) I will move away validation and execution from handle

core/src/smartcontracts/wasm.rs
Arjentix: This is just a handle to a one service running on host on the background. So any smart contract can talk to it, yes

mversic: then you should make sure that you have deleted all live queries after smart contract exits (whether successfully or with error)

core/src/smartcontracts/isi/query.rs
```diff
@@ -56,9 +56,9 @@ impl_lazy! {
 }
 
 /// Query Request statefully validated on the Iroha node side.
-#[derive(Debug, Decode, Encode)]
+#[derive(Debug, Clone, Decode, Encode)]
 #[repr(transparent)]
-pub struct ValidQueryRequest(SignedQuery);
+pub struct ValidQueryRequest(pub SignedQuery);
```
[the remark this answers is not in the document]
Arjentix: Indeed, I just needed a getter, but created a way to break our state machine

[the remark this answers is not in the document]
mversic: try to avoid a getter, that is precisely what would break state machine. Maybe you can implement `From<SignedQuery> for ValidQueryRequest` or have a method on `ValidQueryRequest` that does the job internally?

core/benches/kura.rs
```diff
@@ -42,7 +43,8 @@ async fn measure_block_size_for_n_validators(n_validators: u32) {
         iroha_core::kura::Kura::new(iroha_config::kura::Mode::Strict, dir.path(), false).unwrap();
     let _thread_handle = iroha_core::kura::Kura::start(kura.clone());
 
-    let mut wsv = WorldStateView::new(World::new(), kura);
+    let query_handle = LiveQueryStore::test().start();
+    let mut wsv = WorldStateView::new(World::new(), kura, query_handle);
```
Erigara, line 47: Why might what to reconceptualize what `WSV` is, because currently it contains slightly related things (blocks, data base, wasm engine, query service and so on).

Don't think it should be done in this PR.

cli/src/torii/routing.rs
Arjentix: I did it the way you proposed, but `LazyQuery` returned by a query execution depends on the `wsv` lifetime, so it's not possible.

However I removed validation/execution from `LiveQueryStoreHandle`.

core/benches/kura.rs
Arjentix: Yeah, and now I decided to put `query_handle` into it. I tried keeping it separate but it's just too much pain

cli/src/torii/routing.rs
mversic: > I did it the way you proposed, but LazyQuery returned by a query execution depends on the wsv lifetime, so it's not possible.

yeah, I know. As far as I am aware this lifetime bound cannot be expressed in Rust. But still it's better you have pulled it out

data_model/src/query/cursor.rs
```diff
@@ -27,6 +34,12 @@ pub mod model {
         /// Pointer to the next element in the result set
         pub cursor: Option<NonZeroU64>,
     }
+
+    impl ForwardCursor {
+        pub const fn new(query_id: Option<String>, cursor: Option<NonZeroU64>) -> Self {
```
mversic, line 39: you can use `derive_more::Constructor`

data_model/src/query/mod.rs
```diff
@@ -1308,6 +1344,9 @@ pub mod http {
             /// Payload
             pub payload: QueryPayload,
         }
+
+        /// End type of a query http clients can send to an endpoint.
+        pub type ClientQueryRequest = QueryRequest<SignedQuery>;
```
mversic, line 1349: I don't think there is any need to keep this in data_model?

mversic: it can reside in `iroha_core` and be imported by cli

data_model/src/query/mod.rs
```diff
@@ -65,6 +64,10 @@ pub trait Query: Into<QueryBox> + seal::Sealed {
     type Output: Into<Value> + TryFrom<Value>;
 }
 
+// TODO: Could we make a variant of `Value` that holds only query results?
+/// Type representing Result of executing a query
+pub type QueryOutput = Value;
```
mversic, line 69: I believe this type can reside in `iroha_core` as well

data_model/src/query/mod.rs
```diff
@@ -165,6 +168,24 @@ pub mod model {
     )]
     #[ffi_type]
     pub struct MetadataValue(pub Value);
+
+    /// Request type clients (like http clients or wasm) can send to a query endpoint.
+    #[derive(Debug, Clone, Encode, Decode, Deserialize)]
+    pub enum QueryRequest<Q> {
```
mversic, line 174: I believe this can also reside in `iroha_core` and be imported by cli

wasm/validator/derive/src/token.rs
```diff
@@ -26,17 +26,21 @@ fn impl_token(ident: &syn::Ident, generics: &syn::Generics) -> proc_macro2::Toke
...
                         )
                     ),
                     "Failed to execute `FindPermissionTokensByAccountId` query"
-                ).try_into().unwrap();
+                );
 
-                all_account_tokens
+                account_tokens_cursor
                     .into_iter()
+                    .map(|res| ::iroha_validator::iroha_wasm::debug::DebugExpectExt::dbg_expect(
```
mversic, line 40:
```suggestion
                account_tokens_cursor
                    .map(|res| ::iroha_validator::iroha_wasm::debug::DebugExpectExt::dbg_expect(
```

isn't it already an iterator?

cli/src/torii/mod.rs
```diff
@@ -107,57 +58,52 @@ pub enum Error {
...
 impl Reply for Error {
     fn into_response(self) -> Response {
-        use Error::*;
-        match self {
-            Query(err) => {
-                reply::with_status(utils::Scale(&err), query_status_code(&err)).into_response()
-            }
-            _ => reply::with_status(Self::to_string(&self), self.status_code()).into_response(),
-        }
+        reply::with_status(Self::to_string(&self), self.status_code()).into_response()
```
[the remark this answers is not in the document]
Arjentix: That was actually one of the error in my tests, thanks

core/src/smartcontracts/isi/query.rs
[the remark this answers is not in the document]
Arjentix: This change is no logner needed, so I removed it

core/src/smartcontracts/wasm.rs
Arjentix: By the way, this code is outdated, but `migrate()` can still execute queries

core/src/smartcontracts/wasm.rs
Arjentix: That's a nice idea. Don't you mind if I implement it in another PR?

Arjentix: - Created #3982

core/src/query/store.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    fn remove(&self, query_id: String) -> Result<Option<LiveQuery>> {
+        let mut remove_channel_lock = self.remove_channel.blocking_lock();
+
+        remove_channel_lock
+            .0
+            .blocking_send(query_id)
+            .map_err(|_| Error::ConnectionClosed)?;
+
+        remove_channel_lock
```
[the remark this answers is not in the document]
Arjentix: That's interesting, never thought about that. It's like a mail return address

data_model/src/query/cursor.rs
Arjentix: At first look yes, but in fact I need this `const`, which `derive_more::Cosntructor` cannot provide. I left a comment for this function

data_model/src/query/mod.rs
Arjentix: It's used by `iroha_client` as well

data_model/src/query/mod.rs
Arjentix: I think I was not the one, who introduced it, but just moved it.

Altough I don't find this alias usefull so I removed it at all

data_model/src/query/mod.rs
Arjentix: Oh, I need to rename `WasmQueryRequest` to `SmartContractQueryRequest`

wasm/validator/derive/src/token.rs
Arjentix: Well, no, it's a type which has some other functions like `into_inner()` and `collect()`. It implements `into_iter()` only if its generic is `Vec`.

From the other side I think I could make it to be an iterator it-self, but then I'm not sure about functions I wrote above. Especialy `collect()` which has a bit another meaning than `Iterator::collect()`.

data_model/src/query/mod.rs
mversic: fine if you removed it, we're going to have to thing about relationship between queries and expression system anyhow

Cargo.toml
```diff
@@ -103,6 +103,7 @@ duct = "0.13.6"
 criterion = "0.3.6"
 proptest = "1.0.0"
 expect-test = "1.4.1"
+stacker = "0.1.15"
```
mversic, line 106: don't make it a workspace level dependency. It's only used in config

core/src/queue.rs
```diff
@@ -444,14 +450,16 @@ mod tests {
     }
 
     #[test]
```
Erigara, line 452: You didn't use `tokio::test` is that intentional?

data_model/src/lib.rs
```diff
@@ -1843,55 +1859,13 @@ pub fn current_time() -> core::time::Duration {
...
-        }
-    }
-
-    impl From<BatchedResponseV1<Self>> for QueryOutput {
-        #[inline]
-        fn from(source: BatchedResponseV1<Self>) -> Self {
-            source.batch
-        }
-    }
+declare_versioned_with_scale!(BatchedResponse<T> 1..2, Debug, Clone, iroha_macro::FromVariant, IntoSchema);
```
mversic, line 1862: unlike other structs/enums, now this is not in the `model` module. Is that ok for API?

data_model/src/query/cursor.rs
mversic: comment is not required because anyone who tries to change it will find the reason. Also, I know it `derive_more::Constructor` doesn't make `const` functions

Erigara (changes requested):

data_model/src/query/mod.rs
mversic: I would annotate it with `transparent_api` and make `ClientQueryRequest` a newtype wrapper

smart_contract/derive/src/entrypoint.rs
```diff
@@ -29,7 +29,9 @@ pub fn impl_entrypoint(_attr: TokenStream, item: TokenStream) -> TokenStream {
     block.stmts.insert(
         0,
         parse_quote!(
-            use ::iroha_wasm::{debug::DebugExpectExt as _, ExecuteOnHost as _, QueryHost as _};
+            use ::iroha_wasm::{
```
mversic, line 32:
```suggestion
            use ::iroha_smart_contract::{
```

shouldn't it be this?

mversic: weird, does none of our tests use this macro?

cli/src/lib.rs
```diff
@@ -246,13 +248,25 @@ impl Iroha {
             std::path::Path::new(&config.kura.block_store_path),
             config.kura.debug_output_new_blocks,
         )?;
+        let live_query_store_handle =
+            LiveQueryStore::from_configuration(config.live_query_store).start();
```
mversic, line 252: did you think about making query store shared between client and smart contracts? I just wonder if you have any thoughts around that

mversic, line 252: did you think about making query store shared between client and smart contracts? I just wonder if you have any thoughts around that

core/src/smartcontracts/wasm.rs
```diff
@@ -767,26 +771,36 @@ impl<S: state::LimitsMut> Runtime<S> {
...
+    ) -> Result<BatchedResponse<Value>, ValidationFail> {
+        iroha_logger::debug!(%query_request, "Executing");
 
         let wsv = state.wsv();
 
-        // NOTE: Smart contract (not executor) is trying to execute the query, validate it first
-        // TODO: Validation should be skipped when executing smart contract.
-        // There should be two steps validation and execution. First smart contract
-        // is validated and then it's executed. Here it's validating in both steps.
-        // Add a flag indicating whether smart contract is being validated or executed
```
mversic, line 779: isn't this comment still valid? should we make a ticket?

mversic: or now we're validating while executing?

mversic (changes requested):

cli/src/lib.rs
Erigara: > did you think about making query store shared between client and smart contracts?

I think it is already shared between them, isn't it?

core/src/query/store.rs
```diff
@@ -0,0 +1,279 @@
...
+        const ALL_HANDLERS_DROPPED: &str =
+            "All handler to LiveQueryStore are dropped. Shutting down...";
+
+        let (insert_sender, mut insert_receiver) = mpsc::channel(1);
+        let (remove_sender, mut remove_receiver) = mpsc::channel::<(String, oneshot::Sender<_>)>(1);
+
+        tokio::task::spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = sleep(self.query_idle_time) => {
```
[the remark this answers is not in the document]
Arjentix: Indeed, I think I just copied the old code

core/src/query/store.rs
```diff
@@ -0,0 +1,279 @@
...
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+type LiveQuery = Batched<Vec<Value>>;
+
+/// Service which stores queries which might be non fully consumed by a client.
+///
+/// Clients can handle their queries using [`LiveQueryStoreHandle`]
+#[derive(Debug)]
+pub struct LiveQueryStore {
+    queries: DashMap<String, (LiveQuery, Instant)>,
```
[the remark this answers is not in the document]
Arjentix: Makes sense, it's no longer needed

core/src/queue.rs
Arjentix: I use `tokio::test` in places where `LiveQueryStore` is needed, because it spawns a new task. And this test doesn't need `LiveQueryStore`

Arjentix: I imported `tokio::test`. If you ask why sometimes I don't import it, then it's because not all tests need `LiveQueryStore` which requires `tokio` runtime. But this test requires it and provided `#[test]` attribute fallbacks to `#[tokio::test]`

core/src/smartcontracts/wasm.rs
Arjentix: I think it's outdated.

Basic smart contracts fallback to `default_execute_query()` (you can find this associated function in `core/src/wasm.rs`). This functions does validation by calling `wasm.executor().validate_query()` and then calls `query.execute()`.

But this is not the case for validator. You can check how functions are implemented for any runtime implementing `ExecuteOperationsAsExecutorMut`. Here `execute_query()` will just call `query.execute()` without validation

core/src/queue.rs
Erigara: I missed your import.

data_model/src/query/mod.rs
Arjentix: As we decidec in DM, it doesn't need to be under `transparent_api`, `#[model]` is enough.

I implemented wrappers

data_model/src/lib.rs
Arjentix: No, looks fine to me, I used it in the same way we do for other versioned structs

smart_contract/derive/src/entrypoint.rs
Arjentix: > weird, does none of our tests use this macro?

Looks like so

Arjentix: Created #3988

Erigara (approved):

Issue #1986, opened 2022-03-21
Add trigger-related queries.


appetrosyan:
As of writing triggers can't be queried. We need to add at a minimum:
- FindTriggerById
- FindAllTriggers

Optionally also 
- FindTriggersByEvent
Which returns a list of triggers that _would_ get executed for a specific event, but doesn't actually execute them.

PR #3965, opened 2023-10-06
 [fix] #3516: make FindTrigger queries return original WASM

QuentinI:
## Description

Retain original WASM blobs in `TriggerSet`
Substitute internal representation for original when querying for Triggers
Remove internal representation from data_model

### Linked issue

Closes https://github.com/hyperledger/iroha/issues/3516

### Benefits

No more leaking internal representation
FindTrigger queries return actually useful WASM

### Checklist


core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -39,7 +40,88 @@ pub enum Error {
...
+#[derive(Clone, Debug)]
+pub struct LoadedAction<F> {
+    executable: LoadedExecutable,
+    pub repeats: Repeats,
+    pub authority: AccountId,
+    pub filter: F,
+    pub metadata: Metadata,
+}
+
+pub trait LoadedActionTrait {
```
mversic, line 52: I don't think there is any need to have this trait

data_model/src/lib.rs
```diff
@@ -1469,42 +1429,23 @@ impl TryFrom<f64> for Value {
...
+//     type Error = ErrorTryFromEnum<Value, Self>;
+//
+//     fn try_from(value: Value) -> Result<Self, Self::Error> {
+//         if let Value::Identifiable(IdentifiableBox::Trigger(trigger)) = value {
+//             return Ok(trigger);
+//         }
+//
+//         Err(Self::Error::default())
+//     }
+// }
```
mversic, line 1448: instead of commenting out everything, I would recommend you to use `unimplemented` in the body of the impl block. That way you will not accidentally forget to uncomment

data_model/src/trigger.rs
```diff
@@ -317,11 +293,8 @@ pub mod action {
     /// Trait for common methods for all [`Action`]'s
     #[cfg(feature = "transparent_api")]
     pub trait ActionTrait {
```
mversic, line 295: I don't think this trait is needed. But if it is, please move it's definition into `iroha_core`

mversic (changes requested):
looks good overall

core/src/smartcontracts/isi/triggers/set.rs
QuentinI: It's used because `triggers::Set` stores `Action`s in [specialized maps](https://github.com/hyperledger/iroha/blob/107545cf45af1d457649f1dac63b4f6f2fac02a7/core/src/smartcontracts/isi/triggers/set.rs#L134) but allows to inspect [them regardless of filter type](https://github.com/hyperledger/iroha/blob/107545cf45af1d457649f1dac63b4f6f2fac02a7/core/src/smartcontracts/isi/triggers/set.rs#L554)
I don't really understand why `Action` needs to be parametrized by filter type at all and not have an enum instead, but unless we change that, I don't think there's an easy way to get rid of this trait.

data_model/src/lib.rs
QuentinI: I meant to delete these altogether, they aren't needed anymore

core/src/smartcontracts/isi/triggers/set.rs
mversic: after you have undrafted the PR, if you can remove this trait and replace trait objects with concrete types fine. If not, also fine. It's not a big deal

QuentinI: Opened https://github.com/hyperledger/iroha/issues/3986

data_model/src/query/mod.rs
```diff
@@ -1025,15 +1025,15 @@ pub mod trigger {
     }
 
     impl Query for FindTriggerById {
```
mversic, line 1027:
```suggestion
    impl Query for FindTriggerByHash {
```

and also change it's definition to:
```rs
        pub struct FindTriggerByHash {
            /// The Identification of the trigger to be found.
            pub id: EvaluatesTo<HashOf<Trigger<FilterBox>>>,
        }
```

mversic: I believe we can also change `FindAllActiveTriggerIds` into `FindAllActiveTriggerHashes`

data_model/src/query/mod.rs
```diff
@@ -1025,15 +1025,15 @@ pub mod trigger {
...
-        type Output = Trigger<TriggeringFilterBox, OptimizedExecutable>;
+        type Output = Trigger<TriggeringFilterBox>;
     }
 
     impl Query for FindTriggerKeyValueByIdAndKey {
```
mversic, line 1031: why do we have this query?

QuentinI: I tracked it down to https://github.com/hyperledger/iroha/issues/1986, couldn't find any further motivation
I'd imagine it's what you would use if you receive `ExecuteTriggerEvent` and want to inspect the trigger for example

configs/peer/executor.wasm
mversic, line 1: wow, what happened here? the size doubled somehow

mversic (dismissed):

configs/peer/executor.wasm
QuentinI: It's a product of my environment probably, it's a bit specific.
I can't rebuild it on a usual machine, unfortunately.

mversic: hm, but wasm shouldn't be platform specific?

Arjentix: One of us can build it on our system and push to your branch

mversic: > It's a product of my environment probably, it's a bit specific.

it's true, I tested it now. That's surprising

Arjentix: Are you sure you have run `iroha_wasm_builder_cli` with `--optimize` option?

data_model/src/trigger.rs
```diff
@@ -64,62 +64,38 @@ pub mod model {
     )]
     #[display(fmt = "@@{id}")]
     #[ffi_type]
-    pub struct Trigger<F, E> {
+    pub struct Trigger<F> {
         /// [`Id`] of the [`Trigger`].
         pub id: TriggerId,
```
mversic, line 69: can you go to `TriggerId` definition and swap `name` and `domain_id`. So that `Ord` impl orders first by domain then by name

flake.nix
Arjentix, line 1: Why did you change this file? It should be in a separate meaningful commit, I believe

configs/peer/executor.wasm
QuentinI: > hm, but wasm shouldn't be platform specific?

Linker problem probably

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -507,32 +633,52 @@ impl Set {
...
+        };
+
+        let blob_hash = match event_type {
+            TriggeringEventType::Data => self
+                .data_triggers
+                .remove(id)
+                .map(|trigger| match trigger.executable {
+                    LoadedExecutable::Wasm(LoadedWasm { blob_hash, .. }) => Some(blob_hash),
+                    _ => None,
+                })
```
Arjentix, line 647: I think you can move this part to a separate function to not to repeat this code for every trigger type. Will be something like:

```rust
.map(extract_blob_hash)
```

configs/peer/executor.wasm
QuentinI: > Are you sure you have run `iroha_wasm_builder_cli` with `--optimize` option?

That is indeed the problem, thank you
I had to bypass parts of `iroha_wasm_builder` and inadvertently disabled optimization pass.

flake.nix
QuentinI: Accidental autoformat, will revert

mversic (approved):

configs/peer/executor.wasm
QuentinI: Rebuilt with optimization pass, size back to normal.

mversic (approved):

mversic (approved):

Arjentix (approved):

PR #3969, opened 2023-10-09
[ci] #3889: prune docker-compose files

### LINKED ISSUE #3889: [BUG]: `docker-compose.*.yml` for `iroha2-stable` aren't covered with CI
## Description

Recently we started to use `kagami swarm` ([now `iroha_swarm`](https://github.com/hyperledger/iroha/pull/3841)) to generate Docker Compose configurations in the repo. It was introduced in the `iroha2-dev` branch. Thus, only `docker-compose.dev.*.yml` files were covered by that.

Now `kagami swarm` is integrated into `iroha2-stable`, but CI still targets `dev` branch:

https://github.com/hyperledger/iroha/blob/77d96c8577b148df4e2f000712807dbaca4d7a8a/scripts/check.sh#L59-L61

Docker Compose configurations for `stable` branch were **not updated** and are **not checked** by CI.

### Proposed Solution

Configure CI checks that way so in `iroha2-stable` it checks how `kagami swarm` generated configurations for `stable` channel.

...Or try to get rid of confusing approach of having configurations for all 3 branches in all 3 branches. Currently `iroha2-lts`, `iroha2-dev` and `iroha2-stable` all have docker compose files for `lts`, `stable` and `dev` branches (not sure about `lts` though, haven't checked). Maybe it makes sense in each branch to have configuration related to that branch. That way, the CI will be uniform.

0x009922:
## Description

Currently each branch (`iroha2-<lts|stable|dev>`) has set of docker-compose files not only for itself, but for other branches too. It leads to confusion and goes against how versioning works. Moreover, since `2.0` release will be relatively soon and we are going to get rid of channels at all, there is no sense to keep them in the dev branch.

This PR removes docker-compose files related to other branches/channels, and removes `.dev` suffix from the configurations related to `dev` branch.

### Linked issue

Closes #3889 (will be completely closed after merging into `iroha2-stable`)

### Benefits

Less confusion, natural branch-based versioning and CI checks

### Checklist

- [ ] Add CI check to ensure usage of a proper Dockerhub image according to the branch


Erigara:
@0x009922 afaik initial reason for having multiple files was to prevent situation when new user is trying to run `lts` containers when he was on `dev` branch in git. 
Other than that i think that having 3 sets of files was confusing and this change is for good.

0x009922:
I guess we should first extend this PR with a new CI which will ensure the sync of the Dockerhub image used in the docker files, and the branch the file is located on.

6r1d (approved):

6r1d (changes requested):
Will need an update from me.

appetrosyan:
I'll remind you that the impetus for the original shift towards docker compose files for different branches on the same branch was because of community confusion, and problems caused by merging into `-main` and `-lts` branches. 

If you remove them, please consider adding something instead that fixes those issues. 

I'd advise not doing this, because even a simple docker compose generator needs to be ran by the user and those will often not work. Pulling the `docker-compose` from the tutorial might work, but I'd advise against it, because then the dependency of the tutorial on dev becomes bidirectional.

0x009922:
Closing in favor of #4053.

@appetrosyan, now we have a single set of docker-compose configuration files binded to the branch. The primary intention we have is to reduce any confusion of end users: they come to the repo on whatever branch they want, and they use instructions in the README of that branch, which are usually simply "use `docker-compose.yml`.

We have also introduced a CI check that ensures that compose files in `iroha2-dev` branch contain `image: hyperledger/iroha2:dev`, and `iroha2-stable` - `image: hyperledger/iroha2:stable`. Although the Git pipeline became a little more complicated, hopefully compose files will be always up to date.

We are also planning to add CI checks to ensure that docker compose files are not only valid textually, but that they are also actually runnable.

PR #3960, opened 2023-10-05
Update commands.go

6r1d:
There are several commits to be added.

----

removed duplicate code in goSrc/srcvmCaller/iroha/commands.go: 478

Signed-off-by: Sudhanshu <76694373+sudhanshu-k@users.noreply.github.com>## Description

### Linked issue

Closes #{issue_number} 

### Checklist

- [ ] I've read `CONTRIBUTING.md`
- [ ] I've used the standard signed-off commit format (or will squash just before merging)
- [ ] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

PR #3862, opened 2023-09-02
Trying to upgrade libraries in VCPKG

baziorek:
## Description
In PR https://github.com/hyperledger/iroha/pull/3852/ I needed to revert upgrading VCPKG. It caused libraries to downgrade. So now we need to try to upgrade libraries manually one by one.

Corrections (marked as [x] when CI passed):
- [x] iroha-ed25519 (no updates)
- [x] protobuf 3.15.8 -> 3.21.12
- [x] abseil 2021-03-24 -> 20230125.0
- [x] benchmark 1.5.2 -> 1.7.1
- [x] gflags 2.2.2-1 -> 2.2.2#5
- [x] nlohmann-json 3.9.1 -> 3.11.2
- [x] rapidjson 2020-09-14 -> 2022-06-28#3
- [x] rxcpp 4.1.0-1 -> 4.1.1#1
- [x] rocksdb 6.14.6 -> 7.9.2
- [ ] fmt 7.1.3#4 -> 9.1.0#1 - PROBLEM with upgrading library with patch
- [ ] spdlog 1.8.5#2 -> 1.11.0 - PROBLEM with upgrading library with patch
- [ ] gtest 1.10.0#4 -> 1.13.0 - PROBLEM with upgrading library with patch
- [ ] grpc 1.37.0#1 -> 1.51.1 - PROBLEM with upgrading library with patch
- [ ] soci[boost,postgresql] - PROBLEM with upgrading library with patch
- [ ] prometheus-cpp 0.12.2 -> 1.1.0 - PROBLEM with upgrading library with patch
- [ ] boost-* - PROBLEM with upgrading library with patch

### Linked issue

Closes #{issue_number} 

### Benefits
Newer packages - it will compile longer, less bugs, more security etc.

### Checklist

- [ ] I've read `CONTRIBUTING.md`
- [ ] I've used the standard signed-off commit format (or will squash just before merging)
- [ ] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

# More information:
I was using the tutorial (as inspiration): https://stackoverflow.com/questions/72588408/vcpkg-how-to-edit-package-file-when-compilation-fails-when-installing-package
1. First I've build all packages with VCPKG in newest versions (details: https://github.com/hyperledger/iroha/pull/3852)
2. Then I used older version of vcpkg (+ packages).
3. I've created repository in `./vcpkg-build`
[...]

PR #3977, opened 2023-10-09
Repairing MacOs build on develop

baziorek:
## Description
develop -> master is failing to compile in MacOs: https://github.com/hyperledger/iroha/pull/3960

That is why I need to repair the compilation. It looks that GTest is not compilling for MacOS because compiler is now newer than before (newer compiler usually is more strict). Newer compiler signals some things as warnings, but GTest has `-Werror` compilation flag. So the solution is to create patch for vcpkg port of gtest to remove the flag (how to create patch is here: https://stackoverflow.com/questions/72588408/vcpkg-how-to-edit-package-file-when-compilation-fails-when-installing-package).

Details of what was done: https://github.com/hyperledger/iroha/pull/3977#issuecomment-1771606416

### Linked issue

Closes #{issue_number} 

### Benefits
We will be able to upgrade `main` branch with changes from `develop`

### Checklist

### DIFF
--- a/.github/build-iroha1.src.yml
+++ b/.github/build-iroha1.src.yml
@@ -203,19 +203,15 @@ jobs:
             pull_request)       if   commit_message_body_build_spec FETCH_HEAD >/tmp/comment_body ;then
                                   if git_is_merge_commit FETCH_HEAD ;then
                                     echo ::warning::'/build directive in merge commit overrides default "/build before-merge"'
-                                    echo here2
                                   fi
-                                  echo here1
                                 elif commit_was_merged_build_spec   FETCH_HEAD >/tmp/comment_body ;then
-                                  echo here3
                                   true
                                 else
-                                  echo here4
                                   #echo  >/tmp/comment_body "/build debug; /build ubuntu release debug normal"
                                   echo  >/tmp/comment_body "/build ubuntu debug release normal gcc-10"$'\n' "/build macos debug"
                                 fi ;;
             push)               commit_message_body_build_spec >/tmp/comment_body || {
-                                  echo "/build macos ubuntu debug release normal gcc-10"
+                                  echo "/build ubuntu debug release normal gcc-10"
                                 } >/tmp/comment_body ;;
             workflow_dispatch)  echo >/tmp/comment_body "${{github.event.inputs.build_spec}}" ;;
             *)                  echo >&2 "::error::Unexpected event"; false ;;
--- a/.github/build-iroha1.src.yml
+++ b/.github/build-iroha1.src.yml
@@ -696,6 +696,8 @@ jobs:
       - *step_show_context
 
   build-M:
+    env:
+      CC_NAME: gcc-10
     needs:
       - prepare-macos-env
       - generate_matrixes
--- a/.github/build-iroha1.src.yml
+++ b/.github/build-iroha1.src.yml
@@ -225,6 +225,7 @@ jobs:
         id: matrixes
         run: |
           set -x
+          cat /tmp/comment_body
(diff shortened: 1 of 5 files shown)


mversic (approved):

baziorek:
It is not ready yet. I need some way to run:
image
but I unfortunately can not find the option anywhere, could somebody help?

It will be great if somebody runs also:
image
it is waiting about 12 hours:)

baziorek:
@safinsaf I hope that this PR (https://github.com/hyperledger/iroha/pull/3977) is fix for: https://github.com/hyperledger/iroha/pull/3960, but to make sure we need to enable checks:
image
image

Those checks seems to be enabled only when we want to merge something into `main` branch, but we need to enable them also for other branches. Unfortunately I'm not CI expert and I can't find where it is enabled only for main. Could You help, please?

baziorek:
It looks that everything works expect for building on MacOs:
image
+ DCO:D.   
 
When I check details about Building on MacOS I see that building dependencies is fixed (what was problem here: https://github.com/hyperledger/iroha/pull/3960):
image

Unfortunately Iroha is not building on MacOS (interesting thing is that on Linux it is building). Details of compilation failures:
image

**I see three solutions (I don't know if they will help)**:
1. Try to downgrade protobuf (it was upgraded here: https://github.com/hyperledger/iroha/pull/3862)
2. Can we change compiler on MacOs from clion to gcc10?
3. Try to fix on MacOS with current configuration - but I don't have MacOs, so I can not try that way.

baziorek:
We have problem, because it is not that easy to fix its, according to possible solutions from another post (https://github.com/hyperledger/iroha/pull/3977#issuecomment-1765862908):
1. Downgrading protobuf is not working, so I'll revert reverting commit (https://github.com/hyperledger/iroha/pull/3977/commits/fc4ebec33c5f68d9eb3835dfb07f7d59d7108c40)
2. Changing compiler into g++-10 on MacOs - didn't help.

baziorek:
I noticed that compilation problems are in iroha-lib (what was done during last year internship project): https://github.com/hyperledger/iroha/pull/3977#issuecomment-1765862908

So I decided - let's check if after commenting them everything is going to compile. Unfortunately we faced linkage error:
image
We can come back to clang compiler and maybe it will pass:)

baziorek:
I noticed that compilation problems are in iroha-lib (what was done during last year internship project): https://github.com/hyperledger/iroha/pull/3977#issuecomment-1765862908

So I decided - let's check if after commenting them everything is going to compile. Unfortunately we faced linkage error:
image
We can come back to clang compiler and maybe it will pass:)

baziorek:
**Summary (it looks that we fixed with @safinsaf), but the solution is not best possible**.
MacOS build was failing, because dependencies were not compiling. So I've created path for VCPKG GTest port and dependencies are building. Another problem was compilation of internship project to create [iroha-lib in C++](https://wiki.hyperledger.org/display/INTERN/Implement+iroha-cpp+library+for+Hyperledger+Iroha+1) inside the main repository. As I see probably it never compiled for MacOS (but I can be wrong). The current problem with compilation is connected with examples for the iroha-lib (details of compilation error: https://github.com/hyperledger/iroha/pull/3977#issuecomment-1765862908), unfortunately changing compiler didn't help. So I added some kind of dirty hack - I've added conditional compilation in CMake of the iroha-lib (`if(NOT APPLE) compile`). Now everything passes.
Unfortunately I can't fix this better way, because I don't have MacOS.

**So the question for somebody who is deciding about future of Iroha 1 - can we leave like that?**

safinsaf (approved):

appetrosyan (approved):

PR #3980, opened 2023-10-10
[fix] #3971: Consensus bug fixes

### LINKED ISSUE #3971: [BUG] Fix recently discovered consensus bugs
While working on #3914 I discovered bugs and made several fixes to the iroha consensus. These should be solved in an independent PR.

SamHSmith:
## Description

Several bug fixes for sumeragi.

### Linked issue

Closes #3971


core/src/sumeragi/main_loop.rs
```diff
@@ -241,11 +242,32 @@ impl Sumeragi {
...
+                }
+                .sign(genesis_network.genesis_key_pair)
+                .expect("Must succeed"),
+            ));
+
+            transactions
+                .into_iter()
+                .map(AcceptedTransaction::accept_genesis)
+                .collect()
+        };
```
mversic, line 270:
```suggestion
        let transactions: Vec<_> = {
            // Copy sumeragi config params into chain.
            genesis_network.set_parameters(ParametersBuilder::new()
                .add_parameter(BLOCK_TIME, self.block_time.as_millis())
                .expect("should not fail")
                .add_parameter(COMMIT_TIME_LIMIT, self.commit_time.as_millis())
                .expect("should not fail")
                .add_parameter(MAX_TRANSACTIONS_IN_BLOCK, self.max_txs_in_block as u32)
                .expect("should not fail")
                .into_set_parameters()
            );
```

this is how you should do it, and everything else should be in the genesis crate

genesis/src/lib.rs
```diff
@@ -198,7 +203,7 @@ impl ValidatorPath {
 #[repr(transparent)]
 pub struct GenesisTransactionBuilder {
     /// Instructions
-    isi: Vec<InstructionExpr>,
+    pub isi: Vec<InstructionExpr>,
```
mversic, line 206: make it private again after applying previous comment

core/src/sumeragi/main_loop.rs
```diff
@@ -324,15 +337,14 @@ impl Sumeragi {
...
-        let block_signees = block
-            .signatures()
-            .into_iter()
-            .map(|s| s.public_key())
-            .cloned()
-            .collect::<Vec<PublicKey>>();
         let events = block.produce_events();
 
+        let new_topology = Topology::recreate_topology(
+            &SignedBlock::from(block.clone()),
```
mversic, line 343: now we're cloning a block where previously we weren't

mversic (changes requested):
do we have a test confirming new changes have fixed a bug we previously had?

core/src/sumeragi/main_loop.rs
```diff
@@ -354,13 +366,14 @@ impl Sumeragi {
         // NOTE: This sends "Block committed" event,
         // so it should be done AFTER public facing WSV update
         self.send_events(events);
-        self.update_topology(&block_signees, block_topology);
+
+        self.current_topology = new_topology;
```
Erigara, line 370: If now we recreate topology every time what role `current_topology` has?

core/src/sumeragi/main_loop.rs
```diff
@@ -624,7 +652,14 @@ fn process_message_independent(
...
                     let new_block = match BlockBuilder::new(
                         transactions,
-                        sumeragi.current_topology.clone(),
+                        Topology::recreate_topology(
+                            &sumeragi
+                                .wsv
+                                .latest_block_ref()
+                                .expect("WSV must have blocks"),
+                            current_view_change_index,
+                            current_topology.ordered_peers.iter().cloned().collect(),
```
Erigara, line 661: Can we just grab peers from `wsv`?

core/src/sumeragi/main_loop.rs
```diff
@@ -975,6 +1015,12 @@ fn vote_for_block(
     let role = sumeragi.current_topology.role(&sumeragi.peer_id);
     trace!(%addr, %role, block_hash=%block_hash, "Block received, voting...");
 
+    let block_view_change_index = block.payload().header().view_change_index;
+    if block_view_change_index != current_view_change_index {
+        warn!(%role, %current_view_change_index, %block_view_change_index, "Could not vote for block because of incorrect view change index");
+        return None;
+    }
+
```
Erigara, line 1023: Why peer can't accept block with higher view change index if it has all necessary signatures?

Erigara (changes requested):

core/src/sumeragi/main_loop.rs
SamHSmith: In that case they should not exist then? I think from a UX perspective it's way worse for a network operator to modify genesis.json. I will remove this but it does introduce a bug. Iroha users changing the block size in the config then being upset when nothing is different.

core/src/sumeragi/main_loop.rs
SamHSmith: It makes it easier to debug. It's a rare and unnecessary case and we save ourselves time be not supporting it.

core/src/sumeragi/main_loop.rs
SamHSmith: Yes, they are the same. This makes more sense to me.

core/src/sumeragi/main_loop.rs
SamHSmith: Storing the current topology so that we can refer to it? You could probably refactor to create the topology on the spot when you want to use it. I don't think that this is important for this PR though. We shouldn't delay getting these fixes to iroha users.

genesis/src/lib.rs
SamHSmith: just reverting this since we do not believe in modifying genesis.

SamHSmith:
> do we have a test confirming new changes have fixed a bug we previously had?

No. But if you actually use iroha the difference is obvious. Making a test would be a lot of work and it wouldn't be a test the CI would handle well.

core/src/sumeragi/main_loop.rs
Erigara: > In that case they should not exist then? 

Yes, afaik @mversic and @0x009922 has ongoing discussion to preserve on chain parameter only in `genesis`.

core/src/sumeragi/network_topology.rs
```diff
@@ -139,13 +139,18 @@ impl Topology {
...
+        let mut ordered_peers: Vec<_> = self.ordered_peers.clone().into();
+        ordered_peers.retain(|p| match new_peers.iter().position(|p2| p2 == p) {
+            Some(index) => {
+                new_peers.remove(index);
+                true
+            }
+            None => false,
+        });
+        ordered_peers.append(&mut new_peers);
+        self.modify_peers_directly(|peers| *peers = ordered_peers);
```
Erigara, line 153: We can automatically dedup with extend.

```suggestion
self.modify_peers_directly(|peers| peers.retain(|peer| new_peers.contains(peer)));
self.ordered_peers.extend(new_peers);
```

Erigara:
> No. But if you actually use iroha the difference is obvious. Making a test would be a lot of work and it wouldn't be a test the CI would handle well.

Imo test would allow us to catch possible regression in the future.

core/src/sumeragi/mod.rs
```diff
@@ -251,25 +251,24 @@ impl SumeragiHandle {
...
         for block in block_iter_except_last {
+            if wsv.height() == 0 {
+                current_topology = Topology::new(block.payload().commit_topology.clone());
+                wsv.world_mut().trusted_peers_ids = current_topology.ordered_peers.clone();
+            }
+
+            let new_topology =
+                Topology::recreate_topology(&block, 0, wsv.peers_ids().iter().cloned().collect());
+
+            current_topology.rotate_all_n(block.payload().header().view_change_index);
```
Erigara, line 270: Why not to combine into single step?

SamHSmith:
@Erigara Is going to open an alternative PR for these fixes.

PR #3984, opened 2023-10-12
[feature] #3952: Transfer the concrete type of the Query result over-the-wire instead of Value

### LINKED ISSUE #3952: Remove responsibility of `Value` conversion from client
when server returns query result it returns `Value` enum that is then converted with `TryInto` on the client side into a concrete value. The same thing happens when a query is returned to wasm smart contract. This is not appropriate because the return type of the query is well known. Server should always respond with concrete type and not `Value`

DCNick3:
## Description

On the server side it's done by adding an `EncodedValue` type, which contains a scale-encoded type-erased value and a `QueryEncoder` type, which creates `EncodedValue` from `Value` using dynamic dispatch. An alternative to dynamic dispatch would be to return `EncodedValue` from `QueryBox::execute`, but that is problematic for smart contracts (they do not necessarily want a scale-encoded value) and filters (they operate on `Value`s).

On the client size the expected query result type is already known, so `Decode` is called on the correct type from the get-go.

### Linked issue

Closes #3952

### Benefits

Less communication overhead, one less comparison when decoding on client

### Checklist

- [ ] Make sure CI passes
- [ ] Refactor the `_handle_query_response_base` function on the client to not use generics


data_model/src/query/mod.rs
```diff
@@ -62,7 +62,7 @@ macro_rules! queries {
 /// Trait for typesafe query output
 pub trait Query: Into<QueryBox> + seal::Sealed {
     /// Output type of query
-    type Output: Into<Value> + TryFrom<Value>;
+    type Output: Into<Value> + TryFrom<Value> + Encode;
```
mversic, line 65:
```suggestion
    type Output: Encode;
```

Do we still need these bounds?

client/src/client.rs
```diff
@@ -236,7 +234,7 @@ impl From<ResponseReport> for eyre::Report {
 }
 
 /// Output of a query
-pub trait QueryOutput: Into<Value> + TryFrom<Value> {
+pub trait QueryOutput: Into<Value> + TryFrom<Value> + Decode {
```
mversic, line 237:
```suggestion
pub trait QueryOutput: Decode {
```

I guess the same question

data_model/src/query/mod.rs
DCNick3: `TryFrom<Value>` is still needed for smart contracts AFAIU, but `Into<Value>` is not necessary, yeah

mversic: what is it used for in smart contracts?

mversic: what is it used for in smart contracts?

DCNick3: `host_execute_query`'s result is decoded as `Result<Value, ValidationFail>`, which later needs a conversion into a concrete type https://github.com/hyperledger/iroha/blob/fcd2c28b7f104e95b0e1c228ba39e459c4b1083a/smart_contract/src/lib.rs#L101

mversic: I think this is only a temporary limitation. It happens becaue we upper cast query into `QueryBox`. I'll split this into a separate ticket

mversic:
when client sends a query does it send actual query like `FindAllAccounts` or it sends `QueryBox`?

DCNick3:
It does send the `QueryBox`. Otherwise, I think, it would require us to send a tag along with the query object... Which is exactly what `QueryBox` does?

DCNick3:
Well
This PR definitely introduces some additional complexity into the server (`QueryEncoder`) and gives some more potential for type confisuon.

It does reduce the on-the-wire overhead (and one less branch on the client side), but I am not sure whether it's worth the added complexity.

I guess one thing that this might help is with is removing stuff-that-is-not-really-value from `Value` enum, but that is not really required for that, if we wanted, we could split it into more specific enums.

mversic:
> Well This PR definitely introduces some additional complexity into the server (`QueryEncoder`) and gives some more potential for type confisuon.
> 
> It does reduce the on-the-wire overhead (and one less branch on the client side), but I am not sure whether it's worth the added complexity.
> 
> I guess one thing that this might help is with is removing stuff-that-is-not-really-value from `Value` enum, but that is not really required for that, if we wanted, we could split it into more specific enums.

I get the feeling this was my brain fart and we should close this PR before you waste any more time

PR #4004, opened 2023-10-18
[feature] #3130: Add type-state-pattern to divide client into sync and async parts

### LINKED ISSUE #3130: [suggestion] Fully asynchronous client API

pesterev:
## Description

This PR adds traits to implement type-state-pattern on `Client` to divide it into sync and async parts.
I'm going to open also two pull requests:
- Add async client features
- `iroha_awc` and `iroha_hyper` crates implementations

### Linked issue

Closes #3130

### Checklist

- [ ] Rewrite doc tests and examples
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs 

- [ ] Open a pull request that adds async client features
- [ ] Open a pull request that add implementations of the `iroha_awc` and `iroha_hyper` crates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ include/generated/*
 peers.list
 cmake-build*
 result
+.obsidian/
 
 .gtm
 /.gtm/
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -12,7 +12,6 @@ use std::{
 use derive_more::{DebugCustom, Display};
 use eyre::{eyre, Result, WrapErr};
 use futures_util::StreamExt;
-use http_default::{AsyncWebSocketStream, WebSocketStream};
 use iroha_config::{client::Configuration, torii::uri, GetConfiguration, PostConfiguration};
 use iroha_crypto::{HashOf, KeyPair};
 use iroha_data_model::{
@@ -35,8 +34,11 @@ use url::Url;
 
 use self::{blocks_api::AsyncBlockStream, events_api::AsyncEventStream};
 use crate::{
-    http::{Method as HttpMethod, RequestBuilder, Response, StatusCode},
-    http_default::{self, DefaultRequestBuilder, WebSocketError, WebSocketMessage},
+    http::{
+        Method as HttpMethod, RequestBuilder, RequestManager, Response, StatusCode,
+        SyncRequestManager, SyncSendRequest,
+    },
+    http_default::{DefaultSendRequest, WebSocketError, WebSocketMessage},
 };
 
 const APPLICATION_JSON: &str = "application/json";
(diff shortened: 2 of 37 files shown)


client/src/http.rs
```diff
@@ -69,6 +69,59 @@ pub trait RequestBuilder {
...
+        &'async_trait self,
+        rb: Self::RequestBuilder,
+    ) -> ::std::pin::Pin<
+        Box<
+            dyn ::std::future::Future<Output = Result<http::Response<Vec<u8>>>>
+                + Send
+                + 'async_trait,
+        >,
+    >;
+}
```
Arjentix, line 123: Is this an expanstion of `async_trait` crate?

pesterev: I considered this crate, but it will require users to use this crate as I understand. If we agree with it, then let's use it. 
Btw the native async traits have already stabilized and seem can appear in one of the coming releases of Rust.

.gitignore
```diff
@@ -38,6 +38,7 @@ include/generated/*
 peers.list
 cmake-build*
 result
+.obsidian/
```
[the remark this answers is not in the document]
pesterev: That is a [note-taking app](https://obsidian.md). I use it to render and edit our documentation. Pretty useful.

client/src/client.rs
```diff
@@ -248,16 +250,41 @@ pub struct ResultSet<T> {
     client_cursor: usize,
 }
 
-impl<T: Clone> Iterator for ResultSet<T>
+pub struct SeekResultSet<'a, T, R>
```
DCNick3, line 253: Is there a particular reason this struct is called `Seek`? Also, why does it always uses `SyncRequestManager`? Is it more like `SyncResultSet`?

Issue #3991, opened 2023-10-13
Consider `Arc<str>` for `Name`


Arjentix:
This should reduce the memory footprint when a lot of the same Ids a stored multiple times.

I hope that benches won't degrade.

Arjentix:
In #4013 you can see the results of my measurements. This is idea will downgrade our perfromence too much

PR #4013, opened 2023-10-24
[refactor]: Optimize executor

Arjentix:
## Description

Originally this work was about #3991. But after implementing reference counting for `ConstString` I got these results:

1. 7.5 - 13 % less memory footprint
2. 50 - 75 % slower

I measured memory using the next command:

```bash
gtime -f 'Time passed: %E\nMax memory used: %M kbytes' ../../target/release/examples/apply_blocks
```

Where `gtime` is a `time` utility by GNU, which can be installed on mac using `brew install gnu-time`

So it defenelty doesn't worth it.

You can do your own research. Take a look at my branch: https://github.com/arjentix/iroha/tree/arc_name

- Fixed `apply_blocks` benchmark
- Made `apply_blocks` even more heavy with long ids
- After that this benchmarks just stops woring on `iroha2-dev`, the issue was a very long transaction deallocation inside wasm executor, so fixed that with `core::mem::forget()`
- Some minor refactoring

### Checklist


default_executor/src/lib.rs
```diff
@@ -190,12 +191,14 @@ pub fn validate_instruction(
...
     executor.visit_instruction(&authority, &instruction);
+    core::mem::forget(instruction);
     executor.verdict
 }
 
 #[entrypoint]
 pub fn validate_query(authority: AccountId, query: QueryBox, block_height: u64) -> Result {
     let mut executor = Executor::new(block_height);
     executor.visit_query(&authority, &query);
+    core::mem::forget(query);
```
mversic, line 202: I don't think it would be a bad idea to inject these `forget`s by the `entrypoint`?

Erigara (approved):

mversic (dismissed):

mversic (approved):

PR #4015, opened 2023-10-25
[fix] #4014: Use method(not field) access in client

### LINKED ISSUE #4014: Fix client build
Client cannot be built with `cargo build -p iroha_client` again due to recent changes in #3929

mversic:
### Linked issue

Closes #4014

### Checklist

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -256,7 +256,9 @@ where
 
     fn next(&mut self) -> Option<Self::Item> {
         if self.client_cursor >= self.iter.len() {
-            let iroha_data_model::query::QueryRequest::Cursor(cursor) = &self.query_handler.query_request.request else {
+            let iroha_data_model::query::QueryRequest::Cursor(cursor) =
+                &self.query_handler.query_request.request
+            else {
                 return None;
             };
             if cursor.cursor().is_none() {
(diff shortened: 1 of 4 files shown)


data_model/src/query/mod.rs
```diff
@@ -176,7 +176,7 @@ pub mod model {
 
     /// Query with parameters client can specify.
     #[derive(
-        Debug, Constructor, Getters, Clone, PartialEq, Eq, Encode, Decode, Serialize, Deserialize,
+        Clone, Debug, PartialEq, Eq, Constructor, Getters, Encode, Decode, Serialize, Deserialize,
```
Arjentix, line 179: So is our policy to place std derives first? Didn't know about that

Arjentix (approved):

data_model/src/query/mod.rs
mversic: there is no strict policy. If you look around you will see this pattern

client/src/client.rs
```diff
@@ -387,9 +389,9 @@ impl QueryRequest {
 
         match self.request {
             iroha_data_model::query::QueryRequest::Query(query_with_params) => builder
-                .params(Vec::from(query_with_params.sorting))
-                .params(Vec::from(query_with_params.pagination))
-                .body(query_with_params.query),
+                .params(Vec::from(query_with_params.sorting().clone()))
```
DCNick3, line 392: Not introduced in this PR, but the conversion of `Pagination` and `Sorting` structs to `Vec` is a bit unintuitive.. I feel like explicitly naming it something like `into_request_params`.

Maybe even make it into a `as_request_params` & make it accept a reference (it's cloned here anyways).

mversic: opened #4017 4017

DCNick3 (approved):

PR #4016, opened 2023-10-26
[feature] #3900: `fetch_size` query parameter

### LINKED ISSUE #3900: Server-side cursor `fetch_size` per query
After merge of #3468 we have enabled cursor based pagination for queries. We've also added a way to define the `fetch_size` in `config.json`. The request made recently was that we provide a way to define this parameter forevery  query request.

Arjentix:
### Linked issue

- `fetch_size` is no more a configuration parameter but a query parameter. Clients have to send it as a query parameter.
- Added integration test

- Closes #3900 

### Benefits

- Clients can configure how big batches they want

### Checklist

- [x] (optional) I've written ~~unit tests~~ integration test for the code changes


data_model/src/query/mod.rs
```diff
@@ -40,6 +41,50 @@ pub mod cursor;
 pub mod pagination;
 pub mod sorting;
 
+const FETCH_SIZE: &str = "fetch_size";
+
+/// Default value for `fetch_size` parameter in queries.
+#[cfg(feature = "std")]
+pub static DEFAULT_FETCH_SIZE: once_cell::sync::Lazy<NonZeroU32> =
```
Erigara, line 48: Maybe instead of `once_cell` we can use [small function](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=31efe0ae459721e7df303b92fc7f997c) with a little bit of unsafe? Safety is checked at compile time.

Upside is that it can be used in `std` and `no_std` environment without problem and require lesser amount of code.

Erigara (commented):
Imo we should limit maximum size of `fetch_size`.

data_model/src/query/mod.rs
Arjentix: Nice idea, will do!

Arjentix:
> Imo we should limit maximum size of fetch_size.

But how? Using another config parameter?

data_model/src/query/mod.rs
Arjentix: I think just in-place `unsafe` block would be fine here as well, check out new code, please

Erigara:
> But how? Using another config parameter?

Or hardcode, however we can leave this as is for now.

Erigara (approved):

client/src/client.rs
```diff
@@ -247,6 +247,15 @@ pub struct ResultSet<T> {
...
 
+impl<T> ResultSet<T> {
+    /// Get the length of the batch returned by Iroha.
+    ///
+    /// This is controlled by `fetch_size` parameter of the query.
+    pub fn batch_len(&self) -> usize {
+        self.iter.len()
+    }
+}
+
```
mversic, line 258:
```suggestion
```

I don't think we need this. The one place you use it, you could have just collected into a vec

data_model/src/query/mod.rs
```diff
@@ -40,6 +41,28 @@ pub mod cursor;
 pub mod pagination;
 pub mod sorting;
 
+const FETCH_SIZE: &str = "fetch_size";
+
+/// Default value for `fetch_size` parameter in queries.
+// SAFETY: `100` is greater than `0`
+#[allow(unsafe_code)]
+pub const DEFAULT_FETCH_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(100) };
```
mversic, line 49:
```suggestion
pub const DEFAULT_FETCH_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(10) };
```

in JDBC most drivers use `10`

mversic (changes requested):

0x009922 (changes requested):
Merging **only** after a PR to update [Torii Endpoints](https://hyperledger.github.io/iroha-2-docs/api/torii-endpoints.html) page is created.

client/src/client.rs
Arjentix: No, I can't, because it will collect all batches implicitly

Arjentix:
>>> Imo we should limit maximum size of fetch_size.
>>
>> But how? Using another config parameter?
>
> Or hardcode, however we can leave this as is for now.

@mversic , what do you think about that?

Arjentix:
> Merging only after a PR to update [Torii Endpoints](https://hyperledger.github.io/iroha-2-docs/api/torii-endpoints.html) page is created.

@0x009922 , is it `iroha2-docs` repo?

mversic:
> > > > Imo we should limit maximum size of fetch_size.
> > > 
> > > 
> > > But how? Using another config parameter?
> > 
> > 
> > Or hardcode, however we can leave this as is for now.
> 
> @mversic , what do you think about that?

I think it's ok to just hardcode

0x009922:
> > Merging only after a PR to update [Torii Endpoints](https://hyperledger.github.io/iroha-2-docs/api/torii-endpoints.html) page is created.
> 
> @0x009922 , is it `iroha2-docs` repo?

@Arjentix, yes. At the bottom of the page you might find "Edit this page on GitHub" link for convenience.

Arjentix:
@0x009922 , created https://github.com/hyperledger/iroha-2-docs/pull/425

mversic (approved):

mversic:
> > > > > Imo we should limit maximum size of fetch_size.
> > > > 
> > > > 
> > > > But how? Using another config parameter?
> > > 
> > > 
> > > Or hardcode, however we can leave this as is for now.
> > 
> > 
> > @mversic , what do you think about that?
> 
> I think it's ok to just hardcode

I would like to backport this into RC20. But in the backport we have to leave in the config intact since removing it would break backwards compatibility

Arjentix:
> I would like to backport this into RC20. But in the backport we have to leave in the config intact since removing it would break backwards compatibility

Ok, will do in separate PR

Arjentix:
@mversic , I tried and turns out this PR uses a lot from #3958 which is not merged into stable.

Also this seems to be a feature, while AFAIK only bug-fix backports make sence

Arjentix:
@mversic , I tried and turns out this PR uses a lot from #3958 which is not merged into stable.

Also this seems to be a feature, while AFAIK only bug-fix backports make sence

mversic:
> Also this seems to be a feature, while AFAIK only bug-fix backports make sence

ideally, only bugfixes would be backported. However, this feature was requested by Orillion. 
Nvm, if not possible

Erigara (dismissed):

Erigara (approved):

mversic (approved):

PR #4024, opened 2023-10-30
[refactor] #3982: Clear live queries after smart contract end

### LINKED ISSUE #3982: Remove live queries after smart contract end
then you should make sure that you have deleted all live queries after smart contract exits (whether successfully or with error)

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/3958#discussion_r1347135274_

Arjentix:
## Description

- Queries are removed from `LiveQueryStore` after wasm end
- Added `QueryId` type alias
- Added client crate API to directly use cursor
- Added test

### Linked issue

Closes #3982 

### Benefits

- Optimizes storage
- Less dangerous
- More flexible client API

### Checklist

- [x] (optional) I've written ~unit~ integration tests for the code changes


core/src/smartcontracts/wasm.rs
```diff
@@ -1241,28 +1392,44 @@ impl<'wrld> import_traits::ExecuteOperations<state::executor::ValidateQuery<'wrl
     #[codec::wrap]
     fn execute_query(
         query_request: SmartContractQueryRequest,
-        state: &state::executor::ValidateQuery<'wrld>,
+        state: &mut state::executor::ValidateQuery<'wrld>,
```
mversic, line 1395: seems like reason to not make query store part of the wsv

core/src/smartcontracts/wasm.rs
```diff
@@ -1241,28 +1392,44 @@ impl<'wrld> import_traits::ExecuteOperations<state::executor::ValidateQuery<'wrl
...
-                    &sorting,
-                    pagination,
-                )
+                let batched = {
+                    let wsv = state.wsv();
+                    let output = query.execute(wsv)?;
+
+                    wsv.query_handle().handle_query_output(
+                        output,
+                        NonZeroUsize::new(30_000).expect("30 000 is not zero"),
```
mversic, line 1411: what is this magic number?

core/src/smartcontracts/wasm.rs
```diff
@@ -763,39 +870,61 @@ impl<S: state::LimitsMut> Runtime<S> {
...
+        let mut state = store.into_data();
+        let executed_queries = state.take_executed_queries();
+        forget_all_executed_queries(state.wsv().query_handle(), executed_queries)?;
+        Ok(validation_res)
     }
 }
 
 #[allow(clippy::needless_pass_by_value)]
-impl<S: state::Authority + state::Wsv + state::WsvMut> Runtime<S> {
+impl<S: state::Authority + state::Wsv + state::WsvMut + state::AddExecutedQuery> Runtime<S> {
```
mversic, line 885:
```suggestion
impl<S: state::Authority + state::WsvMut + state::AddExecutedQuery> Runtime<S> {
```

shouldn't `WsvMut` imply `Wsv`?

core/src/smartcontracts/wasm.rs
```diff
@@ -985,33 +1119,49 @@ trait ExecuteOperationsAsExecutorMut<S> {}
...
+                }
+                Ok(batched)
+            }
+            QueryRequest::Cursor(cursor) => {
+                // In a normal situation we already have this `query_id` stored,
+                // so that's a protection from malicious smart contract
+                if let Some(query_id) = &cursor.query_id {
+                    state.add_executed_query(query_id.clone());
+                }
+                state.wsv().query_handle().handle_query_cursor(cursor)
```
mversic, line 1163: this looks so duplicated in 3 places

core/src/smartcontracts/wasm.rs
Erigara: i think this happened not because of query store, since it uses only `&self` methods, but because we now attach `HashSet<QueryId>` to `State` and we need to update it.

core/src/query/store.rs
```diff
@@ -197,13 +199,25 @@ impl LiveQueryStoreHandle {
...
-    fn insert(&self, query_id: String, live_query: LiveQuery) -> Result<()> {
+    /// Remove query from the storage if there is any.
+    ///
+    /// Returns `true` if query was removed, `false` otherwise.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::ConnectionClosed`] if [`QueryService`] is dropped,
+    /// - Otherwise throws up query output handling errors.
+    pub fn forget_query(&self, query_id: QueryId) -> Result<bool> {
```
mversic, line 210:
```suggestion
    pub fn remove_query(&self, query_id: QueryId) -> Result<bool> {
```

`forget` to me implies a memory leak

Erigara:
I think tests is required here, we can postpone until client `fetch_size` will be merged

core/src/query/store.rs
```diff
@@ -197,13 +199,25 @@ impl LiveQueryStoreHandle {
...
+    }
+
+    fn insert(&self, query_id: QueryId, live_query: LiveQuery) -> Result<()> {
         self.insert_sender
             .blocking_send((query_id, live_query))
             .map_err(|_| Error::ConnectionClosed)
     }
 
-    fn remove(&self, query_id: String) -> Result<Option<LiveQuery>> {
+    fn remove(&self, query_id: QueryId) -> Result<Option<LiveQuery>> {
```
mversic, line 220: Considering my previous comment, I would recommend renaming to `take`

core/src/query/store.rs
Erigara, line 210: Maybe `drop_query`? Because as @mversic wrote it has memory leaking connotation.

core/src/smartcontracts/wasm.rs
Arjentix: It was introduced in my previous PR with live queries for wasm. It's fixed in #4016

core/src/smartcontracts/wasm.rs
Arjentix: Fixed

core/src/smartcontracts/wasm.rs
Arjentix: Agree, will try to do something with this

core/src/query/store.rs
Arjentix: Ok. Btw, where did he wrote that?

Erigara: I think @mversic deleted this comment

core/src/smartcontracts/wasm.rs
```diff
@@ -119,9 +122,15 @@ pub mod error {
 
     /// `WebAssembly` execution error type
     #[derive(Debug, thiserror::Error, displaydoc::Display)]
+    #[ignore_extra_doc_attributes]
```
mversic, line 125: why?

Arjentix: It's for `displaydoc::Display` derive. Need because I don't want to pollute user-friendly message with developer-specific info described in the next paragraph for `Finalization` error

Arjentix:
Added test

client/src/client.rs
```diff
@@ -338,12 +338,13 @@ impl_query_output! {
 }
 
 /// Iroha client
-#[derive(Clone, DebugCustom, Display)]
+#[derive(Clone, DebugCustom, Display, getset::Getters)]
```
Erigara, line 341: Why do we need getters here?

client/tests/integration/queries/mod.rs
```diff
@@ -27,3 +30,45 @@ fn too_big_fetch_size_is_not_allowed() {
...
+    let transaction = client.build_transaction(
+        WasmSmartContract::from_compiled(wasm),
+        UnlimitedMetadata::default(),
+    )?;
+    client.submit_transaction_blocking(&transaction)?;
+
+    let metadata_value = client.request(FindAccountKeyValueByIdAndKey::new(
+        client.account_id().clone(),
+        Name::from_str("cursor").unwrap(),
+    ))?;
```
Erigara, line 56: Smart idea

core/src/smartcontracts/wasm.rs
```diff
@@ -327,258 +353,211 @@ pub mod state {
...
+    /// Generic over concrete state kind and containing common data.
+    pub struct CommonState<K> {
         pub(super) authority: AccountId,
         pub(super) store_limits: StoreLimits,
-        pub(super) wsv: &'wrld mut WorldStateView,
         /// Span inside of which all logs are recorded for this smart contract
         pub(super) log_span: Span,
+        pub(super) executed_queries: HashSet<QueryId>,
+        /// Concrete kind of state
+        pub(super) kind: K,
```
Erigara, line 365: Why do you choose to use `CommonState<Kind<Subkind>>` instead of `CommonState<WsvType, ExecutableType>`?

client/src/client.rs
Arjentix: As you can see in your another comment, I use `client.account_id()`. I think it's more readable and makes more sence than `alice@wonderland` here

Arjentix: I think it's okay for `Client` to have getters for fields it's configured with

core/src/smartcontracts/wasm.rs
Arjentix: Interesting, I'll try

Arjentix: Done

mversic (approved):

client/src/client.rs
mversic: and did you consider making the fields public?

client/tests/integration/smartcontracts/executor_with_custom_token/src/lib.rs
```diff
@@ -170,6 +170,7 @@ impl Executor {
     }
 }
 
+// TODO: why it warns about unused function?
```
mversic, line 173: please remove and open a ticket if needed

client/tests/integration/smartcontracts/query_assets_and_save_cursor/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[lib]
+crate-type = ['cdylib']
+
+[dependencies]
+iroha_smart_contract.workspace = true
+
+panic-halt.workspace = true
+lol_alloc.workspace = true
+serde_json = { version = "1.0.108", default-features = false }
```
mversic, line 18: why is this not taken from workspace?

client/src/client.rs
```diff
@@ -873,6 +875,29 @@ impl Client {
         self.build_query(request).execute()
     }
 
+    /// Query API entry point using cursor.
+    ///
+    /// # Errors
+    /// Fails if sending request fails
+    pub fn request_with_cursor<O>(&self, cursor: ForwardCursor) -> QueryResult<O::Target>
```
mversic, line 882: shouldn't you put this into `client/src/query_builder.rs`?

client/tests/integration/smartcontracts/executor_with_custom_token/src/lib.rs
mversic: please also remove `request_with_filter_and_pagination_and_sorting`

client/src/client.rs
mversic: please also remove `request_with_filter_and_pagination_and_sorting`. It must have been left in accidentally

client/src/client.rs
Arjentix: Hm, no

Arjentix: Although it's a crate which will be used by external clients a lot. So I think having methods instead of direct field access is better for future API changes

client/tests/integration/smartcontracts/executor_with_custom_token/src/lib.rs
Arjentix: Actually me and @ilchu have already found the reason on his PR: https://github.com/hyperledger/iroha/pull/4049#discussion_r1391462122

Arjentix: But I'll leave a comment

client/tests/integration/smartcontracts/query_assets_and_save_cursor/Cargo.toml
Arjentix: `smartcontracts` is a separate workspace. There was no `serde_json` in there, so I decided to put it directly in the crate which uses it, as nobody else needs that

client/src/client.rs
mversic: > Although it's a crate which will be used by external clients a lot. So I think having methods instead of direct field access is better for future API changes

I disagree. What is the difference if you use field vs method access?

client/src/client.rs
Arjentix: And it uses some private parts of `Client` so looks like we have to keep it

Arjentix: > shouldn't you put this into client/src/query_builder.rs?

I just tried and I don't think so. There is nothing to customize in a query by cursor, so nothing to build. And for simple queries we have `client.request()`. So I think having `client.request_with_filter()` is also ok

mversic: > I'll make it `pub(crate)`

this is proper solution. This should not be exposed

client/src/client.rs
Arjentix: To me it sounds natural to have `pub` fields only in [POD types](https://en.wikipedia.org/wiki/Passive_data_structure). For complex types with complex logic I'd prefer getters.

Objectively having getters allows us to change structure fields if future while still mimicing the same API for the end users.

Arjentix: It creates opaqueness which is not bad for a crate which will be used by external people

mversic: how is this opaqueness useful? it serves no tangible purpose. If you can tell me how method provides you with more opaqueness than `pub` fields I will accept this

Arjentix: Here it is:

> Objectively having getters allows us to change structure fields if future while still mimicing the same API for the end users.

Actually I don't mind making fields public

mversic (approved):

client/Cargo.toml
```diff
@@ -46,6 +46,7 @@ parity-scale-codec = { workspace = true, default-features = false, features = ["
 tokio = { workspace = true, features = ["rt"] }
 tokio-tungstenite = { workspace = true, features = ["native-tls"] }
 futures-util = "0.3.28"
+getset = { workspace = true }
```
mversic, line 49: remove please

Arjentix: Indeed, outdated now

mversic (approved):

Erigara (approved):

mversic (approved):

Issue #3974, opened 2023-10-09
[suggestion]: Extend functionality of `iroha_client_cli` 


Erigara:
### Feature request

After merge of #3963 there is couple of features missing from `iroha_client_cli`.

- [x] Transferring `Domain` between users
- [x] Setting/Removing metadata for `Domain`
- [ ] Minting/Burning trigger repetitions

### Motivation

Help user to express what they what using `iroha_client_cli`.

### Who can help?

@Erigara

timofeevmd:
@VAmuzing could you tell me, is it possible to add to the documentation what exactly the domain metadata set command takes as input?
It was not obvious to me as a user what type of parameters this command accepts

PR #4031, opened 2023-11-03
[feature] #3974: Add subcommand into `iroha_client_cli` to transfer domains

pesterev:
## Description

This pull request adds command to transfer domains between accounts into `iroha_client_cli`.

### Linked issue

Closes one of subtasks from #3974  

### Benefits

Extends the functionality of `iroha_client_cli`.

### Checklist

### DIFF
--- a/client_cli/src/main.rs
+++ b/client_cli/src/main.rs
@@ -363,11 +363,13 @@ mod domain {
         /// List domains
         #[clap(subcommand)]
         List(List),
+        /// Transfer domain
+        Transfer(Transfer),
     }
 
     impl RunArgs for Args {
         fn run(self, context: &mut dyn RunContext) -> Result<()> {
-            match_all!((self, context), { Args::Register, Args::List })
+            match_all!((self, context), { Args::Register, Args::List, Args::Transfer })
         }
     }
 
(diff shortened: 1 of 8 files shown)


client_cli/src/main.rs
```diff
@@ -420,6 +422,36 @@ mod domain {
...
+        pub id: DomainId,
+        /// Account from which to transfer (in form `name@domain_name')
+        #[structopt(short, long)]
+        pub from: AccountId,
+        /// Account from which to transfer (in form `name@domain_name')
+        #[structopt(short, long)]
+        pub to: AccountId,
+        /// The JSON/JSON5 file with key-value metadata pairs
+        #[structopt(short, long, default_value = "")]
+        pub metadata: super::Metadata,
```
Erigara, line 440: What is this metadata for?

Erigara (changes requested):
Let's cover this functionality in python client tests?

client_cli/src/main.rs
pesterev: Yes, seems this argument is not needed. By the way, I've found out that there are other subcommands which accepts metadata that I don't think is needed there. For instance, `asset transfer` or `peer unregister` shouldn't accept metadata in my opinion. Let's fix them in a separate pull request?

Arjentix: AFAIU this metadata is used in transaction. So why to remove it?

client_cli/src/main.rs
```diff
@@ -420,6 +422,29 @@ mod domain {
...
+    pub struct Transfer {
+        /// Domain name as double-quited string
+        #[structopt(short, long)]
+        pub id: DomainId,
+        /// Account from which to transfer (in form `name@domain_name')
+        #[structopt(short, long)]
+        pub from: AccountId,
+        /// Account from which to transfer (in form `name@domain_name')
+        #[structopt(short, long)]
+        pub to: AccountId,
```
Arjentix, line 437:
```suggestion
        /// Account to which to transfer (in form `name@domain_name')
        #[structopt(short, long)]
        pub to: AccountId,
```

client_cli/src/main.rs
```diff
@@ -420,6 +422,29 @@ mod domain {
...
+    }
+
+    impl RunArgs for Transfer {
+        fn run(self, context: &mut dyn RunContext) -> Result<()> {
+            let Self { id, from, to } = self;
+            let transfer_domain = TransferExpr::new(from, id, to);
+            submit([transfer_domain], UnlimitedMetadata::new(), context)
```
Arjentix, line 444: So that's the place where custom metadata could be.

Yes, I don't know the exact reasons why do we even have metadata for transactions. But I can imagine that for custom chains it could be usefull to attach some additional info for transaction. I.e. it can be checked insise runtime executor. Or analyzed by some off-chain listener.

pesterev: Agreed, this is a very good use case where it is needed.

0x009922 (dismissed):

Erigara (approved):

Arjentix (approved):

Issue #4046, opened 2023-11-09
Make it possible for iroha_crypto to target wasm for JS SDK


DCNick3:
Currently JS SDK has to reimplement parts of `iroha_crypto`: https://github.com/hyperledger/iroha-javascript/tree/iroha2/packages/crypto/crypto-rs. It would be nice to remove the duplication and have the same codebase used for crypto operations between the node and js sdk.

After the PR for #3422 lands, it does make sense to try and make sure it compiles under wasm for js, so that JS SDK can use this code directly, without having to duplicate it.

Arjentix:
In #4181 wasm support was implemented. @0x009922 can you check if it works for JS SDK now?

0x009922:
I have created https://github.com/hyperledger/iroha-javascript/issues/186 to track it.

Issue #4045, opened 2023-11-09
[suggestion]: Use concrete key types in iroha_crypto


DCNick3:
(note that this only makes sense to do after the PR for #3422 lands)

## Background

For most asymmetric cryptographic algorithm, the keys are not just "an array of bytes" but they actually have some internal structure. Constructing them from bytes requires a parsing step that can fail. Currently we do not reflect this structure in `PublicKey` and `PrivateKey` types, storing them as an array of bytes tagged with an algorithm name.

This is unfortunate because we end up copying & deserializing the keys each time we do any cryptographic operation with it. This:
1. might have some performance impact
2. delays the key validation errors to the stage where we actually try to use the key

## The suggestion

I suggest to change the key types to instead be something akin to:

```rust
pub enum PublicKey {
    Ed25519(ed25519_dalek::VerifyingKey),
    Secp256k1(k256::PublicKey),
    ...
}
```

In this case the keys are stored in native representations of the crypto traits we use and do not require any conversion to work with. The conversions would now happen at the serialization/deserialization stage.

## The smart contract problem

This would, however, require a change to how we treat `iroha_crypto` in smart contracts. Currently smart contracts can manipulate the cryptographic keys, but cannot do actual cryptographic operations (relying on the host to validate everything for them). Changing the `PublicKey` type like this would, however, require to actually include at least the parsing part of cryptographic libraries to the smart contracts.

We are, however, focused on reducing the wasm binary size, so this is not feasible as-is.

One solution I can think of is to have separate types for unparsed/parsed key representations. This seems feasible on the surface, but should be investigated further.

PR #4047, opened 2023-11-09
[refactor] #3422: Ursa replacement

### LINKED ISSUE #3422: [refactor]: Remove `ursa` dependency
Hyperledger Ursa is rightfully considered EOL. We have no incentive to use `ursa` over any library given that at this point we're also the only people maintaining it and also the only people using it. 

Under these circumstances it is advised to remove the direct dependency on `ursa` and instead extract the code paths which are used in `iroha_crypto`. Ursa-specific implementations of cryptographic primitives should be added to `iroha_crypto` directly.

DCNick3:
## Description

Ursa is a rust cryptographic library that iroha has been using for a lot of time. It combines several cryptographic libraries from crates.io and provides a unified interface. Unfortunately, it is no longer maintained, so we have to migrate to other ecosystem crates.

This PR does this by importing the parts of ursa code that we have been using into `iroha_crypto`. It then proceeds to clean the code up a bit, removing unused functionality, updating dependencies and simplifying internal APIs.

In the end, as ursa was just a unifying layer, iroha_crypto ends up using almost the same underlying cryptographic function implementations (except for a switch from C-based `secp256k1` to pure-rust `k256` for easier use as a wasm library in JS SDK later #4046).

### Linked issue

Closes #3422

### Benefits

Not having an unmaintained dependency is obviously a plus, allowing us to benefit from updates to upstream libraries.

### Future work

#4046, #4045

### Checklist

- [x] make ci pass


Cargo.toml
```diff
@@ -114,8 +114,7 @@ strum = { version = "0.25.0", default-features = false }
 getset = "0.1.2"
 hex-literal = "0.4.1"
 
-ursa = "0.3.7"
-aead = "0.3.2"
+aead = "0.5.2"
```
mversic, line 117: move it into `crypto` crate

crypto/Cargo.toml
```diff
@@ -34,10 +53,42 @@ parity-scale-codec = { workspace = true, features = ["derive", "full"] }
...
+hkdf = { version = "0.12.3", optional = true }
+amcl = { version = "0.2.0", optional = true, default-features = false, features = ["secp256k1"] }
+amcl_wrapper = { version = "0.4.0", optional = true }
+
+signature = { version = "2.1.0", optional = true }
+ed25519-dalek = { version = "2.0.0", optional = true, features = ["rand_core"] }
+curve25519-dalek = { version = "4.1.1", optional = true }
+x25519-dalek = { version = "2.0.0", optional = true, features = ["static_secrets"] }
+
+rand = { version = "0.8.5", optional = true }
```
mversic, line 73:
```suggestion
rand = { workspace = true, optional = true }
```

why not use from workspace?

crypto/Cargo.toml
```diff
@@ -34,10 +53,42 @@ parity-scale-codec = { workspace = true, features = ["derive", "full"] }
...
+k256 = { version = "0.13.1", optional = true, features = ["ecdsa", "sha256"]}
+
 [dev-dependencies]
 hex-literal = { workspace = true }
 serde_json = { workspace = true }
+
+# these crypto libraries are not used to implement actual crypto algorithms
+# but to test some of the primitives against them
+secp256k1 = { version = "0.28.0", features = ["rand", "serde"] }
+libsodium-sys-stable = { version = "1.20.3", features = [] }
```
mversic, line 93:
```suggestion
libsodium-sys-stable = { version = "1.20.3", default-features = false }
```

no?

crypto/src/encryption/mod.rs
```diff
@@ -0,0 +1,293 @@
...
+
+use std::io::{Read, Write};
+
+use aead::{
+    generic_array::{typenum::Unsigned, ArrayLength, GenericArray},
+    Aead, Error, KeyInit, Payload,
+};
+use rand::{rngs::OsRng, RngCore};
+
+pub use self::chacha20poly1305::ChaCha20Poly1305;
```
mversic, line 28:
```suggestion
pub use chacha20poly1305::ChaCha20Poly1305;
```

DCNick3: This is required for disambiguation, we use a dependency named `chacha20poly1305`

Erigara (commented):
lgtm, not much to add

Cargo.toml
DCNick3: Done. This also prompted me to refactor the error handling code in the symmetric encryption and iroha_p2p a bit.

crypto/Cargo.toml
```diff
@@ -34,10 +53,42 @@ parity-scale-codec = { workspace = true, features = ["derive", "full"] }
...
+openssl = "0.10.59"
```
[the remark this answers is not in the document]
DCNick3: Do you mean between the comment and the deps?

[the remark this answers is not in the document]
mversic: at the EOF

crypto/src/encryption/chacha20poly1305.rs
```diff
@@ -0,0 +1,154 @@
...
+    use super::*;
+
+    #[test]
+    fn encrypt_easy_works() {
+        let cipher = ChaCha20Poly1305::new(&ChaCha20Poly1305::key_gen().unwrap());
+        let aad = Vec::new();
+        let message = b"Hello and Goodbye!".to_vec();
+        let res = cipher.encrypt_easy(&aad, &message);
+        assert!(res.is_ok());
+        let ciphertext = res.unwrap();
```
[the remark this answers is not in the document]
DCNick3: This is an artifact of `ursa`'s code, fixed now

crypto/Cargo.toml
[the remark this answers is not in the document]
DCNick3: done

crypto/Cargo.toml
```diff
@@ -34,10 +55,45 @@ parity-scale-codec = { workspace = true, features = ["derive", "full"] }
 serde = { workspace = true, features = ["derive"] }
 serde_with = { workspace = true, features = ["macros"] }
 hex = { workspace = true, features = ["alloc", "serde"] }
+# TODO: iroha_crypto no longer depends on openssl (did it ever?)
+# currently it's being used by iroha_client through attohttpc and iroha_cli through warp's tokeio-tunstenite
 openssl-sys = { version = "0.9.93", features = ["vendored"], optional = true }
```
mversic, line 60: so why do we have this dependency if it doesn't depend on `openssl`?

DCNick3: I think it's needed to force the `vendored` feature, which will make sure an openssl version bundled with the crate is built and used, instead of relying on system openssl. It also probably helps to ensure static linking is possible.

I haven't researched that much whether do we need it, but it's clear that even if we do, `iroha_crypto` is not the place.

`iroha_client` should probably be configurable by the user (you may want to use rustls, and `attohttpc` exposes features to do that)

mversic: move it to `dev-dependencies`

DCNick3: So, I did some experimentation and here's what I learned:

The `vendored` feature was added to make musl build of iroha work.

When building for musl, one cannot rely on the system openssl and has to build one themselves in order to statically link with it. So, the `vendored` feature was added to `iroha_crypto` to enable the `vendored` feature of `openssl-sys` which ultimately compiled the correct openssl with the correct compiler and linked it in.

But the thing is that `iroha_cli` (which is what gets built with musl in `Dockerfile`) does not depend on `openssl` at all. However other crates from the workspace (like `iroha_client` and `iroha_client_cli`) do. The whole workspace was getting build in the `Dockerfile`, so it failed.

I removed the `vendored` feature and modified the Dockerfile command to only build `iroha_cli`, so the build passes without this feature now. Ping @6r1d because they were the one dealing with Docker images AFAIR?

DCNick3: Update, turns out the feature _is_ needed for the build of the docker image, as `iroha_client_cli` is bundled inside.

I've moved the `vendored` feature to `iroha_client_cli`, where it belongs

Cargo.toml
mversic: you didn't move it to iroha_crypto?

crypto/Cargo.toml
mversic: why would `iroha_client_cli` need `openssl`?

client_cli/Cargo.toml
```diff
@@ -22,6 +22,11 @@ is-it-maintained-issue-resolution = { repository = "https://github.com/hyperledg
 is-it-maintained-open-issues = { repository = "https://github.com/hyperledger/iroha" }
 maintenance = { status = "actively-developed" }
 
+[features]
+vendored = [
+    "dep:openssl-sys"
```
mversic, line 27:
```suggestion
    "dep:iroha_crypto/openssl-sys"
```

I think you should do it like this and remove explicit dependency for `openssl-sys` from this file

crypto/Cargo.toml
mversic: > However other crates from the workspace (like iroha_client and iroha_client_cli) do

yes, but only transitively through `iroha_crypto` crate

client_cli/Cargo.toml
mversic: but why would we need this at all if `openssl` is only used for `dev-dependencies`?

crypto/Cargo.toml
DCNick3: `iroha_crypto` doesn't actually use any `openssl` at all, it's all pure rust!

What does use `openssl` is `native-tls`, which is in turn used by `attohttpc`, the http client used by `iroha-client` and `tokio-tungstenite` the websocket client used by `iroha-client`. The `native-tls` feature is even [explicitly enabled](https://github.com/hyperledger/iroha/blob/e9c6fe4f14dba2376c1ac1af85283b146c7005b1/client/Cargo.toml#L47) in `iroha_client`.

There is a common pattern now in rust libraries supporting HTTPS: they have two backends: `native-tls` (which works through openssl) and `rustls` (which is a pure-rust implementation based on `ring`).

We can ditch `native-tls` and use `rustls`, which brings much less linking trouble. Though, IMO, it only makes sense for binaries we ship; ideally the `iroha_client` should give the user a choice like all the other libraries ([1](https://github.com/elastic/elasticsearch-rs/blob/6c0bcd23eed35ad8492f69fd213bfcee52882156/elasticsearch/Cargo.toml#L25), [2](https://github.com/XAMPPRocky/octocrab/blob/551871a5766339f47119ee84686ce2dd4310e6ba/Cargo.toml#L73), [3](https://github.com/durch/rust-s3/blob/5105a1bff57827f88fb90840f03193665ea34d68/s3/Cargo.toml#L108), ...)

Cargo.toml
DCNick3: Uhh, I think I misunderstood you before. `iroha_p2p` previously depended on `aead`, which is what I thought this was all about.

Now only `iroha_crypto` uses it and I removed it from the workspace `Cargo.toml` :+1:

crypto/Cargo.toml
DCNick3: Went ahead and did this in the latest commit. 

To reiterate: `iroha_crypto` does not use `openssl` (only for tests), but `openssl` is used for HTTPS support in `iroha_client`. This is not the only TLS implementation, `rustls` can be used for painless musl support.

This commit gives the user more freedom in choosing their TLS implementation by providing four `iroha_client` features: `tls-native`, `tls-native-vendored`, `tls-rustls-native-roots` and `tls-rustls-webpki-roots`, which mirror corresponding features of `attohttpc` and `tokio-tungstenite`.

`iroha_client_cli`, which is the only thing that uses `iroha_client` and gets built with musl libc, now uses the `tls-rustls-native-roots` feature, which does not need any tweaks to work with `musl`

Unlike previously, none of the TLS implementations are enabled by default, which is a breaking change. If this gets landed as-is, we'll have to communicate it to users and/or tech writers

crypto/src/lib.rs
```diff
@@ -440,7 +439,7 @@ impl PrivateKey {
     /// - If the given payload is not hex encoded
     /// - If the given payload is not a valid private key
     #[cfg(feature = "std")]
-    pub fn from_hex(digest_function: Algorithm, payload: &[u8]) -> Result<Self, Error> {
+    pub fn from_hex(digest_function: Algorithm, payload: &str) -> Result<Self, Error> {
```
mversic, line 442:
```suggestion
    pub fn from_hex(digest_function: Algorithm, payload: impl AsRef<[u8]>) -> Result<Self, Error> {
```

how about this?

crypto/src/lib.rs
```diff
@@ -325,6 +293,19 @@ ffi::ffi_item! {
...
+    /// Creates a new public key from raw bytes received from elsewhere
+    pub fn from_raw(algorithm: Algorithm, payload: ConstVec<u8>) -> Self {
+        Self {
+            digest_function: algorithm,
+            payload,
+        }
+    }
+
+    /// Extracts the raw bytes from public key
+    pub fn into_raw(self) -> (Algorithm, ConstVec<u8>) {
```
mversic, line 305: I don't quite like this API. If we have to have `from_raw` and `into_raw`, rename them to `from_raw_parts` and `into_raw_parts`

crypto/src/lib.rs
```diff
@@ -122,35 +118,13 @@ impl FromStr for Algorithm {
     allow(unused_tuple_struct_fields)
 )]
 #[derive(Debug, Clone)]
-enum KeyGenOption {
+pub enum KeyGenOption {
```
mversic, line 121:
```suggestion 
enum KeyGenOption {
```

I don't think it has to be `pub`

mversic (changes requested):

crypto/Cargo.toml
mversic: good job

crypto/src/lib.rs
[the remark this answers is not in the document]
DCNick3: It is indeed not exposed in the signature key generation APIs, but I've re-used it for the `KeyExchangeScheme` trait, which is public as it is used in `iroha_p2p`.

mversic (approved):

Erigara (approved):

PR #4053, opened 2023-11-13
[ci]: docker image validator

6r1d:
## Description

Adds a script for https://github.com/hyperledger/iroha/pull/3969: prune `docker-compose.yml` files


.github/workflows/no-incorrect-image.yml
```diff
@@ -0,0 +1,17 @@
+name: I2::CI::check_for_incorrect_images
+
+on: workflow_dispatch
```
0x009922, line 3: Will it run for each PR and for each push into `iroha2-dev` and `iroha2-stable` branches?

.github/scripts/ci_test/ci_image_scan.py
```diff
@@ -0,0 +1,46 @@
...
+        error(f'improper formatting at "{config_file}"')
+        status = 1
+    for _, service in services.items():
+        if service.get('image', '') in FAIL_IMAGES:
+            status = 1
+            break
+    return status
+
+def main():
+    for yml_file in git_root().glob('*.yml'):
```
0x009922, line 39: The glob is too general. In principle, I would remove specific file patterns out of the script.

.github/workflows/no-incorrect-image.yml
```diff
@@ -0,0 +1,17 @@
+name: I2::CI::check_for_incorrect_images
```
0x009922, line 1:
```suggestion
name: I2::CI::consistent_docker_compose_images
```

.github/workflows/no-incorrect-image.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v1
+        with:
+          python-version: "3.11"
+      - uses: actions/checkout@v3
+      - name: Install depenencies
+        run: pip install -r .github/scripts/ci_test/requirements.txt --no-input
+      - name: Check containers
+        run: python .github/scripts/ci_test/ci_image_scan.py
```
0x009922, line 17: I believe that in a flexible design, the script itself should not be aware of the file patterns or the exact image it should ban.

IMO, in a flexible design the script invocation should look like:

```shell
python check-compose-image.py -- docker-compose.yml --allow iroha2:stable
```

- The script doesn't care about which `.yaml` it should check
- The script doesn't know about image names, but only compares the content with the passed argument
- Multi-branch support: in `iroha2-dev` branch we could ensure that our compose files contain `iroha2:dev` image, while in `iroha2-stable` branch they contain `iroha2:stable` image. It could be achieved by using e.g. [`${{ github.ref_name }}` syntax](https://docs.github.com/en/actions/learn-github-actions/expressions#example).

0x009922 (changes requested):

.github/scripts/ci_test/ci_image_scan.py
6r1d: responded in another related comment [below](https://github.com/hyperledger/iroha/pull/4053#discussion_r1410270090)

.github/workflows/no-incorrect-image.yml
6r1d: the current script requires a manual start, I'll fix that part, though

.github/workflows/iroha2-no-incorrect-image.yml
```diff
@@ -0,0 +1,25 @@
...
+          python-version: "3.11"
+      - uses: actions/checkout@v3
+      - name: Install depenencies
+        run: pip install -r .github/scripts/ci_test/requirements.txt --no-input
+      - name: Check containers on iroha2-stable branch
+        if: github.ref == 'refs/heads/iroha2-stable'
+        run: python .github/scripts/ci_test/ci_image_scan.py --allow iroha2:stable -- docker-compose.dev.local.yml docker-compose.dev.single.yml docker-compose.dev.yml docker-compose.single.yml docker-compose.stable.single.yml docker-compose.stable.yml docker-compose.yml
+      - name: Check containers on iroha2-dev branch
+        if: github.ref == 'refs/heads/iroha2-dev'
+        run: python .github/scripts/ci_test/ci_image_scan.py --allow iroha2:dev iroha2:stable -- docker-compose.dev.local.yml docker-compose.dev.single.yml docker-compose.dev.yml docker-compose.single.yml docker-compose.stable.single.yml docker-compose.stable.yml docker-compose.yml
```
0x009922, line 25:
```suggestion
        run: python .github/scripts/ci_test/ci_image_scan.py --allow iroha2:stable -- docker-compose.stable*.yml
      - name: Check containers on iroha2-dev branch
        if: github.ref == 'refs/heads/iroha2-dev'
        run: python .github/scripts/ci_test/ci_image_scan.py --allow iroha2:dev -- docker-compose.dev*.yml
```

That is how I imagined it to work.

.github/workflows/iroha2-no-incorrect-image.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v1
+        with:
+          python-version: "3.11"
+      - uses: actions/checkout@v3
+      - name: Install depenencies
```
0x009922, line 18:
```suggestion
      - name: Install dependencies
```

.github/workflows/iroha2-no-incorrect-image.yml
6r1d: Understood, I'll fix it to pick between the globs and filenames very soon

0x009922 (approved):

0x009922 (approved):

0x009922:
As you can see, you should also update `consistency.sh`.

BAStos525 (approved):

PR #4054, opened 2023-11-14
[feature] #3974: Add subcommands into `iroha_client_cli` to edit domain metadata

pesterev:
## Description

This PR adds subcommands that set and remove metadata from a domain.

### Linked issue

Closes one of subtasks from #3974

### Benefits

The users will be able to edit domain metadata using `iroha_client_cli`.

### Checklist

### DIFF
--- a/client_cli/src/main.rs
+++ b/client_cli/src/main.rs
@@ -20,7 +20,21 @@ use iroha_crypto::prelude::*;
 use iroha_data_model::prelude::*;
 use iroha_primitives::addr::SocketAddr;
 
-/// Metadata wrapper, which can be captured from cli arguments (from user supplied file).
+/// A value wrapper that can be parsed from CLI arguments
+#[derive(Debug, Clone)]
+pub struct MetadataValue(iroha_data_model::Value);
+
+impl FromStr for MetadataValue {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        json5::from_str(s)
+            .map(Self)
+            .wrap_err("Failed to deserialize a value")
+    }
+}
+
+/// Metadata wrapper, which can be captured from CLI arguments (from user supplied file).
 #[derive(Debug, Clone)]
 pub struct Metadata(pub UnlimitedMetadata);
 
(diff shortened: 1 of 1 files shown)


client_cli/src/main.rs
```diff
@@ -420,6 +443,72 @@ mod domain {
...
+        }
+
+        impl RunArgs for Set {
+            fn run(self, context: &mut dyn RunContext) -> Result<()> {
+                let Self {
+                    id,
+                    key,
+                    value: MetadataValue(value),
+                } = self;
+                dbg!(&value);
```
Erigara, line 488: Should be removed before merge.

client_cli/src/main.rs
```diff
@@ -20,7 +20,27 @@ use iroha_crypto::prelude::*;
 use iroha_data_model::prelude::*;
 use iroha_primitives::addr::SocketAddr;
 
-/// Metadata wrapper, which can be captured from cli arguments (from user supplied file).
+/// A value wrapper that can be parsed from CLI arguments
+#[derive(Debug, Clone)]
+pub struct MetadataValue(iroha_data_model::Value);
+
+impl fmt::Display for MetadataValue {
```
Erigara, line 27: I guess this `Display` will be unused after removal of `dbg!`.

Erigara (commented):
LGTM, should be covered by tests as well. 

@pesterev you told during the daily that you don't like smt about API, what exactly?

pesterev:
@Erigara 
> should be covered by tests as well.

sure, I will do it here

> you told during the daily that you don't like smt about API, what exactly?

The problem is that we specify the metadata value as a CLI argument so it's difficult to use. Here is another way - read metadata value from a file so we should specify a path to this file as a CLI argument. What do you think about it?

VAmuzing (changes requested):

client_cli/src/main.rs
```diff
@@ -20,7 +20,21 @@ use iroha_crypto::prelude::*;
...
+/// A value wrapper that can be parsed from CLI arguments
+#[derive(Debug, Clone)]
+pub struct MetadataValue(iroha_data_model::Value);
+
+impl FromStr for MetadataValue {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        json5::from_str(s)
+            .map(Self)
```
VAmuzing, line 32: Currently all impls have something like this:
```rust
let smth: some_type = json5::from_str(..).wrap_err(..)?;
Ok(Self(smth))
```
Can we change everything to map(Self) or make it more explicit like in the example so we have a single style?

pesterev: Both options are the same and do the same thing. Both may be used. My option is shorter and straightforward to read in this case.

VAmuzing: Agree.
Let's then use your solution for other cases?
Otherwise at some point in the future we might add another option for from_str and it'll be very inconvenient.

Erigara:
Closing in favor of #4175

PR #4058, opened 2023-11-15
[fix] #4057: Fix query store message ordering issue

### LINKED ISSUE #4057: [fix]: Fix live query service ordering issue
I've discovered issues with live query service that messages was processed out of order.

Erigara:
## Description

Previously 2 separate queues were used to send `insert`/ `remove` request to live query service which lead to issues when `insert`, `remove` were one right after another so that ordering of message processing was screwed.

Now single queue is used for messsages which guarantee that messages are processed sequentially eliminating possibility of this bug.  

### Linked issue

Closes #4057 

### Benefits

Bug is solved.

### Checklist


mversic (approved):

core/src/query/store.rs
```diff
@@ -288,3 +286,47 @@ impl LiveQueryStoreHandle {
...
+        let query_store_handle = threaded_rt.block_on(async { query_store.start() });
+
+        for i in 0..10_000 {
+            let pagination = Pagination::default();
+            let fetch_size = FetchSize {
+                fetch_size: NonZeroU32::new(1),
+            };
+            let sorting = Sorting::default();
+
+            let query_output = LazyValue::Iter(Box::new((0..100).map(|_| Value::Bool(false))));
```
Arjentix, line 309: Why do you need `Box` here?

Arjentix (approved):

core/src/query/store.rs
Erigara: Because of `Box<dyn Iterator>` type in `LazyValue::Iter`

Issue #3941, opened 2023-10-02
Move all expression evaluation from host into Executor


mversic:
Executor should be the only entity that knows how to evaluate expressions. Host should have no reference to expressions, expressions are strictly an Executor concept and can be changed at runtime by updating Executor.

As a consequence evaluation step will now happen entirely before the validation step.
Currently, Validator/Executor traverses syntax tree and evaluates any expression encountered along the way. That is not good because it results in repeated evaluation of expressions in both executor and on the host. Executor should first evaluate every expression (such as `InstructionExpr`) into a concrete instruction (such as `InstructionBox`) and only then call `visit_instruction`.

This means that `visit_*` method signatures should be modified to operate on evaluated instructions

PR #4068, opened 2023-11-22
Fix ser deser loophole for evaluates to

### LINKED ISSUE #3343: Expression deserialization loophole
#2822 has removed custom deserialization of `EvaluatesTo<V>`. This was an oversight and now incorrect types can be constructed through deserialization because there is no type checking involved

Related to #1995

VAmuzing:
## Description

It appears that this field attribute allows us to use derive impl for `Deserialize` and still have compile-time check for the generic type of `EvaluatesTo<>`

### Linked issue

Closes #3343 

### Benefits

Inability to deserialize into incorrect types.

### Checklist


data_model/tests/evaluates_to_serde_fail/deser_evaluates_to_fails.rs
```diff
@@ -0,0 +1,17 @@
...
+use iroha_data_model::prelude::*;
+
+#[derive(Serialize, Deserialize)]
+struct MyTestStructEvsTo {}
+
+fn main() {
+    // Generic type does not matter in this context
+    let evals_to = EvaluatesTo::<Value>::new_unchecked(
+        Value::Bool(true),
+    );
```
mversic, line 12:
```suggestion
    let evals_to = EvaluatesTo::<Name>::new_unchecked(
        Value::Bool(true),
    );
```

and what if we have a test like this? will this deserialize successfully?

mversic: this check needs to be done at runtime. You cannot make it a compile time check

mversic: what you need to do is inspect `Expression` and verify that the contained value is indeed as declared in the type

VAmuzing: It will not.
But it will if we do it like so:
```rust
let evals_to = EvaluatesTo::<Name>::new_unchecked(
        Value::Bool(true),
);

let serialized = serde_json::to_string(&evals_to).unwrap();

// Or Result<EvaluatesTo<Name>, _>
let deserialized: Result<EvaluatesTo<Value>, _> = serde_json::from_str(&serialized);
```

mversic: > But it will if we do it like so:
> 
> ```rust
> let evals_to = EvaluatesTo::<Name>::new_unchecked(
>         Value::Bool(true),
> );
> 
> let serialized = serde_json::to_string(&evals_to).unwrap();
> 
> // Or Result<EvaluatesTo<Name>, _>
> let deserialized: Result<EvaluatesTo<Value>, _> = serde_json::from_str(&serialized);
> ```

this is what we need to prevent from happening

VAmuzing:
As the current PR might have conflicts with the #3941 I'm temporarily currently closing it.

PR #4069, opened 2023-11-22
[fix] #4065: Prevent pub key spoofing in p2p

### LINKED ISSUE #4065: [BUG]: Prevent peer public key spoofing in p2p
Atm during handshake peer blindly trust other side that it's owns private key for the public key sent during handshake. 

This should be fixed, probably by sending some kind of signature to other side to prove ownership of private key. 

Relevant code: `p2p/src/peer.rs GetKey and SendKey`.

Erigara:
## Description

Instead of sending just pub key it's now required to send pub key + signature of shared session key to prove ownership of this public key in order to prevent spoofing.

### Linked issue

Closes #4065 

### Benefits

Security bug resolved.


p2p/src/network.rs
```diff
@@ -66,7 +66,11 @@ impl<T: Pload, K: Kex + Sync, E: Enc + Sync> NetworkBaseHandle<T, K, E> {
     /// # Errors
     /// - If binding to address fail
     #[log(skip(public_key))]
-    pub async fn start(listen_addr: SocketAddr, public_key: PublicKey) -> Result<Self, Error> {
+    pub async fn start(
+        listen_addr: SocketAddr,
+        public_key: PublicKey,
+        private_key: PrivateKey,
```
mversic, line 72: so why don't we use `KeyPair`?

Erigara: Public and private key lay separately in our configuration, so i decided to preserve it this way.

cli/src/lib.rs
```diff
@@ -228,9 +228,13 @@ impl Iroha {
...
     ) -> Result<Self> {
         let listen_addr = config.torii.p2p_addr.clone();
-        let network = IrohaNetwork::start(listen_addr, config.public_key.clone())
-            .await
-            .wrap_err("Unable to start P2P-network")?;
+        let network = IrohaNetwork::start(
+            listen_addr,
+            config.public_key.clone(),
+            config.private_key.clone(),
+        )
```
mversic, line 235:
```suggestion
        let network = IrohaNetwork::start(
            listen_addr,
            KeyPair::new(config.public_key.clone(), config.private_key.clone())?,
        )
```

p2p/src/network.rs
Erigara: And why pack them if we will just unpack them later.

mversic: because in `KeyPair::new` maintains the invariant that these keys match, i.e. it makes sure they don't belong to 2 different pairs

mversic: pairs should be transferred around in `KeyPair`

Erigara:
> and it's not usually done this way

i'm open to your suggestions if you have smt on your mind

Erigara:
I realized that we already have unique shared value called `disambiguator` which was previously used to decide which of simultaneous connections to preserve.
So i will just use it instead of directly hash of session key.

DCNick3:
Looked at what TLS does for a bit:

In 1.2 they server signs various parts of the key-exchange process, but not the key itself: https://github.com/rustls/rustls/blob/1db4506dcfeb4a08b5517eafba84f16084321786/rustls/src/client/tls12.rs#L724-L726. Client's signature verification seems like a kludge, it just signs all the handshake messages: https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.8

In 1.3 they tried to mitigate some attacks regarding tricking the server into signing some partially-controlled data, so they came up with a pretty contrived scheme, but made it symmetric at least: https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.3

In case of iroha, I think the scheme of 1.2 would make sense? In our case it would just signing the public keys used in kx (`local_public_key` and `remote_public_key`)

DCNick3:
> I realized that we already have unique shared value called disambiguator which was previously used to decide which of simultaneous connections to preserve.
> So i will just use it instead of directly hash of session key.

It's only 64 bit.. I don't think it's a good idea to rely on it for security

Erigara:
> In case of iroha, I think the scheme of 1.2 would make sense? In our case it would just signing the public keys used in kx (local_public_key and remote_public_key)

Yeah, this make sense, i will do that.

p2p/src/peer.rs
```diff
@@ -374,67 +371,75 @@ mod run {
...
-                peer_id,
+                peer_addr,
+                key_pair,
                 mut connection,
             }: Self,
         ) -> Result<SendKey<E>, crate::Error> {
             let key_exchange = K::new();
-            let (local_public_key, local_private_key) = key_exchange.keypair(None)?;
-            let (algorithm, local_public_key_raw) = local_public_key.into_raw();
+            let (local_session_public_key, local_private_key) = key_exchange.keypair(None)?;
```
DCNick3, line 421: The `local_private_key` is also not the real peer private key, but just used for key exchange, so the `local_session_public_key` and `local_private_key` are now somewhat inconsistently named..

Maybe name them `kx_local_public_key` and `kx_local_private_key`? Or shorten them to `kx_local_pk` and `kx_local_sc`? Same for the remote PK

mversic: good point, this was confusing me

DCNick3: Sorry, I made a typo, it should've been `kx_local_sk`, not `kx_local_sc` :sweat:

p2p/src/peer.rs
```diff
@@ -130,7 +136,8 @@ mod run {
                     },
                 cryptographer,
             } = peer;
-            peer_id = new_peer_id;
+            peer_id = Some(new_peer_id.clone());
+            let peer_id = new_peer_id;
```
mversic, line 140:
```suggestion
            let peer_id = peer_id.insert(new_peer_id.clone());
```

mversic (changes requested):

p2p/src/peer.rs
Erigara: Indeed

mversic (approved):

DCNick3 (approved):

Issue #2585, opened 2022-08-03
Tracking Issue for Configuration Overhaul


ilchu:
## What is it

This issue serves as the central point for monitoring the ongoing overhaul of the configuration system.

## RFC

[Rendered](https://github.com/hyperledger/iroha-rfcs/blob/main/text/001-config-overhaul.md)

RFC document related work:

- https://github.com/hyperledger/iroha-rfcs/pull/3
- https://github.com/hyperledger/iroha-rfcs/pull/8

## Tasks

Issues opened before the merging of the RFC:

- #2615 
- #2801 
- #2824 
- #3367
- #3470
- #3499
- #3500 
- #3501 
- #3502 
- #3503
- #3504 
- #3505 
- #3507 
- #3107
- #3327

New subtasks/discussions/PRs:

- #4028 
- #4029 
- #4030 
- #4059 
- #4063 
  - #4076
  - #4100
- #4064 
- #4079
- #4127
- #4136
  - #4153
- #4161
  - #4239
- #4194
- #4287
- #4288
- #4291
- #4292
- #4293
- #4294
- #4295
- #4296
- #4297
- #4299
- #4300
- #4301
- #4320
- #4323
- #4278
- #4383

Tracking of the configuration reference update:

- hyperledger/iroha-2-docs#392

0x009922:
I consider this epic to be mostly done. Some issues and questions are still open; some are going to be closed by some PRs soon; but in its core the overhaul is complete.

PR #4076, opened 2023-11-23
[fix] #4063: update configuration endpoints

### LINKED ISSUE #4063: [discussion] Revise configuration endpoints according to the RFC
## Description

According to the Configuration Overhaul RFC (https://github.com/hyperledger/iroha/issues/2585), the current configuration endpoints should be revised.

### Current Design

> Iroha's API currently provides two endpoints related to configuration:
> 
> - `POST /configuration`: Allows runtime updating of the configuration. As of now, it only supports modifying the log level using a request body structured as `{"LogLevel":"<level>"}`.
> - `GET /configuration`: Serves a dual purpose; it can either return the entire configuration in a JSON format or provide documentation for a specific field.

_From the [Configuration Endpoints](https://github.com/hyperledger/iroha-rfcs/blob/main/text/001-config-overhaul.md#configuration-endpoints) section of the RFC_

### Detailed Proposal

For the reasoning please refer to the [Proposal 9](https://github.com/hyperledger/iroha-rfcs/blob/main/text/001-config-overhaul.md#proposal-9---revise-configuration-endpoints) of the RFC.

- Remove documentation fetching via API
- Use the same naming conventions as used in the config file (or ENV)
- Use `GET`/`POST` `/configuration` for fetching and updating the subset of configuration parameters. The subset and format should be the same between the two endpoints for the sake of uniformity. 
- While updating the configuration accepts _partial_ structure (i.e. user might specify only those fields they need to update), retrieval always returns the whole subset. __Note:__ this is not relevant now because we only have a single dynamic parameter.

The only dynamic parameter as of now is `logger.level` (according to the updated naming of parameters). Therefore, the following JSON request might be used to update the value with `POST /configuration`:

```json
{
  "logger": {
    "level": "DEBUG"
  }
}
```

The same will be returned by `GET /configuration`.

### Caveats

In [that place](https://hyperledger.github.io/iroha-2-docs/guide/configure/peer-management.html#_2-set-up-a-peer) of the docs it is recommended to get peers' configuration in the network through the configuration endpoint. It suggests me that probably retrieval of the configuration might include more fields than only dynamic ones.

### Also

There is an issue:

- https://github.com/hyperledger/iroha/issues/3107

, which suggests using sub-routing in a way it is used in `GET /status` endpoint. In general, I don't see a conflict here and it might be adopted for the new configuration endpoints:

<details>

<summary>Samples</summary>

Getting whole configuration:

```http
GET /configuration
```

```json
{
  "logger": {
    "level": "DEBUG"
  }
}
```

Getting a `logger` part:

```http
GET /configuration/logger
```

```json
{
  "level": "DEBUG"
}
```

Getting `logger.level` precisely:

```http
GET /configuration/logger/level
```

```json
"DEBUG"
```

Updating `logger` parameters:

```http
POST /configuration/logger
```

```json
{
  "level": "DEBUG"
}
```

</details>

0x009922:
## Description

I've implemented the changes according to the linked issue:

- Remove documentation retrieval from the API
- Normalize DTO structure with the configuration file

### Linked issue

Closes #4063 

### Benefits

A little step closer to implementing the Configuration RFC (#2585).

### Checklist

- [x] I've written unit tests for the code changes
- [x] Self-review
- [x] Open corresponding PR updating the docs
  - https://github.com/hyperledger/iroha-2-docs/pull/435
- [x] Write an integration test
- [x] Remove redundant `Result<bool>`
- [ ] Use `PATCH` instead of `POST`

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -13,7 +13,7 @@ use derive_more::{DebugCustom, Display};
 use eyre::{eyre, Result, WrapErr};
 use futures_util::StreamExt;
 use http_default::{AsyncWebSocketStream, WebSocketStream};
-use iroha_config::{client::Configuration, torii::uri, GetConfiguration, PostConfiguration};
+use iroha_config::{client::Configuration, client_api::ConfigurationSubset, torii::uri};
 use iroha_crypto::{HashOf, KeyPair};
 use iroha_data_model::{
     block::SignedBlock,
@@ -29,7 +29,6 @@ use iroha_telemetry::metrics::Status;
 use iroha_version::prelude::*;
 use parity_scale_codec::DecodeAll;
 use rand::Rng;
-use serde::de::DeserializeOwned;
 use url::Url;
 
 use self::{blocks_api::AsyncBlockStream, events_api::AsyncEventStream};
@@ -1073,13 +1072,16 @@ impl Client {
         )
     }
 
-    fn get_config<T: DeserializeOwned>(&self, get_config: &GetConfiguration) -> Result<T> {
+    /// Get value of config on peer
+    ///
+    /// # Errors
+    /// Fails if sending request or decoding fails
+    pub fn get_config(&self) -> Result<ConfigurationSubset> {
         let resp = DefaultRequestBuilder::new(
             HttpMethod::GET,
             self.torii_url.join(uri::CONFIGURATION).expect("Valid URI"),
         )
         .header(http::header::CONTENT_TYPE, APPLICATION_JSON)
-        .body(serde_json::to_vec(get_config).wrap_err("Failed to serialize")?)
         .build()?
         .send()?;
 
@@ -1097,9 +1099,9 @@ impl Client {
     ///
     /// # Errors
     /// If sending request or decoding fails
-    pub fn set_config(&self, post_config: PostConfiguration) -> Result<bool> {
-        let body = serde_json::to_vec(&post_config)
-            .wrap_err(format!("Failed to serialize {post_config:?}"))?;
+    pub fn set_config(&self, config: ConfigurationSubset) -> Result<bool> {
+        let body =
+            serde_json::to_vec(&config).wrap_err(format!("Failed to serialize {config:?}"))?;
         let url = self.torii_url.join(uri::CONFIGURATION).expect("Valid URI");
         let resp = DefaultRequestBuilder::new(HttpMethod::POST, url)
             .header(http::header::CONTENT_TYPE, APPLICATION_JSON)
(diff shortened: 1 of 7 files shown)


client/src/client.rs
```diff
@@ -1097,9 +1099,8 @@ impl Client {
     ///
     /// # Errors
     /// If sending request or decoding fails
-    pub fn set_config(&self, post_config: PostConfiguration) -> Result<bool> {
-        let body = serde_json::to_vec(&post_config)
-            .wrap_err(format!("Failed to serialize {post_config:?}"))?;
+    pub fn set_config(&self, dto: ConfigurationDTO) -> Result<bool> {
```
0x009922, line 1102: Actually, `POST /configuration` doesn't return `true` anymore, so it shouldn't be `Result<bool>` here.

config/src/client_api.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use iroha_config_base::runtime_upgrades::{Reload, ReloadError};
+use iroha_data_model::Level;
+use serde::{Deserialize, Serialize};
+
+use super::{iroha::Configuration as BaseConfiguration, logger::Configuration as BaseLogger};
+
+/// Subset of [`super::iroha`] configuration.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+pub struct ConfigurationDTO {
```
DCNick3, line 20: Maybe call it `PartialConfiguration` or `DynamicConfiguration`?

0x009922: I thought that `DTO` suffix emphasizes the purpose to use this structure to interact with the client side.

`Partial` seems to be too general here (like `Proxy`s we have already), while `Dynamic` might semantically fit.

config/src/client_api.rs
```diff
@@ -0,0 +1,82 @@
...
+            .max_log_level
+            .reload(self.logger.max_log_level)
+    }
+}
+
+/// Subset of [`super::logger`] configuration.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+struct Logger {
+    #[allow(missing_docs)]
+    max_log_level: Level,
```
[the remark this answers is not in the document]
0x009922: I've added a comment after the module-level documentation, saying that this implementation is temporary and should be refactored in the future with the rest of the configuration structures and logic.

Goal of this PR is to make a little step towards config rfc, and to be able to remove auto-generated documentation entirely.

I think that in future this structure shouldn't be written by hand at all, but instead generated by a procmacro from the same structure used for the config file. For now I am avoiding refactoring the configuration macros, but making this little step towards it (very indirectly).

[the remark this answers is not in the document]
DCNick3: Ok, fair enough. If this isn't supposed to be the way to move forward, but just a temporary implementation - it's fine.

0x009922:
The logger & dynamic reloads architecture needs to be refactored first.

0x009922:
Close in favor of

- #4100

PR #4089, opened 2023-11-27
[feature] #3941, #3612: Remove expressions

### LINKED ISSUE #3612: [suggestion] Make illegal instructions unrepresentable

### LINKED ISSUE #3941: Move all expression evaluation from host into Executor

Arjentix:
## Description

The main idea was to move all expressions into executor. After a discussion with @mversic we decided not to implement any of them inside `default_executor`, but to have an example executor to show PoC.

- All expressions were removed
- `#[model]` together with `ffi` have become useless for data_model because in future clients should link against executor_data_model. So I partly removed them. This might need future investigation

I wii squash all my commits before merge.

### Linked issue

- Closes #3941 
- Closes #3612 

### Benefits

- Much less code
- No stillborn expression system inside data_model and core
- Types for ISI now can't contain unsupported parameters

### Follow up tasks

- Reconsider ffi for data_model
- Make block content opaque
- Write example executor with expressions or even DSL

### Checklist


configs/peer/executor.wasm
mversic, line 1: nice size reduction

data_model/src/isi.rs
```diff
@@ -4,865 +4,953 @@
...
+
+    /// Constructs a new [`BurnBox`] to burn [`Asset`] of [`BigQuantity`] type.
+    pub fn asset_big_quantity(quantity: u128, asset_id: <Asset as Identifiable>::Id) -> Self {
+        Self::Asset(AssetBurnBox::BigQuantity(Burn::new(quantity, asset_id)))
+    }
+
+    /// Constructs a new [`BurnBox`] to burn [`Asset`] of [`Fixed`] type.
+    pub fn asset_fixed(quantity: Fixed, asset_id: <Asset as Identifiable>::Id) -> Self {
+        Self::Asset(AssetBurnBox::Fixed(Burn::new(quantity, asset_id)))
     }
```
mversic, line 809:
```suggestion
    /// Constructs a new [`BurnBox`] to burn [`Asset`]
    pub fn asset(quantity: impl Into<AssetType>, asset_id: <Asset as Identifiable>::Id) -> Self {
        Self::Asset(AssetBurnBox::Quantity(Burn::new(quantity, asset_id)))
    }
```

is this not possible?

smart_contract/executor/src/default.rs
```diff
@@ -1170,7 +1159,9 @@ pub mod permission_token {
 
     macro_rules! impl_validate {
         ($executor:ident, $authority:ident, $self:ident, $method:ident) => {
-            let token = $self.object;
+            // If executor could visit instructions by value instead of reference,
+            // we could remove this clone
```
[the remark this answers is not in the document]
mversic: but I would just remove the comment. It's no different than any of the other redundant clones

mversic, line 1163:
```suggestion
            // TODO: https://github.com/hyperledger/iroha/issues/4082
```

schema/derive/tests/ui_pass/derive_into_schema.rs
```diff
@@ -42,4 +42,14 @@ pub enum Enum {
     FortyTwo,
 }
 
+pub trait Trait {
+    type Assoc;
+}
+
+#[derive(IntoSchema)]
+#[schema(bounds = "T: Trait, T::Assoc: IntoSchema")]
```
mversic, line 50: this API looks redundant

default_executor/src/lib.rs
```diff
@@ -19,11 +19,10 @@ static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeList
 /// # Warning
 ///
 /// The defaults are not guaranteed to be stable.
-#[derive(Clone, Constructor, Debug, ValidateEntrypoints, ExpressionEvaluator, Validate, Visit)]
+#[derive(Clone, Constructor, Debug, ValidateEntrypoints, Validate, Visit)]
```
mversic, line 22:
```suggestion
#[derive(Debug, Clone, Constructor, Visit, Validate, ValidateEntrypoints)]
```

Also, unrelated. I dislike seeing `ValidateEntrypoints`. Seems out of place

data_model/src/visit.rs
```diff
@@ -251,436 +274,188 @@ pub fn visit_wasm<V: Visit + ?Sized>(
...
+        visitor, authority, isi,
+        UnregisterBox {
+            Peer => visit_unregister_peer(),
+            Domain => visit_unregister_domain(),
+            Account => visit_unregister_account(),
+            AssetDefinition => visit_unregister_asset_definition(),
+            Asset => visit_unregister_asset(),
+            Role => visit_unregister_role(),
+            Trigger => visit_unregister_trigger(),
+        }
```
mversic, line 345: I'm not sure this is much better than a plain `match` statement. There is no real code reduction, only obfuscation

data_model/Cargo.toml
```diff
@@ -55,7 +55,7 @@ base64 = { workspace = true, features = ["alloc"] }
 once_cell = { workspace = true, optional = true }
 
 [dev-dependencies]
-iroha_client = { workspace = true }
+# iroha_client = { workspace = true }
```
mversic, line 58:
```suggestion
```

data_model/src/trigger.rs
```diff
@@ -41,7 +41,6 @@ pub mod model {
         IntoSchema,
     )]
     #[getset(get = "pub")]
-    #[ffi_type]
```
mversic, line 44: I don't get this removal of `ffi_type`. Can't user use `TriggerId` in their smart contract?

client/tests/integration/multisignature_account.rs
```diff
@@ -17,24 +17,20 @@ fn transaction_signed_by_new_signatory_of_account_should_pass() -> Result<()> {
     // Given
     let account_id: AccountId = "alice@wonderland".parse().expect("Valid");
     let asset_definition_id: AssetDefinitionId = "xor#wonderland".parse().expect("Valid");
-    let create_asset = RegisterExpr::new(AssetDefinition::quantity(asset_definition_id.clone()));
+    let create_asset =
+        RegisterBox::asset_definition(AssetDefinition::quantity(asset_definition_id.clone()));
```
mversic, line 21:
```suggestion
        RegisterBox::asset_definition(AssetType::Quantity, asset_definition_id.clone());
```

I think this makes a better API

data_model/src/isi.rs
```diff
@@ -4,865 +4,953 @@
...
-        /// Where to set this key value.
-        #[serde(flatten)]
-        pub object_id: EvaluatesTo<IdBox>,
-        /// Key string.
-        pub key: EvaluatesTo<Name>,
-        /// Object to set as a value.
-        pub value: EvaluatesTo<Value>,
+impl SetKeyValueBox {
+    /// Constructs a new [`SetKeyValueBox`] for a [`Domain`] with the given `key` and `value`.
+    pub fn domain(domain_id: <Domain as Identifiable>::Id, key: Name, value: Value) -> Self {
```
mversic, line 508:
```suggestion
    pub fn domain(domain_id: <Domain as Identifiable>::Id, key: Name, value: impl Into<Value>) -> Self {
```

do this for all `Value`s. This makes user's life a lot easier

data_model/src/isi.rs
```diff
@@ -4,865 +4,953 @@
...
-        key: K,
+    /// Constructs a new [`MintBox`] to mint [`Asset`] of [`BigQuantity`] type.
+    pub fn asset_big_quantity(quantity: u128, asset_id: <Asset as Identifiable>::Id) -> Self {
+        Self::Asset(AssetMintBox::BigQuantity(Mint::new(quantity, asset_id)))
+    }
+
+    /// Constructs a new [`MintBox`] to mint [`Asset`] of [`Fixed`] type.
+    pub fn asset_fixed(quantity: Fixed, asset_id: <Asset as Identifiable>::Id) -> Self {
+        Self::Asset(AssetMintBox::Fixed(Mint::new(quantity, asset_id)))
+    }
```
mversic, line 752: the same as for `BurnBox`

data_model/src/isi.rs
mversic: ok, I see now. Can we do it like this:

```suggestion
    /// Constructs a new [`BurnBox`] to burn [`Asset`]
    pub fn asset_quantity(quantity: impl Into<AssetValue>, asset_id: <Asset as Identifiable>::Id) -> Self {
        match quantity.into() {
                AssetType::Quantity(quantity) => Self::Asset(AssetBurnBox::Quantity(Burn::new(quantity, asset_id))),
                AssetType::BigQuantity(quantity) => Self::Asset(AssetBurnBox::Quantity(Burn::new(quantity, asset_id))),
                AssetType::Fixed(quantity) => Self::Asset(AssetBurnBox::Quantity(Burn::new(quantity, asset_id))),
        }
    }
```

this is how I would prefer doing it

core/src/smartcontracts/isi/tx.rs
```diff
@@ -66,7 +65,7 @@ impl ValidQuery for FindAllTransactions {
                 .flat_map(BlockTransactionIter::new)
                 .map(|tx| TransactionQueryOutput {
                     block_hash: tx.block_hash(),
-                    transaction: tx.value(),
+                    transaction: Box::new(tx.value()),
```
mversic, line 68: why do we have to box it now?

core/src/smartcontracts/isi/mod.rs
```diff
@@ -67,289 +65,158 @@ impl Execute for InstructionExpr {
...
+    fn execute(
+        self,
+        authority: &AccountId,
+        wsv: &mut WorldStateView,
+    ) -> std::prelude::v1::Result<(), Error> {
+        match_all! {
+            self, authority, wsv,
+            PublicKey,
+            SignatureCheckCondition,
+        }
```
mversic, line 122: again, this macro hardly brings us anything but obfuscation

core/src/smartcontracts/isi/block.rs
```diff
@@ -43,15 +42,12 @@ impl ValidQuery for FindAllBlockHeaders {
 impl ValidQuery for FindBlockHeaderByHash {
     #[metrics(+"find_block_header")]
     fn execute(&self, wsv: &WorldStateView) -> Result<BlockHeader, QueryExecutionFail> {
-        let hash = wsv
-            .evaluate(&self.hash)
-            .wrap_err("Failed to evaluate hash")
-            .map_err(|e| QueryExecutionFail::Evaluate(e.to_string()))?;
+        let hash = &self.hash;
```
mversic, line 45:
```suggestion
        let hash = self.hash;
```

core/src/smartcontracts/isi/mod.rs
```diff
@@ -483,11 +292,11 @@ mod tests {
         let account_id = AccountId::from_str("alice@wonderland")?;
         let (public_key, _) = KeyPair::generate()?.into();
         let asset_definition_id = AssetDefinitionId::from_str("rose#wonderland")?;
-        RegisterExpr::new(Domain::new(DomainId::from_str("wonderland")?))
+        RegisterBox::domain(Domain::new(DomainId::from_str("wonderland")?))
```
mversic, line 295:
```suggestion
        RegisterBox::domain("wonderland")?
```

doesn't this make more sense? We need to simplify API. The same goes for others

Erigara:
> #[model] together with ffi have become useless for data_model because in future clients should link against executor_data_model. So I partly removed them. This might need future investigation

I'm not sure about that because executor and host has to interact somehow and in this case `ffi` could be a way to go, no?

mversic:
> > #[model] together with ffi have become useless for data_model because in future clients should link against executor_data_model. So I partly removed them. This might need future investigation
> 
> I'm not sure about that because executor and host has to interact somehow and in this case `ffi` could be a way to go, no?

no, the point of FFI was to reduce smart contract size. Executor size is irrelevant because upgrading it is a rare operation. So we will use FFI only for communication between smart contract and executor

data_model/src/query/mod.rs
```diff
@@ -179,12 +179,11 @@ pub mod model {
         Debug, Clone, PartialEq, Eq, Getters, Decode, Encode, Deserialize, Serialize, IntoSchema,
     )]
     #[getset(get = "pub")]
-    #[ffi_type]
     pub struct TransactionQueryOutput {
         /// The hash of the block to which `tx` belongs to
         pub block_hash: HashOf<SignedBlock>,
-        /// Transaction
-        pub transaction: TransactionValue,
+        /// Transaction. Boxed to not to have infinite size.
```
Erigara, line 185: Why would it have infinite size? Both fields of `TransactionValue` are sized, no?

schema/derive/tests/ui_pass/derive_into_schema.rs
Arjentix: It's required for instructions. Check out `iri.rs`

schema/gen/src/lib.rs
```diff
@@ -390,6 +354,19 @@ types!(
...
+    MintBox,
+    AccountMintBox,
+    AssetMintBox,
+    BurnBox,
+    AssetBurnBox,
+    TransferBox,
+    AssetTransferBox,
+    GrantBox,
+    RevokeBox,
 );
```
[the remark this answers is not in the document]
Arjentix: Will do. I just use some online tool

default_executor/src/lib.rs
Arjentix: > Also, unrelated. I dislike seeing ValidateEntrypoints. Seems out of place

What would you prefer instead?

data_model/src/visit.rs
Arjentix: Agree, that's some overengineering

client/tests/integration/burn_public_keys.rs
```diff
@@ -27,7 +27,7 @@ fn submit(
 }
 
 fn get(client: &Client, hash: HashOf<SignedTransaction>) -> TransactionValue {
-    client
+    *client
```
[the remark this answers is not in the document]
Arjentix: It's from `clippy fix`, haven't caught that

data_model/src/isi.rs
mversic: > But also will bloat size of code.

yes, but that's a price to be paid for nicer API. For any API that is not exposed to users I would say don't use `impl Into<T>`, but for API exposed to users prefer using generics

data_model/src/trigger.rs
Arjentix: I was just removing `#[model]` macro from the places which broke my compilation. Although it can be restored back. 

Do you think we need model and ffi in data_model?

mversic: if a user can access these structs through both smart contract and client, then yes we need this back

client/tests/integration/multisignature_account.rs
Arjentix: Well initially with `*Box` isi constructors I wanted to reduce boilerplate with sub-boxes.

In that particular case it's done for uniformity. Not sure we should take parameters out of `AssetDefinition` and place them into `RegisterBox`. Kind of abstraction break

data_model/src/isi.rs
Arjentix: It's a double-edged sword. Having a lot of generic may confuse the user (e. g. some of our `client` methods).

But in this case I agree, will do

core/src/smartcontracts/isi/tx.rs
Arjentix: Because without expressions `Value` has become self-containing with unlimited size. `Value` contains `TransactionValue` which contains `Value`. So I had to put `Box` somewhere

client/benches/tps/utils.rs
```diff
@@ -265,42 +267,19 @@ impl MeasurerUnit {
...
-                FindAssetQuantityById::new(asset_id(self.name)).into(),
-            )),
-            1_u32,
-        );
-        let transfer_rose = TransferExpr::new(
-            asset_id(self.name),
-            1_u32.to_value(),
-            account_id(self.next_name),
-        );
+        std::iter::once(self.mint()).chain([self.mint(), self.transfer()].into_iter().cycle())
```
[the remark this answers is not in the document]
Arjentix: I don't get your point. Here `mint()` and `transfer()` are parts of the same transaction

configs/peer/genesis.json
```diff
@@ -77,7 +77,7 @@
       },
       {
         "Register": {
-          "NewAssetDefinition": {
+          "AssetDefinition": {
```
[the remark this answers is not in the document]
Arjentix: This comes from the definition of `RegisterBox` and the fact that `object` inside `Register` is marked as `transparent`.

I think leaving `RegisterBox` definition is better in terms of compliance with other `*Box`'es

Erigara:
@mversic @Arjentix 
Ok, i realized that i probably confuse things. 
Ffi would be used to dynamically link smart contract to data model, but interaction between host and smart contract would still by encoding and decoding data in wasm  linear memory. Is that correct?

schema/derive/tests/ui_pass/derive_into_schema.rs
mversic: I understand it's required, but why do we need to specify these bounds explicitly? At least the `T: Trait` can be inferred, if not both

Arjentix: Well, it's more explicit as macro won't generate anything else, just user input. Also simplier to implement

data_model/src/query/mod.rs
Arjentix: Answered [here](https://github.com/hyperledger/iroha/pull/4089#discussion_r1410278751)

Arjentix:
> So this means that iroha data model should be ffi compatible as well?

Yes, looks like we are back. Also noted by Marin [here](https://github.com/hyperledger/iroha/pull/4089#discussion_r1410238795)

data_model/src/trigger.rs
Arjentix: Done

data_model/src/isi.rs
```diff
@@ -79,790 +65,829 @@ pub mod model {
...
-impl RemoveKeyValueExpr {
-    /// Construct [`RemoveKeyValueExpr`].
-    pub fn new<I: Into<EvaluatesTo<IdBox>>, K: Into<EvaluatesTo<Name>>>(
-        object_id: I,
-        key: K,
+impl MintBox {
+    /// Constructs a new [`MintBox`] to mint [`PublicKey`] for [`Account`].
+    pub fn account_public_key(public_key: PublicKey, account_id: AccountId) -> Self {
+        Self::Account(AccountMintBox::PublicKey(Mint::new(public_key, account_id)))
+    }
```
mversic, line 690: I've been looking at the API and have been very unhappy to see that users have to deal with `MintBox` or any other `*Box`. Let me present an idea how we can simplify it:

```suggestion
impl Mint {
    /// Constructs a new [`MintBox`] to mint [`PublicKey`] for [`Account`].
    pub fn account_public_key(public_key: PublicKey, account_id: AccountId) -> MintBox {
        MintBox::Account(AccountMintBox::PublicKey(Mint::new(public_key, account_id)))
    }
```

Can we define constructors on the `Mint` struct and have it return `MintBox`?

mversic: do we even need `MintBox::account_public_key`. Can't we just have `Mint::new` and an `impl `Into<InstructionBox> for Mint`?

client/tests/integration/transfer_asset.rs
```diff
@@ -34,19 +50,20 @@ fn simulate_insufficient_funds() {
         Fixed::try_from(20_f64).expect("Valid"),
         &Fixed::try_from(200_f64).expect("Valid"),
         AssetDefinition::fixed,
+        MintBox::asset_fixed,
+        TransferBox::asset_fixed,
         10_800,
     )
 }
 
 // TODO add tests when the transfer uses the wrong AssetId.
```
mversic, line 59: can you remove this and make a ticket?

configs/peer/genesis.json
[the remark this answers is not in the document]
mversic: I have the same question as @Erigara. This should be `NewAssetDefinition` and `NewAccount` and `NewDomain`. There is a reason we have these structs. I may be wrong

core/src/smartcontracts/isi/tx.rs
mversic: since there is no more expressions, I don't think we need `Value` in that way anymore. We can consider this in a separate issue

data_model/build.rs
mversic, line 1: this should be restored?

data_model/src/query/mod.rs
mversic:
```suggestion
        /// Transaction
```

we don't need these vague comments. It will be clear to anyone who tries to remove the `Box` why it's needed

data_model/src/query/mod.rs
```diff
@@ -806,28 +807,28 @@ pub mod asset {
 
     impl FindAssetById {
         /// Construct [`FindAssetById`].
-        pub fn new(id: impl Into<EvaluatesTo<AssetId>>) -> Self {
+        pub fn new(id: impl Into<AssetId>) -> Self {
```
mversic, line 810: IMO all these constructors can be derived now. I don't see much point in having `Into<AssetId>` or similar

mversic (changes requested):

schema/derive/tests/ui_pass/derive_into_schema.rs
mversic: ok, stay lazy and make it a ticket

data_model/src/isi.rs
Arjentix: `*Box` aren't just for serialization, they represent the entrypoint to our instructions

Arjentix: I don't very like the approach of constructing different types in another type constructor. I'd like to open a type documentation and see all possible creation methods

mversic: > *Box aren't just for serialization, they represent the entrypoint to our instructions

what do you mean by entrypoint for instructions?

> I don't very like the approach of constructing different types in another type constructor. I'd like to open a type documentation and see all possible creation methods

I can agree with that. That's why I suggested to implement `From<Mint> for InstructionBox`

configs/peer/genesis.json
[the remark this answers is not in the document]
Arjentix: It **is** in fact `NewAssetDefinition` type. Just json serialization works like that. It displays `RegisterBox::AssetDefinition` variant, not the type it contains. I can make it to be displayed like `New*`, but does it make sence to introduce this complexity in json?

default_executor/src/lib.rs
Arjentix: - Created #4123

core/src/smartcontracts/isi/tx.rs
Arjentix: - Created #4124

client/tests/integration/transfer_asset.rs
Arjentix: - Created #4125

data_model/src/isi.rs
Arjentix: I tried `Mint::new()` approach and it doesn't work well because of our generic usage style (type has generic `T`, but takes `T::Id` in `new()`). It forces user to **always** specify generics like `Mint::<u32, Asset>::new(...)`.

So I decided to go with `Mint::asset_quantity()` approach (note that it's not `MintBox`). It has the following advantages:

1. Doesn't require expicit generics
2. We can identify the exact instruction type by reading function name
3. Easy to search in the project for exact instructions usages

`From` approaches works normally. So now users are disallowed to use types like `MintBox` as they serve only for (de-)serialization

schema/derive/tests/ui_pass/derive_into_schema.rs
Arjentix: I tried to implement auto-infering. A stuck at compiling stage and after some time got errors like this:

> error[E0275]: overflow evaluating the requirement `model::Value: iroha_schema::IntoSchema`
  --> data_model/src/account.rs:96:9
   |
96 |         IntoSchema,
   |         ^^^^^^^^^^
   |
   = note: required for `BTreeMap<name::model::Name, model::Value>` to implement `iroha_schema::IntoSchema`
   = note: 1 redundant requirement hidden
   = note: required for `metadata::model::Metadata` to implement `iroha_schema::IntoSchema`
   = help: see issue #48214
   = help: add `#![feature(trivial_bounds)]` to the crate attributes to enable
   = note: this error originates in the derive macro `IntoSchema` (in Nightly builds, run with -Z macro-backtrace for more info)

Looks endless loop of recursive bounds evaluation.

See also https://github.com/rust-lang/rust/issues/48214

Arjentix: I think that's the reason why `derive_more::Display` doesn't do that.

My current implementation of `bounds` feels the same as `bounds` from `derive_more::Display`. Do you think it makes sense to infer `T: Trait` and mix that with user input?

client/tests/integration/add_account.rs
```diff
@@ -13,14 +13,14 @@ fn client_add_account_with_name_length_more_than_limit_should_not_commit_transac
     let pipeline_time = super::Configuration::pipeline_time();
 
     let normal_account_id: AccountId = "bob@wonderland".parse().expect("Valid");
-    let create_account = RegisterExpr::new(Account::new(normal_account_id.clone(), []));
+    let create_account = Register::account(Account::new(normal_account_id.clone(), []));
```
mversic, line 16: What I want to see is this:
```suggestion
    let create_account = Register::new(Account::new(normal_account_id.clone(), []));
```

continuing [this](https://github.com/hyperledger/iroha/pull/4089#discussion_r1421831492) discussion. I strongly disagree with what you wrote there. I don't want to see API surface grow. It can always grow if there is a need, but it cannot ever shrink. We should start with small API and consider extending it. What I would rather do is create a ticket to discuss this and use `<ISI>::new` everywhere for now

you said:
> 1. Doesn't require expicit generics

 why is requiring explicit generics bad? But nevertheless, explicit generics are only required if the type you are providing to a function is not known. You don't have to use turbofish if the type of arguments is known

Arjentix: >  But nevertheless, explicit generics are only required if the type you are providing to a function is not known. You don't have to use turbofish if the type of arguments is known

Yes, but as I said, in most of the time `<ISI>::new()` accepts not just a generic, but an associated type of this generic. That requires to **always** use turbo-fish. To me turbo-fish is less readable than a properly named method.

> I don't want to see API surface grow. It can always grow if there is a need, but it cannot ever shrink. We should start with small API and consider extending it.

I don't understand, where did it grow? All possible instructions are still the same, I haven't expand them.

I think the API is more clear with construction methods because user can open documentation for any instruction and easily see all possible *variants*. They don't need to understand what the heck is `<T as Registered>::With`. That can be a pain especially then for `Account` it's `NewAccount`, but for `Asset` it's just `Asset`.

> What I would rather do is create a ticket to discuss this and use <ISI>::new everywhere for now

Do you still consider this is an option keeping in mind that **all** i.e. transfers will look like:

```rust
let transfer = Transfer::<Asset, u32, Account>::new(asset_id, 1_u32, account_id);
```

Just... Should it be that hard?

data_model/src/isi.rs
```diff
@@ -79,790 +71,1151 @@ pub mod model {
...
 
-impl SetParameterExpr {
-    /// Construct [`SetParameterExpr`].
-    pub fn new<P: Into<EvaluatesTo<Parameter>>>(parameter: P) -> Self {
-        Self {
-            parameter: parameter.into(),
-        }
+isi_box! {
+    /// Enum with all supported [`Mint`] instructions.
+    pub enum MintBox {
```
mversic, line 1122: do you think we need these `*Box` structs other than `InstructionBox`. Would it make sense to inline it all to `InstructionBox`. What I'm asking is, do we have to group all `Mint` instructions under `MintBox` or can we just make them variants of `InstructionBox` directly

client/tests/integration/add_account.rs
mversic: > Yes, but as I said, in most of the time <ISI>::new() accepts not just a generic, but an associated type of this generic. That requires to always use turbo-fish. To me turbo-fish is less readable than a properly named method.

I see. Let's have a vote. We have 2 options:
1.
```rs
let transfer = Transfer::asset_quantity(asset_id, 1_u32, account_id);
```

vs

2.
```rs
let transfer: Transfer<Asset, u32, Account> = Transfer::new(asset_id, 1_u32, account_id);
```

data_model/src/isi.rs
```diff
@@ -79,790 +71,1151 @@ pub mod model {
...
+    impl Instruction for Burn<u32, Asset> {}
+    impl Instruction for Burn<u128, Asset> {}
+    impl Instruction for Burn<Fixed, Asset> {}
+    impl Instruction for Burn<u32, Trigger<TriggeringFilterBox>> {}
+
+    impl Instruction for Transfer<Account, DomainId, Account> {}
+    impl Instruction for Transfer<Account, AssetDefinitionId, Account> {}
+    impl Instruction for Transfer<Asset, u32, Account> {}
+    impl Instruction for Transfer<Asset, u128, Account> {}
+    impl Instruction for Transfer<Asset, Fixed, Account> {}
```
mversic, line 153: we can't transfer store assets?

data_model/src/isi.rs
mversic: maybe we should? flat representation will also take less memory

data_model/src/isi.rs
Arjentix: Interesting... Nobody has mentioned that before. Looks like we have no impls in core...

I'll open a ticket

data_model/src/isi.rs
Erigara: I also thought of such approach

client/tests/integration/add_account.rs
Erigara: Maybe we can remove associated types? 
In such case i would definitely vote for 2.

data_model/src/isi.rs
Arjentix: - Created #4137

data_model/src/isi.rs
Arjentix: Maybe...

The only downside I can see (except repeating instruction names) is that in the Executor there will be no simple way to deny all registrations for example, as will be no `visit_register(RegisterBox)`. Same for all other boxes

client/tests/integration/add_account.rs
mversic: let's leave it as it is

data_model/src/isi.rs
mversic: ok, let's leave this as is then

mversic (approved):

client/benches/tps/utils.rs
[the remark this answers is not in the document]
Arjentix: Indeed, you're right.

I think removing `transfer` migh be a fix, will do that. At the end of the day our goal here is to measure tps. Just mint should be enough

mversic (dismissed):

smart_contract/executor/src/default.rs
```diff
@@ -86,224 +80,87 @@ pub fn visit_transaction<V: Validate + ?Sized>(
...
     #[allow(clippy::needless_pass_by_value)]
     pub fn visit_unregister_peer<V: Validate + ?Sized>(
         executor: &mut V,
         authority: &AccountId,
-        _isi: Unregister<Peer>,
+        isi: &Unregister<Peer>,
     ) {
         if is_genesis(executor) {
-            pass!(executor);
+            execute!(executor, isi);
```
Erigara, line 160: Great, i need this change for permission token task.

Erigara (approved):
LGTM

mversic (approved):

Erigara (approved):

Issue #4090, opened 2023-11-28
Enable Iroha build on stable


mversic:
atm when building workspace we require "nightly" rust toolchain. This is only because wasm blobs in client are being built for testing purposes. This can be bypassed with `IROHA_SKIP_WASM_CHECKS=true cargo build`.

Iroha should build on stable toolchain by default to ease user experience. If "nightly" is required it should be opt-in, not opt-out. Also, in CI we must make sure that all wasm tests keep running

mversic:
closed in favor of #3858

Issue #3858, opened 2023-09-01
[suggestion] Remove `cargo fmt` from the WASM Builder


0x009922:
## Description

I raised a discussion originally here: https://github.com/hyperledger/iroha/pull/3841#discussion_r1312594041

> I'm puzzled about the inclusion of `cargo fmt` in the build process. While it makes sense for `wasm_builder` to handle `cargo check` and `cargo build` due to specific rustc parameters, `cargo fmt` doesn't seem to need such configuration. So why is it part of the build? Is formatting not the developer's responsibility? Are there historical reasons for this design choice?

@ilchu suggested that @Arjentix might help with clarifying it.

### Proposed solution

Remove `cargo fmt` from `iroha_wasm_builder` crate.

PR #4096, opened 2023-11-28
[fix] #4090 #3858: Fix having to pass IROHA_SKIP_WASM_CHECKS env variable with true

Asem-Abdelhady:
## Description

Change the checking parts in build.rs files to check if the env is passed during the build. this should only build the WASM builder if the code snippet will be used and in our case it will not be because the env variable is not passed.


DCNick3:
If we are gonna change that, we should at least rename the variable to have the inverted meaning, smth like `IROHA_ENABLE_WASM_CHECKS`

cli/build.rs
```diff
@@ -13,7 +13,7 @@ fn main() -> Result<()> {
     // HACK: used by Nix, since at the moment
     // the checks are a process that's hard to accomodate
     // in Nix environment
-    if std::option_env!("IROHA_SKIP_WASM_CHECKS").is_none() {
+    if std::option_env!("IROHA_SKIP_WASM_CHECKS").is_some() {
         check_default_executor()?;
     }
```
mversic, line 18:
```suggestion
```

IMO we should remove this completely

client/build.rs
```diff
@@ -15,7 +15,7 @@ fn main() -> Result<()> {
     // HACK: used by Nix, since at the moment
     // the checks are a process that's hard to accomodate
     // in Nix environment
-    if std::option_env!("IROHA_SKIP_WASM_CHECKS").is_none() {
+    if std::option_env!("IROHA_SKIP_WASM_CHECKS").is_some() {
         check_all_smartcontracts()?;
     }
```
mversic, line 20: is it possible to run this check only if run with `dev` profile?

DCNick3: `PROFILE` env variable is set for build scripts: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts, so it is technically possible

mversic: what about `cfg(test)`? that is probably not accessible.

Because in fact we would only want smart_contracts built when testing

cli/build.rs
mversic: we build it manually, you can remove it

Asem-Abdelhady: Ok then I will remove the env variable checking from both files

mversic: no, just this one

mversic: or if it's possible make sure that these smartcontracts are built only when running `cargo test`

Asem-Abdelhady: Ok I will try test cargo feature

Asem-Abdelhady: what about checking if the PROFILE variable is set to test? would that be okay?

DCNick3: There's no `test` profile, `cfg(test)` is actually orthogonal to the profile used (meaning, you can run unit tests in release profile).

Also, I an not sure that `cfg(test)` will work the intended way in `build.rs`..

client/build.rs
DCNick3: Yeah, I think `cfg(test)` wouldn't work in a `build.rs`

But it can be made into an integration test, which seems to be closer to what we actually want?

mversic: > But it can be made into an integration test, which seems to be closer to what we actually want?

can you explain what you mean?

DCNick3: > can you explain what you mean?

remove client's `build.rs`, put the following into `client/tests/check_all_smart_contracts.rs`:

```rust
use eyre::Result;

const TEST_SMARTCONTRACTS_DIR: &str = "tests/integration/smartcontracts";

#[test]
fn check_all_smartcontracts() -> Result<()> {
    iroha_wasm_builder::Builder::new(TEST_SMARTCONTRACTS_DIR)
        .format()
        .check()
}
```

This will invoke `cargo check` on smart contracts as a separate test, which seems to be actually close to what we want to do (check it when testing).

DCNick3: Yeah, I would agree, it should something done deliberately

Asem-Abdelhady: image
 I could see there was a PROFILE variable for test as shown in the picutre above or did I get something wrong here?

DCNick3: > I could see there was a PROFILE variable for test as shown in the picutre above or did I get something wrong here?

The `PROFILE` will indeed be passed to the `build.rs` script, but it would have a value of `test`, it will be `debug`, `release` or another value in case of custom profiles.

Profiles are independent from whether the crate is being built in test mode (what `#[cfg(test)]` ultimately checks).

It's actually by design that there's no way to detect whether crate is being built in test mode in `build.rs`: https://github.com/rust-lang/cargo/issues/4789

mversic: Look [here](https://doc.rust-lang.org/cargo/reference/profiles.html). There is a `test` profile

mversic:
related #3858

client/build.rs
DCNick3: UPD: apparently I was wrong, there _is_ a profile called `test`. However it's a bit weird: it seems that it's only used for an actual test executable (see https://github.com/rust-lang/cargo/issues/8904#issuecomment-734881177) and does not get passed to `build.rs` according to my testing

mversic:
we should remove formatting and make sure in CI static analysis that we check formatting of `client/tests/integration/smartcontracts` workspace and of `default_executor`

client/build.rs
```diff
@@ -24,6 +24,5 @@ fn main() -> Result<()> {
 
 fn check_all_smartcontracts() -> Result<()> {
     iroha_wasm_builder::Builder::new(TEST_SMARTCONTRACTS_DIR)
-        .format()
         .check()
```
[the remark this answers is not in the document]
mversic: just from the two `build.rs` scripts

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -4,10 +4,10 @@ on:
...
     branches: [iroha2-dev]
     paths:
-      - '**.rs'
-      - '**.json'
-      - '**.toml'
-      - '.github/workflows/**.yml'
+      - "**.rs"
+      - "**.json"
+      - "**.toml"
+      - ".github/workflows/**.yml"
```
mversic, line 10: does this make a difference?

.github/workflows/iroha2-dev.yml
```diff
@@ -8,6 +8,26 @@ env:
...
+          cd ./default_executor
+          cargo fmt -- --check
+          cargo check
+
+      - name: Format check Client
+        run: |
+          cd ./client/tests/integration/smartcontracts
+          cargo fmt -- --check
+          cargo check
+
```
mversic, line 30: I think you should put this into `iroha2-dev-pr-static.yml`

.github/workflows/iroha2-dev.yml
```diff
@@ -8,6 +8,26 @@ env:
...
+      image: hyperledger/iroha2-ci:nightly-2023-06-25
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Format check Default Executor
+        run: |
+          cd ./default_executor
+          cargo fmt -- --check
+          cargo check
```
mversic, line 23:
```suggestion
```

this is not necessary

.github/workflows/iroha2-dev.yml
mversic: I think it would be sufficient to just replicate the logic in `iroha2-dev-pr-static.yml`

.github/workflows/iroha2-dev-pr.yml
Asem-Abdelhady: I believe it would be safe to safely remove paths here so no it should not make a difference

.github/workflows/iroha2-dev.yml
Asem-Abdelhady: but from what I understood of our github actions so far isn't the static action's funcitonality analysis only?

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -3,11 +3,6 @@ name: I2::Dev::Tests
 on:
   pull_request:
     branches: [iroha2-dev]
-    paths:
-      - '**.rs'
-      - '**.json'
-      - '**.toml'
-      - '.github/workflows/**.yml'
```
mversic, line 10: this means that workflow is only run when some of these files are modified? You think we don't need this?

.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -17,6 +12,24 @@ env:
   RUSTUP_TOOLCHAIN: nightly-2023-06-25
 
 jobs:
+  check_format_smart_contracts:
+    runs-on: [self-hosted, Linux, iroha2ci]
```
mversic, line 16:
```suggestion
  smart_contracts_analysis:
    runs-on: ubuntu-latest
```

.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -17,6 +12,24 @@ env:
...
+        run: |
+          cd ./default_executor
+          mold --run cargo fmt -- --check
+
+      - name: Format check Client
+        run: |
+          cd ./client/tests/integration/smartcontracts
+          mold --run cargo fmt -- --check
+
   analysis:
```
mversic, line 33:
```suggestion
  workspace_analysis:
```

.github/workflows/iroha2-dev-pr-static.yml
```diff
@@ -17,6 +12,24 @@ env:
...
+
+      - name: Format check Default Executor
+        run: |
+          cd ./default_executor
+          mold --run cargo fmt -- --check
+
+      - name: Format check Client
+        run: |
+          cd ./client/tests/integration/smartcontracts
+          mold --run cargo fmt -- --check
```
mversic, line 31:
```suggestion
          mold --run cargo fmt --all -- --check
```

.github/workflows/iroha2-dev-pr.yml
mversic: I'm not sure, I wouldn't change it in this PR

.github/workflows/iroha2-dev-pr.yml
mversic, line 10: please restore single quotes

cli/build.rs
```diff
@@ -10,13 +10,6 @@ fn main() -> Result<()> {
 
     extract_git_hash()?;
 
-    // HACK: used by Nix, since at the moment
```
mversic, line 13: remove `IROHA_SKIP_WASM_CHECKS` from Nix

BAStos525 (approved):

mversic (approved):

cli/build.rs
DCNick3, line 1: `println!("cargo:rerun-if-changed={DEFAULT_EXECUTOR_PATH}");` and `const DEFAULT_EXECUTOR_PATH: &str = "../default_executor";` are not needed anymore I think (can't comment there)

flake.nix
```diff
@@ -155,7 +155,6 @@
           VERGEN_GIT_SHA = self.rev or "?dirty tree?";
 
           # Temporary workaround
```
DCNick3, line 157: Stray comment, should be removed as well

cli/build.rs
Asem-Abdelhady: i believe both print statements are not needed anymore

DCNick3: Hm, I think you are right. Not sure if `rerun-if-changed=build.rs` would be needed ever

mversic (approved):

PR #4098, opened 2023-11-29
[fix] #4097: Fix warp noise in logs

### LINKED ISSUE #4097: [BUG]: Fix warp noise in logs
Every request to iroha is polluted with the following log entry.

```
warp::filters::trace: unable to serve request (client error) status=404 error=Some(Rejection(NotFound))
```

Erigara:
## Description

- `warp::trace::request()` was attached twice for some endpoints, which lead to noise in logs
- logging error inside queue push was redundant because error should be handled on caller site

### Linked issue

Closes #4097 

### Benefits

Less noise in logs.

### DIFF
--- a/cli/src/torii/routing.rs
+++ b/cli/src/torii/routing.rs
@@ -528,9 +528,7 @@ impl Torii {
                 })
             });
 
-        let ws_router = events_ws_router
-            .or(blocks_ws_router)
-            .with(warp::trace::request());
+        let ws_router = events_ws_router.or(blocks_ws_router);
 
         warp::any()
             .and(
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -8,7 +8,7 @@ use eyre::{Report, Result};
 use iroha_config::queue::Configuration;
 use iroha_crypto::HashOf;
 use iroha_data_model::{account::AccountId, transaction::prelude::*};
-use iroha_logger::{debug, info, trace, warn};
+use iroha_logger::{debug, trace, warn};
 use iroha_primitives::must_use::MustUse;
 use rand::seq::IteratorRandom;
 use thiserror::Error;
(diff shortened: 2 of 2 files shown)


core/src/queue.rs
```diff
@@ -199,8 +198,13 @@ impl Queue {
         let entry = match self.txs.entry(hash) {
             Entry::Occupied(mut old_tx) => {
                 // MST case
+                let signatures_amount_before = old_tx.get().signatures().len();
                 assert!(old_tx.get_mut().merge_signatures(tx));
-                info!("Signature added to existing multisignature transaction");
```
Erigara, line 203: Most of the time this would entry produce logs for just transaction duplicate, received through gossiping, so i decided to only log if signatures amount increased.

core/src/queue.rs
```diff
@@ -189,7 +189,6 @@ impl Queue {
     pub fn push(&self, tx: AcceptedTransaction, wsv: &WorldStateView) -> Result<(), Failure> {
         trace!(?tx, "Pushing to the queue");
         if let Err(err) = self.check_tx(&tx, wsv) {
-            warn!("Failed to evaluate signature check. Error = {}", err);
```
Erigara, line 192: Should be handled on caller site.

cli/src/torii/routing.rs
```diff
@@ -528,9 +528,7 @@ impl Torii {
                 })
             });
 
-        let ws_router = events_ws_router
-            .or(blocks_ws_router)
-            .with(warp::trace::request());
+        let ws_router = events_ws_router.or(blocks_ws_router);
```
DCNick3, line 531: I don't see those handlers having a `warp::trace::request()` middleware attached to them. Are they logging with some other mechanism?

Erigara: They are handled together with other endpoints at line 544.

mversic (approved):

cli/src/torii/routing.rs
DCNick3: Oooh, I see, it was after that.

DCNick3 (approved):

Issue #4079, opened 2023-11-24
[BUG] 


0x009922:
## Description

When updating configuration with the `POST /configuration` endpoint, changes are applied (i.e. the log level is updated), but the updates are not reflected on the subsequent `GET /configuration` requests.

### Version

`iroha2-stable` (`rc.20`) and `iroha2-dev` (https://github.com/hyperledger/iroha/blob/b22ebc474204a91c6ec4bf422397f12df58e7946).

### Steps to Reproduce

1. Run Iroha
2. Make Request 1, observe the value of `LOGGER.MAX_LOG_LEVEL` (e.g. `INFO`)
3. Make Request 2 with a different log level value, e.g. `TRACE`
4. Observe the logs of Iroha, now you can see `TRACE` level
5. Make Request 1 again, and observe `LOGGER.MAX_LOG_LEVEL` in the returned data
    - **Expected:** `TRACE`
    - **Actual:** still `INFO`

**Request 1:**

```http
GET http://localhost:8080/configuration  
Accept: application/json  
Content-Type: application/json  
  
"Value"
```

**Request 2:**

```http
POST http://localhost:8080/configuration  
Content-Type: application/json  
  
{"LogLevel": "TRACE"}
```

### Research

Each time `GET /configuration` is invoked, the whole `Configuration` value is copied into it from `Torii` instance (by `add_state!` macro):

https://github.com/hyperledger/iroha/blob/b22ebc474204a91c6ec4bf422397f12df58e7946/cli/src/torii/routing.rs#L440-L445

`logger.max_log_level` is a type `SyncLevel`, which internally has an `AtomicCell<Level>` with logic that allows to react on the value updates. So, when the log level is updated:

```rust
cfg.logger.max_log_level.reload(new_level)
```

It actually changes the value inside of the `cfg` and actually alters the log level in `tracing`. However, this change is not reflected on other clones of the `cfg`, therefore each time `Torii` clones a configuration to handle an endpoint, it works with the initial value of `logger.max_log_level`.

I guess here is some logic bug: `AtomicCell` provides interior mutability, but doesn't provide a shared state across clones.

Issue #4064, opened 2023-11-21
[suggestion] Simplify logger's behaviour and configuration


0x009922:
## Description

I think the current logger's behaviour & configuration are a little overcomplicated, and I see a way to make its UX more straightforward and useful.

### Current implementation

Iroha prints logs into standard output in default way and optionally into a file in JSON format.

There are these options:

- `level`: logging verbosity
- `compact_mode`: flag to make logs in the standard output less verbose in terms of format
- `terminal_colors`: flag to control ANSI colouring in the standard output
- `log_file_path`: enables JSON logging into a file, **in a addition to standard output**

`log_file_path` seems not useful for me here. It is the only way to get JSON logs, and if you need JSON logs, than you most probably don't need any other logs. However, stdout is always busy with default logging.

Apart from that, I've created a separate issue for removal of `terminal_colors` from the logger configuration:

- https://github.com/hyperledger/iroha/issues/4030

### Proposal

- Remove `terminal_colors` (https://github.com/hyperledger/iroha/issues/4030), `log_file_path` and `compact_mode`
- Use stdout and a single output destination. It is easy to redirect it wherever, e.g. into a file. 
- Introduce `format` option to change the format of logger output. I think we can just adopt the [formats supported by `tracing_subscriber` out of the box](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/index.html):
    - `full`: default format now
    - `compact`: what is now enabled with `compact_mode`
    - `pretty`: I would personally very appreciate having this option, it might be very helpful  during debugging of Iroha and development of SDK
    - `json`: JSON machine-readable output

So, after all there will be only two options to configure: `level` and `format`.

0x009922:
Closed by #4100

PR #3346, opened 2023-03-31
[refactor] #3362: Migrate to tokio actors

Erigara:
## Description

- Remove broker 
- Switch from using `iroha_actor` to tokio actors

### Linked issue

Closes #3273, #3362.

### Benefits

- Actor is encapsulated insise it's `run` function so it's possible to modify it's state without enum.
- Flexibility in terms how actors communicate with each other (mpsc channels, oneshot channels, ...) this way it's possible to omit communication deadlock.
- Possible to use multiple channels and set priorities on them.
 
### Disadvantages

Solution feels a bit more verbose, but it ~ the same size in LoC.

Direct usage of `select!` can be tricky.

### Alternative design

We can use something like `actix` as actor framework, but it has almost the same api as `iroha_actor` so only benefits will be that it is more battle tested solution.

### Additional comments

I measured tps-benchmark before and after change and there is slight performance improvement.
However i should say that variance of tps benchmark is quite high so results could be different on another run.

```text
9dc702f37 (iroha2-dev before changes):
    time:   [1528.0 tps 1582.8 tps 1636.5 tps]
7d7816af3 (fix_peer_freeze after changes):
    time:   [1660.0 tps 1706.7 tps 1762.5 tps]
    change: [+3.4816% +7.8319% +13.344%] (p = 0.00 < 0.05)
```

### Checklist

- [x] I've read `CONTRIBUTING.md`
- [x] I've used the standard signed-off commit format (or will squash just before merging)
- [x] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

PR #4100, opened 2023-11-30
[fix] #4030, #4063, #4064, #4079: re-architect logger and dynamic configuration

### LINKED ISSUE #4030: [suggestion] Move `logger.terminal_colors` setting to the CLI layer
## Description

I propose to move `logger.terminal_colors` out of the configuration file to the layer of application's CLI args parsing, which happens before the config file is read and parsed.

### Why

"Terminal colors" configure whether the overall Command Line Interface should be colorized or not. Therefore, it belongs to the layer of CLI args, not the config file. Moreover, I don't see a reason why logger output colors in terminal should differ from any other terminal output of Iroha.

Also, configuration parsing happens after CLI arguments parsing. The process of configuration parsing itself might involve logging of hints/warnings/errors with or without colors. If `terminal_colors` is set in the config file, it will be only available _after_ the config is parsed. Thus, this option should be known before configuration parsing stage.

### Changes

- Remove `logger.terminal_colors` parameter from config file
- Introduce `--terminal-colors` CLI argument with a corresponding ENV var (e.g. `TERMINAL_COLORS`)

### Also

- Currently terminal colors is automatically enabled/disabled accordingly to TTY color detection (if `logger.terminal_colors` is not set). This behaviour will stay the same.

0x009922:
## Description

Started by updating configuration endpoints according to #4063, this PR expanded to major update of dynamic configuration mechanism, Logger configuration (#4030, #4064) and initialisation.

Logger initialisation is a global one-time thing, which was done with `iroha_logger::init(Configuration)`  function, which might accept the config and initialise logger, or might not, without an error. While initialising, it injected a `ReloadHandle` into the configuration instance, so that all updates to `config.logger.max_log_level` would trigger updates set up within `iroha_logger::init`. Than that `config` was copied many times in `Torii` and log level was actually reloaded on `POST /configuration` updates.

However, the configuration value update was not reflected in subsequent `GET /configuration` requests (#4079), because each time `config` is copied (it was copied for each request), it was editing a copy of `config` without affecting an original one. Since a copy of `config` was still containing `ReloadHandle` from Logger, the log level was updated.

Another problem appeared because of combination of configuration & logger architecture, in tests. It appeared that some tests ran concurrently, each calling `iroha_logger::init` with  independent `Configuration`s. However, only one `iroha_logger::init` might succeed, and another silently pretends that everything is ok. It worked fine, until one of Iroha instances fails to update the log level, because its `Configuration` passed into `iroha_logger::init` was not injected with a `ReloadHandle`.

Shortly, the system was *unsound*, i.e. it was easy to misuse. 

**Changes in this PR:**

- Remove `ReloadHandle` and `SyncValue` from configuration
- Introduce Kiso (jp 基礎 - foundation; basis): an actor responsible for holding the configuration state and broadcasting its updates across the system in a fire-and-forget way
- Update Logger configuration according to #4030 and #4064
- Introduce Logger actor, which might be initialised only once per process, and which is responsible for log level reloading and telemetry subscriptions
- Update Logger initialisation: split `init(Configuration) -> Result<Option<Telemetries>>` into
    - `init_global(Configuration) -> Result<LoggerHandle>`, initialising the Logger actor once per process, and failing if it was done already
    - `test_logger() -> LoggerHandle`, lazily initialising the Logger actor with pre-defined configuration, suitable for tests

**Chores:**

- Remove redundant fields from Kura configuration
- Refactor telemetry broadcasting mechanisms

### Linked issue

Closes #4030 #4063 #4064 #4079

### Benefits

- Improves configuration UX
- Improves system soundness

### Checklist

- [x] Refactor Kiso in a fire-and-forget manner
- [x] Play with logger manually
  - [x] Check if `bunyan` can parse current JSON output
- [x] Update logger config in py test env
- [x] Update README logger documentation
- [x] Use `202 Accepted` for `POST /configuration`

#### Update Logger configuration

In the configuration file, under `LOGGER.*` namespace:

- Rename `MAX_LOG_LEVEL` to `LEVEL`
- Remove `COMPACT_MODE`; use `FORMAT: "full"` (if `false`) and `FORMAT: "compact"` (if `true`) instead
- Remove `LOG_FILE_PATH`; use `FORMAT: "json"` instead and redirect stdout to a file, e.g. `iroha > log.json`
- Remove `TERMINAL_COLORS`; use `--terminal-colors`/`--no-terminal-colors` CLI arguments instead
- Remove `TELEMETRY_CAPACITY`

Env variables:

- `LOG_LEVEL`: instead of `MAX_LOG_LEVEL`
- `LOG_FORMAT`: new parameter
- `LOG_TOKIO_CONSOLE_ADDR`: instead of `TOKIO_CONSOLE_ADDR`

Example file before:

```json
{
  "LOGGER": {
    "MAX_LOG_LEVEL": "INFO",
    "TELEMETRY_CAPACITY": 1000,
    "COMPACT_MODE": false,
    "LOG_FILE_PATH": null,
    "TERMINAL_COLORS": true
  }
}
```

Example file after:

```json
{
  "LOGGER": {
    "LEVEL": "INFO",
    "FORMAT": "full"
  }
}
```

#### Update Kura configuration

In the configuration file, under `KURA.*` namespace, remove `ACTOR_CHANNEL_CAPACITY` and `BLOCKS_PER_STORAGE_FILE` fields.

#### Update configuration endpoints usage

- `GET /configuration` now returns JSON with only `{ logger: { level } }`.
- `POST /configuration` now accepts JSON in the same shape

`POST /configuration` request before:

```json
{"LogLevel":"DEBUG"}
```

`POST /configuration` request after:

```json
{"logger":{"level":"DEBUG"}}
```


.gitignore
```diff
@@ -49,3 +49,6 @@ result
 /test/
 /iroha-java/
 /lcov.info
+
+/rust-toolchain.toml
+.cargo/config.toml
```
Erigara, line 54: Do we really want to ignore this files?

cli/src/main.rs
Erigara, line 1: At some point we might consider to rewrite this using `clap`.

core/src/kiso.rs
```diff
@@ -0,0 +1,143 @@
...
+        tokio::spawn(async move { actor.run().await });
+
+        Self { sender }
+    }
+
+    /// Fetch the [`ConfigurationDTO`] from the actor's state.
+    ///
+    /// # Errors
+    /// If communication with actor fails.
+    pub async fn get_dto(&self) -> Result<ConfigurationDTO, Error> {
```
Erigara, line 52: Why do you choose message sending to propagate configuration? 

Some time ago i discovered [`arc_swap` ](https://github.com/vorner/arc-swap) create, from it's [usage patterns](https://docs.rs/arc-swap/latest/arc_swap/docs/patterns/index.html) i can see some benefits:

- it would allow to decouple configuration actor and configuration, so every configuration user will just have shared configuration type (or wrapper around it), not actor handle
- reduced amount of message sending to load configuration

scripts/test_env.py
```diff
@@ -118,18 +118,19 @@ def run(self, is_genesis: bool = False):
 
         os.environ["KURA_BLOCK_STORE_PATH"] = str(peer_dir.joinpath("storage"))
         os.environ["SNAPSHOT_DIR_PATH"] = str(peer_dir.joinpath("storage"))
-        os.environ["LOG_FILE_PATH"] = str(peer_dir.joinpath("log.json"))
-        os.environ["MAX_LOG_LEVEL"] = "TRACE"
+        os.environ["LOG_LEVEL"] = "TRACE"
+        # FIXME: should it be JSON here?
```
Erigara, line 122: Not sure on this one, for quick debugging at least for me it's more convenient to read plain text.

.gitignore
DCNick3: I don't know _why_ we have been ignoring those files, but I've accomplished this by listing them `.git/info/exclude`, which works the same way as gitignore, but isn't committed.

I think it does make sense to commit them to ensure everybody uses the same toolchain we use in CI and to not make people enable lints feature by hand.

I also have a snippet that  enables the usage of `mold` there:

```toml
[target.x86_64-unknown-linux-gnu]
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=/usr/bin/mold"]
```

README.md
```diff
@@ -178,17 +178,13 @@ For example, if your Iroha instance is running at `127.0.0.1:8080` and you want
 curl -X POST \
     -H 'content-type: application/json' \
     http://127.0.0.1:8080/configuration \
-    -d '{"LogLevel": "DEBUG"}' -i
+    -d '{"logger": {"level": "DEBUG"}}' -i
 ```
 </details>
 
-#### JSON Logging Mode
+Logger supports multiple output formats: `full` (default), `compact`, `pretty`, and `json`.
```
DCNick3, line 185: I think this should mention the name of the setting changing the output format

cli/src/main.rs
DCNick3: Created https://github.com/hyperledger/iroha/issues/4105

.gitignore
Erigara: > but I've accomplished this by listing them .git/info/exclude

Oh, thanks for the tip

cli/src/main.rs
```diff
@@ -109,7 +129,7 @@ async fn main() -> Result<(), color_eyre::Report> {
     }
 
     let config = iroha::combine_configs(&args)?;
-    let telemetry = iroha_logger::init(&config.logger)?;
+    let telemetry = iroha_logger::init_global(&config.logger, terminal_colors)?;
```
DCNick3, line 132: This is no longer telemetry, but a logger handle, probably needs renaming

cli/src/main.rs
0x009922: I was going to rewrite the main CLI as a next step, anyway it is related to the config RFC.

core/src/kiso.rs
0x009922: `arc_swap` seems to be a good option in case of sharing mutable state.

Some notes:

- Actually, now the config itself is almost not shared. The initial values are read and directly stored in actors, and only `Torii` holds the whole structure and interacts with it.
- We have a stateful system with many actors, which is not the same as patterns described in `arc_swap`. If in future we add more dynamic configurations related to internal logic of actors, it wouldn't be enough to just share the state: we will need a notification mechanism so that actors might react to changes. `Kiso` might propagate updates naturally.
- With `arc_swap`, each holder of the config will be able to not only read, but also write changes. It makes mutations decentralised and the architecture might become confusing to maintain and change, some subtle semantic bugs might appear etc. A central responsible point makes things clear, and it seems for me not achievable with `arc_swap`'s design.
- Shared mutable state is not idiomatic in the actor model.
- If in future we will need to share big parts of the config frequently, it is still achievable with `Kiso` and careful design of caching and notifications.
- Currently all dynamic configuration interactions happen only during configuration endpoints handling, which I guess is not that part that will be invoked frequently. So, I don't think message passing overhead is significant. Anyway, caching might always be used on top of that in future.

scripts/test_env.py
0x009922: I think it's better to use `pretty` in this case.

.gitignore
0x009922: Ah, just forgot to remove them from `staged`
Will move to `.git/info/exclude`, although I would like to see them in the repo permanently anyway.

logger/src/telemetry.rs
```diff
@@ -82,70 +82,74 @@ impl Telemetry {
...
+/// A pair of [`Channel`] associated with [`Event`]
+pub struct ChannelEvent(pub Channel, pub Event);
+
+/// Supported telemetry channels
+#[derive(Copy, Clone)]
+pub enum Channel {
+    /// Regular telemetry
+    Regular,
+    /// Future telemetry (?)
+    // FIXME: what is the "future"?
```
DCNick3, line 153: I think the "future" telemetry is collected from futures instrumented with `iroha_futures::TelemetryFuture`. It logs on every poll and reports how long did it take.

I am not sure if this feature is used by anyone though

core/src/kura.rs
```diff
@@ -51,21 +51,23 @@ impl Kura {
     /// to access the block store indicated by the provided
     /// path.
     pub fn new(
-        mode: Mode,
-        block_store_path: &Path,
-        debug_output_new_blocks: bool,
+        config: &Configuration,
+        // mode: Mode,
```
DCNick3, line 55: These commented-out arguments can probably be removed

config/src/kura.rs
```diff
@@ -17,11 +15,11 @@ pub struct Configuration {
     /// Initialization mode: `strict` or `fast`.
     pub init_mode: Mode,
     /// Path to the existing block store folder or path to create new folder.
-    pub block_store_path: String,
-    /// Maximum number of blocks to write into a single storage file.
-    pub blocks_per_storage_file: NonZeroU64,
-    /// Default buffer capacity of actor's MPSC channel.
-    pub actor_channel_capacity: u32,
+    pub block_store_path: PathBuf,
+    // /// Maximum number of blocks to write into a single storage file.
```
DCNick3, line 19: Commented code

as an aside note: curious how we ended up with these unused configuration options

core/src/kiso.rs
Erigara: Thanks for detailed explanation, your arguments seem reasonable to me

client/benches/tps/oneshot.rs
```diff
@@ -20,7 +20,7 @@ fn main() {
         flush_guard = Some(flame_layer.flush_on_drop());
 
         tracing_subscriber::registry().with(flame_layer).init();
-        iroha_logger::disable_logger();
+        iroha_logger::disable_global().expect("Logger wasn't set yet");
```
DCNick3, line 23: I think `disable_global` fails in the opposite way - if the logger was already set

client/tests/integration/config.rs
```diff
@@ -1,27 +1,44 @@
...
+
+    // Retrieving through API
+    let mut dto = test_client.get_config().expect("Client can always get it");
+    assert_eq!(dto.logger.level, init_log_level);
+
+    // Updating the log level
+    dto.logger.level = NEW_LOG_LEVEL;
+    test_client.set_config(dto).expect("New config is valid");
+
+    // FIXME: The updated value is not reflected
```
DCNick3, line 34: Is this still true? The PR title suggest that it addresses #4079

config/src/logger.rs
```diff
@@ -29,77 +22,49 @@ pub fn into_tracing_level(level: Level) -> tracing::Level {
...
-    }
-}
-
-impl Eq for SyncLevel {}
-
 /// 'Logger' configuration.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Proxy, Documented)]
 #[serde(rename_all = "UPPERCASE")]
+#[config(env_prefix = "LOG_")]
+// `tokio_console_addr` is not `Copy`, but warning appears without `tokio-console` feature
```
DCNick3, line 29: Now that it has a type of `SocketAddr`, it _is_ actually `Copy`, so it can be derived here

core/src/kiso.rs
```diff
@@ -0,0 +1,143 @@
...
+        dto: ConfigurationDTO,
+        respond_to: oneshot::Sender<Result<(), Error>>,
+    },
+}
+
+/// TODO
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// TODO
+    #[error("cannot dynamically update the configuration")]
```
DCNick3, line 92:
```suggestion
    #[error("failed to dynamically update the configuration")]
```

core/src/kiso.rs
```diff
@@ -0,0 +1,143 @@
...
+    Update(#[from] iroha_logger::ReloadError),
+    /// TODO
+    #[error("cannot get actor's response")]
```
DCNick3, line 95:
```suggestion
    #[error("failed to get actor's response")]
```

futures/tests/basic.rs
```diff
@@ -32,15 +31,13 @@ async fn test_sleep() {
...
     ];
 
-    let (_, telemetry_future) = iroha_logger::init(
-        &ConfigurationProxy::default()
-            .build()
-            .expect("Default logger config always builds"),
-    )
-    .unwrap()
-    .unwrap();
+    let telemetry_future = iroha_logger::test_logger()
```
DCNick3, line 34:
```suggestion
    let future_telemetry = iroha_logger::test_logger()
```

core/src/kiso.rs
```diff
@@ -0,0 +1,143 @@
...
+            LoggerError::LevelReload(err) => Self::from(err),
+            LoggerError::Communication(err) => Self::from(err),
+        }
+    }
+}
+
+struct Actor {
+    receiver: mpsc::Receiver<Message>,
+    state: Configuration,
+    logger: LoggerHandle,
```
DCNick3, line 111: Just an idea: if we would add dynamic configuration for other more general-purpuse actors, we _might_ want to use a separate "handle" type not to conflate the cfg-related commands/messages with the actual actor functionality

logger/tests/setting_logger.rs
```diff
@@ -1,21 +1,17 @@
...
-    assert!(second_init.unwrap().is_none());
+    let cfg = ConfigurationProxy::default()
+        .build()
+        .expect("Default logger config always builds");
+
+    let first = init_global(&cfg, false);
+    assert!(first.is_ok());
+
+    let second = init_global(&cfg, false);
+    assert!(second.is_ok());
```
DCNick3, line 14: I think this shouldn't succeed, as the logger was already initialized?

scripts/test_env.py
DCNick3: I _think_ with bunyan format we had before this PR it was possible to use a [cli tool](https://www.npmjs.com/package/bunyan) to print it in a pretty form with colors, which is nice. I learned of it just now, after researching what is this format :sweat_smile:

cli/src/lib.rs
```diff
@@ -389,35 +413,34 @@ impl Iroha {
...
     #[cfg(feature = "telemetry")]
     async fn start_telemetry(
-        telemetry: Option<(
-            iroha_logger::SubstrateTelemetry,
-            iroha_logger::FutureTelemetry,
-        )>,
-        config: &Configuration,
+        logger: &LoggerHandle,
+        config: &TelemetryConfiguration,
     ) -> Result<bool> {
```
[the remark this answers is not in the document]
0x009922: Thanks! Actually, I don't really like `Result<bool>` here...

client/benches/tps/oneshot.rs
0x009922: It's true. I put this message with "What assumption I _expect_ to be true?" in mind.

This would be clearer:

```rust
iroha_logger::disable_global().expect("Logger should not be set yet");
```

config/src/kura.rs
0x009922: Don't know the history of `blocks_per_storage_file`. For `actor_channel_capacity` you can look here: https://github.com/hyperledger/iroha-2-docs/issues/392#issuecomment-1722872161

client/benches/tps/oneshot.rs
DCNick3: Hm, I usually think of `expect` not as "this is why it should never fire", but in terms of "what the user sees when this gets fired", hence the confusion. Your proposed message serves both goals fine I think

core/src/kiso.rs
0x009922: Hmm, I can't get wdym. Could you give an example?

core/src/kiso.rs
```diff
@@ -0,0 +1,143 @@
...
+                        logger: LoggerDTO { level: new_level },
+                    },
+                respond_to,
+            } => {
+                if let Err(err) = self.logger.reload_level(new_level).await {
+                    let _ = respond_to.send(Err(err.into()));
+                    return;
+                }
+                self.state.logger.level = new_level;
+                let _ = respond_to.send(Ok(()));
```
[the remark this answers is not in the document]
0x009922: So, you mean to use fire-and-forget configuration updates propagation.

Currently I designed `Kiso` in kind of transactional (but not atomic) way: it first tries to apply all updates, and only then updates the actual state. If something goes wrong during the update, it will return an error. This way, if you `POST /configuration`, `200 OK` will mean that the changes are applied, and if not, an error will return.

With fire-and-forget, `POST /configuration` will not return an error, but will always return `202 ACCEPTED`, and `GET /configuration` will return updated config which might not be really applied in the system.

I am fine either way and not sure which option is better. IMO the current design makes Iroha more reliable.

logger/src/lib.rs
```diff
@@ -1,159 +1,139 @@
...
 
-    if configuration.compact_mode {
-        add_bunyan(configuration, layer.compact())
-    } else {
-        add_bunyan(configuration, layer)
+    match configuration.format {
+        Format::Full => step2(configuration, layer),
+        Format::Compact => step2(configuration, layer.compact()),
+        Format::Pretty => step2(configuration, layer.pretty()),
+        Format::Json => step2(configuration, layer.json()),
```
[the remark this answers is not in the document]
0x009922: I _guess_ `bunyan` might read `tracing`'s JSON logs without issues (I will check). `tracing` docs [says](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Json.html) it might be used by some `jq` or a JSON log viewer.

Anyway, we should communicate this change to users, for sure. They will need to change logger configuration anyway.

logger/tests/setting_logger.rs
0x009922: Indeed, will fix

logger/src/lib.rs
```diff
@@ -1,159 +1,139 @@
...
+/// # Errors
+/// If global logger was already initialised/disabled.
+pub fn disable_global() -> Result<()> {
+    try_set_logger()
 }
 
-fn add_bunyan<L>(configuration: &Configuration, layer: L) -> Result<Telemetries>
+fn step2<L>(configuration: &Configuration, layer: L) -> Result<LoggerHandle>
 where
     L: tracing_subscriber::Layer<Registry> + Debug + Send + Sync + 'static,
```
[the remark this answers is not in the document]
DCNick3: Well, the `L` type is different for each match arm calling it. It would either need dynamic dispatch (is `Layer` object safe?) or a static dispatch via an enum, both of which would bring more problems than a separated generic function for `step2` imo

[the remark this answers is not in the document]
0x009922: I see, thanks.

core/src/kiso.rs
0x009922: Hmm, actually now I see a way how to both use `watch` to propagate changes, but also update the actual config value in Kiso only when the change is applied. 

```mermaid
classDiagram
  class Kiso {
    Config config
    watch::Sender~LogLevelUpdate~ log_level_broadcast
    log_level_watcher() watch::Receiver~LogLevelUpdate~
  }
  class LogLevelUpdate {
    Level value
    oneshot::Sender~none~ respond_to
  }
  class Logger {
    watch_log_level(watch::Receiver~LogLevelUpdate~)
  }

  Logger <-- Kiso: sets log level
  Logger --> Kiso: notifies that log level is updated
```

```rust
let kiso: KisoHandle = _;
let logger: LoggerHandle = _;

logger.watch_log_level(kiso.log_level_watcher())
```

_Not sure how much this diagram helps_



My idea is that receiving part of `watch` channels might receive messages with `oneshot` channels to which they might reply when the change is actually applied. Therefore, when `Kiso` receives that update, it might update its `config` value, so that `GET /configuration` requests before actual value application will not return inaccurate information.

Anyway, this seem to make the whole system a little more complicated...

DCNick3: Yeah, if we don't want it to be "fire and forget", it would probably should work like this. Though, in this case, I think, watch channels don't bring much to the table and can be just a separate `mpsc`.

Fire-and-forget or not, what I want is separation of for the actors: I think configuration updates should not go into the "main" channel, as it would require adding additional enum variants to each actor's messages and instead be handled via a separate and reusable channel type.

Though this change only makes sense if we are planning to add more updatable configuration options, otherwise it is indeed an unnecessary complication.

0x009922: Indeed, sounds convincing. Since implementing atomicity and transaction semantics would be way too complex for a not-so-important subsystem of Iroha, fire-and-forget would be a good fit.

Let me refactor.

cli/src/torii/routing.rs
```diff
@@ -478,7 +451,7 @@ impl Torii {
                     warp::path(uri::TRANSACTION)
                         .and(add_state!(self.queue, self.sumeragi))
                         .and(warp::body::content_length_limit(
-                            self.iroha_cfg.torii.max_content_len.into(),
+                            self.instructions_max_content_length,
```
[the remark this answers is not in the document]
0x009922: Agreed. I was confused because the function itself is still `handle_instructions`. Will rename both.

config/src/logger.rs
0x009922: It is actually not:

```
error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> config/src/logger.rs:26:17
   |
26 | #[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, Proxy, Documented)]
   |                 ^^^^
...
37 |     pub tokio_console_addr: SocketAddr,
   |     ---------------------------------- this field does not implement `std::marker::Copy`
   |
```

And it cannot be `Copy` because `SocketAddr::Host` -> `SocketAddrHost` contains `BoxedString` within.

cli/src/main.rs
0x009922: Fixed

cli/src/lib.rs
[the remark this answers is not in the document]
0x009922: Refactored.

logger/src/telemetry.rs
0x009922: Updated the docstring

core/src/kura.rs
0x009922: Removed

config/src/kura.rs
0x009922: Removed dead code

cli/src/torii/routing.rs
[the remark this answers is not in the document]
0x009922: Renamed both

client/benches/tps/oneshot.rs
0x009922: Updated the message.

logger/tests/setting_logger.rs
0x009922: Fixed.

.gitignore
0x009922: Restored `.gitignore`.

core/src/kiso.rs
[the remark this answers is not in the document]
Erigara: Looks like this error pop up when either logger layer was either poisoned or dropped, not smt we can handle meaningfully.

config/src/logger.rs
DCNick3: Oh, I confused it with std's `SocketAddr`

core/src/kiso.rs
0x009922: Refactored.

My main goal was to loose modules coupling. Now Iroha spawns another task on startup which is responsible for updates broadcasting. It listens for `watch` updates from Kiso and forwards them to Logger, without them knowing about each other.

This seems to be the most flexible way in terms of system components independence. However, Kiso has some boilerplate which I think doesn't make sense to refactor now.

docs/source/references/config.md
```diff
@@ -49,7 +47,8 @@ The following is the default configuration used by Iroha.
     "MAX_TRANSACTIONS_IN_BLOCK": 512,
     "ACTOR_CHANNEL_CAPACITY": 100,
     "GOSSIP_BATCH_SIZE": 500,
-    "GOSSIP_PERIOD_MS": 1000
+    "GOSSIP_PERIOD_MS": 1000,
+    "DEBUG_FORCE_SOFT_FORK": false
```
Erigara: Looks like you generated `config.md` with `debug` profile. Can you please recreate it with `release`, because we don't expose `"DEBUG_FORCE_SOFT_FORK"` parameter.

0x009922: Yup, I fixed it already.

logger/src/lib.rs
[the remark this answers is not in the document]
0x009922: Nope, `bunyan` cannot read it: [it requires at least the `name` field to be set](https://github.com/trentm/node-bunyan#introduction), which `tracing` doesn't provide.

cli/src/torii/routing.rs
```diff
@@ -79,7 +74,7 @@ fn fetch_size() -> impl warp::Filter<Extract = (FetchSize,), Error = warp::Rejec
 }
 
 #[iroha_futures::telemetry_future]
-async fn handle_instructions(
+async fn handle_transaction(
```
Erigara, line 77: Good, should be renamed long ago

Erigara (approved):

README.md
0x009922: Done.

docs/source/references/config.md
```diff
@@ -230,7 +230,7 @@ Has type `Option<Box<kura::ConfigurationProxy>>`[^1]. Can be configured via envi
 
 Path to the existing block store folder or path to create new folder.
 
-Has type `Option<PathBuf>`[^1]. Can be configured via environment variable `KURA_BLOCK_STORE_PATH`
+Has type `Option<String>`[^1]. Can be configured via environment variable `KURA_BLOCK_STORE_PATH`
```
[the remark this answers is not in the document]
0x009922: I'd postpone this change until whole configuration refactoring, which should give flexibility in parsing and types. For now, refactoring of config is painful and sometimes has non-obvious bugs.

core/src/kiso.rs
```diff
@@ -83,37 +92,32 @@ enum Message {
...
-/// TODO
+/// Possible errors might occur while working with [`KisoHandle`]
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    /// TODO
-    #[error("failed to dynamically update the configuration")]
-    Update(#[from] iroha_logger::ReloadError),
-    /// TODO
+    #[allow(missing_docs)]
     #[error("failed to get actor's response")]
```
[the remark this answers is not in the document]
0x009922: Thanks for notice, will use it.

cli/src/lib.rs
```diff
@@ -480,6 +482,34 @@ impl Iroha {
...
+            loop {
+                tokio::select! {
+                    Ok(()) = log_level_update.changed() => {
+                        let value = *log_level_update.borrow_and_update();
+                        if let Err(error) = logger.reload_level(value).await {
+                            iroha_logger::error!("Failed to reload log level: {error}");
+                        };
+                    }
+                };
+                tokio::task::yield_now().await;
```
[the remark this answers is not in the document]
0x009922: I blindly followed the pattern in other `select!`s. It turned out that in other places loops need to yield because their futures are always awailable and it might block other actors from functioning. In this case, however, messages are expected to be rare, so I will remove it.

Original discussion: https://github.com/hyperledger/iroha/pull/3346#discussion_r1155532000

core/src/kiso.rs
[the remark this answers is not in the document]
0x009922: Done.

cli/src/lib.rs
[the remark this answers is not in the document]
0x009922: Removed.

core/src/kiso.rs
```diff
@@ -83,37 +92,32 @@ enum Message {
...
-        }
-    }
-}
-
 struct Actor {
-    receiver: mpsc::Receiver<Message>,
+    handle: mpsc::Receiver<Message>,
     state: Configuration,
-    logger: LoggerHandle,
+    /// Current implementation is somewhat not scalable in terms of code writing: for any
```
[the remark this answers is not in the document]
0x009922: Done.

Erigara (dismissed):

logger/src/telemetry.rs
```diff
@@ -82,70 +82,73 @@ impl Telemetry {
...
     }
 
     /// Create new telemetry layer with specific channel size (via const generic)
     #[allow(clippy::new_ret_no_self)]
     pub fn new<const CHANNEL_SIZE: usize>(
         subscriber: S,
-    ) -> (impl Subscriber, Receiver<Telemetry>, Receiver<Telemetry>) {
+    ) -> (impl Subscriber, mpsc::Receiver<ChannelEvent>) {
         let (sender, receiver) = mpsc::channel(CHANNEL_SIZE);
-        let (sender_future, receiver_future) = mpsc::channel(CHANNEL_SIZE);
```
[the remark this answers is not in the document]
0x009922: Refactored, left only `with_capacity` public method.

Erigara (approved):

DCNick3 (approved):

PR #4115, opened 2023-12-05
[refactor] #3657: Split view changes from normal messages

### LINKED ISSUE #3657: Could we remove view change proof from every message?
Since we now separate queue for service messages could we use remove view change proof from every message?

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/issues/3652#issuecomment-1609659814_

SamHSmith:
### Linked issue

Closes #3657

### Checklist

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -43,7 +43,7 @@ pub struct Sumeragi {
     /// Receiver channel, for control flow messages.
     pub control_message_receiver: mpsc::Receiver<ControlFlowMessage>,
     /// Receiver channel.
-    pub message_receiver: mpsc::Receiver<MessagePacket>,
+    pub message_receiver: mpsc::Receiver<BlockMessage>,
     /// Only used in testing. Causes the genesis peer to withhold blocks when it
     /// is the proxy tail.
     pub debug_force_soft_fork: bool,
@@ -83,12 +83,13 @@ impl Sumeragi {
     /// # Errors
     /// Fails if network sending fails
     #[instrument(skip(self, packet))]
-    fn post_packet_to(&self, packet: MessagePacket, peer: &PeerId) {
+    fn post_packet_to(&self, packet: BlockMessage, peer: &PeerId) {
         if peer == &self.peer_id {
             return;
         }
+
         let post = iroha_p2p::Post {
-            data: NetworkMessage::SumeragiPacket(Box::new(packet)),
+            data: NetworkMessage::SumeragiBlock(Box::new(packet)),
             peer_id: peer.clone(),
         };
         self.network.post(post);
@@ -97,7 +98,7 @@ impl Sumeragi {
     #[allow(clippy::needless_pass_by_value, single_use_lifetimes)] // TODO: uncomment when anonymous lifetimes are stable
     fn broadcast_packet_to<'peer_id>(
         &self,
-        msg: MessagePacket,
+        msg: BlockMessage,
         ids: impl IntoIterator<Item = &'peer_id PeerId> + Send,
     ) {
         for peer_id in ids {
@@ -105,9 +106,16 @@ impl Sumeragi {
         }
     }
 
-    fn broadcast_packet(&self, msg: MessagePacket) {
+    fn broadcast_packet(&self, msg: BlockMessage) {
         let broadcast = iroha_p2p::Broadcast {
-            data: NetworkMessage::SumeragiPacket(Box::new(msg)),
+            data: NetworkMessage::SumeragiBlock(Box::new(msg)),
+        };
+        self.network.broadcast(broadcast);
+    }
+
+    fn broadcast_control_flow_packet(&self, msg: ControlFlowMessage) {
+        let broadcast = iroha_p2p::Broadcast {
+            data: NetworkMessage::SumeragiControlFlow(Box::new(msg)),
         };
         self.network.broadcast(broadcast);
     }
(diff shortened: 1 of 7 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -85,6 +87,16 @@ impl Sumeragi {
         if peer == &self.peer_id {
             return;
         }
+        if matches!(packet, MessagePacket::NormalMessage(_)) {
```
Erigara, line 90: So now for every message we send additional message with view_change_proofs? Why do we need to do that?

core/src/sumeragi/mod.rs
```diff
@@ -202,24 +202,27 @@ impl SumeragiHandle {
...
-                     Incoming control messages have to be dropped due to low processing speed."
-                );
+        match msg {
+            MessagePacket::ControlFlowMessage(msg) => {
+                if let Err(error) = self.control_message_sender.try_send(msg) {
+                    self.metrics.dropped_messages.inc();
+                    error!(
+                        ?error,
+                        "This peer is faulty. \
+                         Incoming control messages have to be dropped due to low processing speed."
```
Erigara, line 212: Not for this PR, but what do you think about instead of dropping new messages we would overwrite old ones?

Erigara (commented):
What version of `rustfmt` do you use? Because i noticed some formatting changes not relevant to this PR.

core/src/sumeragi/main_loop.rs
```diff
@@ -41,12 +41,14 @@ pub struct Sumeragi {
...
     /// Receiver channel.
-    pub message_receiver: mpsc::Receiver<MessagePacket>,
+    pub message_receiver: mpsc::Receiver<Message>,
     /// Only used in testing. Causes the genesis peer to withhold blocks when it
     /// is the proxy tail.
     pub debug_force_soft_fork: bool,
     /// The current network topology.
     pub current_topology: Topology,
+    /// The current view change proofs.
+    pub view_change_proofs: view_change::ProofChain,
```
Arjentix, line 51: Oh, another field in this big struct...

core/src/sumeragi/main_loop.rs
Arjentix: Also this `if` introduces one more *exceptional* behaviour... Can it be refactored with some trait implementation? For example default implemetation for all and specific for the internal type of `NormalMessage` variant?

core/src/sumeragi/message.rs
```diff
@@ -9,22 +9,9 @@ use crate::block::{CommittedBlock, ValidBlock};
...
-    /// Construct [`Self`]
-    pub fn new(view_change_proofs: view_change::ProofChain, message: Option<Message>) -> Self {
-        Self {
-            view_change_proofs,
-            message,
-        }
-    }
+pub enum MessagePacket {
+    ControlFlowMessage(ControlFlowMessage),
+    NormalMessage(Message),
```
Arjentix, line 14:
```suggestion
pub enum Message {
    ControlFlow(ControlFlowMessage),
    Block(BlockMessage),
```

- I think `Packet` suffix doesn't make sense anymore
- `Normal` prefix doesn't give anything to us
- No need to repeat `Message` in variants if enum name already contains it

core/src/sumeragi/main_loop.rs
SamHSmith: has been removed

core/src/sumeragi/main_loop.rs
SamHSmith: has been removed

core/src/sumeragi/message.rs
SamHSmith: Good suggestion.

core/src/sumeragi/mod.rs
SamHSmith: I think that would be worse from a security perspective.

core/src/sumeragi/main_loop.rs
```diff
@@ -881,57 +859,63 @@ pub(crate) fn run(
...
+            handle_message(
+                message,
+                &mut sumeragi,
+                &mut voting_block,
+                current_view_change_index,
+                &mut voting_signatures,
+            );
+        }
+
+        // State could be changed after handling message so it is necessary to reset state before handling message independent step
```
Erigara, line 876: You only partially moved reset state part, did you do this on purpose?

mversic:
please change git commit message to `[refactor]`

core/src/sumeragi/mod.rs
```diff
@@ -201,25 +201,28 @@ impl SumeragiHandle {
...
+                    );
+                }
+            }
+            Message::Block(msg) => {
```
mversic, line 216:
```suggestion
            Message::Data(msg) => {
```

how about this? I would say we have data and control flow messages

core/src/sumeragi/main_loop.rs
SamHSmith: Yes.

core/src/sumeragi/main_loop.rs
```diff
@@ -367,26 +360,29 @@ impl Sumeragi {
     }
 }
 
-fn suggest_view_change(
+fn broadcast_view_change(
     sumeragi: &Sumeragi,
     view_change_proof_chain: &mut ProofChain,
     current_view_change_index: u64,
+    should_append: bool,
 ) {
```
mversic, line 368:
```suggestion
fn broadcast_view_change(
    sumeragi: &Sumeragi,
    view_change_proof_chain: &mut ProofChain,
) {
```

pull other stuff out of this function

core/src/sumeragi/mod.rs
```diff
@@ -204,25 +204,28 @@ impl SumeragiHandle {
...
-            if let Err(error) = self.control_message_sender.try_send(ControlFlowMessage {
-                view_change_proofs: msg.view_change_proofs,
-            }) {
-                self.metrics.dropped_messages.inc();
-                error!(
-                    ?error,
-                    "This peer is faulty. \
-                     Incoming control messages have to be dropped due to low processing speed."
-                );
+    pub fn incoming_message(&self, msg: Message) {
```
mversic, line 207: IMO we should split this into 2 functions and remove `Message` enum. There is no need to group messages in the enum anymore

core/src/sumeragi/main_loop.rs
```diff
@@ -138,48 +146,45 @@ impl Sumeragi {
...
-        control_message_in_a_row_counter: &mut usize,
-    ) -> Option<Message> {
+    ) -> (Option<BlockMessage>, bool) {
         const MAX_CONTROL_MSG_IN_A_ROW: usize = 25;
 
-        if *control_message_in_a_row_counter < MAX_CONTROL_MSG_IN_A_ROW {
-            *control_message_in_a_row_counter += 1;
-            self.control_message_receiver
+        let mut should_sleep = true;
+        for _ in 0..MAX_CONTROL_MSG_IN_A_ROW {
```
mversic, line 153: why do we want to limit the number of sequential control flow messages?

core/src/sumeragi/main_loop.rs
```diff
@@ -138,48 +146,45 @@ impl Sumeragi {
...
+            .map_err(|recv_error| {
+                assert!(
+                    recv_error != mpsc::TryRecvError::Disconnected,
+                    "Sumeragi message pump disconnected. This is not a recoverable error."
+                )
+            })
+            .ok();
+        should_sleep &= msg.is_none();
+        (msg, should_sleep)
     }
```
mversic, line 188: depending on the previously received control flow messages, is it possible that some of the `BlockMessage`s should just be rejected?

core/src/sumeragi/main_loop.rs
Erigara: I think to prevent data message starvation in case of high frequency of service messages

core/src/sumeragi/main_loop.rs
Erigara: Yes, e.g. if enough proof received view change should be updated making block with previous view change irrelevant.

mversic: so we should also listen for block messages here in a loop while invalid block messages are observed

core/src/sumeragi/main_loop.rs
mversic: I get that, but is that necessary?

Erigara: yes imo e.g. malicious peer can spam with control flow messages

Erigara (approved):

mversic (approved):

PR #4130, opened 2023-12-08
[fix] #0000: On-chain predictable iteration order

mversic:
## Description

* replace all `HashMap`/`HashSet` in `iroha_core` with `IndexMap`/`IndexSet`
* add `HashMap`/`HashSet` to `dissallowed_types` 

### Linked issue

Closes [#3](https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/zh/3)

### Checklist

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -68,7 +68,9 @@ displaydoc = { workspace = true }
 wasmtime = { workspace = true }
 parking_lot = { workspace = true, features = ["deadlock_detection"] }
 derive_more = { workspace = true }
+
 uuid = { version = "1.4.1", features = ["v4"] }
+indexmap = "2.1.0"
 
 [dev-dependencies]
 criterion = { workspace = true }
--- a/core/clippy.toml
+++ b/core/clippy.toml
@@ -1,0 +1,1 @@
+disallowed-types = ["std::collections::HashMap", "std::collections::HashSet"]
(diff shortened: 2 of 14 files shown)


core/clippy.toml
mversic, line 1: can we agree this is the safest thing to do

Arjentix: > can we agree this is the safest thing to do?

Yeah, looks like a good feature

> should I do this at workspace level?

I think yes and theh disable in client-facing crates

Arjentix: Because it might be useful for `smart_contract`, `primitives` and etc

Arjentix (approved):
In the issue you wrote that you'll use `BTree*` collections. Why did you change your mind?

core/clippy.toml
mversic: > I think yes and theh disable in client-facing crates

it doesn't work. If there are 2 overlapping configs workspace level wins. Check [here](https://github.com/rust-lang/rust-clippy/issues/7353)

mversic:
> In the issue you wrote that you'll use `BTree*` collections. Why did you change your mind?

it seemed as a more natural replacement. As long as the bug is fixed we can always switch to b-trees.

core/Cargo.toml
```diff
@@ -68,7 +68,9 @@ displaydoc = { workspace = true }
 wasmtime = { workspace = true }
 parking_lot = { workspace = true, features = ["deadlock_detection"] }
 derive_more = { workspace = true }
+
```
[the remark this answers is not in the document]
mversic: yeah

Erigara (approved):

Arjentix (approved):
So index at the end

Issue #4136, opened 2023-12-11
[suggestion] Refactor Iroha CLI


0x009922:
## Description

Apart from not using `clap`, the CLI (`iroha` binary) has other issues:

- It is simultaneously a binary and a library, which looks confusing IMO. It is this way so `test_network` crate can re-use `Iroha` root orchestrator in test env
- As a library, it exports `iroha::samples` with data samples for tests. Doesn't seem to relate to CLI at all.
- `Torii` is defined here; again, it is not really related to CLI and rather deserves to be moved somewhere.

To address this issues and make another steps towards the Configuration Overhaul RFC (https://github.com/hyperledger/iroha/issues/2585), I propose the following actions.

### Actions

- [x] Remove `Torii` from `iroha`; make it a separate `iroha_torii` crate. It is a self-contained module, focusing on Iroha HTTP API. While _core_ enough to belong to `iroha_core` (like Sumeragi or Kura), IMO it would better be a separate crate because its peripheral role, and it's a good candidate to be a standalone compilation unit.
  - #4139
- [x] Remove global panic hook setting from `Iroha` startup; set it up on a higher level (e.g. CLI)
  - Addressed by #4716 
- [ ] Remove `iroha::samples` from `iroha`; maybe, make a separate `test_samples` crate or something. 
- [ ] Consider moving `Iroha` structure & logic out of CLI, e.g. into `iroha_orchestrator` or something.
- [x] https://github.com/hyperledger/iroha/issues/4105 (+ bump `clap` version of `iroha_client_cli`)
- [x] Additionally, remove setting genesis file through CLI and introduce `genesis.file` configuration parameter as outlined in https://github.com/hyperledger/iroha/issues/4029

PR #4139, opened 2023-12-12
[refactor]: split `iroha_torii` from `iroha`

0x009922:
## Description

This PR splits Torii (web server functionality), located in the main CLI, into a separate `iroha_torii` crate.

The rationale is to make a step towards making the main CLI exclusively a binary. Currently `iroha` is a library as well, exposing functionality to run `Iroha` in tests. That includes Torii. Since Torii is quite self-contained, making it a separate crate seem to make sense.

A further step would be to move `Iroha` itself and `iroha::samples` out of the CLI, so that `test_network` doesn't depend on `iroha` anymore, and `iroha` might be exclusively a binary.

Additionally:

- fix `telemetry` feature flag. Torii code didn't compile if this feature was disabled.
- clean dependencies of `iroha` CLI

### Linked issue

Addresses #4136, but doesn't close it.


mversic:
I'm not sure we need this. I'd like to get others opinions on this. Having multiple crates might speed up compilation, on the other hand `lib` and `bin` should be compiled as separate crates

6r1d:
Given the [rationale](https://github.com/hyperledger/iroha/pull/4139#issue-2037569803), I don't see the [compilation speed](https://github.com/hyperledger/iroha/pull/4139#issuecomment-1853426909) as a main aspect. I agree with the rationale, and I'm in favour of the change.

mversic (approved):

cli/Cargo.toml
```diff
@@ -91,10 +76,10 @@ vergen = { workspace = true, features = ["cargo"] }
 
 [package.metadata.cargo-all-features]
 denylist = [
-"bridge",
-"dex",
-"schema-endpoint",
-"telemetry",
-"test-network"
+  "bridge",
+  "dex",
```
Arjentix, line 80: Oh, we should remove that confusing stuff someday

torii/derive/src/lib.rs
Arjentix, line 1: Well, there are no derives here actually. Maybe rename to `macro`?

Arjentix (approved):

cli/Cargo.toml
0x009922: Issued https://github.com/hyperledger/iroha/issues/4152

torii/derive/src/lib.rs
0x009922: Done.

mversic: should we do this consistently across codebase? I thought the practice was to name proc-macro crates `derive`. Whether or not they had `derive` macros, `attribute` macros or `function-like` macros

Erigara:
I just get another idea why using splitting torii might be good idea. 
I think for some network setups it might be useful to have nodes for fault tolerance, but this nodes want serve users requests and will only participate in consensus.

torii/src/routing.rs
```diff
@@ -179,7 +180,7 @@ async fn handle_post_configuration(
 }
 
 #[iroha_futures::telemetry_future]
-async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::Result<()> {
+pub async fn handle_blocks_stream(kura: Arc<Kura>, mut stream: WebSocket) -> eyre::Result<()> {
```
Erigara, line 183: Why this functions have to be public now?

torii/derive/src/lib.rs
Erigara: I also thought that it's a convention, however i'm not sure about this anymore (e.g. `tokio` calls it's proc macro crate `tokio-macros`)

torii/src/routing.rs
0x009922: Because `Torii::new` now defined next to `Torii` itself, in the parent module.

0x009922:
> So torii could be excluded from compilation.

Well, Torii might be excluded with feature flags anyway, without necessarily being a separate crate.

Erigara (approved):

torii/derive/src/lib.rs
0x009922: Created https://github.com/hyperledger/iroha/issues/4154

Arjentix (approved):

Issue #4105, opened 2023-11-30
[suggestion] Rewrite the iroha_cli entrypoint using clap


DCNick3:
Right now in the `iroha_cli` we are using a hand-written parser which already handles a bunch of arguments. We should rewrite it to use a more declarative approach with clap

              At some point we might consider to rewrite this using `clap`.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/4100#discussion_r1410165928_

Issue #3506, opened 2023-05-21
[BUG]: `disable_panic_terminal_colors` is deprecated, but couldn't be omitted


0x009922:
### Bug description

If you set `disable_panic_terminal_colors` configuration parameter, you get deprecation warning. If you omit this parameter in the configuration, you get a configuration parsing error and Iroha doesn't even start.

appetrosyan:
As of #3551, this should no longer be an issue.

Issue #3075, opened 2023-01-23
[suggestion]: Disallow invalid transactions in genesis block


Erigara:
Let me rephrase. If a transaction in a genesis fails we seem to just skip it and log error. Why don't we fail blockchain initialization if there is even one invalid transaction?

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/3074#discussion_r1084054877_

mversic:
The point here is, why do we tolerate invalid transactions in the genesis block? IMHO genesis block should be rejected entirely if it contains even one invalid transaction.

I would like to hear others opinion on this matter. Can you come up with a reason where it's expected to find invalid transaction in the genesis block?

Erigara:
I can't find reason for such behavior especially when in almost all cases our genesis block has exactly 1 transaction.

timofeevmd:
Done

![image.png](https://images.zenhubusercontent.com/6454ddfaa513c57f8883fb16/861dbb70-ac45-4fa1-bdca-c27bbe4d2bef)

Erigara:
I think we can close this since it was verified by @timofeevmd.

PR #4153, opened 2023-12-15
[refactor] #4029, #4105, #4136: refactor CLIs

### LINKED ISSUE #4029: [suggestion] Simplify genesis config
## Description

I propose to simplify genesis configuration a little bit by moving it to config file & env and dropping the related CLI flag.

### How it works now

There are the following configurations:

- `--submit-genesis` CLI flag to enable submitting for a particular field
- `IROHA_GENESIS` env var with path to genesis file (defaults to `genesis.json`)
- `genesis.public_key` (`GENESIS_PUBLIC_KEY`) parameter that must be set for ALL peers
- `genesis.private_key` (`GENESIS_PRIVATE_KEY`) parameter that must be set when `--submit-genesis` is set.

What is wrong:

- `IROHA_GENESIS` can only be set via ENV
- Intention to submit genesis is duplicated by setting `--submit-genesis` and `genesis.private_key`

### Proposed change

- `genesis.public_key`, required for all
- `genesis.private_key` and `genesis.file` for the peer that submits the genesis
- `genesis.file` _might_ have a default value (`genesis.json`) which is applied if only `genesis.private_key` is set. Although it might make behaviour somewhat unclear. I think enforcing `genesis.file` to be set explicitly will be more reliable.

Benefits:

- Flexible: genesis might be configured via only config file, or via only ENV, or mixed
- Simple: no extra CLI flag

0x009922:
## Description

**Changes:**

- **`iroha_genesis`:**
    - Refactor `GenesisNetwork` initialisation, remove side effects and simplify logic
    - Make `transactions` in `GenesisNetwork` non-public, to avoid invariant violation
- **`iroha_config`:**
    - Update Genesis config: instead of `account_public_key` and `account_private_key`, it now has `public_key`, `private_key`, and `file` options (according to #4029). Move some validation logic here (in form of `ParsedConfiguration` struct).
    - Finally, make `disable_panic_terminal_colors` deprecated option truly optional (#3506)
    - Use `PathBuf`s for all path parameters (`genesis.file`, `kura.block_store_path`, `snapshot.dir_path`, `telemetry.file`)
    - Update `Path` (multiple extensions support) behaviour
- **`iroha`:**
    - Use `clap`
    - Update CLI itself: `--config` argument, `--terminal-colors` (or `--terminal-colors=false`). `--submit-genesis` remains as is.
    - Update configuration pipeline:
        - Steps: defaults → file → env → build → validate & parse
        - If file is not specified, Iroha tries both `config.json` and `config.json5` and doesn't fail if they are not find (same as before)
        - If file is user-provided (`--config`, `IROHA_CONFIG=`) and it is not found, Iroha fails (previously it didn't)
        - _Read and parse_ ENV after file, not before
        - Resolve paths in the config file relatively to its location
- **`iroha_client_cli`:**
    - Update `clap` to `v4`, as in the rest of workspace
    - Refactor `--metadata` argument
    - Adopt the same configuration pipeline as in `iroha`
- **`iroha_swarm`:** explicitly specifies `config.json` and `genesis.json` paths in the generated docker-compose configurations
- **`kagami`:** `kagami config peer` now accepts an optional argument `--genesis-file-in-config <PATH>` to set `genesis.file` in the output. Semantic analogous to `kagami genesis --executor-path-in-genesis <PATH>`.

### Linked issues

Closes #4029 #4105, partially addresses #4136

### Benefits

- Improves UX of configuration & usage of `iroha` and `iroha_client_cli`
- Moves more configuration responsibility to `iroha_config`
- Reduces complexity of init logic a little

### Checklist

- [x] Technical writers review on CLIs
- [x] https://github.com/hyperledger/iroha/pull/4139
- [x] Check docker compose files manually


cli/src/main.rs
```diff
@@ -1,235 +1,163 @@
...
+    /// This argument must be set alongside with `genesis.file` and `genesis.private_key`
+    /// configuration options. If not, Iroha will exit with an error.
+    ///
+    /// This argument must be set if the amount of trusted peers in the config file
+    /// (`sumeragi.trusted_peers`) is less than 2, i.e. the network consists only from this peer
+    /// itself. Otherwise it would be impossible to receive genesis topology, and Iroha will exit
+    /// with an error.
+    #[arg(long)]
+    submit_genesis: bool,
+}
```
0x009922, line 72: @outoftardis @yamkovoy please take a look.

client_cli/src/main.rs
```diff
@@ -1,86 +1,97 @@
...
+    /// environment variables. However, if the config path is set explicitly with this argument
+    /// and the file is not found, Iroha Client will exit with an error.
+    #[arg(
+        short,
+        long,
+        value_name("PATH"),
+        value_hint(clap::ValueHint::FilePath),
+        value_parser(Path::user_provided_str)
+    )]
+    config: Option<Path>,
```
0x009922, line 80: @outoftardis @yamkovoy please take a look.

(Essentially the same text as in `cli/src/main.rs`)

cli/src/main.rs
```diff
@@ -1,235 +1,163 @@
...
+    /// is not found, Iroha will exit with an error.
+    #[arg(
+        long,
+        short,
+        env("IROHA_CONFIG"),
+        value_name("PATH"),
+        value_parser(Path::user_provided_str),
+        value_hint(clap::ValueHint::FilePath)
+    )]
+    // TODO: use value parser as in client cli
```
0x009922, line 42: Will remove

config/src/genesis.rs
```diff
@@ -12,22 +16,85 @@ view! {
...
+        /// Genesis account public key
+        public_key: PublicKey,
+    },
+    /// The peer is responsible for submitting of the genesis block
+    Submit {
+        /// Genesis account key pair
+        key_pair: KeyPair,
+        /// Raw genesis block
+        raw_block: RawGenesisBlock,
+    },
```
0x009922, line 54: Better naming ideas?

What I have: `Partial`/`Full`, `PublicKey`/`Full`

README.md
```diff
@@ -157,7 +157,18 @@ A brief overview on how to configure and maintain an Iroha instance:
 
 ### Configuration
 
-**Note:** this section is under development. You can track it in the [issue](https://github.com/hyperledger/iroha-2-docs/issues/392).
+There is a set of configuration parameters that could be passed either through a configuration file or environment variables.
+
+```shell
+# will look up for `config.json` or `config.json5`, without failing if files are not found
```
outoftardis, line 163:
```suggestion
# look for `config.json` or `config.json5` (won't fail if files are not found)
```

README.md
```diff
@@ -198,9 +209,9 @@ prometheus --config.file=configs/prometheus.yml
 
 ### Storage
 
-The blocks are written to the `blocks` sub-folder, which is created automatically by Iroha in the working directory of the peer. Additionally, if specified, the logging file must also be stored in a user-specified directory.
+Iroha stores blocks and snapshots in the `storage` directory, which is created automatically by Iroha in the working directory of the peer. If `kura.block_store_path` is overwritten in the config file, it will be resolved relative to the config file location.
```
outoftardis, line 212:
```suggestion
Iroha stores blocks and snapshots in the `storage` directory, which is created automatically by Iroha in the working directory of the peer. If `kura.block_store_path` is specified in the config file, it overwrites the default one and is resolved relative to the config file location.
```

config/src/iroha.rs
```diff
@@ -20,8 +20,9 @@ view! {
         /// Private key of this peer
         #[view(ignore)]
         pub private_key: PrivateKey,
-        /// Disable coloring of the backtrace and error report on panic
-        pub disable_panic_terminal_colors: bool,
+        /// Disable coloring of the backtrace and error report on panic.
+        /// It is deprecated.
```
outoftardis, line 24: I think this is better

```suggestion
        /// **Deprecated**.
        /// Disable coloring of the backtrace and error report on panic.
```

config/src/genesis.rs
```diff
@@ -12,22 +16,85 @@ view! {
...
-            account_public_key: None,
-            account_private_key: Some(None),
+            public_key: None,
+            private_key: Some(None),
+            file: None,
         }
     }
 }
 
+/// Parsed variant of the (user provided) [`Configuration`]
```
outoftardis, line 39:
```suggestion
/// Parsed variant of the user-provided [`Configuration`]
```

config/src/genesis.rs
```diff
@@ -12,22 +16,85 @@ view! {
...
+// TODO: incorporate this struct into the final, parsed configuration
+//       https://github.com/hyperledger/iroha/issues/3500
+pub enum ParsedConfiguration {
+    /// The peer can only observe the genesis block
+    Default {
+        /// Genesis account public key
+        public_key: PublicKey,
+    },
+    /// The peer is responsible for submitting of the genesis block
```
outoftardis, line 48:
```suggestion
    /// The peer is responsible for submitting the genesis block
```

config/src/genesis.rs
outoftardis: I think `Partial`/`Full` is the best one among the options you suggested.

maybe something like `ObserveGenesis`/`SubmitGenesis`? Or `View`/`Submit`?

cli/src/main.rs
```diff
@@ -1,235 +1,162 @@
...
-    } else {
-        args.genesis_path = None;
-    }
+/// Iroha peer Command-Line Interface.
+#[derive(Parser, Debug)]
+#[command(name = "iroha", version = concat!("version=", env!("CARGO_PKG_VERSION"), " git_commit_sha=", env!("VERGEN_GIT_SHA")), author)]
+struct Args {
+    /// Path to the configuration file, defaults to `config.json`/`config.json5`
+    ///
+    /// Supported extensions are `.json` and `.json5`. By default, Iroha looks up for a
```
outoftardis, line 29:
```suggestion
    /// Supported extensions are `.json` and `.json5`. By default, Iroha looks for a
```

cli/src/main.rs
```diff
@@ -1,235 +1,162 @@
...
+    /// `config` file in the Current Working Directory with both supported extensions.
```
outoftardis, line 30:
```suggestion
    /// `config` file with one of the supported extensions in the current working directory.
```

cli/src/main.rs
```diff
@@ -1,235 +1,162 @@
...
+        default_missing_value("true"),
+        default_value(default_terminal_colors_str()),
+        action(clap::ArgAction::Set),
+        require_equals(true),
+        num_args(0..=1),
+    )]
+    terminal_colors: bool,
+    /// Whether the current peer should submit the genesis block or not
+    ///
+    /// The only one peer in the network should submit the genesis block.
```
outoftardis, line 60:
```suggestion
    /// Only one peer in the network should submit the genesis block.
```

cli/src/main.rs
```diff
@@ -1,235 +1,162 @@
...
+    ///
+    /// This argument must be set alongside with `genesis.file` and `genesis.private_key`
+    /// configuration options. If not, Iroha will exit with an error.
+    ///
+    /// This argument must be set if the amount of trusted peers in the config file
+    /// (`sumeragi.trusted_peers`) is less than 2, i.e. the network consists only from this peer
+    /// itself. Otherwise it would be impossible to receive genesis topology, and Iroha will exit
+    /// with an error.
```
outoftardis, line 68:
```suggestion
    /// This argument must be set if the amount of trusted peers in the config file
    /// (`sumeragi.trusted_peers`) is less than 2, i.e. the network consists only of this one peer.
    /// If the argument is not set, it will be impossible to receive genesis topology, and Iroha
    /// will exit with an error.
```

`to receive genesis topology` -- I don't really understand what it means

client_cli/src/main.rs
```diff
@@ -1,86 +1,97 @@
...
+#[derive(clap::Parser, Debug)]
+#[command(name = "iroha_client_cli", version = concat!("version=", env!("CARGO_PKG_VERSION"), " git_commit_sha=", env!("VERGEN_GIT_SHA")), author)]
 struct Args {
-    /// Sets a config file path
-    #[structopt(short, long)]
-    config: Option<Configuration>,
+    /// Path to the configuration file, defaults to `config.json`/`config.json5`
+    ///
+    /// Supported extensions are `.json` and `.json5`. By default, Iroha looks up for a
+    /// `config` file in the Current Working Directory with both supported extensions.
```
outoftardis, line 69: same suggestion as in client

torii/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+                // **after** `.with(trace)`
+                health_route,
+            )
+            .or(ws_router
+                .or(get_router)
+                .or(post_router)
+                .with(warp::trace::request()))
+    }
+
+    /// Start main api endpoints.
```
outoftardis, line 207:
```suggestion
    /// Start main API endpoints.
```

torii/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+    #[cfg(feature = "telemetry")]
+    /// Error while getting Prometheus metrics
+    Prometheus(#[source] eyre::Report),
+    #[cfg(feature = "telemetry")]
+    /// Internal error while getting status
+    StatusFailure(#[source] eyre::Report),
+    /// Failure caused by configuration subsystem
+    ConfigurationFailure(#[from] KisoError),
+    /// Cannot find status segment by provided path
+    StatusSegmentNotFound(#[source] eyre::Report),
```
outoftardis, line 284:
```suggestion
    /// Failed to get or set configuration
    Config(#[source] eyre::Report),
    /// Failed to push into queue
    PushIntoQueue(#[from] Box<queue::Error>),
    #[cfg(feature = "telemetry")]
    /// Failed to get Prometheus metrics
    Prometheus(#[source] eyre::Report),
    #[cfg(feature = "telemetry")]
    /// Internal error while getting status
    StatusFailure(#[source] eyre::Report),
    /// Failure caused by configuration subsystem
    ConfigurationFailure(#[from] KisoError),
    /// Failed to find status segment by provided path
    StatusSegmentNotFound(#[source] eyre::Report),
```

config/src/genesis.rs
0x009922: > maybe something like `ObserveGenesis`/`SubmitGenesis`? Or `View`/`Submit`?

I like these options too. Let's wait for more opinions.

cli/src/main.rs
0x009922: > `to receive genesis topology` -- I don't really understand what it means

TBH, me neither! I copied the exact phrase from an error message. I will check it.

client_cli/src/main.rs
0x009922: Applied

client_cli/Cargo.toml
```diff
@@ -25,6 +25,7 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 iroha_client = { workspace = true }
 iroha_primitives = { workspace = true }
+iroha_config = { workspace = true }
```
Arjentix, line 28: Shouldn't it be visible from `iroha_client`?

tools/kagami/src/config.rs
```diff
@@ -64,16 +64,31 @@ mod client {
...
 
-    #[derive(ClapArgs, Debug, Clone, Copy)]
-    pub struct Args;
+    #[derive(ClapArgs, Debug, Clone)]
+    pub struct Args {
+        /// Specifies the value of `genesis.file` configuration parameter.
+        ///
+        /// Note: relative paths are not resolved but included as-is.
+        #[arg(long)]
+        genesis_file_in_config: Option<PathBuf>,
```
Arjentix, line 79: What about `submit-genesis` (in accordance to `iroha` cli) or `include-genesis`?

config/src/genesis.rs
```diff
@@ -73,16 +73,19 @@ impl Configuration {
...
 
 /// Error which might occur during [`Configuration::parse()`]
 #[derive(Debug, displaydoc::Display, thiserror::Error)]
 pub enum ParseError {
-    /// `genesis.private_key` and `genesis.file` should be set for the peer that submits the genesis
-    InvalidParametersCombination,
+    /// `--submit-genesis` was provided, but `genesis.private_key` and/or `genesis.file` are missing
+    SubmitIsSetButRestAreNot,
+    /// `--submit-genesis` was not provided, but `genesis.private_key` and/or `genesis.file` are set
+    SubmitIsNotSetButRestAre,
```
Arjentix, line 88: Can we refactor it in that style?

```rust
pub enum SubmitSetError {
    PrivateKeyMissing,
    FileMissing,
}
```

I just don't like to provide uncertainty when we can capture the exact error

tools/swarm/src/cli.rs
```diff
@@ -27,7 +27,7 @@ pub struct Cli {
     pub no_banner: bool,
     /// Path to a directory with Iroha configuration. It will be mapped as volume for containers.
     ///
-    /// The directory should contain `config.json` with `genesis.file` parameter specified.
+    /// The directory should contain `config.json` and `genesis.json`
```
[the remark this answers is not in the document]
0x009922: Yes, but soon enough it will be `toml` without multi-extension support. And it is also possible to put JSON5 with `.json` extension and Iroha will be fine with that.

cli/src/lib.rs
```diff
@@ -494,21 +512,25 @@ pub fn read_config(
...
-                }
-            }
-        }
+        let join_to_config_dir = |x: &mut PathBuf| {
+            *x = config_dir.join(&x);
+        };
+        mutate_nested_option!(cfg, [genesis, file, self], join_to_config_dir);
+        mutate_nested_option!(cfg, [snapshot, dir_path], join_to_config_dir);
+        mutate_nested_option!(cfg, [kura, block_store_path], join_to_config_dir);
+        mutate_nested_option!(cfg, [telemetry, file, self], join_to_config_dir);
```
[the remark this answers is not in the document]
0x009922: It wouldn't. While `snapshot.dir_path` and `kura.block_store_path` are `Option<T>`, `genesis.file` and `telemetry.file` are `Option<Option<T>>` (because they are optional in the final config).

tools/kagami/src/config.rs
0x009922: - `submit-genesis` doesn't really communicate that the value of this argument is the path of the `genesis.file` parameter that will be used as is
- `include-genesis` tells more that Kagami will read the path and include genesis into the config file

(IMO)

So, I chose the same approach as with `kagami genesis --executor-path-in-genesis <PATH>`

config/src/genesis.rs
0x009922: To be exact, there are 6 possible error combinations:

| `--submit-genesis` | `genesis.private_key` | `genesis.file` |
| :-: | :-: | :-: |
| `true` | None | None |
| `true` | Some | None |
| `true` | None | Some |
| `false` | Some | None |
| `false` | None | Some |
| `false` | Some | Some |

It would be great to tell the user about the exact error, I agree. However, I feel reluctant to be very detailed on this stage of config overhaul, since genesis design & config might be changed soon, and I am anyway going to completely re-write `iroha_config` as a next stage, which will also address errors detalisation, not only in genesis.

genesis/src/lib.rs
```diff
@@ -69,23 +58,33 @@ impl GenesisNetwork {
...
-                .ok()
-            }).map(GenesisTransaction)
-            .collect::<Vec<_>>();
+            .map(|(i, raw_transaction)| {
+                raw_transaction
+                    // FIXME: fix underlying chain of `.sign` so that it doesn't
+                    //        consume the key pair unnecessarily 
+                    .sign(genesis_key_pair.clone())
+                    .map(GenesisTransaction)
+                    .wrap_err_with(|| eyre!("Failed to sign transaction #{i}"))
```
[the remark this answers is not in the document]
0x009922: Depends really on personal perception =)
For me zero-based indexing seems more natural in this context. Perhaps it would be clearer to say

> Failed to sign transaction _at index_ 0

cli/src/lib.rs
```diff
@@ -530,32 +483,75 @@ fn genesis_domain(public_key: PublicKey) -> Domain {
...
-        .map_err(Into::into)
+        .build()?;
+
+    // TODO: move validation logic below to `iroha_config`
+
+    if !config.disable_panic_terminal_colors {
+        // FIXME: it shouldn't be logged here; it is a part of configuration domain
+        //        this message can be very simply broken by the changes in the configuration
+        //        https://github.com/hyperledger/iroha/issues/3506
+        eprintln!("The configuration parameter `DISABLE_PANIC_TERMINAL_COLORS` is deprecated. Set `TERMINAL_COLORS=false` instead. ")
```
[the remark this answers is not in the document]
0x009922: Indeed, unnecessary to keep it now.

client_cli/Cargo.toml
0x009922: `iroha_client` doesn't re-export `iroha_config_base` containing necessary traits.

Will change this dependency to `iroha_config_base` instead.

genesis/src/lib.rs
```diff
@@ -34,58 +33,66 @@ pub struct GenesisTransaction(pub SignedTransaction);
...
+            )
+            .into()],
         })
         .chain(raw_block.transactions);
 
         #[cfg(test)]
-        let transactions_iter = raw_block.transactions.into_iter();
+        let transactions_iter = {
+            if raw_block.transactions.is_empty() {
+                eyre::bail!("Genesis transaction set is empty");
```
DCNick3, line 64: Hm, this doesn't error out if `iroha_genesis` is not compiled for unit tests. Is this the logic you intended?

(it's weird that we have different behavior with `cfg(test)`.. Do you know why is that?)

scripts/test_env.py
```diff
@@ -127,12 +129,20 @@ def run(self, is_genesis: bool = False):
...
         os.environ["TORII_API_URL"] = f"{self.host_ip}:{self.api_port}"
 
-        genesis_arg = "--submit-genesis" if is_genesis else ""
+        if submit_genesis:
+            os.environ["IROHA_GENESIS_PRIVATE_KEY"] = self.private_key
+            # Assuming it was copied to the peer's directory
+            os.environ["IROHA_GENESIS_FILE"] = str(self.peer_dir.joinpath("genesis.json"))
+        else:
+            for x in ["IROHA_GENESIS_PRIVATE_KEY", "IROHA_GENESIS_FILE"]:
+                os.environ.pop(x, None)
```
DCNick3, line 138: https://docs.python.org/3/library/subprocess.html#subprocess.Popen accepts an `env` parameter, allowing to specify the process environment variables without relying on global state. 

The caveat is that it will require copying the "base" `os.environ`, as it doesn't have any inheritance logic.

client_cli/src/main.rs
```diff
@@ -1,86 +1,97 @@
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{self:?}")
-    }
+/// Re-usable clap `--metadata <PATH>` (`-m`) argument.
+/// Should be combined with `#[command(flatten)]` attr.
+#[derive(clap::Args, Debug, Clone)]
+// FIXME: `pub` is needed because Rust complains about "leaking private types"
+//        when this type is used inside of modules. I don't know how to fix it.
+pub struct MetadataArgs {
+    // The JSON/JSON5 file with key-value metadata pairs
```
DCNick3, line 34:
```suggestion
    /// The JSON/JSON5 file with key-value metadata pairs
Comment
```

config/iroha_test_config.json
```diff
@@ -4,7 +4,7 @@
     "digest_function": "ed25519",
     "payload": "282ED9F3CF92811C3818DBC4AE594ED59DC1A2F78E4241E31924E101D6B1FB831C61FAF8FE94E253B93114240394F79A607B7FA55F9E5A41EBEC74B88055768B"
   },
-  "DISABLE_PANIC_TERMINAL_COLORS": false,
+  "DISABLE_PANIC_TERMINAL_COLORS": null,
```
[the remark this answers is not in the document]
0x009922: See description of [`d88fa05` (#4153)](https://github.com/hyperledger/iroha/pull/4153/commits/d88fa05aca44a79c3f3efb0780b176747eba11cf) - it is here really to workaround how current configuration works.

The test that checks this JSON relies only on the JSON itself. It doesn't apply defaults, doesn't merge anything else. When it _builds_ the config, it expects all the fields to be presented, including `disable_panic...`. The test fails without this field being at least `null`.

Anyway, as @Arjentix suggested, I'll simply delete this parameter in this PR.

genesis/src/lib.rs
0x009922: I found some weird code in this function:

- In `cfg(test)` it doesn't put Executor in the transactions
- Then it signs each transaction and filters out failing ones, although if one fails, every other one will also fail
- Then, it checks if the resulting transactions set is empty and `bail`s if so

I made a conclusion that it is possible to get empty transactions set only in `cfg(test)`. 

Anyway, I think it is better to remove this logic out of here completely to the `TestGenesis` you mentioned in another discussion.

0x009922: Ah, and more context:

- https://github.com/hyperledger/iroha/issues/3075

scripts/test_env.py
0x009922: Thanks for the catch. I think it would be more clear to opt-out from `os.environ` fully and always specify `env` explicitly.

config/src/path.rs
```diff
@@ -6,113 +6,146 @@ use alloc::borrow::Cow;
...
-            "Default configuration path should have no extension"
-        );
-
-        Self(Default(path.to_owned()))
+    /// # Errors
+    /// If `path` contains an extension.
+    pub fn default(path: impl AsRef<std::path::Path>) -> Result<Self> {
+        let path = path.as_ref();
+
+        path.extension().map_or_else(
```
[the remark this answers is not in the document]
0x009922: Perhaps, checking for extension here is redundant indeed.

client/tests/integration/permissions.rs
```diff
@@ -24,7 +24,7 @@ fn genesis_transactions_are_validated() {
         AccountId::from_str("alice@wonderland").unwrap(),
     );
 
-    let tx_ref = &mut genesis.transactions.last_mut().unwrap().0;
+    let tx_ref = &mut genesis.transactions_mut().last_mut().unwrap().0;
```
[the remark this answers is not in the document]
0x009922: Yes. Last `.0` owns the value for a reason I don't understand. Do you have an idea?

`unwrap()` returns `&mut GenesisTransaction`, which is a newtype around `SignedTransaction`, which doesn't implement `Copy`.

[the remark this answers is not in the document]
0x009922: Yeah, `TestGenesis` returning an `Option` doesn't make sense.

BTW, I am confused by both tests patching constructed `GenesisNetwork`. They are inserting plain instructions into the payload of a signed transaction. By doing so, AFAIK, they invalidate the signature of the transaction being computed using the previous payload. One test checks that Iroha validates genesis by providing an invalid permission token, but Iroha should not accept genesis because of its signatures I guess. Another test checks whether base64 triggers are handled fine. I don't why the second test passes.

config/src/path.rs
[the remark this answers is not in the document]
DCNick3: I think it's fine to check the extension here, just that it should be a panic (as it's a problem with a hard-coded string and should be fixed by the programmer)

scripts/test_env.py
DCNick3: It's not always safe to opt-out from the `os.environ`, there are variables like `SystemRoot` on windows and `PATH` on linux.

I _think_ iroha will be fine, but it might lead to unnecessary and potentially hard-to-debug problems imo..

config/src/genesis.rs
Arjentix: Fair enough

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
Arjentix: > Last .0 owns the value for a reason I don't understand. Do you have an idea?

Oh, I guess it's the same problem as borrowing field value fom `&self` -- Rust always requires to explicitly tell that we are borrowing the value:

```rust
fn foo(&self) -> {
    let a = self.a; // Error if `a` doesn't implement `Copy`
    let a = &self.a; // Works
}
```

I also have no idea why did they implement it like that...

client_cli/src/main.rs
```diff
@@ -31,7 +31,8 @@ use iroha_primitives::addr::SocketAddr;
 // FIXME: `pub` is needed because Rust complains about "leaking private types"
 //        when this type is used inside of modules. I don't know how to fix it.
 pub struct MetadataArgs {
-    // The JSON/JSON5 file with key-value metadata pairs
+    /// The JSON/JSON5 file with key-value metadata pairs
+Comment
```
Arjentix, line 32: I guess @DCNick3 made a typo here...

config/src/genesis.rs
0x009922: Since there are not many more opinions here, I'll use `Partial`/`Full`. Not necessarily in this PR, but as part of https://github.com/hyperledger/iroha/issues/4161

cli/src/lib.rs
```diff
@@ -530,32 +483,75 @@ fn genesis_domain(public_key: PublicKey) -> Domain {
...
+    submit_genesis: bool,
+) -> Result<(Configuration, Option<GenesisNetwork>)> {
+    let config = ConfigurationProxy::default();
+
+    let config = if let Some(actual_config_path) = path.try_resolve() {
+        let mut cfg = config.override_with(ConfigurationProxy::from_path(&*actual_config_path));
+
+        // careful here: `genesis.file` might be a path relative to the config file.
+        // we need to resolve it before proceeding
+        // TODO: move this logic into `iroha_config`
```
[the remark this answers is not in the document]
0x009922: Will add it: https://github.com/hyperledger/iroha/issues/4161

cli/src/lib.rs
```diff
@@ -530,32 +483,75 @@ fn genesis_domain(public_key: PublicKey) -> Domain {
...
+
+    if !config.disable_panic_terminal_colors {
+        // FIXME: it shouldn't be logged here; it is a part of configuration domain
+        //        this message can be very simply broken by the changes in the configuration
+        //        https://github.com/hyperledger/iroha/issues/3506
+        eprintln!("The configuration parameter `DISABLE_PANIC_TERMINAL_COLORS` is deprecated. Set `TERMINAL_COLORS=false` instead. ")
+    }
+
+    if !submit_genesis && config.sumeragi.trusted_peers.peers.len() <= 1 {
+        return Err(eyre!("Only peer in network, yet required to receive genesis topology. This is a configuration error."));
```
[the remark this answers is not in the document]
0x009922: Updated this and other messages majorly, so closing this conversation.

cli/src/lib.rs
```diff
@@ -530,32 +483,75 @@ fn genesis_domain(public_key: PublicKey) -> Domain {
...
+        }
+
+        cfg
+    } else {
+        config
+    };
+
+    // it is not chained to the previous expressions so that config proxy from env is evaluated
+    // after reading a file
+    let config = config
```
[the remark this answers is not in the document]
0x009922: Not sure what do you mean. Could you reframe your comment?

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
0x009922: Created an question:

- https://github.com/hyperledger/iroha/issues/4162

config/iroha_test_config.json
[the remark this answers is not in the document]
0x009922: Removed

scripts/test_env.py
0x009922: Thanks for the catch again. Will refactor in a way so that initial `os.environ` data is also used.

config/src/path.rs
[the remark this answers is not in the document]
0x009922: What is the difference between panicking here, and returning a `Result` and panicking outside?

client_cli/src/main.rs
0x009922: Missed it, fixed.

scripts/test_env.py
0x009922: Refactored.

cli/src/lib.rs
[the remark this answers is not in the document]
0x009922: Linked.

config/src/genesis.rs
0x009922: Changed it now

cli/src/lib.rs
```diff
@@ -483,6 +483,24 @@ fn genesis_domain(public_key: PublicKey) -> Domain {
...
+    ($obj:expr, [$field:ident, $($rest:tt)+], $func:expr) => {
+        $obj.$field.as_mut().map(|x| {
+            mutate_nested_option!(x, [$($rest)+], $func)
+        })
+    };
+    ($obj:tt, [$field:tt], $func:expr) => {
+        mutate_nested_option!($obj, $field, $func)
+    };
+
+}
```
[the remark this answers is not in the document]
0x009922: Applied.

cli/src/lib.rs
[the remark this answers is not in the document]
DCNick3: I guess the problem is that the doc string says that the function "Reads configuration from the specified path and validates it.", while in fact it also reads environment variables, which is surprising.

config/src/path.rs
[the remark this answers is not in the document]
DCNick3: The main difference imo is just better ergonomics (less code to write to call this function):
`Path::default("config")` vs `Path::default("config").expect("this path is valid I swear")`. Again, seeing how in most places the result is effectively constant and it always ends up unwrapped right away, I do not see the benefit of using `Result` (which shines when you actually want to handle errors)

cli/src/lib.rs
```diff
@@ -530,56 +483,250 @@ fn genesis_domain(public_key: PublicKey) -> Domain {
...
+    let config = config
         .build()
-        .map_err(Into::into)
+        .wrap_err("Failed to finalize configuration")?;
+
+    // TODO: move validation logic below to `iroha_config`
+
+    if !submit_genesis && config.sumeragi.trusted_peers.peers.len() < 2 {
+        return Err(eyre!("\
+            The network consists from this one peer only (`sumeragi.trusted_peers` is less than 2). \
```
DCNick3, line 554: The `fails_with_no_trusted_peers_and_submit_role` test fails because of the updated test message

outoftardis (approved):
doc comments lgtm

config/src/path.rs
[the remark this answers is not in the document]
0x009922: That makes sense, thank you

cli/src/lib.rs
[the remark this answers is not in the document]
0x009922: Got it. Expanded the documentation of `read_config` function.

config/src/path.rs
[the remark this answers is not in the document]
0x009922: Changed to `panic!` internally

cli/src/lib.rs
0x009922: Thanks, fixed.

DCNick3 (approved):

Arjentix (approved):

PR #4156, opened 2023-12-15
[fix] #3857: Unify permission tokens in executor

### LINKED ISSUE #3857: [BUG] `Eq`/`Ord` for `PermissionToken` implementations are broken
`PermissionToken` has to implement both `Eq` and `Ord`. This is unfortunate because without having access to the wasm validator smart contract it's impossible to compare permission tokens. 

Take the following example:
```rs
let token1 = PermissionToken::new("Token1", json!({"asset_id": "rose#wonderland#alice@wonderland"}));
let toke2n = PermissionToken::new("Token1", json!({"asset_id": "rose##alice@wonderland"}));
assert_eq!(token1, token2)
```

because host cannot know how to parse the payload except as JSON it's impossible for the host to compare these tokens.

Possible solution
1. Create `UnverifiedPermissionToken` comparable to `NewAccount`:
before executing `Grant<Account, UnverifiedPermissionToken>` it will get converted into `PermissionToken` in the executor by first decoding as a concrete permission token and then serialized back to string before being stored on chain. `UnverifiedPermissionToken`will not have `Eq`/`Ord` implemented and `PermissionToken` will simply via string comparison (not even JSON comparison will be required).

Erigara:
## Description

Add unification of permission tokens in executor when granting permission tokens or registering role.
Unification is serialization and deserialization of permission token to remove ambiguity introduced by json format. 

I've tried approach proposed by original issue initially, but it didn't work well because:
- host in any case unable to check that permission tokens are indeed unified
- bloats amount of code
- it's still required to implement `Eq, Hash, ...` on `UnverifiedPermissionToken` because it's part of value

### Linked issue

Closes #3857 

### Benefits

Avoid certain bugs depending on permission token compassion.


smart_contract/executor/derive/src/token.rs
```diff
@@ -63,5 +63,18 @@ fn impl_try_from_permission_token(ident: &syn2::Ident, generics: &syn2::Generics
...
         }
+
+        impl #impl_generics ::core::convert::TryFrom<#ident #ty_generics> for ::iroha_executor::data_model::permission::PermissionToken #where_clause {
+            type Error = ::iroha_executor::permission::PermissionTokenConversionError;
+
+            fn try_from(token: #ident #ty_generics) -> ::core::result::Result<Self, Self::Error> {
+                let definition_id = #token_id;
+
+                let payload = ::serde_json::to_value::<#ident #ty_generics>(token)
+                    .map_err(::iroha_executor::permission::PermissionTokenConversionError::Serialize)?;
```
mversic, line 74: I think it's ok to panic here. It's assumed that executor is valid in the sense that permission tokens are valid. An error here means that executor is faulty, and that is not a recoverable error

client/tests/integration/roles.rs
```diff
@@ -172,3 +172,42 @@ fn role_with_invalid_permissions_is_not_accepted() -> Result<()> {
...
+    let role = test_client
+        .request(FindRoleByRoleId::new(role_id))
+        .expect("failed to find role");
+
+    // Permission tokens are unified so only one token left
+    assert_eq!(
+        role.permissions().len(),
+        1,
+        "permission tokens for role aren't deduplicated"
+    );
```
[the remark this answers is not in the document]
Erigara: > Maybe that's outside of the scope of current PR

I don't get why this should be scope of another PR.

> Shouldn't we check that on role registration step instead?

No, because we need Executor to perform unification.
Before sending `NewRole` to iroha it's impossible to say that 2 json representations of token corresponds to the same concrete permission token type.

Arjentix (approved):

client/tests/integration/roles.rs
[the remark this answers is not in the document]
Arjentix: - Created #4158

smart_contract/executor/derive/src/token.rs
```diff
@@ -63,5 +63,19 @@ fn impl_try_from_permission_token(ident: &syn2::Ident, generics: &syn2::Generics
...
         }
+
+        impl #impl_generics ::core::convert::From<#ident #ty_generics> for ::iroha_executor::data_model::permission::PermissionToken #where_clause {
+            fn from(token: #ident #ty_generics) -> Self {
+                let definition_id = #token_id;
+
+                let payload = ::iroha_executor::smart_contract::debug::DebugExpectExt::dbg_expect(
+                    ::serde_json::to_value::<#ident #ty_generics>(token)
+                    .map_err(::iroha_executor::permission::PermissionTokenConversionError::Serialize),
+                    "failed to serialize concrete permission token type. This is a bug."
```
mversic, line 74: didn't we agree you will panic here?

mversic, line 74: didn't we agree you will panic here?

smart_contract/executor/src/permission.rs
```diff
@@ -53,6 +53,8 @@ pub enum PermissionTokenConversionError {
     Id(PermissionTokenId),
     /// Failed to deserialize JSON
     Deserialize(serde_json::Error),
+    /// Failed to serialize JSON
+    Serialize(serde_json::Error),
```
mversic, line 57: this is unnecessary, we should panic on invalid Executor

mversic (changes requested):

smart_contract/executor/derive/src/token.rs
Erigara: Overlooked this somehow

Erigara: We still panic using `dbg_expect`

mversic (approved):

PR #4159, opened 2023-12-19
[fix] #4133: Count identical wasm for triggers

### LINKED ISSUE #4133: [BUG]: "No original smartcontract saved for trigger"
```
2023-12-08 10:57:01 thread 'sumeragi thread' panicked at 'No original smartcontract saved for trigger. This is a bug.', core/src/smartcontracts/isi/triggers/set.rs:459:22
```

Originally found by @arndey 

By a quick look at the code it seems like the corresponding field or `trigger::Set` is not (de-)serialized anyhow, which may introduce this bug.

Hower @Mingela pointed that it hapens in a [test](https://github.com/soramitsu/iroha2-tools/blob/2fad47defccc967309f042e545c86cad02050562/wasm-triggers-migration/src/test/kotlin/jp/co/soramitsu/Tests.kt#L37) without loading from snapshot, so the problem needs furthes investigation.

Arjentix:
- See also #4160

### Linked issue

- Closes #4133 

### Checklist

### DIFF
--- a/core/src/smartcontracts/isi/triggers/set.rs
+++ b/core/src/smartcontracts/isi/triggers/set.rs
@@ -9,9 +9,9 @@
 //! trigger hooks.
 
 use core::cmp::min;
-use std::fmt;
+use std::{fmt, num::NonZeroU64};
 
-use indexmap::IndexMap;
+use indexmap::{map::Entry, IndexMap};
 use iroha_crypto::HashOf;
 use iroha_data_model::{
     events::Filter as EventFilter,
@@ -56,6 +56,15 @@ pub struct LoadedAction<F> {
     pub filter: F,
     /// Metadata used as persistent storage for trigger data.
     pub metadata: Metadata,
+}
+
+impl<F> LoadedAction<F> {
+    fn extract_blob_hash(&self) -> Option<HashOf<WasmSmartContract>> {
+        match self.executable {
+            LoadedExecutable::Wasm(LoadedWasm { blob_hash, .. }) => Some(blob_hash),
+            LoadedExecutable::Instructions(_) => None,
+        }
+    }
 }
 
 /// Trait common for all `LoadedAction`s
@@ -133,6 +142,10 @@ impl<F: Filter + Into<TriggeringFilterBox> + Clone> LoadedActionTrait for Loaded
     }
 }
 
+/// [`WasmSmartContract`]s by [`TriggerId`].
+/// Stored together with number to count triggers with identical [`WasmSmartContract`].
+type WasmSmartContractMap = IndexMap<HashOf<WasmSmartContract>, (WasmSmartContract, NonZeroU64)>;
+
 /// Specialized structure that maps event filters to Triggers.
 // NB: `Set` has custom `Serialize` and `DeserializeSeed` implementations
 // which need to be manually updated when changing the struct
@@ -149,7 +162,7 @@ pub struct Set {
     /// Trigger ids with type of events they process
     ids: IndexMap<TriggerId, TriggeringEventType>,
     /// Original [`WasmSmartContract`]s by [`TriggerId`] for querying purposes.
-    original_contracts: IndexMap<HashOf<WasmSmartContract>, WasmSmartContract>,
+    original_contracts: WasmSmartContractMap,
     /// List of actions that should be triggered by events provided by `handle_*` methods.
     /// Vector is used to save the exact triggers order.
     matched_ids: Vec<(Event, TriggerId)>,
@@ -159,7 +172,7 @@ pub struct Set {
 struct TriggersWithContext<'s, F> {
     /// Triggers being serialized
     triggers: &'s IndexMap<TriggerId, LoadedAction<F>>,
-    /// Containing Set, used for looking up origignal [`WasmSmartContract`]s
+    /// Containing Set, used for looking up original [`WasmSmartContract`]s
     /// during serialization.
     set: &'s Set,
 }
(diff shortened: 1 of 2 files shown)


DCNick3:
So, we need to implement reference-counting for when the same wasm smart contract is subscribed to several triggers. Subtle, huh

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -633,52 +654,96 @@ impl Set {
...
+            } else {
+                original_contracts.remove(&blob_hash)?;
+            }
+            Some(())
+        })()
+        .expect("`Set::original_contracts` doesn't contain required hash. This is a bug");
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    fn extract_blob_hash<F>(action: LoadedAction<F>) -> Option<HashOf<WasmSmartContract>> {
```
DCNick3, line 740: Can this be a method on `LoadedAction`?

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -633,52 +654,96 @@ impl Set {
...
+    fn remove_original_trigger(
+        original_contracts: &mut IndexMap<
+            HashOf<WasmSmartContract>,
+            (WasmSmartContract, NonZeroU64),
+        >,
+        blob_hash: HashOf<WasmSmartContract>,
+    ) {
+        (|| {
+            let count = &mut original_contracts.get_mut(&blob_hash)?.1;
+            if let Some(new_count) = NonZeroU64::new(count.get() - 1) {
```
DCNick3, line 729: Weird that NonZero* types do not have `saturating_sub`, but there is some argument against them: https://internals.rust-lang.org/t/checked-sub-and-saturating-sub-for-nonzero-unsigned-integers/19898/7

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -633,52 +654,96 @@ impl Set {
...
+    /// Panics if `blob_hash` is not in the [`Set::original_contracts`].
+    fn remove_original_trigger(
+        original_contracts: &mut IndexMap<
+            HashOf<WasmSmartContract>,
+            (WasmSmartContract, NonZeroU64),
+        >,
+        blob_hash: HashOf<WasmSmartContract>,
+    ) {
+        (|| {
+            let count = &mut original_contracts.get_mut(&blob_hash)?.1;
```
DCNick3, line 728: I think using `.entry` might be nicer here? By pattern matching on it, you would have only one case to raise the "`Set::original_contracts` doesn't contain required hash" panic without having to use 

Something like:

```rust
match original_contracts.entry(blob_hash) {
  Occupied(o) => {
    let (_, count) = o.get(); // I think get_mut won't work here without polonius
    if ... {
      *o.get_mut().1 = new_count;
    } else {
      o.remove_entry();
    }
  }
  Vacant(_) => panic!("..."),
}
```

core/src/smartcontracts/isi/triggers/set.rs
Erigara: Do we need `NonZero` here in the first place? It would save us some space in case of `Option<NonZero>`.
But in this case i don't see many benefits in using it.

DCNick3: That's... Also fair

While it is true that we have an invariant that the u64 must not be zero, `NonZeroU64`'s hit to ergonomics kinda makes it not worth it to enforce the invariant at the type system level..

Erigara:
> So, we need to implement reference-counting for when the same wasm smart contract is subscribed to several triggers. Subtle, huh

Afaik trigger id should be unique so instead of mapping (hash, (wasm,counter)) we can have (id, wasm).
It would have redundant entities, but code should be a bit more clear i guess

core/src/smartcontracts/isi/triggers/set.rs
Arjentix: I think it's more robust than just an integer type, because it introduces invariants which are useful in this scenario. Count should either be > 0 or doesn't exist

Arjentix: It's also a documentation for future developers

Arjentix: > NonZeroU64's hit to ergonomics kinda makes it not worth it to enforce the invariant at the type system level..

I don't think having just an unsigned integer type would help much, because I will still need to check for zero case

DCNick3: I just don't like the `NonZeroU64::new(v.get() - 1)` dance, but no strong feelings, it can stay.

core/src/smartcontracts/isi/triggers/set.rs
Arjentix: You're right. Currently I'm doing the search twice. Don't know how I missed that

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -633,52 +663,92 @@ impl Set {
...
+    /// or remove it if the counter reaches zero.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `blob_hash` is not in the [`Set::original_contracts`].
+    fn remove_original_trigger(
+        original_contracts: &mut IndexMap<
+            HashOf<WasmSmartContract>,
+            (WasmSmartContract, NonZeroU64),
+        >,
```
Erigara, line 733: Can we make type alias for this type? 

```rust
type WasmSmartContractMap = IndexMap<HashOf<WasmSmartContract>,(WasmSmartContract, NonZeroU64)>;
```

Erigara (approved):

DCNick3 (approved):

PR #4163, opened 2023-12-21
[feature] #3453: Implement Store set, remove, get operations in Client CLI

### LINKED ISSUE #3453: [suggestion] Add ways to manipulate a Store asset thorough `client_cli`
### Feature request

During a community discussion yesterday, Vince [wrote](https://discord.com/channels/905194001349627914/905205848547155968/1103694129961119764):

> Is there a way to deal with the `Store` asset (`set`, `update`, `read`, `remove`) with the binary client `iroha_client_cli`?

We intend to expose this feature through FFI later, so I'm adding it as an issue.

Asem-Abdelhady:
## Description

I added extra subcommands to asset which are set-store, get-store, remove-store, update-store

## Use Casees
```bash
./iroha_client_cliasset set-key-value --asset-id="tea##mad_hatter@looking_glass" --key="K5" --value='{ "Vec": [{"String": "a"}, {"String": "b"}] }'
---
./iroha_client_cli asset get-key-value --asset-id="tea##mad_hatter@looking_glass" --key="K5"   
---
./iroha_client_cli asset remove-key-value --asset-id="tea##mad_hatter@looking_glass" --key="K5"   
```

### Linked issue

Closes #3453

### Checklist

- [X] All applicable CI checks pass (or I promised to make them pass later)
- [X] (optional) I've written unit tests for the code changes
- [X] I replied to all comments after code review, marking all implemented changes with thumbs up


Erigara:
What is the difference between `set-store` and `update-store`?

```
./iroha_client_cli asset set-store --key="K4" --asset-id="tea##mad_hatter@looking_glass" --value=true
---
./iroha_client_cli asset update-store --key="K4" --asset-id="tea##mad_hatter@looking_glass" --value=16_u32
```

Erigara:
We don't use merge commits in our workflow, use rebase workflow instead (checkout section in [CONTRIBUTING.md](https://github.com/hyperledger/iroha/blob/iroha2-dev/CONTRIBUTING.md#pull-request-etiquette)).

You can squash your commits locally after review before merge.

data_model/src/asset.rs
```diff
@@ -482,7 +482,7 @@ impl FromStr for AssetId {
         if let Some((asset_definition_candidate, account_id_candidate)) = string.rsplit_once('#') {
             let account_id: AccountId = account_id_candidate.parse()
                 .map_err(|_err| ParseError {
-                    reason: "Failed to parse the `account_id` part of the `asset_id`. Please ensure that it has the form `account@domain`"
+                    reason: "Failed to parse the `account_id` part of the `asset_id`. Please ensure that it has the form `asset##account@domain`"
```
Erigara, line 485: I don't think this change is valid since it talks about `account_id` part of `asset_id`:
`asset##[accound@account_domain]`, `asset#asset_domain#[account@account_domain]`.

client_cli/src/main.rs
```diff
@@ -892,6 +906,130 @@ mod asset {
...
+        pub key: Name,
+        /// The new value to be associated with the key
+        #[structopt(long)]
+        pub value: String,
+        /// Interpret the `value` as a JSON string
+        #[structopt(long = "json", requires = "value")]
+        pub is_json: bool,
+    }
+
+    impl RunArgs for UpdateStore {
```
Erigara, line 998: Imo update is redundant (get, set, del is enough), but we can wait for other opinions.

client_cli/src/main.rs
```diff
@@ -892,6 +906,130 @@ mod asset {
...
+        let asset = client
+            .request(asset::by_id(asset_id))
+            .wrap_err("Failed to get store asset")?;
+        let get_json = json!(&asset);
+        let value = &get_json["value"]["Store"][key.to_string()];
+        match value {
+            serde_json::Value::Null => Err(eyre!("Key '{}' not found in the store.", key)),
+            _ => Ok(value.clone()),
+        }
+    }
```
Erigara, line 1032: You can use `FindAssetKeyValueByIdAndKey` query to get value directly

client_cli/src/main.rs
```diff
@@ -892,6 +906,130 @@ mod asset {
...
+    pub struct SetStore {
+        /// AssetId for the Store asset (in form of `asset##account@domain_name')
+        #[structopt(long = "asset-id")]
+        pub asset_id: AssetId,
+        /// The key for the store value
+        #[structopt(long)]
+        pub key: Name,
+        /// The value to be associated with the key
+        #[structopt(long)]
+        pub value: String,
```
Erigara, line 920: I suggest to use a bit different approach. `clap` can accept anything implementing `FromStr`.

I would suggest to create `ValueArg` (better name probably) which would implement parsing logic for `Value`.
This way we could have parsing shortcuts for some types and fallback to parsing json without json flag and without changing data model.

```rust 

struct ValueArg(Value);

impl FromStr for ValueArg {
    type Err = ...;

        // parse as bool
        
        // parse as ip

        // ..

        // if nothing worked parse as json
    }
}
```

And instead of String you can parse to `ValueArg` directly.

client_cli/src/main.rs
0x009922: And I think _should_, because this query is designed specifically for that. Store assets might have a lot of entries.

client_cli/src/main.rs
```diff
@@ -892,6 +906,130 @@ mod asset {
...
+        /// Interpret the `value` as a JSON string
+        #[structopt(long = "json", requires = "value")]
+        pub is_json: bool,
```
0x009922, line 923: Instead of specifying `--json` flag to control how `--value` should be parsed, I would suggest to use conflicting arguments `--value=<string>` and `--value-json=<json>`.

client_cli/src/main.rs
0x009922: I agree. There are two mutations: set and remove. CLI should reflect the API of Iroha as close as possible, imo.

data_model/src/lib.rs
```diff
@@ -1099,7 +1100,32 @@ impl fmt::Display for Value {
...
+
+        if let Ok(num_val) = NumericValue::from_str(v) {
+            return Ok(Value::Numeric(num_val));
+        }
+
+        if let Ok(pub_val) = PublicKey::from_str(v) {
+            return Ok(Value::PublicKey(pub_val));
+        }
+
+        return Ok(Value::String(v.to_string()));
```
0x009922, line 1126: Perhaps, a chain of `Result::or_else` might look less noisy

client_cli/src/main.rs
```diff
@@ -683,13 +689,21 @@ mod asset {
...
         #[clap(subcommand)]
         List(List),
+        /// Put a key-value in a store Asset defintion
+        SetStore(SetStore),
+        /// Remove a key-value via key in a store Asset definition
+        RemoveStore(RemoveStore),
+        /// Get a value from store Asset definition via key
+        GetStore(GetStore),
+        /// Update a value from store Asset definition via key
+        UpdateStore(UpdateStore),
```
0x009922, line 699: I think the following naming would reflect the API of Iroha more precise:

- `set-key-value` (with underlying `SetKeyValue` ISI)
- `get-key-value` (with underlying `FindAssetKeyValueByIdAndKey` query)
- `remove-key-value` (with underlying `RemoveKeyValue` ISI)

0x009922 (changes requested):
Note: https://github.com/hyperledger/iroha/pull/4153 migrates clap to v4 in `iroha_client_cli`.

0x009922:
While reviewing this, found an issue with `asset` subcommand in general:

- https://github.com/hyperledger/iroha/issues/4167

data_model/src/asset.rs
Asem-Abdelhady: I remember that I had to change it becuse I had a very misleading error. I will see if I could replicate it if not I will revert it back

Erigara (approved):
Nice job, lgtm

Erigara:
@Asem-Abdelhady some python tests should be added to cover new functionality

client_cli/src/main.rs
```diff
@@ -55,6 +55,24 @@ impl MetadataArgs {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        s.parse::<bool>()
+            .map(Value::Bool)
+            .or_else(|_| s.parse::<Ipv4Addr>().map(Value::Ipv4Addr))
+            .or_else(|_| s.parse::<Ipv6Addr>().map(Value::Ipv6Addr))
+            .or_else(|_| s.parse::<NumericValue>().map(Value::Numeric))
+            .or_else(|_| s.parse::<PublicKey>().map(Value::PublicKey))
+            .or_else(|_| serde_json::from_str::<Value>(s).map_err(|e| e.into()))
+            .map(ValueArg)
+    }
```
0x009922, line 74: Please cover with unit tests, so that we can also see how actual values look like.

client_cli/src/main.rs
```diff
@@ -888,6 +915,74 @@ mod asset {
...
+    pub struct SetKeyValue {
+        /// AssetId for the Store asset (in form of `asset##account@domain_name')
+        #[clap(long = "asset-id")]
+        pub asset_id: AssetId,
+        /// The key for the store value
+        #[clap(long)]
+        pub key: Name,
+        /// The value to be associated with the key
+        #[clap(long)]
+        pub value: ValueArg,
```
0x009922, line 929: This argument should be extensively documented, so that when users see it, they can understand how to pass different values, how they will be interpreted.

client_cli/src/main.rs
```diff
@@ -683,13 +704,19 @@ mod asset {
         /// List assets
         #[clap(subcommand)]
         List(List),
+        /// Put a key-value in a store Asset defintion
+        SetKeyValue(SetKeyValue),
+        /// Remove a key-value via key in a store Asset definition
+        RemoveKeyValue(RemoveKeyValue),
+        /// Get a value from store Asset definition via key
+        GetKeyValue(GetKeyValue),
```
0x009922, line 712: I think there commands work with actual assets, not their definitions.

We should ask tech writers (@outoftardis, @yamkovoy) how to document these commands better.

Maybe,

> Create a key-value entry in a store asset

or

> Remove Store asset entry

?

client_cli/src/main.rs
```diff
@@ -888,6 +915,74 @@ mod asset {
...
+
+            let set = iroha_client::data_model::isi::SetKeyValue::asset(asset_id, key, value);
+            submit([set], UnlimitedMetadata::default(), context)?;
+            Ok(())
+        }
+    }
+    #[derive(clap::Args, Debug)]
+    pub struct RemoveKeyValue {
+        /// AssetId for the Store asset (in form of `asset##account@domain_name')
+        #[clap(long = "asset-id")]
```
0x009922, line 948:
```suggestion
```

Clap automatically transforms `asset_id` into `--asset-id` in long form.

data_model/src/asset.rs
```diff
@@ -494,7 +494,7 @@ impl FromStr for AssetId {
                                       })?,
                                       account_id.domain_id.clone())
                 } else {
-                    return Err(ParseError { reason: "The `definition_id` part of the `asset_id` failed to parse. Ensure that you have it in the right format: `name#domain_of_asset#account_name@domain_of_account`." });
+                    return Err(ParseError { reason: "The `definition_id` part of the `asset_id` failed to parse. Ensure that you have it in the right format: `name#domain_of_asset#account_name@domain_of_account` or `name##account_name@domain_of_account` in case of same domain" });
```
0x009922, line 497:
```suggestion
                    return Err(ParseError { reason: "The `definition_id` part of the `asset_id` failed to parse. Ensure that you have it in the right format: `name#domain_of_asset#account_name@domain_of_account`, or `name##account_name@domain_of_account` in case of the same domain" });
```

Could you also add a unit test that checks this error message?

client_cli/src/main.rs
Asem-Abdelhady: nvbnvb

0x009922: I think PRs separation makes sense when other changes are large and not very related. Here, adding unit tests are directly related to the feature you are implementing and it is a good manner (imo) to accompany new feature with its tests.

client_cli/src/main.rs
outoftardis: using "Store asset" instead of "store asset definition" should be fine

client_cli/src/main.rs
```diff
@@ -1118,334 +1118,91 @@ mod json {
...
-        config: ClientConfiguration,
-        skip_mst_check: bool,
+    #[test]
+    fn parse_ipv4_address() {
+        match ValueArg::from_str("192.168.0.1") {
+            Ok(ValueArg(Value::Ipv4Addr(addr))) => {
+                assert_eq!(addr, Ipv4Addr::new([192, 168, 0, 1]))
+            }
+            _ => panic!("Expected Ipv4Addr parsing to succeed."),
+        }
```
0x009922, line 1136: Could be written in a flat way, with less code and with detailed error message:

```rs
let actual = ValueArg::from_str("192.168.0.1").expect("Should not fail with valid IP input");
assert_eq!(actual, ValueArg(Value::Ipv4Addr(Ipv4Addr::new([192, 168, 0, 1]))))
```

Also consider [`assert_matches!`](https://docs.rs/assert_matches/latest/assert_matches/macro.assert_matches.html) (not an `std`'s unstable one).

client_cli/src/main.rs
```diff
@@ -888,6 +915,77 @@ mod asset {
             Ok(())
         }
     }
+
+    #[derive(clap::Args, Debug)]
+    pub struct SetKeyValue {
+        /// AssetId for the Store asset (in form of `asset##account@domain_name')
+        #[clap(long = "asset-id")]
```
Erigara, line 922: The same comment about automatic transform from Dima (we don't need this line).

client_cli/src/main.rs
```diff
@@ -1023,3 +1118,80 @@ mod json {
...
+            actual,
+            ValueArg(Value::PublicKey(
+                PublicKey::from_str(public_key_str).unwrap()
+            ))
+        );
+    }
+
+    #[test]
+    fn parse_json_value() {
+        let json_str = r#"{"Vec":[{"String":"a"},{"String":"b"}]}"#;
```
0x009922, line 1183: You might also consider using `serde_json::json!` macro:

```rs
json!({"Vec": [{"String": "a"}, {"String": "b"}]})
```

data_model/src/asset.rs
```diff
@@ -572,3 +572,13 @@ pub mod prelude {
...
     };
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn test_error_for_asset_id() {
+        let _invalid_asset_id = AssetId::from_str("store#alice@wonderland")
+            .expect_err("Expected an error for assetId, but the operation succeeded");
```
0x009922, line 582:
```suggestion
            .expect_err("store#alice@wonderland should not be a valid AssetId");
```

client_cli/src/main.rs
```diff
@@ -888,6 +915,74 @@ mod asset {
...
+
+            let set = iroha_client::data_model::isi::SetKeyValue::asset(asset_id, key, value);
+            submit([set], UnlimitedMetadata::default(), context)?;
+            Ok(())
+        }
+    }
+    #[derive(clap::Args, Debug)]
+    pub struct RemoveKeyValue {
+        /// AssetId for the Store asset (in form of `asset##account@domain_name')
+        pub asset_id: AssetId,
```
0x009922, line 950: Without `#[clap(long)]` `asset_id` becomes a positional argument:

```sh
iroha_client_cli asset remove-key-value <asset-id> --key <key>
```

client_cli/src/main.rs
```diff
@@ -1023,3 +1118,80 @@ mod json {
...
+        let actual_json: JsonValue = serde_json::to_value(actual_value).unwrap();
+        let expected_json: JsonValue = serde_json::from_str(json_str).unwrap();
+        assert_eq!(actual_json, expected_json);
+    }
+
+    #[test]
+    fn parse_invalid_value() {
+        let invalid_str = "not_a_valid_value";
+        let _invalid_value = ValueArg::from_str(invalid_str)
+            .expect_err("Should fail invalid type from string but passed");
```
0x009922, line 1195: Would be good to see an assertion of the error message here.

client_cli/src/main.rs
```diff
@@ -1023,3 +1118,80 @@ mod json {
...
+    }
+
+    #[test]
+    fn parse_numeric_value() {
+        let actual = ValueArg::from_str("123_u32").expect("Failed to parse valid numeric value");
+        assert!(matches!(
+            actual,
+            ValueArg(Value::Numeric(NumericValue::U32(123)))
+        ));
+    }
```
0x009922, line 1166: What about other supported suffixes?

client_cli/src/main.rs
```diff
@@ -1023,3 +1118,80 @@ mod json {
...
+    use super::*;
+
+    #[test]
+    fn parse_ipv4_address() {
+        let actual = ValueArg::from_str("192.168.0.1").expect("Failed to parse valid IPv4 address");
+        assert_eq!(
+            actual,
+            ValueArg(Value::Ipv4Addr(Ipv4Addr::new([192, 168, 0, 1])))
+        );
+    }
```
0x009922, line 1136: I would suggest to organise all "success" cases into a single test with reduced repetition using a macro. For example:

```rust
#[test]
fn parse_value_arg_cases() {
    macro_rules! case {
        ($input:expr, $expected:expr) => {
            let ValueArg(actual) =
                ValueArg::from_str($input).expect("should not fail with valid input");
            assert_eq!(actual, $expected);
        };
    }

    case!(
        "192.168.0.1",
        Value::Ipv4Addr(Ipv4Addr::new([192, 168, 0, 1]))
    );
    case!("true", Value::Bool(true));
    case!("false", Value::Bool(false));
    // ...
}
```

This way, it is easy to write many-many cases and see them all in a single place, which helps a lot for review and future maintenance.

You wrote many test cases already, but they are not really comprehensive, e.g. it is not obvious what numeric suffixes are supported.

client_cli/src/main.rs
```diff
@@ -888,6 +915,74 @@ mod asset {
...
+    pub struct SetKeyValue {
+        /// AssetId for the Store asset (in form of `asset##account@domain_name')
+        pub asset_id: AssetId,
+        /// The key for the store value
+        #[clap(long)]
+        pub key: Name,
+        /// Enter the value to be associated with the specified key.
+        /// This value can be of various types, including boolean(true),
+        /// numeric(1_u32), IP address(127.0.0.1), public key(ed01207233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0)
+        /// or JSON where the key is the type({"Vec":[{"String":"a"},{"String":"b"}]}).
```
0x009922, line 929: I think this documentation might be improved. Here is my try:

> The value to be associated with the specified key.
> 
> The following types are supported:
> 
> - Numbers: with a suffix, e.g. `42_u32` or `1000_u128`
> - Booleans: `false`/`true`
> - IPv4/IPv6: e.g. `127.0.0.1`, `::1`
> - Iroha Public Key Multihash: e.g. `ed01207233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0`
> - JSON: e.g. `{"Vec":[{"String":"a"},{"String":"b"}]}`

client_cli/src/main.rs
```diff
@@ -683,13 +704,19 @@ mod asset {
         /// List assets
         #[clap(subcommand)]
         List(List),
+        /// Put a key-value entry in a store Asset
```
0x009922, line 707:
```suggestion
        /// Set a key-value entry in a Store asset
```

client_cli/src/main.rs
```diff
@@ -683,13 +704,19 @@ mod asset {
...
+        SetKeyValue(SetKeyValue),
+        /// Remove a key-value entry via key in a store Asset
```
0x009922, line 709:
```suggestion
        /// Remove a key-value entry from a Store asset
```

client_cli/src/main.rs
```diff
@@ -683,13 +704,19 @@ mod asset {
...
+        RemoveKeyValue(RemoveKeyValue),
+        /// Get a value from store Asset definition via key
```
0x009922, line 711:
```suggestion
        /// Get a value from a Store asset
```

0x009922 (changes requested):
Requesting changes again primarily because of complete removal of `#[arg(long)]` for `asset_id` argument, making it a positional argument. It is inconsistent with the rest of the CLI.

client_cli/src/main.rs
Asem-Abdelhady: I think we should leave it as it is to be able to parse it using from_str we implemented. Using this aproach we would have to parse the object into str after we create it then test from_str.

client_cli/src/main.rs
Asem-Abdelhady: the user might enter a value that is not supported so we make sure that this results to an error

client_cli/src/main.rs
appetrosyan: In principle the following are acceptable: 
### Both positional;
```
iroha_client_cli asset remove-key-value <key> <asset-id>
```

### Both `--long` and `-s`hort for both arguments. 
```
iroha_client_cli asset remove-key-value -k <key> -a <asset-id>
```
It should be obvious and probably not much of a head scratcher, plus it only takes a few lines of code. 

### Asset as positional to asset:

```
iroha_client_cli asset <asset-id> remove-key-value <key>
```

Imo this makes more sense but might require reworking how the client_cli works.

client_cli/src/main.rs
appetrosyan: What @0x009922 is suggesting is to also verify that the error message is what you'd expect. 

Personally I would probably not do it, because it unnecessarily couples an irrelevant part of the test to your particular error message, so that if anyone wants to change the error message, they will also need to change the test. Instead consider

```suggestion
        ValueArg::from_str(invalid_str).unwrap();
```

and mark the test as `should_panic`.

data_model/src/asset.rs
appetrosyan: @Asem-Abdelhady Make a good point of only resolving conversations after you've 
1. Addressed the issue as requested (there's still a missing `the`). 
2. Explained why you can't resolve the issue as requested, and provided an alternative (in this you could say, `in case of same domain` is grammatically correct). 
3. Promised to do it later: by opening an issue, assigning it to yourself and providing an estimate of when you'd be able to do it.

0x009922 (approved):

Erigara (approved):

PR #4165, opened 2023-12-21
[fix] #4164: Fix topology update on restart

### LINKED ISSUE #4164: [BUG]: Iroha failed to restart with block view change index > 0
Iroha failed to restart with block view change index > 0 because previously it was assumed to always have block with view change index 0.

Erigara:
## Description

Update current topology with block view change index.

### Linked issue

Closes #4164 

### Benefits

Peer restarts correctly.

### How to test 

Use block store with where blocks have view change index > 0 and try to restart peer.

### DIFF
--- a/core/src/sumeragi/mod.rs
+++ b/core/src/sumeragi/mod.rs
@@ -229,11 +229,14 @@ impl SumeragiHandle {
     fn replay_block(
         block: &SignedBlock,
         wsv: &mut WorldStateView,
-        current_topology: &Topology,
+        mut current_topology: Topology,
     ) -> Topology {
-        let block = ValidBlock::validate(block.clone(), current_topology, wsv)
+        // NOTE: topology need to be updated up to block's view_change_index
+        current_topology.rotate_all_n(block.payload().header.view_change_index);
+
+        let block = ValidBlock::validate(block.clone(), &current_topology, wsv)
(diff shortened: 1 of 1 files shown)


core/src/sumeragi/mod.rs
```diff
@@ -229,11 +229,14 @@ impl SumeragiHandle {
     fn replay_block(
         block: &SignedBlock,
         wsv: &mut WorldStateView,
-        current_topology: &Topology,
+        mut current_topology: Topology,
     ) -> Topology {
-        let block = ValidBlock::validate(block.clone(), current_topology, wsv)
+        // NOTE: topology need to be updated up to block's view_change_index
+        current_topology.rotate_all_n(block.payload().header.view_change_index);
```
VAmuzing, line 235: Could you further explain why do we need to rotate peers here if we're doing it in `Topology::recreate_topology`?
Is it the only case where block validation would fail because of mismatched topology?

Erigara: I need this because `Topology::recreate_topology` always pass 0 as `view_change_index`. 
We need look ahead to the next block to be able get `view_change_index`.

VAmuzing (approved):

6r1d (approved):

Issue #3068, opened 2023-01-19
[suggestion] DSL for event filters


appetrosyan:
### Feature request

I propose designing a small Domain-specific language implemented as a procedural macro crate that generates code that produces the following:

Examples
```rust
let asset_filter: FilterBox = filter!(asset "rose##alice@wonderland" added);
let passthrough: FilterBox = filter!(*);
let domain_metadata: FilterBox = filter!(domain metadata["key"])
let account = filter!(asset "rose##alice@wonderland" metadata_inserted);
```

### Motivation

Composing event filters is extremely cumbersome, because the entire hierarchy of the nets must be remembered by the person writing the filter. 

### Who can help?

@Arjentix

DCNick3:
I've decided to make the DSL not macro-based, but based on builder pattern. For this I would also like to refactor the events and event filters first to be more orthogonal

DCNick3:
This is mostly addressed with the merge of #4240. Some improvements (#4259, #4356) will follow

PR #4170, opened 2023-12-25
[refactor] #3068: remove flattened events

DCNick3:
## Description

This PR removes duplicated flattened events. Those are events like the `DataEvent::Account(_)` and `DataEvent::Asset(_)`. They are already present under other event types (`DataEvent::Domain(DomainEvent::Account(_))`, for example) and were duplicated to simplify the event filter construction (?).

While an immediate effect of this change is ever more increased amount of boilerplate to construct event filters, this provides a more orthogonal events system, helping to build a comprehensive DSL.

### Linked issue

Partially addresses #3068 

### Checklist

- [x] make ci pass


tools/parity_scale_decoder/build.rs
```diff
@@ -23,6 +23,8 @@ where
     path_to.push("samples/");
     path_to.push(filename);
 
+    println!("cargo:rerun-if-changed={}", path_to.to_str().unwrap());
```
Arjentix, line 26: What is the idea behind that? `path_to` will have no file extension here, so no file should be matched

DCNick3: Oh, that's definitely a mistake on my part!

It was supposed to be a path to the json file. I had to edit one in this PR, but the `build.rs` was not getting re-ran so I figured I would do a drive-by fix.

DCNick3: should be fixed now

Erigara (approved):

Arjentix (approved):

PR #4171, opened 2023-12-25
[fix] #4155: ensure the secp256k1 signatures coming out of OpenSSL are normalized

### LINKED ISSUE #4155: Investigate why `signature::secp256k1::test::secp256k1_sign` test from `iroha_crypto` is flaky
Sometimes the signature verification is failing, we should investigate why

DCNick3:
## Description

This is an interoperability problem between OpenSSL and newer cryptographic libraries, see [BIP-0062](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#user-content-Low_S_values_in_signatures) for details

This fixes the flaky secp256k1_sign test

Also includes a drive-by cleanup of useless `assert(result.is_ok())`

### Linked issue

Closes #4155

### Benefits

Fixes the flaky test! Less flaky tests is obviously better 

### Checklist

- [ ] make ci pass

### DIFF
--- a/crypto/src/signature/secp256k1.rs
+++ b/crypto/src/signature/secp256k1.rs
@@ -148,8 +148,8 @@ mod test {
     #[test]
     fn secp256k1_load_keys() {
         let secret = PrivateKey::from_hex(ALGORITHM, PRIVATE_KEY).unwrap();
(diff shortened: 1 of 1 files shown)


Erigara (approved):

0x009922 (approved):
Wouldn't say I 100% understand the issue and that the check is correct, but it seems so. After all, it is a patch of tests which now have more verifies, thus LGTM.

Issue #4174, opened 2023-12-25
[suggestion] Stop encoding X25519 keys as Ed25519


Arjentix:
While looks like it works fine I suggest refactoring that part to make it more clear.

Currently two completely different types are casted between each other in completely implicit and silent way. I suggest to make it clear that X25519 is used only for exchanging purposes.

mversic:
also consider [this](https://github.com/hyperledger/iroha/pull/4181#discussion_r1440276612) comment when handling this issue

Issue #3714, opened 2023-07-18
query validation in no_std


mversic:
Make the `cfg(feature)` external. A no-op verify is dangerous. If we can't do it on `std` it's better to mask the entire function.

_Originally posted by @appetrosyan in https://github.com/hyperledger/iroha/pull/3621#discussion_r1235605832_

mversic:
maybe the situation is now different with Ursa removed from iroha. Hopefully, we can make signature verification work with `no_std`

PR #4181, opened 2023-12-28
[refactor] #4045: Use concrete key types

### LINKED ISSUE #4045: [suggestion]: Use concrete key types in iroha_crypto

### LINKED ISSUE #4199: Consider signing by KeyPair reference not by value
Related fixme: https://github.com/hyperledger/iroha/blob/a3fad9b02b64cbc67e99f0fbe19a4802dda841c8/genesis/src/lib.rs#L63-L69

Also see this discussion: https://github.com/hyperledger/iroha/pull/4181#discussion_r1445734786

Arjentix:
## Description

- `PublicKey` and `PrivateKey` are now enums with explicit key types. No invalid state is possible
- Fixed some tests
- Fixed now useless errors signatures of depended code
- `amcl_wrapper` replaced with `w3f-bls`
- `iroha_crypto` is now compatible with `no-std` and WASM
- Fixed some clippy lints

### Linked issue

- Closes #4045
- Closes #4199 
- Addresses #3714
- Openned #4174 
- Openned #4241 

### Benefits

- More robust solution because now we are sure that stored keys are valid
- No more errors are possible when creating keys or signing
- Test are more real now

### Checklist


crypto/src/lib.rs
```diff
@@ -384,52 +506,69 @@ impl PublicKey {
...
 
 #[ffi_impl_opaque]
 impl PrivateKey {
     /// Key payload
-    // TODO: Derive with getset once FFI impl is fixed
-    pub fn payload(&self) -> &[u8] {
-        self.payload.as_ref()
+    pub fn payload(&self) -> ConstVec<u8> {
+        match self {
+            Self::Ed25519(key) => key.to_keypair_bytes().to_vec().into(),
```
mversic, line 545:
```suggestion
            Self::Ed25519(key) => key.to_keypair_bytes().into(),
```

we should have a shortcut like this instead of creating intermediate `Vec`

crypto/src/lib.rs
```diff
@@ -278,58 +310,52 @@ impl From<KeyPair> for (PublicKey, PrivateKey) {
...
-            Algorithm::Ed25519 => signature::ed25519::Ed25519Sha512::keypair(key_gen_option),
-            Algorithm::Secp256k1 => {
-                signature::secp256k1::EcdsaSecp256k1Sha256::keypair(key_gen_option)
-            }
-            Algorithm::BlsNormal => signature::bls::BlsNormal::keypair(key_gen_option),
-            Algorithm::BlsSmall => signature::bls::BlsSmall::keypair(key_gen_option),
-        }?;
-
-        Ok(public_key)
+    pub fn payload(&self) -> ConstVec<u8> {
```
mversic, line 349:
```suggestion
    pub fn payload(&self) -> &[u8] {
```

is this not possible? otherwise we don't need this method, `into_raw` is sufficient

mversic (changes requested):

core/src/sumeragi/network_topology.rs
```diff
@@ -92,19 +92,19 @@ impl Topology {
         for role in roles {
             match (role, self.is_non_empty(), self.is_consensus_required()) {
                 (Role::Leader, Some(topology), _) => {
-                    public_keys.insert(&topology.leader().public_key);
+                    public_keys.insert(topology.leader().public_key());
```
mversic, line 95: how is this change relevant? in `core` we should prefer direct field access

mversic:
> Currently it's impossible to build Executor or any wasm, because used crypto libraries are feature-gated, meaning that even `PublicKey` and `PrivateKey` cannot be represented as is in WASM.
> 
> Solutions I see:
> 
>     * Let WASM have all crypto code and remove feature-gate. Unused code will likely be optimized by Rust, but I'm not sure.
> 
>     * Have another `PublicKey` and `PrivateKey` for WASM types which will be compatible with types from Host in terms of encoding and serialization
> 
> 
> For any of this solutions I would like to hear other opiniont, especially from @mversic .

I would try to get `PublicKey`/`PrivateKey` work in `no-std`. If that is not possible, then we can try using `std` in `wasm`. The last thing is making FFI calls

genesis/src/lib.rs
```diff
@@ -58,17 +58,15 @@ impl GenesisNetwork {
...
 
         let transactions = transactions_iter
-            .enumerate()
-            .map(|(i, raw_transaction)| {
+            .map(|raw_transaction| {
                 raw_transaction
                     // FIXME: fix underlying chain of `.sign` so that it doesn't
                     //        consume the key pair unnecessarily. It might be costly to clone
                     //        the key pair for a large genesis.
                     .sign(genesis_key_pair.clone())
```
0x009922, line 67: This FIXME is related exactly to the change you made in this PR. I guess it should be removed.

data_model/src/query/mod.rs
```diff
@@ -1352,7 +1348,7 @@ pub mod error {
             /// Parameter with id `{0}` not found
             Parameter(ParameterId),
             /// Failed to find public key: `{0}`
-            PublicKey(PublicKey),
+            PublicKey(Box<PublicKey>),
```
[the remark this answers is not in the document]
DCNick3: I don't think it's recursion, but:

Before this change `PublicKey` stored a vec of bytes comprising the key, so its structure size wasn't that large. Now, however, it contains concrete key types with known sizes and so it stores the keys inline, leading to a larger structure size. 

I think it might even be enough to trigger clippy's [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/index.html#/large_enum_variant) lint?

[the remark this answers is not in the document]
DCNick3: Btw, I am not sure if it's a good idea to actually store they keys inline. It will definitely be a problem for using [zeroize](https://docs.rs/zeroize/latest/zeroize/) (which is needed for #3240), as it is not really possible to ensure a by-value struct gets deleted.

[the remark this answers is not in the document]
mversic: you make a point with `zeroize`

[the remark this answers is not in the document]
mversic: but `PublicKey` is not a secret and it doesn't have to be zeroed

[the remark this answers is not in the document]
Arjentix: Yes, it's not recursion, it's large variant size. `PublicKey` takes a lot of space because of concrete crypto types which holds additional info

crypto/src/lib.rs
```diff
@@ -118,6 +118,7 @@ impl FromStr for Algorithm {
     allow(unused_tuple_struct_fields)
 )]
 #[derive(Debug, Clone)]
+#[allow(variant_size_differences, clippy::large_enum_variant)]
```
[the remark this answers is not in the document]
Arjentix: Indeed this line is useless now

crypto/src/lib.rs
```diff
@@ -340,17 +366,113 @@ impl PublicKey {
...
+    }
+}
+
+impl Ord for PublicKey {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        (self.digest_function(), self.payload()).cmp(&(other.digest_function(), other.payload()))
+    }
+}
+
+#[cfg(not(feature = "ffi_import"))]
```
DCNick3, line 418: Is there an actual need to manually implement the `Encode`/`Decode`? Or did you want to keep the scale codec-level compatibility with the older untyped keys?

crypto/src/lib.rs
```diff
@@ -340,17 +366,113 @@ impl PublicKey {
...
+impl Encode for PublicKey {
+    fn size_hint(&self) -> usize {
+        self.payload().len() + 1
```
DCNick3, line 421: I don't think this size hint is right, the `ConstVec` will be prefixed with length, so it'll always end up at least one byte longer

crypto/src/lib.rs
```diff
@@ -340,17 +366,113 @@ impl PublicKey {
...
+        Self::from_raw(digest_function, &payload).map_err(|_| {
+            parity_scale_codec::Error::from(
+                "Failed to construct public key from digest function and payload",
+            )
         })
     }
 }
 
+#[cfg(not(feature = "ffi_import"))]
+impl IntoSchema for PublicKey {
```
DCNick3, line 446: Same thing, do we want to lie about our actual implementation and just expose the old structure, or are we fine breaking the API here?

crypto/src/signature/bls/mod.rs
```diff
@@ -48,12 +53,19 @@ mod normal {
         fn set_pairs((g1, g2): &(Self::Generator, Self::SignatureGroup)) -> (&G1, &G2) {
             (g1, g2)
         }
+
+        fn extract_private_key(private_key: &crate::PrivateKey) -> Option<&super::PrivateKey> {
```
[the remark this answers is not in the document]
DCNick3: Used in generic context: https://github.com/hyperledger/iroha/blob/b2abf54792c2a2a1cdd5a17c89b40873a71bac78/crypto/src/signature/bls/implementation.rs#L201

crypto/src/signature/ed25519.rs
```diff
@@ -126,7 +117,7 @@ mod test {
                 signature.as_slice().as_ptr(),
                 MESSAGE_1.as_ptr(),
                 MESSAGE_1.len() as u64,
-                p.payload().as_ptr(),
+                p.to_bytes().as_ptr(),
```
DCNick3, line 120: Will the lifetimes be all right here?
It seems that this will pass a pointer to a temporary value that will get destroyed before the function is called (I don't think there's lifetime extension magic like in C++ for cases like these?)

crypto/src/signature/ed25519.rs
```diff
@@ -154,11 +144,10 @@ mod test {
                 std::ptr::null_mut(),
                 MESSAGE_1.as_ptr(),
                 MESSAGE_1.len() as u64,
-                s.payload().as_ptr(),
+                s.to_keypair_bytes().as_ptr(),
```
DCNick3, line 147: Same thing, probably UB

crypto/src/signature/bls/mod.rs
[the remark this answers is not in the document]
Arjentix: It's used as trait method in another place

crypto/src/signature/bls/mod.rs
```diff
@@ -105,17 +117,21 @@ mod small {
         fn set_pairs((g2, g1): &(Self::Generator, Self::SignatureGroup)) -> (&G1, &G2) {
             (g1, g2)
         }
+
+        fn extract_private_key(private_key: &crate::PrivateKey) -> Option<&super::PrivateKey> {
```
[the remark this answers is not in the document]
Arjentix: It's used as trait method in another place

crypto/src/signature/secp256k1.rs
```diff
@@ -1,103 +1,101 @@
...
-
-        pub fn keypair(option: Option<KeyGenOption>) -> Result<(PublicKey, PrivateKey), Error> {
-            let signing_key = match option {
-                Some(mut o) => match o {
+        pub fn keypair(option: Option<KeyGenOption>) -> (PublicKey, PrivateKey) {
+            let signing_key = option.map_or_else(
+                || PrivateKey::random(&mut OsRng),
+                |mut o| match o {
                     KeyGenOption::UseSeed(ref mut seed) => {
                         let mut s = [0u8; PRIVATE_KEY_SIZE];
```
DCNick3, line 53: For `Ed25519Sha512` the seed is first hashed and then fed into `ChaChaRng` to generate the key. Here the order is reversed. Fishy stuff...

crypto/src/signature/secp256k1.rs
```diff
@@ -211,15 +201,14 @@ mod test {
...
+            crate::PrivateKey::Secp256k1(secret),
+        )));
 
-        let sig = EcdsaSecp256k1Sha256::sign(MESSAGE_1, &sk).unwrap();
+        let sig = EcdsaSecp256k1Sha256::sign(MESSAGE_1, &sk);
         let result = EcdsaSecp256k1Sha256::verify(MESSAGE_1, &sig, &pk);
-        // we are returning a `Result<bool>`
-        // unwrap will catch the `Err(_)`, and assert will catch the `false`
-        assert!(result.unwrap());
+        assert!(result.is_ok());
```
DCNick3, line 211: I think just an unwrap will do and will give a marginally better experience (it will print the error message).

Some other tests in `iroha_crypto` also use the `assert!(result.is_ok())` idiom, but I don't think it should be used ever, even if the error message is non-helpful (like in the crypto stuff), it won't tempt you to look at it

core/src/sumeragi/network_topology.rs
Arjentix: Using method lets us not to deref box manually

data_model/src/predicate.rs
```diff
@@ -496,10 +496,10 @@ pub mod string {
                 let (public_key, _) = iroha_crypto::KeyPair::generate()
                     .expect("Should not panic")
                     .into();
-                let id = IdBox::PeerId(peer::PeerId {
-                    address: "localhost:123".parse().unwrap(),
+                let id = IdBox::PeerId(peer::PeerId::new(
+                    "localhost:123".parse().unwrap(),
```
DCNick3, line 500: Same, a good place for the `socket_addr!` macro

data_model/src/predicate.rs
```diff
@@ -1197,10 +1197,7 @@ pub mod value {
 
                 assert!(
                     !pred.applies(&Value::Identifiable(IdentifiableBox::Peer(Peer {
-                        id: peer::PeerId {
-                            address: "localhost:123".parse().unwrap(),
-                            public_key
-                        }
+                        id: peer::PeerId::new("localhost:123".parse().unwrap(), public_key)
```
DCNick3, line 1200: ditto

genesis/src/lib.rs
DCNick3: We should also consider not consuming the key pair by-value, but taking it by-reference.

Key pairs are still non-Copy (due to `BlsSmallPublicKey` being too large and hence being boxed).

Also all of the private keys will need be boxed for `zeroize` support, so even if we ignore the public keys, it's not gonna be cheap.

IMO, there's no benefit to take them by-value, it's just an API nuisance which also has a negative performance effect

DCNick3 (commented):
Very nice to see all those `.unwrap()`'s gone!

config/src/path.rs
```diff
@@ -69,9 +69,10 @@ impl Path {
     /// If the path has an extension.
     pub fn default(path: impl AsRef<std::path::Path>) -> Self {
         let path = path.as_ref().to_path_buf();
-        if path.extension().is_some() {
-            panic!("Default config path is not supposed to have an extension. It is a bug.")
-        }
+        assert!(
+            path.extension().is_some(),
```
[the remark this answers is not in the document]
Arjentix: Oof, missed that

cli/src/lib.rs
```diff
@@ -447,6 +447,7 @@ impl Iroha {
                 // FIXME: don't like neither the message nor inability to throw Result to the outside
                 .expect("Cannot proceed without working subscriptions");
 
+            #[allow(clippy::redundant_pub_crate)]
```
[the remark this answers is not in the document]
DCNick3: You can add links to relevant upstream issues then: https://github.com/tokio-rs/tokio/issues/5616, https://github.com/rust-lang/rust-clippy/issues/10636

crypto/src/lib.rs
Arjentix: Yes, I need to implement it manually, because this crypto types do not implement this traits

DCNick3: Ah, you are right, I completely missed that. Of course the types from other crypto crates don't implement scale codec traits

Arjentix: No particular reasons to keep the same API, that was just the simpliest solution

crypto/src/lib.rs
Arjentix: Same answer:

> No particular reasons to keep the same API, that was just the simpliest solution

To me it looks simplier than exposing all crypto internals. Also this schema should be compatible with `Encode` and `Decode` impls, because SDK rely on that

crypto/src/signature/ed25519.rs
Arjentix: Hm, interesting. Fixed.

crypto/src/signature/secp256k1.rs
Arjentix: Fixed in all `crypto/` code.

I'm kind off dissapointed that there is no idiom to express the logic of *assertions being tested* and *assertions which just expected to work in order to test the first ones*.

That can probably be expressed with panicking for first and with `?` for second, but it's not a common practice as I can see

client_cli/src/main.rs
```diff
@@ -236,12 +236,7 @@ fn submit(
     let iroha_client = Client::new(context.configuration())?;
     let instructions = instructions.into();
     #[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
Arjentix: Oh, yes, initially it was need for error message which I removed

crypto/src/kex/x25519.rs
```diff
@@ -43,30 +49,54 @@ impl KeyExchangeScheme for X25519Sha256 {
...
+
+        let sk = StaticSecret::from(*array_ref!(local_private_key.as_bytes(), 0, 32));
+
+        let pk_slice: &[u8; 32] = remote_public_key.as_bytes();
+        let edwards_compressed =
+            curve25519_dalek::edwards::CompressedEdwardsY::from_slice(pk_slice)
+                .expect("Ed25519 public key has 32 bytes");
+        let edwards = edwards_compressed.decompress().ok_or_else(|| {
+            Error::Parse("Invalid public key: failed to decompress edwards point".to_owned())
+        })?;
```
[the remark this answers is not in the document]
Arjentix: Yes, we get it from another peer and we should not trust it

crypto/src/lib.rs
```diff
@@ -384,52 +506,69 @@ impl PublicKey {
...
     }
 }
 
 #[ffi_impl_opaque]
 impl PrivateKey {
     /// Key payload
-    // TODO: Derive with getset once FFI impl is fixed
-    pub fn payload(&self) -> &[u8] {
-        self.payload.as_ref()
+    pub fn payload(&self) -> ConstVec<u8> {
```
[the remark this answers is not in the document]
Arjentix: Hm, yes, keeping `ConstVec` here no longer makes sense

crypto/src/lib.rs
Arjentix: Created `ToConstVecExt` trait

crypto/src/lib.rs
```diff
@@ -278,58 +310,52 @@ impl From<KeyPair> for (PublicKey, PrivateKey) {
...
 }
 
 #[ffi_impl_opaque]
 impl PublicKey {
     /// Creates a new public key from raw bytes received from elsewhere
-    pub fn from_raw(algorithm: Algorithm, payload: ConstVec<u8>) -> Self {
-        Self {
-            digest_function: algorithm,
-            payload,
+    pub fn from_raw(algorithm: Algorithm, payload: &[u8]) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
Arjentix: I don't get this... If you're talking about const generics, then I can't find any info about their application to functions. Also what would be the difference between such generic and just const fn?

crypto/src/lib.rs
Arjentix: No, not possible because of temporary value.

I've found 14 references for `PublicKey::payload()` and 4 references for `PrivateKey::payload()`.

Do you think it would be more readable to use `key.into_raw().1` in 18 places rather than just `key.payload()`?

@DCNick3, what do you think?

crypto/src/signature/bls/mod.rs
```diff
@@ -105,17 +117,21 @@ mod small {
...
+            if let crate::PrivateKey::BlsSmall(key) = private_key {
+                Some(key)
+            } else {
+                None
+            }
+        }
     }
 
     pub type SmallBls = implementation::BlsImpl<SmallConfiguration>;
     #[cfg(test)]
```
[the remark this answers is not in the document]
Arjentix: Yes, it complains...

crypto/src/multihash.rs
```diff
@@ -154,20 +154,19 @@ impl TryFrom<&Multihash> for Vec<u8> {
     }
 }
 
-impl From<Multihash> for PublicKey {
+impl TryFrom<Multihash> for PublicKey {
+    type Error = crate::Error;
+
     #[inline]
-    fn from(multihash: Multihash) -> Self {
+    fn try_from(multihash: Multihash) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
Arjentix: Maybe I don't understand the whole point of using `Multihash` but to me it looks like a mediator to convert string to `PublicKey` back and forward. So to me it looks like so:

```
String (not strong at all)

    |
    v

Multihash (checks some invariants but still not very strong)

    |
    v

PublicKey (strongest representation)
```

Not any `Multihash` payload is compatible with `PublicKey` (e.g. `ed25519` requires a valid compressed Edwards points). So conversion still might fail.

crypto/src/lib.rs
DCNick3: I created the `into_raw` function with the intention that it would promote looking at both the digest function and the payload with tuple deconstruction (`let (digest_function, payload) = key.into_raw();`). Just never got around to make the other code actually do that..

Briefly looking at the 14 references my IDE finds for `PublicKey`:
- The uses in trait implementations in `iroha_crypto` are all using the `.digest_function()`, except `Encode::size_hint`, which is not that bad
- Code in `iroha_data_model` and `iroha_p2p` that hashes the public keys IMO should really also hash the digest functions. Currently it just ignores the tag, which is fishy and what such an API can actually catch

For `PrivateKey` all the uses are in the trait implementations that also look at `digest_function` except for `Display` (maybe it actually should look at it, but this impl should probably just not exist for private keys in accordance with #3240).

With this in mind, I think it's a good idea to remove the `payload` function, as it just promotes a use that is logically incorrect (getting the key payload while ignoring the crypto algorithm it is actually used with)

P.S. we should probably also stop calling that field a digest function, because it is not. It is digital signature algorithm

crypto/src/signature/secp256k1.rs
DCNick3: Opened as a separate issue #4190

crypto/src/multihash.rs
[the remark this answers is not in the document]
mversic: > Maybe I don't understand the whole point of using Multihash but to me it looks like a mediator to convert string to PublicKey back and forward

I think it's there just for serialization

> Not any Multihash payload is compatible with PublicKey (e.g. ed25519 requires a valid compressed Edwards points). So conversion still might fail.

should this be enforced when creating `Multihash`. To me it doesn't makes sense that it's possible to create invalid `Multihash`.

crypto/src/kex/x25519.rs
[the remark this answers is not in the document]
mversic: shouldn't a thing like that be handled during deserialization? Once constructed `PublicKey` must be valid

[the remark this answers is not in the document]
mversic: shouldn't this be verified during deserialization? Once constructed `PublicKey` must be valid

crypto/src/lib.rs
```diff
@@ -278,79 +308,167 @@ impl From<KeyPair> for (PublicKey, PrivateKey) {
...
+    #[cfg_attr(not(feature="ffi_import"), derive(DeserializeFromStr, SerializeDisplay))]
+    #[cfg_attr(all(feature = "ffi_export", not(feature = "ffi_import")), ffi_type(opaque))]
+    #[allow(missing_docs)]
+    pub enum PublicKey {
+        Ed25519(ed25519::PublicKey),
+        Secp256k1(secp256k1::PublicKey),
+        BlsNormal(bls::BlsNormalPublicKey),
+        BlsSmall(Box<bls::BlsSmallPublicKey>),
     }
 }
```
mversic, line 321:
```suggestion
    pub enum PublicKey<A: Algorithm>(payload: A::Payload)
}
```

I'm just wondering something. Can we make `PublicKey` generic like this?

mversic: the first thing I notice is that this way in `p2p` crate you would see the benefit immediately because there we use a specific public key type. Idk, if there are other places to benefit from this

crypto/src/kex/x25519.rs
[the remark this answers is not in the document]
Arjentix: The problem it that we store X25519 key as Ed25519 key, so not every valid Ed25519 is a valid X25519, see #4174 .

X25519 keys are used only for key exchaging and do not represent real keys used in Iroha. So it's a huge crutch. But this code is here for a long time and goes outside of my current scope because it requires a lot of fixes in downstream p2p code, so a created an issue.

crypto/src/lib.rs
Arjentix: Done

crypto/src/multihash.rs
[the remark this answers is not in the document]
Arjentix: Done

crypto/src/lib.rs
DCNick3: I think only p2p uses a fixed algorithm, while account keys, for example, can be whatever the user wants/needs and this wouldn't be represented all too well.

It would be nice to make smth that would work better with the "we know the type of the key here" for p2p, but other code that is actually dynamic over the algorithm shouldn't struggle with it.

We _may_ consider exposing the individual signature algorithm and the types of their keys, but this is quite a considerable increase of API surface just for the p2p.

Maybe we can just use whatever crates we need in p2p directly? It's not like `iroha_crypto` is bringing a lot of value to p2p in this case, there's nothing to abstract over (unlike `data_model`, which needs to support different signature algos)

genesis/src/lib.rs
Arjentix: I don't think it's currently possible to avoid clonning here. That's because of our `Signature` API design. It stores a public key inside to simplify verification later.

If we want to avoid it we need to to redisign `Signature`. That's outside of the scope of this PR and needs further investigation. I'll create an issue

DCNick3: Ooh, I actually missed that. I just was thinking that it's because `iroha_crypto` consumes the keys, which is not actually the case.

Arjentix: - #4199

crypto/src/lib.rs
Arjentix: I partly agree with @DCNick3 that looks like there are no benefits rather than p2p. The whole point of our key types is that their API is algorithm-agnostic so that it would be easier to introduce new crypto algorithms in future and etc.

However I'd prefer to see all crypto code in just `iroha_crypto` crate. That's easier to audit and more incapsulated.

We might also introduce the support of other crypto key types in p2p as well

primitives/src/const_vec.rs
```diff
@@ -92,6 +92,33 @@ impl<T: IntoSchema> IntoSchema for ConstVec<T> {
...
+    type Item = T;
+
+    type IntoIter = <Vec<T> as IntoIterator>::IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.into_vec().into_iter()
+    }
+}
+
+/// Trait to extend `[T]` with a method to convert it to `ConstVec<T>` by analogy with `[T]::to_vec()`.
```
mversic, line 105: why can't it be inherent impl?

data_model/src/peer.rs
```diff
@@ -58,19 +53,26 @@ pub mod model {
...
     }
+
+    /// Get public key of the peer.
+    #[inline]
+    pub fn public_key(&self) -> &PublicKey {
+        &self.public_key
+    }
+
     /// Serialize the data contained in this Id for use in hashing.
     pub fn payload(&self) -> Vec<u8> {
```
mversic, line 72: please remove this method

primitives/src/const_vec.rs
Arjentix: Not sure what do you mean here. Isn't it the most generic implementation?

mversic: why did you make a trait that is implemented by only 1 type?

DCNick3: You add methods to types that are defined not in your crate (it's in std in this case).

This is a so-called extension trait, a [common pattern in rust](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html)

mversic: yeah, sorry, I thought it was implemented on `ConstVec`. Now I see

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -33,4 +33,4 @@ jobs:
         run: cargo install --path tools/wasm_test_runner
       - name: Run smart contract tests on WebAssembly VM
         working-directory: smart_contract
-        run: mold --run cargo test --tests --target wasm32-unknown-unknown --no-fail-fast --quiet
+        run: mold --run cargo test --release --tests --target wasm32-unknown-unknown --no-fail-fast --quiet
```
mversic, line 36: why would you do this?

crypto/src/lib.rs
```diff
@@ -175,7 +173,6 @@ impl KeyPair {
     ///
     /// # Errors
     /// Fails if decoding fails
-    #[cfg(any(feature = "std", feature = "ffi_import"))]
     pub fn generate() -> Result<Self, Error> {
```
DCNick3, line 177: Maybe this should be gated behind some `cfg` to make sure it isn't available in wasm? Currently it will just panic when the stub `getrandom` impl will get called and panic.

Maybe also add a note in doc that this won't work because randomness in smart contracts is not a thing

crypto/src/lib.rs
```diff
@@ -215,7 +209,6 @@ impl KeyPair {
     ///
     /// # Errors
     /// Fails if decoding fails
-    #[cfg(any(feature = "std", feature = "ffi_import"))]
     pub fn generate_with_configuration(configuration: KeyGenConfiguration) -> Result<Self, Error> {
```
DCNick3, line 213: Similar concerns here as with `generate`, though `KeyGenOption::UseSeed` or `KeyGenOption::FromPrivateKey` are deterministic

crypto/src/hash.rs
```diff
@@ -64,7 +63,6 @@ impl Hash {
     }
 }
 
-#[cfg(any(feature = "std", feature = "ffi_import"))]
 #[ffi_impl_opaque]
```
mversic, line 66: I think you can remove this now if it works in `no_std`

crypto/src/signature/mod.rs
```diff
@@ -63,47 +58,35 @@ impl Signature {
     ///
```
mversic, line 58: I also think you can remove `ffi_impl_opaque` from `Signature`. I can't comment at the correct place

crypto/src/hash.rs
mversic: the reason is that the function doesn't have to be opaque anymore

crypto/src/hash.rs
```diff
@@ -269,7 +268,6 @@ impl<T> HashOf<T> {
     }
 }
 
-#[cfg(any(feature = "std", feature = "ffi_import"))]
```
mversic, line 272: please look at the tests at the end of this file. Should they remain feature gated?

primitives/src/const_vec.rs
```diff
@@ -92,6 +92,33 @@ impl<T: IntoSchema> IntoSchema for ConstVec<T> {
...
+pub trait ToConstVecExt {
```
mversic, line 106:
```suggestion
pub trait ToConstVec {
```

`Ext` suffix implies that there already exists a trait of the same name

crypto/src/lib.rs
mversic: My suggestion here is to guard it behind a feature flag and make that feature part of the `default` feature set. Since smart contracts always use `default-features = false` this will remove friction from all users

mversic (changes requested):
please look into `crypto/merkle.rs`. I think it doesn't have to be `std` feature gated anymore

crypto/src/lib.rs
```diff
@@ -274,89 +299,189 @@ impl From<KeyPair> for (PublicKey, PrivateKey) {
...
+                        name: String::from("payload"),
+                        ty: core::any::TypeId::of::<ConstVec<u8>>(),
+                    },
+                ],
+            }));
+        }
+    }
+}
+
+#[cfg(not(feature = "ffi_import"))]
```
mversic, line 471:
```suggestion
```

this is not necessary?

mversic: also, you can derive `TypeId`

crypto/src/lib.rs
```diff
@@ -274,89 +299,189 @@ impl From<KeyPair> for (PublicKey, PrivateKey) {
...
     }
 
-    /// Extracts the raw bytes from public key
-    pub fn into_raw(self) -> (Algorithm, ConstVec<u8>) {
-        (self.digest_function, self.payload)
+    /// Extracts the raw bytes from public key, copying the payload.
+    ///
+    /// `into_raw()` without copying is not provided because underlying crypto
+    /// libraries do not provide move functionality.
+    pub fn to_raw(&self) -> (Algorithm, Vec<u8>) {
```
mversic, line 340: why is it not called `into_raw` anymore? because it involves cloning?

Arjentix: Yes

mversic: look [here](https://doc.rust-lang.org/std/primitive.str.html#impl-From%3C%26str%3E-for-Box%3Cstr%3E). In `std` there is a `From<&str>` impl that involves cloning. It's obvious that cloning is happening when the receiver type is a reference.

data_model/src/peer.rs
```diff
@@ -27,19 +26,15 @@ pub mod model {
...
-    #[getset(get = "pub")]
     #[ffi_type]
     pub struct PeerId {
         /// Address of the [`Peer`]'s entrypoint.
         // TODO: Derive with getset once FFI impl is fixed
-        #[getset(skip)]
         pub address: SocketAddr,
         /// Public Key of the [`Peer`].
-        pub public_key: PublicKey,
+        pub public_key: Box<PublicKey>,
```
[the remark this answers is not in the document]
Arjentix: Without it `PeerId` takes a lot of space so that clippy starts complaining about `RegisterBox` enum.

I think fixing that warging on `RegisterBox` side would be kind of inconsistent with other instructions

p2p/src/peer.rs
```diff
@@ -463,20 +463,20 @@ mod state {
             }: Self,
         ) -> Result<SendKey<E>, crate::Error> {
             let key_exchange = K::new();
-            let (kx_local_pk, kx_local_sk) = key_exchange.keypair(None)?;
-            let (algorithm, kx_local_pk_raw) = kx_local_pk.clone().into_raw();
+            let (kx_local_pk, kx_local_sk) = key_exchange.keypair(None);
+            let (algorithm, kx_local_pk_raw) = kx_local_pk.to_raw();
```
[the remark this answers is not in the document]
Arjentix: Not so necessary, but future-proof, isn't it? E.g. if we will decide to add a new algorithm in p2p

primitives/src/const_vec.rs
Arjentix: I was thinking more about that: https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html

But yes, my current style follow neither of these, will rename

crypto/src/lib.rs
Arjentix: > this is not necessary?

It still kind of not obvious to me which parts should be under this feature-gate

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -13,6 +13,8 @@ use lol_alloc::{FreeListAllocator, LockedAllocator};
 #[global_allocator]
 static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
 
+getrandom::register_custom_getrandom!(iroha_trigger::stub_getrandom);
```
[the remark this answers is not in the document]
Arjentix: > do we have any documentation on how to write a smart contract? or do the integration tests serve as examples for this?

AFAIK we don't have such documentation now and our existing tests and default validator serve as examples.

@Stukalov-A-M , I think you were working on smart contract API documentation for stable branch, right? If so be aware that after this PR users will also need to add one more boilerplate line...

crypto/src/lib.rs
mversic: the ones that operate on fields of opaque structs

crypto/src/lib.rs
mversic: interesting

data_model/src/peer.rs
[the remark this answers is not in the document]
mversic: ok I guess

p2p/src/peer.rs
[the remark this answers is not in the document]
mversic: I'm more concerned that it might be a bug if we decide to add a new algorithm in p2p

crypto/src/lib.rs
```diff
@@ -185,20 +183,16 @@ impl KeyPair {
 impl KeyPair {
     /// Digest function
     pub fn digest_function(&self) -> Algorithm {
-        self.private_key.digest_function()
+        self.private_key.algorithm()
```
mversic, line 186: so is it `digest_function` or `algorithm`?

crypto/src/lib.rs
```diff
@@ -274,89 +299,189 @@ impl From<KeyPair> for (PublicKey, PrivateKey) {
...
+
+    /// Get the digital signature algorithm of the public key
+    pub fn algorithm(&self) -> Algorithm {
+        match self {
+            Self::Ed25519(_) => Algorithm::Ed25519,
+            Self::Secp256k1(_) => Algorithm::Secp256k1,
+            Self::BlsNormal(_) => Algorithm::BlsNormal,
+            Self::BlsSmall(_) => Algorithm::BlsSmall,
+        }
+    }
```
mversic, line 375: this has become quite useless, since the user can use `match` or `matches`. I would remove it

crypto/src/lib.rs
```diff
@@ -376,79 +501,132 @@ impl PublicKey {
...
-        };
+        Self::from_raw(algorithm, &payload)
+    }
 
-        PublicKey::try_from_private(private_key_candidate).map(|_| Self {
-            digest_function,
-            payload,
-        })
+    /// Get the digital signature algorithm of the private key
+    pub fn algorithm(&self) -> Algorithm {
```
mversic, line 584: also remove

crypto/src/lib.rs
mversic: but we can leave it

crypto/src/lib.rs
```diff
@@ -376,79 +501,132 @@ impl PublicKey {
...
-        #[getset(get_copy = "pub")]
-        digest_function: Algorithm,
-        /// Key payload
-        #[serde(with = "hex::serde")]
-        payload: ConstVec<u8>,
+    #[derive(Clone)]
+    #[cfg_attr(all(feature = "ffi_export", not(feature = "ffi_import")), ffi_type(opaque))]
+    #[allow(missing_docs)]
+    pub enum PrivateKey {
+        Ed25519(Box<ed25519::PrivateKey>),
```
mversic, line 531: I don't quite like seeing a `Box` here

crypto/src/lib.rs
Arjentix: Removed `Option<KeyGenOption>` and added feature-gated `KeyGenOption::Random`

crypto/src/lib.rs
```diff
@@ -376,79 +501,132 @@ impl PublicKey {
...
+        match algorithm {
+            Algorithm::Ed25519 => ed25519::Ed25519Sha512::parse_private_key(payload)
+                .map(Box::new)
+                .map(Self::Ed25519),
+            Algorithm::Secp256k1 => {
+                secp256k1::EcdsaSecp256k1Sha256::parse_private_key(payload).map(Self::Secp256k1)
+            }
+            Algorithm::BlsNormal => bls::BlsNormal::parse_private_key(payload).map(Self::BlsNormal),
+            Algorithm::BlsSmall => bls::BlsSmall::parse_private_key(payload).map(Self::BlsSmall),
+        }
```
mversic, line 568: the thing I mentioned before is that we don't use the `algorithm` value inside this function. This is an indication it could be a generic parameter like so:

```suggestion
    pub fn from_raw<T: Algorithm>(payload: &[u8]) -> Result<Self, ParseError> {
        T::parse_private_key(payload)
    }
```

not saying that we should do it like this. Just a food for thought

mversic: since there are no const generics of enum variants this can only work if we make `Algorithm` a trait and it's variants into empty enums

.github/workflows/iroha2-dev-pr-wasm.yaml
Arjentix: Oh, yes, so for some weird reason now `debug` compilation gives this:

```
Error: WebAssembly translation error

Caused by:
    Invalid input WebAssembly code at offset 2981644: too many locals: locals exceed maximum
```

This seems to be a [known bug](https://github.com/rustwasm/wasm-pack/issues/981) and using `release` [helps](https://github.com/xJonathanLEI/starknet-rs/pull/301#issuecomment-1519113607)

p2p/src/peer.rs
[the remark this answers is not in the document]
Arjentix: Hm, let's add debug assert then?

crypto/src/lib.rs
Arjentix: Algorithm, thx

crypto/src/lib.rs
Arjentix: Most of the time we just need any `Algorithm` value to do further logic, not just check exact value

crypto/src/lib.rs
Arjentix: `ed25519::PrivateKey` contains a lot of math data, it weights `224` bytes, while with this `Box` our main `PrivataKey` weights just `40`.

Not sure what else we can do there

crypto/src/lib.rs
Arjentix: I get what you mean, but still there should be an entrypoint for user to *just* use `PrivateKey` key with algorithm got externally without hassle.

crypto/Cargo.toml
```diff
@@ -42,6 +39,9 @@ std = [
 # Expose FFI API for dynamic linking (Internal use only)
 ffi_export = ["std", "iroha_ffi", "iroha_primitives/ffi_export"]
 
+# Allow creating key using random number generator which is tricky in some environments like Smart Contracts
+generate = []
```
mversic, line 43:
```suggestion
random = []
```

my suggestion

data_model/src/query/mod.rs
```diff
@@ -1231,16 +1231,12 @@ pub mod http {
...
-            self,
-            key_pair: iroha_crypto::KeyPair,
-        ) -> Result<SignedQuery, iroha_crypto::error::Error> {
-            SignatureOf::new(key_pair, &self.payload)
-                .map(|signature| SignedQueryV1 {
-                    payload: self.payload,
-                    signature,
-                })
-                .map(Into::into)
+        pub fn sign(self, key_pair: &iroha_crypto::KeyPair) -> SignedQuery {
```
mversic, line 1234:
```suggestion
        #[must_use]
        pub fn sign(self, key_pair: &iroha_crypto::KeyPair) -> SignedQuery {
```

data_model/src/transaction.rs
```diff
@@ -284,19 +284,17 @@ impl SignedTransaction {
     ///
     /// Fails if signature creation fails
```
mversic, line 285: outdated docs

data_model/src/transaction.rs
```diff
@@ -750,17 +748,14 @@ mod http {
         ///
         /// Fails if signature creation fails
```
mversic, line 749: outdated docs

genesis/src/lib.rs
```diff
@@ -193,11 +185,7 @@ impl GenesisTransactionBuilder {
     ///
     /// # Errors
     /// Fails if signing or accepting fails.
```
mversic, line 187: outdated docs

genesis/src/lib.rs
```diff
@@ -193,11 +185,7 @@ impl GenesisTransactionBuilder {
...
-    fn sign(
-        self,
-        chain_id: ChainId,
-        genesis_key_pair: KeyPair,
-    ) -> core::result::Result<SignedTransaction, iroha_crypto::error::Error> {
+    fn sign(self, chain_id: ChainId, genesis_key_pair: &KeyPair) -> SignedTransaction {
```
mversic, line 188:
```suggestion
    #[must_use]
    fn sign(self, chain_id: ChainId, genesis_key_pair: &KeyPair) -> SignedTransaction {
```

p2p/src/peer.rs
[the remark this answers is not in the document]
mversic: @DCNick3 your thoughts?

data_model/src/block.rs
```diff
@@ -200,12 +200,12 @@ impl SignedBlock {
     /// If given signature doesn't match block hash
```
mversic, line 200: outdated docs

crypto/src/lib.rs
mversic: that's a big difference. I see what you did now, what you could have done is to make a newtype wrapper with Box inside

Arjentix: Yeap, just did it while trying to improve performance

mversic (approved):

p2p/src/peer.rs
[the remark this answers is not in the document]
DCNick3: Yeah, I think an assert would be nice here (not sure if it should be debug-only tbh)

DCNick3 (approved):

PR #4182, opened 2024-01-04
[fix] #4082: Remove cloning for ISI and query execution in wasm

### LINKED ISSUE #4082: [optimization] Executor: validate by value, not by reference
Currently Executor entrypoint functions get all parameters by value. But then validate them by reference because of `Visit` trait. This causes some extra clonning.

This problem can be avoided. Probably need to reconsider `Visit` trait. Especially after #3941

mversic:
## Description

I left `Visit` trait to work with references and have removed redundant cloning through other means. One of the reasons why I left it is so that we can keep calling `mem::forget` at the end of validation but also because it's just not necessary to take them by value

### Linked issue

Closes #4082 

### Benefits

less cloning

### Checklist


configs/peer/executor.wasm
Erigara, line 1: Wow, why the size have increased 18 KB? Not smt critical but i'm just curious

Erigara (commented):
Like this change, often cloning could be avoided when doing serialization.

configs/peer/executor.wasm
mversic: but that's nothing, change is 0%. For instance I noticed that your PR increased size by 20%. I find that interesting

Erigara: > but that's nothing, change is 0%

Ok, i see

> For instance I noticed that your PR increased size by 20%.

You mean token unification one? I guess it could be because we added new function to every permission token.

data_model/src/isi.rs
```diff
@@ -105,65 +107,448 @@ pub mod model {
...
+    impl Instruction for InstructionBox {
+        fn encode_as_instruction_box(&self) -> Vec<u8> {
+            self.encode()
+        }
+    }
+
+    impl Instruction for SetKeyValue<Domain> {
+        fn encode_as_instruction_box(&self) -> Vec<u8> {
+            let mut output = InstructionType::SetKeyValue.encode();
+            SetKeyValueType::Domain.encode_to(&mut output);
```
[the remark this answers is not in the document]
Erigara: I agree, but it would help to find the issue much more quickly. If we can see errors in (de)serialization.

DCNick3:
I kind of don't like how fragile this approach is: any mismatch between the `encode_as_*` function implementation and the derived scale encodings (that can be easily created by accident when modifying some instructions, for example) would lead to hard-to-diagnose encoding issues. But I don't there's a way to resolve this that doesn't rely on macros, and we already have enough of those in iroha..

If we decide to go the macro way we can either:
- automatically derive the `encode_as_*` functions (keeping the approach used in this PR)
- automatically generate the sister `*BoxRef` structures with owned values replaced by references (following the approach suggested by @Erigara)

I _think_ the second one will be easier to implement, especially seeing how heterogeneous the encode methods for instructions are due to different levels of nesting..

Another far-fetched idea w/o using macros is to somehow make instructions/queries cheaper to clone (for example, by moving out all the expensive objects like strings to arenas/bump allocators/...). I don't think those are worth the complexity they bring though.

config/src/path.rs
```diff
@@ -69,9 +69,10 @@ impl Path {
     /// If the path has an extension.
     pub fn default(path: impl AsRef<std::path::Path>) -> Self {
         let path = path.as_ref().to_path_buf();
-        if path.extension().is_some() {
```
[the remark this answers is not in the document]
mversic: yes

data_model/src/isi.rs
```diff
@@ -22,7 +22,10 @@ use crate::{seal, Level, Registered};
 /// Instructions allows to change the state of `Iroha`.
 /// All possible instructions are implementors of this trait, excluding
 /// [`InstructionBox`] which is just a wrapper.
-pub trait Instruction: Into<InstructionBox> + seal::Sealed {}
+pub trait Instruction: Into<InstructionBox> + seal::Sealed {
+    /// [`Encode`] [`Self`] as [`InstructionBox`]
+    fn encode_as_instruction_box(&self) -> Vec<u8>;
```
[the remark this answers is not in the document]
Arjentix: For future code readers I'd add some explanation why do we need this hack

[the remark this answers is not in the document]
Arjentix: > What about to use [EncodeLike](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.EncodeLike.html)?

I see, you probably don't want for `ConcreteIsi::encode()` to be equal to `ConcreteIsi::encode_as_instruction_box()`, right? But why not actually...?

smart_contract/src/lib.rs
```diff
@@ -79,72 +78,70 @@ pub trait ExecuteOnHost: Instruction {
...
 
-impl<Q: Query> From<QueryRequest<Q>> for SmartContractQueryRequest {
-    fn from(query_request: QueryRequest<Q>) -> Self {
-        SmartContractQueryRequest::query(
-            query_request.query.into(),
-            query_request.sorting,
-            query_request.pagination,
-            query_request.fetch_size,
-        )
+impl<Q: Query> Encode for QueryWithParameters<'_, Q> {
```
Arjentix, line 114: I think having common structure in data_model with implemented `EncodeLike` would be more robust.

Currently there's no state that this encoding should be fully compatible with the structure in `core/src/smartcontracts/wasm.rs`

smart_contract/src/lib.rs
```diff
@@ -426,10 +425,39 @@ mod tests {
...
     use super::*;
 
+    #[derive(Decode)]
+    struct QueryWithParameters<Q> {
+        query: Q,
+        sorting: Sorting,
+        pagination: Pagination,
+        #[allow(dead_code)]
+        fetch_size: FetchSize,
+    }
```
Arjentix, line 440: And with structure in data_model you wouldn't need to copy-paste it

Arjentix:
> any mismatch between the encode_as_* function implementation and the derived scale encodings (that can be easily created by accident when modifying some instructions, for example) 
 
 I don't think that could actually happen to us.
`encode_as_*` fallback to `self.encode_to()`, so changing internal fields won't break anything.
And while we keep this enums everything should be fine. Throwing back these enums would be a big change and we certunally catch such problems with encodings (at least because of autogenerated `*Type` enums).

Arjentix:
I suggest to create a derive macro called something like `EncodeLikeInstructionBox` and `EncodeLikeQueryBox`. Same for decoding. They will implement only existing traits from `parity_scale_codec` like `Encode` and `EncodeLike`, no new traits needed.

data_model/src/isi.rs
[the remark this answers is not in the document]
mversic: > I see, you probably don't want for `ConcreteIsi::encode()` to be equal to `ConcreteIsi::encode_as_instruction_box()`, right? But why not actually...?

yes, I don't think that's a good idea. And I can't just implement `EncodeLike`, I have to first implement custom `Encode`/`Decode` for every concrete ISI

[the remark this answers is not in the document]
Arjentix: > And I can't just implement EncodeLike, I have to first implement custom Encode/Decode for every concrete ISI

Yes, but a custom proc macro could simpify that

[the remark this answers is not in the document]
mversic: actually, I'm becoming fond of this idea. I'm not sure if there is any downside except that we will always add one byte to serialized form of concrete ISI. But then again I think that we're always serializing `InstructionBox`, never concrete instruction. This will bring us a lot of benefits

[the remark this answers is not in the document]
mversic: I chose a hybrid approach. This could be a foot gun if we ever introduce individual instructions in the API considering that it wouldn't match the schema we provide SDKs for encoding

mversic:
@Arjentix @DCNick3 @Erigara how do you like the modified approach now?

mversic:
@Arjentix @DCNick3 @Erigara how do you like the modified hybrid approach now? I think it eliminates the previous concerns. I chose to expose `encode_as_instruction_box` rather than expose `InstructionBoxRef` in the public API

data_model/src/isi.rs
```diff
@@ -207,26 +231,22 @@ mod transparent {
     }
 
     macro_rules! impl_into_box {
-        (
-            $($isi:ident $(< $($generic:ident $(< $nested_generic:ident >)?),+ >)?)|*
-            ==> $boxed:ident :: $variant:ident
+        ( $($isi:ty)|*
+          => $middle:ty => $boxed:ty[$variant:ident],
+          => $middle_ref:ty => $boxed_ref:ty[$variant_ref:ident]
```
Arjentix, line 236:
```suggestion
          => $middle_ref:ty => $boxed_ref:ty :: $variant_ref:ident
```

Is that not allowed?

Arjentix, line 236: Can the second line be derived from the first one?

Arjentix (approved):
I like it

data_model/src/isi.rs
mversic: well, type cannot be followed by `::` because it's a path delimiter

data_model/derive/src/lib.rs
```diff
@@ -9,6 +10,49 @@ use iroha_macro_utils::Emitter;
...
+/// pub(crate) enum InnerEnumRef<'a> {
+///     A(&'a u32),
+///     B(&'a i32),
+/// }
+///
+/// pub(crate) enum OuterEnumRef<'a> {
+///     A(&'a String),
+///     B(InnerEnumRef<'a>),
+/// }
+/// */
```
Arjentix, line 46:
```suggestion
/// /* will produce:
/// #[derive(Encode)
/// pub(crate) enum InnerEnumRef<'a> {
///     A(&'a u32),
///     B(&'a i32),
/// }
///
/// #[derive(Encode)
/// pub(crate) enum OuterEnumRef<'a> {
///     A(&'a String),
///     B(InnerEnumRef<'a>),
/// }
/// */
```

I think it should be with derives

data_model/src/query/mod.rs
```diff
@@ -244,6 +260,69 @@ pub mod model {
...
 }
 
+macro_rules! impl_query {
+    ($($ty:ty => $output:ty),+ $(,)?) => { $(
+        impl Query for $ty {
+            type Output = $output;
+
+            fn encode_as_query_box(&self) -> Vec<u8> {
+                QueryBoxRef::from(self).encode()
+            }
```
Arjentix, line 270: If encoding is so easy now and used in only one place then maybe use that construction in that place and remove this method?

Arjentix (approved):

data_model/src/query/mod.rs
mversic: it's just a question of what API do we want to expose. Notice that `QueryBoxRef` is private atm

PR #4185, opened 2024-01-09
[feature] #4126: Add chain_id to prevent replay attacks

### LINKED ISSUE #4126: [BUG]: Prevent cross chain replay attacks
If two iroha networks had the same user (with the same pub key).

Then transaction on one network could be reused on the another network without user intention. 

Proposed solution is to use unique network identificator (configuration parameter) which would be included in the transaction. 

Workaround is avoid situation when user have the same pub key across network.

mversic:
### Linked issue

Closes #4126

### Checklist


client/tests/integration/domain_owner.rs
```diff
@@ -122,7 +122,7 @@ fn domain_owner_asset_definition_permissions() -> Result<()> {
 
     // register asset definitions by "bob@kingdom" so he is owner of it
     let coin = AssetDefinition::quantity(coin_id.clone());
-    let transaction = TransactionBuilder::new(bob_id.clone())
+    let transaction = TransactionBuilder::new(0, bob_id.clone())
```
Erigara, line 125: Maybe add `with_chain_id` with 0 by default? Because it's not really obvious what this 0 represent.

docker-compose.dev.single.yml
```diff
@@ -7,6 +7,7 @@ services:
     build: ./
     platform: linux/amd64
     environment:
+      IROHA_CHAIN_ID: '0'
```
Erigara, line 10: Why quotes was used only here?

Erigara:
Let's in future separate versions bump in different commit

docker-compose.dev.single.yml
mversic: good point

client/src/client.rs
```diff
@@ -344,6 +344,8 @@ impl_query_output! {
 )]
 #[display(fmt = "{}@{torii_url}", "key_pair.public_key()")]
 pub struct Client {
+    /// Unique id of the blockchain. Used for simple replay attack protection.
+    pub chain_id: u16,
```
DCNick3, line 348: Is using a `u16` here a good idea?

It would hard for the deployers to select a unique value, as there isn't really a defined order for various iroha chains unlike Etherium which can just use 1 for mainnet, and other networks just don't matter as much.

Even if deployers would select perfectly random numbers (which they most probably won't), due to birthday paradox they would collide quite often. If my calculations are right, with 300 randomly chosen `u16` numbers, with around 50% at least two will collide.

I propose to make the chain id a string, which is more natural for unordered structures.

If dealing with variable-sized data structure here is a problem, we can hash the user-provided config value into 64 bits with blake2 and then use that.

mversic: sure, it was a stupid choice. Even though `chain_id` is a publicly available information it should be best defined as `UUID` because there is no way to coordinate uniqueness of `chain_id` across different chains

DCNick3: Yeah, a UUID would also work here

mversic: changed to `String` now

Erigara (dismissed):
LGTM, few optional comments

client/benches/tps/utils.rs
```diff
@@ -218,8 +220,9 @@ impl MeasurerUnit {
             for instruction in instructions {
                 match shutdown_signal.try_recv() {
                     Err(mpsc::TryRecvError::Empty) => {
-                        let mut transaction = TransactionBuilder::new(alice_id.clone())
-                            .with_instructions([instruction]);
+                        let mut transaction =
+                            TransactionBuilder::new(chain_id.clone(), alice_id.clone())
```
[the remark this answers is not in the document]
mversic: yes because it's not optional

DCNick3 (dismissed):

Erigara (approved):

data_model/src/lib.rs
```diff
@@ -584,6 +584,25 @@ pub mod parameter {
...
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    #[repr(transparent)]
+    #[serde(transparent)]
+    #[ffi_type(unsafe {robust})]
+    pub struct ChainId(String);
```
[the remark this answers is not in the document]
mversic: done

DCNick3 (approved):

DCNick3 (approved):

data_model/src/lib.rs
```diff
@@ -1002,6 +1011,22 @@ pub mod model {
...
+}
+
+impl Decode for ChainId {
+    fn decode<I: parity_scale_codec::Input>(
+        input: &mut I,
+    ) -> Result<Self, parity_scale_codec::Error> {
+        let boxed: String = parity_scale_codec::Decode::decode(input)?;
+        Ok(Self::new(&boxed))
+    }
 }
```
Erigara, line 1030: This could be improved by using constructor `fn new(inner: impl Into<Box<str>>)`.
I will extract it into separate good first issue.

Erigara (approved):

PR #4187, opened 2024-01-09
[ci] #4151: Enhance the test environment by automatically generating WASM files for the smartcontracts to be tested

### LINKED ISSUE #4151: Enhance `test_env.py` to Automate WASM Smart Contract Generation and Organization
### Feature Request
We propose an enhancement to the `test_env.py` script used in setting up the Iroha test environment. The enhancement involves automating the generation of WebAssembly (WASM) files for smart contracts and organizing these files into a designated directory (`smartcontracts`) within the `test` directory.

### Motivation
The current process of managing smart contracts in the Iroha test environment requires manual generation and organization of WASM files. This is not only time-consuming but also prone to human error. By automating this process, we can ensure consistency, reduce manual effort, and make it easier to manage and add new smart contracts in the future.

Asem-Abdelhady:
## Description

I made a shell script to create test-smartcontracts directory and build the WASM files in that directory
### Linked issue

Closes #4151 

### Benefits

No need to manually generate the WASM files.

### Checklist

- [X] All applicable CI checks pass (or I promised to make them pass later)

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -128,8 +128,11 @@ impl NetworkRelay {
         }
 
         match msg {
-            SumeragiPacket(data) => {
-                self.sumeragi.incoming_message(*data);
+            SumeragiBlock(data) => {
+                self.sumeragi.incoming_block_message(*data);
+            }
+            SumeragiControlFlow(data) => {
+                self.sumeragi.incoming_control_flow_message(*data);
             }
             BlockSync(data) => self.block_sync.message(*data).await,
             TransactionGossiper(data) => self.gossiper.gossip(*data).await,
--- a/client/tests/integration/connected_peers.rs
+++ b/client/tests/integration/connected_peers.rs
@@ -95,6 +95,8 @@ fn connected_peers_with_f(faults: u64, start_port: Option<u16>) -> Result<()> {
     thread::sleep(pipeline_time * 2); // Wait for some time to allow peers to connect
     let (removed_peer, removed_peer_client) = peer_clients.remove(removed_peer_idx);
 
+    thread::sleep(pipeline_time * 2); // Wait for some time to allow peers to disconnect
+
     check_status(&peer_clients, 2);
     let status = removed_peer_client.get_status()?;
     // Peer might have been disconnected before getting the block
--- a/core/src/block_sync.rs
+++ b/core/src/block_sync.rs
@@ -129,7 +129,6 @@ impl BlockSynchronizer {
 pub mod message {
     //! Module containing messages for [`BlockSynchronizer`](super::BlockSynchronizer).
     use super::*;
-    use crate::sumeragi::view_change::ProofChain;
 
     /// Get blocks after some block
     #[derive(Debug, Clone, Decode, Encode)]
@@ -234,12 +233,11 @@ pub mod message {
                     }
                 }
                 Message::ShareBlocks(ShareBlocks { blocks, .. }) => {
-                    use crate::sumeragi::message::{Message, MessagePacket};
+                    use crate::sumeragi::message::BlockMessage;
                     for block in blocks.clone() {
-                        block_sync.sumeragi.incoming_message(MessagePacket::new(
-                            ProofChain::default(),
-                            Some(Message::BlockSyncUpdate(block.into())),
-                        ));
+                        block_sync
+                            .sumeragi
+                            .incoming_block_message(BlockMessage::BlockSyncUpdate(block.into()));
                     }
                 }
             }
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -25,8 +25,9 @@ use parity_scale_codec::{Decode, Encode};
 use tokio::sync::broadcast;
 
 use crate::{
-    block_sync::message::Message as BlockSyncMessage, prelude::*,
-    sumeragi::message::MessagePacket as SumeragiPacket,
+    block_sync::message::Message as BlockSyncMessage,
+    prelude::*,
+    sumeragi::message::{BlockMessage, ControlFlowMessage},
 };
 
 /// The interval at which sumeragi checks if there are tx in the `queue`.
@@ -59,8 +60,10 @@ pub type EventsSender = broadcast::Sender<Event>;
 /// The network message
 #[derive(Clone, Debug, Encode, Decode)]
 pub enum NetworkMessage {
-    /// Blockchain message
-    SumeragiPacket(Box<SumeragiPacket>),
+    /// Blockchain concensus data message
+    SumeragiBlock(Box<BlockMessage>),
+    /// Blockchain concensus control flow message
+    SumeragiControlFlow(Box<ControlFlowMessage>),
     /// Block sync message
     BlockSync(Box<BlockSyncMessage>),
     /// Transaction gossiper message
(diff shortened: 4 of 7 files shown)


Erigara:
Please only add relevant commits

scripts/test_env.py
```diff
@@ -143,6 +143,35 @@ def run(self, shared_env: dict(), submit_genesis: bool = False):
...
+def initialize_test_smart_contracts(root_dir: pathlib.Path, out_dir: pathlib.Path):
+    """
+    Function to make directory called smartcontracts, build every integeration smartcontract,
+    the it moves the result WASM file to the smartcontracts directory. 
+    """
+    smartcontracts_dir = out_dir / "smartcontracts"
+    os.makedirs(smartcontracts_dir, exist_ok=True)
+    source_dir = pathlib.Path("./client/tests/integration/smartcontracts")
+
+    #Iterate through every smartcontract in the inte
```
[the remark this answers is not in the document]
Asem-Abdelhady: was typo somehow the full comment was not committed.

scripts/test_env.py
```diff
@@ -143,6 +143,35 @@ def run(self, shared_env: dict(), submit_genesis: bool = False):
...
+        wasm_file_name = f"{folder_name}.wasm"
+        wasm_file_path = root_dir / wasm_file_name
+
+        # Build the smart contracts in the root director
+        subprocess.run([
+            "cargo", "run", "--bin", "iroha_wasm_builder_cli", "--", "build",
+            str(folder), "--optimize", "--outfile", str(wasm_file_path)
+        ])
+
+        # Move the .wasm file to the smarts contracts folder
```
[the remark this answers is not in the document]
Asem-Abdelhady: Done

scripts/test_env.py
```diff
@@ -143,6 +143,31 @@ def run(self, shared_env: dict(), submit_genesis: bool = False):
...
+def initialize_test_smart_contracts(out_dir: pathlib.Path):
+    """
+    Function to make directory called smartcontracts, build every integeration smartcontract,
+    the it moves the result WASM file to the smartcontracts directory. 
+    """
+    smartcontracts_dir = out_dir / "smartcontracts"
+    os.makedirs(smartcontracts_dir, exist_ok=True)
+    source_dir = pathlib.Path("./client/tests/integration/smartcontracts")
+
+    #Iterate through every smartcontract in the integeration tests folder
```
Erigara, line 155:
```suggestion
    #Iterate through every smartcontract in the integration tests folder
```

Erigara (commented):
I was thinking about should we make smartcontract build optional?

Asem-Abdelhady:
> I was thinking about should we make smartcontract build optional?

@Erigara We can add argument `--wasm_generation` with default true value. We check the arg before the generation step.

Erigara:
I suggested to make this optional is the fact that this functionality is only necessary to run python integration tests. 
So if i don't need wasm files i could skip this step and don't waste my time.

scripts/test_env.py
```diff
@@ -143,6 +143,31 @@ def run(self, shared_env: dict(), submit_genesis: bool = False):
...
+def initialize_test_smart_contracts(out_dir: pathlib.Path):
+    """
+    Function to make directory called smartcontracts, build every integeration smartcontract,
+    the it moves the result WASM file to the smartcontracts directory. 
+    """
+    smartcontracts_dir = out_dir / "smartcontracts"
+    os.makedirs(smartcontracts_dir, exist_ok=True)
+    source_dir = pathlib.Path("./client/tests/integration/smartcontracts")
+
+    #Iterate through every smartcontract in the integration tests directory
```
mversic, line 155:
```suggestion
```

useless

mversic:
> I suggested to make this optional is the fact that this functionality is only necessary to run python integration tests.
So if i don't need wasm files i could skip this step and don't waste my time.

I don't think we should put the test smart contract build into `test_env` script at all

Asem-Abdelhady:
> > I suggested to make this optional is the fact that this functionality is only necessary to run python integration tests.
> > So if i don't need wasm files i could skip this step and don't waste my time.
> 
> I don't think we should put the test smart contract build into `test_env` script at all

Do you think it is better to add it in the CI process? or not to do it at all?

scripts/test_env.py
Asem-Abdelhady: Done

Asem-Abdelhady:
@AlexStroke  We reached an approach to make a separate script and add it in the CI process so those who want to set up the test env don't have to build the smartcontracts. The script will be done using Github artifacts. I will search about how exactly it works then I will commit in the PR accordingly. Is that okay with you?

AlexStroke:
> @AlexStroke We reached an approach to make a separate script and add it in the CI process so those who want to set up the test env don't have to build the smartcontracts. The script will be done using Github artifacts. I will search about how exactly it works then I will commit in the PR accordingly. Is that okay with you?

yes

Asem-Abdelhady:
> > @AlexStroke We reached an approach to make a separate script and add it in the CI process so those who want to set up the test env don't have to build the smartcontracts. The script will be done using Github artifacts. I will search about how exactly it works then I will commit in the PR accordingly. Is that okay with you?
> 
> yes

Github artifacts are not needed anymore. The shell script provided does the job, you can use it by passing the SOURCE-DIR. The script can be integerated into the CI as well.

.gitignore
```diff
@@ -49,3 +49,4 @@ result
 /test/
 /iroha-java/
 /lcov.info
+**/test-smartcontracts/
```
0x009922, line 52: Quite... indeterminate. I guess the target is always at the root?
```suggestion
/test-smartcontracts
```

0x009922 (changes requested):
Please also document somewhere how to run the script, so that a person could simply cope-paste the command into shell.

.gitignore
Asem-Abdelhady: it depends on from where you are running the script. if you are running from root then it will be in root yes

Stukalov-A-M:
@Asem-Abdelhady, the Iroha-java team is already created script for autogenerating WASMs, I think you may just reuse it.
https://github.com/hyperledger/iroha-java/blob/iroha2-dev/modules/client/src/test/executor/build.sh

0x009922:
@Stukalov-A-M, IMO both scripts are quite small and trivial, and it doesn't seem like there is something to reuse really.

Asem-Abdelhady:
> @Stukalov-A-M, IMO both scripts are quite small and trivial, and it doesn't seem like there is something to reuse really.

i agree

.gitignore
Asem-Abdelhady: I believe it is better to leave it this way. maybe someone wanted to create a dir locally to test something so it makes sense if he ran the command within this dir to have test-smartcontracts inside that that dir as well.

scripts/generate_wasm.sh
```diff
@@ -0,0 +1,32 @@
...
+    fi
+done
+
+echo "Smart contracts build complete."
+
+# How to run:
+# make sure you root from the root directrory if iroha or
+# any iroha directory that is not a package in the working space
+# run the following command:
+# ./path/to/script/generate_wasm.sh /path/to/smart_contracts
```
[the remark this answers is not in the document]
Asem-Abdelhady: Done

scripts/generate_wasm.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
Erigara, line 1: If you don't use any `bash` specific functionality i would suggest to use `#!/bin/sh`

Asem-Abdelhady: done

AlexStroke (approved):
LGTM

Erigara (approved):
LGTM

0x009922 (approved):

0x009922 (approved):

Erigara (approved):

PR #4188, opened 2024-01-10
[refactor]: update wasm_codec_derive to use syn2

### LINKED ISSUE #3943: Migrate `iroha_core_wasm_codec_derive` to syn 2.0
Part of https://github.com/hyperledger/iroha/issues/3682

Do the same work that was done in https://github.com/hyperledger/iroha/pull/3727 and https://github.com/hyperledger/iroha/pull/3759 and https://github.com/hyperledger/iroha/issues/3814 , but for `iroha_core_wasm_codec_derive`

VAmuzing:
## Description

Introduction of manyhow & upgrade syn to syn2

### Linked issue

Closes #3943 

### Checklist


wasm_codec/derive/src/lib.rs
```diff
@@ -343,182 +395,195 @@ enum ErrType {
...
-fn parse_state_param(param: &syn::PatType) -> Result<&syn::Type, Diagnostic> {
-    let syn::Pat::Ident(pat_ident) = &*param.pat else {
-        return Err(diagnostic!(
-            param,
-            Level::Error,
-            "State parameter should be an ident"
-        ));
+fn parse_state_param(param: &syn2::PatType) -> Result<&syn2::Type> {
+    let syn2::Pat::Ident(pat_ident) = &*param.pat else {
+        bail!(param, "State parameter should be an ident");
```
Arjentix, line 507: Maybe I don't understand something here, but why here you use `bail!` with `Result` without any `emitter` and `Option`, but you don't do that in other places.

I'm asking because it looks kind of strange to insert additional `emitter` parameter and `Option` return type

DCNick3: The idea of using an emitter and an option is to try and collect as much errors as possible instead of erroring out.

Ideally, one would even try to generate some code with the incomplete information, though it is not always feasible from the complexity standpoint.

For that purpose I think that this function (and other ones that were returning a `Result<..., Diagnostic>`) should be refactored to use the same style.

DCNick3:
As a general comment: I am not sure how feasible it is, but it would be very nice if this code tried to recover from the missing/erroneous values provided by the user, as it seems that right now it bails out pretty fast with a `?`, negating the benefits of using the `Emitter` type

wasm_codec/derive/src/lib.rs
```diff
@@ -1,31 +1,31 @@
-//! This crate provides [`wrap`] and [`wrap_signature`] attribute macros to wrap a host-defined
```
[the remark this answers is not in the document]
VAmuzing: Wow, I don't even know how that happened!
Going to fix it

wasm_codec/derive/src/lib.rs
VAmuzing: I wanted to use `bail!` only in places where it made sense to stop.
But as @DCNick3 pointed out I'm basically doing the same thing with the `?` operator where I could've gathered more errors.
I'm going to redo / rearrange some checks to do that.

appetrosyan:
You need to check if the ports for the integration tests collide, otherwise the integration tests might not pass because of a conflict.

VAmuzing:
> You need to check if the ports for the integration tests collide, otherwise the integration tests might not pass because of a conflict.

It seems like the failure of the integration tests was due to I/O part of the tests. Re-running them fixed the issue. But it definitely is an issue to look into to make the tests more robust.

Arjentix (approved):

Arjentix (approved):

DCNick3 (approved):

PR #4198, opened 2024-01-12
[fix] #4162: SignedTransaction is immutable

### LINKED ISSUE #4162: Are genesis transaction signatures checked?
> BTW, I am confused by both tests patching constructed `GenesisNetwork`. They are inserting plain instructions into the payload of a signed transaction. By doing so, AFAIK, they invalidate the signature of the transaction being computed using the previous payload. One test checks that Iroha validates genesis by providing an invalid permission token, but Iroha should not accept genesis because of its signatures I guess. Another test checks whether base64 triggers are handled fine. I don't why the second test passes.

_Originally posted by @0x009922 in https://github.com/hyperledger/iroha/pull/4153#discussion_r1432095800_

Erigara:
## Description

Remove `payload_mut` method which is considered harmful.

### Linked issue

Closes #4162 

### Benefits

Don't break `SignedTransaction` invariant.


data_model/src/transaction.rs
```diff
@@ -744,6 +734,12 @@ mod http {
             self
         }
 
+        /// Set creation time of transaction
+        pub fn set_creation_time(&mut self, creation_time_ms: u64) -> &mut Self {
```
mversic, line 738: the same

mversic (changes requested):

data_model/src/transaction.rs
```diff
@@ -715,6 +699,12 @@ mod http {
             self
         }
 
+        /// Set executable for this transaction
+        pub fn with_executable(mut self, executable: Executable) -> Self {
```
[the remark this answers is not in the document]
mversic: it's only used for testing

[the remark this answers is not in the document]
Erigara: It's not different from `with_wasm` or `with_instructions`.
Also doesn't break any invariants because we are working with `TransactionBuilder`, not `SignedTransaction`.

data_model/src/transaction.rs
Erigara: I don't see why this method should be private, user might want to specify exact timestamp for his transaction.

mversic: but then you have to modify the signature

data_model/src/transaction.rs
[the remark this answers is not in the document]
mversic: oh, I see it's defined on `TransactionBuilder`. But we already have `with_wasm` and `with_instructions`. I think you can remove both in favor of the `with_executable(mut self, executable: impl Into<Executable>`

data_model/src/transaction.rs
mversic: ok. Is `creation_time` even required to be part of the payload and signed?

data_model/src/transaction.rs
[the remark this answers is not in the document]
Erigara: good point

data_model/src/transaction.rs
Erigara: yes, otherwise it will be hard to distinguish two transfers

data_model/src/transaction.rs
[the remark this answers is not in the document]
Erigara: it won't work unfortunately with `with_instructions` because it's more involved than just call `into`

DCNick3 (approved):

mversic (approved):

data_model/src/transaction.rs
mversic: could the `creation_time` be assigned by the node instead of the client?

Erigara: i don't this so because this will mess up transaction signature

mversic: yes, it won't be a part of transaction signature. I'm questioning whether it makes sense to sign this piece of information? We're trusting a user that they will use a correct time, we might be trusting the peer

Erigara: @mversic if transaction has wrong time it will be just rejected either as being outdated or from future

mversic: > @mversic if transaction has wrong time it will be just rejected either as being outdated or from future

I know this. My intention here is to remove `create_time_ms` from transaction that user submits so that we can remove `pending_transaction` endpoint all together. It's the only thing that stands in the way of MST

Erigara: I'm not sure that this will end up well. 
So currently we use time + nonce to distinguish between otherwise identical transactions and time is convenient way to do so.

If we would remove time from the transaction user have to use unique nonce every time he send transaction. 
And we back to square one where user need to get pending transactions in order to get proper nonce.

We might consider that creation of MST is happens outside iroha: that interested parties can communicate and create proper MST. 
This would allow to simplify:
1. CLI by not querying `/pending` transactions on every command
2. Queue because we can reject transactions with invalid signature check immediately we won't have partial MST which might or might not get additional signatures.

mversic: thanks for the thoughtful response

DCNick3 (approved):

mversic (approved):

PR #4201, opened 2024-01-15
[refactor]: Fix CI lints after bump of toolchain version

mversic:
## Description

Fixed CI lints

`missing_panics_doc` now fires for every public function that calls `Option::expect` or `Result::expect`. In some cases (like `Metadata`) I've extracted the panicking into a special private function because I don't think it's a good solution to just slap `missing_panics_doc` on every function that asserts invariants. We definitely want to get rid of these warnings, but what to do about it?

### Linked issue

Closes #{issue_number} 

### Checklist


smart_contract/executor/derive/src/validate.rs
```diff
@@ -41,7 +41,6 @@ impl FromAttributes for ValidateAttribute {
     // but we still _want_ to validate that each attribute parses successfully
     // this is to ensure that we provide the user with as much validation as possible, instead of bailing out early
     // `Option::or_else` would NOT work here, as it would not validate conditions after the first valid one
-    #[allow(clippy::or_fun_call)]
```
DCNick3, line 44: Please see the comment above, this is NOT correct, this code actually DOES rely on the `or` argument to be executed every time

crypto/src/hash.rs
```diff
@@ -224,6 +222,8 @@ impl<T> Clone for HashOf<T> {
 impl<T> Copy for HashOf<T> {}
 
 impl<T> PartialEq for HashOf<T> {
+    // NOTE: False positive
+    #[allow(clippy::unconditional_recursion)]
```
DCNick3, line 226: Yeeah, this one is definitely because of the night (and fixed already: https://github.com/rust-lang/rust-clippy/issues/12133)

DCNick3 (approved):

crypto/src/hash.rs
mversic: that's why I said if we could somehow upgrade to a nightly version that will become next stable. But I'm not sure that Rust's release cycle works that way

DCNick3: I don't know much about the versioning of clippy, but rustc doesn't work this way. They branch it as a beta, backport bugfixes to it and then promote it to stable. 

Clippy seems to have a somewhat similar process: https://doc.rust-lang.org/nightly/clippy/development/infrastructure/release.html

Cargo.toml
```diff
@@ -144,12 +144,10 @@ rust.nonstandard_style = "deny"
 rust.rust_2018_idioms = "deny"
 rust.trivial_casts = "deny"
 rust.trivial_numeric_casts = "deny"
-rust.unconditional_recursion = "deny"
```
Arjentix, line 147: Because of false-positives, right? Maybe add a todo? This lint seems useful

Cargo.toml
Arjentix, line 1: Have you check newly added lints? I guess we should enable them mannually, we can do a research for new useful lints in another task

Arjentix (approved):

Cargo.toml
mversic: they are reported as warnings

mversic: would you like me to return it to `deny`?

Cargo.toml
mversic: no, maintaining lints is a work in itself with marginal utility. IMHO I would happily just use whatever is `clippy`s default

Cargo.toml
Arjentix: Warnings are ok

Cargo.toml
Arjentix: I'd prefer no enable something like `deny restriction::*` and then allowing everything we like

Arjentix (approved):

DCNick3 (approved):

PR #4203, opened 2024-01-16
[feature] #4183: change endpoint to return Vec of matching transactions

### LINKED ISSUE #4183: Return all matching pending transactions instead of all pending transactions
Currently we're returning `Vec<SignedTransaction>` from torii endpoint and later we're trying to find the first one that matched the equality function `equals_excluding_creation_time`.
As this is the only use case for that endpoint lets change it (or add a separate one) to get the necessary transaction instead of the whole queue.

VAmuzing:
## Description

Update endpoint that returned `Vec<SignedTransaction>` that were in a queue to return `Vec<SignedTransaction>` that are equal (by equality function) to our demanded transaction instead.

### Linked issue

Closes #4183 

### Benefits
It decreases the amount of traffic through the network.

### Checklist

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1018,8 +1018,7 @@ impl Client {
             .join(crate::config::torii::PENDING_TRANSACTION)
             .expect("Valid URI");
         let pagination = pagination.into_query_parameters();
-        let body = serde_json::to_vec(transaction.payload())
-            .wrap_err("Failed to serialize SignedTransaction.payload")?;
+        let body = transaction.encode();
 
         for _ in 0..retry_count {
             let response = DefaultRequestBuilder::new(HttpMethod::POST, url.clone())
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1012,10 +1012,10 @@ impl Client {
         retry_count: u32,
         retry_in: Duration,
         pagination: Pagination,
-    ) -> Result<Option<SignedTransaction>> {
+    ) -> Result<Vec<SignedTransaction>> {
         let url = self
             .torii_url
-            .join(crate::config::torii::PENDING_TRANSACTION)
+            .join(crate::config::torii::PENDING_TRANSACTIONS)
             .expect("Valid URI");
         let pagination = pagination.into_query_parameters();
         let body = transaction.encode();
@@ -1030,11 +1030,11 @@ impl Client {
                 .send()?;
 
             if response.status() == StatusCode::OK {
-                let pending_transaction: Option<SignedTransaction> =
+                let pending_transactions: Vec<SignedTransaction> =
                     DecodeAll::decode_all(&mut response.body().as_slice())?;
 
-                if pending_transaction.is_some() {
-                    return Ok(pending_transaction);
+                if !pending_transactions.is_empty() {
+                    return Ok(pending_transactions);
                 }
                 thread::sleep(retry_in);
             } else {
@@ -1045,7 +1045,7 @@ impl Client {
                 ));
             }
         }
-        Ok(None)
+        Ok(vec![])
     }
 
     /// Find the original transaction in the local pending tx queue.
@@ -1058,7 +1058,7 @@ impl Client {
         transaction: &SignedTransaction,
         retry_count: u32,
         retry_in: Duration,
-    ) -> Result<Option<SignedTransaction>> {
+    ) -> Result<Vec<SignedTransaction>> {
         self.get_original_transaction_with_pagination(
             transaction,
             retry_count,
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1000,17 +1000,6 @@ impl Client {
         )
     }
 
-    /// Check if two transactions are the same. Compare their contents excluding the creation time.
-    fn equals_excluding_creation_time(
-        first: &TransactionPayload,
-        second: &TransactionPayload,
-    ) -> bool {
-        first.authority() == second.authority()
-            && first.instructions() == second.instructions()
-            && first.time_to_live() == second.time_to_live()
-            && first.metadata().eq(second.metadata())
-    }
-
     /// Find the original transaction in the pending local tx
     /// queue.  Should be used for an MST case.  Takes pagination as
     /// parameter.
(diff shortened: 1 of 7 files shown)


client/src/client.rs
```diff
@@ -1024,33 +1013,29 @@ impl Client {
         retry_in: Duration,
         pagination: Pagination,
     ) -> Result<Option<SignedTransaction>> {
+        let url = self
+            .torii_url
+            .join(crate::config::torii::PENDING_TRANSACTION)
+            .expect("Valid URI");
         let pagination = pagination.into_query_parameters();
+        let body = serde_json::to_vec(transaction.payload())
```
DCNick3, line 1021: It's weird that the API handler accepts JSON, but returns scale-encoded data.. I would expect an API handler to be consistent with itself.

Using scale codec for both request and response bodies seems fitting, as we encode all blockchain-related data with scale usually

torii/src/routing.rs
```diff
@@ -151,15 +152,19 @@ pub async fn handle_pending_transactions(
...
+) -> Result<Scale<Option<SignedTransaction>>> {
     let query_response = sumeragi.apply_wsv(|wsv| {
         queue
             .all_transactions(wsv)
             .map(Into::into)
             .paginate(pagination)
-            .collect::<Vec<_>>()
-        // TODO:
-        //.batched(fetch_size)
+            .find(|current_transaction: &SignedTransaction| {
```
DCNick3, line 162: This is much better than scanning all the transaction on client, but this is still doing a linear search of pending transactions. This might get slow in case there a lot of pending transactions.

Do we expect a large (>100) number of pending transactions? If yes, we can maintain a `HashMap` to speed up this search

DCNick3, line 162: What if there are multiple matching transactions? Is returning only one of them a sensible thing to do? (I don't know how the MST use this..)

VAmuzing: I do not have an answer regarding the size of the queue, although it should be important that it is indeed a queue that maintains an order. So we need some kind of `HashMap` that remembers insertion ordering.
I think that the comparison function itself is kinda slow (it compares strings in `AccountId`) and I don't really see any speed gain on such a small number of transactions (if it's indeed in range of 100-1000 or less). is there anyone who can help?

VAmuzing: The behavior is kinda copied from the client side. It was fine back there so I guessed that that is the way.

client/src/client.rs
VAmuzing: That's a good point, I also changed the type of the body so that it's similar to the other endpoints.

torii/src/routing.rs
mversic: > What if there are multiple matching transactions? Is returning only the first one of them a sensible thing to do? (I don't know how the MSTs use this..)

no, there shouldn't be. Transactions are deduplicated in the queue

data_model/src/transaction.rs
```diff
@@ -123,6 +123,19 @@ pub mod model {
...
+        pub fn equals_excluding_creation_time(
+            first: &TransactionPayload,
+            second: &TransactionPayload,
+        ) -> bool {
+            first.authority() == second.authority()
+                && first.instructions() == second.instructions()
+                && first.time_to_live() == second.time_to_live()
+                && first.metadata().eq(second.metadata())
+        }
+    }
```
mversic, line 137: move this code into `torii`. I'm not sure we want to have it in `data_model`. If we do we can always expand our public API

torii/src/routing.rs
mversic: you can make a wrapper over a tranasction and implement custom `Hash` and `PartialEq` and `PartialOrd` over it so that they exclude `creation_time`. Then you can put it into a `HashMap`. We don't need insertion ordering since transactions are guaranteed to be unique in the `queue`

mversic: I'm sorry, I was wrong here. Transctions can indeed be duplicated in the queue

torii/src/routing.rs
```diff
@@ -151,15 +152,19 @@ pub async fn handle_pending_transactions(
...
     sumeragi: SumeragiHandle,
     pagination: Pagination,
-) -> Result<Scale<Vec<SignedTransaction>>> {
+    transaction: SignedTransaction,
+) -> Result<Scale<Option<SignedTransaction>>> {
     let query_response = sumeragi.apply_wsv(|wsv| {
         queue
             .all_transactions(wsv)
             .map(Into::into)
             .paginate(pagination)
```
DCNick3, line 161: Hm, I don't think it makes sense to keep pagination in this endpoint. AFAIU, you can use it to specify a range in which to search for the transaction, which isn't really useful for the only function we use it for.

I would also suggest to remove the corresponding `get_original_transaction_with_pagination` on the client side, leaving all the implementation in `get_original_transaction`

config/src/torii.rs
```diff
@@ -62,7 +62,7 @@ pub mod uri {
     /// The web socket uri used to subscribe to blocks stream.
     pub const BLOCKS_STREAM: &str = "block/stream";
     /// Get pending transactions.
-    pub const PENDING_TRANSACTIONS: &str = "pending_transactions";
+    pub const PENDING_TRANSACTION: &str = "pending_transaction";
```
DCNick3, line 65: With the semantics changed to only return a single matching transaction, I think this endpoint should be renamed to `original_transaction` (open to other name suggestions). `pending_transaction` is very non-descriptive though..

VAmuzing: As we've kinda returned to how it was before but with a twist, do you think this is no longer applicable?

DCNick3: Not sure... It still filters transactions with no way to get all pending transactions, so the concern is still kind of applicable?

Other possible names: `matching_pending_transactions` or just `matching_transactions`

client/src/client.rs
```diff
@@ -1061,21 +1045,21 @@ impl Client {
                 ));
             }
         }
-        Ok(None)
+        Ok(vec![])
```
Erigara, line 1048:
```suggestion
Vec::new()
```

let's not use macros where not required

client/tests/integration/multisignature_transaction.rs
```diff
@@ -83,8 +83,10 @@ fn multisignature_transactions_should_wait_for_all_signatures() -> Result<()> {
     let instructions = [mint_asset];
     let transaction = client_2.build_transaction(instructions, UnlimitedMetadata::new())?;
     let transaction = client_2
-        .get_original_transaction(&transaction, 3, Duration::from_millis(100))?
-        .expect("Found no pending transaction for this account.");
+        .get_original_transactions(&transaction, 3, Duration::from_millis(100))?
+        .last()
```
Erigara, line 87: you could use `pop` here without cloning since you want just the last element

torii/src/routing.rs
Erigara: I agree that pagination here is a bit strange

Erigara (approved):

DCNick3 (approved):

PR #4213, opened 2024-01-18
[fix] #3962: Revoke associated tokens on entity unregistretration

### LINKED ISSUE #3962: [BUG]: Permission tokens aren't removed
Currently permission tokens aren't removed when there associated entity is removed.

This makes scenarios like this one possible: 

1. Account A register asset definition and give account B token to unregister it
2. Account A unregister asset definition
3. Account C register asset definition with the same id
4. Account B can unregister this asset definition

Erigara:
## Description

Hook into `visit_unregister_*` was added to remove associated permission tokens.

I've decided not to add this cleanup logic for the `Asset` entity because asset entity could be added and remove multiple times (for example when asset value reach 0).

Also i've made decision not to use macros here as imo it wouldn't bring much value here. 

### Linked issue

Closes #3962 

### Benefits

Tokens are cleaned up.

### Drawbacks

Uninteresting entities could became potentially expensive as we have to iterate through every token.


smart_contract/executor/src/default.rs
```diff
@@ -295,22 +439,34 @@ pub mod account {
...
+                    account_id: account_id.clone(),
+                };
+                can_unregister_user_account.is_owned_by(authority)
+            }
+        {
+            if executor.verdict().is_ok() {
+                if let Err(err) = isi.execute() {
+                    executor.deny(err);
+                }
+            }
```
mversic, line 458: I would prefer removing a token after it has been revoked from all the entities. Just move this piece of code after the `for loop`. Also in other cases

smart_contract/executor/src/default.rs
```diff
@@ -1168,6 +1465,24 @@ pub mod trigger {
...
+            .or_else(|_err| {
+                tokens::trigger::CanExecuteUserTrigger::try_from(permission.clone())
+                    .map(|permission| &permission.trigger_id == trigger_id)
+            })
+            .or_else(|_err| {
+                tokens::trigger::CanMintUserTrigger::try_from(permission.clone())
+                    .map(|permission| &permission.trigger_id == trigger_id)
+            })
+            .unwrap_or(false)
+    }
```
mversic, line 1485: can't you use `map_token` macro?

Erigara: i don't think so, `map_token` requires uniform `callback` for every token which is not the case here

mversic: how is it not the case? it looks uniform to me

smart_contract/executor/src/permission.rs
```diff
@@ -338,3 +338,20 @@ impl<T: Token> From<&T> for OnlyGenesis {
         Self
     }
 }
+
+/// Iterator over all accounts and theirs permission tokens
+pub fn accounts_permission_tokens() -> impl Iterator<Item = (AccountId, PermissionToken)> {
```
mversic, line 343: let's keep it private for now

smart_contract/executor/src/default.rs
Erigara: `map_token!` works with every default token and if parsed successfully call `callback` on it

in my case i only try small subset of tokens

mversic: got it, then at least use match statement like in the `map_token!` macro

mversic:
I find this implementation very brittle. If we add a new token we have to make sure that we also add the code for it's removal. I would suggest using `map_token!` and implement a callback on every token for every entity. Some of the callbacks won't do anything

Erigara:
> I find this implementation very brittle. If we add a new token we have to make sure that we also add the code for it's removal. 

I will try to figure out how we do this

Erigara:
@mversic take a look, imo this solution is less brittle

smart_contract/executor/src/default/tokens.rs
```diff
@@ -49,6 +49,28 @@ macro_rules! declare_tokens {
             };
         }
 
+        /// Enum with every default token
+        #[allow(clippy::enum_variant_names)]
+        pub(crate) enum AnyPermissionToken {
```
mversic, line 54:
```suggestion
        #[non_exhaustive]
        #[allow(clippy::enum_variant_names)]
        pub(crate) enum AnyPermissionToken {
```

should we make it `non_exhaustive`?

smart_contract/executor/src/default/tokens.rs
```diff
@@ -49,6 +49,28 @@ macro_rules! declare_tokens {
...
+                        let token = <$($token_path::)+$token_ty>::try_from(token)?;
+                        Ok(Self::$token_ty(token))
+                    } )+
+                    _ => Err(Self::Error::Id(token.definition_id().clone()))
+                }
+            }
+        }
+
         pub(crate) use map_token;
         pub(crate) use map_token_type;
```
mversic, line 75: since we have the enum now, do we need these anymore?

mversic:
> I've decided not to add this cleanup logic for the Asset entity because asset entity could be added and remove multiple times (for example when asset value reach 0).

I think this is fine. `Register`/`Unregister` asset operations aren't like others. If the same asset is unregistered and then registered again it will still be the same asset. So it's not a bug to have previously assigned permission token apply.

But what I think you should do is:
1. remove all permission tokens referring to assets when unregistering an asset definition
2. remove all permission tokens referring to accounts and asset definitions when unregistering a domain

Erigara:
> But as far as I can see, you have already done this

Yes, i already do this

smart_contract/executor/src/default/tokens.rs
Erigara: It won't help because this annotation doesn't work inside crate and this type available only within crate.

> Within the defining crate, non_exhaustive has no effect.

smart_contract/executor/src/default/tokens.rs
Erigara: I've managed to remove `map_token`, but `map_token_type` is still required to build schema

smart_contract/executor/src/default.rs
```diff
@@ -1002,29 +1293,27 @@ pub mod role {
...
-                                    &$token,
-                                    $authority,
-                                    $executor.block_height(),
-                                )
-                            {
-                                deny!($executor, error);
-                            }
+            if !is_genesis($executor) {
+                for token in role.permissions() {
+                    if let Ok(token) = AnyPermissionToken::try_from(token.clone()) {
```
mversic, line 1298: I would prefer if you used `match` here as was the case previously

smart_contract/executor/src/default/tokens.rs
```diff
@@ -29,27 +29,64 @@ use crate::permission::{self, Token as _};
...
+        /// Enum with every default token
+        #[allow(clippy::enum_variant_names)]
+        #[derive(Clone)]
+        pub(crate) enum AnyPermissionToken {
+            $(
+                $token_ty($($token_path::)+$token_ty),
+            )*
+        }
+
+        impl TryFrom<::iroha_smart_contract::data_model::permission::PermissionToken> for AnyPermissionToken {
```
mversic, line 47:
```suggestion
        impl TryFrom<crate::data_model::permission::PermissionToken> for AnyPermissionToken {
```

smart_contract/executor/src/default/tokens.rs
```diff
@@ -29,27 +29,64 @@ use crate::permission::{self, Token as _};
...
+            type Error = $crate::permission::PermissionTokenConversionError;
+
+            fn try_from(token: ::iroha_smart_contract::data_model::permission::PermissionToken) -> Result<Self, Self::Error> {
```
mversic, line 50:
```suggestion
            fn try_from(token: crate::data_model::permission::PermissionToken) -> Result<Self, Self::Error> {
```

smart_contract/executor/src/default/tokens.rs
```diff
@@ -29,27 +29,64 @@ use crate::permission::{self, Token as _};
...
+                        Ok(Self::$token_ty(token))
                     } )+
-                    _ => {}
+                    _ => Err(Self::Error::Id(token.definition_id().clone()))
                 }
+            }
+        }
 
-            };
+        impl From<AnyPermissionToken> for ::iroha_smart_contract::data_model::permission::PermissionToken {
```
mversic, line 61:
```suggestion
        impl From<AnyPermissionToken> for crate::data_model::permission::PermissionToken {
```

smart_contract/executor/src/default/tokens.rs
```diff
@@ -29,27 +29,64 @@ use crate::permission::{self, Token as _};
...
+        impl TryFrom<::iroha_smart_contract::data_model::permission::PermissionToken> for AnyPermissionToken {
+            type Error = $crate::permission::PermissionTokenConversionError;
+
+            fn try_from(token: ::iroha_smart_contract::data_model::permission::PermissionToken) -> Result<Self, Self::Error> {
+                match token.definition_id().as_ref() { $(
                     stringify!($token_ty) => {
-                        if let Ok(token) = <$($token_path::)+$token_ty>::try_from($token.clone()) {
-                            $callback!(token);
-                        }
+                        let token = <$($token_path::)+$token_ty>::try_from(token)?;
```
mversic, line 53: I don't like that we have `TryFrom<PermissionToken>` implemented for every token. The reason is that such construction is suboptimal. I would prefer removing it and inline the code here, as it's the only place where it's used

mversic (changes requested):

smart_contract/executor/src/default/tokens.rs
Erigara: @mversic what do you mean by suboptimal?

Inlining code here is not that simple since we generating `TryFrom` conversion for concrete permission inside procmacro. 

> it's the only place where it's used
is it possible that we will expect certain permission type? so that we can directly parse into specific type omitting `AnyPermissionToken`

mversic: > @mversic what do you mean by suboptimal?

how it's been used mostly was that a programmer would just try guessing to which type it can be converted when the match would have done the job in one jump

> Inlining code here is not that simple since we generating TryFrom conversion for concrete permission inside procmacro.

I believe you can literally just copy paste that code

Erigara: Currently `TryFrom` is used to implement `Token` trait (in `is_owned_by` method) where we convert `PermissionToken` to the concrete type.  

And `is_owned_by` use with concrete tokens, not only `AnyPermissionToken`.

I would suggest to extract this refactor into separate task.

mversic (approved):

smart_contract/executor/src/permission.rs
```diff
@@ -338,3 +338,20 @@ impl<T: Token> From<&T> for OnlyGenesis {
         Self
     }
 }
+
+/// Iterator over all accounts and theirs permission tokens
+pub(crate) fn accounts_permission_tokens() -> impl Iterator<Item = (AccountId, PermissionToken)> {
```
Arjentix, line 343: Oh, it takes so much time... We should consider optimizations in future

client/tests/integration/permissions.rs
```diff
@@ -381,3 +382,50 @@ fn permission_tokens_are_unified() {
...
         .expect_err("permission tokens are not unified");
 }
+
+#[test]
+fn associated_permission_tokens_removed_on_unregister() {
+    let (_rt, _peer, iroha_client) = <PeerBuilder>::new().with_port(11_240).start_with_runtime();
+    wait_for_genesis_committed(&[iroha_client.clone()], 0);
+
+    let bob_id: AccountId = "bob@wonderland".parse().expect("Valid");
+    let kingdom_id: DomainId = "kingdom".parse().expect("Valid");
```
Arjentix, line 392: I think we now discourage the use of `expect("Valid")` and prefer just `unwrap()`

smart_contract/executor/src/default.rs
```diff
@@ -192,21 +196,28 @@ pub mod domain {
...
+            }
+            || {
+                let can_unregister_domain_token = tokens::domain::CanUnregisterDomain {
+                    domain_id: domain_id.clone(),
+                };
+                can_unregister_domain_token.is_owned_by(authority)
+            }
+        {
+            for (owner_id, permission) in accounts_permission_tokens() {
+                if is_token_domain_associated(&permission, domain_id) {
```
Arjentix, line 212: I feel like it can (or should in future) be implemented using query filter

Arjentix (approved):

client/tests/integration/permissions.rs
Erigara: I was unaware of this, where it's stated?

smart_contract/executor/src/default.rs
Erigara: Permission tokens are opaque for core, how it should extract fields?

PR #4220, opened 2024-01-19
[feature] #4089: Add profiling support to iroha

### LINKED ISSUE #4083: [suggestion] We need a profiler for iroha
### Feature request

I have created the startup benchmark which is good at consistently measuring iroha2 wsv performance. And it is setup to allow for easy profiling by an external tool. But sofar I have had issues when I tried to use valgrind, so some other tool will have to be used instead.

### Motivation

The profiler would allow us to better understand what parts of the code can be optimized.

### Who can help?

_No response_

Erigara:
## Description

Add compilation profile `profiling` which preserve some debuginfo + force frame pointers (used by porfilers to symbolize functions in the stack trace).

Add dockerfile with profiling support (used glibc instead of musl to be able to integrate with parca).  

Add support to profile running wasm smartcontracts.

### Linked issue

Closes #4083 

### Benefits 

Data driven approach to optimization.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -266,6 +266,45 @@ tokio-console http://127.0.0.1:5555
 
 </details>
 
+### Profiling
+
+<details> <summary> Expand to learn ho to profile iroha. </summary>
+
+To optimize performance it's useful to profile iroha.
+
+To do that you should compile iroha with `profiling` profile and with `wasm_profiling` compilation flag: 
+
+```bash
+RUSTFLAGS="-C force-frame-pointers=on --cfg wasm_profiling" cargo build --profile profiling
+```
+
+Then start iroha and attach profiler of your choice to the iroha pid.
+
+Alternatively it's possible to build iroha inside docker with profiler support and profile iroha this way.
+
+```bash
+docker build -f Dockerfile.glibc --build-arg="PROFILE=profiling" --build-arg='RUSTFLAGS=-C force-frame-pointers=on --cfg wasm_profiling' -t iroha2:profiling .
+```
+
+E.g. using perf (available only on linux):
+
+```bash
+# to capture profile
+sudo perf record -g -p <PID>
+# to analyze profile
+sudo perf report
+```
+
+To be able to observe profile of the executor during iroha profiling, executor should be compiled without stripping symbols.
+It can be done by running:
+
+```bash
+# compile executor without optimizations
+cargo run --bin iroha_wasm_builder_cli -- build ./path/to/executor --outfile executor.wasm
+```
+
+</details>
+
 ## Style Guides
 
 Please follow these guidelines when you make code contributions to our project:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -253,3 +253,7 @@ members = [
 inherits = "release"
 strip = "symbols"
 lto = true
+
+[profile.profiling]
+inherits = "release"
+debug = "line-tables-only"
--- a/Dockerfile.glibc
+++ b/Dockerfile.glibc
@@ -1,16 +1,13 @@
-#base stage
-FROM archlinux:base-devel AS builder
+# base stage
+FROM debian:bookworm-slim AS builder
 
-# Force-sync packages, install archlinux-keyring, repopulate keys
-RUN pacman -Syy
-RUN pacman -S archlinux-keyring --noconfirm --disable-download-timeout
-RUN rm -rf /etc/pacman.d/gnupg/* && pacman-key --init && pacman-key --populate archlinux
+# install required packages
+RUN apt-get update -y && \
+     apt-get install -y curl build-essential mold
 
-# Install updates
-RUN pacman -Syu --noconfirm --disable-download-timeout
-
-# Set up Rust toolchain
-RUN pacman -S rustup mold wget --noconfirm --disable-download-timeout
+# set up Rust toolchain
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
 RUN rustup toolchain install nightly-2024-01-12
 RUN rustup default nightly-2024-01-12
 RUN rustup target add wasm32-unknown-unknown
@@ -19,15 +16,16 @@ RUN rustup component add rust-src
 # builder stage
 WORKDIR /iroha
 COPY . .
-RUN mold --run cargo build --target x86_64-unknown-linux-gnu --profile deploy
+ARG PROFILE="deploy"
+ARG RUSTFLAGS=""
+RUN RUSTFLAGS="${RUSTFLAGS}" mold --run cargo build --target x86_64-unknown-linux-gnu --profile "${PROFILE}"
 
 # final image
-FROM alpine:3.18
+FROM debian:bookworm-slim
 
-ENV  GLIBC_REPO=https://github.com/sgerrand/alpine-pkg-glibc
-ENV  GLIBC_VERSION=2.35-r1
+ARG PROFILE="deploy"
 ARG  STORAGE=/storage
-ARG  TARGET_DIR=/iroha/target/x86_64-unknown-linux-gnu/deploy
+ARG  TARGET_DIR=/iroha/target/x86_64-unknown-linux-gnu/${PROFILE}
 ENV  BIN_PATH=/usr/local/bin/
 ENV  CONFIG_DIR=/config
 ENV  IROHA2_CONFIG_PATH=$CONFIG_DIR/config.json
@@ -39,12 +37,9 @@ ENV  UID=1001
 ENV  GID=1001
 
 RUN  set -ex && \
-     apk --update add libstdc++ curl ca-certificates gcompat && \
-     for pkg in glibc-${GLIBC_VERSION} glibc-bin-${GLIBC_VERSION}; \
-         do curl -sSL ${GLIBC_REPO}/releases/download/${GLIBC_VERSION}/${pkg}.apk -o /tmp/${pkg}.apk; done && \
-     apk add --force-overwrite --allow-untrusted /tmp/*.apk && \
-     rm -v /tmp/*.apk && \
-     addgroup -g $GID $USER && \
+     apt-get update -y && \
+     apt-get install -y curl ca-certificates && \
+     addgroup --gid $GID $USER && \
      adduser \
      --disabled-password \
      --gecos "" \
--- a/core/src/smartcontracts/wasm.rs
+++ b/core/src/smartcontracts/wasm.rs
@@ -274,6 +274,10 @@ fn create_config() -> Result<Config> {
         .consume_fuel(true)
         .cache_config_load_default()
         .map_err(Error::Initialization)?;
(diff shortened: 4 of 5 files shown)


Cargo.toml
```diff
@@ -1,3 +1,5 @@
+cargo-features = [ "profile-rustflags" ]
```
Erigara, line 1: It's available only on `nighlty`, but we anyway compile iroha on `nighltly`.

Cargo.toml
```diff
@@ -253,3 +255,9 @@ members = [
 inherits = "release"
 strip = "symbols"
 lto = true
+
+[profile.profiling]
+inherits = "release"
+debug = "line-tables-only"
```
Erigara, line 261: This saves like 1GB of memory, compared to `debug = true`.

CONTRIBUTING.md
```diff
@@ -266,6 +266,31 @@ tokio-console http://127.0.0.1:5555
...
+
+To do that you should compile iroha with `profiling` profile and with `wasm-profiling` feature: 
+
+```bash
+cargo build --features "wasm_profiling" --profile profiling
+```
+
+Then start iroha and attach profiler of your choice to the iroha pid.
+
+E.g. using perf (available only on linux):
```
mversic, line 283: someone should add a section for profiling on mac

Cargo.toml
mversic: users don't necessarily need to compile on `nightly`, only building smart contracts seems to require `nightly`. Will this force us to always use `nightly` or just when using `profiling` profile?

Cargo.toml
mversic: seems reasonable. What about `debug = "limited"`?

core/src/smartcontracts/wasm.rs
```diff
@@ -274,6 +274,10 @@ fn create_config() -> Result<Config> {
         .consume_fuel(true)
         .cache_config_load_default()
         .map_err(Error::Initialization)?;
+    #[cfg(feature = "wasm_profiling")]
```
mversic, line 277:
```suggestion
    #[cfg(wasm_profiling)]
```

what I would prefer is to make it a compilation flag and then snuck it in with `rustflags` instead of defining a cargo feature

mversic: then you can just build iroha with:
`cargo build --profile profiling`

mversic (changes requested):

core/src/smartcontracts/wasm.rs
Erigara: oh, good point!

Cargo.toml
Erigara: Build will fail on stable.

```
cargo +stable build 
error: failed to parse manifest at `/Users/erigara/Projects/iroha/Cargo.toml`

Caused by:
  the cargo feature `profile-rustflags` requires a nightly version of Cargo, but this is the `stable` channel
  See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
  See https://doc.rust-lang.org/cargo/reference/unstable.html#profile-rustflags-option for more information about using this feature.
```

Erigara: Alternatively we could create `Makefile` or smt like this to pass all required flags.

mversic: then I would remove it from `Cargo.toml` and pass it on the CLI

Cargo.toml
Erigara: Didn't test with limited but when used with `parca` i didn't notice any difference between `true` and `"line-tables-only"`, so decided to select weakest possible option.

Dockerfile.glibc.profiling
```diff
@@ -0,0 +1,66 @@
...
+RUN pacman -S rustup mold wget --noconfirm --disable-download-timeout
+RUN rustup toolchain install nightly-2024-01-12
+RUN rustup default nightly-2024-01-12
+RUN rustup target add wasm32-unknown-unknown
+RUN rustup component add rust-src
+
+# builder stage
+WORKDIR /iroha
+COPY . .
+RUN mold --run cargo build --target x86_64-unknown-linux-gnu --features wasm_profiling --profile profiling
```
mversic, line 22: I assume profile is the only difference from `Dockerfile.glibc`?

Cargo.toml
mversic: makes sense

Dockerfile.glibc.profiling
Erigara: Yes

mversic: how about using [build arguments](https://docs.docker.com/build/guide/build-args/)?

Cargo.toml
Erigara: unfortunately ni, so in such case we have to pass them trough `RUSTFLAGS` on build

mversic: building Iroha on `nightly` must remain optional

Erigara: removed

core/src/smartcontracts/wasm.rs
Erigara: changed, hope copiler flags in cargo.toml would be stabilized soon

CONTRIBUTING.md
```diff
@@ -266,6 +266,37 @@ tokio-console http://127.0.0.1:5555
...
+### Profiling
+
+<details> <summary> Expand to learn ho to profile iroha. </summary> </details>
+
+To optimize performance it's useful to profile iroha.
+
+To do that you should compile iroha with `profiling` profile and with `wasm-profiling` feature: 
+
+```bash
+cargo build --features "wasm_profiling" --profile profiling
```
mversic, line 278: update this please

mversic, line 278:
```suggestion
cargo build --profile profiling
```

Dockerfile.glibc
```diff
@@ -19,15 +19,16 @@ RUN rustup component add rust-src
...
+ARG PROFILE="deploy"
+ARG RUSTFLAGS=""
+RUN RUSTFLAGS="-C target-feature=+crt-static ${RUSTFLAGS}" mold --run cargo build --target x86_64-unknown-linux-gnu --profile "${PROFILE}"
 
 # final image
 FROM alpine:3.18
 
-ENV  GLIBC_REPO=https://github.com/sgerrand/alpine-pkg-glibc
-ENV  GLIBC_VERSION=2.35-r1
+ARG PROFILE="deploy"
```
mversic, line 29: redefined here?

configs/peer/executor.wasm
mversic, line 1: what's the change in executor?

Erigara: no real change i guess it's just hook being run on my machine, i'm going to exclude executor change

Dockerfile.glibc
Erigara: Not redefined, but defined again in the different image

Erigara: I have to do it due to multistage build.

Erigara: > An ARG instruction goes out of scope at the end of the build stage where it was defined. To use an argument in multiple stages, each stage must include the ARG instruction.

From the [reference](https://docs.docker.com/engine/reference/builder/#arg).

Dockerfile.glibc
```diff
@@ -39,11 +40,7 @@ ENV  UID=1001
 ENV  GID=1001
 
 RUN  set -ex && \
-     apk --update add libstdc++ curl ca-certificates gcompat && \
```
Erigara, line 42: Compile statically so we no longer need this.

mversic (dismissed):

CONTRIBUTING.md
```diff
@@ -266,6 +266,44 @@ tokio-console http://127.0.0.1:5555
...
+### Profiling
+
+<details> <summary> Expand to learn ho to profile iroha. </summary> </details>
+
+To optimize performance it's useful to profile iroha.
+
+To do that you should compile iroha with `profiling` profile and with `wasm_profiling` compilation flag: 
+
+```bash
+RUSTFLAGS="-C force-frame-pointers=on --cfg wasm_profiling"cargo build --profile profiling
```
DCNick3, line 278: It would be very nice to specify the `force-frame-pointers` in the cargo configuration, but, unfortunately, this is not stable yet =( https://github.com/rust-lang/cargo/issues/10271

CONTRIBUTING.md
```diff
@@ -266,6 +266,44 @@ tokio-console http://127.0.0.1:5555
...
+sudo perf record -g -p <PID>
+# to analyze profile
+sudo perf report
+```
+
+It's also possible to profile iroha's wasm executor to do then unstriped version should be provided.
+
+```bash
+# compile executor without optimizations
+cargo run --bin iroha_wasm_builder_cli -- build ./peth/to/executor --outfile executor.wasm
```
DCNick3, line 302:
```suggestion
cargo run --bin iroha_wasm_builder_cli -- build ./path/to/executor --outfile executor.wasm
```

CONTRIBUTING.md
```diff
@@ -266,6 +266,44 @@ tokio-console http://127.0.0.1:5555
...
+E.g. using perf (available only on linux):
+
+```bash
+# to capture profile
+sudo perf record -g -p <PID>
+# to analyze profile
+sudo perf report
+```
+
+It's also possible to profile iroha's wasm executor to do then unstriped version should be provided.
```
DCNick3, line 298: I think it should be mentioned that profiling the wasm executor can be done using the same tools as profiling the iroha itself (maybe even link to [wasmtime's docs on the subject](https://docs.wasmtime.dev/examples-profiling.html)?).

It's not obvious that this will work, as it actually requires some support from wasmtime and mentioning this will lead to less confusion IMO.

Dockerfile.glibc
DCNick3: There are some [gotchas with statically linking glibc](https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged), but I guess iroha might happen to use the APIs that rely on the dynamically linked NSS services, so it works? Should be careful with this stuff

mversic: do we have to statically link glibc?

CONTRIBUTING.md
mversic: yes, this is what @Erigara did at first, but we have to keep iroha to build on `stable` so he removed it

Dockerfile.glibc
Erigara: Good catch, it’s not required for parca i just faced issues with building iroha non statically and use it with alpine image. I will try to use different glibc based container for both stages to make sure that glibc versions are compatible.

CONTRIBUTING.md
Erigara: I’ve ment that we could see what is going on inside executor while we run iroha. 
Not that we should invoke some external command.
I will try to make this part more clear.

mversic (approved):

DCNick3 (approved):

PR #4250, opened 2024-02-02
[feature] #4083: Add pprof-rs support

### LINKED ISSUE #4083: [suggestion] We need a profiler for iroha

Erigara:
## Description

Add torii endpoint to serve pprof profiles.

Changed debug to "limited" due to more detailed function names.

Added `-Z build-std` for profiling image in order to force frame pointers in the std as well.

### Linked issue

Closes #4083 

### Benefits

Able to profile iroha without any privileges like (root access of linux cap).

### Downsides

No kernel level traces.

Additional fragile dependency into our codebase.


Cargo.toml
```diff
@@ -256,4 +256,4 @@ lto = true
 
 [profile.profiling]
 inherits = "release"
-debug = "line-tables-only"
+debug = "limited"
```
mversic, line 259: so you've changed to `limited` now?

torii/src/lib.rs
```diff
@@ -279,6 +303,9 @@ pub enum Error {
     #[cfg(feature = "telemetry")]
     /// Failed to get Prometheus metrics
     Prometheus(#[source] eyre::Report),
+    #[cfg(feature = "profiling")]
```
mversic, line 306: hm, we've had issues with conditionally included fields previously

Cargo.toml
Erigara: Yes, i've compared profiles with `lines-tables-only` and `limited`.
And with `lines-tables-only` resulting profile looks very confusing because instead of `sumeragi::main_loop::run` funcion name would be just `run` which make it very confusing. 
Looks like i didn't noticed it before.

torii/src/lib.rs
Erigara: we already have `Prometheus` included conditionally, without conditional compilation this variant would be a dead code

mversic (approved):

torii/src/lib.rs
DCNick3: I don't think the serialization concerns are relevant here: this is an error type that is not encoded or exposed in schema.

DCNick3 (approved):
Huh, I never actually learned of the `build-std` trick. I guess this is why I always had missing frames in my rust programs when profiling

PR #4275, opened 2024-02-12
[refactor] #4174: Stop encoding X25519 keys as Ed25519

### LINKED ISSUE #4174: [suggestion] Stop encoding X25519 keys as Ed25519

Arjentix:
## Description

- Key exchanging trait updated with keys associated types
- X25519 keys are now used as a separate type
- `KeyGenOption` now accepts optional generic

### Linked issue

- Closes #4174 

### Benefits

More robust type-system, less error-handling, simpler ed25519 code

### Checklist


Erigara (dismissed):

crypto/src/lib.rs
```diff
@@ -113,14 +113,14 @@ impl FromStr for Algorithm {
...
-pub enum KeyGenOption {
+pub enum KeyGenOption<K = PrivateKey> {
     /// Use random number generator
     #[cfg(feature = "rand")]
     Random,
     /// Use seed
     UseSeed(Vec<u8>),
     /// Derive from private key
-    FromPrivateKey(Box<PrivateKey>),
+    FromPrivateKey(K),
```
mversic, line 123: this seems weird. There is a variant `FromPrivateKey` but you can parametrize it with something that's not a private key?

Arjentix: The idea was to parametrize it with the private key of concrete type related to the exact algorithm

mversic (dismissed):

mversic (approved):

Erigara (approved):

PR #4281, opened 2024-02-14
Changed "some-host" to "localhost" in example/config.docker

baziorek:
## Description
In one of our config file we had database default address: "some-postgres". Those sample files are for Iroha's beginners to run Iroha as fast as possible. Telling them to change sample configs to run Iroha IMO is not helping them. Also it is making our life harder, because many times somebody was asking about "I'm doing as in instruction and it is not working".

What is more - we have few config files in `example/` directory:
```
ls config*
config.docker  config.postgres.sample  config.sample  config-win.sample
```
in all of config files the host for DB is `localhost`, only in one of them (which I'm changing now) it is `some-postgres`.
- https://github.com/hyperledger/iroha/blob/88de1c6ca640eacf563bc24ae05c11dd14403c38/example/config-win.sample#L5
- https://github.com/hyperledger/iroha/blob/88de1c6ca640eacf563bc24ae05c11dd14403c38/example/config.postgres.sample#L7
- `config.sample` - this is using RocksDB instead of Postgres

### Linked issue

Closes #{issue_number} 

### Benefits
1. Easier for Iroha's beginners to run example
2. Less work for us with explaining to them where to change config files
3. Consequence in config files

### Checklist

### DIFF
--- a/example/config.docker
+++ b/example/config.docker
@@ -4,7 +4,7 @@
   "internal_port" : 10001,
   "database": {
     "type": "postgres",
-    "host": "some-postgres",
+    "host": "localhost",
     "port": 5432,
     "user": "postgres",
     "password": "mysecretpassword",


6r1d (approved):
That looks reasonable to me, thank you for the detailed description.

Issue #4116, opened 2023-12-06
[BUG] Response from /status Endpoint Not in Expected Binary Format


AlexStroke:
### OS and Environment

MacOS, bare metal

### GIT commit hash

0ef18feabeded1839226df8f9d5cdf0253b8fdef

### Minimum working example / Steps to reproduce

```bash
curl --location 'http://127.0.0.1:8080/status/' --header 'Accept: application/x-parity-scale'
```

### Actual result

The response is received in a non-binary JSON format

### Expected result

The response should be in a binary format, specifically encoded using the SCALE (Simple Concatenated Aggregate Little-Endian) encoding.

### Who can help to reproduce?

@AlexStroke

AlexStroke:
resolved in #3964

PR #4284, opened 2024-02-16
[feature] #4116: Add test for plain == scale encoded verification for status request

VAmuzing:
## Description

While checking #4116 I found helpful to verify that scale encoded response was equal to plain with this test.
It also might come in handy when we eventually move from warp to something like axum.

### Linked issue

Relates to  #4116 

### Checklist

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -176,12 +176,17 @@ pub struct StatusResponseHandler;
 
 impl StatusResponseHandler {
     fn handle(resp: &Response<Vec<u8>>) -> Result<Status> {
+        let slice = Self::handle_raw(resp)?;
+        serde_json::from_slice(slice).wrap_err("Failed to decode body")
+    }
+
+    fn handle_raw(resp: &Response<Vec<u8>>) -> Result<&Vec<u8>> {
         if resp.status() != StatusCode::OK {
             return Err(ResponseReport::with_msg("Unexpected status response", resp)
                 .unwrap_or_else(core::convert::identity)
                 .into());
         }
-        serde_json::from_slice(resp.body()).wrap_err("Failed to decode body")
+        Ok(resp.body())
     }
 }
 
@@ -1032,6 +1037,18 @@ impl Client {
         let req = self.prepare_status_request::<DefaultRequestBuilder>();
         let resp = req.build()?.send()?;
         StatusResponseHandler::handle(&resp)
+    }
+
+    /// Gets network status seen from the peer in scale encoding
+    ///
+    /// # Errors
+    /// Fails if sending request or decoding fails
+    pub fn get_status_scale_encoded(&self) -> Result<Vec<u8>> {
+        let req = self
+            .prepare_status_request::<DefaultRequestBuilder>()
+            .header(http::header::ACCEPT, "application/x-parity-scale");
+        let resp = req.build()?.send()?;
+        StatusResponseHandler::handle_raw(&resp).cloned()
     }
 
     /// Prepares http-request to implement [`Self::get_status`] on your own.
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -175,7 +175,7 @@ impl TransactionResponseHandler {
 pub struct StatusResponseHandler;
 
 impl StatusResponseHandler {
-    fn handle(resp: &Response<Vec<u8>>) -> Result<Status> {
+    pub(crate) fn handle(resp: &Response<Vec<u8>>) -> Result<Status> {
         let slice = Self::handle_raw(resp)?;
         serde_json::from_slice(slice).wrap_err("Failed to decode body")
     }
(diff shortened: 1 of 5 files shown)


telemetry/src/metrics.rs
```diff
@@ -32,8 +32,19 @@ impl Encode for Uptime {
     }
 }
 
+impl Decode for Uptime {
+    fn decode<I: parity_scale_codec::Input>(
+        input: &mut I,
+    ) -> Result<Self, parity_scale_codec::Error> {
+        let (secs, nanos) = <(Compact<u64>, u32)>::decode(input)?;
```
Erigara, line 39: Why only `secs` are `Compact`?

client/tests/integration/status_response.rs
```diff
@@ -0,0 +1,54 @@
...
+    lhs.peers == rhs.peers
+        && lhs.blocks == rhs.blocks
+        && lhs.txs_accepted == rhs.txs_accepted
+        && lhs.txs_rejected == rhs.txs_rejected
+        && lhs.view_changes == rhs.view_changes
+}
+
+#[test]
+fn json_and_scale_statuses_equality() -> Result<()> {
+    let (_rt, network, client) = Network::start_test_with_runtime(1, Some(10_945));
```
Erigara, line 19: Port is overlapping with `multisignature_transactions_should_wait_for_all_signatures`, please pick free port.

Erigara (changes requested):
Please pick free port

telemetry/src/metrics.rs
DCNick3: Comment from the Encode impl above:
```
// While seconds are rarely very large, nanos could be anywhere between zero and one billion,
// eliminating the profit of Compact
```

Erigara: Oh, missed it, thx for pointing out

client/tests/integration/status_response.rs
Erigara: @VAmuzing usually i just search for max port and than add 5 ports for every peer in this test, e.g. if previous port was `10_000` with 4 peers, next port for the test would be `10_000 + (4 + 1) * 5 = 10_025`

Erigara (approved):

client/tests/integration/status_response.rs
```diff
@@ -0,0 +1,54 @@
...
+    lhs.peers == rhs.peers
+        && lhs.blocks == rhs.blocks
+        && lhs.txs_accepted == rhs.txs_accepted
+        && lhs.txs_rejected == rhs.txs_rejected
+        && lhs.view_changes == rhs.view_changes
+}
+
+#[test]
+fn json_and_scale_statuses_equality() -> Result<()> {
+    let (_rt, network, client) = Network::start_test_with_runtime(1, Some(10_935));
```
Erigara, line 19: 10_925 is occupied with `unstable_network_stable_after_add_and_after_remove_peer` which establish network of 4 peers, so 10_935 is overlapping.

VAmuzing: Changed. Also changed peers to 2.
I noticed that there are lots of tests that are currently ignored due to slowness. Can reduction of n_peers help with that, what do you think?

0x009922 (commented):
Actually, why not use only SCALE in `iroha_client`'s `get_status` implementation? It's win-win, isn't it?

client/src/client.rs
```diff
@@ -1081,6 +1086,18 @@ impl Client {
...
+    ///
+    /// # Errors
+    /// Fails if sending request or decoding fails
+    pub fn get_status_scale_encoded(&self) -> Result<Vec<u8>> {
+        let req = self
+            .prepare_status_request::<DefaultRequestBuilder>()
+            .header(http::header::ACCEPT, "application/x-parity-scale");
+        let resp = req.build()?.send()?;
+        StatusResponseHandler::handle_raw(&resp).cloned()
+    }
```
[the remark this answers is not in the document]
VAmuzing: I thought of it as an option when I implemented a test.
Do you think that this function should return `SCALE`'d structure and for the outer function to decode it if necessary?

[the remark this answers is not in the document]
0x009922: No objections about tests and the code overall, it's great.

> Do you think that this function should return SCALE'd structure and for the outer function to decode it if necessary?

I think having only `fn get_status(&self) -> Result<Status>` is enough.

0x009922:
> If this feature is unused then we indeed can switch to SCALE only.

I think it is unused.

client/src/lib.rs
```diff
@@ -36,6 +40,16 @@ pub mod samples {
...
+
+    /// Gets network status seen from the peer in json format
+    ///
+    /// # Errors
+    /// Fails if sending request or decoding fails
+    pub fn get_status_json(client: &Client) -> Result<Status> {
+        let req = client.prepare_status_request::<DefaultRequestBuilder>();
+        let resp = req.build()?.send()?;
+        StatusResponseHandler::handle(&resp)
+    }
```
0x009922, line 52: Not sure if this method is needed at all in `iroha_client`..

0x009922 (approved):

Erigara (approved):

client/src/lib.rs
VAmuzing: I think it should be fine as it's only used for the testing purposes in the designated module.

0x009922 (approved):

Erigara (approved):

PR #4305, opened 2024-02-20
[refactor]: remove `Value` enum

mversic:
## Description

* break dependency of `Metadata::Value` and `Query::Output` on `Value`
* remove `Value` enum. It's useless after removal of expressions 

### Linked issue

Closes #{issue_number} 

### Checklist


data_model/src/lib.rs
s8sato, line 1: Any additional context or related PRs would be appreciated.
Is this part of the trend to leave evaluation of complex expressions to Wasm executables?

mversic: we removed expressions from `iroha_core`. First of all nobody was using them and they were just complicating the API which was a big complaint. 

But also, with the introduction of `executor` it's now possible for a user to implement a custom expression system in the executor that evaluates to ISI instructions. We don't have an example of this, but it's quite possible to introduce it

core/src/wsv.rs
```diff
@@ -1126,15 +1126,16 @@ impl WorldStateView {
     where
         I: iroha_primitives::CheckedOp + Copy,
         NumericValue: From<I> + TryAsMut<I>,
+        <NumericValue as TryAsMut<I>>::Error: Debug,
         eyre::Error: From<<NumericValue as TryAsMut<I>>::Error>,
     {
         let domain = self.domain_mut(&definition_id.domain_id)?;
         let asset_total_amount: &mut I = domain
-            .asset_total_quantities.get_mut(definition_id)
-            .expect("Asset total amount not being found is a bug: check `Register<AssetDefinition>` to insert initial total amount")
```
Erigara, line 1134: Why you removed info where to search for exact error place?

data_model/src/query/mod.rs
```diff
@@ -316,24 +337,206 @@ impl_query! {
...
+                .map(TryInto::try_into)
+                .collect::<Result<Vec<_>, _>>()
+                .map_err(|_e| Self::Error::default());
+        }
+
+        Err(Self::Error::default())
+    }
+}
+
+//impl<A: SmallArray> From<SmallVec<A>> for QueryOutputBox
```
Erigara, line 502: Commented code.

data_model/src/lib.rs
```diff
@@ -332,18 +318,64 @@ pub mod parameter {
             /// Unique [`Id`] of the [`Parameter`].
             pub id: ParameterId,
             /// Current value of the [`Parameter`].
-            #[getset(get = "pub")]
-            pub val: Box<Value>,
+            pub val: Box<ParameterValue>,
```
Erigara, line 321: Do we still need boxing here?

Erigara (changes requested):

client/tests/integration/triggers/data_trigger.rs
```diff
@@ -117,13 +118,18 @@ fn domain_scoped_trigger_must_be_executed_only_on_events_in_its_domain() -> Resu
...
+    let new_value = get_u32_asset_value(&test_client, asset_id);
     assert_eq!(new_value, prev_value + 1);
 
     Ok(())
 }
 
-fn get_asset_value(client: &client::Client, asset_id: AssetId) -> Result<u32> {
-    let asset = client.request(client::asset::by_id(asset_id))?;
-    Ok(*TryAsRef::<u32>::try_as_ref(asset.value())?)
+fn get_u32_asset_value(client: &client::Client, asset_id: AssetId) -> u32 {
```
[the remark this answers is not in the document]
mversic: I will wait for your PR to be merged first

data_model/src/lib.rs
mversic: good point

schema/gen/src/lib.rs
```diff
@@ -453,10 +549,8 @@ mod tests {
 
     #[test]
     fn no_extra_or_missing_schemas() {
-        let exceptions: HashSet<_> = RawGenesisBlock::schema()
-            .into_iter()
-            .map(|(type_id, _)| type_id)
-            .collect();
+        // NOTE: Skipping Box<str> until [this PR](https://github.com/paritytech/parity-scale-codec/pull/565) is merged
+        let exceptions: [core::any::TypeId; 1] = [core::any::TypeId::of::<Box<str>>()];
```
s8sato, line 553: Have you considered using static memory instead of heap for `ChainId` ? Not compatible with wasm?

client_cli/src/main.rs
```diff
@@ -55,33 +55,29 @@ impl MetadataArgs {
...
-            .map(Value::Bool)
-            .or_else(|_| s.parse::<Ipv4Addr>().map(Value::Ipv4Addr))
-            .or_else(|_| s.parse::<Ipv6Addr>().map(Value::Ipv6Addr))
-            .or_else(|_| s.parse::<NumericValue>().map(Value::Numeric))
-            .or_else(|_| s.parse::<PublicKey>().map(Value::PublicKey))
-            .or_else(|_| serde_json::from_str::<Value>(s).map_err(Into::into))
-            .map(|value| ValueArg { value })
+            .map(MetadataValue::Bool)
+            .or_else(|_| s.parse::<NumericValue>().map(MetadataValue::Numeric))
+            .or_else(|_| serde_json::from_str::<MetadataValue>(s).map_err(Into::into))
```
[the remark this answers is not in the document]
mversic: idk, that's how it was before my work. @0x009922 maybe you can tell us?

core/src/smartcontracts/isi/query.rs
```diff
@@ -12,16 +15,16 @@ pub trait Lazy {
...
 }
 
-/// Lazily evaluated equivalent of [`Value`]
-pub enum LazyValue<'a> {
-    /// Concrete computed [`Value`]
-    Value(Value),
-    /// Iterator over a set of [`Value`]s
-    Iter(Box<dyn Iterator<Item = Value> + 'a>),
+/// Lazily evaluated equivalent of [`Query::Output`]
+pub enum LazyQueryOutput<'a> {
```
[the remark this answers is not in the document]
mversic: there is no need. Where would you like to use it?

schema/gen/src/lib.rs
mversic: good point, we might try doing it

mversic: opened #4336

core/src/smartcontracts/isi/query.rs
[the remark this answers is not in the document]
s8sato: Then it could be `pub(crate)`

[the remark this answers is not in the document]
mversic: privacy policy is mostly irrelevant for `iroha_core` because it's not a part of the public API. Also, changing this to `pub(crate)` would be a big change all across the core crate because it's used in pub functions

data_model/src/lib.rs
```diff
@@ -281,6 +245,32 @@ pub mod parameter {
...
+        pub enum ParameterValueBox {
+            TransactionLimits(transaction::TransactionLimits),
+            MetadataLimits(metadata::Limits),
+            LengthLimits(LengthLimits),
+            Numeric(
+                #[skip_from]
+                #[skip_try_from]
+                Numeric,
+            ),
+        }
```
mversic, line 272: 1. I think that we might want to split `***Limits` into it's constituents

In any case the design here should be reconsidered

data_model/src/metadata.rs
```diff
@@ -1,27 +1,92 @@
...
+        Bytes(Vec<u8>),
+        Numeric(Numeric),
+        LimitedMetadata(Metadata),
+
+        Vec(
+            #[skip_from]
+            #[skip_try_from]
+            Vec<MetadataValueBox>,
+        ),
+    }
```
mversic, line 88: I think the values in metadata should be reconsidered. I'd be in favor of loosely typing it with `StringWithJson` and defer the parsing of the structure to executor as we do for permission tokens payload

data_model/src/lib.rs
Erigara: Currently values for `Parameter` used not by executor, but used by `wsv` and `sumeragi`. 
So i think having strong typing is beneficial.
But i don't like current design of `Parameter` and `SetParameter` isi. 
Imo it should be singular instruction for setting on-chain configuration of iroha.
I think we already have some issue about that.

primitives/derive/src/numeric.rs
```diff
@@ -6,8 +6,17 @@ use quote::quote;
...
 pub fn numeric_impl(input: TokenStream) -> Result<TokenStream> {
     let input = input.to_string();
-    let numeric = ::iroha_numeric::Numeric::from_str(&input)
+
+    let mut input = input.as_str();
+    for pattern in ["_u8", "_u16", "_u32", "_u64", "_u128"] {
+        if let Some(stripped) = input.strip_suffix(pattern) {
+            input = stripped;
+        }
+    }
```
Erigara, line 15: Why do you want to allow suffixes in macro input?
Imo it looks redundant and doesn't cover all cases (rust allows to write `10_u32` and `10u32`).

primitives/numeric/src/lib.rs
```diff
@@ -212,6 +212,22 @@ impl From<u64> for Numeric {
     }
 }
 
+impl TryFrom<Numeric> for u32 {
+    type Error = <u32 as TryFrom<Decimal>>::Error;
```
Erigara, line 216: We are leaking implementation details here.

primitives/derive/src/numeric.rs
mversic: should I remove it?

primitives/numeric/src/lib.rs
mversic: good point

client_cli/src/main.rs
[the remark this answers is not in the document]
0x009922: I guess it is because if it is deserialized directly using `Value` (or `MetadataValue` now), it requires specifying a full enum repr in CLI, e.g.:

```shell
iroha_client_cli ... --value='{"Bool":false}'
```

Doesn't look CLI friendly, does it? So, in https://github.com/hyperledger/iroha/pull/4163 @Asem-Abdelhady added this chain parsing which allows a simplified form:

```shell
iroha_client_cli ... --value=false
```

s8sato (approved):

primitives/numeric/src/lib.rs
mversic: fixed

s8sato (approved):

data_model/src/metadata.rs
mversic: #4353

primitives/numeric/src/lib.rs
```diff
@@ -72,6 +73,8 @@ pub enum NumericError {
     Negative,
     /// Malformed: expecting number with optional decimal point (10, 10.02)
     Malformed,
+    /// Represents a failure to convert to/from [`Numeric`] to the specified type
+    Conversion,
```
Erigara, line 77: I think it's better to use different error types for different code paths.
In such case we don't have to match against error not possible in given context.

Erigara (dismissed):

primitives/numeric/src/lib.rs
mversic: I made it into a separate `TryFromNumericError`

Erigara (dismissed):

s8sato (dismissed):

s8sato (approved):

Erigara (approved):

PR #4308, opened 2024-02-20
[refactor] #4229: Removed MST aggregation

### LINKED ISSUE #4229: remove MST aggregation from iroha
I think we should remove aggregation of MST signatures out of Iroha. Transaction signing should happen outside Iroha and only after all signatures have been collected should it be submitted to Iroha

This would also mean that we should remove `pending_transactions` endpoint

Stukalov-A-M:
## Description

 * Removed MST aggregation
 * Removed` hash_of_payload` method for Transactions
 *  Removed `merge_signatures` method
 *  Removed `pending_transactions` endpoint and derived methods
 
**Now not fully signed tx will be declined by queue with SignatureCondition error**

### Linked issue

 * Closes #4229 

### Checklist


client_cli/src/main.rs
```diff
@@ -241,29 +220,8 @@ fn submit(
...
-                .interact()
-                .wrap_err("Failed to show interactive prompt.")? => {
-                    original_transactions.into_iter().map(|transaction| {
-                        iroha_client.sign_transaction(transaction)
-                    }).collect()
-                }
-            _ => vec![tx],
-        }
-    };
+    let transactions = vec![tx];
```
VAmuzing, line 223: Initially we decided to defend ourselves from situation where there are multiple transactions returned by 'pending_transactions' endpoint. Currently there is only 1 transaction, so we shouldn't wrap it into Vec.

core/src/queue.rs
```diff
@@ -18,7 +18,7 @@ use crate::prelude::*;
 
 impl AcceptedTransaction {
     // TODO: We should have another type of transaction like `CheckedTransaction` in the type system?
-    fn check_signature_condition(&self, wsv: &WorldStateView) -> Result<MustUse<bool>> {
```
VAmuzing, line 21: Wasn't there a difference between `Error` and `MustUse(false)`? Or do want to use previous `Error` as just `false`

Stukalov-A-M: Well, I just wanted to process err in one place and as I noticed, this and derived methods have low chances to return a system error, but high chances to return business error which is SignatureCondition. So, I decided remove result from returning.

Stukalov-A-M:
@AlexStroke , could you pls. look at client_cli python tests. I removed check_mst flag from args, and assume that cli tests are dropped because of this.

core/src/queue.rs
```diff
@@ -537,27 +525,30 @@ mod tests {
...
+               let mut available = queue.collect_transactions_for_block(&wsv, max_txs_in_block);
+               assert_eq!(available.len(), 1);
+               let tx_from_queue = available.pop().expect("Checked that have one transactions");
+               // Check that transaction from queue pass signature check
+               assert!(matches!(
+                   tx_from_queue.check_signature_condition(&wsv),
+                   MustUse(true)
+               ));
+
+        */
```
Arjentix, line 551: Commented code. Either remove it or uncomment

core/src/tx.rs
```diff
@@ -120,11 +120,6 @@ impl AcceptedTransaction {
         Ok(Self(tx))
     }
 
-    #[inline]
-    pub(crate) fn merge_signatures(&mut self, other: Self) -> bool {
-        self.0.merge_signatures(other.0)
-    }
```
Arjentix, line 126: Looking at this removed code I remembered how much thing in Iroha were just *prior optimizations* which becomed useless...

no_std, expressions, query predicates and now mst

Erigara:
@Stukalov-A-M about pytests, that's right pytests use `--skip-mst-check` be default

core/src/queue.rs
```diff
@@ -275,7 +266,7 @@ impl Queue {
                 continue;
             }
             seen.push(hash);
-            if *tx.check_signature_condition(wsv).unwrap_or(MustUse(false)) {
+            if *tx.check_signature_condition(wsv) {
```
Erigara, line 269: This check look redundant now, because we add transaction in the queue only if it's `check_signature_condition` return `true`.

The only case when `check_signature_condition` became `false` if there is another transaction in the queue which changes signature check condition for the account. 
Not sure how we should handle this case.

Arjentix (dismissed):

VAmuzing (dismissed):

VAmuzing:
@AlexStroke @Stukalov-A-M please fix DCO before merging

core/src/queue.rs
Stukalov-A-M: Refactored by using `check_tx` matcher

VAmuzing (approved):

s8sato:
It seems to me that this change temporarily causes Iroha to lose the feature of multi-signature transactions.
Is that okay for you?
My concerns are noted here https://github.com/hyperledger/iroha/issues/4229#issuecomment-1958880901

Stukalov-A-M:
> It seems to me that with this change, Iroha will temporarily lose the feature of multi-signature transactions. Is that okay for you all? My concerns are noted here [#4229 (comment)](https://github.com/hyperledger/iroha/issues/4229#issuecomment-1958880901)

We have discussed it with @Arjentix exactly an idea to keep MST via triggers and it seems logical to divide the MST question resolving on 2 steps:
1. Remove current MST realization
2. Design a new solution for MST

core/src/queue.rs
```diff
@@ -157,23 +154,20 @@ impl Queue {
...
         if self.is_in_future(tx) {
             Err(Error::InFuture)
         } else if self.is_expired(tx) {
+            debug!("Transaction is expired");
             Err(Error::Expired)
         } else if tx.is_in_blockchain(wsv) {
+            debug!("Transaction is already in blockchain");
             Err(Error::InBlockchain)
+        } else if !tx.check_signature_condition(wsv).into_inner() {
+            debug!("Signature condition appeared");
```
Erigara, line 167: probably? 

```suggestion
            debug!("Signature condition failed");
```

core/src/queue.rs
```diff
@@ -237,10 +224,10 @@ impl Queue {
         Ok(())
     }
 
-    /// Pop single transaction from the queue. Record all visited and not removed transactions in `seen`.
+    /// Pop single transaction from the queue. Removes all transaction that fail the `tx_check`.
```
VAmuzing, line 227:
```suggestion
    /// Pop single transaction from the queue. Removes all transactions that fail the `tx_check`.
```

Erigara (approved):

VAmuzing (approved):

core/src/queue.rs
```diff
@@ -646,7 +615,7 @@ mod tests {
             query_handle,
         ));
         let queue = Queue::from_config(Config {
-            transaction_time_to_live: Duration::from_millis(200),
+            transaction_time_to_live: Duration::from_millis(300),
```
DCNick3, line 618: This change breaks this test for me, and I don't think this change can be right without changing the expected numbers of transactions collected

Erigara: Yes, i fixed this test in #4319

PR #4276, opened 2024-02-13
[ci]: Fix auto-labeling

s8sato:
## Description

- Fix: `api-changes` `config-changes` auto-labeling
- Refactor: merge the above and `iroha2` into one workflow
- Feature: notify someone specific based on the above auto-labels
- Feature: add another group of auto-labels to categorize PRs based on title

### Linked issue

- Closes #4278

### Benefits

- Make PR labels easier to handle
- Reduce workflows by one

### Checklist

- [x] I've read `CONTRIBUTING.md`
- [x] I've used the standard signed-off commit format (or will squash just before merging)
- [x] All applicable CI checks pass (or I promised to make them pass later)
- [ ] (optional) I've written unit tests for the code changes
- [ ] I replied to all comments after code review, marking all implemented changes with thumbs up

PR #4239, opened 2024-01-29
[refactor] #4161: rewrite config, _minimally_

0x009922:
## Description

This PR is a complete overhaul of Iroha & Client CLI configuration. It is based on the [Configuration Overhaul RFC](https://github.com/hyperledger/iroha/issues/2585).

Although this PR has many changes already, there are still big parts of the RFC to be completed separately.

### Configuration changes

- Change format: TOML, lower-case fields
- Restructure and rename parameters for users' convenience
- Localise chain-wide parameters in the `[chain_wide]` section to ease further transition: https://github.com/hyperledger/iroha/issues/4028
- Limit set of ENV aliases for the parameters
- Support composition via `extends` mechanism
- Split user config view from strongly-typed internal config representation
- Emit as many errors at once as possible: collecting all missing fields, all ENV parsing failures, and semantic errors.

### Subsequent changes

- **Revise sample configurations:**
    - remove alt branch configs (`configs/peer/lts|stable`, `configs/client/lts|stable`)
    - put `client.template.toml`, `peer.template.toml`, and `prometheus.template.yaml` in the root of `configs`. These configs are not for direct use, but rather a part of documentation.
    - move docker compose files into `configs/swarm`
    - move docker compose related configurations into `configs/swarm` as well (`client.toml`, `genesis.json`, and `executor.wasm`)
    - some tests & deployment docs relied to `/configs` directory to use client/peer configs. It was implied that the configs are actually workable and filled with everything needed. It was ambiguous. Now those places rely on `/configs/swarm` content.
- **Update test environment:**
    - `test_env.py`:
        - doesn't use ENVs to set configuration
        - generates a shared config, and individual configs (extending the shared one) for each peer
    - Pytests:
        - Split `CLIENT_CLI_DIR` (implying it has `iroha_client_cli` and `config.json` content) setting into `CLIENT_CLI_BINARY` and `CLIENT_CLI_CONFIG`
- **Iroha Swarm:**
    - Doesn't rely on `IROHA_CONFIG` variable anymore, and doesn't expect Iroha configuration to be present in the configs directory; relies only on `genesis.json`, and all necessary configs are set in ENV.
    - TODO: allow users to customise configuration? For all peers or for individual ones somehow?
        - maybe via `--config-shared=shared.toml --config-peer0=peer0.toml` CLI
- **Iroha Torii:**
[...]

Issue #1995, opened 2022-03-22
`serde` validation hook in deserialize


mversic:
It is quite often that we need to write custom `serde::Deserialize` implementations just to validate structure's invariants. If possible, we should write some sort of a hook into serde macro system that would take validation function and generate deserialize implementation for a struct. 

Also, maybe there is some crate offering this functionality or it can be done with serde.

mversic:
[this](https://docs.rs/serde_with/latest/serde_with/rust/display_fromstr/index.html) crate may be of some help

mversic:
`#[serde(from = "FromType")]` annotation might be of great use for this purpose

PR #4309, opened 2024-02-20
[feature] #4212: Prevent account registration without signatures

### LINKED ISSUE #4212: Registration of accounts without signatories
Currently we seem to allow this, but we don't allow burning the last signatory of an account (we have a test ensuring this `public_keys_cannot_be_burned_to_nothing`). Should we also prevent creation of `Account`/`NewAccount` without signatories. If so, then I propose to make modifications to all the constructors `Account::new`/`Decode`/`Deserialize` to only allow creation of `Accounts` with at least one signatory

This might be a potentially API changing modification

Arjentix:
### Linked issue

- Closes #4212 

### Benefits

More logical

### Checklist


client/benches/torii.rs
```diff
@@ -64,7 +64,7 @@ fn query_requests(criterion: &mut Criterion) {
     let create_domain = Register::domain(Domain::new(domain_id.clone()));
     let account_id = AccountId::new(domain_id.clone(), "account".parse().expect("Valid"));
     let (public_key, _) = KeyPair::generate().into();
-    let create_account = Register::account(Account::new(account_id.clone(), [public_key]));
+    let create_account = Register::account(Account::new(account_id.clone(), public_key));
```
mversic, line 67: hm, this is actually, quite a good compromise

data_model/src/account.rs
```diff
@@ -236,6 +260,39 @@ impl HasMetadata for NewAccount {
...
+}
+
+impl Decode for NewAccount {
+    fn decode<I: parity_scale_codec::Input>(
+        input: &mut I,
+    ) -> Result<Self, parity_scale_codec::Error> {
+        let candidate = NewAccountCandidate::decode(input)?;
+        Self::try_from(candidate).map_err(Into::into)
+    }
+}
```
mversic, line 294: put this into `mod candidate`

data_model/src/account.rs
```diff
@@ -214,15 +226,27 @@ impl Account {
...
+    /// Add additional signatures to account.
+    #[inline]
+    #[must_use]
+    pub fn with_additional_signatures(
+        mut self,
+        signatures: impl IntoIterator<Item = PublicKey>,
+    ) -> Self {
+        self.signatories.extend(signatures);
+        self
+    }
```
mversic, line 246: how about:

```suggestion
    pub fn add_signature(
        &mut self,
        signatures: PublicKey,
    ) -> Self {
        self.signatories.extend(signatures);
    }
```

mversic, line 246: how about:

```suggestion
    pub fn add_signature(
        &mut self,
        signatures: PublicKey,
    ) -> Self {
        self.signatories.extend(signatures);
    }
```

data_model/src/account.rs
```diff
@@ -109,15 +109,17 @@ pub mod model {
 
     /// Builder which should be submitted in a transaction to create a new [`Account`]
     #[derive(
-        DebugCustom, Display, Clone, IdEqOrdHash, Decode, Encode, Serialize, Deserialize, IntoSchema,
+        DebugCustom, Display, Clone, IdEqOrdHash, Encode, Serialize, Deserialize, IntoSchema,
     )]
     #[display(fmt = "[{id}]")]
     #[debug(fmt = "[{id:?}] {{ signatories: {signatories:?}, metadata: {metadata} }}")]
     #[ffi_type]
+    #[serde(try_from = "NewAccountCandidate")]
```
mversic, line 117: I would like to see this in other places, we have manually written `Deserialize` in many places. Tracked by #1995

data_model/src/account.rs
```diff
@@ -186,6 +185,19 @@ impl Account {
...
+    /// Construct builder for [`Account`] identifiable by [`Id`] containing random signature.
+    ///
+    /// Without the use of [`with_additional_signatures()`](NewAccount::with_additional_signatures)
+    /// this will make impossible to sign anything from behalf of this account.
+    ///
+    /// Usable for tests where there is no need to use the account directly.
+    #[inline]
+    #[must_use]
+    #[cfg(feature = "rand")]
+    pub fn new_with_random_signature(id: AccountId) -> <Self as Registered>::With {
```
mversic, line 197: I'd rather not have this in public API. Please remove it

mversic (changes requested):

data_model/src/account.rs
mversic, line 117: I think we need both `NewAccountCandidate` and `AccountCandidate`

data_model/src/account.rs
s8sato, line 1: If we want to guarantee at least one signatory at the `Account` level, we may want to make [account.remove_signatory](https://github.com/hyperledger/iroha/blob/64e9e311e57dabae9f01b8453d76c33e8d4b9c1a/data_model/src/account.rs#L211
) fallible and propagate it to [burn.execute](https://github.com/hyperledger/iroha/blob/64e9e311e57dabae9f01b8453d76c33e8d4b9c1a/core/src/smartcontracts/isi/account.rs#L158
)

mversic: this is a good point. `Account` must maintain it's own invariants

client/benches/torii.rs
Arjentix: Yes. I also took a look at some crates which provides things like `NonEmptyIterator` and `NonEmptyVec`, but it looks like having such dependencies doesn't worth it in this case

data_model/src/account.rs
mversic: let's then do this:

```suggestion
    #[must_use]
    pub fn add_signatory(mut self, signatory: PublicKey) -> Self {
        self.signatories.insert(signatory);
        self
    }
```

we use this pattern also

data_model/src/account.rs
Arjentix: Done, but keep in mind that `Account` and `NewAccount` fields are public

Arjentix:
> @BAStos525

Wow, something new... But why was it triggered? I haven't changed any configs...

s8sato:
https://github.com/hyperledger/iroha/pull/4309#issuecomment-1957724414

This is a new feature introduced by #4276 to automatically label and notify someone of specific code changes.
Currently `config-changes` reactions to any change under `configs/`, which may be an inappropriate pattern.
I guess `configs/*.template.*` would be correct one at the current directory structure.
Since things may have changed after #4239 I'd hear your opinion @0x009922 

Apart from that we observed the bot removed the manual `api-changes` label, which will be problematic.
2024-02-22T125101+090000
This is because with the current setting, it always checks the consistency between the changed files and [the labeler configuration](https://github.com/hyperledger/iroha/blob/effdf37d4b1fa6c37802128a78a43e30717b7147/.github/labeler.yml#L4).
I'll disable this synchronization in another PR #

data_model/src/account.rs
```diff
@@ -109,15 +101,17 @@ pub mod model {
...
     #[display(fmt = "[{id}]")]
     #[debug(fmt = "[{id:?}] {{ signatories: {signatories:?}, metadata: {metadata} }}")]
     #[ffi_type]
+    #[serde(try_from = "candidate::NewAccount")]
     pub struct NewAccount {
         /// Identification
         pub id: AccountId,
         /// Signatories, i.e. signatures attached to this message.
+        /// Cannot be empty, guaranteed by constructors.
         pub signatories: Signatories,
```
mversic, line 115:
```suggestion
       signatories: Signatories,
```

can we do this. Also for `Account`? This is so that we maintain the invariant that account has to have at least one signatory

mversic, line 115:
```suggestion
       signatories: Signatories,
```

can we do this. Also for Account? This is so that we maintain the invariant that account has to have at least one signatory

mversic (changes requested):

Arjentix:
> Currently config-changes reacts to any change under configs/, which may be an inappropriate pattern.

Yes, we change `executor.wasm` in pretty much every PR. And I'm sure it doesn't affect the way we deploy Iroha.

s8sato (approved):

mversic (approved):

Issue #4028, opened 2023-11-03
[discussion] Align configuration-related instructions with the RFC


0x009922:
## Description

According to the Configuration Overhaul RFC (#2585), the instructions set related to setting chain-wide configuration parameters should be redefined. For more details and reasoning behind it please read the document.

The goal of this issue is to work on the design first, without implementing it. We should establish the design simultaneously with the updating of the configuration reference (https://github.com/hyperledger/iroha-2-docs/issues/392). Only after the all parts of the reference are consistent and "good enough", we will start implementing it.

### Set of chain-wide parameters

- `sumeragi.block_time`
- `sumeragi.commit_time`
- `sumeragi.max_transactions_in_block`
- `wsv.transaction_limits`
- `wsv.identifier_length_limits`
- `wsv.domain_metadata_limits`
- `wsv.account_metadata_limits`
- `wsv.asset_definition_metadata_limits`
- `wsv.asset_metadata_limits`
- `wsv.wasm_runtime.fuel_limit`
- `wsv.wasm_runtime.memory_limit`

### Thoughts

Current `NewParameter` and `SetParameter` instructions accept any kind of key-value pair. There is a set of "magic keys" which are mapped into config parameters.

The core of Iroha has a limited set of parameters. Therefore, I propose to hard-code this set into the data model, and introduce an instruction to set/update parameters using this strictly defined set.

Notes:

- Remember that chain-wide parameters will be set **exclusively** via ISIs, i.e. they will be excluded from local peer configurations.
- Format: how such an instruction will look like?
- Genesis & Default: these parameters should be set in genesis. Will Iroha have default values for them? If not, then all parameters should be set in genesis.
- Naming & Nesting: should namespaces be flattened into a single namespace for simplicity? E.g. not `wsv.wasm_runtime.fuel_limits`, but just `wasm_fuel_limits`. Also, should the naming be aligned with naming of parameters in local configuration file? (or even in env?)
- Conveniences: according to the RFC, the config file will have some conveniences like setting durations not only as a number of millis, but also as a human-readable string like `"2s 300ms"`. Will the instruction also support it? I think yes, for consistency.

How the instruction might look like? I don't have an idea how to do it better.

A structure might look like:

```rust
enum Instruction {
    // ...
    Configure(Parameters)
}

struct Parameters {
    block_time: Option<Duration>,
    commit_time: Option<Duration>,
    wsv_asset_metadata_limits: Option<MetadataLimits>,
    // ...
}
```

[...]

PR #4310, opened 2024-02-21
[refactor] #3901: delete NewParameter isi and use only SetParameter

### LINKED ISSUE #3901: [suggestion] Move `NewParameter` into Executor

Asem-Abdelhady:
## Description
`NewParameter` is only used during genesis generation. it does not make sense to have it as ISI. Now the admin will generate the parameters during configuration and he can give permessions for other users to allow them to put parameters. 

### Linked issue

Closes #3901

### Checklist


configs/swarm/genesis.json
```diff
@@ -127,43 +127,43 @@
...
+        "SetParameter": "?WSVAccountMetadataLimits=1048576,4096_ML"
       },
       {
-        "NewParameter": "?WSVDomainMetadataLimits=1048576,4096_ML"
+        "SetParameter": "?WSVDomainMetadataLimits=1048576,4096_ML"
       },
       {
-        "NewParameter": "?WSVIdentLengthLimits=1,128_LL"
+        "SetParameter": "?WSVIdentLengthLimits=1,128_LL"
       },
```
mversic, line 155: can we drop the `WSV` prefix? what is it's use?

configs/swarm/genesis.json
```diff
@@ -188,5 +188,5 @@
       }
     ]
   ],
-  "executor": "./executor.wasm"
+  "executor": "./configs/swarm/executor.wasm"
```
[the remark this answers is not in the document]
Asem-Abdelhady: i didn't generate inline in the last amend let me fix this

mversic:
we need a test that shows how we can create a new parameter in executor

configs/swarm/genesis.json
mversic: @0x009922

configs/swarm/genesis.json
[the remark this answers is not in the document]
0x009922: The executor path is relative to the genesis file, so it should be `./executor.wasm`

configs/swarm/genesis.json
0x009922: Well, if we are speaking in the context of #4028, then... `SetParameter` should not accept some magically constructed string. It should accept a hard-coded-into-data-model enum of possible on-chain parameters.

As for prefix, I don't think it is needed. In the issue there are comments about naming.

mversic: created #4325

mversic:
> we need a test that shows how we can create a new parameter in executor and then update it's value

let's say that we want to define a parameter like `max_accounts_per_domain` in the executor and every time we register an account we also check for this parameter

PR #4333, opened 2024-02-27
fix: Don't prematurely execute time-triggers before their `start` timestamp

### LINKED ISSUE #3262: [suggestion] Time trigger first execution time
### Feature request

We want to execute the time trigger in the first block that comes after `start` but before the `start+period` time. 

### Motivation

@arndey Detected that the current implementation is somewhat misleading. As we want for our triggers to be final and stable we want to fix the implemetation sooner rather than later. 

### Who can help?

@appetrosyan @Arjentix @arndey

Arjentix:
## Description

While original issues doesn't state it clearly I think that the problem was premature execution of time triggers.
So that trigger could be executed even before `start`. I fixed this.

- Now it's forbidden to register schedule-based time-trigger with `start` point before `latest_block_timestamp + consensus_estimation` time, where `consensus_estimation = 4 secs` and is a hardcoded constant. Everithing before this timestamp is considered already analyzed by Iroha. In other words -- **time-triggers in the past are forbidden**.
- As a consequence it's forbidden to register such trigger in genesis
- Now premature time-trigger look up is forced to not to execute trigger `start` point is in future
- This behavior is tested by a fixed `change_asset_metadata_after_1_sec` test

Forbidding time-triggers in the past was the simpliest solution and I actually think it was making sence to implement that in the initial solution.

### Linked issue

- Closes #3262 

### Benefits

- Less confusion about time-trigger execution time
- No more painful triggers in the past

### Checklist

### DIFF
--- a/client/tests/integration/triggers/time_trigger.rs
+++ b/client/tests/integration/triggers/time_trigger.rs
@@ -3,10 +3,15 @@ use std::{str::FromStr as _, time::Duration};
 use eyre::Result;
 use iroha_client::{
     client::{self, Client, QueryResult},
-    data_model::{prelude::*, transaction::WasmSmartContract},
+    data_model::{
+        asset::AssetId,
+        events::pipeline::{BlockEventFilter, BlockStatus},
+        prelude::*,
+        transaction::WasmSmartContract,
+        Level,
+    },
 };
 use iroha_config::parameters::defaults::chain_wide::CONSENSUS_ESTIMATION as DEFAULT_CONSENSUS_ESTIMATION;
-use iroha_data_model::events::pipeline::{BlockEventFilter, BlockStatus};
 use iroha_logger::info;
 use test_network::*;
 use test_samples::{gen_account_in, ALICE_ID};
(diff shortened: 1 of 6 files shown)


client/tests/integration/triggers/time_trigger.rs
s8sato, line 1: Can we say we confirmed the following points:

1. the trigger was successfully registered
2. interval (time-event) and schedule (time-event acceptor points) matched
3. but the trigger was not executed because it was premature
4. now that it is past the `schedule.start` time and other conditions met, the trigger was successfully executed

I think at least one more query will be required

s8sato, line 1: I cannot read the role of `PERIOD` here from its name

data_model/src/events/time.rs
```diff
@@ -326,10 +339,10 @@ mod tests {
...
-            let schedule = Schedule::starting_at(Duration::from_secs(TIMESTAMP))
-                .with_period(Duration::from_secs(10));
-            let since = Duration::from_secs(TIMESTAMP + 35);
-            let length = Duration::from_secs(4);
+            let schedule = Schedule::starting_at(Duration::from_secs(TIMESTAMP - 5))
+                .with_period(Duration::from_secs(20));
+            let since = Duration::from_secs(TIMESTAMP);
+            let length = Duration::from_secs(5);
             let interval = TimeInterval { since, length };
             assert_eq!(count_matches_in_interval(&schedule, &interval), 0);
```
s8sato, line 347: Does this change guarantee anything new?
In each case before and after, the intersection of schedule and interval is empty, regardless of where `schedule.start` is

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -115,7 +117,7 @@ fn change_asset_metadata_after_1_sec() -> Result<()> {
         &mut test_client,
         &account_id,
         Duration::from_secs(1),
-        usize::try_from(PERIOD.as_millis() / DEFAULT_CONSENSUS_ESTIMATION.as_millis() + 1)?,
+        2, // One block for trigger registration and one for execution
```
VAmuzing, line 120: What if we introduce something like scopes to show why it's equal to 2? And have a variable that we increment?
I see that it's a test but that value might cause confusions even with the commentary

data_model/src/events/time.rs
Erigara: I think this tests isn't testing anything new.
In each case two interval is located between two schedule points.
I'm going to revert it.

client/tests/integration/triggers/time_trigger.rs
Erigara: Going to remove it

client/tests/integration/triggers/time_trigger.rs
Erigara: No longer relevant.

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -96,46 +101,53 @@ fn time_trigger_execution_count_error_should_be_less_than_15_percent() -> Result
...
 
-    let value = test_client.request(FindAssetDefinitionKeyValueByIdAndKey {
-        id: asset_definition_id,
-        key,
+    let after_registration_quantity = test_client.request(FindAssetQuantityById {
+        id: asset_id.clone(),
     })?;
-    assert_eq!(value, numeric!(3).into());
+    // Schedule start was in the future so trigger isn't executed
+    assert_eq!(init_quantity, after_registration_quantity);
```
s8sato, line 137: To conclude that the trigger execution was filtered out, another block than one for registering trigger would need to be committed

data_model/src/events/time.rs
```diff
@@ -144,14 +144,46 @@ impl EventFilter for TimeEventFilter {
...
+                    // ---------[--[-|-)--)-----
+                    //         p1  c1 s p2 c2
+                    //
+                    // Schedule start is after current block (c1).
+                    // In this case event won't match and it will be handled in the next block.
+                    let since = if Range::from(prev).contains(&schedule.start) {
+                        schedule.start
+                    } else {
+                        prev.since + prev.length
+                    };
```
s8sato, line 182: great pattern analysis and description

client/tests/integration/triggers/time_trigger.rs
Erigara: Added one more block

s8sato (approved):

PR #4341, opened 2024-03-04
[refactor] #4339: refactor `iroha_crypto`, update API

### LINKED ISSUE #4339: Adopting `iroha_crypto` for WASM in `iroha-javascript` - feedback
While rewriting Crypto WASM in `iroha-javascript` using original `iroha_crypto` crate from this repo (https://github.com/hyperledger/iroha-javascript/issues/186), I faced a few issues with `iroha_crypto` API. Some of them are critical and block further process.

Here are my findings:

- **(critical)** Missing `Signature::from_raw`. While there are `PublicKey::from_raw` and `PrivateKey::from_raw` methods, there is no such for `Signature`.
    - Why it is needed:
        - to be consistent with other structures
        - to allow reproducing the `Signature` to then verify payloads. For example, in JS signatures might be represented as data-model objects. In order to translate such representation into crypto, `from_raw` is needed
    - If there is some _reason_ why this method "should not be exposed from `Signature`", I would remind that `Signature` could be still reproduced from raw using JSON deserialization. This way is suboptimal.
- **(critical)** `Signature`'s serialized form (i.e. JSON) contains `payload` as an array of bytes. It is inconsistent with `PrivateKey`'s representation, where `payload` is a hex-encoded string. Consider (de-)serializing `payload` in `Signature` as hex too.
- `Multihash` is no longer visible in Public API. However, `PublicKey`'s (de-)serialization and `FromStr`/`Display` are still based on `Multihash` parsing. This behaviour isn't documented. Apart from mentioning it in the docs, it would be nice to have doctests showing parsing from a multihash string
- `Signature::payload` - confusing documentation:
  https://github.com/hyperledger/iroha/blob/fbdd5914cdc3f792c73a15786b3a8447f14ccaee/crypto/src/signature/mod.rs#L61
- `iroha_crypto::error::Error` - should probably be exposed as `iroha_crypto::Error`, or at least as `iroha_crypto::prelude::Error`.
- `KeyPair::new` effectively has semantics of `from_raw` methods from other structs. Consider renaming.

0x009922:
## Description

See linked issue.

### Linked issue

Closes #4339

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -111,10 +111,24 @@ impl FromStr for Algorithm {
     }
 }
 
-/// Options for key generation
-#[cfg(not(feature = "ffi_import"))]
-#[derive(Debug, Clone)]
-pub enum KeyGenOption<K = PrivateKey> {
+/// Needed since it couldn't be represented in FFI
+#[derive(Clone)]
+#[cfg_attr(not(feature = "ffi_import"), derive(Debug))]
+enum KeyPairGenConfigInner {
+    /// Use random number generator
+    #[cfg(feature = "rand")]
+    Random { algorithm: Algorithm },
+    /// Use seed
+    UseSeed { seed: Vec<u8>, algorithm: Algorithm },
+    /// Derive from a private key
+    FromPrivateKey(PrivateKey),
+}
+
+/// While [`KeyPairGenConfig`] is for external users, this structure is for internal use
+/// to pass it to a specific algorithm.
+/// Parameter `K` is intended to be set as a `PrivateKey` type of that specific algorithm.
+#[derive(Debug)]
+pub enum KeyPairGenOption<K> {
     /// Use random number generator
     #[cfg(feature = "rand")]
     Random,
@@ -124,85 +138,46 @@ pub enum KeyGenOption<K = PrivateKey> {
     FromPrivateKey(K),
 }
 
-mod config_algorithm_state {
-    use crate::Algorithm;
-
-    #[derive(Copy, Clone, Debug)]
-    pub struct Default;
-
-    impl From<Default> for Algorithm {
-        fn from(_: Default) -> Self {
-            use std::default::Default;
-            Self::default()
-        }
-    }
-
-    #[derive(Copy, Clone, Debug)]
-    pub struct Set(pub Algorithm);
-
-    impl From<Set> for Algorithm {
-        fn from(Set(value): Set) -> Self {
-            value
-        }
-    }
-}
-
 ffi::ffi_item! {
-    /// Configuration of key generation
+    /// Configuration for key generation
     #[derive(Clone)]
-    #[cfg_attr(not(feature="ffi_import"), derive(Debug))]
-    pub struct KeyGenConfiguration<A> {
-        /// Options
-        key_gen_option: KeyGenOption,
-        /// Algorithm
-        algorithm: A,
+    #[cfg_attr(not(feature = "ffi_import"), derive(Debug))]
+    pub struct KeyPairGenConfig{
+        inner: KeyPairGenConfigInner
     }
 }
 
 #[ffi_impl_opaque]
-impl KeyGenConfiguration<config_algorithm_state::Default> {
-    /// Construct using random number generation with [`Ed25519`](Algorithm::Ed25519) algorithm
+impl KeyPairGenConfig {
+    /// Generate using random number generation with a specified algorithm
     #[cfg(feature = "rand")]
     #[must_use]
-    pub fn from_random() -> Self {
+    pub fn from_random(algorithm: Algorithm) -> Self {
         Self {
-            key_gen_option: KeyGenOption::Random,
-            algorithm: config_algorithm_state::Default,
+            inner: KeyPairGenConfigInner::Random { algorithm },
         }
     }
 
-    /// Construct using seed with [`Ed25519`](Algorithm::Ed25519) algorithm
+    /// Construct using a seed with a specified algorithm
     #[must_use]
-    pub fn from_seed(seed: Vec<u8>) -> Self {
+    pub fn from_seed(seed: Vec<u8>, algorithm: Algorithm) -> Self {
         Self {
-            key_gen_option: KeyGenOption::UseSeed(seed),
-            algorithm: config_algorithm_state::Default,
+            inner: KeyPairGenConfigInner::UseSeed { seed, algorithm },
         }
     }
 
-    /// With algorithm
+    /// Construct the pair from a private key. Infers the algorithm from the key's one.
     #[must_use]
-    pub fn with_algorithm(
-        self,
-        algorithm: Algorithm,
-    ) -> KeyGenConfiguration<config_algorithm_state::Set> {
-        KeyGenConfiguration {
-            key_gen_option: self.key_gen_option,
-            algorithm: config_algorithm_state::Set(algorithm),
+    pub fn from_private_key(key: PrivateKey) -> Self {
+        Self {
+            inner: KeyPairGenConfigInner::FromPrivateKey(key),
         }
     }
-}
 
-impl KeyGenConfiguration<config_algorithm_state::Set> {
-    /// Construct using private key with [`Ed25519`](Algorithm::Ed25519) algorithm
+    /// Generates a key pair
     #[must_use]
-    pub fn from_private_key(private_key: impl Into<PrivateKey>) -> Self {
-        let key = private_key.into();
-        let algorithm = key.algorithm();
-        Self {
-            key_gen_option: KeyGenOption::FromPrivateKey(key),
-            algorithm: config_algorithm_state::Set(algorithm),
-        }
+    pub fn generate(self) -> KeyPair {
+        KeyPair::generate_with_config(self)
     }
 }
 
@@ -220,12 +195,10 @@ ffi::ffi_item! {
 }
 
 impl KeyPair {
-    /// Generates a pair of Public and Private key with [`Algorithm::default()`] selected as generation algorithm.
+    /// Generates a random key pair using [`Algorithm::default()`]
     #[cfg(feature = "rand")]
     pub fn generate() -> Self {
-        Self::generate_with_configuration(
-            KeyGenConfiguration::from_random().with_algorithm(Algorithm::Ed25519),
-        )
+        Self::generate_with_config(KeyPairGenConfig::from_random(Algorithm::default()))
     }
 }
 
@@ -239,7 +212,7 @@ impl KeyPair {
     /// Construct a [`KeyPair`]
     ///
     /// # Errors
-    /// If public and private key don't match, i.e. if they don't make a pair
+    /// If public and private keys don't match, i.e. if they don't make a pair
     pub fn new(public_key: PublicKey, private_key: PrivateKey) -> Result<Self, Error> {
         let algorithm = private_key.algorithm();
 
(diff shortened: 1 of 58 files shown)


crypto/src/signature/mod.rs
```diff
@@ -633,4 +658,27 @@ mod tests {
         }
         // From the above we can conclude that `SignatureWrapperOf` have consistent behavior for `HashSet` and `BTreeSet`
     }
+
+    #[test]
+    fn signature_serialized_representation() {
+        let input = json!({
+            "public_key": "e701210312273E8810581E58948D3FB8F9E8AD53AAA21492EBB8703915BBB565A21B7FCC",
```
DCNick3, line 665: The fact that pubkey is serialized as uppercase (mostly?), but payload is lowercase is not very nice..

Can we make it consistent?

0x009922: This goes back to https://github.com/hyperledger/iroha/pull/3316. TBH, I don't understand how this

> improves private key validation

@mversic, could you comment on that?

mversic: > This goes back to https://github.com/hyperledger/iroha/pull/3316. TBH, I don't understand how this
> 
> > improves private key validation
> 
> @mversic, could you comment on that?

This was referring to something else. Uppercase/lowercase has no influence on validation

I would prefer making it all uppercase

crypto/src/signature/mod.rs
```diff
@@ -82,10 +81,32 @@ impl Signature {
         }
     }
 
+    /// Creates new signature from its raw payload and public key.
+    ///
+    /// **This method does not sign the payload.** Use [`Signature::new`] for this purpose.
+    ///
+    /// This method exists to allow reproducing the signature in a more efficient way than through
+    /// deserialization.
+    pub fn from_raw(public_key: PublicKey, payload: &[u8]) -> Self {
```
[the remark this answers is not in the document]
mversic: please rename to `from_bytes` elsewhere also

crypto/src/lib.rs
```diff
@@ -697,16 +728,9 @@ impl<'de> Deserialize<'de> for PrivateKey {
...
-            payload: String,
-        }
-
-        // NOTE: Verify that private key is valid
-        let private_key = PrivateKeyCandidate::deserialize(deserializer)?;
-        Self::from_hex(private_key.algorithm, private_key.payload.as_ref())
-            .map_err(D::Error::custom)
+        let PrivateKeySerialized { algorithm, payload } =
+            PrivateKeySerialized::deserialize(deserializer)?;
+        Self::from_raw(algorithm, &payload).map_err(D::Error::custom)
```
mversic, line 733: check out `serde(try_from` in `data_model/src/account.rs`. Is that applicable in this case?

0x009922:
Okay, so what's next:

- rename to `from_bytes` anyway
- change serialization to uppercase everywhere
- create an issue with a bug of `KeyGenConfiguration::from_private_key()`

crypto/src/lib.rs
0x009922: Thanks for pointing this out. `try_from` and `into` work perfectly here.

crypto/src/signature/mod.rs
0x009922: Changed `PrivateKey`'s and `Signature`'s `payload` serialized hex to uppercase.

I haven't bothered to updated every single hex in the workspace to uppercase accordingly. Hope it's fine.

crypto/src/signature/mod.rs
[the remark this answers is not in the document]
0x009922: Done

0x009922:
I've updated `KeyGenConfiguration` using type-state-pattern, but it fails in checks and with ffi features enabled.

Will address it next. Probably, will try to simplify the builder yet remain it type-safe.

mversic (approved):

DCNick3 (dismissed):

crypto/src/lib.rs
```diff
@@ -119,31 +120,54 @@ pub enum KeyGenOption<K = PrivateKey> {
...
 
+mod config_algorithm_state {
+    use crate::Algorithm;
+
+    #[derive(Copy, Clone, Debug)]
+    pub struct Default;
+
+    impl From<Default> for Algorithm {
+        fn from(_: Default) -> Self {
+            use std::default::Default;
```
mversic, line 135:
```suggestion
            #[cfg(not(feature = "std"))]
            use std::default::Default;
```

crypto/src/lib.rs
```diff
@@ -119,31 +120,54 @@ pub enum KeyGenOption<K = PrivateKey> {
...
+            Self::default()
+        }
+    }
+
+    #[derive(Copy, Clone, Debug)]
+    pub struct Set(pub Algorithm);
```
mversic, line 141:
```suggestion
    pub struct Custom(pub Algorithm);
```

maybe?

crypto/src/lib.rs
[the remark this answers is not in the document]
0x009922: Not relevant anymore.

0x009922:
Updated how keys generation is configured. With some refactoring, managed to avoid unnecessary validation and panics in runtime thanks to types.

Renamed `KeyGenConfiguration` and `KeyGenOption` to `KeyPairGenConfig` and `KeyPairGenOption` respectively.

crypto/src/lib.rs
mversic: I believe I can help you work around FFI macros.

crypto/src/lib.rs
```diff
@@ -110,66 +111,73 @@ impl FromStr for Algorithm {
...
+    /// Configuration for key generation
     #[derive(Clone)]
-    #[cfg_attr(not(feature="ffi_import"), derive(Debug))]
-    pub struct KeyGenConfiguration {
-        /// Options
-        key_gen_option: KeyGenOption,
-        /// Algorithm
-        algorithm: Algorithm,
+    #[cfg_attr(not(feature = "ffi_import"), derive(Debug))]
+    pub struct KeyPairGenConfig{
```
mversic, line 145:
```suggestion
    pub struct KeyGenConfig{
```

I think previous name was fine

crypto/src/lib.rs
0x009922: There was a problem that `KeyPair::generate_with_configuration` started to have a generic parameter to accept `KeyGenConfig<A>`, and FFI macros said generics aren't supported.

And this piece errored with a missing generic argument. Didn't succeed with adding one.

https://github.com/hyperledger/iroha/blob/e79f617f5edc53621e4561f057df92b425beeaca/crypto/src/lib.rs#L867-L873

crypto/src/lib.rs
```diff
@@ -110,66 +111,73 @@ impl FromStr for Algorithm {
...
+    UseSeed { seed: Vec<u8>, algorithm: Algorithm },
+    /// Derive from a private key
+    FromPrivateKey(PrivateKey),
+}
+
+/// While [`KeyPairGenConfig`] is for external users, this structure is for internal use
+/// to pass it to a specific algorithm.
+/// Parameter `K` is intended to be set as a `PrivateKey` type of that specific algorithm.
+#[derive(Debug)]
+pub enum KeyPairGenOption<K> {
```
mversic, line 131:
```suggestion
pub enum KeyGenOpt<K> {
```

I think the previous name was fine?

mversic: there is no ambiguity

crypto/src/lib.rs
mversic: | see, we can't have generic objects in FFI.

in this case, you should have specified all concrete types like this:
```rs
 iroha_ffi::handles! { 
     KeyPairGenConfig<Default>, 
     KeyPairGenConfig<Set>,
     PublicKey, 
     PrivateKey, 
     KeyPair, 
     Signature, 
 }
 ```

Comment (mversic) [reply crypto/src/lib.rs]: but then I'm not sure what would have happened with generic functions. I think there was only one potentially problematic `generate_with_configuration`. If you want to go back to this, tell me and we can work something out

File: crypto/src/lib.rs (hunk 7 above)
Comment (0x009922) [reply crypto/src/lib.rs:131]: I think so. Let me revert.

File: crypto/src/lib.rs (hunk 5 above)
Comment (0x009922) [reply crypto/src/lib.rs]: It's fine, I removed `with_algorithm` method and this problem is gone.

Comment (mversic) [reply crypto/src/lib.rs]: I think you found a very good alternative

File: crypto/src/lib.rs [hunk 8]
```diff
@@ -224,21 +230,42 @@ impl KeyPair {
...
-    ) -> Self {
-        match algorithm {
-            Algorithm::Ed25519 => signature::ed25519::Ed25519Sha512::keypair(key_gen_option).into(),
-            Algorithm::Secp256k1 => {
-                signature::secp256k1::EcdsaSecp256k1Sha256::keypair(key_gen_option).into()
-            }
-            Algorithm::BlsNormal => signature::bls::BlsNormal::keypair(key_gen_option).into(),
-            Algorithm::BlsSmall => signature::bls::BlsSmall::keypair(key_gen_option).into(),
+    /// Generates a pair of Public and Private key with the corresponding [`KeyGenOption`].
+    pub fn generate_with_config(config: KeyGenConfig) -> Self {
```
mversic, line 234: instead of having `KeyGenConfig` I think it would be preferable to have 3 methods on `KeyPair`:
```rs
pub fn rand() -> Self;
pub fn from_seed(algorithm: Algorithm, seed: Vec<u8>) -> Self;
impl TryFrom<PrivateKey> for KeyPair {
    //..
}
```

or similar naming

what do you think?

0x009922: Yup, I think it would work totally fine and simplify things:

- `KeyPair::generate_from_random(algorithm)`
- `KeyPair::generate_from_seed(seed, algorithm)`
- `KeyPair::generate_from_private_key(private_key)`

Let me impl it tomorrow.

crypto/src/lib.rs
```diff
@@ -110,66 +111,73 @@ impl FromStr for Algorithm {
...
     #[derive(Clone)]
-    #[cfg_attr(not(feature="ffi_import"), derive(Debug))]
-    pub struct KeyGenConfiguration {
-        /// Options
-        key_gen_option: KeyGenOption,
-        /// Algorithm
-        algorithm: Algorithm,
+    #[cfg_attr(not(feature = "ffi_import"), derive(Debug))]
+    pub struct KeyGenConfig{
+        inner: KeyGenConfigInner
```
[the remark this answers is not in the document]
0x009922: Because `KeyGenConfigInner` is a enum, and it is also restricted in FFI to have either unit or single-tuple variants. So I introduced a layer of indirection here to satisfy FFI macros with `KeyGenConfig`.

crypto/src/lib.rs
0x009922: And yeah, `TryFrom<PrivateKey>` makes sense too, although I still having a method for it is not bad.

Also, I think it would make more sense to use **"derive"** term:

- generate from random (or just "generate")
- derive from seed
- derive from private key

mversic: > And yeah, TryFrom<PrivateKey> makes sense too, although still having a method for it is not bad.

I wouldn't be so eager to increase API surface. This can always be done so, as a compromise, I would make this into a separate issue to discuss and atm just leave `TryFrom`

> Also, I think it would make more sense to use "derive" term:

"derive" doesn't sound quite right to me. I'd vote for either:
1.
```rs
Keypair::random(algorithm: Algorithm)
Keypair::from_seed(algorithm: Algorithm, seed: Vec<u8>) -> Self;
```
2.
```rs
KeyPair::generate_from_random(algorithm: Algorithm)
KeyPair::generate_from_seed(algorithm: Algorithm, seed: Vec<u8>)
```

but I don't find `generate` prefix brings anything but a longer method name

0x009922: AFAIK, _derivation_ is an established term in cryptography. It applies to both deriving a key pair using pRNG + seed, and to deriving of the public key from a private one.

Some references:

- https://en.wikipedia.org/wiki/Key_derivation_function
- https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey

Anyway, `KeyPair::random` and `KeyPair::from_seed` works fine, imo. And will leave `TryFrom` as the only way for `PrivateKey`.

crypto/src/lib.rs
```diff
@@ -186,28 +136,64 @@ ffi::ffi_item! {
...
-    /// Construct a [`KeyPair`]
+    /// Construct a [`KeyPair`].
+    ///
+    /// See [`Self::into_raw_parts`] for an opposite conversion.
     ///
     /// # Errors
-    /// If public and private key don't match, i.e. if they don't make a pair
-    pub fn new(public_key: PublicKey, private_key: PrivateKey) -> Result<Self, Error> {
+    /// If public and private keys don't match, i.e. if they don't make a pair
+    pub fn from_raw_parts(public_key: PublicKey, private_key: PrivateKey) -> Result<Self, Error> {
```
0x009922, line 196: See https://github.com/hyperledger/iroha/issues/4339#issuecomment-1984711378

I finally renamed it to `from_raw_parts`, because there is a mirrored method called `into_raw_parts` and it makes sense.

mversic: the reason why we need a destructuring method is because we want to keep fields of our structs private but also not be required to clone if we want to move a field out of a struct. This is the same in `iroha_crypto` and `iroha_data_model` so we better find a general solution to this. And having constructors named `from_raw_parts` is not what I would like


IMO `raw_parts` implies some sort of raw pointers when you look at `std`. Instead of `KeyPair::into_raw_parts` we already have:
```rs
impl From<KeyPair> for (PublicKey, PrivateKey)
```

so we can just remove `KeyPair::into_raw_parts`. We can also just remove `KeyPair::new` and add:
```rs
impl From<(PublicKey, PrivateKey)> for KeyPair
```

mversic: let's have a discussion on this at standup

mversic: - let's keep `KeyPair::new`
- rename `KeyPair::into_raw_parts` into `KeyPair::into_parts`
- remove `From<KeyPair> for (Pk, Pk)`

0x009922: Addressed in [`ec0fa3a` (#4341)](https://github.com/hyperledger/iroha/pull/4341/commits/ec0fa3a1a4185f11deabb2cffa5c895a50f7569e), and created https://github.com/hyperledger/iroha/issues/4360 as discussed

mversic (dismissed):

mversic (approved):

DCNick3 (approved):

Issue #4296, opened 2024-02-18
[suggestion] Snapshot configuration - ensure it creates a directory for itself


0x009922:
## Description

There is `snapshot.store_dir` config parameter which specifies the directory where snapshots are stored. It's default value is `./storage/snapshot`.

IMO, the behaviour should be the following: the last directory in the path should be automatically created if it doesn't exist; if the parent directory doesn't exist, Iroha should exit with an error on the config validation stage.

For example: with a `./storage/snapshot` path, Iroha will automatically create `snapshot` dir, but will exit with an error if `storage` doesn't exist.

I am not sure what the behaviour is right now. This issue is about ensuring it, and possibly updating.

PR #4365, opened 2024-03-13
[feature] #4059: snapshots "mode"

### LINKED ISSUE #4059: [discussion] Update how Snapshots are configured
## Description

The current way the snapshots system is configured confuses me, and I propose to make this part clearer.

### Current Design

Snapshots are configured with these parameters:

- `creation_enabled` (boolean, set by default)
- `create_every_ms` (number)
- `dir_path` (string, same as default Kura storage path)

When creation is enabled, Iroha tries to read a snapshot on startup to quickly restore the WSV. Then Iroha periodically re-writes the snapshots. 

When creation is disabled, Iroha still tries to read a snapshot on startup, but doesn't periodically update it.

There is no way for users to disable snapshots mechanism entirely. For example, one might want to disable even reading an existing snapshot at all. If they need to do so, they have to discover the path to the snapshot file (which is not documented) and remove it so that Iroha will not recover its WSV from the file.

### Proposal

I propose to give users clearer control over snapshots by doing the following:

- Instead of `creation_enabled` flag, use a parameter called `mode` and use a enum value for it:
    - `normal` (try to read on startup and update periodically, current behaviour with `creation_enabled = true`)
    - `read` (only try to read on startup, current behaviour with `creation_enabled = false`)
    - `disabled` (do nothing at all)
- Instead of having `dir_path` defaulted to the same directory as Kura storage path (i.e. `./storage`), use a different default value, e.g. `./storage/snapshots`. IMO it will give a little better opportunity for users to remove the snapshots by themselves: just remove `./storage/snapshots` directory, without research of which files in `./storage` belong to storage and which to snapshots.

0x009922:
## Description

- Introduce `mode` of snapshots behaviour: `read_write`, `readonly`, and `disabled`
- Run `SnapshotMaker` actor only in `read_write` mode
- Configuration: remove `snapshot.creation_enabled`, add `snapshot.mode`
- Recursively create directories on snapshot writing (#4296)
- Update related logging
- Add tests

### Linked issue

Closes #4059, #4296 

### Benefits

More expected behaviour of snapshots for users?


core/src/snapshot.rs
```diff
@@ -92,79 +88,68 @@ impl SnapshotMaker {
...
 /// - IO errors
 /// - Deserialization errors
 pub fn try_read_snapshot(
-    snapshot_dir: impl AsRef<Path>,
+    store_dir: impl AsRef<Path>,
     kura: &Arc<Kura>,
     query_handle: LiveQueryStoreHandle,
     BlockCount(block_count): BlockCount,
-) -> Result<WorldStateView> {
+) -> Result<TryReadResult> {
```
mversic, line 140:
```suggestion
) -> Result<WorldStateView, TryReadError> {
```

I think this is more in line with Rust conventions. Just group all possible errors in one enum

config/src/snapshot.rs
```diff
@@ -0,0 +1,36 @@
...
+    strum::Display,
+    strum::EnumString,
+    serde_with::SerializeDisplay,
+    serde_with::DeserializeFromStr,
+)]
+#[strum(serialize_all = "snake_case")]
+pub enum Mode {
+    /// Read the snapshot on startup, update periodically
+    #[default]
+    Normal,
```
mversic, line 18:
```suggestion
    ReadWrite,
```

`Normal` seems ambiguous

core/src/snapshot.rs
0x009922: The thing is, it isn't really an error. 

Anyway, would it be fine?

```rust
enum TryReadError {
  NotFound,
  Other(Error), // general `snapshot::Error`
}
```

mversic (changes requested):

core/src/snapshot.rs
mversic: > The thing is, it isn't really an error.

you want to say that `NonFound`is not necessarily an error?

> Anyway, would it be fine?

yes

mversic: > The thing is, it isn't really an error.

it is an error. Because it's returned from `try_read_snapshot`. An alternative would be to check for existence on disk before calling `try_read_snapshot`

0x009922: Done. Also split `Error` into `TryReadError` and `TryWriteError`. They have an overlap, but the difference is bigger.

config/src/snapshot.rs
0x009922: Updated (`read_write` in serialized form)

cli/src/lib.rs
```diff
@@ -298,7 +307,8 @@ impl Iroha {
         }
         .start();
 
-        let snapshot_maker = SnapshotMaker::from_config(&config.snapshot, sumeragi.clone()).start();
+        let snapshot_maker =
+            SnapshotMaker::from_config(&config.snapshot, &sumeragi).map(SnapshotMaker::start);
```
VAmuzing, line 311: I like this change as it makes it clear from the code that `start` might not run without jumping to the `start` implementation

VAmuzing (approved):

mversic (approved):

PR #4368, opened 2024-03-14
[feature] #4285: Verifiable Random Function in Sumeragi

### LINKED ISSUE #4285: VRF based topology choice
atm we're choosing topology of the next round based on the signees of the block in the current round as a seed. This is susceptible to manipulation. A better solution is to introduce a verifiable random function to produce the seed

SamHSmith:
## Description
Adds a crypto primitive and randomization of the topology on block commits.

### Linked issue

Closes #4285

### Checklist


core/src/sumeragi/network_topology.rs
```diff
@@ -181,7 +188,23 @@ impl Topology {
...
-        let mut topology = Topology::new(block.commit_topology().clone());
+        let created_with_vrf_state = block.header().vrf_state.clone();
+        let mut rng = StdRng::seed_from_u64(u64::from_le_bytes(
+            HashOf::new(&created_with_vrf_state).as_ref()[0..8]
+                .try_into()
+                .expect("Cannot fail"),
+        ));
+
+        let mut topology = {
+            let mut shuffle_peers: Vec<PeerId> = block.commit_topology().clone().into();
```
mversic, line 199: is it really a necessity that we store `commit_topology` in the block? Can't we just use the current topology of sumeragi? Because this field can blow the chain size up significantly

SamHSmith: It is. We could imagine some scheme where we store  only a delta and can infer the current topology. But I think it's probably fine since the block headers don't stay in memory. Once 2 blocks have passed this data should only be on disk. I say the chance of bugs or mistakes is a lot higher if we try to be smart here. Does not seem worth it.

Erigara:
I'm not sure that i fully understand the purpose of this change.

Since VRFState looks predetermined from the beginning, so peer that submit genesis could manipulate it.

Erigara:
Can you provide any sources you used while implementing this approach?

crypto/src/lib.rs
```diff
@@ -718,6 +718,65 @@ impl From<PrivateKey> for PrivateKeySerialized {
...
+    Eq,
+    Clone,
+    Hash,
+    Decode,
+    Encode,
+    Serialize,
+    Deserialize,
+    IntoSchema,
+)]
+pub struct VRFState {
```
mversic, line 737: let's not put this into `iroha_crypto`. This crate is considered finished and up for audit

core/src/block.rs
```diff
@@ -841,7 +857,7 @@ mod tests {
         let transactions = vec![tx_fail, tx_accept];
         let topology = Topology::new(UniqueVec::new());
         let valid_block = BlockBuilder::new(transactions, topology, Vec::new())
-            .chain(0, &mut wsv)
+            .chain(0, VRFState::generate_new_random_state(), &mut wsv)
```
mversic, line 860:
```suggestion
            .chain(0, VRFState::random(), &mut wsv)
```

core/src/sumeragi/network_topology.rs
mversic: > But I think it's probably fine since the block headers don't stay in memory

that's true

> We could imagine some scheme where we store only a delta and can infer the current topology

can't we replay topology while doing block replay? I'm saying it will be computed anyhow, so why store it in the block? 

> I say the chance of bugs or mistakes is a lot higher if we try to be smart here. Does not seem worth it

I need a better argument why it's necessary to store it in the block header

SamHSmith: It isn't necessary, you are right that we could replay the topology with the block replay. It didn't seem worth it to me since the topology is quite small. But if you think we should replay it then I could work on that.

SamHSmith:
> I'm not sure that i fully understand the purpose of this change.
> 
> Since VRFState looks predetermined from the beginning, so peer that submit genesis could manipulate it.
> 
> Because it's computed as `signature(signature(rand, genesis), leader)`.

`signature(rand, genesis)` is in practice random. Because if could be predicted then you
could fake signatures. This is the basic verifiable random property. A random number is generated
that only the leader can generate. But everyone else can verify the number so the leader cannot tamper
with it. This means that the topology evolution cannot be tampered with, but only the leader can know
what the next topology will be. This reduces the time for an attacker to coordinate a denial of service attack.

The topology for the next round is known only when a peer receives the block candidate. And nobody is able
to predict the topology two blocks ahead. (Unless by random chance the same leader was picked twice.)

SamHSmith:
Also the seed value that the genesis peers submits is completely unimportant. It just needs to be some value. I've thought about @mversic 's suggestion that we should replay topology from the start and not store it on each block. I think it is a good idea and I will make the change in this PR.

Erigara:
Afaik for the single pair of payload and private key there is more than one valid signature.
For elliptic curve signatures this set is large due to selection of pseudorandom number. 
So i have concern that this could be abused by leader to pick itself as leader again.

SamHSmith:
Good point. I will have to investigate that. It might be that I can encrypt the hash with the private key, that way creating a signing scheme that only has one correct answer. This is the behavior I assumed from our signature primitives. Good that you spotted this.

mversic:
> Also the seed value that the genesis peers submits is completely unimportant. It just needs to be some value. I've thought about @mversic 's suggestion that we should replay topology from the start and not store it on each block. I think it is a good idea and I will make the change in this PR.

@SamHSmith don't remove `commit_topology` in the scope of this PR

config/src/parameters/user.rs
```diff
@@ -387,6 +387,9 @@ impl Sumeragi {
         } = self;
 
         let trusted_peers = construct_unique_vec(trusted_peers.unwrap_or(vec![]))?;
+        for peer in &trusted_peers {
+            assert!(peer.public_key.algorithm() == iroha_crypto::Algorithm::Secp256k1);
```
Erigara, line 391: I think we should return human readable error with explanation that only certain public key types are allowed.

core/src/sumeragi/main_loop.rs
```diff
@@ -1460,3 +1480,19 @@ mod tests {
...
+        .expect("Is not allowed to fail")
+}
+/// Verify the verifiable random function
+pub fn verify_vrf(old_state: &Vec<u8>, new_state: &Vec<u8>, pk: &PublicKey) -> bool {
+    assert!(pk.algorithm() == iroha_crypto::Algorithm::Secp256k1);
+
+    let mut ctx = ECVRF::from_suite(CipherSuite::SECP256K1_SHA256_TAI).expect("Cannot fail");
+    ctx.verify(&pk.to_bytes().1, new_state.as_ref(), old_state.as_ref())
+        .is_ok()
+}
```
Erigara, line 1498: IMO you can create separate file with VRF related functionality

config/src/parameters/user.rs
Erigara: Should we as well check key type of peer in `Register<Peer>` instruction?

docs/source/references/schema.json
```diff
@@ -4490,6 +4494,14 @@
...
     ]
   },
+  "VRFState": {
+    "Struct": [
+      {
+        "name": "signature",
+        "type": "Signature"
+      }
+    ]
+  },
```
Erigara, line 4504: you should update schema

core/src/sumeragi/network_topology.rs
```diff
@@ -181,7 +182,23 @@ impl Topology {
...
+            shuffle_peers.shuffle(&mut rng);
+            let mut ordered_peers = UniqueVec::new();
+            ordered_peers.extend(shuffle_peers);
+            Topology {
+                ordered_peers,
+                created_with_vrf_state,
+            }
+        };
```
Erigara, line 201: I noticed that we still just rotate topology on view change witch create predictable peer order, should we do something about that?

Erigara:
AFAIK using VRFs should provide secrecy property (so it's up to the leader to reveal itself with block proposal and proof of being the leader).
Currently i think we don't have such property, because everyone is know what topology going to look like when they receive the block.
So i have a question do we need VRFs for doing that?

core/src/sumeragi/network_topology.rs
SamHSmith: We can't do anything about that. Any order change will be predictable since all peers must calculate it separately. This is only true for the view change though.

core/src/sumeragi/network_topology.rs
```diff
@@ -169,7 +176,7 @@ impl Topology {
     }
 
     /// Perform sequence of actions after block committed.
-    pub fn update_topology(&mut self, block_signees: &[PublicKey], new_peers: UniqueVec<PeerId>) {
+    fn update_topology(&mut self, block_signees: &[PublicKey], new_peers: UniqueVec<PeerId>) {
```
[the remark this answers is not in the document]
SamHSmith: We don't rely on the order of the signatures, we only use them to form the A set.

core/src/sumeragi/network_topology.rs
SamHSmith: @mversic I misspoke in our call. We would need to sort after we make this change and keep track of Peers without storing it in the block header. Right now we use the commit topology and that will be the same for all peers since it is on the block header. Once we remove the topology from the header we would need track the peers with a UniqueVec for consistent shuffling on different peers. But as of now there is no thing I needed to force push to the PR.

SamHSmith:
Something is wrong with the CI here.

PR #4369, opened 2024-03-15
[fix] #4323: remove `iroha_config/tokio-console` feature

### LINKED ISSUE #4323: [discussion] Remove `tokio-console` feature from `iroha_config`
`iroha_config` crate has `tokio-console` feature, which enables `logger.tokio_console_address` configuration parameter.

It introduces some dev issues:

- If you run `cargo tests` without this feature enables, config snapshot tests will fail, and their feature-gating will be quite... ambiguous and cumbersome, I can imagine.
- It produces some cratches across workspace, like this:
  https://github.com/hyperledger/iroha/blob/964476722e2a219becaacdb3676ca058ec5748cd/logger/src/lib.rs#L85-L90

A benefit of having this feature is that... allowing to compile Iroha without knowing that this field even exists?

From the UI standpoint, I think, even in this case it would be better to have `logger.tokio_console_address` field in the config anyway, but if the actual feature is disabled, Iroha can issue a warning. This is more user-friendly, simpler in the code, and has minimal overhead.

0x009922:
## Description

See the linked issue

### Linked issue

Closes #4323

### DIFF
--- a/config/src/parameters/user.rs
+++ b/config/src/parameters/user.rs
@@ -461,7 +461,6 @@ pub struct Queue {
     pub future_threshold: Duration,
 }
 
-#[allow(missing_copy_implementations)] // triggered without tokio-console
 #[derive(Debug, Clone)]
 pub struct Logger {
     /// Level of logging verbosity
@@ -471,18 +470,15 @@ pub struct Logger {
     pub level: Level,
     /// Output format
     pub format: LoggerFormat,
-    #[cfg(feature = "tokio-console")]
-    /// Address of tokio console (only available under "tokio-console" feature)
+    /// Address of tokio console
     pub tokio_console_address: SocketAddr,
 }
 
-#[allow(clippy::derivable_impls)] // triggers in absence of `tokio-console` feature
 impl Default for Logger {
     fn default() -> Self {
         Self {
             level: Level::default(),
             format: LoggerFormat::default(),
-            #[cfg(feature = "tokio-console")]
             tokio_console_address: super::defaults::logger::DEFAULT_TOKIO_CONSOLE_ADDR,
         }
     }
--- a/config/src/parameters/user/boilerplate.rs
+++ b/config/src/parameters/user/boilerplate.rs
@@ -490,18 +490,11 @@ impl UnwrapPartial for QueuePartial {
 
 impl FromEnvDefaultFallback for QueuePartial {}
 
-/// 'Logger' configuration.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Default, Merge)]
-// `tokio_console_addr` is not `Copy`, but warning appears without `tokio-console` feature
-#[allow(missing_copy_implementations)]
 #[serde(deny_unknown_fields, default)]
 pub struct LoggerPartial {
-    /// Level of logging verbosity
     pub level: UserField<Level>,
-    /// Output format
     pub format: UserField<Format>,
-    #[cfg(feature = "tokio-console")]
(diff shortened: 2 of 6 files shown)


config/src/parameters/user/boilerplate.rs
```diff
@@ -512,9 +505,8 @@ impl UnwrapPartial for LoggerPartial {
         Ok(Logger {
             level: self.level.unwrap_or_default(),
             format: self.format.unwrap_or_default(),
-            #[cfg(feature = "tokio-console")]
             tokio_console_address: self.tokio_console_address.get().unwrap_or_else(|| {
-                super::super::defaults::logger::DEFAULT_TOKIO_CONSOLE_ADDR.clone()
+                defaults::logger::DEFAULT_TOKIO_CONSOLE_ADDR.clone()
```
VAmuzing, line 509: Seems like clippy wants it this way:
```rust
tokio_console_address: self
    .tokio_console_address
    .get()
    .unwrap_or_else(|| defaults::logger::DEFAULT_TOKIO_CONSOLE_ADDR.clone()),
```

config/src/parameters/user.rs
```diff
@@ -471,18 +470,15 @@ pub struct Logger {
     pub level: Level,
     /// Output format
     pub format: LoggerFormat,
-    #[cfg(feature = "tokio-console")]
-    /// Address of tokio console (only available under "tokio-console" feature)
+    /// Address of tokio console
     pub tokio_console_address: SocketAddr,
```
mversic, line 474:
```suggestion
    pub tokio_console_address: Option<SocketAddr>,
```

should it not be optional now?

0x009922: It is optional for users, but since it has a default value, it is always `SocketAddr` here.

mversic: oh, I see. Do you think it should have a default?

0x009922: I haven't dived into the purpose of this feature, to be honest. So, I don't have an opinion on that. Will check it out.

0x009922: What I have found:

- There is a broken code in `iroha_logger`, relying on obsolete name `tokio_console_addr`. Our CI doesn't detect validity of different features presets.
  https://github.com/hyperledger/iroha/blob/f3e7097c9c07dc024f6de32d968f594d284880c8/logger/src/lib.rs#L120-L132
- I think having a default value for this parameter is reasonable, because this functionality is switched with a feature flag of `iroha_logger`. However, `iroha`'s feature flags do not enable this feature of logger in any way.
- Although `tokio_console` is directly related to `tracing`, I think it would make more sense to move the parameter from `logger` to `telemetry` section of the config. I.e. not `logger.tokio_console_address`, but `telemetry.tokio_console.address`.

So, what I propose:

- Move the parameter from `logger` to `telemetry.tokio_console.address`, alongside with "regular" and "dev" telemetries
- Possibly nest "regular" telemetry from `telemetry.*` to `telemetry.substrate` or whatever
- Possibly continue feature-gating this field in `iroha_config`. As well as other telemetry-related fields, because telemetry might be totally disabled by features of `iroha`
- By continuing feature-gating these parameters, it is reasonable for them to have default values when they are enabled, imo.

Your thoughts?

0x009922: A contra point against using feature-gating in `iroha_config`: it complicates snapshot-based testing.

0x009922:
Closing it, because I am going to open a separate PR with different scope to solve this issue.

0x009922:
Moved to #4377

PR #4370, opened 2024-03-15
[refactor] #4291: update genesis config & schema

### LINKED ISSUE #4291: [suggestion] Update executor config in the genesis block
## Description

> TL;DR: drop inline executor support; rename field to `executor_file`.

Genesis block contains `executor` field, which might either represent an inline base64-encoded WASM blob, or a file path to the WASM blob:

https://github.com/hyperledger/iroha/blob/964476722e2a219becaacdb3676ca058ec5748cd/configs/swarm/genesis.json#L185

This is backed by the `ExecutorMode` struct:

https://github.com/hyperledger/iroha/blob/964476722e2a219becaacdb3676ca058ec5748cd/genesis/src/lib.rs#L122-L131

As we discussed in #4239, `executor` field with file path should be renamed to `executor_file` for uniformity.

In addition to that, I propose to remove inline executor mode. It was effectively deprecated a long time ago: there was a warning printed when the inline mode is used (cannot find references for that). The reason is that inlining WASM blob makes the whole genesis file (1) clunky and (2) hardly reproducible due to WASM compilation intricacies.

0x009922:
## Description

- remove inline executor mode
- rename field to `executor_file`
- remove genesis from schema (revert #1391).
  - **Why?**
  - Because it doesn't belong here. It's not a part of the data model, but rather of the configuration system. The original reasoning about "ease for SDK developers" doesn't work (genesis is JSON, but schema is about SCALE; types are straightforward).

### Linked issue

Closes #4291

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -516,10 +516,11 @@ pub fn read_config_and_genesis<P: AsRef<Path>>(
     let genesis = if let Genesis::Full { key_pair, file } = &config.genesis {
         let raw_block = RawGenesisBlock::from_path(file)?;
 
-        Some(
-            GenesisNetwork::new(raw_block, &config.common.chain_id, key_pair)
-                .wrap_err("Failed to construct the genesis")?,
-        )
+        Some(GenesisNetwork::new(
+            raw_block,
+            &config.common.chain_id,
+            key_pair,
+        ))
     } else {
         None
     };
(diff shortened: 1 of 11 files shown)


mversic (dismissed):

VAmuzing (approved):
There is just a small thing to mention: you might want to add a missing newline in `config/tests/fixtures/empty_ok_genesis.json'.

0x009922:
I am not sure about this PR: Java SDK relies on the schema entries, and maybe on inline executor too. Let's get more context.

mversic (approved):

mversic (approved):

PR #4375, opened 2024-03-19
[refactor] #4259: Make Action and Filter non-generic types

### LINKED ISSUE #4259: Make `Action` and `Filter` non-generic
`Action` and `Filter` are currently generic over a type of filter they use. This is used in `iroha_core` to split event triggers for events coming from different sources.

This is `iroha_core` an implementation detail that makes `iroha_data_model` harder to understand for the user.

`Action` should have a concrete event filter type specified and `iroha_core` should have its own types for the specialized event filters which would not be exposed to the user.

DCNick3:
## Description

Make the `Action` and `Filter` types exported from `iroha_data_model` non-generic. This is used by code in `iroha_core`, so make generic copies in there. There's no need to complicate the public API due to the requirements of the implementation.

This also makes the `Action` constructor accept `impl Into<TriggeringEventFilterBox>` as a filter, simplifying most API invocations. Additional shorthand conversions from concrete event filter types are added (for example `AccountEventFilter -> TriggeringEventFilterBox`) to reduce the required boilerplate even more.

### Linked issue

Closes #4259

### Benefits

No unnecessary generic in the API -> better code completion and easier API.

### Checklist

- [ ] make CI pass


core/src/smartcontracts/isi/triggers/mod.rs
```diff
@@ -130,7 +112,26 @@ pub mod isi {
         }
     }
 
-    impl Execute for Burn<u32, Trigger<TriggeringEventFilterBox>> {
+    impl Execute for Unregister<Trigger> {
```
Erigara, line 115: Do you need change order? `Register/Unregister` and then `Mint/Burn` seems reasonable.
I'm also thinking about resolving conflicts.

DCNick3: This was unintentional... I'll revert this

Erigara (dismissed):

data_model/src/events/mod.rs
```diff
@@ -209,6 +209,57 @@ impl EventFilter for TriggeringEventFilterBox {
...
+        DomainEventFilter           => DataEventFilter => EventFilterBox,
+        AccountEventFilter          => DataEventFilter => EventFilterBox,
+        AssetEventFilter            => DataEventFilter => EventFilterBox,
+        AssetDefinitionEventFilter  => DataEventFilter => EventFilterBox,
+        TriggerEventFilter          => DataEventFilter => EventFilterBox,
+        RoleEventFilter             => DataEventFilter => EventFilterBox,
+        ConfigurationEventFilter    => DataEventFilter => EventFilterBox,
+        ExecutorEventFilter         => DataEventFilter => EventFilterBox,
+
+        PeerEventFilter             => DataEventFilter => TriggeringEventFilterBox,
```
mversic, line 251: IMO we should remove `TriggeringEventFilterBox` and have only `EventFilterBox`. Is there a good reason why we have it?

DCNick3: We have it to enforce on type level that `TriggerCompletedEvent` can't trigger a trigger to run to prevent infinite loops (see #2885)

mversic (approved):

data_model/src/events/mod.rs
mversic: ok, got it, but we don't have to prevent it on type level

DCNick3: Yeah, that would work. Can also reject transactions registering such a trigger for a bit better UX

Erigara (approved):

PR #4377, opened 2024-03-21
[refactor] #4323: telemetry features, tokio console, logs, and configuration

### LINKED ISSUE #4323: [discussion] Remove `tokio-console` feature from `iroha_config`

0x009922:
## Description

_This is an improved iteration of https://github.com/hyperledger/iroha/pull/4369_

- Remove `logger.tokio_console_address` configuration parameter entirely: this could be configured in a more refined way using default ENVs (see [`console_subscriber` docs](https://docs.rs/console-subscriber/0.2.0/console_subscriber/struct.Builder.html#method.with_default_env))
- Rename `[telemetry.dev]` config section to `[dev_telemetry]`
- Update `telemetry` and `dev-telemetry` features of `iroha`
- Update `tokio-console` feature of `iroha_logger`: now console seamlessly starts whenever this feature is enabled
- Improve logs & warnings
- Fix dev-telemetry file logs: separate each item with `\n`, and recursively create parent directories
- Improve config validation: handle file paths

### Linked issue

Closes #4323 

### Benefits

- Less config parameters
- More flexible way to configure tokio console
- Less feature mess (?)
- Better warnings
- Refactoring

### Checklist

- [x] Document how to use tokio console
- [x] Figure out what happened to `Cargo.toml`s
- [x] Check manually several cases, how Iroha behaves with `telemetry`/`dev-telemetry` features enabled depending on `telemetry` configuration (warnings, logs, file output)
- [x] Check manually Tokio Console (works, but requires `TRACE` log level)


Cargo.toml
mversic, line 1: this file got messed up

core/Cargo.toml
```diff
@@ -18,15 +18,10 @@ categories.workspace = true
...
 
 # Support lightweight telemetry, including diagnostics
 telemetry = []
-# Support the included CLI
-cli = []
-# Support developer-specific telemetry.
-# Should not be enabled on production builds.
-dev-telemetry = ["telemetry", "iroha_telemetry/dev-telemetry"]
 # Support Prometheus metrics. See https://prometheus.io/.
 expensive-telemetry = ["iroha_telemetry/metric-instrumentation"]
```
mversic, line 26: I think this isn't used and can be removed

0x009922: There are many `cfg`s in `iroha_telemetry_derive`'s generated code relying on this feature. I wonder why it isn't used by `iroha` crate, and how Prometheus works without it...

0x009922: There is still a mess around telemetry features...

- In Torii, `telemetry` enables `/version` (why it is telemetry?), `/status` (again, why?), and `/metrics` (Prometheus data) endpoints;
- In Iroha itself, `telemetry` enables an WebSocket service which submits a telemetry events to some endpoint. `dev_telemetry` extends previous case to writing to a file, and also enables tokio console in `iroha_logger`.
- `Metrics` in Sumeragi in `iroha_core`, which are served as Prometheus metrics, are collected anyway, independent from enabled features
- `iroha_core` has `telemetry` feature, which doesn't enable anything else, and isn't used inside by the crate
- purpose of `iroha_telemetry/metric-instrumentation` isn't really clear

Cargo.toml
0x009922: Fixed

cli/src/lib.rs
```diff
@@ -229,16 +230,16 @@ impl Iroha {
...
                 Some(state)
             }
             Err(TryReadSnapshotError::NotFound) => {
-                iroha_logger::info!("Didn't find a snapshot of state, creating an empty one");
+                iroha_logger::info!("Didn't find a WSV snapshot; creating empty WSV");
                 None
             }
             Err(error) => {
-                iroha_logger::warn!(%error, "Failed to load state from a snapshot, creating an empty one");
+                iroha_logger::warn!(%error, "Failed to load WSV from a snapshot; creating empty WSV");
```
Erigara, line 242: `WSV` is renamed to just `State` now, so referring to `WSV` doesn't really make sense now

Erigara:
Does telemetry used at all? I'm aware that QA and DevOps team using prometheus metrics.
But i'm not aware of using WS for telemetry, if it's not required we can make code simpler.

0x009922:
@Erigara, created a separate discussion for it: https://github.com/hyperledger/iroha/issues/4385

core/Cargo.toml
0x009922: @mversic, should we address it now, or separately, or never?

mversic: let 's open a ticket

0x009922: Opened: #4386

cli/src/lib.rs
0x009922: Fixed.

mversic (approved):

Erigara (approved):

PR #4379, opened 2024-03-21
[refactor] #4075: Move metrics related functionality into MetricsReporter

### LINKED ISSUE #4075: [refactor]: Move metrics into separate service
it would be nice if we could remove `sumeragi` from `torii`. I see that it is only used with `sumeragi.metrics` and `sumeragi.update_metrics`. Maybe we can make a ticket for this to make metrics a separate service

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/4061#discussion_r1401915422_
            
It would be convenient to decouple `torii` and `sumeragii`. And after wsv update (#4061) the last thing that requires `torii` to depend on `sumeragi` is metrics which should be moved into separate service.

Erigara:
## Description

- sumergai no longer responsible for handling metrics for torii
- fix bug with not proper view change index in metrics

### Linked issue

Closes #4075

### DIFF
--- a/core/src/metrics.rs
+++ b/core/src/metrics.rs
@@ -1,0 +1,155 @@
+//! Metrics and status reporting
+
+use std::{sync::Arc, time::SystemTime};
+
+use eyre::{Result, WrapErr as _};
+use iroha_telemetry::metrics::Metrics;
+use parking_lot::Mutex;
+use storage::storage::StorageReadOnly;
+
+use crate::{
+    kura::Kura,
+    queue::Queue,
+    state::{State, StateReadOnly, WorldReadOnly},
+    IrohaNetwork,
+};
+
+/// Responsible for collecting and updating metrics
+#[derive(Clone)]
+pub struct MetricsReporter {
+    state: Arc<State>,
+    network: IrohaNetwork,
+    kura: Arc<Kura>,
+    queue: Arc<Queue>,
+    metrics: Metrics,
+    /// Latest observed and processed height by metrics reporter
+    latest_block_height: Arc<Mutex<u64>>,
+}
+
+impl MetricsReporter {
+    /// Construct [`Self`]
+    pub fn new(
+        state: Arc<State>,
+        network: IrohaNetwork,
+        kura: Arc<Kura>,
+        queue: Arc<Queue>,
+    ) -> Self {
+        Self {
+            state,
+            network,
+            queue,
+            kura,
+            metrics: Metrics::default(),
+            latest_block_height: Arc::new(Mutex::new(0)),
+        }
+    }
+
+    /// Update the metrics on the state.
+    ///
+    /// # Errors
+    /// - Domains fail to compose
+    ///
+    /// # Panics
+    /// - If either mutex is poisoned
+    #[allow(clippy::cast_precision_loss)]
+    pub fn update_metrics(&self) -> Result<()> {
+        let online_peers_count: u64 = self
+            .network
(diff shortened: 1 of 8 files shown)


core/src/sumeragi/mod.rs
```diff
@@ -29,146 +27,25 @@ pub mod message;
...
 #[derive(Clone)]
 pub struct SumeragiHandle {
-    state: Arc<State>,
-    metrics: Metrics,
-    last_update_metrics_mutex: Arc<Mutex<LastUpdateMetricsData>>,
-    network: IrohaNetwork,
-    kura: Arc<Kura>,
-    queue: Arc<Queue>,
+    /// Counter for amount of dropped messages by sumeragi
+    dropped_messages_metric: iroha_telemetry::metrics::DroppedMessagesCounter,
```
mversic, line 37: seems like a useless thing to measure, I'd remove it

Erigara: We can detect that something is wrong with sumeragi, by observing this metric.
But i didn't have particular opinion on this metric.

mversic: @0x009922 your thoughts?

0x009922: Seems like it might be useful, but I don't have a strong opinion on that neither.

Erigara: I would prefer to remove it in separate ticket if we decided to.
Since currently we don't break the API in any way.

mversic (approved):

mversic (dismissed):

core/src/metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+        Self {
+            state,
+            network,
+            queue,
+            kura,
+            metrics: Metrics::default(),
+            last_update_metrics_mutex: Arc::new(Mutex::new(LastUpdateMetricsData {
+                block_height: 0,
+            })),
+        }
```
0x009922, line 55: You can impl `Default` for `LastUpdateMetricsData` and replace a whole thing with `<_>::default()`.

core/src/metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    queue::Queue,
+    state::{State, StateReadOnly, WorldReadOnly},
+    IrohaNetwork,
+};
+
+/*
+The values in the following struct are not atomics because the code that
+operates on them assumes their values does not change during the course of
+the function.
+*/
```
0x009922, line 21: Maybe, it would be clearer to say that "This struct doesn't contain atomics because it is supposed to be wrapped into an atomic guard (e.g. mutex) as a whole."

0x009922 (dismissed):

core/src/metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    ) -> Self {
+        Self {
+            state,
+            network,
+            queue,
+            kura,
+            metrics: Metrics::default(),
+            last_update_metrics_mutex: Arc::new(Mutex::new(LastUpdateMetricsData {
+                block_height: 0,
+            })),
```
0x009922, line 54:
```suggestion
            last_update_metrics_mutex: <_>::default(),
```

You can impl `Default` for `LastUpdateMetricsData` and replace a whole thing with `<_>::default()`.

0x009922 (approved):

0x009922 (approved):

mversic (approved):

PR #4381, opened 2024-03-22
[feature] #4350: Expose event set bitfields in schema

### LINKED ISSUE #4350: Include event set bitmap info in schema
With the implementation of #4240, we now have a family of `*EventSet`, which are encoded with scale codec as bit maps. We should figure out a way to encode bitmap values corresponding to each event in schema for SDKs consumption.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/4240#discussion_r1516029587_

DCNick3:
## Description

Add a new kind of schema metadata - BitmapMeta - and use it to expose `*EventSet` bitfields in schema

### Linked issue

Closes #4350 

### Checklist

- [ ] make ci pass

### DIFF
--- a/docs/source/references/schema.json
+++ b/docs/source/references/schema.json
@@ -94,7 +94,57 @@
       }
     ]
   },
-  "AccountEventSet": "u32",
+  "AccountEventSet": {
+    "Bitmap": {
+      "repr": "u32",
+      "masks": [
+        {
+          "name": "AnyAsset",
+          "mask": 1
+        },
+        {
+          "name": "Created",
+          "mask": 2
+        },
+        {
+          "name": "Deleted",
+          "mask": 4
+        },
+        {
+          "name": "AuthenticationAdded",
+          "mask": 8
+        },
+        {
+          "name": "AuthenticationRemoved",
+          "mask": 16
+        },
+        {
+          "name": "PermissionAdded",
+          "mask": 32
+        },
+        {
+          "name": "PermissionRemoved",
+          "mask": 64
+        },
+        {
+          "name": "RoleRevoked",
+          "mask": 128
+        },
+        {
+          "name": "RoleGranted",
+          "mask": 256
+        },
+        {
+          "name": "MetadataInserted",
+          "mask": 512
+        },
+        {
+          "name": "MetadataRemoved",
+          "mask": 1024
+        }
+      ]
+    }
+  },
   "AccountId": {
     "Struct": [
       {
@@ -310,7 +360,41 @@
       }
     ]
   },
-  "AssetDefinitionEventSet": "u32",
+  "AssetDefinitionEventSet": {
+    "Bitmap": {
+      "repr": "u32",
+      "masks": [
+        {
+          "name": "Created",
+          "mask": 1
+        },
+        {
+          "name": "MintabilityChanged",
+          "mask": 2
+        },
+        {
+          "name": "OwnerChanged",
+          "mask": 4
+        },
+        {
+          "name": "Deleted",
+          "mask": 8
+        },
+        {
+          "name": "MetadataInserted",
+          "mask": 16
+        },
+        {
+          "name": "MetadataRemoved",
+          "mask": 32
+        },
+        {
+          "name": "TotalQuantityChanged",
+          "mask": 64
+        }
+      ]
+    }
+  },
   "AssetDefinitionId": {
     "Struct": [
       {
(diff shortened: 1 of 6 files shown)


0x009922 (approved):

docs/source/references/schema.json
```diff
@@ -1000,7 +1166,17 @@
...
+    "Bitmap": {
+      "repr": "u32",
+      "masks": [
+        {
+          "name": "Upgraded",
+          "mask": 1
+        }
+      ]
+    }
+  },
```
mversic, line 1179:
```suggestion
  "ExecutorEventSet": {
    "Enum": [
      {
        "tag": "Upgraded",
        "discriminant": 1
      }
    ]
  },
```

why don't we just serialize it as an enum?

mversic (changes requested):

docs/source/references/schema.json
```diff
@@ -631,7 +745,25 @@
...
+          "name": "Created",
+          "mask": 2
+        },
+        {
+          "name": "Deleted",
+          "mask": 4
+        }
+      ]
+    }
+  },
```
mversic, line 766:
```suggestion
  "ConfigurationEventSet": {
    "Enum": [
      {
        "tag": "Changed",
        "discriminant": 1
      },
      {
        "tag": "Created",
        "discriminant": 2
      },
      {
        "tag": "Deleted",
        "discriminant": 4
      }
    ]
  },
```

it seems to me that we could have just treated it as an enum. I don't like seeing extra complexity in the schema

mversic: @0x009922 how will you use this in the sdk? does current schema format bring you something that enum doesn't?

0x009922: I think re-using enum in this case would be more problematic than helpful. After all, enums variants are mutually exclusive, while this new structure could have many values set at once.

If we use a enum in this case, it will enforce a hard code on SDK codegen side, saying "this specific enum is not actually a enum". I would rather have a separate structure for bit masks explicitly in the schema.

docs/source/references/schema.json
mversic, line 1179:
```suggestion
  "ExecutorEventSet": {
    "Bitmap": {
      "masks": [
        {
          "name": "Upgraded",
          "mask": 1
        }
      ]
    }
  },
```

I don't think it's necessary to define `repr`. This implementation detail and doesn't belong to a schema

0x009922: It is necessary to know to encode/decode SCALE.

mversic: then how come we don't have `repr` for enums?

mversic: I see, we implicitly assume `u8`

0x009922: In the SCALE specification it is said that enum always take a single byte for discriminant.

> A fixed number of variants, each mutually exclusive and potentially implying a further value or series of values. **Encoded as the first byte identifying the index of the variant that the value is.** Any further bytes are used to encode any data that the variant implies. **Thus, no more than 256 variants are supported.**

https://docs.substrate.io/reference/scale-codec/

docs/source/references/schema.json
mversic: > If we use a enum in this case, it will enforce a hard code on SDK codegen side, saying "this specific enum is not actually a enum". I would rather have a separate structure for bit masks explicitly in the schema.

hm, I respectfully disagree here. This is not something like an enum, this is literally an enum, at least from the perspective of a schema. It's an implementation detail how this enum is actually represented in the code. 

In Rust we cannot express relations (`BitOr` and `BitOrAssign`) between variants of this enum as we think is the most ergonomical, so we fallback to using a combination of a struct(that is made to look like an enum) and associated constants to represent this type. However, in a different language (think of `Haskell` where enum variants are types in their own right) it might be possible to represent this type as an enum

mversic: > If we use a enum in this case, it will enforce a hard code on SDK codegen side, saying "this specific enum is not actually a enum". I would rather have a separate structure for bit masks explicitly in the schema.

hm, I respectfully disagree here. This is not something like an enum, this is literally an enum, at least from the perspective of a schema. It's an implementation detail how this enum is actually represented in the code.

@DCNick3 correct me if I'm wrong, but you could have implemented this type as a regular enum, but you made an optimization. Whether or not an SDK will make this optimization is not something to be required by a schema

DCNick3: `EventSet` is an optimization indeed, but it is not equivalent to an enum, but to a `HashSet<Enum>`. Enums have one and only one variant set at the same time, while `EventSet` can have any combination of set and unset variants. As we know that the set of `Enum` values is low we store them as single bits instead of allocating a whole `HashSet` on the heap.

What `EventSet`s currently have though, is a corresponding enum that can be used to reconstruct the set of masks using the same logic as the derive macro. However, the logic is non-trivial (there are some name mappings applied to the mappings, for example) and would have to duplicate the logic of the derive macro. Exposing it in schema keeps the translation logic in one place (derive macro) and makes SDK codegen straightforward, so I think schema is _the_ place to have this information.

docs/source/references/schema.json
DCNick3: I guess we can fix it as a `u32`, I just wanted to give possible leeway in case the `BitmapMeta` is going to be reused for something that is not an `EventSet`

docs/source/references/schema.json
mversic: ok, I concede. This is the best solution :)

docs/source/references/schema.json
mversic: I think it's fine

mversic (approved):

PR #4382, opened 2024-03-25
[feature] #4225: Remove genesis signing

### LINKED ISSUE #4225: Remove genesis signing from Iroha
Genesis should be signed offline and given to Iroha through config. This way no peer needs to know about genesis private key. The functionality to sign genesis (or any) transaction could be placed into `kagami`. Also this would remove the `--submit-genesis` from the start args because genesis could just be picked up through config

VAmuzing:
## Description

Draft implementation of removing genesis signing, still need to finish removing all the private keys where they're no longer necessary

Added additional option for kagami onto crypto command, now it has both genesis signing and key-pair generation options

### Linked issue

Closes #4225 

### Checklist


config/src/parameters/user/boilerplate.rs
```diff
@@ -184,7 +184,7 @@ impl FromEnv for RootPartial {
 #[serde(deny_unknown_fields, default)]
 pub struct GenesisPartial {
     pub public_key: UserField<PublicKey>,
-    pub private_key: UserField<PrivateKey>,
+    pub public_key_algorithm: UserField<Algorithm>,
```
0x009922, line 187: The public key is represented as a multihash, which includes the algorithm in it. So, the algorithm could be derived directly from `PublicKey`, therefore `public_key_algorithm` field is redundant.

Please remove.

config/src/parameters/user.rs
```diff
@@ -236,10 +224,9 @@ impl Root {
 }
 
 #[derive(Copy, Clone)]
-pub struct CliContext {
-    pub submit_genesis: bool,
-}
+pub struct CliContext {}
```
0x009922, line 227: If there is no context needed from CLI layer for configuration parsing anymore, I think this struct may be removed completely.

config/src/parameters/user.rs
```diff
@@ -308,24 +295,22 @@ pub(crate) fn private_key_from_env<E: Error>(
...
     pub file: Option<PathBuf>,
 }
 
 impl Genesis {
-    fn parse(self, cli: CliContext) -> Result<actual::Genesis, GenesisConfigError> {
-        match (self.private_key, self.file, cli.submit_genesis) {
-            (None, None, false) => Ok(actual::Genesis::Partial {
+    fn parse(self, _cli: CliContext) -> Result<actual::Genesis, GenesisConfigError> {
+        match self.file {
+            None => Ok(actual::Genesis::Partial {
```
0x009922, line 304: As far as I understand, the new design implies that every peer should submit the genesis?

If so, then there shouldn't be `Genesis::Partial`/`Genesis::Full`, it should be always complete.

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+use iroha_data_model::ChainId;
+use iroha_genesis::{GenesisNetwork, RawGenesisBlock};
+use parity_scale_codec::Encode;
 
 use super::*;
 
+#[derive(Subcommand, Debug, Clone)]
+pub enum CryptoMode {
+    GenesisSigning(GenesisSigningArgs),
+    KeyPairGeneration(KeyPairArgs),
```
0x009922, line 22:
```suggestion
    SignGenesis(SignGenesisArgs),
    GenerateKeyPair(GenerateKeyPairArgs),
```

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+#[command(group = ArgGroup::new("public_key").required(true))]
+#[command(group = ArgGroup::new("format").required(false))]
+pub struct GenesisSigningArgs {
+    /// The algorithm of the provided keypair
+    #[clap(default_value_t, long, short)]
+    algorithm: AlgorithmArg,
+    /// Private key (in string format) to sign genesis block
+    #[clap(long, group = "private_key")]
+    private_key_string: Option<String>,
+    /// Path to private key to to sign genesis block
```
0x009922, line 37:
```suggestion
    /// Path to private key to sign genesis block
```

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+            (Some(path_buf), None) => Ok(Cow::Owned(fs::read(path_buf)?)),
+            (None, Some(hex)) => Ok(Cow::Borrowed(hex.as_bytes())),
+            _ => unreachable!("Clap group invariant"),
+        }
+    }
+
+    fn read_keypair<P: AsRef<Path>>(path: P) -> Result<KeyPair> {
+        let bytes = fs::read(path)?;
+        Ok(serde_json::from_slice(bytes.as_slice())?)
+    }
```
0x009922, line 96: Minor thing: these 2 methods don't seem to be related to `GenesisSigningArgs` struct. Consider externalising as plain `fn`s.

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+        if let Some(path) = self.output_path {
+            fs::write(&path, encoded_genesis_network)?;
+
+            writeln!(
+                writer,
+                "Genesis was successfully signed, encoded and written to `{}`",
+                path.display()
+            )?;
+        } else {
+            // TODO: ADD ?
```
0x009922, line 121: I think it would make sense to make `--output-path` a required arg.

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+pub struct GenesisSigningArgs {
+    /// The algorithm of the provided keypair
+    #[clap(default_value_t, long, short)]
+    algorithm: AlgorithmArg,
+    /// Private key (in string format) to sign genesis block
+    #[clap(long, group = "private_key")]
+    private_key_string: Option<String>,
+    /// Path to private key to to sign genesis block
+    #[clap(long, group = "private_key")]
+    private_key_path: Option<PathBuf>,
```
0x009922, line 39:
```suggestion
    private_key_file: Option<PathBuf>,
```

In the configuration, we agreed to use `_file` suffix for file paths, and `_dir` for dir paths. Let's follow this convention everywhere else as well.

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+    /// Public key of the corresponding private key
+    #[clap(long, group = "public_key")]
+    public_key_string: Option<String>,
+    /// Path to public key of the corresponding private key
+    #[clap(long, group = "public_key")]
+    public_key_path: Option<PathBuf>,
```
0x009922, line 45:
```suggestion
    public_key_file: Option<PathBuf>,
```

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+    /// Path to json-serialized keypair
+    #[clap(long, short, group = "private_key", group = "public_key")]
+    keypair_path: Option<PathBuf>,
```
0x009922, line 48:
```suggestion
    keypair_file: Option<PathBuf>,
```

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+    /// Unique id of blockchain
+    #[clap(long)]
+    chain_id: ChainId,
+    /// Path to genesis json file
+    #[clap(long, short)]
+    genesis_path: PathBuf,
```
0x009922, line 54:
```suggestion
    genesis_file: PathBuf,
```

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+    /// Path to signed genesis output file
+    #[clap(long, short, group = "format")]
+    output_path: Option<PathBuf>,
```
0x009922, line 57:
```suggestion
    outfile: Option<PathBuf>,
```

This name is also used by `iroha_swarm`.

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,135 @@
...
+            let (public_key, private_key) = (self.get_public_key()?, self.get_private_key()?);
+            KeyPair::new(public_key, private_key)?
+        };
+
+        let genesis_block = RawGenesisBlock::from_path(&self.genesis_path)?;
+        let encoded_genesis_network =
+            GenesisNetwork::new(genesis_block, &self.chain_id, &key_pair).encode();
+
+        if let Some(path) = self.output_path {
+            fs::write(&path, encoded_genesis_network)?;
```
0x009922, line 113: So, we take a JSON genesis file, and write a binary SCALE-encoded file back.

What are other options, should we consider them? E.g. making JSON output so that it is readable.

tools/kagami/src/crypto.rs
```diff
@@ -102,7 +223,18 @@ impl Args {
...
+    use color_eyre::Result;
+    use iroha_data_model::ChainId;
+    use iroha_genesis::GenesisNetwork;
+    use parity_scale_codec::Decode;
+
     use super::{Algorithm, AlgorithmArg};
+    use crate::{
+        crypto::{GenesisSigningArgs, KeyPairArgs},
+        RunArgs,
+    };
```
0x009922, line 237: Or just...?

```suggestion
    use super::*;
```

tools/kagami/src/crypto.rs
```diff
@@ -111,4 +243,53 @@ mod tests {
...
+            genesis_path: PathBuf::from_str(GENESIS_JSON_PATH)?,
+            output_path: Some(PathBuf::from_str(GEN_SIGNED_ENCODED_GENESIS_PATH)?),
+        };
+
+        let mut genesis_buf_writer = BufWriter::new(Vec::new());
+
+        crypto_genesis_config.run(&mut genesis_buf_writer)?;
+
+        let signed_genesis_from_file_encoded = fs::read(GEN_SIGNED_ENCODED_GENESIS_PATH)?;
+        let _maybe_signed_genesis_from_file: GenesisNetwork =
```
0x009922, line 283: Would be cool to `assert_eq!` the resulting `GenesisNetwork` with an expected one. Shouldn't take many lines for this.

tools/kagami/src/main.rs
```diff
@@ -45,7 +46,9 @@ trait RunArgs<T: Write> {
 #[command(name = "kagami", version, author)]
 enum Args {
     /// Generate cryptographic key pairs using the given algorithm and either private key or seed
-    Crypto(Box<crypto::Args>),
+    /// Or sign genesis block offline
+    #[clap(subcommand)]
+    Crypto(CryptoMode),
```
0x009922, line 51:
```suggestion
    Crypto(crypto::Args),
```

This would align with the surrounding style, imo.

tools/swarm/src/compose.rs
```diff
@@ -317,24 +317,27 @@ struct CompactPeerEnv {
     key_pair: KeyPair,
     genesis_public_key: PublicKey,
     /// Genesis private key is only needed for a peer that is submitting the genesis block
-    genesis_private_key: Option<PrivateKey>,
+    genesis_public_key_algorithm: Algorithm,
```
0x009922, line 320:
```suggestion
```

0x009922 (changes requested):

cli/src/main.rs
```diff
@@ -34,19 +34,6 @@ struct Args {
...
-    ///
-    /// This argument must be set alongside with `genesis.file` and `genesis.private_key`
-    /// configuration options. If not, Iroha will exit with an error.
-    ///
-    /// In case when the network consists only of this one peer, i.e. the amount of trusted
-    /// peers in the configuration (`sumeragi.trusted_peers`) is less than 2, this peer must
-    /// submit the genesis, since there are no other peers who can provide it. In this case, Iroha
-    /// will exit with an error if `--submit-genesis` is not set.
-    #[arg(long)]
-    submit_genesis: bool,
```
Erigara, line 49: Why this is removed? Do we assume that every peer should have genesis block provided in file?

config/src/parameters/actual.rs
```diff
@@ -103,17 +104,17 @@ impl Genesis {
...
     }
 
-    /// Access the key pair, if present
-    pub fn key_pair(&self) -> Option<&KeyPair> {
-        match self {
-            Self::Partial { .. } => None,
-            Self::Full { key_pair, .. } => Some(key_pair),
-        }
-    }
+    // Access the key pair, if present
```
Erigara, line 111: Commented code, don't forget to remove.

hooks/pre-commit.sample
```diff
@@ -2,6 +2,8 @@
 set -e
 cargo +nightly fmt --all -- --check
 cargo +nightly lints clippy --workspace --benches --tests --examples --all-features
-cargo run --bin kagami -- genesis >configs/peer/genesis.json
+cargo run --bin kagami -- genesis >configs/swarm/genesis.json
 cargo run --bin kagami -- schema >docs/source/references/schema.json
-git add configs/peer/genesis.json docs/source/references/schema.json
+cargo run --bin kagami -- crypto key-pair-generation -j >configs/swarm/keypair.json
+cargo run --bin kagami -- crypto --keypair-path configs/swarm/keypair.json --genesis-path configs/swarm/genesis.json --chain-id 01234 -o configs/swarm/signes_genesis.scale
+git add configs/peer/genesis.json configs/swarm/keypair.json configs/swarm/signes_genesis.scale docs/source/references/schema.json
```
Erigara, line 9: I'm not sure that we should store signed genesis in the repo

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,169 @@
...
+}
+
+impl<T: Write> RunArgs<T> for GenesisSigningArgs {
+    fn run(self, writer: &mut BufWriter<T>) -> Outcome {
+        let key_pair = if let Some(path) = self.keypair_path {
+            Self::read_keypair(path)?
+        } else {
+            let (public_key, private_key) = (self.get_public_key()?, self.get_private_key()?);
+            KeyPair::new(public_key, private_key)?
+        };
```
Erigara, line 140: If we always require `KeyPair` maybe we can simplify code to only allow keypair?

tools/kagami/src/crypto.rs
Erigara: I would prefer to be as simple as possible here.

tools/kagami/src/crypto.rs
VAmuzing: I've tried to make it work. It does not due to different creation times thus payload is different.
Do you have something in mind on how to bypass this?

tools/kagami/src/crypto.rs
VAmuzing: Added an option to output signed genesis block as json to stdout

tools/kagami/src/crypto.rs
VAmuzing: I thought that it might be complex to pass KeyPair as a command line argument. And I thought that we don't want to enforce user to pass it through the file, so I decided to support both options

tools/kagami/src/crypto.rs
mversic: either we output signed genesis as JSON or we output it as SCALE. IMHO JSON is better

tools/kagami/src/main.rs
```diff
@@ -57,7 +59,10 @@ impl<T: Write> RunArgs<T> for Args {
         use Args::*;
 
         match self {
-            Crypto(args) => args.run(writer),
+            Crypto(args) => match args {
+                crypto::Args::SignGenesis(args) => args.run(writer),
```
mversic, line 63:
```suggestion
                crypto::Args::SignTransaction(args) => args.run(writer),
```

let's make it a bit more general

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,201 @@
...
+    genesis_file: PathBuf,
+    /// Output signed genesis block in JSON format
+    #[clap(long, short, default_value_t = true, group = "format")]
+    json: bool,
+    /// Encode signed genesis block with SCALE (it is only supported with file output)
+    #[clap(long, short, default_value_t = false, group = "format")]
+    scale: bool,
+    /// Path to signed genesis output file (stdout by default)
+    #[clap(long, short)]
+    outfile: Option<PathBuf>,
```
mversic, line 64:
```suggestion
    out_file: Option<PathBuf>,
```

@0x009922 we seem to be mixing these 2. I prefer `out_file` but it's fine either way as long as it's consistent

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,201 @@
...
+    private_key_file: Option<PathBuf>,
+    /// Public key in multihash format of the corresponding private key
+    #[clap(long, group = "public_key")]
+    public_key_string: Option<String>,
+    /// Path to public key in multihash format of the corresponding private key
+    #[clap(long, group = "public_key")]
+    public_key_file: Option<PathBuf>,
+    /// Path to json-serialized keypair
+    #[clap(long, short, group = "private_key", group = "public_key")]
+    keypair_file: Option<PathBuf>,
```
mversic, line 49: these thing are unnecessary, remove them. You are signing a transaction, not a block

genesis/src/lib.rs
```diff
@@ -27,12 +25,12 @@ pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
...
 /// Genesis transaction
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Decode, Encode, Deserialize, Serialize)]
 #[repr(transparent)]
 pub struct GenesisTransaction(pub SignedTransaction);
 
 /// [`GenesisNetwork`] contains initial transactions and genesis setup related parameters.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Decode, Encode, Deserialize, Serialize)]
 pub struct GenesisNetwork {
```
mversic, line 34: we want genesis to be exactly 1 transaction, not a vec

mversic (changes requested):

genesis/src/lib.rs
mversic: you can remove `GenesisNetwork` and just use `GenesisTransaction` everywhere

tools/kagami/src/crypto.rs
0x009922: I don't like the idea that the final configuration of the peer will consist of the SCALE encoded genesis blob. Instead, I think, it should include the original untouched `genesis.json` with an external file containing signatures.

For example, `kagami crypto sign-genesis` (or `kagami genesis sign`) command might get a `genesis.json` as input and produce `genesis-signatures.json` as the output. It will contain an array of hexes - signatures of each genesis transaction:

```json
["<hex for tx1 (executor)>", "<hex for tx2>", "<...>"]
```

Then, this file with signatures must be specified in the config for the peer started with `--submit-genesis` arg:

```toml
[genesis]
public_key = "..."
block_file = "genesis.json"
signatures_file = "genesis-signatures.json"
```

tools/kagami/src/crypto.rs
0x009922: Oh, that's unfortunate that there is this side effect sneaking deeply inside.

IMO this usage of `current_time()` should be moved away from depths of data model, to some peripheral layers of code:

https://github.com/hyperledger/iroha/blob/96bf9973c24eb7789c5389fc66890fb2f277d563/data_model/src/transaction.rs#L641

Or add a way to mock what `current_time` in tests?

hooks/pre-commit.sample
0x009922: Docker compose configurations contain key pairs and what not. It is necessary to be able to just open the folder and `docker-compose up`, so I think it's inevitable to have such data there.

tools/kagami/src/crypto.rs
mversic: > I don't like the idea that the final configuration of the peer will consist of the SCALE encoded genesis blob. Instead, I think, it should include the original untouched genesis.json with an external file containing signatures.

we can agree that it shouldn't be a SCALE encoded blob. But why take 2 files in the configuration when you can take only 1 containing JSON serialized signed genesis transaction instead of 

> signatures of each genesis transaction:

there should be exactly 1 genesis transaction to avoid the possibility of malicious peer removing or reordering some transactions. Or we can have a fixed amount of genesis transactions where position of each is hardcoded. For example, we can expect 1st transaction to be executor and 2nd to be everything else

tools/kagami/src/crypto.rs
0x009922: I agree. Having both a `--key-pair` and `--private-key`/`--public-key` is good for CLI, IMO.

Although, I just realized that `--public-key` is redundant, as it might be derived from the private key.

mversic: > Although, I just realized that --public-key is redundant, as it might be derived from the private key. Maybe even having only --private-key is enough.

it cannot always be derived depending on which algorithm is used

mversic: > I agree. Having both a --key-pair and --private-key/--public-key is good for CLI, IMO.

yes if they are mutually exclusive

tools/kagami/src/crypto.rs
mversic: > But why take 2 files in the configuration when you can take only 1 containing JSON serialized signed genesis transaction instead of

it's not a bad idea, I just wanted genesis to have the layout of `SignedTransaction` because of sdk issues

0x009922: > there should be exactly 1 genesis transaction

Indeed, I forgot it. So it's even simpler: we need produce a single hex.

> But why take 2 files in the configuration when you can take only 1 containing JSON serialized signed genesis transaction instead

It is confusing in many minor practical aspects:

- you produce almost a copy of the original file, but with a small addition
- if you deploy Iroha (i.e. you deal with all this signing), you start to bother about having both copies, storing the original one somewhere
- oh, I messed something, let me re-sign the block with a different key! Wait.. can I re-sign the signed block, or only the empty one? Let me find the original one, where is it...

Actually, to make things **really** simple:

```toml
[genesis]
public_key = "<hex>"
file = "genesis.json"
signature = "<hex>"
```

tools/kagami/src/crypto.rs
0x009922: Agree. I prefer `--out-file` too.

tools/kagami/src/crypto.rs
VAmuzing: They are mutually exclusive which is enforced by clap grouping. I think it's pretty convenient to have keypair in json and load it from the file. And the same goes for passing public and private key to cli

mversic: why is public key ever necessary when signing genesis?

mversic: I see it's because of the fact that we pack public keys with signatures

tools/kagami/src/crypto.rs
VAmuzing: Left it as optional as we will output genesis in json to stdout as well

cli/src/main.rs
VAmuzing: I thought that initially (https://github.com/hyperledger/iroha/issues/4225#issuecomment-1953390635) but now judging by the discussion in this pr I might restore this option

tools/kagami/src/crypto.rs
VAmuzing: Added `encoded_config` parameter that contains hex string which encodes 3 parameters: `chain_id`, `creation_time` and `signatures`

cli/src/main.rs
VAmuzing: Rolled back these changes

genesis/src/lib.rs
VAmuzing: Changed to operate with GenesisTransaction, although it's more convenient to get GenesisNetwork directly at the final step

genesis/src/lib.rs
```diff
@@ -27,12 +29,81 @@ pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
...
+        raw_block: RawGenesisBlock,
+        chain_id: &ChainId,
+        genesis_key_pair: &KeyPair,
+    ) -> GenesisTransaction {
+        Self(raw_block.unify().sign(chain_id.clone(), genesis_key_pair))
+    }
+}
+
+/// [`SignedGenesisConfig`] contains data that is used for loading signed genesis from config.
+#[derive(Debug, Clone, Deserialize, Serialize, Decode, Encode)]
```
0x009922, line 48: `serde` derives are possibly not needed for this struct?

genesis/src/lib.rs
```diff
@@ -27,12 +29,81 @@ pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
...
+            chain_id,
+            creation_time_ms,
+            signatures,
+        }
+    }
+
+    /// Checks that [`SignedGenesisConfig`] corresponds to [`RawGenesisBlock`] and produces [`GenesisNetwork`] if it does.
+    /// # Errors
+    /// Fails if [`RawGenesisBlock`] does not correspond to [`SignedGenesisConfig`] and it was unable to verify it's integrity
+    pub fn validate(self, genesis_block: RawGenesisBlock) -> Result<GenesisNetwork> {
```
0x009922, line 72: This function seems to belong more to `GenesisNetwork` itself:

```rust
impl GenesisNetwork {
    fn try_parse(raw_block: RawGenesisBlock, signature: RawGenesisBlockSignature) -> Result<Self> {}
}
```

0x009922: Also, if we stick to the article ["Parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), this is the case of __parse__.

genesis/src/lib.rs
```diff
@@ -27,12 +29,81 @@ pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
...
+        chain_id: &ChainId,
+        genesis_key_pair: &KeyPair,
+    ) -> GenesisTransaction {
+        Self(raw_block.unify().sign(chain_id.clone(), genesis_key_pair))
+    }
+}
+
+/// [`SignedGenesisConfig`] contains data that is used for loading signed genesis from config.
+#[derive(Debug, Clone, Deserialize, Serialize, Decode, Encode)]
+pub struct SignedGenesisConfig {
```
0x009922, line 49: I am _really_ confused with this name, as well with the `genesis.encoded_config` configuration parameter.

Conceptually, it is a composite signature of the genesis, as a whole. So, the name `GenesisSignature` may work. And `genesis.signature` in the config would make more sense.

genesis/src/lib.rs
```diff
@@ -86,6 +157,14 @@ impl RawGenesisBlock {
     pub fn from_path<P: AsRef<Path>>(path: P) -> Result<Self> {
         RawGenesisBlockFile::from_path(path)?.try_into()
     }
+
+    fn unify(self) -> GenesisTransactionBuilder {
+        let mut unified_tx = GenesisTransactionBuilder {
+            isi: vec![Upgrade::new(self.executor).into()],
+        };
+        unified_tx.extend_instructions(self.transactions.iter().flat_map(|v| v.isi.clone()));
```
0x009922, line 165: Since genesis block now should have only one transaction, there should not be `transactions` field. It could be possibly renamed to `instructions`.

scripts/test_env.py
```diff
@@ -143,10 +146,25 @@ def __init__(self, args: argparse.Namespace, nth: int):
                 logging.error(f"Some of the config files are missing. \
                                           Please provide them in the `{target}` directory")
                 sys.exit(1)
+
+            sign_command = [self.out_dir / "kagami", "crypto", "sign-transaction", "--chain-id", CHAIN_ID,
```
0x009922, line 150: Choice of `kagami crypto sign-transaction --genesis-file ... ...` looks a little bit out of place. I would more like `kagami genesis sign ...`, because it is closely related to genesis.

configs/peer.template.toml
```diff
@@ -16,7 +16,6 @@
 [genesis]
 # file =
 # public_key =
-# private_key = { algorithm = "", payload = "" }
```
0x009922, line 19: Remember to add the new parameter here.

data_model/src/transaction.rs
```diff
@@ -320,6 +320,29 @@ impl SignedTransaction {
     }
 }
 
+#[cfg(feature = "transparent_api")]
+impl TryFrom<(SignaturesOf<TransactionPayload>, TransactionPayload)> for SignedTransaction {
+    type Error = &'static str;
```
0x009922, line 325: `iroha_data_model` is a highly-domain-specific module, thus it's better to use stricter, domain-focused errors.

In other words, please define this error as an `enum` deriving `thiserror::Error`.

genesis/Cargo.toml
```diff
@@ -12,15 +12,17 @@ workspace = true
 
 [dependencies]
 iroha_crypto = { workspace = true }
-iroha_data_model = { workspace = true, features = ["http"] }
+iroha_data_model = { workspace = true, features = ["http", "transparent_api"] }
```
0x009922, line 15: Is it OK to use `transparent_api` here? @mversic @Erigara

tools/kagami/src/crypto.rs
```diff
@@ -1,14 +1,207 @@
...
+            KeyStorage::FromFile(bytes) => {
+                PublicKey::from_bytes(self.algorithm.0, bytes.as_slice()).wrap_err_with(|| {
+                    eyre!(
+                        "Failed to parse public key from bytes for algorithm `{}`",
+                        self.algorithm
+                    )
+                })
+            }
+            KeyStorage::FromCLI(hex) => {
+                if hex.starts_with(ED25519_PREFIX) {
```
0x009922, line 124: `PublicKey` is always represented as a multihash string. Look at other places where `PublicKey` is parsed from external input and re-use that way, please.

tools/kagami/src/crypto.rs
```diff
@@ -102,7 +295,9 @@ impl Args {
 
 #[cfg(test)]
 mod tests {
-    use super::{Algorithm, AlgorithmArg};
+    // use iroha_data_model::transaction::{SignedTransactionV1, TransactionPayload};
```
0x009922, line 298: Remember to remove.

tools/kagami/src/crypto.rs
```diff
@@ -111,4 +306,76 @@ mod tests {
...
+        Ok(cmp(
+            &signed_genesis_from_config.0,
+            &signed_genesis_manually.0,
+        ))
+    }
+
+    #[test]
+    fn test_genesis_signing_works() {
+        let result = genesis_signing_works();
+        let _ = fs::remove_file(GEN_KEYPAIR_JSON_PATH);
```
0x009922, line 378: Use `tempfile::NamedTempFile`.

tools/swarm/src/compose.rs
```diff
@@ -92,7 +92,7 @@ pub struct DockerComposeServiceBuilder {
     volumes: Vec<(String, String)>,
     trusted_peers: BTreeSet<PeerId>,
     genesis_public_key: PublicKey,
-    genesis_private_key: Option<PrivateKey>,
+    signed_genesis_file: Option<PathBuf>,
```
0x009922, line 95: _Probably, you haven't updated this part yet; if so, ignore this comment._

So, the genesis file remains the same, always unsigned. Just `genesis_file`. What this structure is lacking now is `genesis_signature` field, that should be represented as `GENESIS_SIGNATURE` env var.

0x009922 (changes requested):

genesis/src/lib.rs
VAmuzing: Thank you for the read, I agree with you on that!

scripts/test_env.py
VAmuzing: Initially I thought that it would be better placed within crypto but after all these changes I moved it to genesis

genesis/src/lib.rs
VAmuzing: Renamed the field to `transaction` and changes its type.

genesis/Cargo.toml
VAmuzing: As it was necessary to somehow setup a creation time I decided to make it public in the module with this feature. Maybe I'm missing some feature that would enable me to do just that without `transparent_api`.

config/tests/fixtures/absolute_paths.toml
```diff
@@ -11,4 +11,5 @@ out_file = "/telemetry/file.json"
 
 [genesis]
 file = "/oh/my/genesis.json"
-private_key = { algorithm = "ed25519", payload = "8f4c15e5d664da3f13778801d23d4e89b76e94c1b94b389544168b6cb894f84f8ba62848cf767d72e7f7f4b9d2d7ba07fee33760f79abe5597a51520e292a0cb" }
\ No newline at end of file
+public_key = "ed01208BA62848CF767D72E7F7F4B9D2D7BA07FEE33760F79ABE5597A51520E292A0CB"
```
0x009922, line 14: `public_key` is specified in `base.toml`, so here it is redundant for tests. Please revert.

config/tests/fixtures/minimal_alone_with_genesis.toml
```diff
@@ -2,5 +2,5 @@ extends = "base.toml"
 
 [genesis]
 file = "./empty_ok_genesis.json"
-private_key.algorithm = "ed25519"
-private_key.payload = "8f4c15e5d664da3f13778801d23d4e89b76e94c1b94b389544168b6cb894f84f8ba62848cf767d72e7f7f4b9d2d7ba07fee33760f79abe5597a51520e292a0cb"
+public_key = "ed01208BA62848CF767D72E7F7F4B9D2D7BA07FEE33760F79ABE5597A51520E292A0CB"
```
0x009922, line 5: Same as with another file: redundant `public_key` here.

config/src/parameters/actual.rs
```diff
@@ -100,27 +100,20 @@ pub enum Genesis {
...
     Full {
-        /// Genesis account key pair
-        key_pair: KeyPair,
-        /// Path to the [`RawGenesisBlock`]
+        /// Genesis account public key
+        public_key: PublicKey,
+        /// Path to the genesis file
         file: PathBuf,
+        /// Hex-encoded genesis config
+        signature: String,
```
0x009922, line 108: Consider using a stronger type right away. Possibly a newtype around `iroha_genesis::GenesisSignature` which will implement `Deserialize` and `FromStr` from a hex string.

config/src/parameters/user/boilerplate.rs
```diff
@@ -227,22 +228,20 @@ impl FromEnv for GenesisPartial {
...
-            "GENESIS_PRIVATE_KEY",
-            "genesis.private_key",
-        )
-        .into();
+
         let file =
             ParseEnvResult::parse_simple(&mut emitter, env, "GENESIS_FILE", "genesis.file").into();
 
+        let signature =
+            ParseEnvResult::parse_simple(&mut emitter, env, "SIGNATURE", "genesis.signature")
```
0x009922, line 236: `GENESIS_SIGNATURE` would be consistent with others

core/test_network/src/lib.rs
```diff
@@ -64,6 +64,18 @@ pub fn get_key_pair() -> KeyPair {
     ).unwrap()
 }
 
+/// Get hex string represented genesis_config for ./genesis.json that is tied to hardcoded keys and chain_id
+pub fn get_genesis_encoded_signature() -> String {
```
0x009922, line 68: It's always possible to encode `GenesisSignature` when needed. It's better to operate with stronger types when possible. Thus,

```rust
pub fn get_genesis_signature() -> GenesisSignature {
```

cli/src/samples.rs
```diff
@@ -63,6 +63,7 @@ pub fn get_user_config(
     peers: &UniqueVec<PeerId>,
     chain_id: Option<ChainId>,
     key_pair: Option<KeyPair>,
+    encoded_signature: String,
```
0x009922, line 66: `genesis_signature: GenesisSignature`

config/iroha_test_config.toml
```diff
@@ -8,7 +8,6 @@ address = "127.0.0.1:1337"
 [genesis]
 public_key = "ed01204CFFD0EE429B1BDD36B3910EC570852B8BB63F18750341772FB46BC856C5CAAF"
 file = "./genesis.json"
-private_key = { algorithm = "ed25519", payload = "D748E18CE60CB30DEA3E73C9019B7AF45A8D465E3D71BCC9A5EF99A008205E534CFFD0EE429B1BDD36B3910EC570852B8BB63F18750341772FB46BC856C5CAAF" }
```
0x009922, line 11: If you remove `private_key`, than you should put `signature`, right?

0x009922 (changes requested):

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,248 @@ fn generate_synthetic(
...
+#[command(group = ArgGroup::new("private_key").required(true))]
+#[command(group = ArgGroup::new("public_key").required(true))]
+#[command(group = ArgGroup::new("format").required(false))]
+pub struct SignArgs {
+    /// The algorithm of the provided keypair
+    #[clap(default_value_t, long, short)]
+    algorithm: crypto::AlgorithmArg,
+    /// Private key (in string format) to sign genesis block
+    #[clap(long, group = "private_key")]
+    private_key_string: Option<String>,
```
0x009922, line 281: > (in string format)

Which one? A JSON with `algorithm` and `payload` fields?

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,248 @@ fn generate_synthetic(
...
+    /// Path to private key to sign genesis block
+    #[clap(long, group = "private_key")]
+    private_key_file: Option<PathBuf>,
+    /// Public key in multihash format of the corresponding private key
+    #[clap(long, group = "public_key")]
+    public_key_string: Option<String>,
```
0x009922, line 287: You can put a type which will deserialize to `PublicKey` from a string. (i.e. `PublicKey` itself!)

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,248 @@ fn generate_synthetic(
...
+    chain_id: ChainId,
+    /// Path to genesis json file
+    #[clap(long, short)]
+    genesis_file: PathBuf,
+    /// Output signed genesis config as a hex string
+    #[clap(long, default_value_t = true, group = "format")]
+    hex: bool,
+    /// Encode signed genesis block with SCALE (it is only supported with file output)
+    #[clap(long, short, default_value_t = false, group = "format")]
+    scale: bool,
```
0x009922, line 305: Why `--scale` has `-s` alias, but `--hex` doesn't?

(maybe remove `-s` alias?)

genesis/Cargo.toml
mversic: there is no need to enable `transparent_api` here. You can use `TransactionPayload::new_with_time`. This feature should only be enabled by `iroha_core`

@0x009922 I was actually thinking to rename this feature to `_transparent_api` to give people the idea it's an internal feature and dissuade them from using it. What do you think about this?

tools/kagami/src/genesis.rs
VAmuzing: As there's a field for algorithm I thought of using hex string for private key and multihash for public. I've added `hex` part

config/src/parameters/user.rs
```diff
@@ -347,38 +350,36 @@ pub(crate) fn private_key_from_env<E: Error>(
 #[derive(Debug)]
 pub struct Genesis {
     pub public_key: PublicKey,
-    pub private_key: Option<PrivateKey>,
     pub file: Option<PathBuf>,
+    pub signature: Option<String>,
```
0x009922, line 354: Please use `GenesisSignature` directly, or a type that wraps it. Just `String` is way less useful.

It is as if `public_key` would be just some `String` and not `PublicKey`. Will require more parsing steps further, while it could be done right here, seamlessly, by `serde`.

config/src/parameters/actual.rs
```diff
@@ -100,27 +122,20 @@ pub enum Genesis {
...
     /// The peer is responsible for submitting the genesis block
     Full {
-        /// Genesis account key pair
-        key_pair: KeyPair,
-        /// Path to the [`RawGenesisBlock`]
+        /// Genesis account public key
+        public_key: PublicKey,
+        /// Path to the genesis file
         file: PathBuf,
+        /// Hex-encoded genesis config
```
0x009922, line 129: Obsolete doc comment

config/tests/fixtures/full.env
```diff
@@ -4,8 +4,6 @@ PRIVATE_KEY_ALGORITHM=ed25519
 PRIVATE_KEY_PAYLOAD=8f4c15e5d664da3f13778801d23d4e89b76e94c1b94b389544168b6cb894f84f8ba62848cf767d72e7f7f4b9d2d7ba07fee33760f79abe5597a51520e292a0cb
 P2P_ADDRESS=127.0.0.1:5432
 GENESIS_PUBLIC_KEY=ed01208BA62848CF767D72E7F7F4B9D2D7BA07FEE33760F79ABE5597A51520E292A0CB
-GENESIS_PRIVATE_KEY_ALGORITHM=ed25519
-GENESIS_PRIVATE_KEY_PAYLOAD=8f4c15e5d664da3f13778801d23d4e89b76e94c1b94b389544168b6cb894f84f8ba62848cf767d72e7f7f4b9d2d7ba07fee33760f79abe5597a51520e292a0cb
```
0x009922, line 8: Here, also, if you remove private key variables, I would logically expect you to add `GENESIS_SIGNATURE` here instead.

0x009922: And, by the way, `full.env`'s purpose is to contain __all__ envs supported by Iroha config.

tools/swarm/src/compose.rs
```diff
@@ -316,23 +319,23 @@ struct CompactPeerEnv {
...
-            .map_or((None, None, None), |private_key| {
-                let (algorithm, payload) = private_key.to_bytes();
+        let (genesis_file, genesis_signature) =
+            value.genesis_file.map_or((None, None), |genesis_path| {
                 (
-                    Some(algorithm),
-                    Some(payload),
-                    Some(PATH_TO_GENESIS.to_string()),
+                    Some(genesis_path),
+                    Some(generate_hex_string_signature(
```
0x009922, line 335: Wait, why you generate a signature here?

`CompactPeerEnv` & `FullPeerEnv` are emptyheaded structs purely for data representation purposes. They don't do any such logic. `genesis_signature` should be provided to them from outside.

tools/swarm/src/compose.rs
```diff
@@ -471,6 +477,20 @@ fn generate_key_pair(base_seed: Option<&[u8]>, additional_seed: &[u8]) -> KeyPai
     })
 }
 
+fn generate_hex_string_signature(chain_id: &ChainId, genesis_key_pair: &KeyPair) -> String {
+    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let raw_block = iroha_genesis::RawGenesisBlock::from_path(
+        manifest_dir.join("../../configs/swarm/genesis.json"),
```
0x009922, line 483: `iroha_swarm` is not a tool to work only in our repo. CLI args should provide path to the genesis.

But, in this case, `iroha_swarm` will duplicate the functionality of `kagami genesis sign`. So, I think it would be OK if users provide the signature directly, i.e.:

```shell
iroha_swarm --genesis-signature 0022aafsdfew...
```

0x009922 (changes requested):
Also, do not forget that since you have updated `kagami` and `iroha_swarm`, `consistency.sh` will fail CI unless you also update their generated artifacts. This works as a reminder to us to keep tools and their artifacts in sync.

So, run the exact commands that `consistency.sh` suggests you to do and update contents of `configs/swarm` by them.

Apart from that, there are a couple of critical comments. And it is not clear for me yet how the final setup of configs & CI would look like.

genesis/Cargo.toml
0x009922: @mversic, it would make sense, I think. And if it's really only for `iroha_core`, naming it `_transparent_api_for_iroha_core` would be clearer. I myself haven't understood the scope of it until now.

tools/kagami/src/genesis.rs
0x009922: You mean, SCALE-encoded hex?
It is not a usual format which users use with Iroha for private keys. Unfortunately, we always separate `algorithm` and `payload` somehow :(

This might help: https://github.com/hyperledger/iroha/issues/4412

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,251 @@ fn generate_synthetic(
...
+    genesis_file: PathBuf,
+    /// Output signed genesis config as a hex string
+    #[clap(long, default_value_t = true, group = "format")]
+    hex: bool,
+    /// Encode signed genesis block with SCALE (it is only supported with file output)
+    #[clap(long, default_value_t = false, group = "format")]
+    scale: bool,
+    /// Path to signed genesis output file (stdout by default)
+    #[clap(long, short)]
+    out_file: Option<PathBuf>,
```
0x009922, line 308: Could you also use `--out-file` in `iroha_swarm`? It uses `--outfile` now.

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,251 @@ fn generate_synthetic(
...
+    chain_id: ChainId,
+    /// Path to genesis json file
+    #[clap(long, short)]
+    genesis_file: PathBuf,
+    /// Output signed genesis config as a hex string
+    #[clap(long, default_value_t = true, group = "format")]
+    hex: bool,
+    /// Encode signed genesis block with SCALE (it is only supported with file output)
+    #[clap(long, default_value_t = false, group = "format")]
+    scale: bool,
```
0x009922, line 305: Hm, I think, we can simplify here a little bit. We can replace both `--hex` and `--scale` with a single flag - `--binary`. With `--out-file`, it will write SCALE bytes into the given file, without - will print the same SCALE bytes in a hex format to stdout.

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,251 @@ fn generate_synthetic(
...
+        } else {
+            Vec::default()
+        };
+
+        let hex_genesis_config = if self.hex {
+            genesis_signature_string
+        } else {
+            String::default()
+        };
+
```
0x009922, line 406: Certainly a case to use `Option`s for stronger abstractions.

```rust
let scale_encoded_signature = self.scale.then(|| genesis_signature_string.encode());
```

And `hex_genesis_config`... It doesn't seem to be needed.

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,251 @@ fn generate_synthetic(
...
+            &signed_genesis_manually.0,
+        ))
+    }
+
+    #[test]
+    fn test_genesis_signing_works() {
+        let result = genesis_signing_works();
+        assert!(result.is_ok_and(|result| result));
+    }
+}
```
0x009922, line 509: Please avoid using this pattern with nesting. Just make `#[test]` of `genesis_signing_works` directly (with minor internal changes).

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,251 @@ fn generate_synthetic(
...
+            public_key_string: None,
+            public_key_file: None,
+            keypair_file: Some(PathBuf::from(tmp_keypair_json_file.path())),
+            chain_id: chain_id.clone(),
+            genesis_file: PathBuf::from_str(GENESIS_JSON_PATH)?,
+            out_file: None,
+            scale: false,
+            hex: true,
+        };
+
```
0x009922, line 471: Would be cool to construct `SignArgs` from raw CLI args, in a way I did it here:

https://github.com/hyperledger/iroha/blob/ecd311983a6af96dd4c4c5ca65acc0e89d2d1b0d/cli/src/lib.rs#L997-L1002

tools/kagami/src/genesis.rs
VAmuzing: Forgot to remove when introduced TempFile into the test

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,231 @@ fn generate_synthetic(
...
+        let tmp_keypair_json_file = tempfile::NamedTempFile::new()?;
+        fs::write(tmp_keypair_json_file.path(), keypair_json.buffer())?;
+
+        let chain_id_slice = "0123456";
+        let chain_id = ChainId::from(chain_id_slice);
+        let crypto_genesis_config = SignArgs::try_parse_from([
+            "sign",
+            "--chain-id",
+            chain_id_slice,
```
mversic, line 449: this should be in the genesis file, not given over CLI

mversic, line 449:
```suggestion
```

this should be in the genesis file, not given over CLI

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,231 @@ fn generate_synthetic(
...
+            "--genesis-file",
+            GENESIS_JSON_PATH,
```
mversic, line 451:
```suggestion
            GENESIS_JSON_PATH,
```

why isn't this positional argument?

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,231 @@ fn generate_synthetic(
...
     }
 
     Ok(genesis)
 }
+
+/// Use `Kagami` to sign genesis block.
+#[derive(Clone, Debug, Parser)]
+#[command(group = ArgGroup::new("private_key").required(true))]
+#[command(group = ArgGroup::new("public_key").required(true))]
+pub struct SignArgs {
```
mversic, line 274: This should not be in this module. We want facility to sign all transactions, not just genesis. Genesis is only a special case

mversic: @0x009922 do we put it in `kagami crypto` or will we have another command?

mversic: now that I think about it the command should be like this `kagami crypto sign file`. It can sign any file. Contents of the file should not be relevant. Whether it's transaction or genesis transaction or query

tools/kagami/src/genesis.rs
VAmuzing: Correct me if I wrong, but I don't see genesis file containing chain id.
And it seems like every time we create genesis, be it inside the swarm or core libs, we supply chain id explicitly.
So where do we find it besides taking it over CLI?

tools/kagami/src/genesis.rs
VAmuzing: I don't have an answer to that. It seems like we don't positional args in both swarm and kagami and I thought not to break the "tradition".
I thought specifying the argument explicitly makes it less error prone, don't you agree?
@0x009922 what do you think?

tools/kagami/src/genesis.rs
VAmuzing: At the moment I rely on the fact that the file is indeed genesis file - there are verifications for executor via `RawGenesisBlock` struct. 
What if I move that logic to crypto and leave it somewhat "as is" and we create a task to extend the functionality somewhere down the road? We might never need it and then we just abandon the idea.
Will it be fine?

tools/kagami/src/genesis.rs
mversic: > Correct me if I wrong, but I don't see genesis file containing chain id.

Yes, add it to the file

tools/kagami/src/genesis.rs
0x009922: @mversic, I proposed to move it to `kagami genesis` from `kagami crypto` deliberately.

It's not just signing of a file. In case of `genesis.json`, Kagami needs to know that it is a genesis file, to decoded it from JSON properly, read `executor_file`, combine it all into a single genesis transaction, and then produce a signature of it. It is quite different from signing any file, interpreting it as just some blob of bytes.

Therefore, I still think `kagami genesis` is a proper place for it, since it is tightly coupled to the genesis.

tools/kagami/src/genesis.rs
0x009922: I think either way is fine, but I have a _feeling_ that explicit argument would be more consistent overall.

genesis/src/lib.rs
```diff
@@ -28,16 +31,121 @@ pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
...
 pub struct GenesisTransaction(pub SignedTransaction);
 
+impl GenesisTransaction {
+    /// Construct single genesis transaction with [`Executor`] and transactions unified under a single transaction
+    pub fn new_unified(
+        raw_block: RawGenesisBlock,
+        chain_id: &ChainId,
+        genesis_key_pair: &KeyPair,
+    ) -> GenesisTransaction {
+        Self(raw_block.unify().sign(chain_id.clone(), genesis_key_pair))
```
0x009922, line 45: Leaves allocation to the caller if they need to:

```suggestion
        chain_id: ChainId,
        genesis_key_pair: &KeyPair,
    ) -> GenesisTransaction {
        Self(raw_block.unify().sign(chain_id, genesis_key_pair))
```

genesis/src/lib.rs
```diff
@@ -28,16 +31,121 @@ pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
...
+    HexDecodeError,
+    ///  The config has incorrect scale format and cannot be decoded
+    ScaleDecodeError,
+}
+
+impl std::error::Error for GenesisSignatureParseError {}
+
+/// [`SignedGenesisConfig`] contains data that is used for loading signed genesis from config.
+#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode)]
+pub struct GenesisSignature {
```
0x009922, line 62: As a small note, `iroha_genesis::GenesisSignature` contains name duplication, and may be shortened to `iroha_genesis::Signature`. Up to your preference here.

Same applies to `GenesisSignatureParseError`, `GenesisNetwork` etc.

genesis/src/lib.rs
```diff
@@ -47,24 +155,54 @@ impl GenesisNetwork {
...
+    }
+
+    /// Construct `GenesisSignature` from config
+    /// # Errors
+    /// Fails if the
+    pub fn new_genesis_signature(
+        raw_block: RawGenesisBlock,
+        chain_id: &ChainId,
+        genesis_key_pair: &KeyPair,
+    ) -> GenesisSignature {
```
0x009922, line 173: - It is not related to `impl GenesisNetwork`
- `# Errors` is irrelevant

Consider moving it to `RawGenesisBlock`:

```rust
impl RawGenesisBlock {
  pub fn sign(&self, chain_id: &ChainId, key_pair: &KeyPair) -> GenesisSignature;
}
```

genesis/src/lib.rs
```diff
@@ -275,7 +430,7 @@ impl<S> RawGenesisBlockBuilder<S> {
 
     /// Add instruction to the end of genesis transaction
     pub fn append_instruction(mut self, instruction: impl Into<InstructionBox>) -> Self {
```
0x009922, line 432: I guess it should be renamed too:

```suggestion
    pub fn push_instruction(mut self, instruction: impl Into<InstructionBox>) -> Self {
```

Or better keep using `append_instruction` everywhere - this will reduce code changes!

tools/kagami/src/genesis.rs
0x009922: Please replace `String` with a type that shows _what_ this string is, which shows how it parses it.

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,231 @@ fn generate_synthetic(
...
+    /// Path to public key in multihash format of the corresponding private key
+    #[clap(long, group = "public_key")]
+    public_key_file: Option<PathBuf>,
+    /// Path to json-serialized keypair
+    #[clap(long, short, group = "private_key", group = "public_key")]
+    keypair_file: Option<PathBuf>,
+    /// Unique id of blockchain
+    #[clap(long)]
+    chain_id: ChainId,
+    /// Encode signed genesis block with SCALE (it is only supported with file output)
```
0x009922, line 299: - If it works only with file output, please enforce it with a clap requirement on `--out-file`
- Without file output, you can still print a hex
- Actually, what do we have `--out-file` for?.. The final purpose is to produce a hex string which user will put into the `genesis.signature` config field (or `GENESIS_SIGNATURE` env). I don't see a purpose of having a binary file output. What do I miss?

tools/kagami/src/genesis.rs
```diff
@@ -241,10 +259,231 @@ fn generate_synthetic(
...
+#[derive(Clone, Debug, Parser)]
+#[command(group = ArgGroup::new("private_key").required(true))]
+#[command(group = ArgGroup::new("public_key").required(true))]
+pub struct SignArgs {
+    /// Path to genesis json file
+    #[clap(long)]
+    genesis_file: PathBuf,
+    /// The algorithm of the provided keypair
+    #[clap(default_value_t, long, short)]
+    algorithm: crypto::AlgorithmArg,
```
0x009922, line 280: The algorithm is included into the keypair itself. It is contained in both `PublicKey` and `PrivateKey`.

tools/kagami/src/genesis.rs
0x009922: Umm, duplicating the chain id in the genesis is unfortunate =(

Then we will have to decide what is the source of truth - Chain ID in the config file or in the genesis file? We will also have to validate their equality during config parsing.

So, maybe `--chain-id` is fine? Another option might be to provide a relevant `--config-file`, from where `kagami` can read `chain_id`.

tools/swarm/src/compose.rs
```diff
@@ -303,39 +309,27 @@ struct FullPeerEnv {
     private_key_payload: Vec<u8>,
     p2p_address: SocketAddr,
     api_address: SocketAddr,
-    genesis_public_key: PublicKey,
-    genesis_private_key_algorithm: Option<Algorithm>,
-    #[serde_as(as = "Option<serde_with::hex::Hex>")]
-    genesis_private_key_payload: Option<Vec<u8>>,
     genesis_file: Option<String>,
+    genesis_public_key: PublicKey,
+    genesis_signature: Option<String>,
```
0x009922, line 314:
```suggestion
    #[serde_as(as = "Option<serde_with::hex::Hex>")]
    genesis_signature: Option<iroha_genesis::GenesisSignature>,
```

tools/swarm/src/cli.rs
```diff
@@ -10,6 +10,12 @@ pub struct Cli {
     /// The Unicode `seed` string for deterministic key-generation.
     #[arg(long, short)]
     pub seed: Option<String>,
+    /// Explicit keypair in json format. Takes precedence over seed parameter.
+    #[arg(long)]
+    pub key_pair: Option<String>,
+    /// Genesis signature. Should be set with --key-pair.
+    #[arg(long, requires = "key_pair")]
+    pub signature: Option<String>,
```
0x009922, line 18: Why?

Swarm only needs to support `--genesis-signature` argument, and simply pass it down to the generated `yaml` in a form of `GENESIS_SIGNATURE` env.

tools/swarm/src/compose.rs
0x009922, line 1: This file mostly contains changes I think don't make sense. Please see my comment on `cli.rs`. I think changes from there will cascade down and fix things here.

scripts/tests/consistency.sh
```diff
@@ -20,25 +20,28 @@ case $1 in
...
             eval "$full_cmd"
             diff "$temp_file" "$target" || {
-                echo "Please re-generate \`$target\` with \`$cmd_base --outfile $target\`"
+                echo "Please re-generate \`$target\` with \`$cmd_base --out-file $target\`"
                 exit 1
             }
         }
 
+        genesis_keypair="--key-pair '{\"public_key\": \"ed01208BA62848CF767D72E7F7F4B9D2D7BA07FEE33760F79ABE5597A51520E292A0CB\",\"private_key\": {\"algorithm\": \"ed25519\",\"payload\": \"8f4c15e5d664da3f13778801d23d4e89b76e94c1b94b389544168b6cb894f84f8ba62848cf767d72e7f7f4b9d2d7ba07fee33760f79abe5597a51520e292a0cb\"}}'"
+        genesis_signature="--signature 9030303030303030302d303030302d303030302d303030302d3030303030303030303030308c4b28660000000000811b2c0400808ba62848cf767d72e7f7f4b9d2d7ba07fee33760f79abe5597a51520e292a0cb0101819b61a8c82fc6565c5acc7f92ca6dc07ccd39fcc47dee3d6f85d59ec7b52fe651b20b985b401546b49a50057b4311beaccb822b8ef2cf4e25ce3fab3769c70e"
```
0x009922, line 33: `iroha_swarm` needs to know only the signature

config/src/parameters/user.rs
```diff
@@ -179,8 +181,8 @@ impl Root {
                 emitter.emit(eyre!("\
                     The network consists from this one peer only (no `sumeragi.trusted_peers` provided). \
                     Since `--submit-genesis` is not set, there is no way to receive the genesis block. \
-                    Either provide the genesis by setting `--submit-genesis` argument, `genesis.private_key`, \
-                    and `genesis.file` configuration parameters, or increase the number of trusted peers in \
+                    Either provide the genesis by setting `--submit-genesis` argument, `genesis.file`, \
+                    `genesis.public_key` and `genesis.signature` configuration parameters, or increase the number of trusted peers in \
```
0x009922, line 185: - The only global change is renaming of `genesis.private_key` to `genesis.signature`
- `genesis.public_key` is a required parameter anyway, no need to mention it here IMO.

This will produce a diff minimal to the original:

```suggestion
                    Either provide the genesis by setting `--submit-genesis` argument, `genesis.signature`, \
                    and `genesis.file` configuration parameters, or increase the number of trusted peers in \
```

config/src/parameters/actual.rs
```diff
@@ -100,27 +126,20 @@ pub enum Genesis {
...
     Full {
-        /// Genesis account key pair
-        key_pair: KeyPair,
-        /// Path to the [`RawGenesisBlock`]
+        /// Genesis account public key
+        public_key: PublicKey,
+        /// Path to the genesis file
         file: PathBuf,
+        /// Genesis signature config with the data to reconstruct signed genesis block
+        signature: GenesisSignatureConfig,
```
0x009922, line 134: `parameters::actual` is exposed to the rest of Iroha. Those places don't really need to know about config parsing details.

```suggestion
        signature: iroha_genesis::GenesisSignature,
```

config/src/parameters/actual.rs
```diff
@@ -90,6 +91,31 @@ pub struct Network {
     pub idle_timeout: Duration,
 }
 
+/// Wrapper around `iroha_genesis::GenesisSignature` for type-checking
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct GenesisSignatureConfig(iroha_genesis::GenesisSignature);
```
0x009922, line 96: It is a newtype needed to handle parsing of `GenesisSignature` from the user-provided format. It's place more in the `parameters::user` IMO.

config/src/parameters/actual.rs
```diff
@@ -90,6 +91,31 @@ pub struct Network {
...
+
+impl GenesisSignatureConfig {
+    /// Constructs `GenesisSignatureConfig` from `GenesisSignature`
+    pub fn new(genesis_signature: iroha_genesis::GenesisSignature) -> Self {
+        Self(genesis_signature)
+    }
+    /// Converts `GenesisSignatureConfig` into `iroha_genesis::GenesisSignature`
+    pub fn into_genesis_signature(self) -> iroha_genesis::GenesisSignature {
+        self.0
+    }
```
0x009922, line 106:
```suggestion
    pub fn into_inner(self) -> iroha_genesis::GenesisSignature {
        self.0
    }
```

0x009922 (changes requested):
Sorry for torturing with changes requests, but there are still major questionable things here =(

core/test_network/src/lib.rs
```diff
@@ -61,6 +61,15 @@ pub fn get_key_pair() -> KeyPair {
     ).unwrap()
 }
 
+/// Get hex string represented genesis_config for ./genesis.json that is tied to hardcoded keys and chain_id
+pub fn get_genesis_signature() -> GenesisSignature {
+    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let genesis = RawGenesisBlock::from_path(manifest_dir.join("../../configs/swarm/genesis.json"))
```
0x009922, line 67: Shouldn't genesis signature be computed from a genesis block that is generated within tests? I remember that different tests generate it dynamically depending on their needs.

tools/kagami/src/genesis.rs
mversic: I have a _feeling_ that if you're signing something, it's not optional and should be a positional arg.
CLI apps should have one positional argument, that is the object of operation

tools/kagami/src/genesis.rs
mversic: am I missing something. When did we deviate from [this](https://github.com/hyperledger/iroha/issues/4225#issuecomment-1905464604)? Genesis file should exactly contain one JSON serialized `SignedTransaction`. We don't want no special genesis transaction type if we can avoid it

To deserialize this JSON file, `kagami` only needs to know that it's dealing with any `SignedTransaction` not that it's genesis. So it can work for any transaction. To make it work for multiple different types we `kagami` can try deserializing into multiple types `SignedTransaction`, `SignedQuery` (is there any other?) or it can take a CLI argument such as `--type query`

VAmuzing: Yes.
Here, you approved @0x009922's [idea](https://github.com/hyperledger/iroha/pull/4382/#discussion_r1540642426)

mversic: Yes, but I don't understand how that is related to the discussion here

VAmuzing: Let me know if I don't understand you.
You suggest to sign genesis offline and keep it as a `SignedTransaction`. But previously we agreed on separating "signature" and genesis file itself as a separate entities.

mversic: Just leave the signatures field empty in the file. It's a list so it can be empty

0x009922: > am I missing something. When did we deviate from https://github.com/hyperledger/iroha/issues/4225#issuecomment-1905464604? Genesis file should exactly contain one JSON serialized SignedTransaction. We don't want no special genesis transaction type if we can avoid it

Indeed, we deviated. I pictured the solution differently.

So, @mversic, you propose to make `genesis.json` to be a direct representation of `SignedTransaction`. Which means, it will contain a complete payload and its signature(s) in a final format. It won't have a convenience such as `"executor_file": "./executor.wasm"`.

If so, than I am confused, as it is going to increase complexity of what users need to do to prepare the genesis. They will need to prepare some unsigned transaction, put an "Upgrade" with executor into it, then sign it. Along the way they should also put chain id somewhere in the right place (even I don't know where exactly). The resulting `genesis.json` will also be bloated due to the inlined executor binary.

In the meantime, we can:

- keep current "convenient" `genesis.json` (with an addition of `chain_id` parameter)
- specify it by `genesis.file` config
- produce its signature by `kagami genesis sign genesis.json`
- put the signature into `genesis.signature` config

I might be missing something.

mversic: Ok. The point about executor makes sense. Can someone write an example of how genesis will look like?

0x009922: I expect it to remain essentially the same. Only with addition of `chain_id`, and changing of `transactions: Instruction[][]` to `transaction: Instruction[]`

```json
{
  "chain_id": "0-0",
  "transaction": [
    {
      "Register": {
        "Domain": {
          "id": "wonderland",
          "logo": null,
          "metadata": {
            "key": {
              "String": "value"
            }
          }
        }
      }
    }
  ],
  "executor_file": "./executor.wasm"
}
```

mversic: before I comment on this tell me this. Do you think genesis should be given with `--submit-genesis` or should it be submitted like any other transaction with `iroha_client`? We no longer need to submit genesis at the time of bootstrapping the network. Admin can bootstrap the network and push genesis later to 1 or multiple nodes

mversic: let me disambiguate this. We have:
1. what `kagami sign` takes. This can have any JSON schema
2. what `--submit-genesis` takes. This should have the JSON schema of `SignedTransaction`

As far as I'm concerned 1. can be whatever. But 2. must be `SignedTransaction` or at best `TransactionPayload`. Nothing else should be omitted because the node needs all the info to verify the signature. If you omit a field (such as `time_to_live_ms`) then we have to, for each omitted field, either:
1. provide it as another argument together with `--submit-genesis`
2. we have to agree that every genesis will have some predefined value for this field

option 1. I will not even entertain. Option 2. is too implicit for my taste. This tells me that `kagami sign` should best produce full genesis, not just a signature. And this produced genesis should be SCALE encoded blob because it will have inlined executor or some other smart contract

genesis/src/lib.rs
VAmuzing: I thought that we already have lots of "signatures" and decided to distinguish them like that.
And yeah, we had `GenesisNetwork` for a long time so it's kinda similar in that sense, or so I thought.

tools/kagami/src/genesis.rs
VAmuzing: By specifying `time_to_live_ms` do you talk about signed genesis or just any block in general?
One of the issues for us to have a full `SignedTransaction` as a parameter is its size - right now even SCALE encoded version is around 1MB.
Also, before that we didn't have a parameter for `--signed-genesis`, it was a flag. Do you want to change that?

scripts/tests/consistency.sh
VAmuzing: I agree that `iroha_swarm` does not need to know full key-pair, but it does need to know the public key of genesis, isn't it? Otherwise we cannot verify it.

tools/kagami/src/genesis.rs
0x009922: I've lost thread of what we are trying to achieve...

> Do you think genesis should be given with --submit-genesis or should it be submitted like any other transaction with iroha_client?

If there is no technical need to have the genesis block anymore, then why have it? Let's get rid of it.

Issue #4145, opened 2023-12-13
How to update peer's blockstore when the whole topology changed


mversic:
I wounder how we should deal with more general case when whole topology changed.

Step by step: 
1. Network is started with peer set A
2. After some height peer set A is replaced with peer set B (A intersection B = empty set)
3. New peer is added with trusted peers peers from set B 
4. Peer get block and update it's topology to set A (because at this height only peers from topology A were present)

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/issues/4142#issuecomment-1853566834_

mversic:
@dima74 will #5117 take care of this?

dima74:
> @dima74 will #5117 take care of this?

If all peer public keys are also changed, then no. Since we establish connection only with peers which public keys our peer know

mversic:
> > @dima74 will #5117 take care of this?
> 
> If all peer public keys are also changed, then no. Since we establish connection only with peers which public keys our peer know

public key of a peer cannot change. That requires re-registering a peer

dima74:
From the issue description:
> After some height peer set A is replaced with peer set B (A intersection B = empty set)

I mean if there is no such peer in set B so that its public key equals public key of some peer in set A

PR #4384, opened 2024-03-25
[refactor]: remove signatures from blocks

mversic:
## Description

* remove signatures from blocks stored on chain
Signatures are transient and only applicable during consensus. Since blocks in a blockchain are tied together via block hashes, keeping block signatures after consensus is not only superfluous but also creates issues when replaying blocks when topology has changed (related to #4145)
* removed `consensus_estimation` from block header - it was unused
* removed `event_recommendations` from block - it was unused

## Future work
* Following this PR, block sync protocol should be modified to prevent syncing with malicious peers. This can be done with not too much overhead

### Checklist


Erigara (changes requested):
I don't think we can get rid of storing block's signatures.
Because this way we forcing a new peer who is joining the network to trust that blocks are valid without ability to verify that.

> Since blocks in a blockchain are tied together via block hashes

Without signatures malicious peer can provide any blockchain sequence it want.

mversic:
@Erigara the point is that node needs to contact the network, not just one peer when doing block sync because it cannot trust one node. So it asks the network what the hash of the next block is before requesting next block via block sync

Erigara:
This makes block synchronization much more complex.

And also vulnerable to cases where suppose that only one peer left of the whole topology for some reason.

With signatures (and initial set of peers) we still have an ability to check that blocks are valid, without them there is no way.

mversic:
> This makes block synchronization much more complex.

somewhat more complex. This is why block sync should be avoided if possible. This change shouldn't increase network traffic by a noticeable amount (just circling some extra hashes)

> And also vulnerable to cases where suppose that only one peer left of the whole topology for some reason.

even in the current implementation this lone peer could put anything in the signature set and compromise the blockchain. Note that `genesis_private_key` only signs genesis transactions, it doesn't sign genesis block. Genesis block is signed by a peer so we have to trust this peer to put correct signatures in the block.

> With signatures (and initial set of peers) we still have an ability to check that blocks are valid, without them there is no way.

this is not true again, or true only superficially. As in the previous example, new peer has to trust the peer it's connecting to because that peer can manipulate signatures of the genesis block. With the current approach, new peer trusts the one peer from `trusted_peers` it connects to. With new approach at least this trust is dispersed among all peers of the `trusted_peers` set.

mversic:
>  Afaik this could be fixed by signing genesis block and inclusion of initial topology public keys into genesis (maybe just inserting them as ISIs?).
 
this seems like it would do the trick, except that we should make sure genesis consists of exactly 1 transaction so that the peer submitting genesis cannot remove or reorder any

> All in all i'm suggesting to investigate this question more deeply. Maybe removal of signatures is fine and i'm just extra cautious.

ofc. I'd like this to be thoroughly considered. I'm not that convinced myself

mversic:
this was example. Should be further discussed

PR #4398, opened 2024-03-29
[fix] #4267: Introduce p2p idle timeout

### LINKED ISSUE #4267: [BUG] incorrect metric connected_peers with istio service mesh
### OS and Environment

Linux, K8s

### GIT commit hash

eca8569b

1. Raise five peers in Kubernetes
2. Istio must be turned off
3. Metric `connected_peers` = 4 on all peers
3. Turn on istio, restart the pods. Metric `connected_peers` = 4
4. Restart the pods again.
5. Check the metric for peer0, peer1, peer4 `connected_peers` = 4,
6. Check the `connected_peers` metric for peer2 and peer3. It should be equal to `connected_peers` = 3.

Peculiarities
1. There are no errors in the logs related to the connection between peers
2. Go to peer 3 (where connected_peers = 3), make a transaction - it should receive commit status.
3. We go to another peer, for example peer1, where connected_peers = 4 - the transaction should be displayed in this peer

### Actual result

The metric for peer0, peer1, peer4 connected_peers = 4,
The `connected_peers` metric for peer2 and peer3. It should be equal to connected_peers = 3.

### Expected result

The metric `connected_peers` should be equal 4 in all peers

### Logs in JSON format

<details>
  
  ```json
see attachments
  ```
</details>

@BAStos525

### Notes

_No response_

Erigara:
## Description

Previously iroha tried to connect indefinitely to idle peer.

Introduce idle timeout for p2p communication:

Each peer track the last time it received message (or ping) from particular peer, if this time is larger than timeout peer is disconnected.

If no other messages were produced in time equal to half timeout ping is send to prove that peer isn't idle.

### Linked issue

Closes #4267 

### How to test

1. setup iroha network with multiple peers using `LOGLEVEL=TRACE ./scripts/test_env.py setup`
2. kill one of the peers (e.g. iroha1)
3. start listening on p2p port this peer was using `nc -l 1338`
4. check that peers terminate connection with this "peer" after timeout


Erigara:
Waiting for @BAStos525 conformation that issue is resolved to undraft PR.

Erigara:
Got confirmation that issue is resolved.

config/src/parameters/actual.rs
```diff
@@ -72,6 +72,7 @@ pub struct Common {
     pub chain_id: ChainId,
     pub key_pair: KeyPair,
     pub p2p_address: SocketAddr,
+    pub idle_timeout: Duration,
```
DCNick3, line 75: Maybe group the `p2p_address` with `idle_timeout` into a common struct? Could name it something like `P2p`, `NetworkLayer` or just `Network`

p2p/src/peer.rs
```diff
@@ -175,8 +187,22 @@ mod run {
             let mut message_reader = MessageReader::new(read, cryptographer.clone());
             let mut message_sender = MessageSender::new(write, cryptographer);
 
+            let mut idle_interval = tokio::time::interval_at(Instant::now() + idle_timeout, idle_timeout);
+            let mut ping_interval = tokio::time::interval_at(Instant::now() + idle_timeout / 2, idle_timeout / 2);
+
             loop {
                 tokio::select! {
+                    _ = ping_interval.tick() => {
+                        iroha_logger::trace!("Sending ping");
```
DCNick3, line 196: (probably needs a better way for displaying the duration)

```suggestion
                        iroha_logger::trace!("The connection has been idle for {:?}, pinging to check if it's alive", ping_interval.period());
```

p2p/src/peer.rs
```diff
@@ -175,8 +187,22 @@ mod run {
...
+                        if let Err(error) = message_sender.send_message(Message::<T>::Ping).await {
+                            iroha_logger::error!(%error, "Failed to send ping to peer.");
+                            break;
+                        }
+                    }
+                    _ = idle_interval.tick() => {
+                        iroha_logger::error!(timeout=idle_timeout.as_secs(), "Connection with peer is idle");
```
DCNick3, line 203:
```suggestion
                        iroha_logger::error!(timeout=idle_timeout.as_secs(), "Didn't receive anything from the peer for {:?}, abandoning this connection", idle_interval.period());
```

p2p/src/peer.rs
```diff
@@ -187,10 +213,12 @@ mod run {
                         if post_receiver_len > 100 {
                             iroha_logger::warn!(size=post_receiver_len, "Peer post messages are pilling up");
                         }
-                        if let Err(error) = message_sender.send_message(msg).await {
+                        if let Err(error) = message_sender.send_message(Message::Message(msg)).await {
                             iroha_logger::error!(%error, "Failed to send message to peer.");
                             break;
                         }
+                        // Reset ping timeout as peer send another message
+                        ping_interval.reset();
```
DCNick3, line 221: Does it _always_ make sense to reset the ping interval when sending a message?

AFAIU, the logic is that if we are gonna send some message, the peer's gonna respond and reset our `idle_timer`. But this assumption doesn't hold if there are messages that don't require answering..

IMO it makes more sense to reset the interval at at the same time as we reset the `idle_interval`: when a message is received

Erigara: Logic is that we don’t care if other peer going to respond to our ping, so it’s just a way to send message once in a while to remind about itself.

So these two processes are independent (reminding about ourselves and checking if other side is alive).

That’s why i reset ping interval when any message is sent.

Now i noticed the problem:
Two peers can have different idle timeouts (like 60s and 10s). 
First peer going to send pings once every 30s which is not enough to keep alive connection with the other peer who expects message every 10s.

Not sure what best course of action would be here:
1. hardcoding some value seems limiting
2. we can send desired keep alive value after the handshake to the other side (like “please send me pings once every n seconds”)
3. make it ping-pong like so that other side should reply to our ping

DCNick3: Out of 2 and 3 I prefer the 3, as it would just easier to implement

DCNick3 (dismissed):

DCNick3:
Need to update config snapshots it seems

config/src/parameters/actual.rs
```diff
@@ -71,16 +71,24 @@ impl Root {
...
     }
 }
 
+/// Common options shared between multiple places
+#[allow(missing_docs)]
+#[derive(Debug, Clone)]
+pub struct P2p {
+    pub address: SocketAddr,
+    pub idle_timeout: Duration,
+}
```
0x009922, line 90:
```suggestion
/// Network options
#[allow(missing_docs)]
#[derive(Debug, Clone)]
pub struct Network {
    pub address: SocketAddr,
    pub idle_timeout: Duration,
}
```

config/src/parameters/actual.rs
```diff
@@ -71,16 +71,24 @@ impl Root {
 pub struct Common {
     pub chain_id: ChainId,
     pub key_pair: KeyPair,
-    pub p2p_address: SocketAddr,
+    pub p2p: P2p,
```
0x009922, line 74: It feels confusing, because while the network address is a _common_ option, the idle timeout is not. Could you re-arrange them?

0x009922 (commented):
Please, while changing configuration, also contribute into the reference:

https://github.com/hyperledger/iroha-2-docs/pull/397

config/src/parameters/actual.rs
Erigara: How do you think i should rearrange them?

config/src/parameters/actual.rs
Erigara: Initially i used Network name, problem is that it clashes with other Network structs defined through codebase. 
What is our policy here? rename to Config?

0x009922: I would say, the approach I used is to reduce discrepancy between config modules and actual Iroha core modules. That is why `iroha_config::parameters::actual` even exists - its purpose is to map _user_ configuration in a way that lies down onto the internal structure.

I would appreciate if you take it one step further in this direction. Maybe it would mean to make a minor refactoring on the core side as well. Maybe it would mean to get rid of `actual::Sumeragi` and define just `actual::Network`. I am confused myself.

config/src/parameters/actual.rs
0x009922: Replied in other thread generally.

DCNick3: > How do you think i should rearrange them?

You could put `P2p` into the `Root` struct instead of `Common`

config/src/parameters/actual.rs
Erigara: i don't think we should merge p2p configuration and sumeragi configuration as single entity, they even defined in separate crates

Erigara: what we can do is to duplicate some fields instead of having `Common` struct

Erigara: renamed `P2p` to `Network` and moved it to the `Root` also `Common` now has `PeerId` directly

Erigara: @0x009922 your opinion?

0x009922 (dismissed):

config/src/parameters/actual.rs
0x009922: I think it is a fine compromise, thanks.

0x009922 (approved):

DCNick3 (approved):

Issue #4409, opened 2024-04-04
[suggestion] Remove "genesis account" notion altogether


s8sato:
After genesis, should peers continue to worry about whether the genesis account will appear?
Since the genesis transaction is derived from system rather than user, it might be possible to remove the concept of account authentication from the genesis.
In other words, it does not have to be a genesis transaction to give an initial state, it only has to be a genesis executable.
What if every peer executes it at the initialization time to get a "genesis state" instead of a genesis block, with a "state hash" to agree on that state?

s8sato:
First thought:
After genesis, should peers continue to worry about whether the genesis account will appear?
Since the genesis transaction is derived from system rather than user, it might be possible to remove the concept of account authentication from the genesis.
In other words, it does not have to be a genesis transaction to give an initial state, it only has to be a genesis executable.
What if every peer executes it at the initialization time to get a "genesis state" instead of a genesis block, with a "state hash" to agree on that state?

Why closed：
Such a "genesis state" would not be transparent to clients i.e. cannot be queried nor reproduced

Issue #4410, opened 2024-04-04
[suggestion] Dedup public keys in `transaction.signature` and `transaction.payload.authority`


s8sato:
As of f12077d35 there is a duplication between `transaction.signature.public_key` and `transaction.payload.authority.signatory`.
The former could also serve as the latter role, but the problem is that `authority: AccountId` is required for subsequent process and it consists of not only public key but also `DomainId` which cannot be known from `transaction.signature`.
The solution would be to drive `DomainId` out of `AccountId`, but the main blocker is that `AccountId` should inform of its parent domain for event scoping.
So the first step would be to retrieve `DomainId` somewhere other than `AccountId` at the account event emission

Issue #4373, opened 2024-03-18
Reimplement multi-signature


s8sato:
### Feature request

> 3. Open a separate issue to discuss implementation of multisig accounts

### Motivation

https://github.com/hyperledger/iroha/issues/2085#issuecomment-1991254965

s8sato:
My basic idea keeps coming back to the multi-signature section of https://github.com/hyperledger/iroha/issues/2085#issuecomment-1980557101

Issue #4372, opened 2024-03-18
Account alias resolution


s8sato:
### Feature request

> 2. Open a separate issue to discuss whether we want to implement alias as part of Iroha or delegate this to external service (I would prefer to keep alias resolution as a separate service outside Iroha to not complicate the codebase)

### Motivation

https://github.com/hyperledger/iroha/issues/2085#issuecomment-1991254965

mversic:
closed in favor of #2086

s8sato:
As long as `transaction.payload.authority` is `AccountId`, aliases should be resolved before the transaction building.
Otherwise the transaction signature will be affected by aliases, which means verifiers should consider alias history of the transaction authority

mversic:
you make a good point for the case of having alias resolution as external service

s8sato:
My last post didn't say anything so let me redefine the motive to external name services:
Since the transaction authority is the client itself, it is already in the config in a resolved format.
Rather, the problem with having an internal name service (or [an internal entrypoint for external name services](https://docs.ens.domains/learn/ccip-read)) is that API has to accept aliases as a variant of other account IDs in the request (e.g. transfer destination).
If we want to treat aliases as optional, it should not affect the current API.

The problem then is that such an external name service needs to have the same security as Iroha:
Suppose the name service has been hacked. Your destination alias will resolve to the scammer's address. This would mean that the name service spoiled the security provided by Iroha.
So the name service would be a decentralized system. In blockchains, it seems to be common to use blockchain for name service.

My suggestion is to deploy another chain dedicated to aliases, and having clients communicate with it to resolve aliases as needed.
I think we can learn from [ENSv2](https://blog.ens.domains/post/ensv2), which is moving core functionalities from the mainnet to L2

s8sato:
> My last post didn't say anything

No, it said something:
if clients built a transaction with unresolved names, the signature should be verified with that unresolved names, which means the name resolution would penetrate deeper than the authentication, which would be a cost on every verifier

s8sato:
### Draft happy path

Prerequisites:

- A resolver is a Torii endpoint of a node of another Iroha network deployed as a name service

Resolution on sending requests:

1. Client accepts `AccountAddr` as user input, which is either `AccountAlias` (e.g. `alice@wonderland`) or `AccountId`
2. Client detects `AccountAlias` before building a transaction/query and query the resolver to resolve them
[...]

PR #4411, opened 2024-04-04
feat!: authenticate personal accounts by ID

### LINKED ISSUE #2085: Add primary key to account
Refactor the accounts to only *optionally* have an alias. 

We still need to retain the current semantics of account structure, but most data should be stored in a separate location: hash-table keyed by the account's ss58 (or pub_key) address, which contains actual data, an account alias is only an interface to all of that.

### LINKED ISSUE #3511: [BUG] kogami - after creating new genesis.json file we have 2 same public keys per user by default
### OS and Environment

Docker hub

### GIT commit hash

c4af68c4

simple install via docker-compose file

```
docker-compose -f docker-compose.yml up -d
```

### Actual result

On file /iroha/configs/peer/genesis.json we have a same public key on signatories array
```
{
          "Register": {
            "NewAccount": {
              "id": "alice@wonderland",
              "signatories": [
                "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
              ],
              "metadata": {
                "key": {
                  "String": "value"
                }
              }
            }
          }
        },
        {
          "Register": {
            "NewAccount": {
              "id": "bob@wonderland",
              "signatories": [
                "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
              ],
              "metadata": {
                "key": {
                  "String": "value"
                }
              }
            }
          }
        }
```

### Expected result

On file /iroha/configs/peer/genesis.json we have a different public key on signatories array
```
{
          "Register": {
            "NewAccount": {
              "id": "alice@wonderland",
              "signatories": [
                "ed01204595da8957c5598d4de20fe5f3eb4c14820678e1a1957a535db1fd4b3d1607c5"
              ],
              "metadata": {
                "key": {
                  "String": "value"
                }
              }
            }
          }
        },
        {
          "Register": {
            "NewAccount": {
              "id": "bob@wonderland",
              "signatories": [
                "ed01207233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0"
              ],
              "metadata": {
                "key": {
                  "String": "value"
                }
              }
            }
          }
        }
```

### Logs in JSON format

<details>
  
  ```json
  ```
</details>

@timofeevmd

### Notes

_No response_

s8sato:
#### [feature] #2085: Authenticate personal accounts by ID

This is a milestone for the following commit. The inclusion of the public key in the ID allows the ID to match the verifying key of the signature. This makes it possible to authenticate transactions and queries with the ID alone.

- `api-changes`: single signatory account, single signature transaction
- `api-changes`: remove query `FindAccountsByName`
- `config-changes`: client config `account.id` changes to `account.domain_id`
- add `test_samples` utility for tests
- different notions of signatories "peer", "genesis", and "alice" in the test network

#### ~~[feature] #2085: Auto-register destination account on transfer~~

~~Includes registration of the destination account in the transfer execution. The main use case is described in test `auto_register_account::on_transfer::asset_numeric`~~

- EDIT: left to #4426

### Linked issue

- This is the 1st of the 3-part work of account restructuring:
  - Closes #2085
  	- Opens #4426
  - #4372
  - #4373
- Opens #4410
- Contains suggestions for #4409
- Closes #3511

### Benefits

As long as you know the address (account ID), you can send without worrying whether the destination account is already registered on the chain.
However, the current design requires the sender to be authorized to register a new account within the domain. See https://github.com/hyperledger/iroha/pull/4411#discussion_r1552269607

### Drawbacks

- Conventional `name@domain` notation will be temporary lost
  - To be restored with #4372
- Multi-signature functionality will be temporary lost, coupled with [the queue change](https://github.com/hyperledger/iroha/pull/4308)
  - To be restored and enhanced with #4373

### Checklist

- [x] I've written unit tests for the code changes


data_model/src/query/predicate.rs
```diff
@@ -688,9 +689,10 @@ pub mod string {
                 // TODO: Once #2302 and #1889 are merged, add more cases.
             }
 
+            #[cfg(compile_err)] // FIXME the trait `Alias<AccountId>` is not implemented for `str`
             #[test]
             fn account_id() {
-                let id = IdBox::AccountId("alice@wonderland".parse().expect("Valid"));
+                let id = IdBox::AccountId("alice@wonderland".parse_alias());
```
s8sato, line 695: As an unresolved issue, 5 tests are failing to build and compiler says

> the trait `Alias<AccountId>` is not implemented for `str`

even though the trait `Alias<AccountId>` is implemented for `str`

configs/sample_params/src/alias.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Resolve [`self`] to string that should be parsed to `T: FromStr`.
+    fn resolve(&self) -> impl AsRef<str>;
+}
+
+impl Alias<AccountId> for str {
+    fn resolve(&self) -> impl AsRef<str> {
+        let (name, domain) = self
+            .rsplit_once('@')
+            .expect("name@domain format should be given");
+        let sp = super::SampleParams::default();
```
Erigara, line 37: For the separate issue we can wrap `SampleParams` into smt like [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock) to initialize only once.

configs/sample_params/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
```
Erigara, line 1: Do you think this crate should be here? 
Before `configs` wasn't used to store source code. Maybe use `config` folder?

client/tests/integration/auto_register_account/on_transfer.rs
```diff
@@ -0,0 +1,237 @@
...
+///
+/// # Scenario
+///
+/// 0. alice@wonderland: domain owner
+/// 0. alice -> new carol ... ok (domain owner)
+/// 0. carol -> new dave ... err (permission denied)
+/// 0. grant `CanRegisterAccountInDomain` permission to carol
+/// 0. carol -> new dave ... ok (authorized)
+#[test]
+fn asset_numeric() {
```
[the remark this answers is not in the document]
Erigara: This transfer permissions concerns me a bit, like regular user i probably don't care if person i' trying to transfer assets is registered or not.

Can we do the following?
Transfer to any address is always allowed (what about mint and burn?), but to actually make use of this assets (e.g. transfer them to someone else) you have to register an account.
Thus by registering account you claim ownership of this assets.
Since accounts are identified by public key only owner of the private key can register the account.

data_model/src/query/predicate.rs
Erigara: That's strange everything compiles just fine on my machine.

configs/sample_params/Cargo.toml
Erigara: Or just root folder.

0x009922: <kbd>/configs</kbd> folder is not for crates for sure. I'd propose using <kbd>/sample_params</kbd> for now, and later unify with samples removed from `iroha` crate (https://github.com/hyperledger/iroha/issues/4136)

core/src/smartcontracts/isi/mod.rs
```diff
@@ -270,15 +303,14 @@ mod tests {
         let world = World::with([], PeersIds::new());
         let query_handle = LiveQueryStore::test().start();
         let state = State::new(world, kura.clone(), query_handle);
-        let genesis_account_id = AccountId::from_str("genesis@genesis")?;
-        let account_id = AccountId::from_str("alice@wonderland")?;
-        let (public_key, _) = KeyPair::random().into_parts();
+        let genesis_account_id: AccountId = "genesis@genesis".parse_alias();
```
0x009922, line 306: Maybe a form of `AccountId::from_sample_alias("genesis@genesis")` would also work, and will reduce the diff. (current is also not bad).

client/tests/integration/auto_register_account/on_transfer.rs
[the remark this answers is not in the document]
s8sato: You have a point, I will consider it. If this discussion is controversial or things get too complicated, we may split the 2nd commit into a separate PR

data_model/src/query/predicate.rs
s8sato: Implementing for `&str` did not work. I guess it has something to do with `http` and `transparent_api` features

configs/sample_params/Cargo.toml
s8sato: Ok, it makes sense to rename the crate to more specific one since it is currently only used for aliases

configs/swarm/client.toml
```diff
@@ -5,7 +5,7 @@ torii_url = "http://127.0.0.1:8080/"
 web_login = "mad_hatter"
 password = "ilovetea"
 
-[account]
-id = "alice@wonderland"
-public_key = "ed01207233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0"
-private_key = { algorithm = "ed25519", payload = "9ac47abf59b356e0bd7dcbbbb4dec080e302156a48ca907e47cb6aea1d32719e7233bfc89dcbd68c19fde6ce6158225298ec1131b6a130d1aeb454c1ab5183c0" }
+[account] # alice
+domain_id = "wonderland"
```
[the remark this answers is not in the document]
s8sato: I'm not sure I understand your concern.
Do you have any specific change suggestions for this file?
Is it something other than removing the comment that mentions a specific alias?

configs/sample_params/src/alias.rs
```diff
@@ -0,0 +1,93 @@
...
+    ///
+    /// # Panics
+    ///
+    /// - [`resolve`](Alias::resolve) implementation is not compatible with [`FromStr`] for `T`
+    ///
+    /// # Example
+    ///
+    /// See [`crate`] documentation.
+    #[inline]
+    fn parse_alias(&self) -> T {
```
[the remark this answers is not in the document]
s8sato: Here I'd choose a short method name.
I think "parse_" implies this is an utility for tests because aliases would not be able to simply be parsed in a more authentic environment

client/tests/integration/auto_register_account/on_transfer.rs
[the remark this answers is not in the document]
s8sato: I think the point is the domain membership screening.
How should a new account request and be allowed to join a domain?
In terms of authority, especially in a "private domain", there is a large gap between transferring one's own asset and registering accounts within the domain.
So the admission process should be managed by a different authority than the transaction sender, as you pointed out.

One idea is that each domain has its own admission policy as a trigger that, in response to an account creation event, "activates" the new account (as a personal or a shared account).
Admission policies may vary from domain to domain e.g.

- No screening
- Wait for approval of the domain owner
- Wait for approvals of m of n domain members

In any case, as I implied, I'm going to create a new issue on this and remove the 2nd commit from this PR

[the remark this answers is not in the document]
s8sato: - #4426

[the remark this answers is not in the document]
s8sato: We seem to be in agreement that there are two types of account registration, so to speak, provisional and full registration.

Is it your opinion that the 2nd commit of this PR is acceptable, but its single registration makes an account equivalent to fully registered one, so subsequent PRs should weaken the current single registration to provisional one and newly implement the full registration?

If so, does provisional registration require permission?
If it cannot be executed without the same permission as full registration, then there would be no motive for them to split into two.
Or do we need dedicated permission for provisional registration?

configs/swarm/client.toml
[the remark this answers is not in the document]
0x009922: My concern is that: if I open <kbd>/configs/swarm</kbd> directory and run `docker compose up`, will it just work with these configs?

configs/sample_params/src/alias.rs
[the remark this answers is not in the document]
0x009922: Well, maybe at least make an emphasis in the documentation that this crate and this function is not intended to be used anywhere apart from tests?

data_model/src/query/predicate.rs
mversic: if we disallow strings that parse as multihash to be used as aliases then instead of having separate `parse_alias` method we can put all the logic into `FromStr::from_str`.

data_model/src/transaction.rs
```diff
@@ -9,10 +9,10 @@ use core::{
 };
 
 use derive_more::{DebugCustom, Display};
-use iroha_crypto::SignaturesOf;
+use iroha_crypto::SignatureOf;
 use iroha_data_model_derive::model;
 use iroha_macro::FromVariant;
-#[cfg(feature = "std")]
+#[cfg(all(feature = "std", feature = "transparent_api"))]
 use iroha_primitives::time::TimeSource;
```
mversic, line 16: please inline this import at the call site

data_model/src/block.rs
```diff
@@ -166,6 +166,7 @@ impl SignedBlock {
 
     /// Signatures of peers which approved this block.
     #[inline]
+    #[allow(private_interfaces)]
     pub fn signatures(&self) -> &SignaturesOf<BlockPayload> {
```
mversic, line 170: this is a bit suspicious looking. But I still don't think `BlockPayload` should be public

data_model/src/trigger.rs
```diff
@@ -235,6 +235,7 @@ pub mod action {
     }
 
     impl PartialOrd for Action {
+        #[allow(clippy::non_canonical_partial_ord_impl)]
```
mversic, line 238: it can easily be made canonical. Just swap bodies of `PartialOrd` and `Ord`

genesis/src/lib.rs
```diff
@@ -23,8 +23,16 @@ use serde::{Deserialize, Serialize};
 pub static GENESIS_DOMAIN_ID: Lazy<DomainId> = Lazy::new(|| "genesis".parse().expect("Valid"));
 
 /// [`AccountId`] of the genesis account.
-pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> =
-    Lazy::new(|| AccountId::new(GENESIS_DOMAIN_ID.clone(), "genesis".parse().expect("Valid")));
+/// TODO remove "genesis account" #4409
+pub static GENESIS_ACCOUNT_ID: Lazy<AccountId> = Lazy::new(|| {
+    AccountId::new(
+        GENESIS_DOMAIN_ID.clone(),
+        // FIXME An associated private key can exist and sign transactions even after genesis. Consider removing "genesis account" notion altogether
```
mversic, line 30: it doesn't matter if associated private key exists because if a transaction signed by genesis is encountered after the genesis round it is rejected. So no, this cannot happen

mversic: and I don't think that we can remove the notion of `genesis` account but maybe you have some idea

mversic (changes requested):

core/src/smartcontracts/isi/mod.rs
```diff
@@ -33,6 +36,51 @@ pub trait Registrable {
     fn build(self, authority: &AccountId) -> Self::Target;
 }
 
+/// Validate and execute an additional instruction `T` during `Self` execution.
+/// FIXME This means going backwards from execution phase to validation phase, which is contrary to current design principles
+trait BackValidate<T>
```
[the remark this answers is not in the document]
mversic: IMO it is not the responsibility of `iroha_core` to create destination account if it doesn't exist, it is executor's responsibility. I would move all this logic into default executor and `iroha_core` should plainly reject all instructions that operate on non-existing accounts. In other words, default executor should check whether destination account exists and if not first execute `Register<Account>` which will go through the regular validation process (since validation is done in executor) 

If you do this you wouldn't have the problem you are describing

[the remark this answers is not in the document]
Erigara: @mversic my initial concern was exactly about demanding registration permissions when i.e. making transfer, imo we shouldn't do that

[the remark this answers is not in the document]
mversic: I see. IMO, in any case `iroha_core` should reject this transfer if destination doesn't exist. It is up to the executor to create missing destination before doing the transfer. We can make the default executor not require any permissions when creating a provisional account, but it has to create it before executing the transfer

data_model/src/account.rs
```diff
@@ -66,155 +60,77 @@ mod model {
...
     #[ffi_type]
     pub struct AccountId {
-        /// [`Account`]'s [`Domain`](`crate::domain::Domain`) id.
+        /// [`Domain`](crate::domain::Domain) that the [`Account`] belongs to.
         pub domain_id: DomainId,
-        /// [`Account`]'s name.
-        pub name: Name,
+        /// Sole signatory of the [`Account`].
+        pub signatory: PublicKey,
     }
```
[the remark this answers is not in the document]
s8sato: Yes, I see similar discussions about other entities in #3921 and #4243, so it may be only a matter of time before we are working on it

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
s8sato: I'll try to implement that in the following PR, which will alternate the 2nd commit of this PR.
There, executor would try to provisionally register the account anyway without checking if the account is already registered.
This is because the executor does not refer to the state -- if it needs to determine if the account is existing, the flag should be included in AccountId, which leads to a perversion that the transaction builder should concern about whether the account is registered or not

data_model/src/query/predicate.rs
s8sato: In that case,

- FromStr defined in the data model depends on the sample aliases
- deserializer accepts both multihash and alias, but serializer must decide between them

Do we accept them?

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
Erigara: > I see. IMO, in any case `iroha_core` should reject this transfer if destination doesn't exist. It is up to the executor to create missing destination before doing the transfer. We can make the default executor not require any permissions when creating a provisional account, but it has to create it before executing the transfer
> 
> I guess accounts will then have a flag to indicate whether they are provisional or not

If we are going to settle down with the approach that no additional permissions are required to make transfer, i think it's okay to create necessary for this transfer structures inside core itself.

data_model/src/query/predicate.rs
mversic: > FromStr defined in the data model depends on the sample aliases

I don't understand why

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
mversic: > If we are going to settle down with the approach that no additional permissions are required to make transfer, i think it's okay to create necessary for this transfer structures inside core itself.

I don't think so. We can go about this 2 ways:
1. Iroha is maximally permissible (it will create necessary conditions for instruction to succeed)
2. Iroha is maximally restrictive (if conditions are not met it will reject the instructions)

IMO execution in Iroha should have no side effects and we should go with 2nd approach. All side effects should be contained in the executor

> I'll try to implement that in the following PR, which will alternate the 2nd commit of this PR.

all I'm saying is that you should put this logic into executor. You can deal with "provisional" accounts in a separate PR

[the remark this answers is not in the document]
Erigara: Let's consider that for some extreme case deployment where user using really minimal executor, without additional logic of creating provisional account user can't access feature of offline accounts.
Is it something we want?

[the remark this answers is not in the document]
mversic: > This is a kind of compromise because it goes against the current design principle that all validations should be done prior to any execution.

@s8sato this is not always the case, on `Unregister<Role>` we execute unvalidated `Revoke<Role, Account>` for all accounts that have it granted. As was noted in #4330

> Let's consider that for some extreme case deployment where user using really minimal executor, without additional logic of creating provisional account user can't access feature of offline accounts.
Is it something we want?

ok, I have a different suggestion. Let's keep the current logic in `iroha_core` but have our `default_executor` forbid this by default. I say this since I don't find this is something that permissioned blockchains would want to have enabled.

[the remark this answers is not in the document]
Erigara: ok, it could be the starting point (allowing it later would be backward compatible)

configs/swarm/client.toml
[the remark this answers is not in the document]
s8sato: I ran pytest with docker compose environment and saw the same result as one with `test_env.py` environment, so I think there are no issue specific to swarm.
One of the rest tasks in this PR is providing the same sample aliases as the cargo tests to pytest client, and have client and peer communicate in resolved format.
Does this answer your question?

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
s8sato: @Erigara could you more specific why the `Map<AccountId, Asset>` would achieve an equivalent of provisional accounts?
I understand all we have to do to apply e.g. transfer instruction would be modifying the map only.

I'm not sure it is relevant to the following, but
it reminded me of what I called `@nodomain` approach, which represents being provisional by being under "nodomain".
Why I rejected this approach is because it seemed obviously unreasonable to replace e.g. `@wonderland` with `@nodomain` and remember the original `@wonderland` and later modify `@nodomain` to `@wonderland`.
But an evil here might be non-separation of the account identification and the domain membership.
If they were separated, being under "nodomain" (and being provisional) would have been equal to `account.domain_id: None`.
This approach might rather work well with the shallow objects

data_model/src/query/predicate.rs
s8sato: I assume your suggestion would be something like:

```rust
// data_model/src/account.rs
impl FromStr for AccountId {
        if is_multihash(s) {
            todo!()
        } else {
            use iroha_sample_params::alias::Alias;
            let account_id = s.parse_alias(); // or something equivalent inlined
            Ok(account_id)
        }
    }
}

```

If so, it would be referencing to something like the current `configs/sample_params/src/.toml`, which is only for testing

mversic: you don't have to put your code into `Alias` trait just paste it here

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
s8sato: In that case we still need something indicates if the source account is active or not

data_model/src/query/predicate.rs
s8sato: I mean, going into the `else` block would not be expected except in test cases.
I assume client should have its own alias resolution before building transaction and should communicate with peer in resolved format, where `FromStr` implementation plays an role of the interface.
So for non-multihash (unresolved) input `from_str` should return `Err` immediately

data_model/src/account.rs
```diff
@@ -66,155 +58,77 @@ mod model {
...
-
-    fn check_signatories(signatories: &Signatories) -> Result<(), &'static str> {
-        if signatories.is_empty() {
-            return Err("Signatories cannot be empty");
-        }
-        Ok(())
+impl AccountId {
+    /// Return `true` if the account signatory matches the given `public_key`.
+    #[inline]
+    pub fn signatory_matches(&self, public_key: &PublicKey) -> bool {
```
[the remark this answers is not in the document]
s8sato: Let me keep this function to inform where transactions and queries are checked about consistency of signatory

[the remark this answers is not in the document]
mversic: if you want to keep it, then make it `#[feature = "transparent_api"]`

data_model/src/query/predicate.rs
mversic: >  I assume client should have its own alias resolution before building transaction and should communicate with peer in resolved format, where `FromStr` implementation plays an role of the interface.

ok, I understand you now but I don't like it. I don't want to see 2 different methods for parsing account ids. User shouldn't worry which format is his account id in. And then if they change the format their whole code would break.
I'd rather we put this into `data_model` and introduce some feature flag to control this behavior. We can actually reuse `transparent_api` here. For clients `account_id.parse()` will work for both aliases and keys, but in `iroha_core` it will return error for aliases. How about this?

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
Erigara: Yeah, in that case creating Account and activating it happens at the same time

s8sato:
Updated, see [this comparison](https://github.com/hyperledger/iroha/compare/af55ee4c706e3491f5966b2e4799b62437eb3bc8..fd5a898e8e357c149926a0b26f039786ee817f26). The changes are as follows:

- remove the 2nd commit: Auto-register destination account on transfer
- fix pytest account structure
- move `iroha_sample_params` directory to under the root
- initialize and load `SampleParams` in static memory at the first call
- document to alert `parse_alias()` is for testing only
- provide `(Account/Asset)Id::from_alias()` syntax
- remove imports under features

Then, now rebasing

core/src/smartcontracts/isi/mod.rs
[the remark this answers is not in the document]
Erigara: If there is no need for provisional registration we can check that source account is registered and it should be enough, shouldn't it?

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -43,7 +43,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
       - name: Run tests, with coverage
-        run: mold --run cargo test --all-features --workspace --no-fail-fast
+        run: |
+          mold --run cargo clean
```
mversic, line 47: why are you cleaning?

mversic: if this is about #4488 . Let's please do it in separate PR

s8sato:
Some discussions are now outside the scope of this PR, so they will be folded once and tracked in #4426

.github/workflows/iroha2-dev-pr.yml
s8sato: I know, I'm just taking advantage of the opportunity to run CI

genesis/src/lib.rs
s8sato: The original motivation was that it would be better if that decision process was not necessary, but I reconsidered and closed #4409

data_model/src/query/predicate.rs
s8sato: I still don't understand your point. What's the problem with the following?
- user (end user with client) is allowed to enter either multihash or alias
  - the client should have a different resolver (name service) other than `parse_alias`
    - probably try to `parse` as multihash first and then ask the name service to resolve alias
- user (developer) will be writing tests with `parse_alias` in most cases, as long as our integration tests rely on it
- if any fragility with `parse_alias`, the same with that consolidated `parse`

s8sato: Isn't your suggestion to switch resolver depending on whether it is testing environment or not, rather than called from clients or not?

mversic: how about if we used only public key format of account ids in `data_model`.  You can keep `parse_alias` in `iroha_client`

s8sato: I think `AccountId` in data model is already restricted to the public key (multihash) format as shown [here](https://github.com/hyperledger/iroha/pull/4411#discussion_r1552265003)

mversic: yes, but in `data_model` tests you use `parse_alias`

s8sato: Yeah, currently `iroha_sample_params` appears in
- `[dev-dependencies]` of core, data_model, data_model_derive, genesis
- `[dependencies]` of client, test_network, kagami, parity_scale_cli
Is it your goal to remove dev dependency on `iroha_sample_params` from data_model?

Also, I noticed even if we keep this crate it should be renamed to drop `iroha_` prefix as it is not planed to publish

mversic: > Is it your goal to remove dev dependency on iroha_sample_params from data_model?

no, but it's a nice consequence if it happens. My goal is to not introduce concept of aliases into `data_model`

s8sato: Ok, then I'll just replace alias with multihash in data_model. What do you think about other crate?

mversic: I'd also replace aliases with public keys in other crates

0x009922: I also don't particularly like the idea of having a global `.parse_alias` extension, in `data_model` or not. However, having a pool of aliases for testing purposes seems useful.

A middle ground, IMO, would be having a utility struct like `test_aliases::TestAliases` which will provide a simple and expressive API. For example:

```rust
// voila: 
let alice: AccountId = TestAliases::get("alice").in_domain("wonderland");
```

Or even simpler, just returning a `PublicKey` for a specific account name.

mversic: > However, having a pool of aliases for testing purposes seems useful.

I disagree. Since the value doesn't matter, just create a new data where you need it. Sharing test data across crates looks crazy to me

s8sato: On second thought, the only purpose is to easily get an `AccountId` for testing.
The word "alias" might not be good either.
No need to fix keys except for a few predefined signatories: peer, genesis, alice in test network and bob, carpenter in default genesis

Erigara: Imo having aliases in integration tests is convenient feature.

configs/sample_params/Cargo.toml
0x009922: Maybe not use `iroha_` prefix for it?

We have `test_network` and not `iroha_test_network`. And eventually `iroha::samples` might become `test_samples`. You can make `test_samples` already now and place your code there.

data_model/src/query/predicate.rs
0x009922: > I disagree. Since the value doesn't matter, just create a new data where you need it. Sharing test data across crates looks like an antipattern to me

That point works for me. Indeed, why don't generate random `AccountId`s in tests each time?

configs/sample_params/Cargo.toml
mversic: we will remove `test_network`

data_model/src/query/predicate.rs
s8sato: Considering that account names were originally hard-coded in tests, even fixed keys would serve as a refactoring.
Btw pytests do not require any predefined signatories so far

configs/swarm/client.toml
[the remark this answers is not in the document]
0x009922: No, it doesn't =(

What I am hearing from you:

- Rust tests use sample aliases everywhere with `.parse_alias` helper. Aliases resolution doesn't have anything to do with Iroha internals. It is essentially a shorthand to create `AccountId` in a readable way on client side and then pass it to Iroha.
- ~~You set up the same aliases resolution mechanism in Pytests~~ Upon observing pytests changes, I see that you replaced account names with public keys. Thus, you are not using aliases there.
- You set up test env with `test_env.py` and run pytests, and it works.
- You also run `docker compose up` manually (I guess, with one of the swarm compose files) and it works.

Well, now am not confused... Maybe the diff of the file initially was different, but now it seems fine: client config has `account` with domain id and key pair, and the account id is a combination of public key and domain id.

The only thing remains confusing to me: "alice". Within the scope of swarm configurations, it doens't make sense. Alice exists only within test aliases. **<kbd>/configs/swarm</kbd> is a thing on its own**, not related to tests in any way. The fact that test environment relies on swarm configs is only because it is convenient so far.

I can elaborate more on that, but the request is tiny: please remote mention of Alice here.

data_model/src/query/predicate.rs
mversic: > Here is genesis though for example, so some way (the current toml file would be a candidate) to share predefined signatories across languages would be desired when such tests become not ignored

wdym by share across languages? can you be more specific which crates need to share predefined signatories

Erigara: Ok, random keys might work just as good, i won't insist on aliases.

s8sato: Ok, I think this discussion is settling down:
I'll withdraw the whole `iroha_sample_params`. Instead, since `test_network` will be deprecated and unit tests are still generating account IDs, I'm going to introduce an independent utility crate `test_samples`, which provides some generating function like:

```rust
use test_samples::gen_account_in;

let (mouse_id, mouse_keypair) = gen_account_in("wonderland");
```
As for alice, I might have trouble dealing with it

test_samples/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+
+        static_signatory_ed25519!($kp, $vk, $sk);
+    };
+}
+static_signatory_ed25519!(PEER_KEYPAIR, "ed01207233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0", "9AC47ABF59B356E0BD7DCBBBB4DEC080E302156A48CA907E47CB6AEA1D32719E7233BFC89DCBD68C19FDE6CE6158225298EC1131B6A130D1AEB454C1AB5183C0");
+static_signatory_ed25519!(ALICE_ID, "wonderland", ALICE_KEYPAIR, "ed0120CE7FA46C9DCE7EA4B125E2E36BDB63EA33073E7590AC92816AE1E861B7048B03", "CCF31D85E3B32A4BEA59987CE0C78E3B8E2DB93881468AB2435FE45D5C9DCD53CE7FA46C9DCE7EA4B125E2E36BDB63EA33073E7590AC92816AE1E861B7048B03");
+static_signatory_ed25519!(BOB_ID, "wonderland", BOB_KEYPAIR, "ed012004FF5B81046DDCCF19E2E451C45DFB6F53759D4EB30FA2EFA807284D1CC33016", "AF3F96DEEF44348FEB516C057558972CEC4C75C4DB9C5B3AAC843668854BF82804FF5B81046DDCCF19E2E451C45DFB6F53759D4EB30FA2EFA807284D1CC33016");
+static_signatory_ed25519!(CARPENTER_ID, "garden_of_live_flowers", CARPENTER_KEYPAIR, "ed0120E9F632D3034BAB6BB26D92AC8FD93EF878D9C5E69E01B61B4C47101884EE2F99", "B5DD003D106B273F3628A29E6087C31CE12C9F32223BE26DD1ADB85CEBB48E1DE9F632D3034BAB6BB26D92AC8FD93EF878D9C5E69E01B61B4C47101884EE2F99");
```
s8sato, line 43: - PEER is separated from ALICE (and GENESIS_ACCOUNT_ID), as they would be different in general
- ALICE, BOB, CARPENTER need to be agreed between default genesis and integration tests

client/tests/integration/upgrade.rs
```diff
@@ -3,35 +3,45 @@ use std::{path::Path, str::FromStr as _};
...
     data_model::prelude::*,
 };
+use iroha_crypto::{Algorithm, KeyPair, PrivateKey};
 use iroha_logger::info;
 use serde_json::json;
 use test_network::*;
+use test_samples::ALICE_ID;
+
+const ADMIN_ID: &str =
+    "ed012076E5CA9698296AF9BE2CA45F525CB3BCFDEB7EE068BA56F973E9DD90564EF4FC@admin";
```
s8sato, line 15: ADMIN_ID needs to be agreed between the test `executor_upgrade_should_work` and the executor module `executor_with_admin`

client/tests/integration/smartcontracts/executor_with_admin/src/lib.rs
```diff
@@ -22,7 +22,8 @@ struct Executor {
 }
 
 fn visit_instruction(executor: &mut Executor, authority: &AccountId, isi: &InstructionBox) {
-    if parse!("admin@admin" as AccountId) == *authority {
+    // crate::integration::upgrade::ADMIN_ID
+    if *authority == parse!(AccountId, "ed012076E5CA9698296AF9BE2CA45F525CB3BCFDEB7EE068BA56F973E9DD90564EF4FC@admin") {
```
s8sato, line 26: In attempt to share multihash between the test and the executor module, `parse!` macro was modified to accept `expr`, but no use cases atm

core/src/sumeragi/main_loop.rs
```diff
@@ -1217,12 +1218,11 @@ mod tests {
...
         topology: &Topology,
         leader_key_pair: &KeyPair,
-        tx_signer_key_pair: &KeyPair,
-    ) -> (State, Arc<Kura>, SignedBlock) {
+    ) -> (State, Arc<Kura>, SignedBlock, PublicKey) {
         // Predefined world state
-        let alice_id: AccountId = "alice@wonderland".parse().expect("Valid");
-        let account = Account::new(alice_id.clone(), tx_signer_key_pair.public_key().clone())
-            .build(&alice_id);
+        let (alice_id, alice_keypair) = gen_account_in("wonderland");
```
s8sato, line 1223: Tests irrelevant to default genesis are generating anonymous alices, bobs, etc.

smart_contract/src/lib.rs
```diff
@@ -489,14 +499,12 @@ mod tests {
     const ISI_RESULT: Result<(), ValidationFail> = Ok(());
 
     fn get_test_instruction() -> InstructionBox {
-        let new_asset_id = "tulip##alice@wonderland".parse().unwrap();
-        let register_isi = Register::asset(Asset::new(new_asset_id, 1_u32));
-
-        register_isi.into()
+        let new_asset_id: AssetId = "tulip##ed0120CE7FA46C9DCE7EA4B125E2E36BDB63EA33073E7590AC92816AE1E861B7048B03@wonderland".parse().unwrap();
```
s8sato, line 502: Some tests just have "multihash@domain" literals, being independent of `test_samples`

s8sato (commented):
Finished according to [this policy](https://github.com/hyperledger/iroha/pull/4411#discussion_r1580491931). Then, now rebasing

genesis/src/lib.rs
```diff
@@ -20,12 +21,25 @@ use iroha_data_model::{
...
+});
+
+/// [`KeyPair`] of the genesis account.
+pub static GENESIS_ACCOUNT_KEYPAIR: Lazy<KeyPair> = Lazy::new(|| {
+    KeyPair::new(
+        iroha_crypto::PublicKey::from_str("ed0120E2ECD69DA5833EC10FB3DFAED83A07E5B9CBE9BC39484F0F7DDEC8B46253428B").unwrap(),
+        iroha_crypto::PrivateKey::from_hex(iroha_crypto::Algorithm::Ed25519, "DD61D7A2244A504E78BA80383DFCC0228E25CA131E5A6AF503F71632D23BD76AE2ECD69DA5833EC10FB3DFAED83A07E5B9CBE9BC39484F0F7DDEC8B46253428B").unwrap(),
+    )
+    .unwrap()
+});
```
0x009922, line 42: Is it right that Iroha has a hardcoded magic key pair associated with the genesis account?

I thought that the intention is that Iroha should not know genesis private key at all, and receive the public one via configuration (alongside with pre-signed genesis transaction).

test_samples/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+pub fn gen_account_in(domain: impl core::fmt::Display) -> (AccountId, KeyPair) {
+    let key_pair = KeyPair::random();
+    let account_id = format!("{}@{}", key_pair.public_key(), domain)
+        .parse()
+        .expect("domain name should be valid");
+    (account_id, key_pair)
+}
+
+macro_rules! static_signatory_ed25519 {
+    ( $kp:ident, $vk:expr, $sk:expr ) => {
```
0x009922, line 22: `sk` - secret key? `vk`? I am confused, please use full identifiers.

test_samples/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+#[cfg(feature = "rand")]
+pub fn gen_account_in(domain: impl core::fmt::Display) -> (AccountId, KeyPair) {
+    let key_pair = KeyPair::random();
+    let account_id = format!("{}@{}", key_pair.public_key(), domain)
+        .parse()
+        .expect("domain name should be valid");
+    (account_id, key_pair)
+}
+
+macro_rules! static_signatory_ed25519 {
```
0x009922, line 21: I would split that into two macros with clearer naming: `declare_key_pair` and `declare_account_with_key_pair`.

test_samples/src/lib.rs
s8sato: These predefined signatories are also used when there is a need to agree on them between the prerequisites and the test body e.g. iroha_core::smartcontracts::isi::tests

genesis/src/lib.rs
s8sato: It is not intended to reflect even ongoing changes. I'd keep refactoring to a minimum where it is outside of the main focus

s8sato: Ok, GENESIS_ACCOUNT_ID would be obtained from the configuration. Then GENESIS_ACCOUNT_KEYPAIR would be removed

client_cli/pytests/common/consts.py
```diff
@@ -15,11 +15,11 @@ class Stderr(Enum):
     Enum for standard error messages.
     """
 
-    CANNOT_BE_EMPTY = "cannot be empty\n\nFor more information, try '--help'.\n"
+    EMPTY = "mpty"
```
Erigara, line 18: typo?

mversic (approved):

test_samples/src/lib.rs
mversic: doesn't matter right now because it's just in tests

configs/swarm/genesis.json
```diff
@@ -90,24 +81,33 @@
...
         }
       },
       {
         "Transfer": {
           "Domain": {
-            "source_id": "genesis@genesis",
+            "source_id": "ed01204164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4@genesis",
             "object": "wonderland",
-            "destination_id": "alice@wonderland"
+            "destination_id": "ed0120CE7FA46C9DCE7EA4B125E2E36BDB63EA33073E7590AC92816AE1E861B7048B03@wonderland"
```
s8sato, line 110: Here `genesis.json` has a literal genesis public key, which means `kagami genesis` requires some input of the genesis public key. That's why I made `kagami genesis` read `GENESIS_PUBLIC_KEY` environment variable

scripts/tests/consistency.sh
```diff
@@ -3,8 +3,9 @@ set -e
 
 case $1 in
     "genesis")
-        cargo run --release --bin kagami -- genesis --executor-path-in-genesis ./executor.wasm | diff - configs/swarm/genesis.json || {
-            echo 'Please re-generate the genesis with `cargo run --release --bin kagami -- genesis --executor-path-in-genesis ./executor.wasm > configs/swarm/genesis.json`'
+        GENESIS_PUBLIC_KEY=ed01204164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4 cargo run --release --bin kagami -- genesis --executor-path-in-genesis ./executor.wasm | diff - configs/swarm/genesis.json || {
+            echo 'Please re-generate the default genesis with `GENESIS_PUBLIC_KEY=ed01204164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4 cargo run --release --bin kagami -- genesis --executor-path-in-genesis ./executor.wasm > configs/swarm/genesis.json`'
+            echo 'The assumption here is that the authority of the default genesis transaction is `test_samples::SAMPLE_GENESIS_ACCOUNT_ID`'
```
s8sato, line 8: As long as `genesis.json` is not generated on demand but tracked in git, there is always the problem of which genesis public key should be specified. I've introduced an assumption that the authority of the default genesis transaction is `test_samples::SAMPLE_GENESIS_ACCOUNT_ID`, which is consistent with the preset `GENESIS_PUBLIC_KEY` in swarm

s8sato: So `SAMPLE_GENESIS_ACCOUNT_ID` is deeply involved with the default genesis, but its prefix is not DEFAULT but SAMPLE because it may be used elsewhere e.g. a synthetic genesis

core/src/smartcontracts/isi/domain.rs
```diff
@@ -50,14 +50,9 @@ pub mod isi {
...
-            account_id
-                .name
-                .validate_len(state_transaction.config.ident_length_limits)
-                .map_err(Error::from)?;
-
-            if account_id == *iroha_genesis::GENESIS_ACCOUNT_ID {
+            if *account_id.domain_id() == *iroha_genesis::GENESIS_DOMAIN_ID {
                 return Err(InstructionExecutionError::InvariantViolation(
-                    "Not allowed to register `genesis@genesis` account".to_owned(),
+                    "Not allowed to register account in genesis domain".to_owned(),
```
s8sato, line 55: Not genesis account but genesis domain of the registration target now causes transaction rejection

core/src/tx.rs
```diff
@@ -95,7 +94,7 @@ impl AcceptedTransaction {
             }));
         }
 
-        if *iroha_genesis::GENESIS_ACCOUNT_ID == *tx.authority() {
+        if *iroha_genesis::GENESIS_DOMAIN_ID == *tx.authority().domain_id() {
             return Err(AcceptTransactionFail::UnexpectedGenesisAccountSignature);
```
s8sato, line 98: Not genesis account but genesis domain of the authority now causes transaction rejection

s8sato (commented):
Your feedbacks applied. See [diff](https://github.com/hyperledger/iroha/compare/f90e2eed392ed1206285ecec3f43e97204d02029..6f39591892108fb886a1d1ee531627a4ea9060ae)

- fixed genesis account configuration flow @0x009922 
- please check diff and re-approve @mversic

client_cli/pytests/common/consts.py
s8sato: "mpty" was the intersection of various Empty error messages and was considered enough to identify the cause, but I revised the messages to contain "Empty"

configs/swarm/genesis.json
mversic: I think you already mentioned this recently: we may not need genesis account. It was recently pointed out to me by Makoto that we don't need to sign the genesis file. Instead of starting peers with the genesis public key, we can start them with hash of the genesis block. This completely removes the need to have genesis account for the purposes of signing. 

But can we do without genesis account even in genesis instructions?

client_cli/.gitignore
Erigara, line 1: Is is supposed to be added into repo?

s8sato: I think they are byproduct of pytest and should be ignored as well as that of peers side:
https://github.com/hyperledger/iroha/blob/1606d088fdbfaec3d8cce4a561e2991ba24c49c6/.gitignore#L49

Erigara (dismissed):

tools/kagami/src/genesis.rs
```diff
@@ -76,58 +79,66 @@ impl<T: Write> RunArgs<T> for Args {
...
     builder: RawGenesisBlockBuilder<executor_state::SetPath>,
 ) -> color_eyre::Result<RawGenesisBlockFile> {
+    let genesis_account_id: AccountId = {
+        let multihash =
+            std::env::var("GENESIS_PUBLIC_KEY").wrap_err("GENESIS_PUBLIC_KEY should be set")?;
+        let domain = &*GENESIS_DOMAIN_ID;
+        format!("{multihash}@{domain}")
+            .parse()
+            .wrap_err("GENESIS_PUBLIC_KEY should be valid multihash")?
+    };
```
0x009922, line 89: To make it uniform with the rest of Kagami, please define it as a Clap argument of the `genesis` subcommand (`--genesis-public-key`/`GENESIS_PUBLIC_KEY`)

0x009922 (commented):
Overall no objections, LGTM (except for Kagami UI).

configs/swarm/genesis.json
s8sato: So let's reopen #4409 and discuss it there. Do you think suggested changes in this PR are ok for now?

s8sato:
Made GENESIS_PUBLIC_KEY an arg in kagami. I accidentally mixed [diff](https://github.com/hyperledger/iroha/compare/7493bc2b84107a01fb8098c83cdec90a440912bf..afb1cbb20f495dcde2727ba79a77b6604416a0c3) with other PRs. Sorry for the inconvenience

s8sato:
- Made GENESIS_PUBLIC_KEY an arg in kagami. See [diff](https://github.com/hyperledger/iroha/compare/45189ee7f43e1c6a6eab063740fd5242e9ebde36..0eb152b4cbbce30181419a7c8bdf54f038c97c15)
- `cargo test -p iroha_core --lib -- snapshot::tests::can_read_snapshot_after_writing` seems to be failing since 1606d088f

configs/swarm/genesis.json
mversic: this should be considered in #4555

Erigara:
> cargo test -p iroha_core --lib -- snapshot::tests::can_read_snapshot_after_writing seems to be failing since https://github.com/hyperledger/iroha/commit/1606d088fdbfaec3d8cce4a561e2991ba24c49c6

Going to fix this issue

s8sato:
- Reverted it to be ignored: `cargo test --all-features -p iroha_smart_contract --doc -- parse` has been failing

mversic (approved):

Erigara (approved):

PR #4413, opened 2024-04-10
[fix] #4253: Check genesis pub key in genesis round

### LINKED ISSUE #4253: Genesis is signed with correct key
When genesis is received via listen for genesis via `sumeragi_listen_for_genesis`(`BlockCreated`/`BlockSyncUpdate`) it is not verified that genesis is signed with `IROHA_GENESIS_PUBLIC_KEY`. Not even leader that creates the genesis block verifies that genesis is signed with correct key in `sumeragi_init_commit_genesis`

Erigara:
## Description

Verify that genesis transaction is indeed signed by genesis. 

### Linked issue

Closes #4253 

### Benefits

Not possible to submit wrong genesis.

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -271,10 +271,12 @@ mod valid {
         /// - Block header transaction hashes don't match with computed transaction hashes
         /// - Error during validation of individual transactions
         /// - Topology field is incorrect
+        /// - Transaction in the genesis block is not signed by the genesis public key
         pub fn validate(
             block: SignedBlock,
             topology: &Topology,
             expected_chain_id: &ChainId,
+            genesis_public_key: &PublicKey,
             state_block: &mut StateBlock<'_>,
         ) -> WithEvents<Result<ValidBlock, (SignedBlock, BlockValidationError)>> {
             let expected_block_height = state_block.height() + 1;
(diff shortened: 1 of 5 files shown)


s8sato (dismissed):

core/src/block.rs
```diff
@@ -275,6 +275,7 @@ mod valid {
             block: SignedBlock,
```
DCNick3, line 275: ^ I suggest to add the new error condition into the doc above

`- Transaction in the genesis block is not signed by the genesis public key`

DCNick3:
Other that this small nit, LGTM

Erigara:
> do we have a test for this?

Don't think so, i've tested this manually as described [here](https://github.com/hyperledger/iroha/issues/4253#issuecomment-1929742514).

mversic:
> > do we have a test for this?
> 
> Don't think so, i've tested this manually as described [here](https://github.com/hyperledger/iroha/issues/4253#issuecomment-1929742514).

My fear is that we could easily regress to this bug again

Erigara:
I will add the test.

mversic (dismissed):

DCNick3 (approved):

mversic (approved):

PR #4427, opened 2024-04-12
[fix] #4226: Prevent registering genesis Domain or Account

### LINKED ISSUE #4226: Prevent registering genesis Domain or Account
We have to make sure that instructions that try to register "genesis" domain or "genesis@genesis" account are always rejected. Tests should be added as well. Additionally, if there are none, tests should be added that confirm that any transaction signed by the "genesis@genesis" is rejected

dima74:
## Description

Disallow registering `genesis` domain or `genesis@genesis` account

### Linked issue

Closes #4226

### Checklist


core/src/smartcontracts/isi/domain.rs
```diff
@@ -55,6 +55,12 @@ pub mod isi {
                 .validate_len(state_transaction.config.ident_length_limits)
                 .map_err(Error::from)?;
 
+            if account_id == *iroha_genesis::GENESIS_ACCOUNT_ID {
```
Asem-Abdelhady, line 58: Is the best way to prevent the domain and the account by hardcoding a checking step in the isi?  
what is your opinion @Erigara

mversic: This is the 1st place in the core that this check can be placed at. We could put this check into executor, but I don't think this should be optional so this is the appropriate place for it

On the other hand, you might be on to something. We can place this check into a custom `Deserialize`/`Decode` for `Register<Account>`. But that would require us to hardcode `GENESIS_ACCOUNT_ID` in the `data_model`

so I'm thinking it's fine like this

mversic (approved):

Asem-Abdelhady (approved):

core/src/smartcontracts/isi/domain.rs
Erigara: Yeah, i also think that this place is good

Issue #4412, opened 2024-04-10
[suggestion] Adopt multihash format for private keys


0x009922:
In configurations, we use the following format for public/private keys:

```toml
public_key = "ed012014A2900349A8CB6250D8CE5399D1964766B559A37F48E8C98C1EA83F365D8A72"
private_key = { algorithm = "ed25519", payload = "67AAB998A3951A071324B815BAF425309D6AF8D2D2636A3FC2BEA01C322BC23B14A2900349A8CB6250D8CE5399D1964766B559A37F48E8C98C1EA83F365D8A72" }
```

While multihash format is used for `public_key`, why can't we use it for the private one too?

It would make config more consistent and less confusing, and will also remove necessity to set private key in the environment through 2 split variables `PRIVATE_KEY_ALGORITHM` and `PRIVATE_KEY_PAYLOAD`.

Erigara:
Sounds good to me, what are downsides?

DCNick3:
Same position here. Sounds good and doesn't seem to have any downsides on the first glance.

Issue #4300, opened 2024-02-18
Implement config tracing


0x009922:
Implement tracing as per PRC: [Proposal 8 - Trace Configuration Resolution](https://github.com/hyperledger/iroha-rfcs/blob/main/text/001-config-overhaul.md#proposal-8---trace-configuration-resolution).

0x009922:
Closed by #4456

Issue #4299, opened 2024-02-18
Enhance configuration errors


0x009922:
This might be probably solved as part of #4297.

Points:

- Diagnostics with source spans, hints etc
- Improve the way multiple errors are collected
- Possibly, use an in-house error type instead of an umbrella `eyre::Report`.

0x009922:
[`error_stack`](https://docs.rs/error-stack/latest/error_stack/) might be a good option too.

0x009922:
### Alternative design

So far, I have come up with the following design.

Imagine we have a simplified configuration, having the following structure:

```rust
#[derive(Deserialize, Debug)]  
pub struct Root {  
    pub chain_id: String,  
    pub torii: Torii,  
    pub kura: Kura,  
    pub telemetry: Telemetry,  
    pub logger: Logger,  
}  
  
#[derive(Deserialize, Debug)]  
pub struct Torii {  
    pub address: WithOrigin<SocketAddr>,  
    pub max_content_len: u64,  
}  
  
#[derive(Deserialize, Debug)]  
pub struct Kura {  
    pub store_dir: WithOrigin<PathBuf>,  
    pub debug_force: bool,  
}  
  
#[derive(Deserialize, Debug)]  
pub struct Telemetry {  
    pub out_file: Option<WithOrigin<PathBuf>>,  
}  
  
#[derive(Deserialize, Debug)]  
pub struct Logger {  
    pub level: LogLevel,  
}  
  
#[derive(Deserialize, Debug, Default)]  
pub enum LogLevel {  
    Debug,  
    #[default]  
    Info,  
    Warning,  
    Error,  
}
```

The only new thing here so far is `WithOrigin<T>` struct. It encapsulates the read value with information about where this value comes from, using something like `ParameterOrigin`:

```rust
pub enum ParameterOrigin {  
    File { path: PathBuf, name: String },  
    Env { key: String, name: String },  
    Default { name: String },  
}  
  
pub struct WithOrigin<T> { /* ... */ }
```

This `ParameterOrigin` might be used for the following cases:

- Primarily, to be able to produce useful warnings/errors later, in parts of Iroha which are far from configuration parsing. For example, when Torii fails to bind to a specific address (`failed to listen to localhost:1337`), it will be able to also say
  > btw this value comes from `torii.address` parameter set in `iroha.toml` file
  > ...or `API_ADDRESS` env var
[...]

PR #4456, opened 2024-04-17
feat: improve config debug-ability

### LINKED ISSUE #3470: [suggestion] Enhance configuration parsing with precise field locations
### Feature request

I would like to get meaningful errors when I pass configuration (via JSON, ENV vars or both) to Iroha 2.

Consider the following cases that currently occur:

- I have passed `PUBLIC_KEY` and `PRIVATE_KEY` env vars, and I get an error:

  > Please add `PUBLIC_KEY and PRIVATE_KEY` to the configuration.

  It turned out that yes, in JSON configuration I could provide them as top-level fields with these names, but when it comes to ENV, I should specify them as `IROHA_PUBLIC_KEY` and `IROHA_PRIVATE_KEY`.

  _By the way:_ yes, it prints exactly "`PUBLIC_KEY AND PRIVATE_KEY`", not "`PUBLIC_KEY` and `PRIVATE_KEY`".

  **The descriptive error would be:**

  > Please add `PUBLIC_KEY` (env: `IROHA_PUBLIC_KEY`) and `PRIVATE_KEY` (env: `IROHA_PRIVATE_KEY`) fields to the configuration.

- I forgot to specify the P2P addr:

  > Please add `p2p_addr` to the configuration.

  **The descriptive error would be:**

  > Please add `TORII.P2P_ADDR` (env: `TORII_P2P_ADDR`) to the configuration.

### Thoughts on implementation

Introduce a data structure for configuration parameter location:

```rust
struct Location {
  env_var: &'static str,
  serde: &'static str
}
```

Then refactor the configuration proc macros so they can generate all of the locations in order to use them **statically**.

**Sample macro usage:**

```rust
#[derive(Proxy)]
#[proxy(locations = "config_locations_private", env_prefix = "IROHA_")]
struct Iroha {
  public_key: PublicKey,
  sumeragi: Sumeragi
}

#[derive(Proxy)]
#[proxy(env_prefix = "SUMERAGI_")]
struct Sumeragi {
  p2p_addr: String
}
```

**Generated modules**

```rust
// by proc macro
mod config_locations_private {
  use crate::Location;

  pub const public_key: Location = Location {
    env_var: "IROHA_PUBLIC_KEY",
    serde: "PUBLIC_KEY"
  }

  pub mod sumeragi {
    pub const p2p_addr: Location = Location {
      env_var: "SUMERAGI.P2P_ADDR",
      serde: "SUMERAGI_P2P_ADDR"
    }
  }
}

// visibility control, if needed
pub mod config_locations {
  pub use super::config_locations_private;
}
```

Then, this code could be used in other parts to throw descriptive errors, like that:

```rust
if let (Some(public_key), Some(private_key)) = (&self.public_key, &self.private_key) {
  // ...
} else {
  return Err(ConfigError::MissingParameters {
    locations: vec![config_locations::public_key, config_locations::private_key]
  });
}

```

### Motivation

Iroha will be even more useful if it throws descriptive and helpful errors when misconfiguration happens. 

Also, it will be useful for #2373 - the tool will be able to define ENV variables in a docker-compose configuration with static check.

### Who can help?

_No response_

0x009922:
### Description

This PR contains a major improvement of the work done in https://github.com/hyperledger/iroha/pull/4239. This is a big step towards better debug-ability of Iroha configuration system due to better errors. And yet, there are still many things to improve further!

Key points:

- New `ConfigReader` API, which handles configuration sources in a dynamic manner rather than relying on `serde` completely, thus giving high flexibility in how to handle different cases. It is also built targeting Iroha domain specifically and doesn't have any extra stuff. Much less boilerplate.
- Heavy use of `error_stack` crate, which gives higher control over errors reporting than `eyre` does.
- `--trace-config` works! Uses `log` & `stderrlog` crates for it. Seems to be not a heavy solution for this. It is completely separate from `tracing` stack.

### Further enhancements

There are still many things that might be improved.

- Enhance use of attachments API provided by `error_stack`:
    - Use hooks to colour them conditionally
    - Attach hints and suggestions
    - Unify attachments further as specific structs, not just `format!`
- Use spans to config files, and build reports showing the actual file contents
- Fine-tune API of `ConfigReader`, use better naming and abstractions, perhaps.
- Optimise internals of `ConfigReader`. There is lots of naive code and extra allocations.
- Find a way to reduce extra context changes in `error_stack`. Sometimes it would be really cool to just "forward" underlying context without changing it, but currently it's how `error_stack` is designed. So, we have sometimes stacks like "Failed to load config -> Failed to load config -> failed to read file system -> ..."
- Possibly, introduce a `slim` compile feature for Iroha, which will remove `--trace-config` support alongside with `log` & `stderrlog` dependencies, and more
- Use a single multihash string for private keys, thus removing necessity to set private keys in env as two separate variables (e.g. `PRIVATE_KEY_ALGORITHM` + `PRIVATE_KEY_PAYLOAD`, and removing cratchy `env_custom` API from `ConfigReader`
    - https://github.com/hyperledger/iroha/issues/4412
- Migrate to Axum, finally! Errors handling in Torii might be improved a lot.
    - https://github.com/hyperledger/iroha/issues/3776

### Linked isues

Closes #3470, #4299, #4300

### Gallery

_Note: I've changed some bits after making these screenshots._

<details>

<summary>Some terminal screenshots</summary>

Pasted image 20240417100106

Pasted image 20240417160528

Pasted image 20240417160759

Pasted image 20240417160900

Pasted image 20240417161142

Pasted image 20240417161212

Pasted image 20240417161247

Pasted image 20240417161611

Pasted image 20240417162020

</details>

### TODO

- [x] Document internals of `iroha_config_base`
- [x] Fix `iroha_client_cli` compilation failure
- [x] Enabled failing test back: `fails_with_no_trusted...` in `cli/src/lib.rs`
- [x] Self-review


data_model/src/block.rs
```diff
@@ -88,7 +88,7 @@ mod model {
     )]
     #[display(fmt = "({header})")]
     #[allow(missing_docs)]
-    pub(crate) struct BlockPayload {
+    pub struct BlockPayload {
```
mversic, line 91: can you explain yourself here? This shouldn't be public

data_model/src/transaction.rs
```diff
@@ -98,7 +98,7 @@ mod model {
         Serialize,
         IntoSchema,
     )]
-    pub(crate) struct TransactionPayload {
+    pub struct TransactionPayload {
```
mversic, line 101: the same as [here](https://github.com/hyperledger/iroha/pull/4456/files#r1568427689)

data_model/src/block.rs
0x009922: This warning appears when I run `cargo check -p iroha_swarm` (this is about `TransactionPayload`, but the cause is the same):

```
warning: type `transaction::model::TransactionPayload` is more private than the item `transaction::SignedTransaction::signatures`
   --> data_model/src/transaction.rs:297:5
    |
297 |     pub fn signatures(&self) -> &SignaturesOf<TransactionPayload> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method `transaction::SignedTransaction::signatures` is reachable at visibility `pub`
    |
note: but type `transaction::model::TransactionPayload` is only usable at visibility `pub(crate)`
   --> data_model/src/transaction.rs:101:5
    |
101 |     pub(crate) struct TransactionPayload {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: `#[warn(private_interfaces)]` on by default
```

However, it doesn't appear if I check `iroha`, or `iroha_core`, or `iroha_client` (targetly). I guess that is because they all by the end of the day use `iroha_data_model` with `transparent_api` enabled, while `iroha_swarm` uses it without any features.

I guess the temporary solution would be, first of all, to create an issue like "Fix all `transparent_api` usage across workspace" (you also mentioned `_transparent_api` here: https://github.com/hyperledger/iroha/pull/4382#discussion_r1565317258), but in scope of this PR either:

1. Return `pub(crate)` and silent the warning, temporarily
2. Enable `transparent_api` for `iroha_swarm`, temporarily

What do you think?

mversic: check what @s8sato did [here](https://github.com/hyperledger/iroha/pull/4411#discussion_r1565275625)

if we don't want to leak private types here:
`pub fn signatures(&self) -> &SignaturesOf<TransactionPayload> {`
we can consider changing the function to:
`pub fn signatures(&self) -> &SignaturesOf<SignedTransaction> {`

> **I guess the temporary solution would be, first of all, to create an issue like "Fix all transparent_api usage across workspace" (you also mentioned _transparent_api here: https://github.com/hyperledger/iroha/pull/4382#discussion_r1565317258), but in scope of this PR either:**

there isn't much to fix. We just need to remove `transparent_api` from `test_network` because it's leaking into our integration tests and then client appears to have access to internals when in fact it doesn't. But that's too big of an effort so we have to live with it for now

0x009922:
@DCNick3, I would appreciate if you can take a look at the procmacro I wrote here, in <kbd>config/base/derive</kbd>.

data_model/src/block.rs
0x009922: I tried to change it to `SignedTransaction` (and `SignedBlock`), but I need to transmute the type then. What is the proper way to do so?

0x009922:
Debugging CI: new reports are really helpful!

image

data_model/src/block.rs
DCNick3: > if we don't want to leak private types here:
> pub fn signatures(&self) -> &SignaturesOf<TransactionPayload> {
> we can consider changing the function to:
> pub fn signatures(&self) -> &SignaturesOf<SignedTransaction> {

I don't think it's a good idea, as `SignedTransaction` is not scale-compatible with `TransactionPayload`. I guess we can make another versioned structure for transaction signatures and expose them as newtypes? Or we can expose the `TransactionPayload` itself..

DCNick3:
A bit concerned with the inclusion of source code location in the error trace (the `at some/source/location.rs` lines). I don't think it's useful for users and questionably useful for iroha developers. Maybe remove it/gate being some verbosity flag?

0x009922:
> A bit concerned with the inclusion of source code location in the error trace (the `at some/source/location.rs` lines). I don't think it's useful for users and questionably useful for iroha developers. Maybe remove it/gate being some verbosity flag?

I am concerned too. There are also complete backtraces attached, which are displayed in dev environment. Anyway, we can disable `Location`'s printing using the hooks. I used it in tests to capture snapshots. Will think about it.

config/base/derive/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                                        }
+                                        (Some(lit), next) => {
+                                            (lit, next)
+                                        },
+                                    };
+                                    let expr: syn::Expr = lit
+                                        .parse()
+                                        .map_err(|err| LocalError::BadDefaultExpr(lit.span(), err))?;
+                                    attr_default = Some(AttrDefault::Expr(expr));
+                                    rest = expect_comma_or_none(next)?;
```
DCNick3, line 269: Is there a particular reason you decided agains re-using syn's `syn::punctuated::Punctuated::<YourTokenType, Token![,]>` for parsing commans and decided to roll out your own comma parser?

One unfortunate thing is that you have to add a separate type for a token and then implement `Parse` for it, but as a bonus you can have commas (and optional trailing ones) handled for you

config/base/derive/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                }
+
+                let mut attr_default = None;
+                let mut attr_env = None;
+                let mut attr_env_custom = false;
+                let mut attr_nested = false;
+
+                let mut rest = *cursor;
+
+                while let Some((tt, next)) = rest.token_tree() {
```
DCNick3, line 242: There is a less deeply-nested way to write this:

```rust
while !rest.eof() {
    if let Some((tt, next)) = rest.ident() {
        // ...
    } else {
        // ...
    }
}
```

config/base/derive/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                                    args.args.first().expect("should be exactly 1")
+                                {
+                                    found = Some((Some(ty), &last_segment.ident));
+                                }
+                            }
+                            syn::PathArguments::None => found = Some((None, &last_segment.ident)),
+                            _ => {}
+                        }
+                    }
+                }
```
DCNick3, line 416: You can return value from the match instead of assigning it to the mutable variable

```suggestion
                let found = if let syn::Type::Path(type_path) = ty {
                    if let Some(last_segment) = type_path.path.segments.last() {
                        match &last_segment.arguments {
                            syn::PathArguments::AngleBracketed(args) if args.args.len() == 1 => {
                                if let syn::GenericArgument::Type(ty) =
                                    args.args.first().expect("should be exactly 1")
                                {
                                    Some((Some(ty), &last_segment.ident))
                                }
                            }
                            syn::PathArguments::None => Some((None, &last_segment.ident)),
                            _ => None,
                        }
                    }
                }
```

config/base/src/read.rs
```diff
@@ -0,0 +1,784 @@
...
+};
+
+/// A type that implements reading from [`ConfigReader`]
+pub trait ReadConfig: Sized {
+    /// Returns the [`FinalWrap`] with self and the reader itself, transformed
+    /// throughout the process of reading.
+    ///
+    /// The wrap is guaranteed to unwrap safely if the reader emits
+    /// no error upon [`ConfigReader::into_result`].
+    fn read(reader: ConfigReader) -> (FinalWrap<Self>, ConfigReader);
```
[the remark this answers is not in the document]
0x009922: I think your point is valid. If there was a reason I chose "immutable" pattern initially, it doesn't seem to be relevant after all, and passing a mutable reference would indeed reduce noise.

config/base/derive/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+        generics: syn::Generics,
+        data: darling::ast::Data<(), Field>,
+    }
+
+    impl Input {
+        pub fn lower(self, emitter: &mut Emitter) -> Option<codegen::Ir> {
+            let mut halt_codegen = false;
+
+            for i in self.generics.params {
+                emit!(emitter, i, "generics are not supported");
```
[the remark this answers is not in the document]
0x009922: > Maybe we can actually add a functionality prefixing the macro name to the emitted error 🤔

I think it would be great. Will see what I can do with `Emitter`.

config/base/src/read.rs
```diff
@@ -0,0 +1,784 @@
...
+
+    /// Finish with the value only.
+    #[must_use]
+    pub fn finish(self) -> (FinalWrap<Option<T>>, ConfigReader) {
+        match self {
+            Self::Errored { reader, .. } => (FinalWrap::errored(), reader),
+            Self::WithOptionalValue { reader, value } => {
+                (FinalWrap::value(value.map(WithOrigin::into_value)), reader)
+            }
+            Self::WithValue { .. } => unreachable!(),
```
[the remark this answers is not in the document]
0x009922: Indeed... will try

config/base/derive/src/lib.rs
0x009922: No reason apart from my ignorance of its existence!

Will definitely try to re-use it.

config/base/derive/src/lib.rs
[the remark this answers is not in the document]
0x009922: Eeeh, `Emitter` relies on `ToTokensError` trait, which essentially converts everything to a token stream. Therefore, I see the two options:
1. Patch the resulting token stream in a fine-grained manner, prefixing compile error messages. Implications: it might be a complex thing to do?
2. Introduce an intermediate trait for `iroha_macro_utils::Emitter`, which will "intercept" actual errors (plain strings, `syn::Error`, etc) and put a prefix inside them _before_ their conversion to a token stream. Implications: might be a high burden to re-implement the existing thing for a not-so-crucial feature.

What do you think?

For now, I'll simply prefix this particular error.

config/base/src/read.rs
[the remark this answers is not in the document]
0x009922: Remembered: my initial concern was that Rust wouldn't allow to pass a mutable reference inside `ReadingParameter` struct, and that I have to "drag" the reader internally all the way down.

But, it turned out it works fine. Now code is cleaner, thanks.

config/base/derive/src/lib.rs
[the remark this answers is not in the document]
DCNick3: Approach №2 would be cleaner, but that's for a separate PR.

core/src/sumeragi/mod.rs
```diff
@@ -153,8 +153,10 @@ impl SumeragiHandle {
 
             current_topology = match state_view.height() {
                 0 => {
-                    assert!(!sumeragi_config.trusted_peers.is_empty());
-                    Topology::new(sumeragi_config.trusted_peers.clone())
+                    // we rely on the guarantee of that `sumeragi.trusted_peers` contains
+                    // at least id of the peer itself
+                    // FIXME: find a way to guarantee it statically
+                    Topology::new(sumeragi_config.trusted_peers.value().clone())
```
0x009922, line 159: Any comments on that?

0x009922 (commented):
Self-review

0x009922:
@Erigara @mversic if you can, please take a look at the overall workspace changes, outside of the `ConfigReader` API.

Here is an overview of how changes are spread:

visualization

The major part, which is under <kbd>/config/base</kbd>, is under the review of @DCNick3.

I am asking you about looking at other parts.

client_cli/src/main.rs
```diff
@@ -158,6 +156,7 @@ trait RunArgs {
     ///
     /// # Errors
     /// if inner command errors
+    // TODO: replace `eyre::Result` with `error_stack::Result`
```
Erigara, line 159: Please create ticket if not planned for current PR.

core/src/sumeragi/mod.rs
Erigara: We probably can do smt like that: 

```rust
struct TrustedPeers {
    myself: PeerId, 
    others: UniqueVec<PeerId>
}
```

This way we can be sure that self is present, but i don't think this is high priority issue.

cli/src/lib.rs
```diff
@@ -533,8 +663,46 @@ pub fn read_config_and_genesis(
...
+    let mut emitter = Emitter::new();
+
+    // These cause race condition in tests, due to them actually binding TCP listeners
+    // Since these validations are primarily for the convenience of the end user,
+    // it seems a fine compromise to run it only in release mode
+    #[cfg(release)]
+    {
+        validate_try_bind_address(&mut emitter, &config.network.address);
+        validate_try_bind_address(&mut emitter, &config.torii.address);
+    }
```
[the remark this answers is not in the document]
0x009922: With the way it is currently implemented, it makes the path of error extraction quite complicated. Apart from that, `warp` doesn't return an error and just panics, therefore, extraction of an error from Torii becomes even more complicated.

I found this way to be a fine compromise.

cli/src/lib.rs
```diff
@@ -348,76 +383,112 @@ impl Iroha {
...
+
     /// To make `Iroha` peer work it should be started first. After
     /// that moment it will listen for incoming requests and messages.
     ///
     /// # Errors
     /// - Forwards initialisation error.
     #[iroha_futures::telemetry_future]
-    pub async fn start(&mut self) -> Result<()> {
+    pub async fn start_torii(self) -> Result<Iroha<ToriiStarted>, StartError> {
```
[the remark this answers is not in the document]
0x009922: The current implementation is like that due to the `test_network`. I will explore what I can do.

client_cli/src/main.rs
0x009922: I found it simpler to just remove this comment =)

cli/src/lib.rs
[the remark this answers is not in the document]
0x009922: Created a separate ticket instead: https://github.com/hyperledger/iroha/issues/4562

data_model/src/block.rs
mversic: and why can't we just silence this lint instead of making any code changes?

mversic: I would prefer that you open an issue, so we can merge this

0x009922: Reverted

mversic (dismissed):

mversic (approved):

mversic (approved):

DCNick3 (approved):

PR #4457, opened 2024-04-17
[refactor] #3240: Guard against secrets leakage

### LINKED ISSUE #3240: Guard against secrets leakage

dima74:
## Description

* Added `Secret` to `PrivateKey`. Access to inner `PrivateKeyInner` now requires import of `ExposeSecret` trait via `expose_secret` method
* `Display`, `Debug` and `Serialize` implementations for `PrivateKey` now returns `"[REDACTED]"`
* Added `ExposedPrivateKey` wrapper which can be formatted and serialized as usual
* Note that I don't used `secrecy` crate because it requires to implement `Zeroize` trait, but inner key struct (`ed25519_dalek::SigningKey`) implement only `ZeroizeOnDrop`, and not `Zeroize`. So I used modified version of `Secret` which requires `ZeroizeOnDrop` instead of `Zeroize`. This potentially could be controversal, so any suggestions are welcome

### Linked issue

Closes #3240

### Checklist

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -649,6 +649,20 @@ mod tests {
             base
         }
 
+        fn config_to_toml_value(config: PartialUserConfig) -> Result<toml::Value> {
+            use iroha_crypto::ExposedPrivateKey;
+            let private_key = config.private_key.as_ref().unwrap().clone();
+            let genesis_private_key = config.genesis.private_key.as_ref().unwrap().clone();
+            let mut result = toml::Value::try_from(config)?;
+
+            // private key will be serialized as "[REDACTED PrivateKey]" so need to restore it
+            result["private_key"] = toml::Value::try_from(ExposedPrivateKey(private_key))?;
+            result["genesis"]["private_key"] =
+                toml::Value::try_from(ExposedPrivateKey(genesis_private_key))?;
+
+            Ok(result)
+        }
+
         #[test]
         fn relative_file_paths_resolution() -> Result<()> {
             // Given
@@ -663,7 +677,7 @@ mod tests {
                 cfg.kura.store_dir.set("../storage".into());
                 cfg.snapshot.store_dir.set("../snapshots".into());
                 cfg.dev_telemetry.out_file.set("../logs/telemetry".into());
-                toml::Value::try_from(cfg)?
+                config_to_toml_value(cfg)?
             };
 
             let dir = tempfile::tempdir()?;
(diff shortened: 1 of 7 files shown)


cli/src/lib.rs
```diff
@@ -649,6 +649,20 @@ mod tests {
...
         }
 
+        fn config_to_toml_value(config: PartialUserConfig) -> Result<toml::Value> {
+            use iroha_crypto::ExposedPrivateKey;
+            let private_key = config.private_key.as_ref().unwrap().clone();
+            let genesis_private_key = config.genesis.private_key.as_ref().unwrap().clone();
+            let mut result = toml::Value::try_from(config)?;
+
+            // private key will be serialized as "[REDACTED]" so need to restore it
+            result["private_key"] = toml::Value::try_from(ExposedPrivateKey(private_key))?;
```
DCNick3, line 659: It's unfortunate that this will require us to remember to update this function if we ever move/add new private keys to the config, but I didn't think of any alternative other than making `PartialUserConfig` serialize with exposed keys, which is probably not what we want.

crypto/src/lib.rs
```diff
@@ -884,23 +976,45 @@ mod tests {
...
         }
     }
 
+    #[test]
+    fn private_key_format_or_serialize_redacted() {
+        let key_pair = KeyPair::random();
+        let (_, private_key) = key_pair.into_parts();
+
+        assert_eq!(
+            serde_json::to_string(&private_key).expect("Can't serialize key"),
```
DCNick3, line 1012:
```suggestion
            serde_json::to_string(&private_key).expect("Couldn't serialize key"),
```

cli/src/lib.rs
dima74: Yes, but this function is used only in tests, so in case new private keys added to the config, tests will fail

DCNick3 (approved):

Erigara (approved):

mversic:
I was thinking that this change won't affect `iroha_crypto` crate

mversic:
> I was thinking that this change won't affect `iroha_crypto` crate. IMO This should be an external functionality and not integrated into `iroha_crypto`, the change should be in `iroha_config` mostly

dima74:
To summarize today discussion:
* Current implementation:
  * Add `Secret` inside `PrivateKey`
  * Disadvantages: new public struct `ExposedPrivateKey` in `iroha_crypto` crate
* Alternative implementation:
  * Add secret to usages of `PrivateKey` in other crates, such as `iroha_config`
  * Disadvantages: easy to forget to add `Secret` in case new `PrivateKey` field added to some struct in future

The plan is to merge current implementation as-is if no new concerns arise

PR #4461, opened 2024-04-17
[fix] #4332: Remove corresponding triggers on `Unregister<Domain>`

### LINKED ISSUE #4332: Forbid `Unregister<Domain>` if referenced from a live trigger
In `iroha_core`, we should prevent `Unregister<Domain>` if it's listed in `TriggerId::domain_id`. Custom defined executor can check for this and choose to transfer the trigger ownership to some other domain before executing `Unregister<Domain>`

mirna-versic:
### Description of the Change

Triggers that are living in a domain are now deleted when that domain is deleted.

Closes #4332 

### Benefits

Fixes the bug.

### Possible Drawbacks 
None

### DIFF
--- a/core/src/smartcontracts/isi/world.rs
+++ b/core/src/smartcontracts/isi/world.rs
@@ -119,7 +119,16 @@ pub mod isi {
         ) -> Result<(), Error> {
             let domain_id = self.object_id;
 
+            let triggers_in_domain = state_transaction
+                .world()
+                .triggers()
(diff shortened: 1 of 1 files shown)


Erigara:
Please format commit like: 
```
[fix] #4332: Remove corresponding triggers on `Unregister<Domain>`
```

Erigara (approved):
LGTM

PR #4486, opened 2024-04-19
[feature] #4227: Support JSON in parity_scale_decoder

### LINKED ISSUE #4227: Support encoding/decoding to/from JSON in `parity_scale_decoder`
1. rename `parity_scale_decoder` to `parity_scale_codec` or something similar
2. add support to encode/decode SCALE to/from JSON

atm, we're supporting only decoding to Rust debug format. I think we should just replace that with JSON, but we can also support different formats with some flag

dima74:
## Description

* Renamed `parity_scale_decoder` to `parity_scale_cli` (in issue it was suggested to rename to `parity_scale_codec`, but I am not sure that it is good idea because we use dependency crate with exactly same name - `parity-scale-codec`)
* Added `scale-to-json` and `json-to-scale` subcommands
* Existing functionality (decode to Rust debug format) left as-is, only renamed to `scale-to-rust` subcommand

### Linked issue

Closes #4227

### Checklist


configs/swarm/executor.wasm
0x009922, line 1: Hm, you removed this, but `genesis.json` still points to this file.

Why remove?

tools/parity_scale_cli/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+```bash
+cargo build --bin parity_scale_cli
+```
+
+If your terminal does not support colours, run:
+
+```bash
+cargo build --features no_color --bin parity_scale_cli
+```
```
0x009922, line 17: Instead of feature-gating it, I would suggest following the same pattern that `iroha` (crate in <kbd>/cli/</kbd>) itself follows:

- It detects colors support automatically via `supports_color` crate
- User can override it by `--terminal-colors=false`/`--terminal-colors=true` flag

tools/parity_scale_cli/src/main.rs
```diff
@@ -289,18 +386,50 @@ mod tests {
...
+
+    #[test]
+    fn test_decode_encode_trigger() {
+        test_decode_encode("trigger.bin", "Trigger");
+    }
+
+    fn test_decode_encode(sample_path: &str, type_id: &str) {
+        let mut binary = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        binary.push("samples/");
+        binary.push(sample_path);
```
0x009922, line 422:
```suggestion
        let binary = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
        .join("samples")
        .join(sample_path);
```

tools/parity_scale_cli/src/main.rs
```diff
@@ -56,44 +60,81 @@ use iroha_primitives::{
...
+    fn new() -> Box<Self> {
+        Box::new(Self(PhantomData))
+    }
+}
+
+trait Converter {
+    fn scale_to_rust(&self, input: &[u8]) -> Result<String>;
+    fn scale_to_json(&self, input: &[u8]) -> Result<String>;
+    fn json_to_scale(&self, input: &str) -> Result<Vec<u8>>;
+}
```
0x009922, line 101: Hm, the pattern of a map of types with dynamic `Conveter`s for each type seems to be an overkill for me, because it all goes down to a single `ConverterImpl` anyway.

I think overall implementation would be simpler if we have just a few generic functions:

```rust
mod convert {
	fn scale_to_rust<T: DecodeAll + Debug>(value: T, input: &[u8]) -> Result<String> { todo! }
	
	fn scale_to_json<T: DecodeAll + Serialize>(value: T, input: &[u8]) -> Result<String> { todo! }
	
	fn json_to_scale<T: DeserializeOwned + Encode>(value: T, input: &str) -> Result<Vec<u8>> { todo! }
}
```

(and also to replace `String`, `String`, and `Vec<u8>` with some newtypes like `RustDebug`, `Json`, and `Scale` to reduce blindness a bit)

Is there an issue I cannot recognise? Why did you chose the way that you chose?

mversic: Just to say that we don't need `to_rust`. `to_json` is sufficient

Erigara: @0x009922 i'm not sure how you gonna convert string name of type the function you want to call. 
It could be some big `match` statement, but map in this case seems more readable to me.

Since our converter type doesn't have any state instead of `dyn` object we can just use function pointers.

```rust
struct Converter {
    scale_to_rust: fn(input: &[u8]) -> RustDebug,
    scale_to_json: fn(input: &[u8]) -> Json,
    json_to_scale: fn(input: &[u8]) -> Scale,
}
```

But since this program is essentially one off it doesn't really matter and i'm fine with `dyn` object approach.

0x009922: Aah, that is the part I missed, thanks.

Than I am fine too.

configs/swarm/executor.wasm
0x009922: I see, fine then.

Issue #4352, opened 2024-03-11
dynamic parameter value


mversic:
1. I think that we might want to split `***Limits` into it's constituents
2. I think that having `Numeric` is sufficient because values should be loosely typed and checked in executor

Given the previous 2 points, I think that instead of `ParameterValueBox` we could just have `StringWithJson` as we do for permission tokens. This is because the value for parameters defined in executor can indeed be anything. Even though there are some hard coded, I don't think it's necessarily worth to have them strongly typed inside an enum

This is something I think should be explored in a separate PR

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/4305#discussion_r1515992597_

Erigara:
Relevant here #4028.

We should consider that there might be arbitrary parameters used by executor and we have hard-coded parameters used by wasm engine, sumeragi. 

Maybe we can define all on-chain parameters which look like smt like this:

```
struct Parameters {
    block_time: Option<Duration>,
    commit_time: Option<Duration>,
    wsv_asset_metadata_limits: Option<MetadataLimits>,
    // ...
    /// Can contain arbitrary key-value pairs to be used by executor
    executor_params: Map<Name, StringWithJson>,
}
```

What do you think @mversic @0x009922?

I should notice that executor already can kinda have arbitrary parameters by either using `Parameter`s as they exist right now or executor can create domain and use it's metadata for storing it's state.

mversic:
> We should consider that there might be arbitrary parameters used by executor and we have hard-coded parameters used by wasm engine, sumeragi.

what's the problem here?

Erigara:
> what's the problem here?

It's not exactly a problem, but it would be nice if crucial on-chain parameters would be strongly typed with exact types (not stored as random JSON strings)

Erigara:
> Are you saying that we can just reuse Domain::Metadata or any other metadata for custom parameters?

Yes, it can place them in any metadata, since permissions for the executor aren't checked.

0x009922:
I don't mind any format for executor parameters as long as WASM/Sumeragi parameters are strictly specified and typed.

mversic:
> > what's the problem here?
> 
> It's not exactly a problem, but it would be nice if crucial on-chain parameters would be strongly typed with exact types (not stored as random JSON strings)

that's internally, no? Externally (from the client side), we should do it in the same way, i.e. as JSON string?

Erigara:
[...]

PR #4513, opened 2024-04-25
[refactor] #3901: Add method to define parameter in executor `migrate` entrypoint

### LINKED ISSUE #3901: [suggestion] Move `NewParameter` into Executor

dima74:
## Description

* Added method `add_parameter` which can be called from executor (in `migrate` entrypoint) to define new executor parameter
* Added test which demonstrates how executor can use it, as suggested [here](https://github.com/hyperledger/iroha/pull/4310#issuecomment-1964247631)
* Note that this PR was mostly implemented before discussion in #4352, so potentially this should be implemented differently. E.g. if it will be decided to store executor parameters in `Domain::metadata`, then there will be no need for `add_parameter` method since executor parameter will be defined using `SetKeyValue` ISI
* Note that I didn't removed `NewParameter` ISI since it is currently used to define chain-wide parameters (for wasm engine and sumeragi), and potentially chain-wide parameters should be defined/stored differently as discussed in https://github.com/hyperledger/iroha/issues/4352#issuecomment-2076735651

### Linked issue

Closes #3901
Related: #4352
Related: #4310

### Checklist


mversic:
superseeded by #4597

Issue #4264, opened 2024-02-07
Remove predictable ordering hack from sumeragi


mversic:
This was introduced in #4263 to prevent consensus from operating if there is a bug in the implementation. We should make sure that consensus operation is well covered with tests and remove this backdoor

Erigara:
Going to be removed in #4518

AlexStroke:
I see in the code that that hack has been removed and it looks like iroha has been running successfully without it. But I can't understand if the sorting of peers has become unpredictable, which seems to be the main problem why this task was opened.

mversic:
this is ok, we can close it

Issue #4346, opened 2024-03-06
Refactor sumeragi into a state machine


mversic:
Current sumeragi is an unmaintanable mess and full of potential loopholes. Additionally, we want this piece of code to be exemplary for both outside reviewers and auditors. We need to rewrite it into a state machine. This refactor should be done carefully and, if possible, incrementally

Issue #3522, opened 2023-05-23
[refactor]: Consider forbid empty topology


Erigara:
At the moment it is possible to hit panic by calling some methods on empty topology.

We should consider forbidding empty topology.

PR #4518, opened 2024-04-26
refactor!: don't send public key with signature

### LINKED ISSUE #4393: Should we pack public keys with signatures?
our `Signature` is represented as:

```rs
pub struct Signature {
    public_key: PublicKey,
    payload: ConstVec<u8>,
}
```

but in the case that we know who's signed the message having `public_key` in the signature is unnecessary. It is unnecessary because the verifier still has to know what public key it expects in the signature. I noticed this due to the fact that we require a key pair to sign genesis. In that case there is only 1 signee and there can't be any doubt as to what public key must have been used.

Another example is the consensus. We don't have to send full blown public key with every signature. Since all nodes have to agree on the topology I think it is sufficient to just send index of the node in the current topology

I'm not sure what we can do with MST

mversic:
### Linked issue

Closes #4393 #4410 #4264

### Checklist


data_model/src/block.rs
```diff
@@ -92,13 +91,29 @@ mod model {
...
+        Eq,
+        PartialOrd,
+        Ord,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    pub struct BlockSignature(pub u64, pub SignatureOf<BlockPayload>);
```
Erigara, line 115: I think you should write doc comment here to specify what `u64` means here. 
You might as well consider not using tuple struct here.

data_model/src/block.rs
```diff
@@ -189,10 +205,10 @@ impl SignedBlock {
     /// Add additional signatures to this block
     #[must_use]
     #[cfg(feature = "transparent_api")]
-    pub fn sign(mut self, key_pair: &KeyPair) -> Self {
+    pub fn sign(mut self, private_key: &PrivateKey, node_pos: u64) -> Self {
         let SignedBlock::V1(block) = &mut self;
-        let signature = iroha_crypto::SignatureOf::new(key_pair, &block.payload);
-        block.signatures.insert(signature);
+        let signature = SignatureOf::new(private_key, &block.payload);
+        block.signatures.push(BlockSignature(node_pos, signature));
```
Erigara, line 211: Implication here is that now it's possible to store multiple signatures under the same `node_pos`.
Can't this be used by malicious peer to add irrelevant signatures to block?

core/src/sumeragi/network_topology.rs
```diff
@@ -36,90 +35,111 @@ pub struct ConsensusTopology<'topology> {
...
+                    filtered.extend(topology.proxy_tail_index() + 1..topology.0.len());
                 }
                 _ => {}
             };
         }
+
         signatures
             .into_iter()
-            .filter(|signature| public_keys.contains(signature.public_key()))
+            .filter(|signature| filtered.contains(&(signature.0 as usize)))
```
Erigara, line 135: Looks like check that signature actually match peer's public key should be checked here as well

core/src/block.rs
```diff
@@ -474,8 +474,11 @@ mod valid {
 
         /// Add additional signatures for [`Self`].
         #[must_use]
-        pub fn sign(self, key_pair: &KeyPair) -> ValidBlock {
-            ValidBlock(self.0.sign(key_pair))
+        pub fn sign(self, key_pair: &KeyPair, topology: &Topology) -> ValidBlock {
+            let node_pos = topology
+                .position(key_pair.public_key())
+                .expect("BUG: Node is not in topology");
```
Erigara, line 480: Why you decided to panic here? Maybe allow downstream users to handle error condition?

Erigara (changes requested):

core/src/sumeragi/network_topology.rs
mversic: ugh, good point. This will take me some time to fix

mversic: should be fixed now, but the change propagated

core/src/block.rs
mversic: I was thinking that it indicates a bug in consensus. The node that is signing must be present in the topology. This must never happen, I consider it the same as accessing array out of bounds.

I will reconsider

mversic: encountering this panic means that `Role::Undefined` is trying to sign a block. This mustn't happen in a correct consensus so panic is appropriate

core/src/sumeragi/view_change.rs
```diff
@@ -41,90 +42,92 @@ pub struct ProofBuilder(SignedProof);
...
-                self.signatures.insert(signature);
+    fn merge_signatures(&mut self, other: Vec<ViewChangeProofSignature>, topology: &Topology) {
+        for (signatory_idx, signature) in other {
+            let public_key = topology.as_ref()[signatory_idx as usize].public_key();
+
+            if signature.verify(public_key, &self.payload).is_ok() {
+                self.signatures.push((signatory_idx, signature));
             }
         }
     }
```
s8sato, line 77: I don't know what the real harm will be, but attackers might be able to exchange each other's positions in the topology with the promise that they will exchange public keys and signatures and then declare them to the network:

```
topology: .., (i, key_j), .., (j, key_i), ..
other:    .., (i, sig_j), .., (j, sig_i), ..
```

this would pass the verification

core/src/sumeragi/network_topology.rs
```diff
@@ -36,229 +34,205 @@ pub struct ConsensusTopology<'topology> {
...
-            peers.sort_by_cached_key(|peer| !to_lift_up.contains(&peer.public_key));
+    pub fn lift_up_peers(&mut self, to_lift_up: impl IntoIterator<Item = usize>) {
+        let to_lift_up: IndexSet<_> = to_lift_up.into_iter().collect();
+
+        let mut signatory_idx = 0;
+        self.0.sort_by_cached_key(|_| {
+            let res = !to_lift_up.contains(&signatory_idx);
+            signatory_idx += 1;
+            res
         });
```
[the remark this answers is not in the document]
mversic: good point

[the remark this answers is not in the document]
mversic: fixed

core/src/sumeragi/view_change.rs
s8sato: I posted just a thought to bring it to the attention of other reviewers. I'll try a little more thought.

I think this is an issue that appears only after unpacking signature. If there were any vulnerability, it would be in referencing the topology as a map from index to public key

s8sato: After thinking about it, such a situation is unlikely to occur since there would be no chance to misidentify one's public key to other peers after the peer registration

core/src/sumeragi/network_topology.rs
```diff
@@ -36,229 +34,210 @@ pub struct ConsensusTopology<'topology> {
...
-    pub fn update_topology(&mut self, block_signees: &[PublicKey], new_peers: UniqueVec<PeerId>) {
-        self.lift_up_peers(block_signees);
+    fn lift_up_peers(&mut self, to_lift_up: impl IntoIterator<Item = usize>) {
+        let to_lift_up =
+            to_lift_up
+                .into_iter()
+                .fold(Vec::with_capacity(self.0.len()), |mut acc, x| {
+                    assert!(x < self.0.len(), "Unknown committed block signature");
+                    acc.extend(core::iter::repeat(false).take(x - acc.len()));
+                    acc.push(true);
```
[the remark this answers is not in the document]
mversic: fixed again

s8sato (approved):

core/src/sumeragi/view_change.rs
```diff
@@ -28,103 +30,153 @@ struct ProofPayload {
...
+            })
+        }
+
+        fn validate_signatures(&self) -> Result<(), &'static str> {
+            self.signatures
+                .iter()
+                .try_fold(IndexSet::new(), |mut acc, elem| {
+                    if !acc.insert(elem) {
+                        return Err("Duplicate signature in proof");
+                    }
```
mversic, line 67: I'm not sure whether proofs are duplicated if peer is repeatedly requesting view change?

@Erigara @SamHSmith

core/src/sumeragi/network_topology.rs
```diff
@@ -784,20 +667,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn observing_peers_empty() {
```
Erigara, line 788: This case is still possible for topology size of 1-3.

core/src/sumeragi/network_topology.rs
```diff
@@ -713,20 +610,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn validating_peers_empty() {
```
Erigara, line 717: Possible for topology size of 1-2.

core/src/sumeragi/view_change.rs
Erigara: Since you've changed type from set to vector singnatures now can duplicate. 

You can look at `merge_signatures` function.

data_model/src/query/mod.rs
```diff
@@ -1203,12 +1203,28 @@ pub mod http {
...
+            Eq,
+            PartialOrd,
+            Ord,
+            Decode,
+            Encode,
+            Deserialize,
+            Serialize,
+            IntoSchema,
+        )]
+        pub struct QuerySignature(pub PublicKey, pub SignatureOf<ClientQueryPayload>);
```
Erigara, line 1220: After Sato's work pub key can be extracted from `AccoundId`.

data_model/src/block.rs
```diff
@@ -90,13 +89,34 @@ mod model {
...
+        PartialOrd,
+        Ord,
+        Decode,
+        Encode,
+        Deserialize,
+        Serialize,
+        IntoSchema,
+    )]
+    pub struct BlockSignature(
+        /// Index of the block in the topology
```
Erigara, line 114:
```suggesion
    /// Index of the peer in the topology
```

Erigara (changes requested):

core/src/block.rs
```diff
@@ -253,17 +250,146 @@ mod valid {
...
+                        == leader_index =>
+                {
+                    let mut additional_leader_signatures =
+                        topology.filter_signatures_by_roles(&[Role::Leader], block_signatures);
+
+                    if additional_leader_signatures.next().is_some() {
+                        return Err(SignatureVerificationError::DuplicateSignatures {
+                            signatory: leader_index,
+                        });
+                    }
```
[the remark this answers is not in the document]
mversic: malicious peer can send all sorts of invalid blocks. What you're saying is true, but is no less dangerous than sending other types of invalid blocks. We just reject a block sent from this peer, there is no harm in that IMO

core/src/block.rs
```diff
@@ -253,17 +250,146 @@ mod valid {
...
+
+                    signature
+                }
+                _ => {
+                    return Err(SignatureVerificationError::LeaderMissing);
+                }
+            };
+
+            leader_signature
+                .verify(topology.leader().public_key(), block.payload())
```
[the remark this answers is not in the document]
mversic: yes, I agree

core/src/smartcontracts/isi/query.rs
```diff
@@ -567,10 +565,12 @@ mod tests {
         let tx_limits = &state_block.transaction_executor().transaction_limits;
         let va_tx = AcceptedTransaction::accept(tx, &chain_id, tx_limits)?;
 
-        let topology = Topology::new(UniqueVec::new());
+        let (peer_public_key, _) = KeyPair::random().into_parts();
+        let peer_id = PeerId::new("127.0.0.1:8080".parse().unwrap(), peer_public_key);
+        let topology = Topology::new(vec![peer_id]);
```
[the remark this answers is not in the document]
mversic: I didn't think about that though. I didn't strictly impose the limit that there must be 1 peer in the topology, rather it's just that new topology has to be created with 1 peer at least. Do you agree with this change? Do you think we should impose a limit to prevent topology from going to 0?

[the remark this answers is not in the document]
Erigara: >  Do you think we should impose a limit to prevent topology from going to 0?

Right now we would panic in `Topology::new` if topology is empty. 
As well with calls to `leader` and `proxy_tail`.
But it's possible to get empty topology through `block_committed` call afaik.

So i think we should bring consistency.
Essentially empty topology is possible but it's the end of the blockchain since no new blocks would be produced ever.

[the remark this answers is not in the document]
Erigara: Found ticket #3522 created by my a while ago

core/src/sumeragi/main_loop.rs
```diff
@@ -648,107 +618,100 @@ impl Sumeragi {
...
-                                }
-                                Err(error) => error!(%addr, role=%Role::Leader, ?error),
-                            };
-                        }
+                    if created_in > self.pipeline_time() / 2 {
+                        warn!("Creating block takes too much time. This might prevent consensus from operating. Consider increasing `commit_time` or decreasing `max_transactions_in_block`");
                     }
-                }
-            }
-            Role::ProxyTail => {
```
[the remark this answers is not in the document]
mversic: now block will only be committed on observing `BlockCommitted` or `BlockSyncUpdate` messages

core/src/sumeragi/main_loop.rs
```diff
@@ -430,325 +435,378 @@ impl Sumeragi {
...
-                info!(%addr, %role, block=%block_hash, "Block sync update received");
+                info!(
+                    peer_id=%self.peer_id,
+                    role=%self.role(),
+                    block=%block.hash(),
+                    "Block sync update received"
+                );
+
+                // FIX: Release writer before handling block sync
+                let curr_block = voting_block.take();
```
mversic, line 454: I have this issue in 4 places. It happens either on `BlockSyncUpdate` or `BlockCreated`. At the time of receiving a new block it's possible that the block is invalid and I would like to keep the old block while validating the new one. However, this creates deadlock as I have not released previous writer

@Erigara do you have some suggestions?

Erigara: Unfortunately it’s current limitation of our state layout. 
That it’s single writer and locks state for duration of block.

When i’ve designed state i made some decisions for making optimisations for happy path (block committed) without problems. 
We might reconsider some of this decisions if necessary.

Maybe we can be smarter about when to reject incoming block? Like checking signatures, height, hashes this checks doesn’t need ability to modify state.

mversic: well, I'll just drop the current voting block in this PR as it was before and open an issue for this

core/src/block.rs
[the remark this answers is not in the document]
mversic: > I think what @Erigara is saying is that the block validation will globally and repeatedly fail unless it has a mechanism to identify and ignore (or expel from the validator set) the peer that spoof index zero with its signature.

spoofing 0 index is just one way in which malicious peer can affect consensus. Worst case, malicious peer can withhold sending any messages to other peers. It will only fail consensus repeatedly until the malicious peer becomes observing peer. 0 index spoofing is no special case and I don't think we should be more forgiving in this case

core/src/sumeragi/main_loop.rs
```diff
@@ -430,325 +449,381 @@ impl Sumeragi {
...
+
+                if let Some(mut voted_block) = voting_block.take() {
+                    add_signatures::<true>(&mut voted_block, valid_signatures, &self.topology);
+                    *voting_block = self.try_commit_block(voted_block, is_genesis_peer);
                 } else {
                     // NOTE: Due to the nature of distributed systems, signatures can sometimes be received before
                     // the block (sent by the leader). Collect the signatures and wait for the block to be received
                     voting_signatures.extend(valid_signatures);
                 }
             }
```
Erigara, line 657: By looking at this code and `try_commit` it look like we don't check that signature indeed belong to the peer. 
So any peer (including malicious leader itself) can send `BlockSigned` message with random block mathing signatures for enough validating/obsering peers and block will be committed. 

Another problem is that on this path signatures are not checked for duplication.
So as before peer can submit `BlockSigned` message with the same index and signature and block will be committed.

core/src/sumeragi/main_loop.rs
```diff
@@ -430,325 +449,381 @@ impl Sumeragi {
...
+        let votes_count = voting_block.block.as_ref().signatures().len();
+        if votes_count + 1 >= self.topology.min_votes_for_commit() {
+            voting_block.block.sign(&self.key_pair, &self.topology);
+
+            let committed_block = voting_block
+                .block
+                .commit(&self.topology)
+                .unpack(|e| self.send_event(e))
+                .expect("INTERNAL BUG: Proxy tail failed to commit block");
+
```
Erigara, line 745: Maybe instead of unwrapping we can try to commit the block in the `if` statement?
I think it would be more reliable in long term if we decide that to commit block it's not enough to just receive enough signatures.
And proxy tail can add it's signature on `BlockCreated` message.

core/src/block.rs
```diff
@@ -389,91 +513,84 @@ mod valid {
...
         pub fn commit(
             self,
             topology: &Topology,
         ) -> WithEvents<Result<CommittedBlock, (ValidBlock, BlockValidationError)>> {
-            if !self.0.header().is_genesis() {
-                if let Err(err) = self.verify_signatures(topology) {
+            if !self.as_ref().header().is_genesis() {
+                if let Err(err) = Self::verify_proxy_tail_signature(self.as_ref(), topology) {
                     return WithEvents::new(Err((self, err.into())));
                 }
```
Erigara, line 569: We probably should check that whole set of sigantures is valid.

core/src/block.rs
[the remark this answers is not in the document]
s8sato: Ok, at least this is not a problem specific to this PR. It used to trust public keys instead of indices

core/src/block.rs
mversic: it's checked on `replace_signatures`

mversic: because `ValidBlock` is guaranteed to have all signatures valid

core/src/sumeragi/main_loop.rs
```diff
@@ -322,34 +320,26 @@ impl Sumeragi {
...
-        self.current_topology = new_topology;
-        self.connect_peers(&self.current_topology);
+        self.topology
+            .block_committed(block.as_ref(), state_block.world.peers().cloned());
+        self.connect_peers(&self.topology);
 
         // Commit new block making it's effect visible for the rest of application
         state_block.commit();
+        Strategy::kura_store_block(&self.kura, block);
+
```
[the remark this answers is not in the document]
mversic: fixed

data_model/src/query/mod.rs
mversic: done, although why do we even sign queries?

core/src/sumeragi/main_loop.rs
mversic: opened #4643

data_model/src/query/mod.rs
Erigara: We have query permissions if we want to make sure who makes query we have to sign it.

core/src/sumeragi/main_loop.rs
mversic: 1. `ValidBlock` is guaranteed to have all valid signatures except for the `ProxyTail` signature
2. `valid_bloc.commit` will check that `ProxyTail` signature is valid
3. `try_commit` is only invoked by the `ProxyTail`. You can see there is an `assert` statement inside

core/src/sumeragi/main_loop.rs
mversic: this function is only callable by the `ProxyTail` and `ValidBlock` is guaranteed to have all signatures valid. Consequently, this cannot and must not fail unless the `ProxyTail` is itself faulty

core/src/sumeragi/main_loop.rs
s8sato: I see no problem with peer-signature correspondence, but `let valid_signatures` would be premature naming since the validity is guaranteed in `add_signatures`

I agree that signatures can duplicate

core/src/sumeragi/mod.rs
```diff
@@ -265,7 +273,7 @@ pub struct VotingBlock<'state> {
     /// At what time has this peer voted for this block
     pub voted_at: Instant,
     /// [`WorldState`] after applying transactions to it but before it was committed
-    pub state_block: StateBlock<'state>,
+    pub block_state: StateBlock<'state>,
```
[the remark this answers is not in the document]
mversic: yeah, I revert incorrectly

core/src/sumeragi/network_topology.rs
```diff
@@ -19,10 +20,7 @@ use iroha_primitives::unique_vec::UniqueVec;
 ///
 /// Above is an illustration of how the various operations work for a f = 2 topology.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Topology {
-    /// Current order of peers. The roles of peers are defined based on this order.
-    pub(crate) ordered_peers: UniqueVec<PeerId>,
-}
+pub struct Topology(Vec<PeerId>, usize);
```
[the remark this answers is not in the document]
mversic: true, it's the current view change index of the topology. Every topology starts with index of 0

core/src/sumeragi/view_change.rs
```diff
@@ -20,111 +23,115 @@ pub enum Error {
...
 
 /// The proof of a view change. It needs to be signed by f+1 peers for proof to be valid and view change to happen.
-#[derive(Debug, Clone, Decode, Encode)]
-pub struct SignedProof {
-    signatures: SignaturesOf<ProofPayload>,
-    /// Collection of signatures from the different peers.
-    payload: ProofPayload,
+#[derive(Debug, Clone, Encode)]
+pub struct SignedViewChangeProof {
+    signatures: Vec<ViewChangeProofSignature>,
```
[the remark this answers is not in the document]
mversic: because I want the deserialization to fail if there are duplicates

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
mversic: fixed

core/src/sumeragi/mod.rs
```diff
@@ -138,70 +142,74 @@ impl SumeragiHandle {
...
-        let mut recreate_topology: RecreateTopologyByViewChangeIndex;
+        let mut topology;
 
         {
             let state_view = state.view();
-            let skip_block_count = state_view.block_hashes.len();
+            let skip_block_count: usize = state_view
+                .height()
+                .try_into()
+                .expect("Blockchain height should fit into usize");
```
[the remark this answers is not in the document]
mversic: tbh, idk. I opened #4647 to deal with this

core/src/sumeragi/main_loop.rs
mversic: fixed

core/src/sumeragi/message.rs
```diff
@@ -78,20 +77,14 @@ impl From<&ValidBlock> for BlockSigned {
 #[derive(Debug, Clone, Decode, Encode)]
 #[non_exhaustive]
 pub struct BlockCommitted {
-    /// Hash of the block being signed.
-    pub hash: HashOf<SignedBlock>,
```
Erigara, line 82: I think that hash of block is still helpful for diagnostic, we can reject message not because signature is wrong, but because it was sent for different block entirely.

Same for `BlockSigned` message.

data_model/src/block.rs
```diff
@@ -193,76 +219,85 @@ impl SignedBlock {
...
+
+        if !self
+            .signatures()
+            .collect::<BTreeSet<_>>()
+            .insert(&signature)
+        {
+            return Err(iroha_crypto::Error::Signing(
+                "Duplicate signature".to_owned(),
+            ));
+        }
```
Erigara, line 246: Afaik we compare `BlockSingature` by both fields `index` and `signature`, but some siganture schemes can produce multiple valid results for the same output. 
So malicious peer still can produce multiple votes for a block.
Can we compare just by index?

s8sato: I agree with comparing only index in `BlockSignature`

core/src/sumeragi/main_loop.rs
```diff
@@ -378,377 +406,475 @@ impl Sumeragi {
...
+                        Role::Undefined => error!(
+                            peer_id=%self.peer_id,
+                            role=%self.role(),
+                            "Unknown signatory"
+                        ),
+                        Role::ObservingPeer if view_change_index == 0 => error!(
+                            peer_id=%self.peer_id,
+                            role=%self.role(),
+                            "Signatory is observing peer"
+                        ),
```
s8sato, line 667: I'd like to take this opportunity to ask, why after view changes does the proxy tail accept block signatures from _observing_ peers as well as validating peers? Shouldn't the roles change along with the view changes?

Erigara: I think this was done as optimization, to reduce amount of view changes required to push faulty peer from voting set.

s8sato: Ok, I can see #3110

core/src/sumeragi/message.rs
mversic: it's not quite the same for `BlockSigned` since those can come before `BlockCreated`. I can agree about the `BlockCommitted` though

mversic: done

core/src/sumeragi/main_loop.rs
mversic: > Shouldn't the roles change along with the view changes?

I think they should but the current implementation is such that they don't

core/src/sumeragi/main_loop.rs
```diff
@@ -378,377 +406,487 @@ impl Sumeragi {
...
                 }
             }
-            (BlockMessage::BlockSigned(BlockSigned { hash, signatures }), Role::ProxyTail) => {
-                trace!(block_hash=%hash, "Received block signatures");
+            (BlockMessage::BlockSigned(BlockSigned { signature }), Role::ProxyTail) => {
+                info!(
+                    peer_id=%self.peer_id,
+                    role=%self.role(),
+                    "Received block signatures"
+                );
```
Erigara, line 647: Imo having hash of block from message here is anyway helpful.

data_model/src/block.rs
```diff
@@ -271,16 +307,26 @@ mod candidate {
...
+                .try_fold(BTreeSet::new(), |mut acc, elem| {
+                    if !acc.insert(elem) {
+                        return Err("Duplicate signature in block");
+                    }
+
+                    Ok(acc)
+                })?;
+
+            Ok(())
+        }
```
s8sato, line 329: This seems to check only for duplicates. Where is the signature verification?

data_model/src/block.rs
```diff
@@ -271,16 +307,26 @@ mod candidate {
...
+            self.signatures
+                .iter()
+                .map(|signature| signature.0)
+                .try_fold(BTreeSet::new(), |mut acc, elem| {
+                    if !acc.insert(elem) {
+                        return Err("Duplicate signature in block");
+                    }
+
+                    Ok(acc)
+                })?;
```
s8sato, line 326: The same as

core/src/sumeragi/view_change.rs
```diff
@@ -204,3 +212,69 @@ impl ProofChain {
...
+            self.signatures
+                .iter()
+                .map(|signature| &signature.0)
+                .try_fold(IndexSet::new(), |mut acc, elem| {
+                    if !acc.insert(elem) {
+                        return Err("Duplicate signature in proof");
+                    }
+
+                    Ok(acc)
+                })?;
```
s8sato, line 241: The same as . Also, boundary between `validate` and `validate_signatures` is inconsistent with that of `SignedBlockCandidate`

core/src/sumeragi/message.rs
```diff
@@ -57,19 +58,21 @@ impl From<ValidBlock> for BlockCreated {
 #[non_exhaustive]
 pub struct BlockSigned {
     /// Hash of the block being signed.
-    pub hash: HashOf<BlockPayload>,
-    /// Set of signatures.
-    pub signatures: SignaturesOf<BlockPayload>,
+    pub hash: HashOf<SignedBlock>,
```
Erigara, line 61: Why it's now `HashOf<SignedBlock>`? 
`BlockSignature` itself is `SigantureOf<BlockPayload>`

data_model/src/block.rs
mversic: here it's different because we should compare signature every to every other

core/src/sumeragi/view_change.rs
mversic: it's different because we have to compare every signature to every other. Collecting into a set is the easiest

core/src/sumeragi/message.rs
s8sato: `HashOf<SignedBlock>` does not serve as block identifier during consensus as we recently discussed in #4641

mversic: yes, I used `hash_of_payload` now

core/src/sumeragi/message.rs
```diff
@@ -38,74 +38,126 @@ impl ControlFlowMessage {
...
     }
 }
 
 /// `BlockCommitted` message structure.
-#[derive(Debug, Clone, Decode, Encode)]
-#[non_exhaustive]
+#[derive(Debug, Clone, Encode)]
 pub struct BlockCommitted {
     /// Hash of the block being signed.
     pub hash: HashOf<SignedBlock>,
```
s8sato, line 82: Not this PR issue, but shouldn't `hash` in block messages be unified to `HashOf<BlockPayload>`?

mversic: let's discuss in #4641

core/src/sumeragi/main_loop.rs
mversic: this should be considered in #4346

data_model/src/block.rs
```diff
@@ -132,31 +153,35 @@ impl BlockHeader {
...
+impl SignedBlockV1 {
     #[cfg(feature = "std")]
     fn hash(&self) -> iroha_crypto::HashOf<SignedBlock> {
         iroha_crypto::HashOf::from_untyped_unchecked(iroha_crypto::HashOf::new(self).into())
     }
 }
 
 impl SignedBlock {
+    /// Block payload. Used for tests
+    #[cfg(feature = "transparent_api")]
```
Erigara, line 179: Why we put getters under `transparent_api` flag?

mversic: just this one. Because `BlockPayload` is not supposed to be exposed

Erigara (dismissed):

Cargo.lock
s8sato, line 1: pytest is failing. Otherwise, I can approve

s8sato (approved):

Erigara (approved):

core/src/block.rs
```diff
@@ -790,6 +891,14 @@ mod event {
 
     impl EventProducer for BlockValidationError {
         fn produce_events(&self) -> impl Iterator<Item = PipelineEventBox> {
+            // TODO:
```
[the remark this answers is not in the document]
mversic: created #4722

PR #4553, opened 2024-05-06
ci: Remove executor.wasm from git

### LINKED ISSUE #4549: Build `config/swarm/executor.wasm` in CI
Currently, every developer builds executor on his own machine and commits it into git. Then this executor is used in tests. This creates too many unnecessary git conflicts and prevents me from merging backports directly into `main` branch. What was the reason for doing this in the first place?

We should build executor in the CI every time before running tests in the CI and not track `executor.wasm` in git at all

Erigara:
## Description

Remove `executor.wasm` from git, executor is built in the CI.

### Linked issue

Closes #4549


Erigara:
@BAStos525 ci change for building and providing executor for tests

Erigara:
Shouldn't proceed before figuring out how to ship executor for iroha deployment.

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -35,14 +35,38 @@ jobs:
...
+  build_executor:
+    runs-on: [self-hosted, Linux, iroha2]
+    container:
+      image: hyperledger/iroha2-ci:nightly-2024-04-18
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+      - name: Build iroha executor
+        run: cargo run --bin iroha_wasm_builder_cli -- build ./default_executor --optimize --outfile configs/swarm/executor.wasm
```
mversic, line 48:
```suggestion
        run: cargo run --release --bin iroha_wasm_builder_cli -- build ./default_executor --optimize --outfile configs/swarm/executor.wasm
```

should it run in release mode?

Erigara: probably doesn't metter since it will build `iroha_wasm_builder`, actual executor always builds in release mode

0x009922 (changes requested):
This PR removes `executor.wasm`, but doesn't change anything about `genesis.json`, which still has `"executor_file": "./executor.wasm"`.

One of the intentions behind having the WASM blob stored in Git is to allow users "just open the repo and run `docker compose up`". Perhaps we need to revise this use case in some way, in its root. Until then, this PR breaks things.

Erigara:
Close as not planed for now.

Issue #4500, opened 2024-04-23
Proper black-box integration tests MVP


mversic:
We've long had an issue that our `client/tests/integration` are not proper black box tests. This means the have `iroha` as dependency  when they shouldn't.

The main goal is to remove `iroha` and `test_network` as `dev-dependencies` of `iroha_client`. However, the scope of this task is to produce an MVP with only several tests migrated to this new scheme of running integration tests. Later we can migrate all of the tests, but it's important to provide a framework for new tests first.

We should use [this](https://docs.rs/testcontainers/latest/testcontainers/) to bring up dockers running iroha for every test.

nxsaken:
Questions on the Dockerfile I struggle to find answers to:

- Why do we use `archlinux` as the base image for the build stage?
  - this probably isn't that simple, but can't we use something that supports arm64, like alpine?
- What difference does adding `rust-src` to the toolchain make?
- Can we get away with only installing the minimal toolchain (without rustfmt, clippy and rust-docs)?
  - asking because specifying the image to be `amd64/archlinux` (for testing reasons) makes `rustup` fail to install `rust-docs` for some reason

Erigara:
> Why do we use archlinux as the base image for the build stage? 

Afaik it was chosen just because it's easy to install rust packages, so imo we can replace it.

0x009922:
For the reference, `iroha-javascript` has an internal utility for end-to-end testing: [`test-peer`](https://github.com/hyperledger/iroha-javascript/tree/main/packages/test-peer). It sets up actual `iroha` binary & clean environment for each test, and tests with it work real fast, surprisingly. It has direct library interface and even a simple stateful HTTP server to control peer execution state.

0x009922:
> We should use [this](https://docs.rs/testcontainers/latest/testcontainers/) to bring up dockers running iroha for every test.

Do we really need `testcontainers` for this purpose? While trying to integrate it in tests of `iroha-javascript`, I found it quite heavy and excessive in comparison to plain child-process execution. Using Docker in tests also opens up an infinite hole of possible Docker-related problems which are usually very hard to debug.

dmitrivenger:
@0x009922 can we move this task to "Done"?

PR #4573, opened 2024-05-09
ci: fix `with_coverage`, isolate not `unstable` but `extra_functional`

### LINKED ISSUE #4488: Fix `with_coverage`
After merging #4414 `with_coverage` is failing

s8sato:
## Description

1. redefine `with_coverage` step to exclude client integration tests 
   - and to include p2p integration tests
2. expand the exception of the client integration tests from `unstable_network` to `extra_functional`

### Linked issue

1. closes https://github.com/hyperledger/iroha/issues/4488#issuecomment-2103228798
	- opens https://github.com/hyperledger/iroha/issues/4584
2. sometimes `integration` step fails due to some flaky `extra_functional` tests

### Benefits

2. isolation of flaky tests more in line with current realities

### Checklist

### DIFF
--- a/.github/workflows/iroha2-dev-pr.yml
+++ b/.github/workflows/iroha2-dev-pr.yml
@@ -70,7 +70,7 @@ jobs:
           path: lcov.info
 
   # include: client/tests/integration/
-  # exclude: "unstable_network"
+  # exclude: client/tests/integration/extra_functional
   integration:
     runs-on: [self-hosted, Linux, iroha2]
     container:
@@ -80,10 +80,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
       - name: Run tests, with no-default-features
-        run: mold --run cargo test --no-default-features --no-fail-fast -p iroha_client integration -- --skip unstable_network
+        run: mold --run cargo test --no-default-features --no-fail-fast -p iroha_client integration -- --skip extra_functional
 
-  # include: client/tests/integration/, "unstable_network"
-  unstable:
+  # include: client/tests/integration/extra_functional
+  extra_functional:
     runs-on: [self-hosted, Linux, iroha2]
     container:
       image: hyperledger/iroha2-ci:nightly-2024-04-18
@@ -92,7 +92,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
       - name: Run tests
-        run: mold --run cargo test --no-default-features --no-fail-fast -p iroha_client unstable_network
+        run: mold --run cargo test --no-default-features --no-fail-fast -p iroha_client extra_functional
 
   # Run the job to check that the docker containers are properly buildable
   pr-generator-build:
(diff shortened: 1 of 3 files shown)


.github/workflows/iroha2-dev-pr.yml
```diff
@@ -44,7 +45,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
       - name: Run tests, with coverage
-        run: mold --run cargo test --all-features --workspace --no-fail-fast
+        run: |
+          mold --run cargo test --all-features --no-fail-fast --workspace --exclude iroha_client
+          mold --run cargo test --all-features --no-fail-fast -p iroha_client -- --skip integration
```
s8sato, line 50: image

However, the result of 18% would not indicate that coverage has returned:

image
https://coveralls.io/builds/67405583

Something is wrong. There are some weird items like `target/` and `usr/`

s8sato: Tried several patterns to little effect. Reopened since at least `with_coverage` step works

nxsaken (dismissed):

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -66,7 +69,10 @@ jobs:
           name: lcov.info
           path: lcov.info
 
-  integration:
+  # migrating to pytest
+  # include: client/tests/integration/
+  # exclude: client/tests/integration/extra_functional
+  client_integration:
```
mversic, line 75: I would keep the old name. They are integration tests for all of Iroha, we may move them outside of `client` crate and pull to the root of the project

.github/workflows/iroha2-dev-pr.yml
```diff
@@ -66,7 +69,10 @@ jobs:
           name: lcov.info
           path: lcov.info
 
-  integration:
+  # migrating to pytest
```
mversic, line 72: #4500 we're not migrating to pytest, rather pytest is an addition to iroha integration tests. The reason is that it's too involving for Rust devs to write pytests with allure. We would also want to move pytests out of this repository

mversic (changes requested):

.github/workflows/iroha2-dev-pr.yml
s8sato: Renaming is ok, and what do you think about which step p2p/tests/integration should be included in?

s8sato: Ah, I meant renaming from `client_integration` to `integration` is ok. Keeping the old name is fine

mversic: > what do you think about which step should p2p/tests/integration be included in?

it could be argued both ways. Maybe you could open a separate issue for this?

s8sato: #4584

nxsaken (approved):

.github/workflows/iroha2-dev-pr.yml
s8sato: - Left to #4584

mversic (approved):

PR #4596, opened 2024-05-16
feat!: unify executor data model

### LINKED ISSUE #4352: dynamic parameter value

0x009922:
## Description

This PR solves a few problems:

- Allows executor define its config parameters through a single `ExecutorDataModel` entity.
- Unifies parameters and permission tokens to `ExecutorDataModelObject` trait, having a `definition_id` and an arbitrary JSON `payload` (possible to support SCALE too): #4352
- Parameters no longer configure Sumeragi/WASM/limits. I.e. chain wide configuration could only be set through the config file. **It is temporary:** this PR is a preparation for a proper implementation of chain-wide configuration: #4028
- Removes extra `PermissionTokenSchemaUpdated` event. Since this action only happens during executor migration, which emits its own event `ExecutorEvent::Upgraded`, I also included the updated `ExecutorDataModel` in this event.

## Linked issue

Closes #4352


0x009922:
GitHub doesn't see my updated after force push... closing to reopen.

PR #4597, opened 2024-05-16
feat!: unify executor data model

### LINKED ISSUE #3901: [suggestion] Move `NewParameter` into Executor

### LINKED ISSUE #4352: dynamic parameter value

0x009922:
_Reopening #4596 due to some GitHub force-push branch-recreate issues_

---

## Description

This PR solves a few problems:

- Allows executor define its config parameters through a single `ExecutorDataModel` entity.
- Unifies parameters and permission tokens to `ExecutorDataModelObject` trait, having a `definition_id` and an arbitrary JSON `payload` (possible to support SCALE too):
  - #4352
  - #3901
- Parameters no longer configure Sumeragi/WASM/limits. I.e. chain wide configuration could only be set through the config file. **It is temporary:** this PR is a preparation for a proper implementation of chain-wide configuration: #4028
- Removes extra `PermissionTokenSchemaUpdated` event. Since this action only happens during executor migration, which emits its own event `ExecutorEvent::Upgraded`, I also included the updated `ExecutorDataModel` in this event.

## Linked issue

Closes #4352
Closes #3901


data_model/src/executor.rs
```diff
@@ -47,10 +52,87 @@ mod model {
...
+        Deserialize,
+        Serialize,
+        IntoSchema,
+        Display,
+    )]
+    #[ffi_type]
+    #[display(fmt = "{self:?}")]
+    pub struct ExecutorDataModel {
+        /// Data model JSON schema, typically produced by [`IntoSchema`].
+        pub schema: JsonString,
```
mversic, line 84: make it last field of the struct. `Ord` and `Eq` can exit eagerly if you do so

data_model/src/executor.rs
```diff
@@ -47,10 +52,87 @@ mod model {
...
+        /// Permission tokens supported by the executor.
+        ///
+        /// These IDs refer to the types in the schema.
+        pub permission_token_ids: BTreeSet<PermissionTokenId>,
+        /// Configuration parameters supported by the executor.
+        ///
+        /// These IDs refer to the types in the schema.
+        pub parameter_ids: BTreeSet<ParameterId>,
```
mversic, line 92:
```suggestion
        pub permission_tokens: BTreeSet<PermissionTokenId>,
        /// Configuration parameters supported by the executor.
        ///
        /// These IDs refer to the types in the schema.
        pub parameters: BTreeSet<ParameterId>,
```

how about this? it's simpler and it's not really necessary to emphasize they are ids

data_model/src/executor.rs
```diff
@@ -47,10 +52,87 @@ mod model {
...
+    type DefinitionId: PartialEq + Clone + Into<crate::name::Name>;
+
+    /// Constructor
+    fn new(definition_id: Self::DefinitionId, payload: JsonString) -> Self;
+
+    /// Getter for the object id
+    fn object_definition_id(&self) -> &Self::DefinitionId;
+
+    /// Getter for the object payload
+    fn object_payload(&self) -> &JsonString;
```
mversic, line 133:
```suggestion
    fn id(&self) -> &Self::DefinitionId;

    /// Getter for the object payload
    fn schema(&self) -> &JsonString;
```

data_model/src/permission.rs
```diff
@@ -1,212 +1,124 @@
...
     #[ffi_type]
-    pub struct PermissionTokenSchema {
-        /// Ids of all permission tokens, sorted.
-        pub token_ids: Vec<PermissionTokenId>,
-        /// Type schema of permission tokens
+    #[display(fmt = "PERMISSION TOKEN `{definition_id}`")]
+    pub struct PermissionToken {
+        /// Refers to a type defined in [`crate::executor::ExecutorDataModel`].
+        #[id]
+        pub definition_id: PermissionTokenId,
```
mversic, line 62:
```suggestion
        pub definition_id: PermissionTokenId,
```

you cannot compare permission tokens only with their id because 2 permission tokens can have the same ID but different payloads. Example:
`CanUnregisterDomain {domain_id: "wonderland"}`  != `CanUnregisterDomain {domain_id: "noland"}`

data_model/src/executor.rs
```diff
@@ -47,10 +52,87 @@ mod model {
...
+    }
+}
+
+/// Defines an object that is part of the [`ExecutorDataModel`].
+///
+/// An object consists of id and payload.
+/// ID is a unique type identifier in the executor's data model.
+/// Payload is JSON-serialized type structure, according to the type schema.
+// TODO: support both JSON and SCALE for payload.
+pub trait ExecutorDataModelObject: Sized {
```
mversic, line 122: why don't you put this trait into `smart_contract`?

data_model/src/visit.rs
```diff
@@ -84,6 +82,7 @@ pub trait Visit {
         visit_find_trigger_by_id(&FindTriggerById),
         visit_find_trigger_key_value_by_id_and_key(&FindTriggerKeyValueByIdAndKey),
         visit_find_triggers_by_domain_id(&FindTriggersByDomainId),
+        visit_find_executor_data_model(&FindExecutorDataModel),
```
mversic, line 85: I think end users would appreciate more if we had `FindPermissionTokens` and `FindParameters`. But we can discuss that in a separate issue

data_model/src/query/mod.rs
```diff
@@ -970,18 +962,39 @@ pub mod peer {
...
+        #[ffi_type]
+        pub struct FindExecutorDataModel;
 
-        /// [`FindAllParameters`] Iroha Query finds all [`Peer`]s parameters.
+        /// [`FindAllParameters`] Iroha Query finds all defined executor configuration parameters.
         #[derive(Copy, Display)]
         #[display(fmt = "Find all peers parameters")]
-        // TODO: Unused query. Remove?
         #[ffi_type]
         pub struct FindAllParameters;
```
mversic, line 992:
```suggestion
        pub struct FindAllParameters;
```

so what does this query do now? It finds values of parameters?

mversic: if yes, it shouldn't be in the `executor` module because it also returns values for core parameters

data_model/src/events/data/events.rs
```diff
@@ -575,11 +533,38 @@ mod executor {
...
+        )]
+        #[ffi_type]
+        #[repr(transparent)]
+        pub struct ExecutorUpgraded {
+            /// Updated data model
+            pub new_data_model: ExecutorDataModel,
+        }
+    }
+
+    impl ExecutorUpgraded {
```
mversic, line 565:
```suggestion
    impl ExecutorUpgrade {
```

data_model/src/permission.rs
```diff
@@ -1,212 +1,124 @@
...
-    pub struct PermissionToken {
-        /// Token identifier
-        pub definition_id: PermissionTokenId,
-        /// JSON encoded token payload
-        pub payload: JsonString,
+    #[getset(get = "pub")]
+    #[serde(transparent)]
+    #[repr(transparent)]
+    #[ffi_type(opaque)]
+    pub struct PermissionTokenId {
```
mversic, line 48:
```suggestion
    pub struct PermissionId {
```

I think we can rename `PermissionToken` to just `Permission`

mversic (changes requested):

data_model/src/permission.rs
0x009922: So, the primary purpose of `#[id]` is to quickly compare of equality?

Does this mean that `PermissionToken` & `Parameter` cannot implement `Identifiable`?

data_model/src/query/mod.rs
0x009922: Yes, it returns all currently set executor parameters. In this PR, I haven't yet introduced the concept of core parameters.

As we discussed previously (if memory doesn't fail me), after proper core params implementation, we are going to have a single composite query `FindParameters` which will return both executor's key-value map and strictly typed set of core params.

data_model/src/visit.rs
0x009922: Created #4618

data_model/src/executor.rs
0x009922: As for `id`, well, let's rename this field from `definition_id` to `id` in `PermissionToken` and `Parameter` structs as well then. **But,** the reason I added `object_` prefix here is because it created ambiguity for Rust, because structs themselves also had `id()` getters in their impls.

But that's not going to be an issue if I move this trait into `iroha_smart_contract`, as you suggested.

As for `schema`, it's not, isn't it? It's a JSON-encoded permission structure, i.e. payload, not the schema of it.

data_model/src/executor.rs
0x009922: Indeed.

data_model/src/permission.rs
0x009922: Done. Bloated the overall diff significantly...

data_model/src/executor.rs
mversic: > As for schema, it's not, isn't it? It's a JSON-encoded permission structure, i.e. payload, not the schema of it.

true, then `payload` instead of `object_payload`

data_model/src/permission.rs
mversic: > Does this mean that `PermissionToken` & `Parameter` cannot implement Identifiable?

yes, it means that `PermissionToken` cannot implement `Identifiable`
no, it doesn't mean that `Parameter` cannot implement `Identifiable`

mversic: > yes, it means that `PermissionToken` cannot implement Identifiable

user can have assigned multiple (as I pointed out in the example) permissions of the same id with different payloads.

>  no, it doesn't mean that Parameter cannot implement Identifiable

`ParameterId` uniquely identifies a parameter, it's not contingent on the payload

data_model/src/lib.rs
```diff
@@ -212,66 +211,20 @@ impl<EXPECTED, GOT> EnumTryAsError<EXPECTED, GOT> {
...
-            LengthLimits(LengthLimits),
-            Numeric(
-                #[skip_from]
-                #[skip_try_from]
-                Numeric,
-            ),
-        }
-
-        /// Identification of a [`Parameter`].
+        /// Identification of [`Parameter`].
```
DCNick3, line 227:
```suggestion
        /// Identifies a global on-chain [`Parameter`]. Available parameter names are defined by the executor
```

data_model/src/lib.rs
```diff
@@ -281,97 +234,51 @@ pub mod parameter {
...
-            Decode,
-            Encode,
-            DeserializeFromStr,
-            SerializeDisplay,
-            IntoSchema,
+            Debug, Display, Clone, IdEqOrdHash, Decode, Encode, Deserialize, Serialize, IntoSchema,
         )]
-        #[display(fmt = "?{id}={val}")]
-        /// A chain-wide configuration parameter and its value.
+        #[display(fmt = "PARAMETER `{id}`")]
```
DCNick3, line 261: Why did you remove the value from the display impl? I feel this might reduce usefulness of the log messages...

data_model/src/executor.rs
0x009922: Done

data_model/src/lib.rs
0x009922: I did it to unify `Parameter` and `Permission`: the latter one also displays only `id`, without `payload`.

Shall we then display payloads of both?

smart_contract/executor/src/lib.rs
```diff
@@ -172,34 +182,111 @@ macro_rules! deny {
...
-        if let Some(pos) = self.0.iter().position(|token_id| *token_id == to_remove) {
-            self.0.remove(pos);
-            <T as iroha_schema::IntoSchema>::remove_from_schema(&mut self.1);
+        if *object.id() != expected_id {
+            return Err(ConvertDataModelObjectError::Id(object.id().clone().into()));
         }
+
+        object
+            .payload()
+            .deserialize_to()
```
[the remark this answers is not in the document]
0x009922: No objection at all, makes sense, renamed.

smart_contract/executor/src/permission.rs
```diff
@@ -3,27 +3,55 @@
...
+
+    /// Convert into [`PermissionObject`]
+    fn into_object(self) -> PermissionObject {
+        TokenConvertWrap(self).into_object()
+    }
+}
+
+/// Utility to convert between `T`: [`Permission`] and [`PermissionObject`].
+#[derive(Serialize, Deserialize)]
+pub(crate) struct TokenConvertWrap<T>(pub T);
```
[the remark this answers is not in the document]
0x009922: Made just `ConvertWrap`.

data_model/src/events/data/events.rs
```diff
@@ -575,11 +533,38 @@ mod executor {
...
+        pub fn new_data_model(&self) -> &ExecutorDataModel {
```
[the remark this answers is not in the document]
0x009922: Tried:

- `ExecutorUpgrade`: worked
- `Permission` and `Parameter`: worked for `id`, but not for `payload`
- `ExecutorDataModel`: failed for all fields

Left appropriate comments.

client/tests/integration/asset.rs
```diff
@@ -296,7 +296,7 @@ fn find_rate_and_make_exchange_isi_should_succeed() {
     let alice_id = ALICE_ID.clone();
     let alice_can_transfer_asset = |asset_id: AssetId, owner_key_pair: KeyPair| {
         let instruction = Grant::permission(
-            PermissionToken::new(
+            Permission::new(
```
mversic, line 299: please open a separate PR for this change. We'll merge it and then you can rebase this PR on top of it

data_model/src/lib.rs
DCNick3: I would prefer it this way, yeah. I don't think they will be _too large_, so it probably won't lead to any problems

client/tests/integration/upgrade.rs
```diff
@@ -225,7 +225,45 @@ fn migration_fail_should_not_cause_any_effects() {
...
+            .map(Result::unwrap)
+        {
+            if let EventBox::Data(DataEvent::Executor(ExecutorEvent::Upgraded(ExecutorUpgrade {
+                new_data_model,
+            }))) = event
+            {
+                let _ = sender.send(new_data_model);
+            }
+        }
+    });
```
Erigara, line 250: Instead of spawning threads we can use async `listen_for_events_async` and `rt.block_on`.
But since this is a test i don't think it's critical here.

core/src/smartcontracts/wasm.rs
```diff
@@ -1657,14 +1654,15 @@ impl<'wrld, S: StateReadOnly> RuntimeBuilder<state::executor::ValidateQuery<'wrl
...
-                export::SET_PERMISSION_TOKEN_SCHEMA => |caller: ::wasmtime::Caller<state::executor::ValidateQuery<'_, S>>, offset, len| Runtime::set_permission_token_schema(caller, offset, len),
+                export::SET_DATA_MODEL => |caller: ::wasmtime::Caller<state::executor::ValidateQuery<'_, S>>, offset, len| Runtime::set_data_model(caller, offset, len),
             )?;
             Ok(linker)
         })
     }
 }
 
 impl<'wrld, 'block, 'state> RuntimeBuilder<state::executor::Migrate<'wrld, 'block, 'state>> {
+    // FIXME: is this accurate doc?
```
Erigara, line 1665: Doesn't look like accurate doc...

data_model/src/events/data/filters.rs
```diff
@@ -226,8 +223,8 @@ mod model {
         IntoSchema,
     )]
     pub struct ConfigurationEventFilter {
-        /// If specified matches only events originating from this configuration
-        pub(super) id_matcher: Option<super::ParameterId>,
+        // /// If specified matches only events originating from this configuration
+        // pub(super) id_matcher: Option<super::ParameterId>,
```
Erigara, line 227: Should be removed?

client/tests/integration/asset.rs
0x009922: Opened #4635, will rebase on top of it

data_model/src/permission.rs
```diff
@@ -13,200 +14,116 @@ pub use self::model::*;
...
-    pub fn new(token_ids: Vec<PermissionTokenId>, schema: String) -> Self {
-        Self { token_ids, schema }
-    }
-
-    /// Return id of this token
-    pub fn token_ids(&self) -> &[PermissionTokenId] {
-        &self.token_ids
+        /// It is JSON-encoded, and its structure must correspond to the structure of
+        /// the type defined in [`crate::executor::ExecutorDataModel`].
+        #[getset(skip)] // TODO: derive with getset once FFI impl is fixed
```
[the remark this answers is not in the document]
0x009922: There is a getter already, maybe I should move this comment to it instead.

data_model/src/events/data/filters.rs
0x009922: Yes, thanks.

client/tests/integration/upgrade.rs
```diff
@@ -225,7 +225,45 @@ fn migration_fail_should_not_cause_any_effects() {
...
+        &client,
+        "tests/integration/smartcontracts/executor_with_custom_permission",
+    )
+    .unwrap();
+
+    let data_model = receiver
+        .recv_timeout(Duration::from_secs(60))
+        .expect("should receive upgraded event immediately after upgrade");
+
+    assert_eq!(data_model.permissions.len(), 40);
```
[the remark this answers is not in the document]
0x009922: Not _flaky_, but rather refactoring-sensitive, yeah.

Changed the check to just ensure it isn't empty. I think this check isn't really necessary at all.

client/tests/integration/upgrade.rs
0x009922: Thanks for notice, refactored with `async`.

data_model/src/lib.rs
0x009922: Done!

0x009922:
#4635 should be merged first. This PR is currently rebased on top of it.

core/src/smartcontracts/wasm.rs
0x009922: I would like to leave this FIXME comment as is... or maybe create an issue?

data_model/src/permission.rs
```diff
@@ -17,196 +18,113 @@ pub type Permissions = BTreeSet<Permission>;
...
 }
 
-impl PartialOrd for JsonString {
-    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
-        Some(self.cmp(other))
+impl From<PermissionId> for Name {
+    fn from(value: PermissionId) -> Self {
+        value.name
     }
 }
```
mversic, line 117: are these impls used somewhere? We don't usually assume that id can be converted to name or str

data_model/src/query/mod.rs
mversic, line 992: I feel we have a small discrepancy here. On the one hand we have `FindAllParameters` but we don't have `FindAllPermissions` (I assume this would be for some account)?

data_model/src/lib.rs
```diff
@@ -864,6 +583,124 @@ mod model {
...
+        let json = serde_json::Value::from_str(&self.0).map_err(serde::ser::Error::custom)?;
+        json.serialize(serializer)
+    }
+}
+
+impl iroha_schema::TypeId for JsonString {
+    fn id() -> iroha_schema::Ident {
+        "JsonString".to_owned()
+    }
+}
```
mversic, line 680: you can derive this

client/tests/integration/parameters.rs
```diff
@@ -0,0 +1,83 @@
...
+    let _err = client
+        .submit_blocking(SetParameter::new(Parameter::new(
+            "WrongNonExistingParameter".parse().unwrap(),
+            json!({ "prefix": "whatever" }),
+        )))
+        .expect_err("should still not work");
+
+    assert_eq!(
+        client
+            .request(client::parameter::all())?
```
mversic, line 65: this query should return core parameters as well

client/tests/integration/smartcontracts/executor_with_custom_parameter/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    execute!(executor, isi);
+}
+
+fn visit_register_domain(executor: &mut Executor, _authority: &AccountId, isi: &Register<Domain>) {
+    // FIXME: unwrap is ok here?
+    let required_prefix = FindAllParameters
+        .execute()
+        .unwrap()
+        .into_iter()
+        .map(Result::unwrap)
```
mversic, line 38: yes, because it would indicate a bug in iroha. It should be considered impossible to get Err here

client/tests/integration/smartcontracts/executor_remove_permission/src/lib.rs
```diff
@@ -25,13 +26,11 @@ struct Executor {
 pub fn migrate(_block_height: u64) -> MigrationResult {
     // Note that actually migration will reset token schema to default (minus `CanUnregisterDomain`)
     // So any added custom permission tokens will be also removed
-    let mut schema = default_permission_schema();
-    schema.remove::<iroha_executor::default::tokens::domain::CanUnregisterDomain>();
+    let mut data_model = DataModelBuilder::new();
+    data_model.extend_with_default_permissions();
```
mversic, line 30:
```suggestion
    let mut data_model = DataModelBuilder::with_default_permissions();
```

mversic (changes requested):

client/tests/integration/smartcontracts/executor_with_custom_permission/src/lib.rs
```diff
@@ -199,14 +201,13 @@ fn visit_unregister_domain(
...
+    data_model.extend_with_default_permissions();
+    data_model
+        .remove_permission::<iroha_executor::default::permissions::domain::CanUnregisterDomain>();
+    data_model.add_permission::<token::CanControlDomainLives>();
 
-    let (token_ids, schema_str) = schema.serialize();
-    iroha_executor::set_permission_schema(
-        &iroha_executor::data_model::permission::PermissionSchema::new(token_ids, schema_str),
-    );
+    iroha_executor::set_data_model(&data_model.serialize());
```
mversic, line 210:
```suggestion
    iroha_executor::set_data_model(&data_model);
```

can't we have api like this?

client_cli/src/main.rs
```diff
@@ -608,21 +608,22 @@ mod account {
...
 
         fn from_str(s: &str) -> Result<Self> {
             let content = fs::read_to_string(s)
                 .wrap_err(format!("Failed to read the permission token file {}", &s))?;
-            let permission: DataModelPermission = json5::from_str(&content).wrap_err(format!(
-                "Failed to deserialize the permission token from file {}",
-                &s
-            ))?;
-            Ok(Self(permission))
+            let permission_token: PermissionObject =
```
mversic, line 621:
```suggestion
            let permission: PermissionObject =
```

core/src/smartcontracts/isi/world.rs
```diff
@@ -407,33 +382,47 @@ pub mod isi {
...
+            .world
+            .parameters
+            .get_mut()
+            .retain(|parameter| {
+                state_transaction
+                    .world
+                    .executor_data_model
+                    .parameters
+                    .contains(&parameter.id)
+            });
```
mversic, line 396: this looks as if it also removes core parameters?

core/src/smartcontracts/isi/world.rs
```diff
@@ -299,41 +298,14 @@ pub mod isi {
...
 
-            let world = &mut state_transaction.world;
-            if !world.parameters.insert(parameter) {
-                return Err(RepetitionError {
-                    instruction_type: InstructionType::NewParameter,
-                    id: IdBox::ParameterId(parameter_id),
-                }
-                .into());
+            if !world.executor_data_model.parameters.contains(&parameter.id) {
+                return Err(FindError::Parameter(parameter.id).into());
```
mversic, line 304: what if the parameter being set is a core parameter?

default_executor/src/lib.rs
```diff
@@ -52,11 +52,9 @@ impl Executor {
 pub fn migrate(block_height: u64) -> MigrationResult {
     Executor::ensure_genesis(block_height)?;
 
-    let schema = default_permission_schema();
-    let (token_ids, schema_str) = schema.serialize();
-    iroha_executor::set_permission_schema(
-        &iroha_executor::data_model::permission::PermissionSchema::new(token_ids, schema_str),
-    );
+    let mut data_model = DataModelBuilder::new();
+    data_model.extend_with_default_permissions();
```
mversic, line 56:
```suggestion
    let data_model = DataModelBuilder::with_default_permissions();
```

smart_contract/executor/derive/src/permission.rs
```diff
@@ -0,0 +1,35 @@
...
+        impl #impl_generics ::iroha_executor::permission::Permission for #ident #ty_generics #where_clause {
+            fn is_owned_by(&self, account_id: &::iroha_executor::data_model::account::AccountId) -> bool {
+                let account_tokens_cursor = ::iroha_executor::smart_contract::debug::DebugExpectExt::dbg_expect(
+                    ::iroha_executor::smart_contract::ExecuteQueryOnHost::execute(
+                        &::iroha_executor::data_model::query::permission::FindPermissionsByAccountId::new(
+                            account_id.clone(),
+                        )
+                    ),
+                    "Failed to execute `FindPermissionsByAccountId` query"
+                );
```
mversic, line 22: We knows for a fact that this query can't fail unless there is a bug in Iroha. You can just `unwrap` here to reduce the bloat

smart_contract/executor/src/default/permissions.rs
```diff
@@ -43,23 +48,23 @@ macro_rules! declare_tokens {
         }
 
         impl TryFrom<&$crate::data_model::permission::Permission> for AnyPermission {
-            type Error = $crate::permission::PermissionConversionError;
+            type Error = $crate::ConvertDataModelObjectError;
 
             fn try_from(token: &$crate::data_model::permission::Permission) -> Result<Self, Self::Error> {
-                match token.definition_id().as_ref() { $(
+                match token.id().borrow() { $(
```
mversic, line 54:
```suggestion
                match token.id().as_ref() { $(
```

you shouldn't use `borrow` manually. It is just for use in `HashMap`s

smart_contract/src/lib.rs
```diff
@@ -264,6 +264,58 @@ where
     }
 }
 
+/// Defines an object that is part of the [`ExecutorDataModel`].
+///
+/// An object consists of id and payload.
+/// ID is a unique type identifier in the executor's data model.
+/// Payload is JSON-serialized type structure, according to the type schema.
+// TODO: support both JSON and SCALE for payload.
+pub trait ExecutorDataModelObject: Sized {
```
mversic, line 273: I don't think there is much point in having this trait just for 2 types. I'd rather if we had 2 traits - `Permission` and `Parameter`

smart_contract/executor/src/parameter.rs
```diff
@@ -0,0 +1,64 @@
...
+/// #[derive(IntoSchema, Serialize, Deserialize)]
+/// struct DomainPrefix {
+///     prefix: String,
+/// }
+///
+/// impl Parameter for DomainPrefix {}
+/// ```
+pub trait Parameter: Serialize + DeserializeOwned + IntoSchema {
+    /// Parameter definition id, according to [`IntoSchema`].
+    fn definition_id() -> ParameterId {
```
mversic, line 29:
```suggestion
    fn id() -> ParameterId {
```

smart_contract/executor/src/permission.rs
```diff
@@ -3,27 +3,51 @@
...
-    /// Return name of this permission token
-    fn name() -> Name {
-        <Self as iroha_schema::IntoSchema>::type_name()
-            .parse()
-            .dbg_expect("Failed to parse permission token as `Name`")
+    /// Check if the account owns this token
+    fn is_owned_by(&self, account_id: &AccountId) -> bool;
+
+    /// Token's id, according to [`IntoSchema`].
+    fn definition_id() -> PermissionId {
```
mversic, line 20:
```suggestion
    fn id() -> PermissionId {
```

smart_contract/executor/src/lib.rs
```diff
@@ -172,34 +182,106 @@ macro_rules! deny {
...
+#[derive(Debug)]
+pub enum ConvertDataModelObjectError {
+    /// Unexpected object id
+    Id(data_model::prelude::Name),
+    /// Failed to deserialize object payload
+    Deserialize(serde_json::Error),
+}
+
+/// A convenience to build [`ExecutorDataModel`] from within the executor
+#[derive(Debug, Clone, Default)]
```
mversic, line 234:
```suggestion
#[derive(Debug, Clone)]
```

I'd keep it without `Default` so as not to confuse with `with_default_permissions`

mversic (changes requested):

client/tests/integration/permissions.rs
```diff
@@ -315,10 +315,9 @@ fn stored_vs_granted_token_payload() -> Result<()> {
     // Allow alice to mint mouse asset and mint initial value
     let mouse_asset = AssetId::new(asset_definition_id, mouse_id.clone());
     let allow_alice_to_set_key_value_in_mouse_asset = Grant::permission(
-        Permission::from_str_unchecked(
+        Permission::new(
             "CanSetKeyValueInUserAsset".parse().unwrap(),
-            // NOTE: Introduced additional whitespaces in the serialized form
-            &*format!(r###"{{ "asset_id" : "xor#wonderland#{mouse_id}" }}"###),
+            json!({ "asset_id": format!("xor#wonderland#{mouse_id}") }),
```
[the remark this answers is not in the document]
mversic: yes, you can define `serde_json::Value` directly from a string

smart_contract/executor/src/default/permissions.rs
```diff
@@ -85,63 +90,63 @@ macro_rules! declare_tokens {
 macro_rules! token {
```
dima74, line 90: Consider also renaming `token!` and `declare_tokens!` macros in this file

tools/kagami/Cargo.toml
```diff
@@ -13,7 +13,7 @@ license.workspace = true
 workspace = true
 
 [dependencies]
-iroha_crypto = { workspace = true }
+iroha_crypto = { workspace = true, features = ["rand"] }
```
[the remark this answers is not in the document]
DCNick3: `iroha_crypto` is used in workspace without `default-features` (because it also gets used in wasm crates)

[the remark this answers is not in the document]
DCNick3: Oh, it's actually not a bug, `default-features = true` in workspace crate makes the default features always enabled. The fix to that issue was to add a warning when you try to disable default features on a workspace dependency..

Bottom line: either the `rand` feature or default features as a whole must be enabled explicitly when using workspace dependencies like this (at least the ones which have default features disabled in the workspace). I think it's fine to have only `rand` enabled, though putting `default-features = true` here would also work

data_model/src/query/mod.rs
0x009922: > (I assume this would be for some account)

Yes, currently there is `FindPermissionsByAccountId`, and that is the only permission-related query. There are also roles-related though.

data_model/src/lib.rs
```diff
@@ -864,6 +583,124 @@ mod model {
...
+    /// This string is guaranteed to be parsed as JSON.
+    #[derive(Display, Default, Debug, Clone, Eq, Encode, Decode)]
+    #[ffi_type(unsafe {robust})]
+    #[repr(transparent)]
+    #[display(fmt = "{}", "0")]
+    pub struct JsonString(pub(super) String);
+}
+
+/// A helper trait for polymorphism, implemented for various types
+pub trait IntoJsonString {
```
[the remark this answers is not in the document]
0x009922: It is not necessary.

It allows all of that:

```rust
let id = "SomeId".parse().unwrap();

Permission::new(id, JsonString::from(&json!(null)));
Permission::new(id, &json!(null));
Permission::new(id, json!(null));
```

I find `iroha_data_model` to lack polymorphism in constructors a lot, actually.

client/tests/integration/parameters.rs
0x009922: This PR removes "core parameters" entirely from Iroha, temporarily. `Parameter` is now related only to executor-defined ones.

After merging of this PR, I will implement core parameters properly, according to https://github.com/hyperledger/iroha/issues/4028#issuecomment-2099618024.

For now chain-wide parameters are only configured via config file.

client/tests/integration/smartcontracts/executor_with_custom_permission/src/lib.rs
0x009922: Well, we can, but that would be an unnecessary opinionated API IMO, because `set_data_model` really needs only a borrowed `ExecutorDataModel`. It isn't exactly necessary for developers to rely on the builder, they might want to pass their own `ExecutorDataModel`.

However, this might be better?

```rust
data_model.set(); // invokes `iroha_executor::set_data_model` inside
```

0x009922: Revised builder API, here is before vs after:

```rust
let mut data_model = DataModelBuilder::new();
data_model.extend_with_default_permissions();
data_model
    .remove_permission::<iroha_executor::default::permissions::domain::CanUnregisterDomain>();
data_model.add_permission::<token::CanControlDomainLives>();

iroha_executor::set_data_model(&data_model.serialize());
```

```rust
DataModelBuilder::with_default_permissions()
    .remove_permission::<iroha_executor::default::permissions::domain::CanUnregisterDomain>()
    .add_permission::<token::CanControlDomainLives>()
    .set();

```

core/src/smartcontracts/isi/world.rs
0x009922: Replied in other thread (in sum: there are no core parameters, _yet_)

core/src/smartcontracts/isi/world.rs
0x009922: Replied in other thread.

smart_contract/executor/derive/src/permission.rs
0x009922: Replaced with plain `expect`.

smart_contract/src/lib.rs
0x009922: Well, yeah, it seems to bring more boilerplate than benefit.

I wanted to formalize that "data model objects are those which have id and payload" and base impls on that. But, it turns out I can remove a lot and just leave that duplicated fragment for both `Parameter` and `Permission` traits in `iroha_executor`:

```rust
/// Try to convert from [`ParameterObject`]
/// # Errors
/// See [`TryFromDataModelObjectError`]
fn try_from_object(
    object: &ParameterObject,
) -> crate::prelude::Result<Self, TryFromDataModelObjectError> {
    if &object.id() != Self::id() {
        return Err(TryFromDataModelObjectError::Id(object.id().name().clone()));
    }
    object
        .payload()
        .deserialize()
        .map_err(TryFromDataModelObjectError::Deserialize)
}

/// Convert into [`ParameterObject`]
fn to_object(&self) -> ParameterObject {
    ParameterObject::new(
        Self::id(),
        JsonString::serialize(&self)
            .expect("failed to serialize concrete data model entity; this is a bug"),
    )
}

```

client/tests/integration/extra_functional/unregister_peer.rs
```diff
@@ -114,24 +112,25 @@ fn init() -> Result<(
...
-        .add_parameter(MAX_TRANSACTIONS_IN_BLOCK, 1u32)?
-        .into_set_parameters();
     let create_domain = Register::domain(Domain::new("domain".parse()?));
     let (account_id, _account_keypair) = gen_account_in("domain");
     let create_account = Register::account(Account::new(account_id.clone()));
     let asset_definition_id: AssetDefinitionId = "xor#domain".parse()?;
     let create_asset =
         Register::asset_definition(AssetDefinition::numeric(asset_definition_id.clone()));
-    let instructions = parameters.into_iter().chain([
+    let isi: [InstructionBox; 3] = [
```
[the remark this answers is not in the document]
0x009922: rustc bug?

```
error: in expressions, `_` can only be used on the left-hand side of an assignment
   --> client/tests/integration/extra_functional/unregister_peer.rs:128:31
    |
128 |     let isi: [InstructionBox; _] = [
    |                               ^ `_` not allowed here
``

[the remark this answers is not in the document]
Comment (0x009922) [reply client/tests/integration/extra_functional/unregister_peer.rs:128]: Ah, it's just unstable: https://github.com/rust-lang/rust/issues/85077

File: smart_contract/executor/src/default/permissions.rs (hunk 32 above)
Comment (0x009922) [reply smart_contract/executor/src/default/permissions.rs:54]: Replaced with `token.id().name().as_ref()`

File: tools/kagami/Cargo.toml (hunk 39 above)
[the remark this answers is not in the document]
Comment (0x009922) [reply tools/kagami/Cargo.toml:16]: > `rand` is in default-features of `iroha-crypto`, looks like no need to add it here

Doesn't compile with `cargo check -p kagami`

> I think it's fine to have only `rand` enabled, though putting `default-features = true` here would also work

Not sure what is better, leaving as-is.

File: smart_contract/executor/src/default/permissions.rs (hunk 38 above)
Comment (0x009922) [reply smart_contract/executor/src/default/permissions.rs:90]: Done

File: data_model/src/lib.rs [hunk 42]
```diff
@@ -864,6 +583,124 @@ mod model {
...
+impl PartialOrd for JsonString {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for JsonString {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.0.cmp(&other.0)
+    }
```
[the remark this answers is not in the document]
0x009922: Did it.

data_model/src/permission.rs
0x009922: Cleaned these.

client/tests/integration/smartcontracts/executor_with_custom_parameter/src/lib.rs
0x009922: Got it, replaced with explanatory `.expect("...")`ations

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
0x009922: I noticed that whitespaces were introduced in order to check that that payloads are compared by `serde_json::Value` equality rather than by JSON str comparison.

But... in this particular test there is no reason to do that. This test registers a single permission and doesn't check any comparisons.

There is another test below which actually grants the same permission with technically the same content, although they even differ in JSON:

```rust
let allow_alice_to_transfer_rose_1 = Grant::permission(
    Permission::new(
        "CanTransferUserAsset".parse().unwrap(),
        json!({ "asset_id": format!("rose#wonderland#{alice_id}") }),
    ),
    alice_id.clone(),
);

let allow_alice_to_transfer_rose_2 = Grant::permission(
    Permission::new(
        "CanTransferUserAsset".parse().unwrap(),
        // different content, but same meaning
        json!({ "asset_id": format!("rose##{alice_id}") }),
    ),
    alice_id,
);
```

1. They have different JSON payloads! Therefore, equality by `serde_json::Value` would fail to detect duplication.
2. Semantically, they have the same payloads, because it is still `AssetId`, but written in a different form.
3. And indeed, the second instruction fails as duplicate! Therefore, the comparison is made in executor-native types, not JSON.

I am _not sure_ why it works, but this test passes.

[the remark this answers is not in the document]
mversic: previously tokens were compared as strings, not as JSON objects. `stored_vs_granted` test was introduce specifically to ensure that we compare it as JSON objects, so it's enough that you keep it only in there. Check #3853

core/src/smartcontracts/wasm.rs
Erigara: I think you can leave it as is, not to stretch scope of PR.

smart_contract/executor/src/permission.rs
```diff
@@ -3,27 +3,49 @@
...
+    }
+
+    /// Convert into [`PermissionObject`]
+    fn to_object(&self) -> PermissionObject {
+        PermissionObject::new(
+            <Self as Permission>::id(),
+            JsonString::serialize(&self)
+                .expect("failed to serialize concrete data model entity; this is a bug"),
+        )
+    }
```
mversic, line 48: what was wrong with having implemented `TryFrom` and `From`?

smart_contract/executor/src/lib.rs
```diff
@@ -78,7 +79,11 @@ pub fn get_migrate_payload() -> payloads::Migrate {
     unsafe { decode_with_length_prefix_from_raw(host::get_migrate_payload()) }
 }
 
-/// Set new [`PermissionSchema`].
+// pub trait IntoDataModel {
+//     fn into_data_model(&self) -> &ExecutorDataModel;
+// }
```
mversic, line 84: can be removed?

smart_contract/executor/derive/src/default.rs
```diff
@@ -142,21 +142,20 @@ pub fn impl_derive_visit(emitter: &mut Emitter, input: &syn::DeriveInput) -> Tok
         "fn visit_transfer_asset_definition(operation: &Transfer<Account, AssetDefinitionId, Account>)",
         "fn visit_set_asset_definition_key_value(operation: &SetKeyValue<AssetDefinition>)",
         "fn visit_remove_asset_definition_key_value(operation: &RemoveKeyValue<AssetDefinition>)",
-        "fn visit_grant_account_permission(operation: &Grant<Permission, Account>)",
-        "fn visit_revoke_account_permission(operation: &Revoke<Permission, Account>)",
+        "fn visit_grant_account_permission(operation: &Grant<PermissionObject, Account>)",
```
mversic, line 145:
```suggestion
        "fn visit_grant_account_permission(operation: &Grant<Permission, Account>)",
```

I would prefer if we kept the naming. I know it collides with the trait name, but the 2 will never intersect in user code

smart_contract/executor/derive/src/entrypoint.rs
```diff
@@ -144,6 +144,7 @@ fn impl_migrate_entrypoint(fn_item: syn::ItemFn) -> TokenStream {
 
     let migrate_fn_name = syn::Ident::new(export::EXECUTOR_MIGRATE, proc_macro2::Span::call_site());
 
+    // FIXME: is doc accurate?
```
mversic, line 147: which doc?

smart_contract/executor/src/lib.rs
```diff
@@ -222,16 +288,22 @@ pub mod prelude {
     pub use alloc::vec::Vec;
 
     pub use iroha_executor_derive::{
-        entrypoint, Constructor, Token, Validate, ValidateEntrypoints, ValidateGrantRevoke, Visit,
+        entrypoint, Constructor, Permission, Validate, ValidateEntrypoints, ValidateGrantRevoke,
+        Visit,
+    };
+    pub use iroha_smart_contract::prelude::{
+        Parameter as ParameterObject, Permission as PermissionObject, *,
```
mversic, line 295: I'm not a fan of this renaming to `*Object`. If we have to do this I'd prefer we do it for the trait

smart_contract/executor/src/permission.rs
0x009922: 🤔 good question

client/tests/integration/permissions.rs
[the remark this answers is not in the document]
0x009922: Got it, returned JSON with extra whitespaces

smart_contract/executor/src/permission.rs
0x009922: Ah, yes. Because:

```rust
impl<T: Parameter> TryFrom<&ParameterObject> for T {
//                 ^^^^^^^^^^^^^^^^^^^^^^^^^
//                 Only traits defined in the current crate can be implemented for arbitrary types
    type Error = TryFromDataModelObjectError;

    fn try_from(value: &ParameterObject) -> core::result::Result<Self, Self::Error> {
        todo!()
    }
}
```

smart_contract/executor/src/lib.rs
0x009922: Done.

smart_contract/executor/src/permission.rs
mversic: you can't make a blanket implementation, but you can derive for each type on `#[derive(PermissionTrait)]`

0x009922: Then we also need to introduce a derive macro for `ParameterTrait`, and I wouldn't want it just for the sake of this conversion.

0x009922:
Drafted due to some questions related to chain-wide config design. More on that commented here: https://github.com/hyperledger/iroha/issues/4028#issuecomment-2133096935

0x009922:
Close in favor of #4658

PR #4611, opened 2024-05-17
fix(kura): Inline kura initialization into constructor

### LINKED ISSUE #4603: Kura panics after restart
After restarting peer kura fails with a message: "The block to be written cannot be None, see store_block function."

Looks like it happens because `Kura::start` is called before `Kura::init`.
It was the opposite before.

Erigara:
## Description

Due `init` was separate functions error was introduced where `start` was called before initialization which cause panic in kura.

Fix for that is to inline `init` inside constructor.

### Linked issue

Closes #4603 

### How to reproduce

1. start iroha
2. submit few blocks
3. restart iroha

Previously kura would crash, but iroha would continue working without ability to write new blocks to the disk.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -255,11 +255,9 @@ impl Iroha<ToriiNotStarted> {
                 .into_non_empty_vec(),
         );
 
-        let kura = Kura::new(&config.kura).change_context(StartError::InitKura)?;
+        let (kura, block_count) = Kura::new(&config.kura).change_context(StartError::InitKura)?;
         let kura_thread_handler = Kura::start(Arc::clone(&kura));
         let live_query_store_handle = LiveQueryStore::from_config(config.live_query_store).start();
-
-        let block_count = kura.init().change_context(StartError::InitKura)?;
 
         let state = match try_read_snapshot(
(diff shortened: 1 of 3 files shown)


cli/src/lib.rs
```diff
@@ -215,12 +215,10 @@ impl Iroha {
             config.sumeragi.trusted_peers.clone(),
         );
 
-        let kura = Kura::new(&config.kura)?;
+        let (kura, block_count) = Kura::new(&config.kura)?;
         let kura_thread_handler = Kura::start(Arc::clone(&kura));
```
mversic, line 219: can this also be inlined?

mversic (dismissed):

cli/src/lib.rs
Erigara: yeah, but my thinking was that for testing purposes we don't need starting write to disk loop

VAmuzing (dismissed):

mversic (approved):

VAmuzing (approved):

Erigara:
There is failling test `executor_upgrade_should_run_migration` with error introduced by 9d91c77c1 commit. 
Going to merge as is and fix issue later since already have approvals.

PR #4616, opened 2024-05-19
[Fix] #4586: Added register trigger permissions check

### LINKED ISSUE #4586: Check authority on trigger registration
# TL;DR

Atm in the default executor permission for trigger is not checked. 

This creates security loophole allowing malicious user to execute any instruction on behalves of another user (e.g. stealing assets).

Instead we should only allow register trigger if: 
1. Account trying register trigger is the same as trigger authority
2. Or account is the the owned of the account (domain owner)
3. Or account has `CanRegisteretUserTrigger` permission token 

# Minimum working example

1. Run iroha with default genesis and executor 
```bash
./scripts/test_env.py setup
```
2. Submit the following transaction to the iroha on behalf of alice
```bash
cd test
cat trigger.json | ./iroha_client_cli --config client.toml json 
```

```json
[
   {
      "Register" : {
         "Trigger" : {
            "action" : {
               "authority" : "carpenter@garden_of_live_flowers",
               "executable" : {
                  "Instructions" : [
                     {
                        "Mint" : {
                           "Asset" : {
                              "destination_id" : "cabbage#garden_of_live_flowers#carpenter@garden_of_live_flowers",
                              "object" : 1
                           }
                        }
                     }
                  ]
               },
               "filter" : {
                  "Time" : "PreCommit"
               },
               "metadata" : {},
               "repeats" : "Indefinitely"
            },
            "id" : "mint_every_block"
         }
      }
   }
]
```

5. Check that `cabbage` is minted every block for `carpenter`.

It is expected that such transaction would be rejected.

Stukalov-A-M:
### Linked issue

* Closes #4586 

### Checklist


Stukalov-A-M:
@mversic @Erigara 
Proposal:
I think we should remove connection of a trigger with a domain_id. It doesn't make sense because each trigger's authority already has this connection. Moreover, it creates some redundant difficulties in cases of permission granting and checking ref `is_trigger_owner()`

client/tests/integration/domain_owner_permissions.rs
```diff
@@ -325,7 +325,7 @@ fn domain_owner_trigger_permissions() -> Result<()> {
     let bob_id: AccountId = "bob@wonderland".parse()?;
     let token = PermissionToken::new(
         "CanUnregisterUserTrigger".parse().unwrap(),
-        &json!({ "trigger_id": trigger_id }),
+        &json!({ "account_id": bob_id }),
```
mversic, line 328: now this gives a permission to remove any trigger, instead of only one

Erigara:
> I think we should remove connection of a trigger with a domain_id. It doesn't make sense because each trigger's authority already has this connection. Moreover, it creates some redundant difficulties in cases of permission granting and checking ref is_trigger_owner()

Sounds reasonable.

client/tests/integration/domain_owner_permissions.rs
Erigara: Yeah, we discussed this with @Stukalov-A-M and decided that it make sense to feel `CanRegister/CanUnregister` permissions more uniformly. 

Like there is not much value in token `CanRegisterUserTrigger` which accepts trigger id.

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -266,10 +268,83 @@ fn trigger_should_be_able_to_modify_its_own_repeats_count() -> Result<()> {
...
+    let found_trigger = test_client.request(find_trigger)?;
+
+    assert_eq!(found_trigger.id, trigger_id);
+
+    Ok(())
+}
+
+#[test]
+fn unregister_trigger() -> Result<()> {
+    let (_rt, _peer, test_client) = <PeerBuilder>::new().with_port(10_044).start_with_runtime();
```
Erigara, line 345: We pick every 5th port (10_040, 10_045, ...). You have to find free one.

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -266,10 +268,83 @@ fn trigger_should_be_able_to_modify_its_own_repeats_count() -> Result<()> {
...
+    );
+
+    // Register rabbit's account
+    test_client.submit_blocking(Register::account(rabbit_account))?;
+
+    let find_rabbit = FindAccountById {
+        id: rabbit_account_id.clone(),
+    };
+    test_client.request(find_rabbit).expect("Account not found");
+    info!("Rabbit is found.");
```
Erigara, line 313: This check is redundant imo, we check account registration ability in other test.

smart_contract/executor/src/permission.rs
```diff
@@ -212,31 +212,40 @@ pub mod account {
...
-        if trigger.action().authority() == authority {
+        let trigger = find_trigger(trigger_id)?;
+
+        if trigger.action().authority() == authority
+            || is_domain_owner(trigger.action().authority().domain_id(), authority)?
+            || match trigger_id.domain_id() {
+                Some(domain) => is_domain_owner(domain, authority)?,
+                None => false,
+            }
+        {
```
Erigara, line 236: If look unnecessary here because we essentially do here 

```rust
if true {
    return Ok(true);
} else {
    return Ok(false);
}
``` 

Instead we can use predicate directly:

```suggestion
        Ok(
            trigger.action().authority() == authority
            || is_domain_owner(trigger.action().authority().domain_id(), authority)?
            || match trigger_id.domain_id() {
                Some(domain) => is_domain_owner(domain, authority)?,
                None => false,
            }
        )
```

client/tests/integration/triggers/by_call_trigger.rs
Stukalov-A-M: This check was created only for debugging convenience and helped me a lot. So I would like to keep it if you don't mind.

smart_contract/executor/src/permission.rs
Stukalov-A-M: Good advice, thanks!

client/tests/integration/triggers/by_call_trigger.rs
Erigara: I don’t insist on removing this check, feel free to resolve discussion.

Stukalov-A-M:
> > I think we should remove connection of a trigger with a domain_id. It doesn't make sense because each trigger's authority already has this connection. Moreover, it creates some redundant difficulties in cases of permission granting and checking ref is_trigger_owner()
> 
> Sounds reasonable.

The ticket is created
https://github.com/hyperledger/iroha/issues/4630#issue-2307500287

Erigara (approved):

PR #4636, opened 2024-05-22
feat: add get_original_transaction_by_hash

enxtur:
## Description

Added a method to retrieve a pending partially signed transaction by its transaction hash from the queue. Previously, existing methods required the entire transaction payload, which was cumbersome for other users who needed to build the entire transaction for signing. Now, only the transaction hash is needed.

### Benefits

This improvement simplifies the process for users who need to sign transactions, as they no longer need to reconstruct the entire transaction payload. Instead, they can simply use the transaction hash. This change aligns more closely with real-world use cases, making the signing process more practical and efficient.

### Checklist


mversic:
can you tell us more on how do you use `PENDING_TRANSACTIONS`?

mversic:
> > can you tell us more on how do you use `PENDING_TRANSACTIONS`?
> 
> Well, simply for fetching those encoded pending transactions, if I understand correctly. In my personal experience, I was able to fetch pending transactions using API directly and filtered with hash. So I thought this new method will make life easier.

ok. But what do you do with pending transactions after you fetch them? I'm just trying to understand your use-case, not related to this PR

enxtur:
> > > can you tell us more on how do you use `PENDING_TRANSACTIONS`?
> > 
> > 
> > Well, simply for fetching those encoded pending transactions, if I understand correctly. In my personal experience, I was able to fetch pending transactions using API directly and filtered with hash. So I thought this new method will make life easier.
> 
> ok. But what do you do with pending transactions after you fetch them? I'm just trying to understand your use-case, not related to this PR

I would like another account sign it.

mversic (approved):

Issue #4569, opened 2024-05-08
Restrict types of predicates applied to queries depending on the query output type (type-safe queries)


DCNick3:
To implement #3671 we would need to add more predicate types like account predicates or asset predicates.

However our current API allows the clients to specify any predicate filter, without any regard to whether the predicate type would be compatible or not.

In the client-facing API we should restrict the types of predicates allowed for a less error-prone API

PR #4638, opened 2024-05-22
feat!: Add traits for iterable and non-iterable queries

DCNick3:
## Description

This PR is a first step towards more type-safe queries (#4569). It adds traits for singular and iterable queries and prohibits supplying filters, sorting, pagination and batching parameters in compile time (it is already a runtime error to do so).

### Linked issue

Partially addresses #4569

### Benefits

- less error prone API
- lays foundation to allow type-checking query filters

### Checklist

- [ ] make CI pass


0x009922 (approved):
LGTM.

I noticed that `smart_contract` and `client` duplicate impls of query execution, and both have type logic to guard against incorrect use of queries. But this compile-time guard is tested only for client. Would it make sense to make a ui test for smartcontracts as well?

Also, do I understand correctly that the runtime check you are talking about is this code?

https://github.com/hyperledger/iroha/blob/09995afd357d74617d34500a5535f8ca2f5d9de7/core/src/smartcontracts/isi/query.rs#L52-L63

DCNick3:
> I noticed that smart_contract and client duplicate impls of query execution, and both have type logic to guard against incorrect use of queries. But this compile-time guard is tested only for client. Would it make sense to make a ui test for smartcontracts as well?

I thought about that too, though I just got lazy doing that :sweat_smile:. I guess I'll update the PR to add that

> Also, do I understand correctly that the runtime check you are talking about is this code?

Yes, this is the check I added in #4544 as per [your suggestion](https://github.com/hyperledger/iroha/pull/4544#discussion_r1596110415) =)

client/tests/ui_fail/cant_filter_singular_query.stderr
Erigara, line 1: I wonder would `#[diagnostic]` be helpful here to suggest user that this query is not iterable

smart_contract/Cargo.toml
```diff
@@ -32,3 +32,4 @@ webassembly-test = "0.1.0"
 # Not used directly but required for compilation
 getrandom = { version = "0.2", features = ["custom"] }
 
+trybuild = { workspace = true }
```
Erigara, line 35: New line is missing

Erigara:
Other than comments LGTM

client/tests/ui_fail/cant_filter_singular_query.stderr
DCNick3: Tried that, but it seems it works only for [`E0277`](https://doc.rust-lang.org/error_codes/E0277.html) (passing a type that doesn't implement a trait), not for [E0599](https://doc.rust-lang.org/error_codes/E0599.html) that is raised here (using a method which has unsatisfied trait bounds).

There might be some chance I did something wrong, but simply adding `#[diagnostic::on_unimplemented(...)]` didn't change the diagnostic at all

Erigara (approved):

0x009922 (approved):

Issue #3671, opened 2023-07-03
[suggestion] Remove redundant non-iterable queries


mversic:
### Feature request

Our `QueryBox` contains some superfluous queries which could be composed by using a more general query and a filter. For example, there is both `FindAllAccounts` and `FindAccountByName` where you could have combined former and filter it by name.

In general we should remove all queries that return a single result and fallback to using more general queries with filter. Iroha query would correspond to `SELECT` statement in the relational database and filter would correspond to `WHERE` clause

### Motivation

It's simpler and removes redundancy

appetrosyan:
To add some clarification as per our discussion with @DCNick3. 

For now, we only want to reduce the surface area of the API and reduce the number of queries. In this example, we want to replace `FindAccountByName` with a `FindAllAccounts` query, combined with a `StringPredicate::is(name_of_account)` to get the same result. 

Due to part of the context only being known at runtime, the latter operation is likely going to result in an $O(n)$ lookup, in contrast to the $O(1)$ for `FindAccountByName`. We can **ooptionally** counteract this problem by doing a short-circuiting heuristic. 

So I'd do things in three steps.

### Step 1: Enumerate replacement queries

This is just adding a new file to `client/tests` that runs through the Rust API queries and compares the results on a synthetic `wsv`. We need to ensure that the behaviours are identical, which can be false due to bugs in either the `predicate` or the sending/receiving logic. 

### Step 2: Identify the short-circuiting heuristics

It might be as simple as saying that any `StringPredicate::is` applied to an identifiable will result in either `0` (not found), or at most `1` element in a `vec`, because identifiers are unique per class of object and our queries return homogeneous data. 

As long as we find a neat way to execute these heuristics on the backend generically. This should prevent a redundant clone after #3621 is merged, and potentially work better if we replace hash tables with prefix tries (no that's not a spelling mistake). 

### Step 3: Implement new logic with heuristics

### Step 4: Deprecate old queries

At this point we might want to consider adding new predicates to express the missing information. In principle we should be able to bridge the gap between what Iroha can do now and what the GraphQl API should be post-launch (#1455). 

As this is a complex task, I'd recommend that @DCNick3 take up another smaller task until this one is done.

mversic:
> Due to part of the context only being known at runtime, the latter operation is likely going to result in an lookup, in contrast to the for FindAccountByName. We can ooptionally counteract this problem by doing a short-circuiting heuristic.

[...]

PR #4640, opened 2024-05-23
refactor: remove domain from trigger identity

### LINKED ISSUE #4630: Remove a redundant trigger-domain_id connection
# Remove a redundant trigger-domain_id connection

**ref**. https://github.com/hyperledger/iroha/pull/4616#issuecomment-2119188143

I think we should remove connection of a trigger with a domain_id. It doesn't make sense because each trigger's authority already has this connection. Moreover, it creates some redundant difficulties in cases of permission granting and checking ref is_trigger_owner()

# Minimum working example

Proposed a new trigger structure
```
    pub struct Trigger {
        /// [`Id`] of the [`Trigger`].
        pub id: Name,
        /// Action to be performed when the trigger matches.
        pub action: action::Action,
    }
```

nxsaken:
## Description

Remove `DomainId` from `TriggerId`.

### Linked issue

Closes #4630.

### Checklist


nxsaken:
I'm having a bad time with integration tests, many of which are failing with a cryptic error, even those unrelated to triggers:

`domain_owner_trigger_permissions`:

```
ERROR iroha_core::tx: Internal error occurred during transaction validation, is Runtime Executor correct?, error: "error while executing at wasm backtrace:\n    0: 0x24ccf - <unknown>!<wasm function 235>: wasm trap: wasm `unreachable` instruction executed"
```

`grant_revoke_role_permissions`:

```
ERROR iroha_core::tx: Internal error occurred during transaction validation, is Runtime Executor correct?, error: "error while executing at wasm backtrace:\n    0: 0x13a80 - <unknown>!<wasm function 104>\n    1: 0x562f - <unknown>!<wasm function 43>\n    2: 0x24c89 - <unknown>!<wasm function 235>: Could not decode `SmartContractQueryRequest.0`:\n\tCould not decode `QueryRequest::Query.0`:\n\t\tCould not decode `SmartContractQuery::query`:\n\t\t\tCould not decode `QueryBox::FindRolesByAccountId.0`:\n\t\t\t\tCould not decode `FindRolesByAccountId::id`:\n\t\t\t\t\tCould not decode `AccountId::signatory`:\n\t\t\t\t\t\tout of range decoding Compact<u32>\n: Could not decode `QueryRequest::Query.0`:\n\tCould not decode `SmartContractQuery::query`:\n\t\tCould not decode `QueryBox::FindRolesByAccountId.0`:\n\t\t\tCould not decode `FindRolesByAccountId::id`:\n\t\t\t\tCould not decode `AccountId::signatory`:\n\t\t\t\t\tout of range decoding Compact<u32>\n: Could not decode `SmartContractQuery::query`:\n\tCould not decode `QueryBox::FindRolesByAccountId.0`:\n\t\tCould not decode `FindRolesByAccountId::id`:\n\t\t\tCould not decode `AccountId::signatory`:\n\t\t\t\tout of range decoding Compact<u32>\n: Could not decode `QueryBox::FindRolesByAccountId.0`:\n\tCould not decode `FindRolesByAccountId::id`:\n\t\tCould not decode `AccountId::signatory`:\n\t\t\tout of range decoding Compact<u32>\n: Could not decode `FindRolesByAccountId::id`:\n\tCould not decode `AccountId::signatory`:\n\t\tout of range decoding Compact<u32>\n: Could not decode `AccountId::signatory`:\n\tout of range decoding Compact<u32>\n: out of range decoding Compact<u32>"
```

Erigara (commented):
Ping for approval after executor recompile

Erigara (approved):

core/src/smartcontracts/isi/world.rs
```diff
@@ -121,23 +121,11 @@ pub mod isi {
...
-            let triggers_in_domain = state_transaction
-                .world()
-                .triggers()
-                .inspect_by_domain_id(&domain_id, |trigger_id, _| trigger_id.clone())
-                .collect::<Vec<_>>();
-
             let world = &mut state_transaction.world;
-            for trigger_id in &triggers_in_domain {
-                assert!(world.triggers.remove(trigger_id.clone()));
-            }
```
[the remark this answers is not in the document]
nxsaken: I think you're right, thanks for the catch.

core/src/smartcontracts/isi/triggers/set.rs
```diff
@@ -966,12 +962,9 @@ impl<'block, 'set> SetTransaction<'block, 'set> {
     // Passing by value to follow other `handle_` methods interface
     #[allow(clippy::needless_pass_by_value)]
     pub fn handle_data_event(&mut self, event: DataEvent) {
-        self.data_triggers
-            .iter()
-            .filter(|(id, _)| id.domain_id.is_none() || id.domain_id.as_ref() == event.domain_id())
```
VAmuzing, line 971: Is it fine to remove this filter? I don't see this check in previous calls.

client/src/client.rs
```diff
@@ -1507,6 +1507,11 @@ pub mod trigger {
     //! Module with queries for triggers
     use super::*;
 
+    /// Construct a query to get a trigger by its id
+    pub fn by_id(trigger_id: TriggerId) -> FindTriggerById {
+        FindTriggerById::new(trigger_id)
+    }
+
     /// Construct a query to get triggers by domain id
     pub fn by_domain_id(domain_id: DomainId) -> FindTriggersByDomainId {
```
mversic, line 1516: I don't think having this query makes sense anymore?

mversic: Maybe we can have `FindTriggersByAccountId`?

mversic:
can we say this. There are 2 accounts, the one that registers the trigger and the one that executes(owns?) the trigger? They can but don't have to be the same. 

In this case would it make more sense to remove `Action::authority` and add `TriggerId::account_id`?

client/src/client.rs
nxsaken: The query is still functional in the sense that it will find all triggers that are supposed to be executed on behalf of anyone belonging in the domain. It should at least be renamed to reflect what `DomainId` means (it's the authority's domain), if not removed as you suggest. And we can have something like `FindTriggersByAuthority`, which takes an `AccountId`.

mversic: ok, I think we can keep `FindTriggersByDomainId`. But let's also add `FindTriggersByAccountId`. 

Instead of all this, we should have `FindTriggers` combined with a filter to select only for some `DomainId`/`AccountId`? @DCNick3 no?

DCNick3: Yeah, that's what #3671 is about, but it's still some ways off..

nxsaken:
@mversic yes, the registering authority and the executing authority are not necessarily the same. I guess if we move the executing authority's `AccountId` inside `TriggerId`, it will allow registering the same trigger name under different accounts (not sure if it's an error to register a duplicate object or it just rewrites the current thing). Trigger IDs would look like `my_trigger$alice@wonderland` instead of just `my_trigger`. Do we want that?

Also, I don't think there is any sort of validation on whether one account can make another account the executing authority for a trigger. I see that the (registering) `authority` field passed to the `execute` method is unused. Maybe that should be a permission?

nxsaken:
You're right, I forgot about the validator visitors.

Stukalov-A-M:
> > Also, I don't think there is any sort of validation on whether one account can make another account the executing authority for a trigger. I see that the (registering) `authority` field passed to the `execute` method is unused. Maybe that should be a permission?
> 
> I think @Stukalov-A-M took care of this in #4616? if not let's make a separate ticket

Yea, I've made this validation. ref. `client/tests/integration/triggers/by_call_trigger.rs` 
`fn only_account_with_permission_can_register_trigger()`

There are only 3 scenarios when a trigger can be registered:
1. The transaction signer is an Action.authority
2. The transaction signer is a domain owner for Action.authority
3. The transaction signer has a permission to register trigger with Action.authority on behalf of Action.authority

Stukalov-A-M:
> can we say this. There are 2 accounts, the one that registers the trigger and the one that executes(owns?) the trigger? They can but don't have to be the same.


There is a mess with authorities )))

It's possible only if account that registers trigger has an appropriate permission.

> Would it make more sense to remove `Action::authority` and add `TriggerId::account_id`?

Sound good! But not here `TriggerId::account_id`, here `Trigger::owner`

Stukalov-A-M:
> @mversic yes, the registering authority and the executing authority are not necessarily the same. I guess if we move the executing authority's `AccountId` inside `TriggerId`, it will allow registering the same trigger name under different accounts (not sure if it's an error to register a duplicate object or it just rewrites the current thing). Trigger IDs would look like `my_trigger$alice@wonderland` instead of just `my_trigger`. Do we want that?

There is no need to complicate a `TriggerId` imho

Erigara:
I agree that there is no need to complicate `TriggerId`

nxsaken:
How should a trigger be unregistered when its authority or its domain gets unregistered? What if someone unregisters an account that was the authority for a trigger, yet the one unregistering the account doesn't have the authority to unregister the trigger?

And should unregistering those orphaned triggers emit events as if `Unregister<Trigger>` was executed directly from `Unregister<Account/Domain>`?

s8sato:
Speaking of preventing orphaned triggers, I guess trigger will be able to be identified as an account. See #4670

nxsaken:
@mversic I think I have [here](https://github.com/hyperledger/iroha/pull/4640/files#diff-075e74b164d4b37493e4fe8f7592c29bc610b8e7c609b4e0e424aa6ca2953789R86)

mversic (dismissed):

mversic:
@nxsaken rebase and ping for approval

core/src/smartcontracts/isi/world.rs
```diff
@@ -303,7 +318,7 @@ pub mod isi {
             let parameter_id = parameter.id.clone();
 
             let world = &mut state_transaction.world;
-            if !world.parameters.remove(&parameter) {
+            if !world.parameters.swap_remove(&parameter) {
```
mversic, line 321: in some cases ordering might matter. For parameters I think it's fine

mversic: that being said, we should just use `BTreeSet` to avoid confusion

nxsaken: `remove` is identical to `swap_remove`, just deprecated.

mversic (approved):

client/tests/integration/triggers/data_trigger.rs
```diff
@@ -41,72 +51,14 @@ fn must_execute_both_triggers() -> Result<()> {
...
 
-    test_client.submit_blocking(Register::account(Account::new(
-        gen_account_in("neverland").0,
-    )))?;
+    test_client.submit_blocking(Register::domain(Domain::new("neverland".parse()?)))?;
 
-    let new_value = get_asset_value(&test_client, asset_id);
-    assert_eq!(new_value, prev_value.checked_add(Numeric::ONE).unwrap());
+    let newer_value = get_asset_value(&test_client, asset_id);
+    assert_eq!(newer_value, new_value.checked_add(numeric!(1)).unwrap());
```
dima74, line 61: Test is called `must_execute_both_triggers`, but looks like now only first trigger get executed. Is it expected?

nxsaken: Both get executed, I check the decrement twice (`new_value` and `newer_value`).

dima74: Yes, sorry

dima74 (approved):

PR #4645, opened 2024-05-24
feat: Custom instructions in executor

### LINKED ISSUE #4599: Custom ISI/expressions in executor
I'd like if we could make one integration test that covers this use case.

### Rationale
Previously we had expression system but we removed it for the sake of simplicity and promised that it would become possible for blockchain admins to roll out their own expression system by upgrading executor

dima74:
## Description

Added ISI for custom instructions, and two tests demonstrating possible use:
* Test with one custom instruction `MintAssetForAllAccounts`
* Test with simple expression system

### Linked issue

Closes #4599

### Checklist


client/tests/integration/smartcontracts/executor_custom_data_model/src/complex.rs
```diff
@@ -0,0 +1,258 @@
...
+    use crate::complex::expression::EvaluatesTo;
+
+    #[derive(Debug, Deserialize, Serialize)]
+    pub enum CustomInstructionExpr {
+        Core(CoreExpr),
+        If(Box<ConditionalExpr>),
+        // Other custom ISI
+    }
+
+    // Built-in Iroha ISI (can be evaluated based on query values, etc)
```
nxsaken, line 25: I think this terminology is confusing, since "Iroha ISI" is a tautology, and "Custom ISI" is a contradiction (an instruction that's both custom and special)

0x009922 (changes requested):
So, the approach of custom instruction in this PR is the following, as I understand:

- In Executor, users can define whatever parsing logic around `Custom { payload: String }` instruction
- On client side, users _must_ know how to serialize the `Custom` instruction, and there is no way to know it via Iroha API

This goes against the current design. We had a discussion with @mversic, and the idea is that all executor-defined types must be defined in some schema and clients should be able to retrieve it and use it.

So, I guess this PR should came after #4597 and extend `ExecutorDataModel` with a set of instruction types, and then to have `Custom` instruction somehow this way:

```rust
struct Custom {
    id: CustomInstructionId, // or just Name?
    payload: JsonString
}
```

client/tests/integration/smartcontracts/executor_custom_data_model/src/complex.rs
mversic: > I think this terminology is confusing, since "Iroha ISI" is a tautology

agreed

> "Custom ISI" is a contradiction (an instruction that's both custom and special)

can you propose better naming in this case?

data_model/src/isi.rs
```diff
@@ -997,6 +1000,20 @@ mod transparent {
...
+        /// Custom instruction with arbitrary payload.
+        /// Should be handled in custom executor, where it will be translated to usual ISIs.
+        /// Can be used to extend instruction set or add expression system.
+        /// See `executor_custom_isi_simple` and `executor_custom_isi_complex`
+        /// examples in `client/tests/integration/smartcontracts`.
+        #[derive(Constructor, Display)]
+        #[display(fmt = "CUSTOM({payload})")]
+        pub struct Custom {
+            /// Custom payload
+            pub payload: String,
```
mversic, line 1014:
```suggestion
            pub payload: JsonString,
```

we have this type

client/tests/integration/upgrade.rs
```diff
@@ -195,6 +199,123 @@ fn executor_upgrade_should_revoke_removed_permissions() -> Result<()> {
...
+                FindAssetQuantityById::new(bob_rose.clone()).into(),
+            )),
+            Numeric::from(5u32),
+        );
+        let then = Burn::asset_numeric(Numeric::from(1u32), bob_rose.clone());
+        let then: InstructionBox = then.into();
+        let then = CustomInstructionExpr::Core(CoreExpr::new(then));
+        let isi = CustomInstructionExpr::If(Box::new(ConditionalExpr::new(condition, then)));
+        let isi = Custom::new(serde_json::to_string(&isi)?);
+        client.submit_blocking(isi)?;
```
mversic, line 297:
```suggestion
        client.submit_blocking(isi)?;
```

is it possible to implement `From<CustomInstructionExpr> for InstructionBox`? That would reduce friction in the API and make the conversion to json implicit

data_model/src/isi.rs
```diff
@@ -997,6 +1000,20 @@ mod transparent {
...
         }
     }
+
+    isi! {
+        /// Custom instruction with arbitrary payload.
+        /// Should be handled in custom executor, where it will be translated to usual ISIs.
+        /// Can be used to extend instruction set or add expression system.
+        /// See `executor_custom_isi_simple` and `executor_custom_isi_complex`
+        /// examples in `client/tests/integration/smartcontracts`.
+        #[derive(Constructor, Display)]
```
mversic, line 1010:
```suggestion
        #[derive(Display)]
```

take a look at `Permission::new` and mirror it

client/tests/integration/upgrade.rs
mversic: also add impl `From<CustomInstructionExpr> for Custom`

mversic:
great work

mversic:
> So, I guess this PR should came after #4597 and extend `ExecutorDataModel` with a set of instruction types, and then to have `Custom` instruction somehow this way:
> 
> ```rust
> struct Custom {
>     id: CustomInstructionId, // or just Name?
>     payload: JsonString
> }
> ```

I think you should make `ExecutorDataModel` a separate pr. And yeah, I was thinking that `Custom` should have an `id` field as you present here, but maybe there are some drawbacks?

client/tests/integration/smartcontracts/executor_custom_isi_complex/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Constructor, ValidateEntrypoints, Validate, Visit)]
+#[visit(custom(visit_custom))]
+struct Executor {
+    verdict: Result,
+    block_height: u64,
+}
+
+fn visit_custom(executor: &mut Executor, _authority: &AccountId, isi: &Custom) {
+    let Ok(isi) = serde_json::from_str::<CustomInstructionExpr>(&isi.payload()) else {
```
mversic, line 35:
```suggestion
    let Ok(isi) = CustomInstructionExpr::try_from(&isi.payload()) else {
```

client/tests/integration/smartcontracts/executor_custom_data_model/src/complex.rs
mversic: `syn` uses `Verbatim`. Maybe that would be more appropriate

0x009922:
> I think you should make `ExecutorDataModel` a separate pr.

Created https://github.com/hyperledger/iroha/pull/4658

> And yeah, I was thinking that `Custom` should have an `id` field as you present here, but maybe there are some drawbacks?

IMO, technically, there is no need to have an id. This would suffice:

```rust
struct ExecutorDataModel {
  schema: String,

  // -- snip --

  // type id in the schema
  custom_instruction_definition_id: Option<Name>
}

enum Instruction {
  // -- snip --
  Custom(JsonString)
}
```

Clients will know how to serialize `Custom` if they look into the data model.

client/tests/integration/smartcontracts/executor_custom_data_model/src/complex.rs
nxsaken: `Custom Instructions`, `External Instructions`, `User Instructions` (my favorite probably). Also, what makes `Iroha Special Instructions` "special", is it more of a marketing term? Maybe it would be clearer to call the essential ones `Builtin Instructions` or `Intrinsic Instructions`, so the totality of Intrinsics/Externals could form ISI.

mversic:
> IMO, technically, there is no need to have an id. This would suffice:

there is obviously no need in this PR, but I still maintain it might be useful. Maybe for migrations. Will `iroha_core` ever have to know that two custom instructions are of the same type, i.e. will it have to compare them by id. When `Custom` instruction contains only payload it's not possible to do so without invoking executor to evaluate the whole instruction. I think `id` brings a little more flexibility and makes it more future proof

0x009922:
> I think `id` brings a little more flexibility and makes it more future proof

I see your point, though, it sounds vague and this possible future-proof solution doesn't sound _very_ robust to me. Due to that, I would prefer a minimal implementation now, as we are not 100% sure that having `id` would help in future.

mversic:
> > I think `id` brings a little more flexibility and makes it more future proof
> 
> I see your point, though, it sounds vague and this possible future-proof solution doesn't sound _very_ robust to me. Due to that, I would prefer a minimal implementation now, as we are not 100% sure that having `id` would help in future.

ok then, I don't have strong arguments for adding instruction id. Then we should remove id from permissions as well. Not sure about parameters since they are `Identifiable`

I'd like some more opinions here: @Erigara @DCNick3

mversic:
> I see your point, though, it sounds vague and this possible future-proof solution doesn't sound very robust to me. Due to that, I would prefer a minimal implementation now, as we are not 100% sure that having id would help in future.

I didn't want to focus on future-proofing. It was more about `iroha_core` being able to categorize instructions. Because now there is no way for it to know whether 2 `Custom` instructions are of the same type nominally

Erigara:
Looks cool, give ability to implement custom instructions without too much changes.

One concern i can think of is that custom instructions work more like an extension, so executor can't for example completely replace isi with own DSL for example.

Alternative design i was thinking about was to make transaction payload opaque to the core iroha (just bytes essentially).
So executor has complete freedom in how to interpret this payload.
But this would involve much more work and i'm not sure we need such extensibility in the first place.

Erigara:
> It was more about iroha_core being able to categorize instructions. 

Do we need this ability since whole custom isi execution is on the executor?
In general i don't see strong need in adding `id`.

mversic:
> > It was more about iroha_core being able to categorize instructions.
> 
> Do we need this ability since whole custom isi execution is on the executor? In general i don't see strong need in adding `id`.

@dima74 keep it as it is then

mversic:
> > > It was more about iroha_core being able to categorize instructions.
> > 
> > 
> > Do we need this ability since whole custom isi execution is on the executor? In general i don't see strong need in adding `id`.
> 
> @dima74 keep it as it is then

moreover, remove `Permission::id` in a separate PR to be consistent with this one here

client/tests/integration/smartcontracts/executor_custom_data_model/src/complex.rs
dima74: Used `Custom instructions` and `Built-in instructions`

client/tests/integration/upgrade.rs
dima74: To use `client.submit()` for `CustomInstructionExpr`, it is necessary for `CustomInstructionExpr` to implement `Instruction` and `Sealed` traits. This will require some changes to iroha_data_model (at least make `Sealed` trait public). I am not sure if it is worth it.

Currently I added `pub fn into_instruction(self) -> InstructionBox` method to `CustomInstructionExpr`, so it is possible to write:
```rust
let isi = CustomInstructionExpr::If(...);
client.submit_blocking(isi.into_instruction())?;
```

dima74:
> This goes against the current design. We had a discussion with @mversic, and the idea is that all executor-defined types must be defined in some schema and clients should be able to retrieve it and use it.

I added `custom_instruction_definition_id` to `ExecutorDataModel`, however it is currently not used, e.g. user may forget to set it in executor `migrate` entrypoint, and there will be no errors when using custom instructions. Is it ok or should I add some validation of custom instruction payload against `custom_instruction_definition_id`?

mversic:
> I added `custom_instruction_definition_id` to `ExecutorDataModel`

please name it just `instruction_id`

> user may forget to set it in executor migrate entrypoint, and there will be no errors when using custom instructions. Is it ok or should I add some validation of custom instruction payload against `custom_instruction_definition_id`?

I think it's fine. It only exists for users to be able to query executor to find which custom instructions are available

mversic:
> moreover, remove `Permission::id` in a separate PR to be consistent with this one here

might not be possible to do because of migrations. What about migrations for custom instructions? I suppose it's not needed

client/tests/integration/smartcontracts/executor_custom_instructions_complex/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+            QueryOutputBox::Numeric(value) => Ok(Value::Numeric(value)),
+            _ => unimplemented!(),
+        }
+    }
+}
+
+#[entrypoint]
+pub fn migrate(_block_height: u64) -> MigrationResult {
+    DataModelBuilder::with_default_permissions()
+        .with_custom_instruction_type::<CustomInstructionExpr>()
```
mversic, line 82:
```suggestion
        .with_custom_instruction::<CustomInstructionExpr>()
```

data_model/src/executor.rs
```diff
@@ -81,6 +83,9 @@ mod model {
         ///
         /// These IDs refer to the types in the schema.
         pub permissions: BTreeSet<PermissionId>,
+        /// Type id in the schema.
+        /// Corresponds to payload of `InstructionBox::Custom`.
+        pub custom_instruction_definition_id: Option<Ident>,
```
mversic, line 88:
```suggestion
        pub instruction_id: Option<Ident>,
```

mversic: it's a bit weird that it's an `Option` and not a `Vec`

client/tests/integration/smartcontracts/executor_custom_instructions_simple/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+        Mint::asset_numeric(isi.quantity, asset_id).execute()?;
+    }
+    Ok(())
+}
+
+#[entrypoint]
+pub fn migrate(_block_height: u64) -> MigrationResult {
+    DataModelBuilder::with_default_permissions()
+        .with_custom_instruction_type::<CustomInstructionBox>()
+        .build_and_set();
```
mversic, line 61:
```suggestion
    let data_model = DataModelBuilder::with_default_permissions()
        .with_custom_instruction_type::<CustomInstructionBox>()
        .build();
        
    set(data_model);
```

Is API similar to this possible?

Erigara:
> moreover, remove Permission::id in a separate PR to be consistent with this one here 

having `id`s for permissions might be still helpful because we do lookups on then which is not the case for instructions

mversic:
u

> > moreover, remove Permission::id in a separate PR to be consistent with this one here
> 
> having `id`s for permissions might be still helpful because we do lookups on then which is not the case for instructions

but the lookups are completely unnecessary. They only happen because we keep list of permission ids in wsv. There is no need for that

client/tests/integration/smartcontracts/executor_custom_instructions_simple/src/lib.rs
dima74: It is possible, but what is wrong with `build_and_set`?

data_model/src/executor.rs
dima74: Potentially it is possible to use arbitrary type for payload of custom instruction. If it is enum, in schema there will be `iroha_schema::Metadata::Enum` with vector of variants

client/tests/integration/smartcontracts/executor_custom_instructions_simple/src/lib.rs
mversic: which API do you find better?

dima74: I like current version with `build_and_set`

mversic (approved):

Erigara (dismissed):

0x009922 (approved):

mversic (dismissed):

mversic:
> > > moreover, remove Permission::id in a separate PR to be consistent with this one here
> > 
> > 
> > having `id`s for permissions might be still helpful because we do lookups on then which is not the case for instructions
> 
> but the lookups are completely unnecessary. They only happen because we keep list of permission ids in wsv. There is no need for that

created #4702 for further discussion, but I don't see any reason to keep `Permission::id` if we don't have it for instructions

client/tests/integration/upgrade.rs
mversic: I created #4703

0x009922 (dismissed):

mversic (dismissed):

mversic (approved):

0x009922 (approved):

PR #4648, opened 2024-05-27
ci: Remove coveralls

BAStos525:
## Description
1. Remove `coveralls` tool for coverage since it's useless while having it in the workflow trigger type (btw, it's possible to fix it). However, coverage is uploading in Sonarqube now.
2. Modify concurrency.

### Benefits

Remove `coveralls` that doesn't work.
Reduce the possible amount of parallel runs of `I2::Dev::CodeQuality` workflow.

### Checklist

### DIFF
--- a/.github/workflows/iroha2-dev-code-quality.yml
+++ b/.github/workflows/iroha2-dev-code-quality.yml
@@ -3,75 +3,120 @@ name: I2::Dev::CodeQuality
 on:
   workflow_run:
     workflows: ["I2::Dev::Tests", "I2::Dev::Static"]
-    types: [in_progress]
+    types: [completed]
 
 concurrency:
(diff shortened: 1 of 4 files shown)


.github/workflows/iroha2-dev-code-quality.yml
```diff
@@ -3,10 +3,10 @@ name: I2::Dev::CodeQuality
 on:
   workflow_run:
     workflows: ["I2::Dev::Tests", "I2::Dev::Static"]
-    types: [requested]
+    types: [in_progress]
```
BAStos525, line 6: May be it's possible even to set  `types: completed` to reduce parallels runs stronger? Btw, I think we want to start clippy and coverage jobs without waiting for finishing all other tests successfully.

PR #4650, opened 2024-05-27
fix(crypto): do not panic if the passed Ed25519Sha512 public key is of invalid length

### LINKED ISSUE #4649: `iroha_crypto` panic
when I try to decode `04000001aeeba99f825cf8e4e6d7270b30fd927aa4c5b030df2ff891f28c2cc5b6d5d7990000
` hex string with `parity_scale_decoder` I get a panic like this:

```
thread 'main' panicked at crypto/src/signature/ed25519.rs:31:31:
range end index 32 out of range for slice of length 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This happens on `v2.0.0-pre-rc.21.3` branch so it may not be a problem anymore

DCNick3:
## Description

Change the conversion of unsized slice to array reference to be infallible and form a better error message.

This prevents a panic during `parity_scale_cli`'s type probing

### Linked issue

Closes #4649 

### Checklist

- [ ] make CI pass

### DIFF
--- a/crypto/src/signature/ed25519.rs
+++ b/crypto/src/signature/ed25519.rs
@@ -13,7 +13,10 @@ pub type PrivateKey = ed25519_dalek::SigningKey;
 #[cfg(not(feature = "std"))]
 use alloc::{format, string::ToString as _, vec::Vec};
 
-fn parse_fixed_size<T, E, F, const SIZE: usize>(payload: &[u8], f: F) -> Result<T, ParseError>
+fn parse_fixed_size<T, E, F, const SIZE: usize>(
+    payload: &[u8],
+    fixed_parser: F,
+) -> Result<T, ParseError>
 where
     F: FnOnce(&[u8; SIZE]) -> Result<T, E>,
     E: core::fmt::Display,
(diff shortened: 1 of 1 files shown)


crypto/src/signature/ed25519.rs
```diff
@@ -11,7 +11,23 @@ pub type PublicKey = ed25519_dalek::VerifyingKey;
 pub type PrivateKey = ed25519_dalek::SigningKey;
 
 #[cfg(not(feature = "std"))]
-use alloc::{string::ToString as _, vec::Vec};
+use alloc::{format, string::ToString as _, vec::Vec};
+
+fn parse_fixed_size<T, E, F, const SIZE: usize>(payload: &[u8], f: F) -> Result<T, ParseError>
```
VAmuzing, line 16: We have both `f`'s and custom names for `Fn...` in the code. Do you think we should name it `f` here or maybe use some custom name?
Besides that everything looks good.

VAmuzing (dismissed):

crypto/src/signature/ed25519.rs
DCNick3: Yeah, naming things is good. Renamed to `fixed_parser` (because it parses a fixed-size array)

VAmuzing (approved):

mversic (approved):

Issue #4353, opened 2024-03-11
dynamic metadata value


mversic:
I think the values in metadata should be reconsidered. I'd be in favor of loosely typing it with `StringWithJson` and defer the parsing of the structure to executor as we do for permission tokens payload. But this is something I think should be explored in a separate PR

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/4305#discussion_r1515996223_

Stukalov-A-M:
@mversic , @Erigara 
After some exploration I found a few options how to realize it:
1. Add Custom(jsonString) into MetadataValueBox and keep all current functions and realization.
2. Keep `pub struct Metadata(pub(super) BTreeMap<String, JsonString>)` and remove all nested operation methods.
3. Refactor the Metadata in `Metadata(JsonString)`, implement basic CRUD methods and move parsing implementation to the client side;

I like the #3 the most 

P.S.  (Set/Remove)KeyValue will be changed to (Set/Remove)Metadata

Erigara:
I'm for option 2. because imo option 3. might be too coarse for some use cases, it might be costly to update whole metadata in one go.

On other hand nested operations look redundant and can be modeled with specific keys.

mversic:
I like 2 as well. I don't see a point in having nested metadata. And it can always be implemented in a custom executor, no? Is there some old issue talking about the reason behind introducing nested metadata? @s8sato 

I find that 1. might be good option for performance reasons but then again I don't think we will see any noticeable benefit

I don't think 3 is even possible because of Store Assets. And I don't think we should change the concept of metadata into something else but key-value store

s8sato:
I think the nested metadata just aimed a standard datastore. For nested structures I think `JsonString` would be enough, so I stand for 2. or 3.

Before accepting 2. I'd consider the reason for keeping the first keys outside `JsonString` when `JsonString` can represent complete key-value store:

- currently `SetKeyValue` `RemoveKeyValue` instructions and `Find*By*Key` queries rely on `key: Name`
  - a key (and a key-value) could be extended to a path in json
- currently `CanSetKeyValue*` permission allows changing entire metadata of the entity
  - could be more granular with the same structure as metadata that represents accessible paths in the metadata

Are there requirements that metadata should hold in terms of use cases we should support @Mingela ?

- related closed issues as far as I can see: #1144, #1371, #3082

PR #4658, opened 2024-05-28
feat!: generalize permissions schema to executor data model

0x009922:
## Description

This is a version of #4597, but without changing anything about parameters. What to do with parameters is not yet clear.

### Benefits

Gives a foundation for executor-defined...

- Parameters: #4352 
- Instructions: #4599 and #4645 
- Metadata: #4353 

### Linked issue

None


client/tests/integration/permissions.rs
```diff
@@ -315,10 +316,14 @@ fn stored_vs_granted_token_payload() -> Result<()> {
...
+        Permission::new(
             "CanSetKeyValueInUserAsset".parse().unwrap(),
-            // NOTE: Introduced additional whitespaces in the serialized form
-            &*format!(r###"{{ "asset_id" : "xor#wonderland#{mouse_id}" }}"###),
+            JsonString::from_json_string_unchecked(format!(
+                // Introducing some whitespaces
+                // This way, if the executor compares just JSON strings, this test would fail
+                r##"{{ "asset_id"   :   "xor#wonderland#{}" }}"##,
+                mouse_id
+            )),
```
mversic, line 326:
```suggestion
            JsonString::from_json_string_unchecked(format!(
                // Introducing some whitespaces
                // This way, if the executor compares just JSON strings, this test would fail
                r##"{{ "asset_id"   :   "xor#wonderland#{}" }}"##,
                mouse_id
            )),
```

was it not possible to create string with `json!` and then edit it's contents?

client/tests/integration/smartcontracts/executor_remove_permission/src/lib.rs
```diff
@@ -25,13 +26,9 @@ struct Executor {
...
-    let mut schema = default_permission_schema();
-    schema.remove::<iroha_executor::default::tokens::domain::CanUnregisterDomain>();
-
-    let (token_ids, schema_str) = schema.serialize();
-    iroha_executor::set_permission_schema(
-        &iroha_executor::data_model::permission::PermissionSchema::new(token_ids, schema_str),
-    );
+    DataModelBuilder::with_default_permissions()
+        .remove_permission::<CanUnregisterDomain>()
+        .set();
```
mversic, line 31:
```suggestion
        .build();
```

since it's a builder?

client/tests/integration/smartcontracts/executor_with_custom_permission/src/lib.rs
```diff
@@ -95,7 +95,9 @@ impl Executor {
                 })?;
 
                 if let Ok(can_unregister_domain_token) =
-                    iroha_executor::default::tokens::domain::CanUnregisterDomain::try_from(&token)
+                    iroha_executor::default::permissions::domain::CanUnregisterDomain::try_from_object(
```
mversic, line 98: I was hoping that you would leave `try_from`

client/tests/integration/upgrade.rs
```diff
@@ -144,9 +144,9 @@ fn executor_upgrade_should_revoke_removed_permissions() -> Result<()> {
 
     // Check that permission exists
     assert!(client
-        .request(FindPermissionSchema)?
+        .request(FindExecutorDataModel)?
         .permissions()
-        .contains(&can_unregister_domain_token.definition_id));
+        .contains(&can_unregister_domain_token.id));
```
mversic, line 149:
```suggestion
        .contains(&can_unregister_domain_token.id()));
```

you should use method access in client

core/src/state.rs
```diff
@@ -618,9 +619,9 @@ impl<'world> WorldBlock<'world> {
     /// Commit block's changes
     pub fn commit(self) {
         // IMPORTANT!!! Commit fields in reverse order, this way consistent results are insured
+        self.executor_data_model.commit();
         self.executor.commit();
```
mversic, line 623:
```suggestion
        self.executor.commit();
```

why isn't executor data model inside executor?

data_model/src/lib.rs
```diff
@@ -864,6 +866,113 @@ mod model {
...
+    /// This string is guaranteed to be parsed as JSON.
+    #[derive(Display, Default, Debug, Clone, Eq, Encode, Decode, TypeId, Ord, PartialOrd)]
+    #[ffi_type(unsafe {robust})]
+    #[repr(transparent)]
+    #[display(fmt = "{}", "0")]
+    pub struct JsonString(pub(super) String);
+}
+
+/// A helper trait for polymorphism, implemented for various types
+pub trait IntoJsonString {
```
mversic, line 881: please don't introduce it in this PR

data_model/src/lib.rs
```diff
@@ -864,6 +866,113 @@ mod model {
...
+    where
+        S: serde::Serializer,
+    {
+        let json = serde_json::Value::from_str(&self.0).map_err(serde::ser::Error::custom)?;
+        json.serialize(serializer)
+    }
+}
+
+impl IntoSchema for JsonString {
+    fn type_name() -> iroha_schema::Ident {
```
mversic, line 967: I believe you can derive this

data_model/src/lib.rs
```diff
@@ -864,6 +866,113 @@ mod model {
...
+        let serialized = serde_json::to_string(value)?;
+        // the string was obtained from serde_json serialization,
+        // so it should be a valid JSON string
+        Ok(Self(serialized))
+    }
+
+    /// Create without checking whether the input is a valid JSON string.
+    ///
+    /// The caller must guarantee that the value is valid.
+    pub fn from_json_string_unchecked(value: String) -> Self {
```
mversic, line 928: if you want to have this method, mark it as `transparent_api` or sth

data_model/src/permission.rs
```diff
@@ -17,196 +18,101 @@ pub type Permissions = BTreeSet<Permission>;
...
 }
 
-impl serde::ser::Serialize for JsonString {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        let json = serde_json::Value::from_str(&self.0).map_err(serde::ser::Error::custom)?;
-        json.serialize(serializer)
+impl Borrow<str> for Permission {
```
mversic, line 101: this doesn't make sense to implement

data_model/src/permission.rs
```diff
@@ -17,196 +18,101 @@ pub type Permissions = BTreeSet<Permission>;
...
 }
 
-impl<'de> serde::de::Deserialize<'de> for JsonString {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        let json = serde_json::Value::deserialize(deserializer)?;
-        Ok(Self::new(&json))
+impl Borrow<str> for PermissionId {
```
mversic, line 95: do you need this impl?

mversic (changes requested):

smart_contract/executor/src/default.rs
```diff
@@ -40,21 +41,10 @@ pub use trigger::{
...
-        };
-    }
-
-    tokens::map_token_type!(add_to_schema);
-
-    schema
-}
+use crate::{
+    permission::Permission as _,
+    prelude::{Permission as PermissionObject, *},
```
mversic, line 46: since you're importing trait as `_` do you need to rename the struct?

client/tests/integration/permissions.rs
0x009922: You mean, how? `json!` creates `serde_json::Value`.

core/src/state.rs
0x009922: For the same reason why permissions schema was not inside executor. I just followed the original pattern.

Why should it be a single entry in the `World`?

data_model/src/lib.rs
0x009922: Why?

mversic: I just didn't want to have it in the public API. Maybe this method can take `JsonString` and `JsonString` to have unchecked constructor?

mversic: because we'll use JsonString for different things

client/tests/integration/domain_owner_permissions.rs
```diff
@@ -48,7 +48,7 @@ fn domain_owner_domain_permissions() -> Result<()> {
     // Granting a respective token also allows "bob@kingdom" to do so
     let token = Permission::new(
         "CanRegisterAssetDefinitionInDomain".parse().unwrap(),
-        &json!({ "domain_id": kingdom_id }),
+        json!({ "domain_id": kingdom_id }).into(),
```
mversic, line 51: you should make:
```rust
fn new(id, PermissionId, impl Into<JsonValue>)
```

client/tests/integration/upgrade.rs
```diff
@@ -197,7 +199,7 @@ fn executor_upgrade_should_revoke_removed_permissions() -> Result<()> {
 
 #[test]
 fn migration_fail_should_not_cause_any_effects() {
-    let (_rt, _peer, client) = <PeerBuilder>::new().with_port(10_995).start_with_runtime();
+    let (_rt, _peer, client) = <PeerBuilder>::new().with_port(10_999).start_with_runtime();
```
dima74, line 202: Wouldn't it conflict with `connected_peers_with_f_1_0_1` which used port 11_000?

core/src/smartcontracts/isi/world.rs
```diff
@@ -377,18 +384,15 @@ pub mod isi {
...
-        permissions_before: Vec<PermissionId>,
+        permissions_before: BTreeSet<PermissionId>,
     ) -> Result<(), Error> {
         let world = state_transaction.world();
-        let permissions_after = world
-            .permission_schema()
-            .token_ids
-            .iter()
-            .collect::<IndexSet<_>>();
+        // permissions_before.reta
```
dima74, line 390:
```suggestion
```

client/tests/integration/upgrade.rs
0x009922: It might... changed to 10_998

data_model/src/lib.rs
0x009922: > Maybe this method can take `JsonString` and `JsonString` to have unchecked constructor?

I don't get it... This is already an unchecked constructor of `JsonString` - `JsonString::from_json_string_unchecked`.

data_model/src/query/mod.rs
```diff
@@ -356,8 +356,6 @@ impl TryFrom<QueryOutputBox> for u64 {
 /// Implements [`Query`] and, additionally, either [`SingularQuery`] or [`IterableQuery`],
 ///     depending on whether the output type is wrapped into a Vec (purely syntactically)
 macro_rules! impl_queries {
-    // base case for the tt-muncher
-    () => {};
```
0x009922, line 360: @DCNick3 rustc complained that it is unused 🤔

data_model/src/lib.rs
mversic: my bad. Then:
```suggestion
    pub fn from_string_unchecked(value: String) -> Self {
```

data_model/src/lib.rs
```diff
@@ -864,6 +864,83 @@ mod model {
...
+    }
+}
+
+impl From<serde_json::Value> for JsonString {
+    fn from(value: serde_json::Value) -> Self {
+        Self::from(&value)
+    }
+}
+
+impl PartialEq for JsonString {
```
mversic, line 919: it's ok to derive this

data_model/src/lib.rs
```diff
@@ -864,6 +864,83 @@ mod model {
...
+
+    /// Serializes a value into [`JsonString`]
+    /// # Errors
+    /// See [`serde_json::to_string`].
+    pub fn serialize<T: serde::Serialize>(value: &T) -> serde_json::Result<Self> {
+        let serialized = serde_json::to_string(value)?;
+        // the string was obtained from serde_json serialization,
+        // so it should be a valid JSON string
+        Ok(Self(serialized))
+    }
```
mversic, line 897: why do you have these 2 special functions when they are implemented as serde traits?

data_model/src/lib.rs
mversic: look at docs for `Permission` about being compared by payload and copy that here on `JsonString`

mversic: as you wish actually, but then you have to make `PartialOrd` consistent with `PartialEq`. Just make sure to transfer the docs here

mversic: the thing is that `asset##account@domain` will compare as not equal to `asset#domain#account@domain` so I thought we may as well not care about whitespaces

data_model/src/lib.rs
0x009922: Because these functions explicitly use `serde_json` serialization, and the caller doesn't need to care about it.

mversic (dismissed):

mversic (approved):

dima74 (dismissed):

mversic (approved):

dima74 (approved):

PR #4666, opened 2024-05-30
fix: avoid early drop of `ConfigReader` in `iroha_client`

### LINKED ISSUE #4623: forgot to call `ConfigReader::finish()`, didn't you?
in `Config::load` every early exit with `?` will produce this error because `Config` is dropped. It can be most easily reproduced if you send into `Config::load` name of a file that doesn't exist. If there are other constructors, they will suffer from the same issue.

0x009922:
## Description

Fixed usage of `ConfigReader` in `iroha_client`.

### Linked issue

Closes #4623

### DIFF
--- a/client/src/config.rs
+++ b/client/src/config.rs
@@ -78,8 +78,9 @@ impl Config {
     /// - unable to load config from a TOML file
     /// - the config is invalid
     pub fn load(path: impl AsRef<Path>) -> error_stack::Result<Self, LoadError> {
+        let toml = TomlSource::from_file(path).change_context(LoadError)?;
         let config = ConfigReader::new()
-            .with_toml_source(TomlSource::from_file(path).change_context(LoadError)?)
+            .with_toml_source(toml)
             .read_and_complete::<user::Root>()
             .change_context(LoadError)?
             .parse()
@@ -90,6 +91,7 @@ impl Config {
 
 #[cfg(test)]
 mod tests {
+    use assertables::{assert_contains, assert_contains_as_result};
     use iroha_config_base::env::MockEnv;
 
     use super::*;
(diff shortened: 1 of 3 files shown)


client/src/config.rs
```diff
@@ -90,6 +91,7 @@ impl Config {
 
 #[cfg(test)]
 mod tests {
+    use assertables::{assert_contains, assert_contains_as_result};
```
nxsaken, line 94: Unused import?

0x009922: It's used by the output of `assert_contains`. Maybe it should be patched, actually... Out of scope of this PR anyway.

nxsaken (approved):

VAmuzing (approved):

nxsaken:
You should edit the commit message to be the same as the PR title to make the check pass.

client/src/config.rs
```diff
@@ -78,8 +78,9 @@ impl Config {
     /// - unable to load config from a TOML file
     /// - the config is invalid
     pub fn load(path: impl AsRef<Path>) -> error_stack::Result<Self, LoadError> {
+        let toml = TomlSource::from_file(path).change_context(LoadError)?;
         let config = ConfigReader::new()
-            .with_toml_source(TomlSource::from_file(path).change_context(LoadError)?)
+            .with_toml_source(toml)
             .read_and_complete::<user::Root>()
             .change_context(LoadError)?
```
mversic, line 85: what if we get an error here? Won't this also fail with drop bomb?

nxsaken: shouldn't since the `ConfigReader` wouldn't have been constructed yet

0x009922: > what if we get an error here? Won't this also fail with drop bomb?

`read_and_complete` defuses the bomb, so no, it shouldn't. I _believe_ tests cover it.

mversic (approved):

Issue #4295, opened 2024-02-18
Implement human-readable bytes in the config


0x009922:
## Description

Actually implement parsing from a human-readable string for `HumanBytes`:

https://github.com/hyperledger/iroha/blob/964476722e2a219becaacdb3676ca058ec5748cd/config/base/src/lib.rs#L36-L38

### Specification

Here is an excerpt from the config reference draft:

> Bytes amount is specified as a human-readable string:
> 
> ```toml
> # 42 bytes
> value1 = "42B"
> 
> # 1 kilobyte = 1000 bytes
> value2 = "1KB"
> 
> # 1 kilobyte (binary format) = 1024 bytes
> value3 = "1KiB"
> 
> # Sum of multiple
> value4 = "1GB 5MB"
> ```
> 
> Iroha can parse sizes in bytes, kilobytes (`K`), megabytes (`M`), gigabytes (`G`), terabytes (`T`),
> and petabytes (`P`).
> 
> The format of suffixes:
> 
> - **`{size}iB`:** Binary size
> - **`{size}B`:** Decimal size

I used [`humanfriendly`](https://humanfriendly.readthedocs.io/en/latest/api.html?highlight=parse_size#humanfriendly.parse_size) Python package as a reference.

## Also

- [kb, kB, KiB… What’s Up With That? | Pacoup.com](https://web.archive.org/web/20150324153922/https://pacoup.com/2009/05/26/kb-kb-kib-whats-up-with-that/)
- #4294

mversic:
as discussed https://github.com/hyperledger/iroha-2-docs/issues/392 I'm not sure we want to complicate config files. Simple fixed suffix on the field name should suffice IMO. We need to discuss this before proceeding

DCNick3:
I am in favor of having human-readable suffixes in config, this makes it much easier to write large values by hand.

Oftentimes, when a large value is needed, you would still end up having a comment to explain what the value actually is, like `max_transfer_bytes = 1048576 # 1 MiB`, so having this natively supported by the config system is better UX IMO.

Same opinion on the #4294 with the same arguments, `time_value = "1h"` is better than `time_value_ms = 3600000 # 1 hour`

Issue #4294, opened 2024-02-18
Implement human-readable durations in the config


0x009922:
## Description

Enhance `HumanDuration` type to actually parse a human-readable string:

https://github.com/hyperledger/iroha/blob/964476722e2a219becaacdb3676ca058ec5748cd/config/base/src/lib.rs#L23-L27

### Specification

Here is an excerpt from the configuration reference draft:

> Duration is specified as a human-readable string:
> 
> ```toml
> value1 = "1sec"
> value2 = "1hour 12min 5s"
> value3 = "2years 2min 12us"
> value4 = "550ms"
> ```
> 
> The duration string is a concatenation of time spans. Each time span is an
> integer number and a suffix. Supported suffixes:
> 
> - `nsec`, `ns` &mdash; nanoseconds
> - `usec`, `us` &mdash; microseconds
> - `msec`, `ms` &mdash; milliseconds
> - `seconds`, `second`, `sec`, `s`
> - `minutes`, `minute`, `min`, `m`
> - `hours`, `hour`, `hr`, `h`
> - `days`, `day`, `d`
> - `weeks`, `week`, `w`
> - `months`, `month`, `M` &mdash; defined as $30.44$ days
> - `years`, `year`, `y` &mdash; defined as $365.25$ days

I used [`humantime`](https://docs.rs/humantime/latest/humantime/fn.parse_duration.html) crate for reference. Maybe it can be used as is. Otherwise, it is straighforward (and quite interesting!) to write our own parser with something like [`winnow`](https://docs.rs/winnow).

mversic:
as discussed [here](https://github.com/hyperledger/iroha-2-docs/issues/392) I'm not sure we want to complicate config files. Simple fixed suffix on the field name should suffice IMO. We need to discuss this before proceeding

0x009922:
There was a discussion in favor of human-readable strings here:

- https://github.com/hyperledger/iroha-2-docs/issues/392

PR #4667, opened 2024-05-30
refactor!: revert `_ms` and `_bytes` suffixes in config

0x009922:
## Description

Since there is still no consensus about using human readable types in configuration (#4294, #4295), I decided to at least make it explicit that they aren't supported.

- Reverted use of `_ms` suffix for millisecond numbers
- Used `_bytes` suffix for <code>torii.max_content_length<b>_bytes</b></code>
- Updated config templates
- Refactored internals:
  - Renamed `HumanDuration` to `DurationMs`, and `HumanBytes` to just `Bytes`
  - Chore refactors

### DIFF
--- a/config/src/parameters/defaults.rs
+++ b/config/src/parameters/defaults.rs
@@ -46,14 +46,15 @@ pub mod snapshot {
 }
 
 pub mod chain_wide {
+    use iroha_config_base::util::Bytes;
+
     use super::*;
 
     pub const MAX_TXS: NonZeroU32 = nonzero!(2_u32.pow(9));
     pub const BLOCK_TIME: Duration = Duration::from_secs(2);
     pub const COMMIT_TIME: Duration = Duration::from_secs(4);
     pub const WASM_FUEL_LIMIT: u64 = 55_000_000;
-    // TODO: wrap into a `Bytes` newtype
-    pub const WASM_MAX_MEMORY_BYTES: u32 = 500 * 2_u32.pow(20);
+    pub const WASM_MAX_MEMORY: Bytes<u32> = Bytes(500 * 2_u32.pow(20));
 
     /// Default estimation of consensus duration.
     pub const CONSENSUS_ESTIMATION: Duration =
@@ -82,7 +83,9 @@ pub mod chain_wide {
 pub mod torii {
     use std::time::Duration;
 
-    pub const MAX_CONTENT_LENGTH: u64 = 2_u64.pow(20) * 16;
+    use iroha_config_base::util::Bytes;
+
+    pub const MAX_CONTENT_LENGTH: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
     pub const QUERY_IDLE_TIME: Duration = Duration::from_secs(30);
 }
 
--- a/config/src/parameters/defaults.rs
+++ b/config/src/parameters/defaults.rs
@@ -85,7 +85,7 @@ pub mod torii {
 
     use iroha_config_base::util::Bytes;
 
-    pub const MAX_CONTENT_LENGTH: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
+    pub const MAX_CONTENT_LEN: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
     pub const QUERY_IDLE_TIME: Duration = Duration::from_secs(30);
 }
 
--- a/config/src/parameters/user.rs
+++ b/config/src/parameters/user.rs
@@ -512,7 +512,7 @@ impl ChainWide {
 pub struct Torii {
     #[config(env = "API_ADDRESS")]
     pub address: WithOrigin<SocketAddr>,
-    #[config(default = "defaults::torii::MAX_CONTENT_LENGTH")]
+    #[config(default = "defaults::torii::MAX_CONTENT_LEN")]
     pub max_content_len: Bytes<u64>,
     #[config(default = "defaults::torii::QUERY_IDLE_TIME.into()")]
     pub query_idle_time_ms: DurationMs,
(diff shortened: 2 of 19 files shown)


nxsaken (approved):

0x009922:
CI failed due to a single `unused` import.

I made this rule a warning locally. And, I don't think it should be a hard error in general. Warning is just fine in all circumstances.

nxsaken (dismissed):

nxsaken (dismissed):

VAmuzing (dismissed):

nxsaken (approved):

config/src/parameters/user.rs
```diff
@@ -512,21 +512,21 @@ impl ChainWide {
 pub struct Torii {
     #[config(env = "API_ADDRESS")]
     pub address: WithOrigin<SocketAddr>,
-    #[config(default = "defaults::torii::MAX_CONTENT_LENGTH.into()")]
-    pub max_content_length: HumanBytes<u64>,
+    #[config(default = "defaults::torii::MAX_CONTENT_LENGTH")]
+    pub max_content_length_bytes: Bytes<u64>,
```
mversic, line 516:
```suggestion
    pub max_content_len: Bytes<u64>,
```

I don't think a suffix is needed here

config/src/parameters/user.rs
```diff
@@ -458,20 +458,20 @@ pub struct ChainWide {
...
-    #[config(default = "defaults::chain_wide::WASM_MAX_MEMORY_BYTES")]
-    pub executor_max_memory: u32,
+    #[config(default = "defaults::chain_wide::WASM_MAX_MEMORY")]
+    pub executor_max_memory_bytes: Bytes<u32>,
     #[config(default = "defaults::chain_wide::WASM_FUEL_LIMIT")]
     pub wasm_fuel_limit: u64,
-    #[config(default = "defaults::chain_wide::WASM_MAX_MEMORY_BYTES")]
-    pub wasm_max_memory: u32,
+    #[config(default = "defaults::chain_wide::WASM_MAX_MEMORY")]
+    pub wasm_max_memory_bytes: Bytes<u32>,
```
mversic, line 466:
```suggestion
    pub wasm_max_memory: Bytes<u32>,
```

likewise

mversic (changes requested):

config/src/parameters/user.rs
mversic: @nxsaken what do you think? Just as an example rust's std has `string.len()` and returns len in bytes

nxsaken: Generally agree, if the type conveys that info, then the field name should stay minimal IMO. But these fields seem to get de/serialized into an untyped format, so maybe they should contain the full info.

mversic: I would assume that length is always in bytes or we could have that as the default assumption

torii/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub struct Torii {
     notify_shutdown: Arc<Notify>,
     query_service: LiveQueryStoreHandle,
     kura: Arc<Kura>,
-    transaction_max_content_length: u64,
+    transaction_max_content_length: Bytes<u64>,
```
nxsaken, line 55:
```suggestion
    transaction_max_content_len: Bytes<u64>,
```

torii/src/lib.rs
```diff
@@ -86,7 +86,7 @@ impl Torii {
             #[cfg(feature = "telemetry")]
             metrics_reporter,
             address: config.address.into_value(),
-            transaction_max_content_length: config.max_content_len_bytes,
+            transaction_max_content_length: config.max_content_len,
```
nxsaken, line 89:
```suggestion
            transaction_max_content_len: config.max_content_len,
```

torii/src/lib.rs
```diff
@@ -168,7 +168,7 @@ impl Torii {
                     warp::path(uri::TRANSACTION)
                         .and(add_state!(self.chain_id, self.queue, self.state.clone()))
                         .and(warp::body::content_length_limit(
-                            self.transaction_max_content_length,
+                            self.transaction_max_content_length.get(),
```
nxsaken, line 171:
```suggestion
                            self.transaction_max_content_len.get(),
```

config/src/parameters/defaults.rs
```diff
@@ -82,7 +83,9 @@ pub mod chain_wide {
 pub mod torii {
     use std::time::Duration;
 
-    pub const MAX_CONTENT_LENGTH: u64 = 2_u64.pow(20) * 16;
+    use iroha_config_base::util::Bytes;
+
+    pub const MAX_CONTENT_LENGTH: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
```
nxsaken, line 88:
```suggestion
    pub const MAX_CONTENT_LEN: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
```

config/src/parameters/user.rs
```diff
@@ -512,21 +512,21 @@ impl ChainWide {
 pub struct Torii {
     #[config(env = "API_ADDRESS")]
     pub address: WithOrigin<SocketAddr>,
-    #[config(default = "defaults::torii::MAX_CONTENT_LENGTH.into()")]
-    pub max_content_length: HumanBytes<u64>,
+    #[config(default = "defaults::torii::MAX_CONTENT_LENGTH")]
```
nxsaken, line 515:
```suggestion
    #[config(default = "defaults::torii::MAX_CONTENT_LEN")]
```

nxsaken (changes requested):

mversic (approved):

nxsaken (dismissed):

mversic (approved):

nxsaken (approved):

Issue #1212, opened 2021-07-15
Research the defi functionality


takemiyamakoto:
In order to finish the ISI design we need to think of Defi instruction design. Main focus is the most popular use cases for defi which can be implemented inside ISI.

https://wiki.hyperledger.org/display/iroha/Set+of+OOB+ISIs

appetrosyan:
Consider implementing it via WASM instead.

takemiyamakoto:
I wouldn't do this via wasm. Rather, I would put this in the core and add other functions too, similar to unreal engine Blueprint, which has many built-in functions that people can combine.

PR #4668, opened 2024-05-30
feat!: recognize and activate accounts

### LINKED ISSUE #4426: Permissionless accounts: registration-free model
> I think the point is the domain membership screening.
> How should a new account apply for and be allowed to join a domain?
> In terms of authority, there will be a gap between transferring one's own asset to a new account and registering a new account, especially in "private domain".
> So as you pointed out, the admission process should be managed by a different authority than the sender of the first transaction that mentioned the account.
> 
> One idea is that each domain has its own admission policy as a trigger that "activates" a new account in response to the account creation event.
> Admission policies may vary from domain to domain e.g.
> 
> - No screening
> - Wait for approval of the domain owner
> - Wait for approvals of m of n domain members
> 
> In any case, as I implied, I'm going to create a new issue on this and remove the 2nd commit from this PR

_Originally posted by @s8sato in https://github.com/hyperledger/iroha/pull/4411#discussion_r1562173269_

s8sato:
BREAKING_CHANGE:

- `Account` contains `is_active` field
- `AccountEvent::Created` divides into `Recognized` and `Activated`
- `ValidationFail` contains `UnrecognizedAuthority` and `InactiveAuthority` variants

## Description

When an account is _recognized_,

- it becomes able to hold assets, permissions, roles, and metadata
- it cannot yet execute any queries or transactions

When an account is _activated_,

- it becomes able to execute queries and transactions

Recognize and activate an account when targeted as:

- `Register<Account>` object

Recognize an account when targeted as:

- `Register<Asset>` object.account
- `Mint<Numeric, Asset>` destination.account
- ~~`Transfer<Account, DomainId, Account>` destination~~ See [comment](https://github.com/hyperledger/iroha/pull/4668#discussion_r1619777320)
- `Transfer<Account, AssetDefinitionId, Account>` destination
- `Transfer<Asset, Numeric, Account>` destination
- `Transfer<Asset, Metadata, Account>` destination
- `SetKeyValue<Account>` object
- `SetKeyValue<Asset>` object.account
- `Grant<Permission, Account>` destination
- `Grant<Role, Account>` destination

Let me call these _creative instructions_ in documentation

### Linked issue

- closes #4426

### Benefits

- who targets an account does not need to care if it has been registered or not
- depending on use case, accounts can be registered implicitly/automatically or explicitly/manually

### Checklist


client/tests/integration/account_registration/public.rs
```diff
@@ -0,0 +1,102 @@
...
+/// - recognized when targeted as a destination of a transfer of numeric asset e.g. "rose"
+/// - automatically activated by an administrative trigger
+///
+/// # Scenario
+///
+/// 0. ... administrative trigger registered
+/// 0. new carol targeted ... carol recognized
+/// 0. clock forward by one block ... carol activated
+/// 0. carol tries query ... ok
+/// 0. carol tries transaction ... ok
```
s8sato, line 23: This public scenario describes the main goal of this PR

client/tests/integration/account_registration/public.rs
```diff
@@ -0,0 +1,102 @@
...
+    .show_output()
+    .build()?
+    .optimize()?
+    .into_bytes()?;
+    let register_activator = Register::trigger(Trigger::new(
+        "activator$wonderland".parse().unwrap(),
+        Action::new(
+            WasmSmartContract::from_compiled(wasm),
+            Repeats::Indefinitely,
+            ALICE_ID.clone(),
```
s8sato, line 43: The trigger authority is tentatively set to alice, but once the account ID is expanded to off-curve multihash, this should be a system (non-personal) authority

client/tests/integration/account_registration/private.rs
```diff
@@ -0,0 +1,90 @@
...
+/// - activated when manually registered by an administrative account e.g. "alice"
+///
+/// # Scenario
+///
+/// 0. new carol targeted ... carol recognized
+/// 0. carol tries query ... err
+/// 0. carol tries transaction ... err
+/// 0. register carol ... carol activated
+/// 0. carol tries query ... ok
+/// 0. carol tries transaction ... ok
```
s8sato, line 23: For private scenarios, the aim is no impact on membership and privacy while relaxing restrictions on the order of registering and targeting accounts

client/tests/integration/account_registration/mod.rs
s8sato, line 1: The idea is to realize public/private characteristics not by different executors, but by different account activators (trigger/administrator). In this way, each domain can have different activation policies under the common executor

core/src/smartcontracts/isi/domain.rs
```diff
@@ -339,6 +359,9 @@ pub mod isi {
             } = self;
 
             let _ = state_transaction.world.account(&source_id)?;
+            // Exceptionally, the destination account should not be recognized here.
+            // Otherwise, the risk is that the previous owner can no longer activate the current owner who cannot yet take any action by oneself.
+            // Thus, the destination account should be explicitly registered before this transfer
             let _ = state_transaction.world.account(&destination_id)?;
```
s8sato, line 365: It may be dangerous to recognize an account when targeted as `Transfer<Account, DomainId, Account>` destination. The risk is that the previous owner can no longer activate the current owner who cannot yet take any action by oneself

data_model/src/account.rs
s8sato, line 1: This PR keeps the nested structure of IDs as it is focused on feat. My another commit eliminated dependency between entity IDs and the event hierarchy, so hopefully entities can be placed in storages directly under the world. However, I'd leave it to another perf. PR with proper benchmarking and profiling

s8sato (commented):
- [ ] unit tests for all patterns of account recognition

client/tests/integration/account_registration/public.rs
nxsaken: Is there an issue for it?

core/src/executor.rs
```diff
@@ -185,6 +198,15 @@ impl Executor {
     ) -> Result<(), ValidationFail> {
         trace!("Running instruction validation");
 
+        if 0 != state_transaction.height() {
+            let Ok(authority_account) = state_transaction.world.account(authority) else {
```
nxsaken, line 202: This could be a function to reduce code duplication

core/src/smartcontracts/isi/query.rs
```diff
@@ -316,7 +316,8 @@ mod tests {
     fn world_with_test_domains() -> World {
         let domain_id = DomainId::from_str("wonderland").expect("Valid");
         let mut domain = Domain::new(domain_id).build(&ALICE_ID);
-        let account = Account::new(ALICE_ID.clone()).build(&ALICE_ID);
+        let mut account = Account::new(ALICE_ID.clone()).build(&ALICE_ID);
```
nxsaken, line 319: This frequent pattern of creating a new account and immediately activating it is asking for an `Account::new_active` constructor (marked with `#[cfg(test)]` probably).

client/tests/integration/account_registration/public.rs
s8sato: I don't think it has been broken down into issues yet.
If we want to support use cases where a trigger needs different permission than the registrant person (e.g. swap between two accounts, #1212), the trigger should reference its own account (or should be identified as an account).
The idea of _system level triggers_ and _technical accounts_ was suggested in as early as [Triggers ADR](https://wiki.hyperledger.org/display/iroha/Triggers).
Similar concepts include [Contract Address](https://info.etherscan.com/what-is-an-ethereum-address/) and [Program Derived Addresses](https://solanacookbook.com/core-concepts/pdas.html)

s8sato: I found my first comment off topic. In this use case the trigger manages domain membership, so alice (domain owner) would be an appropriate authority

data_model/src/account.rs
Erigara: I agree, let's stay focused

client/tests/integration/account_registration/public.rs
Erigara, line 1: Public case look like automation of private case.
What should distinguish this 2 cases?
That in public case account is guaranteed to be registered?

client/tests/integration/account_registration/public.rs
s8sato: - created #4670

client/tests/integration/account_registration/public.rs
s8sato: The idea is architecture does not force public or private use. It depends on what activator authority the founders of the chain or the domain will set

core/src/executor.rs
```diff
@@ -142,6 +144,15 @@ impl Executor {
...
         trace!("Running transaction validation");
 
+        if 0 != state_transaction.height() {
+            let Ok(authority_account) = state_transaction.world.account(authority) else {
+                return Err(ValidationFail::UnrecognizedAuthority);
+            };
+            if !authority_account.is_active {
+                return Err(ValidationFail::InactiveAuthority);
+            }
+        }
```
mversic, line 154: this should be verified before calling executor

core/src/executor.rs
mversic: IMO this should be verified before calling executor. Iroha should never call `validate_*` in the 1st place if authority doesn't exist or is inactive

core/src/smartcontracts/isi/account.rs
```diff
@@ -340,7 +338,7 @@ pub mod isi {
                 .into_iter()
                 .map(|token| token.id);
 
-            state_transaction.world.account(&account_id)?;
+            recognize_account(account_id.clone(), authority, state_transaction)?;
```
mversic, line 341:
```suggestion
            recognize_account(&account_id, authority, state_transaction)?;
```

IMO you should move the clone inside the function because the most common execution path will just return

core/src/smartcontracts/isi/domain.rs
```diff
@@ -47,28 +47,48 @@ pub mod isi {
...
+                Ok(existing_account) => {
+                    if existing_account.is_active {
+                        return Err(RepetitionError {
+                            instruction_type: InstructionType::Register,
+                            id: IdBox::AccountId(account_id),
+                        }
+                        .into());
+                    }
+
+                    existing_account.activate();
```
mversic, line 68: this part is now weird because `NewAccount` can carry more information (like some metadata) that you just disregard

mversic (changes requested):

mversic:
This is not how I would implement it. I would rather just control the access through executor as follows:
1. By default all instructions that target accounts are allowed:
* this covers **public use** 
* requires no executor (public case shouldn't require executor)
2. Custom executor rejects instructions that target accounts that have not yet been registered:
* this covers **private use** 
* should be part of our default executor

I also believe this approach reduces the implementation complexity

s8sato:
As I understand it, according to @mversic  your suggestion, the process when an account is targeted is as follows:

- public use
  - the target is immediately registered no matter what domain the target belongs to
    - may be reasonable, since the domain boundaries may not make sense in public chains
- private use
  - as before, only registered accounts can be valid targets

This means that no intermediate state is needed for account registration. If so, let's remove `is_active` flag and forget about recognize/activate terms.

Now, there are two possible approaches in terms of granularity of the account auto-registration policy:

1. core processes back to executor for auto-registering the target
   - executor is responsible for registering targeted accounts
     - returns uniform results regardless of authority of the instruction
   - chain level public/private characterization
   - needs a flag to indicate if `Register<Account>` is called back from core for auto-registration
2. each domain has a flag, and core determines whether to auto-register the target from its domain, without processing back to executor
   - core is responsible for registering targeted accounts
   - domain level public/private characterization
   - how should the flag be modified, if it should be?

Dealing with the concept of domains and public/private uses at the same time is new to me, and I seek further input

core/src/smartcontracts/isi/mod.rs
```diff
@@ -230,6 +230,28 @@ impl Execute for RevokeBox {
...
+
+    let account = Account::new(id.clone()).build(authority);
+    state_transaction
+        .world
+        .domain_mut(&id.domain_id)?
+        .add_account(account);
+
+    state_transaction
+        .world
+        .emit_events(Some(DomainEvent::Account(AccountEvent::Recognized(id))));
```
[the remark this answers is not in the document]
s8sato: The design required emitting this data event to trigger the activator

core/src/executor.rs
s8sato: Could you paraphrase, what seems like optional?

core/src/smartcontracts/isi/query.rs
s8sato: This pattern only appears in core unit tests, so I provided a utility in its `tests` module

mversic:
> * the target is immediately registered no matter what domain the target belongs to

wasn't that how you implemented, or was there something different?

> as before, only registered accounts can be valid targets

yes

> This means that no intermediate state is needed for account registration. If so, let's remove is_active flag and forget about recognize/activate terms.

yes

mversic:
> 1. core processes back to executor for auto-registering the target 

can't executor make sure account is registered before calling `isi.execute()` and handing over execution to the core

Erigara:
> I don't think there is any need to support inactive accounts in private blockchains. 

What about offline transactions? do you think they are required only in public case?

Is full registration even possible automatically? 

I mean that executor/trigger might have not enough data to perform registration (i.e. metadata)

Erigara:
Requirement for auto registration emerge from our implementation details (account structs holds assets).

Shouldn't we reconsider this? 

In this case we would allow isi where target account is not registered yet and if required we can prohibit this isi by checking if account is registered in the executor.

s8sato:
The feature might require the separate and relational entities after all.
As I understand it, an example where a new account "carol" comes to exercise its authority in public use is as follows:

1. the world has separate assets and accounts maps

    ```yml
    - world:
      - numeric_assets:
        - rose:
          - alice: 13
      - accounts:
        - alice:
          - metadata:
    ```

2. (public use) alice transfers 3 roses to carol, who is not registered yet

    ```yml
    - world:
      - numeric_assets:
        - rose:
          - alice: 10
          - carol:  3 # appeared
      - accounts: # not checked when carol is targeted
        - alice:
          - metadata:
    ```

3. (public use) carol registers herself

    ```yml
    - world:
      - numeric_assets:
        - rose:
          - alice: 10
          - carol:  3
      - accounts:
        - alice:
          - metadata:
        - carol: # registered
          - metadata:
    ```

4. carol successfully transfers 2 roses to alice

    ```yml
    - world:
      - numeric_assets:
        - rose:
          - alice: 12
          - carol:  1
      - accounts: # checked when carol is authority
        - alice:
          - metadata:
        - carol:
          - metadata:
    ```

I'll extend this idea to other ISIs and other entity maps and consider consistency.
For example, I can immediately see that `SetKeyValue<Account>` would be invalid before account registration

mversic:
> What about offline transactions? do you think they are required only in public case?

IMHO yes, but that should be confirmed. In private use-case it's expected users are registered before becoming entities in the network

s8sato:
So the idea is to deliberately introduce an inconsistency, the absence of a primary entity in `ACCOUNTS` in [this diagram](https://github.com/hyperledger/iroha/issues/4672#issuecomment-2201948148) while other storages referencing it, and thereby represent a pre-registration state of accounts. I'll convert this to draft (and maybe close it) and work on it after #4792

s8sato:
Closed as the issue should be addressed by a different approach such as https://github.com/hyperledger/iroha/pull/4668#issuecomment-2165073957

Issue #4698, opened 2024-06-06
Improve handling of task panics


0x009922:
Currently Iroha spawns many `tokio::task`s in a detached way, without checking whether they panic somewhere along execution.

Here are a few examples (just cases when the handle is completely ignored; there are many other when the handle is used, but not to check whether it resolves in an error):

https://github.com/hyperledger/iroha/blob/5085ff2435bf2e412dec76649b19d3a7bf091239/p2p/src/network.rs#L111

https://github.com/hyperledger/iroha/blob/f5e3c493a6f2f336da09d75c8d00562aac8168a5/cli/src/lib.rs#L387

https://github.com/hyperledger/iroha/blob/f5e3c493a6f2f336da09d75c8d00562aac8168a5/cli/src/lib.rs#L137

https://github.com/hyperledger/iroha/blob/f5e3c493a6f2f336da09d75c8d00562aac8168a5/cli/src/lib.rs#L494

It has a few issues:

- When some vital task panics, we don't even see that it happened, and don't see an error message.
- It results in a non-graceful "crumbling" of the system due to some tasks relying on failed ones.

For example, in case of the `NetworkBase` task panicking, we don't see anything except `NetworkBase must accept messages until there is at least one handle to it: SendError { .. }` (which doesn't tell the cause).

Instead, it would be good:

- To see panic messages of _most_ tasks (ideally - of all tasks)
- When a vital task panics, gracefully shutdown all other tasks
- _Ideally_ (overkill for now), adopt Supervision Tree design principle: monitor tasks execution in a centralised way with restart/shutdown strategies. This must contribute into overall fault-tolerance of Iroha significantly. For more info, see [Erlang/OTP](https://www.erlang.org/doc/system/design_principles.html) or [Supervisor in Elixir](https://hexdocs.pm/elixir/1.17.0-rc.1/Supervisor.html)

Tools to help: [`TaskTracker`](https://docs.rs/tokio-util/0.7.11/tokio_util/task/task_tracker/struct.TaskTracker.html#), [`JoinSet`](https://docs.rs/tokio/latest/tokio/task/join_set/struct.JoinSet.html), [`CancellationToken`](https://docs.rs/tokio-util/0.7.11/tokio_util/sync/struct.CancellationToken.html).

Erigara:
[...]

PR #4699, opened 2024-06-06
refactor(irohad): always start torii as task

### LINKED ISSUE #4562: Refactor Iroha - start `Torii` with all other tasks
I feel strange that torii is used to "block" iroha.

Imo it would be simpler design if iroha would always spawn `torii` in sparate task as any other service. 

And on iroha `start` we would just listen on `notify_sender`: 

```rust
async fn start(self) {
    self.notify_sender.notified().await    
}
```

This way iroha user is free to either block on this and wait on shutdown or start it as task.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/4456#discussion_r1580903592_

Erigara:
## Description

Simplify iroha startup + fix in integration test.

### Linked issue

Closes #4562 

### Benefits

Simpler code.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -55,18 +55,7 @@ pub mod samples;
 /// [Orchestrator](https://en.wikipedia.org/wiki/Orchestration_%28computing%29)
 /// of the system. It configures, coordinates and manages transactions
 /// and queries processing, work of consensus and storage.
-///
-/// # Usage
-/// Construct and then use [`Iroha::start_torii`] or [`Iroha::start_torii_as_task`]. If you experience
-/// an immediate shutdown after constructing Iroha, then you probably forgot this step.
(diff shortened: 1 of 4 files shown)


cli/src/lib.rs
```diff
@@ -384,81 +369,40 @@ impl Iroha<ToriiNotStarted> {
             metrics_reporter,
         );
 
+        tokio::spawn(async move {
+            torii
+                .start()
+                .await
+                .into_report()
+                .map_err(|report| report.change_context(StartError::StartTorii))
+        });
```
Erigara, line 378: This looks like should be part of #4698

0x009922 (dismissed):

nxsaken (dismissed):

nxsaken (approved):

0x009922 (approved):

PR #4714, opened 2024-06-10
refactor!: remove `_id` suffix for entity ids in public API

mversic:
## Description

Remove `_id` suffix from tokens, config files and data model

### Linked issue

Closes #{issue_number} 

### Checklist


nxsaken (approved):

scripts/test_env.py
```diff
@@ -247,7 +247,7 @@ def setup(args: argparse.Namespace):
 
 def cleanup(out_dir: pathlib.Path):
     logging.info("Killing peer processes...")
-    subprocess.run(["pkill", "-9", "irohad"])
+    subprocess.run(["pkill", "-9", "iroha"])
```
0x009922, line 250: Is it right?

smart_contract/executor/src/default.rs
```diff
@@ -251,7 +251,7 @@ pub mod domain {
         authority: &AccountId,
         isi: &SetKeyValue<Domain>,
     ) {
-        let domain_id = isi.object_id();
+        let domain_id = isi.object();
```
0x009922, line 254: Although IMO it's totally fine to not rename local variables.

```suggestion
        let domain = isi.object();
```

0x009922 (approved):

scripts/test_env.py
mversic: it is because processes are named `iroha0` etc

smart_contract/executor/src/default.rs
mversic: ah, I didn't go renaming local variables because that would sometimes conflict with entity names

PR #4718, opened 2024-06-11
refactor: migrate to axum

### LINKED ISSUE #3776: [refactor] Migrate from warp to axum/actix
### Feature request

Replace `warp` crate to a more batteries included kind of web framework, main candidates being `axum` or `actix`. Do some researching/benchmark to determine what's better, if possible.

### Motivation

Currently the `warp` crate is used throughout the `cli` and `version` parts of the codebase, but it's been known as a pain point for a while with unwieldy filter type signatures occupying the whole screen and awkward proc macro metaprogramming to reach our goals.

### Who can help?

@appetrosyan

Erigara:
## Description

Migrate from `warp` to `axum`.

With current implementation i tried to preserve the same behavior as before.

I would suggest further improvements in the separate PRs.

### Linked issue

Closes #3776  

### Benefits

Bring more simplicity to torii.


torii/src/utils.rs
```diff
@@ -37,48 +24,73 @@ impl<T: Encode + Send> Reply for Scale<T> {
...
-impl<O: Reply, E: Reply> Reply for WarpResult<O, E> {
-    fn into_response(self) -> warp::reply::Response {
-        match self {
-            Self(Ok(ok)) => ok.into_response(),
-            Self(Err(err)) => err.into_response(),
+        async fn from_request(req: Request, state: &S) -> Result<Self, Self::Rejection> {
+            let (mut parts, body) = req.into_parts();
+            let cursor = Query::<ForwardCursor>::from_request_parts(&mut parts, &state)
+                .await
+                .map(|Query(cursor)| ClientQueryRequest::cursor(cursor));
```
Erigara, line 85: I don't think it's a right thing to extract cursor from query params, since anyone can fetch cursor.
So it should be signed and passed as body.

0x009922: Created https://github.com/hyperledger/iroha/issues/4720

torii/src/lib.rs
```diff
@@ -92,171 +95,171 @@ impl Torii {
...
-                .or(post_router)
-                .with(warp::trace::request()))
+            ;
+
+        router.layer((
+            TraceLayer::new_for_http()
+                .make_span_with(DefaultMakeSpan::default().include_headers(true)),
+            // Graceful shutdown will wait for outstanding requests to complete.
+            // Add a timeout so requests don't hang forever.
+            TimeoutLayer::new(Duration::from_secs(60)),
```
0x009922, line 236: Please define a `const`

```suggestion
            TimeoutLayer::new(SERVER_SHUTDOWN_TIMEOUT),
```

torii/src/lib.rs
```diff
@@ -92,171 +95,171 @@ impl Torii {
...
-                //        handle this properly, report address origin after Axum
-                //        migration: https://github.com/hyperledger/iroha/issues/3776
-                let (_, serve_fut) =
-                    warp::serve(api_router).bind_with_graceful_shutdown(addr, signal_fut);
 
-                task::spawn(serve_fut)
+                let signal = async move { torii.notify_shutdown.notified().await };
+
+                let serve_fut = async move {
+                    let listener = TcpListener::bind(addr).await?;
```
0x009922, line 256: Here, it is possible to attach the configuration source of the address to the error.

0x009922 (commented):
Finally!

torii/src/lib.rs
Erigara: What do you mean exactly by "attach the configuration source of the address"? And how to do that?

Erigara:
> But Torii tests are failing.

Tests are fixed + added extra test for different wrong body errors, the problem was mainly because pytests didn't set proper headers.

Erigara:
@AlexStroke please take a look at torii pytests

torii/src/lib.rs
0x009922: I mean to return a report with an information on where the `addr` came from. For example:

https://github.com/hyperledger/iroha/blob/056cae387d51d120264facaf8bc0c13c1d671784/cli/src/lib.rs#L662-L668

Erigara: Fixed, `torii` now exit early if fails to bind address

```plain
Error: Could not start Iroha
│
├─▶ Unable to start Torii (Iroha HTTP API Gateway)
│
╰─▶ Address already in use (os error 48)
    ╰╴config origin: parameter `torii.address` with value `127.0.0.1:8080` in file `test/peers/iroha0/config.toml`
```

Erigara:
Ignore first commit, it's extracted in separate PR and here only to make unit tests work

torii/src/lib.rs
```diff
@@ -85,177 +94,182 @@ impl Torii {
...
+                        ))
+                    }
+                }),
+            )
+            .route(
+                uri::STATUS,
+                get({
+                    let metrics_reporter = self.metrics_reporter.clone();
+                    move |headers: axum::http::header::HeaderMap| {
+                        let accept = headers.get(axum::http::header::ACCEPT);
```
DCNick3, line 136: I see this is repeated several times.. Maybe it's worth making an `Accept` [extractor](https://docs.rs/axum/latest/axum/extract/index.html#defining-custom-extractors)?

UPD: or we can go one step further and make something akin to https://docs.rs/axum-content-negotiation/ for json/scale responses as I remember being discussed some time ago, though that is probably for a separate PR

torii/src/lib.rs
```diff
@@ -85,177 +94,182 @@ impl Torii {
...
+                get({
+                    let metrics_reporter = self.metrics_reporter.clone();
+                    move || core::future::ready(routing::handle_metrics(&metrics_reporter))
+                }),
+            )
+            .route(
+                uri::API_VERSION,
+                get({
+                    let state = self.state.clone();
+                    move || routing::handle_version(state)
```
DCNick3, line 152: Hm, so instead of using [`State`](https://docs.rs/axum/latest/axum/extract/struct.State.html) extractor you just keep the required state bundled with the handler? Don't really see a problem with that, just never considered that you can pass closures to routes instead of bare functions.

torii/src/lib.rs
```diff
@@ -85,177 +94,182 @@ impl Torii {
...
+                }
+            }))
+            .route(uri::BLOCKS_STREAM,
+                post({
+                    let kura = self.kura.clone();
+                    move |ws: WebSocketUpgrade| {
+                        core::future::ready(
+                            ws.on_upgrade(|ws| async move {
+                                if let Err(error) = routing::handle_blocks_stream(kura, ws).await {
+                                    iroha_logger::error!(%error, "Failed to subscribe to blocks stream");
```
DCNick3, line 229: nit: This error can be logged not only due to "failure to subscribe", but also due to some error happening when handling the subscription (the future lives as long as the client is interested in the stream).

Cargo.toml
```diff
@@ -117,10 +116,10 @@ getset = "0.1.2"
 hex-literal = "0.4.1"
 
 rand = { version = "0.8.5", default-features = false, features = ["getrandom", "alloc"] }
-warp = { version = "0.3.7", default-features = false }
+axum = { version = "0.7.5", default-features = false }
 wasmtime = "15.0.1"
 
-tracing = "0.1.40"
+tracing = { version = "0.1.40", features = ["log"] }
```
DCNick3, line 122: Is this an intended change? AFAIK, it enables some adapter subscribers bridging between `log` and `tracing` infrastructure, but I don't see them being used anywhere..

torii/src/lib.rs
Erigara: Yeah, it's described as one [method](https://docs.rs/axum/latest/axum/index.html#using-closure-captures) to share state with handler.

Cargo.toml
Erigara: it's used in `config/src/lib.rs`, i  think it was silently enabled by `warp`

DCNick3: That's right, it was https://github.com/seanmonstar/warp/blob/ce8114b50cadedac089d496235fe9d18596f944e/Cargo.toml#L37.

torii/src/lib.rs
Erigara: i've added `ExtractAccept` header, would like to handle content negotiation in separate PR

torii/src/utils.rs
```diff
@@ -37,48 +24,97 @@ impl<T: Encode + Send> Reply for Scale<T> {
...
+            let body = Bytes::from_request(req, state)
+                .await
+                .map_err(IntoResponse::into_response)?;
+
+            T::decode_all_versioned(&body)
+                .map(ScaleVersioned)
+                .map_err(|err| {
+                    (
+                        axum::http::StatusCode::BAD_REQUEST,
+                        format!("Transaction Rejected (Malformed), Reason : '{err}'"),
```
DCNick3, line 61: Not sure it should say "transaction" here. AFAIU, it's also used for queries too..

```suggestion
                        format!("Could not decode request: {err}"),
```

DCNick3 (dismissed):

DCNick3 (approved):

0x009922 (approved):

DCNick3 (dismissed):

0x009922 (approved):

DCNick3 (approved):

0x009922 (approved):

PR #4719, opened 2024-06-11
feat(swarm): reimplement, optimize compose config and its generation

nxsaken:
## Description

Reimplement most of Swarm generation logic, optimize the generated Compose config.

### Benefits

- declutters schema code
- removes almost all clones
- optimizes Compose config: builds/pulls image only once instead of per peer (this has caused sigkills during builds)
- allows writing to any `std::io::Write` instead of just files
  - adds an option to print the config to stdout (used in CI to check config updates)

### Checklist

- [x] I've written unit tests for the code changes=
- [x] Update usage of Swarm in CI


configs/swarm/docker-compose.yml
```diff
@@ -1,46 +1,60 @@
-# This file is generated by iroha_swarm.
-# Do not edit it manually.
+# target/release/iroha_swarm -p 4 -s xyzzy -c ./configs/swarm -i hyperledger/iroha:dev -o ./configs/swarm/docker-compose.yml -HF
```
0x009922, line 1: There was a discussion about putting all parameters into the banner: https://github.com/hyperledger/iroha/pull/3872#discussion_r1318327619

I am not sure if it is a good idea still, as it depends on CWD and contains `target/release/iroha_swarm`, which is environment-dependent.  Option `-F` (`--force`) also doesn't make sense here.

configs/swarm/docker-compose.yml
```diff
@@ -1,46 +1,60 @@
...
       interval: 2s
       timeout: 1s
       retries: 30
       start_period: 4s
+    command: |-
+      /bin/sh -c "
+        kagami genesis sign /config/genesis.json \
+          --public-key ed012018CC218A5830112811C85287E43CEE37A81CBACEA72CFAB6A855E4A89BD60A91 \
+          --private-key 802640268A94EC4A0FD1BEEC8ABD0A6CC184BCC5090698ADFF08758016D509F75FB6EC18CC218A5830112811C85287E43CEE37A81CBACEA72CFAB6A855E4A89BD60A91 \
+          --out-file /tmp/genesis.signed.scale && \
```
0x009922, line 42:
```suggestion
          --out-file $GENESIS_SIGNED_FILE && \
```

configs/swarm/docker-compose.yml
```diff
@@ -1,46 +1,60 @@
-# This file is generated by iroha_swarm.
-# Do not edit it manually.
+# target/release/iroha_swarm -p 4 -s xyzzy -c ./configs/swarm -i hyperledger/iroha:dev -o ./configs/swarm/docker-compose.yml -HF
+# 
+# This configuration has been generated by `iroha_swarm` using the above settings.
+# You should not edit this manually.
 
-version: '3.8'
 services:
+  image_provider:
```
0x009922, line 7: Is image provider needed when irohad services have `image`, not `build`?

tools/swarm/src/main.rs
```diff
@@ -1,63 +1,218 @@
...
+    ///
+    /// If the file exists, the app will prompt its overwriting. If the TTY is not
+    /// interactive, the app will stop execution with a non-zero exit code.
+    /// To overwrite the file anyway, pass the `--force` flag.
+    #[arg(long, short, value_name = "FILE")]
+    out_file: std::path::PathBuf,
+    /// Print the generated configuration to stdout
+    /// instead of writing it to the target file.
+    #[arg(long, short = 'P', conflicts_with = "force")]
+    print: bool,
```
0x009922, line 58: It also conflicts with `out_file`, doesn't it?

tools/swarm/src/main.rs
```diff
@@ -1,63 +1,218 @@
...
+    /// Overwrite the target file if it already exists.
+    #[arg(long, short = 'F')]
+    force: bool,
+    /// Do not include the banner with the generation notice in the file.
+    ///
+    /// The banner includes the passed arguments in order to help with reproducibility.
+    #[arg(long, short = 'B')]
+    no_banner: bool,
```
0x009922, line 66: - `-B` alias for `--no-banner` option is confusing, imo
- Probably not a very common option to use to have an alias for it

```suggestion
    /// Do not include the banner with the generation notice in the file.
    ///
    /// The banner includes the passed arguments in order to help with reproducibility.
    #[arg(long)]
    no_banner: bool,
```

tools/swarm/src/main.rs
```diff
@@ -1,63 +1,218 @@
...
+
+    let (mut stdout, mut file);
+
+    let writer: &mut dyn std::io::Write = if args.print {
+        stdout = std::io::stdout();
+        &mut stdout
+    } else {
+        file = std::fs::File::create(&args.out_file)?;
+        &mut file
     };
```
0x009922, line 128: That's cool

tools/swarm/src/lib.rs
0x009922, line 1: I am not in favor of combining `lib.rs` with `main.rs` in a single crate.

tools/swarm/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+        .build()
+        .write(&mut buffer, None)
+        .unwrap();
+        let actual = std::str::from_utf8(&buffer).unwrap();
+        assert_yml_eq(EXPECTED, actual);
+    }
+
+    #[test]
+    fn multiple_pull_healthcheck_nocache() {
+        const EXPECTED: &str = include_str!("../samples/multiple-pull-healthcheck-nocache.yml");
```
0x009922, line 229: Having separate `yml` files with fixtures is convenient from files structuring point, but it forces to manually update them whenever they are updated. The reason I used `expect_test` is that it can rewrite the snapshots automatically with `UPDATE_EXPECT=1` env variable.

Actally, just discovered existence of [`expect_test::expect_file!`](https://docs.rs/expect-test/latest/expect_test/macro.expect_file.html) macro. I would prefer if you use it.

tools/swarm/src/main.rs
```diff
@@ -1,63 +1,218 @@
...
+    /// Always pull or rebuild the image even if it is cached locally.
+    #[arg(long, short = 'C')]
+    no_cache: bool,
+    /// Path to the target Compose configuration file.
+    ///
+    /// If the file exists, the app will prompt its overwriting. If the TTY is not
+    /// interactive, the app will stop execution with a non-zero exit code.
+    /// To overwrite the file anyway, pass the `--force` flag.
+    #[arg(long, short, value_name = "FILE")]
+    out_file: std::path::PathBuf,
```
0x009922, line 54: If swarm supports both file output and STDOUT, I think it would be more CLI-idiomatic to use `--output` for file output and `-` for STDOUT. Not sure how to achieve it with clap.

0x009922:
@BAStos525, could you take a look at new compose files and check whether they work as expected in CI/CD?

tools/swarm/src/main.rs
nxsaken: No, we still need this to resolve paths. Maybe renaming this to `--target` would be clearer

tools/swarm/src/lib.rs
nxsaken: I will move the binary to `src/bin/cli.rs`

configs/swarm/docker-compose.yml
nxsaken: What if the cwd is included relative to the swarm binary?

```
# ../../..
# target/release/iroha_swarm ...
#
# [...] from the above working directory and settings
```

nxsaken: I'll probably change this to only include the seed as the first line

configs/swarm/docker-compose.yml
nxsaken: Not really, I see how this can be simplified.

tools/swarm/src/lib.rs
nxsaken: Going back to using `expect!` because `expect_file!` doesn't support custom asserts, and the files will break on formatting due to the way `serde_yaml` serializes lists.

configs/swarm/docker-compose.yml
nxsaken: Apparently, Docker Compose doesn't recognize variables from the `environment` field, expecting them to be set in the host environment.

BAStos525:
> @BAStos525, could you take a look at new compose files and check whether they work as expected in CI/CD?

Add `hyperledger/iroha:local` tag [here](https://github.com/hyperledger/iroha/blob/86d3cbc54d6e5d9c012653dfedade3dcc7f6c935/.github/workflows/iroha2-dev.yml#L29)

configs/swarm/docker-compose.local.yml
```diff
@@ -1,46 +1,57 @@
...
-      "
     healthcheck:
       test: test $(curl -s http://127.0.0.1:8080/status/blocks) -gt 0
       interval: 2s
       timeout: 1s
       retries: 30
       start_period: 4s
+    command: |-
+      /bin/sh -c "
+        kagami genesis sign /config/genesis.json --public-key $$GENESIS_PUBLIC_KEY --private-key 80264082B3BDE54AEBECA4146257DA0DE8D59D8E46D5FE34887DCD8072866792FCB3AD4164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4 --out-file $$GENESIS_SIGNED_FILE &&
```
BAStos525, line 39: Could we move --private-key value to env var for convince?

nxsaken: Yes, but I wonder if there is a reason why it hasn't been done earlier.
@dima74 wdyt?

dima74: It can be used to ensure that `irohad` doesn't use genesis private key. I don't mind moving it to env var

nxsaken: Thank you, I'll add it to the genesis env only.

tools/swarm/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+        .unwrap()
+        .build()
+        .write(&mut buffer, banner)
+        .unwrap();
+        String::from_utf8(buffer).unwrap()
+    }
+
+    #[test]
+    fn single_build_banner() {
+        expect_test::expect!([r##"
```
0x009922, line 177: `expect_file!` didn't work? I think it could be better to move these snippets to dedicated files

0x009922 (dismissed):

tools/swarm/src/lib.rs
nxsaken: Yeah, the yaml generator doesn't indent lists. They get properly indented on reformatting if placed in a file, breaking the tests. https://github.com/hyperledger/iroha/pull/4719#discussion_r1637611496

configs/swarm/docker-compose.local.yml
dima74: Note that genesis private key was moved to env var in #4739 (and there are some other changes to docker-compose files)

tools/swarm/src/lib.rs
0x009922: Ah, okay :(

BAStos525 (approved):

PR #4725, opened 2024-06-13
fix: Make `iroha_smart_contract_utils` `log` and `dbg` functions work outside of wasm

### LINKED ISSUE #4721: [BUG] linking fails on `cargo test -p iroha_executor --features debug`
### OS and Environment

macOS M1, happens in CI also (Ubuntu)

### GIT commit hash

started in 7a41ae8, still going as of 86d3cbc

`cargo test -p iroha_executor --features debug`

### Actual result

```
error: linking with `cc` failed: exit status: 1
  = note: Undefined symbols for architecture arm64:
            "_dbg", referenced from:
                iroha_smart_contract_utils::debug::dbg::h090653f2ad4ed03b in libiroha_smart_contract-dd8fcf4b8ebee85c.rlib[15](iroha_smart_contract-dd8fcf4b8ebee85c.2kke7rkpsclar55l.rcgu.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Expected result

Builds and runs tests normally

### Logs

Output from one of the failed jobs: https://github.com/hyperledger/iroha/actions/runs/9481834871/job/26125652260#step:5:1050

@mversic

### Notes

_No response_

DCNick3:
## Description

This PR adds support for `iroha_smart_contract_utils`'s `log` and `dbg` functions to be used outside of wasm. This allows writing tests to be ran against the host target and to still use those wasm-specific functions.

Both implementations just print the requested object to stderr.

As a drive-by, this also makes sure to run wasm tests from `iroha_smart_contract_utils`, which was omitted in CI for some time.

### Linked issue

Fixes #4721

### Benefits

- [ ] make CI pass

### DIFF
--- a/smart_contract/utils/src/debug.rs
+++ b/smart_contract/utils/src/debug.rs
@@ -4,6 +4,9 @@
 use alloc::format;
 use core::fmt::Debug;
 
+use cfg_if::cfg_if;
+
+#[cfg(target_family = "wasm")]
 #[cfg(not(test))]
 mod host {
     #[cfg(feature = "debug")]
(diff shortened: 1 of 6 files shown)


smart_contract/utils/src/debug.rs
```diff
@@ -23,18 +26,23 @@ mod host {
 /// Print `obj` in debug representation to the stdout.
 ///
 /// Do nothing if `debug` feature is not specified
```
Erigara, line 28: Docs are not correct now, because we print in any case if `target_family = "wasm"`

Erigara (approved):

smart_contract/utils/src/debug.rs
DCNick3: True, I'll update those

Erigara (approved):

mversic (approved):

PR #4732, opened 2024-06-14
refactor: Make the metadata dynamic via implementing JsonString

### LINKED ISSUE #4353: dynamic metadata value

Stukalov-A-M:
## Description

* Removed MetadataValueBox
* Added JsonString

### Linked issue

* Closes #4353  

### Checklist


client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -43,7 +43,9 @@ fn main(_id: TriggerId, _owner: AccountId, _event: EventBox) {
         )
         .parse()
         .dbg_unwrap();
-        metadata.insert_with_limits(name, true, limits).dbg_unwrap();
+        metadata
+            .insert_with_limits(name, JsonString::new(&true), limits)
```
mversic, line 47:
```suggestion
            .insert_with_limits(name, JsonString::new(&true), limits)
```

the constructor shouldn't take a reference to a `Copy` type

client/tests/integration/sorting.rs
```diff
@@ -55,7 +55,7 @@ fn correct_pagination_assets_after_creating_new_one() {
         asset_metadata
             .insert_with_limits(
                 sort_by_metadata_key.clone(),
-                i as u32,
+                JsonString::new(&i),
```
mversic, line 58:
```suggestion
                JsonString::new(&i),
```

same, reference to a `Copy` type

data_model/src/metadata.rs
```diff
@@ -378,18 +196,21 @@ impl Metadata {
     pub fn insert_with_limits(
         &mut self,
         key: Name,
-        value: impl Into<MetadataValueBox>,
+        value: JsonString,
```
mversic, line 199:
```suggestion
        value: impl Into<JsonString>,
```

and add different implementations of `From<JsonString>`. This will look much nicer for the client

mversic (changes requested):

data_model/src/metadata.rs
mversic: you will be able to replace every `JsonString::new(x)` with just `x` in function calls

primitives/src/lib.rs
```diff
@@ -10,13 +10,13 @@
 
 #[cfg(not(feature = "std"))]
 extern crate alloc;
-
```
mversic, line 13: it would be nice to keep this whitespace :)

primitives/src/json.rs
```diff
@@ -0,0 +1,166 @@
...
+use serde_json::Value;
+
+#[allow(missing_docs)]
+#[derive(Debug, Clone, Eq, Encode, Decode)]
+pub struct JsonString(pub String);
+
+/// A candidate for a valid `JsonString`.
+/// Is used for generalizing ser/de any types to `JsonString` and vise versa
+#[derive(Serialize, Deserialize, Clone)]
+pub struct JsonCandidate<T>(pub T);
```
mversic, line 28: make it private. We usually put it into `mod candidate`. Take a look at `block.rs` for example

smart_contract/executor/src/permission.rs
```diff
@@ -117,7 +117,7 @@ pub mod asset_definition {
 
     /// Check if `authority` is the owner of asset definition
 
-    /// `authority` is owner of asset definition if:
+    /// `authority` is owner of `asset_definition` if:
```
mversic, line 120:
```suggestion
    /// `authority` is owner of asset definition if:
```

primitives/src/json.rs
mversic: I can see you're using it in one test, but you really shouldn't

primitives/Cargo.toml
```diff
@@ -42,6 +42,8 @@ smallstr = { version = "0.3.0", default-features = false, features = ["serde", "
 thiserror = { workspace = true, optional = true }
 displaydoc = { workspace = true }
 parking_lot = { workspace = true, optional = true }
+serde_json = { workspace = true }
```
mversic, line 45: does this mean that this dependency can be removed from some crates, like `iroha_data_model`?

smart_contract/executor/src/permission.rs
Stukalov-A-M: Clippy proposed that :sassy_man:

mversic: i know

data_model/src/metadata.rs
Stukalov-A-M: > and add different implementations of `From<JsonString>`. This will look much nicer for the client

I did it before, but then I & @Erigara decided to reduce the code by generalizing via TryFrom<JsonString> for JsonCandidate<T> because serde_json may throw an error.

mversic: implement for the ones that are infallible. For example, you can implement `From<bool> for JsonString`, you can also implement for integers

nxsaken:
Please use the imperative mood in commit messages / PR titles to keep it consistent

primitives/src/json.rs
Stukalov-A-M: > make it private. We usually put it into `mod candidate`. Take a look at `block.rs` for example

I've made it private, but there is no need to put it into separate module

client_cli/src/main.rs
```diff
@@ -1110,12 +1118,13 @@ mod tests {
...
-        case!("false", false.into());
+        case!("true", JsonString::new(true));
+        case!("false", JsonString::new(false));
 
         // Numeric values
-        case!("123", numeric!(123).into());
-        case!("123.0", numeric!(123.0).into());
+        // ToDo: Bug with Json serialization. Hasn't covered by CI
+        case!("123", JsonString::new(numeric!(123)));
+        case!("123.0", JsonString::new(numeric!(123.0)));
```
Stukalov-A-M, line 1127: Researching how to fix

client/tests/integration/permissions.rs
```diff
@@ -313,16 +313,17 @@ fn stored_vs_granted_token_payload() -> Result<()> {
...
 
     // Allow alice to mint mouse asset and mint initial value
+    // Introducing some whitespaces
+    // This way, if the executor compares just JSON strings, this test would fail
+    let value_json: Value = serde_json::de::from_str(
+        &*r#"{ "asset"    :   "xor#wonderland#{id}" }"#
+            .replace("{id}", mouse_id.to_string().as_str()),
+    )
+    .unwrap();
+
```
mversic, line 323: this is still not good because in the parsed form you've removed the whitespaces

client/tests/integration/roles.rs
```diff
@@ -242,7 +242,7 @@ fn grant_revoke_role_permissions() -> Result<()> {
     let set_key_value = SetKeyValue::account(
         mouse_id.clone(),
         Name::from_str("key").expect("Valid"),
-        "value".to_owned(),
+        JsonString::from_str("value")?,
```
mversic, line 245:
```suggestion
        "value".parse()?,
```

client/tests/integration/roles.rs
```diff
@@ -83,7 +83,7 @@ fn register_and_grant_role_for_metadata_access() -> Result<()> {
     let set_key_value = SetKeyValue::account(
         mouse_id,
         Name::from_str("key").expect("Valid"),
-        "value".to_owned(),
+        JsonString::from_str("value")?,
```
mversic, line 86:
```suggestion
        "value".parse()?,
```

client/tests/integration/queries/smart_contract.rs
```diff
@@ -33,7 +33,7 @@ fn live_query_is_dropped_after_smart_contract_end() -> Result<()> {
         client.account.clone(),
         Name::from_str("cursor").unwrap(),
     ))?;
-    let cursor: String = metadata_value.try_into()?;
+    let cursor: String = metadata_value.0;
```
mversic, line 36: this field should be private

client/tests/integration/smartcontracts/query_assets_and_save_cursor/src/lib.rs
```diff
@@ -47,7 +43,7 @@ fn main(owner: AccountId) {
     SetKeyValue::account(
         owner,
         parse!(Name, "cursor"),
-        MetadataValueBox::String(
+        JsonString(
```
mversic, line 46: you must use some constructor

client/tests/integration/queries/smart_contract.rs
mversic: this is allowed only because our current testing infra includes `transparent_api`. This should and will be disallowed

client/tests/integration/triggers/time_trigger.rs
```diff
@@ -189,7 +189,7 @@ fn pre_commit_trigger_should_be_executed() -> Result<()> {
         let sample_isi = SetKeyValue::account(
             account_id.clone(),
             "key".parse::<Name>()?,
-            String::from("value"),
+            JsonString::from_str("value")?,
```
mversic, line 192: please use `parse`

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+/// A candidate for a valid `JsonString`.
+/// Is used for generalizing ser/de any types to `JsonString` and vise versa
+#[derive(Serialize, Deserialize, Clone)]
+struct JsonCandidate<T>(T);
+
+impl JsonString {
+    /// Construct [`JsonString`]
+    pub fn new<T: Serialize>(payload: T) -> Self {
+        JsonCandidate(payload).try_into().unwrap()
+    }
```
mversic, line 38: doesn't clippy complain that you have no `# Panics` in the documentation? Also, I don't approve of a panicking constructor, but other may disagree

client/tests/integration/roles.rs
Stukalov-A-M: Is it really necessary? I mean, I can implement From<&str>, but why if we have FromStr implementation? Plus imo current realization is more explicit.

mversic: You have it already implemented. It's just a different form. You used it elsewhere

Stukalov-A-M: It doesn't work with an explicit type definition. So I see no difference what to use: either `"s".parse<JsonString>()? or `JsonString::from_str("s")`
image

primitives/src/json.rs
Stukalov-A-M: Will try to figure out something

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    /// Tries cast [`JsonString`] to any value.
+    ///
+    /// # Errors
+    /// - if String == invalid Json type
+    pub fn try_into_any<T: DeserializeOwned>(&self) -> Result<T, serde_json::Error> {
```
mversic, line 44: I get that you can't make a blanket impl but I don't like API like this. Let's consider if we have some other options?

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl PartialOrd for JsonString {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for JsonString {
```
mversic, line 106: is this `Ord` implementation congruent with `PartialEq` implementation?

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn try_into_any<T: DeserializeOwned>(&self) -> Result<T, serde_json::Error> {
+        serde_json::from_str(&self.0)
+    }
+}
+impl PartialEq for JsonString {
+    fn eq(&self, other: &Self) -> bool {
+        serde_json::from_str::<Value>(&self.0).unwrap()
+            == serde_json::from_str::<Value>(&other.0).unwrap()
+    }
+}
```
mversic, line 53: IMO you can derive this. I think there is no point of having a special `PartialEq` if it can't be for all scenarios. The scenarios for which it fundamentally can't be implemented is `rose##alice@wonderland` vs `rose#wonderland@alice#wonderland`. How about you derive it and we open an issue to discuss whether we want to prevent. Whatever you decide, please open an issue so we can discuss the implementation of `Eq` and `Ord` for `JsonString`

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let json: Value = serde_json::from_str(&self.0).map_err(serde::ser::Error::custom)?;
+        json.serialize(serializer)
+    }
+}
+
+impl iroha_schema::TypeId for JsonString {
```
mversic, line 75: you can derive this

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn id() -> iroha_schema::Ident {
+        "JsonString".to_owned()
+    }
+}
+
+impl IntoSchema for JsonString {
```
mversic, line 81: can't this be derived also?

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+use serde_json::Value;
+
+#[allow(missing_docs)]
+#[derive(Debug, Clone, Eq, Encode, Decode)]
+pub struct JsonString(pub String);
+
+/// A candidate for a valid `JsonString`.
+/// Is used for generalizing ser/de any types to `JsonString` and vise versa
+#[derive(Serialize, Deserialize, Clone)]
+struct JsonCandidate<T>(T);
```
mversic, line 32: please create a `mod candidate` and put all candidate related stuff inside. It's a sort of convention we have

client/tests/integration/roles.rs
mversic: > So I see no difference what to use: either `"s".parse<JsonString>()?` or `JsonString::from_str("s")`

The difference is that for the 1st case you don't need to bring in `FromStr` trait into the scope. Since our tests are also our documentation this makes it simpler

Stukalov-A-M: > > So I see no difference what to use: either `"s".parse<JsonString>()?` or `JsonString::from_str("s")`
> 
> The difference is that for the 1st case you don't need to bring in `FromStr` trait into the scope. Since our tests are also our documentation this makes it simpler

Haven't known that, thanks!

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+        let l: Value = serde_json::from_str(&self.0).unwrap();
+        let r: Value = serde_json::from_str(&other.0).unwrap();
+
+        match (l, r) {
+            (Value::Number(left), Value::Number(right)) => {
+                if left.is_u64() && right.is_u64() {
+                    left.as_u64().cmp(&right.as_u64())
+                } else {
+                    left.as_f64().unwrap().total_cmp(&right.as_f64().unwrap())
+                }
```
Erigara, line 117: I'm afraid this part might panic where we don't want to.

primitives/src/json.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn from(value: f64) -> Self {
+        JsonString::new(value)
+    }
+}
+
+impl From<bool> for JsonString {
+    fn from(value: bool) -> Self {
+        JsonString::new(value)
+    }
+}
```
Erigara, line 168: Can't we use single implementation instead?

This way we can avoid panicking constructor.

```
impl<T: Serialize> From<T> for JsonString {
    // ...
}
```

Stukalov-A-M: > Can't we use single implementation instead?
> 
> This way we can avoid panicking constructor.
> 
> ```
> impl<T: Serialize> From<T> for JsonString {
>     // ...
> }
> ```

No, we can not. It's prohibited to implement generic because of the trait implementation conflict

primitives/src/json.rs
Stukalov-A-M: I can add if is_f64 check. Then there should be no panic variants

client_cli/src/main.rs
```diff
@@ -61,20 +62,27 @@ pub struct MetadataValueArg {
...
 impl FromStr for MetadataValueArg {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        s.parse::<bool>()
-            .map(MetadataValueBox::Bool)
-            .or_else(|_| s.parse::<Numeric>().map(MetadataValueBox::Numeric))
-            .or_else(|_| serde_json::from_str::<MetadataValueBox>(s).map_err(Into::into))
-            .map(|value| MetadataValueArg { value })
+        match Value::from_str(s)? {
```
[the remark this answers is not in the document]
Stukalov-A-M: > create JsonString directly from String to have extra whitespaces

My main idea is to forbid creatin `JsonString` from a raw string to avoid invalid Jsons. Only `new(), from_str(), parse::<JsonString>()` are allowed because they implement ser/de chain "str->Value->String"

primitives/src/json.rs
mversic: the problem there was that `Ord` implementation was not consistent with `Eq` implementation. They can compare just as strings, but they have to both do so

client/tests/integration/permissions.rs
Stukalov-A-M: > create `JsonString` directly from `String` to have extra whitespaces

My main idea is to forbid creatin JsonString from a raw string to avoid invalid Jsons. Only new(), from_str(), parse::<JsonString>() are allowed because they implement ser/de chain "str->Value->String"

primitives/src/json.rs
Erigara: Which traits are in conflict?

Erigara: Conflicting with `impl From<JsonString> for JsonString`... because `JsonString` itself implement `Serialize`

primitives/src/json.rs
Erigara: What are other options? Calling unwrap in every single case won't be convenient and not sure if it would have any benefits.

Currently we only can fail in the following [case](https://docs.rs/serde_json/latest/serde_json/fn.to_string.html):
> Serialization can fail if T’s implementation of Serialize decides to fail, or if T contains a map with non-string keys.

primitives/src/json.rs
Erigara: When we discussed this with @Stukalov-A-M there we not much other options.

1. `impl<T: DeserializeOwned> TryFrom<JsonString> for T` is not possible due to `T` non-local type
2. `impl<T: DeserializeOwned> TryInto<T> for JsonString` has conflicts
3. we might introduce additional trait like `MyTryFrom` but this would require to import this trait (not sure about benefits)
4. introduce helper `Wrapper<T>` and implement `impl<T: DeserializeOwned> TryFrom<JsonString> for Wrapper<T>` (again seems not that usable)
5. we might have direct method (this case)

mversic: I was thinking it should be th other way around. Every type that wants to be parseable from `JsonString` should implement `TryFrom<JsonString>`. It should be responsibility of the type not of `JsonString`. But that may be inconvenient?

Erigara: It would be a lot of repetition + we can't access private private fields of the `JsonString` to implement `TryFrom`

client/tests/integration/permissions.rs
Stukalov-A-M: Returned all back

primitives/Cargo.toml
Stukalov-A-M: Not yet. data-model still has a lot of dependencies on serde_json

primitives/src/json.rs
Stukalov-A-M: @mversic , @Erigara , yep, I couldn't find any other options

primitives/src/json.rs
mversic: @Stukalov-A-M it looks to me as if `Eq` and `Ord` are not incongruent

mversic (changes requested):

client/tests/integration/permissions.rs
```diff
@@ -334,7 +333,11 @@ fn stored_vs_granted_token_payload() -> Result<()> {
         .expect("Failed to grant permission to alice.");
 
     // Check that alice can indeed mint mouse asset
-    let set_key_value = SetKeyValue::asset(mouse_asset, Name::from_str("color")?, "red".to_owned());
+    let set_key_value = SetKeyValue::asset(
+        mouse_asset,
+        Name::from_str("color")?,
```
mversic, line 338:
```suggestion
        "color".parse::<Name>()?,
```

primitives/src/json.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn from(value: &Value) -> Self {
+        JsonString(value.to_string())
+    }
+}
+
+impl From<Value> for JsonString {
+    fn from(value: Value) -> Self {
+        JsonString(value.to_string())
+    }
+}
```
mversic, line 129: IMO, I would keep only one of these

core/src/smartcontracts/isi/mod.rs
```diff
@@ -285,25 +284,15 @@ mod tests {
         let account_id = ALICE_ID.clone();
         let asset_definition_id = AssetDefinitionId::from_str("rose#wonderland")?;
         let asset_id = AssetId::new(asset_definition_id, account_id.clone());
-        SetKeyValue::asset(
-            asset_id.clone(),
-            Name::from_str("Bytes")?,
-            vec![1_u32, 2_u32, 3_u32],
-        )
-        .execute(&account_id, &mut state_transaction)?;
+        let key = Name::from_str("Bytes")?;
```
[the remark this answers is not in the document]
Stukalov-A-M: @mversic this is not my code and its related to Name module not to JsonString. It's ok for me to get rid of all `::from_str` in tests, but not in this task. This is out of scope. I'll create a separate issue

primitives/src/json.rs
Stukalov-A-M: > @Stukalov-A-M it looks to me as if `Eq` and `Ord` are not incongruent

Didn't catch the thought. Current implementation of `Ord` is fully consistent and realizes correct comparing for most of potential cases.
Other questions that regard to congruent we agreed to discuss in a separate issue https://github.com/hyperledger/iroha/pull/4732#discussion_r1642358856, isn't it?

mversic: there is no such thing as correct comparing. Whatever you implement in `Ord` is "correct". What you have to make sure though is that when `Ord` returns `Ordering::Equal`, `Eq` must return true. Otherwise you break implementations of `BTreeSet` or `HashSet` etc. Since your `Ord` implementation looks specialized, I assume `Eq` should be specialized as well. You can read about which conditions must hold [here](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)

Stukalov-A-M: > there is no such thing as correct comparing. Whatever you implement in `Ord` is "correct"

How so?
```
let str1 = r#"{ "a" : 1.23}"#;
    let str2 = r#"{ "a" :      1.23}"#;

    println!("{:?}", str1.cmp(&str2));

    let str1 = serde_json::from_str::<Value>(str1).unwrap();
    let str2 = serde_json::from_str::<Value>(str2).unwrap();

    let value_1 = str1.to_string();
    let value_2 = str2.to_string();

    println!("{:?}", value_1.cmp(&value_2));

Greater
Equal

```

Stukalov-A-M: @mversic , the congruity between `Ord`, `PartialOrd` and `PartialEq` is fixed.

Stukalov-A-M:
Created a a task for unit tests creation
https://github.com/hyperledger/iroha/issues/4748

primitives/src/json.rs
```diff
@@ -0,0 +1,220 @@
...
+            }
+            (Value::Array(left), Value::Array(right)) => {
+                let left = left.iter().map(JsonString::from);
+                let right = right.iter().map(JsonString::from);
+                left.cmp(right)
+            }
+            (Value::Object(left), Value::Object(right)) => {
+                let left = left.iter().map(|(k, v)| (k, JsonString::from(v)));
+                let right = right.iter().map(|(k, v)| (k, JsonString::from(v)));
+                left.cmp(right)
```
Erigara, line 118: I think with `serde_json` feature `preserve_order` it might give us weird results.
Like two objects like `{"a": 0, "b": 0}` and `{"b": 0, "a": 0}` won't be equal, but afaik we don't use this feature.

mversic: after discussing with @Erigara we agreed it would be best to just derive both `Ord` and `Eq`. This custom implementation is not necessary

mversic (changes requested):

core/src/smartcontracts/isi/query.rs
```diff
@@ -610,7 +603,7 @@ mod tests {
             let mut metadata = Metadata::new();
             metadata.insert_with_limits(
                 Name::from_str("Bytes")?,
-                MetadataValueBox::Vec(vec![1_u32.into(), 2_u32.into(), 3_u32.into()]),
+                JsonString::new(vec![1_u32, 2_u32, 3_u32]),
```
mversic, line 606:
```suggestion
                vec![1_u32, 2_u32, 3_u32],
```

data_model/src/metadata.rs
```diff
@@ -463,116 +266,15 @@ mod tests {
...
-        Ok(())
-    }
-
     #[test]
     fn insert_exceeds_entry_size() -> Result<(), TestError> {
         let mut metadata = Metadata::new();
         let limits = Limits::new(10, 5);
         assert!(metadata
-            .insert_with_limits(Name::from_str("1")?, "2".to_owned(), limits)
+            .insert_with_limits(Name::from_str("1")?, JsonString::new("2"), limits)
```
mversic, line 274:
```suggestion
            .insert_with_limits(Name::from_str("1")?, 2_u32, limits)
```

data_model/src/metadata.rs
```diff
@@ -463,116 +266,15 @@ mod tests {
...
             .is_ok());
         assert!(metadata
-            .insert_with_limits(Name::from_str("1")?, "23456".to_owned(), limits)
+            .insert_with_limits(Name::from_str("1")?, JsonString::new("23456"), limits)
```
mversic, line 277:
```suggestion
            .insert_with_limits(Name::from_str("1")?, 23_456_u32, limits)
```

client/tests/integration/domain_owner_permissions.rs
```diff
@@ -58,7 +59,7 @@ fn domain_owner_domain_permissions() -> Result<()> {
 
     // check that "alice@wonderland" as owner of domain can edit metadata in her domain
     let key: Name = "key".parse()?;
-    let value: Name = "value".parse()?;
+    let value = JsonString::new("value");
     test_client.submit_blocking(SetKeyValue::domain(kingdom_id.clone(), key.clone(), value))?;
```
mversic, line 63:
```suggestion
    let value = "value";
    test_client.submit_blocking(SetKeyValue::domain(kingdom_id.clone(), key.clone(), value))?;
```

I think we should implement `From<&str> for JsonString` and have it just like this

core/src/smartcontracts/isi/mod.rs
```diff
@@ -378,26 +346,20 @@ mod tests {
         let mut state_transaction = state_block.transaction();
         let domain_id = DomainId::from_str("wonderland")?;
         let account_id = ALICE_ID.clone();
+        let key = Name::from_str("Bytes")?;
         SetKeyValue::domain(
             domain_id.clone(),
-            Name::from_str("Bytes")?,
-            vec![1_u32, 2_u32, 3_u32],
+            key.clone(),
+            JsonString::from(vec![1_u32, 2_u32, 3_u32]),
```
mversic, line 353:
```suggestion
            vec![1_u32, 2_u32, 3_u32],
```

core/src/smartcontracts/isi/mod.rs
```diff
@@ -347,26 +321,20 @@ mod tests {
         let mut state_transaction = state_block.transaction();
         let definition_id = AssetDefinitionId::from_str("rose#wonderland")?;
         let account_id = ALICE_ID.clone();
+        let key = Name::from_str("Bytes")?;
         SetKeyValue::asset_definition(
             definition_id.clone(),
-            Name::from_str("Bytes")?,
-            vec![1_u32, 2_u32, 3_u32],
+            key.clone(),
+            JsonString::from(vec![1_u32, 2_u32, 3_u32]),
```
mversic, line 328:
```suggestion
            vec![1_u32, 2_u32, 3_u32],
```

primitives/src/json.rs
Stukalov-A-M: In this case this test will fail `correct_pagination_assets_after_creating_new_one()`

mversic: ouch, but I don't think this is how it should be resolved

client/tests/integration/domain_owner_permissions.rs
```diff
@@ -233,7 +234,7 @@ fn domain_owner_asset_permissions() -> Result<()> {
 
     // check that "alice@wonderland" as owner of domain can edit metadata of store asset in her domain
     let key: Name = "key".parse()?;
-    let value: Name = "value".parse()?;
+    let value = JsonString::new("value");
```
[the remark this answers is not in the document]
Stukalov-A-M: Will be done here: https://github.com/hyperledger/iroha/issues/4749

data_model/src/metadata.rs
Stukalov-A-M: `"2"` is not a `2_u32`

data_model/src/metadata.rs
Stukalov-A-M: `"23456"` is not a `23_456_u32`

primitives/src/json.rs
Stukalov-A-M: I think to test the solution here but before keep as is:
https://github.com/hyperledger/iroha/issues/4748

Erigara: i think we can adjust the test

client/tests/integration/domain_owner_permissions.rs
Stukalov-A-M: Integration tests will be refactored here

Will be done here: https://github.com/hyperledger/iroha/issues/4749

primitives/src/json.rs
mversic: > I think to test the solution here but before keep it as is: #4748

you should derive `Eq` and `Ord`. you can ignore the test and fix in another ticket

mversic (approved):

mversic:
@Stukalov-A-M IMO you can add `#[ignore` to the failing test and fix it in a separate PR

Erigara (approved):

nxsaken:
@Stukalov-A-M the integration test `correct_pagination_assets_after_creating_new_one ` has been failing since merging this. Can you fix or ignore the test as suggested above?

Stukalov-A-M:
> @Stukalov-A-M the integration test `correct_pagination_assets_after_creating_new_one ` has been failing since merging this. Can you fix or ignore the test as suggested above?

I don't think it's possible in the closed PR. You may add `ignore` if u have an opened PR.

nxsaken:
@Stukalov-A-M I meant in a separate PR

Issue #4672, opened 2024-05-30
cascade or protect on delete entities?


s8sato:
Currently, domains, accounts, and assets are cascaded on unregister due to nested structure, but for shallow objects #3921 we might choose one of the following unregister policies:

- autoremove orphans (ON DELETE CASCADE)
  - entity should have a set of references (IDs) of other entities
- do not let any entity be orphaned  (ON DELETE PROTECT)
  - entity should have a reference counter that indicates how many other entities are referencing it

s8sato:
- domains, accounts, and assets would keep cascading if that is implied in CanUnregisterDomain
- should an account be dominant for the triggers it has registered?
  - related to #4670

s8sato:
To consider entity relationships, I drew what if the storages under the world were tables of relational database:

```mermaid
erDiagram
    %% parent tables
    DOMAINS {
        _ id PK
        _ logo
        _ metadata
        _ admin FK
    }
    ACCOUNTS {
        _ id PK
        _ metadata
    }
    ASSETS {
        _ id PK
        _ total_quantity
        _ is_mintable
        _ logo
        _ metadata
        _ admin FK
    }
    %% child tables
    DOMAIN_ACCOUNTS {
        _ domain_id PK, FK
        _ account_id PK, FK
    }
    DOMAIN_ASSETS {
        _ domain_id PK, FK
        _ asset_id PK, FK
    }
    ASSET_ACCOUNTS_NUMERIC {
        _ asset_id PK, FK
        _ account_id PK, FK
        Numeric balance
    }
    ASSET_ACCOUNTS_STORE {
        _ asset_id PK, FK
        _ account_id PK, FK
        Metadata datastore
    }
    %% relationships
    DOMAINS ||--o{ DOMAIN_ACCOUNTS : has
    DOMAINS ||--o{ DOMAIN_ASSETS : has
    ACCOUNTS ||--o{ DOMAINS : administrates
    ACCOUNTS ||--o{ ASSETS : administrates
    ACCOUNTS ||--o{ DOMAIN_ACCOUNTS : has
    ACCOUNTS ||--o{ ASSET_ACCOUNTS_NUMERIC : has
    ACCOUNTS ||--o{ ASSET_ACCOUNTS_STORE : has
    ASSETS ||--o{ DOMAIN_ASSETS : serves
    ASSETS ||--o{ ASSET_ACCOUNTS_NUMERIC : serves
[...]

PR #4734, opened 2024-06-15
[refactor] 3921: Make domain structure shallow

SamHSmith:
### Linked issue

Closes #{issue_number} 

### Checklist


core/src/smartcontracts/isi/account.rs
```diff
@@ -578,7 +576,13 @@ pub mod query {
...
             iroha_logger::trace!(%id);
             Ok(Box::new(
-                state_ro.world().domain(id)?.accounts.values().cloned(),
+                state_ro
+                    .world()
+                    .domain(id)?
+                    .accounts
+                    .iter()
+                    .filter_map(|a_id| state_ro.world().account(a_id).ok())
+                    .cloned(),
```
Erigara, line 585: I think instead of storing account ids inside domain we can take advantage of the fact that accounts are stored in order so that all accounts belong to the same domain are adjacent. 

This would be similar to how it's done in `account_roles_iter`.

Erigara (commented):
Looks good in general, minor suggestions

core/src/smartcontracts/isi/domain.rs
```diff
@@ -105,7 +109,12 @@ pub mod isi {
                 .world
                 .domain_mut(&account_id.domain)?
                 .remove_account(&account_id)
-                .is_none()
+                == false
+                || state_transaction
+                    .world
+                    .accounts
+                    .remove(account_id.clone())
+                    .is_none()
```
DCNick3, line 117: This shouldn't fail if the check to the left of `||` has succeeded, otherwise iroha's storage is in some inconsistent state. Should this be an assert?

DCNick3:
AFAIU, unregistering domain will leave iroha in a somewhat inconsistent state, where it has orphaned accounts in its world structure. Related: #4672

data_model/src/domain.rs
```diff
@@ -74,7 +80,7 @@ mod model {
         /// Identification of this [`Domain`].
         pub id: DomainId,
         /// [`Account`]s of the domain.
-        pub accounts: AccountsMap,
+        pub accounts: btree_set::BTreeSet<AccountId>,
```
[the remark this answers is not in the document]
SamHSmith: I think we do for speed reason. Otherwise searching the global btree will be very expensive I think.

core/src/smartcontracts/isi/account.rs
SamHSmith: Oh interesting, I had not thought of that.

data_model/src/domain.rs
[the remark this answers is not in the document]
SamHSmith: Well I take this back given what you said about them being in order.

core/src/smartcontracts/isi/domain.rs
SamHSmith: When writing an application I would assert. But here I think we want to do our best without crashing. Perhaps I can minimize the potential for this inconsistent state.

data_model/src/domain.rs
```diff
@@ -73,8 +72,6 @@ mod model {
     pub struct Domain {
         /// Identification of this [`Domain`].
         pub id: DomainId,
-        /// [`Account`]s of the domain.
-        pub accounts: AccountsMap,
         /// [`Asset`](AssetDefinition)s defined of the `Domain`.
         pub asset_definitions: AssetDefinitionsMap,
```
Erigara, line 76: We might consider excluding `asset_definitions` as well (probably in separate PR).

Erigara (approved):

data_model/src/domain.rs
mversic: separate issue for sure. This was POC

mversic (approved):
I thought we said this change will not affect public API, but it does

core/src/smartcontracts/isi/account.rs
```diff
@@ -609,11 +607,8 @@ pub mod query {
             Ok(Box::new(
                 state_ro
                     .world()
-                    .map_domain(&asset_definition_id.domain.clone(), move |domain| {
-                        domain.accounts.values().filter(move |account| {
-                            account.assets.contains_key(&asset_definition_id)
-                        })
-                    })?
+                    .accounts_iter()
```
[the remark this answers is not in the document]
Erigara: It would clash with `WorldReadOnly` field getter i think

Issue #4696, opened 2024-06-05
Allow multiple peers submit genesis


dima74:
As described in https://github.com/hyperledger/iroha/issues/4225#issuecomment-2087976958

dima74:
Currently we have following behaviour:
* GenesisTransaction is provided to Leader peer
* Leader peer signs it and creates SignedBlock
* Leader peer sends SignedBlock to all peers
* Finally all peers have same SignedBlock (signed only by Leader peer)

Initially I planned to implement this task as follows:
* GenesisTransaction is provided to all peers
* Each peer extract executor and parameters from GenesisTransaction (they should be first and second instructions)
* Leader peer signs GenesisTransaction and creates SignedBlock
* Leader peer sends SignedBlock to all peers
* Peers receive SignedBlock and check that block payload equals to GenesisTransaction they have
* Finally all peers have same SignedBlock (signed only by Leader peer)

We discussed with @Erigara and found problem with this approach: all peers have all necessary data (GenesisTransaction), but still have to wait SignedBlock from the Leader. So we agreed on the following approach:
* `kagami genesis sign` will produce SignedBlock instead of GenesisTransaction. To do so `topology` field will be added to `genesis.json`
* SignedBlock is provided to all peers. All peers will use that SignedBlock directly as a first block without any communication. Note that first block will not be signed by any peer, that is `SignedBlock::signatures` will be empty.

@mversic @0x009922 what do you think?

mversic:
>  `kagami genesis sign` will produce SignedBlock instead of GenesisTransaction. To do so `topology` field will be added to `genesis.json`

yes, this is also better because it hardcodes the initial topology

> SignedBlock is provided to all peers. All peers will use that SignedBlock directly as a first block without any communication. Note that first block will not be signed by any peer, that is SignedBlock::signatures will be empty.

why does it have to be provided to all initial peers and can't be also circulated via block sync? what happens when a new peer is added to the network?

dima74:
> why does it have to be provided to all peers and can't be also circulated via block sync? what happens when a new peer is added to the network, how does it get genesis?

This is according to https://github.com/hyperledger/iroha/issues/4225#issuecomment-2087976958. @0x009922 did you propose to provide genesis to all peers including peers that are not in initial topology (added to the network later), or did I misunderstand your comment?

mversic:
> > why does it have to be provided to all peers and can't be also circulated via block sync? what happens when a new peer is added to the network, how does it get genesis?
[...]

Issue #4555, opened 2024-05-06
Don't sign genesis blocks


mversic:
signing of genesis isn't required. Just produce a `HashOf<SignedTransaction>` and distribute it to all users to verify the block. This means to replace `GENESIS_PUBLIC_KEY` with `GENESIS_HASH` in the config file

dima74:
https://github.com/hyperledger/iroha/pull/4793#issuecomment-2208774337

PR #4739, opened 2024-06-18
refactor!: Supply `SignedBlock` instead of `SignedTransaction` to peer

dima74:
## Description

This is first part of #4696:
* Added `topology` field to `genesis.json`
* Changed `kagami genesis sign` to generate `SignedBlock`
* Adapted `iroha_swarm` and `test_env.py` to those changes

Planned changes for next PRs:
* Provide hash of genesis block to all peers (#4555)
* Allow multiple peers to submit genesis

### Checklist


tools/swarm/src/compose.rs
```diff
@@ -757,7 +784,12 @@ mod tests {
                 init: true
                 command: |-
                   /bin/sh -c "
-                  kagami genesis sign /config/genesis.json --public-key $$GENESIS_PUBLIC_KEY --private-key 8026405A6D5F06A90D29AD906E2F6EA8B41B4EF187849D0D397081A4A15FFCBE71E7C73420F48A9EEB12513B8EB7DAF71979CE80A1013F5F341C10DCDA4F6AA19F97A9 --out-file $$GENESIS_SIGNED_FILE &&
+                  kagami genesis sign /config/genesis.json \\
+                      --public-key $$GENESIS_PUBLIC_KEY \\
```
mversic, line 788: seems superfluous logically

genesis/src/lib.rs
```diff
@@ -23,9 +26,15 @@ pub static GENESIS_DOMAIN_ID: Lazy<DomainId> = Lazy::new(|| "genesis".parse().un
 #[repr(transparent)]
 pub struct GenesisTransaction(pub SignedTransaction);
 
+/// Genesis block.
+/// First instruction should be [`Upgrade`].
```
mversic, line 30: Is this enforced? Also I believe genesis can only have 1 transaction?

genesis/Cargo.toml
```diff
@@ -12,7 +12,7 @@ workspace = true
 
 [dependencies]
 iroha_crypto = { workspace = true }
-iroha_data_model = { workspace = true, features = ["http"] }
+iroha_data_model = { workspace = true, features = ["http", "transparent_api"] }
```
mversic, line 15:
```suggestion
iroha_data_model = { workspace = true, features = ["http"] }
```

If clients are expected to use this library, this is not ok

mversic (changes requested):

tools/swarm/src/compose.rs
mversic: I'm saying this only as a reminder

dima74: It is about newlines, right? They are not required, but when arguments are on separate lines, it is much easier to read or check generated docker compose files

mversic: No, I was talking about the public key. Nvm

genesis/Cargo.toml
dima74: I can move this function to `iroha_data_model`, do you think it is ok?
```rust
fn create_genesis_block(
    genesis_transaction: SignedTransaction,
    topology: Vec<PeerId>,
) -> SignedBlock {
```

genesis/src/lib.rs
dima74: It is checked in `ValidBlock::validate` that genesis block has single transaction. Added check that first instruction is `Upgrade`

genesis/Cargo.toml
mversic: Maybe BlockBuilder can have build_genesis?

dima74: It will be cyclic dependency if we use `BlockBuilder` in `iroha_genesis`. Added method `SignedBlock::new_genesis_block`

mversic: not the best but fine

mversic (dismissed):

data_model/src/block.rs
```diff
@@ -278,6 +278,42 @@ impl SignedBlock {
             SignatureOf::new(private_key, &block.payload),
         ));
     }
+
+    /// Creates genesis block without signatures (not signed by any peer)
+    pub fn new_genesis_block(
```
mversic, line 283:
```suggestion
    pub fn genesis(
```

mversic:
A thing that I have noticed. We don't validate genesis instructions at all anymore because there is no initial executor. This is a bug, only the 1st instruction (which is upgrade) should not be validated, all others should

mversic (approved):

configs/swarm/docker-compose.local.yml
```diff
@@ -22,7 +22,12 @@ services:
     init: true
     command: |-
       /bin/sh -c "
-      kagami genesis sign /config/genesis.json --public-key $$GENESIS_PUBLIC_KEY --private-key 80264082B3BDE54AEBECA4146257DA0DE8D59D8E46D5FE34887DCD8072866792FCB3AD4164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4 --out-file $$GENESIS_SIGNED_FILE &&
+      kagami genesis sign /config/genesis.json \\
+          --public-key $$GENESIS_PUBLIC_KEY \\
+          --private-key 80264082B3BDE54AEBECA4146257DA0DE8D59D8E46D5FE34887DCD8072866792FCB3AD4164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4 \\
```
Erigara, line 27: Why do we get public key from environment, but private key is always the same?

configs/swarm/docker-compose.local.yml
```diff
@@ -22,7 +22,12 @@ services:
...
+          --topology '[{\"address\":\"irohad0:1337\",\"public_key\":\"ed012082528CCC8727333530C8F6F19F70C23882DEB1BF2BA3BE4A6654C7E8A91A7731\"},{\"address\":\"irohad1:1338\",\"public_key\":\"ed012083C85E315776FD2DDC187ECB23E608F800B313A1D614B108078EC048D5013D2D\"},{\"address\":\"irohad2:1339\",\"public_key\":\"ed0120A37B7B758C952FE9429E9E35D1D71E2D8BB9364EDD077B5027ABAAC798D3230E\"},{\"address\":\"irohad3:1340\",\"public_key\":\"ed0120B23E14F659B91736AAB980B6ADDCE4B1DB8A138AB0267E049C082A744471714E\"}]' \\
```
Erigara, line 28: Does `--topology` is used to override what is already defined in `genesis.json`?

Erigara: I'm a bit concerned that `kagami genesis sign` is not single purpose tool.
Maybe we can use smt like `jq` to replace `"topology"` in separate step?

```
jq '.topology=[...]' /config/genesis.json > $$GENESIS_SIGNED_FILE
kagami genesis sign $$GENESIS_SIGNED_FILE ... --out-file $$GENESIS_SIGNED_FILE
```

core/src/block.rs
```diff
@@ -655,6 +678,34 @@ mod valid {
...
     }
 
+    fn check_genesis_block(
+        block: &SignedBlock,
+        genesis_account: &AccountId,
+    ) -> Result<(), InvalidGenesisError> {
+        let transactions = block.payload().transactions.as_slice();
+        let [transaction] = transactions else {
+            return Err(InvalidGenesisError::NotSingleTransaction);
+        };
```
Erigara, line 688: Cool way to check!

data_model/src/block.rs
```diff
@@ -278,6 +278,40 @@ impl SignedBlock {
...
+            error: None,
+        };
+        let payload = BlockPayload {
+            header,
+            commit_topology: topology,
+            transactions: vec![transaction],
+            event_recommendations: vec![],
+        };
+        SignedBlockV1 {
+            signatures: vec![],
```
Erigara, line 310: Shouldn't we include genesis signature here?
So that other peers can verify that genesis block was indeed signed by genesis. 
I think this is not checked right now only on transaction level, so sender can easily replace other fields.

tools/kagami/src/genesis/generate.rs
```diff
@@ -85,9 +85,29 @@ impl<T: Write> RunArgs<T> for Args {
     }
 }
 
+// This should be consistent with `scripts/test_env.py`
```
Erigara, line 88: Can't we modify topology inside `scripts/test_env.py`? 
Because it allows running peer with any number of peers (not only 4).

configs/swarm/docker-compose.local.yml
mversic: good point. This should be defined in `genesis.json`

dima74: > Maybe we can use smt like `jq` to replace `"topology"` in separate step?

Good idea, but this requires to add `jq` to docker image, is it ok?
Alternatively I can add command like `kagami genesis replace-topology`

Erigara: Imo it’s okay, we might ask our DevOps

configs/swarm/docker-compose.local.yml
dima74: Originally it was made this way because genesis private key is used only to sign genesis, and genesis public key is used also by `irohad`. But there are plans (https://github.com/hyperledger/iroha/pull/4719#discussion_r1638122530) to move genesis private key to env, so I moved genesis private key to env in this PR

configs/swarm/docker-compose.local.yml
dima74: Used jq to update topology and executor path

data_model/src/block.rs
dima74: Changed as you suggested. By the way there are plans to use genesis hash instead of signing (#4555)

mversic (approved):

Erigara (approved):

PR #4740, opened 2024-06-18
fix: fix unit tests

Erigara:
## Description

Fix failing unit tests.

Checked tests locally to verify if errors are resolved.

```
cargo test --all-features --no-fail-fast --workspace --exclude iroha_executor
```


core/src/smartcontracts/isi/mod.rs
```diff
@@ -485,23 +482,4 @@ mod tests {
         ));
         Ok(())
     }
-
-    #[test]
-    async fn transaction_signed_by_genesis_account_should_be_rejected() -> Result<()> {
```
Erigara, line 490: This check was removed from `accept` function.

Erigara: @mversic any opinion on this one?

mversic: I can't recall why I removed it

mversic: I think it should be returned

Erigara: returned

0x009922 (approved):

mversic (approved):

PR #4744, opened 2024-06-19
fix(sumeragi): early exit from handle_block_sync

Erigara:
## Description

Exit early from `handle_block_sync` if received block is the same or has smaller or equal view change index.

## Benefits

Improve performance of block sync by avoid doing work where not necessary.


mversic (approved):
should it be ported to `main` branch also?

Mingela (approved):

PR #4757, opened 2024-06-21
fix!: Upgrade executor in separate transaction in genesis block

### LINKED ISSUE #4743: Validate genesis instructions
> A thing that I have noticed. We don't validate genesis instructions at all anymore because there is no initial executor. This is a bug, only the 1st instruction (which is upgrade) should not be validated, all others should

_Originally posted by @mversic in https://github.com/hyperledger/iroha/issues/4739#issuecomment-2178400860_

dima74:
## Description

`GenesisBlock` now have two transactions, first with single `Upgrade` instruction to set executor, and second with all other instructions. If there are no other instructions, second transaction will be omitted.

### Linked issue

Closes #4743

### Checklist


core/src/block.rs
```diff
@@ -695,23 +694,34 @@ mod valid {
...
+        for transaction in transactions {
+            if transaction.value.authority() != genesis_account {
+                return Err(InvalidGenesisError::UnexpectedAuthority);
+            }
+            if transaction.error.is_some() {
+                return Err(InvalidGenesisError::ContainsErrors);
+            }
+            let Executable::Instructions(_) = transaction.value.instructions() else {
+                return Err(InvalidGenesisError::InvalidInstructions);
+            };
```
mversic, line 706: I think you could have put many of these checks into `SignedBlock::deserialize` and reject eagerly

mversic (approved):

data_model/src/block.rs
```diff
@@ -280,16 +280,18 @@ impl SignedBlock {
 
     /// Creates genesis block signed with genesis private key (and not signed by any peer)
     pub fn genesis(
-        genesis_transaction: SignedTransaction,
+        genesis_transactions: Vec<SignedTransaction>,
```
0x009922, line 283: [Parse, don’t validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

Maybe it would be good to define a stricter type for `GenesisTransactions`, which will verify on construction/deserialization that the first TX has only one `Upgrade` ISI?

```rs
// (De-)serializes as `Vec<SignedTransaction>` with validation
struct GenesisTransactions {
  upgrade: SignedTransaction,
  rest: Vec<SignedTransaction>
}

impl GenesisTransactions {
  fn new(upgrade: SignedTransaction, rest: Vec<SignedTransaction>) -> Result<Self, SomeError> {
    // validate that `upgrade` is appropriate
  }

  fn get_upgrade(&self) -> Upgrade {
    // can `unwrap` `self.upgrade` safely thanks to validation on creation
  }
}
```

0x009922 (approved):
Anyway, clean work.

data_model/src/block.rs
```diff
@@ -339,13 +345,50 @@ mod candidate {
...
         }
 
+        fn validate_genesis(&self) -> Result<(), &'static str> {
+            let transactions = self.payload.transactions.as_slice();
+            for transaction in transactions {
+                if transaction.error.is_some() {
+                    return Err("Genesis transaction must not contain errors");
+                }
+                let Executable::Instructions(_) = transaction.value.instructions() else {
+                    return Err("Genesis transaction must contain instructions");
```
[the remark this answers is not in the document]
mversic: It's fine atm. We don't even have a way to represent it in json

data_model/src/block.rs
mversic: Let's make it a separate issue and merge as is

Issue #4729, opened 2024-06-13
Replace use of `Duration` for timestamps with an actual timestamp, e.g. `chrono::DateTime`


0x009922:
In some places in `iroha_data_model`, structures use `Duration` for data which doesn't mean a _span of time_, but actually means a timestamp. Examples:

- `start`: https://github.com/hyperledger/iroha/blob/4362c9427fa8b4090676ac46e8171ef5f1995ff6/data_model/src/events/time.rs#L100-L105
- `since`: https://github.com/hyperledger/iroha/blob/4362c9427fa8b4090676ac46e8171ef5f1995ff6/data_model/src/events/time.rs#L126-L131

These fields are timestamps, not time spans.

I propose to use a structure that is designed specifically for timestamps, e.g. [`chrono::DateTime`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html).

hollermay:
Can I work on this issue?
I can help around this

nxsaken:
Hi @hollermay, feel free to open a PR.

mversic:
`start` should become `start_ms` and `since` should become `since_ms`. However, getters should return something like `chrono::DateTime`

@hollermay should I assign you?

hollermay:
> `start` should become `start_ms` and `since` should become `since_ms`. However, getters should return something like `chrono::DateTime`
> 
> @hollermay should I assign you?

Sure I can give a shot

mversic:
also do the same for `SignedTransactionV1::creation_time_ms` and `SignedBlockV1::timestamp_ms`

hollermay:
@mversic please check the pull request

dima74:
`u64` is used for timestamps after #4841:
https://github.com/hyperledger/iroha/blob/a27bc2eaa7e3cfab6c35cae3b2bcace63a3fee2b/data_model/src/events/time.rs#L100-L105
@mversic should this issue be closed?

s8sato:
https://github.com/hyperledger-iroha/iroha/pull/5015#issuecomment-3050490455

PR #4762, opened 2024-06-22
fix: Replace use of Duration for timestamps with an actual timestamp, e.g. chrono::DateTime

hollermay:
## Description

This PR resolves issue number #4729 

### Before resolving:
Screenshot_23-6-2024_24127_github com
Screenshot_23-6-2024_2427_github com

### After resolving the issue

image
image
image

Also, apart from this I made the required changes that involved changing of start to start_ms and since to since_ms.


data_model/src/events/time.rs
```diff
@@ -99,7 +99,7 @@ mod model {
     )]
     pub struct Schedule {
         /// The first execution time
-        pub start: Duration,
+        pub start_ms: Duration,
```
mversic, line 102:
```suggestion
        pub start_ms: u64,
```

data_model/src/events/time.rs
```diff
@@ -125,7 +125,7 @@ mod model {
     #[ffi_type]
     pub struct TimeInterval {
         /// The start of a time interval
-        pub since: Duration,
+        pub since_ms: Duration,
```
mversic, line 128:
```suggestion
        pub since_ms: u64,
```

mversic (changes requested):

PR #4773, opened 2024-06-26
fix: fix apply/validate blocks benchmarks

Erigara:
## Description

Fix number of issues with block benchmarks:
- correct block signing
- generating accounts correctly
- correctly get accounts by domain id
- remove account permissions on account removal
- update storage crate to fix issues with range iterator and removal of accounts 

## Benchmark results

After making account shallow we have the following result:

```plain
apply_blocks/apply_blocks
    time:   [18.491 ms 19.988 ms 21.691 ms]
    change: [+7.2201% +19.955% +34.299%] (p = 0.01 < 0.10)
    Performance has regressed.

validate_blocks/validate_blocks
    time:   [11.090 s 11.425 s 11.751 s]
    change: [-32.585% -29.986% -27.411%] (p = 0.00 < 0.10)
    Performance has improved.
```


core/src/smartcontracts/isi/domain.rs
```diff
@@ -102,6 +102,11 @@ pub mod isi {
                         .expect("should succeed")
                 });
 
+            state_transaction
+                .world
+                .account_permissions
+                .remove(account_id.clone());
```
s8sato, line 108: I think this is a good catch, and why not the same for `world.account_roles` ?

core/benches/blocks/validate_blocks_oneshot.rs
```diff
@@ -16,7 +18,12 @@ fn main() {
         .expect("Failed building the Runtime");
     {
         let _guard = rt.enter();
-        iroha_logger::test_logger();
+        let config = ConfigReader::new()
+            .with_env(std_env)
+            .read_and_complete()
+            .expect("Failed to load config");
+        let config = InitConfig::new(config, true);
+        let _ = iroha_logger::init_global(config).expect("Failed to initialize logger");
```
mversic, line 26: is this really how we construct config?

mversic: this is not representative of how clients will do it?

mversic (approved):

core/benches/blocks/validate_blocks_oneshot.rs
Erigara: You can check how config is constructed in `read_config_and_genesis`.

Here i just wanted to use some default values and override them with env if necessary (not to spam log messages).

core/src/state.rs
```diff
@@ -2015,15 +2004,41 @@ mod tests {
...
+
+        let domain_id = "kingdom".parse().unwrap();
+        let view = map.view();
+        let range = view
+            .range(AccountByDomainBounds::new(&domain_id))
+            .collect::<Vec<_>>();
+        assert_eq!(range.len(), 0);
+        for (account, ()) in range {
+            assert_eq!(&account.domain, &domain_id);
+        }
```
s8sato, line 2042: I think this iterates over an empty vec and asserts nothing

core/benches/blocks/apply_blocks.rs
```diff
@@ -20,17 +23,22 @@ impl StateApplyBlocks {
     /// - Failed to generate [`KeyPair`]
     /// - Failed to create instructions for block
     pub fn setup(rt: &tokio::runtime::Handle) -> Self {
-        let domains = 100;
-        let accounts_per_domain = 1000;
-        let assets_per_domain = 1000;
+        let domains = 10;
+        let accounts_per_domain = 100;
+        let assets_per_domain = 100;
```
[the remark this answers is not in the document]
Erigara: it just took too much time...

core/src/state.rs
Erigara: oh, right, it's leftover from my testing

core/src/smartcontracts/isi/domain.rs
Erigara: good point

core/src/state.rs
Erigara: fixed it to just count + added extra check

s8sato (approved):

mversic (approved):

PR #4775, opened 2024-06-26
feat: Allow multiple peers submit genesis block

### LINKED ISSUE #4696: Allow multiple peers submit genesis

dima74:
## Description

Continuation of #4739:
* Topology stored in genesis block will be used as initial topology, as a result multiple peers can submit genesis
* Removed `--submit-genesis` flag
* Changed `scripts/test_env.py`, it will submit genesis for all peers

### Linked issue

Closes #4696

### Checklist

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -410,6 +410,8 @@ impl Sumeragi {
             // Consider our peer has genesis,
             // and some other peer has genesis and broadcast it to our peer,
             // then we can ignore such genesis block because we already has genesis.
+            // Note: `ValidBlock::validate` also checks it,
+            // but we don't want warning to be printed since this is correct behaviour.
             return None;
         }
 
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -318,6 +318,8 @@ impl Sumeragi {
             "Genesis contains invalid transactions"
         );
 
+        self.topology = Topology::new(genesis.as_ref().commit_topology().cloned());
+
         let msg = BlockCreated::from(&genesis);
         let genesis = genesis
             .commit(&self.topology)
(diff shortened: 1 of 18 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -404,6 +406,13 @@ impl Sumeragi {
     ) -> Option<VotingBlock<'state>> {
         let mut state_block = state.block();
 
+        if state_block.height() == 1 && block.header().height == 1 {
```
mversic, line 409: I believe this is already checked in `ValidBlock::validate`

scripts/test_env.py
```diff
@@ -138,7 +129,8 @@ def __init__(self, args: argparse.Namespace, nth: int):
             #     "tokio_console_addr": f"{self.host_ip}:{self.tokio_console_port}",
             # }
         }
-        if nth == 0:
+        self.submit_genesis = nth == 0 or args.all_peers_submit_genesis
+        if self.submit_genesis:
```
mversic, line 133: IMO you could have just have it submitted by all peers and avoid having a flag

mversic (changes requested):
I think you should remove the `--submit-genesis` flag? A peer should check if the genesis file has been set and based on that submit a genesis

core/src/sumeragi/main_loop.rs
dima74: Yes but `ValidBlock::validate` will print warning about it, so I added this check to avoid the warning

mversic: ok, but you can inline this check into `ValidBlock::validate` and skip writing warning in there then

mversic (approved):

core/src/sumeragi/main_loop.rs
mversic: it's fine

nxsaken (approved):

dima74:
> * Removed `--submit-genesis` flag

@BAStos525 this is the only change needed from the deployment point of view

PR #4778, opened 2024-06-26
refactor: improve custom instruction usage

mversic:
### Linked issue

Closes #{issue_number} 

### Checklist


data_model/src/query/mod.rs
```diff
@@ -109,10 +109,13 @@ macro_rules! queries {
 pub type QueryId = String;
 
 /// Trait for typesafe query output
-pub trait Query: Into<QueryBox> + seal::Sealed {
+pub trait Query: Into<QueryBox> {
```
DCNick3, line 112: Can you please the query changes for now? We don't have custom queries yet (and I don't think they'll land before #4569), so this landing this change will just lead to additional conflicts for me (🙁), while not having any benefit (yet)

DCNick3 (commented):
As far as I can see, this PR:
- adds a custom parameter and an empty file for custom permissions to the `executor_custom_data_model` test smartcontract
- removes `Fail` instruction; replaces its uses in tests with unregistering of a non-existent domain
- renames `Custom` instruction to `CustomInstruction`
- in addition to `Instruction` trait, adds `BuiltInInstruction`, allowing `Instruction` to be not sealed
- in addition to `Query` trait, adds a sealed `BuiltInQuery` trait, by analogy with the `Instruction` trait. 
- makes it so that in smart contracts, only `BuiltInInstruction` and `BuiltInQuery` can be executed.

My guess that the main point of it is the last change: to not allow custom instructions to be executed from wasm. While I see how it's beneficial to prevent this for executor (it's kind of recursive otherwise), I am not sure that regular smart-contracts & triggers should be prohibited from doing so...

client/tests/integration/smartcontracts/executor_custom_data_model/src/simple_isi.rs
```diff
@@ -23,7 +23,9 @@ pub struct MintAssetForAllAccounts {
     pub quantity: Numeric,
 }
 
-impl From<CustomInstructionBox> for Custom {
+impl Instruction for CustomInstructionBox {}
```
dima74, line 26: Please do same changes for `complex_isi.rs` (add `impl Instruction` and remove `fn into_instruction()`)

client/tests/integration/smartcontracts/executor_custom_data_model/src/parameters.rs
```diff
@@ -0,0 +1,23 @@
+//! Module with custom parameters
```
dima74, line 1: Looks like `parameters.rs` and `permissions.rs` are not used anywhere, at least there are no mod declarations for them in `lib.rs`

core/src/queue.rs
```diff
@@ -669,9 +667,7 @@ pub mod tests {
         let mut queue = Queue::test(config_factory(), &time_source);
         let (event_sender, mut event_receiver) = tokio::sync::broadcast::channel(1);
         queue.events_sender = event_sender;
-        let instructions = [Fail {
-            message: "expired".to_owned(),
-        }];
+        let instructions = [Register::domain(Domain::new("domain".parse().unwrap()))];
```
dima74, line 670:
```suggestion
        let instructions = [Unregister::domain(Domain::new("dummy".parse().unwrap()))];
```

data_model/src/query/mod.rs
mversic: sure

client/tests/integration/smartcontracts/executor_custom_data_model/src/simple_isi.rs
mversic: weird, I thought I did

client/tests/integration/smartcontracts/executor_custom_data_model/src/parameters.rs
mversic: my bad, committed by accident

mversic:
> adds a custom parameter and an empty file for custom permissions to the executor_custom_data_model test smartcontract

that was an accidental leftover

> removes Fail instruction; replaces its uses in tests with unregistering of a non-existent domain

this can now be expressed with a custom ISI

> in addition to Query trait, adds a sealed BuiltInQuery trait, by analogy with the Instruction trait.

reverted, since as you point out it has no benefit

> My guess that the main point of it is the last change: to not allow custom instructions to be executed from wasm. While I see how it's beneficial to prevent this for executor (it's kind of recursive otherwise), I am not sure that regular smart-contracts & triggers should be prohibited from doing so...

I tried initially but there were practical problems. Like I had to derive `Decode/Encode`, implement `encode_as_instruction_box` which depended on `EnumRef`. It was becoming a rabbit hole

mversic:
@DCNick3 I agree it's quite unfortunate that this doesn't work from within smart contracts, but it's still an improvement. Like was previously possible, from within smart contracts user would have to convert instruction into `CustomInstruction` and then execute. I will open a ticket to explore this

DCNick3 (approved):

dima74 (approved):

PR #4788, opened 2024-06-28
feat!: reimplement multisignature transactions

### LINKED ISSUE #4373: Reimplement multi-signature

Erigara:
## Description

This PR provide an example how multisignature transactions could be implemented using iroha triggers.
In this PR i've tried to make as small invasive changes into the core as possible.

How this work?

There is 2 triggers (thanks @s8sato  for inspiration in account activation feature): 
1. `multisig_register` - responsible for registering new multisig user and trigger `multisig`
2. `multisig` - responsible for executing instructions on behalf of multisig user when enough votes are collected  

Assumptions: 
- user have to come up with they own public key for multisig account:
    - it might be ok to have primary private key in case of emergency or smt
    - it could be generated at random or obtained through key aggregation
    - it's currently not possible to generate public key inside wasm smart-contracts
- `multisig_register` should be registered either as part of genesis or by domain owner to have enough permissions for it's actions
    - currently in default executor there is no way to allow certain trigger to be executable by anyone (workaround would be to have trigger which would grant this permission on account registration)

### Linked issue

Closes #4373 

### Benefits

Get back multisig functionality without too much changes on iroha side.

### Downsides

Executing multisig instructions through tirgger probably slower that backed in solution.


client/tests/integration/multisig.rs
```diff
@@ -0,0 +1,170 @@
...
+            ExecuteTriggerEventFilter::new().for_trigger(multisig_register_trigger_id.clone()),
+        ),
+    );
+
+    // Register trigger which would allow multisig account creation in wonderland domain
+    // Access to call this trigger shouldn't be restricted
+    test_client.submit_blocking(Register::trigger(trigger))?;
+
+    // Create multisig account id and destroy it's private key
+    let multisig_account_id = gen_account_in("wonderland").0;
```
s8sato, line 80: - this should be a variant of `AccountId` (essentially a variant of `PublicKey`) to reject transactions that claim to have this multisig authority, which could be left to another issue
- since multisig account ID is not associated with its signatories, they would need some way to know that the multisig account is theirs

s8sato: - the former holds if multisig account is created on user side as this
  - if created on system side, such a case could be statistically ignored that user restores the private key and correctly signs
- the latter holds if created on system side with some tweak with account registration

client/tests/integration/multisig.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/// Mirror trigger args for `multisig`
+#[derive(Serialize, Deserialize)]
+enum MultisigArgs {
+    /// Accept instruction and vote
+    Instruction(InstructionBox),
+    /// Accept vote for certain instruction
+    Vote(HashOf<InstructionBox>),
+}
```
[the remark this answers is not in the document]
Erigara: > and why not an executable but an instruction? 

Afaik currently we currently don't have ability to execute `Executable` from the `wasm` only one instruction at a time.

> Do we limit to one instruction at a time? Or does this just intend simplicity as a prototype?

Yes and yes, i want to make implementation simple. 
In case we want more sophisticated schemes we can accept `Vec<InstructionBox>`. 

This way user would be able to have smt like `[register by call trigger, execute trigger]` which is kinda workaround to execute arbitrary executable.

client/tests/integration/smartcontracts/multisig/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+        ),
+        _ => dbg_panic("only work as by call trigger"),
+    };
+
+    let isi_hash = match &args {
+        Args::Instruction(isi) => HashOf::new(isi),
+        Args::Vote(isi_hash) => *isi_hash,
+    };
+    let votes_metadata_key: Name = format!("{isi_hash}/votes").parse().unwrap();
+    let isi_metadata_key: Name = format!("{isi_hash}/isi").parse().unwrap();
```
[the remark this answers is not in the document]
Erigara: That's right, as for now it's possible to discard instruction by perfroming 2 rounds of voting to remove associated isi (if we would extend voting from `isi` to `Vec<isi>` this would be single round of voting).

However we might consider might sophisticated schemes like you suggested as well.

Extend trigger arguments to accept negative votes or have associated trigger which would track metadata of multisig trigger and discard isi after some amount of blocks.

Do you think it's functionality we should implement in this PR?

client/tests/integration/smartcontracts/multisig/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    let signatories: BTreeSet<AccountId> =
+        FindTriggerKeyValueByIdAndKey::new(id.clone(), "signatories".parse().unwrap())
+            .execute()
+            .dbg_unwrap()
+            .into_inner()
+            .try_into_any()
+            .dbg_unwrap();
+
+    // Require N of N signatures
+    if votes.is_superset(&signatories) {
```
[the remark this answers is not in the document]
Erigara: It's not hard to modify this code to work for `m` of `n` case. 
It would be some trigger metadata which would contain voting mode like `All`, `Majority`, `SuperMajority`, ...

Another way i was think about is to use this multisig trigger as some kind of gateway. 
So that signatories of multisig account can register another trigger to perform voting (in a different way) and then unregistered this trigger if no longer required.

client/tests/integration/multisig.rs
Erigara: > since multisig account ID is not associated with its signatories, they would need some way to know that the multisig account is theirs

My thinking was that creation of multisig account is always initialized as user action (submitting transaction) so interested party already know that they are signatories of a given account.

Erigara: > this should be a variant of AccountId (essentially a variant of PublicKey) to reject transactions that claim to have this multisig authority, which could be left to another issue

There is might be valid use cases to have multisig for regular workflow and private key as backup solution i.e. regular workers have to gather votes and boss can use private key in case of any problem. 

Alternatively signatories might derive public key as sum of theirs public keys (or generated by them for this particular multisig account) so unless every signatory would have to share there private keys with each other otherwise it should be infeasible to find aggregated private key.

client/tests/integration/smartcontracts/multisig/src/lib.rs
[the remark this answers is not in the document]
s8sato: Could you explain how cancellations are currently supposed to be done?

I meant it would be nice if there was an equivalent to transaction ttl, but I don't think this PR should necessarily address this

client/tests/integration/smartcontracts/multisig/src/lib.rs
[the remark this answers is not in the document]
s8sato: If you are referring to a dedicated ephemeral trigger corresponding to each multisig transaction, then yes, that would be an alternative. But so far the current design seems to be achieving an equivalent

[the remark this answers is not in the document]
s8sato: As far as the m of n feature is concerned,
one solution would be to expand "authentication state trie" during multisig proposals so that each node is `(quorum, is_pass)` and update the trie for each vote. This would also support recursive multisig:

```mermaid
flowchart
  classDef fail fill:#8c4351,fill-opacity:0.5
  classDef pass fill:#166775,fill-opacity:0.5

  012345678(1/2):::pass -->|012345| 012345(all):::fail
  012345678 -->|678| 678(2/3):::pass
  012345 -->|012| 012(2/3):::fail
  012345 -->|345| 345(2/3):::pass
  012 -->|0| 0( ):::pass
  012 -->|1| 1( ):::fail
  012 -->|2| 2( ):::fail
  345 -->|3| 3( ):::pass
  345 -->|4| 4( ):::pass
  345 -->|5| 5( ):::fail
  678 -->|6| 6( ):::pass
  678 -->|7| 7( ):::fail
  678 -->|8| 8( ):::pass

```

client/tests/integration/multisig.rs
s8sato: I see, it makes sense to me to leave it up to user to decide whether or not to have the corresponding private key

client/tests/integration/smartcontracts/multisig/src/lib.rs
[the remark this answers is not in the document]
Erigara: > Could you explain how cancellations are currently supposed to be done?

It certainly a hack, but users can submit 2 ISIs with `[remove hash/vote metadata, remove hash/votes]` for any isi which would result in effectively removing this transaction.

s8sato (approved):

client/tests/integration/smartcontracts/multisig/src/lib.rs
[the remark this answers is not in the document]
s8sato: Oh, they vote for `RemoveKeyValue<Trigger>` instruction. That would work

client/tests/integration/smartcontracts/multisig/src/lib.rs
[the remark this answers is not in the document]
Erigara: > If you are referring to a dedicated ephemeral trigger corresponding to each multisig transaction, then yes, that would be an alternative.

I probably explained my point not precisely.

If multisig account signatories want to change voting scheme they can do the following:
1) multisig account `A` has multisig manager trigger `T1` which requires `N of N` votes
2) signatories of `A` make proposal and vote to register trigger `T2` which is also could be used to manage multisig account but requires `M of N` votes (or other voting scheme)
3) in the same proposal they also submit isi to unregister `T1` (and some other technical stuff like giving permission for signatories to execute `T2`)
 
 So now multisig account `A` has replaced voting scheme.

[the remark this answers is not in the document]
Erigara: > authentication state trie

Can you please explain a bit more about this trie? 

Is this would be a trie of proposal hashes?
So when new proposal is submitted we add a new leaf is added to the trie?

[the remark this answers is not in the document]
s8sato: Let me explain:

- the entire trie represents the readiness of one multisig transaction
- accounts are keys to nodes
  - `0` .. `8` are personal and the rest are multisig accounts
- node indicates whether the multisig transaction has been approved up to the corresponding account

client/tests/integration/smartcontracts/multisig_register/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+            "{}_{}_signatories",
+            account_id.signatory(),
+            account_id.domain()
+        )
+    );
+
+    let can_execute_multisig_trigger = Permission::new(
+        "CanExecuteUserTrigger".parse().unwrap(),
+        serde_json::json!({ "trigger": trigger_id }),
+    );
```
mversic, line 88: you can now statically type this

data_model/src/isi.rs
```diff
@@ -924,13 +924,30 @@ mod transparent {
...
-        #[derive(Constructor, Display)]
+        #[derive(Display)]
         #[display(fmt = "EXECUTE `{trigger}`")]
-        #[serde(transparent)]
-        #[repr(transparent)]
         pub struct ExecuteTrigger {
             /// Id of a trigger to execute
             pub trigger: TriggerId,
+            /// Arguments to trigger execution
+            pub args: Option<JsonString>,
```
mversic, line 933:
```suggestion
            pub args: Vec<JsonString>,
```

why isn't it a list of args?

client/tests/integration/smartcontracts/multisig_register/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+panic-halt.workspace = true
+lol_alloc.workspace = true
+getrandom.workspace = true
+
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true, default-features = false }
+
+[build-dependencies]
+iroha_wasm_builder = { version = "=2.0.0-pre-rc.21", path = "../../../../../wasm_builder" }
```
mversic, line 25:
```suggestion
iroha_wasm_builder = { workspace = true }
```

should we bring it into workspace?

mversic (changes requested):

client/tests/integration/multisig.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// Mirror trigger args for `multisig`
+#[derive(Serialize, Deserialize)]
+enum MultisigArgs {
+    /// Accept instruction proposal and initialize votes with the proposer's one
+    Instruction(Vec<InstructionBox>),
+    /// Accept vote for certain instruction
+    Vote(HashOf<Vec<InstructionBox>>),
+}
```
mversic, line 38: put it in `executor_custom_data_model/trigger.rs` use as dependency

data_model/src/isi.rs
Erigara: Because it easily could be encoded as json array if required or as object to have named fields.

client/tests/integration/smartcontracts/multisig_register/Cargo.toml
Erigara: not sure, it's used only here now

client/tests/integration/smartcontracts/executor_custom_data_model/src/multisig.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub signatories: BTreeSet<AccountId>,
+}
+
+/// Arguments to multisig account manager trigger
+#[derive(Serialize, Deserialize)]
+pub enum MultisigArgs {
+    /// Accept instruction proposal and initialize votes with the proposer's one
+    Instruction(Vec<InstructionBox>),
+    /// Accept vote for certain instruction
+    Vote(HashOf<Vec<InstructionBox>>),
```
s8sato, line 23:
```suggestion
    /// Accept instructions proposal and initialize votes with the proposer's one
    Instructions(Vec<InstructionBox>),
    /// Accept vote for certain instructions
    Vote(HashOf<Vec<InstructionBox>>),
```

s8sato (dismissed):

client/tests/integration/smartcontracts/multisig/src/lib.rs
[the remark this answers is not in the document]
s8sato: If we take this approach, we need a way to know its members from the multisig account in order to expand the trie

configs/swarm/executor.wasm
mversic, line 1: why do we have a 50% size increase?

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -648,3 +649,51 @@ fn build_register_trigger_isi(
...
+    assert_eq!(new_value, prev_value.checked_add(numeric!(42)).unwrap());
+
+    Ok(())
+}
+
+/// Mirror trigger args for `mint_rose_trigger_args`
+#[derive(Serialize, Deserialize)]
+struct Args {
+    val: u32,
+}
```
mversic, line 699: should be part of the `executor_custom_data_model`

client/tests/integration/triggers/by_call_trigger.rs
```diff
@@ -648,3 +649,51 @@ fn build_register_trigger_isi(
...
+        iroha_wasm_builder::Builder::new("tests/integration/smartcontracts/mint_rose_trigger_args")
+            .show_output()
+            .build()?
+            .optimize()?
+            .into_bytes()?;
+    let wasm = WasmSmartContract::from_compiled(wasm);
+    let trigger = Trigger::new(
+        trigger_id.clone(),
+        Action::new(
+            wasm.clone(),
```
mversic, line 674:
```suggestion
            wasm,
```

why clone it?

client/tests/integration/smartcontracts/mint_rose_trigger_args/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+getrandom::register_custom_getrandom!(iroha_trigger::stub_getrandom);
+
+#[derive(Serialize, Deserialize)]
+struct Args {
+    val: u32,
+}
+
+/// Mint 1 rose for owner
+#[iroha_trigger::main]
+fn main(_id: TriggerId, owner: AccountId, event: EventBox) {
```
mversic, line 26:
```suggestion
fn main(_id: TriggerId, owner: AccountId, event: ExecuteTriggerEvent) {
```

can't we take the correct event in the function definition?

client/tests/integration/smartcontracts/multisig/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+use iroha_trigger::{debug::dbg_panic, prelude::*};
+use lol_alloc::{FreeListAllocator, LockedAllocator};
+
+#[global_allocator]
+static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
+
+getrandom::register_custom_getrandom!(iroha_trigger::stub_getrandom);
+
+#[iroha_trigger::main]
+fn main(id: TriggerId, _owner: AccountId, event: EventBox) {
```
mversic, line 21:
```suggestion
fn main(id: TriggerId, _owner: AccountId, event: ExecuteTriggerEvent) {
```

the same. I think you can do this and conversion will be implicit

client/tests/integration/smartcontracts/multisig_register/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+#[global_allocator]
+static ALLOC: LockedAllocator<FreeListAllocator> = LockedAllocator::new(FreeListAllocator::new());
+
+getrandom::register_custom_getrandom!(iroha_trigger::stub_getrandom);
+
+// Trigger wasm code for handling multisig logic
+const WASM: &[u8] = core::include_bytes!(concat!(core::env!("OUT_DIR"), "/multisig.wasm"));
+
+#[iroha_trigger::main]
+fn main(_id: TriggerId, _owner: AccountId, event: EventBox) {
```
mversic, line 25:
```suggestion
fn main(_id: TriggerId, _owner: AccountId, event: ExecuteTriggerEvent) {
```

client/tests/integration/smartcontracts/mint_rose_trigger_args/src/lib.rs
Erigara: this is not supported:

```plain
error[E0308]: mismatched types
  --> mint_rose_trigger_args/src/lib.rs:25:1
   |
25 | #[iroha_trigger::main]
   | ^^^^^^^^^^^^^^^^^^^^^^ expected `ExecuteTriggerEvent`, found `EventBox`
26 | fn main(_id: TriggerId, owner: AccountId, event: ExecuteTriggerEvent) {
   |    ---- arguments to this function are incorrect
   |
note: function defined here
  --> mint_rose_trigger_args/src/lib.rs:26:4
   |
26 | fn main(_id: TriggerId, owner: AccountId, event: ExecuteTriggerEvent) {
   |    ^^^^                                   --------------------------
   = note: this error originates in the attribute macro `iroha_trigger::main` (in Nightly builds, run with -Z macro-backtrace for more info)
```

mversic (changes requested):

client/tests/integration/smartcontracts/multisig_register/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    .dbg_expect("failed to parse role");
+
+    let can_execute_multisig_trigger = CanExecuteUserTrigger {
+        trigger: trigger_id.clone(),
+    };
+    let role = Role::new(role_id.clone()).add_permission(can_execute_multisig_trigger);
+
+    Register::role(role)
+        .execute()
+        .dbg_expect("failed to register multisig role");
```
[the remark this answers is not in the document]
Erigara: after what? we need this role for the whole duration of multisig trigger existence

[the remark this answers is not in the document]
mversic: after the instruction has received enough votes and was executed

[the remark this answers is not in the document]
mversic: nvm, I got it

client/tests/integration/smartcontracts/executor_custom_data_model/src/multisig.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub signatories: BTreeSet<AccountId>,
+}
+
+/// Arguments to multisig account manager trigger
+#[derive(Serialize, Deserialize)]
+pub enum MultisigArgs {
+    /// Accept instructions proposal and initialize votes with the proposer's one
+    Instructions(Vec<InstructionBox>),
+    /// Accept vote for certain instructions
+    Vote(HashOf<Vec<InstructionBox>>),
```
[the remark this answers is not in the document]
Erigara: > You could have sent the same instruction repeatedly.

Yes, but than you have to store all this repetition on-chain which is not great.

We lenient on `creation_time` because we were merging transactions in the queue. 
Right not we don't have such problem, we might still want to de-duplicate payloads (i.e. multiple the same mints) in this case we might want to introduce nonce.
I would prefer to deal with it in separate PR if necessary.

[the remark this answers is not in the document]
mversic: is there a reason to collect instructions and not the whole transaction?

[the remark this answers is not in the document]
Erigara: Currently you call trigger using ExecuteTrigger isi, you can't really pass transaction inside itself. 

Additionally i don't see a reason to pass transaction, since only part relevant to execution is payload.

configs/swarm/executor.wasm
Erigara: i forgot `--optimize` flag

Erigara:
> great piece of work. It would be interesting to explore whether we can make a custom instruction for MultiSigRegister, might be even simpler to use

i think it should be possible, in this case we would move logic of multisig register trigger into the executor which would be triggered on custom instruction

mversic (dismissed):

mversic (approved):

s8sato (approved):

PR #4793, opened 2024-07-01
refactor!: Use hash to validate genesis block

### LINKED ISSUE #4555: Don't sign genesis blocks

dima74:
## Description

* Added `genesis.hash` config parameter and removed `genesis.public_key`
* Added `genesis` field (account id) to `genesis.json`. It will be authority of genesis transactions
* Genesis transactions will not be signed (actually signed by some random key pair and peers will ignore signatory for genesis transactions and instead check genesis block hash match)
* Replaced `kagami genesis sign` with `kagami genesis prepare` which converts `genesis.json` to `genesis.scale` (`SignedBlock`) and outputs genesis hash to stdout
* Genesis domain and account is now added to the `World` after receiving genesis block (since we now don't know genesis account id at startup but need to use authority of genesis block)

Todo:
* [ ] Fix docker-compose. Currently each peer generates genesis block independently, so its time and hash will be different

### Deployment changes
1. Add `genesis` field to `genesis.json` (genesis account id)
2. Change `kagami genesis sign` invocation

Before:
```sh
kagami genesis sign genesis.json --public-key ... --private-key ... --out-file ...
```

After:
```sh
kagami genesis prepare ... --out-file ...
```
This will also output genesis hash

3. Provide genesis hash using `GENESIS_HASH` environment variable to `irohad`

### Linked issue

Closes #4555

### Checklist


configs/swarm/docker-compose.local.yml
```diff
@@ -64,7 +65,10 @@ services:
       P2P_ADDRESS: 0.0.0.0:1338
       API_ADDRESS: 0.0.0.0:8081
       GENESIS_PUBLIC_KEY: ed01204164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4
+      GENESIS_PRIVATE_KEY: 80264082B3BDE54AEBECA4146257DA0DE8D59D8E46D5FE34887DCD8072866792FCB3AD4164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4
```
mversic, line 68: why is this required?

both `GENESIS_PUBLIC_KEY` and `GENESIS_PRIVATE_KEY` should be removed from docker-compose files. Instead, if this is required, nodes should pick it from the authority of genesis transactions

mversic (changes requested):

configs/swarm/docker-compose.local.yml
mversic: but preferably, `kagami` can put a random public key as genesis account id. Then peers just take that from genesis transaction authority

tools/kagami/src/genesis.rs
```diff
@@ -5,19 +5,22 @@ use clap::Subcommand;
 use crate::{Outcome, RunArgs};
 
 mod generate;
+mod hash;
 mod sign;
 
 #[derive(Debug, Clone, Subcommand)]
 pub enum Args {
     Sign(sign::Args),
```
mversic, line 13:
```suggestion
```
we want to remove signing of genesis

tools/kagami/src/genesis.rs
```diff
@@ -5,19 +5,22 @@ use clap::Subcommand;
...
     Generate(generate::Args),
+    Hash(hash::Args),
```
mversic, line 15:
```suggestion
    Generate(generate::Args),
```

could we generate both genesis block and it's hash with a single command? hash could be printed to `stdout`

client/examples/register_1000_triggers.rs
```diff
@@ -75,15 +75,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let chain_id = get_chain_id();
     let genesis_key_pair = get_key_pair(test_network::Signatory::Genesis);
     let topology = vec![peer.id.clone()];
+    let genesis = generate_genesis(1_000_u32, chain_id.clone(), &genesis_key_pair, topology)?;
     let configuration = get_config(
         unique_vec![peer.id.clone()],
         chain_id.clone(),
         get_key_pair(test_network::Signatory::Peer),
-        genesis_key_pair.public_key(),
+        genesis.0.hash(),
```
mversic, line 83:
```suggestion
        genesis.hash(),
```

maybe make it private

config/iroha_test_config.toml
```diff
@@ -6,7 +6,7 @@ private_key = "802640282ED9F3CF92811C3818DBC4AE594ED59DC1A2F78E4241E31924E101D6B
 address = "127.0.0.1:1337"
 
 [genesis]
-public_key = "ed01204164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4"
+hash = "b66bcd327cd7f29fe3b6e677de734326a7bf89e3d2775c927c8a60496ff000a9"
 file = "./genesis.signed.scale"
```
mversic, line 10:
```suggestion
file = "./genesis.scale"
```

configs/swarm/docker-compose.local.yml
mversic: > Instead, if this is required, nodes should pick it from the authority of genesis transactions

but preferably peers should not care about which account is the authority of genesis transactions

mversic (changes requested):

configs/swarm/docker-compose.local.yml
dima74: In `genesis.json` there can be genesis account id, in e.g. `Transfer` instructions. So do I understand correctly that you propose that `kagami` will generate new genesis account id and replace existing genesis account id in `genesis.json` with generated one?

mversic: > In genesis.json there can be genesis account id, in e.g. Transfer instructions. So do I understand correctly that you propose that kagami will generate new genesis account id and replace existing genesis account id in genesis.json with generated one?

this is what I was thinking since I see genesis account as some dummy value that is more like a placeholder for `Transaction::authority`. But since genesis can create entities, those entities need to have some owner as you point out. How about we define a field `genesis` in `genesis.json` that holds `AccountId` of genesis account?

mversic: it's not important that transactions are signed by this genesis account. Genesis transactions don't even have to be signed

dima74: > it's not important that transactions are signed by this genesis account. Genesis transactions don't even have to be signed

Yes but in current data model `SignedTransaction` must have some `signature`, if it is not genesis account signature, then how peers will know what is genesis account id?

I think users may use `genesis@genesis` as alias for genesis account id in `genesis.json`, and `kagami` will generate genesis key pair, use it as authority for genesis transaction, so in this case users will not need to know anything about genesis key pair. But `genesis@genesis` can't be used as account id, so another option I think of is that users use some genesis account id in `genesis.json`, and then `kagami` generates new genesis account id and replace with it in `genesis.json`

Erigara:
So if we still have to use genesis account afaik what benefits do we gain by now also checking hash of genesis block?

configs/swarm/docker-compose.local.yml
mversic: > Yes but in current data model SignedTransaction must have some signature, if it is not genesis account signature, then how peers will know what is genesis account id?

from `Transaction::authority`

mversic: we could also consider introducing a separate type `GenesisBlock` in `data_model` if we can't make it as yet another `SignedBlock`

dima74:
> So if we still have to use genesis account afaik what benefits do we gain by now also checking hash of genesis block?

As I see `kagami genesis sign` might be simplified a bit (no genesis key pair needed) if we [figure out](https://github.com/hyperledger/iroha/pull/4793#discussion_r1660814153) how to do it. But I don't know how genesis account id can be removed altogether since it is used in `genesis.json` instructions like `Transfer`

configs/swarm/docker-compose.local.yml
Erigara: introduction of `GenesisBlock` as separate type would make block sync more complex because now there is 2 types of block we want to sync.

core/test_network/src/lib.rs
```diff
@@ -141,13 +138,13 @@ impl TestGenesis for GenesisBlock {
 pub struct NetworkBuilder {
     n_peers: u32,
     port: Option<u16>,
-    config: Option<Config>,
```
Erigara, line 144: Is this change related to rebasing?

configs/swarm/docker-compose.local.yml
dima74: > > Yes but in current data model SignedTransaction must have some signature, if it is not genesis account signature, then how peers will know what is genesis account id?
> 
> from `Transaction::authority`, and `Transaction::authority` should be taken from `genesis` field in `genesis.json`. Genesis private key will never be needed by any of the peers

So should it work like this:
* Add `genesis` field to `genesis.json` with genesis `AccountId`
* `kagami` conversion from `genesis.json` to `SignedBlock`:
  * Remove genesis key pair arguments
  * Genesis transactions authority will be genesis account specified in `genesis.json`
  * Genesis transactions will be signed by some random key pair
  * Match of transaction `authority` and `signatory` will not be checked for genesis transactions

core/test_network/src/lib.rs
dima74: No it is because genesis block hash was added to the `Config`. Only one test set custom config so I added specific method for this test instead (`with_debug_force_soft_fork`)

configs/swarm/docker-compose.local.yml
mversic: > introduction of `GenesisBlock` as separate type would make block sync more complex because now there is 2 types of block we want to sync

yes, but different block types will never interchange. So you can think of block sync operating in 2 phases which could be modeled as a state machine. First there is genesis phase and only after it do all other blocks come.

dima74: Changed implementation according to [this comment](https://github.com/hyperledger/iroha/pull/4793#discussion_r1661199227) (PR description also updated). Please re-review

tools/kagami/src/genesis.rs
dima74: Introduced `kagami genesis prepare` command which converts `genesis.json` to `genesis.scale` and prints genesis hash to stdout

tools/kagami/src/genesis.rs
```diff
@@ -5,18 +5,18 @@ use clap::Subcommand;
...
 
 mod generate;
-mod sign;
+mod prepare;
 
 #[derive(Debug, Clone, Subcommand)]
 pub enum Args {
-    Sign(sign::Args),
+    Prepare(prepare::Args),
     Generate(generate::Args),
```
mversic, line 13:
```suggestion
    Prepare(prepare::Args),
    Generate(generate::Args),
```

do we need both?

data_model/src/transaction.rs
```diff
@@ -329,7 +339,6 @@ mod candidate {
     impl SignedTransactionCandidate {
         fn validate(self) -> Result<SignedTransactionV1, &'static str> {
             self.validate_instructions()?;
-            self.validate_signature()?;
```
mversic, line 332: I don't like this. Transactions with incorrect signatures might creep into Iroha. I'd rather that you make another `SignedTransactionCandidate` in `data_model::block` which will have different implementation, i.e. it will not verify genesis signatures

data_model/src/block.rs
```diff
@@ -367,6 +370,9 @@ mod candidate {
                 let Executable::Instructions(_) = transaction.value.instructions() else {
                     return Err("Genesis transaction must contain instructions");
                 };
+                if transaction.value.authority().domain.name.as_ref() != "genesis" {
+                    return Err("Genesis transaction authority must be from `genesis` domain");
+                }
```
mversic, line 375: I don't think we should care about this?

core/src/tx.rs
mversic, line 1: I think we can remove `GENESIS_DOMAIN_ID` check from this module. Also I think that we can this constant alltogether and with it the notion of "genesis" domain

configs/swarm/docker-compose.local.yml
```diff
@@ -57,8 +51,9 @@ services:
       PRIVATE_KEY: 8026403ECA64ADC23DC106C9D703233375EA6AC345AD7299FF3AD45F355DE6CD1B55109897952D14BDFAEA780087C38FF3EB800CB20B882748FC95A575ADB9CD2CB21D
       P2P_ADDRESS: 0.0.0.0:1338
       API_ADDRESS: 0.0.0.0:8081
-      GENESIS_PUBLIC_KEY: ed01204164BF554923ECE1FD412D241036D863A6AE430476C898248B8237D77534CFC4
+      GENESIS: /tmp/genesis.scale
       TRUSTED_PEERS: '[{"address":"irohad2:1339","public_key":"ed01204EE2FCD53E1730AF142D1E23951198678295047F9314B4006B0CB61850B1DB10"},{"address":"irohad0:1337","public_key":"ed0120A98BAFB0663CE08D75EBD506FEC38A84E576A7C9B0897693ED4B04FD9EF2D18D"},{"address":"irohad3:1340","public_key":"ed0120CACF3A84B8DC8710CE9D6B968EE95EC7EE4C93C85858F026F3B4417F569592CE"}]'
+      TOPOLOGY: '[{"address":"irohad2:1339","public_key":"ed01204EE2FCD53E1730AF142D1E23951198678295047F9314B4006B0CB61850B1DB10"},{"address":"irohad1:1338","public_key":"ed01209897952D14BDFAEA780087C38FF3EB800CB20B882748FC95A575ADB9CD2CB21D"},{"address":"irohad0:1337","public_key":"ed0120A98BAFB0663CE08D75EBD506FEC38A84E576A7C9B0897693ED4B04FD9EF2D18D"},{"address":"irohad3:1340","public_key":"ed0120CACF3A84B8DC8710CE9D6B968EE95EC7EE4C93C85858F026F3B4417F569592CE"}]'
```
mversic, line 56: what was the reason you introduced genesis to all peers? is it because of the integration tests?
On the other hand, I would prefer keeping it as simple as possible because of external users

core/src/block.rs
```diff
@@ -417,7 +412,7 @@ mod valid {
             block: SignedBlock,
             topology: &Topology,
             expected_chain_id: &ChainId,
-            genesis_account: &AccountId,
+            genesis_hash: &HashOf<SignedBlock>,
```
mversic, line 415:
```suggestion
            genesis_hash: HashOf<SignedBlock>,
```

it's a copy type

core/src/block.rs
```diff
@@ -684,27 +675,32 @@ mod valid {
...
+            .insert(genesis_account.id.clone(), genesis_account);
+    }
+
+    fn genesis_account_and_domain(public_key: PublicKey) -> (Account, Domain) {
+        let genesis_account_id =
+            AccountId::new(iroha_genesis::GENESIS_DOMAIN_ID.clone(), public_key);
+        let genesis_account = Account::new(genesis_account_id).into_account();
+        let genesis_domain =
+            Domain::new(iroha_genesis::GENESIS_DOMAIN_ID.clone()).build(&genesis_account.id);
+        (genesis_account, genesis_domain)
```
mversic, line 703:
```suggestion
    fn genesis_account_and_domain(public_key: PublicKey) -> (Account, Domain) {
        let genesis_account_id =
            AccountId::new(iroha_genesis::GENESIS_DOMAIN_ID.clone(), public_key);
        let genesis_account = Account::new(genesis_account_id).into_account();
        let genesis_domain =
            Domain::new(iroha_genesis::GENESIS_DOMAIN_ID.clone()).build(&genesis_account.id);
        (genesis_account, genesis_domain)
```

IMO we should take both from the genesis authority. And remove the `GENESIS_DOMAIN_ID`

genesis/src/lib.rs
```diff
@@ -23,14 +23,31 @@ pub static GENESIS_DOMAIN_ID: Lazy<DomainId> = Lazy::new(|| "genesis".parse().un
...
+pub struct GenesisBlock(SignedBlock);
+
+impl GenesisBlock {
+    /// Constructor
+    pub fn new(block: SignedBlock) -> GenesisBlock {
+        Self(block)
+    }
+
+    /// Extracts inner `SignedBlock`
+    pub fn into_inner(self) -> SignedBlock {
```
mversic, line 37: why not just implement `From<GenesisBlock> for SignedBlock`

genesis/src/lib.rs
```diff
@@ -23,14 +23,31 @@ pub static GENESIS_DOMAIN_ID: Lazy<DomainId> = Lazy::new(|| "genesis".parse().un
 /// First transaction should contain single [`Upgrade`] instruction to set executor.
 /// Second transaction should contain all other instructions.
 /// If there are no other instructions, second transaction will be omitted.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Encode)]
```
mversic, line 26: why do we need `Encode` now?

genesis/src/lib.rs
```diff
@@ -93,64 +112,65 @@ impl RawGenesisTransaction {
...
-    );
-    TransactionBuilder::new(chain_id, genesis_account_id)
+    let genesis_account = AccountId::new(GENESIS_DOMAIN_ID.clone(), genesis_public_key.clone());
+    TransactionBuilder::new(chain_id, genesis_account)
         .with_instructions(instructions)
-        .sign(genesis_key_pair.private_key())
+        // There is no need to sign genesis transactions since we check it by hash,
+        // but our data model requires SignedTransaction to be signed,
+        // so we sign it with random key pair
+        .sign(KeyPair::random().private_key())
```
mversic, line 173: Just to bring attention for other reviewers. This is a bit contentious, but I think it's a good approach.

genesis/src/lib.rs
```diff
@@ -40,6 +57,8 @@ pub struct RawGenesisTransaction {
     instructions: Vec<InstructionBox>,
     /// Initial topology
     topology: Vec<PeerId>,
+    /// Genesis account
+    genesis: AccountId,
```
mversic, line 61: please put it after `chain`. Generated json file will look nicer

mversic (changes requested):

core/src/tx.rs
dima74: There are checks like "not allowed to register account in genesis domain". Should we keep those checks?

mversic: I don't think we need to keep those checks. Because if we do, we have to track it in runtime which is complicated. Remove checks and open an issue to discuss

data_model/src/transaction.rs
dima74: I agree, but what to do with `validate_transaction` in executor? It will deserialize `SignedTransaction` at WASM side, so it will fail for genesis transaction if we keep validate signature during deserialization

mversic: > One option I can think of is to make `SignedTransaction` validation `#[cfg(not(target_family = "wasm"))]`

I like this suggestion. Can we say that since data model entities are always provided by the host there is no need to validate them on the wasm side? If so, we can can open a ticket to generalize this approach to all entities, although it seems kinda hacky and I'm not sure if this won't bite us in the ass

mversic: Another possibility is to hardcode some public key as genesis and say that every Iroha deployment will have this as genesis account_id. In that case we can remove `account` from `genesis.json`. This is more or less how we had it except that `kagami` wouldn't have to receive private key but it already knows it's value. @Erigara @s8sato ?

mversic: let's discuss tomorrow on standup

Erigara: I was thinking about such opportunity and for example pick a key with all 0s, however as turned out such public key is not secure in `Ed25519`, because it's kinda identity element.

So if we going to pick some hardcoded public key users might have some trust issues if we picked it on purpose to have some kind of backdoor.

Alternative might be to extend `AccountId` to either be public key or arbitrary address which doesn't have corresponding private key and can't be used to sign transactions. It was [suggested](https://github.com/hyperledger/iroha/issues/2085#issuecomment-1980557101) by @s8sato for multisig accounts.

Erigara: We can use the same multihash format  but with [`identity`](https://github.com/multiformats/multicodec/blob/8fc4918c9e7c6cb8f1d0501523d5a0943c46a157/table.csv#L2) code

Erigara: After discussion it was noted that we still want genesis to sign transaction/block in which case my proposal is not relevant.

dima74:
Discussed with @mversic and decided that currently it is not worth to implement #4555. Originally it was expected that it will simplify things, but looks like there is no good and clean implementation, so will keep current approach with genesis public and private key (note that genesis private key is used only in kagami)

data_model/src/transaction.rs
s8sato: The complexity seems to come from trying to strip away the functionality of digital signatures while taking advantage of the transaction processing flow.
There might not have been any problem in the first place: if the privilege comes from block height 0, what is the actual harm even if the genesis account with the correct private key appears after genesis?

core/src/tx.rs
s8sato: I agree, after genesis there doesn't seem to be any particular harm in referring a domain named "genesis"

data_model/src/transaction.rs
mversic: > what is the actual harm even if the genesis account with the correct private key appears after genesis?

I like this point. Especially if someone has misconfigured the blockchain in genesis.json, for example they haven't transfered ownership

s8sato: Yeah, currently custom genesis can grant any privileges to any account. But they agreed anyway. They joined the network while still being able to see that genesis block

PR #4794, opened 2024-07-01
refactor(swarm): build with `irohad0` instead of dummy builder

nxsaken:
## Description

Image for Compose is now built by the first peer service instead of the dummy builder. Solves [the CI failure](https://github.com/hyperledger/iroha/actions/runs/9739223415/job/26874072004) when testing generated Compose configs caused by https://github.com/docker/compose/issues/10596.

### Benefits

No need for a dummy service to build images.

### Checklist

### DIFF
--- a/configs/swarm/docker-compose.local.yml
+++ b/configs/swarm/docker-compose.local.yml
@@ -3,15 +3,9 @@
 # Seed: Iroha
 
 services:
-  builder:
+  irohad0:
     image: hyperledger/iroha:local
     build: ../..
-    pull_policy: never
-    command: echo ok
-  irohad0:
-    depends_on:
-    - builder
-    image: hyperledger/iroha:local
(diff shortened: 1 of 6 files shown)


configs/swarm/docker-compose.local.yml
```diff
@@ -54,7 +48,7 @@ services:
       "
   irohad1:
     depends_on:
-    - builder
+    - irohad0
```
mversic, line 51: why do other services depend on the 1st?

BAStos525 (approved):

configs/swarm/docker-compose.local.yml
nxsaken: If they don't, they will all build the image redundantly, causing the OOM killer kicking in

mversic (approved):

PR #4802, opened 2024-07-04
fix: Smarter transaction expire logic

SamHSmith:
### Linked issue

Closes #{issue_number} 

### Checklist

### DIFF
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -106,12 +106,17 @@ impl Queue {
         }
     }
 
-    fn is_pending(&self, tx: &AcceptedTransaction, state_view: &StateView) -> bool {
-        !self.is_expired(tx) && !tx.is_in_blockchain(state_view)
+    fn is_pending(
+        &self,
+        tx: &AcceptedTransaction,
+        state_view: &StateView,
+        time_padding: Duration,
+    ) -> bool {
+        !self.is_expired(tx, time_padding) && !tx.is_in_blockchain(state_view)
     }
 
     /// Checks if the transaction is waiting longer than its TTL or than the TTL from [`Config`].
-    pub fn is_expired(&self, tx: &AcceptedTransaction) -> bool {
+    pub fn is_expired(&self, tx: &AcceptedTransaction, time_padding: Duration) -> bool {
         let tx_creation_time = tx.as_ref().creation_time();
 
         let time_limit = tx.as_ref().time_to_live().map_or_else(
@@ -120,7 +125,7 @@ impl Queue {
         );
 
         let curr_time = self.time_source.get_unix_time();
-        curr_time.saturating_sub(tx_creation_time) > time_limit
+        curr_time.saturating_sub(tx_creation_time) + time_padding > time_limit
     }
 
     /// If `true`, this transaction is regarded to have been tampered to have a future timestamp.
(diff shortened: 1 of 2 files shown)


core/src/queue.rs
```diff
@@ -120,7 +125,7 @@ impl Queue {
         );
 
         let curr_time = self.time_source.get_unix_time();
-        curr_time.saturating_sub(tx_creation_time) > time_limit
+        curr_time.saturating_sub(tx_creation_time) + time_padding > time_limit
```
s8sato, line 128: This change argues that more transactions should be expired in sumeragi, right?
Where does this issue originate?

SamHSmith: Yes. The original issue is that the leader can create a block with transactions that will expire before the consensus round is over. When iroha is under transaction load and producing many blocks this becomes very common. This change makes sure that block production remains smooth by throwing away transactions that are not guaranteed to live long enough.

s8sato: I thought it was at the time of block creation that determines whether transactions expire or not, and transactions only need to survive until the block creation.

Btw do we have some consensus mechanism on whether or not the leader should include a certain transaction in block, as noted [here](https://github.com/hyperledger/iroha/blob/dc6a0c642a3af2ca89974e0d9d01029ca28cacbf/docs/source/iroha_2_whitepaper.md?plain=1#L185-L189
) and [here](https://github.com/hyperledger/iroha/blob/dc6a0c642a3af2ca89974e0d9d01029ca28cacbf/docs/source/iroha_2_whitepaper.md?plain=1#L247-L252
)?

Erigara: > I thought it was at the time of block creation that determines whether transactions expire or not, and transactions only need to survive until the block creation.

By looking at the code i think Sato is correct, we don't reject transaction from the block based on expired it or not.

Erigara: > Btw do we have some consensus mechanism on whether or not the leader should include a certain transaction in block, as noted [here](https://github.com/hyperledger/iroha/blob/dc6a0c642a3af2ca89974e0d9d01029ca28cacbf/docs/source/iroha_2_whitepaper.md?plain=1#L185-L189) and [here](https://github.com/hyperledger/iroha/blob/dc6a0c642a3af2ca89974e0d9d01029ca28cacbf/docs/source/iroha_2_whitepaper.md?plain=1#L247-L252)?

I far as i remember there were discussion that this mechanism is not really working because faulty peer can ignore sending recites or smt like that.

mversic: > `is_expired` is Queue method and we newer call in on block revalidation.

which seems like a bug because malicious peer can put expired transactions into the new block. This should be checked during consensus by other peeers. But only during consensus and not during block replay

Erigara: this change might cause some problems because time is not synchronized across nodes

s8sato: Regardless of this PR, correct implementation of TTL might not be possible without time consensus:
The leader imposes a time window (with `Queue.tx_time_to_live` and `Queue.future_threshold`) on transactions from the client, but since there is no time consensus, even if the leader drops all transactions with its system time far outside this window, no one can sue the leader with certainty

s8sato: But it may also be true that there is no incentive for leaders to cheat in such a way. For example, they can't just ignore certain clients this way. Ignoring all transactions, they would simply result in being replaced immediately

PR #4810, opened 2024-07-08
refactor!: rename wasm to smart contract

mversic:
## Description

The fact that we work with wasm smart contracts is a sort of implementation detail.
Let's not specify implementation details in the API.

### Linked issue

Closes #{issue_number} 

### Benefits

If we change from wasm to some other VM, we will have a reduced effect on the API.
If we don't it is implied that we use wasm, so no harm done there

### Checklist


client_cli/src/main.rs
```diff
@@ -108,8 +108,8 @@ enum Subcommand {
     /// The subcommand related to event streaming
     #[clap(subcommand)]
     Events(events::Args),
-    /// The subcommand related to Wasm
-    Wasm(wasm::Args),
+    /// The subcommand related to smart contracts
+    SmartContract(smart_contract::Args),
```
nxsaken, line 112: This changes the subcommand, README needs to be updated

nxsaken (approved):

nxsaken (approved):

nxsaken (approved):

nxsaken (approved):

client/src/client.rs
```diff
@@ -460,7 +459,9 @@ impl Client {
 
         let mut tx_builder = match instructions.into() {
             Executable::Instructions(instructions) => tx_builder.with_instructions(instructions),
-            Executable::Wasm(wasm) => tx_builder.with_wasm(wasm),
+            Executable::SmartContract(smart_contract) => {
```
s8sato, line 462: Imo what corresponds to smartcontracts is triggers in Iroha, and the execution part of a trigger (and a transaction) can be Wasm or ISIs. So I don't think "smartcontract" is a good candidate to rename "wasm"

nxsaken: While working on #4850, I noticed that Iroha has both triggers and smart contracts, at least we have two separate macros for defining the entry point. Not sure what the difference is.

nxsaken: Ok, so smart contracts only have an owner, but triggers also specify the trigger id, owner and event. What happens when a trigger wasm is executed as a smart contract, and vice versa?

mversic: > While working on #4850, I noticed that Iroha has both triggers and smart contracts, at least we have two separate macros for defining the entry point. Not sure what the difference is.

how I see it.

1. `Transaction::executable` is either a `Vec<Instruction>` or `SmartContract`
You can execute an executable with `client.submit`. This executes the instructions or the smart contract immediately on submition
2. `Action::executable` is either a `Vec<Instruction>` or `SmartContract`
You have to first register a trigger and then invoke it with `InstructionBox::Execute`. The trigger can consist of  `Vec<InstructinoBox`> or a be a smart contract

mversic: > Ok, so smart contracts only have an owner, but triggers also specify the trigger id, owner and event

no, smart contracts don't have an owner. The transaction they are part of has an owner

nxsaken: I mean a smart contract entrypoint has an owner field, while a trigger entrypoint has that plus additional fields

mversic:
@s8sato has a good point. I'm closing this

PR #4812, opened 2024-07-08
feat!: speicfy on-chain parameters explicitly in genesis

Erigara:
## Description

Currently it's not possible to submit genesis which doesn't pass under default parameter values (i.e. large genesis). 
This PR fixes that allowing to specify parameters inside genesis.json so that they are applied before processing rest of instructions.


genesis/src/lib.rs
```diff
@@ -40,6 +40,9 @@ pub struct RawGenesisTransaction {
     instructions: Vec<InstructionBox>,
     /// Initial topology
     topology: Vec<PeerId>,
+    /// Parameters
+    #[serde(default)]
+    parameters: Vec<Parameter>,
```
mversic, line 45:
```suggestion
    /// Parameters
    #[serde(default)]
    parameters: Vec<Parameter>,
    instructions: Vec<InstructionBox>,
    /// Initial topology
    topology: Vec<PeerId>,
```

configs/swarm/genesis.json
mversic, line 1: I would like for `parameters` to look like this:
```json
  "parameters": [
    {
      "Sumeragi": {
        "BlockTimeMs": 2000
        "CommitTimeMs": 4000
      }
    },
    {
      "Block": {
        "MaxTransactions": 512
      }
    },
    {
      "Transaction": {
        "MaxInstructions": 4096
        "SmartContractSize": 4194304
      }
    },
    {
      "Executor": {
        "Fuel": 55000000
        "Memory": 55000000
      }
    },
    {
      "SmartContract": {
        "Fuel": 55000000
        "Memory": 55000000
      }
    }
```

mversic (changes requested):

configs/swarm/genesis.json
mversic: maybe you can just deserialize to `Parameters`?

genesis/src/lib.rs
mversic: this affects how `genesis.json` looks when generated

configs/swarm/genesis.json
Erigara: I think this is doable, but would require annotating whole parameter stucts and fields with serde(default). 
Additionally it would result in redundant SetParsmeter instructions.

mversic: you can make a ticket to simplify parameters API in `genesis.json`

mversic: leave as is

Erigara: Created #4813

mversic:
the downside of this approach is that `SetParameter` instruction will not get validated with the executor specified in the `genesis.json`

Erigara:
I see, another downside is that custom parameters is not registered yet.
In this case we can left `Upgrade` to be first instruction, then we should make sure that set parameters works fine under new executor default parameters.

mversic:
> I see, another downside is that custom parameters is not registered yet.
> In this case we can left `Upgrade` to be first instruction, then we should make sure that set parameters works fine under new executor default parameters.

We can make Upgrade be the first instruction but not check it for executor limits. Then apply parameters and then all other

Erigara:
> > I see, another downside is that custom parameters is not registered yet.
> > In this case we can left `Upgrade` to be first instruction, then we should make sure that set parameters works fine under new executor default parameters.
> 
> We can make Upgrade be the first instruction but not check it for executor limits. Then apply parameters and then all other

Regestring executor itself is not that heavy. 
So i want expect any problems under default parameters. 
Executor limits is a problem when we validate transaction with a lot of instructions in it.

mversic:
Also it's possible to set parameters on executor migration. But I like the simplicity of providing them through `genesis.json`

configs/swarm/genesis.json
s8sato, line 1: LGTM, I honestly still don't understand what needs to be validated in genesis.
It seems to me that it is the phase of defining the validator arbitrarily and the concept of validate is not yet established there

s8sato (approved):

data_model/src/block.rs
```diff
@@ -383,9 +383,9 @@ mod candidate {
                 );
             };
 
-            if transactions.len() > 2 {
+            if transactions.len() > 3 {
                 return Err(
-                    "Genesis block must have one or two transactions (first with executor upgrade)",
+                    "Genesis block must have 1 to 3 transactions (executor upgrade, parameters, other isi)",
```
mversic, line 388: you need to verify that all ISIs are `SetParameter` in the 2nd transaction

mversic (changes requested):

mversic (approved):

data_model/src/block.rs
Erigara: if there is no parameters in genesis than 2nd transaction would be regular transaction

PR #4828, opened 2024-07-09
fix: Keep voting block if possible

### LINKED ISSUE #4643: Keep voting blocks
> I have this issue in 4 places. It happens either on `BlockSyncUpdate` or `BlockCreated`. At the time of receiving a new block it's possible that the block is invalid and I would like to keep the old block while validating the new one. However, this creates deadlock as I have not released previous writer

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/4518#discussion_r1612166021_

dima74:
## Description

Implements idea from https://github.com/hyperledger/iroha/pull/4518#discussion_r1612185825:

> Maybe we can be smarter about when to reject incoming block? Like checking signatures, height, hashes this checks doesn’t need ability to modify state.

This is relatively small change which I think will keep voting block in most cases (however not in all).

---

Alternative approach might be to change our state block to accumulate changes localy and take write lock just before applying changes. However it will require large refactoring and might affect performance.

### Linked issue

Closes #4643

### Checklist


core/src/block.rs
```diff
@@ -420,7 +425,63 @@ mod valid {
...
+                Self::validate_header(&block, topology, genesis_account, &state.view())
+            {
+                return WithEvents::new(Err((block, error)));
+            }
+
+            // Release block writer before creating new one
+            let _ = voting_block.take();
+            let mut state_block = state.block();
+
+            if let Err(error) = Self::validate_transactions(
```
Erigara, line 466: Cases when voting block going to be discarded falsely:

Case 1:
1. peer already have voting block
2. new block is received through block created message
3. this block has proper height, matching topology, proper leader signature
4. BUT incorrect transactions in the new block

This means that leader send two blocks.

I don't think this is issue because after such event current leader would loose it's place in the validators set. 

And no one other than current leader can perform this operation.

Case 2:
1. peer already have voting block
2. new block received through block sync message
3. this block has proper height (for soft-fork or ordinary condition), view change index larger than current one (!!!), proper leader signature for this larger signature
4. here is two options why this block might be incorrect:
    - incorrect transactions
    - missing proxy tail signature, not enough voting peers signatures

In case when peer conclude that transactions are incorrect but block has all valid signatures indicates that something is certainly wrong with peer itself.

Case when block has missing signatures (and has incorrect transactions) opens possibility for the following attack by any peer in the topology:
1. find view change index (higher than current once)
2. create block 
3. sign it
4. send it through block sync to other peers
5. peers will discard current voting block and this block going to be rejected
6. current view change index will fail 
7. repeat the same process for the new view change index

To mitigate this issue we could check all signatures before trying to execute transactions for the given block (and discarding previous voting block).

I think this can be done in separate ticket or in the current one as you see fit.

core/src/sumeragi/main_loop.rs
```diff
@@ -445,11 +445,10 @@ impl Sumeragi {
                     "Block sync update received"
                 );
 
-                // Release block writer before creating new one
-                // FIX: Restore `voting_block` if `handle_block_sync` returns Err
-                // Currently it's not possible because block writer needs to be released
-                // Look at https://github.com/hyperledger/iroha/issues/4643
-                let _ = voting_block.take();
+                if categorize_block_sync(&block, &state.view()).is_ok() {
```
Erigara, line 448: maybe we can pass result of `categorize_block_sync` into `handle_block_sync` to avoid doing the same check twice?

in the tests you can define functions which would combine this two function

core/src/block.rs
dima74: #4834

core/src/block.rs
```diff
@@ -420,7 +425,63 @@ mod valid {
...
+            }
+
+            WithEvents::new(Ok(ValidBlock(block)))
+        }
+
+        /// Same as `validate` but:
+        /// * Block header will be validated with read-only state
+        /// * If block header is valid, `voting_block` will be released,
+        ///   and transactions will be validated with write state
+        pub fn validate_keep_voting_block<'state>(
```
mversic, line 447: why do we still need `Validate::validate`?

mversic: it seems to me that we could have just one method

dima74: It is used in cases when block is not received from other peers, e.g. in `Sumeragi::init_commit_genesis` for genesis block (provided offline) or `SumeragiHandle::replay_block` (block read from kura)

mversic: even in those cases, wouldn't you be able to just pass `None` for `voting_block`?

mversic (approved):

core/src/block.rs
dima74: > even in those cases, wouldn't you be able to just pass `None` for `voting_block`?

It will be not enough. These functions are also different in `StateBlock` handling. `validate` takes `StateBlock`, but `validate_keep_voting_block` takes `State` and additionally returns `StateBlock` in success case

mversic: it looks to me as if you could send `State` and return `StateBlock` in those cases also?

Erigara (approved):

core/src/block.rs
dima74: Probably yes, but it will complicate usual usages of `validate`, isn't it? Like `validate_keep_voting_block` is specialized function for specific use case thus have a bit more complicated API, and `validate` is general function with usual (easier to understand) API. There are no duplication between these functions (all common parts are extracted to `validate_header` and `validate_transactions`), so I am not sure what could be wrong with having two different functions

mversic: my concern was that someone will use `validate` when they should have used `validate_keep_voting_block`

Issue #4832, opened 2024-07-10
Make methods accepting ids by value generic over its referencedness


DCNick3:
In a lot of places we accept an id (like `AccountId`) by value, requiring spelling out `.clone()` in a lot of places where the id cannot actually be consumed.

A very similar problem is faced when passing strings in rust. `Into<String>` is a generally acceptable way to receive an owned string value, while allowing implicit clones of `&String` and `&str`.

I think we can get a similar benefits by implementing `From<&Id>` for `Id` for all out id types and using `Into<Id>` in function signatures. Then instead of `whiff(floof.clone())` we would be able to write `whiff(&floof)`.

I feel the client-facing APIs (like constructing instructions and queries) could benefit greatly from this: our integration tests have a lot of clutter in form of `.clone()`s. Some parts of `iroha_core` might benefit too.

Issue #4720, opened 2024-06-12
Move query cursor into query params


0x009922:
I don't think it's a right thing to extract cursor from query params, since anyone can fetch cursor.
So it should be signed and passed as body.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/4718#discussion_r1634991206_

DCNick3:
Closed in #4833

Issue #3720, opened 2023-07-18
[refactor] Replace `FindAccountsWithPermissionTokenDefinition` with filtered `FindAllAccounts`


Arjentix:
- Originally suggested by @appetrosyan: https://github.com/hyperledger/iroha/pull/3707#discussion_r1266314964

Should be done after the merge of #3707 as a part of #3671 or in a separate PR.

### Who can help?
@appetrosyan , @DCNick3

mversic:
while we're at it, we should remove `FindAccountsByName` and `FindAssetsByName(FindAssetsByName)`

DCNick3:
I believe this is done as of the merge #4833

PR #4833, opened 2024-07-10
refactor: query filters

### LINKED ISSUE #4569: Restrict types of predicates applied to queries depending on the query output type (type-safe queries)

DCNick3:
# Description

This PR introduces a lot of changes to the query subsystem. I tried to make the changes gradually in smaller PRs, but it ended up being very intertwined =(

## Summary of changes
- Queries are now split into two categories: iterable queries which return sequences of results and singular queries that return a single value. They are now fully disjoint traits stored in different Box enums, as they require very different handling.
- Query filters (aka predicates) are now typed depending on the type of the iterable query they are applied to (#4569). Instead of a single `PredicateBox` there is a predicate type for each queryable type (like `AccountPredicateBox`), with `CompountPredicate<T>` providing the logical operations on top.
- A new rust DSL to construct the predicates is introduced. It is designed to look very close to how one would write an `Iterator::filter` closure (`|obj| obj.field.inner_field.starts_with("hello")`) and is inspired by diesel's DSL.
- HTTP clients and wasm code (smart contracts, triggers and executor) are now using a common query builder struct, generic over a new `QueryExecutor` trait
- The HTTP API is majorly revamped to pass all of the params as scale-encoded values. This simplifies the torii handler, makes it more type-safe (can't add pagination to singular query) and allows sharing more code with wasm
- All of the queries that can be implemented as filtered versions of other queries are now removed. Here's a rough google table I made to keep track of all of them: https://docs.google.com/spreadsheets/d/1Cx9cU4wLk1ZyIa3Igx0NN9azl9dfqcXXMJMLKTMxes4/edit?usp=sharing

### The primary key question

Currently we have a bunch of queries that give you an object using its Primary Key (like `FindAccountById`). They can be implemented by filtering a basis query that returns all objects of such type (like `FindAllAccounts`). However, with naïve implementation (which this PR currently does) this would result in `O(n)` complexity instead of `O(log n)` that we currently have.

There is a way to do these lookups fast: just pattern-match the filter on the shape `|obj| obj.id.eq(expected_id)` and use the primary key index instead of linear scan. We can even use a similar approach to speed up other types of queries, if deemed necessary.

However, merging it as-is would mean that the primary queries would be slow. I can probably add the PK queries back for now, if that's unacceptable.

## Future work

These are improvements I see that can be made to this PR, but I would prefer them to be implemented in separate PRs. I'll convent them to issues when this PR gets closer to getting merged.

Perf:
- the Primary Key optimization (see above, #4895)
- make queries truly lazy (put them into the query store without collecting them all in a vec); Needs investigation whether that'll actually be beneficial and some clever implementation (need to store an iterator with lifetime tied to iroha state snapshot).

Ergonomics:
- remove the `client::account::all()` et al, as this API doesn't provide much value compared to `FindAccounts::new()` and doesn't exist in smart contracts (#4892)
- provide helpers like `client::domain::by_id`, with them still relying on filtered queries
- `impl Into<String>` for ids (#4832)
- introduce a client-like object for smart contracts to make the API more similar to that of the Rust client (#4898)
- expose query types in schema for the ergonomic improvements to translate to SDKs (#4902)
- provide an async client for query execution (#4935)

Misc:
- `Display` representations for queries & filters (right now printing them with `Debug`)
- check that the iterable query continuation is executed by the same authority as the one that started it
- get rid of more singular queries (#4933)

## Linked issue

Closes #4569, #4720, #3720, #3671.

## Checklist

I still have a lot of doc strings, an overview of the predicate DSL implementation and some tests to write, hence marking this PR as a draft.

- [x] add more docs and remove all the `#[warn(missing_docs)]`
- [x] add more tests of the DSL
- [x] add a test checking the validation performed on `SignedQueryCandidate`
- [x] fix client cli tests
- [x] fix ui tests
- [x] make CI pass


Erigara:
Are projections smt we can introduce later? 
To get rid of queries like `FindAllRoleIds` into smt like `query_iter(FindAllRoles).with_projection(|role| role.id)`

DCNick3:
> Are projections smt we can introduce later?

Yeah, adding projections in the future is possible. Would also shave off a few specialized queries we have. Although I haven't really been thinking about possible designs.

Erigara:
> However, merging it as-is would mean that the primary queries would be slow. I can probably add the PK queries back for now, if that's unacceptable.

Degradation in PK quires might introduce overhead into permission checks in the executor, so we should be careful with it.
Imo we can return PK queries for now and than gradually replace them `filters` after optimization will be introduced.

data_model/src/query/mod.rs
```diff
@@ -31,199 +24,179 @@ use self::{
...
+        pub batch: IterableQueryOutputBatchBox,
+        pub continue_cursor: Option<ForwardCursor>,
+    }
 
-        Vec(
-            #[skip_from]
-            #[skip_try_from]
-            Vec<Self>,
-        ),
+    /// A type-erased iterable query, along with all the
```
Erigara, line 155: Looks like you forgot to complete sentence.

data_model/src/query/predicate/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    Or(Vec<Self>),
+}
+
+impl<Atom> CompoundPredicate<Atom> {
+    /// A compound predicate that always evaluates to `true`.
+    pub const PASS: Self = Self::And(Vec::new());
+    /// A compound predicate that always evaluates to `false`.
+    pub const FAIL: Self = Self::Or(Vec::new());
+
+    pub fn and(self, other: Self) -> Self {
```
Erigara, line 36: Should `or` be added as well?

Erigara:
Can you describe briefly highlevel idea behind new predicates? 

Like i can see now that we have `AstPredicate`, `AtomPredicate`, ...

What are stages predicate go though?

Erigara (commented):
I like that now quires are handled more uniformly between client and wasm. 

Also that now each query is coupled with specific filters is convenient as well/

core/src/smartcontracts/isi/query.rs
```diff
@@ -6,283 +6,329 @@ use eyre::Result;
...
+    smartcontracts::{wasm, ValidIterableQuery},
     state::{StateReadOnly, WorldReadOnly},
 };
 
-/// Represents lazy evaluated query output
-pub trait Lazy {
-    /// Type of the lazy evaluated query output
-    type Lazy<'a>;
+/// Allows to generalize retrieving the metadata key for all the query output types
+pub trait SortableQueryOutput {
```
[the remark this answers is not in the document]
DCNick3: Yeah, I do find it weird too, but I decided not to leave it as-is in this PR

client/tests/integration/queries/query_errors.rs
```diff
@@ -15,16 +10,15 @@ fn non_existent_account_is_specific_error() {
...
 
     let err = client
-        .request(client::account::by_id(gen_account_in("regalia").0))
+        .iter_query(client::account::all())
+        .with_filter(|account| account.id.eq(gen_account_in("regalia").0))
+        .execute_single()
         .expect_err("Should error");
 
+    // NOTE: with the current implementation, the error returned is less specific than it was before
+    // (we no longer get the context that it is the entire domain that doesn't exist, not just the account)
```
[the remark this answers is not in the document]
DCNick3: It was more as a comment for the review and for myself, yeah. I'll remove it

client/tests/integration/queries/role.rs
```diff
@@ -91,7 +89,10 @@ fn find_role_by_id() -> Result<()> {
     let register_role = Register::role(new_role.clone());
     test_client.submit_blocking(register_role)?;
 
-    let found_role = test_client.request(client::role::by_id(role_id))?;
+    let found_role = test_client
+        .iter_query(client::role::all())
```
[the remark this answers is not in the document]
DCNick3: Are you talking about `execute_single`? I found them to be quite useful in a bunch of places, as they reduce boilerplate and provide convenient way to error out if unexpected number of results is returned

client/tests/integration/sorting.rs
```diff
@@ -107,13 +99,11 @@ fn correct_pagination_assets_after_creating_new_one() {
         .expect("Valid");
 
     let queried_assets = test_client
-        .build_query(client::asset::all())
-        .with_filter(xor_filter)
+        .iter_query(client::asset::all())
+        .with_raw_filter(xor_filter)
```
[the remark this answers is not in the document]
DCNick3: `with_filter` accepts a closure that constructs a predicate (`|asset| ..`), while `with_raw_filter` accepts an already built predicate (which was built by `AssetPredicateBox::build(|asset| ...)` before and got reused in several queries).

Maybe there's a more clear name, but this is the best I've got..

client/tests/integration/queries/role.rs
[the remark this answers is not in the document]
Erigara: no, execute_single is fine imo

i mean client::role::all() and etc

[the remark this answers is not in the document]
DCNick3: Hm.. I just strived for consistency with those tbh: I think the original intention was to use them in all client tests. You think I should remove all of them in this PR?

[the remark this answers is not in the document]
Erigara: No, i don't think we should stretch the scope of you pr. 
It's just something to consider.

core/src/smartcontracts/isi/domain.rs
```diff
@@ -391,32 +391,34 @@ pub mod isi {
...
-    }
-
-    impl ValidQuery for FindDomainById {
-        #[metrics(+"find_domain_by_id")]
-        fn execute(&self, state_ro: &impl StateReadOnly) -> Result<Domain, Error> {
-            let id = &self.id;
-            iroha_logger::trace!(%id);
-            Ok(state_ro.world().domain(id)?.clone())
+        ) -> std::result::Result<impl Iterator<Item = Domain> + 'state, Error> {
+            Ok(state_ro
```
Erigara, line 413: I've checked and it look like that some simple cases indeed could be optimized quite easily!
This can be left for follow up PR.

```suggestion
            let iter: Box<dyn Iterator<Item = Domain>> = match filter {
                CompoundPredicate::Atom(DomainPredicateBox::Id(DomainIdPredicateBox::Equals(
                    domain_id,
                ))) => Box::new(Some(state_ro
                    .world()
                    .domain(&domain_id)?
                    .clone())
                    .into_iter()),
                filter => Box::new(
                    state_ro
                        .world()
                        .domains_iter()
                        .filter(move |&v| filter.applies(v))
                        .cloned(),
                ),
            };

            Ok(iter)
```

core/src/smartcontracts/isi/domain.rs
```diff
@@ -391,32 +391,34 @@ pub mod isi {
...
+                .world()
+                .domains_iter()
+                .filter(move |&v| filter.applies(v))
+                .cloned())
         }
     }
 
-    impl ValidQuery for FindDomainKeyValueByIdAndKey {
+    impl ValidSingularQuery for FindDomainKeyValueByIdAndKey {
```
Erigara, line 421: What do you think about replacing this type of queries with `FindDomainMetadata` and then use filters to get specific keys?

DCNick3:
@Erigara 
> Can you describe briefly describe idea behind new predicates?
> Like i can see now that we have AstPredicate, AtomPredicate, ...
> What are stages predicate go though?

I've pushed some docs, including explanations of this topic. You would probably want to take a look at `query::predicate::{self, prototypes, projectors}`. If I missed any of the topics or any of the explanations are bad - please let me know!

core/src/smartcontracts/isi/domain.rs
DCNick3: I am kinda split on this... This won't reduce the number of queries we have, but it will improve flexibility just a little bit..

Though I feel projections are a more general solution for this

core/src/smartcontracts/isi/domain.rs
DCNick3: I've also envisioned a helper functions for this matching, probably this can make it a bit less boilerplaty

core/src/smartcontracts/isi/query.rs
```diff
@@ -6,283 +6,329 @@ use eyre::Result;
...
+    fn get_metadata_sorting_key(&self, key: &Name) -> Option<JsonString>;
 }
 
-/// Lazily evaluated equivalent of [`Query::Output`]
-pub enum LazyQueryOutput<'a> {
-    /// Concrete computed [`Query::Output`]
-    QueryOutput(QueryOutputBox),
-    /// Iterator over a set of [`Query::Output`]s
-    Iter(Box<dyn Iterator<Item = QueryOutputBox> + 'a>),
+impl SortableQueryOutput for Account {
```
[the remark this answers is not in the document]
DCNick3: No, because of this limitation: https://github.com/rust-lang/rfcs/issues/2758.

Probably would've worked if this were not cross-crate (the type definitions are from data model, while the impl is in core), but alas

data_model/src/query/predicate/mod.rs
DCNick3: Added along with `not` method, and also improved representation of and/or chains by more aggressively flattening them and actually using these methods in the AST combinator normalization routines.

client/src/query.rs
```diff
@@ -0,0 +1,302 @@
...
+        Q: IterableQuery,
+        Q::Item: HasPredicateBox<PredicateBoxType = P>,
+    {
+        IterableQueryBuilder::new(self, query)
+    }
+
+    /// Make a request to continue an iterable query with the provided raw [`ForwardCursor`]
+    ///
+    /// You probably do not want to use this function, but rather use the [`Self::iter_query`] method to make a query and iterate over its results.
+    pub fn raw_continue_iterable_query(
```
mversic, line 251:
```suggestion
    fn raw_continue_iterable_query(
```

is there a reason to have it public?

mversic:
I think that we should get rid of singular queries. We should have queries that return multiple results but can be filtered to return just one. Would you agree?

data_model/src/query/builder.rs
```diff
@@ -0,0 +1,304 @@
...
+            (Some(result), None) => Ok(Some(result)),
+            (Some(_), Some(_)) => Err(SingleQueryError::ExpectedOneOrZeroGotMany.into()),
+            (None, Some(_)) => {
+                unreachable!()
+            }
+        }
+    }
+
+    /// Execute the query, constraining the number of results to exactly one. Errors if zero or more than one result is returned.
+    pub fn execute_single(self) -> Result<Q::Item, E::SingleError> {
```
mversic, line 273: do we need to support this special case? seems to me a client should deal with it themselves

mversic (changes requested):

client/src/query.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        let request = QueryRequest::Singular(query);
+
+        let response = request_head.assemble(request).build()?.send()?;
+        let response = decode_singular_query_response(&response)?;
+
+        Ok(response)
+    }
+
+    fn start_iterable_query(
```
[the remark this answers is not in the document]
mversic: `iterable` should be the default and the fact that it's query is implicit in the trait name

data_model/src/query/builder.rs
mversic: also, it's not necessary to check that 2nd result doesn't exist. That's an invariant of the blockchain. A user will execute a query and just take the 1st result. They will not be checking if there is more results. So I don't think method is needed in the API

smart_contract/src/lib.rs
```diff
@@ -107,329 +111,118 @@ impl<I: BuiltInInstruction + Encode> ExecuteOnHost for I {
...
-            batch: typed_value,
-            cursor,
-        })
-    }
+    IterableQueryBuilder::new(&SmartContractQueryExecutor, query)
 }
 
-impl<Q> SmartContractQuery<'_, Q>
+/// Run a singular query in a smart contract.
+pub fn query<Q>(query: Q) -> Result<Q::Output, ValidationFail>
```
mversic, line 212:
```suggestion
pub fn query_single<Q>(query: Q) -> Result<Q::Output, ValidationFail>
```

smart_contract/src/lib.rs
```diff
@@ -107,329 +111,118 @@ impl<I: BuiltInInstruction + Encode> ExecuteOnHost for I {
...
+        Ok((
+            batch,
+            cursor.map(|cursor| SmartContractQueryCursor { cursor }),
+        ))
     }
 }
 
-impl<Q> SmartContractQuery<'_, Q>
+/// Build an iterable query for execution in a smart contract.
+pub fn iter_query<Q>(
```
mversic, line 196:
```suggestion
pub fn query<Q>(
```

data_model/src/query/mod.rs
```diff
@@ -31,199 +23,207 @@ use self::{
...
-/// A [`Query`] that returns an iterable collection of values
-pub trait IterableQuery: Query {
-    /// A type of single element of the output collection
-    type Item;
+/// A query that returns an iterable collection of values
+///
+/// Iterable queries logically return a stream of items.
+/// In the actual implementation, the items collected into batches and a cursor is used to fetch the next batch.
+/// [`builder::IterableQueryIterator`] abstracts over this and allows the query consumer to use a familiar [`Iterator`] interface to iterate over the results.
+pub trait IterableQuery: Sealed {
```
mversic, line 53:
```suggestion
pub trait Query: Sealed {
```

in line with other changes I would make the fact that it's `Iterable` implicit and have `Singular` as a special case

client/src/query.rs
DCNick3: Yes, it is used in `live_query_is_dropped_after_smart_contract_end` test. It purposefully breaks abstraction, and this method enables it.

client/src/query.rs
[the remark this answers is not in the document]
DCNick3: Can agree on the iterable part, but removing query from the name is a bad idea. `Client` implements this trait, so, if one imports the query, this method can get exposed on it (even though it is not the main API to use). Because of that I made the name more specific.

data_model/src/query/builder.rs
DCNick3: > do we need to support this special case? seems to me a client should deal with it themselves

This _is_ a way for the client to deal with it themselves, just reducing boilerplate. The check is fully implemented on the client side.

> it's not necessary to check that 2nd result doesn't exist. That's an invariant of the blockchain.

Just so that we are on the same page: this methods executes an _iterable_ query (which can return different amount of results depending on the parameters). If iroha returns more than one result it may as well be an ill-specified query, and so we should notify the user that their intention is not met. This also has zero additional cost of network traffic in common case of batch size not being 1, so I don't see why shouldn't we check this.

DCNick3:
> I think that we should get rid of singular queries. We should have queries that return multiple results but can be filtered to return just one. Would you agree?

I've already removed all singular queries that can be implemented as a query + filter. Some of them can be removed (by introducing projections into the query system), but others, like `FindExecutorDataModel`, just don't make sense as iterable queries.

data_model/src/query/mod.rs
DCNick3: I am fine with making iterable queries the default case in the API methods, but would prefer to keep the trait names explicit. You would use them more on the implementation side, where I feel it's more important to think of the type of the query you are working with

core/src/smartcontracts/isi/query.rs
```diff
@@ -6,283 +6,329 @@ use eyre::Result;
...
-    ) -> Result<ProcessedQueryOutput, Error> {
-        let query = &self.0;
-
-        query.query().execute(state_ro)?.apply_postprocessing(
-            query.filter(),
-            query.sorting(),
-            query.pagination(),
-            query.fetch_size(),
-        )
+    /// The validation logic is defined by the implementation of the [`ValidateQueryOperation`] trait.
```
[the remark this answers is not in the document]
DCNick3: Yep, AFAIU, that's what `ValidateQueryOperation` impls do:

https://github.com/hyperledger/iroha/blob/7771577c6d34766687175763977a230f69d81b92/core/src/smartcontracts/wasm.rs#L565-L593

client/tests/integration/queries/account.rs
```diff
@@ -66,8 +67,8 @@ fn find_accounts_with_asset() -> Result<()> {
     );
 
     let found_accounts = test_client
-        .request(client::account::all_with_asset(definition_id))?
-        .collect::<QueryResult<Vec<_>>>()?;
+        .query(client::account::all_with_asset(definition_id))
+        .execute_all()?;
```
mversic, line 71:
```suggestion
        .execute()?;
```

not sure if I mentioned this already, but also this should be the default IMO

data_model/src/query/builder.rs
mversic: > If iroha returns more than one result it may as well be an ill-specified query, and so we should notify the user that their intention is not met

ok, I'm getting your point. But why is this a recoverable error? This indicates a bug and there is no reasonable way to recover from it

client/tests/integration/queries/account.rs
mversic: or how about just:
```suggestion
        .all()?;
```

client/tests/integration/pagination.rs
```diff
@@ -13,14 +13,13 @@ fn limits_should_work() -> Result<()> {
 
     register_assets(&client)?;
 
-    let vec = &client
-        .build_query(asset::all_definitions())
+    let vec = client
+        .query(asset::all_definitions())
```
mversic, line 17: I don't like it that on the http client we have API like `asset::all_definitions()`. Because it differs from what we have in smart contracts. IMO there is nothing wrong in constructing a query like this:

```suggestion
    let vec = client
        .query(FindAllAssetDefinitions::new())
```

Definitely not intended for this PR, but would like to get an opinion on this

client/tests/integration/queries/account.rs
mversic: wait, I'll collect my remarks separately

data_model/src/query/builder.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        Ok(iterator)
+    }
+
+    /// Execute the query, returning all the results collected into a vector.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the query execution fails.
+    pub fn execute_all(self) -> Result<Vec<Q::Item>, E::Error> {
```
mversic, line 279: I would like to remove this method. Anybody can collect an iterator into a struct they want to, doesn't have to be a vec. And this method barely reduces any boilerplate

data_model/src/query/builder.rs
```diff
@@ -0,0 +1,339 @@
...
+                sorting: self.sorting,
+                fetch_size: self.fetch_size,
+            },
+        };
+
+        let (first_batch, continue_cursor) = self.query_executor.start_query(query)?;
+
+        let iterator = IterableQueryIterator::<E, Q::Item>::new(first_batch, continue_cursor)
+            .expect(
+                "iroha returned unexpected type in iterable query. Is there a schema mismatch?",
```
mversic, line 268:
```suggestion
                "INTERNAL BUG: iroha returned unexpected type in iterable query. Is there a schema mismatch?",
```

mversic (changes requested):

client/tests/integration/sorting.rs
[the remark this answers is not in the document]
mversic: I'd suggest that we have `filter_with` which takes closure and `filter` which takes preconstructed filter

core/src/smartcontracts/isi/domain.rs
mversic: at the very least let's rename `FindDomainKeyValueByIdAndKey` to `FindDomainMetadata`.
But my question is, if we have `FindAllDomains`, isn't it possible to get rid of this query?

data_model/src/query/mod.rs
```diff
@@ -31,199 +23,207 @@ use self::{
...
-        FindAllAssetsDefinitions(FindAllAssetsDefinitions),
-        FindAssetById(FindAssetById),
-        FindAssetDefinitionById(FindAssetDefinitionById),
-        FindAssetsByName(FindAssetsByName),
-        FindAssetsByAccountId(FindAssetsByAccountId),
-        FindAssetsByAssetDefinitionId(FindAssetsByAssetDefinitionId),
-        FindAssetsByDomainId(FindAssetsByDomainId),
-        FindAssetsByDomainIdAndAssetDefinitionId(FindAssetsByDomainIdAndAssetDefinitionId),
+    pub enum IterableQueryBox {
+        FindAllDomains(IterableQueryWithFilterFor<FindAllDomains>),
```
mversic, line 89:
```suggestion
        FindDomains(IterableQueryWithFilterFor<FindAllDomains>),
```

I think that now it makes sense to drop `All` in iterable queries

data_model/src/query/mod.rs
```diff
@@ -31,199 +23,207 @@ use self::{
...
+        TriggerId(Vec<TriggerId>),
+        Block(Vec<SignedBlock>),
+        BlockHeader(Vec<BlockHeader>),
+    }
+
+    /// An enum of all possible singular queries
+    #[derive(
+        Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema, FromVariant,
+    )]
+    pub enum SingularQueryBox {
```
mversic, line 134: from what I can see, it seems to me that, in the end, only `FindExecutorDataModel` will remain as a singular query? Is that indeed the case?

mversic (changes requested):

client/tests/integration/smartcontracts/create_nft_for_every_user_trigger/src/lib.rs
```diff
@@ -19,7 +19,7 @@ getrandom::register_custom_getrandom!(iroha_trigger::stub_getrandom);
 fn main(_id: TriggerId, _owner: AccountId, _event: EventBox) {
     iroha_trigger::log::info!("Executing trigger");
 
-    let accounts_cursor = FindAllAccounts.execute().dbg_unwrap();
+    let accounts_cursor = query(FindAllAccounts).execute().dbg_unwrap();
```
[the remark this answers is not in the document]
DCNick3: I think it would be a nice improvement bridging the gap between smart contracts and http clients

client/tests/integration/smartcontracts/executor_custom_data_model/src/complex_isi.rs
```diff
@@ -155,7 +174,7 @@ mod expression {
         /// Greater expression.
         Greater(Greater),
         /// Query to Iroha state.
-        Query(QueryBox),
+        Query(NumericQuery),
```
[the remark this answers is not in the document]
DCNick3: Mostly the second, I prefered to have explicit types of the query and remove the need for a downcast from `SingularQueryOutputBox`.

client/tests/integration/pagination.rs
DCNick3: I have no emotional attachment to these helpers, but I think a similar syntax can be used if we decide to ever provide helpers for constructing pre-made queries with filters, to make actions like "get object by primary key" less boilerplaty. Although this API won't work as-is, not sure how exactly it would look like (and if we even want those boilerplate-reduction utilities).

client/tests/integration/smartcontracts/executor_custom_data_model/src/complex_isi.rs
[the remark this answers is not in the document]
mversic: than that is ok, I wanted to know if some limitation was introduced

mversic (changes requested):

core/src/smartcontracts/isi/domain.rs
DCNick3: > But my question is, if we have FindAllDomains, isn't it possible to get rid of this query?

It is possible, but only if you don't consider the additional traffic generated by this: you need to fetch the full `Domain` object and then query the metadata KV store on the client side. `FindDomainKeyValueByIdAndKey`, on the other hand, does its lookup on the peer's side and only sends the queried metadata value.

data_model/src/query/builder.rs
DCNick3: I personally find `.execute().collect::<Result<Vec<_>, _>>()` a bit too unwieldy to type when this is such a common case. In most cases `Vec` of results is much easier to work with, so I believe it is worthwhile to make this easier to do.

data_model/src/query/builder.rs
mversic: let's keep it

client/tests/integration/sorting.rs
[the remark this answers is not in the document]
DCNick3: Not a fan of the more common method having a longer name, but it rhymes nicely with stdlib naming conventions (`_with` functions take closures), so I'll implement it

data_model/src/query/mod.rs
DCNick3: Yes, I think with projections implemented and some refactors to the queries, all the other ones will be representable as iterable queries. Although this assumes "smart" projections, like "get metadata value with key `42`" and not just field accesses.

mversic: which means that there will no longer be any dichotomy between singular and iterable queries. It doesn't make sense to have 1 singular query.

This is the reason why I want to subjugate defaults to iterable queries already now, so we don't change api later

smart_contract/src/lib.rs
```diff
@@ -107,329 +111,122 @@ impl<I: BuiltInInstruction + Encode> ExecuteOnHost for I {
...
-    fn sort(&self, sorting: Sorting) -> SmartContractQuery<Self>;
-
-    /// Apply pagination to a query
-    fn paginate(&self, pagination: Pagination) -> SmartContractQuery<Self>;
-
-    /// Set fetch size for a query. Default is [`DEFAULT_FETCH_SIZE`]
-    fn fetch_size(&self, fetch_size: FetchSize) -> SmartContractQuery<Self>;
+/// An error that can occur when executing a singular query in a smart contract.
+#[derive(Debug, displaydoc::Display, FromVariant)]
+pub enum SmartContractSingleQueryError {
```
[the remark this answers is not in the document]
DCNick3: This name conflicts with the error this type wraps as one of the variants:

```rust
pub enum SingleQueryError {
    /// Expected exactly one query result, got none
    ExpectedOneGotNone,
    /// Expected exactly one query result, got more than one
    ExpectedOneGotMany,
    /// Expected one or zero query results, got more than one
    ExpectedOneOrZeroGotMany,
}
```

I think we can rename this to `SingleError` and then make `SmartContractSingleQueryError` into `SingleQueryError`, but this feels a bit confusing to me..

DCNick3:
Sonar-Dojo job is failing with some very cryptic error: https://github.com/hyperledger/iroha/actions/runs/10081046371/job/27872356479#step:6:2149

Is this OK?

client/tests/integration/pagination.rs
mversic: #4892

data_model/src/query/builder.rs
mversic: Unless popular vote is against me, I'm insisting on this one

Erigara:
> Sonar-Dojo job is failing with some very cryptic error: https://github.com/hyperledger/iroha/actions/runs/10081046371/job/27872356479#step:6:2149
> 
> Is this OK?

I had the same error, it was fixed after i rebased my PR, maybe it will help you as well.

data_model/src/query/builder.rs
Erigara: Maybe it's possible to do through extension trait?
Than our bare-bone api will be minimal with just `execute` and other `execute_single`, `execute_all` could be part of extension trait?

smart_contract/src/lib.rs
```diff
@@ -474,74 +271,33 @@ pub mod prelude {
...
-    };
+    pub use crate::{data_model::prelude::*, ExecuteOnHost};
 }
 
 #[cfg(test)]
 mod tests {
     use core::{mem::ManuallyDrop, slice};
 
-    use data_model::{prelude::numeric, query::asset::FindAssetQuantityById, BatchedResponseV1};
+    // use data_model::{prelude::numeric, query::asset::FindAssetQuantityById};
```
Erigara, line 281: Commented code.

data_model/src/query/predicate/predicate_atoms/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            StringPredicateBox::EndsWith(content) => input.ends_with(content),
+            StringPredicateBox::Equals(content) => *input == *content,
+        }
+    }
+}
+
+/// A predicate that can be applied to [`Metadata`].
+#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize, IntoSchema)]
+pub enum MetadataPredicateBox {
+    // TODO: populate this with something. Seeing as how we can change it to be just a JsonString, not populating it right now
```
Erigara, line 149: already a JsonString

data_model/src/query/predicate/predicate_atoms/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+impl_predicate_box!(Metadata: MetadataPredicateBox);
+
+impl PredicateTrait<Metadata> for MetadataPredicateBox {
+    fn applies(&self, _input: &Metadata) -> bool {
+        match *self {}
```
Erigara, line 156: Thought for some time how this result in `bool`...
Before realizing that `!` can be substitute any type.

data_model/src/query/builder.rs
DCNick3: Hmm, an extension trait does seem like a good fit here. @mversic what yo you think?

DCNick3:
> > Sonar-Dojo job is failing with some very cryptic error: https://github.com/hyperledger/iroha/actions/runs/10081046371/job/27872356479#step:6:2149
> > Is this OK?
> 
> I had the same error, it was fixed after i rebased my PR, maybe it will help you as well.

Didn't help, it's still failing =(

https://github.com/hyperledger/iroha/actions/runs/10097405719/job/27922177879#step:6:2149

data_model/src/query/predicate/predicate_atoms/mod.rs
mversic: me too

mversic (changes requested):

core/src/smartcontracts/isi/domain.rs
mversic: we can resolve this?

mversic (changes requested):

docs/source/references/schema.json
```diff
@@ -3739,6 +4016,109 @@
...
+        "discriminant": 0,
+        "type": "FindAssetQuantityById"
+      },
+      {
+        "tag": "FindExecutorDataModel",
+        "discriminant": 1,
+        "type": "FindExecutorDataModel"
+      },
+      {
+        "tag": "FindAllParameters",
```
0x009922, line 4032: This name stands out within singular queries and is confusing. `FindParameters`? `FindConfiguration`? @mversic

data_model/src/query/builder.rs
0x009922: From my experience with ORMs and "repo"-like APIs (which provide convenienes to query databases), they often expose split methods like 

- `.one() -> Option<T>`
- `.one!() -> T`
- `.all() -> T[]`

as they reflect the most common use cases.

So, I am in favor of having these shortcuts (which also give nice errors!), in a form of direct methods or an extension trait.

data_model/src/query/predicate/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+//! There are also two representations of the predicates:
+//! - Normalized representation, which is designed for serialization and evaluation
+//! - AST representation, which is designed for type-checking and easy & efficient composition
+//!
+//! Normalized representation consists of [`CompoundPredicate<T>`], with `T` being an atomic predicate box for that type (like [`predicate_atoms::account::AccountIdPredicateBox`]).
+//! The [`CompoundPredicate`] layer implements logical operators on top of the atomic predicate, while the projections are handled with the atomic predicate itself, with variants like [`predicate_atoms::account::AccountIdPredicateBox::DomainId`].
+//!
+//! Normalized representation aims to reduce the number of types not to bloat the schema and reduce redundancy.
+//!
+//! Predicates in the normalized representation can be evaluated using the [`PredicateTrait`] trait.
```
0x009922, line 30:
```suggestion
//! Predicates in the normalized representation can be evaluated using the [`EvaluatePredicate`] trait.
```
```suggestion
//! Predicates in the normalized representation can be evaluated using the [`ApplyPredicate`] trait.
```

data_model/src/query/mod.rs
```diff
@@ -250,303 +250,331 @@ mod model {
...
-
-impl<V: Into<QueryOutputBox>> From<Vec<V>> for QueryOutputBox {
-    fn from(values: Vec<V>) -> Self {
-        Self::Vec(values.into_iter().map(Into::into).collect())
-    }
+impl_iter_queries! {
+    FindRoles => crate::role::Role,
+    FindRoleIds => crate::role::RoleId,
+    FindRolesByAccountId => crate::role::RoleId,
+    FindPermissionsByAccountId => crate::permission::Permission,
```
0x009922, line 551: This information should be provided in the schema, so that SDK can auto-generate these mappings too.

For a different PR, I guess.

data_model/src/query/predicate/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+                .parse()
+                .unwrap();
+        let alice_domain_id: DomainId = "wonderland".parse().unwrap();
+
+        let account_predicate_denorm = AccountPredicateBox::build_fragment(|account| {
+            let account_id_predicate = AccountIdPredicateBox::build_fragment(|account_id| {
+                account_id
+                    .signatory
+                    .eq(alice_signatory.clone())
+                    .and(account_id.domain_id.eq(alice_domain_id.clone()))
```
0x009922, line 252: You mentioned in the PR description the use of `impl Into<T>` to reduce clutter. I guess it will also apply to these methods (if implemented in future)

```suggestion
                    .eq(&alice_signatory)
                    .and(account_id.domain_id.eq(&alice_domain_id))
```

core/src/smartcontracts/isi/account.rs
```diff
@@ -570,31 +558,31 @@ pub mod query {
...
-    impl ValidQuery for FindAccountsWithAsset {
+    impl ValidIterableQuery for FindAccountsWithAsset {
         #[metrics(+"find_accounts_with_asset")]
         fn execute<'state>(
-            &self,
+            self,
+            filter: CompoundPredicate<AccountPredicateBox>,
             state_ro: &'state impl StateReadOnly,
-        ) -> Result<Box<dyn Iterator<Item = Account> + 'state>, Error> {
+        ) -> std::result::Result<impl Iterator<Item = Account> + 'state, Error> {
```
nxsaken, line 567: `+ 'state` is too restrictive (it misleadingly means that the Iterator has to live _at least_ as long as the state, not _at most_ as long). It might be worth relaxing this bound using [the captures trick](https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html#the-captures-trick), which is strictly more correct:

```rust
trait Capture<U> {}
impl<T: ?Sized, U> Capture<U> for T {}

impl Iterator<Item = Account> + 'state -> impl Iterator<Item = Account> + Capture<&'state ()>
```

0x009922 (commented):
Magnificent work!

However, the "type-safe and easy & efficient composition" part doesn't translate to SDKs via schema. And it is a huge ergonomics feature for client-side. I am going to research how to represent it natively in TypeScript.

data_model/src/query/predicate/mod.rs
DCNick3: It is only a future work, I haven't actually done this for non-string types =(

data_model/src/query/mod.rs
DCNick3: Opened #4902

docs/source/references/schema.json
mversic: yes, we should remove `All` from query names, I commented on that. also remember to update `visit` method name

data_model/src/query/mod.rs
mversic: should be done sooner rather than later

data_model/src/query/builder.rs
```diff
@@ -0,0 +1,341 @@
...
+
+use crate::query::{
+    parameters::{FetchSize, IterableQueryParams, Pagination, Sorting},
+    predicate::{projectors, AstPredicate, CompoundPredicate, HasPredicateBox, HasPrototype},
+    IterableQuery, IterableQueryBox, IterableQueryOutputBatchBox, IterableQueryWithFilter,
+    IterableQueryWithFilterFor, IterableQueryWithParams, SingularQueryBox, SingularQueryOutputBox,
+};
+
+/// A trait abstracting away concrete backend for executing queries against iroha.
+pub trait QueryExecutor {
```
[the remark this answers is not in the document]
DCNick3: Currently this trait is used for both singular and iterable queries. Do you want to split them?

data_model/src/query/builder.rs
```diff
@@ -0,0 +1,341 @@
...
+    // bound introduced to inject `SingularQueryError` in builder's `execute_single` and `execute_single_opt` methods
+    /// An error type that combines `SingleQueryError` and `Self::Error`. Used by helper methods in query builder that constrain the number of results of an iterable query to one.
+    type SingleError: From<SingleQueryError> + From<Self::Error>;
+
+    /// Executes a singular query and returns its result.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the query execution fails.
+    fn execute_singular_query(
```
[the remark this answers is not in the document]
DCNick3: I thought we wanted to keep singular queries explicit? Feels a bit confusing to have `start_query` and `execute_query` which operate on completely different query types..

docs/source/references/schema.json
DCNick3: Whoops, I forgot to update this enum variant.

And also the metric names... Will fix it

core/src/smartcontracts/isi/account.rs
DCNick3: Huh, this is an unexpected sharp edge, thanks for pointing this out.

I am not sure if we should use the trick though: AFAIU, this leads to problems only if we use multiple lifetime parameters or have use another type with a different captured lifetime. We don't have any compile-time errors, so I am a bit hesitant to complicate the signature..

nxsaken: Agree, no need if this hasn't caused issues so far

docs/source/references/schema.json
```diff
@@ -3714,7 +3991,7 @@
       },
       {
         "name": "payload",
-        "type": "ClientQueryPayload"
+        "type": "QueryRequestWithAuthority"
```
0x009922, line 3994:
```suggestion
        "name": "request",
        "type": "QueryRequestWithAuthority"
```

DCNick3: I think the `payload` is fitting, as it signifies the part that is signed (signature payload). Not sure that `request` is better, especially considering `QueryRequestWithAuthority` itself also has a `request` field..

data_model/src/query/builder.rs
DCNick3: I've pushed a commit that makes these into an extension trait. Now that I've implemented it, I am not sure it's a good idea (it's a downgrade in ergonomics for no apparent benefit), but a positive change I see is decoupling of the singular logic from the `QueryExecutor` trait by making the `SingleQueryError` wrap the query error and not the other way around.

smart_contract/src/lib.rs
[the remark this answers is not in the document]
DCNick3: Done, no more `SmartContractSingleQueryError`

0x009922:
While trying to integrate changes made in this PR in the block explorer (https://github.com/soramitsu/iroha2-block-explorer-backend), faced an issue: `QueryBuilder` relies on `QueryExecutor` trait which doesn't support `async`. Therefore, it isn't usable there.

Not a big issue, it's still very easy to implement a custom client and query builder, but shall Iroha expose something like `QueryAsyncExecutor` trait?

DCNick3:
How was async handled before? I don't remember seeing any async query methods..

data_model/src/query/mod.rs
mversic: If we're going to remove `SingularQuery`, then I think you should rename it. But if you disagree, resolve the conversation

core/src/smartcontracts/isi/domain.rs
mversic: should we open a new issue for this?

client/src/query.rs
```diff
@@ -0,0 +1,310 @@
...
+            key_pair: self.key_pair.clone(),
+        }
+    }
+
+    /// Execute a singular query and return the result
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the query execution fails.
+    pub fn query_single<Q>(&self, query: Q) -> Result<Q::Output, ClientQueryError>
```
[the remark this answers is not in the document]
DCNick3: I feel like whether or not we implement projections, this will be a wasted effort.

If we implement projections, we are likely to be able to get rid of all singular queries (#4933), making the `FindExecutorDataModel` a special case. With no singular queries - no need for singular builders.

If we don't implement projections then I don't see any options we can define for the singular queries..

data_model/src/query/mod.rs
DCNick3: I am not confident that singular queries are possible to be removed (#4933), so I am ready to commit to the rename.

client/src/query.rs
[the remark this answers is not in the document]
mversic: let's close this

mversic (changes requested):

client/src/query.rs
```diff
@@ -0,0 +1,307 @@
...
+            "Got unexpected type of query response from the node (expected iterable)"
+        )
+        .into());
+    };
+    Ok(resp)
+}
+
+/// An iterable query cursor for use in the client
+#[derive(Debug)]
+pub struct ClientQueryCursor {
```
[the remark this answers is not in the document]
DCNick3: My idea was to make it different from `SmartContractQueryCursor` (which serves a similar function in smart contracts) and `ForwardCursor` (which is a lower level implementation detail).

I think it's fine to rename both `SmartContractQueryCursor` and `ClientQueryCursor` to `QueryCursor` as they won't get used in the same context. I see some potential for confusion with `ForwardCursor`, but let's hope this won't happen.

client/src/query.rs
```diff
@@ -0,0 +1,307 @@
...
+    http::{Method as HttpMethod, RequestBuilder},
+    http_default::DefaultRequestBuilder,
+};
+
+#[derive(Debug)]
+struct ClientQueryRequestHead {
+    pub torii_url: Url,
+    pub headers: HashMap<String, String>,
+    pub account_id: AccountId,
+    pub key_pair: KeyPair,
```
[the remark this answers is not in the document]
DCNick3: It doesn't for some reason. I agree that it should be private.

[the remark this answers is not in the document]
DCNick3: On that note, all fields of `Client` are also public. Not sure if it's intended or not.

smart_contract/executor/src/default.rs
```diff
@@ -1121,20 +1121,33 @@ pub mod parameter {
...
-            let find_role_query_res = match FindRoleByRoleId::new(role_id.clone()).execute() {
-                Ok(res) => res.into_inner(),
-                Err(error) => {
-                    deny!($executor, error);
-                }
-            };
+            let find_role_query_res =
+                match crate::data_model::query::builder::QueryBuilderExt::execute_single(
+                    iroha_smart_contract::query(FindRoles)
+                        .filter_with(|role| role.id.eq(role_id.clone())),
```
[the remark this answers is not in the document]
DCNick3: This is a limitation (or a feature) of rust: `==` doesn't allow overloading return type, as it's bound to the `Eq`/`PartialEq` traits.

mversic (dismissed):

client/tests/integration/queries/role.rs
```diff
@@ -38,8 +36,8 @@ fn find_roles() -> Result<()> {
 
     // Checking results
     let found_role_ids = test_client
-        .request(client::role::all())?
-        .collect::<QueryResult<Vec<_>>>()?
+        .query(client::role::all())
+        .execute_all()?
```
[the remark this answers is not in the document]
DCNick3: No, not really.

Looking at prior art, diesel has `load` which collects into a `Vec` and `load_iter` which gives you an iterator. 

I like the `_all` name more (as it specifies _what_ it does instead of specifying what type it returns), but I am not very averse to the `into_vec` name

client/tests/integration/pagination.rs
```diff
@@ -33,14 +32,14 @@ fn fetch_size_should_work() -> Result<()> {
...
         .with_pagination(Pagination {
             limit: Some(nonzero!(7_u32)),
-            start: Some(nonzero!(1_u64)),
+            offset: Some(nonzero!(1_u64)),
         })
         .with_fetch_size(FetchSize::new(Some(nonzero!(3_u32))))
         .execute()?;
-    assert_eq!(iter.batch_len(), 3);
+    assert_eq!(iter.remaining_in_current_batch(), 3);
```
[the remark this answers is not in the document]
DCNick3: The fix turned out to be not that hard, I've pushed a commit that removes it.

mversic (approved):

0x009922 (approved):

PR #4837, opened 2024-07-11
feat(logger): use RUST_LOG-like EnvFilter for logging

### LINKED ISSUE #4829: feat: introduce RUST_LOG like log level format
This will help us to reduce noisy logs from parts of the system we are not interested in.

Erigara:
## Description

Replace level filter with `EnvFilter` which support setting specific log level for different iroha modules.

## Examples

The same can be done through configuration in config.toml.

```bash
# set global logging level to debug
LOG_LEVEL="debug" ./scripts/test_env.py setup
# set global logging level to error and info for iroha_core
LOG_LEVEL="error,iroha_core=info" ./scripts/test_env.py setup
# receive only logs from sumerag with trace level or heigher
LOG_LEVEL="iroha_core::sumeragi" ./scripts/test_env.py setup
# receive all logs at trace level, except axum for which use error
LOG_LEVEL="trace,axum=error" ./scripts/test_env.py setup
```

### Linked issue

Closes #4829 

### Benefits

More granular control over logging.

### Downsides

It's harder to check correctness for example if someone would write `LOG_LEVEL=infa` iroha won't return any error because this would be parsed as accept logs from module `infa` at `trace` level.

I believe we can mitigate this, by custom parsing, but it would make our type more strict than orginal `EnvFilter`.
It's smt i would prefer to leave for separate PR.


logger/src/lib.rs
```diff
@@ -121,12 +120,15 @@ pub fn disable_global() -> Result<()> {
...
 }
 
+#[allow(clippy::needless_pass_by_value)]
 fn step2<L>(config: InitConfig, layer: L) -> Result<LoggerHandle>
 where
     L: tracing_subscriber::Layer<Registry> + Debug + Send + Sync + 'static,
 {
-    let level: tracing::Level = into_tracing_level(config.base.level);
-    let level_filter = tracing_subscriber::filter::LevelFilter::from_level(level);
+    // NOTE: unfortunately constructing EnvFilter from vector of Directive is not part of public api
```
DCNick3, line 128: There's a [PR open](https://github.com/tokio-rs/tracing/pull/2763), not sure why it wasn't merged

config/src/logger.rs
```diff
@@ -40,6 +35,68 @@ pub enum Format {
...
+        for directive in directives_iter {
+            write!(f, ",{directive}")?;
+        }
+        Ok(())
+    }
+}
+
+impl Debug for Directives {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self, f)
```
DCNick3, line 73: `tracing`'s `Debug` implementation for `Directive` is actually done with `#[derive(Debug)]` (looks like `Directive { in_span: None, fields: [], target: None, level: LevelFilter::INFO }`), should we follow them maybe?

Or do we want to pretty-print it as part of some structure?

config/src/logger.rs
```diff
@@ -40,6 +35,68 @@ pub enum Format {
...
+}
+
+impl Default for Directives {
+    fn default() -> Self {
+        Self::from(Level::INFO)
+    }
+}
+
+/// Convert [`Level`] into [`tracing::Level`]
+pub fn into_tracing_level(level: Level) -> tracing::Level {
```
DCNick3, line 90: Should this be public API? The only place I see it used is right in this file..

Erigara:
> > It's harder to check correctness for example if someone would write LOG_LEVEL=infa iroha won't return any error because this would be parsed as accept logs from module infa at trace level.
> 
> 
> 
> why is default `TRACE`?

idk, lowest possible level is assumed by tracing crate

config/src/parameters/user.rs
```diff
@@ -346,14 +346,14 @@ impl Queue {
...
 }
 
-#[derive(Debug, Clone, Copy, Default, ReadConfig)]
+#[derive(Debug, Clone, Default, ReadConfig)]
 pub struct Logger {
     /// Level of logging verbosity
     // TODO: parse user provided value in a case insensitive way,
     //       because `format` is set in lowercase, and `LOG_LEVEL=INFO` + `LOG_FORMAT=pretty`
     //       looks inconsistent
     #[config(env = "LOG_LEVEL", default)]
```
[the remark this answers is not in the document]
Erigara: I remember some discussion about that some time ago, but don't remember why we settled in with this name. 
RUST_LOG name sounds ok to me, but i'm not sure it's wroth all renaming for devops team.

config/src/logger.rs
Erigara: i've used pretty printing for debug, because it it used in one of our configuration tests

config/src/parameters/user.rs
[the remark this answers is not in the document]
DCNick3: Probably not worth it, yeah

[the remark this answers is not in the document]
mversic: I think `LOG_LEVEL` is quite appropriate here

config/tests/fixtures.rs
```diff
@@ -172,7 +172,7 @@ fn minimal_config_snapshot() {
                 idle_time: 30s,
             },
             logger: Logger {
-                level: INFO,
+                level: info,
                 format: Full,
```
mversic, line 176: I notice a difference in casing

mversic: let's open an issue to standardize this

Erigara: you can still specify TRACE, DEBUG, INFO, WARN, ERROR

lower-case affect only serialisation

core/src/kiso.rs
```diff
@@ -29,7 +29,7 @@ impl KisoHandle {
     /// Spawn a new actor
     pub fn new(state: Config) -> Self {
         let (actor_sender, actor_receiver) = mpsc::channel(DEFAULT_CHANNEL_SIZE);
-        let (log_level_update, _) = watch::channel(state.logger.level);
+        let (log_level_update, _) = watch::channel(state.logger.level.clone());
```
mversic, line 32: not sure if this is appropriate place to comment. What happens in the following scenario:
1. We set LOG_LEVEL="error,iroha_core=info"
2. We dynamically update LOG_LEVEL="debug"

what is `iroha_core` log level now?

Erigara: new `LOG_LEVEL` completely overwrites previous one, so `iroha_core` new log level is `debug` as for every other module

logger/src/lib.rs
Erigara: Let's hope it will be merged soon and we would have more ergonomic API

core/src/kiso.rs
mversic: I thought so, and I find this behavior unexpected

config/tests/fixtures.rs
mversic: I meant that we have `info` for `level` on the one hand and `Full` for `format` on the other hand. I don't mind having all lowercase or all uppercase, but I don't like seeing cammel case. I wasn't complaining about your change

core/src/kiso.rs
mversic: ok, this wasn't a good example:
1. We set LOG_LEVEL="error,iroha_core=info"
2. We dynamically update LOG_LEVEL="iroha_core=debug"

will this affect only `iroha_core`, or will every module be set to default?

config/tests/fixtures.rs
Erigara: i see, i thought you said you noticed the difference between `info` and `INFO` :)

Erigara: Created #4840

core/src/kiso.rs
mversic: I think that dynamic update should be granular, what do you think? But it's fine for me to merge this PR

mversic (approved):

DCNick3 (approved):

core/src/kiso.rs
Erigara: for granular updates i think we should use different endpoint like `PATCH /configuration` instead of `POST /configuration`

DCNick3: > I think that dynamic update should be granular

I find the current behavior (updating the full filter string, overwriting whatever was there before, as if you set a new configuration value) more intuitive (and simpler to implement).

If the worst comes to the worst, one can still get the configuration, add the necessary filters and set it back

Issue #4848, opened 2024-07-16
Consider supporting non-graceful shutdown in Kura


dima74:
Consider a non-graceful shutdown occurring to Iroha in the middle of Kura saving data to disk (for example, due to power loss or being killed by the OOM killer). In this case, the data on disk might be in an incomplete or inconsistent state, leading to data corruption. This can be solved using custom [write-ahead logging](https://en.wikipedia.org/wiki/Write-ahead_logging) in Kura or crates like [this](https://github.com/khonsulabs/okaywal).

Issue #4830, opened 2024-07-10
Investigate how Iroha operates under load under Ubuntu/Debian


dima74:
It was reported that:
* Iroha container under Ubuntu restarts after ~10 minutes
* Iroha container under Debian restarts after ~7 hours
* After restart Iroha fails to start because of kura.lock
Need to investigate it

dima74:
Reproduced the problem, here are some results
* Preconditions:
  * Testing only `FindAllAccounts` query
  * Using [genesis with 2500 accounts](https://github.com/soramitsu/iroha2-perf/blob/54bcff91b0e5a899881c043839e93d00dcd5f542/src/test/resources/iroha2_config/genesis.json)
  * Using iroha version 50e81042aaca79527beae8ae66ced372a61b2e53
  * Using [iroha2-perf](https://github.com/soramitsu/iroha2-perf) load generator version 44b2499839ee6a6851dd9f39bb5bfa8d02b4fa6c
  * Testing is performed locally, results in the docker/k8s should be similar
* Observed problem that iroha consumes a lot of memory even under small load:
  * 10 queries per second ~2.6GB
  * 20 queries per second ~6.6GB
  * 30 queries per second ~11.3GB
* Background: how queries are handled
  * Consider iroha received query `FindAllAccounts`. It executes it and finds all 2500 accounts. Then it splits response to batches (default batch size is 10), returns first batch to the client, and saves remaining batches to `LiveQueryStore`. When client requests second batch, it is removed from `LiveQueryStore`. Normally client requests all batches, so after it query data will be removed from `LiveQueryStore`. However if client doesn't request batches, `LiveQueryStore` has special pruning mechanism - every 30 seconds it removes old queries (inactive more then 30 seconds).
* Background: how load generator works
  * When making query request in the load generator, [currently only first batch is requested](https://github.com/soramitsu/iroha2-perf/blob/44b2499839ee6a6851dd9f39bb5bfa8d02b4fa6c/src/test/java/requests/Queries.java#L65-L71).
* So iroha uses a lot of memory in such scenario because it stores all queries results for last 30 seconds in `LiveQueryStore`
* So:
  * I think load generator should be modified to request the whole data, not only the first batch (either retrieve all batches in multiple requests, or increase batch size)
  * Iroha can be easily DDOSed by requesting only first batch of heavy query. This might be a security problem

mversic:
>> Iroha can be easily DDOSed by requesting only first batch of heavy query. This might be a security problem

[...]

PR #4849, opened 2024-07-16
fix: Remove kura.lock

dima74:
## Description

Originally `kura.lock` [was introduced](https://github.com/hyperledger/iroha/issues/3027) to prevent multiple iroha instances running with the same data directory. However if iroha doesn't remove `kura.lock`, e.g. in case of non-graceful shutdown, it causes problem that iroha couldn't start after restart. Also usually iroha is deployed using k8s cluster where each peer lives in different container so original problem is not relevant in this case. So it is proposed to remove `kura.lock`.

### Linked issue

Related: #4830
Related: #4848

### Benefits

Remove `kura.lock`, allowing Iroha to restart after non-graceful shutdown

### Checklist

### DIFF
--- a/core/benches/kura.rs
+++ b/core/benches/kura.rs
@@ -7,7 +7,7 @@ use criterion::{criterion_group, criterion_main, Criterion};
 use iroha_config::{base::WithOrigin, parameters::actual::Kura as Config};
 use iroha_core::{
     block::*,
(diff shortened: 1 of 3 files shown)


nxsaken (approved):

Erigara (approved):
lgtm

PR #4853, opened 2024-07-16
refactor: preserve voting block during block sync

### LINKED ISSUE #4834: Keep voting block in more cases
> Cases when voting block going to be discarded falsely:
> 
> Case 1:
> 
> 1. peer already have voting block
> 2. new block is received through block created message
> 3. this block has proper height, matching topology, proper leader signature
> 4. BUT incorrect transactions in the new block
> 
> This means that leader send two blocks.
> 
> I don't think this is issue because after such event current leader would loose it's place in the validators set.
> 
> And no one other than current leader can perform this operation.
> 
> Case 2:
> 
> 1. peer already have voting block
> 2. new block received through block sync message
> 3. this block has proper height (for soft-fork or ordinary condition), view change index larger than current one (!!!), proper leader signature for this larger signature
> 4. here is two options why this block might be incorrect:
>    
>    * incorrect transactions
>    * missing proxy tail signature, not enough voting peers signatures
> 
> In case when peer conclude that transactions are incorrect but block has all valid signatures indicates that something is certainly wrong with peer itself.
> 
> Case when block has missing signatures (and has incorrect transactions) opens possibility for the following attack by any peer in the topology:
> 
> 1. find view change index (higher than current once)
> 2. create block
> 3. sign it
> 4. send it through block sync to other peers
> 5. peers will discard current voting block and this block going to be rejected
> 6. current view change index will fail
> 7. repeat the same process for the new view change index
> 
> To mitigate this issue we could check all signatures before trying to execute transactions for the given block (and discarding previous voting block).

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/4828#discussion_r1672594229_

Erigara:
## Description

This change allow to preserve voting block during block sync until iroha have to drop it to process transactions.

It is done in two ways: 
1. check block signatures before verification
2. try verify block without discarding voting block

### Linked issue

Closes #4834 

### Benefits

Preserve voting block in more cases.

### Downsides

Resulting code is not pretty.
Because it has to handle soft-fork case explicitly and not just revert previous block and proceed.

### How To Test

```
cargo test --package iroha_core --lib -- sumeragi::main_loop::tests::block_sync_commit_err_keep_voting_block --exact --show-output
```

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -402,6 +402,7 @@ impl Sumeragi {
             genesis_account,
             state,
             existing_voting_block,
+            false,
         )
         .unpack(|e| self.send_event(e))
         .map(|(block, state_block)| VotingBlock::new(block, state_block))
@@ -449,11 +450,6 @@ impl Sumeragi {
                 );
 
                 let block_sync_type = categorize_block_sync(&block, &state.view());
-                if block_sync_type.is_ok() {
-                    // Release block writer before creating new one
-                    let _ = voting_block.take();
-                }
-
                 match handle_categorized_block_sync(
                     &self.chain_id,
                     block,
@@ -461,6 +457,7 @@ impl Sumeragi {
                     genesis_account,
                     &|e| self.send_event(e),
                     block_sync_type,
+                    voting_block,
                 ) {
                     Ok(BlockSyncOk::CommitBlock(block, state_block, topology)) => {
                         self.topology = topology;
@@ -1256,6 +1253,7 @@ fn handle_block_sync<'state, F: Fn(PipelineEventBox)>(
         genesis_account,
         handle_events,
         block_sync_type,
+        &mut None,
     )
 }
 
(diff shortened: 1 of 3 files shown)


dima74 (approved):

core/src/sumeragi/main_loop.rs
```diff
@@ -1266,32 +1264,48 @@ fn handle_categorized_block_sync<'state, F: Fn(PipelineEventBox)>(
...
+        topology.nth_rotation(block.header().view_change_index as usize);
+        topology
+    };
+
+    // Check if block have enough signatures and proxy tail signature before potentially discard previous voting block
+    if let Err(err) = ValidBlock::is_commit(&block, &topology) {
+        return Err((block, BlockSyncError::BlockNotValid(err)));
+    }
+
+    ValidBlock::validate_keep_voting_block(
```
mversic, line 1293: I would prefer if you would make a new constructor method on `ValidBlock` and put the `is_commit` inside it and make `is_commit` private

mversic (changes requested):

core/src/sumeragi/main_loop.rs
mversic, line 1293:
```suggestion
    ValidBlock::commit_keep_voting_block(
```

how about we do it like this and produce `CommitBlock` directly. Because you have a redundant check afterwards `block.commit`. Or this way:

```suggestion
    CommitedBlock::commit_keep_voting_block(
```

Erigara: I've decided to make this method of `ValidBlock` to keep all related logic nearby

mversic (approved):

dima74 (approved):

PR #4866, opened 2024-07-19
refactor: remove `TriggeringEventFilterBox`

### LINKED ISSUE #4376: Should we remove `TriggeringEventFilterBox`?
IMO we should remove `TriggeringEventFilterBox` and have only `EventFilterBox`. Is there a good reason why we have it?

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/4375#discussion_r1530447568_

mversic:
## Description

I love static typing. It prevents many errors.
But there is a tradeoff between static typing and an API that is easy to use.
Sometimes too much static typing is difficult for end users

### Linked issue

Closes #4376

### Checklist

### DIFF
--- a/core/src/smartcontracts/isi/triggers/mod.rs
+++ b/core/src/smartcontracts/isi/triggers/mod.rs
@@ -49,19 +49,19 @@ pub mod isi {
             let triggers = &mut state_transaction.world.triggers;
             let trigger_id = new_trigger.id().clone();
             let success = match &new_trigger.action.filter {
-                TriggeringEventFilterBox::Data(_) => triggers.add_data_trigger(
+                EventFilterBox::Data(_) => triggers.add_data_trigger(
                     &engine,
                     new_trigger
                         .try_into()
                         .map_err(|e: &str| Error::Conversion(e.to_owned()))?,
                 ),
-                TriggeringEventFilterBox::Pipeline(_) => triggers.add_pipeline_trigger(
+                EventFilterBox::Pipeline(_) => triggers.add_pipeline_trigger(
                     &engine,
                     new_trigger
                         .try_into()
                         .map_err(|e: &str| Error::Conversion(e.to_owned()))?,
                 ),
-                TriggeringEventFilterBox::Time(time_filter) => {
+                EventFilterBox::Time(time_filter) => {
                     if let ExecutionTime::Schedule(schedule) = time_filter.0 {
                         match last_block_estimation {
                             // Genesis block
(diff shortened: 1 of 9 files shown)


core/src/smartcontracts/isi/triggers/mod.rs
```diff
@@ -91,12 +91,15 @@ pub mod isi {
...
                 }
-                TriggeringEventFilterBox::ExecuteTrigger(_) => triggers.add_by_call_trigger(
+                EventFilterBox::ExecuteTrigger(_) => triggers.add_by_call_trigger(
                     &engine,
                     new_trigger
                         .try_into()
                         .map_err(|e: &str| Error::Conversion(e.to_owned()))?,
                 ),
+                EventFilterBox::TriggerCompleted(_) => {
+                    unreachable!("Disallowed during deserialization");
```
Erigara, line 101: Maybe return error explicitly? Because if we would do some kind of unit test and construct `EventFilterBox` by hand it won't be rejected during deserialization.

Erigara (commented):
I still not sure why we should treat trigger completion event separately. 

It was requested like that in #2885, but reasons are not specified.

Erigara:
In general i agree that strict typing in this case create more inconvenience than value.

core/src/smartcontracts/isi/triggers/mod.rs
mversic: test case will then fail, won't it?

Erigara: hmm, indeed
my thought was that error is easier to check

Erigara (approved):

data_model/src/trigger.rs
```diff
@@ -244,6 +249,57 @@ pub mod action {
...
+            pub authority: AccountId,
+            pub filter: EventFilterBox,
+            pub metadata: Metadata,
+        }
+
+        impl ActionCandidate {
+            pub(super) fn validate(self) -> Result<Action, &'static str> {
+                if matches!(self.filter, EventFilterBox::TriggerCompleted(_)) {
+                    return Err("TriggerCompleted cannot be used as filter for triggering actions");
+                }
```
0x009922, line 270: Just as an idea: Iroha can have an internal newtype around `EventFilterBox` which ensures it never holds the `TriggerCompleted` variant.

0x009922 (approved):
LGTM, good tradeoff.

PR #4868, opened 2024-07-19
fix(snapshot): verify all blocks while reading snapshot

Erigara:
## Description

Changes:
- relative path to snapshot folder is resolved in the same way as for `kura`
- last block hash check of snapshot was missed before:
    - not it is checked
    - if check fail last changes to the state are discarded without losing the whole snapshot

### Benefits

- Behavior of configuration is consistent 
- All blocks are checked

### How to test

1. Run iroha
2. Wait until iroha create snapshot
3. Stop iroha
4. Modify last hash in `block_hashes`.`blocks` (any character except for last one)
5. Load iroha
6. Verify that iroha is loaded from snapshot
7. Check that last block was discarded (in logs)
8. Check that iroha can continue to commit new blocks

### Checklist


mversic (dismissed):

core/src/snapshot.rs
```diff
@@ -182,6 +182,20 @@ pub fn try_read_snapshot(
...
+            .and_then(|height| kura.get_block_hash(height))
+            .expect("Kura has height at least as large as state height");
+        let snapshot_block_hash = state_view.block_hashes[snapshot_height - 1];
+        if kura_block_hash != snapshot_block_hash {
+            iroha_logger::warn!(
+                "Snapshot has incorrect latest block hash, discarding changes made by this block"
+            );
+            state.block_and_revert().commit();
+        }
+    }
```
0x009922, line 197: This could be combined with the loop just behind, reducing extra duplication (and confusion):

```rust
for height in 1..=snapshot_height {
    let kura_block_hash = NonZeroUsize::new(height)
        .and_then(|height| kura.get_block_hash(height))
        .expect("Kura has height at least as large as state height");
    let snapshot_block_hash = state_view.block_hashes[height - 1];
    if kura_block_hash != snapshot_block_hash {
        if height == snapshot_height {
            // If last block hash is different it might mean that snapshot was crated for soft-fork block so just drop changes made by this block
            iroha_logger::warn!(
                "Snapshot has incorrect latest block hash, discarding changes made by this block"
            );
            state.block_and_revert().commit();
        } else {
            return Err(TryReadError::MismatchedHash {
                height,
                snapshot_block_hash,
                kura_block_hash,
            });
        }
    }
}
```

0x009922 (commented):
LGTM in general.

I think PR's title is not accurate; maybe, "fix(snapshot): allow last block mismatch; relative store path"?

core/src/snapshot.rs
Erigara: My reasoning was to exclude unnecessary if check out of loop, but since it's one off function it doesn't matter that much.

SamHSmith (approved):

0x009922 (approved):

PR #4870, opened 2024-07-22
fix(kura): correctly handle replace_top_block

Erigara:
## Description

I've noticed multiple issues with kura which this PR is aimed to fix.

1. Quick sequential call of `replace_top_block` and `store_block` cause kura to omit overwriting soft-fork block
    - this happened because kura checked latest_written_block hash against latest_block_hash which is not correct because between kura loop iteration more than one block could be added
2. Call `store_block` before or just after `start` cause kura to incorrectly assume amount of already written blocks

### Benefits

Fewer bugs.

### How to test

To check this issues `create_blocks` test was added.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -69,6 +69,7 @@ indexmap = "2.2.6"
 
 [dev-dependencies]
 iroha_executor_data_model = { workspace = true }
+iroha_wasm_builder = {workspace = true }
 test_samples = { workspace = true }
 
 criterion = { workspace = true }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -69,7 +69,6 @@ indexmap = "2.2.6"
 
 [dev-dependencies]
 iroha_executor_data_model = { workspace = true }
-iroha_wasm_builder = {workspace = true }
 test_samples = { workspace = true }
 
 criterion = { workspace = true }
--- a/core/src/kura.rs
+++ b/core/src/kura.rs
@@ -32,13 +32,14 @@ pub struct Kura {
     /// The block storage
     block_store: Mutex<BlockStore>,
     /// The array of block hashes and a slot for an arc of the block. This is normally recovered from the index file.
-    #[allow(clippy::type_complexity)]
-    block_data: Mutex<Vec<(HashOf<BlockHeader>, Option<Arc<SignedBlock>>)>>,
+    block_data: Mutex<BlockData>,
     /// Path to file for plain text blocks.
     block_plain_text_path: Option<PathBuf>,
     /// Amount of blocks loaded during initialization
     init_block_count: usize,
 }
+
+type BlockData = Vec<(HashOf<BlockHeader>, Option<Arc<SignedBlock>>)>;
 
 impl Kura {
     /// Initialize Kura and start a thread that receives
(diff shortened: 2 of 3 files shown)


core/src/kura.rs
```diff
@@ -962,4 +990,209 @@ mod tests {
...
+                [account, genesis_account],
+                [],
+                UniqueVec::new(),
+            ),
+            Kura::blank_kura_for_testing(),
+            live_query_store,
+        );
+
+        let executor = {
+            let wasm_blob = iroha_wasm_builder::Builder::new("../default_executor")
```
mversic, line 1098: this seems to heavy for a unit test

Erigara: i haven't find a way around it

Erigara: Because in order for genesis block to be valid it have to have valid `Upgrade` instruction.

core/Cargo.toml
```diff
@@ -69,6 +69,7 @@ indexmap = "2.2.6"
 
 [dev-dependencies]
 iroha_executor_data_model = { workspace = true }
+iroha_wasm_builder = {workspace = true }
```
DCNick3, line 72: nit (formatting)

```suggestion
iroha_wasm_builder = { workspace = true }
```

core/src/kura.rs
DCNick3: Would it be possible to write a mock executor with wasm text representation in the same way it's done in core unit tests? https://github.com/hyperledger/iroha/blob/58c943e44eea09e6f0c80c487547d0ff11f9eccb/core/src/smartcontracts/wasm.rs#L1776

Or maybe use the same executor blob that is used in integration tests' genesis (`defaults/executor.wasm`)?

core/src/kura.rs
```diff
@@ -29,17 +29,19 @@ const SIZE_OF_BLOCK_HASH: u64 = Hash::LENGTH as u64;
 /// The interface of Kura subsystem
 #[derive(Debug)]
 pub struct Kura {
-    /// The mode of initialisation of [`Kura`].
-    mode: InitMode,
     /// The block storage
     block_store: Mutex<BlockStore>,
     /// The array of block hashes and a slot for an arc of the block. This is normally recovered from the index file.
     #[allow(clippy::type_complexity)]
```
DCNick3, line 35: I think this lint should be no longer triggered with the type alias used

core/src/kura.rs
```diff
@@ -193,9 +192,12 @@ impl Kura {
         kura: &Kura,
         mut shutdown_receiver: tokio::sync::oneshot::Receiver<()>,
     ) {
-        let (mut written_block_count, mut latest_block_hash) = {
-            let block_data_guard = kura.block_data.lock();
-            (block_data_guard.len(), block_data_guard.last().map(|d| d.0))
+        let mut written_block_count = kura.init_block_count;
+        let mut latest_written_block_hash = {
+            let block_data = kura.block_data.lock();
+            written_block_count
```
DCNick3, line 198: Do I understand it right that this is basically the same behavior as before? Seeing how `init_block_count` gets set to `block_data.len()` in the init.. Or can something actually modify the `block_data` vec in the meantime?

UPD: I see now that you mention that `store_block` being called before or quickly after spawning the `receive_blocks_loop` thread, makes sense.

DCNick3 (dismissed):

core/Cargo.toml
Erigara: removed since `iroha_wasm_builder` is no longer required

core/src/kura.rs
```diff
@@ -962,4 +988,203 @@ mod tests {
...
+        let state = State::new(
+            World::with([domain, genesis_domain], [account, genesis_account], []),
+            Kura::blank_kura_for_testing(),
+            live_query_store,
+        );
+
+        let executor = {
+            let mut executor_file = std::fs::File::open("../defaults/executor.wasm").unwrap();
+            let mut executor_blob = Vec::new();
+            executor_file.read_to_end(&mut executor_blob).unwrap();
```
DCNick3, line 1093: Why not use `std::fs::read` here?

DCNick3 (dismissed):

Erigara:
@mversic @DCNick3 updated, reapprove pls

mversic (approved):

PR #4880, opened 2024-07-23
refactor: remove committed_topology from block

Erigara:
## Description

- committed topology is removed from block
- committed topology for latest and previous blocks are stored in the state
    - it's required for blocks sync
    - and for initializing state from the snapshot + block store
- initial topology in the genesis block is passed as transaction with `Register<Peer>` instructions
- trusted peers are removed from `World` constructor

### Linked issue

Closes #4879 

### Benefits

Less space used in the block store, easier to derive block hash.

### How to test

Run, try to submit new blocks, try restart peers check that iroha operates as usual.

PR #4890, opened 2024-07-24
refactor: derive block hash from header

### LINKED ISSUE #4641: How block hash should be derived?
Look like we sign over whole block not just header.
I see few problems with that:
- redundant becasue in the block header merkle root of transactions is alredy included
- might harm ability to verify block for light client, since in order to verify inclusion of transation we need whole transacaction set

I will create a ticket for discussion.

_Originally posted by @Erigara in https://github.com/hyperledger/iroha/pull/4518#discussion_r1611265234_

Erigara:
## Description

Derive block hash from the header only.

### Linked issue

Closes #4641 

### Benefits

- block hash is the same for the whole lifetime of the block
- block hash is less expensive as only header is used
- light client is able to verify block chain by only headers and corresponding signatures (without obligation to load the whole block)


core/src/sumeragi/message.rs
```diff
@@ -118,7 +118,7 @@ mod candidate {
     #[derive(Decode)]
     struct BlockCommittedCandidate {
         /// Hash of the block being signed.
-        pub hash: HashOf<BlockPayload>,
+        pub hash: HashOf<SignedBlock>,
```
Erigara, line 121: Alternative would be to use `HashOf<BlockHeader>`, but it's not versioned.

mversic (dismissed):

mversic (approved):

s8sato (approved):
Along with #4880 that made the block header reflect the whole block payload, this PR would streamline the block hash

PR #4893, opened 2024-07-25
fix: Add limits to LiveQueryStore to prevent high memory usage

### LINKED ISSUE #4830: Investigate how Iroha operates under load under Ubuntu/Debian

dima74:
## Description

Added two new config parameters which limits number of live queries (global and per-user). Default value for both is 128. This is needed to prevent high memory usage in cases when client sends a lot of queries with large output, but do not fully consume those queries (that is obtain only first batch and doesn't send requests for the next batches). Normally clients should not do this, but malicious third-party can send such requests to DDOS iroha peer. In case limits are exceeded, iroha peer will return error "429 too many requests" for the query request.

* Names of parameters:
```toml
[torii]
query_store_capacity = 128
query_store_capacity_per_user = 128
```

Also:
* Changed default value of query idle time from 30s to 10s (for the same reason to decrease potential peak memory usage)
* Changed default value of fetch size from 10 to 100 (I think it is strange that when query output has e.g. 1000 items, client currently has to send 100 requests to get full query result)

Some notes about implementation:
* Currently when adding new query, the message is send to `mpsc::channel`, and `LiveQueryStore` thread does the addition. I changed it to perform addition directly without `mpsc::channel` since we now need to check capacity and return error to client if it exceeds.

### Linked issue

Closes #4830

### Benefits

Add limit to number of live queries to prevent high memory usage in cases when client don't fully consume query output.

### Checklist


config/src/parameters/defaults.rs
```diff
@@ -45,12 +45,15 @@ pub mod snapshot {
...
+    use std::{num::NonZeroUsize, time::Duration};
 
     use iroha_config_base::util::Bytes;
+    use nonzero_ext::nonzero;
 
     pub const MAX_CONTENT_LEN: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
-    pub const QUERY_IDLE_TIME: Duration = Duration::from_secs(30);
+    pub const QUERY_IDLE_TIME: Duration = Duration::from_secs(10);
+    pub const QUERY_STORE_CAPACITY: NonZeroUsize = nonzero!(256usize);
+    pub const QUERY_STORE_CAPACITY_PER_USER: NonZeroUsize = nonzero!(256usize);
```
mversic, line 56: you said it's 128 in pr description?

mversic (approved):

config/src/parameters/defaults.rs
```diff
@@ -45,12 +45,15 @@ pub mod snapshot {
...
+    use std::{num::NonZeroUsize, time::Duration};
 
     use iroha_config_base::util::Bytes;
+    use nonzero_ext::nonzero;
 
     pub const MAX_CONTENT_LEN: Bytes<u64> = Bytes(2_u64.pow(20) * 16);
-    pub const QUERY_IDLE_TIME: Duration = Duration::from_secs(30);
+    pub const QUERY_IDLE_TIME: Duration = Duration::from_secs(10);
+    pub const QUERY_STORE_CAPACITY: NonZeroUsize = nonzero!(128usize);
+    pub const QUERY_STORE_CAPACITY_PER_USER: NonZeroUsize = nonzero!(128usize);
```
mversic, line 56:
```suggestion
    pub const QUERY_STORE_CAPACITY: NonZeroUsize = nonzero!(128usize);
    pub const QUERY_STORE_CAPACITY_PER_USER: NonZeroUsize = nonzero!(128usize);
```

isn't this a bit small for global capacity?

mversic:
@dima74 `with_coverage` is failing if you haven't noticed

mversic (approved):

config/src/parameters/defaults.rs
dima74: Based on my testing, with limit 128 and query with output size 2500, iroha will consume around 2GB. Memory consumption should increase lineary with the limit, e.g. with limit 256 I expect memory consumption around 4GB.  If we make per-user limit 128, and increase global limit, I am wondering whether malicious third-party could utulize multiple accounts to DDOS iroha

nxsaken (approved):

PR #4909, opened 2024-07-27
fix: prevent redundant blocksync block messages

SamHSmith:
## Description

There was a problem during TPS testing where block sync could not occur quickly because it would re-request the
same blocks from each peer. Here is a solution that should be robust against that scenario but doesn't introduce
any new issues.

### Linked issue

Closes #{issue_number} 

### Checklist

### DIFF
--- a/core/src/block_sync.rs
+++ b/core/src/block_sync.rs
@@ -1,5 +1,11 @@
 //! This module contains structures and messages for synchronization of blocks between peers.
-use std::{fmt::Debug, num::NonZeroU32, sync::Arc, time::Duration};
+use std::{
+    collections::BTreeSet,
+    fmt::Debug,
+    num::{NonZeroU32, NonZeroU64},
+    sync::Arc,
+    time::Duration,
+};
 
 use iroha_config::parameters::actual::BlockSync as Config;
 use iroha_crypto::HashOf;
@@ -44,6 +50,8 @@ pub struct BlockSynchronizer {
     gossip_size: NonZeroU32,
     network: IrohaNetwork,
     state: Arc<State>,
+    seen_blocks: BTreeSet<(NonZeroU64, HashOf<SignedBlock>)>,
+    latest_height: usize,
 }
 
 impl BlockSynchronizer {
@@ -74,6 +82,17 @@ impl BlockSynchronizer {
 
     /// Sends request for latest blocks to a random peer
     async fn request_block(&mut self) {
+        let now_height = self.state.view().height();
+
+        // This guards against a softfork and adds general redundancy.
+        if now_height == self.latest_height {
+            self.seen_blocks.clear();
+        }
+        self.latest_height = now_height;
+
+        self.seen_blocks
+            .retain(|(height, _hash)| height.get() >= now_height as u64);
+
         if let Some(random_peer) = self.network.online_peers(Self::random_peer) {
             self.request_latest_blocks_from_peer(random_peer.id().clone())
                 .await;
@@ -98,6 +117,10 @@ impl BlockSynchronizer {
         message::Message::GetBlocksAfter(message::GetBlocksAfter::new(
             latest_hash,
             prev_hash,
+            self.seen_blocks
+                .iter()
+                .map(|(_height, hash)| hash.clone())
+                .collect(),
             self.peer_id.clone(),
         ))
         .send_to(&self.network, peer_id)
@@ -121,6 +144,8 @@ impl BlockSynchronizer {
             gossip_size: config.gossip_size,
             network,
             state,
+            seen_blocks: BTreeSet::new(),
+            latest_height: 0,
         }
     }
 }
@@ -138,6 +163,8 @@ pub mod message {
         pub latest_hash: Option<HashOf<SignedBlock>>,
         /// Hash of second to latest block
         pub prev_hash: Option<HashOf<SignedBlock>>,
+        /// The block hashes already seen
+        pub seen_blocks: BTreeSet<HashOf<SignedBlock>>,
         /// Peer id
         pub peer_id: PeerId,
     }
@@ -147,11 +174,13 @@ pub mod message {
         pub const fn new(
             latest_hash: Option<HashOf<SignedBlock>>,
             prev_hash: Option<HashOf<SignedBlock>>,
+            seen_blocks: BTreeSet<HashOf<SignedBlock>>,
             peer_id: PeerId,
         ) -> Self {
             Self {
                 latest_hash,
                 prev_hash,
+                seen_blocks,
                 peer_id,
             }
         }
@@ -190,6 +219,7 @@ pub mod message {
                 Message::GetBlocksAfter(GetBlocksAfter {
                     latest_hash,
                     prev_hash,
+                    seen_blocks,
                     peer_id,
                 }) => {
                     let local_latest_block_hash = block_sync.state.view().latest_block_hash();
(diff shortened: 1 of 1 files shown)


core/src/block_sync.rs
```diff
@@ -138,6 +163,8 @@ pub mod message {
         pub latest_hash: Option<HashOf<SignedBlock>>,
         /// Hash of second to latest block
         pub prev_hash: Option<HashOf<SignedBlock>>,
+        /// The block hashes already seen
+        pub seen_blocks: BTreeSet<HashOf<SignedBlock>>,
```
Erigara, line 167: I'm  a bit concerned that now `GetBlocksAfter` can grow potentially large but can't think of other good solution which would achieve the same thing.

One option could be to used randomized data structure like `LRUCache` to have bounded size for `GetBlocksAfter` and possibility of duplicated message send.

Erigara (commented):
I think this PR is step in the right direction, minor fixes and i would approve.

core/src/block_sync.rs
```diff
@@ -44,6 +44,8 @@ pub struct BlockSynchronizer {
     gossip_size: NonZeroU32,
     network: IrohaNetwork,
     state: Arc<State>,
+    seen_blocks: BTreeSet<(u64, HashOf<SignedBlock>)>,
+    last_sumeragi_height: u64,
```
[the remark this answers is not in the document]
mversic: it's not sumeragi height, it's taken from the state

mversic, line 48:
```suggestion
    seen_blocks: BTreeSet<(NonZeroUsize, HashOf<SignedBlock>)>,
    last_height: usize,
```

core/src/block_sync.rs
```diff
@@ -74,6 +76,23 @@ impl BlockSynchronizer {
...
     /// Sends request for latest blocks to a random peer
     async fn request_block(&mut self) {
+        let now_height = {
+            let state_view = self.state.view();
+            state_view.height()
+        } as u64;
+        if now_height == self.last_sumeragi_height {
+            self.seen_blocks.clear();
+        }
+        self.last_sumeragi_height = now_height;
```
mversic, line 86:
```suggestion
        let now_height = self.state.view().height();
```

this looks like a micro optimization to me. Since you're later clearing `seen_blocks` anyways

core/src/block_sync.rs
```diff
@@ -225,6 +255,7 @@ pub mod message {
                                 .and_then(|height| block_sync.kura.get_block_by_height(height))
                         })
                         .skip_while(|block| Some(block.hash()) == *latest_hash)
+                        .filter(|block| !seen_blocks.contains(&block.hash()))
```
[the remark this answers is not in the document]
mversic: yes, but only if height is also compared. Hashes are not sorted properly

core/src/block_sync.rs
```diff
@@ -244,6 +275,9 @@ pub mod message {
                     use crate::sumeragi::message::BlockSyncUpdate;
 
                     for block in blocks.clone() {
+                        block_sync
+                            .seen_blocks
+                            .insert((block.header().height().get(), block.hash()));
```
mversic, line 280: what if we receive a block with the height that is already present in the `seen_blocks` but with a different hash?

mversic (changes requested):

mversic:
this should be backported to rc22

core/src/block_sync.rs
mversic: malicious peer can send `GetBlocksAfter` with a huge list of invalid `seen_blocks`

core/src/block_sync.rs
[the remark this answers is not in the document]
SamHSmith: That guarantee is not enforced. So I don't want to rely on it.

core/src/block_sync.rs
SamHSmith: A malicious peer can send large amounts of invalid data in most of our network messages. So solving this kind of issue generally should be a separate ticket.

core/src/block_sync.rs
SamHSmith: No it's actually essential. If the block height is not changing we need to stop filtering the incoming blocks in case we need to receive the same block again. I added a comment.

core/src/block_sync.rs
mversic: then open a ticket for this

mversic:
you have a failing test in torii api

core/src/block_sync.rs
SamHSmith: Here is the ticket. https://github.com/hyperledger/iroha/issues/4944

SamHSmith:
> you have a failing test in torii api

It has now been fixed.

Erigara (dismissed):

core/src/block_sync.rs
mversic: I see. This handles soft fork. Another approach would be to always request the latest block again, wouldn't it?

mversic (dismissed):

core/src/block_sync.rs
```diff
@@ -44,6 +50,8 @@ pub struct BlockSynchronizer {
     gossip_size: NonZeroU32,
     network: IrohaNetwork,
     state: Arc<State>,
+    seen_blocks: BTreeSet<(NonZeroU64, HashOf<SignedBlock>)>,
+    latest_height: NonZeroU64,
```
mversic, line 54:
```suggestion
    latest_height: NonZeroUsize,
```

mversic, line 54:
```suggestion
    seen_blocks: BTreeSet<(NonZeroUsize, HashOf<SignedBlock>)>,
    latest_height: NonZeroUsize,
```

core/src/block_sync.rs
```diff
@@ -74,6 +82,23 @@ impl BlockSynchronizer {
...
+            if now_height == self.latest_height {
+                self.seen_blocks.clear();
+            }
+            self.latest_height = now_height;
+
+            self.seen_blocks
+                .retain(|(height, _hash)| *height >= now_height);
+        } else {
+            self.seen_blocks.clear();
+        }
```
mversic, line 100: I don't think you need to worry about the genesis

core/src/block_sync.rs
```diff
@@ -74,6 +82,23 @@ impl BlockSynchronizer {
...
     /// Sends request for latest blocks to a random peer
     async fn request_block(&mut self) {
+        let now_height = self.state.view().height() as u64;
+
+        if now_height > 0 {
+            let now_height = now_height.try_into().expect("Should not be zero");
+
+            // This guards against a softfork and adds general redundancy.
+            if now_height == self.latest_height {
+                self.seen_blocks.clear();
```
mversic, line 92: don't you only need to filter out the last block?

core/src/block_sync.rs
```diff
@@ -121,6 +150,8 @@ impl BlockSynchronizer {
             gossip_size: config.gossip_size,
             network,
             state,
+            seen_blocks: BTreeSet::new(),
+            latest_height: NonZeroU64::new(1).expect("1 is not 0"),
```
mversic, line 154:
```suggestion
            latest_height: nonzeroext!(1),
```

core/src/block_sync.rs
mversic: `latest_height` can be 0. That was a mistake

core/src/block_sync.rs
SamHSmith: Okay true.

core/src/block_sync.rs
SamHSmith: Yes. But they should be rare and so we don't want to send a wasted block on each block sync.

core/src/block_sync.rs
SamHSmith: The seen_blocks will be NonZeroU64 to match the block header data structure.

Erigara (approved):

mversic (approved):

PR #4922, opened 2024-07-31
fix(sumeragi): remove expired transaction from cache

Erigara:
## Description

- fix a bug in retain that actually non expired transactions were removed
- qol improvment to print transactions by hash in the logs
    - i find it really hard to debug iroha on TRACE level in presence of heavy transactions because all logs are polluted by tx printing 

### Benefits

- fewer bugs
- more clear logs

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -1024,7 +1024,7 @@ pub(crate) fn run(
             .retain(|tx| {
                 let expired = sumeragi.queue.is_expired(tx);
                 if expired {
-                    debug!(?tx, "Transaction expired")
+                    debug!(tx=%tx.as_ref().hash(), "Transaction expired")
                 }
                 !expired
             });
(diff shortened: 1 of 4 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -1024,9 +1024,9 @@ pub(crate) fn run(
             .retain(|tx| {
                 let expired = sumeragi.queue.is_expired(tx);
                 if expired {
-                    debug!(?tx, "Transaction expired")
+                    debug!(tx=%tx.as_ref().hash(), "Transaction expired")
                 }
-                expired
+                !expired
```
Erigara, line 1029: The error was here...

nxsaken (approved):

core/src/tx.rs
```diff
@@ -229,7 +229,7 @@ impl TransactionExecutor {
             ));
         }
 
-        debug!("Validating transaction: {:?}", tx);
+        debug!(tx=%tx.as_ref().hash(), "Validating transaction");
```
nxsaken, line 232: Does `rustfmt` complain about whitespace in these macros?

Erigara: Not really, it's ok with both options

DCNick3: It's custom syntax, so it doesn't touch it

DCNick3 (approved):

PR #4923, opened 2024-07-31
fix(p2p): prevent deadlock on simultaneous sending large message

Erigara:
## Description

When 2 peers try to send each other huge messages they end up deadlocked.
I suspect that this happen when message is large enough that it's not possible to fully put it in the OS buffer.

Here is minimal [example](https://github.com/Erigara/tcp_deadlock) i was able to came up with which suffers from the same issue.

Solution was to lift sending data up to the `select!` statement so peer can either read or write data simultaneously. 

### Benefits

No deadlock.

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -196,8 +196,8 @@ mod run {
                             ping_period=?ping_interval.period(),
                             "The connection has been idle, pinging to check if it's alive"
                         );
-                        if let Err(error) = message_sender.send_message(Message::<T>::Ping).await {
-                            iroha_logger::error!(%error, "Failed to send ping to peer.");
+                        if let Err(error) = message_sender.prepare_message(Message::<T>::Ping) {
+                            iroha_logger::error!(%error, "Failed to encrypt message.");
                             break;
                         }
                     }
@@ -218,8 +218,8 @@ mod run {
                         if post_receiver_len > 100 {
                             iroha_logger::warn!(size=post_receiver_len, "Peer post messages are pilling up");
                         }
-                        if let Err(error) = message_sender.send_message(Message::Data(msg)).await {
-                            iroha_logger::error!(%error, "Failed to send message to peer.");
+                        if let Err(error) = message_sender.prepare_message(Message::Data(msg)) {
+                            iroha_logger::error!(%error, "Failed to encrypt message.");
                             break;
                         }
                     }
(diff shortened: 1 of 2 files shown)


p2p/src/peer.rs
```diff
@@ -379,38 +388,40 @@ mod run {
...
-                "Buffer must be empty at this point"
-            );
+            self.queue.put_u32(size as u32);
+            self.queue.put_slice(encrypted.as_slice());
+            Ok(())
+        }
+
+        /// Send bytes of byte-encoded messages piled up in the message queue so far.
+        /// On the other side peer will collect bytes and recreate original messages from them.
+        ///
```
DCNick3, line 416:
```suggestion
        ///
        /// Sends only as much data as the underlying writer will accept in one `.write` call, so must be called in a loop to ensure everything will get sent.
```

p2p/src/peer.rs
```diff
@@ -379,38 +388,40 @@ mod run {
...
         }
 
-        /// Send byte-encoded message to the peer
+        /// Prepare message for the delivery and put it into the queue to be sent later
         ///
         /// # Errors
         /// - If encryption fail.
-        /// - If write to `stream` fail.
-        async fn send_message<T: Pload>(&mut self, msg: T) -> Result<(), Error> {
+        fn publish_message<T: Pload>(&mut self, msg: T) -> Result<(), Error> {
```
DCNick3, line 400: Maybe this name?

```suggestion
        fn prepare_message<T: Pload>(&mut self, msg: T) -> Result<(), Error> {
```

p2p/src/peer.rs
```diff
@@ -379,38 +388,40 @@ mod run {
...
-            let encoded_size = self.buffer.remaining();
-            let encrypted = self.cryptographer.encrypt(&self.buffer[..encoded_size])?;
-            self.buffer.advance(encoded_size);
-            assert!(
-                !self.buffer.has_remaining(),
-                "Buffer must be empty at this point"
-            );
-            let encrypted_size = encrypted.len();
+            msg.encode_to(&mut self.buffer);
+            let encrypted = self.cryptographer.encrypt(&self.buffer)?;
```
DCNick3, line 404: (not for this PR): Maybe we can get rid of this allocation by encrypting in-place?

Erigara: I was thinking about exactly the same thing, but currently Encryptor trait provide only allocating methods...
Also tried to use only single buffer, but resulting code was tricky and buggy so i‘ve decided to leave it for separate PR.

DCNick3 (approved):

dima74 (approved):

PR #4928, opened 2024-08-02
feat: bounded discrete time

### LINKED ISSUE #4962: Bounded time
Any timestamp on the chain should be bounded:
1. block creation time - lower bound is previous block time (hard limit), upper bound is current peer time (soft limit)
2. transaction creation time - upper bound is current peer time (soft limit) or creation time of it's block (hard limit)

hard limit must not be crossed, whereas soft limit can be exceeded by a negligibly small amount

mversic:
## Description

* put a lower bound on block time (previous block time)
* put upper bound on the block time (less than peer current time)

### Linked issue

Closes #4962

### Benefits

now have a bounded concept of time that is synchronized across peers.
the time is discrete with increment intervals being the time between 2 blocks

### Checklist


data_model/src/block.rs
```diff
@@ -325,10 +313,81 @@ mod candidate {
...
+    }
+
+    impl BlockHeaderCandidate {
+        const fn creation_time(&self) -> Duration {
+            Duration::from_millis(self.creation_time_ms)
+        }
+
+        fn validate(self) -> Result<BlockHeader, &'static str> {
+            #[cfg(feature = "std")]
+            if std::time::Instant::now().elapsed() < self.creation_time() {
```
Erigara, line 340: I'm a bit concerned that block validity now depends on when block was received.

Erigara (commented):
Can you provide more context for this change.

Also i can't see synchronization part of this PR, am i missing something?

data_model/src/block.rs
```diff
@@ -325,10 +313,81 @@ mod candidate {
...
+                consensus_estimation_ms: self.consensus_estimation_ms,
+            })
+        }
+    }
+
+    impl BlockPayloadCandidate {
+        fn validate(self) -> Result<BlockPayload, &'static str> {
+            self.validate_header()?;
+
+            // TODO: Validate Event recommendations somehow?
```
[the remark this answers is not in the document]
mversic: removed

mversic:
> Can you provide more context for this change.

Sequence of blocks is now guaranteed to have monotonically increasing time. The leader can't just completely fabricate the voted block's creation time because the creation time has to be within some limits (after previous block creation time, but before current validator/observer/proxy node time). If the leader tells a lie within the imposed limits, given enough rotations we should again arrive to the correct value of the time. On average, I'd say that time should remain quite close to the current time (a little bit behind, never in the future).

> Also i can't see synchronization part of this PR, am i missing something?

I removed that word. I meant that it's synchronized in the sense that everybody can agree when the block was created (within an interval). Peers can use blocks creation time to reason about current time. Whatever is the last creation time is the current time across all peers and it increases discretely in intervals.

### Example
1. A block was created at 11:25AM and committed
2. 11:25AM becomes the current time (the network is certain that the actual time is no less than that)
3. Block was committed at 11:35
4. 11:35AM becomes the current time (the network is certain that the actual time is no less than that)

time is discrete in the blockchain because 11:30 is not a valid time, only 11:25, 11:35, etc are valid timestamps

### Implementing time based logic in a smart contract
A smart contract can be given latest block creation time as the current time (a sort of synchronization point) and have logic based on that (understanding the limitations). In the previous example, if the smart contract was written so that it performs a certain action if time is past 11:30AM, it will do it at 11:35AM. But the time of the execution is guaranteed to never be less than 11:30AM

data_model/src/block.rs
Erigara: i don't think it compromise security, but might harm liveliness if peers clocks drift from each other.

Erigara:
To avoid block rejection can't leader select block time to be as close to the previous block time as possible?
I think it would be considered valid because it's in the range (previous_block_time, min(voting_peer_current_time_1, voting_peer_current_time_n)].

mversic:
> To avoid block rejection can't leader select block time to be as close to the previous block time as possible? I think it would be considered valid because it's in the range (previous_block_time, min(voting_peer_current_time_1, voting_peer_current_time_n)].

yes, but then the next leader will correct it, or the one after that. The error doesn't compound or get propagated indefinitely. Network time can drift from the actual current time but not too far.

data_model/src/block.rs
mversic: This is upper bound. What do you think about having the lower bound? I can only define lower bound if I define both lower and upper. Because if lower bound is defined and block time is set far into the future no new blocks can be produced

core/src/block.rs
s8sato, line 1: Do we accept that consensus might not be reached due to system time gap between peers?

data_model/src/block.rs
```diff
@@ -325,10 +313,88 @@ mod candidate {
...
+            {
+                use std::time::SystemTime;
+
+                let now = SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .expect("INTERNAL BUG: Failed to get the current system time");
+
+                if now <= self.creation_time() {
+                    return Err("Block creation time is in the future");
+                }
```
s8sato, line 350: What is the reason for validating the upper and lower bounds in separate places?

core/src/block.rs
s8sato: In other words, should having a unsynced clock be considered as being faulty? To me it sounds reasonable

mversic: > having a unsynced clock be considered as being faulty? To me

if the clock drifts too far away, than I would consider it faulty

data_model/src/block.rs
mversic: I prefer validating as soon as possible (i.e. on deserialization). But I can't validate lower bound on deserialization since I don't have access to the previous block

mversic: However, validating on deserialization is redundant for clients. They should trust the integrity of the blockchain. In that case, I agree that all block check should be done only at the time of consensus. Never again later. I can make a separate ticket for that?

data_model/src/block.rs
```diff
@@ -325,10 +313,88 @@ mod candidate {
...
+        fn validate(self) -> Result<BlockHeader, &'static str> {
+            #[cfg(feature = "std")]
+            {
+                use std::time::SystemTime;
+
+                let now = SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .expect("INTERNAL BUG: Failed to get the current system time");
+
+                if now <= self.creation_time() {
```
mversic, line 348: I'm not sure if this is too tight? For transactions we use `future_threshold`. @Erigara @s8sato

data_model/src/block.rs
mversic: #4950

core/src/block.rs
s8sato: More precisely, it would be considered faulty if the clock drifts by a non-negligibly large amount relative to the average period between blocks. I think Roman's concern about liveliness is about this point i.e. depending on the frequency of block updates, the system time lag may become non-negligible

data_model/src/block.rs
s8sato: What if we introduce a buffer to allow for system time differences, which is negligible relative to the average block period?

s8sato: Implying the more frequently a block is committed, the more buffer is reduced and the more likely a peer is considered faulty

mversic: Do you mean like `future_threshold` in the queue

s8sato: Yes, it would be an analogy of `future_threshold`. Do you think there should be a common variable for the block and transaction timestamps to allow for system time differences?

core/src/block.rs
s8sato: For the upper bound (referencing only the voting block) only the system time differences matter, while for the lower bound (referencing also the last committed block) the average block period also matters

s8sato: As with the current proposed change 31872905167c5205f92a670af6a5b2f2ae1e1626, a buffer should be introduced only for the upper bound. If introduced for the lower bound, timestamps cannot be guaranteed to monotonically increase.

My slight concern is, if the buffer for the upper bound is too large compared to the block period, it might damage the soundness of the election:
As an extreme example, suppose that a leader with an outstandingly advanced system time created a block in the previous round, and most validators barely approved the block thanks to the large buffer. In this case, other leaders in the current round might not be able to create timestamps that meet the lower bound, which means view changes will be repeated until the same amount of time as the buffer has elapsed or the same leader as in the previous round is elected

data_model/src/block.rs
```diff
@@ -334,10 +330,53 @@ mod candidate {
...
+                .ok_or("Block is empty")?;
+
+            if expected_txs_hash != actual_txs_hash {
+                return Err("Transactions' hash incorrect");
+            }
+
+            self.transactions.iter().try_for_each(|tx| {
+                if tx.value.creation_time() >= self.header.creation_time() {
+                    return Err("Transaction creation time is ahead of block creation time");
+                }
```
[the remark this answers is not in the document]
mversic: you are correct, if this is an invariant it must be upheld everywhere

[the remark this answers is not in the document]
mversic: and yes, I think we want to be strict about the lower bound, but fuzzy about upper bound

core/benches/kura.rs
```diff
@@ -30,23 +39,13 @@ async fn measure_block_size_for_n_executors(n_executors: u32) {
...
         .with_instructions([transfer])
         .sign(alice_keypair.private_key());
-    let txn_limits = TransactionParameters {
-        max_instructions: nonzero!(4096_u64),
-        smart_contract_size: nonzero!(1_u64),
+    let (max_clock_drift, tx_limits) = {
+        let state_view = state.world.view();
+        let params = state_view.parameters();
+        (params.sumeragi().max_clock_drift(), params.transaction)
     };
```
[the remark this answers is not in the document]
mversic: I don't think it's worth it, but we can make it a ticket

data_model/src/block.rs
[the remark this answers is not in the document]
mversic: I updated the code for block creation in `core/block.rs`

core/benches/kura.rs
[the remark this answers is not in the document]
s8sato: I won't insist on it

core/src/queue.rs
```diff
@@ -36,8 +36,8 @@ pub struct Queue {
     events_sender: EventsSender,
     /// The queue for transactions
     tx_hashes: ArrayQueue<HashOf<SignedTransaction>>,
-    /// [`AcceptedTransaction`]s addressed by `Hash`
-    accepted_txs: DashMap<HashOf<SignedTransaction>, AcceptedTransaction>,
+    /// [`ValidTransaction`]s addressed by `Hash`
+    txs: DashMap<HashOf<SignedTransaction>, AcceptedTransaction>,
```
dima74, line 40:
```suggestion
    /// [`AcceptedTransaction`]s addressed by `Hash`
    txs: DashMap<HashOf<SignedTransaction>, AcceptedTransaction>,
```

torii/src/routing.rs
```diff
@@ -29,14 +29,19 @@ pub async fn handle_transaction(
...
+        let params = state_view.parameters();
+        (params.sumeragi.max_clock_drift(), params.transaction)
+    };
+
+    let accepted_tx = AcceptedTransaction::accept(tx, &chain_id, max_clock_drift, tx_limits)
         .map_err(Error::AcceptTransaction)?;
+
     queue
-        .push(transaction, &state_view)
+        .push(accepted_tx, state.view())
```
dima74, line 44: Isn't it possible to reuse first `state_view` here?

s8sato (dismissed):

torii/src/routing.rs
mversic: yes, but then lock is held longer then necessary

s8sato (approved):

dima74 (approved):

PR #4929, opened 2024-08-02
refactor(view_change): only send current and previous view change proof

### LINKED ISSUE #4926: refactor(view_change): send only latest view change proof
I've noticed that right now every proof within current block height is send on each view change suggestion. 
Which create redundant data transfer and it became heavier and heavier with each new view change.

Erigara:
## Description

Instead of sending every view change proof only view change proof for current round and previous verified proof are sent.

### Linked issue

Closes #4926 

### Benefits

This help peer which is lagging behind to directly jump to most recent view change and participate in consensus.

### DIFF
--- a/core/src/sumeragi/main_loop.rs
+++ b/core/src/sumeragi/main_loop.rs
@@ -138,12 +138,12 @@ impl Sumeragi {
                 })
             {
                 should_sleep = false;
-                if let Err(error) = view_change_proof_chain.merge(
-                    msg.view_change_proofs,
+                if let Err(error) = view_change_proof_chain.insert_proof(
+                    msg.view_change_proof,
                     &self.topology,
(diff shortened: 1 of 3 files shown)


SamHSmith (approved):
Looks great.

core/src/sumeragi/main_loop.rs
```diff
@@ -1135,8 +1135,28 @@ pub(crate) fn run(
                     .unwrap_or_else(|err| error!("{err}"));
```
mversic, line 1135:
```suggestion
                    .expect("INTERNAL BUG: Invalid view change proof");
```

this can't fail, can it?

mversic (changes requested):

mversic (approved):

PR #4931, opened 2024-08-02
ci: Build executor.wasm in CI

### LINKED ISSUE #4549: Build `config/swarm/executor.wasm` in CI

BAStos525:
## Description
1. Compile on CI runners and get whatever it's required the `executor.wasm` file.
2. Include the necessary iroha files into image on iroha image building stage.
3. Modify `Dockerfile` and `docker-compose` files.
4. Decorate `Dockerfile.glibc` up to `Dockerfile`.
5. Remove `executor.wasm` and `Dockerfile.build.glibc` files from the repository.
6. Remove `archive_binaries_and_schema` job.
7. Save build artifacts inside `registry` job.

### Linked issue
Closes #4549 #4942

### Benefits
1. Developers shouldn't compile the `executor.wasm` file locally and push it to the repository more.
2. Keeping iroha configs already in the image will help to DevOps do not copy genesis and executor from git iroha repo and manage egress policies.

### Checklist


BAStos525:
@0x009922 please take a look on docker-compose files generation tool.

nxsaken:
I'm on it

nxsaken (approved):

tools/swarm/src/schema.rs
```diff
@@ -204,17 +204,66 @@ impl<'a> GenesisEnv<'a> {
...
+const CONTAINER_SIGNED_GENESIS: ContainerFile =
+    ContainerFile(CONTAINER_TMP_DIR, GENESIS_SIGNED_SCALE);
+
+#[derive(Copy, Clone, Debug)]
+struct ReadOnly;
 
 /// Mapping between `host:container` paths.
 #[derive(Copy, Clone, Debug)]
-struct PathMapping<'a>(HostPath<'a>, ContainerPath<'a>);
+struct PathMapping<'a>(HostFile<'a>, ContainerFile<'a>, ReadOnly);
```
Erigara, line 263: Why do you need `ReadOnly` struct? should `bool` be used instead to indicate whatever file is read only or not?

Erigara (commented):
How am i supposed to run iroha locally without docker or run tests which require executor?
Compile it locally?

tools/swarm/src/schema.rs
nxsaken: We don't need it, but it makes it explicit that `PathMapping` is always read-only for now.

.gitignore
s8sato, line 1: A similar question as https://github.com/hyperledger/iroha/pull/4931#pullrequestreview-2218446270, are developers supposed to checkout a certain commit, build executor, and run tests?

.github/workflows/iroha2-dev-nightly.yml
```diff
@@ -2,13 +2,39 @@ name: I2::Dev::Nightly::Publish
 
 on: workflow_dispatch
 
+env:
+  DOCKER_COMPOSE_PATH: defaults
+
 jobs:
+  build_executor:
```
s8sato, line 9: Can we somehow avoid this repetition i.e. can we have the only `build_executor` job somewhere and make the subsequent jobs call it?

.gitignore
BAStos525: Executor is built every time when developer commits a code to PR triggers. If there are some cases where executor shouldn't be built, please welcome to remove it then. And it also is built when we compile iroha image

mversic: > A similar question as [#4931 (review)](https://github.com/hyperledger/iroha/pull/4931#pullrequestreview-2218446270), are developers supposed to checkout a certain commit, build executor, and run tests?

yes, but we can make it part of the build script for client tests?

.github/workflows/iroha2-dev-nightly.yml
BAStos525: Yes, but it's not a good decision. We might do it, but from my нexperience it requires to move workflows from PR trigger to `workflow run` trigger that will break some things. And also we might encounter GitHub API rate limit. That is, too significant disadvantages.

mversic:
> > How should i run iroha locally without docker or run tests which require executor? Compile it locally?
> 
> If you need to compile iroha locally, I guess you have to compile executor locally as well or to download it from Actions artifacts if it's appropriate?

it would be nice to add things as artifacts. executor, configs, genesis, etc. Let's make it a separate issue

.gitignore
mversic: let's make a tracking issue for this

mversic:
> > > How should i run iroha locally without docker or run tests which require executor? Compile it locally?
> > 
> > 
> > If you need to compile iroha locally, I guess you have to compile executor locally as well or to download it from Actions artifacts if it's appropriate?
> 
> it would be nice to add things as artifacts. executor, configs, genesis, etc. Let's make it a separate issue



> > > How should i run iroha locally without docker or run tests which require executor? Compile it locally?
> > 
> > 
> > If you need to compile iroha locally, I guess you have to compile executor locally as well or to download it from Actions artifacts if it's appropriate?
> 
> it would be nice to add things as artifacts. executor, configs, genesis, etc. Let's make it a separate issue

opened #4942

.gitignore
mversic: opened #4943

.github/workflows/iroha2-dev-nightly.yml
s8sato: Ok, I found it is not simple to achieve an equivalent of the combined condition of `on.workflow_run` and `on.pull_request.paths` when we want to take the single build approach.

On the other hand, what if we move the `build_executor` job to a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) e.g. `build-executer.yml` and replace every `build_executor` job with a step that `uses: ./.github/workflows/build-executer.yml`? It would at least save the literal repetition, but I don't know if it hits the API rate limit

.gitignore
```diff
@@ -50,3 +50,5 @@ result
 /iroha-java/
 /lcov.info
 **/test-smartcontracts/
+
+defaults/executor.wasm
```
mversic, line 54:
```suggestion
*.wasm
```

mversic (approved):

.github/workflows/iroha2-dev-nightly.yml
```diff
@@ -2,13 +2,39 @@ name: I2::Dev::Nightly::Publish
...
+    runs-on: ubuntu-latest
+    container:
+      image: hyperledger/iroha2-ci:nightly-2024-04-18
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+      - name: Build iroha executor
+        run: cargo run --bin iroha_wasm_builder -- build ./wasm_samples/default_executor --optimize --out-file ${{ env.DOCKER_COMPOSE_PATH }}/executor.wasm
```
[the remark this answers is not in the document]
BAStos525: What about those tests that don't use iroha in containers?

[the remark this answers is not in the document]
mversic: these are `tori-api-client-cli` tests?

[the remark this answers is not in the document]
BAStos525: `torii-api-client-cli` use containers, right. But we have many other tests jobs that use cargo without containers, so what about them?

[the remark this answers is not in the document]
mversic: and which tests are those? can you give an example?

.github/workflows/iroha2-pr-ui.yml
```diff
@@ -15,10 +15,29 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
+  DOCKER_COMPOSE_PATH: defaults
 
 jobs:
+  build_executor:
```
mversic, line 21: I don't think executor is required to execute `ui` tests

.github/workflows/iroha2-profiling-image.yml
```diff
@@ -11,10 +11,29 @@ env:
   IROHA2_RUSTFLAGS: -C force-frame-pointers=on
   IROHA2_FEATURES: profiling
   IROHA2_CARGOFLAGS: -Z build-std
+  DOCKER_COMPOSE_PATH: defaults
 
 jobs:
+  build_executor:
```
mversic, line 17: why do you need executor to publish image?

BAStos525: We need to keep executor in image and do not copy it from GitHub while deployment

.github/workflows/iroha2-dev-nightly.yml
BAStos525: Considered with Sato that it's not acceptable for our CI case

nxsaken:
@BAStos525 did you intend to wipe my commit with the swarm fix?

.github/workflows/iroha2-pr-ui.yml
nxsaken: What is the resolution here? I see it was agreed that the executor is not needed, but the job is still here

BAStos525: It was removed [here](https://github.com/hyperledger/iroha/pull/4931/commits/c1490f414a4b103676d56109a8254476b24ee4e9#diff-f4db2cb3367415c444df62a5359b0ca64b4930f0dfb52e29c652ebc921804722) and reverted [here](https://github.com/hyperledger/iroha/pull/4931/commits/bca9adbd20620e4cb4fb9bd96981e57620c87dc8)

nxsaken: So we do need the executor for UI tests?

Dockerfile.glibc
```diff
@@ -37,26 +37,32 @@ ENV  USER=iroha
...
-     mkdir -p $CONFIG_DIR && \
-     mkdir -p $STORAGE && \
-     mkdir -p $WASM_DIRECTORY && \
-     chown $USER:$USER $STORAGE && \
-     chown $USER:$USER $WASM_DIRECTORY
+RUN <<EOT
+  set -ex
+  apt-get update -y && \
+    apt-get install -y curl ca-certificates jq
+  addgroup --gid $GID $USER &&
```
nxsaken, line 44:
```suggestion
  apt-get update -y
  apt-get install -y curl ca-certificates jq
  addgroup --gid $GID $USER
```

.github/workflows/iroha2-dev.yml
```diff
@@ -42,40 +69,19 @@ jobs:
...
-          cargo run --bin kagami -- schema >target/schema/schema.json
-      - name: Archive schema
-        uses: actions/upload-artifact@v3
+          mkdir -p ${{ env.ARTIFACTS_DIR }}
+          docker run --rm --entrypoint cat hyperledger/iroha:dev ${{ env.BIN_PATH }}/iroha > ${{ env.ARTIFACTS_DIR }}/iroha
+          docker run --rm --entrypoint cat hyperledger/iroha:dev ${{ env.BIN_PATH }}/irohad > ${{ env.ARTIFACTS_DIR }}/irohad
+          docker run --rm --entrypoint cat hyperledger/iroha:dev ${{ env.BIN_PATH }}/kagami > ${{ env.ARTIFACTS_DIR }}/kagami
+      - name: Generate schema
+        run: cargo run --bin kagami -- schema > ${{ env.ARTIFACTS_DIR }}/schema.json
+      - name: Archieve artifacts
```
nxsaken, line 80:
```suggestion
      - name: Archive artifacts
```

.github/workflows/iroha2-pr-ui.yml
BAStos525: Removed

Dockerfile.glibc
nxsaken: This whole block is already treated like one operation because of `set -e` short-circuiting on failure

s8sato (approved):

.github/workflows/iroha2-dev.yml
```diff
@@ -42,40 +69,19 @@ jobs:
...
+          docker run --rm --entrypoint cat hyperledger/iroha:dev ${{ env.BIN_PATH }}/kagami > ${{ env.ARTIFACTS_DIR }}/kagami
+      - name: Generate schema
+        run: cargo run --bin kagami -- schema > ${{ env.ARTIFACTS_DIR }}/schema.json
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v4
         with:
-          name: schema
-          path: target/schema
+          name: iroha_binaries_and_schema
+          path: ${{ env.ARTIFACTS_DIR }}
```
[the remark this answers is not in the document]
BAStos525: I think our goal is to store executor, kagami, iroha and irohad as artifacts. And also schema.json. It doesn't touch genesis and configs. We just merged two jobs in one do not repeat the same things.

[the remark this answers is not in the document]
s8sato: It sounds reasonable to me to upload only binaries, excluding generative non-binaries tracked in git atm. What do you think @mversic

s8sato (dismissed):

nxsaken (dismissed):

nxsaken:
@BAStos525 I will merge after the conflicts are resolved

0x009922:
> 1. Developers shouldn't compile the `executor.wasm` file locally and push it to the repository more.

It is not fully correct: developers still need to compile the `executor.wasm` locally in order to run integration tests which rely on it.

mversic:
> > 1. Developers shouldn't compile the `executor.wasm` file locally and push it to the repository more.
> 
> It is not fully correct: developers still need to compile the `executor.wasm` locally in order to run integration tests which rely on it.

We need to add this to the `build.rs`

DCNick3:
> We need to add this to the `build.rs`

I don't believe this is a good idea unless we also put integration tests into a different crate (and even then it's... questionable). Otherwise, every `cargo check` on the client will cause unnecessary (and quite long) rebuilds.

I think we should only build the executor if the integration tests are actually being ran

Issue #4727, opened 2024-06-13
[BUG] bandwidth degradation


timofeevmd:
### OS and Environment

linux, k8s

### GIT commit hash

0275cfa

### Minimum working example / Steps to reproduce

Test preconditions

- Infrastructure
- iroha version: version="2.0.0-pre-rc.21.3" git_commit_sha="0275cfa"
- java sdk version: commit: commit: 69c6515
- 5 peers

Create:
- 5 domains;
- 250 unique users for each domain;
- for each user mint 10,000 units of assets;

Test
- transfer assets from one user to another;
- Longevity state of the stand before starting the test
- clearing storage

### Actual result

Throughput does not increase:

[profiler](https://parca.tachi.soramitsu.co.jp/?search_string=&dashboard_items=icicle&filter_by_function=&compare_a=true&expression_a=process_cpu%3Acpu%3Ananoseconds%3Acpu%3Ananoseconds%3Adelta%7B%7D&from_a=1718113800000&to_a=1718115300000&time_selection_a=absolute:1718113800000-1718115300000&merge_from_a=1718113800000&merge_to_a=1718115300000&compare_b=true&expression_b=process_cpu%3Acpu%3Ananoseconds%3Acpu%3Ananoseconds%3Adelta%7B%7D&from_b=1718115300000&to_b=1718116800000&time_selection_b=absolute:1718115300000-1718116800000&merge_from_b=1718115300000&merge_to_b=1718116800000)

[grafana](https://grafana.soramitsu.co.jp/d/BD5RpSFVk/kubernetes-deployments?orgId=1&from=1718113303000&to=1718136003000&var-namespace=iroha2-test)

[queue](https://grafana.soramitsu.co.jp/d/1kmsWQQ4z/iroha2-qa?orgId=1&var-namespace=iroha2-test&from=1718113303134&to=1718136003879)

Problems detected with modules
iroha_core::gossiper::TransactionGossiper::run
iroha_core::_::<impl parity_scale_codec::codec::Decode for iroha_core
iroha_core::sumeragi::main_loop::handle_block_sync

### Expected result

Throughput increases in proportion to the load

### Logs

<details>
  <summary>Log contents</summary>
  
  ```json
  Replace this text with a JSON log,
  so it doesn't grow too large and has highlighting.
  ```
</details>

### Who can help to reproduce?

@timofeevmd

Erigara:
Should be noticed as well that degradation happens in case of single peer in the network as well.

[...]

PR #4947, opened 2024-08-06
fix: Improve queue transactions handling

dima74:
## Description

Currently in some cases tps numbers (transactions per second) may be very unstable for network of a single peer. See https://github.com/hyperledger/iroha/issues/4727#issuecomment-2268694679 and https://github.com/hyperledger/iroha/issues/4727#issuecomment-2269086095 for detailed description of the reasons. This PR adjusts how transactions are handled in the queue, so tps will be more stable. Note that average tps should not change (maybe slightly improvement).

Here is comparison of tps:

| Before                                                                                            | After                                                                                          |
|---------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------|
| 1600_main | 1600_fixed |

### Linked issue

Related: #4727

### Benefits

Improve stability of the tps in case when there are a lot of requests and transaction queue is full

### Checklist


core/src/sumeragi/main_loop.rs
```diff
@@ -375,7 +375,13 @@ impl Sumeragi {
 
     fn cache_transaction(&mut self, state_block: &StateBlock<'_>) {
         self.transaction_cache.retain(|tx| {
-            !state_block.has_transaction(tx.as_ref().hash()) && !self.queue.is_expired(tx)
+            let has_transaction = state_block.has_transaction(tx.as_ref().hash());
+            let retain = !has_transaction && !self.queue.is_expired(tx);
+            if !retain {
+                let is_expired = !has_transaction;
```
Erigara, line 381: Shouldn't be `let is_expired = self.queue.is_expired(tx)`?

Erigara (commented):
LGTM, other than small comment

Erigara (dismissed):

core/src/queue.rs
```diff
@@ -304,12 +306,10 @@ impl Queue {
...
         }
 
-        let mut seen_queue = Vec::new();
         let mut expired_transactions = Vec::new();
 
-        let txs_from_queue = core::iter::from_fn(|| {
-            self.pop_from_queue(&mut seen_queue, state_view, &mut expired_transactions)
-        });
+        let txs_from_queue =
+            core::iter::from_fn(|| self.pop_from_queue(state_view, &mut expired_transactions));
```
mversic, line 312:
```suggestion
        let txs_from_queue =
            core::iter::from_fn(|| self.pop_from_queue(state_view, &mut expired_transactions));
```

could `pop` return a `TransactionGuard` instead? so that when the guard is dropped the transaction is also removed from the queue. I suppose `TransactionGuard` would keep a reference to the `Queue` and have an implementation of `Drop` that calls `remove_stale_transaction`

core/src/queue.rs
```diff
@@ -335,6 +330,31 @@ impl Queue {
...
+    /// 1. Transaction is added to queue using [`Queue::push`] method.
+    /// 2. Transaction is moved to [`Sumeragi::transaction_cache`] using [`Queue::pop_from_queue`] method.
+    ///    (transaction is removed from [`Queue::tx_hashes`], but kept in [`Queue::accepted_tx`])
+    /// 3. When transaction is removed from [`Sumeragi::transaction_cache`]
+    ///    (either because it was expired, or because transaction is commited to blockchain),
+    ///    we should remove transaction from [`Queue::accepted_tx`].
+    ///
+    /// `is_expired` is passed as a small optimization to not again call `self.is_expired(tx)`.
+    pub fn remove_stale_transaction(&self, tx: &AcceptedTransaction, is_expired: bool) {
+        let removed = self.accepted_txs.remove(&tx.as_ref().hash());
```
mversic, line 343: why don't we remove transaction from `accepted_txs` at the time when we remove `tx_hashes`?

dima74: It is used to return `Error::IsInQueue` when pushing same transaction twice

mversic: neat, I didn't see that

core/src/queue.rs
```diff
@@ -335,6 +351,29 @@ impl Queue {
...
     }
 
+    /// Overview:
+    /// 1. Transaction is added to queue using [`Queue::push`] method.
+    /// 2. Transaction is moved to [`Sumeragi::transaction_cache`] using [`Queue::pop_from_queue`] method.
+    ///    (transaction is removed from [`Queue::tx_hashes`], but kept in [`Queue::accepted_tx`])
+    /// 3. When transaction is removed from [`Sumeragi::transaction_cache`]
+    ///    (either because it was expired, or because transaction is commited to blockchain),
+    ///    we should remove transaction from [`Queue::accepted_tx`].
+    pub fn remove_stale_transaction(&self, tx: &AcceptedTransaction) {
```
mversic, line 361:
```suggestion
    fn remove_stale_transaction(&self, tx: &AcceptedTransaction) {
```

core/src/queue.rs
mversic: can you add a comment about that?

mversic (approved):

core/src/queue.rs
```diff
@@ -82,6 +82,27 @@ pub struct Failure {
     pub err: Error,
 }
 
+/// Will remove transaction from the queue on drop.
+/// See [`Queue::remove_stale_transaction`] for details.
+pub struct TransactionGuard {
+    tx: AcceptedTransaction,
+    queue: Arc<Queue>,
+}
```
mversic, line 90:
```suggestion
pub struct TransactionGuard<'q> {
    tx: AcceptedTransaction,
    queue: &'q Queue,
}
```

wouldn't this have been possible?

SamHSmith (approved):
LGTM

mversic (approved):

mversic:
failing tests

core/src/queue.rs
dima74: If we remove `Arc` from `TransactionGuard`, basically we will have this:
```rust
struct Queue {}

struct Sumeragi<'q> {
    queue: Arc<Queue>,
    transaction_cache: Vec<&'q Queue>,
}

fn run(mut sumeragi: Sumeragi) {
    sumeragi.transaction_cache.push(&sumeragi.queue);
}
```

I don't see how this can pass borrow checker

mversic (approved):

PR #4955, opened 2024-08-08
perf: Fix cloning of executor

dima74:
## Description

I am debugging tps of a single peer (#4727), measured performance of `Sumeragi::try_create_block`, and noticed that executor cloning takes about 25% of `try_create_block`. This is because of `LoadedExecutor::raw_executor` which was added for snapshot creation. Could be fixed by adding `Arc`.

It is also interesting that other wasm-related code takes a lot of time, e.g. `wasmtime::Linker::instantitate` and `wasmtime::Store::into_data`.

image

### Linked issue

Related: #4727

### Benefits

Slightly optimize block creation performance

### Checklist

### DIFF
--- a/core/src/executor.rs
+++ b/core/src/executor.rs
@@ -1,14 +1,12 @@
 //! Structures and impls related to *runtime* `Executor`s processing.
 
-use std::sync::Arc;
-
 use derive_more::DebugCustom;
 use iroha_data_model::{
     account::AccountId,
     executor as data_model_executor,
     isi::InstructionBox,
     query::{AnyQueryBox, QueryRequest},
-    transaction::{Executable, SignedTransaction},
+    transaction::{base64_util::Base64Wrapper, Executable, SignedTransaction},
     ValidationFail,
 };
 use iroha_logger::trace;
@@ -247,7 +245,7 @@ impl Executor {
     /// - Failed to execute entrypoint of the WASM blob.
     pub fn migrate(
         &mut self,
-        raw_executor: data_model_executor::Executor,
+        raw_executor: &data_model_executor::Executor,
         state_transaction: &mut StateTransaction<'_, '_>,
         authority: &AccountId,
     ) -> Result<(), wasm::error::Error> {
@@ -278,21 +276,21 @@ impl Executor {
 #[derive(DebugCustom, Clone, Serialize)]
 #[debug(fmt = "LoadedExecutor {{ module: <Module is truncated> }}")]
 pub struct LoadedExecutor {
-    #[serde(skip)]
+    #[serde(serialize_with = "wasm::serialize_module_base64")]
     module: wasmtime::Module,
-    /// Arc is needed so cloning of executor will be fast.
-    /// See [`crate::tx::TransactionExecutor::validate_with_runtime_executor`].
-    raw_executor: Arc<data_model_executor::Executor>,
 }
 
 impl LoadedExecutor {
+    fn new(module: wasmtime::Module) -> Self {
+        Self { module }
+    }
+
     fn load(
         engine: &wasmtime::Engine,
-        raw_executor: data_model_executor::Executor,
+        raw_executor: &data_model_executor::Executor,
     ) -> Result<Self, wasm::error::Error> {
         Ok(Self {
             module: wasm::load_module(engine, &raw_executor.wasm)?,
-            raw_executor: Arc::new(raw_executor),
         })
     }
 }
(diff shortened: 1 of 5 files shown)


mversic (dismissed):

Erigara (dismissed):

core/src/executor.rs
```diff
@@ -269,7 +271,9 @@ impl Executor {
 pub struct LoadedExecutor {
     #[serde(skip)]
     module: wasmtime::Module,
-    raw_executor: data_model_executor::Executor,
+    /// Arc is needed so cloning of executor will be fast.
+    /// See [crate::tx::TransactionExecutor::validate_with_runtime_executor].
```
dima74, line 275: I2::Dev::Sonar-Dojo failing at this line for some reason

mversic (approved):

mversic:
so [this](https://github.com/hyperledger/iroha/blob/main/core/src/tx.rs#L280) comment was correct until executor wasm was added to snapshot. Was using `wasmtime::Module::serialize` considered?

dima74:
> Was using `wasmtime::Module::serialize` considered as a solution instead of adding `LoadedExecutor::raw_executor`?

`wasmtime::Module::serialize` returns something different from what `LoadedExecutor::raw_executor` currently stores (I compared lengths of inner `Vec<u8>` and it differs)

mversic:
> > Was using `wasmtime::Module::serialize` considered as a solution instead of adding `LoadedExecutor::raw_executor`?
> 
> `wasmtime::Module::serialize` returns something different from what `LoadedExecutor::raw_executor` currently stores (I compared lengths of inner `Vec<u8>` and it differs)

yes, I know that. What I'm saying is if you can directly serialize/deserialize the `wasmtime::Module`, you don't need `raw_executor` in `LoadedExecutor`

core/src/smartcontracts/wasm.rs
```diff
@@ -251,6 +255,21 @@ pub fn load_module(engine: &Engine, bytes: impl AsRef<[u8]>) -> Result<wasmtime:
...
+{
+    let bytes = module.serialize().map_err(serde::ser::Error::custom)?;
+    let wrapper = Base64Wrapper(bytes);
+    wrapper.serialize(serializer)
+}
+
+/// Deserialize [`Module`]. Accepts only output of [`Module::serialize`].
+#[allow(unsafe_code)]
+pub(crate) fn deserialize_module(engine: &Engine, bytes: impl AsRef<[u8]>) -> Result<Module> {
+    unsafe { Module::deserialize(engine, bytes).map_err(Error::ModuleDeserialization) }
```
mversic, line 270: I think you should put a comment here. Why is it acceptable to use this unsafe method?

dima74: Added

mversic (approved):

Erigara:
I think changing to `wasmtime::Module` could cause problems because `serialize/deserialize` only works for the same version of `wasmtime` and in other case can produce errors.
It might be dangerous as well because random changes in serialized snapshot can cause memory safety issues.

core/src/smartcontracts/wasm.rs
```diff
@@ -251,6 +255,23 @@ pub fn load_module(engine: &Engine, bytes: impl AsRef<[u8]>) -> Result<wasmtime:
...
+    let bytes = module.serialize().map_err(serde::ser::Error::custom)?;
+    let wrapper = Base64Wrapper(bytes);
+    wrapper.serialize(serializer)
+}
+
+/// Deserialize [`Module`]. Accepts only output of [`Module::serialize`].
+#[allow(unsafe_code)]
+pub(crate) fn deserialize_module(engine: &Engine, bytes: impl AsRef<[u8]>) -> Result<Module> {
+    // SAFETY: `Module::deserialize` is safe when calling for bytes received from `Module::serialize`.
+    // We store serialization result on disk and then load it back so should be ok.
```
Erigara, line 271: I think we can't grantee that...

mversic:
> I think changing to `wasmtime::Module` could cause problems because `serialize/deserialize` only works for the same version of `wasmtime` and in other case can produce errors.

How can we end in this situation that we serialize with one and deserialize with another version of `wasmtime`?

Erigara:
> How can we end in this situation that we serialize with one and deserialize with another version of wasmtime

@mversic update of iroha version which in theory should preserve backward compatibility for block store and snapshot.

dima74:
So we should change back to `LoadedExecutor::raw_executor` approach

mversic:
> So we should change back to `LoadedExecutor::raw_executor` approach

I think so. Just do `git revert` and switch back to `Arc`

dima74:
Opened #5009

Issue #4952, opened 2024-08-08
bug: self dos of transactions gossip


Erigara:
While working on #4265 i've noticed that in case of heavy transaction increasing commit time does not help.
Additionally time between leader creating the block and other nodes receiving it increased as view changes grow.

Turned out it was due to repeatedly rebroadcasted transaction gossip which overloaded communication channels between 2 peers. 
This issue is similar to one discovered by @SamHSmith  in #4909.
However i'm not sure that solution used in this PR would work here because sending hash of every tx with each gossip request is too much of data.

One possible solution is to only gossip transaction on it's arrival from the client.

Q: What if not delivered?
A: Than one who received it should put it in the block.

Q: What if this peer is malicious?
A: To be sure that your transaction not going to be censored you should submit it to at least f + 1 nodes.

But i going to do additional research in how to handle transaction gossip.

Erigara:
Simplest solution is to just increase `transaction_gossip_period_ms`, but it still has a problem that the same transactions are sent over and over again.

Erigara:
Transaction i've sent to iroha:

transfer.json:
```json
{
    "Transfer": {
        "Asset": {
            "Numeric": {
                "source": "rose##ed0120CE7FA46C9DCE7EA4B125E2E36BDB63EA33073E7590AC92816AE1E861B7048B03@wonderland",
                "object": "10",
                "destination": "ed0120CE7FA46C9DCE7EA4B125E2E36BDB63EA33073E7590AC92816AE1E861B7048B03@wonderland"
            }
        }
    }
}
```

transaction.json:
```bash
yes transfer.json | head -n 4096 | xargs cat | jq --slurp '.' > transaction.json
```

```bash
cat transaction.json | /iroha --config client.toml json transaction
```

s8sato:
> Q: What if this peer is malicious?

I think this cannot be a drawback of the solution, even if the malicious peer is the only one who could gossip a certain transaction. It would be the same as before that a transaction might be dropped by a malicious entrypoint.

I'm more concerned about if transactions are assumed to be forwarded in one shot to all of validators

Erigara:
> I'm more concerned about if transactions are assumed to be forwarded in one shot to all of validators

Can you clarify your concern?

PR #4957, opened 2024-08-09
feat(sumeragi): dynamic commit time based on view change index

### LINKED ISSUE #4265: Consensus should continue working in the presense of heavy transactions
Invalid transactions take a lot of time to process. This is probably because they require cloning of `wsv`. When there is a batch of invalid transactions in a block the time to create a block takes long. If `max_transactions_in_block` is large or `commit_time` is low this can prevent consensus from operating because it will never create a new block in required time.

What I can suggest at this point is that we can implement some sort of backoff mechanism to reduce the amount of transactions in a block in such cases. We can also consider straightout rejecting transactions that take too long to process. The main thing is that consensus continues operating under all circumstances

Erigara:
## Description

- Commit time is now linear function of view change index (capped at Duration::MAX)
- Add more logs to track messages arriving to sumeragi

Note: #4929 would also nice to have because it prevents degradation of view change updates.

### Linked issue

Closes #4265 

### Benefits

This allow to consensus to pass heavy transactions which cause infinite loop previously.

### How to test?

Currently it's not possible to modulary test consensus :(

So i've did my tests on bare metal iroha with 4 nodes.

Parameters:

config.toml:
```toml
[network]
transaction_gossip_period_ms = 10_000

[transaction]
time_to_live_ms = 3_600_000
```

genesis.json:
```json
    {
      "Sumeragi": {
        "BlockTimeMs": 100
      }
    },
    {
      "Sumeragi": {
        "CommitTimeMs": 100
      }
    }
```

Notice that `transaction_gossip_period_ms` was crucial here, because without it i faced #4952.  

Check this [comment](https://github.com/hyperledger/iroha/issues/4952#issuecomment-2275820415) for the transaction i've submitted to iroha.

### DIFF
--- a/crates/iroha/tests/integration/triggers/time_trigger.rs
+++ b/crates/iroha/tests/integration/triggers/time_trigger.rs
@@ -18,7 +18,7 @@ use iroha_test_samples::{gen_account_in, load_sample_wasm, ALICE_ID};
 pub fn pipeline_time() -> Duration {
     let default_parameters = SumeragiParameters::default();
 
-    default_parameters.pipeline_time()
+    default_parameters.pipeline_time(0, 0)
 }
 
 fn curr_time() -> core::time::Duration {
--- a/crates/iroha_core/src/sumeragi/main_loop.rs
+++ b/crates/iroha_core/src/sumeragi/main_loop.rs
@@ -178,21 +178,23 @@ impl Sumeragi {
                 })
                 .ok()?;
 
-            let block_vc_index = match &block_msg {
+            match &block_msg {
                 BlockMessage::BlockCreated(bc) => {
-                    Some(bc.block.header().view_change_index as usize)
+                    if (bc.block.header().view_change_index as usize) < current_view_change_index {
+                        trace!(
+                            ty="BlockCreated",
+                            block=%bc.block.hash(),
+                            "Discarding message due to outdated view change index",
+                        );
+                        // ignore block_message
+                        continue;
+                    }
                 }
                 // Signed and Committed contain no block.
                 // Block sync updates are exempt from early pruning.
                 BlockMessage::BlockSigned(_)
                 | BlockMessage::BlockCommitted(_)
-                | BlockMessage::BlockSyncUpdate(_) => None,
-            };
-            if let Some(block_vc_index) = block_vc_index {
-                if block_vc_index < current_view_change_index {
-                    // ignore block_message
-                    continue;
-                }
+                | BlockMessage::BlockSyncUpdate(_) => {}
             }
             return Some(block_msg);
         }
@@ -538,6 +540,13 @@ impl Sumeragi {
                 }
             }
             (BlockMessage::BlockCreated(block_created), Role::ValidatingPeer) => {
+                info!(
+                    peer_id=%self.peer_id,
+                    role=%self.role(),
+                    block=%block_created.block.hash(),
+                    "Block received"
+                );
+
                 let topology = &self
                     .topology
                     .is_consensus_required()
@@ -555,10 +564,23 @@ impl Sumeragi {
                     let msg = BlockSigned::from(&v_block.block);
                     self.broadcast_packet_to(msg, [topology.proxy_tail()]);
 
+                    info!(
+                        peer_id=%self.peer_id,
+                        role=%self.role(),
+                        block=%v_block.block.as_ref().hash(),
+                        "Voted for the block"
+                    );
                     *voting_block = Some(v_block);
                 }
             }
             (BlockMessage::BlockCreated(block_created), Role::ObservingPeer) => {
+                info!(
+                    peer_id=%self.peer_id,
+                    role=%self.role(),
+                    block=%block_created.block.hash(),
+                    "Block received"
+                );
+
                 let topology = &self
                     .topology
                     .is_consensus_required()
@@ -581,7 +603,7 @@ impl Sumeragi {
                             peer_id=%self.peer_id,
                             role=%self.role(),
                             block=%v_block.block.as_ref().hash(),
-                            "Block signed and forwarded"
+                            "Voted for the block"
                         );
                     }
 
@@ -619,6 +641,7 @@ impl Sumeragi {
                 info!(
                     peer_id=%self.peer_id,
                     role=%self.role(),
+                    block=%hash,
                     "Received block signatures"
                 );
 
@@ -658,6 +681,7 @@ impl Sumeragi {
                                         ?actual_hash,
                                         "Block hash mismatch"
                                     );
+                                    *voting_block = Some(voted_block);
                                 } else if let Err(err) =
                                     voted_block.block.add_signature(signature, &self.topology)
                                 {
@@ -667,6 +691,7 @@ impl Sumeragi {
                                         ?err,
                                         "Signature not valid"
                                     );
+                                    *voting_block = Some(voted_block);
                                 } else {
                                     *voting_block =
                                         self.try_commit_block(voted_block, is_genesis_peer);
@@ -698,6 +723,12 @@ impl Sumeragi {
                 BlockMessage::BlockCommitted(BlockCommitted { hash, signatures }),
                 Role::Leader | Role::ValidatingPeer | Role::ObservingPeer,
             ) => {
+                info!(
+                    peer_id=%self.peer_id,
+                    role=%self.role(),
+                    block=%hash,
+                    "Received block committed",
+                );
                 if let Some(mut voted_block) = voting_block.take() {
                     let actual_hash = voted_block.block.as_ref().hash();
 
@@ -799,7 +830,10 @@ impl Sumeragi {
                     .world
                     .parameters()
                     .sumeragi
-                    .pipeline_time();
+                    .pipeline_time(
+                        self.topology.view_change_index(),
+                        self.topology.max_faults() + 1,
+                    );
 
                 std::thread::sleep(pipeline_time * 2);
             } else {
@@ -1000,7 +1034,10 @@ pub(crate) fn run(
     let mut should_sleep = false;
     let mut view_change_proof_chain = ProofChain::default();
     // Duration after which a view change is suggested
-    let mut view_change_time = state.world.view().parameters().sumeragi.pipeline_time();
+    let mut view_change_time = state.world.view().parameters().sumeragi.pipeline_time(
+        sumeragi.topology.view_change_index(),
+        sumeragi.topology.max_faults() + 1,
+    );
     // Instant when the previous view change or round happened.
     let mut last_view_change_time = Instant::now();
 
@@ -1049,7 +1086,12 @@ pub(crate) fn run(
 
         reset_state(
             &sumeragi.peer_id,
-            state.world.view().parameters().sumeragi.pipeline_time(),
+            state
+                .world
+                .view()
+                .parameters()
+                .sumeragi
+                .pipeline_time(view_change_index, sumeragi.topology.max_faults() + 1),
             view_change_index,
             &mut sumeragi.was_commit,
             &mut sumeragi.topology,
@@ -1154,12 +1196,22 @@ pub(crate) fn run(
 
             // NOTE: View change must be periodically suggested until it is accepted.
             // Must be initialized to pipeline time but can increase by chosen amount
-            view_change_time += state.world.view().parameters().sumeragi.pipeline_time();
+            view_change_time += state
+                .world
+                .view()
+                .parameters()
+                .sumeragi
+                .pipeline_time(view_change_index, sumeragi.topology.max_faults() + 1);
         }
 
         reset_state(
             &sumeragi.peer_id,
-            state.world.view().parameters().sumeragi.pipeline_time(),
+            state
+                .world
+                .view()
+                .parameters()
+                .sumeragi
+                .pipeline_time(view_change_index, sumeragi.topology.max_faults() + 1),
             view_change_index,
             &mut sumeragi.was_commit,
             &mut sumeragi.topology,
--- a/crates/iroha_core/src/sumeragi/mod.rs
+++ b/crates/iroha_core/src/sumeragi/mod.rs
@@ -44,6 +44,7 @@ pub struct SumeragiHandle {
 impl SumeragiHandle {
     /// Deposit a sumeragi control flow network message.
     pub fn incoming_control_flow_message(&self, msg: ControlFlowMessage) {
+        trace!(ty = "ViewChangeProofChain", "Incoming message");
         if let Err(error) = self.control_message_sender.try_send(msg) {
             self.dropped_messages_metric.inc();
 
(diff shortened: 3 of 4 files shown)


SamHSmith (approved):

core/src/sumeragi/main_loop.rs
```diff
@@ -936,7 +976,8 @@ fn reset_state(
         *voting_block = None;
         voting_signatures.clear();
         *last_view_change_time = Instant::now();
-        *view_change_time = pipeline_time;
+        *view_change_time =
+            pipeline_time.saturating_mul((view_change_index + 1).try_into().unwrap_or(u32::MAX));
```
s8sato, line 980: Why increase the transaction complexity allowance based on `pipeline_time` which includes `block_time`, when block creation seems independent of transaction complexity?

core/src/sumeragi/main_loop.rs
```diff
@@ -1138,7 +1179,13 @@ pub(crate) fn run(
...
+            view_change_time += state
+                .world
+                .view()
+                .parameters()
+                .sumeragi
+                .pipeline_time()
+                .saturating_mul((view_change_index + 1).try_into().unwrap_or(u32::MAX));
         }
 
         reset_state(
```
s8sato, line 1191: Here `view_change_time` is nonlinear with `view_change_index`.
Also, doesn't this increment overlap with the one in the `reset_state()`?

Erigara: > Also, doesn't this increment overlap with the one in the reset_state()?

Inside `reset_state` `view_change_time` is reset to trigger view change index after `pipeline_time` * `view_change_index`.

> Here view_change_time is nonlinear with view_change_index.

Why not? `view_change_time` is increased in chunks of `pipeline_time` * `view_change_index` size.
So it would be like: 2, 4, 6, 8, ... so it's linear.

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
Erigara: Eventually heavy transaction would be either rejected or committed so infinite loop is prevented.
But i think you right that having unknown time might be poor user experience. 

If we would have a way to have impose bounded block execution time than this change is not necessary.

core/src/sumeragi/main_loop.rs
s8sato: [my branch]: https://github.com/s8sato/iroha/commit/54062b8fe5b5c1cffea59bc1d6818b739a94fad5

In [my branch] I found it was indeed linear when logged at line 981 but strange when logged at line 1189.

- [54062b8fe.log](https://github.com/user-attachments/files/16760619/54062b8fe.log)

This is part of my ERROR level log. I intentionally killed 2 peers to cause view changes

<details><summary>details</summary>

```txt
$ cat 54062b8fe.log | grep -B1 main_loop.rs:981

  2024-08-27T09:37:37.382184Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 1, view_change_time: 12s
    at core/src/sumeragi/main_loop.rs:981
--
  2024-08-27T09:37:49.410953Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 2, view_change_time: 18s
    at core/src/sumeragi/main_loop.rs:981
--
  2024-08-27T09:38:07.507803Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 3, view_change_time: 24s
    at core/src/sumeragi/main_loop.rs:981
--
  2024-08-27T09:38:31.535742Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 4, view_change_time: 30s
    at core/src/sumeragi/main_loop.rs:981
--
  2024-08-27T09:39:01.703077Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 36s
    at core/src/sumeragi/main_loop.rs:981
```

```txt
$ cat 54062b8fe.log | grep -B1 main_loop.rs:1189

  2024-08-27T09:37:36.437753Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 18s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.451240Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 24s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.464787Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 30s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.478254Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 36s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.491821Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 42s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.505859Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 48s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.522229Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 54s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.536978Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 60s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.551447Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 66s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.565179Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 72s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.579937Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 78s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:36.593612Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 0, view_change_time: 84s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:37:49.395329Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 1, view_change_time: 24s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:38:07.419794Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 2, view_change_time: 36s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:38:31.520076Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 3, view_change_time: 48s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:39:01.547381Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 4, view_change_time: 60s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:39:37.704183Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 72s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:40:13.707816Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 108s
    at core/src/sumeragi/main_loop.rs:1189
--
  2024-08-27T09:40:49.703807Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 144s
    at core/src/sumeragi/main_loop.rs:1189
```

</details>

core/src/sumeragi/main_loop.rs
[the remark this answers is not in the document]
s8sato: > bounded block execution time

If it times out, what happens to the other good transactions involved?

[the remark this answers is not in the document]
Erigara: So if block time is bounded for example by max amount of fuel per block, then leader is responsible to provide block which satisfy this limitation (picking proper transactions).
If block exceed this limit it should be rejected by other peers.

core/src/sumeragi/main_loop.rs
Erigara: About line 1189, you can see that increments within the same view change index are constant. 
For:
- 0 it's 6 seconds
- 1 it's 12 seconds
- 2 it's 18 seconds
- 3 it's 24 seconds
- ...
So `view_change_time` is incremented with constant values within given `view_change_index`.
Sumeragi suggest a view change when since `last_view_change` elapsed more time than `view_change_time`.
Lets look for example at index 5:
- `adjusted_pipeline_time = pipeline_time * (view_change_index + 1) = 6s * 6 = 36s`
- so first view change would be suggested after 36 seconds since view change started:
```
  2024-08-27T09:39:01.703077Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 36s
    at core/src/sumeragi/main_loop.rs:981
  2024-08-27T09:39:37.704183Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 72s
    at core/src/sumeragi/main_loop.rs:1189
```
- you can see that ~36 seconds elapsed since between this 2 records
- the next attempt would be after 72 seconds after view change started (or 36 seconds after previous attempt):
```
  2024-08-27T09:39:37.704183Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 72s
    at core/src/sumeragi/main_loop.rs:1189

  2024-08-27T09:40:13.707816Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 108s
    at core/src/sumeragi/main_loop.rs:1189

  2024-08-27T09:40:49.703807Z ERROR iroha_core::sumeragi::main_loop: view_change_index: 5, view_change_time: 144s
    at core/src/sumeragi/main_loop.rs:1189
```
- again you can see that since first and the last record about 72 seconds elapsed (or 36 between last and previous)

s8sato: I see, so increments per index and increments per attempt within the same index are different things, and in the latter case `view_change_time` indicates not a duration between attempts but a point in time within the same index because the origin (`last_view_change_time`) is not reset

s8sato (approved):

mversic (changes requested):

core/src/sumeragi/main_loop.rs
```diff
@@ -858,7 +889,25 @@ impl Sumeragi {
...
+            let pipeline_time = state
+                .world
+                .view()
+                .parameters()
+                .sumeragi
+                .pipeline_time()
+                .saturating_mul(
+                    (self.topology.view_change_index() + 1)
+                        .try_into()
+                        .unwrap_or(u32::MAX),
```
mversic, line 909:
```suggestion
                .pipeline_time(view_change_index),
```

how about?

mversic:
this will increase `pipeline_time` even in the presence of malicious blocks (not just too complex blocks). How about if we add a shift so that first increase in pipeline time is after `x` first view changes? Thereafter the increase should happen on every view change

core/src/sumeragi/view_change.rs
```diff
@@ -22,8 +22,9 @@ pub enum Error {
     ViewChangeNotFound,
 }
 
-#[derive(Debug, Clone, Decode, Encode)]
-struct ViewChangeProofPayload {
+/// Payload of view change.
+#[derive(Debug, Clone, Copy, Decode, Encode)]
+pub struct ViewChangeProofPayload {
```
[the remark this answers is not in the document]
Erigara: accidental, due to refactor on view change

going to revert

core/src/sumeragi/main_loop.rs
Erigara: yes, good suggestion, also implemented what @s8sato asked [here](https://github.com/hyperledger/iroha/pull/4957#discussion_r1724813277)

Erigara:
> this will increase `pipeline_time` even in the presence of malicious blocks (not just too complex blocks). How about if we add a shift so that first increase in pipeline time is after `x` first view changes? Thereafter the increase should happen on every view change

Added `shift` parameter and made it `max_faults + 1` by default, because we can have at most `max_faults` malicious blocks in a row.

core/src/sumeragi/main_loop.rs
```diff
@@ -1140,12 +1184,22 @@ pub(crate) fn run(
 
             // NOTE: View change must be periodically suggested until it is accepted.
             // Must be initialized to pipeline time but can increase by chosen amount
-            view_change_time += state.world.view().parameters().sumeragi.pipeline_time();
+            view_change_time += state
+                .world
+                .view()
+                .parameters()
+                .sumeragi
+                .pipeline_time(view_change_index, sumeragi.topology.max_faults() + 1);
```
mversic, line 1192:
```suggestion
                .pipeline_time(view_change_index, sumeragi.topology.max_faults() + 1);
```

IMO `max_faults` is too much. Let's hardcode to some smaller number for now

Erigara: It would be 2 for 4 network and 3 for 7 node network.
What in your opinion would be reasonable number?

mversic: and 33 for 100. That's pretty big. I'd suggest maybe 5 as the upper limit? We could try to make some calculation. What would have to be the value of `shift` so that probability that failure is due to a malicious node is less than p? or something along those lines

Erigara: If we assume that every faulty node try to submit genesis peer than probability that given peer would submit incorrect block is ~1/3.
And shift can't increase or decrease this probability only change time when block will be committed.

If we indeed have large block than approach with shift is actually worse by factor `shift` * `commit_time`.

In case where we have `f` wrong blocks in a row and than valid approach with a shift behave better by:
`diff = commit_time * (f + 2) * (f +1) / 2 - commit_time * (f + 1) = commit_time * (f + 1) * (f - 1) / 2`.

So the difference is quadratic of `f`.

Examples (commit_time=1s):
- f=1 => diff=0s
- f=2 => diff=1.5s
- f=3 => diff=4s
- f=4 => diff=7.5s
- f=33 => diff=544s

Erigara: So having fixed shift might theoretically slightly improve latency for the case where we have multiple invalid blocks in a row.

mversic:
I approve that the growth is linear with `view_change_index`. We don't want to have exponential growth because malicious peers can have larger impact on the network liveness in that case

core/src/sumeragi/main_loop.rs
s8sato: When faulty peers (including faulty leader i.e. malicious block) are least concerned, the shift would be 0. When most concerned, the shift would be 2f+1 (sufficient rotations to exclude every faulty peer from the validating set).

After all, since no theoretical weights can be assigned between the causes (complexity and disagreement), the only way would be to determine a constant

mversic: @SamHSmith

SamHSmith: @Erigara @mversic My suggestion is that we add a parameter that represents the target maximum amount of faulty nodes. Then the equation can use that. `total_time = block_time + commit_time * max(1.0, view_change_index / max_faulty_count)`

The result is that pipeline time is constant when f < max_faulty_count. The time to commit the block is worst case linear with f. Until we surpass the target max fault count. Then pipeline time because non constant because something is clearly wrong. And the total time to commit the block becomes quadratic.

Erigara: Can you specify wdym by "represents the target maximum amount of faulty nodes"?

SamHSmith: I mean that the Iroha customer could be able to specify how many faulty nodes they want to tolerate in the normal case. We assume that all nodes are online all the time and something is very wrong if they are faulty. It could be that you expect 1-3 nodes to be down for maintenance at any given time in your 20 node network.

On the other hand, it might be that the correct course of action is having smart contracts that automatically remove faulty peers from the network. In that case we are targeting no faulty peers in an iroha network. We can still produce blocks at 1/3 number of peers being faulty but we don't expect to stay at that high amount of faultyness. Then the speed problem becomes less of an issue because we only need to produce a few blocks to start kicking the faulty nodes.

SamHSmith: And maybe for public chains that is combined with some kind of staking or something.

Erigara: I'm concerned that this might complicate things for users. 
And other than that we already have max allowed number of faulty peers (`f` out of `3f + 1`).

mversic: we don't need this to be configurable

mversic:
let's merge this and split that discussion into a separate ticket

mversic (approved):

s8sato (approved):

PR #4959, opened 2024-08-09
feat(test-env): Preserve test env logs on crash

### LINKED ISSUE #4814: `test-env.py` cleanup behaviour makes it hard to debug issues leading to genesis not being committed
# TL;DR

When there's a problem with starting iroha (for example, if it crashed due to port being occupied), `test_env.py` just prints `Genesis block wasn't created within 20 seconds. Aborting...`, attempts to kill all processes and deletes all directories.

This makes it hard to debug what went wrong with iroha, as all the logs also get removed too.

IMO, the script would be more helpful if it left all the directories as-is, while still killing the iroha peers. It can also invite the user to investigate what has happened, telling them where the logs are

DCNick3:
## Description

1. Kill iroha but keep logs if waiting for genesis fails
2. Rename log files to make them not hidden by UNIX conventions

### Linked issue

Closes #4814 

### Benefits

1. Allows you to debug problems that lead to genesis not being committed
2. Log files not being hidden reduces surprises (I had to look into code two times already to find them)

### Checklist

- [ ] make CI pass

### DIFF
--- a/scripts/test_env.py
+++ b/scripts/test_env.py
@@ -150,8 +150,8 @@ class _Peer:
         logging.info(f"Running peer {self.name}...")
 
         # FD never gets closed
-        stdout_file = open(self.peer_dir / ".stdout", "w")
-        stderr_file = open(self.peer_dir / ".stderr", "w")
+        stdout_file = open(self.peer_dir / "log.stdout", "w")
+        stderr_file = open(self.peer_dir / "log.stderr", "w")
         # These processes are created detached from the parent process already
         subprocess.Popen([self.name, "--config", self.config_path],
                     executable=self.out_dir / "peers/irohad", stdout=stdout_file, stderr=stderr_file)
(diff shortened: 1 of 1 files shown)


scripts/test_env.py
```diff
@@ -148,8 +150,8 @@ def run(self):
         logging.info(f"Running peer {self.name}...")
 
         # FD never gets closed
-        stdout_file = open(self.peer_dir / ".stdout", "w")
-        stderr_file = open(self.peer_dir / ".stderr", "w")
+        stdout_file = open(self.peer_dir / "log.stdout", "w")
+        stderr_file = open(self.peer_dir / "log.stderr", "w")
```
Erigara, line 154: is it possible to put logs in the single file?

DCNick3: Yeah, that's probably more convenient.

mversic (approved):

0x009922 (approved):

PR #4961, opened 2024-08-11
feat(client_cli): Integrating CLI API to set and retrieve parameters.

### LINKED ISSUE #4820: API to set parameters through CLI
We should add support to set parameters through `iroha` CLI

divyaranjan1905:
## Description

This PR would integrate the necessary patchwork needed to ensure that one could use the `client_cli` of `iroha` to perform operations to set new values for parameters and retrieve existing ones.

### Linked issue

Closes #4820 

### Benefits

This creates an easier way for the user to modify parameter values.

### Checklist

### DIFF
--- a/client_cli/src/main.rs
+++ b/client_cli/src/main.rs
@@ -1,5 +1,4 @@
 //! Iroha client CLI
-
 use std::{
     fs::{self, read as read_file},
     io::{stdin, stdout},
@@ -101,6 +100,9 @@ enum Subcommand {
     /// The subcommand related to p2p networking
     #[clap(subcommand)]
     Peer(peer::Args),
+    /// The subcommand related to adding config parameters.
+    #[clap(subcommand)]
+    Parameters(param::Args),
     /// The subcommand related to event streaming
     #[clap(subcommand)]
     Events(events::Args),
@@ -165,7 +167,7 @@ macro_rules! match_all {
 impl RunArgs for Subcommand {
     fn run(self, context: &mut dyn RunContext) -> Result<()> {
         use Subcommand::*;
-        match_all!((self, context), { Domain, Account, Asset, Peer, Events, Wasm, Blocks, Json })
+        match_all!((self, context), { Domain, Account, Asset, Peer, Parameters, Events, Wasm, Blocks, Json })
     }
 }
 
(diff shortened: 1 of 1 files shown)


divyaranjan1905:
Okay, so @nxsaken and others. This draft PR might have several errors and the code might not be the best, bare with me as I fix these things.

client_cli/src/main.rs
```diff
@@ -1060,6 +1062,196 @@ mod peer {
...
+                Self::Inner(param) => client
+                    .build_query(client::parameter::all())
+                    .with_filter(ParamCmd::Sumeragi)
+                    .execute()
+                    .wrap_err("Failed to get the parameter value"),
+            }?;
+            context.print_data(&vec.collect::<QueryResult<Vec<_>>>()?)?;
+            Ok(())
+        }
+    }
```
divyaranjan1905, line 1115: So the first thing is here, I tried to replicate this from another commit, but I'm stuck on how to get the specific parameters that are in `ParamCmd`. I'd like some suggestion on that.

client_cli/src/main.rs
```diff
@@ -1060,6 +1062,196 @@ mod peer {
...
+        Inner(ParamCmd),
+    }
+
+    impl RunArgs for Set {
+        fn run(self, context: &mut dyn RunContext) -> Result<()> {
+            let set_param_value = SetParameter::new(0);
+	    // Not sure what to do.
+	    Ok(())
+        }
+    }
```
divyaranjan1905, line 1131: And similarly, how do we exactly take the parameters from the hierarchy we have and actually set them? I believe we need to use `SetParameter` from `data_model::parameter`, but don't know how to go further on it.

nxsaken: `SetParameter` is an `Instruction`, which can be submitted to an Iroha `Client`. You need to turn `self` into a `SetParameter` here, and then call `submit` (defined in this file).

client_cli/src/main.rs
nxsaken: You will also need to make a version of `ParamCmd` and its nested types without the `value` field for this.

divyaranjan1905: > You will also need to make a version of `ParamCmd` and its nested types without the `value` field for this.

@nxsaken I would need to duplicate it again? That seems weird. Maybe `ParamCmd` could be refactored for the better?

mversic:
@nxsaken what's the state with this?

nxsaken:
@divyaranjan1905 hi, do you plan to finish this PR?

divyaranjan1905:
I do, I've simply been on a hiatus due to some stuff. I plan to get back on this within a week or so.

mversic:
should I assume this abandoned?

divyaranjan1905:
> should I assume this abandoned?

It isn't, I have been busy with guiding someone's PhD, I would eventually be pushing this for merge.

Issue #4914, opened 2024-07-29
Sumeragi Leader sends only transaction list


SamHSmith:

SamHSmith:
Currently during a consensus round the leader creates and verifies a block before sending it to the other peers. They then "run" the block by validating it, then sending signatures to the proxy tail. The idea is that we can have the leader select the transactions that go into the block and send only those to be assembled into a block on the receiving side. Then all the peers will be "running" the block at the same time instead of the leader then the other peers. The advantage of this is that it reduces total round latency and can therefore improve total TPS by up to 2x. This change was inspired by issue https://github.com/hyperledger/iroha/issues/4733 .

mversic:
good idea

PR #4967, opened 2024-08-13
fix: Leader as transaction ordering service

SamHSmith:
## Description

#4914 

### Linked issue

Closes #{issue_number} 

### Checklist

### DIFF
--- a/crates/iroha_core/src/block.rs
+++ b/crates/iroha_core/src/block.rs
@@ -153,6 +153,31 @@ mod pending {
             Self(Pending { transactions })
         }
 
+        /// Create new BlockPayload
+        pub fn new_unverified(
+            prev_block: Option<&SignedBlock>,
+            view_change_index: usize,
+            transactions_a: Vec<AcceptedTransaction>,
+            consensus_estimation: Duration,
+        ) -> BlockPayload {
+            let transactions = transactions_a
+                .into_iter()
+                .map(|tx| CommittedTransaction {
+                    value: tx.clone().into(),
+                    error: None,
+                })
+                .collect::<Vec<_>>();
+            BlockPayload {
+                header: Self::make_header(
+                    prev_block,
+                    view_change_index,
+                    &transactions,
+                    consensus_estimation,
+                ),
+                transactions,
+            }
+        }
+
         fn make_header(
             prev_block: Option<&SignedBlock>,
             view_change_index: usize,
@@ -264,7 +289,7 @@ mod chained {
     pub struct Chained(pub(super) BlockPayload);
 
     impl BlockBuilder<Chained> {
-        /// Sign this block and get [`SignedBlock`].
+        /// Sign this block as Leader and get [`SignedBlock`].
         pub fn sign(self, private_key: &PrivateKey) -> WithEvents<ValidBlock> {
             WithEvents::new(ValidBlock(self.0 .0.sign(private_key)))
         }
@@ -415,7 +440,8 @@ mod valid {
             Ok(())
         }
 
-        /// Validate a block against the current state of the world.
+        /// Validate a block against the current state of the world. Individual transaction
+        /// errors will be updated.
         ///
         /// # Errors
         ///
@@ -430,7 +456,7 @@ mod valid {
         /// - Error during validation of individual transactions
         /// - Transaction in the genesis block is not signed by the genesis public key
         pub fn validate(
-            block: SignedBlock,
+            mut block: SignedBlock,
             topology: &Topology,
             expected_chain_id: &ChainId,
             genesis_account: &AccountId,
@@ -442,9 +468,12 @@ mod valid {
                 return WithEvents::new(Err((block, error)));
             }
 
-            if let Err(error) =
-                Self::validate_transactions(&block, expected_chain_id, genesis_account, state_block)
-            {
+            if let Err(error) = Self::validate_transactions(
+                &mut block,
+                expected_chain_id,
+                genesis_account,
+                state_block,
+            ) {
                 return WithEvents::new(Err((block, error.into())));
             }
 
@@ -456,7 +485,7 @@ mod valid {
         /// * If block header is valid, `voting_block` will be released,
         ///   and transactions will be validated with write state
         pub fn validate_keep_voting_block<'state>(
-            block: SignedBlock,
+            mut block: SignedBlock,
             topology: &Topology,
             expected_chain_id: &ChainId,
             genesis_account: &AccountId,
@@ -480,7 +509,7 @@ mod valid {
             };
 
             if let Err(error) = Self::validate_transactions(
-                &block,
+                &mut block,
                 expected_chain_id,
                 genesis_account,
                 &mut state_block,
@@ -577,7 +606,7 @@ mod valid {
         }
 
         fn validate_transactions(
-            block: &SignedBlock,
+            block: &mut SignedBlock,
             expected_chain_id: &ChainId,
             genesis_account: &AccountId,
             state_block: &mut StateBlock<'_>,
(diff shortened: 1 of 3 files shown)


core/src/sumeragi/main_loop.rs
```diff
@@ -383,40 +365,34 @@ impl Sumeragi {
...
-            return None;
-        }
+        BlockCreated {
+            transactions,
+            creation_time_ms,
+        }: BlockCreated,
+    ) -> VotingBlock<'state> {
+        let transactions = transactions
+            .into_iter()
+            .map(|stx| AcceptedTransaction(stx))
```
mversic, line 375:
```suggestion
            .map(|stx| AcceptedTransaction::accept(stx))
```

don't ever skip validation. The inner field of `AcceptedTransaction` should have been private

mversic:
you should follow the style of the code and add a new state to the block state machine. Let's consider this in more detail:
1. remove `categorize_transactions` function
2. make `BlockBuilder::new` -> `BlockBuilder::chain` -> `BlockBuilder::sign` produce `SignedBlock`(not `ValidBlock`)

crates/iroha_core/src/sumeragi/message.rs
```diff
@@ -39,17 +42,12 @@ impl ControlFlowMessage {
 /// `BlockCreated` message structure.
 #[derive(Debug, Clone, Decode, Encode)]
 pub struct BlockCreated {
+    /// The leaders peer id. Helps for cheap pruning of messages from previous rounds.
+    pub leader_signature: SignatureOf<Vec<SignedTransaction>>,
     /// The corresponding block.
-    pub block: SignedBlock,
```
mversic, line 43: There is no need to change this. Leader can keep sending `SignedBlock` as before

mversic: I say this because transaction validation will not change anything except modify the `CommittedTransaction::error` value

mversic: IMO leader can send a block where `CommittedTransaction::error` is `None` and update the value later during block validation

mversic (changes requested):

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -841,57 +839,74 @@ impl Sumeragi {
...
+
+            let new_voting_block = self
+                .validate_block(
+                    state,
+                    &self.topology,
+                    genesis_account,
+                    block_created_msg,
+                    voting_block,
+                )
+                .expect("We just created this block ourselves, it has to be valid.");
```
Erigara, line 895: I think this place might panic if any transaction is invalid, isn't it?

Also validators would reject block if any transaction is invalid.

Erigara (commented):
Is there is any performance boost measurements?

mversic (changes requested):

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -269,7 +267,10 @@ impl Sumeragi {
 
                     // NOTE: By this time genesis block is executed and list of trusted peers is updated
                     self.topology = Topology::new(state_block.world.trusted_peers_ids.clone());
-                    self.commit_block(block, state_block);
+                    self.commit_block(block.clone(), state_block);
+                    self.broadcast_packet(BlockSyncUpdate {
+                        block: block.into(),
+                    });
```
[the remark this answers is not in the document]
SamHSmith: I think I mixed up the genesis functions when I was editiing the code. Thanks for spotting this.

crates/iroha_core/src/sumeragi/main_loop.rs
SamHSmith: My understanding is that the transaction should have been rejected by the queue already. If it were true that a transaction could fail the block. Then we could be perpetually failing to generate a block in all rounds because the bad transaction would be gossiped around to all peers. This would be true already. Therefore it is much safer for our sanity to crash here because we should never be constructing a bad block ourselves.

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -132,10 +132,9 @@ impl Sumeragi {
                 .control_message_receiver
                 .try_recv()
                 .map_err(|recv_error| {
-                    assert!(
-                        recv_error != mpsc::TryRecvError::Disconnected,
-                        "INTERNAL ERROR: Sumeragi control message pump disconnected"
-                    )
+                    if recv_error == mpsc::TryRecvError::Disconnected {
+                        error!("INTERNAL ERROR: Sumeragi control message pump disconnected");
+                    }
```
mversic, line 137: I think panic was appropriate here, we need to shut down iroha

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -171,10 +170,9 @@ impl Sumeragi {
                 .message_receiver
                 .try_recv()
                 .map_err(|recv_error| {
-                    assert!(
-                        recv_error != mpsc::TryRecvError::Disconnected,
-                        "INTERNAL ERROR: Sumeragi message pump disconnected"
-                    )
+                    if recv_error == mpsc::TryRecvError::Disconnected {
+                        error!("INTERNAL ERROR: Sumeragi message pump disconnected");
+                    }
```
mversic, line 175: also here

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -841,57 +838,74 @@ impl Sumeragi {
...
-            if created_in > pipeline_time / 2 {
-                warn!(
-                    role=%self.role(),
-                    peer_id=%self.peer_id,
-                    "Creating block takes too much time. \
-                    This might prevent consensus from operating. \
-                    Consider increasing `commit_time` or decreasing `max_transactions_in_block`"
-                );
+            let pre_signed_block = BlockPayload {
+                header: BlockBuilder::make_header(
```
mversic, line 856: there's a `new` constructor. Please use it

crates/iroha_core/src/block.rs
```diff
@@ -153,7 +153,8 @@ mod pending {
             Self(Pending { transactions })
         }
 
-        fn make_header(
+        /// Create a block header
+        pub fn make_header(
```
mversic, line 157:
```suggestion
       fn make_header(
```

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -841,57 +838,74 @@ impl Sumeragi {
...
 
         if tx_cache_full || (deadline_reached && tx_cache_non_empty) {
             let transactions = self
                 .transaction_cache
                 .iter()
-                .map(|tx| tx.deref().clone())
+                .map(|tx| CommittedTransaction {
+                    value: tx.deref().clone().into(),
+                    error: None,
+                })
```
mversic, line 852: this should happen inside `BlockBuilder::new`. Restore it and move this code there

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -312,13 +313,11 @@ impl Sumeragi {
         // NOTE: By this time genesis block is executed and list of trusted peers is updated
         self.topology = Topology::new(state_block.world.trusted_peers_ids.clone());
 
-        let msg = BlockCreated::from(&genesis);
         let genesis = genesis
             .commit(&self.topology)
             .unpack(|e| self.send_event(e))
             .expect("Genesis invalid");
 
-        self.broadcast_packet(msg);
```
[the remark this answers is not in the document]
mversic: yes, please restore this. It's better that all peers are notified immediately and block sync can take over later.
For genesis all transactions have `CommittedTransaction::error = None`

mversic (changes requested):
I don't see that `CommittedTransaction::error` is being updated anywhere

crates/iroha_core/src/sumeragi/main_loop.rs
```diff
@@ -841,57 +838,74 @@ impl Sumeragi {
             .expect("INTERNAL BUG: transactions in block exceed usize::MAX");
         let block_time = state.world.view().parameters.sumeragi.block_time();
         let tx_cache_full = self.transaction_cache.len() >= max_transactions.get();
-        let deadline_reached = self.round_start_time.elapsed() > block_time;
+        let deadline_reached =
+            self.topology.view_change_index() > 0 || self.round_start_time.elapsed() > block_time;
```
mversic, line 842:
```suggestion
        let deadline_reached =
            self.topology.view_change_index() > 0 || self.round_start_time.elapsed() > block_time;
```

this is abuse of the word `deadline`. view change index is not related to deadline so keep it out

crates/iroha_core/src/sumeragi/main_loop.rs
Erigara: You can check `validate_keep_voting_block` and `validate_transactions` inside it.
If there is mismatch in transaction error (and we set it to `None` despite it's actual status) we got rejection.
It's not possible to fully validate transaction in the queue without executing it.

mversic: We want to categorize transactions and reject a block if it's `CommittedTransaction::error` is incorrect from what the current peer's evaluation produced

mversic: @erigara but how can there be a mismatch in the transaction error when the received block has it set to None for all transactions?

Erigara: Look [here](https://github.com/hyperledger/iroha/blob/002803facb9d2ae9fb414f5428e6356510fe094c/crates/iroha_core/src/block.rs#L619), if transaction has `None` but after execution we discovered that it not actually `None` error is returned

mversic: > Look [here](https://github.com/hyperledger/iroha/blob/002803facb9d2ae9fb414f5428e6356510fe094c/crates/iroha_core/src/block.rs#L619), if transaction has `None` but after execution we discovered that it not actually `None` error is returned   

Shouldn't we remove that check there?

SamHSmith: I have removed that check and replaced it with a mutation of the error value in the block.

mversic (approved):

Erigara (approved):

Issue #2933, opened 2022-11-03
Publish Iroha on Crates.io


appetrosyan:

nxsaken:
Waiting for `storage` to be published: https://github.com/Erigara/storage/issues/2

nxsaken:
Also waiting for `pprof`: https://github.com/tikv/pprof-rs/pull/241

Erigara:
@nxsaken you are no longer blocked by `pprof` they made fix without my PR (just udpate pprof should be enough).

PR #4970, opened 2024-08-13
feat!: reorganize crates

### LINKED ISSUE #4915: Consider renaming the bin crates for consistency
To install binaries via `cargo install`, crate names are used. This is confusing because the `iroha` binary belongs to the `iroha_client_cli` crate, while the `iroha` crate contains the client SDK. The WASM builder suffers from the same awkwardness. I propose we rename the crates once more:

* `iroha` -> `iroha_sdk`
* `iroha_client_cli` -> `iroha` (or merge into a lib+bin crate)
* Merge `iroha_wasm_builder_cli` and `iroha_wasm_builder` into a lib+bin crate similarly to `iroha_swarm`

If we want to eventually publish Iroha (#2933), we should also take into account that `iroha` and some related names like `iroha_derive` have been taken on crates.io already. We should ask the author of the crate to consider giving up the name, or come up with something else.

nxsaken:
## Description

- closes #4915 
  - rename `iroha_client_cli` to `iroha_cli`
  - merge `iroha_wasm_builder` and `iroha_wasm_builder_cli`
  - rename `kagami` to `iroha_kagami` (crate only, not binary)
  - rename `parity_scale_cli` crate to `iroha_codec`
- rename directories to match crate names
- flatten crate structure, move all crates to `/crates`
- move pytests to `/pytests`

### Linked issue

Closes #4915, part of #2933.

### Benefits

- `iroha_cli` a bit more convenient to install
- `iroha_wasm_builder` more convenient to install
- `kagami` no longer conflicts with an existing crate
- directories predictably match crate names
- `iroha_trigger` and `iroha_executor` no longer hidden inside `iroha_smart_contract`

### Checklist


mversic:
Let's name it `iroha_scale_cli`

mversic:
Let's name it `iroha_decode` instead of `iroha_parity_scale_cli`

nxsaken:
`iroha_codec` maybe?

mversic (dismissed):

.github/dependabot.yml
```diff
@@ -21,7 +21,7 @@ updates:
 
   - package-ecosystem: "pip"
     target-branch: "main"
-    directory: "/client_cli/pytests/"
+    directory: "/iroha_cli/pytests/"
```
mversic, line 24: can you remind what was the conclusion, why not merge this binary with it's library?

nxsaken: We don't want to force users to pull in CLI dependencies, and if we separate them with optional features, building or installing the binary would be even less convenient because the feature flag must be specified.

.github/workflows/iroha2-dev-pr-wasm.yaml
```diff
@@ -4,15 +4,15 @@ on:
   pull_request:
     branches: [main]
     paths:
-      - 'data_model/**.rs'
-      - 'data_model/**.yml'
-      - 'data_model/**.json'
-      - 'data_model/**.toml'
+      - 'iroha_data_model/**.rs'
```
mversic, line 7: it's one thing to rename crates but why are you renaming directories?

nxsaken: Consistency, predictability, less cognitive load

mversic: I think what we had was fine. Crates that are in the `iroha` repository are prefixed with `iroha` but the prefix is implied for directories. And this will create so many conflicts for PRs

nxsaken: This is not true for `iroha`, which had the `client` directory. IMO, it's worth changing and going through a brief patch of conflicts to have a standard naming scheme long-term. It's not uncommon for Rust projects to follow this, e.g. [bevy](https://github.com/bevyengine/bevy/tree/main/crates).

0x009922: I agree that have directories named with full crate names is less confusing and more consistent overall. It is currently a mess - some directories are crates, some are not, some have nested crates. Some crates are named in one way as a directory, but have a different actual crate name. As someone who is not always focused on Iroha repository, I find it confusing time and time again to deal with.

0x009922: > And this will create so many conflicts for PRs

Well that's true.

0x009922 (commented):
I would advocate to go even further with flattening: to lift crates such as `iroha_config/base`, `iroha_schema/derive` etc to the top level as well and keep the crates structure completely flat.

Also what about nesting all crates specifically into `/crates` directory or something?

I also don't see a good reason why pytests (for client and torii) are located next to the respective crates. I mean, _yes_, the crates are indeed respective, but related pytests have nothing to do with the crates - they are integrational tests interacting with bare-metal running Iroha peer. Thus, I'd advocate to put them somewhere on the top level as well and avoid their unnecessarily deep nesting.

Cargo.toml
```diff
@@ -193,65 +193,51 @@ clippy.wildcard_dependencies = "deny"
...
+    "crates/kura_inspector",
+    "crates/iroha_codec",
+    "crates/iroha_wasm_test_runner",
+    "crates/iroha_torii",
+    "crates/iroha_torii_const",
+    "crates/iroha_version",
+    "crates/iroha_version_derive",
+    "crates/iroha_wasm_codec",
+    "crates/iroha_wasm_codec_derive",
+    "crates/iroha_wasm_builder",
```
mversic, line 240:
```suggestion
    "crates/*",
```

Cargo.toml
```diff
@@ -193,65 +193,51 @@ clippy.wildcard_dependencies = "deny"
...
+    "crates/iroha_genesis",
+    "crates/iroha_logger",
+    "crates/iroha_p2p",
+    "crates/iroha_futures",
+    "crates/iroha_futures_derive",
+    "crates/iroha_primitives",
+    "crates/iroha_primitives_derive",
+    "crates/iroha_numeric",
+    "crates/iroha_ffi",
+    "crates/iroha_ffi_derive",
```
mversic, line 214: why do we want to make derive crates adjacent?

docs/README.md
```diff
@@ -11,13 +11,13 @@ This is the main Iroha 2 documentation that you will find useful:
 
 Documentation for Iroha 2 tools:
 
-- [Kagami](../tools/kagami/README.md)
-- [Kura Inspector](../tools/kura_inspector/README.md)
-- [Parity Scale Decoder Tool](../tools/parity_scale_cli/README.md)
+- [Kagami](../crates/iroha_kagami/README.md)
+- [Kura Inspector](../crates/kura_inspector/README.md)
+- [Parity Scale Decoder Tool](../crates/iroha_codec/README.md)
```
mversic, line 16: I think we might want to have tools in a tools directory

Cargo.toml
nxsaken: There is no benefit to nesting crates, it only adds mental overhead. Cargo doesn't have nested workspaces

Cargo.toml
```diff
@@ -193,65 +193,51 @@ clippy.wildcard_dependencies = "deny"
...
+    "crates/iroha_schema",
+    "crates/iroha_schema_gen",
```
mversic, line 216: could this be 1 crate?

docs/README.md
nxsaken: What's the benefit? They're still just plain old crates

Cargo.toml
nxsaken: Possibly, but I think we better merge stuff in a separate PR. I think `iroha_macro`/`iroha_derive` could also be simplified if we use `FromVariant` from another crate

docs/README.md
nxsaken: I think it only does so for tools used in CI, while Iroha tools are still part of Iroha

mversic (approved):

0x009922 (dismissed):

mversic (approved):

docs/README.md
mversic: #5043

0x009922 (dismissed):

mversic (approved):

mversic (approved):

0x009922 (approved):

PR #4978, opened 2024-08-14
fix(data-model)!: more consistent query parameters

0x009922:
## Context

There are some minor inconsistencies within query parameters (pagination, fetch size, sorting):

- Impossible to construct these structures when using `iroha_data_model` as a library
- Pagination limit and fetch size was `u32`, while pagination offset is `u64`. It doesn't make sense, since they are all technically limited to the maximum vec size in Iroha, which is `usize`, which is commonly represented as `u64` in the data model.
- Pagination offset is `Option<NonZero<u64>>`, where  `None` value doesn't give any additional meaning and is just treated as zero, which is also a default value for `u64`.

### Solution

- Add missing constructors
- Use `u64`/`NonZero<u64>` everywhere
- Make pagination offset just `u64` instead of `Option<NonZero<u64>>`
- Re-arrange pagination fields - first is `offset`, second is `limit` (affects ordering in the constructor)

---

### Review notes (optional)

- For complex PRs, try to provide some information on how to approach the review more effectively.
- For example, is there a natural order in which the affected files should be reviewed?

### Checklist

- [x] I've read [`CONTRIBUTING.md`](../CONTRIBUTING.md).


data_model/src/query/parameters.rs
```diff
@@ -97,7 +107,7 @@ mod model {
         /// Inner value of a fetch size.
         ///
         /// If not specified then [`DEFAULT_FETCH_SIZE`] is used.
-        pub fetch_size: Option<NonZeroU32>,
+        pub fetch_size: Option<NonZeroU64>,
```
mversic, line 110: I don't see the need to use 64-bit number for fetch size

mversic:
> since they are all technically limited to the maximum vec size in Iroha, which is `usize`, which is commonly represented as `u64` in the data model.

that is the hard upper limit, we can choose to make it tighter. Using a `u64` for fetch size is unnecessary, especially when you consider that `MAX_FETCH_SIZE = 10000`

0x009922:
> that is the hard upper limit, we can choose to make it tighter.

I am fine either way, just want to make it consistent.

> Using a `u64` for fetch size is unnecessary, especially when you consider that `MAX_FETCH_SIZE = 10000`

I see, maybe then make it even smaller, like `u16`?

0x009922:
@mversic as was discussed, let's keep `u64`.

data_model/src/query/parameters.rs
```diff
@@ -56,22 +55,33 @@ mod model {
...
         "limit.map_or(\".inf\".to_owned(), |n| n.to_string())"
     )]
     pub struct Pagination {
-        /// limit of indexing
-        pub limit: Option<NonZeroU32>,
         /// start of indexing
-        pub offset: Option<NonZeroU64>,
+        pub offset: u64,
+        /// limit of indexing
+        pub limit: Option<NonZeroU64>,
```
mversic, line 69: why reorder?

mversic (dismissed):

data_model/src/query/parameters.rs
0x009922: The idea was that the `offset` is primary and `limit` is on top of it.

But

1. Breaks schema a bit
2. SQL uses LIMIT and then OFFSET

So I will revert.

mversic: yes, the feature is known as `LIMIT OFFSET`

0x009922: Reverted

0x009922:
Tests seem to work on my machine.

nxsaken (approved):

mversic (approved):

PR #4980, opened 2024-08-19
fix: Fix bug in test harness for unstable network

SamHSmith:
### Solution

The bug was that the node selection for the client was random. This meant that the node the client was using to observe the tests were being frozen. I fixed it so that the client always is connected to the genesis peer which never gets frozen.

### DIFF
--- a/client/tests/integration/extra_functional/connected_peers.rs
+++ b/client/tests/integration/extra_functional/connected_peers.rs
@@ -57,7 +57,7 @@ fn register_new_peer() -> Result<()> {
         .1
         .submit_blocking(register_peer)?;
     peer_clients.push((&new_peer, Client::test(&new_peer.api_address)));
-    thread::sleep(pipeline_time * 2); // Wait for some time to allow peers to connect
+    thread::sleep(pipeline_time * 2 * 20); // Wait for some time to allow peers to connect
 
     check_status(&peer_clients, 2);
 
--- a/core/test_network/src/lib.rs
+++ b/core/test_network/src/lib.rs
@@ -27,7 +27,7 @@ use iroha_primitives::{
     unique_vec::UniqueVec,
 };
 use irohad::Iroha;
-use rand::{prelude::SliceRandom, seq::IteratorRandom, thread_rng};
+use rand::{prelude::SliceRandom, thread_rng};
 use tempfile::TempDir;
 use test_samples::{ALICE_ID, ALICE_KEYPAIR, PEER_KEYPAIR, SAMPLE_GENESIS_ACCOUNT_KEYPAIR};
 use tokio::{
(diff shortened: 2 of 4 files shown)


SamHSmith:
Currently the github runners are too slow for all the integration tests to pass.

nxsaken (dismissed):

Erigara:
Is increased timeouts necessary for a fix?

SamHSmith:
> Is increased timeouts necessary for a fix?

Yes.

nxsaken:
Note that #4996 switches to a faster runner, and adds retries for `extra_functional` tests

SamHSmith:
> Note that #4996 switches to a faster runner, and adds retries for `extra_functional` tests

The faster runner would be a good addition. But I think retries is a very bad idea.

client/src/config.rs
```diff
@@ -20,7 +20,7 @@ use crate::{
 mod user;
 
 #[allow(missing_docs)]
-pub const DEFAULT_TRANSACTION_TIME_TO_LIVE: Duration = Duration::from_secs(100);
+pub const DEFAULT_TRANSACTION_TIME_TO_LIVE: Duration = Duration::from_secs(1000);
 #[allow(missing_docs)]
 pub const DEFAULT_TRANSACTION_STATUS_TIMEOUT: Duration = Duration::from_secs(15);
```
Erigara, line 25: We set default ttl to 1000 seconds, but client will wait only for 15 seconds for response, should this value be increased as well.

Also do you thinks ~15 minutes is reasonable default value for general case?

SamHSmith: I undid the change and I think it may have even the torii api tests.

client/tests/integration/extra_functional/connected_peers.rs
```diff
@@ -57,7 +57,7 @@ fn register_new_peer() -> Result<()> {
         .1
         .submit_blocking(register_peer)?;
     peer_clients.push((&new_peer, Client::test(&new_peer.api_address)));
-    thread::sleep(pipeline_time * 2); // Wait for some time to allow peers to connect
+    thread::sleep(pipeline_time * 2 * 20); // Wait for some time to allow peers to connect
 
     check_status(&peer_clients, 2);
```
s8sato, line 62: What if poll status periodically instead of waiting for a fixed amount of time?

core/test_network/src/lib.rs
```diff
@@ -378,7 +369,7 @@ impl Network {
 /// # Panics
 /// When unsuccessful after `MAX_RETRIES`.
 pub fn wait_for_genesis_committed(clients: &[Client], offline_peers: u32) {
-    const MAX_RETRIES: u32 = 40;
+    const MAX_RETRIES: u32 = 200;
```
s8sato, line 372: Do we have to wait 1000 seconds for the conclusion?

client/tests/integration/extra_functional/connected_peers.rs
Erigara: Waiting mechanisms is what was done in #5013 so instead on relying on `sleep`s or fixed time polling we listen for events from iroha.
So i think we can postpone this change and merge this PR because it indeed reduce flakiness of `extra_functional` tests.

Erigara (approved):

s8sato (approved):

PR #4989, opened 2024-08-20
ci: Unify CI to build profiling and glibc iroha image

BAStos525:
## Context
Sometimes we need to build iroha2 `glibc` image without `profiling` features for longevity testing. Currently we do it manually and locally only.

### Solution

- Modify `I2::Profiling::Publish` workflow to build either `profiling` image (by new git tag publishing) or vanilla `glibc` image by `workflow_dispatch` event.
- Consider to chose the needed git SHA to build `glibc` image when is required.

### Checklist

- [ ] I've read [`CONTRIBUTING.md`](../CONTRIBUTING.md).

### DIFF
--- a/.github/workflows/iroha2-profiling-image.yml
+++ b/.github/workflows/iroha2-profiling-image.yml
@@ -4,6 +4,24 @@ on:
   push:
     tags:
       - 'v2*'
+  workflow_dispatch:
+    inputs:
+      BUILD_GLIBC_IMAGE:
+        description: "Select \"true\" to build and push the standard \"glibc\" image"
+        type: choice
+        required: true
+        default: 'false'
+        options:
+          - true
+          - false
+      CHECKOUT_REF:
+        description: "The branch, tag or SHA to checkout"
+        required: true
+        default: main
+      PROFILER_NAME:
+        description: "Profiler value as the image tag part"
+        required: true
+        default: glibc
 
 env:
   IROHA2_DOCKERFILE: Dockerfile.glibc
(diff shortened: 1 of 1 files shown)


RamilMus (approved):

.github/workflows/iroha2-profiling-image.yml
mversic, line 1: but the file says it's for profiling?

nxsaken (approved):

.github/workflows/iroha2-profiling-image.yml
nxsaken: Good point, maybe rename this to `iroha-docker.yml`/`I2::Docker`, and instead of `BUILD_GLIBC_IMAGE` provide a choice `IMAGE_TYPE` between `profiling`, `glibc` and `ci` (merge with `I2::CI::Publish`)?

Btw, do we need the `I2` prefix in our workflow names?

nxsaken: Makes sense to separate into its own file because there are additional inputs that only make sense if `BUILD_GLIBC_IMAGE` is true.

BAStos525: "because there are additional inputs that only make sense if BUILD_GLIBC_IMAGE is true": not really. When it's true, it uses the default build-args from Dockerfile. Actually `profiling` case requires the additional build args.
I would try do not create additional workflows when it's possible do not do it.
"Publish is vague IMO": well, for now it's a common part name for all our workflows where image build & publish is there.

BAStos525: My idea is to bind "build & publish iamges" workflows type to Dockerfile that it uses when it's possible

nxsaken: As far as I understand, `PROFILER_NAME` and `CHECKOUT_REF` are only used in the second job that only runs when `BUILD_GLIBC_IMAGE` is true. I don't mind if it stays one workflow, just pointing out ways to simplify

Issue #3716, opened 2023-07-18
[suggestion] Persistent Executor


appetrosyan:
### Feature request

The executor should persist between execution of different transactions. Ideally it should be brought up either after an upgrade or after a quick recovery (from a power outage). It should not have significant persistent storage (caching is fine, storing information that can affect the next verdict is not). 

@mversic suspects that `wasmtime` could have an out-of-memory condition if WASM is not periodically blanked, so the suggestion is to start with a block scope and then extend it further to longer periods of time. I Suggest making it a configuration parameter. 

In detail we need the following: 
- [ ] Persistence configuration parameter, in `configs/peer/config.json` that can be `Every n transactions`, `every n blocks`, `until peer crashes`, defaulting to `every 1 transaction`. 
- [ ] Infrastructure to periodically purge the Validator/executor every `n` of `x` entity. 
- [ ] Tests that produce an out-of-memory condition. Install guard rails to prevent Iroha from crashing in that situation. 
- [ ] Tests that verify that the current executor does not cause an out-of-memory condition under load. 
- [ ] A test that verifies that the current executor does not leak memory (by unloading itself, and comparing resident memory with the last executor). 

### Motivation

The operation of loading and unloading an executor affects the performance of regular transaction processing, so it makes sense to optimise the process and avoid unnecessary loading and blanking of memory if the old memory does a good-enough job. 

### Who can help?

@mversic @appetrosyan

mversic:
related #4914

dima74:
After #5048 merged, executor related things (linker initialization, module instantiation, memory free) started to take noticeable amount of flamegraph. Potentially we could get good tps improvement with persistent executor, but there are problems with its implementation related to lifetimes.

![image](https://github.com/user-attachments/assets/afd2eff2-7a0b-43f5-9cdf-17c65dbb6718)

DCNick3:
Using [`Linker::instantiate_pre`](https://docs.wasmtime.dev/api/wasmtime/struct.Linker.html#method.instantiate_pre) might help the performance without downsides of re-using, since, AFAIU, it has all the imports resolved already and can be instantiated multiple times.

PR #4995, opened 2024-08-21
perf: Don't validate transactions inside WASM

### LINKED ISSUE #4803: Consider removing validation of transactions inside WASM
Do we need to validate transactions inside WASM? Potentially validation on the host should be enough and validation at WASM side could be removed as a small perofrmance optimization. 

> > One option I can think of is to make `SignedTransaction` validation `#[cfg(not(target_family = "wasm"))]`
> 
> I like this suggestion. Can we say that since data model entities are always provided by the host there is no need to validate them on the wasm side? If so, we can can open a ticket to generalize this approach to all entities, although it seems kinda hacky and I'm not sure if this won't bite us in the ass

_Originally posted by @mversic in https://github.com/hyperledger/iroha/pull/4793#discussion_r1664602953_

dima74:
## Context

I am investigating single peer tps performance (#4727), and it turns out that actual execution of wasm code takes most of the time per transaction (https://github.com/hyperledger/iroha/issues/3716#issuecomment-2299147066). 

### Solution

This PR removes validation of transactions at WASM side - there is no need to revalidate transaction which comes from Iroha Rust code. This gives approximately **50% tps increase** (single peer). Also size of default executor is reduced from 492KB to 391KB.

Comparison of tps:

compare

Fixes #4803
Related: #4727
Related: #3716

### Checklist

- [ ] I've read [`CONTRIBUTING.md`](../CONTRIBUTING.md).


data_model/src/transaction.rs
```diff
@@ -335,7 +335,9 @@ mod candidate {
 
     impl SignedTransactionCandidate {
         fn validate(self) -> Result<SignedTransactionV1, &'static str> {
+            #[cfg(not(target_family = "wasm"))]
```
mversic, line 338: this is a bit hackish but I think it's ok if it's well documented

mversic (commented):
I think you should apply this to all `validate_***` methods in this crate

data_model/src/transaction.rs
dima74: Added comment

mversic:
> I think you should apply this to all `validate_***` methods in this crate

you still didn't apply it to `Name::validate_str`

mversic:
> > I think you should apply this to all `validate_***` methods in this crate
> 
> you still didn't apply it to `Name::validate_str`

also `Parameter`, `Trigger` and  `event_set`

data_model/src/transaction.rs
mversic: this should best be documented on the module level

mversic: even on the crate level. Somewhere where it will be visible to users that it applies to all structures deserialization

dima74:
* Removed validation for `Name::validate_str`
* For `Parameter`, the only validation is `NonZeroUsize::try_from(/* NonZeroU64 */)`, I think it should be kept as is
* For `Trigger` there is method `ActionCandidate::validate` which does really small match check, and also is used in `ActionCandidate::new`. Do you think it is worth to change it?
*  For `event_set` didn't find validation method, could you clarify?

mversic:
> * Removed validation for `Name::validate_str`
> 
>     * For `Parameter`, the only validation is `NonZeroUsize::try_from(/* NonZeroU64 */)`, I think it should be kept as is
> 
>     * For `Trigger` there is method `ActionCandidate::validate` which does really small match check, and also is used in `ActionCandidate::new`. Do you think it is worth to change it?
> 
>     * For `event_set` didn't find validation method, could you clarify?

it's fine, I only looked around for `deserialize` functions

Can you update the numbers in the PR description?

dima74:
> Can you update the numbers in the PR description?

Update executor size. Tps numbers are approximately [the same](https://github.com/user-attachments/assets/fcb91700-785e-4d97-824c-067ef6f03556)

PR #5000, opened 2024-08-22
refactor: sign only block's header, not the whole payload

mversic:
## Context

it's not necessary to sign the whole block payload since block header holds the merkle tree root hash

### Reason

this will enable us to work on #4914


SamHSmith (approved):
Logic is sound

s8sato (approved):

